Issue #3165, opened 2019-06-20
Improve event handler typings for all components


sradevski:
Currently most (if not all) components' event handlers have the following typing: `((...args: any[]) => any);`. This should be fixed and modified to the appropriate handler typing for each event.

TODO: List all components concerned and fix progressively.
- [ ] ...

jamiedavenport:
Personally, not having types for the values from `Form`s has been the biggest problem for me. I'm happy to contribute a fix for this.

My solution would be to make `FormProps` generic and to change the `onSubmit` definition to be:
```js
onSubmit?: ((event: {value: V}) => void);
```

britt6612:
## Listing all of the components in which we need to improve event handler typing's 

**AccordionProps** 
- onActive [ ]

**AnchorProps**
- onClick [ ]

**Button** 
- onClick [ ]

**Calendar** 
- onSelect [ ]
- onReference [ ]

**ChartProps**
- onClick [ ]
- onHover [ ]

**CheckBoxProps**
- onChange [ ]

**ClockProps**
- onChange [ ]

**DataTableProps**
- onMore [ ]
- onSearch [ ]

**DistributionProps**
- children [ ]

**DropProps**
- onClickOutside [ ]
- onEsc [ ]

**DropButtonProps**
- onClose [ ]
- onOpen [ ]

**FormProps**
- onChange [ ]
- onSubmit [ ]

**InfiniteScrollProps**
- onMore [ ]
- children [ ]
- renderMarket [ ]

**KeyBoardProps** 
- onBackspace [ ]
- onComma [ ]
- onDown? [ ]
- onEnter [ ]
- onEsc [ ]
- onKeyDown [ ]
- onLeft [ ]
- onRight [ ]
- onShift [ ]
- onSpace [ ]
- onTab [ ]
- onUp [ ]

**MarginType** 
- onClickOutside [ ]
- onEsc [ ]

**MaskedInputProps** 
- onChange [ ]
- onBlur [ ]

**RadioButtonProps**
- onChange [ ]

**RadioButtonGroupProps**
- onChange

**RangeInputProps**
- onChange [ ]

**RangeSelectorProps** 
- onChange [ ]

**SelectProps** 
- children [ ]
- icon [ ]
- onChnage [ ]
- onClose [ ] 
- onMore [ ] 
- onOpen [ ]
- onSearch [ ] 

**TabsProps** 
- onActive [ ]

**TextAreaProps** 
- onChange [ ]

[...]

PR #3335, opened 2019-09-10
Changed onBlur type to match with react focusevent

britt6612:
#### What does this PR do?
Update onBlur type to be more specific to the event 
maskedInput/index.d.ts
manually 
write onBlur function in TS app 
We are using ``` JSX.IntrinsicElements['input'] ``` this includes the 
HTML input elements so we just need to add the react event handling
of focus event
#3165 
no
no
backwards compatible

### DIFF
--- a/src/js/components/MaskedInput/index.d.ts
+++ b/src/js/components/MaskedInput/index.d.ts
@@ -11,7 +11,7 @@ export interface MaskedInputProps {
   }>;
   name?: string;
   onChange?: ((...args: any[]) => any);
-  onBlur?: ((...args: any[]) => any);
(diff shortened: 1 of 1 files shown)


ShimiSun:
>We are using JSX.IntrinsicElements['input'] this includes the
HTML form elements so we just need to add the react event handling
of focus event

did you mean '...the HTML input elements'?

Can you attach the source you used to solve the problem?

britt6612:
Yes we are already have the HTML elements from the input so we just need to also support react events 

https://stackoverflow.com/questions/46950418/using-onblur-to-set-focus-in-a-react-component

PR #3338, opened 2019-09-11
Adjust Keyboard TS Definitions

taysea:
#### What does this PR do?
This changes the TS definitions for all Keyboard props from ((...args): any => any) to ((event: React.KeyboardEvent<HTMLElement>) => void).

src/js/components/Keyboard/index.d.ts

Testing has been done in a local TS project. Keyboard was wrapped around content in a local TS file, and the type definitions were checked for the different keyboard interactions. First, I created a separate handler and defined the event type to match the TS definition. I expected there to be no errors, and there weren't. Then, I tried breaking it by providing the function with different types such as a string to be sure it would flag this error. It did.

In a project with TS added.

When TS definitions are "any", it defeats its purpose. In this case, the event is a KeyboardEvent and this will happen on an HTML Element. Therefore, React.KeyboardEvent is the event type. Void is the return type because the function doesn't return anything or get assigned to a variable.

See React's documentation about Keyboard Events: https://reactjs.org/docs/events.html#keyboard-events

Issue #3165 

No

No

Yes, backwards compatible.

### DIFF
--- a/src/js/components/Keyboard/index.d.ts
+++ b/src/js/components/Keyboard/index.d.ts
@@ -2,18 +2,18 @@ import * as React from "react";
 
 export interface KeyboardProps {
   target?: "component" | "document";
-  onBackspace?: ((...args: any[]) => any);
-  onComma?: ((...args: any[]) => any);
-  onDown?: ((...args: any[]) => any);
-  onEnter?: ((...args: any[]) => any);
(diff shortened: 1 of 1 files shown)


ShimiSun:
Fantastic explanations and descriptions! Thank you for your contribution

PR #3343, opened 2019-09-13
Removed padding with a plain menu

nomadus:
What does this PR do?
This PR fixes the issue with padding when a menu component is set to plain

src/js/components/components/Menu/Menu.js

I did a quick test (Google Chrome), with plain set to true and with no plain property set. The extra padding has disappeared in the former case.

Set plain={true} for a menu component and see the decrease in the height and width of the box around the component.

I needed to have a thin header on my page, but the menu component does not allow it as it is surrounded with a Box with fixed 'small' padding.

What are the relevant issues?
No

Screenshots (if appropriate)
No. The plain property implies that no extra padding and extra margins should stay when this property is set to true.

No.

Is this change backward compatible or is it a breaking change?
Yes, backward compatible.

### DIFF
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -123,7 +123,7 @@ class Menu extends Component {
         direction="row"
         justify={justifyContent}
         align="center"
-        pad="small"
+        pad={plain ? "none": "small"}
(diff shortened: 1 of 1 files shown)


nomadus:
The change not in the code, not sure what happened

PR #3345, opened 2019-09-13
Added Chromatic for Visual Testings

ShimiSun:
Adding Chromatic visual testings to Grommet. https://www.chromaticqa.com/ 

According to the current design, the tool will run the chroma build on every commit and will create snapshots for each component's stories, hence leveraging the power of existing stories to aslo be used as tests (snapshots). The build results could be accessed directly from GitHub details on the CI workflow 
Screen Shot 2019-10-01 at 3 47 03 PM
(See this PR for example)

The build includes *45 components, 212 stories, 212 snapshots*.

The app results can be found on https://www.chromaticqa.com/builds?appId=5d7a63beda813e00202faf9d, the chroma build-summary shows a comparison of the current and previous build results, and it alerts the admins on any changes that occurred between them.

Terminology  - when running the chromatic build you may see the following:

1. *Errors* - when a chroma test fails with an error, it means the following:
>Your story couldn’t be captured because it exceeds our 25,000,000px limit. Its dimensions are 1,200x21,900px. Possible ways to resolve: Separate pages into components or Minimize the number of very large elements in a story

2. *Changes* - the concept of 'Changes' is similar to the snapshots changes we see on jest, changes can be auto-approved or approved manually by an admin via the chroma app.

Grommet testing results on Chroma will always have 'Changes' due to stories with realtime data (date, clock) that create different DOM values on every run. 
And in order to not break the build on every run, I added `--exit-zero-on-changes`, this will help to exit gracefully when changes are detected on the build results. The build will still fail in case of an error, and any Changes will still require manual acceptance.

After a PR is filed, chroma changes can be reviewed on its app, similarly to how we review jest snapshots but with a nicer UI :) https://www.chromaticqa.com/build?appId=5d7a63beda813e00202faf9d&number=32

Another flag we could use, in case the review of the tests becomes over-bearing is `--auto-accept-changes` to auto accept the changed files without review, and the test will be flagged only on errors. IMO, it will defeat the purpose and we should use `--exit-zero-on-changes` for now, so we will still have to review the changes but the build itself won't fail.

To run `yarn chroma` locally, add 
`"chroma": "chromatic test --app-code=<appcode> --exit-zero-on-changes",`
To your `package.json` scripts.

Currently, the `CHROMATIC_APP_CODE` is part of grommet NPM environment variables.

>NOTE: The app code cannot be used to read story data, it can only be used to create new builds.


oorestisime:
I guess we can add the npm run chromatic in the circleci and instead of having the code in the repo we can have it there as an env variable. would that make sense?

ShimiSun:
Terminology  - when running the chromatic build you may see the following:

1. *Errors* - when a chroma test fails with an error, it means the following:
>Your story couldn’t be captured because it exceeds our 25,000,000px limit. Its dimensions are 1,200x21,900px. Possible ways to resolve: Separate pages into components or Minimize the number of very large elements in a story

2. *Changes* - the concept of 'Changes' is similar to the snapshots changes we see on jest, changes can be auto-approved or approved manually by an admin via the chroma app.

Grommet testing results on Chroma will always have 'Changes' due to stories with realtime data (date, clock) that create different DOM values on every run. 
And in order to not break the build on every run, I added `--exit-zero-on-changes`, this will help to exit gracefully when changes are detected on the build results. The build will still fail in case of an error, and any Changes will still require manual acceptance.

After a PR is filed, chroma changes can be reviewed on its app, similarly to how we review jest snapshots but with a nicer UI :) https://www.chromaticqa.com/build?appId=5d7a63beda813e00202faf9d&number=2.

Another flag we could use, in case the review of the tests becomes over-bearing is `--auto-accept-changes` to auto accept the changed files without review, and the test will be flagged only on errors. IMO, it will defeat the purpose and we should use `--exit-zero-on-changes` for now, so we will still have to review the changes but the test won't fail.

To run `yarn chroma` locally, add 
`"chroma": "chromatic test --app-code=<appcode> --exit-zero-on-changes",`
To your `package.json` scripts.

Issue #3292, opened 2019-08-22
ts: export actual prop types (and TableCell issue)


atanasster:
The full component props need to be exported from the .ts files.
example for the export:
https://github.com/grommet/grommet/blob/e23667c953bc871b18ef9689f73609e47d283fdf/src/js/components/Box/index.d.ts#L39
the JSX.IntrinsicElements['div'] part is not exported, while any component that would like to use BoxProps will need to eventually duplicate those div props
so I would suggest 
```
export interface BoxOwnProps {
  a11yTitle?: A11yTitleType;
  alignSelf?: AlignSelfType;
...
}

export type BoxProps = BoxOwnProps & JSX.IntrinsicElements['div']>;
```

example for usage: TableCell is using the BoxProps
https://github.com/grommet/grommet/blob/e23667c953bc871b18ef9689f73609e47d283fdf/src/js/components/TableCell/TableCell.js#L57

it should allow for the BoxProps to be specified (currently typescript apps can not pass Box props to TableCell due to type checking): 

```
export interface TableCellOwnProps {
  plain?: boolean;
  scope?: "col" | "row";
...
}

export type TableCellProps = TableCellOwnProps & BoxProps & JSX.IntrinsicElements['td'] ;
```

PR #3346, opened 2019-09-13
Added Box types to tableCell to accept

britt6612:
#### What does this PR do?
Since Table Cell accepts Box props it should also accept the TS types from Box
This PR allows Table Cell to also accept Box TS properties.
TableCell/index.d.ts
Manually tested by branching out grommet with ts project and testing it there

also can see error in https://codesandbox.io/s/grommet-ts-lp7bp

``` 
              <TableCell key={c.property} align="center" margin="small">
              </TableCell> 

```
Pass any of Box Props to TableCell and you will see an TS error without fix 

issue #3292 
no
no 
Backwards Compatible


src/js/components/TableCell/index.d.ts
```diff
@@ -7,6 +8,6 @@ export interface TableCellProps {
   verticalAlign?: "top" | "middle" | "bottom";
 }
 
-declare const TableCell: React.FC<TableCellProps & JSX.IntrinsicElements['td']>;
+declare const TableCell: TableCellProps & Box & JSX.IntrinsicElements['td'];
```
ShimiSun: what does it mean removing the wrapper of `React.FC<>`?

britt6612: React.fc is a short for a function component we will actually need this and have to export Box without the component class. Table cell would be a functional component where box would need to be a component class in which it already is. The class component which is box would be used just to return a component.

src/js/components/Box/index.d.ts
```diff
@@ -39,5 +39,6 @@ export interface BoxProps {
 }
 
 declare const Box: React.ComponentClass<BoxProps & JSX.IntrinsicElements['div']>;
+declare const BoxType: BoxProps & JSX.IntrinsicElements['div'];
```
taysea: Could you simplify this by saying the following?
```
declare const BoxType: BoxProps & JSX.IntrinsicElements['div'];
declare const Box: React.ComponentClass<BoxType>;
```

britt6612: @taysea that will work and make it cleaner still need to investigate if we need decalre const or a type for this

ShimiSun: hi ladies, any progress on this one?

britt6612: Yes Ill push up my findings and changes later today!

britt6612: @ShimiSun I went a head and pushed up my latest changes. Since In Box index.d.ts we have declare const Box which is declaring a new type for Box which is then exported out of Grommet. We also need to use the same box props as well as the props from HTML div in TabelCell since TabelCell is using box props ` <Box {...tableContextTheme} {...boxProps}> ` for this we can easily export the type BoxType in which we can have this equal the BoxProps as well as the HTML div props and pass this to the tableCell index.d.ts file

We are not creating a new type we just want to give another name to the type in which is being passed to TabelCell.

Hopefully this clears up any questions

britt6612: @ShimiSun this is ready for review not sure if you saw my comment above 👍

Issue #3326, opened 2019-09-05
Table caption always hidden?


robinandeer:
### Expected Behavior

When I provide a caption to the `Table` component, it shows up in the DOM.

### Actual Behavior

The Table caption is always hidden.
See here: https://github.com/grommet/grommet/blob/master/src/js/components/Table/StyledTable.js#L57

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-kw9xt?fontsize=14

### Steps to Reproduce

Import `Table` and supply the "caption" property.

### Your Environment

- Grommet version: 2.7.7 (or current master)
- Browser Name and version: Chrome 77
- Operating System and version (desktop or mobile): Mac desktop macOS 10.14.6

ShimiSun:
YEa, I agree it doesn't look right. looks like you already found the root cause of a problem, feel like contributing a PR?

robinandeer:
Sure I can try to do that 🙂

ShimiSun:
Hi buddy, @robinandeer let us know how it goes and if you need any help!

robinandeer:
Yeah thanks, didn't have to get to it until now but I've just opened a new PR :-)

PR #3348, opened 2019-09-14
Display table caption

robinandeer:
This PR addresses #3326.

#### What does this PR do?

Displays the table caption when it's passed as a prop to the table component.

Check out the updates in Storybook.

I've only updated the snapshots to match the new styles/classnames.

Review if the look/feel for the table caption is inline with the grommet styles.

Nope.

- #3326 

Screenshot 2019-09-14 at 11 15 04

Nope, previously caption was just never displayed even if the prop was passed to the component.

Yeah maybe as a fix.

Yes. It enables functionality that was supposed to be there in the first place so I would say it's backwards compatible.


ShimiSun:
I'm trying to troubleshoot a concern regarding accessibility for screen-readers when a table is using a caption. 
From a read on https://webaim.org/techniques/tables/data#caption, I found this 

<i>Table Captions
Data tables very often have brief descriptive text before or after the table that indicates the content of that table. This text should be associated with its respective table using the <caption> element. The <caption> element must be the first thing after the opening `<table>` tag.
```
<table>
<caption>Shelly's Daughters</caption>
```
While it is not necessary for each table to have a caption, a caption is generally very helpful. If present, it should be associated to the table using the <caption> element.</i>

It seems like it is encouraged to use `caption` to help screen readers. 

If the caption is `display: none;` like it is now, then the Text To Speech (TTS) will skip the caption whether it is assigned or not. 
For example, turn on TTS (I use chromeVox) on this story https://storybook.grommet.io/?path=/story/table--custom, in this case, the caption is assigned, but since it is `display: none;` the TTS will skip it and will only read the existing title text without reading the caption itself, hence, the caption doesn't show its value.

To me, the flow that makes sense for grommet is as follows: 
- a caption is assigned --> caption should be visible --> it should be read with TTS, 
- a caption isn't assigned --> not visible --> it won't be part of the screen reading 
Hence, I think we should allow displaying the caption in case it is not empty. 
Please let me know if you have more thoughts on this @ericsoderberghp, or would like to share more areas that need to be explored.

ericsoderberghp:
Thanks for the detailed analysis. My only remaining hesitation is if the caller should be given more flexibility with how the caption is rendered. Perhaps we can allow `caption` to also be a react element that it will allow the caller to control rendering? I think we can wait on this last suggestion until the need arises though.

PR #3349, opened 2019-09-16
Extract keyboard type to single variable.

taysea:
#### What does this PR do?
Extracts Keyboard type into singe variable. Previously, many lines were copy and pasted. This change places the definition in a single place which will makes any future adjustments for streamline.

src/js/utils/index.d.ts

Testing has been done on local TS project.

On a project with TS enabled.

Issue #3165 
No.

No.

Yes, backwards compatible.

### DIFF
--- a/src/js/components/Layer/index.d.ts
+++ b/src/js/components/Layer/index.d.ts
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { AnimateType, MarginType } from "../../utils";
+import { AnimateType, MarginType, KeyboardType } from "../../utils";
 
 export interface LayerProps {
   animate?: AnimateType;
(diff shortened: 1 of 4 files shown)


src/js/components/Layer/index.d.ts
```diff
@@ -8,7 +8,7 @@ export interface LayerProps {
   margin?: MarginType;
   modal?: boolean;
   onClickOutside?: ((...args: any[]) => any);
-  onEsc?: ((...args: any[]) => any);
+  onEsc?: KeyboardType;
   plain?: boolean;
   position?: "bottom" | "bottom-left" | "bottom-right" | "center" | "hidden" | "left" | "right" | "top" | "top-left" | "top-right";
```
ShimiSun: Regardless of this PR, can we do some cleanups for the position prop?

PR #3358, opened 2019-09-18
Added TS handlers for onChange

britt6612:
#### What does this PR do?
This changes the TS definitions for all of the onChange from ((...args): any => any)
These changes are in the index.d.ts of the components that had onChange 
Testing has been done manually using a local TS project. I created a separate handler
for each of the components  something like `onChange={event => console.log(event)}`
and received no errors then changed it to another type such as string and it recognized the error.

In a project with TS 

```
          <FormField
            label="Size"
            name="size"
            component={Select}
            onChange={event => console.log(event)}
          />
```
Any component that has onChange 

Before the definition was any which does nothing for someone who is using TS this not have any type checking involved. onChange is a event type. The event type is the react.changeEvent with the HTML Input element since ChangeEvent more suitable for typing form events.

Issue #3165 
no
no
backwards compatible


ShimiSun:
Does this fix also takes care of the issue that eric brought up?

britt6612:
@ShimiSun  There is a suggestion to allow something like this for `Function` to be assignable to `(...args: any[]) => any` in the TypeScript Repo. 

`The original intention of Function is to not be callable. in other words, Function to function types should be like unknown to other types, but not callable. we have since relaxed this restriction giving Function a callable behavior in the compiler through special casing. We have talked about making this a --noImplicitAny error since, it is really unsafe to call Functions.`

https://github.com/Microsoft/TypeScript/issues/20007

Usually it is suggested that users stay away from `Function` its not a type-safe away to express function signatures

Im curious to hear from @ericsoderberghp on his take on using Function.

ericsoderberghp:
I think I ran into this when just stubbing out the `onChange` in a TypeScript project I was working on. I had seen that syntax in some reference somewhere and was trying to just make progress without having to learn more TypeScript details in the moment. If those who know TypeScript more feel that isn't the right approach, I'm happy to defer to them.

ShimiSun:
Thank you for sharing your inputs and contribution!

Issue #3360, opened 2019-09-19
Bug in `step` parameter of <RangeSelector />


saidelmark:
### Expected Behavior
`<RangeSelector />` with a given `step` parameter should select only select `n`*`step` values

### Actual Behavior
If `step` equals `2`, then everything works as expected.
Otherwise I can select `5` when `step` is `3` and available range is `[0, 100]`. But if I change the selection using arrow keys, then every keystroke increases/decreases selected value by the exact value of `step`.

### URL, screen shot, or Codepen exhibiting the issue

I also created a codepen so you don't have to: https://codepen.io/saidelman/pen/ExYdwbM?editors=0010
It's based on [grommet's storybook](https://storybook.grommet.io/?path=/story/rangeselector--thin).

This is an example with `step` 3, while you can see that I can move the selector to values like 7 or 8:
![RangeSelector](https://user-images.githubusercontent.com/9284083/65163447-9e006200-da43-11e9-92b3-a44421aa0224.gif)

### Steps to Reproduce

1. Create a `<RangeSelector />`
1. Set `step` parameter to 2
1. Check that it works fine
1. Set `step` to a bigger value
1. Move the sliders with your mouse, the selected value will have incorrect values
1. Move the sliders using arrow keys to see that this time the difference in selected value changes by `step` each time.

I initially noticed the issue when I was working with huge numbers: I wanted to have `[0, 1e9]` range and `1e6` step. Expected to be able to select only values with 6 zeros ending. And then, when started to investigate the issue, found similar problems with `[0, 100]` range and `5` step.

The most interesting part is that there is no this bug in a similar `Grommet` component, `<RangeInput />` (I added this component to my codepen as well to illustrate that).

### Your Environment
- Grommet version: tested with `2.7.6` and `2.7.8`
- Browser Name and version: Google Chrome Version 77.0.3865.75 (Official Build) (64-bit)
- Operating System and version (desktop or mobile): Debian 10
- TypeScript version: 3.6.2
- React version: 16.9.0

taysea:
Thanks for raising this issue @saidelmark! I submitted a PR for this. It looked like the calculation for the value of the range selector based on the mouse position wasn't working as expected.

ShimiSun:
Thank you for reporting buddy!

PR #3363, opened 2019-09-19
Fix range selector step issue

taysea:
#### What does this PR do?
Fixes issue with step on range selector with mouse interaction.

src/js/components/RangeSelector/RangeSelector.js

The issue noted in Issue #3360 was recreated in a local project. Then after the fix, the changes were tested in a storybook example and the example provided in order to confirm successful changes.

Create a RangeSelector with a step value and ensure that as the mouse moves either end of the range selector that the value is adhering to the defined step.

The RangeSelector was creating a glitchy effect when a step was provided. For example, if the step was 3, the RangeSelector output value would jump from 0 -> 2 -> 4 -> 3 -> 5 -> 7 -> 6 and so on. This does not achieve the intended goal of only allowing the RangeSelector to reach values in increments of 3.

Issue #3360 

No.

No.


ShimiSun:
I'd like to get some information on why the new formula is working, and where the previous one was failing.

taysea:
The existing calculation would create the following (where x is the value of the mouse location and y is the value of the range selector):
```
x | y
0 | 0
1 | 2
2 | 4
3 | 3
4 | 5
...
```
Because of this, you can see how the output does not create the expected step output and instead creates a glitchy output. What we expect to see is:
```
x | y
0 | 0
1 | 3
2 | 3
3 | 3
4 | 6
5 | 6
6 | 6
...
```
The updated calculation achieves that.

ShimiSun:
@taysea Thank you for working on this PR! 
I'm ok with merging it, but I'm concerned about two things:
1. How come our tests didn't catch it? (all snapshots haven't changed)
2. I need a more asserted confirmation then just a review to confirm this fixed behavior.

Would you be willing to add tests on src/js/components/RangeSelector/__tests__/RangeSelector-test.js with a couple of tricky uses cases that this PR is solving?

taysea:
Yes, @ShimiSun I can do that. It does seem strange that the existing was passing the tests.

taysea:
I looked into the existing test, and this is how the step is being tested:
```
  test('step', () => {
      <Grommet>
        <RangeSelector step={10} values={[20, 30]} />
      </Grommet>,
    );
  });
```
It doesn't seem to be testing how the position of the range selector shifts when the mouse drags an edge. That must be why the current version is passing the test. I'll work on creating a new test and confirm that the existing version fails but my change passes.

taysea:
Verified with Chromatic: https://www.chromaticqa.com/build?appId=5d7a63beda813e00202faf9d&number=39

ShimiSun:
Great! thank you @taysea !! 🎉

PR #3368, opened 2019-09-23
Adjust Heading/Responsive Context Grommet docs

taysea:
#### What does this PR do?
This PR adds a simple elaboration about how Heading level is dependent on the responsive breakpoints. It also clarifies that there are default breakpoints that are determined by the screen resolution.

src/js/components/Heading.js

This change only effects docs, but the text has been compared with other documentation on the site to match the phrasing/tone.

There was confusion in the community about how Heading dynamically/automatically adjusts size when screen size is changed. It seemed like a small bit of documentation would help clarify these confusions.

N/A

Yes, but this PR is updating them.

No.

Yes, backwards compatible.


src/js/components/Heading/doc.js
```diff
@@ -72,7 +72,7 @@ export const themeDoc = {
   },
   'heading.level': {
     description:
-      'The level that impacts line height, max width, font size, weight and family of the Heading.',
+      'The level that impacts line height, max width, font size, weight and family of the Heading. Heading level is automatically adjusted at different screen sizes. These screen sizes are derived from the global.breakpoints entry in the theme object.',
```
ShimiSun:
```suggestion
      'The level that impacts line-height, max-width, font size, weight and family of the Heading. Heading level is automatically adjusted at different screen sizes. These screen sizes are derived from its component breakpoints .',
```
I adjusted the text to be more descriptive about the prop rather than getting into the role of the theme here. In my mind, the user could scroll and look for theme breakpoints info in case she is looking for additional information.

taysea: I think this might be a little deceiving because it makes the phrasing sound like this component controls the breakpoints for screen sizes in general. What if we just simplify it to:
`These screen sizes are derived from global.breakpoints.` The reason being, global.breakpoints exists on the page so if the user did control F or just scrolled they could find elaborated information about it.

ShimiSun: I think we should avoid the `global` reference on the description hence avoid this declared dependency.
I'm ok with `These screen sizes are derived from breakpoints.`
Or `These screen sizes are derived from the components breakpoints defenitions` or anything along those lines.

taysea: Okay, that makes sense about avoiding declared dependencies.

ShimiSun:
Let's resolve conflicts on this one and its good to go 🚀

src/js/contexts/ResponsiveContext/doc.js
```diff
@@ -28,8 +28,9 @@ export const doc = ResponsiveContext => {
 
 export const themeDoc = {
   'global.breakpoints': {
-    description:
-      'The possible breakpoints that could affect border, direction, gap, margin, pad, and round. The default values help to optimize content for mobile, tablet, and computer.',
+    description: `The possible breakpoints that could affect border,
```
ShimiSun: why does your line break after 70 characters? The standard for prettier (grommet) is 80 characters.

taysea: Prettier wasn't breaking the text for some reason and I was getting an eslint error. Let me go back and make sure prettier is working correctly and make a new commit.

ShimiSun: Any updates on this one?

taysea: Still trying to fix my format on save.. can't figure out why it's working perfectly fine for everything else except this. Will work through that today.

ShimiSun:
Thank you for making sure the small things don't slip! 👍

Issue #2973, opened 2019-03-23
Ability to unselect / clear selected value with <Select>


markjackson02:
Currently there is no built-in way for a user to unselect an option after selecting (resetting to original state). Proposed enhancement would be to add a "clearValue" prop that adds a grey "x" on the right side of the Select input to clear the current value.

My current workaround is to put the first option as a blank string, but this doesn't look great from a UI perspective.

oorestisime:
hm, maybe the solution here is to approach the selection closer to the select multiple handling. In multiple when you re-select the option it will unselect it. Unselecting all the selected values clears all the values.

oorestisime:
I ve been thinking about this a bit. not sure this should be supported on the core library. there are so many ways of doing this (handle re-clicking, adding a checkbox, adding a close icon, adding an empty...) it will end up not being generic enough.
We could have an example with a few ways of doing this though.

ShimiSun:
@JetLim what do you think of a template example for this one? 
@MikeWalrath can you add it to out templates list.

JetLim:
@oorestisime 
We'll discuss this with a checklist of possible selector solutions. 

@ShimiSun 
@MikeWalrath

markjackson02:
@JetLim @oorestisime any progress here?

oorestisime:
Thanks for reminding us! We haven't got the chance to talk about this. will in the next few days.

markjackson02:
Hi @ShimiSun I noticed the "in progress" tag. Is the team moving forward with a specific solution to this? If so I am interested to know what it is!

markjackson02:
@ShimiSun @JetLim any updates on this?

ShimiSun:
Yes, we will be discussing the designs today, hopefully we'll have something to share soon. @MikeWalrath @L0ZZI FYI

JetLim:
Yes, we have something that is in the works. @ericsoderberghp @ShimiSun @markjackson02

@MikeWalrath @L0ZZI

ShimiSun:
Hi buddy, thanks for your feedback and patience with this one.

There are a couple of approaches here I would like to share:
1. Unselect could be done today with clearing the selected options, whether it is a multiple or single Select option. @britt6612 contributed a story to demonstrate this implementation (Thanks @britt6612). https://github.com/grommet/grommet/pull/3371
https://deploy-preview-3371--sad-tereshkova-173d07.netlify.com/?path=/story/components--all
[...]

PR #3371, opened 2019-09-24
Adding UnSelect storybook ex

britt6612:
#### What does this PR do?
This PR adds an example of how to clear selected options in a select component
using hooks
Select/stories/UnSelect.js
storybook
There was a request that grommet add a unselect button 
or clear button that would clear the selected options. In the mean time 
grommet should provide an example of how this can be done.
#2973 
no
no
backwards compatible


britt6612:
Currently there is a trash can icon next to the select. @JetLim is looking at another icon or other options we can use.

src/js/components/Select/stories/UnSelect.js
```diff
@@ -0,0 +1,55 @@
...
+        pad="medium"
+        direction="row-responsive"
+        align="center"
+        justify="center"
+      >
+        <Select
+          options={options}
+          onChange={e => onChange(e)}
+          value={value}
+          placeholder="select"
```
ShimiSun:
```suggestion
          placeholder="Select multiple options"
```

src/js/components/Select/stories/UnSelect.js
```diff
@@ -0,0 +1,55 @@
+import React, { useState } from 'react';
+import { storiesOf } from '@storybook/react';
+
+import { Box, Grommet, Select, Button } from 'grommet';
+import { grommet } from 'grommet/themes';
+import { Trash } from 'grommet-icons';
+
+const option = [
```
ShimiSun: I think we should use a plural, or something like `optionList`, thoughts?

src/js/components/Select/stories/UnSelect.js
```diff
@@ -0,0 +1,55 @@
...
+        justify="center"
+      >
+        <Select
+          options={options}
+          onChange={e => onChange(e)}
+          value={value}
+          placeholder="select"
+          multiple
+        />
+        <Button onClick={onClick} icon={<Trash color="control" size="medium" />} />
```
ShimiSun:
```suggestion
        <Button onClick={onClickClearOptions} icon={<Trash color="control" size="medium" />} />
```
I'm suggesting to rename the function, so when its defenition and implemntation appears on the top of the page, it will be implied immediatly what is its purpose.

leonardobork:
There's just only one issue. This is not working :).

https://storybook.grommet.io/?path=/story/select--unselect

ShimiSun:
We are taking a look, thanks @leonardobork

PR #3197, opened 2019-07-13
fix(formfield): missing select props

roger-king:
#### What does this PR do?

Potentially fix types for formfield 

#### How should this be manually tested?

Create a formfield using `Select` as component and setting `options` prop.

#### Is this change backwards compatible or is it a breaking change?

Backwards compatible

PR #3372, opened 2019-09-24
TypeScript FormField- Omit size and onSelect

### LINKED ISSUE #3370: [TS] Incompatible FormField "size" and "onSelect" prop types
Passing `size` and `onSelect` props to `FormField` down to `TextInput` (via `component` prop) results in typescript complaining about inconsistent types. I narrowed this down to  `HTMLInputElement['size']` and `HTMLInputElement['onselect']` respectively, passed down (I suppose) to `<input>` element by styled components as valid HTML attributes. 

A possible solution would be to filter and override those two with the appropriate type spec in `FormField` (similar to #3256) and eventually ensure they are not passed down to the `<input>` element as valid DOM parameters by styled.

### Expected Behavior

No compile error.

### Actual Behavior

Compile error.
- `size`:
```
Type 'string' is not assignable to type 'number'.
```
- `onSelect`: 
```
Type '(x: { target: HTMLElement; suggestion: any; }) => void' is not assignable to type '(event: SyntheticEvent<HTMLInputElement, Event>) => void'.
  Types of parameters 'x' and 'event' are incompatible.
    Property 'suggestion' is missing in type 'SyntheticEvent<HTMLInputElement, Event>' but required in type '{ target: HTMLElement; suggestion: any; }'
```

### Code sample
https://codesandbox.io/s/grommet-typescript-v2-formfield-issue-xqh05

### Your Environment
- Grommet version: 2.7.8

britt6612:
#### What does this PR do?
This PR omits the size and onselect for the formfield typescript 
file.
FormField/index.d.ts
branched out in a project that uses TS
Following code can be used

```

const App = () => {
  const onSelect: TextInputProps["onSelect"] = () => {};

  return (
    <Grommet theme={grommet}>
      <Box pad="small">
        <Form>
          <FormField
            size="large"
            onSelect={onSelect}
            label="Test Input"
            name="test"
            value="test"
            component={TextInput}
          />
        </Form>
      </Box>
    </Grommet>
  );
};

render(<App />, document.getElementById("root"));

```
See above
This was done for TextInput previous that onselect and size were both 
omitted from the HTML 'input' since grommet states its own prop types for both size and onSelect. If someone was using the same TextInput Prop types for Form field they were getting an error because these were not omited from FormField. Also see issue filed provides a great explanation.
closes #3370 
no
no
backwards compatible

### DIFF
--- a/src/js/components/FormField/index.d.ts
+++ b/src/js/components/FormField/index.d.ts
@@ -16,6 +16,6 @@ export interface FormFieldProps {
   validate?: {regexp?: object,message?: string} | ((...args: any[]) => any);
 }
 
-declare const FormField: React.ComponentClass<FormFieldProps & Omit<JSX.IntrinsicElements['input'], 'placeholder'>>;
+declare const FormField: React.ComponentClass<FormFieldProps & Omit<JSX.IntrinsicElements['input'], 'placeholder', 'size', 'onSelect' >>;
 
 export { FormField };


ShimiSun:
@murbanowicz I'd love to hear your thoughts on this one, the issue is coming up on different use cases when using FormField. 
FormField should enable typings of child component props, e.g. #3197 is trying to patch missing props of Select within the FormField, but I'm thinking we need a more generic approach here as those missing typings may come from all child components of FormField (CheckBox, RadioButton or any other input type). Please share your inputs.

ShimiSun:
@dallas I'd love your inputs on this thread/issue.

dallas:
It seems like for a limited set of possible child components you could perhaps create a list of overloads for the `FormField` component. I’m not sure how well TypeScript overloads work with classes, though, as opposed to functions. Another option might be generics with inferred types, but I’m struggling to figure out how to make it work since the component being passed in to the `component` prop or rendered as the first `children` item isn’t the same as the props that that component accepts. I think we can [use generics directly in JSX/TSX](https://mariusschulz.com/blog/passing-generics-to-jsx-elements-in-typescript), so the solution might be to just tell `FormField` what additional props you expect it to be able to take:

```TypeScript
<FormField<TextAreaProps>
    label="Label"
    htmlFor="text-area"
    {...props}
    component={TextArea}
    placeholder="placeholder"
/>
```

and

```TypeScript
<FormField<SelectProps> label="Label" htmlFor="select" {...props}>
    <Select
        id="select"
        placeholder="placeholder"
        options={allOptions}
        value={value}
    />
</FormField>
```

Which isn’t ideal – the ideal is that we could simply infer the new set of types by noticing whatever `component` is set. I don’t know how to jump from `TextArea` the component to `TextAreaProps` the interface, though. I’m also not sure how it would be done in the case of passing in a child component rather than using the `component` prop.

dallas:
Ahh, we can extract/infer the props interface for a given component:

```TypeScript
type ExtractProps<C> =  C extends React.FunctionComponent<infer P> ? P :
                        C extends React.ComponentClass<infer P> ? P :
                        never;
```

Where `C` is the `typeof` some component and we either get `P` returned (as the interface of props the component accepts) or the special TypeScript value `never`, indicating that the component does not accept any props (or is not a component). Example:

```TypeScript
type TextAreaProps = ExtractProps<typeof TextArea>;
```

maslennikov:
As a quick and dirty workaround I created a `global.d.ts` with the following contents redeclaring the original `FormField` spec:
```typescript
declare module "grommet" {
  declare const FormField: React.ComponentClass<
    FormFieldProps
    & Omit<JSX.IntrinsicElements['input'], 'placeholder'>
    & SelectProps
    // whatever else props you need
    >;
}
```

ShimiSun:
@maslennikov thanks for sharing buddy! 
The potential problem I see with this idea is the override/collision of props, for example, if Select is using `value` with a certain type and another component is using different types for the same prop of `value` then this solution will break.

maslennikov:
@ShimiSun oh yeah that will break for sure. It solves the problem only for simple cases in a controlled environment of a single project and doesn't suit for a library as a whole.

PR #3373, opened 2019-09-24
Refactor Tabs to use hooks

taysea:
#### What does this PR do?
This PR refactors Tab.js to use hooks.

src/js/components/Tab/Tab.js

Testing on all storybook examples of Tabs has been tested on Chrome, Safari, and Edge to ensure that keyboard/mouse engagement works correctly.

On a local project or with storybook.

N/A

N/A

No.

No.


src/js/components/Tab/Tab.js
```diff
@@ -13,118 +13,107 @@ import { normalizeColor } from '../../utils';
...
+  const {
+    active,
+    forwardRef,
+    plain,
+    title,
+    onMouseOver,
+    onMouseOut,
+    theme,
+    ...rest
+  } = props;
```
oorestisime: can we move the destructure in the function declaration?

src/js/components/Tab/Tab.js
```diff
@@ -13,118 +13,107 @@ import { normalizeColor } from '../../utils';
...
+
+  delete rest.onActivate;
+
+  let normalizedTitle = title;
+  const tabStyles = {};
+
+  const [state, setState] = useState({
+    active: props && active ? active : false,
```
oorestisime: not sure what props is meaning here.

oorestisime: hm, i am more inclined to have two useState one for active and one for over. not sure what you think here. https://reactjs.org/docs/hooks-state.html#tip-using-multiple-state-variables
if i am reading this correctly the setState would completely replace the object hence when you change only one of the two states (active or over) in the way you did it would replace the other value. if we want to handle all in one state then we need to repeat the existing values in all the setStates. (I am personally leaning over having decoupled useState it would make the code clearer.)

src/js/components/Tab/Tab.js
```diff
@@ -13,118 +13,107 @@ import { normalizeColor } from '../../utils';
...
+    setState({ over: undefined });
     if (onMouseOut) {
       onMouseOut(event);
     }
   };
 
-  onClickTab = event => {
-    const { onActivate } = this.props;
+  const onClickTab = event => {
+    const { onActivate } = props;
```
oorestisime: don't  have time to test this right now but i am not sure why we do this. why don't we destructure onActivate always and just call the onActivate if it exists? do we forward this later to something?

src/js/components/Tab/Tab.js
taysea: This makes sense. Let me create a new version with the state variables decoupled.

src/js/components/Tab/Tab.js
```diff
@@ -13,118 +13,107 @@ import { normalizeColor } from '../../utils';
...
-    return null;
-  }
-
-  state = {};
-
-  onMouseOver = event => {
-    const { onMouseOver } = this.props;
-    this.setState({ over: true });
+const Tab = props => {
+  const {
```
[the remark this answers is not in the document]
taysea: You're right. I think we should destructure. Let me create a new version and test that it still works.

src/js/components/Tab/Tab.js
```diff
@@ -13,118 +13,107 @@ import { normalizeColor } from '../../utils';
...
+    forwardRef,
+    plain,
+    title,
+    onMouseOver,
+    onMouseOut,
+    theme,
+    ...rest
+  } = props;
+
+  delete rest.onActivate;
```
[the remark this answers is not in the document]
taysea: That was existing from the Tabs.js as it was prior to any of my changes. I would have to look more into the reasoning for having it in general, but it isn't specific to using a function component.

src/js/components/Tab/Tab.js
taysea: I believe you're right. Let me try doing that and confirm it still works.

ShimiSun:
linking to #3394

ShimiSun:
Verified with chroma on https://www.chromaticqa.com/build?appId=5d7a63beda813e00202faf9d&number=36.

Issue #2975, opened 2019-03-25
Add `plain` prop to FormField


MaxMillington:
### Expected Behavior

I would like there to be a `plain` prop in the `FormField` component which will remove the default underline from the FormField

### Actual Behavior

Currently, there is no way to remove the underline from a FormField without removing it from every FormField

sahithikol:
Is some one working on this
I can work on this if noone is working on it

oorestisime:
Doesnt seem so! go for it :)

sahithikol:
how can i assign this to my self?

oorestisime:
It seems you cant. and i cant find you in the list either! but go for it don't worry!

ShimiSun:
@sahithikol are you still on it? should I assign you?

sahithikol:
@ShimiSun  i have not started working on this yet
I can be on it in a few days, i am on vacation currently
if there is any one else who can look into please feel free to assign them

ShimiSun:
No worries, first and foremost enjoy your vacation :)
Feel free to communicate if you start working on it, so I'll assign you on it, I think it can wait for you to be back :)

felipebrunetti94:
is this issue still open ?

bdesouza:
Yes. Is there any way for me to upvote this?

ShimiSun:
@bdesouza can you share what you are experiencing exactly?

fuyucn:
any update for this one? the default underline is very annoying

jashanshah33:
can i work on this,  if no one is working on it ?

jcfilben:
Hi @jashanshah33 feel free to work on this

theankith:
hi,
Is this issue closed?

MikeKingdom:
I would consider it open. It hasn't had interest for a couple years

rickyentzer:
> Currently, there is no way to remove the underline from a FormField without removing it from every FormField

I think in this instance, the use of a CSS class would be a good solution. 

```
.FormField-no-underline {
    border-bottom: none
}
```

MikeKingdom:
You can use `contentProps` rather than resorting to CSS:
```jsx
<FormField contentProps={{ border: { color: "transparent" } }} ... >
```

PR #3096, opened 2019-05-09
fix: #2975:  plain prop in the FormField component which will remove the defa…

sahithikol:
…ult underline from the FormField

#### Is this change backwards compatible or is it a breaking change?
yes it is backward compatible

PR #3374, opened 2019-09-25
Added border prop to FormField - DON'T MERGE YET

taysea:
#### What does this PR do?
This PR allows users to add a `border` prop to FormField. By default, the border styling defined by the theme will be used. However, is border is specified, it will override the theme styling. Additionally, if `border={false}`, there will be no border. The border prop takes all of the same options that Box border accepts, in addition to a position which can be 'inner' or 'outer'.

The docs were updated, a storybook example was made, and two tests were added (one to test no border, and one to test a custom border).

This change has been tested on a local version of storybook. The prop border={false} was applied to a form field to ensure the bottom border was removed. Additionally, a custom border containing an array of objects was tested to ensure all recent additions to Box border props also worked for FormField.

In any local project, create a FormField and manipulate the values for border. Some suggestions would be:
```
<FormField border={false} />
```
OR
```
<FormField
    border={[
        {
            color: 'brand',
            side: 'left',
            position: 'outer',
            style: 'dashed',
            size: 'medium',
        },
        { color: 'neutral-3', side: 'bottom', size: 'medium' },
    ]}
/>
```

Many users were seeking to be able to remove this border without having to use Theme.Extend every time. It made sense that in addition to being able to remove the default theme border, they should be able to control border in the same way that is allowed on Box.

Issue #2975 

Yes, the updates are included in this PR.

No.


src/js/components/FormField/FormField.js
```diff
@@ -106,10 +106,13 @@ class FormFieldContent extends Component {
       onBlur,
       onFocus,
       margin,
+      underline,
```
ShimiSun: I'm not sure about the naming of the prop, doesn't seem consistent with grommet props.
A suggestion, let's try to get a `border` prop and if it is "none" than the underline will be removed, or any other given border on that prop will be applied correspondingly. thoughts?

ericsoderberghp: Agreed. We should do something with a `border` property.

taysea: Okay, this feedback makes sense. I will work on doing a `border` property.

ShimiSun:
Following our discussion, the checklist for adding a new prop:
[ ] add the new prop and its description on doc.js
[ ] create its updated snapshots `yarn test-update`
[ ] add typescript support on `index.d.ts`
[ ] create a new test for the new prop
Nice to have:
[ ] provide a storybook example that demonstrates the new prop's behavior. It doesn't have to be a new story, it could be embedded as part of an existing story.

src/js/components/FormField/FormField.js
oorestisime: this looks similar to this https://github.com/grommet/grommet/pull/3096 maybe we can finish that PR?

taysea: @oorestisime You're right. I think this PR would solve the issues noted in that PR.

taysea:
The following has been done:
[x] add the new prop and its description on doc.js
[x] create its updated snapshots yarn test-update
[x] add typescript support on index.d.ts
[x] create a new test for the new prop

Nice to have:
[x] provide a storybook example that demonstrates the new prop's behavior. It doesn't have to be a new story, it could be embedded as part of an existing story.

src/js/components/FormField/stories/CustomBorder.js
```diff
@@ -0,0 +1,70 @@
...
+const FormFieldCustomBorder = props => {
+  const [checkBox1, setCheckBox1] = useState(false);
+  const [checkBox2, setCheckBox2] = useState(false);
+  const [checkBox3, setCheckBox3] = useState(false);
+
+  return (
+    <Grommet theme={grommet}>
+      <Box justify="center" pad="large" direction="row" gap="medium">
+        <FormField label="Default border" htmlFor="check-box" {...props}>
+          <Box pad={{ horizontal: 'small', vertical: 'xsmall' }}>
```
ShimiSun: Since `pad={{ horizontal: 'small', vertical: 'xsmall' }}` is repeated three times, we can extract it to be a variable.

src/js/components/FormField/doc.js
```diff
@@ -17,6 +52,20 @@ export const doc = FormField => {
...
+        'left',
+        'bottom',
+        'right',
+        'horizontal',
+        'vertical',
+        'all',
+      ]),
+      BORDER_SHAPE,
+      PropTypes.arrayOf(BORDER_SHAPE),
+    ]).description('Include a border.'),
```
ShimiSun: Should we also mention it defaults to `bottom` here?

src/js/components/FormField/FormField.js
```diff
@@ -156,27 +192,41 @@ class FormFieldContent extends Component {
...
+        : border.position === 'outer' &&
+          (border.side === 'all' ||
+            border.side === 'horizontal' ||
+            !border.side);
       if (abut) {
         // marginBottom is set to overlap adjacent fields
         abutMargin = { bottom: '-1px' };
         if (margin) {
           abutMargin = margin;
+        } else if (Array.isArray(border)) {
```
ShimiSun: why does the border rendering is coupled to the `margin` assignment?

src/js/components/FormField/index.d.ts
```diff
@@ -1,7 +1,10 @@
 import * as React from "react";
-import { Omit, PlaceHolderType, MarginType } from "../../utils";
+import { BoxSideType, BoxSizeType, BoxStyleType, ColorType, Omit, PlaceHolderType, MarginType, SizeType, SideType, StyleType } from "../../utils";
+
+type PositionType = "inner" | "outer";
 
 export interface FormFieldProps {
+  border?: boolean | BoxSideType | {color?: ColorType, position?: PositionType, side?: BoxSideType, size?: BoxSizeType, style?: BoxStyleType} | ({color?: ColorType, position?: PositionType, side?: BoxSideType, size?: BoxSizeType, style?: BoxStyleType})[];
```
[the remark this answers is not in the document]
taysea: I believe that FormField is using border from box under the hood because the formfield container (`FormFieldBox`) is defined as a styled Box. See here: https://github.com/grommet/grommet/blob/a6a9ab1e575bb8ff262d2e0b364586e81d892dd8/src/js/components/FormField/FormField.js#L32

src/js/components/FormField/stories/CustomBorder.js
```diff
@@ -0,0 +1,70 @@
...
+              onChange={() => setCheckBox3(!checkBox3)}
+            />
+          </Box>
+        </FormField>
+      </Box>
+    </Grommet>
+  );
+};
+
+storiesOf('FormField', module).add('Custom Border', () => (
```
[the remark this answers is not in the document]
taysea: Agreed. I'll change this.

src/js/components/FormField/FormField.js
taysea: This is existing from the previous code, but essentially this section is to help align FormFields when they are vertically stacked and use the size of the border to make them flush with each other. However, if the user specifies margin as a prop to FormField, it will use that value instead of the border size. Nothing new here about coupling with margin except that I made it able to handle when the border provided is an array.

src/js/components/FormField/__tests__/__snapshots__/FormField-test.js.snap
```diff
@@ -432,6 +432,102 @@ exports[`renders abut with forced margin 1`] = `
...
+}
+
+<div
+  className="c0"
+>
+  <div
+    className="c1"
+  >
+    <div
+      className="c2"
```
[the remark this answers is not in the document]
taysea: The reason for this is because there are two different boxes that can receive border. One outer box is the main `FormFieldBox` and is marked by `c1`. However, there is also the inner box which contains the `contents`; this is marked as `c2`. This class structure matches the structure of all other FormField tests that existed prior to any changes. Let me know if this makes sense. In the case of this test, the left dashed border has position "outer" which causes it to effect the styling of the outer box while the other border affects the inner box.

src/js/components/FormField/index.d.ts
```diff
@@ -1,7 +1,10 @@
 import * as React from "react";
-import { Omit, PlaceHolderType, MarginType } from "../../utils";
+import { BoxSideType, BoxSizeType, BoxStyleType, ColorType, Omit, PlaceHolderType, MarginType, SizeType, SideType, StyleType } from "../../utils";
+
+type PositionType = "inner" | "outer";
```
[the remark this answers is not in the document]
taysea: I think it should just remain as I have it. `const type` gives an eslint error. This is the same as what you had done here: https://github.com/grommet/grommet/blob/8c64ea2f413cc6a3b7d5562d32cde9626793a207/src/js/components/Box/index.d.ts#L14

taysea:
Validated with chromatic: https://www.chromaticqa.com/build?appId=5d7a63beda813e00202faf9d&number=37

src/js/components/FormField/index.d.ts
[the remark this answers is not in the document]
ShimiSun: great! thanks for verifying.

src/js/components/FormField/index.d.ts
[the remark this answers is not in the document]
ShimiSun: thanks for enlighting me :)

ShimiSun:
Thanks for building with chroma! 
Congrats, you have a new screenshot added 
image
The snapshots review were all accepted 🥇

ShimiSun:
I want another pair of eyes on this PR before I merge it.

src/js/components/FormField/FormField.js
```diff
@@ -156,27 +192,41 @@ class FormFieldContent extends Component {
             this.childContainerRef = ref;
           }}
           border={
-            border.position === 'inner'
-              ? {
+            Array.isArray(border)
+              ? this.createBorder(border).innerBorder
```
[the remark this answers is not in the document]
taysea: You're right, I'll work on the logic to minimize how many times this gets calculated.

ShimiSun:
Hi @taysea, any updates on this PR?

taysea:
@ShimiSun I'm going to close this for now because there was inconsistency with how the focus would be handled with custom border. @ericsoderberghp, @L0ZZI, and I have been discussing FormField this past week.

Issue #3356, opened 2019-09-18
Form onChange returns two objects instead of only value


cscognamiglio:
### Expected Behavior
```
<Form onSubmit={({value}) => this.save(value)}
            onChange={(value) => console.log(value)}
            value={value}
>
```

onChange should log the content of value, that is the changed form values object.

### Actual Behavior

onChange seems to be called twice, first time with an event object and second time with the actual value object.

### URL, screen shot, or Codepen exhibiting the issue
![Screenshot 2019-09-18 at 12 31 44](https://user-images.githubusercontent.com/11227537/65141289-9a071c80-da10-11e9-9ec1-e2c2d95421db.png)

Reproduced here:

https://codesandbox.io/s/grommet-v2-template-slom4

### Steps to Reproduce

1. Create a Form
2. Add an onChange prop
3. log value

### Your Environment

- Grommet version: 2.7.8
- Browser Name and version: Firefox Quantum 69
- Operating System and version (desktop or mobile): MacOS Mojave 10.14.6

maslennikov:
I confirm with Chrome (ver 79 - latest for the time being) on Mac. Looks like some keypress event is leaking from TextInput (both as regular form field children, and nested deeper inside other form components (like search input inside a select).

For example, within a searchable select, normal click upon an option from the list will trigger only a valid Form.onChange event. But after any key press inside of the search field it will trigger two successive Form.onChange invocations (first - with a keypress event passed as a param, second - with a valid Form.onChange param).

Also, local onChange() callbacks passed to FormField components won't catch this buggy event, it happens only with Form.onChange.

johnny1K:
Same issue which @maslennikov has with grommet ^2.7.11
With the latest grommet (2.11.1) as of now, the synthetic event is gone, but the form object arrives twice.

ShimiSun:
Hi @maslennikov can you share a [codesandbox](https://codesandbox.io/s/m7mml8l0zj) that reproduces the problem? It will help us to quickly troubleshoot.

britt6612:
There has been many changes to `form` in the past few months 

I created an updated codesandbox to show that onChange is now just called once. 

https://codesandbox.io/s/onchange-form-1efo4?file=/src/App.js

If there are no objections I believe this issue can be closed. The Grommet 2.11.0 release with the massive improvements to `form` and `formfield` this issue was no longer happening

cscognamiglio:
Yes I can confirm that the problem is fixed, thanks! I will close the issue.

[...]

PR #3375, opened 2019-09-25
HTML onchange

britt6612:
#### What does this PR do?
This PR potentially fixes a bug found with onchange in form  
form.js
storybook and grommet project linked to grommet core
```
class App extends Component {
  render() {
    let value = {
      title_fr: "Title fr",
      title_en: "Title en"
    };
    return (
      <Grommet theme={grommet}>
        <Box pad="small">
          <Form
            onSubmit={({ value }) => this.save(value)}
            onChange={value => console.log(value)}
            value={value}
          >
            <FormField label="Title (FR)" name="title_fr" />
            <FormField label="Title (EN)" name="title_en" />
          </Form>
        </Box>
      </Grommet>
    );
  }
}
```
Currently onChange was added into onReset and upDate functions. 
Since we were using ` <form {...rest} onReset={this.onReset} onSubmit={this.onSubmit}>`
it was using passing the HTML onchange. Since this was the case in the original issue you can see that the object was being console.log twice which the second object the correct event. 
There is no actual onChange function which may be something we discuss.
#3356 
N/A
No
No
backwards compatible


ShimiSun:
Let's test this one with onChange on Form, I want us to make sure it works properly and not creates a backward compatibility issue for users who had been using onChange for Form.

britt6612:
@ShimiSun sounds good I will test this also with that as well!

britt6612:
@ShimiSun after some more digging into this problem turns out that
the warning about the `SyntheticEvent` object is because React reuses the `SyntheticEvent` by polling the synthetic events all together that is why in the `syntheticevent` it is returning null after an event is called back. 

Essentially, `SyntheticEvent` cannot be used asynchronously, because the event will no longer exist after the event callback has been invoked.

Im investaging another way in which we can leverage the event.persist() to maybe fix this issue as well as looking at other things.

ShimiSun:
OK, sounds good, feel free to reopen the PR or file a new one when relevant.

PR #3378, opened 2019-09-26
Add an option to show a checkmark when Checkbox toggled

sradevski:
#### What does this PR do?
Adds an opt-in checkmark when the Checkbox (in toggle mode) is checked.

Tested on storybook and ran the existing tests

Check the added storybook page

Depending on the knob color, it is not always clear whether a toggle
is on or off. This just adds additional context.

When toggle icon color is passed (otherwise it is transparent):
Screen Shot 2019-09-26 at 1 17 40 PM

Yes

Yes

It is backwards compatible

### DIFF
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -123,7 +123,7 @@ export interface ThemeType {
         borderSize?: BreakpointBorderSize;
         edgeSize?: BreakpointEdgeSize;
         size?: BreakpointSize;
-      };
+      } | undefined;
     };
     deviceBreakpoints?: {
       phone?: string;


src/js/components/CheckBox/StyledCheckBox.js
```diff
@@ -20,7 +20,12 @@ const StyledCheckBoxIcon = styled.svg`
   box-sizing: border-box;
   stroke-width: ${props => props.theme.checkBox.check.thickness};
   stroke: ${props =>
-    normalizeColor(props.theme.checkBox.color || 'control', props.theme)};
+    props.toggle
+      ? normalizeColor(
+          props.theme.checkBox.toggle.icon.color || 'transparent',
```
ShimiSun: I think there might be a better way to assign the icon and the icon color.
Currently, when the icon isn't used or colored, it is spamming the dom with an unneeded SVG, and a redundant nested `span` element.
I think the icon shouldn't be applied on the DOM if the user wasn't asking for it.

ShimiSun:
We should try to have a version for on/off and a theme option for on/off while the custom theme will have the `<path>`

sradevski:
I'll close this for now as I don't have time to focus on it, (and it seems I overwrote the changes with another PR I wanted to do), I'll just reopen when I have a bit more time

ShimiSun:
Thanks for the follow up @sradevski

PR #3380, opened 2019-09-26
Refractor Button.js to use hooks

britt6612:
What does this PR do?

This PR refactors Button.js to use hooks.

Button/Button.js
storybook examples of buttons 
storybook
On a local project or with storybook.

N/A
What are the relevant issues?

N/A
Screenshots (if appropriate)
No
No
backwards compatible


src/js/components/Button/Button.js
```diff
@@ -1,8 +1,9 @@
-import React, { cloneElement, Children, Component } from 'react';
+import React, { cloneElement, Children, useState } from 'react';
 import { compose } from 'recompose';
 
 import { withTheme } from 'styled-components';
 
+import PropTypes from 'react-desc/lib/PropTypes';
```
ShimiSun: why do we suddenly need the PropTypes? we didn't need it before the hooks.

britt6612: @ShimiSun we don't necessarily need it I just put it out there since we were declaring the default PropTypes at the bottom of this file so added the prop types for those values.

oorestisime: proptypes are handled by the doc file if i am not mistaken. i think we shouldn't add them here as well.

src/js/components/Button/Button.js
```diff
@@ -26,134 +27,129 @@ const isDarkBackground = props => {
...
+    onMouseOut,
+    onMouseOver,
+    plain,
+    primary,
+    reverse,
+    theme,
+    type,
+    as,
+    ...rest
+  } = props;
```
oorestisime: can we move those in the props directly ? would be much easier to read.

src/js/components/Button/Button.js
```diff
@@ -26,134 +27,129 @@ const isDarkBackground = props => {
...
+  gap: 'small',
+};
 
 Object.setPrototypeOf(Button.defaultProps, defaultProps);
 
 let ButtonDoc;
 if (process.env.NODE_ENV !== 'production') {
-  // eslint-disable-next-line global-require
-  ButtonDoc = require('./doc').doc(Button);
+ ButtonDoc = require('./doc').doc(Button); // eslint-disable-line global-require
```
oorestisime: why did this change?

src/js/components/Button/Button.js
britt6612: Sounds good I got rid of the proptypes!

src/js/components/Button/Button.js
britt6612: I will work on this and test to make sure everything is still working correctly! In Progress

src/js/components/Button/Button.js
britt6612: This must of been something on my end! Fixed

ShimiSun:
linking to #3394

PR #3390, opened 2019-10-01
fix spell word

yogaap24:


templates/grommet-v2-template/public/index.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en - id">
```
ShimiSun: why is this change required?

PR #3391, opened 2019-10-01
Refactor Keyboard.js to use hooks

shabtaisharon:
#### What does this PR do?
Refactor `Keyboard.js` to use hooks
`src/js/components/Keyboard/Keyboard.js`
Running `yarn test`
Running `yarn test`
N/A
N/A
No
No
backwards compatible


ShimiSun:
linking to #3394

thidasapankaja:
Is this up for grabs ? Can I start working on this ?

src/js/components/Keyboard/Keyboard.js
```diff
@@ -14,47 +14,37 @@ const KEYS = {
...
-    const { target } = this.props;
-    if (target === 'document') {
-      document.removeEventListener('keydown', this.onKeyDown);
-    }
-  }
-
-  onKeyDown = (event, ...rest) => {
-    /* eslint-disable-next-line react/prop-types */
-    const { onKeyDown } = this.props;
+function Keyboard(props) {
```
oorestisime: this is looking good, wonder if we can make a tiny bit of improvement and turn this into const Keyboard = ({destructure, any, prop}) => 

it will match the rest of the components being ported right now!

shabtaisharon: I could change it to `const Keyboard = props => {` to be more consistent.
Since the function is doing a dynamic lookup based on `callbackName` in runtime, then using static destruction will not work. Please let me know if you know of any other way to do that and I'll be happy to make the change.

oorestisime: i see i didn't even notice the callbackName, was actually talking for `props.onKeyDown` or `props.target` etc. for callbackName you would need to destructure all the values and this, i agree is not nice, but for props we are directly using it might be a nice tweak. what do you think?

ShimiSun:
Nice work! thank you for your contribution!

Issue #3394, opened 2019-10-02
Refactoring components and stories to use hooks


ShimiSun:
Tracking the effort for the hooks migration starts here... see the below PRs as references.

aravind1078:
@ShimiSun  is there anything left to pick ?

thidasapankaja:
@ShimiSun  Can I start working on this #3391 ?

ShimiSun:
@aravind1078 Go to any storybook example that has Class usage and refactor to class. look for example src/js/components/CheckBox/stories/Indeterminate.js

ShimiSun:
Hi @thidasapankaja, #3391  is a PR in progress, so maybe try to pick another Grommet class or story?

thidasapankaja:
@ShimiSun ok sure. I'll work on refactoring Class usage in Select story

vitorfreitas:
#3424

vitorfreitas:
#3424

shakti97:
I m trying to do but as i committed the code it starts some test and then i just wrote the git status and it showed me a huge amount of README files modified. Don't know why it changes the readme.md file. Can anybody know the reason behind this

sofiageo:
> I m trying to do but as i committed the code it starts some test and then i just wrote the git status and it showed me a huge amount of README files modified. Don't know why it changes the readme.md file. Can anybody know the reason behind this

Are you using Windows and git autocrlf set to auto? The project is committed with LF line endings and I don't see any gitattributes committed to the repository to enforce the line endings. Your git might be changing the line endings to CRLF.

shakti97:
yes i m on windows

sofiageo:
In my opinion you should not commit anything with CRLF endings if the project is using LF endings, because it will create more confusion later. Another problem is that git is converting these files on "git clone" so even adding a gitattributes file don't be enough to fix it. Try to see if you can change the line endings from your IDE - if you are using IDEA products I think it's possible. - then only commit your changes.

p.s that's just my opinion and expertise and maybe grommet has no issues with the different line endings.

sdowding:
is there any other hooks stuff i can work on? it seems like its all finished!

ShimiSun:
@sdowding plenty...
Accordion.js
Chart.js
All of Chart stories  under src/js/components/Chart/stories
Collapsible.js
Resizer.js
Form.js
And more...

Look for imports of  `{ Component } from 'react'`, that will be a good place to start.

[...]

PR #3397, opened 2019-10-02
Hookified Meter/stories

scfoxcode:
Signed-off-by: Stephen J Surtees <stephen.publicmail@gmail.com>

#### What does this PR do?
Converts the 4 class components in Meter/stories to use hooks

Code was built and the Meter unit tests were ran, all passed

Create the individual components, particularly Circle.js as it is using a timeout


ShimiSun:
linking #3394

src/js/components/Meter/stories/Bar.js
```diff
@@ -4,12 +4,14 @@ import { storiesOf } from '@storybook/react';
...
 import { grommet } from 'grommet/themes';
 
-const BarMeter = () => (
-  <Grommet theme={grommet}>
-    <Box align="center" pad="large">
-      <Meter type="bar" background="light-2" values={[{ value: 30 }]} />
-    </Box>
-  </Grommet>
-);
+function BarMeter() {
```
nolleto: There is a reason to change from `const` to `function`?
Also, you could create a variable to store `[{ value: 30 }]` and avoid unnecessary re-render:
```js
const meterValues = [{ value: 30 }];

const BarMeter = () => (
  <Grommet theme={grommet}>
      <Meter type="bar" background="light-2" values={meterValues} />
    </Box>
  </Grommet>
);
```

src/js/components/Meter/stories/Circle.js
```diff
@@ -1,37 +1,32 @@
-import React from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import { storiesOf } from '@storybook/react';
 
 import { Grommet, Box, Meter } from 'grommet';
 import { grommet } from 'grommet/themes';
 
-class CircleMeter extends React.Component {
-  state = { value: 20 };
+function CircleMeter() {
```
nolleto: I believe the default is to use `const` insead `function`.

src/js/components/Meter/stories/Labelled.js
```diff
@@ -4,26 +4,30 @@ import { storiesOf } from '@storybook/react';
...
+      <Box align="center" pad="large">
+        <Stack anchor="center">
+          <Meter
+            type="circle"
+            background="light-2"
+            values={[{ value: meterValue }]}
+            size="xsmall"
+            thickness="small"
+          />
+          <Box direction="row" align="center" pad={{ bottom: 'xsmall' }}>
```
nolleto: You could store `{ bottom: 'xsmall' }` to avoid unnecessary re-render.

src/js/components/Meter/stories/Labelled.js
```diff
@@ -4,26 +4,30 @@ import { storiesOf } from '@storybook/react';
...
+  const meterValue = 30;
+
+  return (
+    <Grommet theme={grommet}>
+      <Box align="center" pad="large">
+        <Stack anchor="center">
+          <Meter
+            type="circle"
+            background="light-2"
+            values={[{ value: meterValue }]}
```
nolleto: You could store in a variable `[{ value: meterValue }]` to avoid unnecessary re-render.

src/js/components/Meter/stories/Labelled.js
```diff
@@ -4,26 +4,30 @@ import { storiesOf } from '@storybook/react';
...
-            30
-          </Text>
-          <Text size="small">%</Text>
-        </Box>
-      </Stack>
-    </Box>
-  </Grommet>
-);
+function LabelledMeter() {
+  const meterValue = 30;
```
nolleto: I believe you can declare this variable outside this component.

src/js/components/Meter/stories/Bar.js
ShimiSun: why was it changed to a function rather than a const?

src/js/components/Meter/stories/MultipleValues.js
```diff
@@ -1,57 +1,54 @@
-import React, { Component } from 'react';
+import React, { useState } from 'react';
 import { storiesOf } from '@storybook/react';
 
 import { Grommet, Box, Meter, Stack, Text } from 'grommet';
 import { grommet } from 'grommet/themes';
 
-class MultipleValues extends Component {
-  state = { total: 100 };
+function MultipleValues() {
```
ShimiSun:
```suggestion
const MultipleValues = () => {
```

src/js/components/Meter/stories/MultipleValues.js
```diff
@@ -1,57 +1,54 @@
...
+  const [ total ] = useState(100);
```
ShimiSun: why does `total` need to use state rather than just be a const?

src/js/components/Meter/stories/Bar.js
scfoxcode: I see now that other refactored code is using lambdas so I will do that for consistency, though personally I normally only use those if it's getting passed as an event handler to avoid the need for .bind(this)

I'm curious about the values though. It's never being changed in the parent component, so whether I pass it as a literal or a variable, it wouldn't effect how often it is re-rendered?

I did create a variable in labelled.js as it was being used in more than one location, but again for consistency I will fix this :)

src/js/components/Meter/stories/MultipleValues.js
scfoxcode: You are right, it should just be const, unsure why it was state in the first place

src/js/components/Meter/stories/Labelled.js
scfoxcode: Related to your other replies about these causing re-renders, I'm not sure I understand.

As far as I can see, the difference between storing that in a variable outside the render function, and declaring it as an object inline, is with your approach you aren't creating a new object instance when the component re-renders.

However, that component would only be re-rendered if either
1. The parent got re rendered and the child had no shouldComponentUpdate, in which case it's irrelevant what prop we pass, it's getting re-rendered either way
2. The box component had internal state change forcing a re-render, again, the props being passed are not relevant.

Please correct me if I'm missing the point, my whole team does this a lot at work so I'm interested to know if there is some performance issue here, but if there is I cannot see it

src/js/components/Meter/stories/Circle.js
scfoxcode: Agreed

src/js/components/Meter/stories/Labelled.js
scfoxcode: I could, but it's such a minor point and there are pros and cons both ways.

The only pro I can think of with it being outside of that component is it will only ever be declared and initialised once, so you are saving that initialisation whenever that function gets called. If we were writing a kernel here I may agree :)

The pro of leaving it inside, is that in general, variables should have the smallest scope they need to do their job. The meterValue variable is only relevant to that component so it should be encapsulated by it

src/js/components/Meter/stories/Bar.js
scfoxcode: I've changed them now, didn't notice that the default was lambda's in this project

src/js/components/Meter/stories/Labelled.js
oorestisime: There is no performance issue here and it won't cause any unnecessary re-renders. the prop doesn't change and assinging this to a variable doesn't affect this.

src/js/components/Meter/stories/Labelled.js
oorestisime: Same answer with above. LGTM

src/js/components/Meter/stories/Labelled.js
oorestisime: agreed lets keep it this way. 
1) no need to be reviewing such minor details especially in stories. 
2) the user should be able to get the code from the story in an easy way. putting this elsewhere will reduce readibility in storybook

src/js/components/Meter/stories/Circle.js
```diff
@@ -1,37 +1,32 @@
...
 import { storiesOf } from '@storybook/react';
 
 import { Grommet, Box, Meter } from 'grommet';
 import { grommet } from 'grommet/themes';
 
-class CircleMeter extends React.Component {
-  state = { value: 20 };
+const CircleMeter = () => {
+  const [ value, setValue ] = useState(20);
+  const [ timer ] = useRef(setInterval(() => {
```
oorestisime: 1) haven't tested this and i might be wrong but i think useRef returns a value not an array. 
2) why are we doing a useRef with a setInterval? seems weird to me. you might want to look at this article https://overreacted.io/making-setinterval-declarative-with-react-hooks/#just-show-me-the-code

src/js/components/Meter/stories/Circle.js
```diff
@@ -1,37 +1,32 @@
...
+  }, 2000));
 
-  componentDidMount() {
-    this.timer = setInterval(() => {
-      const { value } = this.state;
-      this.setState({ value: value < 100 ? value + 8 : 20 });
-    }, 2000);
-  }
+  useEffect(() => {
+    return () => {
```
oorestisime: same comment from above besically. i wonder if the setinterval should be here.

nolleto:
Hey @scfoxcode, about storing the variable: https://medium.com/dailyjs/react-is-slow-react-is-fast-optimizing-react-apps-in-practice-394176a11fba#5e63

When you are passing an object or function like this:
```js
const FatherComponent = () => (
  <Component style={{ position: 'relative' }} />
)
```
Everytime `FatherComponent` render a new object will be created (`{ position: 'relative' }`), then `Component` will receive new props and re-render too.

It will receive new props because:
```js
{ position: 'relative' } !== { position: 'relative' }
```
but
```js
const style = { position: 'relative' }
style === style
```

Storing, in this example, `style` will avoid `Component` re-render again with the same props. ;D

oorestisime:
While i see the perf interest to do this specifically on style prop so that the styles are not generating on each render it seems pretty irrelevant for any of the other props we are using. 

IMHO this is premature optimization and should definitely not be on the story. code will be less readable and if we need to strive for something in storybook then its readability for the users. The performance gains from this approach is really minimal.

scfoxcode:
> Hey @scfoxcode, about storing the variable: https://medium.com/dailyjs/react-is-slow-react-is-fast-optimizing-react-apps-in-practice-394176a11fba#5e63
> 
> When you are passing an object or function like this:
> 
> ```js
> const FatherComponent = () => (
>   <Component style={{ position: 'relative' }} />
> )
> ```
> 
> Everytime `FatherComponent` render a new object will be created (`{ position: 'relative' }`), then `Component` will receive new props and re-render too.
> 
> It will receive new props because:
> 
> ```js
> { position: 'relative' } !== { position: 'relative' }
> ```
> 
> but
> 
> ```js
> const style = { position: 'relative' }
> style === style
> ```
> 
> Storing, in this example, `style` will avoid `Component` re-render again with the same props. ;D

Hi @nolleto

Thanks for the detailed reply. I thought you may have been getting at the more than one object instance issue. You are correct that every time the parent re-renders a new object instance will be created, and that the resulting prop change would trigger a re-render.

I was however under the impression that in react, children usually re-render if their parent re-renders even if no props change, so in this case there would be no difference? I mocked up a simple example on codesandbox of what I mean. The inner component takes props that do no change, yet it still gets re-rendered when the parent does. (unless the setter that useState returns changes though I very much doubt it)
https://codesandbox.io/s/gifted-wescoff-r96zd?fontsize=14

I'm not trying to be awkward about this btw :) I'm just curious about what the actual behaviour is for my work.

Thanks

src/js/components/Meter/stories/Circle.js
scfoxcode: I think you are right about useRef being wrong, it should be useState.

I initially created a useEffect with a setState, but that was one extra re-render and more convoluted. I think doing the setInterval in the initialisation of useState is fine, as it will only be called once ever, which will have the same effect as the old code.

Using timeouts and intervals though does trigger some uneasiness, especially in something like react. It may indeed be worth refactoring. A custom hook may well make sense. I resisted that as I'm not too familiar with building my own hooks and I was short on time last night. Will give it another look soon, thanks for the feedback

nolleto:
Hi @scfoxcode, I believe I misunderstood. This is only valid when the component, `Inner` in your example, is a pure component or using `React.memo`. 

https://stackoverflow.com/questions/54946933/how-can-i-prevent-my-functional-component-from-re-rendering-with-react-memo-or-r

But, since there are no performance issue we don't need to worry about it. 
😄

scfoxcode:
@nolleto Thanks for the update. That makes sense. I was getting worried I'd been messing up my teams project at work for the past 2 years :D

src/js/components/Meter/stories/Circle.js
scfoxcode: @oorestisime 
I've fixed this now, ran it in storybook to verify (cool tool by the way). useState didn't work either. The problem is trickier than it first appears, as the setInterval ends up using old state values and things get confusing.

I've refactored it to use a clearTimeout and setTimeout on re-render for now. The custom hook you linked looks like the right solution if this problem crops up in the future and you need a more generic solution in multiple places.

I couldn't see an obvious location for custom hooks in the project, something to maybe think about, but beyond the scope of this pull request I feel.

Thanks for spotting the error

ShimiSun: hi @scfoxcode check out this example to see if it works, it loads the timer once when the component mounts https://github.com/ShimiSun/grommet-graphs/blob/f078d7aa34c64404debbf456bd5ca4bd670be04d/src/components/analytics/geo/GeographyByClient.js#L12

ShimiSun:
Thank you guys for collaborating on this, sending love from Colorado ❤️

PR #3404, opened 2019-10-03
Refactoring Stack component

VictorKunzler:
**What does this PR do?**
Refactor /src/js/components/Stack/Stack.js using Hooks

/src/js/components/Stack/Stack.js

Running yarn test

N/A

#3394 

N/A

No

No

Backwards compatible


VictorKunzler:
Hi @ShimiSun,
Can you help me? Do you know why ci/circleci: chromatic failed? I didn't understand.

ShimiSun:
@VictorKunzler I'm taking a look at it now, it seems that CircleCI wasn't able to fetch environment variables.

ShimiSun:
verified with chroma https://www.chromaticqa.com/build?appId=5d7a63beda813e00202faf9d&number=44

src/js/components/Stack/Stack.js
```diff
@@ -1,69 +1,82 @@
...
 
-class Stack extends Component {
-  toChildIndex = child => {
-    const { children } = this.props;
-    let index = child;
-    if (index === 'first' || !index) {
-      index = 0;
-    } else if (index === 'last') {
-      index = React.Children.count(children) - 1;
+const buildStyleChildren = ({
```
ShimiSun:
```suggestion
const buildStyledChildren = ({
```

src/js/components/Stack/Stack.js
```diff
@@ -1,69 +1,82 @@
...
+  interactiveChild,
+  ...rest
+}) => {
+  const guidingIndex = toChildIndex(guidingChild, children);
+  const interactiveIndex =
+    interactiveChild && toChildIndex(interactiveChild, children);
+
+  const styledChildren = Children.map(
+    children,
+    buildStyleChildren({
```
ShimiSun:
```suggestion
    buildStyledChildren({
```

src/js/components/Stack/Stack.js
```diff
@@ -1,69 +1,82 @@
...
-  render() {
-    const {
-      anchor,
-      children,
-      fill,
-      guidingChild,
-      interactiveChild,
-      ...rest
-    } = this.props;
+const toChildIndex = (child, children) => {
```
ShimiSun: What are your thoughts on moving `toChildIndex` inside of the functional component of Stack, and remove the arguments of `child` and `children`?

ShimiSun:
BTW, for testing your changes, you can also run `yarn storybook` and make sure there are no regressions on the Stack component examples.

src/js/components/Stack/Stack.js
VictorKunzler: I think is not a good ideia, because Stack is calling toChildIndex twice with differents child argument. I can move toChildIndex to inside of the Stack component and remove the second argument, but in my opinion is not the best solution.

ShimiSun: Thanks for your inputs, I overlooked the fact that `child` should remain an argument, but I still think we can move the function inside of Stack, `children` could be removed and the only argument would be `child`. why would you think this is not a good idea?

VictorKunzler: In my opinion the better solution always is the component just render the element, all the others logics should be outside the component, but I can change, no problems.

VictorKunzler: Done :blush:

ShimiSun:
Thank you for your patience with this PR buddy! and mostly thank you for your contribution!

PR #3414, opened 2019-10-04
Refactor Anchor component using hooks

VictorKunzler:
**What does this PR do?**
Refactor /src/js/components/Anchor/Anchor.js using Hooks

/src/js/components/Anchor/Anchor.js

Running yarn test

N/A

#3394 

N/A

No

No

Backwards compatible


ShimiSun:
I've pushed a fix to master few minutes ago and I need this PR to include this fix (to enable some testings), hence the recent merge with master.

src/js/components/Anchor/Anchor.js
```diff
@@ -11,77 +11,70 @@ import { withFocus, withForwardRef } from '../hocs';
...
+  href,
+  icon,
+  focus,
+  label,
+  onClick,
+  reverse,
+  theme,
+  ...rest
+}) => {
+  useEffect(() => {
```
oorestisime: hm, not sure we need a useEffect hook here. you can just add the test without it, if prop is changing then the component will re-render anyway.

VictorKunzler: I used useEffect here because this logic was in constructor, with useEffect I guarantee that will exec just if these props been changed

oorestisime: ok fine you are also avoiding to check if state changes. fine by me

PR #3415, opened 2019-10-04
Refactoring AccordionPanel to use hooks

VictorKunzler:
**What does this PR do?**
Refactor /src/js/components/AccordionPanel/AccordionPanel.js using Hooks

/src/js/components/AccordionPanel/AccordionPanel.js

Running yarn test

N/A

#3394

N/A

No

No

Backwards compatible


src/js/components/AccordionPanel/AccordionPanel.js
```diff
@@ -14,133 +14,125 @@ import { withForwardRef } from '../hocs';
...
-                )}
-              </Button>
-              <Box border={theme.accordion.border}>
-                {animate ? (
-                  <Collapsible open={active}>{children}</Collapsible>
-                ) : (
-                  active && children
-                )}
-              </Box>
+const onHandleMouseOver = ({ onMouseOver, theme, setHover }) => (...args) => {
```
oorestisime: any particular reason to have those functions outside of the component? the code will be a bit simpler without needing to provide those parameters in all the functions. not sure the performance gains are even notable to justify having them outside. thoughts?

VictorKunzler: In my opinion the better solution always is the component just render the element, all the others logics should be outside the component.

ShimiSun: I agree with @oorestisime, l think the code could be more readable and clean having the functions inside the component.

src/js/components/AccordionPanel/AccordionPanel.js
```diff
@@ -60,6 +32,34 @@ const AccordionPanel = ({
     theme,
   );
 
+  const onHandleMouseOver = () => (...args) => {
```
oorestisime: may i suggest to remove the () and in the caller to just add `onHandleMouseOver` seems a tiny bit more readable. same case with the rest of them. thanks for taking care of the changes.

VictorKunzler: Fixed

ShimiSun:
Thank you for your contributions! 🎉

PR #3419, opened 2019-10-05
Refactor Checkbox Component to be functional

negreirosleo:
#### What does this PR do?
Refactor `src/js/components/RangeSelector/RangeSelector.js` to be a Functional Component

`src/js/components/RangeSelector/RangeSelector.js`

Running `yarn teste` and `yarn storybook` to test its behaviour

Running `yarn test`

No

#3394 

N/A

No

No

No


src/js/components/CheckBox/CheckBox.js
```diff
@@ -45,145 +56,131 @@ class CheckBox extends Component {
         'Checkbox of type toggle does not have "indeterminate" state.',
       );
     }
+  };
+
+  useEffect(() => {
```
ShimiSun: I might be missing something, but I don't think we need `useEffect` in this case,  any reason you used it?

negreirosleo: Those checks are being called at the constructor of the class, so I used useEffect to call them at the component render,
do you think it would be better iflike this?
```js
const warnings() => { 
  ... 
}
warnings();

...

return ( <Component />);
```

what do you suggest?

oorestisime: so i don't think we need a function for this. i am thinking the useEffect is ok and we can add checked and indeterminate in the array dependencies.

ShimiSun:

Issue #3248, opened 2019-08-05
Enable Select to grow the list with new options


juanlanus:
Top-gun Select controls have options to add new items to the list. 
Like for example https://react-select.com/creatable 
As I see it, it should be available only in searched Select's. 
The functionality is to pass the search value typed by the user to the caller so she can handle the new value (for example popping-up a dialog, or whatever). 
This should happen when the user leaves the Select with enter or tab, not with esc or clicktouching outside. The user action should be one of the args of the handling function. 
I need this in my site to handle collaborative data creation. 

This is a feature request. I'm sorry I'm not smart enough to provide a PR even for this, that seems so simple to me. 
I would be happy to update the docs if this feature is added.

markjackson02:
Yeah I am using react-select to accomplish this currently. This would be a welcome addition.

nik72619c:
I would like to work on this

juanlanus:
Hi Nik,
There are two kinds of options list: the simple strings array, and an array of objects.  
As I see it now (working with an array of objects), this feature would be practical to implement in the case of the strings array only, because the other case implies showing the user form interactions that will always be very application-specific. 
I would like to perform as a tester for you  :-)

nik72619c:
cool, thanks !

nik72619c:
Can I please know which component is exactly facing this issue?
Which one of the select component is this?

juanlanus:
Select is a single component. 
It can be provided two kind of options lists: plain strings array, or objects array. 
IMO what you want to do makes sense only in the plain strings array case.

nik72619c:
so basically the searched selects here, must work like react-select about which you mentioned. I believe the current behaviour is that the searched select inputs only display the current value being clicked or selected. Am I thinking in the right direction @juanlanus ?

juanlanus:
Yes, Select displays a list of values and lets the user to choose one of the displayed values. 
The text the user types into the control is used as a search string, not as an input value. 
This issue is asking to allow the user to add the typed text to the list as a new item when it doesn't matches any of the existing items and the user wants to.

oorestisime:
Hey folks before jumping into the implementation lets take some time to think a bit more on this. This will change the select behaviour so we need to think a bit on the API before. @ericsoderberghp @ShimiSun would love your thoughts as well.

nik72619c:
#3430 please have a look and suggest changes :)

ShimiSun:
[...]

PR #3430, opened 2019-10-06
Added functionality to handle new searched value

nik72619c:
Added functionality to handle new searched value

#### What does this PR do?
this solves issue #3248. it detects the enter key and for now, an alert box popover is displayed. 

The reviewer should start from the file Search.js. The Select component inside it now has an onkeypress event.

The manual forms of testing like breaking change detections etc are checked. It is ensured that the existing code does not break

The search select component inside the storybook examples consist of a search select dropdown. It can be tested by typing an option and then clicking enter.

No

None

nik1234

Yes

Yes

Yes


juanlanus:
IMO there should be at least a ^Z option just after having added the new list item in case the user realizes just after enter that there is a typo (as usually happens). 
Consider opening a new issue to enable deleting items from the list using the keyboard or with the mouse/finger hitting an [x] button shown only at the end of the selected/hovered choice. 
This is to follow the rule that says that the user must be allowed to undo everything (almost) so she can use the UI without the fear of introducing an irreversible unrecoverable fatal modification in the data.

ShimiSun:
@nik72619c hi buddy! thanks for working on this.
I'm running the deployment of your PR https://deploy-preview-3430--sad-tereshkova-173d07.netlify.com/?path=/story/select--search and I don't get the alert as you described.
I've opened the Select component on storybook -> Search -> typed 'shimi' as a string and Enter, but nothing happened. what am I missing?

nik72619c:
I'll just look into it

nik72619c:
main
The build seems to work fine in my case .
I believe its because you didnt enter a valid option which was present in the list.

ShimiSun:
I've done some clean-ups on this PR and I think it is good to go, @ericsoderberghp I'll appreciate your second review.

ShimiSun:
I think we already took care of this functionality, @IanKBovard can you please verify it already exists so we can close this PR?

IanKBovard:
@nik72619c just double checking the intention of this pr. It's to add enter key functionality to the select component? If that's the case, we can close this pr.

IanKBovard:
@nik72619c I know it's been a while, just checking in to see where you are with this pr.

ShimiSun:
I'll be closing this PR since I think we already approached a fix elsewhere, feel free to reopen if still relevant.
Thank you!

PR #3436, opened 2019-10-07
Refactor Diagram Stories to be functional components

negreirosleo:
#### What does this PR do?
It refactor `src/js/components/Diagram/stories/` to be functional components

No particular order required.

Tested with the storybook checking if its behaviour was as expected
Testing with `yarn storybook`
No
#3394 
N/A
No
No
No

### DIFF
--- a/src/js/components/Diagram/stories/Animated.js
+++ b/src/js/components/Diagram/stories/Animated.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useReducer, useEffect } from 'react';
 import { storiesOf } from '@storybook/react';
 
 import { Box, Diagram, Grommet, grommet, Stack, Text } from 'grommet';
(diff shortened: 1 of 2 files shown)


src/js/components/Diagram/stories/Progressing.js
```diff
@@ -28,57 +28,58 @@ const connection = (fromTarget, toTarget, { color, ...rest } = {}) => ({
...
-  componentDidMount() {
-    this.timer = setInterval(() => {
-      const { topRow } = this.state;
-      this.setState({
-        topRow: fullTopRow.slice(
-          0,
-          topRow.length < fullTopRow.length ? topRow.length + 1 : 1,
-        ),
-      });
+const SimpleDiagram = () => {
```
ShimiSun:
```suggestion
const ProgressingDiagram = () => {
```
Renaming the component to use a proper name that is aligned with the story.

src/js/components/Diagram/stories/Progressing.js
```diff
@@ -28,57 +28,58 @@ const connection = (fromTarget, toTarget, { color, ...rest } = {}) => ({
...
+            </Box>
+          </Box>
+          <Diagram connections={connections} />
+        </Stack>
+      </Box>
+    </Grommet>
+  );
+};
 
 storiesOf('Diagram', module).add('Progressing', () => <SimpleDiagram />);
```
ShimiSun:
```suggestion
storiesOf('Diagram', module).add('Progressing', () => <ProgressingDiagram />);
```
Renaming the component to use a proper name that is aligned with the story.

src/js/components/Diagram/stories/Animated.js
```diff
@@ -66,58 +66,53 @@ const Container = ({ node, index }) => (
...
 
-  componentDidMount() {
-    this.timer = setInterval(() => {
-      const { draw } = this.state;
-      this.setState({
-        draw: !draw,
-      });
-    }, 3000);
-  }
+  const [draw, toogleDraw] = useReducer(reducer, true);
```
ShimiSun: I think that using a `useReducer` as an alternative to `useState` makes sense for a more complex state logic.
Hence I'm just wondering why you chose it vs `useState`, keeping in mind that the storybook components supposed to provide simple examples. let me know your thoughts.

negreirosleo: Imagine the following scenario:
```js
useEffect(() => {
    const id = setInterval(() => {
      setDraw(!draw);
    }, 3000);
    return () => clearInterval(id);
  }, []);
```

This should create the setInterval at the start right? But if you look closely you arent passing the `draw` dependency to useEffect, you are kinda cheating react, this will create an unexpected behaviour.
It will work like this: 
```
setValue(!true) every 2seconds

first State -> draw === true
2s after -> draw === false
2s after -> draw === false
```

The component isnt re-rerendering, the useEffect its always use the `draw === true` state to change the state every interval tick.

The solution would be:
```js
useEffect(() => {
    const id = setInterval(() => {
      setDraw(!draw);
    }, 3000);
    return () => clearInterval(id);
  }, [draw]);
```
Right??
However it wouldnt work because you are triggering the effect to change `draw` when `draw` is changed. See the problem?
By using `useReducer`, at render it calls the dispatcher to alters its state. This way you keep the dependencies on useEffect clear.

with useReducer it work like this:
```
Every 3 seconds the toogleDraw function will be called and it will invert the current `draw` state
```

src/js/components/Diagram/stories/Progressing.js
negreirosleo: Oh sure, thanks for noticing :+1:

src/js/components/Diagram/stories/Animated.js
ShimiSun: Thank you for the detailed explanation, nice work

PR #3439, opened 2019-10-08
update contribution docs

aravind1078:
#### What does this PR do?
update contribution docs
contributing docs doesn't really show how to run/see our local changes. I got this info from one of my reviewers till then I could not figure out how to run my changes. so this should help the future contributors.

no
no
backward compatible

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -147,6 +147,7 @@ started you should:
 1. fork the `grommet` repository
 1. clone it `git clone https://github.com/<your-username>/grommet.git`
 1. install dependencies using: `yarn install`
+1. start running the storybook locally using: `yarn storybook`
 
 The components code lives in `src/js/components`. The structure of the
 project is a bit particular since it is using lots of internal tooling to try to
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -147,6 +147,7 @@ started you should:
 1. fork the `grommet` repository
 1. clone it `git clone https://github.com/<your-username>/grommet.git`
 1. install dependencies using: `yarn install`
+1. start running the storybook locally using: `yarn storybook`
```
ShimiSun: I don't think you need to run a storybook in order to file a PR. Can you please share your thoughts on why you think it is needed?

aravind1078: Yes, we don't need to run a storybook in order to file a PR. But when we make any changes locally there is no way we know how to run the storybook. It can be optional step before going and submitting the pr. I didn't know how to run locally until you mentioned on one of my PR review.so I thought it would be useful new contributers/developer coming, When trying to make changes locally.

ShimiSun: Gotcha. It is detailed in the README.md page https://github.com/grommet/grommet#explore, but I might agree it should be mentioned on the Contribution page as well.
Would you be able to add a section for testing your code and explain it there?

aravind1078: ok sure, will give it a try.

aravind1078: @ShimiSun  added as a seperate section

ShimiSun: Great thanks for the follow up!

Issue #3186, opened 2019-07-04
Creating image links: wrapping <Image> with <Anchor> causes image to render full size, ignoring parent <Box>


markjackson02:
### Expected Behavior
Anchor should have no effect on sizing of children

### Actual Behavior
Wrapping `<Image>` with `<Anchor>` causes image to render full size, ignoring parent `<Box>` height and width.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-github-issue-3186-0m2dl (see src/Image.js and https://0m2dl.codesandbox.io/image)

Mr-Magnificent:
I believe `max-width: 100%; max-height: 100%` on `<img>` in `src/js/components/Image/StyledImage.js` could solve the issue? Can I take this up?

britt6612:
@Mr-Magnificent I'm sure @ShimiSun will be grateful if you take this up! 👍 🥇

Mr-Magnificent:
How can I test if my changes solves the issue?

oorestisime:
You can create a new story on storybook with a reproduction of the issue.

Mr-Magnificent:
Should the patch include the story?

oorestisime:
If you feel this could be interesting for users why not!

Mr-Magnificent:
> If you feel this could be interesting for users why not!

To be honest, I am not entirely sure. Maybe check if it adds value during review? :sweat_smile:

oorestisime:
sure will be easier to review! :)

IanKBovard:
If #3443 solved this issue we can close this.

markjackson02:
Retested on 2.14.0. This still seems to be an issue.

halocline:
@markjackson02 -- I looked into this and think there are a couple of options here, but my recommendation would be to move the `Anchor` so that it contains the `Box` and `Image`, rather than the `Anchor` being placed in between the `Box` and `Image`.
```
<Anchor>
  <Box height="small" width="small" >
    <Image fit="cover" />
  </Box>
</Anchor>
```

I have modified your CodeSandbox example to demonstrate: https://codesandbox.io/s/grommet-github-issue-3186-gingn?file=/src/Image.js

**Explanation**
[...]

PR #3298, opened 2019-08-26
feature(Paragraph): Adds support for `fill` property

Jakobo:
#### What does this PR do?
This adds support for `fill` to the Paragraph component. When enabled, paragraphs no longer are constrained by the theme's max-width, instead taking up 100% of their parent container. Setting `fill` takes precedence over your size's max-width setting.

#### Where should the reviewer start?
`components/Paragraph`

#### What testing has been done on this PR?
Storybook, snapshot tests updated, README for Paragraph updated.

#### How should this be manually tested?
Launch storybook, verify story.

#### What are the relevant issues?
Resolves #3222

#### Do the grommet docs need to be updated?
Yes. Doc updates included in PR.

#### Should this PR be mentioned in the release notes?
Yes.

#### Is this change backwards compatible or is it a breaking change?
Backwards Compatible.

**DCO Included in Commit Message**

Issue #3222, opened 2019-07-24
Paragraph to support "fill" prop


Jakobo:
Just an observation on `Paragraph`.

9/10 times the max-width makes total and complete sense. Then, there's that _one_ time when you want the paragraph's width controlled by its parent `Box`. Currently the only fix is to replace the Paragraph calls with the non-semantic `Text` component.

### Proposed Solution
There is a property, we'll call it `fill` for now, which behaves like the box's `fill` property, allowing for the values `["horizontal", "vertical", true, false]`. The default value of this would be `false`, and provides the current behavior.

* A value of `horizontal` will set `width` to `100%`
* A value of `vertical` will set `height` to `100%`
* A value of `true` sets both above
* A value of `false` takes the `max-width` setting from the paragraph's definition, along with the other Paragraph defaults

#### Box+Text
This is probably the cleanest workaround, although it does not satisfy the requirements of a semantic `<p>` tag in output. `<Box as="p"><Text>...content</Text></Box>` will give you all of the flex/box support you'd want with an inner text object. The only drawback is the non-semantic `<span>` added inside the paragraph.

If we opted for this solution, we would want to update the grommet documentation to explain how to do paragraphs that don't force a maximum width for readability.

#### Text as=
This is a variant of the above, using the `as` prop to change `Text` from its default span to a `p` tag. It has some drawbacks though, as using `Text` by itself forgoes grommet's support for `border`, `animation`, `background`, and others.

### Reference
* Reference on behavior of `fill` in the `StyledBox` https://github.com/grommet/grommet/blob/master/src/js/components/Box/StyledBox.js#L120
* #1775 we actually talked about this once before. Paragraph has strong opinions on its default width (good!) but no way for someone to create a semantic `<p>` tag and override the Paragraph in specific instances (bad!). The solution of manually doing `<p><Text>...content</Text></p>` or `<Box><Text>...content</Text></Box>` are workarounds.

ShimiSun:
Thanks for the detailed observation, suggested solution and workarounds! with the way you put it it does make sense for me to have a `fill` prop. 
I'd like to hear @ericsoderberghp thoughts here as well.

oorestisime:
I like this as well.. I find myself doing the Box+Text alternative a few times as well :)

ericsoderberghp:
I'm OK with the basic concept of being able to override the max-width for special circumstances.

One other options is to do `style={{ maxWidth: 'none' }}`. This might be sufficient if the frequency is low enough.

[...]

PR #3443, opened 2019-10-08
Creating image links #3186

Mr-Magnificent:
Signed-off-by: Ayush Poddar <ayush_poddar@live.com>

#### What does this PR do?
Images within Anchor acquire the size of the container
Storybook
#3186 

Screenshot_20191008_202840

Maybe
NA
NA


ShimiSun:
Thank you for filing this PR and addressing this issue!
I'm thinking whether those styling should be applied directly as you suggested or via an additional `fill` prop.
Adding those styling to existing used styles might impact current users and I'm not sure if it will be backward compatible in those cases.
I'd love a second opinion on this one @ericsoderberghp.

oorestisime:
Yeap was thinking the same with @ShimiSun here. Probably will make images stretched if the container is bigger than the image and this won't be desired by users (or at least would be a breaking change) so i am inclining for a fill prop as well.
Nice investigation and thanks for working on this

Mr-Magnificent:
If possible, I'd love to continue working on this but I'd need some help and directions on how to implement the above method. Maybe point to an example in the codebase? :smile:

oorestisime:
For sure! We are not taking over :)

So you can have a look at Paragraph component it has recently acquired the fill prop :)
I think fill should be by default false for backwards compatibility.

Generally the files to change should be:

- doc.js for the documentation and prop types
- index.d.ts for the typescript
- Image file to pass in the fill prop to styled image
- styled image file to take into account the fill prop and do something along the following https://github.com/grommet/grommet/blob/master/src/js/components/Paragraph/StyledParagraph.js#L16

We would also love if you can can add a test in the tests folder making use of the fill prop. 

If you have any question feel free to ask here or better join the slack channel :)

ShimiSun:
Thank you for owning this one @Mr-Magnificent, as @oorestisime mentioned above, this PR will be a useful reference for you https://github.com/grommet/grommet/pull/3298
Let us know if you need any further help here!

Mr-Magnificent:
Actually, I was preoccupied with some college work. Will be a couple of days before I can jump back in. I hope that is not a problem.

ShimiSun:
@Mr-Magnificent sure thing, let us know if you need any further guidance when you are back at it.

ericsoderberghp:
I think I am missing something, why isn't the `fit` property sufficient to address this?

ShimiSun:
@ericsoderberghp Interesting comment.
I took a look at `object-fit` CSS, and we might want to enhance our `fit` prop. 
At the moment `fit` supports `cover` and `contain` but it doesn't support `fill`, which might answer this issue. 
The difference between `cover` and `fill` of the `fit` values is
>cover
>The replaced content is sized to maintain its aspect ratio while filling the element’s entire content box. If the object's aspect ratio does not match the aspect ratio of its box, then the object will be clipped to fit.
>fill
>The replaced content is sized to fill the element’s content box. The entire object will completely fill the box. If the object's aspect ratio does not match the aspect ratio of its box, then the object will be stretched to fit.

https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit

ericsoderberghp:
I'm just wondering when `fill` might be used, since it doesn't preserve the aspect ratio.

Mr-Magnificent:
From what I could test out from the codebase `object-fit` doesn't seem to work because the `<a>` which wraps the image because of `<Anchor>` in the DOM acquires the complete size of the image because of it's inline-nature and all of the `object-fit` seems to turn non-functional.

Screenshot_20191015_132625

src/js/components/Image/doc.js
```diff
@@ -26,6 +26,9 @@ export const doc = Image => {
       PropTypes.string,
       PropTypes.bool,
     ]).description('Transparency of the image.'),
+    fill: PropTypes.bool.description(
+      'Set maximum width to that of parent element',
+    ),
```
ShimiSun: I think you should mention both width and height.

src/js/components/Image/doc.js
```diff
@@ -26,6 +26,9 @@ export const doc = Image => {
       PropTypes.string,
       PropTypes.bool,
     ]).description('Transparency of the image.'),
+    fill: PropTypes.bool.description(
```
ShimiSun: I'm inclining for consistency of the fill prop.
When `fill` is defined on other components it allows the values of 'vertical' | 'horizontal' | boolean.
It's been done a few days ago in a similar way for WorldMap.
Do you feel like it is needed for Image as well?

ShimiSun:
@Mr-Magnificent thank you for following up on this!
I'm not sure I understand the image you uploaded, is this image created from your branch and uses the `fill` for image?
Regardless, I think it is worth asking the question of whether we want to have the `fill` as an independent prop as you suggested, or is it more semantically accurate to add it as on option for `fit` since internally being used as part of `object-fit`.

Mr-Magnificent:
Yes, I had made the changes to see if `object-fit = fill` worked in my local branch by adding the necessary code. It can be seen that `object-fit` was indeed applied on the image by checking the `CSS pane`, but the image still overflowed the container (highlighted).

Yes, I do agree about `fill` and `fit` since both of them seem so similar and I wasn't even sure about the wordings to put in the docs since they try to achieve similar things logically. I would be glad if you could point out the course of action.

src/js/components/Image/doc.js
Mr-Magnificent: Sure thing! I had referred to [https://github.com/grommet/grommet/issues/3222#issuecomment-516176939](url) to assume that `fill` should assumes only `true` or `false`. Should all the values be supported?

ShimiSun:
I think that as much as it will be consistent with other components to use `fill` as a prop, it will be more semantically accurate to use it as one of `fit` options. but I'd love a second opinion cause I'm not that convinced either...

src/js/components/Image/StyledImage.js
```diff
@@ -14,9 +14,15 @@ const fitStyle = css`
   object-fit: ${props => FIT_MAP[props.fit]};
 `;
 
+const fillStyle = css`
+  max-width: 100%;
```
ericsoderberghp:
```suggestion
  width: 100%;
```

src/js/components/Image/StyledImage.js
```diff
@@ -14,9 +14,15 @@ const fitStyle = css`
...
+  max-height: 100%;
```
ericsoderberghp:
```suggestion
  height: 100%;
```

src/js/components/Image/__tests__/__snapshots__/Image-test.js.snap
```diff
@@ -1,5 +1,32 @@
...
+  max-width: 100%;
+  max-height: 100%;
+}
+
+<div
+  className="c0"
+>
+  <img
+    className="c1"
+    fill={true}
```
ericsoderberghp: We shouldn't be leaking the `fill` into the DOM. We handle this in other areas by renaming the prop in Image to something like `fillProp` and passing that to StyledImage.

src/js/components/Image/doc.js
ericsoderberghp: I think we're OK to just have it be a boolean for now.

src/js/components/Image/doc.js
```diff
@@ -26,6 +26,9 @@ export const doc = Image => {
...
+      'Set maximum width to that of parent element',
```
ericsoderberghp:
```suggestion
      'Whether the image expands to fill all of the available width and height.',
```

src/js/components/Image/stories/Fill.js
```diff
@@ -0,0 +1,26 @@
...
+    <Grommet theme={grommet}>
+      <Box align="start" gap="small">
+        <Box height="small" width="small" border>
+          <Anchor href="#">
+            <Image fit="cover" fill src="//v2.grommet.io/assets/IMG_4245.jpg" />
+          </Anchor>
+        </Box>
+        <Box height="small" width="small" border>
+          <Anchor href="#">
+            <Image fit="cover" src="//v2.grommet.io/assets/IMG_4245.jpg" />
```
ericsoderberghp:
```suggestion
            <Image fit="contain" fill src="//v2.grommet.io/assets/IMG_4245.jpg" />
```

ShimiSun:
hi @Mr-Magnificent, will you be able to follow up on the last small comments? we appreciate your work on this issue.

Mr-Magnificent:
Yeah, Sorry was just slightly preoccupied with some work. I have made the reviewed changes. Kindly check.

src/js/components/Image/__tests__/__snapshots__/Image-test.js.snap
ShimiSun: @Mr-Magnificent we shouldn't have our props to appear on the DOM in order to keep the DOM nice and clean, can you please take a look at this last comment?

Mr-Magnificent:
Hi @ShimiSun, I was wondering whether these changes are okay?

ShimiSun:
Yep, @Mr-Magnificent , thank you for your friendly reminder. reviewing again now

src/js/components/Image/doc.js
```diff
@@ -26,6 +26,9 @@ export const doc = Image => {
       PropTypes.string,
       PropTypes.bool,
     ]).description('Transparency of the image.'),
+    fillProp: PropTypes.bool.description(
```
ShimiSun: I think the prop should still be called `fill` for the users (consistent across grommet) and it should use the `fillProp` internally as an alias.

Mr-Magnificent:
@ShimiSun kindly check whether this is okay?

Mr-Magnificent:
Hi @ShimiSun, Does this require anything else too?

ShimiSun:
This change is included on grommet 2.8.1 🎉

Issue #3389, opened 2019-10-01
Improve documentation of extend theme concept


brianbancroft:
The extend theme object is a useful way to add other non-common attributes to Grommet elements such as the RangeInput, but lack of applied examples in the documentation may preclude their adoption among other developers

### Expected Behavior

Somewhere in the documentation (or storybook), there should be a fleshed-out example of an extend theme object so that developers who code through example can use it with less friction. 

### Actual Behaviour

In documentation pages such as [that of the RangeInput](https://v2.grommet.io/rangeinput#theme), there are two examples which are not clear. 

![image](https://user-images.githubusercontent.com/6225122/65983990-9a051480-e433-11e9-9032-4590b5aab28f.png)

#### "Any CSS"
This doesn't tell me whether I should write the rules directly, or whether I should encase them within a class or similar. While Grommet does use styled-components, it doesn't address whether tools available to that library (such as the '&' operator) are available. 

#### "(props) => {}"
This does not show what could exist in the object, or where it could be applied. Or how to pull from the props as well. 

This is not to detract from the library itself. It is a lovely set of abstractions that I've come to enjoy the more I use them. Thank you for all of your hard work

ShimiSun:
I think we can close this one now unless @brianbancroft thinks we can improve elsewhere.

brianbancroft:
Concurred. Closing.

rodmartinezmedina:
Hi, So where can i find more detailed explanation about the extend property to clarify? I've checked all the links in this and the other post and can't find it.
Thanks

IanKBovard:
Here is an example of TextInput using extend
https://storybook.grommet.io/?path=/story/textinput--themed
and Heading using extend with dynamic variables
https://storybook.grommet.io/?path=/story/heading--extend

PR #3446, opened 2019-10-09
📝  Add applied example to rangeinput extends

brianbancroft:
## Description
This PR proposes an example of how to document the `extends` option in the Grommet components that is less abstract and more applied, in support of #3389 

This PR only affects documentation, so no testing took place aside static checking. 

If the general idea is welcome, I would go and extend this to the other components in either this PR or in others.


ShimiSun:
Thanks, @brianbancroft ! can you please push the results of the updated snapshots so we could see how the text was auto-generated? 
run `yarn test-update` and commit the snapshots that were created following these changes.
This should also fix the `jest` failures.

brianbancroft:
@ShimiSun thanks for the direction. This is now done

src/js/components/RangeInput/doc.js
```diff
@@ -47,7 +47,9 @@ export const themeDoc = {
   },
   'rangeInput.extend': {
     description: 'Any additional style for the RangeInput.',
-    type: 'string | (props) => {}',
+    type: `string | (props) => \`
+      color: #c9c19f;
```
ShimiSun:
```suggestion
      color: #44444;
```
Choosing a color that is less distracting.

ShimiSun: Or another idea is 
```suggestion
      any CSS styling;
```
so users won't be confused it is only for colors, thoughts?

brianbancroft: Actually, the second one looks better. I'm going to have a little time on holiday monday, so I'll be sure to do that. 

Thanks!

ShimiSun:
Thank you for your contribution and for helping the community with improving the docs!

Issue #3408, opened 2019-10-04
Menu items onClick handlers do not fire on Firefox as of v2.7.9


ryanjchr:
### Expected Behavior
Clicking on a `<Menu items={[{ onClick: ... }]}>` item should fire its `onClick` handler if provided on all browsers.

### Actual Behavior
The `onClick` handler does not fire on the latest version of Firefox, but does on Chrome.
Reverting to Grommet 2.7.8 fixes the issue.

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/s/grommet-v279-menu-items-broken-bhr5w

### Steps to Reproduce
1. Have a `<Menu items={[{ onClick: ... }]}>` component with an item containing an `onClick` property
2. Click the item with the `onClick` property
3. Note that the handler does not execute if using Firefox

### Your Environment
- Grommet version: 2.7.9
- Browser Name and version: Firefox 69.0.2, Chrome 77.0.3865.90
- Operating System and version (desktop or mobile): macOS 10.14.6

taysea:
Thanks for raising this issue! It looks like potentially Firefox is having an issue with the nested onClick? https://stackoverflow.com/questions/24103072/reactjs-onclick-handler-not-firing-when-placed-on-a-child-component

I'm going to look more into this to understand why the recent changes to Menu are causing this to occur.

ShimiSun:
linking #3334

PR #3334, opened 2019-09-10
Fix menu keyboard accessibility

taysea:
#### What does this PR do?
Fixes keyboard accessibility for Menu. The behavior now matches what is recommended here: https://www.w3.org/TR/wai-aria-practices/#menu. Comments have been added in the code to note the expected behavior in case of future changes.

**Expected behavior:**

**To open menu when menu button is focused:**
- Space/Enter/Up/Down will open menu

**To navigate within menu:**
- Up/down arrow keys can be used and will loop through options (keeping focus within the Menu)
- Tab can be used, but once the last menu item is reached, Tab will close the Menu and continue through page content.

**To close the menu:**
- Tabbing beyond the first or last menu item.
- Esc will close the menu

**To make a selection:**
- Enter key is pressed.
- Space is pressed.

#### Where should the reviewer start?
src/js/components/Menu/Menu.js

#### What testing has been done on this PR?
Local testing in storybook with keyboard and screen reader.

#### How should this be manually tested?
Local testing using keyboard and/or screen reader.

#### Any background context you want to provide?
Keyboard interaction was not properly supported by Menu.

#### What are the relevant issues?
Issue #3225

#### Do the grommet docs need to be updated?
No.

#### Should this PR be mentioned in the release notes?
No.

#### Is this change backwards compatible or is it a breaking change?
Backwards compatible. Users could previously tab between menu items and they still can. However, now the styling is updated to provide the hover state when menu items are in focus.

PR #3449, opened 2019-10-09
Fix onClick behavior in Menu

taysea:
#### What does this PR do?
This PR fixes the cross browser onClick functionality that was removed in the previous update to Menu. The mouse can be used to open and select menu items. For keyboard accessibility, space/enter/down/up can be used to open the menu and tab can be used to navigate into and out of the menu. If the user tabs out of the menu, the menu will close. These functions work on Chrome, Safari, FireFox, Edge, and IE11.

These changes have been tested by creating an updated test using react testing library. They have also been tested manually by adding a console statement when the onClick handler of a menu item is called. This console statement printed out which item was clicked to ensure the correct item was being registered. This was tested using both mouse and keyboard.

To manually test, you can add this line `console.log(item)` to line 232 of Menu.js and then interact with the Menu as noted above.

In the process of trying to improve keyboard accessibility on the Menu component, the onClick functionality was affected. This PR fixes the regression while also maintaining some of the improved accessibility. Future steps will need to be taken to improve accessibility for arrow keys.

Issue #3408 

No.

No.

Yes, backwards compatible.


ShimiSun:
linking to #3334

src/js/components/Menu/__tests__/Menu-test.js
```diff
@@ -140,39 +140,36 @@ describe('Menu', () => {
...
-
-    // moves to the first suggestion
-    fireEvent.keyDown(getByLabelText('Open Menu'), {
-      key: 'Up',
-      keyCode: 38,
-      which: 38,
+    fireEvent.keyDown(getByLabelText('Close Menu'), {
+      key: 'Tab',
+      keyCode: 9,
+      which: 9,
```
ShimiSun: Are we closing the Menu three times using Tab? 
Before these changes, it made sense to use the Tab in order to navigate within the items of the 'Open Menu', but when the Menu is closed how does that makes sense?

src/js/components/Menu/__tests__/__snapshots__/Menu-test.js.snap
```diff
@@ -2492,9 +2489,6 @@ exports[`Menu open and close on click 3`] = `
   border: none;
   padding: 0;
   text-align: inherit;
-  background-color: rgba(221,221,221, 0.4);
```
ShimiSun: Please, doubles check all these styles didn't exist prior to your previous changes as well. I want to make sure we are not introducing regression from the style aspect.

src/js/components/Menu/__tests__/Menu-test.js
```diff
@@ -140,39 +140,36 @@ describe('Menu', () => {
...
     );
     expect(container.firstChild).toMatchSnapshot();
 
-    // pressing down 3x: first opens the drop,
-    // second moves to the first suggestion
-    // third moves to the last suggestion
+    // pressing space: opens drop
+    // tab moves to menu label
+    // second moves to first item
+    // third moves to second item
```
ShimiSun: This description no longer describes the current code flow, I think it should be adjusted.

src/js/components/Menu/Menu.js
```diff
@@ -45,6 +45,10 @@ To make a selection:
 - Space is pressed.
 */
 
+const INDEX = {
+  NONE: 'none',
```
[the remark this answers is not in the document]
taysea: You're right, we don't the the extra abstraction. I will change it.

src/js/components/Menu/Menu.js
```diff
@@ -71,54 +75,45 @@ const Menu = props => {
...
-          buttonRefs[items.length].focus();
-          setActiveItemIndex(-1);
-        }
-      }
-    }
-  }, [isManagingFocus]);
 
   const onDropClose = () => {
-    setActiveItemIndex(-1);
+    setActiveItemIndex(INDEX.NONE);
```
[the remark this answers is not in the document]
taysea: When the menu opens with the mouse, I don't believe we want any of the menu buttons to have an active state styling. If the active item index is -1 by default, when the menu has `align.top === 'top'`, then the menu label button will receive active styling upon opening the menu even if it not being hovered over. By making the default active index NONE, this styling will not occur until keyboard is used to interact with the menu or the mouse hovers over a menu item.

src/js/components/Menu/__tests__/Menu-test.js
taysea: We're not closing the menu, but the 'Close Menu' button refers to the Drop div that gets added to the DOM when a menu opens, and it seemed that the keyDown focus needed to refer to this element in order to advance through the menu items which are contained in this Drop. So, the menu is not closing but using that label text is helping to tell the test where the keyDown should occur. I am still strengthening my understanding of test writing, so another set of eyes may be helpful.

src/js/components/Menu/__tests__/Menu-test.js
taysea: This description does refer to the expected flow, but if this is not what is happening please let me know. But this describes the test steps as:
1. space opens the menu
2. tab to the menu label
3. tab to the first item
4. tab to the second item
Then, the second item is selected with 'Enter'.

src/js/components/Menu/__tests__/__snapshots__/Menu-test.js.snap
taysea: I confirmed these styles didn't exist prior to my previous changes.

ShimiSun:
Hi there! Not sure why all my comments weren't posted yesterday, sorry for the late post

ShimiSun:
Actually, I think we are working on double threads with a similar PR #3464. Can we close this PR?

Issue #10, opened 2015-05-01
Unable to install sample app using npm on windows


donwise:
I was able to install grommet successfully on Windows 7, but when I try to build the sample app using npm, I get these build errors:

VCBUILD : error Message:  [c:\Tutorials & Examples\grommet\sample-app\node_modu
les\webpack-dev-server\node_modules\socket.io\node_modules\engine.io\node_modul
es\ws\build\binding.sln]
VCBUILD : error Message:  [c:\Tutorials & Examples\grommet\sample-app\node_modu
les\webpack-dev-server\node_modules\socket.io\node_modules\engine.io\node_modul
es\ws\build\binding.sln]

Then, when I run gulp dev, I get this error:

ERROR in ./~/css-loader!./~/sass-loader?outputStyle=expanded&includePaths[]=c:/
utorials & Examples/grommet/sample-app/node_modules!./src/scss/index.scss
Module build failed: file to import not found or unreadable: grommet/scss/gromm
t-core/index
Current dir:  (1:9)
 @ ./src/scss/index.scss 4:14-214 12:19-219

If I try to load the app in my browser, it does not appear to load correctly and I get a similar error in the debug console.

Any help would be appreciated!
Don

PR #3455, opened 2019-10-11
Refactor Accordion.js to use hooks

husseyexplores:
#### What does this PR do?
This PR refactors Accordion.js to use hooks.

`/src/js/components/Accordion/Accordion.js`

Storybook

Storybook

Nothing, but I do have a question. Why are we using `getDerivedStateFromProps` here? What is the use case? I don't see any difference in storybook examples even after removing that method. I implemented the hooks version of it anyway though. The results still make no difference with or without it while testing it manually in storybook examples.

#3394 

No

No

Backwards compatible


husseyexplores:
Some Accordion snapshot tests are failing (were already failing before I started working on it) because of the swapping CSS classes.
For example, snapshot contains this: ` className="c10 c1"` , but the render in the test contains this: `className="c1 c10"`. I'm not sure what to do regarding this, update the snapshots or not.

src/js/components/Accordion/Accordion.js
```diff
@@ -7,85 +7,76 @@ import { AccordionContext } from './AccordionContext';
...
-    const { animate, children, messages, ...rest } = this.props;
-    const { activeIndexes } = this.state;
-
-    delete rest.onActive;
-
-    return (
-      <Box role="tablist" {...rest}>
-        {Children.toArray(children).map((panel, index) => (
-          <AccordionContext.Provider
-            key={`accordion-panel_${index + 0}`}
```
husseyexplores: One more question here, does `index + 0` won't make any difference I believe? Should I remove that?

husseyexplores: Oh I get it now. It is there to silence `eslint(react/no-array-index-key)` error message.

src/js/components/Accordion/Accordion.js
```diff
@@ -7,85 +7,76 @@ import { AccordionContext } from './AccordionContext';
...
+          }}
+        >
+          {panel}
+        </AccordionContext.Provider>
+      ))}
+    </Box>
+  );
 }
 
+Accordion.defaultProps = {
```
ShimiSun: Can we embed the defaultProps values as part of the props declaration on lines #10-#18

husseyexplores: I thought it was not the recommended way of doing it (idk why though) and just came to know it is indeed the preferred way. Thanks for letting me know. I just made the change and also rearranged the props alphabetically.

ShimiSun:

PR #65, opened 2015-06-30
Switch from jsx loader to babel

alansouzati:
This PR aims to move from jsx loader to babel, which is now the recommended way to transpile future JS code.

PR #3461, opened 2019-10-13
Refactor Layer component to use hooks

husseyexplores:
#### What does this PR do?
This PR refactors Layer component to use hooks.

`/src/js/components/Layer/Layer.js`
`/src/js/components/Layer/LayerContainer.js`

Storybook

Storybook

Test `should be null prior to mounting, displayed after mount` is failing because it uses implementation details to test the component. Since `Layer` is now a functional component, we can no longer use `ref` to access the instance and update state manually in our test or call `useEffect`. Any ideas about how to make that work or is this test even a valid test now?
[(This is the test I'm referring to)](https://github.com/grommet/grommet/blob/master/src/js/components/Layer/__tests__/Layer-test.js#L215)

#3394 

No

No

Backwards compatible


ShimiSun:
I think you need to fix the failing test, could you please take a look?
```
Summary of all failing tests
 FAIL  src/js/components/Layer/__tests__/Layer-test.js
  ● Layer › should be null prior to mounting, displayed after mount

    TypeError: Cannot read property 'setState' of null

      213 |     );
      214 | 
    > 215 |     ref.current.setState({ islayerContainerAvailable: false });
          |                 ^
      216 |     expect(queryByTestId(document, 'test-layer-container')).toBeNull();
      217 | 
      218 |     ref.current.componentDidMount();

      at Object.setState (src/js/components/Layer/__tests__/Layer-test.js:215:17)


Test Suites: 1 failed, 51 passed, 52 total
Tests:       1 failed, 446 passed, 447 total
Snapshots:   556 passed, 556 total
Time:        37.295s
```

src/js/components/Layer/Layer.js
```diff
@@ -1,76 +1,78 @@
...
-      : null;
-  }
-}
+Layer.defaultProps = {
+  full: false,
+  margin: 'none',
+  modal: true,
+  position: 'center',
+  responsive: true,
+};
```
ShimiSun: Can we please move the default props assignment to be part of the functional component declaration (line #9)?

src/js/components/Layer/LayerContainer.js
```diff
@@ -17,137 +17,120 @@ const HiddenAnchor = styled.a`
...
+  }
+  return content;
+};
+
+LayerContainer.defaultProps = {
+  full: false,
+  margin: 'none',
+  modal: true,
+  position: 'center',
+};
```
ShimiSun: Same comment here to move the assignment to be part of the declaration.

husseyexplores:
> #### Any background context you want to provide?
> Test `should be null prior to mounting, displayed after mount` is failing because it uses implementation details to test the component. Since `Layer` is now a functional component, we can no longer use `ref` to access the instance and update state manually in our test or call `useEffect`. Any ideas about how to make that work or is this test even a valid test now?

src/js/components/Layer/Layer.js
husseyexplores: If we define all the `defaultProps` on line #9, in the function parameters, then we would also need to pass every single prop into the `LayerContainer` component manually. (line #65)
That feels a bit awkward to me.
This is the only reason I didn't destructure the two props needed on line #9 but did it on line #10 so we could just pass along the props as-is to the `LayerContainer` component.

If being explicit is a requirement, I have no problem doing it the way you recommend. I just expressed my reasoning.

src/js/components/Layer/LayerContainer.js
husseyexplores: Made the change.

ShimiSun:
I need to find time to dive into this, unfortunately, I don't have a straight answer.

ShimiSun:
Thanks for contributing! I'll be closing this PR since refactoring Layer to use hooks was taken care of in other efforts.

PR #3396, opened 2019-10-02
Refactored TextInput story in FormField using Hooks

thidasapankaja:
#### What does this PR do?
Refactor `src/js/components/FormField/stories/TextInput.js` using Hooks

#### Where should the reviewer start?
`src/js/components/FormField/stories/TextInput.js`

#### What testing has been done on this PR?
Running `yarn test`

#### How should this be manually tested?
Running `yarn test`

#### Any background context you want to provide?
N/A

#### What are the relevant issues?
https://github.com/grommet/grommet/issues/3394

#### Screenshots (if appropriate)
N/A

#### Do the grommet docs need to be updated?
No

#### Should this PR be mentioned in the release notes?
No

#### Is this change backwards compatible or is it a breaking change?
Backwards compatible

Issue #119, opened 2015-08-04
Distribution not covering text


afong94:
For a distribution with 1 item, the box isn't filled with color and only shows a small bar.
![capture](https://cloud.githubusercontent.com/assets/5715687/9065938/a4c01ff8-3a89-11e5-9948-1dba829c358e.PNG)

ericsoderberghp:
This has been fixed in commit 49a36620b7ef515392d5422790aae7b29f81c005.

PR #3462, opened 2019-10-13
Refactoring Chart to use hooks

terryBaz:
Signed-off-by: terryBaz <et_bazine@esi.dz>

#### What does this PR do?

- Refactoring Chart to use hooks

- fix state in TextInput story (selecting a suggestion)

`src/js/components/Chart/Chart.js`
`src/js/components/FormField/stories/TextInput.js`

/
/
/
https://github.com/grommet/grommet/issues/3394
#3396
/
/
/
/


src/js/components/Chart/Chart.js
```diff
@@ -111,125 +116,138 @@ const renderArea = (
...
+    overflow,
+    round,
+    size,
+    theme,
+    thickness,
+    type,
+    values,
+    bounds,
+    ...rest
+  } = props;
```
ShimiSun: All of these could be part of the deconstructing props on line #119
Something like 
`const Chart = ({ color, onClick ...}) => { ...`

src/js/components/Chart/Chart.js
```diff
@@ -111,125 +116,138 @@ const renderArea = (
...
 
-  containerRef = createRef();
+  const containerRef = useRef(null);
```
ShimiSun:
```suggestion
  const containerRef = useRef();
```
Could you use a cleaner version?

src/js/components/Chart/Chart.js
```diff
@@ -111,125 +116,138 @@ const renderArea = (
...
+  );
+};
+
+Chart.defaultProps = {
+  color: 'accent-1',
+  overflow: false,
+  size: { width: 'medium', height: 'small' },
+  thickness: 'medium',
+  type: 'bar',
+};
```
ShimiSun: Can we initialize the variables part of deconstructing the props? so we will be more consistent with the behavior of other components.

src/js/components/Chart/Chart.js
```diff
@@ -7,7 +7,12 @@ import { normalizeColor, parseMetricToNum } from '../../utils';
 import { defaultProps } from '../../default-props';
 
 import { StyledChart } from './StyledChart';
-import { normalizeValues, normalizeBounds } from './utils';
+import {
+  normalizeValues,
+  normalizeBounds,
+  areNormalizedValuesEquals,
+  areNormalizedBoundsEquals,
```
ShimiSun: Can we please use alphabetical order on the imports?

src/js/components/Chart/Chart.js
```diff
@@ -111,125 +116,138 @@ const renderArea = (
...
+  const [state, setState] = useState({
+    containerWidth: 0,
+    containerHeight: 0,
+  });
+  const {
+    containerWidth,
+    containerHeight,
+    values: valuesState,
+    bounds: boundsState,
+  } = state;
```
ShimiSun: I'm inclining to separate the one variable `state` to multiple states, thoughts?

ShimiSun:
Nice work buddy! Dropped a few comments.

src/js/components/Chart/Chart.js
terryBaz: I did not do that because `props` is been used in line 217 and 219

src/js/components/Chart/Chart.js
terryBaz: maybe three `containerState, valuesState, boundsState`

terryBaz:
Hello, waiting for your reply regarding my comments

src/js/components/Chart/Chart.js
terryBaz: @ShimiSun

src/js/components/Chart/Chart.js
ShimiSun: sure

ShimiSun:
@terryBaz thanks for reaching out, I resolved some of the threads, and left those I think still relevant opened. let me know if that works.

src/js/components/Chart/Chart.js
```diff
@@ -111,125 +116,138 @@ const renderArea = (
...
-  onResize = () => {
-    const containerNode = this.containerRef.current;
+  const onResize = () => {
+    const containerNode = containerRef.current;
     if (containerNode) {
       const { parentNode } = containerNode;
       if (parentNode) {
         const rect = parentNode.getBoundingClientRect();
-        this.setState({
+        setState({
```
husseyexplores: Can we do this instead?
```js
setState(prevState => ({
  ...prevState,
  containerWidth: rect.width,
  containerHeight: rect.height,
}));
```

Otherwise, the spreaded state (`...state`) will always be the initial state values, not the latest.

This is kinda hard to explain for me. But I'll try.

`onResize` will be redefined every time the component is rendered.
On the first render, we define `onResize`, then our `useEffect(() => { ... },[])` runs and attaches the listener to the `window`.
Then our state/props change and the component rerenders. We define our `onResize` again, but we never use it, since it is not listed in our `useEffect` dependencies array.

So, whenever our `onResize` runs later on, it has stale `state` values.
So the simple fix is to use the functional of `setState` here.
If we were using state or props values to conditionally update the state, then it'd become a bit messier.

A true refactor according to [this detailed article](https://overreacted.io/a-complete-guide-to-useeffect) would be
```js
const onResize = useCallback(() => { ... }, []);
useEffect(() => { /* attach listener here */ }, [onResize]); 
```
Now if we every use state/props in `onResize` function, we would add it as the dependency in `useCallback` and our effect will rerun.

I think there should be `eslint-plugin-react-hooks` in the library. `exhaustive-deps` rule will help to catch unnoticed bugs like these.

terryBaz:
@ShimiSun @ericsoderberghp any updates

ShimiSun:
Thank you for your contribution buddy! 🎉

Issue #118, opened 2015-08-03
Format issue, seen in Firefox version 39


tran2519:
![image](https://cloud.githubusercontent.com/assets/13039669/9047304/dfc248fa-39e5-11e5-8a06-f637d4fe26a7.png)

tran2519:
The format issue is also seen in Chrome and IE11

alansouzati:
I believe this should be fixed, I was not able to reproduce in the first place, but I guess I know what is going on. Could you please check it again? Thanks for reporting this.

tran2519:
It's fixed.  Thanks.

PR #3464, opened 2019-10-14
Restore onClick Menu functionality

taysea:
#### What does this PR do?
This PR fixes the cross browser onClick functionality that was removed in the previous update to Menu. The mouse can be used to open and select menu items. For keyboard accessibility, space/enter/down/up can be used to open the menu and tab can be used to navigate into and out of the menu. If the user tabs out of the menu, the menu will close. These functions work on Chrome, Safari, FireFox, Edge, and IE11.

These changes have been tested by creating an updated test using react testing library. They have also been tested manually by adding a console statement when the onClick handler of a menu item is called. This console statement printed out which item was clicked to ensure the correct item was being registered. This was tested using both mouse and keyboard.

To manually test, you can add this line `console.log(item)` to line 232 of Menu.js and then interact with the Menu as noted above.

In the process of trying to improve keyboard accessibility on the Menu component, the onClick functionality was affected. This PR fixes the regression while also maintaining some of the improved accessibility. Future steps will need to be taken to improve accessibility for arrow keys.

Issue #3408 

No.

No.

Yes, backwards compatible.


src/js/components/Menu/Menu.js
```diff
@@ -70,55 +70,45 @@ const Menu = props => {
   const iconColor = normalizeColor('control', theme);
   const align = dropProps.align || dropAlign;
   const buttonRefs = {};
+  const NONE = 'none';
```
ShimiSun:
```suggestion
  const constants = {none: 'none'};
```
I'll be doing more refactoring around this area of constant strings, but for now, I think we can use this structure, so in this case, the constant will be approached as `constants.none`. If you have more constants, put them on there as well.

src/js/components/Menu/Menu.js
```diff
@@ -70,55 +70,45 @@ const Menu = props => {
...
-          buttonRefs[items.length].focus();
-          setActiveItemIndex(-1);
-        }
-      }
-    }
-  }, [isManagingFocus]);
 
   const onDropClose = () => {
-    setActiveItemIndex(-1);
+    setActiveItemIndex(NONE);
```
ShimiSun: why do we need this line at all now? If the Menu Button itself shouldn't have hoverindication than this line may be redundant.

src/js/components/Menu/__tests__/Menu-test.js
```diff
@@ -140,39 +140,36 @@ describe('Menu', () => {
...
-      which: 40,
+      key: 'Space',
+      keyCode: 32,
+      which: 32,
     });
-    fireEvent.keyDown(getByLabelText('Open Menu'), {
-      key: 'Down',
-      keyCode: 40,
-      which: 40,
+    fireEvent.keyDown(getByLabelText('Close Menu'), {
```
ShimiSun:
```suggestion
    fireEvent.keyDown(getByLabelText('Move to Menu label'), {
```

src/js/components/Menu/__tests__/Menu-test.js
```diff
@@ -140,39 +140,36 @@ describe('Menu', () => {
...
+      key: 'Tab',
+      keyCode: 9,
+      which: 9,
     });
-    fireEvent.keyDown(getByLabelText('Open Menu'), {
-      key: 'Down',
-      keyCode: 40,
-      which: 40,
+    fireEvent.keyDown(getByLabelText('Close Menu'), {
```
ShimiSun:
```suggestion
    fireEvent.keyDown(getByLabelText('Move to first Menu item'), {
```

src/js/components/Menu/__tests__/Menu-test.js
```diff
@@ -140,39 +140,36 @@ describe('Menu', () => {
...
+      keyCode: 9,
+      which: 9,
     });
-
-    // moves to the first suggestion
-    fireEvent.keyDown(getByLabelText('Open Menu'), {
-      key: 'Up',
-      keyCode: 38,
-      which: 38,
+    fireEvent.keyDown(getByLabelText('Close Menu'), {
```
ShimiSun:
```suggestion
    fireEvent.keyDown(getByLabelText('Move to second Menu item'), {
```

src/js/components/Menu/__tests__/Menu-test.js
```diff
@@ -140,39 +140,36 @@ describe('Menu', () => {
...
+      key: 'Tab',
+      keyCode: 9,
+      which: 9,
```
ShimiSun: I think we should add one more fireevent block to close the drop once all items were tabbed.

src/js/components/Menu/Menu.js
taysea: Good question. In this component, the active item index is changed based on what item has focus (not hover). So, it can be affected when the user uses keyboard to move through menu items. Hover indication doesn't actually change the active item index. What this line does is ensures that the active item index returns to its default value when the menu closes.

src/js/components/Menu/__tests__/__snapshots__/Menu-test.js.snap
```diff
@@ -605,6 +602,7 @@ exports[`Menu close by clicking outside 2`] = `
       <button
         aria-label="Close Menu"
         class="c5"
+        tabindex="-1"
```
taysea: The addition of these lines in the snapshots refers to the change in line 209 of Menu.js. Essentially, when the menu first opens, the user's tab should go to the first menu item not the control button. What this means is at first, the control button receives a tabindex of -1 to remove it from the tab order (otherwise it naturally receives focus because it's a button and if it is on top of the menu, it will take the focus first). However, once the user has tabbed once into the menu, the control button tabindex of -1 is removed.

src/js/components/Menu/Menu.js
```diff
@@ -69,57 +69,72 @@ const Menu = props => {
   const MenuIcon = theme.menu.icons.down;
   const iconColor = normalizeColor('control', theme);
   const align = dropProps.align || dropAlign;
+  let controlButtonIndex;
+  if (align.top === 'top') {
+    controlButtonIndex = -1;
+  } else if (align.bottom === 'bottom') {
```
ShimiSun: Can we try to use the constants here?

src/js/components/Menu/Menu.js
```diff
@@ -69,57 +69,72 @@ const Menu = props => {
   const MenuIcon = theme.menu.icons.down;
   const iconColor = normalizeColor('control', theme);
   const align = dropProps.align || dropAlign;
+  let controlButtonIndex;
+  if (align.top === 'top') {
```
ShimiSun: Can we try to use the constants here?

src/js/components/Menu/Menu.js
```diff
@@ -69,57 +69,72 @@ const Menu = props => {
...
 
   const onNextMenuItem = event => {
     event.preventDefault();
     if (!isOpen) {
       onDropOpen();
+    } else if (
+      (event.keyCode === constants.tab || event.which === constants.tab) &&
+      ((!constants.menuControlIncludedBottom &&
+        activeItemIndex === items.length - 1) ||
+        (constants.menuControlIncludedBottom &&
```
ShimiSun: Can we think of a shorter form for this name? maybe bottomMenu? we can add a more elaborating comment where it is defined.

src/js/components/Menu/__tests__/Menu-test.js
```diff
@@ -140,40 +139,32 @@ describe('Menu', () => {
...
     });
-
-    // moves to the first suggestion
-    fireEvent.keyDown(getByLabelText('Open Menu'), {
-      key: 'Up',
-      keyCode: 38,
-      which: 38,
+    fireEvent.keyDown(document.activeElement, {
```
ShimiSun: Nice refinement! glad you figured it out!

src/js/components/Menu/__tests__/__snapshots__/Menu-test.js.snap
ShimiSun: Thank you for anticipating my next question...

src/js/components/Menu/Menu.js
```diff
@@ -69,57 +69,70 @@ const Menu = props => {
...
+  } else {
+    controlButtonIndex = undefined;
+  }
   const buttonRefs = {};
+  const constants = {
+    none: 'none',
+    tab: 9,
+    // Menu control button included on top of menu items
+    controlTop: align.top === 'top' || undefined,
+    // Menu control button included on bottom of menu items
```
taysea:
```suggestion
    // Menu control button included on the bottom of menu items
```

src/js/components/Menu/Menu.js
```diff
@@ -143,10 +149,12 @@ const Menu = props => {
 
   const onPreviousMenuItem = event => {
     event.preventDefault();
+    const isTab =
```
ShimiSun: Can we avoid the duplication of this line and line #118, and maybe use a function the accepts event?

taysea: I think that's a good idea

taysea: I think that's a good idea.

src/js/components/Menu/Menu.js
```diff
@@ -112,14 +112,16 @@ const Menu = props => {
     }
   };
 
+  const isTab = event => {
+    return event.keyCode === constants.tab || event.which === constants.tab;
+  };
```
ShimiSun: Could be replaced with a cleaner way
```
const isTab = event =>
    event.keyCode === constants.tab || event.which === constants.tab;
```

PR #3473, opened 2019-10-17
Update type definitions for Functional Components

taysea:
#### What does this PR do?
This PR updates the type definitions for components that have been refactored to hooks for Hacktober Fest. Before the changes were made, there were some components that had been refactored to hooks but their type definitions still said `React.ComponentClass`. This changes those to say `React.FC`.

Any of the index.d.ts files listed in files changed.

These changes have been tested in a local typescript project. While the components that had mismatched type definitions were not being flagged in the editor, some where not rendering onto the screen. Once their type definitions were updated, the components all work as expected.

In a local typescript project, use any of the components whose types have been updated in the PR. There should be no flags in the editor, and the components should render to the screen.

Some components had been refactored to hooks, but their type definitions had not been updated.

#3394 

No.

No.

Yes, it's backwards compatible.

### DIFF
--- a/src/js/components/CheckBox/index.d.ts
+++ b/src/js/components/CheckBox/index.d.ts
@@ -12,6 +12,6 @@ export interface CheckBoxProps {
   indeterminate?: boolean;
 }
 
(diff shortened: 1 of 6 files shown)


ShimiSun:
Awesome! thanks!

PR #3478, opened 2019-10-17
Add hover indicator to control button when menu opens

taysea:
#### What does this PR do?
This PR adds hover styling to the menu control button when it is included in the menu drop. Initially, this change was going to be included in a different Menu PR, but for clarity, I separated it into its own PR.

This has been testing in local storybook examples by opening the menu then hovering over the control button with the mouse to check that it receives the same hover indicator as the menu items.

By using the storybook examples where the control button is included in the drop (such as "Custom"), open the menu and hover over the control button. It should receive the same hover indicator that the menu items receive.

The control button was not receiving any styling when it was being hovered over even though it does receive styling when given keyboard focus. For clarity and consistency of styling when the control button is included within the menu drop, it should receive a hover indicator when the mouse hovers over.

No.

No.

This introduces new styling on the control button when hovered, but it does not affect any of the functionality of the button.


src/js/components/Menu/Menu.js
```diff
@@ -204,6 +204,7 @@ const Menu = props => {
         a11yTitle={messages.closeMenu || 'Close Menu'}
         active={activeItemIndex === controlButtonIndex}
         focusIndicator={false}
+        hoverIndicator="background"
```
ShimiSun: what happens if the user don't want this hover behavior? shouldn't we allow it to be customized instead of forcing it at this point?

ShimiSun: oh wait, gotcha, it is only for the Button which is currently missing the hoverIndicator behavior.

taysea: Yes, just the control button when it is included in the drop with the menu items, so the hoverIndicator here will match the hover indicator that appears on the other menu items as well. I think it's important to consider though, would there be a case where someone wouldn't want a hover on this button? I think from a consistency in behavior across the drop, @ericsoderberghp and I felt that it should be receiving hover indication but I'm open to discussing it further.

ShimiSun:

PR #3479, opened 2019-10-18
Added test for when FormField receives prop of pad

taysea:
#### What does this PR do?
This adds a test for when FormField is provided with prop `pad`.

src/js/components/FormField/__tests__/FormField-test.js

I looked into the styling created by the snapshot to check that it matches that of the storybook example with a FormField.

The FormField had tests that checked margin and other properties were rendered correctly, but there was not one for pad.

No.

No.


ShimiSun:
Thank you for your contribution 🎉

Issue #3237, opened 2019-08-01
Typescript definitions - extract types


atanasster:
The types for common properties like `align`, `gap`, `margin` etc. need to be extracted in a core types file. The current typescript implementation leads to code duplications, both in grommet itself and in the applications using grommet.
https://github.com/grommet/grommet/blob/9a2663082bb21bb5d4a947cedfdc769451c72522/src/js/components/Box/index.d.ts#L7

for example 
`export type GapProp = "xxsmall" | "xsmall" | "small" | "medium" | "large" | "xlarge" | string;`

`
export interface BoxProps {
...
  gap?: GapProp,
...
}
`

ShimiSun:
@britt6612 / @atanasster any other types that needs to be extracted?

britt6612:
@ShimiSun there will be more especially once we start handling all of the event handlers for issue #3165

ShimiSun:
Any updates on this issue?

taysea:
I opened one more PR (see above), but after that I think we can close this.

ShimiSun:
Thank you @taysea & @britt6612 for working on this issue! 
I'm closing the issue but let's keep our eyes open to keep optimizing the typescript coding.

PR #3490, opened 2019-10-22
TS Extraction: justifyContent & Elevation

taysea:
#### What does this PR do?
This PR extracts `elevation` and `justifyContent` into variables that can be used across various components.

src/js/utils/index.d.ts

These changes have been tested in a local TS project on the components that have these properties to ensure it only accepts the types defined in the utils/index.d.ts file.

In a local TS project, use one of the components that has these props and put a value that matches the prop type (no flag should appear and the app should render) and then provide a value that doesn't match the prop type (should cause an error).

#3237 

No.

No.

Yes, backwards compatible.

### DIFF
--- a/src/js/components/Box/index.d.ts
+++ b/src/js/components/Box/index.d.ts
@@ -4,6 +4,7 @@ import {
   AlignContentType, 
   AlignSelfType,
   BackgroundType,
+  ElevationType,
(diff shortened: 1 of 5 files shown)


ShimiSun:
Thank you @taysea ! 🎉

PR #3491, opened 2019-10-23
Add eslint-plugin-react-hooks

husseyexplores:
Signed-off-by: husseyexplores <husseyexplores@gmail.com>

#### What does this PR do?
Adds `eslint-plugin-react-hooks` into the project dev dependencies and adds the rule in `.eslintrc` file.

`.eslintrc`
`package.json`

\-

\-

Since the codebase is now shifting to use react hooks, I thought this rule should be added to avoid the bugs related to hooks. I shared my thoughts with @ShimiSun which she approved.

\-

\-

No

No

Backwards compatible


.eslintrc
```diff
@@ -45,7 +45,9 @@
     "react/state-in-constructor": 0,
     "react/static-property-placement": 0,
     "react/jsx-curly-newline": 0,
-    "max-len": [2, {"ignoreUrls": true, "ignoreRegExpLiterals": true}]
+    "max-len": [2, {"ignoreUrls": true, "ignoreRegExpLiterals": true}],
+    "react-hooks/rules-of-hooks": 2,
+    "react-hooks/exhaustive-deps": 1
```
ShimiSun: why do you think `react-hooks/rules-of-hooks` should be an error rule while the other rule of `react-hooks/exhaustive-deps` is a warning rule?

husseyexplores: I just followed the official docs example. And I kinda agree with their explanation. Erroring out on 'rules-of-hooks' will prevent any wrong usage of hooks like using in regular functions instead of React Components. Or calling `useEffect, useState` (or any other react hook) conditionally.
Reference: https://reactjs.org/docs/hooks-rules.html

Let me what you think since I definitely not have any strong opinion here.

ShimiSun:

PR #3499, opened 2019-10-28
Changed Keyboard to fix an issue with how hooks are used

ericsoderberghp:
#### What does this PR do?

Changed Keyboard to fix an issue with how hooks are used.

The problem was when a caller supplied a new `onKeyDown`, the effects weren't using the new `onKeyDownHandler`.

Keyboard.js

grommet-designer

no

yes

backwards compatible


ShimiSun:
Can we please add a test to cover this scenario?

PR #3500, opened 2019-10-28
Adjust onSearch type definition

taysea:
#### What does this PR do?
This PR adjusts the onSearch type definition for Select and DataTable to be: `(search: string) => void`

src/js/components/Select/index.d.ts or src/js/components/DataTable/index.d.ts

This PR has been tested on a local TS project where the storybook example of Select Search was copied into my local TS project. The same was done for DataTable with searchable columns. There were no flags in the editor, and the application worked as expected. 

In a local TS project, copy and paste the code from the Select Search storybook example and make sure you are using a grommet build that contains this TS change. Do the same for the Tunable DataTable. There should be no flags in your code and the application should compile successfully.

I found the information about what was being accepted as a parameter for onSearch in the SelectContainer.js file:
```
onSearch = debounce(search => {
    const { onSearch } = this.props;
    onSearch(search);
  }, debounceDelay(this.props));
```
I found for DataTable:
```
if (onSearch) {
      onSearch(nextFilters);
    }
```
I could see that it was accepting a search term variable. When I printed the `search` or `nextFilters` to the console, I saw that it was the string containing the search terms.

#3165 

No.

No.


src/js/components/DataTable/index.d.ts
```diff
@@ -12,7 +12,7 @@ export interface DataTableProps {
   data?: {}[];
   groupBy?: string | { property: string, expand: Array<string>, onExpand: ((...args: any[]) => any) };
   onMore?: ((...args: any[]) => any);
-  onSearch?: ((...args: any[]) => any);
+  onSearch?: ((search: string) => void);
```
ShimiSun: I noticed onSearch work properly with DataTable as is, why do we nee this fix? to eliminate the 'any'?

taysea: The reason to change it is because when the type is `any`, then it is not placing any restriction on what types can be accepted. This won't cause any error, but it also won't help prevent errors for people using Typescript. From the perspective of Typescript, having `any` is like having no type enforcement at because everything will always pass. By changing it, it will make sure that onSearch is only receiving a string.

ShimiSun: Thank you and of course. I just wanted to make sure the 'any' is the motivation behind the change rather than ts failure that drove this change.

ShimiSun:
Thank you!! 💟

PR #3501, opened 2019-10-28
Fix react hooks flags

taysea:
#### What does this PR do?
This PR contains changes that react hooks eslint plugin caught after it was added to package.json. I am placing these in a separate PR so they are not included in other issues I'm working on.

src/js/components/Keyboard/Keyboard.js

After accepting the suggestions from the hooks eslint plugin, I ran storybook and made sure the stories/components affected functioned that same as before.

Run storybook and engage with the changed stories.

A new plugin was added to the package.json, so it flagged some hooks errors that weren't previously caught. The changes essentially just affect dependency arrays.

N/A

No.

No.


ShimiSun:
Hi, I think the Keyboard issue is fixed, care to resolve the conflicts and only keep relevant files?

PR #3505, opened 2019-10-29
Adjust onReference TS definition

taysea:
#### What does this PR do?
This PR adjusts the type definition for onReference to be `(reference: string) => void`.

src/js/components/Calendar/index.d.ts

This has been tested in a local TS project where a Calendar storybook example (Dual) was copied into my local project to check that the onReference handler had no flags. It did not, and that app compiled successfully.

Copy a storybook example containing onReference into a local TS project and make sure there are no errors in your editor.

This line in Calendar.js shows how onReference is used. It accepts a reference which is a string, and converts it to an ISO string.
```
if (onReference) {
          onReference(reference.toISOString());
        }
```
 I used this to know how to create the type definition.

#3165 

No.

No.


taysea:
Hey Shimi, similar to the onActive TS changes I made, there was no error caused by having `any`. These changes simply enforce a specific type for onReference. With `any` as the type, this is like having no type enforcement because everything will be accepted. My changes specify the type declaration. To confirm, when you use my type definition, you still are not seeing any error right?

ShimiSun:
Thank you for all the refinements! 💯

PR #3506, opened 2019-10-29
Adjust onActive TS definitions

taysea:
#### What does this PR do?
This PR adjusts on the onActive TS definitions for Accordion and Tabs.

src/js/components/Accordion/index.d.ts

These changes have been tested on a local TS project where storybook examples were copy and pasted to confirm that there were no errors caused by the changes. The examples work the same on a typescript project as they do on a non-typescript project.

Test these changes on a local TS project by copy & pasting existing storybook examples that use the onActive handler or creating your own.

For Accordion.js, the following code informed me of how to define the nextActiveIndexes as a number[]:
```
 const [activeIndexes, setActiveIndexes] = useState([]);
...
let nextActiveIndexes = [...(activeIndexes || [])];

if (onActive) {
      onActive(nextActiveIndexes);
}
```
For Tabs.js, the following code informed me of how to define index as a number:
```
const [activeIndex, setActiveIndex] = useState(rest.activeIndex || 0);
...
if (onActive) {
      onActive(index);
}
```
#3165 

No.

No.


src/js/components/Accordion/stories/typescript/CustomHeader.tsx
```diff
@@ -0,0 +1,68 @@
...
+          </Box>
+        </AccordionPanel>
+      </Accordion>
+    </Grommet>
+  );
+};
+
+storiesOf('Accordion', module).add('Custom Header', () => (
+  <CustomHeaderAccordion />
+));
```
ShimiSun: I think we can remove this part of the js code

src/js/components/Accordion/stories/typescript/CustomHeader.tsx
```diff
@@ -0,0 +1,68 @@
...
+);
+
+const CustomHeaderAccordion = () => {
+  const [activeIndex, setActiveIndex] = useState([0]);
+
+  return (
+    <Grommet theme={grommet}>
+      <Accordion
+        activeIndex={activeIndex}
+        onActive={newActiveIndex => setActiveIndex(newActiveIndex)}
```
ShimiSun:
```suggestion
        onActive={(newActiveIndex: number) => setActiveIndex(newActiveIndex)}
```
should we be more typescript oriented and add the type?

src/js/components/Tabs/stories/typescript/Controlled.tsx
```diff
@@ -0,0 +1,42 @@
...
+          <Box margin="small" pad="large" align="center" background="accent-3">
+            <Car size="xlarge" />
+          </Box>
+        </Tab>
+      </Tabs>
+    </Grommet>
+  );
+};
+
+storiesOf('Tabs', module).add('Controlled', () => <ControlledTabs />);
```
ShimiSun: we can remove the js story reference.

src/js/components/Accordion/stories/typescript/CustomHeader.tsx
taysea: Hey Shimi, my inclination here is that we don't need to redefine the type because this type has already been defined in the index.d.ts, so in the code we wouldn't have to set it again. Because it's inline, if newActiveIndex was coming in as a string, for example, it would flag it. It seems like this is similar to this example on the typescript docs:
Screen Shot 2019-11-01 at 10 26 14 AM
They define the types in their interface, then in the code they don't redefine them. However, in the case that the handler was defined in a separate line of code, then the editor will flag you to specify the type. For example, if we have it separate like this then it will ask us to specify the type:
```
const onActive = (newActiveIndex: number[]) => { setActiveIndex(newActiveIndex) }

<Accordion onActive={onActive}>
...
```
When it is inline, though, there is no flag, so it seems like it is able to infer the type from the index.d.ts file.

src/js/components/Accordion/index.d.ts
```diff
@@ -9,7 +9,7 @@ export interface AccordionProps {
   activeIndex?: number | number[];
   animate?: AnimateType;
   children?: React.ReactNode;
-  onActive?: ((...args: any[]) => any);
+  onActive?: ((nextActiveIndexes: number[]) => void);
```
ShimiSun:
```suggestion
  onActive?: ((activeIndex: number[]) => void);
```

Issue #3452, opened 2019-10-10
Refactor color namespace


ericsoderberghp:
This is a proposal for how to refactor the grommet color namespace.

There are three motivations for this:
1) "accent" and "neutral" are intended to be generic but many themes don't naturally think in these terms.
2) a desire to have variations on non-grey colors, such as more or less saturation.
3) a desire to have dark and light variations of a theme built into a single theme.

Add a `dark` property to the `Grommet` component, allowing the entire app to be toggled between dark and light modes from the top.

We should allow both wide and narrow colorspaces, where the names have meaning primarily within the theme. For example, maybe a theme has just a single brand color, or a brand color with a single secondary color, or maybe a brand color with three shade variations on it (like Target).

Alias color names for standard contexts, dark+light variants: border, control, focus, hover, active

global.colors {
    // base colors, single color value or light/dark object
    brand,
    background,
    text,
    // greyscale variations on background and text colors, single value or light/dark object
    background-strong,
    background-weak,
    background-xweak,
    text-strong,
    text-weak,
    text-xweak,
    // broad context, typically the name of an existing color
    border: ‘background-xweak’, // default border color
    control: ‘brand',
    focus: ‘brand',
    hover: {
        background: ‘background-weak’,
        text: ’text-strong’,
    },
    active: {
        background: ‘background-strong’,
        text: ’text-strong’,
    },
    // the following are used for Meter, Chart, Distribution, etc.
    graphic: ‘brand’,
    graphic-1: ‘graphic',
    graphic-2: ’text-strong’,
    graphic-3: ’text-weak’,
    // theme specific color names, light and dark tints or single color value
    // these can be used in the above context color names as desired
    blue: { light: ‘0x00F’, dark: ‘0x00C'  },
    red: ‘0xF00’,
    // per-component coloring is done within the component objects in the theme
}

Existing color names, like `accent-1`, in grommet’s base theme would be preserved for backwards compatibility.

Feedback appreciated.

ShimiSun:
1. Linking to #2940.

2. I think below is a great idea.
     >Add a dark property to the Grommet component, allowing the entire app to be toggled between 
     dark and light modes from the top.

3. 
```
// greyscale variations on background and text colors, a single value or light/dark object
background-strong,
background-weak,
background-xweak,
```
For these cases, the user would be able to set it up himself, but I think it will be interesting if we can also provide those default colors out of the box according to `background`.

ericsoderberghp:
[...]

PR #3507, opened 2019-10-29
Changed Grommet to add themeMode

ericsoderberghp:
#### What does this PR do?

Changed Grommet to add themeMode

Grommet/doc.js

Storybook all components

Storybook all components

automatic?

yes

backwards compatible


ShimiSun:
https://github.com/grommet/grommet/issues/3452

src/js/components/Grommet/doc.js
```diff
@@ -28,6 +28,10 @@ export const doc = Grommet => {
     theme: PropTypes.object.description(
       'Custom styles for Grommet app component.',
     ),
+    themeMode: PropTypes.oneOf(['dark', 'light']).description(
+      `Dark vs. light theme variation. Default is unspecified and left to
+      theme.`,
+    ),
```
ShimiSun: Few thoughts here.
Is there a way to set the default `themeMode` from the theme?
How will this behave if a theme isn't supporting `dark` or `light`?

ericsoderberghp: Good questions. If a theme doesn't support dark/light, `themeMode` has no effect. I've updated how the default `theme.dark` is calculated to be clearer as well.

PR #3508, opened 2019-10-29
Adjust onClick TS definition for Anchor

taysea:
#### What does this PR do?
This PR adjusts the type definition for Anchor onClick type definition to `onClick?: ((event: React.MouseEvent<HTMLAnchorElement>) => void);`

src/js/components/Anchor/index.d.ts

This has been tested on a local TS project with the following:
```
<Anchor icon={<Add />} href="https://v2.grommet.io/components" target="blank" onClick={(event) => console.log(event)}/>
```

With the above code in a local TS project.

This type definition matches that of the intrinsic elements of anchor, but if we don't omit `onClick` from the intrinsic elements, then in a TS project, the type will appear duplicated which is not wanted:
```
onClick?: ((event: React.MouseEvent<HTMLAnchorElement>) => void) & (event: React.MouseEvent<HTMLAnchorElement>) => void) | undefined
```
To maintain clarity about what properties the Anchor component handles, I left our onClick handler in the interface and omitted the intrinsic property so that the declaration only appears once:
```
onClick?: ((event: React.MouseEvent<HTMLAnchorElement>) => void) | undefined
```

#3165 

No.

No.

### DIFF
--- a/src/js/components/Anchor/index.d.ts
+++ b/src/js/components/Anchor/index.d.ts
@@ -19,11 +19,12 @@ export interface AnchorProps {
   icon?: JSX.Element;
   label?: React.ReactNode;
   margin?: MarginType;
+  onClick?: ((event: React.MouseEvent<HTMLAnchorElement>) => void);
   reverse?: boolean;
   size?: "xsmall" | "small" | "medium" | "large" | "xlarge" | "xxlarge" | string;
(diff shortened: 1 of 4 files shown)


ShimiSun:
Following our discussion, I think we can remove the `onClick` from the index.d.ts as long as it has the same typing as the instinct element.

PR #3516, opened 2019-10-31
Remove onClick and leave intrinsic for Anchor

taysea:
#### What does this PR do?
This PR removes the type definition for Anchor onClick type definition and just leaves the definition from the intrinsic elements since we are not adding any new/different behavior.

src/js/components/Anchor/index.d.ts

This has been tested on a local TS project with the following:
```
      <Anchor icon={<Add />} href="https://v2.grommet.io/components" target="blank" onClick={(event) => console.log(event)}/>
```

With the above code in a local TS project.

We are not adding any new/different behavior to the onClick handler from what the intrinsic element already provides, so we should remove it.

#3165 

No.

No.

### DIFF
--- a/src/js/components/Anchor/index.d.ts
+++ b/src/js/components/Anchor/index.d.ts
@@ -19,7 +19,6 @@ export interface AnchorProps {
   icon?: JSX.Element;
   label?: React.ReactNode;
   margin?: MarginType;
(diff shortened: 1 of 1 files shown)


ShimiSun:
Thank you!! 💟

Issue #3481, opened 2019-10-19
ts: Select on change event error


atanasster:
The onChange event for `<Select />` does not correspond to the actual use of the type
ie: 
https://github.com/grommet/grommet/blob/d8f7eafcccaa38b79d810d36f8f9734bf0636cd4/src/js/components/Select/index.d.ts#L25

while: https://github.com/grommet/grommet/blob/d8f7eafcccaa38b79d810d36f8f9734bf0636cd4/src/js/components/Select/SelectContainer.js#L163 
has `value` and `option`

example to reproduce: any of the storybook examples convert to typescript. ie
```
<Select
  options={options}
  onChange={({ option }) => setValue(option)}
/>
```

will result in error 
`Property 'option' does not exist on type 'ChangeEvent<HTMLSelectElement>'.`

taysea:
I've been looking into this and found that onChange has these items:
```
onChange({
        option,
        value: nextValue,
        selected: nextSelected,
})
```
I am thinking that the type definition should be something like:
```
  onChange?: ((event: { option, selected, value }) => void);
```
However, the piece I am working on is how to declare types for `option`, `selected`, and `value`. If I give each of these the type that is defined in the Select interface (for example: `option: string | boolean | number | JSX.Element | object;`), but in a user's application the Select options are strings, then something like this will have the following error:
```
<Select
          id="select"
          name="select"
          placeholder="Select"
          value={value}
          options={options}
          onChange={({ option } ) => setValue(option)}
        />
```
Error: `Argument of type 'string | number | boolean | object | Element' is not assignable to parameter of type 'SetStateAction<string>'.
  Type 'number' is not assignable to type 'SetStateAction<string>'`

However, if we don't declare any types on option, selected, or values, then it seems to defeat the purpose of declaring types. @britt6612 any thoughts?

ryanjchr:
I've had success with this by re-declaring Select and SelectProps as generic types as follows:
```tsx
/** /src/@types/grommet.d.ts */
import React from 'react';
import { DropProps } from 'grommet';
import {
  A11yTitleType,
  AlignSelfType,
  GridAreaType,
  MarginType,
  PlaceHolderType,
} from 'grommet/utils';

declare module 'grommet' {
  interface SelectProps<T extends string | boolean | number | JSX.Element | object> {
    // ... all the other SelectProps omitted for clarity
    disabledKey?: string | ((value: T) => any);
    labelKey?: string | ((value: T) => any);
[...]

PR #3519, opened 2019-10-31
Revert onChange for Select and update onChange for intrinsic components

taysea:
#### What does this PR do?
This PR reverts changes to the onChange type definitions back to `(...args: any[]) => any` for components that do not have an intrinsic onChange behavior that we are matching. For components with an onChange prop that we are not modifying, I have removed the onChange from our type definitions so the intrinsic definition is used.

src/js/components/CheckBox/index.d.ts

These changes have been tested on the allComponents TS storybook file. An onChange handler was added to each to observe that the types appeared from the intrinsic properties.

In a local TS project or in the TS storybook, add an onChange handler to the changed components.

Previous changes to the onChange type definitions was causing errors for some Typescript users. In the interest of allowing them to continue to develop their apps, this rolls back those changes. However, this will remain a focus to update the types again with the correct definitions.

#3481 

No.

No.


ShimiSun:
>An onChange handler was added to each to observe that the types appeared from the intrinsic properties.

Can you please add your testing to the allComponents.tsx file? it can be part of our regression testings.

taysea:
Yes, and @britt6612 and I are just double checking some of the intrinsic properties to make sure Grommet has no use cases that differ. Let's keep this PR on hold through tomorrow until Brittany and I can confirm.

taysea:
@ShimiSun I've added `onChange` handlers to the TSX all components files. Also, I added MaskedInput to that file because it was not there before.

ShimiSun:
Thank you @taysea let's keep making tsx files awesome! 💟

PR #3520, opened 2019-11-01
Refactor Layer component

nolleto:
#### What does this PR do?

Refactor `Layer` component with Hooks and update the broken tests for this component.

`src/js/components/Layer/Layer.js`

I've run the storybook and check the `Layer` behavior and also run `yarn test`

`yarn test`

N/A

#3394

N/A

No

No

Backwards compatible


src/js/components/Layer/__tests__/Layer-test.js
```diff
@@ -203,20 +203,20 @@ describe('Layer', () => {
...
-    ref.current.setState({ islayerContainerAvailable: false });
-    expect(queryByTestId(document, 'test-layer-container')).toBeNull();
-
-    ref.current.componentDidMount();
     expect(queryByTestId(document, 'test-layer-container')).toMatchSnapshot();
+    expect(container.firstChild).not.toBeNull();
+
+    unmount();
+
+    expect(container.firstChild).toBeNull();
```
nolleto: Please, if you had a better way to test it just let me know 😄

src/js/components/Layer/Layer.js
```diff
@@ -1,78 +1,82 @@
...
-  render() {
-    const { islayerContainerAvailable } = this.state;
-
-    return islayerContainerAvailable
-      ? createPortal(<LayerContainer {...this.props} />, this.layerContainer)
-      : null;
-  }
-}
+    };
+  }, [animate, animation]);
```
ericsoderberghp: I think this means we will animate the Layer out and change the focus back if the user changes the `animate` or `animation` properties. While I don't think those are likely to change, something feels a bit odd about that behavior. It seems like we shouldn't touch the focus or animation unless the Layer is being unmounted. Would it be possible to leave the dependency array empty?

nolleto: Sure, I will do it.

ShimiSun:
hi buddy, so you mind taking a look at the tabIndex change on the test snapshot?

nolleto:
Apparently, jest snapshot is adding `tabindex` in theses two `input`s here. 🤔 
`src/js/components/SkipLinks/__tests__/SkipLink-test.js`
```js
const { container } = render(
  <Grommet>
    <SkipLinks id="skip-links">
      <SkipLink id="main" label="Main Content" />
      <SkipLink id="footer" label="Footer" />
    </SkipLinks>
    <div>
      <SkipLinkTarget id="main" />
      Main Content
      <input type="text" value="main content" onChange={() => {}} /> // <--- HERE
    </div>
    <footer>
      <SkipLinkTarget id="footer" />
      <input type="text" value="footer" onChange={() => {}} /> // <--- HERE
    </footer>
  </Grommet>,
);
```

So, I've updated the snapshot.

ShimiSun:
Thanks for contributing! I'll be closing this PR since refactoring Layer to use hooks was taken care of in other PR effort.

Issue #3212, opened 2019-07-23
Storybook structure 


ShimiSun:
This direction and issue came up from our #pattern-ideas channel on slack.
I thought it is worth tracking it because it does offer a good direction.
https://grommet.slack.com/archives/CEKR61G2D/p1560114072001400

from @crashuniverse:

I spent a few weeks with grommet and I had an idea this evening that I thought I can contribute to. I'll put my notes here to generate any reactions.
Improvements with how we write stories:

1. Usage: Every story must have the first story as the simplest one. This helps a developer
   to see the first story and look at the simplest props to make it work. This also reflects the core principle - 80% of grommet by learning the simplest way of using a component if user doesn't need much customization.
2. Storage of stories: Currently stories use filenames {component}.stories.js (Anchor & 34 more) 
   or stories/{use_case}.stories.js (Accordion & n more). We can restructure everything to follow
   2nd way i.e. using stories/ directory that seems to be more efficient by grouping and moreover
   separating them from actual components.
   To find stories following pattern 1:
   `~/D/l/grommet ❯❯❯ find src | grep -i .stories.js`
3. Fixing links in documentation: Story links of pattern https://storybook.grommet.io/?selectedKind=${availableAt}&full=0&addons=0&stories=1&panelRight=0
   are no longer valid so it should be updated with new pattern https://storybook.grommet.io/?path=/story/${availableAt}
   This would still not work completely unless we do 1 and have a "basic" story for every
   component and then use https://storybook.grommet.io/?path=/story/${availableAt}--basic
   Needs more research for a proper fix.
4. Add missing stories: Assess if some components like AccordionPanel (& 12 more) do need a story.

ShimiSun:
Some components have stories named `Custom` because they are demonstrating the capabilities of the 'Custom Theme' of the component, and 'Custom' was the shortcut name for the story.
I'm thinking of renaming all the story names of `Custom` to `Theme` or some other short name that implies 'theme' context. 
This way the users will be more exposed to the Theme concept and get immediate theme examples per component. Custom doesn't imply 'theme' context because it can imply other types of customizations.
Thoughts and ideas will be appreciated.

ericsoderberghp:
Excellent notion.

ShimiSun:
Wonderful work on the storybook refactoring efforts!

PR #3523, opened 2019-11-04
Make Storybook Example Titles Capitalization Consistent

taysea:
#### What does this PR do?
This PR capitalizes storybook examples into 'Title Case' and renames some stories to clearer names.

src/js/components/Accordion/stories/Simple.js

There isn't consistency in how storybook examples were capitalized. This PR aims to establish a style guide that storybook examples should be in title case (first letter of each word capitalized).

No.

No.

### DIFF
--- a/src/js/components/Accordion/stories/Simple.js
+++ b/src/js/components/Accordion/stories/Simple.js
@@ -35,7 +35,7 @@ const SimpleAccordion = props => {
 
 storiesOf('Accordion', module)
   .add('Simple', () => <SimpleAccordion />)
-  .add('Dark no animation', () => (
+  .add('Dark, No Animation', () => (
     <SimpleAccordion animate={false} background="dark-2" />
   ))
   .add('Multiple', () => <SimpleAccordion multiple />);
--- a/src/js/components/Box/stories/Fixed.js
+++ b/src/js/components/Box/stories/Fixed.js
@@ -47,4 +47,4 @@ const FixedSizesBox = () => (
   </Grommet>
 );
 
-storiesOf('Box', module).add('Fixed sizes', () => <FixedSizesBox />);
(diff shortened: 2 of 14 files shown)


ShimiSun:
Thank you for working on this, I'm not sure if capitalizing is the right approach, why do you think it is better than how it is today?
I'm looking at the storybook official example, and it doesn't look capatalized to me https://storybooks-official.netlify.com/?path=/story/ui-panel--default

taysea:
I guess that the aim of this PR was more to propose consistency in the way storybook examples are named because in ours, some stories are in title case 'Story Example' and others are like 'Story example'. Maybe capitalizing everything isn't the perfect solution (I'm very open to other options), but I think some discussion about a style guide for story naming could be beneficial moving forward. 

Some concepts for discussion could be:
- If a storybook example is the name of the prop, do we leave lowercase to match how it would be typed in code? (`border` or `onChange` as opposed to `Border` or `OnChange`)
- If a storybook is combining multiple props or concepts, what punctuation do we use to indicate this? (`Controlled and Grouped` or `Controlled & Grouped` or `Controlled, Grouped` or `Controlled + Grouped`)

I could open this as a discussion issue if you think it's fitting. These changes don't affect how the code works, but I think consistency in naming conventions would make for a smoother user experience and create a more polished look.

ShimiSun:
agreed.
I'm inclining to follow the standards of https://storybooks-official.netlify.com/?path=/story/ui-panel--default.

Feel free to open for a discussion.

ShimiSun:
I'm just dropping a comment here so I won't forget, I noticed on storybook ^5.2 the title case behavior on storybook UI stories is enforced out of the box.

ShimiSun:
linking to #3212

ShimiSun:
We came to a resolution that all stories should use sentence case.

Issue #3525, opened 2019-11-05
Menu component does not honor a11yTitle text


LiveFreeAndRoam:
Attempts to set the a11yTitle on a menu component does not work.  Instead, the aria-label appears to be hardcoded to "Open Menu"
 
### Expected Behavior
I expected the HTML button's aria-label to be set the same as my Grommet's Menu a11yTitle text.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/menu-a11ytitle-not-honored-48mq7

Use developer tools to observe the HTML and aria-label.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: v2
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): Windows 10.

taysea:
Hey there, thanks for bringing this up. At the current time, the way to adjust the aria-label would be to use the `messages` prop (https://v2.grommet.io/menu#messages). By default the control button has 'Open Menu' so that users with screen readers can receive an indication of what pressing the button will do, but you are welcome to change it to the label on your menu or anything you'd like. Hopefully that allows you to make a fix for now. I will look into why we also accept an a11yTitle on this component even though it has no effect. Thanks!

taysea:
@ShimiSun I think we can close this now that my PR merged : )

PR #3527, opened 2019-11-05
Add a11yTitle prop to Menu

taysea:
#### What does this PR do?
This PR allows users to supply an a11yTitle prop to Menu. While it already existed in the docs, when users provided this prop it did not have an effect. 

This has been tested by adding `a11yTitle="My Menu"` in a local project. The DOM was inspected and the button had an aria-label of "My Menu". Additionally, a new jest test has been added to avoid any future regressions.

Add `a11yTitle="My Menu"` to a Menu and use the inspector tools to view the aria-label.

Currently, the aria-label on Menu was being controlled by the messages prop. By default, the menu button had an aria-label of 'Open Menu' when closed, and 'Close Menu' when open. If users supplied a custom a11yTitle, it was not changing the aria-labels. Now, if an a11yTitle is provided, then the aria-labels will take the value of this string. With this change, the user should provide either an a11yTitle or messages as a prop but not both. If both are provided, then the a11yTitle will override custom/default messages.

#3525 

Yes, I have updated them in this PR.

No.


ShimiSun:

Issue #3517, opened 2019-10-31
Grid component a11yTitle is documented but not clear


LiveFreeAndRoam:
Using `<Grid a11yTitle="ARIA label">...</Grid>` does not produce a div tag wth an `aria-label`.

### Expected Behavior
It is unclear what the expected behaviour ought to be.  Perhaps the `a11yTitle` attribute ought to be removed from the `<Grid>` component?

It is unclear.  The documentation describes it, so I was only experimenting with it to understand how it worked.

### Actual Behavior
`aria-label` was not rendered.

Despite using `a11yTitle` attribute in my `<Grid>` component, it did not render an `aria-label` attribute.
### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: v2
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): Windows 10.

ShimiSun:
Thank you for reporting, the fix will be available on our next release! (soon :) )

PR #3528, opened 2019-11-06
Add a11ytitle support for Grid

taysea:
#### What does this PR do?
This PR allows users to provide an a11yTitle to Grid. Even though the prop was listed in the documentation, when users provided this prop it was not appearing in the DOM.

src/js/components/Grid/Grid.js

This change has been tested in a local project by adding the prop `a11yTitle="My Grid"` to a Grid. Using the inspector tools, I could see that the aria-label was added. I also turned on voiceover to check that a screen reader would be able to pick up on the label. Additionally, I've added a new jest test that checks that the aria-label is created.

In a local project create a grid and add the `a11yTitle` prop. Using the inspector tools, check that the aria-label appears in the DOM.

While the docs said you could supply an a11yTitle to Grid, the code wasn't supporting this. So, when users tried to add this prop, it wouldn't have an effect on their code.

#3517 

No.

No.


ShimiSun:
I'm wondering how come this change didn't impact any of the snapshots, will you be willing to create a test that is capturing this prop? we can see from the snapshots if it behaves as expected.

taysea:
Hey Shimi, I did create a test that uses this prop (see the changes on src/js/components/Grid/__tests__/Grid-test.js). The reason it doesn't affect the other snapshots is because I didn't add the a11yTitle prop to those tests. If no a11yTitle is provided, then there will be no aria-label added to the DOM.

ShimiSun:

PR #3530, opened 2019-11-06
Clarify a11yTitle docs

taysea:
#### What does this PR do?
This PR clarifies what is added to the DOM when an a11yTitle prop is provided. The name of this prop is slightly deceiving because ARIA has attributes called `aria-label` and `title`. While our prop is called `a11yTitle`, it is actually affecting the `aria-label`. While it wouldn't be smart to change the name of our prop due to backwards compatibility, a clarification to the docs may be helpful to users.

src/js/utils/prop-types.js

The tests were updated to reflect the change in wording.

While our prop is called `a11yTitle`, it is actually affecting the `aria-label`. While it wouldn't be smart to change the name of our prop due to backwards compatibility, a clarification to the docs may be helpful to users.

#3517 

Yes, this PR updates docs only.

No.


src/js/utils/prop-types.js
```diff
@@ -1,7 +1,8 @@
 import { PropTypes } from 'react-desc';
 
 export const a11yTitlePropType = PropTypes.string.description(
-  'Custom title to be used by screen readers.',
+  `Custom label to be used by screen readers. When provided, an aria-label will
+   be added to the element.`,
```
ShimiSun: @ericsoderberghp or @alansouzati, was there a reason we avoided mentioning that `a11yTitle` is actually being translated to `aria-label` under the hood?

ericsoderberghp: I can't remember. :(

ShimiSun:
@taysea I think some updated snapshots are missing here 
Screen Shot 2019-11-15 at 8 23 11 AM

taysea:
Let me update those! @ShimiSun

ShimiSun:
Thank you! 🎉

Issue #3340, opened 2019-09-12
Boxy Buttons


ericsoderberghp:
### Expected Behavior

It should be easier to create Buttons with padding and rounding.

```
<Button icon={<Add />} round="full" hoverIndicator />
```

```
<Button label="Dashboard" pad={{ horizontal: 'medium', vertical: 'small' }} size="large" />
```

### Actual Behavior

When creating rounded icon Buttons or tile Buttons you have to wrap the Button in a Box that applies the rounding and sets `overflow="hidden"` to prevent the Button hoverIndicator from being square.

```
<Box round="full" overflow="hidden">
    <Button icon={<Add />} />
</Box>
```

Also, when using Button in a sidebar menu, the Button needs to contain a Box with padding and text sizing.

```
<Button>
    <Box pad={{ horizontal: 'medium', vertical: 'small' }}>
        <Text size="large">Dashboard</Text>
    </Box>
</Button>
```

I'm actually wondering if we should allow Button to accept all Box props. Crazy idea, I know.

ShimiSun:
Conversation from Slack around this "crazy" idea.

>David Edelhart Today at 11:58 AM
Thanks. Your button class is far too opinionated. Deciding for us when a button has a border and a fill and when it doesn't is extremely arbitrary and unworkable in production. Nice layout stuff though.

>David Edelhart  3 hours ago
you can fill a button's area --- IF its primary. You can stroke its border -- IF its not primary . Basically you don't have all the options; if you allowed for a style (button.primary, button.primary.active) with a fill and/or stroke then the developer can decide. Yes you have a lot of properties.

>David Edelhart  3 hours ago
Also, it seems to me there are basically three styles: primary, base, and plain. having two flags just confuses the issue.

https://grommet.slack.com/archives/C04LMHN59/p1570471095246800

bingomanatee:
As a follow up: You have a good extended styling convention around Box. In fact its fair to say there is very little difference between a box and a button other than that it has hover/mouse interactions. 

If you used the theming overlay as a "factory" with mouse over, down, and inactive variations, then button could be a function that produces mutable boxes; you wouldn't really need a Button "component" as such.  I.e., you can have a static function Button.defineVariant ('active', {
labelStyle, borderStyle, 
disabled: { labelStyle, borderStyle},
hover: { labelStyle, borderStyle},
down: {labelStyle, borderStyle}
}) and then you could have <Button variant="active">foo</Button>

[...]

PR #3535, opened 2019-11-08
Changed Box to add onClick

ericsoderberghp:
#### What does this PR do?

Changed Box to add onClick

Box/doc.js

Storybook stories with Safari, Chrome, and Firefox.
unit tests

same

See the issue below.

#3340 

automatic?

yes

backwards compatible


src/js/components/Box/Box.js
```diff
@@ -1,114 +1,156 @@
...
+import React, {
+  Children,
+  forwardRef,
+  useContext,
+  useMemo,
+  useState,
+} from 'react';
 
-import { withForwardRef, withTheme } from '../hocs';
 import { ThemeContext } from '../../contexts';
```
ShimiSun:
```suggestion
import { ThemeContext } from 'style-components';
```

ericsoderberghp:
> 3. Missing types on base.d.ts for Box hover.

Can you clarify this? base.d.ts already has types for global.hover, which is what Box uses.

> 4. I think `hover` should have an `extend` option on the theme, for example, if a user has a Box with a rounded corner, I'm not sure she can round the corners on hover to maintain the look and feel without an `extend` option, but I might be missing something.

The hover styling does not set rounding, so whatever rounding is set on the Box will be used when hovering.

PR #3399, opened 2019-10-02
Refactor Components/All story to hooks

dustykeyboard:
As requested in https://github.com/grommet/grommet/issues/3394 the story in Components/All is converted to use the `useState` hook.

#### What does this PR do?
The story in Components/All is converted to a functional component using the `useState` hook.

#### Where should the reviewer start?
Check the Components/All story

#### What testing has been done on this PR?
Visual testing in storybooks

#### How should this be manually tested?
Visual testing in storybooks

#### Any background context you want to provide?
As requested in https://github.com/grommet/grommet/issues/3394

#### What are the relevant issues?
https://github.com/grommet/grommet/issues/3394

#### Do the grommet docs need to be updated?
No

#### Should this PR be mentioned in the release notes?
No

#### Is this change backwards compatible or is it a breaking change?
Yes

PR #3547, opened 2019-11-13
Changed RadioButtonGroup to use hooks

ericsoderberghp:
#### What does this PR do?

Changed RadioButtonGroup to use hooks.

The snapshot change indicates that the old version was incorrect. We should not have been passing the value down to the div in the DOM.

storybook, unit

storybook

#3399 

no

no

backwards compatible


ShimiSun:
linking to #3394

PR #3548, opened 2019-11-14
Changed RadioButton to add children rendering

ericsoderberghp:
#### What does this PR do?

Changed RadioButton to add children rendering
Also restructured the stories and use `describe()` for unit tests.

RadioButton/doc.js

storybook and unit testing

storybook

This change is a precursor to making a similar change in RadioButtonGroup, which will then be used in grommet-designer.

automatic

yes

backwards compatible


src/js/components/RadioButton/doc.js
```diff
@@ -18,6 +18,14 @@ export const doc = RadioButton => {
 
   DocumentedRadioButton.propTypes = {
     checked: PropTypes.bool.description('Same as React <input checked={} />'),
+    children: PropTypes.func.description(
```
ShimiSun: I think we should also add `children` prop to index.d.ts.

ericsoderberghp: Isn't it already there because of `React.ComponentClass`?

ShimiSun: I took a look at an example, and you are right. 
I suspect we should do a clean up on all the components that are defining `children` on their index.d.ts file. thanks for your comment.

Issue #3546, opened 2019-11-13
gap prop ignored for items in Menu


blakehilliard:
For Menu "items", the documentation says `The object values can be any Button prop`. But if I specify a "gap", it is ignored.

### Expected Behavior

The gap should be used to put space between the icon and label of the item, just like in a "Button".

### Actual Behavior

No gap is used.

### URL, screen shot, or Codepen exhibiting the issue

Here is a codesandbox demonstrating the issue: https://codesandbox.io/s/grommet-menu-item-no-gap-86omy

### Steps to Reproduce

```
      <Menu
        label="Menu"
        items={[
          {
            icon: <Power />,
            label: "Power",
            gap: "xlarge"
          }
        ]}
      />
```

### Your Environment

- Grommet version: 2.8.0
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): Windows desktop

ShimiSun:
Thank you for filing this issue. It's a bug with how the icon is rendered separately from the button. We'll take a look at that.

ShimiSun:
Thank you for reporting, the fix will be available on our next release.

PR #3550, opened 2019-11-14
Allow gap to affect menu items

taysea:
#### What does this PR do?
This PR improves menu item functionality by extending `gap` to the menu items. If a label and icon are provided with gap, then the gap will be placed between these two items.

This change has been testing on a local project by using the following code:
```
<Menu
        label="actions"
        items={[
          { label: 'Launch', icon: <Power />, gap: 'xlarge', onClick: () => {} },
          { label: 'Abort', onClick: () => {} },
          { label: 'Disabled', disabled: true },
        ]}
      />
```
Additionally, a test has been added to Menu-test.js to ensure that the gap is rendered in the DOM.

Using the above code snippet, make sure that the expected gap appears between the label and icon.

Menu items are said to accept any prop that button accepts. Prior to these changes, gap was not having an effect on menu items.

#3546 

No.

No.


package.json
```diff
@@ -50,6 +50,7 @@
   },
   "dependencies": {
     "@testing-library/dom": "^6.1.0",
+    "@testing-library/jest-dom": "^4.2.4",
```
ShimiSun: with package.json changes we need to follow with updating yarn.lock file. that is the reason the PR's build broke.

src/js/components/Menu/__tests__/Menu-test.js
```diff
@@ -58,6 +59,31 @@ describe('Menu', () => {
...
+            { label: 'Item 2' },
+          ]}
+        />
+      </Grommet>,
+    );
+
+    const firstItem = getByText('Item 1');
+    expect(
+      firstItem.getElementsByClassName(
+        'StyledBox__StyledBoxGap-sc-13pk1d4-1 bPCnCu',
```
ShimiSun: I wonder if this class name is something we can count on with the testing, I think it may change once snapshots classes are updated (once a package is updated foe=r example).
maybe we should just test the prefix of the class name?

taysea: I think that should work now that I've drilled down to the `firstItem`. Let me update the test and just try using the `StyledBox__StyledBoxGap` portion.

ShimiSun:

PR #3555, opened 2019-11-15
Support reverse prop on menu items

taysea:
#### What does this PR do?
This PR allows users to add 'reverse' to their Menu items, this results in the order of the label and icon being reversed from the default. Menu items are supposed to support all props that button supports. Reverse is a prop that should be supported but previously was not.

This PR has been tested in a local project with the following code:
```
<Menu
        dropProps={{ align: { top: 'bottom', left: 'left' } }}
        label="actions"
        items={[
          { label: 'Logout', icon: <Power />, gap: 'small', reverse: true },
        ]}
      />
```
Additionally, a test has been added to check that when `reverse` is provided, the icon renders after the label.

Using the above code snippet in a local project.

Menu documentation states that menu items should support all the same props that Button supports. `Reverse` is a prop that is accepted by button but was not being supported by Menu items.

#3546 

No.

No.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -48,7 +48,8 @@
     "react/jsx-curly-newline": 0,
     "max-len": [2, { "ignoreUrls": true, "ignoreRegExpLiterals": true }],
     "react-hooks/rules-of-hooks": 2,
-    "react-hooks/exhaustive-deps": 1
+    "react-hooks/exhaustive-deps": 1,
+    "react/sort-comp": 0
   },
   "overrides": [
     {
--- a/package.json
+++ b/package.json
@@ -50,6 +50,7 @@
   },
   "dependencies": {
     "@testing-library/dom": "^6.1.0",
+    "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.1.3",
     "css": "^2.2.3",
     "grommet-icons": "^4.2.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "grommet",
-  "version": "2.8.0",
+  "version": "2.8.1",
   "main": "index.js",
   "module": "es6/index.js",
   "jsnext:main": "es6/index.js",
--- a/src/@types/storybook-chromatic/index.d.ts
+++ b/src/@types/storybook-chromatic/index.d.ts
@@ -1,0 +1,3 @@
+declare module 'storybook-chromatic' {
+  export default function isChromatic(): boolean;
+}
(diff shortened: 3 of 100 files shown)


ShimiSun:
Great catch and quick turnaround!! 🚀 
I've tried the `reverse` on the ItemIcon.js story, and it looks a little off with its the margin, will you be willing to take a look?
Screen Shot 2019-11-15 at 6 24 58 PM

taysea:
Ooo I see. Let me look into that!

taysea:
Hey @ShimiSun! So I checked out the code for that example and the reason things get misaligned is because the label and icon in these cases are wrapped in independent boxes:
```
{
            label: <Box alignSelf="center">Slack</Box>,
            onClick: () => {},
            icon: (
              <Box pad="medium">
                <Slack size="large" />
              </Box>
            ),
          },
```
The way the code works is that it just puts on element before or after the other but doesn't look into its contents. In this case, since icon has a pad applied but the label doesn't, when their order is reverse the spacing looks off from the rest of the items. Do you think we should try to account for this and swap the props of the boxes? I'm wondering if it would be worth it to try to account for this or if the user would just need to adjust the box styling if they choose to reverse. Let me know your thoughts!

ShimiSun:
I think that as long we are able to provide a story that demonstrates this functionality in a user-friendly way we are good. maybe you can create a story for this prop that demonstrates it.

taysea:
I can do that! : )

ShimiSun:
One failing tests:
```
        .c2 {
          padding: 6px;

      373 |       expect.stringMatching(/^Item 1/),
      374 |     );
    > 375 |     expect(container).toMatchSnapshot();
          |                       ^
      376 |   });
      377 | });
      378 | 

      at Object.<anonymous> (src/js/components/Menu/__tests__/Menu-test.js:375:23)


Snapshot Summary
 › 1 snapshot failed from 1 test suite. Inspect your code changes or run `yarn test -u` to update them.

Test Suites: 1 failed, 52 passed, 53 total
Tests:       1 failed, 476 passed, 477 total
Snapshots:   1 failed, 588 passed, 589 total
Time:        32.116s
Ran all test suites.
error Command failed with exit code 1.
```

taysea:
Looking into the failing test!

taysea:
Is it possible that the updated snapshots are coming from some other recently merged changes? I merged master into my local branch just now then undid the changes in the PR and there are still snapshots being updated. This makes me think it could be due to something else.

ShimiSun:
@taysea can you run `yarn test-update` and update the snapshots?

ShimiSun:

PR #3560, opened 2019-11-19
Fix typescript support

dallas:
#### What does this PR do?

It fixes TypeScript support. It removes the `skipLibCheck` option in the `tsconfig` as that was avoiding type-checking any and all `*.d.ts` files, which is all that Grommet uses to add TypeScript support. It also adds a custom type declaration file for the `storybook-chromatic` add-on as none seems to exist elsewhere on the interwebs.

I think the diff is simple enough and speaks fairly well for itself.

The only question I really have is around the `TextProps` type: where it should live (as one file was expecting it to be in `utils`) and should it have the `extends` property?

Manual testing running the `tsc` command to spit out TypeScript errors.

Just run the `tsc` command.

I love Grommet and I love TypeScript. It’s nice to help them co-exist :)

None, but I raised a bit of a stink in some commit comments 😬 

N/A

I don’t think so.

Sure.

It is an unbreaking change 😛


dallas:
Hmm… this has some unwanted changes in it. Let me make a new one.

PR #3561, opened 2019-11-19
Fix TypeScript support

dallas:
* Remove `skipLibCheck` from tsconfig
* Add custom type declaration file for storybook-chromatic add-on

Signed-off-by: Dallas Reedy <code@dallasreedy.com>

#### What does this PR do?

It fixes TypeScript support. It removes the `skipLibCheck` option in the `tsconfig` as that was avoiding type-checking any and all `*.d.ts` files, which is all that Grommet uses to add TypeScript support. It also adds a custom type declaration file for the `storybook-chromatic` add-on as none seems to exist elsewhere on the interwebs.

I think the diff is simple enough and speaks fairly well for itself.

The only question I really have is around the `TextProps` type: where it should live (as one file was expecting it to be in `utils`) and should it have the `extends` property?

Manual testing running the `tsc` command to spit out TypeScript errors.

Just run the `tsc` command.

I love Grommet and I love TypeScript. It’s nice to help them co-exist :)

None, but I raised a bit of a stink in some commit comments 😬 

N/A

I don’t think so.

Sure.

It is an un-breaking change 😛


src/js/components/Text/index.d.ts
```diff
@@ -9,12 +9,14 @@ import {
   PolymorphicType,
   TextAlignType 
 } from "../../utils";
+import { ExtendType } from "src/js/themes/base";
```
dallas: This is the main thing I have a question about. In src/js/components/FormField/stories/typescript/CustomTheme.tsx, the `extend` property is used and it was complaining because `label` uses `TextProps` which did not recognize the `extend` property. So I added it. I’m not sure if that was what was desired of if, perhaps, the example in `CustomTheme` isn’t accurate.

dallas: (NOTE: I did update this import to come from `"../../themes/base"` in a later commit.)

thenormalsquid:
<3 Thank you. This issue was preventing us from upgrading to Grommet 2.8 in our typescript app.

dallas:
Just prefixed the title with `WIP: ` because `yarn link`ing this version of Grommet into my project is still having TypeScript issues. So there’s more work to be done than simply getting `tsc` to run cleanly.

dallas:
Removing the `WIP` prefix as I believe this is ready for review now.

src/js/components/Text/index.d.ts
```diff
@@ -9,12 +9,14 @@ import {
...
   TextAlignType 
 } from "../../utils";
+import { ExtendType } from "../../themes/base";
 
 export interface TextProps {
   a11yTitle?: A11yTitleType;
   alignSelf?: AlignSelfType;
   as?: PolymorphicType;
   color?:  ColorType;
+  extend?: ExtendType;
```
ShimiSun: @dallas  why do you think we should have an `extend` prop for typing while it doesn't exist for Text? It is declared on the theme styling and I think it should be enough.

ShimiSun: I see your related question on top ⬆️  I'm on it!

ShimiSun:
Fantastic PR!! 
I added quick commits for adjusting/cleaning it, but overall we appreciate this contribution!
Looking forward to seeing more typescript improvements and alerts from you guys @dallas @thenormalsquid 
Thanks again @dallas !

src/js/components/Text/index.d.ts
ShimiSun: I Fixed it!

ShimiSun: good catch!

dallas: Thanks! 😊

src/js/utils/index.d.ts
```diff
@@ -35,16 +35,17 @@ declare const removeUndefined: <T extends object>(obj: T) => NonUndefinedProps<T
 export {isObject, deepFreeze, deepMerge, removeUndefined};
 
 // Extracting types for common properties among components
+type BoxSideType = "top" | "left" | "bottom" | "right" | "horizontal" | "vertical" | "all";
+type BoxSizeType = "xsmall" | "small" | "medium" | "large" | "xlarge" | string;
+type BoxStyleType = "solid" | "dashed" | "dotted" | "double" | "groove" | "ridge" | "inset" | "outset" | "hidden";
```
dallas: I understand why this got moved up as they are only being used internally and not exported like the other types below, but I’m not sure I fully understand why they got renamed and why the naming chosen was `Box` instead of `Border`. Not a big deal, obviously, I’m just seeking understanding from people who have the bigger picture in view.

dallas:
Thanks, @ShimiSun! Thanks for all the work you all do on this library all the time! You all rock!

src/js/utils/index.d.ts
ShimiSun: Good question. 
It was important to me to make it clear that the types you see there of BoxSideType, BoxSizeType, BoxStyleType should only be associated with Box component. 
For example, some other components other than Box might have `size` but it will use different values and not the exact Box sizing, so I added the extra description to type so no one will be tempted to use those props for components that are not underlined with the Box component.
I agree this might need further clean-ups, as the code grows we grow to understand the delicate handling of typescript as well.

BTW, not sure if you saw the slack message, but grommet 2.8.1 is out and this PR was a great reason to trigger it 🎉

PR #3588, opened 2019-12-03
Menu - added ts example and a fix missing Button prop types

ShimiSun:


ShimiSun:
`plain` prop is coming from Button props, this PR needs to be adjusted to take all props of Button.

ShimiSun:
@ericsoderberghp I've updated Menu to accept all Button prop types, and not just plain. the tsx story is working as expected w/o errors.

PR #3592, opened 2019-12-04
Update README.md / Added standalone componnets

jonisar:
Added a link to bit.dev/grommet/grommet for the installation of standalone components, in the micro-frontends/atomic-design spirit.
It also lets people import the components into their own projects for modifications when needed. Even though there are examples there too, I added it under "Installation" and not "Explore" since the real value is in the flexible distribution and consumption of standalone components.

#### What does this PR do?

It adds a link to the flexible consumption of standalone components.

-

-

-

Through bit.dev developers can export, distribute, consume and collaborate on standalone components: https://docs.bit.dev/docs/quick-start

-

-

no

no

-

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,10 @@ For Yarn users:
   $ yarn add grommet styled-components
 ```
 
+Install standalone components, or import for local code modifications:
+
+[bit.dev/grommet/grommet](https://bit.dev/grommet/grommet)
+
 There are more detailed instructions in the [Grommet Starter] app tutorial for
 new apps. For incorporating Grommet into an existing app, see the [Existing App]
 version.


ShimiSun:
Hi buddy @jonisar , thanks for the PR! I was thinking it might work better if we'll place this information on https://github.com/grommet/grommet-starter-new-app instead of the Grommet README page.
One more thing, how can I update the version number of the components on bit? I noticed they all point to an older version.

ShimiSun:
We should add a separate wiki page to how bit can be used and supported by grommet.

ShimiSun:
linking to #3887

ShimiSun:
This change would be relevant once we will have an automated bit process. currently, we do not and the components on bit are not updated as frequently as we release.

Issue #3580, opened 2019-12-03
Proposal: Add support for more semantic React Elements


iMerica:
### Background

One pain point of Grommet currently is readability and debuggability due to the overloading of `Box`, which communicates zero intent. 

For example, let's say all your margins are inconsistent between pages on a site and you want to fix it. In today's Grommet, this is a frustrating experience because, when looking in your Web Inspector all you will see is Style Component random classes. If you look in the Chrome React Toolbar all Boxes are indistinguishable. So we are left with reading through the source, parsing each `<Box/>` trying to guess what its intent is. 

This is a slow and frustrating way to develop in React, so I would like to propose support for some kind of way to address this either with built-in semantic elements with sensible defaults or with `as` props to annotate each `Box`. 

@ericsoderberghp proposed this example

```jsx
const Header = props =>
  <Box as="header" direction="row" align="center" justify="between" gap="medium" {...props} />;
const NavBar = props =>
  <Box as="nav" direction="row" align="center" {...props} />;
const Nav = props =>
  <Box as="nav" {...props} />;
const Main = props =>
  <Box as="main" fill="vertical" overflow="auto" {...props} />;
const Footer = props =>
  <Box as="footer" direction="row" align="center" justify="between" gap="medium" {...props} />;
```

My opinion:

### Good Improvement

Grommet  provides an `as` prop and makes this value prefix styled component classNames. Sort of like an annotation.

### Better Improvement

Grommet provides `Header`, `NavBar`, `Nav`, `Main`, `Footer` components (and more) that already have sensible default styles/props and use the component names in the generated styled component classNames

ShimiSun:
Thanks again for filing this issue! The components we agreed to implement are all up now. Feel free to shoot us more great ideas.

iMerica:
👍

PR #3598, opened 2019-12-05
Added Footer Component

ShimiSun:
#3580
Let's discuss it more tomorrow.
I want to add one more story example of a sitemap.

### DIFF
--- a/src/js/components/Footer/Footer.js
+++ b/src/js/components/Footer/Footer.js
@@ -1,0 +1,24 @@
+import React from 'react';
+
+import { Box } from '../Box';
+
+const Footer = ({ ...rest }) => (
+  <Box
+    as="footer"
+    direction="row"
+    flex={false}
+    gap="medium"
+    justify="between"
+    {...rest}
+  />
+);
+
+let FooterDoc;
+if (process.env.NODE_ENV !== 'production') {
+  // eslint-disable-next-line global-require
+  FooterDoc = require('./doc').doc(Footer);
+}
+
+const FooterWrapper = FooterDoc || Footer;
+
+export { FooterWrapper as Footer };
--- a/src/js/components/Footer/Footer.js
+++ b/src/js/components/Footer/Footer.js
@@ -5,6 +5,7 @@ import { Box } from '../Box';
 const Footer = ({ ...rest }) => (
   <Box
     as="footer"
+    align="center"
     direction="row"
     flex={false}
     gap="medium"
--- a/src/js/components/Footer/stories/Simple.js
+++ b/src/js/components/Footer/stories/Simple.js
@@ -12,14 +12,9 @@ const Simple = () => (
       <Box flex />
     </Main>
     <Footer background="light-4" justify="center" pad="small">
-      <Box>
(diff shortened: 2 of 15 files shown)


src/js/components/Footer/Footer.js
```diff
@@ -0,0 +1,24 @@
...
+
+import { Box } from '../Box';
+
+const Footer = ({ ...rest }) => (
+  <Box
+    as="footer"
+    direction="row"
+    flex={false}
+    gap="medium"
+    justify="between"
```
ericsoderberghp: any reason not to include `align="center"`?

src/js/components/Footer/stories/Simple.js
```diff
@@ -0,0 +1,32 @@
...
+      elevation="large"
+      pad={{ top: 'large' }}
+      gap="large"
+    >
+      <Text margin="small" size="xsmall">
+        Main Content
+      </Text>
+      <Box flex />
+      <Footer background="light-4" justify="center">
+        <Box pad="small">
```
ericsoderberghp: why not put the pad on the Footer?

src/js/components/Footer/stories/Simple.js
```diff
@@ -0,0 +1,32 @@
...
+      background="light-1"
+      elevation="large"
+      pad={{ top: 'large' }}
+      gap="large"
+    >
+      <Text margin="small" size="xsmall">
+        Main Content
+      </Text>
+      <Box flex />
+      <Footer background="light-4" justify="center">
```
ericsoderberghp: Did you try not setting `justify`?

ShimiSun: Footer has a default of `justify="between"`, which is better for links/sitemap content.
In this case, we need  `justify="center"` because that otherwise the content will align to the right. 
Screen Shot 2019-12-05 at 10 37 31 AM

src/js/components/Footer/stories/Simple.js
ShimiSun: Because essentially it works the same and I thought it will be better to assign it only for the relevant component?!
If you are asking because you potentially wants to remove the Box, the Bos is needed to default to `direction="column"`.

src/js/components/Footer/Footer.js
ShimiSun: Thinking about it...

ShimiSun:
@ericsoderberghp Updated PR with 4 Footer examples on storybook following our conversation today.

src/js/components/Footer/Footer.js
ShimiSun: Added.

iMerica:
Consider adding `style={{marginTop: 'auto'}}` so that by default the footer is sticky.

ShimiSun:
Thanks for your comment @iMerica.
I don’t think Footer should be sticky by default and it is for the user to decide according to the use case. we can provide a storybook example of how this could be achieved with Main that contains a Box with `flex="grow"` before Footer, rather than using `margin-top: auto`. feel free to make a story contribution :)

PR #3602, opened 2019-12-07
Changed Form stories to include FormField stories

ShimiSun:
Semantically, it is more correct to use FormField inside of a Form, hence this PR is a suggestion of moving all of FormField stories inside Form (including tsx story).

A suggestion or a next step proposal will be to wrap the FormField stories with a Form component although it may be simpler to see how an individual FormField can be customized or used. open to more ideas.


ericsoderberghp:
Agreed on changing the stories to always show FormField inside a Form, for a follow on change.

PR #3604, opened 2019-12-07
Fixed an issue with dark background detection when using alpha

ericsoderberghp:
#### What does this PR do?

Fixed an issue with dark background detection when using alpha.

The RangeSelector snapshot changes reflect handling the alpha better. We never have text on the selected range stack layer, but the color value there should reflect the inherited light background context anyway.

Also, you can see the improvements to the contrast in the Box Background story for "brand opacity" in the Chromatic changes.

colors-test

Added a new colors-test unit test

unit test

no

yes

backwards compatible


src/js/components/Video/__tests__/__snapshots__/Video-test.js.snap
```diff
@@ -163,8 +163,7 @@ exports[`Video autoPlay renders 1`] = `
 
 .c5:hover {
   background-color: rgba(221,221,221,0.4);
-  color: #444444;
-  color: #FFFFFF;
+  color: #000000;
```
ShimiSun: So I looked into this change, and the background color `rgba(221,221,221,0.4)` is translated to `#dddddd`, and the color-hex website using white color for its text for contrast instead of black as suggested in this line. 
I'm not a contrast expert but it might worth taking a look at this suggestion https://www.color-hex.com/color/dddddd

ShimiSun: regardless, this change is a potential backward compatibility candidate.

src/js/utils/background.js
```diff
@@ -85,20 +85,21 @@ export const backgroundStyle = (backgroundArg, theme, textColorArg) => {
...
-          };`}
+            ] || textColor,
+            theme,
+          )};`}
       `);
     }
     if (background.dark === false) {
       styles.push(css`
-        color: ${textColor.light};
+        color: ${textColor.light || textColor};
```
ShimiSun: Is this one missing `normalizeColor` as well?

src/js/utils/background.js
```diff
@@ -85,20 +85,21 @@ export const backgroundStyle = (backgroundArg, theme, textColorArg) => {
...
       `);
     } else if (background.dark) {
       styles.push(css`
-        color: ${textColor.dark};
+        color: ${textColor.dark || textColor};
```
ShimiSun: Is this one missing `normalizeColor` as well?

src/js/utils/background.js
ericsoderberghp: Good catch! fixed.

src/js/utils/background.js
ericsoderberghp: Good catch! fixed

src/js/components/Video/__tests__/__snapshots__/Video-test.js.snap
ericsoderberghp: I don't think this is a visible change, as the Video control Buttons don't have any visible text.

I will point out that https://contrast-ratio.com/#%23FFF-on-%23DDDDDD isn't accessible, but https://contrast-ratio.com/#%23000000-on-%23DDDDDD is.

PR #3606, opened 2019-12-09
fix TS definition

### LINKED ISSUE #3581: Property 'index' does not exist on type 'MouseEvent<Element, MouseEvent>'
Typescript error in "onClickItem" in https://storybook.grommet.io/?path=/story/list--selection `Property 'index' does not exist on type 'MouseEvent<Element, MouseEvent>'`.

As I can see from typings type of onClickItem is: `onClickItem?: (event: React.MouseEvent) => void;` and `React.MouseEvent` does not have `index`. And as I can guess you adding `index` prop on this event.

### Expected Behavior

No type error.

### Actual Behavior

Type error.

image

### Steps to Reproduce

Copy-paste example from [here](https://storybook.grommet.io/?path=/story/list--selection) to CRA with typescript.

### Your Environment

- Grommet version: ^2.8.1
- Browser Name and version: chrome 78.0.3904.108
- Operating System and version (desktop or mobile): mac

britt6612:
#### What does this PR do?
The PR changes TS definition for onClickItem for List
List/index.d.ts
storybook example that was in issue #3581 
storybook example with onClickItem
In the onClick in list.js it takes an event as well as sets the item to an object and the index to a number which needed to be included in the onClickItem for the TS definitions 
closes #3581 
no
no
backwards compatible


src/js/components/List/index.d.ts
```diff
@@ -18,7 +18,7 @@ export interface ListProps {
   itemProps?: { [_:string]: { background?: string, border?: BorderType, pad?: PadType}};
   margin?: MarginType;
   onMore?: () => void;
-  onClickItem?: (event: React.MouseEvent) => void;
+  onClickItem?: ((event: {item?: {}, index?: number}) => void);
```
ShimiSun: This will solve the issue, thanks! 
The only thing that may be a problem here is that the event will include more props than item & index and we should be able to allow all keys of the event as part of the TS definition.
for example, if the user will try event.target the typing will fail.

britt6612: @ShimiSun yes you are correct if event.target was used this would throw an error. Also after looking at list docs you can also pass in datum so I will include these two in the TS definitions

src/js/components/List/index.d.ts
```diff
@@ -18,7 +18,7 @@ export interface ListProps {
   itemProps?: { [_:string]: { background?: string, border?: BorderType, pad?: PadType}};
   margin?: MarginType;
   onMore?: () => void;
-  onClickItem?: (event: React.MouseEvent) => void;
+  onClickItem?: ((event: {item?: {}, index?: number, datum?: {}, target?: HTMLElement}) => void);
```
ShimiSun: `target` was just an example, I think we should allow all child props and not just `item`, `index`, or `target`, something to let it include all Mouse Event childs or args.

britt6612: okay I believe that React.MouseEvent covers the children but since we are re-declaring item and index its not included in the React.MouseEvent. Im testing out some of the other children to make sure they are included in the MouseEvent

ShimiSun:
Looks good, tested it locally.
@britt6612 Thanks for getting it done for the coming release!

britt6612:
@ShimiSun thank you for being patient yes I tested it locally so for reference in list.js the onClickItem included the event object so having the React.MouseEvent includes all of those props in the object and since we are delcaring adjustedEvent.item = item as well as index so adding those to that object we just needed to account for those 2 on the onClickItem

PR #3608, opened 2019-12-10
Remove theme.global.graph and added theme.global.colors.graph-*

ericsoderberghp:
#### What does this PR do?

Remove `theme.global.graph` and added ability to use `theme.global.colors.graph-*`.

`theme.global.graph.colors` didn't lend itself to making graph keys easy enough. So, removing this, which hasn't been part of an official grommet release yet, thankfully.

base.js

unit
theme designer

unit

no

yes

backwards compatible

### DIFF
--- a/src/js/components/Chart/doc.js
+++ b/src/js/components/Chart/doc.js
@@ -176,15 +176,23 @@ const data = calcs(<values>, { coarseness: 5, steps: [1, 1] });`,
 };
 
 export const themeDoc = {
+  'chart.color': {
+    description: 'Color of the Chart.',
+    type: 'string | (props) => {}',
+    defaultValue: 'accent-1',
+  },
   'chart.extend': {
     description: 'Any additional style for the Chart.',
     type: 'string | (props) => {}',
     defaultValue: undefined,
   },
   'global.colors': {
-    description: 'color options used for Chart fill area.',
+    description: 'color options',
     type: 'object',
-    defaultValue: 'accent-1',
+    defaultValue: `{
+      "accent-1": "#6FFFB0",
+      "graph-0": "accent-1",
+    }`,
   },
   'global.edgeSize': {
     description: 'The possible sizes for the thickness in the Chart.',
@@ -201,14 +209,14 @@ export const themeDoc = {
         responsiveBreakpoint: 'small',
     }`,
   },
-  'global.graph.colors': {
-    description: 'The color to use when not specified via color.',
-    type: '[string] or { dark: [string], light: [string] }',
-  },
   'global.opacity': {
     description: 'The opacity of the Chart stroke.',
     type: 'string',
-    defaultValue: undefined,
+    defaultValue: `{
+      strong: 0.8,
+      medium: 0.4,
+      weak: 0.1,
+    }`,
   },
   'global.size': {
     description: 'The possible sizes for Chart width and height.',
--- a/src/js/components/Chart/doc.js
+++ b/src/js/components/Chart/doc.js
@@ -178,7 +178,7 @@ const data = calcs(<values>, { coarseness: 5, steps: [1, 1] });`,
 export const themeDoc = {
   'chart.color': {
     description: 'Color of the Chart.',
-    type: 'string | {dark: string, light: string},
+    type: 'string | {dark: string, light: string}',
     defaultValue: 'accent-1',
   },
   'chart.extend': {
@@ -187,11 +187,12 @@ export const themeDoc = {
     defaultValue: undefined,
   },
   'global.colors': {
-    description: 'color options',
+    description: 'Color options.',
     type: 'object',
     defaultValue: `{
       "accent-1": "#6FFFB0",
       "graph-0": "accent-1",
+      ...
     }`,
   },
   'global.edgeSize': {
@@ -211,7 +212,7 @@ export const themeDoc = {
   },
   'global.opacity': {
     description: 'The opacity of the Chart stroke.',
-    type: 'string',
+    type: 'object',
     defaultValue: `{
       strong: 0.8,
       medium: 0.4,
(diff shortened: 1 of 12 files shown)


src/js/components/Chart/doc.js
```diff
@@ -176,15 +176,23 @@ const data = calcs(<values>, { coarseness: 5, steps: [1, 1] });`,
 };
 
 export const themeDoc = {
+  'chart.color': {
+    description: 'Color of the Chart.',
+    type: 'string | (props) => {}',
```
ShimiSun: shouldn't this be:
```suggestion
    type: 'string | {dark: string, light: string},
```

src/js/components/Chart/doc.js
```diff
@@ -201,14 +209,14 @@ export const themeDoc = {
         responsiveBreakpoint: 'small',
     }`,
   },
-  'global.graph.colors': {
-    description: 'The color to use when not specified via color.',
-    type: '[string] or { dark: [string], light: [string] }',
-  },
   'global.opacity': {
     description: 'The opacity of the Chart stroke.',
     type: 'string',
```
ShimiSun:
```suggestion
    type: 'string | number',
```

src/js/components/Chart/doc.js
```diff
@@ -201,14 +209,14 @@ export const themeDoc = {
...
-    defaultValue: undefined,
+    defaultValue: `{
+      strong: 0.8,
+      medium: 0.4,
+      weak: 0.1,
```
ShimiSun: Not sure I understand which one is the default though?

src/js/components/Diagram/doc.js
```diff
@@ -53,6 +53,25 @@ export const doc = Diagram => {
...
+    type: 'string | (props) => {}',
+    defaultValue: undefined,
+  },
+  'diagram.line.color': {
+    description: 'The color of the connection line.',
+    type: 'string',
+    defaultValue: 'accent-1',
+  },
+  'global.colors': {
+    description: 'color options',
```
ShimiSun:
```suggestion
    description: 'Color options.',
```

src/js/components/Meter/doc.js
```diff
@@ -70,6 +70,14 @@ export const doc = Meter => {
 };
 
 export const themeDoc = {
+  'global.colors': {
+    description: 'color options',
+    type: 'object',
+    defaultValue: `{
+      "accent-1": "#6FFFB0",
+      "graph-0": "accent-1",
```
ShimiSun: why `"graph-1": "neutral-1",` isn't included in this case?

src/js/themes/base.js
```diff
@@ -62,6 +62,10 @@ const colors = {
     light: 'brand',
   },
   focus: focusColor,
+  // "graph-0": 'accent-1',
+  // "graph-1": 'neutral-1',
+  // "graph-2": 'accent-2',
+  // "graph-3": 'neutral-2',
```
ShimiSun: we'll need to add these on the base.d.ts file

src/js/components/Meter/doc.js
ericsoderberghp: The list can be rather long. I wanted to include at least a couple to give an example. Perhaps add a `...`?

src/js/themes/base.js
ericsoderberghp: Yep

src/js/components/Chart/doc.js
ericsoderberghp: Well, the `global.opacity` property have a value that is an object, as shown. I was following the examples I was seeing for `global.edgeSize`.

src/js/components/Chart/doc.js
ericsoderberghp: Actually, `global.opacity` is an object.

src/js/components/Meter/doc.js
ShimiSun: yes, the reason I brought it up was that it wasn't consistent with other places that showed the default value. `...` should work.

src/js/components/Chart/doc.js
ShimiSun: correct, but in some cases when we actually have a default value we will add something like (assuming the default is `medium`):
```
  'global.opacity.medium': {
    description: 'The opacity value used for the Chart color.',
    type: 'number',
    defaultValue: '0.4',
  },
```

ericsoderberghp: I think the more interesting part is what the default value for the `opacity` property of Chart is, which is undefined/false. I thought showing all opacity values from the theme here aligned better with how the `opacity` property is documented. And, the default value of the `opacity` property should be documented with the component property, not with the theme.

src/js/components/Meter/doc.js
ericsoderberghp: done__

src/js/components/Chart/doc.js
```diff
@@ -176,15 +176,24 @@ const data = calcs(<values>, { coarseness: 5, steps: [1, 1] });`,
 };
 
 export const themeDoc = {
+  'chart.color': {
```
ShimiSun: Let's add `color` to Chart typing, here https://github.com/grommet/grommet/blob/e6bac4512e0ae18fdfce17ba65f4a7d920f1000a/src/js/themes/base.d.ts#L319

src/js/components/Chart/doc.js
ShimiSun: I understand your point.

src/js/components/Diagram/doc.js
```diff
@@ -53,6 +53,26 @@ export const doc = Diagram => {
...
 
 export const themeDoc = {
+  'diagram.extend': {
+    description: 'Any additional style for Diagram.',
+    type: 'string | (props) => {}',
+    defaultValue: undefined,
+  },
+  'diagram.line.color': {
+    description: 'The color of the connection line.',
+    type: 'string',
```
ShimiSun:
```suggestion
    type: 'string | {dark: string, light: string}',
```

src/js/components/Chart/doc.js
ericsoderberghp: done

PR #3619, opened 2019-12-13
Footer + Main - Updated docs to refer to Box

ShimiSun:
Until we find an improved way to communicate the inherit props of Box, I'm adding this description for Footer and Main. It already in place for Header.
I think we are good for a release after this change. one thing that is still missing is document the default props of the components.

### DIFF
--- a/src/js/components/Footer/README.md
+++ b/src/js/components/Footer/README.md
@@ -1,5 +1,5 @@
 ## Footer
-Footer for a document or section
+is a Box container for a document or a section footer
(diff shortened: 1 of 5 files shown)


ShimiSun:
@ericsoderberghp let me know if you think like this refinement is still needed or maybe another wording suggestion, otherwise we can close this PR.

ericsoderberghp:
I think we can close this PR and just run with what we have for a while.

Issue #3489, opened 2019-10-22
grommet enterprise components and design-system, need your comments


ShimiSun:
Grommet is inclining to allow an extended version of components that may be serving the enterprise industry.

The current list of components we had in mind for the extended design-system:
Avatar 
Navbar
Header
Footer
Sidebar
List
Onboard
Breadcrumbs
wizard
pagination
Tip
Tags
Toggle
Popover
Card
Login
Content/Section
Tree
TreeTable
Validation
Dialog Forms
NumberInput
Tiles
Charting
Toast

You can find more context here https://github.com/grommet/grommet/issues/3271

We are not committing to any component yet, and we'll initially like to hear your feedback!
Feel free to share with us any more ideas for components or directions.

Linked to #3354

nomadus:
Happy to help with the priorities. My list of priorities would be (in descending order):
Pagination, breadcrumbs, tip (tooltip), charts, tags, numberinput, datetime (it is not in the list). As I am already using grommet-controls, tags, date and number inputs are at the bottom.

atanasster:
first priorities list:
https://v1.grommet.io/docs/components

sergeylukin:
Thanks for asking! My preference would be as follows:

- Masonry grid (missing!)
- Cards
- Plug-n-play mobile navigation
- All-the-forms (inputs, login, validation, etc.)
- Navbar
- all the rest mostly the ones from v1

jcfilben:
Oof the components listed above the following have been added to grommet:
Avatar
Navbar
Header
Footer
Sidebar
List
Pagination
Tip
Tags
Card
Toast (Notification)

Moving forward we would like to track new components by request in separate issues. I will be closing this issue and separate issues can be created for new components as needed.

PR #3622, opened 2019-12-13
New Nav component

ShimiSun:
New Nav component that will render a collection of anchors.
Nav has an `items=[{label: 'my link', href='#'}...]` prop, but in case the user wants to use custom links or Buttons he can leverage the component's children instead. 

I moved the Sidebar example from Button to Nav, and refactored it to use Nav instead of a Box, I think it makes better sense for the example to be placed on Nav component.

Added on TS story example but tested all the other stories with TS as well.


src/js/components/Nav/Nav.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+import { Anchor } from '../Anchor';
+import { Box } from '../Box';
+
+
+ // TODO resolve whether children should be allowed with items
```
ShimiSun: Opening this for discussion. I think there shouldn't be a problem to allow both.

src/js/components/Nav/Nav.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+import { Anchor } from '../Anchor';
+import { Box } from '../Box';
+
+// TODO resolve whether children should be allowed with items
+const Nav = ({ children, items = [], ...rest }) => (
+  <Box as="nav" flex={false} gap="medium" {...rest}>
+    {items.map(item => (
+      <Anchor href={item.href} label={item.label} key={item.label} />
```
ericsoderberghp: Why Anchor and not Button?

ericsoderberghp: How would the caller integrate this with a router?

ShimiSun: I found Anchor to be more simplified than Button. 
You can leverage the children to show a Button instead (or even a Box), see for example, the CustomChild story on storybook, so the option for alternative components is still there.

ShimiSun: Not sure why should she/he integrate with a router. 
I think the Nav should be separated from the routing logic, just like Anchor/Button doesn't know which router you are using, and just direct to the `href` provided. 
I might be missing something though, so feel free to elaborate.

ericsoderberghp: The routers have to tie in to the `onClick` property. So, you typically want to do something like `<RouterAwareAnchor path="/internal/path" label="Name" />` with the router taking care of `onClick` and `href`.

ShimiSun: I made the component less opinionated by removing `items`.

ShimiSun: I made the component less opinionated by removing `items`, I think it should solve the routing integration issue.

src/js/components/Nav/Nav.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+import { Box } from '../Box';
+
+const Nav = ({ ...rest }) => (
+  <Box as="nav" flex={false} gap="medium" {...rest} />
```
ericsoderberghp: What's the rationale for `gap` defaulting to `medium`? I was envisioning using this in a Sidebar with Buttons, in which case I would want gap to be undefined.

src/js/components/Nav/stories/typescript/CustomAnchor.tsx
```diff
@@ -0,0 +1,50 @@
...
+      light: 'neutral-2',
+    },
+    hover: {
+      textDecoration: 'none',
+      fontWeight: 700,
+    },
+  },
+});
+
+const CustomAnchor = () => (
```
ericsoderberghp: Can we remove this story, now that we don't have `items`?

src/js/components/Nav/Nav.js
ShimiSun: I agree it would have made more sense to use with `items`

src/js/components/Nav/stories/typescript/CustomAnchor.tsx
ShimiSun: done.

src/js/components/Nav/Nav.js
ericsoderberghp: So, are we going to remove the `gap="medium"` here?

src/js/components/Nav/stories/CustomChild.js
```diff
@@ -30,10 +30,10 @@ const SidebarButtons = () => {
...
           ))}
-        </Box>
+        </Nav>
       </Box>
     </Grommet>
   );
 };
 
-storiesOf('Button', module).add('Sidebar', () => <SidebarButtons />);
+storiesOf('Nav', module).add('Custom child', () => <SidebarNav />);
```
ericsoderberghp: Is it a "custom" child? Or is it just a normal use of Nav?

src/js/components/Nav/stories/CustomAnchor.js
```diff
@@ -0,0 +1,45 @@
...
+
+const navItems = [
+  { label: 'HTML', href: '#' },
+  { label: 'JS', href: '#' },
+  { label: 'CSS', href: '#' },
+  { label: 'REACT', href: '#' },
+];
+
+const customTheme = deepMerge(grommet, {
+  anchor: {
```
ericsoderberghp: This seems orthogonal to Nav, more about customizing Anchor.

src/js/components/Nav/stories/CustomChild.js
ShimiSun: I've replaced it with the name `Sidebar`, it doesn't make sense to have `Custom child` when we don't have `items`.

src/js/components/Nav/stories/CustomAnchor.js
ShimiSun: yet, I think it will be a helpful example. I can remove it if you strongly think it should be placed elsewhere. (I didn't find a Custom Anchor example on Anchor.)

ShimiSun:
linking to #3580 #3489

PR #3627, opened 2019-12-18
Make styled-components external.

ericsoderberghp:
#### What does this PR do?

Make styled-components external.

grommet-designer using react, react-do, styled-components, grommet, and grommet-icons all loaded via `<script />` tags.

no

yes

backwards compatible

### DIFF
--- a/webpack.config.babel.js
+++ b/webpack.config.babel.js
@@ -23,6 +23,7 @@ export default {
   externals: {
     react: 'React',
     'react-dom': 'ReactDOM',
+    'styled-components': 'styled',
   },
   resolve: {
     extensions: ['.js', '.json'],


webpack.config.babel.js
```diff
@@ -23,6 +23,7 @@ export default {
   externals: {
     react: 'React',
     'react-dom': 'ReactDOM',
+    'styled-components': 'styled',
```
ShimiSun: shouldn't this be `'styled-components': 'styled-components'?

Also, I noticed the following addition and thought I'll share it for awareness:
https://github.com/styled-components/styled-components/issues/1770#issuecomment-394101268

ericsoderberghp: The correct variable name is `styled`, as in `window.styled`, which aligns with the default export name.

I think that comment is just if using `npm link`.

Issue #3579, opened 2019-12-03
Select options vanish when the page has been scrolled down


ryanjchr:
### Expected Behavior
A Select dropdown that is open should show its options in the dropdown no matter the user's Y-Scroll position on the page.

### Actual Behavior
As of Grommet 2.8.1, when the user has scrolled a page down a certain amount and opens a Select dropdown there will be no options in the dropdown. Likewise, if a user opens a Select while not scrolled too far down the page, the options will appear, but if the user scrolls down while the options are visible they will disappear once the page scrolls past a certain threshold.

### URL, screen shot, or Codepen exhibiting the issue
This CodeSandbox requires this accompanying GIF demonstrating the issue.

![grommet-select-bug](https://user-images.githubusercontent.com/8773259/70026243-6725fb00-15f3-11ea-83d2-4558b3d6a4fa.gif)

https://codesandbox.io/s/select-long-page-grommet-generic-xkouw

### Steps to Reproduce
1. Create any Select component on a page that can scroll on the Y axis.
2. Open the React app in any browser and shrink the window on the Y axis somewhat (see the gif in the section above)
3. Open the Select dropdown to view the options
4. Scroll the page up and down and notice that the options disappear/reappear from the Select dropdown when scrolling the page down and up

### Your Environment
- Grommet version: 2.8.1
- Browser Name and version: Firefox 70.0.1 and Chrome 78
- Operating System and version (desktop or mobile): macOS

felipebrunetti94:
is this one up for grab ?

emmerich:
If it helps, I believe this was introduced in 2.8.1. Rolling back to 2.8.0 fixes the problem.

ShimiSun:
Thanks @emmerich , can you confirm if this is the commit https://github.com/grommet/grommet/commit/9d1b60527db18af6de090b0312d61201851f7a19?

litewarp:
I think the issue is with the use of React.memo to wrap the SelectOption. It looks like the commit moved the old props comparison from shouldComponentUpdate to the second argument for React.Memo. But those are not equivalent functions. As the React Docs note, [memo shouldn't be used to prevent a render](https://reactjs.org/docs/react-api.html#reactmemo)

Specifically though, I think what is happening is that because the option only updates when the dependency array has changed, and in this case, the dependency array is one boolean value, the options are not updating correctly. 

[...]

PR #3637, opened 2019-12-30
fix: sort out an edge case where search dropdown goes invisible

### LINKED ISSUE #3636: Select dropdown options become empty while in a scrollable modal due to InfiniteScroll
### Expected Behavior
Select dropdown with search enabled becomes empty when its placed in a Layer that has scroll in it and body has scroll enabled.

### Actual Behavior
Select dropdown with search enabled and placed inside a modal with fewer options that results in no scroll, InfiniteScroll.js seems to compute **nextBeginPage** & **nextEndPage** to be same (1 in my case) which results in array slicing to be off and becomes empty array. 

- Grommet version: v2.9.0

aga5tya:
Signed-off-by: aga5tya <aga5tya@yahoo.com>

#### What does this PR do?
Fixes #3636


ShimiSun:
LGTM. @ericsoderberghp I'd appreciate your review as well.

src/js/components/InfiniteScroll/InfiniteScroll.js
```diff
@@ -113,6 +113,9 @@ const InfiniteScroll = ({
         ),
       );
 
+      // Do not update the next if both are same
+      if (nextBeginPage === nextEndPage) return;
```
ericsoderberghp: Suppose the pageHeight is (2 * height), beginPage is 0, and endPage is 1. As the user scrolls down, there is a point in time where nextBeginPage will be 1 and nextEndPage will be 1. At this point, we should be updating beginPage to be 1 and leaving endPage at 1. With the proposed change above, that will not happen and beginPage will remain at 0.

I am trying to understand what the issue is. Can you provide a code example that demonstrates the issue?

aga5tya: I could be in the wrong direction. 

The issue is happening in my specific case, where the search drop down in inside a scrollable modal. In my case the `items` array of size 6 and step defaulting at 20, Initially, `beginPage` is 0 and `endPage` is at 1 and loads fine, InfiniteScroll triggers an update and now the `beginPage` & `endPage` are arrived to be 1, due to this the items slicing goes for a toss on line below because `firstIndex` goes outside boundary array and results in them being empty.

https://github.com/grommet/grommet/blob/626f1e345d1aa3096748ce7c96180dfb96f2e61f/src/js/components/InfiniteScroll/InfiniteScroll.js#L168

aga5tya: @ericsoderberghp  Here's a minimum reproduction on [codesandbox](https://codesandbox.io/s/dreamy-elgamal-obbhm?fontsize=14&hidenavigation=1&theme=dark) - https://obbhm.csb.app/

On the full desktop view
 - Without scrolling, click the toggle on right corner, now the options show up properly
 - Now hide layer, scroll the document to bottom, and toggle the layer again. This time, the options disappear.

This is happening because InfiniteScroll looks upto document as its scroll parent and hence the mismatch. Just curious, wouldn't `beginPage` being same as `lastPage` result in wrong slice of array?

aga5tya: @ericsoderberghp  Any update over this?

aga5tya:
Anything specific holding this ? or kindly suggestive alternative. I tried rolling back to older version, but seems to have the same issue.

patricknick:
Any update to this? Just ran into this issue as well and had to downgrade to 2.8.0.

src/js/components/InfiniteScroll/InfiniteScroll.js
ericsoderberghp: Apologies for the way too long delay on this. I just tried the codesandbox link provided above and couldn't seem to reproduce it. I couldn't scroll anywhere. I will try to reproduce still, just wanted to comment on the sandbox.

ericsoderberghp:
I think I have isolated the root of the problem. We are not calculating the lastPage correctly. Currently, it is: `Math.ceil(items.length / step)` but it should be `Math.floor(items.length / step)`. If the length was a factor of step, it was working fine. Making this change fixed the issue in the way I was reproducing it. It would be great if anyone else who has a means of reproducing it could try this fix.

ShimiSun:
Might be related to https://github.com/grommet/grommet/pull/3784

patricknick:
@ericsoderberghp I pulled the current master branch and was able to reproduce the issue in the Grommet storybook.

I used this code in `src/js/components/Select/stories/Basics.js`. I saw the behaviour described in #3579 .
```
const SimpleSelect = ({ theme, ...rest }) => {
  const [value, setValue] = useState('');
  return (
    <Box align="center" justify="center" pad="large" height="1280px">
      <Select
        id="select"
        name="select"
        placeholder="Select"
        value={value}
        options={options}
        {...rest}
      />
    </Box>
  );
};
```

When changing to `Math.floor`, the options stayed and didn't vanish. 👍

IanKBovard:
im unable to reproduce the issue. it looks like pr #3784 fixed this as well.

ShimiSun:
Thank you for this PR and contribution! the issue was solved with another PR effort, I'll be closing this PR then

PR #3640, opened 2020-01-02
fix: remove conflicting position style while layer is responsive

### LINKED ISSUE #3639: Position of modal container is in conflict when responsive flag is enabled on Layer component
When on mobile responsive view, the modal container does not show up if the user has scrolled to some extent already. This happens due to the container position being relative while it is expected to be fixed.

Also, this could be browser specific issue as it seems to work as expected in firefox not able to reproduce in codesandbox. Given the styles on container, making container position to relative does not make sense to me.

### Expected Behavior

The modal should show work irrespective of the scroll position while on mobile with responsive on Layer enabled.

### Actual Behavior

If the user has significantly scrolled, the modal does not show up.

### Your Environment

- Grommet version: v2.9.0
- Operating System and version (desktop or mobile): Desktop Responsive Mode

aga5tya:
Fixes #3639


IanKBovard:
@aga5tya  Looks good to me!

ShimiSun:
I'm suspecting this fix might cause breaking changes for some use cases that rely on the `position: relative` value. I'm wondering if there is a more generic way to do it. @ericsoderberghp please share your thoughts here.

MikeKingdom:
@ShimiSun I haven't been able to reproduce this issue either. My opinion is we should close this PR and issue unless someone has an example that can show the problem.

ShimiSun:
@MikeKingdom Thanks.  Can you please drop a comment on the issue itself? we'll close this one as well.

PR #3643, opened 2020-01-03
fix exiting calendar with tab

### LINKED ISSUE #3609: Calendar traps keyboard users
### Expected Behavior

My personal opinion is that tabbing once into the calendar sets focus on the calendar, and then from there the arrow keys navigate through days. A second tab should then tab the keyboard user out of the calendar.
However, minimal expected behavior is that there should be _a_ way to get out of the calendar with a keyboard.

### Actual Behavior

Currently, once you start tabbing into the days of the calendar, you're basically trapped inside the calendar with no way to navigate outside of the calendar, as the tab key also navigates between days, and tabbing to the next month's days switches months, essentially trapping a keyboard user. While it definitely could be fixed by passing in the showAdjacentDays={false} prop, this kind of doesn't look that great if you don't have a transparent background, as the height doesn't decrease when hiding adjacent days.

Below is a gif of being stuck in the calendar pressing tab:
trappedInTheCalendar

Below is an image of the calendar with showAdjacentDays set to false, with the height not adjusting:
image (1)

Here's a Codesandbox demonstrating the issue: https://codesandbox.io/s/pedantic-zhukovsky-5fhk6?fontsize=14&hidenavigation=1&theme=dark

### Steps to Reproduce

1. Tab into the Calendar 
2. Tab through the days 
3. Try and tab to the button below the calendar (in my codesandbox)

### Your Environment

- Grommet version: 2.8.1
- Browser Name and version: Chrome version 78.0.3904.108
- Operating System and version (desktop or mobile): Windows desktop

felipebrunetti94:
#### What does this PR do?
fixes #3609 
#### Screenshots 
Gravação de tela de 03-01-2020 11_33_07

### DIFF
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -112,7 +112,9 @@ class Calendar extends Component {
     return null;
   }
 
-  state = {};
+  state = {
+    focusOnCalendar: false,
+  };
 
   dayRefs = {};
 
@@ -188,6 +190,7 @@ class Calendar extends Component {
   onFocus = day => () => {
     const { bounds } = this.props;
     const { reference } = this.state;
+    this.setState({ focusOnCalendar: true });
     if (betweenDates(day, bounds)) {
       this.setState({ focused: day }, () => {
         if (day.getMonth() !== reference.getMonth()) {
@@ -308,7 +311,16 @@ class Calendar extends Component {
       theme,
       ...rest
     } = this.props;
-    const { date, dates, focused, start, reference, end, slide } = this.state;
+    const {
+      date,
+      dates,
+      focused,
+      start,
+      reference,
+      end,
+      slide,
+      focusOnCalendar,
+    } = this.state;
 
     // We have to deal with reference being the end of a month with more
     // days than the month we are changing to. So, we always set reference
@@ -363,7 +375,10 @@ class Calendar extends Component {
               disabled={dayDisabled}
               onClick={this.onClickDay(dateString)}
               onFocus={this.onFocus(day)}
-              onBlur={() => this.setState({ focused: false })}
+              onBlur={() =>
+                this.setState({ focused: false, focusOnCalendar: false })
+              }
+              tabIndex={focusOnCalendar ? -1 : 0}
             >
               <StyledDay
                 inRange={inRange}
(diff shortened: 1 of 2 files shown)


ericsoderberghp:
Thanks for tackling this issue. My inclination is to resolve this by setting the tabIndex of each day to -1, to prevent tab navigation between days and leave tab navigation to only be used to focus the calendar as a whole. This would align better with List (https://storybook.grommet.io/?path=/story/list--onclickitem) and how the keyboard interaction works with drop downs like Select. Thoughts?

felipebrunetti94:
@ericsoderberghp , huumm... I think it's a good solution, I haven't seen the code from the Select and Dropdown components.
The problem of changing every day to tabIndex: -1, is that you can't enter the calendar with tab, by changing the tabIndex, only when focused we resolve this problem...
Will give it a try. :smile:

ericsoderberghp:
> @ericsoderberghp , huumm... I think it's a good solution, I haven't seen the code from the Select and Dropdown components.
> The problem of changing every day to tabIndex: -1, is that you can't enter the calendar with tab, by changing the tabIndex, only when focused we resolve this problem...
> Will give it a try. 😄

I would give the Calendar itself a tabIndex, so it can receive focus, and then manage the active day via keyboard arrows while the calendar has focus.

felipebrunetti94:
>I would give the Calendar itself a tabIndex, so it can receive focus, and then manage the active day via keyboard arrows while the calendar has focus.
Seems good to me, much more simple kk Doing the changes right now. Thanks for the hint :smile: .

felipebrunetti94:
Just tried experimenting, only with the tabIndex: 0 in the container and the buttons with index -1, and doesn't work, it focus only in the container, and not the buttons, that way we can't navigate with the arrows. 

Gravação de tela de 10-01-2020 10_34_32

ericsoderberghp:
I was actually thinking the tabIndex should be on `StyledWeeks`.

felipebrunetti94:
The problem with a StyledWeeks with tabIndex, is that the tab will move around the weeks until exit the calendar.

ericsoderberghp:
I think this has been superseded by https://github.com/grommet/grommet/pull/3692

ShimiSun:
Thank you for the PR! the issue had been fixed. Feel free to reopen or file a new PR if needed.

Issue #3578, opened 2019-12-03
TypeScript: ExtendType does not accept styled-components css helper


JasonRoss-bt:
### Expected Behavior

Previously we were able to use the `css` helper for `extend`ing components in the theme; however, the new typed theme does not seem to allow for it.

```typescript
const theme: ThemeType = {
  ...grommet,
  anchor: {
    ...grommet.anchor,
    extend: css`
      background-color: red;
    `
  }
};
```

### Actual Behavior

TypeScript build fails with:

```
(property) extend?: ExtendType
Type 'FlattenSimpleInterpolation' is not assignable to type 'ExtendType'.
  Type 'FlattenSimpleInterpolation' is not assignable to type 'string'.ts(2322)
base.d.ts(235, 5): The expected type comes from property 'extend' which is declared here on type '{ color?: ColorType; extend?: ExtendType; fontWeight?: number; hover?: { extend?: ExtendType; textDecoration?: string; }; textDecoration?: string; }'
```

### URL, screen shot, or Codepen exhibiting the issue

Codepen: https://codesandbox.io/s/grommet-typescript-theme-extend-issue-cbpo8

### Your Environment

- Grommet version: 2.8.1
- React version: 16.8.4
- TypeScript version: 3.7.2
- styled-components version: 4.4.1

oorestisime:
Do you know which version was working?

ShimiSun:
hi buddy, @JasonRoss-bt, care to make a contribution and extend the `ExtendType`? and should go here https://github.com/grommet/grommet/blob/e6bac4512e0ae18fdfce17ba65f4a7d920f1000a/src/js/themes/base.d.ts#L18

JasonRoss-bt:
@ShimiSun Thanks for pointing me to the right file... saved me some time :)

I'm working on a fix and hope to have a PR soon.

ShimiSun:
Thank you again for your contribution! ❤️

PR #3647, opened 2020-01-06
Update ExtendType to support styled-components interpolations

jasnross:
#### What does this PR do?

Adds TypeScript support for using styled-components interpolations as `extend` values in the `ThemeType`, fixing #3578.

src/js/themes/base.d.ts
src/js/all/stories/typescript/ExtendTheme.tsx

* Created a TypeScript story to type-check various usages of `extend` in a custom theme.
* Updated my own project using locally linked `grommet` and confirmed our custom theme now passed type-checking

* Ensure that `yarn tsc --project .` reports no type-checking errors for the ExtendTheme story.

1) I added a utility type called `PropsOf` which makes use of a TypeScript 2.8 feature called conditional types. This means that TypeScript Grommet users should be using TypeScript >= 2.8, which I think is a reasonable tradeoff. If one of the goals of Grommet's TypeScript support is to be compatible with TypeScript pre-2.8 let me know and I can come up with another solution to cover what that utility type does.

2) I spent some time also investigating the possibility of adding more types for the `extend` props, which could be convenient for TypeScript users since they would get intellisense about what `props` are available when `extend`ing the theme; however, I couldn't see a 100% reliable way of determining which props would be passed to the `styled` version of the component, so most of the extend props will be typed as `Record<string, any>` instead. It should be possible to progressively type more of the props given the new definition of `ExtendType`, but if you have any suggestions on how to tell with accuracy which props will be available in `extend` I'd be happy to look into adding more of them.

3) One other thing to note is that there is an open issue for conflicting react / react-native types at https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33311/. To work around this I've added a .yarnclean to remove `@types/react-native` post-install. There are a few other alternatives to work around this problem suggested in that issue, and if desired we can take another approach:

* Fix @types/styled-components version to an older version
* Declare types explicitly in tsconfig.json

Overall I thought the .yarnclean seemed like the most straightforward way, so long as this project is not planning to make use of `@types/react-native` at this time

#3578 

I don't believe any updates are necessary.

Potentially worth mentioning as a bug fix, but probably not worth drawing a large amount of attention to it or anything :)

Backward compatible for TypeScript >= 2.8.

### DIFF
--- a/.yarnclean
+++ b/.yarnclean
@@ -1,3 +1,0 @@
-# Working around styled-components types issue as suggested in
-# https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33311/#issuecomment-546727329
-@types/react-native
--- a/.yarnclean
+++ b/.yarnclean
@@ -1,0 +1,3 @@
+# Working around styled-components types issue as suggested in
+# https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33311/#issuecomment-546727329
+@types/react-native
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "@types/react": "^16.9.11",
     "@types/react-dom": "^16.9.3",
     "@types/storybook__react": "^4.0.2",
-    "@types/styled-components": "^4.4.2",
+    "@types/styled-components": "^5.0.1",
     "awesome-typescript-loader": "^5.2.1",
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^10.0.1",
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "@types/react": "^16.9.11",
     "@types/react-dom": "^16.9.3",
     "@types/storybook__react": "^4.0.2",
+    "@types/styled-components": "^4.4.2",
     "awesome-typescript-loader": "^5.2.1",
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^10.0.1",
--- a/src/js/all/stories/typescript/ExtendTheme.tsx
+++ b/src/js/all/stories/typescript/ExtendTheme.tsx
@@ -1,0 +1,62 @@
+import React from 'react';
+import { css } from 'styled-components';
+import { storiesOf } from '@storybook/react';
+import { ThemeType, BoxTypes, Grommet, Box, Anchor, Button } from 'grommet';
+
+// Custom theme to verify that various `extend` types work correctly
+const custom: ThemeType = {
+  box: {
+    extend: css`
+      cursor: ${(props: BoxTypes) => (props.onClick ? 'cursor' : 'inherit')};
+    `,
+  },
+  anchor: {
+    extend: props => css`
+      color: ${props.href ? 'green' : 'red'};
+    `,
+  },
+  button: {
+    extend: props => {
+      let extraStyles = '';
+      if (props.primary) {
+        extraStyles = `
+          text-transform: uppercase;
+          color: white;
+        `;
+      }
+      return `
+        color: green;
+        font-size: 12px;
+        font-weight: bold;
+
+        ${extraStyles}
+      `;
+    },
+  },
+  icon: {
+    // Components without typed theme props should pass type-checking
+    extend: props => css`
+      color: ${props.untypedProp === 'some value' ? 'red' : 'green'};
+    `,
+  },
+};
+
+const ExtendTheme: React.FC = () => {
+  return (
+    <Grommet theme={custom}>
+      <Box pad="small" gap="medium" width="medium">
+        <Anchor href="#">Anchor (href)</Anchor>
+        <Anchor>Anchor (no href)</Anchor>
+
+        <Button label="Standard Button" onClick={() => {}} />
+        <Button label="Primary Button" primary onClick={() => {}} />
+
+        <Box onClick={() => {}}>Clickable box should use pointer cursor.</Box>
+
+        <Box>Un-clickable box should use standard cursor.</Box>
+      </Box>
+    </Grommet>
+  );
+};
+
+storiesOf('TypeScript/Theme', module).add('Extend', () => <ExtendTheme />);
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -21,8 +21,8 @@ import {
   PropsOf,
 } from '../utils';
 
+import { BoxProps } from '../components/Box';
 import { Anchor } from '../components/Anchor';
-import { Button } from '../components/Button';
 import { Box } from '../components/Box';
 import { Text, TextProps } from '../components/Text';
 import { ReactComponentElement } from 'react';
@@ -52,7 +52,7 @@ type ExtendValue<TProps> = string | FlattenSimpleInterpolation | FlattenInterpol
 type ExtendFn<TProps> = (props: ExtendProps<TProps>) => ExtendValue<TProps>;
 
 /**
- * ExtendType represents a valid value for `extend` in the theme.
+ * ExtendType represents the type for `extend` values in the theme.
  *
  * Acceptable values for `extend` are one of:
  *
@@ -112,7 +112,7 @@ type Colors = typeof colors & {
   'graph-3'?: ColorType;
   'graph-4'?: ColorType;
   'graph-5'?: ColorType;
-  [x: string]: ColorType | undefined;
+  [x: string]: ColorType;
 };
 
 interface ButtonKindType {
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -1,4 +1,10 @@
 import {
+  FlattenInterpolation,
+  FlattenSimpleInterpolation,
+  ThemedStyledProps,
+} from 'styled-components';
+
+import {
   BackgroundType,
   BorderType,
   BreakpointBorderSize,
@@ -12,10 +18,13 @@ import {
   OpacityType,
   RoundType,
   PadType,
+  PropsOf,
 } from '../utils';
 
-import { BoxProps } from '../components/Box';
-import { TextProps } from '../components/Text';
+import { Anchor } from '../components/Anchor';
+import { Button } from '../components/Button';
+import { Box } from '../components/Box';
+import { Text, TextProps } from '../components/Text';
 import { ReactComponentElement } from 'react';
 
 export declare const base: DeepReadonly<ThemeType>;
@@ -24,7 +33,34 @@ export declare const generate: (
   scale?: number,
 ) => DeepReadonly<ThemeType>;
 
-type ExtendType = string | ((...args: any) => void);
+/**
+ * ExtendProps represents the props that will be provided to an ExtendType.
+ */
+type ExtendProps<TProps> = ThemedStyledProps<TProps, ThemeType>;
+
+/**
+ * ExtendValue represents a valid `extend` value, which can be a CSS string or a
+ * styled-components interpolation. In the theme an ExtendValue can be provided
+ * directly to `extend` or it can be computed as the result of an ExtendFn.
+ */
+type ExtendValue<TProps> = string | FlattenSimpleInterpolation | FlattenInterpolation<ExtendProps<TProps>>;
+
+/**
+ * ExtendFn represents a function passed to `extend`. These functions receive
+ * props and produce an ExtendValue.
+ */
+type ExtendFn<TProps> = (props: ExtendProps<TProps>) => ExtendValue<TProps>;
+
+/**
+ * ExtendType represents a valid value for `extend` in the theme.
+ *
+ * Acceptable values for `extend` are one of:
+ *
+ * * A string containing css
+ * * An array of styled-components interpolations (usually returned by the styled-components `css` helper function)
+ * * A function taking props and returning one of the above values
+ */
+type ExtendType<TProps = Record<string, any>> = ExtendValue<TProps> | ExtendFn<TProps>;
 
 declare const colors: {
   active?: ColorType;
@@ -76,7 +112,7 @@ type Colors = typeof colors & {
   'graph-3'?: ColorType;
   'graph-4'?: ColorType;
   'graph-5'?: ColorType;
-  [x: string]: ColorType;
+  [x: string]: ColorType | undefined;
 };
 
 interface ButtonKindType {
@@ -277,10 +313,10 @@ export interface ThemeType {
   };
   anchor?: {
     color?: ColorType;
-    extend?: ExtendType;
+    extend?: ExtendType<PropsOf<typeof Anchor>>;
     fontWeight?: number;
     hover?: {
-      extend?: ExtendType;
+      extend?: ExtendType<PropsOf<typeof Anchor>>;
       textDecoration?: string;
     };
     textDecoration?: string;
@@ -923,10 +959,8 @@ export interface ThemeType {
       margin?: MarginType;
     };
     options?: {
-      container?: BoxProps;
-      text?: {
-        margin?: MarginType;
-      };
+      container?: PropsOf<typeof Box>;
+      text?: PropsOf<typeof Text>;
     };
     // https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37506
     searchInput?: ReactComponentElement<any>;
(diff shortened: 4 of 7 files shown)


ShimiSun:
@jasnross awesome, thanks buddy! 
I'm working now on upgrading to styled-components 5.0.0 I think we'll wait with this PR until the migration is done.

JasonRoss-bt:
@ShimiSun Cool, sounds good to me! Let me know once that's done and I can make sure to update this PR accordingly if needed.

jasnross:
@ShimiSun Cool, sounds good to me! Let me know once that's done (or if I can help in any way) and I can make sure to update this PR accordingly if needed.

ShimiSun:
@jasnross thanks buddy, styled-components upgrade is done, feel free to update this PR (merge with the master branch) so we can continue with its review. I'd love to get it in for our next release.

jasnross:
Thanks @ShimiSun! I have a bit of a busy week but I'm planning on getting to this before Monday.

jasnross:
@ShimiSun Took a little longer than I thought to get back to this, thanks for waiting! The PR is updated now.

ShimiSun:
@jasnross exciting PR, thanks for the contribution!
@dallas would appreciate your type-script savviness with reviewing :)

src/js/all/stories/typescript/ExtendTheme.tsx
```diff
@@ -0,0 +1,62 @@
...
+  },
+  icon: {
+    // Components without typed theme props should pass type-checking
+    extend: props => css`
+      color: ${props.untypedProp === 'some value' ? 'red' : 'green'};
+    `,
+  },
+};
+
+const ExtendTheme: React.FC = () => {
```
ShimiSun: LOVE this example, THANK YOU

dallas:
@ShimiSun, yeah, I’ll take a look when I get a chance. Probably won’t be until Tuesday (Feb. 18) at the earliest.

dallas:
@ShimiSun, the TypeScript all looks good to me. I haven’t had a chance to actually play with the project with these updates, though.

src/js/themes/base.d.ts
```diff
@@ -809,14 +833,9 @@ export interface ThemeType {
...
     options?: {
-      container?: {
-        align?: string,
-        pad?: string,
-      },
-      text?: {
-        margin?: MarginType,
-      },
-    },
+      container?: PropsOf<typeof Box>;
```
ShimiSun: I might not be fully understanding what `PropsOf` is doing here, but shouldn't it allow only `align` and `pad` instead of a whole type of Box?

src/js/themes/base.d.ts
```diff
@@ -809,14 +833,9 @@ export interface ThemeType {
...
+      text?: PropsOf<typeof Text>;
```
ShimiSun: I think we only allow `margin` for `drop.options.text` other text syles won't work.

package.json
```diff
@@ -83,6 +83,7 @@
     "@types/react": "^16.9.11",
     "@types/react-dom": "^16.9.3",
     "@types/storybook__react": "^4.0.2",
+    "@types/styled-components": "^4.4.2",
```
ShimiSun: Should the package.json also include constraint/enforcment for using typescript 2.8?

src/js/themes/base.d.ts
dallas: I think you’re right in assuming that `PropsOf<typeof Box>` is grabbing all of the props of `Box` and allowing them here. I wonder if we need to also do a `Pick` here:

```TypeScript
Pick<PropsOf<typeof Box>, 'align' | 'pad'>;
```

src/js/themes/base.d.ts
dallas: Could use `Pick` here as well, I think.

package.json
dallas: Specifically 2.8 and 2.8 only, or 2.8 or greater?

ShimiSun: I would think greater than 2.8, right @jasnross?

src/js/themes/base.d.ts
ShimiSun: Or, we can just import PadType from "grommet/utils", and use it instead.
```
import { PadType } from 'grommet/utils';
```
Sounds right?

jasnross:
Thanks for the comments, much appreciated! I’ll take a closer look and respond at some point in the next couple days.

package.json
JasonRoss-bt: It's definitely a good idea, but I'm not aware of a way to do so yet, although there is some discussion about it [here](https://github.com/microsoft/TypeScript/issues/32166). 

Do you know of an existing way to enforce the minimum TypeScript version required by consuming packages?

ShimiSun: I've been thinking about it and since typescript isn't peerDependency and grommet is installing it as part of its devDependencies we should be good with it as it.

src/js/themes/base.d.ts
JasonRoss-bt: Sounds good, I can update that to one of the above... I think I got the idea that it could take any of the Box props from the sentence in the docs here, although now I do see the example only mentioned align and pad:

image

JasonRoss-bt:
Thanks again for the feedback and for waiting for me to get back to this. I think all the feedback should be addressed now.

src/js/themes/base.d.ts
ShimiSun: my bad, I think you are right! I checked the code and select.options.container can accept any Box prop, align and pad are just an example.

src/js/themes/base.d.ts
```diff
@@ -22,7 +33,20 @@ export declare const generate: (
   scale?: number,
 ) => DeepReadonly<ThemeType>;
 
-type ExtendType = string | ((...args: any) => void);
+type ExtendProps<TProps> = ThemedStyledProps<TProps, ThemeType>;
+type ExtendValue<TProps> = string | FlattenSimpleInterpolation | FlattenInterpolation<ExtendProps<TProps>>;
+type ExtendFn<TProps> = (props: ExtendProps<TProps>) => ExtendValue<TProps>;
```
ShimiSun: can we add a comment for each type and explain its purpose and value?

src/js/themes/base.d.ts
```diff
@@ -812,8 +836,8 @@ export interface ThemeType {
     };
     options?: {
       container?: {
-        align?: string;
-        pad?: string;
+        align?: BoxProps['align'];
+        pad?: BoxProps['pad'];
```
ShimiSun: my bad, let's roll it back to how it was.

src/js/themes/base.d.ts
JasonRoss-bt: Cool, no worries! It's back to using PropsOf again

src/js/themes/base.d.ts
JasonRoss-bt: Sounds good, I've added some comments. Hope I explained well enough, let me know if anything doesn't make sense.

src/js/themes/base.d.ts
JasonRoss-bt: Thanks, all set :)

jasnross:
Hey @ShimiSun I just pushed a commit to update to @types/styled-components@5.0.1 to align with the new version of styled-components.

Just checking, is there anything I missed that is needed for this to be merged?

ShimiSun:
@jasnross thanks buddy, I need to dive into this PR again. are you on the grommet slack where I can reach out easily?

jasnross:
@ShimiSun I didn't have Grommet open in the desktop Slack client, but now I do so I'll be able to get messages there without opening the web app. Feel free to reach out!

.yarnclean
```diff
@@ -0,0 +1,3 @@
+# Working around styled-components types issue as suggested in
+# https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33311/#issuecomment-546727329
+@types/react-native
```
jasnross, line 3: @ShimiSun Not sure if there were concerns about this workaround or not, but one alternative to this could be to set the following in tsconfig.json:

```
"types": [],
"lib": ["dom", "es6", "es2016"],
```

My understanding is that this way TypeScript will not implicitly load all type definitions, but only ones listed in "types".

* Set types to  `[]` so no types are loaded implicitly
* When a type is needed we `import` it instead (this is what is already being done in this project, as well as what I do in my own projects)
* Add in the ECMAScript and DOM APIs that we support in "lib" since they are no longer loaded implicitly

ShimiSun:
@negreirosleo I will appreciate another pair of eyes on this one.

jasnross:
@ShimiSun Any updates on this one?

For what it's worth, we're using [patch-package](https://www.npmjs.com/package/patch-package) to patch Grommet with the types from this PR and they've been working great in our own project since around the end of April.

.yarnclean
jasnross: Thinking about this a little more, I do like the solution of setting types to `[]` a little better than relying on yarn to remove the @types/react-native package after install.

If there are no concerns with that I can update this PR with that change.

src/js/themes/base.d.ts
```diff
@@ -22,7 +33,34 @@ export declare const generate: (
   scale?: number,
 ) => DeepReadonly<ThemeType>;
 
-type ExtendType = string | ((...args: any) => void);
+/**
+ * ExtendProps represents the props that will be provided to an ExtendType.
```
carolineartz, line 37: I see how this all works and its great. 

i'm wondering if we can adjust the wording of the comments slightly around "value" and "type"...currently there is both

> ExtendType represents a valid value for `extend` in the theme

and 

> ExtendValue represents a valid `extend` value


I realize that this does make sense since an `ExtendType` may directly be an `ExtendValue`, but I still wonder if this can be clarified a bit. Perhaps changing 

`ExtendType` represents ~~a valid value~~ [the type] for `extend` values in the theme.

might be sufficient.

src/js/utils/index.d.ts
```diff
@@ -34,6 +34,19 @@ declare const removeUndefined: <T extends object>(obj: T) => NonUndefinedProps<T
...
 
+/*
+ * Utility type for inferring the props type of a component.
+ *
+ * Example:
+ *
+ * ```typescript
+ * import { Box } from 'grommet';
+ *
+ * type BoxProps = PropsOf<typeof Box>;
```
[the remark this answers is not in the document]
jasnross: Good point! Let me see if I can adjust the example to be more clear here.

One thing I've been wondering is if ultimately the props types that Grommet provides (like `BoxProps`) should be updated to include all of the props including intrinsic props (typically this is what I'm used to in my own projects). I'm not sure if there was a historical reason why it was done that way nor am I sure if it would be a breaking change for some TS users who rely on it that way though. Probably outside the scope of this PR but could be worth considering in another one.

src/js/themes/base.d.ts
```diff
@@ -12,8 +18,13 @@ import {
   OpacityType,
   RoundType,
   PadType,
+  PropsOf,
 } from '../utils';
-import { TextProps } from '../components/Text';
+
+import { Anchor } from '../components/Anchor';
+import { Button } from '../components/Button';
```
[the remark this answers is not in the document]
jasnross: Nice catch! I've removed the import.

RE: ExtendType type param

Originally I was thinking about adding the type param everywhere but after some investigation I found that in many components a different set of props are passed to the `styled` component. For example, in [Button](https://github.com/grommet/grommet/blob/1cf786ddafe7bf3329f9a37bb0bc51ebac4b2ec7/src/js/components/Button/Button.js#L280) it looks like a prop called `sizeProp` is passed instead of `size`, etc.

I wasn't sure yet how to handle that, so I changed my mind about adding all the types until it's clearer how we can tell which props will be present when the extend function is called, that way the types aren't saying that props will be present that won't really be there.

To simplify things I thought about just removing the type param altogether, which would leave the props typed as `ThemeStyledProps<any, ThemeType>`, and leaving it up to callers to determine what props are available, but it seemed useful to progressively add more typing to these in the future.

What do you think?

ShimiSun:
@jasnross huge thanks for working on this.
@carolineartz I'll appreciate another round of review to get this PR in for our next release!

jasnross:
> @jasnross huge thanks for working on this.
> @carolineartz I'll appreciate another round of review to get this PR in for our next release!

Sorry pushed a little too soon, still working on the feedback but I'll have the rest updated shortly

src/js/themes/base.d.ts
jasnross: Took your suggestion here and I think it's a little clearer. Thanks for the feedback!

jasnross:
> > @jasnross huge thanks for working on this.
> > @carolineartz I'll appreciate another round of review to get this PR in for our next release!
> 
> Sorry pushed a little too soon, still working on the feedback but I'll have the rest updated shortly

Ok I think this is ready for another look now. Thanks again :)

src/js/themes/base.d.ts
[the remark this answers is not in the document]
carolineartz: I was torn on this whole bit too, went back and forth but everything i could think of came with additional questions. I think its good for now!

src/js/utils/index.d.ts
```diff
@@ -34,6 +34,19 @@ declare const removeUndefined: <T extends object>(obj: T) => NonUndefinedProps<T
 
 export {isObject, deepFreeze, deepMerge, removeUndefined};
 
+/*
+ * Utility type for inferring the props type of a component.
+ *
+ * Example:
+ *
+ * ```typescript
```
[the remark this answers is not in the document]
jasnross: I usually include them since a lot of editors will pick up the formatting differently when you hover over values to see the documentation:

image

It doesn't seem to pick them up for the .d.ts files in this project for some reason, but I suppose it's worth keeping in case it gets moved around.

src/js/all/stories/typescript/ExtendTheme.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <Grommet theme={custom}>
+      <Box pad="small" gap="medium" width="medium">
+        <Anchor href="#">Anchor (href)</Anchor>
+        <Anchor>Anchor (no href)</Anchor>
+
+        <Button label="Standard Button" onClick={() => {}} />
+        <Button label="Primary Button" primary onClick={() => {}} />
```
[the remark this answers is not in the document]
jasnross: This was to test that we can define extra styles and that it passes type checking. I wanted to have a few working examples rather than just have a theme definition alone, so figured Buttons were a good test of that.

src/js/themes/base.d.ts
```diff
@@ -76,7 +112,7 @@ type Colors = typeof colors & {
   'graph-3'?: ColorType;
   'graph-4'?: ColorType;
   'graph-5'?: ColorType;
-  [x: string]: ColorType;
+  [x: string]: ColorType | undefined;
```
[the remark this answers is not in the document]
jasnross: Huh you're right, nice catch. Maybe that wasn't there when I first added this?

Either way it's removed now! :)

package.json
```diff
@@ -81,6 +81,7 @@
     "@types/react": "^16.9.11",
     "@types/react-dom": "^16.9.3",
     "@types/storybook__react": "^4.0.2",
+    "@types/styled-components": "^5.0.1",
```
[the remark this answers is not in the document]
jasnross: We don't need styled-components itself as a depedency, but we do need the types so that we can import them into the d.ts files otherwise the TypeScript build would fail.

We don't need it as a peerDependency though since it's not required that consumers install the types for grommet to work.

jasnross:
Great feedback @carolineartz :) I replied to a few comments and made a couple updates. I think we're probably good to go but let me know if you had any other concerns here.

ShimiSun:
@carolineartz last round of review on this gem?

src/js/utils/index.d.ts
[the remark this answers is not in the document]
carolineartz: ahhh very good point! didn't think of that

carolineartz:
this looks awesome, thanks so much for tackling this :)

ShimiSun:
Thank you @jasnross for your contribution and your patient, you'll make the typescript community super happy with this fix 🙌 ❤️ 
@carolineartz thank you for your fantastic collaboration on this!

PR #3648, opened 2020-01-06
Allow tabindex to be restored to original value

taysea:
#### What does this PR do?
This PR fixes an issue that was occurring when FocusContainer was changing the tabindex of elements outside of the container. Essentially, certain elements were toggled to a tabIndex of "-1" but their original tab index was 0 or a value specified by the user. Then, when the function of "makeNodeFocusable" was called to restore their original tabIndex, these elements were not being included.

src/js/utils/DOM.js

I tested other Drop examples to make sure the keyboard functionality had not changed.

Using storybook, add a focusable element after the TextInput Custom story. 

Before my changes: Tab through the elements once (TextInput and your focusable element should receive focus). Then, type into the text input to render the suggestions and tab to leave (You'll notice that your focusable element no longer receives focus because its tabindex is locked at -1).

After my changes: The focusable element will have its tabindex returned back to 0 after the textinput suggestions and tabbing through the elements will work as expected.

After tabIndex was changed on focusable elements when the TextInput drop opened (FocusContainer was handling the modulation of the tabIndex), there was no way for the tab index to be restored. Even though we were already storing the original tab index value in something called 'data-g-tabindex', these elements were not being included in having their tab index restored from -1 to their original value. This fixes that.

N/A

No.
No.

### DIFF
--- a/src/js/utils/DOM.js
+++ b/src/js/utils/DOM.js
@@ -13,8 +13,12 @@ export const filterByFocusable = elements =>
     if (currentTag === 'svg' || currentTag === 'div') {
       return (
         isValidTag &&
-        element.hasAttribute('tabindex') &&
-        element.getAttribute('tabindex') !== '-1'
+        ((element.hasAttribute('tabindex') &&
+          element.getAttribute('tabindex') !== '-1') ||
+          // If element was previously focusable, we want to be able
+          // to access it in makeNodeFocusable
+          (element.hasAttribute('data-g-tabindex') &&
+            element.getAttribute('data-g-tabindex') === '0'))
       );
     }
     return isValidTag;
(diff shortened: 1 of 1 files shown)


src/js/utils/DOM.js
```diff
@@ -13,8 +13,12 @@ export const filterByFocusable = elements =>
     if (currentTag === 'svg' || currentTag === 'div') {
       return (
         isValidTag &&
-        element.hasAttribute('tabindex') &&
-        element.getAttribute('tabindex') !== '-1'
+        ((element.hasAttribute('tabindex') &&
+          element.getAttribute('tabindex') !== '-1') ||
+          // If element was previously focusable, we want to be able
+          // to access it in makeNodeFocusable
+          (element.hasAttribute('data-g-tabindex') &&
```
ericsoderberghp: Hmm... This doesn't match the name and intent of the function anymore, since elements without tab index aren't focusable.

ericsoderberghp:
Superseded by https://github.com/grommet/grommet/pull/3650

PR #3651, opened 2020-01-08
Changed Box to add border='between'

ericsoderberghp:
#### What does this PR do?

Changed Box to add border='between'.
This makes it easy to add visual separators between children of Box.

Box/doc.js

Enhanced Box Border story

Storybook

yes

yes

backwards compatible


src/js/components/Box/README.md
```diff
@@ -279,6 +279,7 @@ right
 horizontal
 vertical
 all
+between
 {
   color: 
     string
```
negreirosleo: If it has the possibility of `border.side === between` , I think would be a good Idea to add it here aswell

negreirosleo:
```
....
all
between
{
  color: 
    string
    {
      dark: string,
      light: string
    },
  side: 
    top
    left
    bottom
    right
    horizontal
    vertical
    all
    between
  size: 
...
```

ShimiSun: https://github.com/grommet/grommet/blob/a3c75c2310ac789120b137f6310a432ffb98fb92/src/js/components/Box/doc.js#L40

src/js/components/Box/doc.js
```diff
@@ -168,10 +168,11 @@ export const doc = Box => {
         'horizontal',
         'vertical',
         'all',
+        'between',
       ]),
       BORDER_SHAPE,
       PropTypes.arrayOf(BORDER_SHAPE),
-    ]).description('Include a border.'),
+    ]).description('Include a border. You must have "gap" to use "between"'),
```
ShimiSun: I wonder if we can add some explanation about `between`, something like 'between will place a border between child elements' or '...where a gap is applied'.

src/js/components/Box/README.md
ericsoderberghp: agreed

src/js/components/Box/doc.js
ericsoderberghp: agreed, done

PR #3652, opened 2020-01-08
Fix how low opacity backgrounds affect text color

taysea:
#### What does this PR do?
This PR fixes how low opacity backgrounds control the text color that appears on them.

src/js/utils/background.js

Stories were added to test how this would affect text on light and dark backgrounds.

Use stories Box Background.

When low opacities were used for backgrounds, the background behind them should drive if text renders as light or dark. This currently wasn't being done, so in some cases, text on dark backgrounds was being rendered dark.

N/A

No.

No.

Yes, backwards compatible.

### DIFF
--- a/src/js/utils/background.js
+++ b/src/js/utils/background.js
@@ -113,8 +113,8 @@ export const backgroundStyle = (backgroundArg, theme, textColorArg) => {
       `;
     }
     const backgroundColor = normalizeColor(background, theme);
-    const backgroundDark = colorIsDark(backgroundColor);
     if (backgroundColor) {
+      const backgroundDark = colorIsDark(backgroundColor);
       return css`
         background: ${backgroundColor};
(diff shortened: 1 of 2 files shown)


src/js/utils/background.js
```diff
@@ -117,8 +117,12 @@ export const backgroundStyle = (backgroundArg, theme, textColorArg) => {
       return css`
         background: ${backgroundColor};
         color: ${normalizeColor(
-          textColor[colorIsDark(backgroundColor) ? 'dark' : 'light'] ||
-            textColor,
+          textColor[
+            colorIsDark(backgroundColor) ||
```
ericsoderberghp: Rather than calling `colorIsDark` twice, how about saving the result in a `const` after line 116?

taysea: good call.

Issue #3657, opened 2020-01-13
Placeholder in TextInput does not disappear if I use a react node instead of a string


c3s4:
If in a TextInput I use a react node as placeholder, the placeholder itself does not disappear when I write something inside the input text.

https://codesandbox.io/s/placeholder-does-not-disappear-w41o5

thaydds:
i just have finished and soon i'm going to open a PR!!!

ShimiSun:
Fixed and will be available on our next release!

PR #3659, opened 2020-01-16
fix custom placeholder bug (#3657)

thaydds:
#### What does this PR do?
Fix a bug when a React.node is passed at placeholder props in TextInput.
Created a isInputEmpty flag in TextInput state to not render a StyledPlaceholder when a React.node is passed by props in placeholder and the input value is empty.
TextInput.js

Placeholder Story TextInput

StoryBook
no
https://github.com/grommet/grommet/issues/3657

|Before|After|
|-|-|
|||

yes, a story  was added in TextInput

yes, 'cause it's a bug a fix.

no

### DIFF
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -269,11 +269,7 @@ class TextInput extends Component {
 
   onChange = event => {
     const { onChange } = this.props;
-    if (event.target.value.length > 0) {
-      this.setState({ isInputEmpty: false });
-    } else {
-      this.setState({ isInputEmpty: true });
-    }
+    this.setState({ isInputEmpty: !Boolean(event.target.value) });
     this.resetSuggestions();
     if (onChange) {
       onChange(event);
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -356,7 +356,7 @@ class TextInput extends Component {
     delete rest.onSuggestionsClose;
     const { showDrop, isInputEmpty } = this.state;
     const showStyledPlaceholder =
-      placeholder && typeof placeholder !== 'string' && !value && isInputEmpty;
+      placeholder && typeof placeholder !== 'string' && isInputEmpty;
     // needed so that styled components does not invoke
     // onSelect when text input is clicked
     delete rest.onSelect;
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -269,7 +269,7 @@ class TextInput extends Component {
 
   onChange = event => {
     const { onChange } = this.props;
-    this.setState({ isInputEmpty: !Boolean(event.target.value) });
+    this.setState({ isInputEmpty: !event.target.value });
     this.resetSuggestions();
     if (onChange) {
       onChange(event);
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -72,6 +72,7 @@ class TextInput extends Component {
   state = {
     activeSuggestionIndex: -1,
     showDrop: false,
+    isInputEmpty: true,
   };
 
   inputRef = React.createRef();
@@ -268,6 +269,11 @@ class TextInput extends Component {
 
   onChange = event => {
     const { onChange } = this.props;
+    if (event.target.value.length > 0) {
+      this.setState({ isInputEmpty: false });
+    } else {
+      this.setState({ isInputEmpty: true });
+    }
     this.resetSuggestions();
     if (onChange) {
       onChange(event);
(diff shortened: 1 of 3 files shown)


src/js/components/TextInput/TextInput.js
```diff
@@ -371,10 +379,11 @@ class TextInput extends Component {
       );
     }
     return (
+      // { render placeHoldr && <Styled...}
```
negreirosleo: You should delete this comment :eyes:

src/js/components/TextInput/stories/Placeholder.js
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { Box, Form, TextInput, Text } from 'grommet';
+
+const Placeholder = () => {
```
negreirosleo: I think you could improve this name, more like "StyledPlaceholder"

src/js/components/TextInput/stories/Placeholder.js
```diff
@@ -0,0 +1,16 @@
...
+  return (
+    <Box>
+      <Form>
+        <TextInput name="name" placeholder={<Text>footer</Text>} />
+        <TextInput name="name" placeholder="foobar" />
```
negreirosleo: Also, you could delete this example :thinking: , letting just the first one

negreirosleo:
Just those litte note, but congrats, nice job :)

src/js/components/TextInput/TextInput.js
thaydds: u right!!
thanks man

src/js/components/TextInput/TextInput.js
```diff
@@ -348,7 +354,9 @@ class TextInput extends Component {
     delete rest.onChange; // se we can manage in this.onChange()
     delete rest.onSuggestionsOpen;
     delete rest.onSuggestionsClose;
-    const { showDrop } = this.state;
+    const { showDrop, isInputEmpty } = this.state;
+    const showStyledPlaceholder =
+      placeholder && typeof placeholder !== 'string' && !value && isInputEmpty;
```
hjJunior: I think you may not need this conditional `!value` since you already are using `isInputEmpty`

thaydds: make sense!!
tnks man!! :)
i'll fix it

src/js/components/TextInput/stories/Placeholder.js
thaydds: u right, focus on styled placeholders.

tnks!!

src/js/components/TextInput/TextInput.js
```diff
@@ -268,6 +269,11 @@ class TextInput extends Component {
 
   onChange = event => {
     const { onChange } = this.props;
+    if (event.target.value.length > 0) {
```
murilovarela: What do you think of `this.setState({ isInputEmpty: !Boolean(event.target.value) })` ?

hjJunior: Good idea, or maybe just `!!event.target.value`

thaydds: so more elegant!!
tnks :)

src/js/components/TextInput/TextInput.js
```diff
@@ -72,6 +72,7 @@ class TextInput extends Component {
   state = {
     activeSuggestionIndex: -1,
     showDrop: false,
+    isInputEmpty: true,
```
ericsoderberghp: What if an initial value is specified? It seems like the placeholder will be shown in that case?

If that's correct, I'm wondering if we don't need the check for value length in `onChange` or a state, but just a change to the value test in the `render` function, such as?:
`placeholder && typeof placeholder !== 'string' && ((typeof value === 'string' && value.length === 0) || value === undefined)`

src/js/components/TextInput/TextInput.js
```diff
@@ -352,7 +352,7 @@ class TextInput extends Component {
     delete rest.onSuggestionsClose;
     const { showDrop, isInputEmpty } = this.state;
     const showStyledPlaceholder =
-      placeholder && typeof placeholder !== 'string' && isInputEmpty;
+      placeholder && typeof placeholder !== 'string' && isInputEmpty && !value;
```
ericsoderberghp: I don't think we need `isInputEmpty` still.

src/js/components/TextInput/TextInput.js
thaydds: nice point.
i didn't test for a controlled component.
in this case i not showing the placeholder is the expected behavior.

src/js/components/TextInput/TextInput.js
thaydds: in an uncontrolled component case we need to use isInputEmpty because we wont have the value prop.
do you think that situation could happen?

ericsoderberghp: Ah, good point. Agreed.

PR #3661, opened 2020-01-17
Fix TS errors for ColorType and sizes

AlexJukes:
#### What does this PR do?

Fixes some TS issues related to the fact custom added properties to size and colour can and should be able to be undefined. 

These are small one line changes in the themes/base.d.ts and utils/index.ts files. 

Currently compiles using Typescript v. `6.13.1`

Compile project using Typescript v. `6.13.1`

Currently this is causing our project to fail typescript compilation.

Custom properties that users can set in objects should also be able to be have undefined values, otherwise this causes a type error for the other defined optional properties on that object.

Screenshot 2020-01-17 at 14 54 27

Screenshot 2020-01-17 at 14 56 37

Don't think so

🤷‍♀️ up to you!


ShimiSun:
Great, thanks for the contribution, right on time for the next release!

Issue #3664, opened 2020-01-19
onKeyDown does not fire for TextArea component


gtrogers:
Hey, enjoying playing around with Grommet, thanks for the great effort and focus on accessibility. I came across this today and thought it was strange and a bit unintuitive. I'm not sure if it's intentional but it and prevents certain use cases that can't be done with `onChange` (like detecting backspace). 

### Expected Behavior

`onKeyDown` handlers added to `TextArea` should fire when the text area receives key down events (similar to `TextInput` and html `textareas`).

### Actual Behavior

`onKeyDown` handlers don't get called for `TextArea`

Here's a [CodePen](https://codesandbox.io/s/grommet-v2-template-b17d1expanddevtools=1&fontsize=14&hidenavigation=1&theme=dark) with a TextArea, TextInput and textarea all logging their onKeyDown events to the console.

gtrogers:
I started work on this last night: https://github.com/gtrogers/grommet/commit/89882148165af21e0611d4e48760407269f9addc

Will raise a PR once I've given it a bit more testing and maybe added the other key handler events, assuming it's something that fits in with Grommet?

gtrogers:
I started work on this last night: https://github.com/gtrogers/grommet/commit/89882148165af21e0611d4e48760407269f9addcv

Seems to work (naively). I'll add the other key handlers and hopefully raise a PR soon.

nateq314:
Has there been any further work on this? Ran into this today. Tried wrapping the `TextArea` in a `Keyboard` but that's not picking it up either.

PR #3666, opened 2020-01-20
Pass onKeyDown handler to TextArea

gtrogers:
#### What does this PR do?

- Allows the TextArea component to receive an onKeyDown handler
- Add tests for TextArea receiving onKeyDown and onKeyUp events

Take a look at the TextArea component. Strangely the onKeyUp handler was already getting called with no additional changes but the onKeyDown handler needed to be passed to the Keyboard component in order for it to work.

Added two unit tests for the TextArea to the existing test file.

I've tested this on a projected where I needed to detect the backspace button being pressed in a textarea and it works. There's possibly something I'm not understanding because onKeyUp was already working so it feels strange that it would only be an issue of onKeyDown - maybe it was disabled for a reason I don't know about?

https://github.com/grommet/grommet/issues/3664

Don't think so

Possibly counts as a bug fix

AFAIK this shouldn't break any existing code

### DIFF
--- a/src/js/components/TextArea/TextArea.js
+++ b/src/js/components/TextArea/TextArea.js
@@ -15,7 +15,7 @@ const TextArea = ({ fill, forwardRef, ...rest }) => {
   };
 
   return (
-    <Keyboard onEsc={onEsc}>
+    <Keyboard onEsc={onEsc} onKeyDown={rest.onKeyDown}>
       <StyledTextArea ref={forwardRef} fillArg={fill} {...rest} />
     </Keyboard>
   );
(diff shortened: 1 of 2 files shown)


src/js/components/TextArea/TextArea.js
```diff
@@ -15,7 +15,7 @@ const TextArea = ({ fill, forwardRef, ...rest }) => {
   };
 
   return (
-    <Keyboard onEsc={onEsc}>
+    <Keyboard onEsc={onEsc} onKeyDown={rest.onKeyDown}>
```
ericsoderberghp: It doesn't look correct to pass `rest.onKeyDown` to both `Keyboard` and to `StyledTextArea`. I'm wondering why the textarea isn't delivering the onKeyDown event since the handler is getting passed via `{...rest}`?

gtrogers: Thanks @ericsoderberghp, I'll admit this was a pretty optimistic PR. I'll do some more investigation if I get time.

nateq314:
Submitted a revision PR: https://github.com/grommet/grommet/pull/4161

PR #3672, opened 2020-01-22
Changed FormField to set aria-invalid when validation fails

ericsoderberghp:
#### What does this PR do?

Changed FormField to set aria-invalid when validation fails

FormField.js

Form Storybook stories "All" and "Validate on blur"

Form Storybook stories "All" and "Validate on blur"

no

yes

backwards compatible

### DIFF
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -50,7 +50,7 @@ class FormFieldContent extends Component {
     }
   }
 
-  renderChildren = (value, update) => {
+  renderChildren = (value, invalid, update) => {
     const {
       name,
       checked,
@@ -72,6 +72,7 @@ class FormFieldContent extends Component {
             update(name, event.target.checked);
             if (onChange) onChange(event);
           }}
+          aria-invalid={invalid || undefined}
           {...rest}
         />
       );
@@ -86,6 +87,7 @@ class FormFieldContent extends Component {
         }}
         plain
         focusIndicator={false}
+        aria-invalid={invalid || undefined}
         {...rest}
       />
     );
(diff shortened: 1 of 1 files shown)


src/js/components/FormField/FormField.js
```diff
@@ -72,6 +72,7 @@ class FormFieldContent extends Component {
             update(name, event.target.checked);
             if (onChange) onChange(event);
           }}
+          aria-invalid={invalid || undefined}
```
ShimiSun: I don't think `aria-invalid` should accept `undefined`, from what I see on its docs it should be set to `false` instead. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-invalid_attribute

src/js/components/FormField/FormField.js
```diff
@@ -86,6 +87,7 @@ class FormFieldContent extends Component {
         }}
         plain
         focusIndicator={false}
+        aria-invalid={invalid || undefined}
         {...rest}
```
ShimiSun: If the user can leverage the `...rest` to set `aria-invalid`, why do grommet needs to add it manually?

src/js/components/FormField/FormField.js
ericsoderberghp: `undefined` means we will remove it from the DOM, which matches the initial state.

src/js/components/FormField/FormField.js
ericsoderberghp: Because grommet is managing the validation. We will set `aria-invalid` based on our evaluation of the caller supplied regex or validation function result.

Issue #3662, opened 2020-01-17
Clock onChange called with non-changing time


standamo:
When using the below, the time printed is not changing:
```
<Clock type="digital"  onChange={t => console.log(t)} />
```

### Expected Behavior

Time should increment on every printout

### Actual Behavior

Startup time is printed over and over again.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-sandbox-0e6bh
https://0e6bh.csb.app/clock

### Steps to Reproduce

Run the above code and check console

### Your Environment

- Grommet version: 2.9.0
- Browser Name and version: Chrome (latest), FF (latest)
- Operating System and version (desktop or mobile): Win10

### Possible fix

Changed run() of Clock to fix:
```
      this.setState({ elements: nextElements }, () => {
        if (onChange) {
          if (elements.duration) {
            onChange(
              `P${nextElements.hours}H${nextElements.minutes}M${nextElements.seconds}S`,
            );
          } else {
            onChange(
              `T${nextElements.hours}:${nextElements.minutes}:${nextElements.seconds}`,
            );
          }
        }
```

ShimiSun:
The fix for this issue is merged and will be available on our next release. Thank you for reporting

PR #3674, opened 2020-01-22
Fix Clock.onChange bug

edumoreira1506:
#### What does this PR do?
Fixes clock bug. 

Testing.

-

-

- Look issue #3662 

- #3662


src/js/components/Clock/Clock.js
```diff
@@ -159,11 +159,20 @@ class Clock extends Component {
         if (onChange) {
           if (elements.duration) {
             onChange(
-              `P${elements.hours}H${elements.minutes}M${elements.seconds}S`,
+              `
+              P${nextElements.hours}
```
ericsoderberghp: This will introduce newlines into the ISO time format, which shouldn't be there.

Also, instead of referencing the old `nextElements`, I think we should be referring to the latest ones by re-getting them from the latest state in the callback, perhaps?:

```
this.setState({ elements: nextElements }, () => {
  if (onChange) {
    const { elements: e2 } = this.state;
    if (elements.duration) {
      onChange(`P${e2.hours}H${e2.minutes}M${e2.seconds}S`);
    } else {
      onChange(`T${e2.hours}:${e2.minutes}:${e2.seconds}`);
    }
  }
});
```

edumoreira1506: Nice! I'll update it. Thanks!

ShimiSun:
Obrigado amigo! Thank you for the contribution @edumoreira1506 !

PR #3680, opened 2020-01-23
Changed RangeInput to not use hocs

ericsoderberghp:
#### What does this PR do?

Changed RangeInput to not use hocs

RangeInput.js

storybook

storybook

no

no

backwards compatible


src/js/components/RangeInput/RangeInput.js
```diff
@@ -1,22 +1,34 @@
...
-const RangeInput = ({ forwardRef, ...rest }) => (
-  <StyledRangeInput {...rest} ref={forwardRef} type="range" />
-);
+const RangeInput = forwardRef(({ onFocus, onBlur, ...rest }, ref) => {
+  const [focus, setFocus] = useState();
+  return (
+    <StyledRangeInput
+      ref={ref}
+      focus={focus}
+      {...rest}
```
ShimiSun: Semantically, I think that `...rest` should be placed as the last prop so the user can override any default behavior our component provides out of the box.
In this case, I noticed you moved the `...rest` to be the third prop rather than the first, but I'm curious why it is not the latter?

ericsoderberghp: The two props after `...rest` are `onFocus` and `onBlur`, which we cannot let the caller replace. I know they won't be in `rest`, because they are excluded above, but this approach seemed to make it even clearer.

PR #3686, opened 2020-01-24
Changed MaskedInput to use hooks

ericsoderberghp:
#### What does this PR do?

Changed MaskedInput to use hooks

MaskedInput.js

storybook

storybook

no

yes

backwards compatible


src/js/components/MaskedInput/stories/Simple.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { Box, Grommet, MaskedInput } from 'grommet';
+import { grommet } from 'grommet/themes';
+
+const SimpleMaskedInput = () => {
```
ShimiSun: I'm not sure what am I seeing on this Simple MaskedInput example. Is the point of it is seeing only one character from any type?

ericsoderberghp: Mostly to make sure that we can not crash when a mask isn't provided.

ShimiSun: I that case I think it will be preferred to put it as a test case and not as an example for the users.

ericsoderberghp: good idea, removing ...

PR #3688, opened 2020-01-24
Changed input components to use FormContext

ericsoderberghp:
#### What does this PR do?

Changed input components to use FormContext.
This makes it possible to control the layout of FormField contents while still being tied in to the Form value.

TextArea.js gives a good example.

Storybook, added Form -> All children

Storybook

no

yes

backwards compatible


src/js/components/Form/stories/CustomTheme.js
```diff
@@ -16,21 +29,60 @@ const customFormFieldTheme = {
...
+              name="email"
+              mask={[
+                { regexp: /^[\w\-_.]+$/, placeholder: 'example' },
+                { fixed: '@' },
+                { regexp: /^[\w]+$/, placeholder: 'my' },
+                { fixed: '.' },
+                { regexp: /^[\w]+$/, placeholder: 'com' },
+              ]}
+            />
+          </FormField> */}
```
ShimiSun: Can we remove this potion of the code?

src/js/components/RangeInput/RangeInput.js
```diff
@@ -1,26 +1,49 @@
...
-        setFocus(false);
-        if (onBlur) onBlur(event);
-      }}
-      type="range"
-    />
-  );
-});
+const RangeInput = forwardRef(
+  ({ name, onChange, onFocus, onBlur, value: valueProp, ...rest }, ref) => {
+    const formContext = useContext(FormContext);
```
ShimiSun: Not sure it makes sense to me that RangeInput is using FormContext.

src/js/components/Form/stories/CustomTheme.js
ericsoderberghp: I de-commented it now that MaskedInput is a hook component too.

src/js/components/RangeInput/RangeInput.js
ericsoderberghp: RangeInput is just a `<input type="range" />`, which provides a numeric input control. So, it sets a numeric property value.

PR #3691, opened 2020-01-25
[TS] Fixed typing for Diagram connections

ShimiSun:
The issue is coming from slack https://grommet.slack.com/archives/C04LMHN59/p1579887376150400

Fixing types of the connections object and adding the example shared on slack.

### DIFF
--- a/src/js/components/Diagram/index.d.ts
+++ b/src/js/components/Diagram/index.d.ts
@@ -3,16 +3,20 @@ import * as React from "react";
 import { 
   ColorType 
 } from "../../utils";
+
+export type DiagramConnectionAnchor = "center" | "vertical" | "horizontal";
+export type DiagramConnectionType = "direct" | "curved" | "rectilinear";
+
 export interface DiagramProps {
   connections: {
-    anchor?: "center" | "vertical" | "horizontal" | string, 
+    anchor?: DiagramConnectionAnchor, 
     color?: ColorType,
     fromTarget: string | object,
     label?: string,
     offset?: "xsmall" | "small" | "medium" | "large" | string,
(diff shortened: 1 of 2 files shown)


src/js/components/Diagram/index.d.ts
```diff
@@ -5,14 +5,15 @@ import {
 } from "../../utils";
 export interface DiagramProps {
   connections: {
-    anchor?: "center" | "vertical" | "horizontal", 
+    anchor?: "center" | "vertical" | "horizontal" | string,
```
carolineartz: This will fix the issue I see in the editor/storybook! The only thing is that now it's not taking advantage of a type system since any string is valid but not necessarily meaningful

that is, "center" is now the same as "asfasdfsfd" (which is, i think actually, the same as `undefined` as well).
https://github.com/grommet/grommet/blob/master/src/js/components/Diagram/Diagram.js#L152

if there were exported type aliases defined for these unions, it would allow easily allow the cosumer to define the type on their own object and enforce a valid value.

```ts
export type DiagramConnectionAnchor = "center" | "vertical" | "horizontal"
export type DiagramConnectionType = "direct" | "curved" | "rectilinear"
```
it would also be nice to have the connection object available easily since a `Connection` type is really the 🍗 and 🥔 s here.

🤷‍♀ I dunno if you folks are building out the type system further or no?

carolineartz:
❤️ 🎉 🍨 🍨 🍨

src/js/components/Diagram/index.d.ts
ShimiSun: AGREED!

PR #3692, opened 2020-01-27
Changed Calendar to use hooks

ericsoderberghp:
#### What does this PR do?

Changed Calendar to use hooks.
These changes also change the keyboard interaction to tie focus to the grid of days and then allow traversing the days via arrow keys. Tab just enters and exits the grid of days itself.

Calendar.js

storybook

storybook

#3609 

no

yes

backwards compatible


src/js/components/Calendar/Calendar.js
```diff
@@ -68,262 +52,293 @@ const buildState = props => {
...
+      const PreviousIcon =
+        size === 'small'
+          ? theme.calendar.icons.small.previous
+          : theme.calendar.icons.previous;
+
+      const NextIcon =
+        size === 'small'
+          ? theme.calendar.icons.small.next
+          : theme.calendar.icons.next;
+
```
ShimiSun: can we add a condition to if the theme exists before we are attempting `theme.calendar.icons.small.next`

ericsoderberghp: I just carried these over from before. And I think we've always got `theme.calendar.icons` in the base theme. Are you suggesting that if there is no theme at all we should be more defensive and have default icons imported in Calendar.js? I had thought we were moving away from that pattern and doing the icon import in the theme?

ShimiSun: agreed.

PR #3695, opened 2020-01-28
Work on clear selection

oorestisime:
started some work here @markjackson02  you can check storybook to see how it works on the clear selection item in selects.

#### What does this PR do?

Implements #2973 solution

storybook

storybook

storybook

#2973 
done
yes
compatible


oorestisime:
@ericsoderberghp @ShimiSun not really happy with my naming there. open to suggestions. i also tried to avoid polluting a lot the api this is why i didn't add a showClearSelection true|false but don't feel strongly about it.

markjackson02:
@oorestisime this looks great!!!

src/js/components/Select/SelectContainer.js
```diff
@@ -342,9 +352,35 @@ class SelectContainer extends Component {
     return result;
   };
 
+  renderClear = () => {
+    console.log('in here');
```
ShimiSun: Yes, you are :) 
I think we can remove this console.log.

src/js/components/Select/doc.js
```diff
@@ -26,6 +26,13 @@ export const doc = Select => {
     closeOnChange: PropTypes.bool
       .description('Wether to close the drop when a selection is made.')
       .defaultValue(true),
+    clearValueLabel: PropTypes.string
+      .description('Label for the clear selection item')
+      .defaultValue('Clear selection'),
+    clearValuePosition: PropTypes.oneOf(['bottom', 'top']).description(
+      `Add a clear value item add the top or at the bottom of the container.
+      By default no unselect option is present.`,
+    ),
```
ShimiSun: Cards on the table, I think that Select is one of the most complicated components in Grommet, lots of props and options it can get overwhelming for users.
In the case above, I feel like we are adding two new props but one of them wouldn't make sense without the other, hence, IMO it should be only one object prop that will gather all the `clear` value attributes, so if the prop is undefined, the clear option wouldn't be available.
So something like this:
```
<Select clear={{label: `Clear`, value: 'top'}} ... />
```

We can even take it one step forward and combine the `renderClearValue` prop with it as well, something like:
```
<Select clear={{label: `Clear`, value: 'top', action: ()=> ()}} ... />
```

I would love to hear your thoughts here, @markjackson02  welcome to join as well.

src/js/components/Select/SelectContainer.js
```diff
@@ -342,9 +352,35 @@ class SelectContainer extends Component {
...
+    }
+    return (
+      <Box flex={false}>
+        <Button
+          tabIndex="-1"
+          hoverIndicator="background"
+          onClick={this.clearSelection}
+        >
+          <Box
+            background="light-4"
```
ShimiSun: Not a must at this point, but we are trying to stay away from the `light-*` colors these days, it is not released officially yet but you can read about it in the *NEW* sections on this wiki https://github.com/grommet/grommet/wiki/Color-Properties

bottom line, it might be better to use `background-back` color?

src/js/components/Select/SelectContainer.js
```diff
@@ -390,6 +427,7 @@ class SelectContainer extends Component {
               />
             </Box>
           )}
+          {value && clearValuePosition === 'top' ? this.renderClear() : null}
```
ShimiSun: would it be better to use undefined instead of null?

src/js/components/Select/doc.js
oorestisime: yeah great suggestion. i think it totally makes sense and i should do that for render prop as well.

src/js/components/Select/SelectContainer.js
oorestisime: i think you can't return undefined for a jsx. i ll need to test this

src/js/components/Select/SelectContainer.js
oorestisime: sure thing i ll have a look :)

src/js/components/Select/SelectContainer.js
markjackson02: i always return null, but undefined would have the same effect https://stackoverflow.com/a/42083262/711672

src/js/components/Select/doc.js
markjackson02: i agree!

PR #3699, opened 2020-01-29
restore focus to provided ref and tests

halocline:

#### What does this PR do?

Fixes bug experienced when a user provides their own `ref` to `<TextInput>` as shown below:
Kapture 2020-01-24 at 9 20 15

If the `ref` is removed, TextInput w/ suggestion functions as normal. However, when `ref` is provided, the following error is thrown: 
Screen Shot 2020-01-24 at 2 55 16 PM

TextInput.js line 192

Local project. Plus additional tests written.

Create a `<TextInput>` with a `ref`

Wondering if `const inputRef = useRef();` should be updated to `const inputRef = ref || useRef()`, then all executions of `(ref || inputRef)` could be simplified to `inputRef` and make this less bug prone. Thoughts?

Kapture 2020-01-24 at 9 20 15

Screen Shot 2020-01-24 at 2 55 16 PM

No

🤷‍♂ Probably not necessary

Backwards compatible.

### DIFF
--- a/src/js/components/MaskedInput/__tests__/__snapshots__/MaskedInput-test.js.snap
+++ b/src/js/components/MaskedInput/__tests__/__snapshots__/MaskedInput-test.js.snap
@@ -528,10 +528,12 @@ exports[`MaskedInput event target props are available option via mouse 4`] = `
   <input
     autocomplete="off"
     class="StyledMaskedInput-sc-99vkfa-0 gFvqdW"
+    data-g-tabindex="none"
     data-testid="test-input"
     id="item"
     name="item"
     placeholder="!"
+    tabindex="-1"
     value=""
   />
 </div>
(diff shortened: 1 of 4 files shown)


halocline:
It looks like two MaskedInput snapshots are not matching and causing Jest test check to fail. The new snapshots were created after `yarn test-update` and have tabindex attributes added, whereas they are not present in the previous snapshot.

Screen Shot 2020-01-29 at 10 48 23 AM

The snapshot diff seems to be unrelated to the work on this PR.

Thoughts?

taysea:
> It looks like two MaskedInput snapshots are not matching and causing Jest test check to fail. The new snapshots were created after `yarn test-update` (and prior to any changes to TextInput.js) and have tabindex attributes added, whereas they are not present in the previous snapshot.
> 
> Screen Shot 2020-01-29 at 10 48 23 AM
> 
> The snapshot diff seems to be unrelated to the work on this PR.
> 
> Thoughts?

I ran into the same thing when working on the TextInput keyboard navigation. I opened PR #3700 to try to just update those snapshots separately (assuming they're coming from something that got committed to master without test updates), but the test is failing for those as well.

ericsoderberghp:
Thanks for fixing this!

PR #3703, opened 2020-01-30
Improve TextInput keyboard accessibility

taysea:
#### What does this PR do?
This PR allows the user to navigate the TextInput suggestions with keyboard and keeps the active suggestion in view by scrolling through the options as the down/up arrow is pressed. The portion of the functions that check if the suggestion is before/after scroll were leveraged from grommet Select.

This has been tested in Chrome, Safari, and FireFox using the "Suggestions" storybook example. As you use down arrow key or up arrow key to navigate through the results, the drop should scroll to keep the active suggestion in view. Also, it supports a trade off between mouse/keyboard navigation where when the mouse hovers over an option it makes that the new active index and keyboard navigation will be up from there.

A note: I wanted to add in a test to check for this, but ran into an issue with being able to check what is truly "visible" in the browser as opposed to just present in the DOM. Here is a link to others with a similar issue: https://github.com/testing-library/dom-testing-library/issues/196. This PR includes a fix, but not an added test.

Using TextInput Suggestions story book, use arrow keys to navigate through the suggestions and make sure the drop scroll with you.

As users used arrow keys to navigate down or up in the suggestions, the drop wouldn't scroll so the active suggestion could fall out of view.

N/A.

No.

No.

Yes, backwards compatible.


ericsoderberghp:
I've already approved this, but I have an "extra credit" thought ...

Leave the code for setting the `activeSuggestionIndex` as was and change the scroll handling to be a separate effect hook that runs when `activeSuggestionIndex` changes. The hook would check for the active option being visible either direction and scroll appropriately. Ideally, we could re-use this hook inside SelectContainer as well ;)

taysea:
> I've already approved this, but I have an "extra credit" thought ...
> 
> Leave the code for setting the `activeSuggestionIndex` as was and change the scroll handling to be a separate effect hook that runs when `activeSuggestionIndex` changes. The hook would check for the active option being visible either direction and scroll appropriately. Ideally, we could re-use this hook inside SelectContainer as well ;)

Nice idea. Let me try that out.

PR #3707, opened 2020-01-31
Changed FormField to support background styling tied to error and disabled states

ericsoderberghp:
#### What does this PR do?

Changed FormField to support background styling tied to error and disabled states

FormField/doc.js

storybook, theme designer

storybook

no

yes

backwards compatible

### DIFF
--- a/src/js/components/FormField/doc.js
+++ b/src/js/components/FormField/doc.js
@@ -113,7 +113,7 @@ export const themeDoc = {
   },
   'formField.disabled.background.opacity': {
     description: 'The opacity of the FormField background when it is disabled.',
-    type: 'string',
+    type: 'string | boolean | number',
     defaultValue: undefined,
   },
   'formField.error.background.color': {
@@ -125,7 +125,7 @@ export const themeDoc = {
   'formField.error.background.opacity': {
     description:
       'The opacity of the FormField background when there is an error.',
-    type: 'string',
+    type: 'string | boolean | number',
     defaultValue: undefined,
   },
   'formField.error.color': {
--- a/src/js/components/FormField/doc.js
+++ b/src/js/components/FormField/doc.js
@@ -28,6 +28,9 @@ export const doc = FormField => {
       of name, value, onChange (event => {}), while event has either event.value
       or event.target.value.`,
     ),
+    disabled: PropTypes.bool.description(
+      'Whether the field should look disabled.',
+    ),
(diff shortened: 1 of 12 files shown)


src/js/themes/base.js
```diff
@@ -517,17 +517,18 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
       },
       content: {
-        pad: {
-          horizontal: 'small',
-          bottom: 'small',
-        },
+        pad: 'small',
+      },
+      disabled: {
+        // background: undefined,
```
ShimiSun: `base.d.ts` should include 
```
 disabled?: {
    background?: BackgroundType,
}
```
here https://github.com/grommet/grommet/blob/c4f9d42dc862bddde612172237a57aa7003c648a/src/js/themes/base.d.ts#L475

src/js/themes/base.js
```diff
@@ -517,17 +517,18 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
       },
       error: {
         color: {
           dark: 'status-critical',
           light: 'status-critical',
         },
         margin: { vertical: 'xsmall', horizontal: 'small' },
+        // background: undefined,
```
ShimiSun: Should be added in a similar way to https://github.com/grommet/grommet/blob/c4f9d42dc862bddde612172237a57aa7003c648a/src/js/themes/base.d.ts#L478

src/js/components/FormField/doc.js
```diff
@@ -101,7 +104,29 @@ export const themeDoc = {
...
+    defaultValue: 'small',
+  },
+  'formField.disabled.background.color': {
+    description: 'The color of the FormField background when it is disabled.',
+    type: "string | {'dark': string, 'light': string}",
+    defaultValue: undefined,
+  },
+  'formField.disabled.background.opacity': {
+    description: 'The opacity of the FormField background when it is disabled.',
+    type: 'string',
```
ShimiSun: According to https://github.com/grommet/grommet/blob/c4f9d42dc862bddde612172237a57aa7003c648a/src/js/utils/background.js#L81

I think that `background.opacity` could also be boolean and a number.

src/js/themes/base.js
ericsoderberghp: Ah, yes. I forgot typescript for base.js. Good catch!

src/js/themes/base.js
ericsoderberghp: yep

Issue #3702, opened 2020-01-29
Cursor loses position in TextInput


csclyde:
### Expected Behavior
When editing text in a TextInput, I enter a string (abc). Then, I move the cursor to the beginning of the string, and start typing (123). My cursor should remain when I put it, giving me "123abc"

### Actual Behavior
When editing text in a TextInput, I enter a string (abc). Then, I move the cursor to the beginning of the string, and start typing (123). After entering one character, my cursor jumps to the end of the input. Giving me "1abc23".

### URL, screen shot, or Codepen exhibiting the issue
The problem is visible in the storybook of the most recent version of Grommet.

ShimiSun:
Thanks for reporting, the fix will be available on our next release.

IanKBovard:
This can be closed. addressed in #3709

csclyde:
@ShimiSun Thanks for the quick work. Will this only be going into Grommet v2, or will it make it into v1 as well? We still have a project using v1. Thanks.

ShimiSun:
Hi @csclyde , thanks for your patience.
Looking at the code for this fix, it is very specific for v2 and wouldn't work on v1, more specifically v1 doesn't use React hooks (react version of ^16.8) so the couldn't apply there.

If you have an idea for a fix on v1, feel free to directly fix it on your local version, we do not have the resources to upgrade and have new releases on v1 at this point, unfortunately.

PR #3709, opened 2020-01-31
Issue #3702 TextArea, TextInput, MaskedInput cursor bug

IanKBovard:
I have a few questions about the changes made to textInput and textArea when refactoring. This will help me decide on a fix for these components.

The issue here, it seems to me, is that both components are being updated under the useEffect hook, which isn't saving the cursor position just updating the field with an entire string and throwing the cursor at the end. The solution here would be to use setValue(event.target.value) instead.

The final post here describes the solution.
https://github.com/facebook/react/issues/955

I'd like to know the reasoning behind using formContext.set() and get for forms(at the moment both if's run in forms, not sure if that was intentional), and not just setting the value. And the reasoning behind setting the value in the useEffect hook. 

I just want to make sure that any changes I make can keep the intention of the refactor without breaking functionality.


IanKBovard:
https://github.com/grommet/grommet/issues/3702

src/js/components/TextArea/TextArea.js
```diff
@@ -49,9 +50,11 @@ const TextArea = forwardRef(
           }}
           onChange={event => {
```
ericsoderberghp: I think we should do:
```
const nextValue = event.target.value;
setValue(nextValue);
```

IanKBovard:
MaskedInput shares this issue as well.

IanKBovard:
i'm curious, what is the significance of formContext?

ericsoderberghp:
> i'm curious, what is the significance of formContext?

It's what the input components use to interact with the `Form` component. When used inside a `Form`, the `Form` holds the official value for each input. But, the input components can be used without a `Form`.

src/js/components/TextArea/TextArea.js
```diff
@@ -17,10 +17,6 @@ const TextArea = forwardRef(
         ? valueProp
         : (formContext && name && formContext.get(name)) || '',
     );
-    useEffect(() => setValue(valueProp), [valueProp]);
-    useEffect(() => {
-      if (formContext && name) setValue(formContext.get(name) || '');
```
ericsoderberghp: We still need something like this to initialize the value from the Form `value` object.

src/js/components/TextArea/TextArea.js
```diff
@@ -48,11 +44,13 @@ const TextArea = forwardRef(
...
           }}
           onChange={event => {
+            const nextValue = event.target.value;
             if (formContext && name) {
-              formContext.set(name, event.target.value);
+              formContext.set(name, nextValue);
+              setValue(nextValue);
             }
             if (onChange) {
-              onChange(event);
```
ericsoderberghp: We still need to call `onChange`, in case the user passes in an `onChange`.

src/js/components/TextArea/TextArea.js
```diff
@@ -48,11 +44,13 @@ const TextArea = forwardRef(
             if (onBlur) onBlur(event);
           }}
           onChange={event => {
+            const nextValue = event.target.value;
             if (formContext && name) {
-              formContext.set(name, event.target.value);
+              formContext.set(name, nextValue);
+              setValue(nextValue);
```
ericsoderberghp: I think we can move this outside of the `if` statements, after line 47.

src/js/components/TextArea/TextArea.js
IanKBovard: gotcha, was just going to ask about this change

Issue #3712, opened 2020-02-03
Stories Missing in Windows


sivakar12:
When running yarn run storybook, Webpack gives a lot of warning for the TypeScript files in stories folders. 
The warnings are all similar
 

> Module parse failed: Unexpected token (22:2)
> You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file.

### Expected Behavior

Storybook should compile find and all the stories should be available in the Storybook that opens in the browser
### Actual Behavior

Compilation give a lot of warnings and most of the stories are not available
### Steps to Reproduce

1. Clone the repository to a Windows machine
2. Run 'yarn install' to install packages
3. Run 'yarn storybook' to start Storybook

### Your Environment

- Grommet version: 2.10.0
- Browser Name and version: Microsoft Edge 79
- Operating System and version (desktop or mobile): Windows 10

sivakar12:
If I change the test regex in the Storybook Webpack config file from  `/stories\/.*\.(ts|tsx)$/`  to `/stories\\.*\.(ts|tsx)$/`, it works.

PR #3713, opened 2020-02-03
Change Webpack loader's test to work on both Windows and *nix

sivakar12:
Signed-off-by: Sivakar Sithamparanathan <sivakar12@outlook.com>

#### What does this PR do?
Fixes a problem with Storybook's Webpack config
In the storybook/webpack.config.js file
This has been manually tested
By running 'yarn storybook' on both a Windows and a Unix based machine
Running Storybook in Windows made Webpack give warnings about missing loader, and not all stories were there
Issue #3712 
No
No
It is backwards compatible


ShimiSun:
Can you give a little background on why there was an issue and how this fix solves it?

sivakar12:
Storybook on Windows does not have most of the stories. That is because of errors in the Webpack build of Storybook. 

The TypeScript loader is needed for transpile TypeScript files used by the stories. 

The TypeScript loader is not configured to look for for files correctly. Windows uses backward slash (\) as path separator while Linux and Mac uses /. 

I changed the regular expression used to search files so that it works with both types of paths.

storybook/webpack.config.js
```diff
@@ -17,7 +17,7 @@ module.exports = async ({ config }) => {
   });
 
   config.module.rules.push({
-    test: /stories\/.*\.(ts|tsx)$/,
+    test: /stories(\\|\/).*\.(ts|tsx)$/,
```
ShimiSun: why not adding this fix to non-typescript story files as well? 
see line #5 on this file.

sivakar12: Sorry I missed that. Just realized that sources are not there in Windows because of the other regex. 
I have changed this and  added a new commit

ShimiSun: Can you verify it fixes the issue fo you?
@markjackson02 can you verify as well? I think you encountered a similar issue.

sivakar12: It fixes the issue for me

ShimiSun:

storybook/webpack.config.js
sivakar12: It fixes the issue for me

PR #3714, opened 2020-02-03
added storybook example for themecontext.extend with global theme

IanKBovard:
Storybook example added.

How to modify global theme to alter focus color, how to use ThemeContext.Extend to modify individual components.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,9 +49,6 @@
     "styled-components": ">= 5.X"
   },
   "dependencies": {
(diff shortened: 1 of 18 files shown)


ShimiSun:
@IanKBovard overall I think it is good. I will suggest separating the stories so they won't be placed on the same file and make it more difficult for the user to navigate between options.

ShimiSun:
Great!! Thank you for this story example!  🚀

PR #3719, opened 2020-02-04
Changed Select to use hooks

ericsoderberghp:
#### What does this PR do?

Changed Select to use hooks

Select.js

storybook

storybook

no

yes

backwards compatible


ShimiSun:
Chromatic found visual differences with those changes, I the Select is wider and the text is bigger, the stories visualization shouldn't change.

PR #3721, opened 2020-02-04
Fixes test error of uncontrolled to controlled input

IanKBovard:
Removes the error
```
Warning: A component is changing a controlled input of type undefined to be uncontrolled. Input elements should not switch from controlled to uncontrolled (or vice versa).
```
when testing by giving a default value of an empty string to the components causing the error.
The issue was caused by the components starting at a value of undefined making the input uncontrolled.


ShimiSun:
Great, thank you!

PR #3723, opened 2020-02-04
Changed FormField to use hooks

ericsoderberghp:
#### What does this PR do?

Changed FormField to use hooks

FormField.js

Storybook

Storybook

no

yes

backwards compatible


ericsoderberghp:
Hmm... partly. The RangeInput doesn't look right there. But, the other fields are shrinking appropriately.

PR #3730, opened 2020-02-05
Changed FormField to add 'round' to theme

ericsoderberghp:
#### What does this PR do?

Changed FormField to add 'round' to theme

FormField/index.d.ts

theme designer

manually

automatic

yes

backwards compatible

### DIFF
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -207,6 +207,7 @@ const FormField = forwardRef(
                 }
               : undefined
           }
+          round={border.position === 'inner' ? formField.round : undefined}
         >
           {contents}
         </Box>
@@ -267,6 +268,7 @@ const FormField = forwardRef(
         }
         background={outerBackground}
         margin={abut ? abutMargin : margin || { ...formField.margin }}
+        round={border.position === 'outer' ? formField.round : undefined}
         style={outerStyle}
         onFocus={event => {
           setFocus(true);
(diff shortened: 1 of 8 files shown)


src/js/components/FormField/FormField.js
```diff
@@ -267,6 +268,7 @@ const FormField = forwardRef(
         }
         background={outerBackground}
         margin={abut ? abutMargin : margin || { ...formField.margin }}
+        round={border.position === 'outer' ? formField.round : undefined}
```
ShimiSun: A couple of thoughts:
1. Can we replace this current logic with `border.position === 'outer' && formField.round` ? (logic applies to line 210 as well)
2. In line 210 we are also applying `round` to the inner radius, wouldn't it look awkward? I'm having a hard time to imagine how it will look.

ericsoderberghp: Regarding 1. if we did that then we'd end up with `round={false}` when border.position is 'inside', but I think we'd rather have `round={undefined}`.

Regarding 2. line 210, we are setting the border and rounding there when border.position is 'inner'. Here, we are setting border and rounding when border.position is 'outer'.

ericsoderberghp:
> Overall this PR looks good and will open the door for applying radius on the FormField theme 🎉
> For me, it was hard to visualize how it will work with various borders, I think we need to enhance the CustomTheme story with an example that demonstrates this new styling.

Well, I kind of like having CustomTheme use the outer border position with all sides. But, that wont look as nice with any rounding on the boxes.

Issue #3568, opened 2019-11-25
Consider adding `fill` to Markdown


vrde:
### Expected Behavior
As a user of the library I would like to have a `fill` option when rendering markdown using the `Markdown` component.

The `fill` option can be used in the `Paragraph` component, and I would expect to have the same behavior using `Markdown`.

### Actual Behavior
Right now I have no way to make the paragraph fill the width of the container :crying_cat_face: 

A workaround would be to use component override, but I'd still prefer to have the same behavior as the `Paragraph` component.

### URL, screen shot, or Codepen exhibiting the issue
Visit the Storybook [under the Markdown section](https://storybook.grommet.io/?path=/story/markdown--simple).

### Steps to Reproduce
Use the `Markdown` component :smiley: 

### Your Environment
- Grommet version: v2
- Browser Name and version: any
- Operating System and version (desktop or mobile): any

oorestisime:
With the addition of the fill prop in Paragraph https://v2.grommet.io/paragraph#fill 
You can component override with the Paragraph component and set the fill there. 

Something along the following should work
```
<Markdown components={{ p: (props)<Paragraph {...props} fill/> }}>{CONTENT}</Markdown>
``` 

I am not sure it makes sense to set this as a default

vrde:
Thank you for the tip! I've tried it out and it works.

As a newbie, I think `fill` would avoid confusion and align the behavior of the `Markdown` component to the one of `Paragraph`.

oorestisime:
Do you mean adding a fill prop in the Markdown component that we forward to the Paragraph?

vrde:
> Do you mean adding a fill prop in the Markdown component that we forward to the Paragraph?

Exactly :+1:

ShimiSun:
Thank you for reporting, you can use the options prop to control the fill property.

PR #3734, opened 2020-02-07
Exposed `options` on Markdown

edumoreira1506:
#### What does this PR do?
This PR adds a `fill` prop to markdown component.

Resolves issue #3568


ShimiSun:
Hi @edumoreira1506, thank you for picking up this issue and. helping us start a conversation here.

After looking into it,  and reviewing https://github.com/probablyup/markdown-to-jsx/blob/master/README.md#optionsforceinline, I think the user should be allowed to set the `forceInLine` prop via `options` and not as a new grommet prop.
This conclusion leads me to a few questions:
1. Why aren't we exposing the usage of the React Markdown component 'markdown-to-jsx',  if the user will know it, it could be simpler to research.
2. why aren't we exposing the `options` prop? it is there to use but we are not exposing it on the docs. 
I was able to solve the issue this PR is trying to solve only adding `options={{ forceInline: true }}` to the Markdown thanks to the `options` prop and the knowledge of markdown-to-jsx.

thoughts here @ericsoderberghp ? @oorestisime ?

edumoreira1506:
Maybe I could change the docs to show for user than we are using `markdown-to-jsx`.

oorestisime:
yeah i agree it makes sense to add the options in the docs. and mention we use that component so they will know what to search

ericsoderberghp:
I'm supportive of adding `fill` to Markdown as it seems consistent with the other grommet components that have `fill`.

ericsoderberghp:
After an excellent discussion with @ShimiSun, I am withdrawing my support for adding `fill` to Markdown. As, this isn't filling anything you might have in Markdown, such as an Image. It's just changing the behavior of Paragraph. For that, I think @oorestisime's suggestion to customize Paragraph props is the right approach. However, I am supportive of documenting the `options` property.

ShimiSun:
@edumoreira1506 , after applying the changes to doc.js, please run locally `yarn test-update` so it will create updated snapshots with the new content, push those snapshots as part of the PR and it will resolve the jest failures.

src/js/components/Markdown/README.md
```diff
@@ -1,6 +1,9 @@
 ## GrommetMarkdown
 Markdown formatting using Grommet components.
 
+Grommet uses 'markdown-to-jsx' in Markdown component,
+      you can see all the options in the documentation.
```
ShimiSun: thanks for adding this, can you also add the new prop of `options` (similarly to `components` prop) and direct to 'markdown-to-jsx' with a link?

edumoreira1506: Yes, sure!

edumoreira1506:
Updated.

src/js/components/Markdown/README.md
```diff
@@ -17,6 +20,10 @@ Custom components and props to override html elements such as 'img'
       or 'pre'. By default 'a', 'p', 'img', and table elements are overridden
       with grommet components
 
+**options**
+
+You can see all the options [here](https://github.com/probablyup/markdown-to-jsx).
+
```
ShimiSun: The README file is auto-generated when running `yarn test-update` and it shouldn't be manually edited.
A way to achieve it is to add `options` prop on the doc.js file and then auto-generate the readme to include the new prop details.

ShimiSun:

PR #3735, opened 2020-02-08
Add background prop to Grommet

taysea:
#### What does this PR do?
This PR allows users to define a `background` on the Grommet component. This is beneficial with the new theme concept of 'background-back' and 'background-front'. Right now, Grommet component only pulls `background` for background color, but users can benefit from having control of this.

Updated:
- Docs
- Storybook
- Typescript definitions

The prop accepts all of the same props that Box background accepts.

src/js/components/Grommet/Grommet.js

A story was added and tests were updated and no snapshots changed. A test has been added that applies a background to Grommet.

Apply a background to Grommet component.

In instances where `background-back` is the ideal background color, the user would need to wrap their entire layout in an additional Box. It would be more beneficial/intuitive if the Grommet component could receive a background value.

Yes, and I have updated them in this PR.

Yes.


src/js/components/Grommet/grommet.stories.js
```diff
@@ -51,4 +77,5 @@ const GrommetVars = () => (
 storiesOf('Grommet', module)
   .add('Plain', () => <Plain />)
   .add('Theme', () => <Themed />)
-  .add('Vars', () => <GrommetVars />);
+  .add('Vars', () => <GrommetVars />)
+  .add('Background', () => <Background />);
```
ShimiSun: we are trying to stay away from having all stories on the same file, can you put the story on a separate file?

src/js/components/Grommet/Grommet.js
```diff
@@ -19,26 +20,28 @@ class Grommet extends Component {
...
-        const color =
-          background[
-            themeMode || nextTheme.defaultMode || Object.keys(background)[0]
-          ];
-        nextTheme.dark = color ? colorIsDark(color) : false;
-      } else if (nextTheme.dark === undefined) {
-        nextTheme.dark = (background && colorIsDark(background)) || false;
-      }
+
+      nextTheme.dark = (themeMode || theme.defaultMode) === 'dark';
```
ericsoderberghp: I think we would like a comment here for when we read this months from now and forget why we are setting `nextTheme.dark` twice.

src/js/components/Box/doc.js
```diff
@@ -29,6 +29,31 @@ const ANIMATION_SHAPE = PropTypes.shape({
   size: PropTypes.oneOf(['xsmall', 'small', 'medium', 'large', 'xlarge']),
 });
 
+const BACKGROUND_SHAPE = PropTypes.oneOfType([
```
ericsoderberghp: Can we put this in utils/prop-types.js?

src/js/components/Grommet/Grommet.js
```diff
@@ -19,26 +20,32 @@ class Grommet extends Component {
...
+
+      // get initial value for dark so we can normalize background color
+      nextTheme.dark = (themeMode || theme.defaultMode) === 'dark';
+      const color = normalizeColor(
+        backgroundProp || themeBackground,
+        nextTheme,
+      );
+
+      // after normalizing, we want to check again if the background color
+      // is dark in case it differs from the previous assignment
```
ShimiSun: Is it necessary to have two assignments? can we normalize and then just assign once?

src/js/components/Box/doc.js
```diff
@@ -29,6 +29,31 @@ const ANIMATION_SHAPE = PropTypes.shape({
...
+      PropTypes.oneOf(['weak', 'medium', 'strong']),
+    ]),
+    repeat: PropTypes.oneOfType([
+      PropTypes.oneOf(['no-repeat', 'repeat']),
+      PropTypes.string,
+    ]),
+    size: PropTypes.oneOfType([
+      PropTypes.oneOf(['cover', 'contain']),
+      PropTypes.string,
+    ]),
```
ShimiSun: Interestingly we are not exposing these options on Box `background` prop. 
Might be a separate PR, but should we add them? https://v2.grommet.io/box#background

src/js/components/Grommet/stories/Background.js
```diff
@@ -0,0 +1,58 @@
...
+
+const Background = () => {
+  const themeColor = 'background-back';
+  const hexValue = '#DCD0FF';
+  const cssColor = 'gold';
+  return (
+    <Box gap="medium">
+      <Grommet>
+        <Box pad="medium">
+          <p>Grommet with no theme or background prop</p>
```
ShimiSun: Any reason we are not using Grommet's Paragraph?

taysea: This should be paragraph, thank you.

src/js/components/Box/doc.js
taysea: Yes, I will move it.

src/js/components/Box/doc.js
taysea: Hmm, that's a good observation. I think I'll check first into why they're not being pulled into the docs site. Seems like if they're defined here that they should be exposed. Agreed, it should be a separate PR regardless.

src/js/components/Grommet/Grommet.js
ericsoderberghp: It is necessary that we do it twice. We have to handle two cases:
1. caller passes in a color object or a color name that resolves an object. In this case, we want to set `dark` as line 38 shows. The second set, in line 46, is a no-op.
2. caller passes a specific color value or a color name that resolve to a specific color value. In this case, we want dark to be set based on that color, which line 46 will do.
The double set of `dark` allows us to handle both cases here without having to duplicate color object + name + dark mode detection code here that is already in `normalizeColor` and `backgroundIsDark`.
Maybe we should just put all of this in a comment, to be more verbose.

src/js/components/Box/doc.js
taysea: I will adjust the docs site in a separate PR to expose these options.

src/js/components/Grommet/Grommet.js
taysea: Thanks for the detailed explanation. I will transfer this into a comment for clarity.

src/js/components/Box/doc.js
```diff
@@ -1,6 +1,7 @@
 import { describe, PropTypes } from 'react-desc';
 
 import {
+  BACKGROUND_DOC,
```
ericsoderberghp: All upper case seems inconsistent.

src/js/components/Meter/doc.js
```diff
@@ -1,10 +1,17 @@
 import { describe, PropTypes } from 'react-desc';
 
-import {
-  backgroundPropType,
-  genericProps,
-  getAvailableAtBadge,
-} from '../../utils';
+import { genericProps, getAvailableAtBadge } from '../../utils';
+
+const backgroundPropType = PropTypes.oneOfType([
```
ericsoderberghp: It would be more consistent to put it inline below.

src/js/components/Meter/doc.js
```diff
@@ -26,10 +15,21 @@ export const doc = Meter => {
...
-      opacity: 'medium',
-    }),
+    background: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.shape({
+        color: PropTypes.string,
+        opacity: PropTypes.oneOfType([
+          PropTypes.oneOf(['weak', 'medium', 'strong']),
+          PropTypes.bool,
+        ]),
```
ShimiSun: I think opacity should also accept a number or a string, e.g. `opacity=0.8`

taysea: Gotcha, I'll check into how Meter works -- these are just the values that already existed and I was moving them into the Meter file. Will address this in a separate PR.

PR #3739, opened 2020-02-10
Added Button sizing

ShimiSun:
This PR is a conversation starter for the `size` prop in Button.
At the moment the `size` prop will impact `font-size`, `line-height` the same way that Text sizes were impacting button before, and `size` also has the ability to control the border `radius`.
If we like this direction I can enhance the size to control the `pad` as well.

TODOs:
* docs.js and index.d.ts are still WIP, I wanted to collect feedback before I continue.
* Button without a label that only has an icon isn't being impacted by the size, we can add it as well if you like the idea.
* tests
* icon + label needs to be impacted as well.

*****************************************************

Recent update:

We decided not to apply the `size` on the icon sizing. The users will still be able to set the size of the icon themselves. The reason behind it, is mostly due to the scenario of label & icon used together, if we'll take an example of `size="small"` the icon will be un-proportionally small comparing to the small text, or un-proportionally large for `size="large"` text.
The scenario of a single icon rendering would be intact, we wouldn't apply radius/pad/icon size on it at all since the pad of horizontal and vertical are different and the padding would look awkward on a symmetric icon.

I still need to address my own review comments and add typescript support after we agree on functionality.

*********************

Update:

Added TS support on both theme props and new prop.
remove size handling on plain mode.


src/js/components/Button/StyledButton.js
```diff
@@ -11,13 +11,20 @@ import {
 } from '../../utils';
 import { defaultProps } from '../../default-props';
 
+const radiusStyle = props => {
+  if (props.size) {
+    return props.theme.button.size[props.size].border.radius;
```
ericsoderberghp: If there isn't a size in the theme, I think we should fall back to `button.border.radius`.

src/js/components/Button/StyledButton.js
```diff
@@ -34,9 +41,24 @@ const primaryStyle = props => css`
     props.theme,
     props.theme.button.color,
   )}
-  border-radius: ${props.theme.button.border.radius};
+  border-radius: ${radiusStyle(props)};
 `;
 
+const sizeStyle = props => {
```
ericsoderberghp: Can we rename this to `text-style`, since that's what it's controlling?

src/js/themes/base.js
```diff
@@ -328,6 +328,26 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       // extend: undefined,
     },
     button: {
+      size: {
+        small: {
+          ...fontSizing(-1),
+          border: {
+            radius: `${baseSpacing * 0.7}px`,
```
ericsoderberghp: We should truncate to be whole pixel values.

src/js/components/Button/StyledButton.js
```diff
@@ -34,9 +41,24 @@ const primaryStyle = props => css`
...
+  const size = props.size || 'medium';
+  const data = props.theme.button.size[size];
```
ericsoderberghp: I'm wondering if we should connect this more directly to Text sizing in the theme, rather than allowing Button to have different text sizing? We should discuss with the designers.

src/js/components/Button/StyledButton.js
ShimiSun: Currently, it is also controlling `border.radius` and I was hoping to hear whether we want `pad` in as well, if we will include border and pad it wouldn't make sense to make it specific for `textStyle`.

src/js/themes/base.js
ShimiSun: To be honest, I was just putting placeholder's values for the radius, except the default 'medium' that behaves as it used to. 
If we will proceed with this direction I would ask @L0ZZI or @gregfuruiye for the border.radius accurate values.

src/js/components/Button/StyledButton.js
ShimiSun: I think you are right.

src/js/components/Button/StyledButton.js
ShimiSun: Good catch. done.

src/js/components/Button/StyledButton.js
ShimiSun: Rethinking this, I think I understand that your intention was just for the function name and not about the overall prop name. I agree and I'll address it.

ShimiSun: Done.

src/js/components/Button/StyledButton.js
ShimiSun: Done.

src/js/components/Button/StyledButton.js
```diff
@@ -11,18 +11,32 @@ import {
 } from '../../utils';
 import { defaultProps } from '../../default-props';
 
+const radiusStyle = props => {
+  if (props.size && props.theme.button.size) {
```
ericsoderberghp: I think we should check for the specific size here. If the caller passes in `xxsmall` as the size, this will cause an exception.

src/js/themes/base.js
ericsoderberghp: sounds good

src/js/components/Button/Button.js
```diff
@@ -82,10 +83,13 @@ const Button = forwardRef(
     };
 
     let buttonIcon = icon;
+
+    if (icon && size) buttonIcon = cloneElement(buttonIcon, { size: 'medium' });
```
ericsoderberghp: Shouldn't it just be `{ size }`? Also, this prevents  caller from overriding the size of an icon. I wonder if we should only explicitly set the size of the icon if the icon doesn't already have a size prop set? Similar to how we check for color below.

src/js/components/Button/StyledButton.js
```diff
@@ -11,18 +11,45 @@ import {
...
+  const size = props.size || 'medium';
+  const data = props.theme.text[size];
+  return css`
+    font-size: ${data.size};
+    line-height: ${data.height};
+  `;
+};
+
+const padStyle = props => {
+  if (props.size && props.theme.button.size) {
```
ericsoderberghp: Similar to radius comment above, I think we should check for `props.theme.button.size[props.size]` too.

src/js/components/Button/__tests__/Button-test.js
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import renderer from 'react-test-renderer';
 import { cleanup, fireEvent, render } from '@testing-library/react';
 import 'jest-styled-components';
+// import { Add, Next } from 'grommet-icons';
```
ericsoderberghp: Clean up commented out code?

src/js/components/Button/Button.js
ShimiSun: Now that we decided to have an impact on the icon size, it is cleaner to remove this addition. It was an exploration until now.

src/js/components/Button/__tests__/Button-test.js
ShimiSun: I've put it back in as I added back the tests.

src/js/components/Button/StyledButton.js
ShimiSun: Done, thanks.

src/js/components/Button/__tests__/__snapshots__/Button-test.js.snap
```diff
@@ -1206,6 +1216,677 @@ exports[`Button reverse icon label 1`] = `
...
+      </svg>
+    </div>
+  </button>
+  <button
+    className="c3"
+    onBlur={[Function]}
+    onFocus={[Function]}
+    onMouseOut={[Function]}
+    onMouseOver={[Function]}
+    size="large"
```
ShimiSun: Just noticed this guy, I'll remove it from the DOM.

src/js/components/Button/doc.js
```diff
@@ -77,40 +65,57 @@ with plain Buttons.`,
...
       )
       .defaultValue(false),
     reverse: PropTypes.bool
       .description(
         `Whether an icon and label should be reversed so that the icon is at the
-end of the anchor.`,
+              end of the anchor.`,
       )
       .defaultValue(false),
+    size: PropTypes.oneOf(['small', 'medium', 'large']).description(
```
ShimiSun: This is the only section that was added to this file. 
The other diffs are organization changes.

src/js/components/Calendar/__tests__/__snapshots__/Calendar-test.js.snap
```diff
@@ -131,6 +131,8 @@ exports[`Calendar date 1`] = `
   border: none;
   padding: 0;
   text-align: inherit;
+  font-size: 18px;
+  line-height: 24px;
```
ShimiSun: Not sure if this is a problem. thoughts? I. see the chromatic changes, so. it can't be good.

ericsoderberghp: Agreed. Good job Chromatic!
I think the issue is related to Calendar setting the font size at the Calendar level and expecting that to be inherited. But, the Button is always setting a font size. I'm thinking that `<Button plain />` should not set the font size.

src/js/components/Button/__tests__/__snapshots__/Button-test.js.snap
ShimiSun: Done.

src/js/components/Calendar/__tests__/__snapshots__/Calendar-test.js.snap
ShimiSun: Fixed, Chromatic is looking good now.

PR #3740, opened 2020-02-11
[TS] fix: rangeInput onChange prop type

moacirguedes:
#### What does this PR do?

This PR adjusts the onChange type definition for RangeInput

src/js/components/RangeInput/index.d.ts

#3165 

No

No

Yes

### DIFF
--- a/src/js/components/RangeInput/index.d.ts
+++ b/src/js/components/RangeInput/index.d.ts
@@ -5,7 +5,7 @@ export interface RangeInputProps {
   min?: number | string;
   max?: number | string;
   name?: string;
-  onChange?: ((...args: any[]) => void);
+  onChange?: ((event: React.ChangeEvent) => void);
   step?: number;
   value?: number | string;
 }
--- a/src/js/components/RangeInput/stories/typescript/CustomRange.tsx
+++ b/src/js/components/RangeInput/stories/typescript/CustomRange.tsx
@@ -1,0 +1,54 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import isChromatic from 'storybook-chromatic/isChromatic';
+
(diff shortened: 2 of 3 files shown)


src/js/components/RangeInput/index.d.ts
```diff
@@ -5,7 +5,7 @@ export interface RangeInputProps {
   min?: number | string;
   max?: number | string;
   name?: string;
-  onChange?: ((...args: any[]) => void);
+  onChange?: ((event: React.ChangeEvent) => void);
```
ShimiSun: I think @taysea reverted this event on purpose, @taysea can you review and remind us what was the issue that it had to be reverted?

taysea: There had been various typescript issues with certain onChange handlers (specifically Select I think), so to be safe with other components that hadn't been validated yet, @britt6612 and I reverted the types to "any". I don't believe RangeInput had caused any errors, but we had reverted it just to be safe since the "onChange" handler had been causing problems for various components.

I think if there is clear documentation that this is the correct type, then it can be added back in.

ShimiSun:
Hi @moacirguedes , thanks for this PR, to make sure we are not causing any regressions, would you be willing to add a TS story file, basically duplicating src/js/components/RangeInput/rangeinput.stories.js to a `ts` file (similarly to how it's done in Menu src/js/components/Menu/stories/typescript/Custom.tsx) and make sure the compilation passes with your suggested fix? 
If you feel like you can't handle it, let me know and I'll try to find some time to do it.
Thanks for the PR.

moacirguedes:
Hello @ShimiSun, thanks for the feedback, i will do that.

ShimiSun:
PERFECT!! Thank you buddy @moacirguedes looking to see more TS contributions from you!

PR #3741, opened 2020-02-11
Changed Chart to pick graph-0 color in case it is defined

ShimiSun:
I think we discussed in the past that if `graph-0` is defined then the theme will pick the `graph-0` color over the `accent-1` color, please let me know if my memory isn't accurate (and that was the reason we didn't feel like we should explicitly state the Chart & Meter colors on the HPE theme). 

In case I'm on the right track, then there is an issue with Chart rendering `accent-1` color instead of `graph-0` on HPE theme where `graph-0` is defined.
Screen Shot 2020-02-11 at 9 47 27 AM

The logic (before this PR) would always choose the `accent-1` color because
```
... if (theme.chart && theme.chart.color) colorName = theme.chart.color;
```
will always be true and as long the theme doesn't explicitly specify the color to be `graph-0` (and override base.js color) then the base.js color of `accent-1` will be picked as the chosen color.

I wanted us to discuss this issue since there are a few approaches to solve it:
1. Fix the HPE theme and all other themes that want to have `graph-*` colors and explicitly mention the color in its theme for every component (Meter, Chart, Diagram, Distribution)
2.  Continue with my current direction and fixing the logic in grommet.


ShimiSun:
@ericsoderberghp I've incorporated the inputs from our discussion.
I ran into the snapshot changes of Meter once putting the `graph`  variables back, and as much as I've tried to make them behave I wasn't able to get to where they were initially. 
Let's continue our discussion from this point.

I would say the theme display looks much better now for HPE theme.

ericsoderberghp:
Something doesn't look right in Chromatic

ShimiSun:
>Also, there is one chromatic change that doesn't look good on Meter, I was wondering how you picked the default values.

PR #3747, opened 2020-02-13
Changed Chart to allow onClick to work for bar and point types

ericsoderberghp:
#### What does this PR do?

Changed Chart to allow onClick to work for bar and point types

Chart.js

grommet designer

hmm???

automatic

yes

backwards compatible

### DIFF
--- a/src/js/components/Chart/doc.js
+++ b/src/js/components/Chart/doc.js
@@ -59,7 +59,8 @@ export const doc = Chart => {
       is required if more than one Chart is shown and they use color
       gradients.`),
     onClick: PropTypes.func.description(`Called when the user clicks on the
-     visualization. values[].onClick should be used for types bar and point.`),
+     visualization. Clicking on individual bars or points are handled via
+     values[].onClick for those types of charts.`),
     onHover: PropTypes.func.description(`Called with a boolean argument
       indicating when the user hovers onto or away from it.
       This is only available when the type is line or area.`),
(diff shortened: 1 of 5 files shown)


src/js/components/Chart/doc.js
```diff
@@ -58,8 +58,8 @@ export const doc = Chart => {
     id: PropTypes.string.description(`A unique identifier for the Chart. This
       is required if more than one Chart is shown and they use color
       gradients.`),
-    onClick: PropTypes.func.description(`Called when the user clicks on it.
-      This is only available when the type is line or area.`),
+    onClick: PropTypes.func.description(`Called when the user clicks on the
+     visualization. values[].onClick should be used for types bar and point.`),
```
ShimiSun: what happens if I use it for line chart?

ericsoderberghp:
> The code looks good.
> Since we are allowing `<path>` to be an interactive element now, I would like to see how it performs and whether it is compliant with accessibility standards of interactive elements.

We already had `onClick` on path elements for line and area types, this is just adding it for bar and point types.

src/js/components/Chart/doc.js
ericsoderberghp: Clicking on the line will trigger the function.

ShimiSun: So why does it say it should be used for 
>values[].onClick should be used for types bar and point

...and not for a line chart? I feel like I'm missing something.

ericsoderberghp: I've revised the documentation for `onClick` in an attempt to make it clearer.

PR #3758, opened 2020-02-18
Add tree component

edumoreira1506:
#### What does this PR do?
This PR adds a Tree component.

Design

Related to issue #3489 

Preview: 
Peek 2020-02-19 09-55

### DIFF
--- a/src/js/components/Tree/README.md
+++ b/src/js/components/Tree/README.md
@@ -1,0 +1,16 @@
+## Tree
+A tree.
+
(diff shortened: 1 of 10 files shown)


IanKBovard:
@edumoreira1506 . Some very nice work, but, this is just my opinion, but I don't see a need for a tree component in grommet. If we were to move forward on this there are a few things that need to be addressed.

I need to be able to control the component with a keyboard, as it stands i can't tab into it at all.

For the children, it's better to use the components grommet provides. This will allow users to pass in props to style the child nodes. RIght now, i see style.div, styled.h3, and so on.

We should be able to customize most of this component through theming and props(like ones found in Box, Heading...).

There are a few other things like testing, documentation, and cleaning up. But, this is where i would start improving this pr.

ShimiSun:
This PR is already covered by #4810

nnurmano:
Is there any chance to have this component?

PR #3763, opened 2020-02-19
added column-reverse to also calcualte gap height

britt6612:
#### What does this PR do?
This PR fixes the gapStyle to include column-reverse in the way that height is calculated for the box gap
Box/StyledBox.js
Storybook testing was done I used this simple code to check
```

  <Grommet theme={grommet}>
  <Box
      background="brand"
      gap="medium"
      direction="column-reverse"
      pad="medium"
    >
      <Text>Copyright</Text>
      <Anchor label="About" />
    </Box>
  </Grommet>

```

StoryBook

gapStyle function was taken into account if direction prop was column then push the height for the gap otherwise it was calculating the width which was being used for row 
I made a quick code sandBox in order to show my issue. 
https://codesandbox.io/s/elated-sea-2m3cs
no
no
backwards compatible

### DIFF
--- a/src/js/components/Box/StyledBox.js
+++ b/src/js/components/Box/StyledBox.js
@@ -576,7 +576,7 @@ const gapStyle = (directionProp, gap, responsive, border, theme) => {
   const hasBetweenBorder =
     border === 'between' || (border && border.side === 'between');
   const styles = [];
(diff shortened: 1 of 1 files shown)


ShimiSun:
Good catch! thank you for the contribution 🚀

PR #3766, opened 2020-02-20
Heading storybook example using extend, reorganize heading storybook files

IanKBovard:
Added a storybook example using extend property for heading.
Gave each storybook example for heading an individual file.


ShimiSun:
Looking good, thanks buddy!

PR #3767, opened 2020-02-20
selectTest error fix

IanKBovard:
Removed setValue in the onSelectChange in Select.js that was causing the issue. The test was expecting a string( 'one' ), but it received and object( {test: 'one} )

### DIFF
--- a/src/js/components/Select/stories/Basics.js
+++ b/src/js/components/Select/stories/Basics.js
@@ -45,7 +45,7 @@ const customRoundedTheme = deepMerge(grommet, {
 const SimpleSelect = ({ theme, ...rest }) => {
   const options = ['one', 'two'];
   const [value, setValue] = useState('');
-
+  console.log(value);
   return (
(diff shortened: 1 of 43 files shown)


src/js/components/Select/stories/Basics.js
```diff
@@ -45,7 +45,7 @@ const customRoundedTheme = deepMerge(grommet, {
 const SimpleSelect = ({ theme, ...rest }) => {
   const options = ['one', 'two'];
   const [value, setValue] = useState('');
-
+  console.log(value);
```
ShimiSun: Can we remove this console.log?

src/js/components/Select/stories/ManyOptions.js
```diff
@@ -16,7 +16,7 @@ class Option extends PureComponent {
   }
 }
 
-const dummyOptions = Array(2000)
+const dummyOptions = Array(90)
```
ShimiSun: why did this number changed?

Issue #3772, opened 2020-02-21
Select in Form not working properly in 2.11.0


mlonn:
The Selected value does not get selected

### Expected Behavior

The selected value should be selected
### Actual Behavior

The Selected value does not get selected

### Steps to Reproduce

1. https://codesandbox.io/s/form-issue-mzb5r 
1. Select a value, it doesn't save
1.  Change the version of grommet to 2.10.0
1. Select a value. now it does save.

### Your Environment

- Grommet version:
    2.11.0
- Browser Name and version:
    Chrome Version 79.0.3945.130 (Official Build) (64-bit)
- Operating System and version (desktop or mobile):
    macOS catalina  10.15.3

ShimiSun:
Thanks for reporting!

copperwall:
I took a shot at fixing this, but it looks like the `setValue` functionality was removed a week or two ago so I'm happy to change the fix if it's not quite right. I left a little more context on the PR #3773

ShimiSun:
Hi buddy, thanks again for the prompt report of the issue, I'll release a hotfix 2.11.1 to include this fix.
In the meanwhile the issue is fixed on the stable branch, so you can point your grommet package.json to "https://github.com/grommet/grommet/tarball/stable" instead of 2.11.0 and it should work.

ShimiSun:
Released officially on grommet v2.11.1, let us know if we can help with anything else. https://github.com/grommet/grommet/releases/tag/v2.11.1

mlonn:
Thanks for the quick response 😊

PR #3773, opened 2020-02-21
fix: select dropdown entry when the input isnt controlled

copperwall:
#### What does this PR do?

In 2.11.0, when using a select without passing value as a prop, selecting
a dropdown entry won't update the Select component to have that entry
selected. This sets the value using `setValue` from the `useFormContext` hook.

Before this change, the Select in the All storybook story would not update when a dropdown entry was clicked. Now it does.

This can be tested by creating a Select component with no value prop and no onChange prop and trying to click it.

It looks like this was intentionally removed in https://github.com/grommet/grommet/commit/caa63f3715dfd7f21a1a7f7bbce898d6fef87c92#diff-19aa4813d5217c4120837104c9b593ff, so I'm not sure if adding this back in is the desired fix, but I thought I'd give it a shot.

#3772 

I don't think so

It's a hotfix, so I don't think it would break any existing functionality

### DIFF
--- a/src/js/components/Select/Select.js
+++ b/src/js/components/Select/Select.js
@@ -83,7 +83,7 @@ const Select = forwardRef(
     const inputRef = useRef();
     const formContext = useContext(FormContext);
 
-    const [value] = formContext.useFormContext(name, valueProp);
+    const [value, setValue] = formContext.useFormContext(name, valueProp);
 
     const [open, setOpen] = useState(propOpen);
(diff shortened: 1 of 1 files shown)


IanKBovard:
looks good to me

ShimiSun:
Yes, that looks about right, I'll add a test for it on a separate PR. Thank you for your contribution buddy!

PR #3775, opened 2020-02-22
Refactor Carousel to use Hooks

sivakar12:
#### What does this PR do?
Refactors the Carousel component to use Hooks
src/js/components/Carousel.js
Ran the jest tests.
The snapshots had to be updated.
By viewing storybook stories
I made a small change to the logic. Previously, boundaries are checked and onLeft and onRight functions are added to DOM elements conditionally. Now, the boundary conditions are checked inside the onLeft and onRight callbacks. That caused the snapshots to be updated.
I thought this might be cleaner. If you think otherwise, I can change it back.
#3394 
No
No
Backwards compatible


src/js/components/Carousel/__tests__/__snapshots__/Carousel-test.js.snap
```diff
@@ -323,6 +323,7 @@ exports[`Carousel basic 1`] = `
           className="c8"
           disabled={true}
           onBlur={[Function]}
+          onClick={[Function]}
```
ShimiSun: Usually, when snapshots changed on refactoring it means that the functionality might have changed, will you be willing to take a look at why this snapshot changed? might mean we are doing something different with `onClick()`

sivakar12: I explained it in the background context

ShimiSun: thanks, I'll re-read

sivakar12:
I noticed that all the functional components that are wrapped with `withTheme` and `withFocus` also have `withForwardRef`.
Should I wrap this with `withForwardRef` as well?

ericsoderberghp:
> I noticed that all the functional components that are wrapped with `withTheme` and `withFocus` also have `withForwardRef`.
> Should I wrap this with `withForwardRef` as well?

We're slowly moving away from higher order components and recompose. With hooks, we're using `useContext()` for theme, and `forwardRef()` for refs. You can see an example of this in Button.js https://github.com/grommet/grommet/blob/master/src/js/components/Button/Button.js

sivakar12:
> > I noticed that all the functional components that are wrapped with `withTheme` and `withFocus` also have `withForwardRef`.
> > Should I wrap this with `withForwardRef` as well?
> 
> We're slowly moving away from higher order components and recompose. With hooks, we're using `useContext()` for theme, and `forwardRef()` for refs. You can see an example of this in Button.js https://github.com/grommet/grommet/blob/master/src/js/components/Button/Button.js

OK.
I want to know whether this particular component needs to forward the ref? I am getting some warning when I run the tests? The previous class component was not doing anything with the refs, however

ericsoderberghp:
If there were no refs before, no need to add them just to convert to hooks. ref can be added later.

sivakar12:
```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
      
      Check the render method of `WithTheme(Carousel)`.
          in Carousel (created by WithTheme(Carousel))
          in WithTheme(Carousel) (created by FocusableComponent)
          in FocusableComponent (created by WithTheme(Carousel))
          in WithTheme(Carousel)
          in div (created by StyledGrommet)
          in StyledGrommet (created by Grommet)
          in Grommet
```

This is the warning I get. withFocus or withTheme is assigning a ref to this component. When it was a class component it was fine. But functional components can't have one.
Shall I leave it like this since this is only a warning and our tests are all passing? Or shall I take the forwardRef as give it to some child (the Stack?)?

ericsoderberghp:
My apologies. I missed the detail that `recompose` was still being used. We should be removing `recompose`, `withTheme`, and `withFocus`. You can take a look at Button.js to see how this is done. Basically, replace `withTheme` with `const theme = useContext(ThemeContext);` and replace `withFocus` with `const [focus, setFocus] = useState();` along with `onFocus` and `onBlur` handlers.

src/js/components/Carousel/Carousel.js
```diff
@@ -204,9 +224,6 @@ if (process.env.NODE_ENV !== 'production') {
   // eslint-disable-next-line global-require
   CarouselDoc = require('./doc').doc(Carousel);
 }
-const CarouselWrapper = compose(
-  withFocus(),
-  withTheme,
-)(CarouselDoc || Carousel);
+const CarouselWrapper = CarouselDoc || Carousel;
```
ericsoderberghp: Also need `Carousel.displayName = 'Carousel';` around line 221.

ericsoderberghp:
Thanks for your contribution!

PR #3779, opened 2020-02-24
Add margin to List

taysea:
#### What does this PR do?
Allows user to put a margin on List item.

This was tested locally on storybook, and two tests have been added that test margin as a string and an object.

Apply a margin to List.

Margin was already listed in the documentation for List as a prop, however applying it had no effect.

N/A.

No.

No.

Yes, backwards compatible.


src/js/components/List/__tests__/__snapshots__/List-test.js.snap
```diff
@@ -1515,6 +1515,347 @@ exports[`List itemProps 1`] = `
...
+  -webkit-flex: 0 0 auto;
+  -ms-flex: 0 0 auto;
+  flex: 0 0 auto;
+}
+
+.c1 {
+  list-style: none;
+  margin: 0;
+  padding: 0;
+  margin: 48px;
```
ShimiSun: It doesn't look good that two margins are being applied, can we be sure it will pick the latest `margin` (line 1788) rather than the first (1786)?

taysea: Good call, yes let me adjust.

Issue #3536, opened 2019-11-08
Carousel bug: can't click child element


TimPchelintsev:
### Expected Behavior

Thanks for a great lib! I embed Facebook video as one of Carousel children. I want to play it.

### Actual Behavior

Can't play video. Click is not propagated, checked with console.log inside onClick, no output.

### URL, screen shot, or Codepen exhibiting the issue

[![Edit Grommet v2 Template](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/grommet-v2-template-2q4tc?fontsize=14)

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.7.10
- Browser Name and version: Chrome 79.0.3945.8
- Operating System and version (desktop or mobile): macOS Catalina 10.15

ShimiSun:
hi buddy, I don't see the video link or attachment, what am I missing?
It will be easier for us to provide a quicker solution if you can reproduce the issue on a codesandbox (here is a template https://codesandbox.io/s/m7mml8l0zj) or on one of our existing storybook examples https://storybook.grommet.io/?path=/story/all--all

TimPchelintsev:
Thank you for quick response, I have added link to initial comment, but anyway:
[![Edit Grommet v2 Template](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/grommet-v2-template-2q4tc?fontsize=14)

wortkotze:
any update for this ?

simonhambly:
Looking at the code of Carousel, it's using the Stack component. The browser pointer events are being blocked by the Stack layer that holds the carousel controls (and by child layers higher in the Stack).

As a workaround you can wrap this custom theme around the Carousel

```jsx
const pointerEventsFriendlyTheme = {
  box: {
    extend: 'pointer-events: none;'
  },
  anchor: {
    extend: 'pointer-events: auto;'
  },
  button: {
    extend: 'pointer-events: auto;'
  }
}

// in the component
<ThemeContext.Extend value={pointerEventsFriendlyTheme}>
  <Carousel >
    {/* carousel children*/}
  </Carousel>
</ThemeContext.Extend>
```


This disables pointer-events on Box elements (and all children), but re-enables them specifically for Anchors or Buttons.

simonhambly:
[...]

PR #3782, opened 2020-02-25
Carousel clickable elements fix :: Don't Merge ::

IanKBovard:
A proposed solution to carousel elements not being clickable. Meant to start a conversation about acceptable solutions. This example wraps the carousel controls around the cards, and updates the z-index of the active card on change. Can see it working in the Simple carousel storybook example.

### DIFF
--- a/src/js/components/Carousel/Carousel.js
+++ b/src/js/components/Carousel/Carousel.js
@@ -27,6 +27,9 @@ class Carousel extends Component {
 
   componentDidUpdate(prevProps) {
     const { play } = this.props;
+    if (typeof window !== 'undefined') {
+      this.handleActive();
+    }
     if (play && (!prevProps.play || play !== prevProps.play)) {
       this.play();
     } else if (!play && prevProps.play) {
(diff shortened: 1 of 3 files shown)


src/js/components/Carousel/Carousel.js
```diff
@@ -223,3 +238,19 @@ const CarouselWrapper = compose(
...
+            justify="between"
+          >
+            {showSelectors && (
+              <Box justify="end" fill={!showArrows && 'horizontal'}>
+                <Box direction="row" justify="center">
+                  {selectors}
+                </Box>
+              </Box>
+            )}
+          </Box> */
```
ShimiSun: Comments can be removed

src/js/components/Carousel/Carousel.js
```diff
@@ -38,6 +41,20 @@ class Carousel extends Component {
     clearInterval(this.timer);
   }
 
+  handleActive = () => {
+    const { activeIndex, priorActiveIndex } = this.state;
+    const activeCard = document.getElementById(`active${activeIndex}`);
+    if (activeCard) {
+      const activeParent = activeCard.parentElement;
+      activeParent.style.zIndex = 10;
```
ShimiSun: Can we ad some comments that explain the logic approach?

IanKBovard:
This pr redesigns carousel in a way, and was hoping to have a conversation about that. The workaround proposed here https://github.com/grommet/grommet/issues/3536 does not fix the issue completely, and has a few problems. 

My proposal is to wrap the controls around the carousel card, and change the z-index of each card as it comes up so that they can be clickable. However, im not sure this is the best approach.

The current implementation of carousel stacks the controls on top of the cards, meaning that they can never be clicked. That's how i understand it at least.

IanKBovard:
closing this issue.

Issue #3771, opened 2020-02-21
Grommet Layer onEsc and onClickOutside not working when modal is set false


karunakarareddyp:
Click on a button displaying Layer form at Top Right side corner of the window. And i am trying to close the Layer form by using onEsc and onClickOutside attributes but is not working when modal is set false. 
If i set modal is true then it's working. 

### Expected Behavior
The Layer form should close on by pressing Esc key and Click on outside of the layer form.

 <Layer
        onEsc={this.onCloseLayer}
        onClickOutside={this.onCloseLayer}
        modal={false}
        position='top-right'>
    <div>Content</div>
</Layer>
- Grommet version: 2.11.0
- Browser Name and version: Chrome 80.0.3987.116 
- Operating System and version (desktop or mobile): Mac OS (deesktop

IanKBovard:
i can confirm that this is an existing bug, not a regression. Using grommet v2.10 the issue still exists.

https://codesandbox.io/s/grommet-sandbox-cvmyo?fontsize=14&hidenavigation=1&module=%2Fsrc%2FLayer.js&theme=dark

karunakarareddyp:
Is there any plan to fix in upcoming releases?

ShimiSun:
Ye. does the codesandbox shared above reproduce the exact same thing you are describing?

mlonn:
Looks like the onEsc bug came here 
https://github.com/grommet/grommet/commit/595b3a05a47a843f0f22b133f4f5d109ba0f2a8c
when focus was removed for non modals. I changed the keyboard to be on the document instead. and moved the onClickOutside to not rely on the overlay

karunakarareddyp:
@ShimiSun Yes it is exactly the same issue.

IanKBovard:
working on this now, as well

IanKBovard:
It looks like onClickOutside never worked when layer modal was set to false. Was this intentional? Only the overlay when modal was set to true has the onclickoutside event.

mlonn:
It was more complicated than I thought..

ericsoderberghp:
What is the design scenario driving this? If the non-modal Layer is something like a notification toast, I don't see Esc or clicking outside being a reasonable way to close it. So, what is a scenario where the Layer isn't modal, meaning it doesn't have focus and the user is free to interact with the page underneath, where interacting with the page should make the Layer go away?

IanKBovard:
[...]

PR #3783, opened 2020-02-25
Fixes Grommet Layer onEsc and onClickOutside not working when modal is set false #3771

mlonn:
#### What does this PR do?
Fixes
Grommet Layer onEsc and onClickOutside not working when modal is set false #3771

Test the a layer in the storybook with/without the modal prop with both onEsc and onClickOutside

Tested in the Center story in the storybook

Test the a layer in the storybook with/without the modal prop with both onEsc and onClickOutside

No

No breaking change


ShimiSun:
Hi buddy, any chance you can merge the PR with the master branch? we refactored the Layer component to use hooks if that is too much for you, let us know!
Thanks for filing this PR!

src/js/components/Layer/LayerContainer.js
```diff
@@ -36,6 +36,26 @@ const LayerContainer = forwardRef(
     const containerRef = useRef();
     const layerRef = useRef();
 
+    useEffect(() => {
+      console.log('hej');
```
ShimiSun: Can we remove the console.log?

mlonn: woops

mlonn:
@ShimiSun Do you know why the tests in the MaskedInput are failing even though the snapshot is updated and the test pass locally?

ShimiSun:
@mlonn I'm not sure, what node version are you using? 
Can you try to remove the updated snapshot of MaskedInput and only keep the Layer changes?

mlonn:
@ShimiSun I was using node 10 at work. Ran tests again with node 12 now and it seems to work

mlonn:
Any plans on merging this?

ericsoderberghp:
> Any plans on merging this?

We started a discussion yesterday on this. So, it is on the radar screen for us. Apologies for the delays. I am hoping we will be able to get it in in the next couple of days.

src/js/components/Layer/LayerContainer.js
```diff
@@ -44,6 +44,25 @@ const LayerContainer = forwardRef(
     const containerRef = useRef();
     const layerRef = useRef();
 
+    useEffect(() => {
+      const handleClick = event => {
+        if (containerRef.current.contains(event.target)) {
```
ericsoderberghp: This approach won't work for a Layer that uses a Drop. For instance, if the Layer has a Select control, when the use opens the Select drop down and clicks on it, this will close the Layer. You can see this in the Layer -> Center story. I think this is why https://v2.grommet.io/layer#onClickOutside documents that it is only available for modal Layers.

mlonn: I changed it to Only work on modal layers and if you click on the overlay

src/js/components/Layer/LayerContainer.js
```diff
@@ -43,6 +43,23 @@ const LayerContainer = forwardRef(
     const anchorRef = useRef();
     const containerRef = useRef();
     const layerRef = useRef();
+    const overlayRef = useRef();
+    useEffect(() => {
+      const handleClick = event => {
+        console.log(event.target);
```
ericsoderberghp: Need to remove `console.log()`

src/js/components/Layer/LayerContainer.js
```diff
@@ -43,6 +43,23 @@ const LayerContainer = forwardRef(
...
+        if (onClickOutside && overlayRef.current === event.target) {
```
ericsoderberghp: Don't need to check for `onClickOutside` here, we already checked at line 54.

src/js/components/Layer/LayerContainer.js
```diff
@@ -127,8 +144,8 @@ const LayerContainer = forwardRef(
           dir={theme.dir}
         >
           <StyledOverlay
+            ref={overlayRef}
             plain={plain}
-            onMouseDown={onClickOutside}
```
ericsoderberghp: Can we just leave this as `onMouseDown={onClickOutsize}`? I'm not sure what adding the `useEffect()` is doing beyond this. Maybe I'm missing something?

src/js/components/Layer/LayerContainer.js
```diff
@@ -137,7 +154,11 @@ const LayerContainer = forwardRef(
     }
 
     if (onEsc) {
-      content = <Keyboard onEsc={onEsc}>{content}</Keyboard>;
+      content = (
+        <Keyboard target="document" onEsc={onEsc}>
```
ericsoderberghp: This change also breaks the behavior of the Layer -> Center story when the user has opened the Select drop down and presses Esc.

IanKBovard:
From #3771 

> A requirement of a non-modal dialogue is to let users interact with content beneath it(toasts, notifications, alerts). onclickoutside and onesc would break that.

this pr can be closed.

ShimiSun:
Hi buddy, thank you for the effort on this PR, the issue is already fixed and will be available on the following release of 2.16.2. Thank you!

Issue #3665, opened 2020-01-20
Select component flashes when having random length options


yoomury:
Hey, I found a pretty weird behaviour on the *Select* component. My use case randomly has *82* elements in the options array and apparently this causes some weird flashes if you scroll really fast to the bottom. If you change the array to has *80* or *100* elements, everything is fine.

### Expected Behavior

Select component works fine.
### Actual Behavior

Select component has some weird flashes if you use some randomly length arrays. 

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/s/elegant-tesla-hjipo

### Steps to Reproduce

1.  Scroll fast to the bottom of the options

IanKBovard:
@yoomury This _**is**_ strange! Checking it out now. It looks like there some extra length at the bottom of the list.

IanKBovard:
![Screen Shot 2020-01-29 at 12 52 17 PM](https://user-images.githubusercontent.com/21990914/73387441-3205f400-4296-11ea-892b-9ef6d877b733.png)

This div constantly resizing from 0 height to its original height when it should stay at 0 is causing the flashing.

![Kapture 2020-01-29 at 13 01 13](https://user-images.githubusercontent.com/21990914/73388247-7f369580-4297-11ea-975b-5b0910335398.gif)

anoopadvaitha:
Facing same issue. If the length of the object is more than 100 it seems to be working fine.

christophrath:
Facing the same issue here. actually dropdown must be higher than viewport for reproducing that issue

ShimiSun:
Thanks for reporting @christophrath

ShimiSun:
The fix will be available on our next release. thank you all for reporting and @IanKBovard for fixing!

PR #3784, opened 2020-02-25
Select flicker fix

IanKBovard:
Proposed solution to select flickering at certain option lengths. It looks like top changes its value at times and causes nextPageEnd to change value at the end of a selection list causing the flicker. Changing the calculation to math.ceil instead of math.floor seemed to have fixed the issue. But, I want to clarify the reason for each calculation.

### DIFF
--- a/src/js/components/InfiniteScroll/InfiniteScroll.js
+++ b/src/js/components/InfiniteScroll/InfiniteScroll.js
@@ -209,6 +209,7 @@ const InfiniteScroll = ({
   });
 
   if (endPage < lastPage || replace || onMore) {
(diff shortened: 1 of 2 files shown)


IanKBovard:
issue #3665

IanKBovard:
endpage fluctuates from being less than to equal to lastpage. causing the rerender of the end div. with math.ceil it fluctuates from equal to to greater than. I'm assuming this is why it works.

ericsoderberghp:
I think the real root of the problem is with how `lastPage` is being calculated. See https://github.com/grommet/grommet/pull/3637#issuecomment-591668393.

Issue #3785, opened 2020-02-25
Allow user to specify target for Layer positioning


taysea:
It'd be useful to be able to tell Layer what box to position relative to. Similar to how Drop has dropTarget. Right now, it only positions relative to the entire window.

### Expected Behavior

Be able to apply layerTarget to Layer which refers to an element that it should position itself relative to.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

ShimiSun:
Thank you for reporting! The issue is fixed and will be available on our next release.

Issue #2877, opened 2019-03-04
Support target on Layer component


wirtzdan:
Currently, Layer's don't support the target property (How its used in the Drop Component) and therefore can't be positioned relative to an element. 

### Expected Behavior

With target on a layer, it would be possible to add something like the Google Calendar popup for event creation. The background has an overlay similar to what the Layer already offers, what's missing though is the possibility to target the element. Therefore its only possible to align the Layer in the middle or on the side of the page. 

![image](https://user-images.githubusercontent.com/18514829/53721335-4d3c6980-3e63-11e9-9248-827b7d753e0f.png)

oorestisime:
I gave this a second look and what you are trying to achieve could be done with the Drop component.
Have a look at the this example https://storybook.grommet.io/?selectedKind=Drop&selectedStory=Progressive&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Fstories%2Fstories-panel

Let me know if this is what you might be looking for

wirtzdan:
But I'm not seeing a dark overlay in the given example of the Drop component. This is what's missing for me.

oorestisime:
True but that's just adding an overlay, i wonder whether we should add an option for drop to render like an overlay. lets wait to see what others think!

oorestisime:
@MikeWalrath maybe to discuss this at some point, i think it might be a good idea to expose the Overlay as a component. So that people could use it and for this case combine the drop with the overlay to reach the desired UI instead of changing the way layer works to support more positions.

ericsoderberghp:
For a general overlay, you could use Stack with `interactiveChild` to control where the interaction happens. But, you'd still need a way to position a drop/layer as requested. For the scenario mentioned, I'm not sure how the blue box on the left would be positioned, as that appears to be above the overlay, presumably positioned directly over the event on the calendar grid underneath. That's a bit of tricky alignment that can't be done with Drop or Layer in grommet today. But, I think the blue box is sort of necessary from a design standpoint to anchor the drop/layer visually.

I'm trying to find another example of Drop + overlay. Layer + overlay seems OK because the Layer is not connected tightly to an underlying component, just the center, side, or corner of the overall window. Drop, on the other hand, is tightly connected to the target element, but without an overlay, so the target element connection is easily visible.

Anyone have any other examples?

ShimiSun:
@wirtzdan, @oorestisime any other examples or inputs, or can we close the issue?

[...]

PR #3789, opened 2020-02-26
Changed Layer to add `target`

ericsoderberghp:
#### What does this PR do?

Changed Layer to add `target`.
This allows Layer to be used in an embedded context, such as for examples or within the grommet designer.

Layer/doc.js

new Layer -> Target storybook story

new Layer -> Target storybook story

automatic

yes

backwards compatible


src/js/components/Layer/__tests__/__snapshots__/Layer-test.js.snap
```diff
@@ -246,7 +226,7 @@ exports[`Layer animation false 1`] = `
   outline: none;
   pointer-events: all;
   z-index: 15;
-  position: fixed;
+  position: absolute;
```
ShimiSun: I noticed a couple of potential backward compatibility issues on `position` value that replaces 'fixed' with 'absolute'.

ericsoderberghp: Correct, that's because we need the LayerContainer to be positioned with respect to the outer Layer and not with respect to the whole window.

src/js/components/Layer/stories/Target.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { Box, Button, Grid, Grommet, Layer, Select } from 'grommet';
+import { grommet } from 'grommet/themes';
+
+const positions = ['center', 'start', 'end', 'top', 'bottom'];
```
ShimiSun: I noticed that both `end` and `start` are displaying the layer on the same position of `start`. I might be missing something. we might want to choose right/left.

ericsoderberghp: Ah, yes. I got ahead of myself. You must not be using Chrome or Firefox. :)

ShimiSun:
Linking to https://github.com/grommet/grommet/issues/3785

ShimiSun:
Linking to https://github.com/grommet/grommet/issues/2877

PR #3794, opened 2020-02-27
Test that checked if layer component unmounts from DOM

IanKBovard:
#### What does this PR do?
Adds a test that checks if layer unmounts from dom

Final test in Layer-test.js

Checked if test fails if layer is not unmounted

Can change the setShowLayer from true to false in SimpleLayer component.

None

No

No

backwards compatible


ShimiSun:
Awesome, thanks @IanKBovard

Issue #3807, opened 2020-03-01
MaskedInput allows input outside of the mask


copperwall:
When controlling the MaskedInput state using the `onChange` handler, the input will allow an extra character. A good example of this is the [Date](https://storybook.grommet.io/?path=/story/maskedinput--date) story in the storybook.

The mask specifies an input format for dd-mm-yyyy, but the user can enter a value of dd-mm-yyyyy.

### Expected Behavior

When a MaskedInput specifies a mask for a date (dd-mm-yyyy), it shouldn't let users add an additional character to the end of the input value

### Actual Behavior

When controlling the MaskedInput state using the `onChange` handler, the input will allow an extra character. A good example of this is the [Date](https://storybook.grommet.io/?path=/story/maskedinput--date) story in the storybook.

The mask specifies an input format for dd-mm-yyyy, but the user can enter a value of dd-mm-yyyyy.

### URL, screen shot, or Codepen exhibiting the issue

![2020-02-29 17 19 12](https://user-images.githubusercontent.com/2539016/75617732-16517000-5b18-11ea-9b57-ca5fbacf67e0.gif)

### Steps to Reproduce

1. Go to https://storybook.grommet.io/?path=/story/maskedinput--date
2. Enter 01/01/19999

### Your Environment

- Grommet version: 2.11.1
- Browser Name and version: Chrome 80
- Operating System and version (desktop or mobile): MacOS Catalina

ShimiSun:
Thank you for the contribution buddy.
adding a reference of a playground to this issue https://codesandbox.io/s/grommet-v2-template-tiuuf, switching between grommet versions will demonstrate the issue clearly.

ShimiSun:
Thanks again for reporting and making a contribution, sending some grommet love ❤️

PR #3808, opened 2020-03-01
fix(MaskedInput): set nextValue as event's value in onChangeInput

copperwall:
#### What does this PR do?

This PR fixes a bug in `MaskedInput` where a user can input a value that doesn't match the mask given for the input.

For example, if a mask was given for a date (dd-mm-yyyy) https://storybook.grommet.io/?path=/story/maskedinput--date, the user could input 01-01-19999, which is one character too many for the mask to be valid.

The issue is because the event given to the onChange function is the original change event from the user typing into the input. The `event.target.value` doesn't reflect the validation and masking done by `parseValue`. `onChange` does fire an event with the updated text on the input ref, but then also calls the `onChange` prop provided with the old event which doesn't have the updated input value.

This updates `event.target.value` with the `nextValue` build by `parseValue`.
https://github.com/grommet/grommet/blob/master/src/js/components/MaskedInput/MaskedInput.js#L207 

src/js/components/MaskedInput/MaskedInput.js

I've tested this manually

Going to http://localhost:9001/?path=/story/maskedinput--date on this branch's storybook and trying to enter a year on the Date story larger than 4 digits

So mutating the `event.target.value` seems to do the trick, but if you have other ideas about how to accomplish I'm totally open to hearing those out!

#3807 

Master:

This branch:

No

If you think that'd be useful

It's backwards compatible

### DIFF
--- a/src/js/components/MaskedInput/MaskedInput.js
+++ b/src/js/components/MaskedInput/MaskedInput.js
@@ -203,9 +203,14 @@ const MaskedInput = forwardRef(
         // Align with the mask.
         const nextValueParts = parseValue(mask, event.target.value);
         const nextValue = nextValueParts.map(part => part.part).join('');
-        if (value !== nextValue) setInputValue(nextValue);
-        if (onChange) onChange(event);
-        setValue(nextValue);
+        // eslint-disable-next-line no-param-reassign
+        event.target.value = nextValue;
+
+        if (value !== nextValue) {
+          setInputValue(nextValue);
+          if (onChange) onChange(event);
+          setValue(nextValue);
+        }
       },
       [mask, onChange, setInputValue, setValue, value],
     );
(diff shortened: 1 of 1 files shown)


src/js/components/MaskedInput/MaskedInput.js
```diff
@@ -203,9 +203,14 @@ const MaskedInput = forwardRef(
...
-        if (onChange) onChange(event);
-        setValue(nextValue);
+        // eslint-disable-next-line no-param-reassign
+        event.target.value = nextValue;
+
+        if (value !== nextValue) {
+          setInputValue(nextValue);
+          if (onChange) onChange(event);
+          setValue(nextValue);
+        }
```
ericsoderberghp: I like the direction to not notify if the value isn't being changed. But, I'm hesitant to reach into the event object and modify it. I think if we remove lines 206 and 207, it will still behave correctly. Can we try that?

copperwall: Oh wow, that totally worked, thanks for the suggestion! I wasn't really happy with those lines, but I thought they were necessary haha.

I'll make a commit to remove those.

ericsoderberghp:
Thanks for your contribution!

PR #3818, opened 2020-03-04
Update Meter test to fix console warning

taysea:
#### What does this PR do?
Changes Meter test to use 0.2 instead of '0.2' for opacity of background. The test should check that opacity accepts a number. This removes the console warning.

src/js/components/Meter/__tests__/Meter-test.js

Tests have been updated and no warning was printed to the console.

yarn test-update

A console warning was being printed since '0.2' (as a string) is not a valid value for opacity.

No.

No.

Yes, backwards compatible.


ShimiSun:
Thank you 💟

Issue #3788, opened 2020-02-26
Box "between" border places borders at end of gaps on mobile viewports


myktra:
The new *between* value for the `border` prop on boxes aims to place borders between items in a box, centered within the space specified by the `gap` prop. This is great! However, because of the way the border is specified in CSS as fixed pixel distance (rather than a percentage?) it can appear at the *end* of a gap on mobile viewports.

### Expected Behavior

The default behavior (arguably?) ought to be that the border remains placed in the *middle* of the gap.

### Actual Behavior

The border is placed at the end of the gap.

### URL, screen shot, or Codepen exhibiting the issue

![2020-02-26 09 58 14](https://user-images.githubusercontent.com/653521/75356970-a2486b00-587e-11ea-9072-cca6f77075af.gif)

### Steps to Reproduce

Use this code:

```jsx
<Box direction="column" gap="medium" border="between">
  <Box background="light-2">
    <Text>Hi!</Text>
  </Box>
  <Box background="light-2">
    <Text>Hi!</Text>
  </Box>
</Box>
```

Notice that when the viewport is shrunk, the gap decreases in height by about 50%. Because the `top` attribute of the CSS rule is set to a fixed `23.5px` (likely a formula based on spacings specified by the theme), the border appears at the end of the gap. If the `top` attribute of the CSS rule were set to `50%` it would appear in the middle of the gap as expected.

### Your Environment

- Grommet version: 2.11.1
- Browser Name and version: Chrome 79
- Operating System and version (desktop or mobile): macOS 10.15

ShimiSun:
Thanks for reporting buddy! I'm glad you are using a recent release and giving this new prop-value a try, looks like the issue starts here https://github.com/grommet/grommet/blob/8fbdf884f339944c5bbf94d6bbc99c228b7af386/src/js/components/Box/StyledBox.js#L589
Feel free to start a conversation with filing a potential PR fix.

myktra:
Thanks back! Right, it would appear in this code that for some reason we're not using the responsive gap/edge size, as we otherwise are elsewhere in this block.

See: https://github.com/grommet/grommet/blob/8fbdf884f339944c5bbf94d6bbc99c228b7af386/src/js/components/Box/StyledBox.js#L580

This would apply to the `top` and `left` attributes depending on the box `direction` prop:

https://github.com/grommet/grommet/blob/8fbdf884f339944c5bbf94d6bbc99c228b7af386/src/js/components/Box/StyledBox.js#L600
[...]

PR #3824, opened 2020-03-09
Changed Box to use align-self for gap element, so border between works

ericsoderberghp:
#### What does this PR do?

Before this change, the only way `border="between"` worked was if `align="stretch"`. This change fixes the gap elements to use `align-items: self;` so the between border will be seen and allow any value of `align` on the Box to work.

manual

no

yes

backwards compatible


ShimiSun:
linking to #3788

Issue #3706, opened 2020-01-30
Button focus persists after clicking it and retains an unfinished animated state


iMerica:
### Summary
Very strange button behavior. The animation is not completed until you click elsewhere. Which suggests buttons implicitly have "state". 

### Expected Behavior

I expect for buttons to never surprise me, maybe an animation on click or a color change on hover. Once you start trying to reinvent the wheel with surprising animations and finite state machines, buttons become too surprising and frustrating to use.

⚠️ **Clicking one button should not animate another button**  🚨

### Actual Behavior
Total confusion about something that is supposed to be dead simple.

### URL, screen shot, or Codepen exhibiting the issue
![weird](https://user-images.githubusercontent.com/487897/73498613-63102280-4383-11ea-936a-d18d5e760072.gif)

### Steps to Reproduce

```jsx
            <Button
              hoverIndicator={false}
              label={'A'}
              plain={true}
              focusIndicator={false}/>
            <Button
              color={'dark-1'}
              plain={false}
              primary
              focusIndicator={false}>
              B
            </Button>
```

- Click Sign up.
- Observe the awkward behavior that occurs: theres two aimations hovering and clicking.
- Clicking one button, animates the other

### Your Environment

- Grommet version: 2.10.0
- Browser Name and version: Chrome  79.0.3945.130 
- Operating System and version (desktop or mobile): macOS

iMerica:
Temporary Fix:

https://gist.github.com/iMerica/c1112b61509ddef1693b0abecf9726c7

ShimiSun:
Regardless of the related fix for this issue, I'm wondering if we can add a control to the `transition` i.e. animation of the Button. According to StyledButton, the transition is hardcoded (unless using plain). would that be something you'll be interested in?

larshisken:
@ShimiSun 

I know this is still a discussion. But I'd very much like to use this feature. Therefore I already created a PR.

ShimiSun:
@larshisken, yes! really nice follow-up, I'll get your PR in after the feedback is incorporated,  hoping you'll be able to use it soon!

jcfilben:
Looks like #3828 only solved part of the issue. When you click on a button and the move your mouse away and then hover on the button the hover styling doesn't get applied unless you click elsewhere before hovering https://codesandbox.io/s/grommet-v2-template-forked-b74uv?file=/index.js

taysea:
[...]

PR #3828, opened 2020-03-09
feat(Button): add transition to button theme

larshisken:
Signed-off-by: Lars Hisken <lars.hisken@bettyblocks.com>

#### What does this PR do?

It makes Button transitions configurable via the theme.

At the StyledButton; and by testing the `button.transition` theme keys.

I tested it manually with Storybook.

Add the transition keys (as stated in the updated docs) to the Button's theme. Inspect the button and check the updated CSS transition keys.

We're using Grommet in our product (with great success) and wish to theme it according to our style-guide. Animations are also part of that.

#3706

Yes.

Not sure.

Backwards compatible.


src/js/components/Button/StyledButton.js
```diff
@@ -153,13 +153,11 @@ const StyledButton = styled.button`
...
     `
-    transition-property: color,
-      background-color,
-      border-color,
-      box-shadow;
-    transition-duration: 0.1s;
-    transition-timing-function: ease-in-out;
+    transition-property: ${props.theme.button.transition.properties.join(',')};
+    transition-duration: ${props.theme.button.transition.duration}s;
+    transition-timing-function: ${props.theme.button.transition.curve};
```
ShimiSun: why not call it `timing` rather than `curve`?

larshisken: I agree, timing sounds more appropriate, I changed it.

src/js/components/Button/StyledButton.js
```diff
@@ -153,13 +153,11 @@ const StyledButton = styled.button`
...
     !props.plain &&
+    props.theme.button.transition &&
     `
-    transition-property: color,
-      background-color,
-      border-color,
-      box-shadow;
-    transition-duration: 0.1s;
-    transition-timing-function: ease-in-out;
+    transition-property: ${props.theme.button.transition.properties.join(',')};
```
ShimiSun: Is it possible to add space after the comma? just to the snapshot changes will be reverted to their default, and for the text to be more readable.

src/js/components/Button/doc.js
```diff
@@ -259,6 +259,21 @@ export const themeDoc = {
     type: 'string',
     defaultValue: '8px',
   },
+  'button.transition.timing': {
+    description: 'The transition timing function',
+    type: 'string',
+    defaultValue: 'ease-in-out',
+  },
```
ShimiSun: Can we organize the styles with alphabetical order and move this block after the `button.transition.properties` block?

src/js/components/Button/doc.js
```diff
@@ -259,6 +259,21 @@ export const themeDoc = {
...
+  'button.transition.duration': {
+    description: 'The transition duration in seconds',
```
ShimiSun:
```suggestion
    description: 'The length of time it will take for the element to transition between two defined states.',
```

src/js/components/Button/doc.js
```diff
@@ -259,6 +259,21 @@ export const themeDoc = {
...
+    type: 'number',
+    defaultValue: 0.1,
+  },
+  'button.transition.properties': {
+    description: 'The transition properties',
```
ShimiSun:
```suggestion
    description: 'The property you want to add the effect to.',
```

src/js/components/Button/doc.js
```diff
@@ -259,6 +259,21 @@ export const themeDoc = {
     type: 'string',
     defaultValue: '8px',
   },
+  'button.transition.timing': {
+    description: 'The transition timing function',
+    type: 'string',
```
ShimiSun:
```suggestion
    description: 'Describes how a transition will progress over one cycle of its duration and allowing it to change speed during its course.',
    type: 'string',
```

src/js/components/Button/StyledButton.js
larshisken: I tried this. But adding a space to the join() doesn't seem to have any effect on the output. Do you have a clue why this might happen?

ShimiSun: Not at the top of my head, have you checked this https://stackoverflow.com/questions/5080165/join-an-array-by-a-comma-and-a-space

larshisken: The code I'm using is:
```
transition-property: ${props.theme.button.transition.properties.join(', ')};
```
Means they should be separated by a comma and a space. Still styled-components doesn't seem to respect this, and trims the spaces in the generated CSS.

The same happens when I try this in a sandbox: https://codesandbox.io/s/hopeful-herschel-bzdk7. The logged output is separated by spaces, yet the contents of the style tag added by styled-components aren't. Weird!

Apparently I cannot control the generated CSS when joining the properties like this. Since this is about formatting in the generated snapshots, are you sure this is worth the effort?

src/js/components/Button/StyledButton.js
```diff
@@ -153,13 +153,11 @@ const StyledButton = styled.button`
...
     !props.plain &&
+    props.theme.button.transition &&
     `
-    transition-property: color,
-      background-color,
-      border-color,
-      box-shadow;
-    transition-duration: 0.1s;
-    transition-timing-function: ease-in-out;
+    transition-property: ${props.theme.button.transition.properties.join(', ')};
```
ShimiSun:
```suggestion
    transition-property: ${props.theme.button.transition.properties.join(',')};
```
Removing the space since it doesn't have any affect.

src/js/components/Button/doc.js
```diff
@@ -259,6 +259,23 @@ export const themeDoc = {
     type: 'string',
     defaultValue: '8px',
   },
+  'button.transition.duration': {
+    description: `The length of time it will take for the element to transition
+between two defined states.`,
```
ShimiSun:
```suggestion
between two states.`,
```

src/js/components/Button/doc.js
```diff
@@ -259,6 +259,23 @@ export const themeDoc = {
...
+    type: 'number',
+    defaultValue: 0.1,
+  },
+  'button.transition.properties': {
+    description: 'The property you want to add the effect to.',
```
ShimiSun:
```suggestion
    description: 'The CSS properties you want to add the transition to.',
```

src/js/components/Button/StyledButton.js
ShimiSun: I'm OK with how it is now.

ShimiSun:
Update the snapshots and we should be good to go with this one

Issue #3829, opened 2020-03-09
TextInput: onSuggestionsOpen/Close not functioning


taysea:
### Expected Behavior

I should be able to provide a function to onSuggestionsOpen/Close and have it execute when the suggestions opens or closes. The expected functionality works with grommet 2.10.0 but no longer works on 2.11.0, so it looks like something happened between those releases.

### Actual Behavior

These props don't execute the function I provide to them.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/onsuggestionsclose-y566t

Note: on 2.10.0, the console statements print, but once I move up to 2.11.0, they don't.

### Steps to Reproduce

1. In grommet 2.11.0, Open the codesandbox and click on the textinput.
2. Check the console (nothing should be printed)
3. Change to grommet 2.10.0, and click on the textinput again.
4. There should be a console statement saying "Suggestions opening"
5. Click out of the textinput so the suggestions close, and check there is a console statement saying "Suggestions closing"

### Your Environment

- Grommet version: 2.11.1
- Browser Name and version: Chrome Version 80.0.3987.132 (Official Build) (64-bit)
- Operating System and version (desktop or mobile): MacOS Mojave 10.14.6

PR #3830, opened 2020-03-10
Fix onSuggestionsOpen/Close

taysea:
#### What does this PR do?
Adds in conditions to allow onSuggestionsOpen/Close to be called when the suggestions drop opens.

Tests have been added to the testing suite in this PR. I reverted my changes, and the tests I wrote failed. However, with the changes added in, they pass.

In storybook, you can add this code to the TextInput Suggestions story. If you click on the text input, you'd expect a console statement to be printed but there isn't one. Similarly, if you click outside of the suggestions to close them, no statement is printed.

```
onSuggestionsOpen={() => console.log('opening suggestions')}
onSuggestionsClose={() => console.log('closing suggestions')}
```

#3829 

No.

No.

### DIFF
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -363,8 +363,7 @@ const TextInput = forwardRef(
               setFocus(true);
               if (suggestions && suggestions.length > 0) {
                 announce(messages.suggestionsExist);
-                setShowDrop(true);
-                if (onSuggestionsOpen) onSuggestionsOpen();
+                openDrop();
               }
               if (onFocus) onFocus(event);
             }}
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -176,6 +176,7 @@ const TextInput = forwardRef(
     const closeDrop = () => {
       setShowDrop(false);
       if (messages.onSuggestionsClose) onSuggestionsClose();
+      if (onSuggestionsClose) onSuggestionsClose();
     };
 
     const onNextSuggestion = event => {
@@ -363,6 +364,7 @@ const TextInput = forwardRef(
               if (suggestions && suggestions.length > 0) {
                 announce(messages.suggestionsExist);
                 setShowDrop(true);
+                if (onSuggestionsOpen) onSuggestionsOpen();
               }
               if (onFocus) onFocus(event);
             }}
--- a/src/js/components/TextInput/__tests__/TextInput-test.js
+++ b/src/js/components/TextInput/__tests__/TextInput-test.js
@@ -170,7 +170,6 @@ describe('TextInput', () => {
     );
 
     fireEvent.focus(getByTestId('test-input'));
-    fireEvent.change(getByTestId('test-input'), { target: { value: ' ' } });
     setTimeout(() => {
       expectPortal('text-input-drop__item').toMatchSnapshot();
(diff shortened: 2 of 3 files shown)


src/js/components/TextInput/TextInput.js
```diff
@@ -363,6 +364,7 @@ const TextInput = forwardRef(
               if (suggestions && suggestions.length > 0) {
                 announce(messages.suggestionsExist);
                 setShowDrop(true);
+                if (onSuggestionsOpen) onSuggestionsOpen();
```
ShimiSun: I think this call, and the call on line 366 are redundant. 
what about leveraging the `openDrop` function instead of those two?
```suggestion
                openDrop();
```
`openDrop();` will take care of `setShowDrop(true);`  and calling onSuggestionsOpen.
I noticed no one else is calling this function and that may be the right place to use it. 
thoughts?

src/js/components/TextInput/__tests__/TextInput-test.js
```diff
@@ -155,6 +155,63 @@ describe('TextInput', () => {
     }, 50);
   });
 
+  test('calls onSuggestionsOpen', done => {
```
ShimiSun: Thank you, we needed it :)

src/js/components/TextInput/__tests__/TextInput-test.js
```diff
@@ -155,6 +155,63 @@ describe('TextInput', () => {
...
+          id="item"
+          name="item"
+          suggestions={['test', 'test1']}
+          onSuggestionsOpen={onSuggestionsOpen}
+        />
+      </Grommet>,
+    );
+
+    fireEvent.focus(getByTestId('test-input'));
+    fireEvent.change(getByTestId('test-input'), { target: { value: ' ' } });
```
ShimiSun: I'm thinking this call might be redundant, `expect(onSuggestionsOpen).toBeCalled();` should happen just by focus and not necessarily by a change event. If it tests something else we can put it on a separate test case.

src/js/components/TextInput/__tests__/TextInput-test.js
```diff
@@ -155,6 +155,63 @@ describe('TextInput', () => {
...
+          name="item"
+          suggestions={['test', 'test1']}
+          onSuggestionsClose={onSuggestionsClose}
+        />
+      </Grommet>,
+    );
+    expect(container.firstChild).toMatchSnapshot();
+
+    fireEvent.focus(getByTestId('test-input'));
+    fireEvent.change(getByTestId('test-input'), { target: { value: ' ' } });
```
ShimiSun: I think the same comment applies here.

src/js/components/TextInput/TextInput.js
ShimiSun: for quick reference https://github.com/grommet/grommet/blob/f4611c750e297e560d35a6403f2eae41a20a4ad7/src/js/components/TextInput/TextInput.js#L166

src/js/components/TextInput/__tests__/TextInput-test.js
taysea: Yes, you're right. Just focus is sufficient.

src/js/components/TextInput/TextInput.js
taysea: Good call.

PR #3833, opened 2020-03-10
IE11 layer background fix #3781

### LINKED ISSUE #3781: Layer opaque background in IE11
Layer opaque background in ie11
### Expected Behavior
When selecting an item that pops up an overlay, you should continue to see the screen behind it:
Screen Shot 2020-02-25 at 8 36 44 AM

### Actual Behavior
When clicking that item it instead causes the background to become gray (opaque):
Grommet_Overlay_Issue

### Steps to Reproduce

1. In Internet Explorer 11, click on an item that causes an overlay to pop up
2. Observe the screen behind the overlay is gray (opaque)

### Your Environment

- Grommet version 2.11.0
- IE11 11.1685.16299.0
- Windows 10 Desktop 10.0.16299 Build 16299

IanKBovard:
solves issue #3781 

StyledLayer in StyledLayer.js had a background property of unset, causing it to have a white background in ie11. changing this value to transparent will fix this issue.

However, I am not sure how this will affect any intended functionality of the layer component.  I assume it's background is set to unset because it inherits its color from the theme, and if it doesn't get it from theme it'll go with it's initial value.

This can be tested in IE11 with yarn storybook. I used browserstack. 

Might not be backwards compatible.


ShimiSun:
Can you update the snapshots to fix jest? @IanKBovard

ShimiSun:
https://stackoverflow.com/questions/23793627/css-unset-initial-in-internet-explorer

IanKBovard:
ie11 does not support initial or unset. i'm guessing that's why it doesn't work.

https://caniuse.com/#feat=css-unset-value

ShimiSun:
yea, the stackoverflow link I attached is agreeing with that. 
Can you confirm Safari/Chrome and FF are intact with his change? and IE is behaving as expected?

IanKBovard:
tested with storybook on chrome, safari, and ff, and they all work the same. Having trouble running storybook with browserstack on ie11. My question is, why was unset used in the first place? If you plan on having theme control the background, wouldn't this solution be unacceptable?

ShimiSun:
>If you plan on having theme control the background, wouldn't this solution be unacceptable?

why would you think that? is this solution blocking the theme from setting the background?

IanKBovard:
this solution changes the layerContainer's background to transparent. I'm just trying to clarify why unset was used at all, and if changing it breaks any intended functionality.

ShimiSun:
Thanks @IanKBovard !

Issue #3825, opened 2020-03-09
Bug: Select in Form, the value is not controlled by value property.


jim-king-2000:
### Expected Behavior

    const [a, setA] = useState();
    <Form>
      <Select name='some_name' value={a} />
      <Button onClick={() => setA('a')} />
    </Form>

We use state `a` to control the value of the Select component. When clicking the button, we set state `a` to string 'a'.

The content of Select component should be controlled by the value property. The value changes, the content changes.

### Actual Behavior

Click the button, the state will be set to 'a'. But the string in the Select control wouldn't change.

By the way, the icon `<Qr />` looks weird.

### Steps to Reproduce

A. Sandbox
  1. See https://codesandbox.io/s/distracted-taussig-hdvjb
  2. click "text" button, see no effect.

B. Github
  1. goto https://github.com/jim-king-2000/nextbugrepro
  2. clone the code
  3. `cd nextbugrepro && npm i && npm run dev`
  4. access http://localhost:3000
  5. click "text" button, see no effect.

### More Information

1. The bug occurs with two conditions. First, there is a `name` property in Select. Second, Select is in a Form. Without any of the two, it works.
2. It works when we select a value in the dropdown menu of the Select component.
3. It can't be reproduced on grommet@2.10.0. (But grommet@2.10.0 contains other bugs)

### Your Environment

- Grommet version: 2.11.1
- Browser Name and version: Chrome 80
- Operating System and version (desktop or mobile): win10

ShimiSun:
Fixed on the master branch. Hoping we'll have a release soon. thanks for reporting!

ShimiSun:
Fixed on the master branch. Hoping we'll have a release soon. thanks for reporting!

jim-king-2000:
:smiley:

jim-king-2000:
What about the `<Qr />` icon bug?

ShimiSun:
Our designers may need to take a look at that.

Issue #3812, opened 2020-03-03
FormField : onChange parameter undefined in "component" property


emmanuelgratuze:
Since version 2.11.1, the props.onChange parameter seems to be undefined when giving a function to component property in FormField.

### Expected Behavior
In the previous release (2.10.0), the component property callback was returning the parameter onChange :
```
<FormField
  component={({ onChange, ...props }) => {
    // onChange was defined here
    return <CustomInput onChange={onChange} {...props} />;
  }
/>
```

### Actual Behavior
onChange property is not defined anymore when component property is being called.

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/s/custom-input-onchange-undefined-k9zml

In the codesandbox, changing the grommet version from 2.11.1 to 2.10.0 solves the problem.

### Your Environment
- Grommet version: 2.11.1
- Browser Name and version: Google Chrome 80.0.3987.122
- Operating System and version (desktop or mobile): macOS Catalina 10.15.3, desktop

Thank you!

ShimiSun:
Thank you for reporting buddy, I was able to fix it by adding a `name` prop to the component field https://codesandbox.io/s/custom-input-onchange-undefined-fixed-qftc8. we will add documentation for it. Thanks.
linking to #3777 #3791

ShimiSun:
...That being said, we are working on a fix that potentially won't need the workaround. I'll keep this issue updated as we go.

ShimiSun:
This issue is fixed, we will post a release soon to include this fix. thanks for reporting!

PR #3836, opened 2020-03-11
Changed Form to fix an issue with setting value from component

ericsoderberghp:
#### What does this PR do?

Changed Form to fix an issue with setting value from component

Form.js

black-box story

#3825 

no

yes

backwards compatible


ShimiSun:
Can we add a test for this changed behavior?

ShimiSun:
linking to #3812

Issue #3728, opened 2020-02-05
Ability to overwrite onSort in DataTables


francescodist:
As for the `onSearch` method I would like to be able to pass a method to the `onSort` property that, when defined, will overwrite the default behaviour for sorting. 
### Expected Behavior
The property would be used to just send to callback from the parent the property and direction (basically the current sort configuration) without making any change to the data in the table. This way the developer is free to do any transformation to the data.

### Actual Behavior
There is no way to have custom behaviors for sorting

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.10.0
- Browser Name and version:
- Operating System and version (desktop or mobile):

If you accept external PRs, I can work on this as I will need to do it anyway for the project I am currently working on

ShimiSun:
Hi @francescodist, in Grommet v2.11.2 we released new props for `sort` and `onSort` https://v2.grommet.io/datatable#onSort. 
Feel free to reopen the issue with a concrete example in case it doesn't meet your needs, from what you are describing it should work as expected. good luck!

ShimiSun:
#3840

PR #3840, opened 2020-03-11
Changed DataTable to add sort and onSort

ericsoderberghp:
#### What does this PR do?

Changed DataTable to add sort and onSort

DataTable/doc.js

Added DataTable Sort story

DataTable Sort story

automatic

yes

backwards compatible

### DIFF
--- a/src/js/components/DataTable/DataTable.js
+++ b/src/js/components/DataTable/DataTable.js
@@ -34,12 +34,14 @@ const DataTable = ({
   onClickRow, // removing unknown DOM attributes
   onMore,
   onSearch, // removing unknown DOM attributes
+  onSort: onSortProp,
   replace,
   pad,
   primaryKey,
   resizeable,
   rowProps,
   size,
+  sort: sortProp,
   sortable,
   step = 50,
   ...rest
@@ -62,7 +64,7 @@ const DataTable = ({
   const [filters, setFilters] = useState(initializeFilters(columns));
 
   // which column we are sorting on, with direction
-  const [sort, setSort] = useState({});
+  const [sort, setSort] = useState(sortProp || {});
 
   // the data filtered and sorted, if needed
   const adjustedData = useMemo(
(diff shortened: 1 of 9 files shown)


src/js/components/DataTable/DataTable.js
```diff
@@ -105,9 +107,13 @@ const DataTable = ({
...
-    const ascending =
-      sort && property === sort.property ? !sort.ascending : true;
-    setSort({ property, ascending });
+    let direction;
+    if (!sort || property !== sort.property) direction = 'asc';
+    else if (sort.direction === 'asc') direction = 'desc';
+    else direction = 'asc';
+    const nextSort = { property, direction };
+    setSort(nextSort);
+    if (onSortProp) onSortProp(nextSort);
```
ShimiSun: I'm guessing this is an internal calculation on our end and there are no backward compatibility issues here.

src/js/components/DataTable/doc.js
```diff
@@ -159,6 +159,13 @@ export const doc = DataTable => {
       names and values which are the search text strings. This is typically
       employed so a back-end can be used to search through the data.`,
     ),
+    onSort: PropTypes.func.description(
+      `When supplied, this function will be called with an event object
+      that includes a 'sort' property containing an object with a
+      'property' property that indicates which property
+      is being sorted on and a 'direction' property that will either be
+      'asc' or 'desc'. onChange={({ sort: { property, direction }}) => {}}`,
```
ShimiSun: why is the example not stating `onSort` to connect the context better? I think the onChange should be replaced.

src/js/components/DataTable/DataTable.js
ericsoderberghp: correct

src/js/components/DataTable/doc.js
ericsoderberghp: Good catch! I've updated the documentation.

ericsoderberghp:
Yep, `sortable` is just there for backwards compatibility.

ShimiSun:
Should we mark `sortable` as deprecated and provide a warning? Also, maybe just a comment on the docs that it is not the best practice and we recommend using onSort/sort instead.

ericsoderberghp:
Hmm... Now that I think about it, if you want the current behavior, where DataTable does it all and you don't want to bother with it, you wouldn't want to set `sort` or `onSort`, to have them be functional.

ericsoderberghp:
So, `sortable` does provide a, very minor, capability that you can't achieve with the other two. Unless, maybe we allow `sort={true}`. Seems minor.

ShimiSun:
linking to https://github.com/grommet/grommet/issues/3728

PR #3842, opened 2020-03-13
Fix Storybook TypeScript errors

jecassis:
When testing a different pull request, I encountered Storybook build failures due to TypeScript errors.

#### What does this PR do?
This pull request clears the errors and allows the build to finish and Storybook to be available in the browser. The errors are cleared by initializing `useState()` to something other than `undefined`.

The `useState()` initialization in each of these files:

- `src/js/components/List/stories/typescript/selected.tsx`
- `src/js/components/Meter/stories/typescript/MultipleValues.tsx`
- `src/js/components/Tabs/stories/typescript/Controlled.tsx`

Storybook builds and components function as expected in the browser.

Storybook builds and List, Meter, and Tabs function in the browser.

None.

None at the moment.

Not available.

No.

No preference.

Not applicable.

### DIFF
--- a/src/js/components/List/stories/typescript/selected.tsx
+++ b/src/js/components/List/stories/typescript/selected.tsx
@@ -5,30 +5,28 @@ import isChromatic from 'storybook-chromatic/isChromatic';
 import { Grommet, Box, List } from 'grommet';
 import { grommet } from 'grommet/themes';
 
-
 export const locations = [
-    'Boise',
-    'Fort Collins',
-    'Los Gatos',
-    'Palo Alto',
-    'San Francisco',
-  ];
-  
-  export const data = [];
-  
-  for (let i = 0; i < 40; i += 1) {
-    data.push({
-      entry: `entry-${i + 1}`,
-      location: locations[i % locations.length],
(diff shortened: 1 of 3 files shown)


ShimiSun:
What may be the reason I'm not experiencing those errors on my end?

jecassis:
Honestly, I'm not familiar enough with TypeScript to give you a good answer. The errors were blocking on my end to test PR #3843 manually using Storybook. So I worked around them and these are those workarounds. CI seems unaffected in either PR as well.

For reference, this was encountered on Windows 10 1909 with TypeScript 3.7.5.

ShimiSun:
Thanks, I tested it locally and it looks good.
Wondering how come we didn't catch it sooner, @dallas any thoughts?

dallas:
Not sure. It definitely seems like TypeScript should complain for everybody in regards to these implicit `undefined` values (without setting up any type expectations). It could be the version of `@types/react` that got installed or even the version of Storybook (or some version of Storybook’s dependencies) that’s complaining for @jecassis but isn’t complaining for you locally, or me locally, or in GitHub automated builds and whatnot.

Issue #3832, opened 2020-03-10
required FormField with Select child always rejects submit


blakehilliard:
required FormField with Select child always rejects submit

### Expected Behavior

If you do `<FormField required ...><Select .../></FormField>`, it should only be rejected if you have not selected anything from the drop-down menu.

### Actual Behavior

Even if you select something, submission is rejected saying that the field is required.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-required-select-bzoi1

### Steps to Reproduce

Open the above codepen. Select something from the drop-down menu, then click "Submit".

### Your Environment

- Grommet version: 2.11.1
- Browser Name and version: Chrome 79.0.3945.130 (Official Build) (64-bit)
- Operating System and version (desktop or mobile): Windows 10, desktop

ericsoderberghp:
I think this is due to the Select not having a `name` property. Here's a fork showing it: https://codesandbox.io/s/grommet-required-select-nxfle

blakehilliard:
That works, thanks. "name" isn't listed here, so I didn't know I could use it with Select: https://v2.grommet.io/select

ShimiSun:
@blakehilliard thanks for raising the docs issue, I added a PR to enhance the documentation.

ShimiSun:
I'll be closing this issue now that the docs are improved, thanks for reporting!

PR #3845, opened 2020-03-13
Added name prop for FormField inputs

ShimiSun:
#### What does this PR do?
Adds `name` prop for input fields that are used in FormField.

doc.js(s)

lot's of issues filed around missing `name` on FormField, I thought it will be good to document it.

https://github.com/grommet/grommet/issues/3832
#3812 
auto

nope
backwards compatible


src/js/components/Select/doc.js
```diff
@@ -102,6 +102,9 @@ export const doc = Select => {
     multiple: PropTypes.bool.description(
       'Whether to allow multiple options to be selected.',
     ),
+    name: PropTypes.string.description(
+      `The name of the attribute when in a Form or FormField..`,
```
negreirosleo: hey, why two dots instead of one or three?

ShimiSun: nice catch! ❤️

PR #3848, opened 2020-03-13
Support hidden input labels with FormField

halocline:
#### What does this PR do?
Allows for a `FormField` label to be visually hidden, yet remain in the DOM to be associated with the input element and assist users utilizing screen readers.

`FormField.js`

Storybook story, plus jest tests

Review Storybook story

Labels for input fields assist with accessibility, however a lot of designs present form fields without labels. Making it easy to include a label, yet visually hide it, will help developers ensure the forms they build are accessible.

Screen Shot 2020-03-13 at 3 17 52 PM

Yes.

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/Button/README.md
+++ b/src/js/components/Button/README.md
@@ -560,6 +560,36 @@ Defaults to
 undefined
 ```
 
+**button.primary.extend**
+
+Any additional style for a primary button. Expects `string | (props) => {}`.
+
+Defaults to
+
+```
+undefined
+```
+
+**button.primary.active.border.color**
+
+The color of the border for active primary buttons. Expects `string | { dark: string, light: string }`.
+
+Defaults to
+
+```
+undefined
+```
+
+**button.primary.active.extend**
+
+Any additional style for an active primary button. Expects `string | (props) => {}`.
+
+Defaults to
+
+```
+undefined
+```
+
 **button.size.small.border.radius**
 
 The border corner radius. Expects `string`.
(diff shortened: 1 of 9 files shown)


src/js/components/FormField/FormField.js
```diff
@@ -49,6 +49,21 @@ const validateField = (required, validate, messages) => (value, data) => {
 };
 
 const FormFieldBox = styled(Box)`
+  ${props =>
+    props.label &&
+    props.label.hidden &&
+    // Visually hiding label, but leaving in DOM for screen readers.
+    // For background on approach see https://www.w3.org/WAI/tutorials/forms/labels/#note-on-hiding-elements
+    `label {
```
ericsoderberghp: Is this selector specific enough? For instance, if I have custom children of the FormField that includes label elements for other things, I don't think this should match those.

Instead, how about doing something like `const LabelText = styled(Text)` and use that?

halocline: Implemented this suggestion.

halocline:
> Hm not sure i understand the need of these changes, why use formfield if we don't need a label? we are effectively adding it in the dom even if it is not there.
> probably missing the use cases here so open for discussion :)

The intent is to still meet accessibility standards for users using screen readers, even when a design relies on placeholder text instead of a label. However, per the [Web Accessibility Initiative](https://www.w3.org/WAI/tutorials/forms/instructions/#placeholder-text) "placeholder text is not a replacement for labels." From the same resource, here are their recommendations for [hiding label text](https://www.w3.org/WAI/tutorials/forms/labels/#hiding-label-text).

I am also open to alternate approaches or to whether meeting accessibility standards is important in these scenarios.

PR #3856, opened 2020-03-17
Changed Accordion and AccordionPanel to use hooks

ericsoderberghp:
#### What does this PR do?

Changed Accordion and AccordionPanel to use hooks

Accordion.js

unit + storybook

storybook

no

no?

backwards compatible

### DIFF
--- a/src/js/components/AccordionPanel/AccordionPanel.js
+++ b/src/js/components/AccordionPanel/AccordionPanel.js
@@ -1,136 +1,116 @@
-import React, { useState } from 'react';
-import { compose } from 'recompose';
-
-import { withTheme } from 'styled-components';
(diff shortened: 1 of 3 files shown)


src/js/components/AccordionPanel/AccordionPanel.js
```diff
@@ -1,136 +1,116 @@
...
+    );
+
+    return (
+      <Box ref={ref} flex={false} onClick={onClick}>
+        <Button
+          role="tab"
+          aria-selected={active}
+          aria-expanded={active}
+          onClick={event => {
+            console.log('!!! clicked');
```
ShimiSun: I think we can remove the console message.

ericsoderberghp: oops, good catch!

PR #3858, opened 2020-03-17
Changed DataTable to add verticalAlign under columns

ericsoderberghp:
#### What does this PR do?

Changed DataTable to add verticalAlign under columns

doc.js

none

automatic

yes

backwards compatible

### DIFF
--- a/src/js/components/DataTable/index.d.ts
+++ b/src/js/components/DataTable/index.d.ts
@@ -21,13 +21,14 @@ export interface DataTableProps {
   gridArea?: GridAreaType;
   margin?: MarginType;
   columns?: {
-    align?: "center" | "start" | "end", 
+    align?: "center" | "start" | "end",
     aggregate?: "avg" | "max" | "min" | 'sum',
     footer?: React.ReactNode | {aggregate?: boolean},
     header?: string | React.ReactNode | {aggregate?: boolean},
     primary?: boolean,property: string,
     render?: ((...args: any[]) => any),
-    search?: boolean,sortable?: boolean}[];
+    search?: boolean,sortable?: boolean,
+    verticalAlign?: "middle" | "top" | "bottom"}[];
   data?: {}[];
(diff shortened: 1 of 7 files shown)


src/js/components/TableCell/TableCell.js
```diff
@@ -10,6 +10,12 @@ import { Box } from '../Box';
 import { TableContext } from '../Table/TableContext';
 import { StyledTableCell } from '../Table/StyledTable';
 
+const verticalAlignToJustify = {
+  middle: 'center',
+  top: 'start',
+  bottom: 'end',
```
ShimiSun: Would it make more sense to allow the Box `justify` values and allow all `justify` [options](https://v2.grommet.io/box#justify), this way we can avoid the dispatching of the values above, and the user is already familiar with the potential values? 
we can also expose just a portion of the `justify` values ('center', 'start', 'end') if it wouldn't make sense to allow all values.

ericsoderberghp: I actually started down that path but then realized that `TableCell` uses `verticalAlign` for the `<td />`, where we can't use `justify`. So, I think tying to how `<td />` works makes more sense.

src/js/components/DataTable/index.d.ts
```diff
@@ -21,13 +21,14 @@ export interface DataTableProps {
...
   columns?: {
-    align?: "center" | "start" | "end", 
+    align?: "center" | "start" | "end",
     aggregate?: "avg" | "max" | "min" | 'sum',
     footer?: React.ReactNode | {aggregate?: boolean},
     header?: string | React.ReactNode | {aggregate?: boolean},
     primary?: boolean,property: string,
     render?: ((...args: any[]) => any),
-    search?: boolean,sortable?: boolean}[];
+    search?: boolean,sortable?: boolean,
```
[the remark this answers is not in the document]
ericsoderberghp: good catch

PR #3859, opened 2020-03-17
Changed Chart to add dash

ericsoderberghp:
#### What does this PR do?

Changed Chart to add dash

Chart/doc.js

Add unit test and Chart -> Dash story

Chart -> Dash story

automatic

yes

backwards compatible

### DIFF
--- a/src/js/components/Chart/doc.js
+++ b/src/js/components/Chart/doc.js
@@ -58,6 +58,9 @@ export const doc = Chart => {
     id: PropTypes.string.description(`A unique identifier for the Chart. This
       is required if more than one Chart is shown and they use color
       gradients.`),
+    dash: PropTypes.oneOfType([PropTypes.bool, PropTypes.string])
+      .description(`Whether to use dashed lines for line or bar charts.`)
+      .defaultValue(false),
     onClick: PropTypes.func.description(`Called when the user clicks on the
      visualization. Clicking on individual bars or points are handled via
      values[].onClick for those types of charts.`),
--- a/src/js/components/Chart/doc.js
+++ b/src/js/components/Chart/doc.js
@@ -58,7 +58,7 @@ export const doc = Chart => {
     id: PropTypes.string.description(`A unique identifier for the Chart. This
       is required if more than one Chart is shown and they use color
       gradients.`),
-    dash: PropTypes.oneOfType([PropTypes.bool, PropTypes.string])
+    dash: PropTypes.bool
       .description(`Whether to use dashed lines for line or bar charts.`)
       .defaultValue(false),
     onClick: PropTypes.func.description(`Called when the user clicks on the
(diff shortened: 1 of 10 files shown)


src/js/components/Chart/doc.js
```diff
@@ -58,6 +58,9 @@ export const doc = Chart => {
     id: PropTypes.string.description(`A unique identifier for the Chart. This
       is required if more than one Chart is shown and they use color
       gradients.`),
+    dash: PropTypes.oneOfType([PropTypes.bool, PropTypes.string])
+      .description(`Whether to use dashed lines for line or bar charts.`)
```
ShimiSun: This description is clear for the boolean value, but not for the string value.
I'm thinking we should add that the string will be used as a CSS `stroke-dasharray`, otherwise, it is not clear what are the string values that are possible for `dash`.

src/js/components/Chart/stories/Dash.js
```diff
@@ -0,0 +1,29 @@
...
+import { grommet } from 'grommet/themes';
+
+const Example = () => (
+  <Grommet theme={grommet}>
+    <Box align="center" pad="large" gap="large">
+      <Chart type="line" dash values={[20, 30, 15]} />
+      <Chart type="line" dash round values={[20, 30, 15]} />
+      <Chart type="line" dash thickness="xsmall" values={[20, 30, 15]} />
+      <Chart type="line" dash round thickness="xsmall" values={[20, 30, 15]} />
+      <Chart type="line" dash="8 2" values={[20, 30, 15]} />
```
ShimiSun: If you are not familiar with the CSS styling for dasharray, It is not clear what the `dash` values of "8 2" mean, we need to elaborate on that on doc.js.

src/js/components/Chart/doc.js
ericsoderberghp: Simplified to remove the string capability for now.

src/js/components/Chart/stories/Dash.js
ericsoderberghp: Simplified to remove the string capability for now.

src/js/components/Chart/doc.js
```diff
@@ -58,6 +58,9 @@ export const doc = Chart => {
     id: PropTypes.string.description(`A unique identifier for the Chart. This
       is required if more than one Chart is shown and they use color
       gradients.`),
+    dash: PropTypes.bool
+      .description(`Whether to use dashed lines for line or bar charts.`)
+      .defaultValue(false),
```
ShimiSun: would it be more accurate to state `undefined`? or actually set the default value of `dash` to false.

ericsoderberghp: I was following the pattern of other props, like `round`.

PR #3863, opened 2020-03-17
Fix type error in theme breakpoints field

sradevski:
Since the literal breakpoint fields are optional
(and can be undefined), the string index type has to be
optional as well.

This broke as part of https://github.com/grommet/grommet/pull/3793/files

Signed-off-by: Stevche Radevski <stevche@balena.io>

#### What does this PR do?
Fixes a typing error

N/A

N/A

N/A

N/A

N/A

N/A

No

Yes

Yes

### DIFF
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -123,7 +123,7 @@ export interface ThemeType {
         borderSize?: BreakpointBorderSize;
         edgeSize?: BreakpointEdgeSize;
         size?: BreakpointSize;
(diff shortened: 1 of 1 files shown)


sradevski:
@ShimiSun It would be great if we can cut a release once this is merged as it prevents us from updating to the latest version, and the Select component was fixed in 2.11.2

ShimiSun:
@sradevski, yes I agree, I'll try to target a release soon.

Issue #3851, opened 2020-03-16
Select component does not set value to selected option


rohit8:
### Expected Behavior
An option selected should set the Select value on first click.

### Actual Behavior
If you scroll down the list of options and select an option, the Select component doesn't set the value, and scrolls up the list instead. However, selecting the same value the second time works.

### URL, screen shot, or Codepen exhibiting the issue

https://storybook.grommet.io/?path=/story/form--select

### Steps to Reproduce
1. Click the dropdown
2. Scroll down the list of options.
3. Select an option

### Your Environment
- Grommet version: 2.11.2
- Browser Name and version: Google Chrome Version 80.0.3987.132 (Official Build) (64-bit)
- Operating System and version (desktop or mobile): MacOS Mojave 10.14.6 (Desktop)

![captured](https://user-images.githubusercontent.com/2968479/76722539-689aaf80-671a-11ea-9b95-058ecaa0574b.gif)

ShimiSun:
Thanks for reporting buddy!

bcliffor:
Thanks for working on Grommet all. Just wanted to bump this issue since it's been causing a lot of annoyance for my team too.

ShimiSun:
@bcliffor thanks for the bump, few released a new version last week that seems to have fixed the issue as you can see https://codesandbox.io/s/form-select-many-options-h1r9e?file=/index.js
On the codesandbox, if you'll change the version of grommet to 2.11.2 you'll see the issue reproduces, but on the recent release of 2.14.0 the issue is already fixed. Please confirm.

bcliffor:
Definitely fixed!

Thanks for the very quick response @ShimiSun !

ShimiSun:
Woohoo, I'll be closing the issue, thank you all for the collaboration.

Issue #3854, opened 2020-03-16
MaskedInput scroll bars goes back to the top


mrnguuyen:
### Expected Behavior
Scroll bar should stay within place inside the MaskedInput and show allow user to click an input at the bottom of the scroll.

### Actual Behavior
Scroll bar goes back to the top and does not allow user to click on the input at the bottom of the scroll.

### URL, screen shot, or Codepen exhibiting the issue
![image](https://user-images.githubusercontent.com/12700313/76793413-3de44180-6782-11ea-9713-f6f8dfb3d0c2.png)

Storybook: https://storybook.grommet.io/?path=/story/maskedinput--date

### Steps to Reproduce
Make the browser small enough to make the MaskedInput scrollable as shown in the screenshot above and try to click on the month 12.

### Your Environment
Web browser from storybook and grommet website

- Grommet version: latest
- Browser Name and version: chrome and latest
- Operating System and version (desktop or mobile): windows 10

mrnguuyen:
Some issues that I am experiencing with this:
When I try to select the month (12) when the browser is minimized to the screenshot above with a scroll bar, the scroll bar goes back to the top the moment I try to click on 12.

mrnguuyen:
![image](https://user-images.githubusercontent.com/12700313/76882511-0041f000-6838-11ea-869b-27a53a2eba6a.png)

IanKBovard:
Looks like this pr caused the issue
https://github.com/grommet/grommet/pull/3776

hooks are causing renrenders on every mouseover.

mrnguuyen:
@ShimiSun, how's the progress on this issue?

ShimiSun:
The fix is available on the [stable branch ](https://github.com/grommet/grommet/wiki/What-is-grommet-stable-and-how-to-use-it%3F) and it will be officially published on our next release. Thanks for reporting!

PR #3867, opened 2020-03-17
Select component does not set value to selected option Fix

IanKBovard:
#### What does this PR do?
Fixes issue #3851 by adding a condition to FormFields onBlur event. onBlur will not setState if the current node clicked has a role of menubar.

FormField.js
FormFieldBox onBlur event.

Tested in storybook, using my own custom stories.

Storybook. Should wrap every necessary component in FormField to see if focus behaves as it should.

When the Select component was wrapped in a FormField component it would take two clicks to run onChange. This was because FormField has setState during an onBlur event which caused a rerender. 

#3851 

No

No

If any other component has a role of menubar, it would affect them.


IanKBovard:
Not sure, but it might affect this issue as well.
https://github.com/grommet/grommet/issues/3854

IanKBovard:
To test i mostly used this section of storybook. Tabbed through and click on everything to see if the components gained and lost focus.
https://storybook.grommet.io/?path=/story/form--all

IanKBovard:
More explanation:

The issue doesn't lie with the value not setting as described in issue #3851. The problem is that when setState is called within an onBlur event the component re-renders. And, more often than not onBlur triggers before onClick so onClick never fires.

https://www.eventbrite.com/engineering/a-story-of-a-react-re-rendering-bug/

What, I believe, is happening is that when you click on an option in a select dropdown onBlur is being called triggering the setFocus(true). This causes a re-render of the dropdown, which is why the scrollbar resets to the top. When you click a dropdown option again onClick fires, because the dropdown already has focus so onBlur doesn't fire again.

The solution here adds a conditional within onBlur to ignore dropdown items.

src/js/components/FormField/FormField.js
```diff
@@ -327,7 +327,10 @@ const FormField = forwardRef(
           if (onFocus) onFocus(event);
         }}
         onBlur={event => {
-          setFocus(false);
+          const role = event.target.getAttribute('role');
```
ericsoderberghp: This seems like a fragile approach. I think we should look for an alternate approach to addressing this.

ericsoderberghp:
I'm wondering why the FormField is getting focus when clicking an option, especially why it is tied to a long options list.

src/js/components/FormField/FormField.js
IanKBovard: gotcha, can do.

IanKBovard:
The problem isn't FormField gaining focus, it's that FormField loses focus causing the onBlur to trigger. When you setState in onBlur it causes a rerender.

Here's what happens.
 - I click the dropdown in formfield
 - my list comes up
- i scroll through the list
 - i select on option
- formfield loses focus and the select options gain it
- formfield's onblur fires with setstate causing a rerender
- the scrollbar resets to the top of the selection
- i scroll back down to my option and select it again
- onclick fires

IanKBovard:
updated logic. checking for tag name instead. 
```
if (event.target.tagName === 'INPUT')  setFocus(false);
```

IanKBovard:
will come back to this issue when my work on discover has completed

Issue #2820, opened 2019-02-23
<Button /> icon does not respect theme color


atanasster:
### Expected Behavior
when theme.button.color is changed, the icon color should match the label color. 

### Actual Behavior
Button icon currently uses text color, regardless of the theme. The better behavior would be to work like Anchor https://github.com/grommet/grommet/blob/fd803e9bd3e1c871cd02e14e322cf03b3eb2dd41/src/js/components/Anchor/Anchor.js#L46

### URL, screen shot, or Codepen exhibiting the issue
example with Button and Anchor 
https://codesandbox.io/s/y07v9k2mkv

britt6612:
@oorestisime @ShimiSun  currently the button icon depends on the background color for it to be light or dark I can change this so that the icon will take the color of the theme instead of looking for the background color to determine the icon color if this sounds good to you guys I can do this!

oorestisime:
sounds good to me. i wonder whether we should fallback to text though if theme value is not defined.

atanasster:
imho, changing the icon color only for primary is also an issue, and the code should look something like
```
if (icon && !icon.props.color) {
      buttonIcon = cloneElement(icon, {
        color:normalizeColor(color || theme.button.color  || 'text', theme),
      });
    }
```

atanasster:
sorry, we need to account for primary color in theme:
```
if (icon && !icon.props.color) {
  const themeColor = primary ? theme.button.primary.color : theme.button.color;
  buttonIcon = cloneElement(icon, {
        color:normalizeColor(color || themeColor || 'text', theme),
      });
}
```

britt6612:
@atanasster  thankyou I just saw that there was both a primary color in the themes under the button I will account for that as well

ShimiSun:
Thank you for filing the issue! @britt6612 thank you for dedicating time to solve it!

ShimiSun:
Applying a direct fix for this issue requires code changes that might cause backward compatibility issues. Hence, the fix was implemented as part of the new 'kind' Button structure. once the user has `theme.button.default` (along with other kind of buttons `theme.button.secondary` or `theme.button.primary`) defined on the theme, for those types of buttons the color of the icon will respect the theme color. For more context, use https://v2.grommet.io/button

PR #3868, opened 2020-03-17
Add simple prop to Button

taysea:
#### What does this PR do?
Adds `simple` prop to Button which has default styling of no background fill but a fill on hover as discussed here: https://github.com/hpe-design/design-system/issues/466

src/js/components/Button/StyledButton.js

Stories have been made and tests have been added.

Using storybook, apply `simple` to a Button.

Background explained in the above mentioned thread.

Yes, they are updated here.

Yes.


src/js/components/Button/Button.js
```diff
@@ -97,7 +110,12 @@ const Button = forwardRef(
     let contents;
     if (first && second) {
       contents = (
-        <Box direction="row" align="center" justify="center" gap={gap}>
+        <Box
+          direction="row"
+          align="center"
+          justify={normalized ? 'start' : 'center'}
```
ShimiSun: I'm not sure why `justify` should be impacted by the `normalized` prop. what am I missing?

src/js/components/Button/README.md
```diff
@@ -260,6 +260,15 @@ Label text to place in the button.
 node
 ```
 
+**normalized**
+
+Whether this is a normalized button. Normalized buttons have consistent 
+      styling regardless of if they contain only an icon or an icon and label.
```
ShimiSun: I would add a little more description of what will be a good use case to use this prop. this description is not telling much.

src/js/components/Button/StyledButton.js
```diff
@@ -122,6 +132,20 @@ const plainStyle = props => css`
...
+const normalizedStyle = props => css`
+  ${backgroundStyle(
+    normalizeColor(
+      props.colorValue || props.theme.button.normalized.color || undefined,
+      props.theme,
+    ),
+    props.theme,
+    props.theme.button.color,
+  )}
+  border: none;
```
ShimiSun: would this affect the styling of normalized + focus?

src/js/components/Button/__tests__/Button-test.js
```diff
@@ -205,6 +205,30 @@ describe('Button', () => {
...
+        <Button onClick={() => {}} normalized>
+          normalized
+        </Button>
+      </Grommet>,
+    );
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  test('normalized primary', () => {
```
ShimiSun: should we also add test for icon w/o a label?

src/js/components/Button/stories/Normalized.js
```diff
@@ -0,0 +1,69 @@
...
+        border: {
+          color: '#99ccff',
+        },
+      },
+    },
+  },
+});
+
+const Normalized = () => (
+  <>
```
ShimiSun: I don't think you need a fragment.

src/js/themes/base.js
```diff
@@ -364,6 +364,14 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
       },
       // color: { dark: undefined, light: undefined }
+      normalized: {
+        // color: { dark: undefined, light: undefined }
+        hover: {
+          border: {
+            // color: {dark: undefined, light: undefined},
+          },
+        },
+      },
```
ShimiSun: The new props need to be added to base.d.ts.

src/js/components/Button/doc.js
```diff
@@ -78,6 +78,10 @@ with plain Buttons.`,
     ),
     icon: PropTypes.element.description('Icon element to place in the button.'),
     label: PropTypes.node.description('Label text to place in the button.'),
+    normalized: PropTypes.bool.description(
```
ShimiSun: Let's add the normalized type to index.d.ts as well.

src/js/components/Button/StyledButton.js
taysea: The focus should still appear around the button as normal.

src/js/components/Button/__tests__/Button-test.js
taysea: Yes, we should. Good call.

src/js/components/Button/stories/Normalized.js
taysea: I used two `Grommet` components to be able to demonstrate with different themes, so I need the fragment to wrap around the two of them.

src/js/components/Button/Button.js
taysea: Otherwise, the icon + label would be centered -- good call that maybe it shouldn't be forced. What do you think about giving users control of the justify? 

I think in some designs we have seen (such as sidebar), you would want the icon and label justified to the start as opposed to center.

src/js/components/Button/Button.js
```diff
@@ -52,18 +54,20 @@ const Button = forwardRef(
     }
 
     const isDarkBackground = () => {
+      const buttonColor =
+        normalized && !primary
+          ? theme.button.normalized.color
+          : theme.button.primary.color;
```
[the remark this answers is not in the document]
taysea: It gets called below in line 63

src/js/components/Button/Button.js
```diff
@@ -17,6 +17,7 @@ import { StyledButton } from './StyledButton';
 const Button = forwardRef(
   (
     {
+      active,
```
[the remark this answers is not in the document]
taysea: I don't that the icon color was not being affected by active styling, only label color. I agree it's not directly related to this PR, should I pull it out into its own PR?

src/js/components/Button/Button.js
ShimiSun: Interesting. can you share the designs? it will help with the context.

taysea: Icon only:
Screen Shot 2020-03-18 at 9 35 13 AM

Icon + Label with justify="start":
Screen Shot 2020-03-18 at 9 35 24 AM

src/js/components/Button/Button.js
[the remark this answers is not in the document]
ShimiSun: I think you are using it later on... so I'm not sure yet.

src/js/components/Button/stories/Normalized.js
ShimiSun: Gotcha, I'll dive deeper into the implementation later on today.

src/js/components/Button/Button.js
```diff
@@ -84,11 +88,20 @@ const Button = forwardRef(
 
     let buttonIcon = icon;
     // only change color if user did not specify the color themselves...
-    if (primary && icon && !icon.props.color) {
+    if (
```
ericsoderberghp: Again, maybe `props.color || theme.button[flavor].color` if !icon.props.color?

src/js/components/Button/Button.js
ericsoderberghp: What about adding `align` to Button, taking values of `start`, `center`, `end`? I don't think the caller should care that Button has a Box in it, and I think `align` will seem a more natural term than `justify`.

src/js/components/Button/README.md
```diff
@@ -260,6 +260,17 @@ Label text to place in the button.
 node
 ```
 
+**normalized**
+
+Whether this is a normalized button. By default, icon only buttons have
```
ericsoderberghp: Nicely described!

src/js/themes/base.js
```diff
@@ -364,6 +364,14 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         radius: `${baseSpacing * 0.75}px`,
       },
       // color: { dark: undefined, light: undefined }
+      normalized: {
```
ericsoderberghp: In order to allow more theme flexibility, I think we should allow the theme to control background and border for all flavors of icon. So: `theme.button[flavor].background`, `theme.button[flavor].color`, `theme.button[flavor].border`. And then per-flavor: `theme.button[flavor].active.background`, `theme.button[flavor].hover.background`, etc. and flavor-agnostic state variations: `theme.button.disabled.background = "opacity: 0.7"`.

src/js/components/Button/Button.js
taysea: I agree `align` feels more natural, but it feels like it contradicts the language/behavior established by `align` vs `justify` in that in row/horizontal things, `justify` controls the horizontal positioning.

src/js/themes/base.js
taysea: Great, thanks. Let me dive into this more.

src/js/components/Button/Button.js
ericsoderberghp: I think we are more inclined to think that way because we know we have implemented this with a Box. I'm less sure that callers will think that justify fits better. But, I don't have a strong opinion on this. And, I don't think we have a strong precedent elsewhere to align with. Text uses `textAlign`, which does use the `align` term but with the `text` prefix, is different enough. So, I'm OK with `justify`. I think we should add that as a property for Button. I'm not sure that `normalized` implies `justify="start"` though. I'm wondering if `justify="center"` should be the default. Thoughts?

taysea:
> I think we're close on this one. Just one more thing to iron out, I think?

Agreed that `normalized` shouldn't be the prop that drives it. I see what you're saying about `justify` and `textAlign` etc.. I would be flexible for the prop to be something other than `justify` if there are strong opinions in another direction. For now since you are OK with it, I will update Button to take justify, and agreed that `center` should be default.

src/js/components/Button/Button.js
taysea: Agreed that normalized shouldn't be the prop that drives it. I see what you're saying about justify and textAlign etc.. I would be flexible for the prop to be something other than justify if there are strong opinions in another direction. For now since you are OK with it, I will update Button to take justify, and agreed that center should be default because that is how Button currently works.

ericsoderberghp:
The only remaining question for me is whether `normalized` is the right word, similar to Greg's reaction in Slack. Otherwise, I'm OK with this PR.

I've been thinking recently of Button having styling strength levels, where default is the middle ground, `primary` is the strongest, and this PR is adding a weaker variant. It's a bit like our T-shirt sizing philosophy but applied to the level of attention the Button wants to call to itself. If we didn't already have primary, this concept might best be illustrated via a `strength` property, where values would be `strong`, `medium`, and `weak`, like `opacity` elsewhere. So, instead of this being about aligning padding between icon-only and label-only buttons, it's more about a vehicle to enable consistent and themed styling of weaker Buttons in sidebars, headers, footers, nav lists, etc.

ShimiSun:
>So, instead of this being about aligning padding between icon-only and label-only buttons, it's more about a vehicle to enable consistent and themed styling of weaker Buttons in sidebars, headers, footers, nav lists, etc

I agree with this approach, and I think this direction has more real use cases we should address.

I was actually playing with the thought of allowing `primary=false` to do the logic :face_with_hand_over_mouth: but not sure how backward compatible it will be.

taysea:
Summarizing from an offline discussion, because it feels like `normalized` is aiming to do a few tasks that are orthogonal from each other (have simple styling but also manage the padding), we are going to break this PR apart.

We will move forward with the prop name of `simple`, but this will handle purely the styling of border/background/hover, and leave out padding. By default, `simple` has no background color and only hover behavior, but these values can be controlled by a custom theme.

src/js/components/Button/doc.js
```diff
@@ -77,7 +77,24 @@ with plain Buttons.`,
...
+      'center',
+      'end',
+      'evenly',
+      'start',
+      'stretch',
+    ]).description(
+      'How to align the contents along the horizontal axis in the button.',
+    ),
     label: PropTypes.node.description('Label text to place in the button.'),
+    normalized: PropTypes.bool.description(
```
ericsoderberghp: should be `simple`?

src/js/components/Button/index.d.ts
```diff
@@ -27,7 +27,9 @@ export interface ButtonProps {
   href?: string;
   target?: "_self" | "_blank" | "_parent" | "_top";
   icon?: JSX.Element;
+  justify?: "around" | "between" | "center" | "end" | "evenly" | "start" | "stretch";
   label?: React.ReactNode;
+  normalized?: boolean;
```
ericsoderberghp: `simple`

taysea: Thank you! Committed these changes yesterday, but forgot to push!

ShimiSun:
Reviewing the PR concept along with other Button ideas with eric today, and I'm sharing a couple of questions that came up regarding this PR. 
The description of `simple` isn't elaborating enough with the styles that are being applied (border/background/hover), and the fact the styles are fetched from the theme.
Also, simple + primary should have a corresponding/similar style setups on the theme.

One last thing is the `justify` prop, why was it exposed as a prop and not set on the theme? and what is the value of `justify` without `align` prop.

src/js/themes/base.js
```diff
@@ -378,8 +378,29 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         radius: `${baseSpacing * 0.75}px`,
       },
       // color: { dark: undefined, light: undefined }
+      simple: {
```
ShimiSun, line 381: `simple`/`primary` & default buttons should include `disabled`, `active`, `hover` and `extend`.
Anything that is styled on the default button should have a style option on the simple/primary objects.
padding should also be an option for each type of button.

src/js/themes/base.js
```diff
@@ -378,8 +378,29 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
+        border: {
+          // color: { dark: undefined, light: undefined }
+        },
+        // color: { dark: undefined, light: undefined }
+        hover: {
+          border: {
+            // color: {dark: undefined, light: undefined}
+          },
+          color: 'active',
```
ShimiSun, line 390: Is this a color or background color? and let's think if it should be `active-background`?

src/js/components/Button/doc.js
```diff
@@ -102,6 +113,10 @@ with plain Buttons.`,
               end of the anchor.`,
       )
       .defaultValue(false),
+    simple: PropTypes.bool.description(
```
ShimiSun, line 116: we're thinking of having `emphasis` prop with either strong or weak values. strong will be translated to `primary` while `weak` will be translated to `simple`. 
we can replace the existing `primary` examples to show the usage of the new prop while still supporting `primary` for backward compatibility.

ShimiSun:
Is it possible to decouple the fix for aligning the icon color with the text color on a separate PR #2820? or is it too embedded with the newly introduced functionality?

src/js/themes/base.d.ts
```diff
@@ -301,17 +301,71 @@ export interface ThemeType {
...
+      active?: {
+        border?: {
+          color: ColorType;
+        },
+        color?: ColorType;
+        extend?: ExtendType;
+      };
+      disabled?: {
+        border?: {
+          color: ColorType;
```
ShimiSun, line 324:
```suggestion
          color?: ColorType;
```

src/js/themes/base.d.ts
```diff
@@ -301,17 +301,71 @@ export interface ThemeType {
...
-      active?: {
+      hover?: {
+        color?: ColorType;
         border?: {
           color?: ColorType;
         };
+      };
+      active?: {
+        border?: {
+          color: ColorType;
```
ShimiSun, line 353:
```suggestion
          color?: ColorType;
```

src/js/themes/base.d.ts
```diff
@@ -301,17 +301,71 @@ export interface ThemeType {
...
+        },
+        color?: ColorType;
+        extend?: ExtendType;
+      };
+      disabled?: {
+        border?: {
+          color: ColorType;
```
ShimiSun, line 360:
```suggestion
          color?: ColorType;
```

src/js/themes/__tests__/__snapshots__/theme-test.js.snap
```diff
@@ -1579,7 +1579,7 @@ exports[`Grommet hpe theme 1`] = `
   box-sizing: border-box;
   max-width: 100%;
   background-color: #01A982;
-  color: #EEEEEE;
+  color: #FFFFFF;
```
ShimiSun, line 1582: Can you verify what is this color change impacting? I see too many similar changes across the snapshot and we need to know the root cause for this change.

src/js/components/Select/__tests__/__snapshots__/Select-test.js.snap
```diff
@@ -508,7 +508,7 @@ exports[`Select applies custom global.hover theme to options 2`] = `
   padding: 0;
   text-align: inherit;
   background-color: rgba(221,221,221,0.4);
-  color: #000000;
+  color: #444444;
```
ShimiSun, line 511: Same here regarding color change.

src/js/components/Distribution/README.md
```diff
@@ -154,10 +154,10 @@ string
 **children**
 
 Function that will be called when each value is rendered. Defaults to `function children(value) {
-    return _react["default"].createElement(_Box.Box, {
+    return /*#__PURE__*/_react["default"].createElement(_Box.Box, {
```
ShimiSun, line 157: which npm version are you using? where does this change is coming from?

src/js/components/Button/stories/Secondary.js
```diff
@@ -0,0 +1,255 @@
...
+    },
+    primary: {
+      background: {
+        color: 'green',
+      },
+    },
+  },
+});
+
+const Secondary = () => (
```
ShimiSun, line 67: This example is a little overwhelming. 
Can we make it leaner and remove the usage of the hpe theme? I think the hpe theme will be confusing for both hpe and grommet users, we can just show how to use the new props merged with grommet theme. 

I'm also not inclining to show multiple usages of Grommet component in the same file, it might cause confusion on how the component should be used. In case we do want to demonstrate a comparison between different themes ThemeContext might be an option.

src/js/components/Button/__tests__/__snapshots__/Button-test.js.snap
```diff
@@ -22,7 +22,7 @@ exports[`Button active + primary 1`] = `
   color: #f8f8f8;
   border-radius: 18px;
   background-color: rgba(221,221,221,0.4);
-  color: #000000;
+  color: #444444;
```
ShimiSun, line 25: Same regarding the color impact.

src/js/components/Button/__tests__/Button-test.js
```diff
@@ -235,6 +235,55 @@ describe('Button', () => {
     expect(tree).toMatchSnapshot();
   });
 
+  test('normalized label only', () => {
+    const component = renderer.create(
+      <Grommet>
+        <Button onClick={() => {}} normalized>
```
ShimiSun, line 241: I don't think we are using `normalized`.

src/js/components/Button/__tests__/Button-test.js
```diff
@@ -235,6 +235,55 @@ describe('Button', () => {
...
+
+  test('normalized icon only', () => {
+    const component = renderer.create(
+      <Grommet>
+        <Button icon={<svg />} onClick={() => {}} normalized />
+      </Grommet>,
+    );
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
+  });
```
ShimiSun, line 258: I don't think we are using normalized.

src/js/components/Button/__tests__/Button-test.js
```diff
@@ -235,6 +235,55 @@ describe('Button', () => {
...
+          icon={<svg />}
+          label="normalized"
+          onClick={() => {}}
+          normalized
+        />
+      </Grommet>,
+    );
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
+  });
```
ShimiSun, line 273: I don't think we are using normalized.

src/js/components/Button/__tests__/Button-test.js
```diff
@@ -235,6 +235,55 @@ describe('Button', () => {
...
+
+  test('normalized primary', () => {
+    const component = renderer.create(
+      <Grommet>
+        <Button onClick={() => {}} normalized primary>
```
ShimiSun, line 278: I don't think we are using normalized.

README.md
```diff
@@ -67,6 +67,7 @@ From your package.json point to stable.
 ```
 "grommet": "https://github.com/grommet/grommet/tarball/stable",
 ```
+
```
ShimiSun, line 70: I don't think README changes are relevant on this PR.

taysea:
Closing because this behavior is handled by https://github.com/grommet/grommet/pull/4050/files

PR #3869, opened 2020-03-18
Bug in Form Validate Fix

### LINKED ISSUE #3803: Bug in Form Validate
In a form, when we have a TextInput formField & a checkbox formField & the form's validate prop is onBlur.
Now, when the formField comes to an error state & when we focus on checkbox, the earlier error state  disappears.

Please check the following codesandbox link,
https://codesandbox.io/s/bold-cray-8n2mf

Please do check the gif where I reproduced the same,
issues1

IanKBovard:
#### What does this PR do?
Fixes #3803 
Added conditionals to form validation so only the specific field is affected by onBlur and onChange effects.

Form.js update function in the form component.

Added a checkbox to validateOnBlur storybook example.

Using the same example should work. I would check if validation behaves as expected.

#3803 

Kapture 2020-03-18 at 12 07 05

No

No

Backwards compatible


IanKBovard:
ill see what i can do. in this pr or a new one?

ShimiSun:
@IanKBovard This PR. thanks.

IanKBovard:
tests added

src/js/components/Form/Form.js
```diff
@@ -85,15 +85,21 @@ const Form = forwardRef(
...
         // re-run any validations, in case the validation
         // is checking across fields
         setErrors(prevErrors => {
           const nextErrors = { ...prevErrors };
           Object.keys(prevErrors).forEach(errName => {
             if (validations.current[errName]) {
-              const nextError = validations.current[errName](data, nextValue);
-              updateErrors(nextErrors, errName, nextError);
+              const nextError = validations.current[errName](nextValue, data);
+              if (!nextError && nextErrors[name]) {
```
ericsoderberghp: Isn't `nextErrors[name]` always going to be true, since we built `nextErrors` from `prevErrors` at line 91 and are only going through the keys in `prevErrors` at line 92?

IanKBovard: ill double check

src/js/components/Form/Form.js
```diff
@@ -85,15 +85,21 @@ const Form = forwardRef(
...
-
         // re-run any validations, in case the validation
         // is checking across fields
         setErrors(prevErrors => {
           const nextErrors = { ...prevErrors };
           Object.keys(prevErrors).forEach(errName => {
             if (validations.current[errName]) {
-              const nextError = validations.current[errName](data, nextValue);
-              updateErrors(nextErrors, errName, nextError);
+              const nextError = validations.current[errName](nextValue, data);
```
ericsoderberghp: This seems backwards, the two arguments to the validation should be `(data, nextValue)`.

src/js/components/Form/Form.js
IanKBovard: No not always. It can come out as undefined.

src/js/components/Form/Form.js
IanKBovard: console.log of validations.current 
```
name: ƒ validateField(value2, data)
email: ƒ validateField(value2, data)
```

i could be misunderstanding something.

ericsoderberghp: Take a look at all of the other calls, such as `validations.current[name](value[name], value)`. At line 94, `nextValue` is the same kind of thing as `value` and `value[name]` is the same thing a `data` in this context.

IanKBovard: gotcha, thanks for clearing that up.

IanKBovard: changes made

src/js/components/Form/Form.js
```diff
@@ -85,15 +85,21 @@ const Form = forwardRef(
...
         // re-run any validations, in case the validation
         // is checking across fields
         setErrors(prevErrors => {
           const nextErrors = { ...prevErrors };
           Object.keys(prevErrors).forEach(errName => {
             if (validations.current[errName]) {
               const nextError = validations.current[errName](data, nextValue);
-              updateErrors(nextErrors, errName, nextError);
+              if (!nextError && nextErrors[name]) {
+                updateErrors(nextErrors, name, nextErrors[name]);
```
ericsoderberghp: I think `nextErrors[name]` is already set before we make this call. Do we need this?

src/js/components/Form/Form.js
```diff
@@ -85,15 +85,21 @@ const Form = forwardRef(
...
+              } else if (nextError) {
+                updateErrors(nextErrors, errName, nextError);
+              }
+            }
+            if (nextValue[name]) {
+              updateErrors(nextErrors, name, undefined);
```
ericsoderberghp: Why do we clear any errors just because there is a value? Won't this potentially clear a validation error set at line 98?

IanKBovard:
working on a new iteration of this. will update you when i have it up

IanKBovard:
Updated logic, added a check so the validate wont for each field on input.

src/js/components/Form/__tests__/Form-test.js
```diff
@@ -481,6 +481,85 @@ describe('Form', () => {
       }),
     );
   });
+  test('validate on blur', () => {
+    const Test = () => (
+      <Form
+        validate="blur"
+        onReset={event => console.log(event)}
```
ShimiSun, line 488: we can remove this event, since it's not being used.

src/js/components/Form/__tests__/Form-test.js
```diff
@@ -481,6 +481,85 @@ describe('Form', () => {
...
+        onSubmit={({ value }) => console.log('Submit', value)}
```
ShimiSun, line 489: remove console log

src/js/components/Form/__tests__/Form-test.js
```diff
@@ -481,6 +481,85 @@ describe('Form', () => {
...
+    });
+    fireEvent.click(getByText('submit'));
+    expect(queryAllByText('required')).toHaveLength(1);
+
+    // fields reset to required error message
+    fireEvent.change(getByPlaceholderText('name'), {
+      target: { value: '' },
+    });
+    fireEvent.click(getByText('submit'));
+    expect(queryAllByText('required')).toHaveLength(2);
```
ShimiSun, line 546: Check if this is needed since it is referring to reset.

IanKBovard: changes made

src/js/components/Form/Form.js
```diff
@@ -90,7 +91,7 @@ const Form = forwardRef(
           setErrors(prevErrors => {
             const nextErrors = { ...prevErrors };
             Object.keys(prevErrors).forEach(errName => {
-              if (validations.current[errName]) {
+              if (validations.current[errName] && errName === name) {
```
ericsoderberghp, line 94: I'm OK with this, since it works.

But, it might be cleaner to:
```
if (prevErrors[name] && validations.current[name]) {
  const nextErrors = { ...prevErrors };
  ...
}
```
rather than looping though all `prevErrors`.

IanKBovard: changes made

src/js/components/Form/Form.js
```diff
@@ -88,14 +89,22 @@ const Form = forwardRef(
...
-            Object.keys(prevErrors).forEach(errName => {
-              if (validations.current[errName]) {
-                const nextError = validations.current[errName](data, nextValue);
-                updateErrors(nextErrors, errName, nextError);
-              }
-            });
-            return nextErrors;
+            if (prevErrors[name] && validations.current[name]) {
+              console.log(prevErrors[name], validations.current[name]);
+              console.log('hello');
```
ShimiSun, line 94: should be removed.

IanKBovard: removed

src/js/components/Form/Form.js
```diff
@@ -88,14 +89,20 @@ const Form = forwardRef(
...
-            Object.keys(prevErrors).forEach(errName => {
-              if (validations.current[errName]) {
-                const nextError = validations.current[errName](data, nextValue);
-                updateErrors(nextErrors, errName, nextError);
-              }
-            });
-            return nextErrors;
+            if (prevErrors[name] && validations.current[name]) {
+              const nextErrors = { ...prevErrors };
+              Object.keys(prevErrors).forEach(errName => {
```
ericsoderberghp, line 94: I don't think we need to iterate over the errors, since we already know the name in question:
```
if (prev && current) {
  const next = { ...prev };
  updateErrors(next, name, validations.current[name](data, nextValue));
  return next;
}
```

IanKBovard: gotcha, made changes and tested.

Issue #3864, opened 2020-03-17
Select with valueKey not working as expected


mrcleanandfresh:
When I give `Select` an array of objects then use the `labelKey` and `valueKey` to try and display a really basic use-case across our app, it doesn't work.

### Expected Behavior

I would expect that if I defined the `labelKey` and `valueKey` that when I select an item from the dropdown, that when I receive an `onChanges` event from the `Form`'s context, that it would contain the properties value for the `valueKey` specified. For example, if `valueKey` was set to `id` and that id was `10` I would expect that to be the value. `{myField: 10}`

### Actual Behavior

It gives me the the label as the value `{myField: "Item Label"}`

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-select-issue-ol1p3

### Steps to Reproduce

1. Select an item from the dropdown
2. The JSON Data on the page will match the label of the selected item, and not be a numerical value.

### Your Environment

- Grommet version: 2.11.2
- Browser Name and version: Brave Version 1.4.96 Chromium: 80.0.3987.132 (Official Build) (64-bit)
- Operating System and version (desktop or mobile): macOS Catalina 10.15.3

ericsoderberghp:
Part of the root cause of this is due to `Select` using `TextInput` internally and both trying to drive the form context.

mrcleanandfresh:
I did notice the Select rendered a text input. In my case the value attribute is empty. The other part of the problem seems to be that the value attribute is what is displayed to the user, instead of a traditional `option` element where it has a label and a value. It’s not trivial for sure.

ShimiSun:
Looks like this issue is fixed, feel free to reopen if needed.

PR #3870, opened 2020-03-19
Changed Select to fix an issue with value propagation.

ericsoderberghp:
#### What does this PR do?

Changed Select to fix an issue with value propagation.

Select.js, probably easier to read the new version without the diff at some point

unit, storybook

storybook

#3864 

no

yes

backwards compatible


src/js/components/Form/stories/Select.js
```diff
@@ -1,24 +1,32 @@
...
-        <FormField label="Label" htmlFor="select" {...props}>
-          <Select
-            id="select"
-            placeholder="placeholder"
-            options={allOptions}
-            value={value}
-            onChange={({ option }) => setValue(option)}
-          />
-        </FormField>
+        <Form value={value} onChange={onChange}>
```
ShimiSun: Let's support the old version of this story (just FormField and no Form) on a test for backward compatibility.

src/js/components/TextInput/TextInput.js
```diff
@@ -99,7 +100,12 @@ const TextInput = forwardRef(
     const suggestionsRef = useRef();
     const suggestionRefs = {};
 
-    const [value, setValue] = formContext.useFormContext(name, valueProp);
+    // if this is a readOnly property, don't set a name with the form context
+    // this allows Select to control the form context for the name.
+    const [value, setValue] = formContext.useFormContext(
+      readOnly ? undefined : name,
+      valueProp,
```
ShimiSun: Reading through those comments, I don't see how Select is using the readOnly prop to control formContext for the name, what am I missing?

ShimiSun:
I'm not sure what [this story example ](https://deploy-preview-3870--sad-tereshkova-173d07.netlify.com/?path=/story/select--multiple) is supposed to show, but I'm not sure that Multiple is the right naming for it, since you can't switch between options.

src/js/components/TextInput/TextInput.js
ericsoderberghp: That was existing already: https://github.com/grommet/grommet/blob/546e18c9fea9d4e888d6437cf39b1a3453f1eb0c/src/js/components/Select/Select.js#L246

src/js/components/Form/stories/Select.js
ericsoderberghp: Since the prior story didn't even have `<Form/>` in it, I put the old story content in `FormField/stories/Select.js` and kept this one as the `Form` version.

ericsoderberghp:
> I'm not sure what [this story example ](https://deploy-preview-3870--sad-tereshkova-173d07.netlify.com/?path=/story/select--multiple) is supposed to show, but the name Multiple is probably not the right naming for it since you can't switch between options.

Good catch! That was left over from a white-box test I was playing with. Fixed.

src/js/components/FormField/stories/Select.js
```diff
@@ -0,0 +1,27 @@
+import React, { useState } from 'react';
+import { storiesOf } from '@storybook/react';
+
+import { grommet, Box, FormField, Select, Grommet } from 'grommet';
+import { allOptions } from './data';
```
ShimiSun:
```suggestion

const allOptions = Array(100)
  .fill()
  .map((_, i) => `option ${i + 1}`);
```
Users have requested no to hide the data in the stories, or add a link to see the data.
In this case, the data is only taking 3 lines, so I think we can embedded it in the file.

ericsoderberghp: fixed

PR #3871, opened 2020-03-19
Avatar component

ShimiSun:
#### What does this PR do?
Added Avatar to the Grommet library.
I have more things in mind for the Avatar component, but I thought to get initial feedback before I proceed:
* Support ref and show example of a clickable avatar with Menu
* Done (with theming) - Set more specific t-shirt sizes on the theme, because the current scale of Box might be too big for avatars. 
* Done - Allow a new prop for textual Avatar, so the examples on the Basic story could get leaner.

doc.js
unit tests and storybook
 storybook - all the new Avatar stories
I found myself writing this component too many times, so I thought that as long as it has modularity it could be part of the grommet library.

#3489 
done
yep, bumps the minor
backwards compatible


src/js/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,37 @@
...
+
+import { Box } from '../Box';
+
+const Avatar = ({
+  align = 'center',
+  round = 'full',
+  size = 'xxsmall',
+  justify = 'center',
+  src,
+  height, // for warning check and discarding the value
```
oorestisime: not sure i see the reason for this, i mean if they aren't documented why would people use this?

src/js/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,37 @@
...
+}) => {
+  const avatarProps = {
+    align,
+    justify,
+    height: size,
+    round,
+    width: size,
+  };
+
+  if (height || width) {
```
oorestisime: if we do intend to leave this then lets only warn in development,

src/js/components/Avatar/Avatar.js
ShimiSun: They will be documented on the site as part of referring to all the Box props.

oorestisime: ah yeah ok. wondering whether we shouldn't care though, what difference does it make if users set non squared avatars or whatever?

ShimiSun: I don't want the user to have unexpected side-effects they can't understand, so I rather call it out.

src/js/components/Avatar/Avatar.js
oorestisime: actually no need to make this check.browser won't have any env to know when to show or not.

src/js/components/Avatar/Avatar.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+
+import { Box } from '../Box';
+import { Text } from '../Text';
+import { validEmail, getMd5 } from '../../utils';
+
+const Avatar = ({
+  align = 'center',
+  children,
+  email,
```
oorestisime: i am a little sceptical for this. adding another dependency (bundle size) for something that not everybody will use (actually not sure who would use this). plus this is something the user could do on their own by providing a background how they want it to be whether its gravatar or any other similar service.

ericsoderberghp: Good comment. I was hoping generating an MD5 hash wouldn't be too heavy. If it is, I agree with removing this optimization.

oorestisime: well bundlesize wise its not that heavy, 3.3kb minified (well sounds heavy for just an md5 actually) but feature wise as well its a maybe something not  alot of people would use to justify this 3.3kb.
just my opinion not sure it makes sense :)

ShimiSun: Good comments. Let me see if I can find a way to md5-encrypt locally on grommet rather than using a packaged. otherwise, I'll remove the email logic.

ericsoderberghp: Is this one lighter? https://preview.npmjs.com/package/md5

oorestisime:
Kind of!
https://bundlephobia.com/result?p=md5@2.2.1

src/js/components/Avatar/Avatar.js
ShimiSun: [md5]( https://preview.npmjs.com/package/md5) is 30% lighter and stands on 2.2kB.
[pure-md5](https://www.npmjs.com/package/pure-md5) is even lighter than that, and stands on 1.9kB.    

Adding a new package, no matter its size, is still a commitment from our end, and its purpose      
of a single use-case that wasn't requested yet isn't too appealing to me.

I'm wondering if we should go the direction of adding a package or let the users that use Avatar + email do it on their end and maybe wait to see if anyone requests this feature in the future.

ericsoderberghp: Agreed with waiting to see where the demand is before adding md5.

ShimiSun: Thanks, removed `email` prop for now.

PR #3878, opened 2020-03-23
Changed Box to ensure background.dark forces color choice

ericsoderberghp:
#### What does this PR do?

Changed Box to ensure background.dark forces color choice.
When a background color name is used that has both dark and light variants, passing the `dark` sub-property now controls which variant is used. `<Box background={{ color: 'background', dark: true }} />`

backgroundStyle() and normalizeColor()

Added two new examples in the Box -> Background story

Storybook Box -> Background

no

yes

backwards compatible

### DIFF
--- a/src/js/utils/colors.js
+++ b/src/js/utils/colors.js
@@ -1,4 +1,6 @@
-export const normalizeColor = (color, theme, forceDark) => {
+// Returns the specific color that should be used according to the theme.
+// If 'dark' is supplied, it takes precedence over 'theme.dark'.
+export const normalizeColor = (color, theme, dark) => {
   const colorSpec =
     theme.global && theme.global.colors[color] !== undefined
       ? theme.global.colors[color]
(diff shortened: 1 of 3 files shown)


src/js/utils/colors.js
```diff
@@ -1,14 +1,20 @@
...
   const colorSpec =
     theme.global && theme.global.colors[color] !== undefined
       ? theme.global.colors[color]
       : color;
   // If the color has a light or dark object, use that
   let result = colorSpec;
   if (colorSpec) {
-    if (theme.dark && colorSpec.dark !== undefined) {
+    if (
+      (forceDark === true || (forceDark === undefined && theme.dark)) &&
```
ShimiSun: I would think that `(forceDark === false && theme.dark)` should also be accepted in this condition; If I'm not forcing the dark mode but the theme itself is already dark, then the result will be the dark color.

ericsoderberghp: If you're not forcing the dark mode, then forceDark will be undefined. If it is explicitly set to `false`, then it takes precedence.

ShimiSun: Maybe the naming is confusing to me then, I think that if `forceDark === false` it means I'm not forcing it to happen, but if it is happening by itself (i.e. theme.dark) then that is ok. Would it make sense to rename `forceDark` to `isDark` or `setDark`?

ericsoderberghp: Hmm... I was thinking "forceDark" was even stronger, that you were saying you wanted "dark" to be forced to whatever "forceDark" indicated. Maybe making the argument to be just "dark" would be clearer?

ShimiSun: Yes, it will make more sense to me if it is just `dark`. Once `dark` is set to false then it will make sense not to pick the dark theme.

ericsoderberghp: OK. Renamed, and added a comment explaining more.

PR #3882, opened 2020-03-24
Changed Form FormContext to fix an issue with caret positioning

ericsoderberghp:
#### What does this PR do?

Changed Form FormContext to fix an issue with caret positioning.

Form.js

Added three unit tests to cover the three flavors of value control: controlled at the Form level, controlled at the Input level, and uncontrolled.
Also added two stories for additional coverage as well.

new Form stories

no

yes

backwards compatible


ericsoderberghp:
> The Reset Button isn't working on this story example
> image

Fixed

Issue #3881, opened 2020-03-24
Make root element configurable for Drop/Layer containers


brentropy:
### Expected Behavior

I have a case where I need the container elements for `Drop`s and `Layer`s to be inserted somewhere other than `document.body`. There may be other uses, but my particular use cases is using grommet inside a shadow root (using [ReactShadow](https://github.com/Wildhoney/ReactShadow)). I am working on a browser extension where I need complete style isolation for UI rendered within inside any web page. For the most part grommet styled components work great in the shadow root, but because the Drop/Layer containers are rendered outside of the shadow root that don't have any styles. If those elements could instead be inserted in the shadow root it would be solved.

I would be happy to submit a PR for this, but wanted to get feedback first. My initial thoughts on this would be to add an optional `containerRoot` prop on `Grommet` which would use a context to make the container root available to child components. It would continue  to default to `document.body` so it would be a non-breaking change. I would create a `useContainer` hook so the components would not need to be aware of the context directly. The `getNewContainer` function in `utils/DOM.js` would take an optional param for the root. It could also default to `document.body` in case anyone is using it directly.

Would you be good with this being available in grommet? What are your thoughts on how I would attack it in a PR?

### Actual Behavior

Drop/Layer containers are always inserted in `document.body`.

### Demo

https://codesandbox.io/s/blissful-murdock-cbrer

### Steps to Reproduce

1. Use grommet inside a shadow root
2. Use a Drop/Layer component.

### Your Environment

- Grommet version: `2.11.2`
- Browser Name and version: `Firefox 74.0`
- Operating System and version (desktop or mobile): `Ubuntu 19.10`

brentropy:
Since this was blocking me, I went ahead and got this working for me and submitted a PR (#3883). If this isn't the best way that's fine, but since I got it working for myself I figured I would share my solution.

ShimiSun:
Thanks for reporting + fixing! awesome work

PR #3883, opened 2020-03-25
Add option to change the root node for Drop/Layer containers

brentropy:
#### What does this PR do?

Makes the element Drop/Layer containers appended to configurable through the `Grommet` component instead of always doing directly in `document.body`. My particular use case was with a browser extension rendered in a shadow root. See #3881 for more information.

I didn't really wait for feedback on my initial issue, so feedback on the general approach would be great. I would also make sure there isn't anything I should have considered that was overlooked.

I included tests for `Drop` and `Layer` and have also manually tested this in my project using grommet inside of a shadow root.

I was specifically thinking about working in a shadow root, but it could just as easily be tested by using any other element as the `containerRoot`. 

I created #3881, but since this is currently blocking progress on my project, I got a working solution going without waiting for a response. If there is another approach that makes more sense I don't feel bad about throwing this away. At the very least this is meeting my needs to keep developing on my other project.

I updated the the `doc.js` file for `Grommet`. It looks the the documentation site generates based on that. There may be other place that should be updated though.

Sure.

I made sure to keep this backwards compatible.


brentropy:
I am not sure why I am getting snapshot differences between my machine and CI for `data-g-tabindex` and `tabindex` :confused: I will fix those.

ericsoderberghp:
Nice addition. I think it just needs to be merged with the latest master.

brentropy:
@ericsoderberghp Thanks, it should be good to go.

src/js/components/Grommet/doc.js
```diff
@@ -37,9 +37,14 @@ export const doc = Grommet => {
...
     ),
     userAgent: PropTypes.string.description(
-      `User agent used to detect the device width for setting the initial 
+      `User agent used to detect the device width for setting the initial
       breakpoint.`,
     ),
+    containerRoot: PropTypes.object.description(
+      `The node where Drop and Layer containers are inserted. Defaults to
+      document.body which is almost always the right choice. This is used
+      for less common cases like rendering within a shadow root.`,
```
ShimiSun:
```suggestion
      for less common cases like rendering within an internal node (e.g shadow root).`,
```
Would this description make more sense? I'm trying to keep it as generic as possible.

src/js/components/Layer/__tests__/Layer-test.js
```diff
@@ -297,4 +297,32 @@ describe('Layer', () => {
...
+      <Grommet>
+        <Layer data-testid="layer">Test</Layer>
+      </Grommet>,
+    );
+    const layer = getByTestId(document, 'layer');
+    const actualRoot = layer.parentNode.parentNode.parentNode.parentNode;
+    expect(actualRoot).toBe(document.body);
+  });
+
+  test('custom containerRoot', () => {
```
ShimiSun: Thanks for adding tests!

src/js/components/Grommet/doc.js
```diff
@@ -37,9 +37,14 @@ export const doc = Grommet => {
       theme.`,
     ),
     userAgent: PropTypes.string.description(
-      `User agent used to detect the device width for setting the initial 
+      `User agent used to detect the device width for setting the initial
       breakpoint.`,
     ),
+    containerRoot: PropTypes.object.description(
```
ShimiSun: Thoughts about renaming the prop to `containerTarget`? It will be more consistent with the overall grommet naming.

src/js/components/Grommet/doc.js
brentropy: Looks good to me.

ShimiSun: Great, approve this fix and update the snapshots.

src/js/components/Grommet/doc.js
brentropy: That works for me, I can get it change in all of the places.

ShimiSun:
Thank you for your contribution! good luck with your project and hope we'll see you around more

brentropy:
Thank you @ShimiSun

PR #3884, opened 2020-03-25
MaskedInput scroll bars goes back to the top fix

IanKBovard:
#### What does this PR do?
Addresses issue #3854 where the scrollbar would move to the top on every mouse over. There is a useEffect hook in DropContainer.js that looks like, to me, just creates and cleans up events. That hook had a dependency array with dependencies passed through to it causing it to run on every update. I removed the dependencies from the array causing the hook to only run on mount, and unmount(which, i think, is what was intended).

Manual testing in storybook

Tested with storybook.

issue #3854 

no
no
backwards compatible


IanKBovard:
One more thing, before the change to hooks DropContainer had componentDidUpdate with place(true). I don't see anywhere in this file that place ever gets called with true.

ericsoderberghp:
All of those properties in the dependency array are there because the functions inside use them. So, if a caller changes any of them, the behavior should align with the changes. I'd be curious to know what prop was changing, causing the effect to be re-evaluated. I didn't see any obvious ones in the list but perhaps I'm missing something.

IanKBovard:
Thanks for responding. I understand that, what I don't understand is when you need to remake the event handlers. Before the change to hooks, events were created onMount and removed on unMount. When refactoring to hooks isn't the intention to keep previous functionality?

compare : before hooks to now.
https://github.com/grommet/grommet/blob/f4611c750e297e560d35a6403f2eae41a20a4ad7/src/js/components/Drop/DropContainer.js

the only thing that changed onUpdate was this.place(true); which is not represented in the current code.

ericsoderberghp:
Superseded by https://github.com/grommet/grommet/pull/4051

PR #3886, opened 2020-03-26
Make sure theme always exists in all components

cozmo:
#### What does this PR do?
When rendering a `FormField` with react server side rendering it was throwing an error - 
```
TypeError: Cannot read property 'formField' of undefined
  at Object.render (/app/node_modules/grommet/components/FormField/FormField.js:191:25)
```

I'll be the first to admin I haven't dived deep into the styled components/react context/server side rendering interplay, but looking at the line, I noticed that it was missing the `|| defaultProps.theme` fallback case that most of the other components had, and adding that fixed the issues. 

Is there a reason components inconsistently had this fallback case? Should this defaulting be done higher up so that folks can't make this mistake in the future? 

Tested that it fixed the rendering described above.

I think the most valuable thing to understand is - are these defaults always safe to fallback to? I feel I don't have 

N/A

Non filed that I could find. 

N/A

No. 

Probably not?

Backwards compatible.


ShimiSun:
Hi buddy, thanks for this PR.
I agree there is a lack of inconsistency but in order to understand what is the best solution for the problem we need to understand the root cause of the issue better. will you be willing to share a small code-example project that reproduces the problem you are describing?
I'm wondering if your FormField isn't wrapped with the Grommet component that sets the theme and that may be the cause of the issue. Let us know.

ShimiSun:
The `defaultProps` is enabling the usage of themes outside of the Grommet component.
We recommend the best practice of using the Grommet component as a wrapper, and fetch its theme via a prop instead of not wrapping with the Grommet component.
In order to accommodate the use case of fetching a theme outside of the Grommet wrapper, we will accept this PR.

Issue #3876, opened 2020-03-23
TextInput drop issues


ShimiSun:
### Expected Behavior

1. Drop should open as you focus on it - https://storybook.grommet.io/?path=/story/textinput--custom
2. In case of multiple selections of suggestions (like in tags), after making one selection the focus remains on the TextInput but the drop isn't opening again. the drop should keep opening to allow making another selection - https://storybook.grommet.io/?path=/story/textinput--tag

### Your Environment

storybook latest

- Grommet version: master
- Browser Name and version: any
- Operating System and version (desktop or mobile): desktop (probably on mobile as well)

PR #3888, opened 2020-03-27
Changed TextInput to fix an issue with drop opening

ericsoderberghp:
#### What does this PR do?

Changed TextInput to fix an issue with drop opening

TextInput.js

assorted stories

stories

#3876 

no

yes

Requires suggestions to not change on every render, should `useMemo` or put in state as needed.


ericsoderberghp:
> Looks good!
> One comment, on the [Custom story](https://deploy-preview-3888--sad-tereshkova-173d07.netlify.com/?path=/story/textinput--custom), the drop doesn't open on focus, it only opens when a string that matches one of the suggestions was entered.

That appears to be the design of the story, the filter on `suggestedFolks` only matches if there's something in the `value`.

Issue #3893, opened 2020-03-28
Calendar not showing individual selected dates


eamrhein:
### Expected Behavior

When I click a date I can add it to my array of selected dates or remove it on a double click, but Individually selected dates are not highlighted at all in the calendar. Shouldn't all selected dates  be  highlighted in some way?
### Actual Behavior
only the current selected date is being highlighted

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/s/grommet-v2-template-ff1wp

### Your Environment

- Grommet version: v2
- Browser Name and version:  google-chrome 82.0.4085.12
- Operating System and version (desktop or mobile):
Desktop Linux

ShimiSun:
Thanks for reporting! The fix is available now on grommet v2.11.4

PR #3905, opened 2020-04-01
Changed Calendar to fix an issue with selecting multiple dates

ericsoderberghp:
#### What does this PR do?

Changed Calendar to fix an issue with selecting multiple dates.

Added Calendar -> Multiple story.
Added unit tests. Added `animate={false}` to the unit tests to avoid complaints about calling setState() in useEffect() without act().

Calendar.js

storybook

storybook

#3893 

no

yes

backwards compatible

### DIFF
--- a/src/js/components/Calendar/__tests__/Calendar-test.js
+++ b/src/js/components/Calendar/__tests__/Calendar-test.js
@@ -141,7 +141,7 @@ describe('Calendar', () => {
     const onSelect = jest.fn();
     const { getByText, container } = render(
       <Grommet>
-        <Calendar dates={DATES} onSelect={onSelect} />
+        <Calendar dates={DATES} onSelect={onSelect} animate={false} />
       </Grommet>,
     );
     expect(container.firstChild).toMatchSnapshot();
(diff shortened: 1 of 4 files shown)


src/js/components/Calendar/__tests__/Calendar-test.js
```diff
@@ -113,10 +116,37 @@ describe('Calendar', () => {
...
+        <Calendar date={DATE} onSelect={onSelect} animate={false} />
+      </Grommet>,
+    );
+    expect(container.firstChild).toMatchSnapshot();
+    fireEvent.click(getByText('17'));
+    expect(onSelect).toBeCalledWith(expect.stringMatching(/^2018-01-17T/));
+    expect(container.firstChild).toMatchSnapshot();
+  });
+
+  test('select dates', () => {
```
ShimiSun, line 140: This test looks the same as the previous test except it is missing the `animate`, wasn't sure if that was on purpose.
I would think that this test should simulate more than one date selection since the test name has `dates` (plural).

ericsoderberghp: All should have `animate={false}`.

Calendar `onSelect` always provides the date the user clicked on. It's up to the caller to decide what to do with it. But, internally it is different whether you start with a single date or multiple dates.

ShimiSun: got it, thanks for clarifying.

ShimiSun: ...added `animate={false}`.

PR #3913, opened 2020-04-02
Changed side-effects to include new context - don't merge yet

ShimiSun:
Suspecting the new containerTarget context is causing build issues on SSR.

Don't merge yet - I haven't verified it is actually working yet, I'm still waiting for the official issue and errors to be shared.


ericsoderberghp:
Seems reasonable. Any reason AnnounceContext isn't there as well?

PR #3915, opened 2020-04-02
added overflow to Box for background color to show correctly

britt6612:
#### What does this PR do?
This PR adds `overflow= 'hidden'` to `box` when there is a border
formfield.js
storybook 
if you add something similar to 

```
formfield: {
   error: { 
      background: {
         color: ''pink"
      }
   }
}

```

You will see that when on a required `formfield` with `textinput` it will make the whole background color pink if there is a error

The same can be done above using all Children in `Form` stories
Currently the corners are cut off
Screen Shot 2020-04-02 at 4 37 15 PM

With the fix
Screen Shot 2020-04-02 at 2 26 52 PM

no
no
backwards compatible


ericsoderberghp:
Looks like maybe another snapshot needs updating?

britt6612:
yeah I believe so trying to pull from master.

Issue #3914, opened 2020-04-02
SSR (Next.js) build fails when using Grommet/Stable branch


halocline:
### Expected Behavior

SSR (Next.js) build should complete successfully.

### Actual Behavior
Summary: Build fails
As Next.js attempts to build, it proceeds to pre-render each page in the site. During this step, logs report an "Error occurred prerendering page" & "ReferenceError: document is not defined" for each page in the site. The end result is a failed build.

Notes: 
- Error first occurred 5:30 MDT
- Build successfully completes when using Grommet v2.12.0
- [Error message documentation](https://github.com/zeit/next.js/blob/master/errors/prerender-error.md)
- Hunch: Grommet is trying to reference `document` during SSR which doesn't exist since `document` is client side only.
- Potentially related to the `ContainerTargetContext` depending on `document` introduced in [this PR](https://github.com/grommet/grommet/pull/3883)

### URL, screen shot, or Codepen exhibiting the issue

Screenshot of log (full log in following comment)
![Screen Shot 2020-04-02 at 10 11 17 AM](https://user-images.githubusercontent.com/1756948/78290831-92f1b880-74e1-11ea-855d-84a58aa34918.png)

### Steps to Reproduce

1. Next.js + Grommet stable branch as dependency (Here's a quick repo I set up as an [environment to reproduce](https://github.com/halocline/grommet-nextjs-starter))
2. 
```
yarn
yarn build
``` 

### Your Environment

- Grommet version: https://github.com/grommet/grommet/tarball/stable

Next.js - v9.1.2

- Browser Name and version: N/A
- Operating System and version (desktop or mobile): N/A

halocline:
5:30:29 PM: Build ready to start
5:30:31 PM: build-image version: 2dbd444fcdce00cf06325060a8238d5ae3e86774
5:30:31 PM: build-image tag: v3.3.7
5:30:31 PM: buildbot version: bd31cf433d2d899021d41733f271c39fdf2f8880
5:30:32 PM: Fetching cached dependencies
5:30:32 PM: Starting to download cache of 229.9MB
5:30:33 PM: Finished downloading cache in 1.271218109s
5:30:33 PM: Starting to extract cache
5:30:46 PM: Finished extracting cache in 13.21818613s
5:30:46 PM: Finished fetching cache in 14.555191645s
5:30:46 PM: Starting to prepare the repo for build
5:30:47 PM: Preparing Git Reference pull/629/head
[...]

PR #3918, opened 2020-04-03
Fix SSR with document referenced in Grommet render

brentropy:
Signed-off-by: Brent Burgoyne <brent47@gmail.com>

#### What does this PR do?

Fixes broken SSR (#3914) caused by me referencing `document` in the `Grommet` render method in a previous commit.

I have tested this with a reproduced failing nextjs build and it is working now. I started to go down the rabbit whole of adding better SSR tests, but with `styled-components` i was running in to issues with having those tests in the same test suite/babel config as everything else so I bailed. It may be worth revisiting something like that in another commit. Maybe a separate jest suite with `environment: node` that just does a sanity check `ReactDOMServer.renderToString()`.

This can be tested with a basic nextjs build rendering `<Grommet>`. For some reason I wasn't able to reproduce the issue with gatsby.

No

#3914

N/A

 No

No

Yes

### DIFF
--- a/src/js/components/Grommet/Grommet.js
+++ b/src/js/components/Grommet/Grommet.js
@@ -123,7 +123,9 @@ class Grommet extends Component {
     const {
       children,
       full,
-      containerTarget = document.body,
+      containerTarget = typeof document === 'object'
+        ? document.body
+        : undefined,
       ...rest
     } = this.props;
     delete rest.theme;


brentropy:
I also discovered this tiny [browser-monads](https://github.com/Jense5/browser-monads) library that provides a SSR safe way to refer to `document` and `window` and chain anything off of them without errors. What do you all think? I am happy to change my fix to use this if you would like.

Issue #3861, opened 2020-03-17
List with 'action' property generates a warning


jim-king-2000:
### Expected Behavior

```javascript
      <List
        primaryKey="name"
        data={[
          { name: 'Alan', percent: 20 },
          { name: 'Bryan', percent: 30 },
          { name: 'Chris', percent: 40 },
          { name: 'Eric', percent: 80 },
        ]}
        action={d => <Menu icon={<More />} hoverIndicator items={[{ label: "one" }]} />}
      />
```

No warning should be emitted.

### Actual Behavior

There's a warning:

index.js:1 Warning: Each child in a list should have a unique "key" prop.

Check the render method of `InfiniteScroll`. See https://fb.me/react-warning-keys for more information.
    in Box (created by InfiniteScroll)
    in InfiniteScroll (created by List)
    in ul (created by List__StyledList)
    in List__StyledList (created by List)
    in Keyboard (created by List)
    in List (created by List)
    in List (created by FocusableComponent)
    in FocusableComponent (created by List)
    in List (created by WithTheme(List))
    in WithTheme(List) (at pages/index.js:7)
    in div (created by StyledGrommet)
    in StyledGrommet (created by Grommet)
    in Grommet (at pages/index.js:6)
    in Unknown (created by App)
    in App
    in Container (created by AppContainer)
    in AppContainer

**If the 'action' property is removed, there is no warning.**

### Steps to Reproduce

See sandbox here: https://codesandbox.io/s/distracted-taussig-hdvjb

### Your Environment

- Grommet version: 2.11.2
- Browser Name and version: Chrome v80
- Operating System and version (desktop or mobile): Win10

IanKBovard:
Changing action to this fixes the issue.

```
action={(d, index) => <Menu key={index} icon={<More />} hoverIndicator items={[{ label: "one" }]} />}
```

Menu needs a unique key prop because I'm assume list is mapping.

jim-king-2000:
Great! It should be documented or in a sample of storybook. Feel free to close it.

IanKBovard:
Gotcha, thanks for double checking for me. I thought I reloaded the console when I made the change :man_facepalming: I'll take a look at it. Seems this issue has existed since, at least, 2.10

Rashi-jain21:
Hi I want to resolve this issue.

bsdis:
Any news on resolving this? Any help needed?

IanKBovard:
[...]

PR #3927, opened 2020-04-06
List with 'action' property generates a warning #3861 fix

IanKBovard:
#### What does this PR do?
Fixes issue #3861 where adding the action prop to a List component would generate the react unique key error. 

I added a key of actionContainer[index] the content when action is passed in. I did not simply make it index because a key must still be provided in the action prop; shown in the list storybook example.

Both need a unique key to get rid of the warning. This issue seems to have been around since, at least 2.9.
```
        action={(elem, index) => {
          return (
            <Menu
             // this is needed as well to prevent the error
              key={index}
              icon={<More />}
              hoverIndicator
              items={[{ label: 'one' }]}
            />
          );
        }}
      />
```

List.js
```
if (action) {
              content = [
                //added key here
                <Box align="start" key={`actionContainer${index}`}>
                  {content}
                </Box>,
                action(item, index),
              ];
```
Tested in the List storybook example action.js

Same as above

No

No

Backwards compatible

### DIFF
--- a/src/js/components/List/stories/action.js
+++ b/src/js/components/List/stories/action.js
@@ -12,7 +12,8 @@ const ActionList = () => (
     <Box pad="large">
       <List
         data={data.slice(0, 10)}
-        action={(elem, index) => {
+        pad={{ left: 'small', right: 'none' }}
+        action={(item, index) => {
           return (
             <Menu
               key={index}
--- a/src/js/components/List/stories/action.js
+++ b/src/js/components/List/stories/action.js
@@ -12,10 +12,16 @@ const ActionList = () => (
     <Box pad="large">
       <List
         data={data.slice(0, 10)}
-        pad={{ left: 'small', right: 'none' }}
-        action={() => (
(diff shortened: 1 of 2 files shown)


src/js/components/List/stories/action.js
```diff
@@ -12,10 +12,16 @@ const ActionList = () => (
     <Box pad="large">
       <List
         data={data.slice(0, 10)}
-        pad={{ left: 'small', right: 'none' }}
-        action={() => (
-          <Menu icon={<More />} hoverIndicator items={[{ label: 'one' }]} />
-        )}
+        action={(elem, index) => {
```
ShimiSun, line 15: why was the `pad` removed?

IanKBovard: it created an error message, invalid prop type

IanKBovard: i can throw it back in

IanKBovard: ok, back in

Issue #3931, opened 2020-04-07
Input - Allow input padding to have top-right-bottom-left


halocline:
### Expected Behavior

Ability to specify padding on input fields by top-right-bottom-left, horizontal, vertical, etc.

### Actual Behavior

Currently, the padding is one-size and consistent across each direction. Would like to specify padding independently.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #3939, opened 2020-04-08
Allow global input padding to accept an object

taysea:
#### What does this PR do?
This PR allows the theme.global.input.padding to accept an object as opposed to just a single string. This will allow for specification of padding that can differ between sides instead of always being uniform.

src/js/utils/styles.js

A test has been added to `theme-test` to ensure backwards compatibility of the old behavior. These changes have been tested in storybook TextInput Themed.

Adjust the theme.global.input.padding to be an object.

We want to allow different padding between the different sides of an input and right now only a string is accepted.

#3931 

Yes, added in this PR. 

No.

Yes, backwards compatible.


src/js/themes/base.js
```diff
@@ -274,7 +276,12 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         },
       },
       input: {
-        padding: `${baseSpacing / 2}px`,
```
ericsoderberghp, line 277: Just checking, did we have to change this? I'm assuming the snapshots would have aligned either way.

src/js/utils/styles.js
```diff
@@ -208,18 +234,56 @@ export const focusStyle = css`
...
+
+  // for backwards compatibility we need to add back the control border width
+  // in each case, check first if the value is an edgeSize,
+  // then if it's a string value.
+  const adjustedPad =
+    parseMetricToNum(`${props.theme.global.edgeSize[pad] || pad}px`) +
+    parseMetricToNum(`${props.theme.global.control.border.width}px`);
+
+  // if still undefined, there should be no pad.
+  return `${adjustedPad}px` || '0px';
```
ericsoderberghp, line 261: This will never be `0px` as the first string will always be `true`.

src/js/utils/styles.js
```diff
@@ -208,18 +234,56 @@ export const focusStyle = css`
...
+  const pad =
+    props.theme.global.input.padding[side] ||
+    props.theme.global.input.padding[orientation];
+
+  // for backwards compatibility we need to add back the control border width
+  // in each case, check first if the value is an edgeSize,
+  // then if it's a string value.
+  const adjustedPad =
+    parseMetricToNum(`${props.theme.global.edgeSize[pad] || pad}px`) +
+    parseMetricToNum(`${props.theme.global.control.border.width}px`);
```
ericsoderberghp, line 258: I thought we weren't going to do anything with border if the `theme.global.input.padding` is an object?

src/js/utils/styles.js
```diff
@@ -208,18 +234,56 @@ export const focusStyle = css`
   }
 `;
 
+export const getInputPadBySide = (props, side) => {
+  if (typeof props.theme.global.input.padding !== 'object') {
+    return props.theme.global.input.padding;
```
ShimiSun, line 239: would that work if the `props.theme.global.input.padding` is a string of t-shirt sizes?

src/js/themes/base.js
```diff
@@ -274,7 +276,12 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
+        padding: {
```
ShimiSun, line 279: I believe we need to adjust it on base.d.ts.

src/js/themes/base.js
taysea: No, we didn't need to change it. Just was updating based on conversation that it was okay to change base.js to use an object.

src/js/themes/base.js
taysea: Yes, you're right. Let me update.

src/js/utils/styles.js
taysea: Gotcha, let me remove the check then.

src/js/utils/styles.js
taysea: This is used on StyledPlaceholder in TextInput and MaskedInput. Based on how grommet was functioning prior to these changes, the padding was subtracting the border width from the theme value, but the placholder was not. So, for example, if the value defined in the theme for `global.input.control.padding` was `12px`, then the padding value would be calculated to `11px`, but the styledplaceholder pad would still be `12px`. Because we're now placing the subtraction into the theme itself, we have to add back in the border width.

Unless, we don't want these values to be different, but for backwards compatibility this addition needs to remain.

src/js/utils/styles.js
taysea: Good call, I've updated to allow this to work.

src/js/themes/base.js
ShimiSun: why are we changing it if we don't have to?

src/js/utils/styles.js
ericsoderberghp: still removing?

src/js/utils/styles.js
ericsoderberghp: Ah, thanks for the details. I think it would help to add some of this as a comment above the function here, mentioning that it is for placeholder and icon placement with TextInput and MaskedInput.

src/js/themes/base.js
taysea: I am fine leaving it as it was, and just allowing grommet to support an object if a theme (such as grommet-theme-hpe) wanted to do that. I'll do that.

src/js/utils/styles.js
taysea: Sounds good, let me add that comment in.

src/js/themes/base.js
taysea: Actually, for backwards compatibility we have a check in styles L281 to maintain backwards compatibility for themes using just a string. However, we do need to update base.js to be an object.

PR #3941, opened 2020-04-08
Allow theme to specify active border color for primary button

taysea:
#### What does this PR do?
Adds control of the active border color for primary buttons. As shown in this design, we want to be able to control the border of primary buttons when active: https://www.figma.com/file/Oi5UEEQ33VCAyOKQcZ8Nr8/hpe-button?node-id=0%3A1

src/js/components/Button/StyledButton.js

A button was added to the `Basic` story for buttons.

Create a custom theme and adjust the value for `button.primary.active.border.color`

Allow theme to have more control of various button types to allow us to align the Design System designs.

Yes, they've been updated here.

No.


src/js/components/Button/__tests__/Button-test.js
```diff
@@ -127,6 +127,26 @@ describe('Button', () => {
...
+    const component = renderer.create(
+      <Grommet>
+        <Button active label="Button" />
+      </Grommet>,
+    );
+    const tree = component.toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  test('active + primary', () => {
```
ShimiSun, line 140: thanks for adding a test

src/js/themes/base.js
```diff
@@ -379,6 +379,11 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       },
       // color: { dark: undefined, light: undefined }
       primary: {
+        active: {
+          border: {
+            // color: { dark: undefined, light: undefined }
```
ShimiSun, line 384: According to the documentation you added, this value should default to `control` color, according to this impl it should default to undefined.

taysea: Gotcha -- so right now, to maintain backwards compatibility, the border falls back to 'control' color unless overridden by the theme. Should I changes the docs to say 'undefined', or should I update the default color to 'control'?

ShimiSun: I think it will be better to set it to `control` on the primary.active.border theme, and remove the default fallback on the code, otherwise, the color wouldn't be documented anywhere since it is a code side-effect and not part of the theme or props default behavior.

taysea: Sounds good. I like that approach.

src/js/components/Button/doc.js
```diff
@@ -214,6 +214,11 @@ export const themeDoc = {
     description: `The color of the background for primary buttons.`,
     type: 'string | { dark: string, light: string }',
   },
+  'button.primary.active.border.color': {
+    description: `The color of the border for active primary buttons.`,
+    type: 'string | { dark: string, light: string }',
+    defaultValue: 'control',
```
ShimiSun, line 220: what was the `active` border-color before this PR? asking for backward compatibility reasons.

src/js/components/Button/StyledButton.js
```diff
@@ -54,7 +54,12 @@ const padStyle = props => {
 const basicStyle = props => css`
   border: ${props.theme.button.border.width} solid
     ${normalizeColor(
-      props.colorValue || props.theme.button.border.color || 'control',
+      (props.active &&
+        props.primary &&
+        props.theme.button.primary.active.border.color) ||
+        props.colorValue ||
+        props.theme.button.border.color ||
+        'control',
```
ShimiSun, line 62: Since we changed the base.js to default to `control` I don't think we will get to the other conditions on this list, I wonder if it should be the default and documented to `undefined`. thoughts?

taysea: I think we still need the other conditions because the first set of conditions just checks in the case where a button is `active` and `primary`, otherwise it should do they same series of conditions as before.

taysea: Simplified this section by creating a `activeButtonStyle` that does this check. It makes more sense for it to live in an `active` function as opposed to the `basicStyle`. Using the mindset we had applied to disabledStyle in that it uses the `activeStyle` from utils and then adds on any button specific stlying.

src/js/components/Button/README.md
```diff
@@ -529,6 +529,16 @@ Defaults to
 undefined
 ```
 
+**button.primary.active.border.color**
```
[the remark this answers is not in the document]
taysea: Done.

src/js/components/Button/__tests__/__snapshots__/Button-test.js.snap
```diff
@@ -24,6 +24,7 @@ exports[`Button active + primary 1`] = `
   border-radius: 18px;
   background-color: rgba(221,221,221,0.4);
   color: #000000;
+  border: 2px solid #7D4CDB;
```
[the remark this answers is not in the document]
taysea: Resolved.

src/js/components/Button/stories/Custom.js
```diff
@@ -27,6 +27,12 @@ const customTheme = {
     },
     primary: {
       color: '#2196f3',
+      active: {
+        border: {
+          color: 'pink',
+        },
+        extend: `border: 2px dashed red;`,
+      },
```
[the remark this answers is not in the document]
ShimiSun: Done. thanks!

src/js/components/Button/stories/Basic.js
```diff
@@ -17,6 +17,15 @@ const BasicButtons = props => (
...
     </Box>
+    <Box align="center" pad="medium">
+      <Button
+        primary
+        label="Active Primary"
+        active
+        onClick={() => {}}
+        {...props}
+      />
+    </Box>
```
ShimiSun, line 28: Later on, we should restructure the stories so Active may have its own story for all types of buttons (primary, default...) same for disabled and hover. labeling for `interns`

Issue #3916, opened 2020-04-03
DataTable: The sorted column arrow is always pointing up.


alexzicat:
While the column is still being sorted, the arrow beside the column name is always pointing up.

### Expected Behavior
The arrow should be pointing up or down depending if the sort direction is ascending or descending.

### Actual Behavior
The arrow is always pointing up.

### URL, screen shot, or Codepen exhibiting the issue
![Grommet V2 - DataTable issue](https://user-images.githubusercontent.com/6243503/78313940-6613c480-7526-11ea-8935-f81804f17227.gif)

### Steps to Reproduce

1. Render a DataTable.
2. Use the sortable prop.
3. Click on the column name multiple times.

### Your Environment

- Grommet version: 2.11.2
- Browser Name and version: Chrome 80.0.3987.163
- Operating System and version (desktop or mobile): macOS Mojave 10.14.6

ShimiSun:
Thanks for reporting, a PR is in process.

PR #3943, opened 2020-04-09
Changed sort condition when resolving the icon on DataTable

ShimiSun:
#### What does this PR do?
Changed sort condition when resolving the icon on DataTable

before the fix, the sort icon on https://storybook.grommet.io/?path=/story/datatable--sort wasn't resolved properly, now it should work as expected.
Also tested the `sortable` examples on https://storybook.grommet.io/?path=/story/all--all to make sure they are not being impacted by this change, and it seems to work as expected as well.

fixes https://github.com/grommet/grommet/issues/3916

no
yea
backwards compatible

### DIFF
--- a/src/js/components/DataTable/__tests__/DataTable-test.js
+++ b/src/js/components/DataTable/__tests__/DataTable-test.js
@@ -109,7 +109,7 @@ describe('DataTable', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
-  test('sort', () => {
+  test('sortable', () => {
     const { container, getByText } = render(
       <Grommet>
         <DataTable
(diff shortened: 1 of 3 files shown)


src/js/components/DataTable/__tests__/DataTable-test.js
```diff
@@ -109,7 +109,7 @@ describe('DataTable', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
-  test('sort', () => {
+  test('sortable', () => {
```
ShimiSun, line 112: Changed the name of the test, since it is testing `sortable` and not `sort`.

PR #3950, opened 2020-04-10
Allow Select Icon to accept up icon in theme or on prop

taysea:
#### What does this PR do?
Allows Select icon to accept an `icon.up` in addition to `icon.down` (which already exists). This icon will display if the Select is open. This can be applied in the theme or as a prop on Select.

Tests have been added to snapshots. Additionally, this was tested in Storybook by creating custom Theme and also applying a icon object as a prop to Select.

Create a custom theme and add:
```
select: {
   icons: {
      down: FormDown,
      up: FormUp,
   },
}
```
or add it as a prop if it's only wanted on a certain Select:
```
<Select icon={{down: FormDown, up: FormUp}} />
```

Needed to align with HPE Design System designs: https://www.figma.com/file/fZJdRcNrvZ1JjdPRYHpFFU/hpe-select?node-id=0%3A1

Yes, they've been updated here.

Maybe?


src/js/components/Select/Select.js
```diff
@@ -161,10 +161,13 @@ const Select = forwardRef(
...
       case undefined:
-        SelectIcon = theme.select.icons.down;
+        SelectIcon =
+          open && theme.select.icons.up
+            ? theme.select.icons.up
+            : theme.select.icons.down;
         break;
       default:
-        SelectIcon = icon;
+        SelectIcon = open && icon.up ? icon.up : icon.down || icon;
```
ShimiSun, line 170: I don't think that the `icon` prop knows what are 'up' and 'down', so I'm not sure if `icon.up` or `icon.down` will result anything but undefined.

src/js/components/Select/README.md
```diff
@@ -577,6 +577,16 @@ Defaults to
...
 ```
 
+**select.icons.up**
+
+The up icon to use for closing the Select. Expects `React.Element`.
+
+Defaults to
+
+```
+undefined
```
ShimiSun, line 587: Good call for backward compatibility.

src/js/components/Select/stories/Basics.js
```diff
@@ -84,4 +94,5 @@ for (let i = 1; i <= 200; i += 1) {
 
 storiesOf('Select', module)
   .add('Simple', () => <SimpleSelect />)
-  .add('Custom Theme', () => <SimpleSelect open theme={customRoundedTheme} />);
+  .add('Custom Theme', () => <SimpleSelect open theme={customRoundedTheme} />)
+  .add('Custom Theme Icons', () => <SimpleSelect theme={iconTheme} />);
```
ShimiSun, line 98: I think we have enough stories for Select, so we can just put this on a Custom Theme example rather than creating a new story.

src/js/components/Select/Select.js
taysea: This checks for if the icon provided is an object. So, if the user gives an icon as a prop and it's an object, then it sees if they provided up or down, if they're undefined that means the user didn't provide an object and it should just behave as before.

src/js/components/Select/stories/Basics.js
taysea: Sounds good. I'll update that story.

src/js/components/Select/stories/Basics.js
```diff
@@ -39,6 +40,10 @@ const customRoundedTheme = deepMerge(grommet, {
         border: '1px solid #7D4CDB',
       },
     },
+    icons: {
+      down: FormDown,
+      up: FormUp,
+    },
```
ShimiSun, line 46: According to chromatic, this added theme is flipping the up/down logic https://www.chromaticqa.com/build?appId=5d9774839a6eff00203f5cbf&number=1688

taysea: Chromatic doesn't seem to be picking up that the Select is open by default in that story. The up/down logic is actually correct. Prior, this story allows showed a down arrow, but since this story shows the Select open by default, we would expect it to show the up arrow. 

If you open this story in the deployment, you'll see it's correct. I'm not sure why chromatic doesn't show that the items are displaying: https://deploy-preview-3950--sad-tereshkova-173d07.netlify.com/?path=/story/select--custom-theme

src/js/components/Select/Select.js
ShimiSun: After discussions on this bullet, we agreed that we do not want to inflate the `icon` prop API with an object that contains up/down attributes.
Users can control the icon that is being sent to this prop and can have the logic internally on their application in case they want a separate icon for open/close of the drop.
We will roll this line up to how it was.

src/js/themes/base.js
```diff
@@ -735,6 +735,7 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       // extend: undefined,
       icons: {
         down: FormDown,
+        // up: undefined,
```
ShimiSun, line 738:
```suggestion
```

Issue #3972, opened 2020-04-17
Allow styling of FormField border & label when disabled


halocline:
### Expected Behavior

When `<FormField>` is `disabled`, in addition to the background, I would like to be able to style the FormField's label and border colors.

### Actual Behavior

When FormField has a disabled prop, the disabled prop is not passed to the `<FormFieldBox>` and therefor cannot extend disabled state styling via theme.

### URL, screen shot, or Codepen exhibiting the issue

Here is a design I am attempting to implement for a disabled state. The design calls for a change in the input's border color and label color.
![Screen Shot 2020-04-17 at 2 10 08 PM](https://user-images.githubusercontent.com/1756948/79610701-49df5e00-80b6-11ea-981d-4c5372d311eb.png)

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #3973, opened 2020-04-17
Formfield disabled state styling via theme

halocline:
#### What does this PR do?
Allows for a FormField's disabled state styling via the theme.

- FormField.js
- base.js

- Storybook

- Storybook

https://github.com/grommet/grommet/issues/3972

https://github.com/grommet/grommet/issues/3972

Screen Shot 2020-04-17 at 2 26 00 PM

Yes

Yes

Backwards compatible


src/js/themes/base.js
```diff
@@ -623,6 +623,12 @@ export const generate = (baseSpacing = 24, scale = 6) => {
           color: 'status-disabled',
           opacity: 'medium',
         },
+        border: {
+          // color: undefined,
+        },
+        label: {
+          // color: undefined,
+        },
```
ShimiSun, line 631: Can we please add those types to base.d.ts to enhance typescript support?

ShimiSun: And since those objects are empty I'm thinking they should be commented out.

halocline: Will do.

halocline:
> I'll share some background on FormField stories; we were trying to stay away from independent FormFiled stories since FormField should always be used in the context of Form and we don't want to recommend otherwise.

I can adjust them to be in a Form. Do you want the Form to be presented with a button as well? Within Storybook, do you see them living under Form or FormField?

ShimiSun:
>I can adjust them to be in a Form. Do you want the Form to be presented with a button as well? Within Storybook, do you see them living under Form or FormField?

Do you think that wrapping it with a Form will add complexity to the story? I think it should be ok but wanted to hear your thoughts as well. If we'll add a button it might add that complexity, but if the Form won't make sense without it will be better to have the button on the example as well.

src/js/themes/base.js
halocline: Done.

src/js/themes/base.d.ts
```diff
@@ -535,6 +535,12 @@ export interface ThemeType {
     };
     disabled?: {
       background?: BackgroundType;
+      border: {
```
ShimiSun, line 538:
```suggestion
      border?: {
```

src/js/themes/base.d.ts
```diff
@@ -535,6 +535,12 @@ export interface ThemeType {
...
+        color?: ColorType;
+      },
+      label: {
```
ShimiSun, line 541:
```suggestion
      label?: {
```

halocline:
> Do you think that wrapping it with a Form will add complexity to the story? I think it should be ok but wanted to hear your thoughts as well. If we'll add a button it might add that complexity, but if the Form won't make sense without it will be better to have the button on the example as well.

I think the purpose of the stories are: 1) to illustrate the various visual states an input may have, and 2) to demonstrate that I may style those states to my liking. I think enclosing the fields in a Form would not detract and reinforce that FormField should live in a Form. I think including a button would potentially detract and shouldn't be required for a form, so would recommend omitting the button if we are okay with that.

ShimiSun:
>I think the purpose of the stories are: 1) to illustrate the various visual states an input may have, and 2) to demonstrate that I may style those states to my liking. I think enclosing the fields in a Form would not detract and reinforce that FormField should live in a Form. I think including a button would potentially detract and shouldn't be required for a form, so would recommend omitting the button if we are okay with that.

I like that, let's move in this direction, Thanks!

src/js/components/FormField/__tests__/__snapshots__/FormField-test.js.snap
```diff
@@ -558,6 +558,7 @@ exports[`FormField disabled 1`] = `
 >
   <div
     className="c1 "
+    disabled={true}
```
ShimiSun, line 561: I don't think we should add the HTML disabled attribute to the DOM, the DOM might override the grommet styling for the disabled state which I'm not sure we should allow. 
Was there a reason this was added?  I might be missing something.

halocline:
> According to Chromatic, there is one change that doesn't look like it is backward compatible https://www.chromaticqa.com/snapshot?appId=5d9774839a6eff00203f5cbf&id=5e9e0d734116c200225d43f5
> the snapshot should be intact for the disabled field.

I will investigate.

halocline:
> > According to Chromatic, there is one change that doesn't look like it is backward compatible https://www.chromaticqa.com/snapshot?appId=5d9774839a6eff00203f5cbf&id=5e9e0d734116c200225d43f5
> > the snapshot should be intact for the disabled field.
> 
> I will investigate.

Issue found and corrected.

Screen Shot 2020-04-22 at 10 38 59 AM

src/js/components/FormField/__tests__/__snapshots__/FormField-test.js.snap
halocline: You are right. This was unnecessary and has been removed.

src/js/components/FormField/FormField.js
```diff
@@ -221,13 +221,26 @@ const FormField = forwardRef(
...
     } else if (normalizedError) {
       borderColor = (border && border.error.color) || 'status-critical';
     } else {
       borderColor = (border && border.color) || 'border';
     }
+
+    const labelStyle = formField.label;
+
+    if (disabled) {
+      labelStyle.color = formField.disabled.label
```
ShimiSun, line 239:
```suggestion
      labelStyle.color = formField.disabled && formField.disabled.label
```
Added this precheck in case the user decides to override the formField object unless you think we can live without it.

src/js/components/FormField/FormField.js
```diff
@@ -221,13 +221,26 @@ const FormField = forwardRef(
...
+        ? formField.disabled.label.color
+        : labelStyle.color;
```
ShimiSun, line 241: Do we have a story or a test to verify backward compatibility of `formField.label` object, and color in specific?

src/js/components/FormField/stories/FieldStatesCustom.js
```diff
@@ -0,0 +1,138 @@
...
+              />
+            </FormField>
+          </Box>
+        </Form>
+      </Box>
+    </Grommet>
+  );
+};
+
+storiesOf('FormField', module).add('Field States Custom', () => (
```
ShimiSun, line 136: Since you added a Form to the example, I think it will be better to put it on the Form module, our goal is to clean up the FormField stories.
```suggestion
storiesOf('Form', module).add('Field States Custom', () => (
```
And also move the file to Form stories, same for the other stories.

src/js/components/FormField/FormField.js
ShimiSun: I took care of this one on https://github.com/grommet/grommet/pull/3998

src/js/components/FormField/stories/FieldStatesCustom.js
halocline: Addressed. Done.

src/js/components/FormField/FormField.js
halocline: Thank you. Worked like a charm.

halocline:
@ShimiSun Ran test suite against the new FormField tests and moved stories to live under Form stories.

src/js/components/FormField/__tests__/__snapshots__/FormField-test.js.snap
```diff
@@ -836,7 +836,7 @@ exports[`FormField disabled with custom label 1`] = `
   margin: 6px;
   font-size: 14px;
   line-height: 20px;
-  color: red;
+  color: teal;
```
ShimiSun, line 839: Great!

src/js/components/Form/stories/FieldStatesCustom.js
```diff
@@ -0,0 +1,138 @@
+import React, { useRef, useEffect } from 'react';
+import { storiesOf } from '@storybook/react';
+
+import { grommet, Box, Form, FormField, TextInput, Grommet } from 'grommet';
+import { deepMerge } from '../../../utils';
+
+const customTheme = deepMerge(grommet, {
+  formField: {
+    border: {
+      error: {
```
ericsoderberghp, line 10: This seems like it should be `error.border`, to align with `error.background`.

src/js/components/FormField/FormField.js
```diff
@@ -221,13 +221,26 @@ const FormField = forwardRef(
...
     } else if (normalizedError) {
       borderColor = (border && border.error.color) || 'status-critical';
     } else {
       borderColor = (border && border.color) || 'border';
     }
+
+    const labelStyle = formField.label;
+
+    if (disabled) {
+      labelStyle.color = formField.disabled && formField.disabled.label
```
ericsoderberghp, line 239: This looks like it is modifying the actual theme structure, since we aren't doing `const labelStyle = { ...formField.label }` above.

src/js/components/Form/stories/FieldStatesCustom.js
ShimiSun: We'll track this one on https://github.com/grommet/grommet/issues/4004

halocline: 👍

src/js/components/FormField/FormField.js
halocline: Corrected and implemented.

PR #3974, opened 2020-04-17
Allow input to accept size, height, and weight

taysea:
#### What does this PR do?
Allows theme to specify the size, height, and weight of an input.

src/js/utils/styles.js

Adjust theme-test by adding:
```
font: {
   height: '50px',
   size: 'large',
   weight: 'bold',
}
```
to global.input in custom theme.

Add values for `global.input.font.height` `global.input.font.size` `global.input.font.weight` to a theme.

Line-height was not being respected in the input, this was causing dimensions between designs and dev to differ even when padding was matching.

Yes, they're updated here.

Maybe?

Yes, backwards compatible.

### DIFF
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -296,6 +296,13 @@ export interface ThemeType {
     };
     primary?: {
       color?: ColorType;
+      active?: {
+        border?: {
+          color?: ColorType;
+        };
+        extend?: ExtendType;
+      };
+      extend?: ExtendType;
     };
     size?: {
       small?: {
@@ -828,7 +835,7 @@ export interface ThemeType {
     };
     control?: {
       extend?: ExtendType;
-      open?: boolean;
+      open?: string | object;
     };
     extend?: ExtendType;
     icons?: {
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -204,9 +204,9 @@ export interface ThemeType {
       font?: {
         height?: string;
         size?: string;
-        weight?: string;
+        weight?: number | string;
       };
-      weight?: number;
+      weight?: number | string;
     };
     opacity?: {
       strong?: number;
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -275,6 +275,11 @@ export interface ThemeType {
       radius?: string;
     };
     disabled?: {
+      border?: {
+        color: ColorType;
+      },
+      color?: ColorType;
+      extend?: ExtendType;
       opacity?: OpacityType;
     };
     extend?: ExtendType;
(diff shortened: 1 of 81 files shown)


src/js/themes/base.d.ts
```diff
@@ -201,6 +201,11 @@ export interface ThemeType {
     };
     input?: {
       padding?: string | { top?: string, bottom?: string, left?: string, right?: string, horizontal?: string, vertical?: string};
+      font?: {
+        height?: string;
+        size?: string;
+        weight?: string;
```
ericsoderberghp, line 207: string | number?

src/js/utils/themeDocUtils.js
```diff
@@ -84,9 +84,19 @@ export const themeDocUtils = {
...
+      defaultValue: undefined,
+    },
+    'global.input.font.size': {
+      description: 'The size of the text.',
+      type: 'string',
+      defaultValue: undefined,
+    },
+    'global.input.font.weight': {
+      description: 'The font-weight of the text.',
+      type: 'string',
```
ShimiSun, line 99:
```suggestion
      type: 'string | number',
```

src/js/utils/styles.js
```diff
@@ -294,9 +303,11 @@ export const inputStyle = css`
           props.theme,
         )}
   ${props =>
-    props.theme.global.input.weight &&
+    // for backwards compatibility, fallback to props.theme.global.input.weight
```
ShimiSun, line 306: great, thanks.

ShimiSun:
Looks good, can you take a look at the Chromatic changes and why the styles of the inputs field had changed?

taysea:
> Looks good, can you take a look at the Chromatic changes and why the styles of the inputs field had changed?

Thanks, reviewed and made an update. Going to let the checks finish then confirm that those changes have been reverted to what they already were.

src/js/utils/styles.js
```diff
@@ -294,9 +303,11 @@ export const inputStyle = css`
           props.theme,
         )}
   ${props =>
-    props.theme.global.input.weight &&
+    // for backwards compatibility, check if props.theme.global.input.weight
+    (props.theme.global.input.weight || props.theme.global.input.font.weight) &&
```
ShimiSun, line 307: Communicating here our conversation resolutions.
In case the user defines `theme.global.input.weight`, applying  `props.theme.global.input.font.weight` won't work. we will add to documentation and communicate effectively that those two shouldn't be used together.

src/js/utils/themeDocUtils.js
```diff
@@ -84,9 +84,19 @@ export const themeDocUtils = {
   },
 
   inputStyle: {
-    'global.input.weight': {
-      description: 'The font weight of the text entered.',
-      type: 'number',
+    'global.input.font.height': {
```
ShimiSun, line 87: we should still keep `global.input.weight` and document how it is being replaced.

Issue #3755, opened 2020-02-18
dropProps being ignored in Menu


shovelmn12:
https://github.com/grommet/grommet/blob/afdc12fa0b5311ce037bf5410b32b83463e85660/src/js/components/Menu/Menu.js#L234

dropProps not being passed to DropButton in Menu Component.
need to add to DropButton:
<DropButton
           dropProps={dropProps} //maybe remove align or merge all drop props here
           .../>

IanKBovard:
@shovelmn12 could you give me an example of your problem on codesandbox?

shovelmn12:
Simplest is when i want to set direction of the drop with style, now i cant do it so i rebuilt it using dropbutton.

What i tried is to pass dropProps={{style: { direction: "rtl" } }} but it didnt work and i didnt know why until I checked the source code

IanKBovard:
have you tried style={{direction: 'rtl'}}? yeah, could you give me an example i can toy around with?

ShimiSun:
Menu has an RTL support built-in, not sure what you are trying to achieve but Menu has `margin` prop and using either `start` & `end` should play with the RTL property, give it a try and if it doesn't work share with us a codesandbox with your problem description.

shovelmn12:
https://codesandbox.io/s/frosty-hofstadter-rhqkd

look at the drop when its open. Grommet is set to dir="rtl" but the drop is ltr.
btw its a problem i encountered with drop everywhere. because the drop is mounted outside <Grommet/> i need to specify direction manually for all drops. usually i am building a drop of my own which checks the current language of the app and set direction accordingly.

ShimiSun:
Thanks for elaborating, we'll use your example to enhance Drop to work with RTL.

shovelmn12:
Sure no problems, i love Grommet. Cant build a website without it now, make things so much easier.
I still think that Menu should pass dropProps to DropButton as i originally showed.
That way users can control better the Drop for Menu.

ShimiSun:
We'll take a look into it as well, thanks again!

ShimiSun:
Thanks for the fix @IanKBovard

PR #3629, opened 2019-12-19
Improved RTL support by adding pad, margin, and border 'start' and 'end'

ericsoderberghp:
#### What does this PR do?

Improved RTL support by adding pad, margin, and border 'start' and 'end'

Note, Layer positioning isn't quite all the way there, waiting for more browser support for `insert-inline-start`.

#### Where should the reviewer start?

doc.js related files

#### What testing has been done on this PR?

storybook, unit, designer

#### How should this be manually tested?

storybook, see new RTL stories.

#### Do the grommet docs need to be updated?

yes, we should add a page on right-to-left.

#### Should this PR be mentioned in the release notes?

yes

#### Is this change backwards compatible or is it a breaking change?

backwards compatible

PR #3977, opened 2020-04-20
Menudrop not receiving rtl from theme

IanKBovard:
#### What does this PR do?
Fixes issue #3755 by adding ThemeContext.Consumer to drop js

Drop.js
Tested on storybook and using the codesandbox example here.
https://codesandbox.io/s/frosty-hofstadter-rhqkd

Storybook and codesanbox is best

I'm assume this issue is occuring because of createPortal(). The drop container is created outside of root, and misses out on some props that are passed into grommet.

#3755 

No

No

### DIFF
--- a/src/js/components/Drop/Drop.js
+++ b/src/js/components/Drop/Drop.js
@@ -2,7 +2,7 @@ import React, { forwardRef, useEffect, useState, useContext } from 'react';
 import { createPortal } from 'react-dom';
 
 import { getNewContainer, setFocusWithoutScroll } from '../../utils';
-
+import { ThemeContext } from '../../contexts';
 import { DropContainer } from './DropContainer';
 import { ContainerTargetContext } from '../../contexts/ContainerTargetContext';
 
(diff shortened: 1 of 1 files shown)


ShimiSun:
Is the suggested fix work similar to Layer? I think they should behave the same.

IanKBovard:
ill double check

IanKBovard:
it looks like rtl is applied to components within the layer. is that what you mean?

src/js/components/Drop/Drop.js
```diff
@@ -46,12 +46,17 @@ const Drop = forwardRef(
 
     return dropContainer
       ? createPortal(
-          <DropContainer
-            ref={ref}
-            dropTarget={dropTarget}
-            restrictFocus={restrictFocus}
-            {...rest}
-          />,
+          <ThemeContext.Consumer>
```
ShimiSun, line 49: let's use 
```
```
instead

IanKBovard:
Changes made

ShimiSun:
linking to #3629.

PR #3978, opened 2020-04-20
Upgrade storybook version

negreirosleo:
#### What does this PR do?
It upgrades the storybook version to 5.3.18

Anywhere, no particular order is required

I manually checked a couple of stories and looked for bugs

Just check the stories and assert that everything looks fine

--

#3212

No

No

It is backwards compatible


package.json
```diff
@@ -72,10 +72,10 @@
     "@babel/preset-env": "^7.3.4",
     "@babel/preset-react": "^7.0.0",
     "@emotion/core": "^10.0.7",
-    "@storybook/addon-options": "~5.1.11",
-    "@storybook/addon-storysource": "~5.1.11",
-    "@storybook/addons": "~5.1.11",
-    "@storybook/react": "~5.1.11",
+    "@storybook/addon-options": "~5.3.18",
```
ShimiSun, line 75: would it work if we'll use the caret?
```suggestion
    "@storybook/addon-options": "^5.3.18",
```

negreirosleo: let me check

negreirosleo: I think theres no problem using `^`

negreirosleo: Unless storybook releases a breaking change out of the blue (which is very improbable), I see no issue, im updating this rn

ShimiSun:
We are getting errors from https://www.chromaticqa.com/build?appId=5d9774839a6eff00203f5cbf&number=1785, can you take a look?

negreirosleo:
@ShimiSun Looking into those erros rn

negreirosleo:
@ShimiSun we should plan something for the files that arent stories but fall into the stories folder, cause we are now receiving console warns about it
image

ShimiSun:
I agree. we should exclude them from the loader.

negreirosleo:
@ShimiSun I address that problem in this PR or I leave it for the next one?

ShimiSun:
@negreirosleo I think it can wait for the next PR.

package.json
```diff
@@ -72,10 +72,10 @@
     "@babel/preset-env": "^7.3.4",
     "@babel/preset-react": "^7.0.0",
     "@emotion/core": "^10.0.7",
-    "@storybook/addon-options": "~5.1.11",
-    "@storybook/addon-storysource": "~5.1.11",
-    "@storybook/addons": "~5.1.11",
-    "@storybook/react": "~5.1.11",
+    "@storybook/addon-options": "^5.3.18",
+    "@storybook/addon-storysource": "^5.3.18",
+    "@storybook/addons": "^5.3.18",
```
ShimiSun, line 77: Can you check if we really need all those packages of addons? I think we need the story source one, not sure why we will need the others, see if we can do some package optimizations here.
Also, I think I read somewhere that story source might be deprecated and it is recommended to use another package, but I can't find it anywhere so I hope I'm not imagining anything.

negreirosleo: Sure!!

negreirosleo:
@ShimiSun I removed storybook-options cause it was deprecated but storysource is being used :+1:

ShimiSun:
Hi buddy, can you check why the Chromatic Chart examples are broken? example: https://deploy-preview-3978--sad-tereshkova-173d07.netlify.app/?path=/story/chart--zoom

ShimiSun:
@negreirosleo any progress with this PR?

negreirosleo:
@ShimiSun I'm having issues with the `data.js` through the stories, there are plenty of them, do I just move them one folder up?

ShimiSun:
@negreirosleo Can we nest them on a folder under 'stories' and have that folder ignored by the loader? that will be best.

negreirosleo:
@ShimiSun I totally forgot we could do that lol, thanks!

ShimiSun:
Any updates on this PR @negreirosleo ?

ShimiSun:
Can you please update the lock file?

negreirosleo:
@ShimiSun I will, just noticed some issues with some chart stories I overlooked

storybook/preview.js
```diff
@@ -1,6 +1,5 @@
-import { addParameters, configure } from '@storybook/react';
+import { configure } from '@storybook/react';
```
ShimiSun, line 1: where is the theme defined now that `addParameters` got removed?

src/js/components/Chart/stories/Scan.js
```diff
@@ -123,5 +123,5 @@ const ScanChart = props => {
 };
 
 storiesOf('Chart', module).add('Scan', () => (
-  <ScanChart data={generateData(30, 100)} max={100} />
+  <ScanChart data={generateData()(30, 100)} max={100} />
```
ShimiSun, line 126: I wondering if there is a simpler way to use generateData, it feels a little verbose.

storybook/preview.js
ShimiSun, line 1: Should you remove the `grommetLight` code if it's not being used?

negreirosleo: That theme wasnt being used!
the only theme used on stories was this one: `src/js/all/stories/theme.js`

negreirosleo: It was removed, wasnt it?

negreirosleo:
@ShimiSun every file that was loaded by the storybook that isnt a story turns into this:
image

So in order to use the: `src/js/components/Chart/stories/data.js` and `src/js/all/stories/theme.js` I had to change their call.

On `src/js/all/stories/theme.js` I had to do:
```js
export const { colors } = hpe().global
```
because `hpe` was used as the decorated function that returned the object, so: `hpe().global`

On `src/js/components/Chart/stories/data.js` I had to create a function that returns a function because of this too.
```js
export const generateData = (count, max) => {} // Even if I call generateData(10,100) max will be undefined, because generateData turns into a decorated function.
export const generateData = () => (count,max) => {} // This way it generateData is the decorated function and generateData() returns the function that actually do what we want
```
To reduce verbosity I think we could make:
```js
export const decorator = () => (count,max) => {} // accepting name suggestions for this 
export const generateData = decorator()
```

and then call: 
```js
<ScanChart data={generateData(30, 100)} max={100} />
```

But i dont think its much better :thinking:

ShimiSun:
> To reduce verbosity I think we could make:
> 
> export const decorator = () => (count,max) => {} // accepting name suggestions for this 
> export const generateData = decorator()
> and then call:
> 
> <ScanChart data={generateData(30, 100)} max={100} />
> But i dont think its much better 🤔

I actually think it will be cleaner this way.

negreirosleo:
@ShimiSun Ok, fixing that to be cleaner!

negreirosleo:
@ShimiSun the solution I suggested didnt work, but I made it cleaner by wraping the function call in an object, that way avoiding the creation of a decorated function

storybook/preview.js
ShimiSun: Nice clean up.

ShimiSun:
I'm super excited about this PR, thank you @negreirosleo ❤️

PR #3980, opened 2020-04-20
Allow Tab to accept icon and reverse

taysea:
#### What does this PR do?
Allows user to provide an `icon` to the Tab and control its placement with `reverse`.

src/js/components/Tab/Tab.js

A new storybook has been added, and a new jest-test has been added as a baseline.

Add an icon to a `Tab` and `reverse` if you want.

To align with HPE Design system designs, teams would need to create complex `title` on a Tab to achieve an icon with the tab label -- this opens up lots of room for inconsistency. Because `Tab` is technically a Button, it felt well-suited to allow Tab to accept icon and reverse props like Button does.

Yes, they're updated here.

Yes.

Yes, backwards compatible.

### DIFF
--- a/src/js/components/Tab/Tab.js
+++ b/src/js/components/Tab/Tab.js
@@ -16,11 +16,13 @@ import { StyledTab } from './StyledTab';
 const Tab = ({
   active,
   forwardRef,
+  icon,
   plain,
   title,
   onActivate,
   onMouseOver,
   onMouseOut,
+  reverse,
   theme,
   ...rest
 }) => {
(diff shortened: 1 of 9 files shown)


ShimiSun:
closes #3980

src/js/components/Tab/Tab.js
```diff
@@ -114,9 +146,6 @@ let TabDoc;
 if (process.env.NODE_ENV !== 'production') {
   TabDoc = require('./doc').doc(Tab); // eslint-disable-line global-require
 }
-const TabWrapper = compose(
-  withTheme,
-  withForwardRef,
-)(TabDoc || Tab);
+const TabWrapper = compose(withTheme, withForwardRef)(TabDoc || Tab);
```
ericsoderberghp, line 149: As long as we're here, can we use `const theme = withContext(ThemeContext);` and also just wrap with React's `forwardRef()`? That allows us to remove `compose()`.

taysea: Yes, let me look at that

taysea: I've implemented this @ericsoderberghp, can you confirm it's been done correctly? storybooks functioned the same and no snapshots changed so I think it is correcy.

ericsoderberghp:
nicely done, thanks :)

PR #3984, opened 2020-04-20
Add disabled to Tab docs

taysea:
#### What does this PR do?
Adds disabled to Tab docs. This functionality already works on the grommet end since Tab is built on top of Button, but it is not a prop we document for Tab.

src/js/components/Tab/doc.js

### DIFF
--- a/src/js/components/Tab/index.d.ts
+++ b/src/js/components/Tab/index.d.ts
@@ -2,6 +2,7 @@ import * as React from "react";
 import { Omit } from "../../utils";
 
 export interface TabProps {
+  disabled?: boolean;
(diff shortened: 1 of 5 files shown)


src/js/components/Tab/index.d.ts
```diff
@@ -2,6 +2,7 @@ import * as React from "react";
 import { Omit } from "../../utils";
 
 export interface TabProps {
+  disabled?: boolean;
```
ShimiSun, line 5: If you are using JSX.IntrinsicElements['button'] then the type of `disabled` is derived from the HTML element, so I'm not sure if this is needed.

PR #3990, opened 2020-04-21
Toggle zindex to keep focus intact

taysea:
#### What does this PR do?
This PR ensures that the focus outline around a tab does not get hidden by the hover styling of adjacent tabs. When a tab receives focus, it is given a zindex of 1 to pull it on top of surrounding tabs. onblur, the zindex returns to undefined.

src/js/components/Tab/Tab.js

Tested in storybook.

Use the Icon example on Tabs in storybook. Tab onto one of the tabs, then hover on adjacent tabs. The focus outline should remain visible on all sides.

Hover styling was covering one edge of the tab when there was no space between tabs.

No.

No.

### DIFF
--- a/src/js/components/Tab/Tab.js
+++ b/src/js/components/Tab/Tab.js
@@ -27,7 +27,7 @@ const Tab = forwardRef(
   ) => {
     const theme = useContext(ThemeContext) || defaultProps.theme;
     const [over, setOver] = useState(undefined);
-    const [zIndex, setZIndex] = useState();
+    const [focus, setFocus] = useState();
     let normalizedTitle = title;
     const tabStyles = {};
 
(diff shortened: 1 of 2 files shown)


src/js/components/Tab/Tab.js
```diff
@@ -128,8 +130,17 @@ const Tab = forwardRef(
...
+          // ensure focus outline is not covered by hover styling
+          // of adjacent tabs
+          setZIndex(1);
+          if (onMouseOver) onMouseOver();
+        }}
+        onBlur={() => {
+          setZIndex();
+          if (onMouseOut) onMouseOut();
+        }}
+        style={zIndex && { zIndex }}
```
ShimiSun, line 143: would it make more sense to have the logic around 'focus'  rather than 'zindex'? so something like `style={ focus && {Index: 1 }}`

taysea: I like that mindset. makes line 143 more readable. I'll update : )

taysea: Done. : )

PR #3995, opened 2020-04-22
add background hover to theme for checkbox

britt6612:
#### What does this PR do?
This PR adds to the theme for checkbox for a background hover
checkbox/doc.js
storybook
```

    hover: {
      background: {
        color: 'light-2',
      },
    },

```

design system designs 
design system designs 
no
no
 backwards compatible


src/js/themes/base.js
```diff
@@ -493,6 +493,9 @@ export const generate = (baseSpacing = 24, scale = 6) => {
             light: 'black',
           },
         },
+        background: {
+          color: { dark: undefined, light: undefined },
+        },
```
ShimiSun, line 498: Why is it not commented out if the value is undefined?

src/js/themes/base.d.ts
```diff
@@ -407,6 +407,9 @@ export interface ThemeType {
       border?: {
         color?: ColorType;
       };
+      background?: {
+        color?: ColorType;
```
ShimiSun, line 411: great, thanks!

src/js/components/CheckBox/doc.js
```diff
@@ -97,6 +97,11 @@ export const themeDoc = {
     type: 'string',
     defaultValue: undefined,
   },
+  'checkBox.hover.background.color': {
+    description: 'The background color on hover.',
```
ShimiSun, line 101: Can we add an explanation whether the background color applies on both the label and the box-indicator.

src/js/components/CheckBox/StyledCheckBox.js
```diff
@@ -40,6 +44,7 @@ const StyledCheckBoxContainer = styled.label`
   ${props => props.disabled && disabledStyle}
   ${props => !props.disabled && 'cursor: pointer;'}
   ${props => props.theme.checkBox.hover.border.color && hoverStyle}
+  ${props => props.theme.checkBox.hover.background.color && hoverStyle}
```
ShimiSun, line 47: This line would break if the user won't have `props.theme.checkBox.hover.border.color` defined.

ShimiSun, line 47: With this logic, `hoverStyle` is being called twice.

src/js/components/CheckBox/StyledCheckBox.js
```diff
@@ -14,6 +14,11 @@ const hoverStyle = css`
     border-color: ${props =>
```
taysea, line 14: Since the check got removed on line 42, we will need to check for it here somehow. I implemented some checks like this on this PR at line 128 of StyledButton.js https://github.com/grommet/grommet/pull/3877/files

src/js/components/CheckBox/stories/CustomToggle.js
```diff
@@ -35,6 +35,9 @@ const customToggleTheme = {
       border: {
         color: undefined,
       },
+      background: {
+        color: 'light-2',
```
ShimiSun, line 39: Putting a default value will break the backward compatibility for users that currently not using background or had the background set in a different way.
the object should be undefined and comment out, and this code should go on the hpe theme.

britt6612: This is added to the custom toggle story ? the object is undefined in the base theme file

britt6612: unless you want me to take this out I totally can just added it to custom story to see an example

ShimiSun: sorry, my bad. we are good.

src/js/components/CheckBox/StyledCheckBox.js
```diff
@@ -12,8 +12,14 @@ const hoverStyle = css`
   :hover input:not([disabled]) + div,
   :hover input:not([disabled]) + span {
     border-color: ${props =>
+      props.theme.checkBox.hover.border.color &&
       normalizeColor(props.theme.checkBox.hover.border.color, props.theme)};
   }
+  :hover {
```
ShimiSun, line 18: Why does the `background` have different hover conditions than `border`? 
the conditions for the border are on lines 12-13. I would think they should all be placed together.

src/js/components/CheckBox/StyledCheckBox.js
ShimiSun: Is this comment still relevant @taysea ?

src/js/components/CheckBox/StyledCheckBox.js
britt6612: The conditions for the `hover.border.color` are not applied to the whole box container just the checkbox 

The `hover.background.color` is being applied to the whole box containing both label and checkbox.

ShimiSun:
@britt6612 Is this PR still needed now that we were able to get the styling from the theme?

britt6612:
@ShimiSun yes we still need the hover to the entire box this was not completed in the theme.

britt6612:
@ShimiSun this can be dont with the extend 

```

    extend: ({ theme }) => `
      : hover {
        background-color: pink;
      }

```

for example

Issue #3993, opened 2020-04-22
Select component in React.StrictMode, component is throwing errors


kkgowtamasa:
When we use the Select component in React strict mode then we could see the following errors:
`findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Ref which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://fb.me/react-strict-mode-find-node.`

### Expected Behavior
-The component should render without any console errors.

### Actual Behavior
- Getting error:
`findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Ref which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://fb.me/react-strict-mode-find-node.`

### URL, screen shot, or Codepen exhibiting the issue
![image](https://user-images.githubusercontent.com/9958655/79961996-17ae7180-84a5-11ea-9194-f9f0f38764a2.png)

Code sandbox: https://codesandbox.io/s/objective-glitter-0dzqt?file=/src/App.tsx

### Steps to Reproduce

1. Click on the Select box.
2. See the console errors

### Your Environment
1. Windows and browser is Chrome.

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

kkgowtamasa:
@ShimiSun @ericsoderberghp

IanKBovard:
@kkgowtamasa Taking a look at this issue now

MarkMarine:
Seems to exist in the DataTable as well

Yoruba:
Same error with List

jimwhurr:
People, myself included, seem to be incurring the same fault after it has been 'closed'

ShimiSun:
Thank you for reporting guys, I've reopened the issue. Feel free to make a contribution and help us improve the open-source for StrictMode users.

PR #3996, opened 2020-04-22
Remove findDomNode from infinite scroll

IanKBovard:
#### What does this PR do?
Removes findDomNode from infiniteScroll.js that causes an error in React.Strictmode. Changes Ref component to use react.forwardref.

InfiniteScroll.js

Tested with infinite scroll examples in storybook and using this
```
    <React.StrictMode>
      <Box height="small" overflow="auto">
        <Select
          options={["Device 1", "Device 2", "Device 3"]}
          value={value}
        />
      </Box>
    </React.StrictMode>
```
I would test to check if inifinite scroll behaves as it should, and with the code pasted above.

I'm a little unsure if this is the correct solution. I compared my local storybook to the current on up, and everything behaved the same way. I did some research, and the accepted replacement for findDomNode was to use forwardRef.

#3993 

No

No

Should be backwards compatible.

### DIFF
--- a/src/js/components/InfiniteScroll/InfiniteScroll.js
+++ b/src/js/components/InfiniteScroll/InfiniteScroll.js
@@ -1,5 +1,11 @@
 /* eslint-disable react/no-find-dom-node */
-import React, { useEffect, useMemo, useRef, useState } from 'react';
+import React, {
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+  useCallback,
+} from 'react';
 import {
   findScrollParent,
   findScrollParents,
@@ -8,9 +14,10 @@ import {
 } from '../../utils';
 import { Box } from '../Box';
 
-const Ref = React.forwardRef((props, ref) => (
-  <Box ref={ref}>{props.children}</Box>
-));
+const Ref = React.forwardRef((props, ref) => {
+  const { children } = props;
+  return <Box ref={ref}>{children}</Box>;
+});
 
 const InfiniteScroll = ({
   children,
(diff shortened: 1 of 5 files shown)


IanKBovard:
Not sure if this is the best approach, as this is my first time dealing with something like this. I'm open to make any changes.

ericsoderberghp:
We can't take this approach because we are adding an extra Box element into the DOM. For the case where this is used inside a List (`<ol />`) and the items in the list are `<li />`, this wraps a `<div />` around the `<li />`. Instead, what we want to do is ensure that the caller uses the `ref` argument in their child render function, which we call here: `let child = children(item, itemsIndex, ref);`.

src/js/components/InfiniteScroll/InfiniteScroll.js
```diff
@@ -8,9 +14,10 @@ import {
 } from '../../utils';
 import { Box } from '../Box';
 
-const Ref = React.forwardRef((props, ref) => (
-  <Box ref={ref}>{props.children}</Box>
-));
+const Ref = React.forwardRef((props, ref) => {
+  const { children } = props;
+  return <Box ref={ref}>{children}</Box>;
```
ericsoderberghp, line 19: This approach will not work. The Box will insert a `<div />` into the DOM, interrupting `<tr />` and `<li />` usage. The general approach was to have callers use the `ref` argument to the child render function as a way to avoid using `findDOMNode`.

IanKBovard: Just to clarify. I want to make sure this section of code receives ref?
```javascript
  items.slice(firstIndex, lastIndex + 1).forEach((item, index) => {
    const itemsIndex = firstIndex + index;
    // We only need page refs if we don't know the pageHeight
    // The new way, we pass the ref we want to the children render function.
    let ref;
    if (!pageHeight && itemsIndex === 0) {
      ref = firstPageItemRef;
    } else if (!pageHeight && (itemsIndex === step - 1 || itemsIndex === lastIndex)
    ) {
      ref = lastPageItemRef;
    } else if (show && show === itemsIndex) {
      ref = showRef;
    }

    let child = children(item, itemsIndex, ref);
```

IanKBovard:
closing this issue for now, will come back to it when my work on discover has completed.

robeberhardt:
seeing this error with DataTable as well

PR #3998, opened 2020-04-22
Added FormField tests for disabled state and custom theme.

ShimiSun:
Following a conversation with @halocline, this PR is supporting #3973 and making sure we are maintaining backward compatibility with the coming changes of `disabled` FormField.
The second test has a placeholder code that should be un-commented in order to test the new functionality of #3973, it also has instructions of what to expect from the snapshots after merging it with branch halocline:formfield-disabled-styling.

### DIFF
--- a/src/js/components/FormField/__tests__/__snapshots__/FormField-test.js.snap
+++ b/src/js/components/FormField/__tests__/__snapshots__/FormField-test.js.snap
@@ -274,6 +274,172 @@ exports[`FormField custom formfield 1`] = `
 </div>
 `;
 
+exports[`FormField custom label 1`] = `
+.c0 {
+  font-size: 18px;
+  line-height: 24px;
+  box-sizing: border-box;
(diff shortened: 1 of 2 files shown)


src/js/components/FormField/__tests__/__snapshots__/FormField-test.js.snap
```diff
@@ -274,6 +274,172 @@ exports[`FormField custom formfield 1`] = `
...
+    onSubmit={[Function]}
+  >
+    <div
+      className="c1 "
+      onBlur={[Function]}
+      onFocus={[Function]}
+    >
+      <label
+        className="c2"
+        size="small"
```
ericsoderberghp, line 414: Not part of this review per-se, but we shouldn't be leaking `size` into the DOM.

ShimiSun: Good catch, we'll be tracking it on #4003

PR #4001, opened 2020-04-23
added an option in  theme to take out hover color for accordion header

britt6612:
Starting a discussion on how we can be able to change color of hover or get rid of hover completely 
#### What does this PR do?
This PR adds a prop to the theme of accordion 
`accordion.heading.hover.color `
This will allow the user to set their own values for the color of the 
hover as well as set to undefined if they do not want the hover color to change
accordionPanel/doc.js
storybook
added in 

```
    heading: {
      level: 3,
      hover: {
        color: 'accent-2',
      },
    },

```

storybook
Currently in AccordionPanel.js if there is a `string` provided for the label prop 
in accordion-panel then it uses a `Heading` for the label. Within this heading 
there are 2 props level and color. Level is default to 4 but can be changed within the 
theme. color is set to `hover` which is defined at `onMouseOver` and `onFocus` 
the colors are set to `dark: 'light-4, light: 'dark-3` 

Im open to discussion about other ways this can be achieved as well. I figured I start the conversation here.

There is also the option of not passing a `string` to label for `accordionpanel`
something like 
```
        <AccordionPanel
          label={
            <Text size="xlarge" margin="vertical">
              Panel 2 
            </Text>
          }
        />

```

but figured we wanted to keep it limited to what users can change in design system 

In the design system the hover colors are not wanted but the `heading` is still wanted. 
no
no
backwards compatible


src/js/components/Accordion/stories/Custom.js
```diff
@@ -7,7 +7,12 @@ import { Accordion, AccordionPanel, Box, Grommet, Text } from 'grommet';
 
 const CustomAccordionTheme = {
   accordion: {
-    heading: { level: '3' },
+    heading: {
+      level: 3,
+      hover: {
```
ShimiSun, line 12: I'm not sure if it is semantically correct to place a hover on a heading element vs the interactive element. Is there a reason the `hover` is nested with `heading` and not directly under accordion?

src/js/components/AccordionPanel/AccordionPanel.js
```diff
@@ -48,15 +48,31 @@ const AccordionPanel = forwardRef(
...
+            );
             if (onMouseOver) onMouseOver(event);
           }}
           onMouseOut={event => {
             setHover(undefined);
             if (onMouseOut) onMouseOut(event);
           }}
           onFocus={event => {
-            setHover(theme.dark ? 'light-4' : 'dark-3');
+            setHover(
```
ShimiSun, line 67: Setting hover while focus doesn't feel right to me. but I might be missing something.

src/js/components/AccordionPanel/doc.js
```diff
@@ -16,6 +16,11 @@ export function doc(Panel) {
 }
 
 export const themeDoc = {
+  'accordion.heading.hover.color': {
+    description: 'The hover color used for the accordion.',
+    type: 'string | { dark: string, light: string }',
+    defaultValue: '{ dark: "light-4", light: "dark-3" }',
```
ShimiSun, line 22: The `defaultValue` here is contradicting the value shown on base.js, which is undefined.

src/js/themes/base.js
```diff
@@ -316,7 +316,12 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         side: 'bottom',
         color: 'border',
       },
-      heading: { level: '4' }, // level ranges from 1-6
+      heading: {
+        level: '4', // level ranges from 1-6
+        // hover: {
+        //   color: { dark: undefined, light: undefined },
+        // },
+      }, // level ranges from 1-6
```
ShimiSun, line 324:
```suggestion
      },
```

src/js/components/Accordion/stories/Custom.js
britt6612: With the current code `heading` is used under the hood for the accordion I was just thinking of a way to get rid of the hover since it is already built in and not defined in the theme. I can move to be accordion.hover. I think my original thought process was this hover effects the heading size in the accordion

src/js/components/AccordionPanel/AccordionPanel.js
britt6612: That was the original code. When the accordion is tabbed through the heading title still gets hover behavior I can look more behind this about best practices

britt6612: Im not sure how others would feel if we took out the hover on focus if they are already having it there and since the code has been that way. I did look at some other accordions and how they were being used as well as some accessibility with accordion https://www.w3.org/TR/wai-aria-practices/examples/accordion/accordion.html for instance and saw mostly just the focus not hover but again since this was here before Im open for either way.

src/js/components/Accordion/stories/Custom.js
```diff
@@ -7,7 +7,12 @@ import { Accordion, AccordionPanel, Box, Grommet, Text } from 'grommet';
 
 const CustomAccordionTheme = {
   accordion: {
-    heading: { level: '3' },
+    heading: {
+      level: 3,
+    },
+    hover: {
+      color: 'accent-2',
```
ShimiSun, line 14: I looked at the storybook example, and I don't see this hover color.

src/js/components/AccordionPanel/doc.js
```diff
@@ -16,6 +16,11 @@ export function doc(Panel) {
 }
 
 export const themeDoc = {
+  'accordion.heading.hover.color': {
```
ShimiSun, line 19: I think this should be fixed to the correct nesting of hover.

src/js/components/Accordion/stories/Custom.js
britt6612: Its just on the last one because the first two are not passing a `string` to heading so thats not getting the hover either way.  I just ran it again on mine and it shows if you can just check the last one ? 

Screen Shot 2020-04-28 at 4 31 04 PM

src/js/components/AccordionPanel/AccordionPanel.js
```diff
@@ -48,15 +48,15 @@ const AccordionPanel = forwardRef(
           aria-expanded={active}
           onClick={onPanelChange}
           onMouseOver={event => {
-            setHover(theme.dark ? 'light-4' : 'dark-3');
+            setHover(theme.accordion.hover && theme.accordion.hover.color);
```
ericsoderberghp, line 51: This will set `hover` to `false` if there is no value in the theme. I think we we should set it to undefined instead:
`setHover((theme.accordion.hover && theme.accordion.hover.color) || undefined);`

src/js/components/AccordionPanel/doc.js
```diff
@@ -16,6 +16,11 @@ export function doc(Panel) {
 }
 
 export const themeDoc = {
+  'accordion.hover.color': {
+    description: 'The hover color used for the accordion.',
```
ericsoderberghp, line 20: Just to be a bit more precise in describing this: "The hover color used for an accordion panel control."

PR #4006, opened 2020-04-23
Refined docs for RadioButtonGroup

ShimiSun:
Simple documentation changes I applied while going over the component functionalities.

### DIFF
--- a/src/js/components/RadioButtonGroup/doc.js
+++ b/src/js/components/RadioButtonGroup/doc.js
@@ -30,7 +30,7 @@ export const doc = RadioButtonGroup => {
     ).isRequired,
     onChange: PropTypes.func.description(
       `Function that will be called when the user clicks on one of the radio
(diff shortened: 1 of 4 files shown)


src/js/components/RadioButtonGroup/doc.js
```diff
@@ -29,8 +29,8 @@ export const doc = RadioButtonGroup => {
       elements.`,
     ).isRequired,
     onChange: PropTypes.func.description(
-      `Function that will be called when the user clicks on of the radio
-      buttons. It will be passed a React event object.`,
+      `Function that will be called when the user clicks on one of the radio
+      buttons. It will be passed as a React event object.`,
```
ericsoderberghp, line 33: The "one" change is right. The "as" change means "The function will be passed as a React event object.", which I don't think is right. I think the original wording for this sentence is better.

PR #4012, opened 2020-04-24
onEsc and onClickOutside does not work when modal set to false in layers :: Don't Merge Yet

IanKBovard:
#### What does this PR do?
Nothing yet. The goal is to address issue #3771 with either a code change or through examples.

LayerContainer.js. I have some coments.

None yet. Will test with storybook.

Same as above

onEsc functionality of layers with modal set to false was removed with this PR #3632. I assume it prevents the layer from taking keystrokes when it is used as a type of drop.

To fix this i can revert the changes made in that pr. The layer would still steal keystrokes, but maybe we can provide an example with drop(or any other component) that achieves the same thing that the pr was trying to address.

onClickOutside was never available for layers with modal set to false. This seems intentional to me. My suggestion would be to update the documentation(saying that outclickoutside no longer functions when modal is set to false), and adding an example that shows how to create your own onclickoutside handler.

#3771 

No

No

It will be when it is complete.

### DIFF
--- a/src/js/components/Layer/LayerContainer.js
+++ b/src/js/components/Layer/LayerContainer.js
@@ -63,6 +63,7 @@ const LayerContainer = forwardRef(
           }
           element = element.parentElement;
         }
+        // this is keeping onEsc from working with modal set to false
         if (modal && !element && anchorRef.current) {
           anchorRef.current.focus();
         }
@@ -131,6 +132,8 @@ const LayerContainer = forwardRef(
           tabIndex="-1"
           dir={theme.dir}
         >
+          {/* This handles to onclickoutside which is not present 
+          when modal is set to false */}
           <StyledOverlay
             plain={plain}
             onMouseDown={onClickOutside}


IanKBovard:
Need some input before I make the changes.

ShimiSun:
`interns` to add more tests to cover the accessibility aspect of the component.

IanKBovard:
I could be wrong, but this issue's main focus isn't the accessibility of the component, but how it's supposed to function. ie. should layer steal keystrokes when modal is set to false, should onClickOutside work when modal is set to false(as it was never implemented). 

I opened this PR because this issue needed clarification. I think the intern issue of adding tests should be a new issue, and separate PR.

ericsoderberghp:
One example of how interaction outside of a Layer should not affect the Layer is when Layers are used for notifications. Such as this one: https://storybook.grommet.io/?path=/story/layer--notification

I think the original design for modal Layers was that they become the focus of interaction. So, the user needs a variety of ways to remove the Layer that has taken over their interaction surface. The Layer should have a control in the Layer but there also needs to be a way to use the keyboard and users are used to "clicking outside" to dismiss the Layer.

I think the original design for non-modal Layers was that they were for things like notifications, asynchronous and non-disruptive to the current interaction. Closing those Layers was either automatic, such as removing after a period of time, or requiring the user to interact with a control in the Layer to dismiss it.

I think the issue behind this change is somewhere in between these two design directions. It would be great to see an example from a design standpoint illustrating the intent. But, I don't have any fundamental issues with refactoring LayerContainer to handle `onEsc` and `onClickOutside` for non-modal cases, since they are optional properties.

IanKBovard:
@ericsoderberghp sorry, just saw your comment while looking through old issues. 
Thanks, that does clear it up quite a bit for me. I appreciate it.

IanKBovard:
closing this in favor of #4452

PR #4018, opened 2020-04-28
Enhanced ability to style focus with theme

ericsoderberghp:
#### What does this PR do?

Enhanced ability to style focus with theme.

`theme.global.focus` now supports `outline` and `shadow` as keys, each with `color` and `size` keys.

`FormField` and `EdgeControl` no longer embed focus styling within them. Everywhere we use `focusStyle()` now.

`focusStyle` is now a function that is called with `focusStyle(true)` from FormField to indicate that the focus should only affect the border color if the theme has `global.focus.border`.

The most notable change in the snapshots is removing `outline: none;` as a base style. This allows the theme to control the focus behavior using outline as an option. Now, we only set `outline: none;` when a component has focus and doesn't want it to use an outline.

base.d.ts to see the theme structure
`focusStyle` in styles.js for the core changes

Added the anticipated HPE NEXT focus theme to the AllComponents story. This allows switching themes and then tabbing through the various components to observe the focus behavior. This story was extremely helpful for making these changes.

AllComponents story

yes, automatic from `themeDocUtils`

yes

mostly backwards compatible

The one visual difference is that the grommet theme no longer changes the border color of Button, TextInput, CheckBox, and TextArea. The Select component already behaved this way, so grommet was inconsistent. This unifies the behavior around how Select behaved.

### DIFF
--- a/src/js/utils/styles.js
+++ b/src/js/utils/styles.js
@@ -208,72 +208,68 @@ export const fillStyle = fillProp => {
   return undefined;
 };
 
-export const focusStyle = justBorder => css`
-  ${props =>
-    // focus also supports clickable elements inside svg
-    props.theme.global.focus &&
-    props.theme.global.focus.border &&
-    props.theme.global.focus.border.color &&
-    `> circle,
-    > ellipse,
-    > line,
-    > path,
-    > polygon,
-    > polyline,
-    > rect {
-      outline: ${normalizeColor(
-        props.theme.global.focus.border.color,
+const focusStyles = (props, { justBorder } = {}) => {
+  const {
+    theme: {
+      global: { focus },
+    },
+  } = props;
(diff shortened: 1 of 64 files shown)


src/js/utils/styles.js
```diff
@@ -208,27 +208,72 @@ export const fillStyle = fillProp => {
...
+    props.theme.global.focus.border.color &&
+    `> circle,
+    > ellipse,
+    > line,
+    > path,
+    > polygon,
+    > polyline,
+    > rect {
+      outline: ${normalizeColor(
+        props.theme.global.focus.border.color,
```
ShimiSun, line 225: Should use outline property from the theme.

ericsoderberghp: done

PR #4020, opened 2020-04-28
Accordion - add pad and margin options to theme

britt6612:
#### What does this PR do?
This PR add the option to change how the pad and margin sizes are being applied to the accordion panel heading
accordionPanel.js
storybook
storybook
When a label was passed into accordionPanel there was no say for the users on what size the margin or pad around it was. This was all done within the AccordionPanel.js starting at Line 70
There was really no way for the users to customize what they wanted for heading. 

The pad was being set to `horizontal: 'xsmall'`
and the margin was being taken from the `Heading line size` for level 4
design system
auto
no
backwards compatible


src/js/components/AccordionPanel/doc.js
```diff
@@ -21,6 +21,16 @@ export const themeDoc = {
     type: 'number',
     defaultValue: '4',
   },
+  'accordion.heading.margin': {
+    description: 'The margin around heading in the accordion.',
+    type: 'string',
+    defaultValue: 'undefined',
+  },
+  'accordion.heading.pad': {
```
ShimiSun, line 29: Semantically, the pad isn't assigned to heading, so I think it should be accordion.pad

src/js/components/AccordionPanel/doc.js
```diff
@@ -21,6 +21,16 @@ export const themeDoc = {
...
+    description: 'The pad size around heading in the accordion.',
+    type: 'sting',
+    defaultValue: 'horizontal: xsmall',
```
ShimiSun, line 32:
```suggestion
    defaultValue: '{ horizontal: xsmall }',
```

src/js/components/AccordionPanel/doc.js
britt6612: I saw that as it was on `box` Ill change thanks

ShimiSun: what about fixing it on the docs as well?

ShimiSun:
Can you please fix this PR? The Changed file doesn't make sense and I suspect its a merge issue.

PR #4021, opened 2020-04-28
add border around accordion to theme

britt6612:
#### What does this PR do?
This PR adds a option to put a border around the box containing accordion panels
accordion.js
storybook
```
    outer: {
      border: {
        side: 'horizontal',
        color: 'border',
      },
    },

```
I added this is custom theme but it doesnt look the prettiest so I can take out but wanted an example for you to look at 
The current border option is only relevant  to the box inside of the accordion panel there was no option to have a border on the outside if users wanted that there
design system
auto
no
backwards compatible


src/js/components/AccordionPanel/doc.js
```diff
@@ -37,13 +37,23 @@ export const themeDoc = {
...
     defaultValue: 'border',
   },
   'accordion.border.side': {
-    description: 'The border side to use in the accordion.',
+    description: 'The panel border side to use in the accordion panel.',
     type: 'string',
     defaultValue: 'bottom',
   },
+  'accordion.outer.border.color': {
+    description: 'The outer border color around to use in the accordion.',
```
ShimiSun, line 50: Not sure this description makes sense, it will be helpful to understand around when. otherwise not sure what is the difference between the outer and default border. 
If outer is only referring or associated with border it should be nested inside the border, i.e. `border.outer`. But I'm not sure I understand what `outer` means from this description.

britt6612: This border color is referring to the main `box` that holds the whole accordion panel within. The accordion border is referring to the border of the accordion panel so maybe I can change these two descriptions to be border of accordion panel vs border of accordion..

ShimiSun:
After a few discussions on the Accordion border, and mainly the `theme.accordion` structure, I think the current objects nested under theme.accordion should be all nested under `theme.accordion.panel`. 
The [requirements](https://www.figma.com/file/UvTBUQdhk07wV79nijsrtr/HPE-Accordion-Component?node-id=88%3A637) of this PR drives a need for a border on the actual accordion component rather than the panel, so IMO the best structure for base.js that would address those two types of border are:
```
    accordion: {
      border: {
        side: 'bottom',
        color: 'border',
      },
      panel: {
        border: {
          side: 'bottom',
          color: 'border',
        },
        heading: {
          level: '4', // level ranges from 1-6
        },
        hover: {
          color: { dark: 'light-4', light: 'dark-3' },
        },
        icons: {
          collapse: FormUp,
          expand: FormDown,
          // color: { dark: undefined, light: undefined },
        },
      },
    },
```

This change will require backward compatibilty adjustments.

PR #4023, opened 2020-04-29
Refactored backgroundStyle to split out backgroundAndTextColors

ericsoderberghp:
#### What does this PR do?

Refactored backgroundStyle to split out backgroundAndTextColors.
This will make it easier to share the colors between properties and styles.
The snapshot changes are primarily due to the new code only setting `background-color` instead of `background` when we are only providing a color value.

Storybook

Storybook

no

no

backwards compatible

### DIFF
--- a/src/js/components/Grommet/__tests__/Grommet-test.js
+++ b/src/js/components/Grommet/__tests__/Grommet-test.js
@@ -85,7 +85,7 @@ describe('Grommet', () => {
 
   test('background', () => {
     const component = renderer.create(
-      <Grommet full background="blue">
+      <Grommet full background="#0000ff">
         Grommet App
       </Grommet>,
     );
--- a/src/js/components/Video/__tests__/__snapshots__/Video-test.js.snap
+++ b/src/js/components/Video/__tests__/__snapshots__/Video-test.js.snap
@@ -158,7 +158,7 @@ exports[`Video autoPlay renders 1`] = `
 
 .c5:hover {
   background-color: rgba(221,221,221,0.4);
-  color: #000000;
+  color: #FFFFFF;
 }
 
 .c16 {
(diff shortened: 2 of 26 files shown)


src/js/components/Grommet/__tests__/Grommet-test.js
```diff
@@ -85,7 +85,7 @@ describe('Grommet', () => {
 
   test('background', () => {
     const component = renderer.create(
-      <Grommet full background="blue">
+      <Grommet full background="#0000ff">
```
ShimiSun, line 88: why is this needed? for easier comparison with the snapshot result?

ericsoderberghp:
> What was the drive for these changes?

It will help the Button changes so the Button can set the icon color to match the label color, but with those colors driven more by the theme.

src/js/components/Video/__tests__/__snapshots__/Video-test.js.snap
```diff
@@ -158,7 +158,7 @@ exports[`Video autoPlay renders 1`] = `
 
 .c5:hover {
   background-color: rgba(221,221,221,0.4);
-  color: #000000;
+  color: #FFFFFF;
```
ShimiSun, line 161: I'm guessing this is the desired behavior (switching the color from black to white), but wondering what would be the impact on the users.

src/js/components/Grommet/__tests__/Grommet-test.js
ericsoderberghp: We are now more accurate setting the text color based on the background color. For example, if you give a background color of "red", we don't know whether to choose light or dark text color, so we leave the text color unset in that case.

I've updated the stories and tests to use hex color values more often, so we can check the mapping more. But, there are still some cases where we leave CSS color names, so we can verify that we don't set a corresponding text color in those cases.

src/js/components/Video/__tests__/__snapshots__/Video-test.js.snap
ericsoderberghp: I verified interacting with Video that it's behaving correctly. And I couldn't find any other situations that seemed like a change wasn't correct.

PR #4028, opened 2020-04-29
add border and background option for formfield

britt6612:
#### What does this PR do?
This PR adds 2 options for background color and border color when in focus 
Formfield.js
storybook 

I tested with following in the custom story in form 
```

      focus: {
         background: {
           color:  'pink',
         },
         border: {
          color: 'purple,
         },
       },

```

storybook
`formfield` uses `Box` props so in order to be able to set a specific color for 
when `formfield` is in focus we need to add the option 
```
if(focus){
set border color 
}

```
I also went ahead and added background color as an option since it was also on disabled 

design system
auto
no
backwards compatible


ShimiSun:
Need to resolve conflicts

src/js/components/FormField/FormField.js
```diff
@@ -234,6 +234,11 @@ const FormField = forwardRef(
     } else if (normalizedError) {
       borderColor =
         (themeBorder && themeBorder.error.color) || 'status-critical';
+    } else if (focus) {
```
ericsoderberghp, line 237: This comment goes beyond just this change.

If the theme does have `formField.focus` but does not set `border` in it, we won't be using the default border color, line 243 in the new code. I would think I could set the border color in the theme for the default case and have that work for disabled, error, and focus cases if I don't wan't the border to change for them. Maybe I'm changing the background instead as an indicator? Or, maybe focus is reflected only via shadow, which is already taken care of.

src/js/components/FormField/FormField.js
```diff
@@ -309,6 +314,8 @@ const FormField = forwardRef(
     if (themeBorder.position === 'outer') {
       if (normalizedError && formFieldTheme.error) {
         outerBackground = formFieldTheme.error.background;
+      } else if (focus && formFieldTheme.focus) {
```
ericsoderberghp, line 317: Similar comment as with `border` above.

src/js/components/FormField/FormField.js
britt6612: This is a good point if formFieldTheme.focus was defined but the color. was not defined your correct it would not set border color to. line 243.. Instead Im wondering if within border color for focus we can have 

```

      borderColor =
        (formFieldTheme.focus &&
        formFieldTheme.focus.border &&
        formFieldTheme.focus.border.color) || (themeBorder && themeBorder.color) || 'border';

```

This would be a little similar to border color when there is an error the fall back is `status-critical` 
and then doing something similar with `disabled-borderColor`

src/js/components/FormField/FormField.js
britt6612: Currently for `FormField` background there is no default so if none of these checks exists then it would be undefined would you want `formfield` to have a default background prop to fall back on ?

ericsoderberghp: no

src/js/components/FormField/FormField.js
ericsoderberghp: You're on the right track, we just need to restructure the logic appropriately. My inclination is to do something like:
```
let borderColor;
if (focus && formFieldTheme.focus && formFieldTheme.focus.border.color && formFieldTheme.focus.border.color) borderColor = formFieldTheme.focus.border.color;
else if (...)
```

ericsoderberghp: When I get to publishing my Button pull request, I've got some patterns that might apply as well here. But, not quite ready yet.

src/js/components/FormField/FormField.js
britt6612: you just. want the same checks as above.. sorry I think I just overthought you comment

PR #4009, opened 2020-04-23
Align button theme structure with new convention

taysea:
#### What does this PR do?
Aligns button theming with convention of `button.active.primary.background.color`, etc.

The general structure is `component.state.prop.styleObject`.

#### Where should the reviewer start?
src/js/components/Button/StyledButton.js

#### What testing has been done on this PR?
All combinations added to the documentation were tested in storybook manually.

#### How should this be manually tested?
Checkout the Custom Button storybook example and alter any values there.

#### Do the grommet docs need to be updated?
Yes, they're updated here.

#### Should this PR be mentioned in the release notes?
No.

#### Is this change backwards compatible or is it a breaking change?
Yes, backwards compatible.

PR #4029, opened 2020-04-29
Add secondary prop to Button (and default to theme)

taysea:
#### What does this PR do?
This adds a new prop to Button `secondary` which can be styled in the theme. Additionally, a theme object of `button.default` allows users to create default styling that would apply to buttons. This will give more flexibility in theming because each button state (disabled, hover, and active) can be controlled broadly for all buttons or fine-tuned for each button type.

This PR folds in the changes that had been addressed on [this PR](https://github.com/grommet/grommet/pull/4009) since the styling was necessary for the completion of this PR. If this PR merges, we can close that one.

This will help achieve HPE Design System button designs.

https://github.com/grommet/grommet/pull/4029/files#diff-aa26cdee9d653531833f0f421d7ac1c6

A storybook has been created that shows how all of these new elements would be successfully implemented to achieve the various HPE Design System Buttons.

I will be adding tests to this PR as well.

Create a custom theme and create styles for button.secondary or button.default/

Yes, they're updated here.

Yes.

I am refining checks here, but wanted to put the PR up to give a chance for review.


ShimiSun:
For more context, attaching a slack thread:

>Here’s my current thinking on “default” Buttons:
Stick with the guidance of “default”, “secondary”, and “primary” as the three flavors of Button.
*default* - the majority of Buttons on the page. Their affordance is primarily driven by their context, such as in Tabs, in Sidebar+Nav, a toolbar, the corner of a tile or card, etc.
*primary* - the primary call to action. There should typically be at most one primary Button on the screen. In a Form context, it signifies the action bound to the enter key that will submit the form.
*secondary* - When there isn’t enough context to indicate its affordance, or when it should call extra attention to itself but isn’t the primary call to action, use a secondary Button.
If the theme has `button.default`, <Button /> will use that.
If the theme doesn’t have `button.default`, <Button /> will behave as it does today.
Any HPE code using the NEXT theme will get the new default behavior for <Button />.
Any HPE code using the v1.0 hpe theme will get the current outline behavior for <Button />.
Anyone wanting the new NEXT styling while still using the v1.0 hpe theme will have to customize the Button themselves, likely via a plain Button with children or via their own local theme extension.
The grommet Sidebar examples can use a custom theme with `button.default` to augment the `grommet`

taysea:
Closing because this behavior is handled by https://github.com/grommet/grommet/pull/4050/files

PR #4031, opened 2020-04-30
Allow menu icon color to be controlled by the theme

taysea:
#### What does this PR do?
This PR allows users to style the icon color of a Menu in the theme using `menu.icons.color`. Currently, only `global.colors.control` has affect on the menu icon and so putting a custom color on the icon requires complex process like [this example](https://storybook.grommet.io/?path=/story/menu--custom) which isn't beneficial for efficiency/scalability across an application.

This would allow the HPE Design System, for example, to specify `menu.icons.color` as `text` for example.

NOTE: the theme-test.js snap changes are coming from master, not due to changes in this PR.

This was tested in storybook, and a jest test has been added that applies a custom color in the theme.

Create a custom theme and define a color for `menu.icons.color`

Currently, the theme can't style menu icon color aside from changing the value of `control`.

Yes, they are updated here.

Maybe.


src/js/components/Menu/doc.js
```diff
@@ -103,10 +103,10 @@ The object values can be any Button prop, for example: label and onClick.`,
 };
 
 export const themeDoc = {
-  'global.colors.control': {
+  'menu.icons.color': {
```
ShimiSun, line 106: I'm thinking we should still keep 'global.colors.control' with the actual base.js default value. to be more explicit.

taysea: > Looks good, one tiny comment.
> we might want to solve the test issues on a separate PR.

I believe this PR is aiming to solve the test issues: https://github.com/grommet/grommet/pull/4030

PR #4032, opened 2020-04-30
Updated dependencies and changed snapshots

ShimiSun:
Updating dependencies of yarn.lock requires some snapshot changes.
Closes #4030.
Should unblock #4031.

The snapshot changes are occurring on the theme-test for the hpe theme testing, we didn't change anything on the hpe theme (^1.0.0) recently, I'm wondering if the recent #4023 PR changes are somehow related to this snapshot update.

### DIFF
--- a/src/js/themes/__tests__/__snapshots__/theme-test.js.snap
+++ b/src/js/themes/__tests__/__snapshots__/theme-test.js.snap
@@ -1579,7 +1579,7 @@ exports[`Grommet hpe theme 1`] = `
   box-sizing: border-box;
   max-width: 100%;
   background-color: #01A982;
-  color: #EEEEEE;
+  color: #FFFFFF;
   min-width: 0;
   min-height: 0;
   -webkit-flex-direction: column;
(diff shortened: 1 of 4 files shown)


src/js/themes/__tests__/__snapshots__/theme-test.js.snap
```diff
@@ -1658,7 +1658,7 @@ exports[`Grommet hpe theme 1`] = `
   display: flex;
   box-sizing: border-box;
   max-width: 100%;
-  background-color: #F8F8F8;
+  background-color: #00E8CF;
```
ericsoderberghp, line 1661: seem right? Related to updates to the HPE NEXT theme?

ShimiSun: I don't think so, the test is using hpe v1.0.0 and not the NEXT theme, so I'm not sure why this change occurs. we didn't have hpe-theme changes or releases in the last few weeks.

ShimiSun:
This PR is needed to update the integrity check for `yarn.lock`.
I'll be merging it to unblock pending PRs, but there is a need to understand the root cause of these changes.

I ran a local test reverting back to branch `feat/all-story-themes` which is when we introduced the NEXT theme and probably changed the integrity check, and I ran `yarn test-update` to see if the snapshot had changed, and yes, even back then I had the same snapshot changes as this PR suggests, so I'm guessing that it is eliminating my suspicion that the changes on  [backgroundStyle PR](https://github.com/grommet/grommet/pull/4023) are related to this theme snapshot changes.

PR #4033, opened 2020-04-30
New CheckBoxGroup component to core library

ShimiSun:
#### What does this PR do?
Added CheckBoxGroup to core library.
I was thriving for a similar behavior between CheckBoxGroup and Select multiple (with the reduced valueKey), mostly in the datastructure and Form management area.
The component is compliant with the WCAG 2.0 standards of accessibility, verified with https://webaim.org/techniques/forms/controls#checkbox
https://www.w3.org/WAI/tutorials/forms/grouping/#checkboxes

Few disclaimers on the current behavior I wanted to call out:
1.  I haven't demonstrated or tested the component with `toggle` or `indeterminate` CheckBox, I wasn't sure if it will be something we want to recommend using.
1. Snapshot testes had been done. Once the approach is approved I can add more behavioral testing as well.
1. Unlike Select, `valueKey` isn't accepting a key+reduce object (CheckBoxGroup valueKey is using `reduce: true` by default), and it is not accepting a function. `labelKey` is also not accepting a function. we can add it later as the need arises. I haven't seen it implemented much in grommet stories or tests for Select, and I wanted to reduce the level of complexity.
1. The component isn't supporting children, I haven't seen a use case for it and we can add it later if we'll have requests coming.
 If you think that any of the disclaimers should be taken care of as part of this PR, I'll get on it.

doc.js to see the new props and then CheckBoxGroup.js 
New test file and multiple stories added to CheckBoxGroup and edited on Form.
storybook Form + CheckBoxGroup.
The need for the component was raised on the design-system work.
#3489 
https://github.com/hpe-design/design-system/issues/743

Screen Shot 2020-04-29 at 4 18 43 PM

Done.
For sure, bump up minor version.
backwards compatible


src/js/components/CheckBoxGroup/doc.js
```diff
@@ -0,0 +1,53 @@
...
+    .availableAt(getAvailableAtBadge('CheckBoxGroup'))
+    .description('A group of CheckBoxes.')
+    .usage(
+      `import { CheckBoxGroup } from 'grommet';
+        <CheckBoxGroup />`,
+    )
+    .intrinsicElement('div');
+
+  DocumentedCheckBoxGroup.propTypes = {
+    checked: PropTypes.arrayOf(PropTypes.string).description(
```
ericsoderberghp, line 16: Might want to allow `PropTypes.oneOfType([PropTypes.string, PropTypes.number])` in case the valueKey maps to a number.

src/js/components/CheckBoxGroup/doc.js
```diff
@@ -0,0 +1,53 @@
...
+      .defaultValue(undefined),
+    labelKey: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.func,
+    ]).description(
+      `When the options array contains objects, this property indicates how
+        to determine the label of each option. If a string is
+        provided, it is used as the key to retrieve each option's label.`,
+    ),
+    name: PropTypes.string.description(
```
ericsoderberghp, line 31: How does this work when each CheckBox should have a `name`? I don't think `<div />` elements can have `name`, just input-like things. I think what I mean to suggest is that `name` is used for the Form context but not as a DOM attribute.

src/js/components/CheckBoxGroup/doc.js
```diff
@@ -0,0 +1,53 @@
...
+    ),
+    onChange: PropTypes.func.description(
+      `Function that will be called when the user clicks on a CheckBox option. 
+      It will be passed a React event object with a value.`,
+    ),
+    options: PropTypes.oneOfType([
+      PropTypes.arrayOf(PropTypes.string),
+      PropTypes.arrayOf(PropTypes.shape({})),
+    ]).description(
+      `Options can be either a string or an object of CheckBox props.`,
```
ericsoderberghp, line 43: Is it CheckBox props or an arbitrary data object? For Select, it is an arbitrary data object. We do not pass that object as props to the underlying options.

src/js/components/CheckBoxGroup/index.d.ts
```diff
@@ -0,0 +1,18 @@
+import * as React from "react";
+import { BoxProps } from '../Box';
+import { CheckBoxProps } from '../CheckBox' ;
+
+
+export interface CheckBoxGroupProps {
+  checked?: string | object;
```
ericsoderberghp, line 7: `checked?: string[] | number[];`?

ShimiSun: yes, `object` is definitely not the right type.

src/js/components/Form/stories/AllChildren.js
```diff
@@ -40,14 +40,21 @@ const Example = () => (
...
+            />
           </FormField>
           <FormField name="ampm">
             <RadioButtonGroup name="ampm" options={['morning', 'evening']} />
           </FormField>
           <FormField label="Size" name="size">
-            <Select name="size" options={['small', 'medium', 'large']} />
+            <Select
+              name="size"
+              multiple
```
ShimiSun, line 55: This was just for the sake of comparing the data structures of Select multiple with CheckBoxGroup, I'm ok removing `multiple` from this example.

src/js/components/CheckBoxGroup/doc.js
ShimiSun: I agree, `name` isn't an input element.

ShimiSun: Overall I agree. 
Though, the `name` isn't being assigned on the `div` level but on the individual `input` (Checkbox) level. why would that be a problem?

src/js/components/CheckBoxGroup/doc.js
ShimiSun: It made sense to me to make `options` an object of CheckBox props, this way the user can leverage the propagation of props on an individual CheckBox via `options` objects. (See CheckBoxGroup.js line 81)

RadioButtonGroup is performing this type of propagation in a similar fashion for its nested RadioButton(s).
In the case of CheckBoxGroup `options`, I see more potential similarity with RadioButtonGroup rather than with Select, but I might be missing something.

src/js/components/CheckBoxGroup/doc.js
ericsoderberghp: I think it will be a problem. A good way to think about it is how would `<form action="" />` behave if the default browser behavior submitted it, via a POST. The form data would pick out the contained values keyed by "name" attribute of the input elements.

src/js/components/CheckBoxGroup/doc.js
ShimiSun: Fixed.

src/js/components/CheckBoxGroup/index.d.ts
ShimiSun: Fixed.

src/js/components/CheckBoxGroup/doc.js
ShimiSun: Good observation. 
I'll clean up the documentation and code with the direction of this comment (I'll remove the actual assignment of `name` from CheckBox and only use `name` when calling the Form context).
The user will still be available to assign the `name` prop per each CheckBox via the `options` object.

src/js/components/CheckBoxGroup/doc.js
ericsoderberghp: So, for RadioButtonGroup, the options are more focused on the underlying RadioButton. Whereas for Select, the options are more focused on the data list of choices.
For example, if I make a REST call to GET /deployment-targets, which returns a JSON array of objects, I can feed that directly to select and give it labelKey and valueKey appropriate to pick the fields of interest. I cannot do this with CheckBoxGroup or RadioButtonGroup. I would have to map() the result to convert it into something appropriate for the UI presentation.
The UI presentation is less important for Select, since there is no input component that each option maps to. For Select, the children render function is where any translation to UI presentation can happen.
Just thinking about these two approaches and thought I write these notes here for memory.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -0,0 +1,99 @@
...
+            : option;
+        }),
+      [optionsProp, disabledProp],
+    );
+
+    // 'checked' is an array of checked valueKeys
+    const [checked, setChecked] = formContext.useFormContext(
+      name,
+      checkedProp,
+      '',
```
ericsoderberghp, line 43: I don't think the initial value would be `''`. I think it can be omitted, or maybe `[]`.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -0,0 +1,99 @@
...
+    // Logic is necessary to maintain a proper data structure for Form logic
+    const onCheckBoxChange = (event, optionChecked) => {
+      const nextChecked = checked || [];
+      const optionIndex = nextChecked.indexOf(optionChecked);
+
+      if (optionIndex < 0) nextChecked.push(optionChecked);
+      else nextChecked.splice(optionIndex, 1);
+
+      setChecked(nextChecked);
+      if (onChange) onChange(event);
```
ericsoderberghp, line 55: We want to pass the same kind of thing we'll receive in the `checked` property as part of the event, so the caller can save this in his state and pass it back as `checked`.  Take a look at this from Select.js:
```
if (onChange) {
  event.persist();
  const adjustedEvent = event;
  adjustedEvent.target = inputRef.current;
  adjustedEvent.value = nextValue;
  adjustedEvent.option = option;
  adjustedEvent.selected = nextSelected;
  onChange(adjustedEvent);
}
```

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -0,0 +1,99 @@
...
+    return (
+      <Box ref={ref} gap={gap} {...rest}>
+        {options.map(
+          ({
+            label: labelOption,
+            disabled,
+            checked: checkedOption,
+            ...optionRest
+          }) => {
+            const label = optionRest[labelKey] || labelOption;
```
ericsoderberghp, line 67: This won't work if `labelKey` is "label" because "label" isn't in `optionRest`.

ericsoderberghp: I'm thinking, since we are saying that the options are tied to CheckBox, that we remove `labelKey` and `valueKey` and rely on the caller supplying `label` and `value` keys, like RadioButtonGroup.

ShimiSun: Fixed.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
ShimiSun: Done.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
ShimiSun: It worked with `[]` better than omitting.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -0,0 +1,99 @@
...
+          const checkedOption =
+            option.checked || (checkedProp && checkedProp.indexOf(value) >= 0);
+          return (
+            <CheckBox
+              checked={checkedOption}
+              disabled={disabledProp || option.disabled}
+              label={label}
+              key={label}
+              onChange={event => onCheckBoxChange(event, value, option)}
+              {...option}
```
ericsoderberghp, line 81: if `option.label` exists, it will override `label` set on line 78. But, I think if `labelKey` is provided, it should take precedence.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -0,0 +1,99 @@
...
+      }
+    };
+
+    return (
+      <Box ref={ref} gap={gap} {...rest}>
+        {options.map(option => {
+          const label = labelKey ? option[labelKey] : option.label;
+          const value = valueKey ? option[valueKey] : option.id;
+          const checkedOption =
+            option.checked || (checkedProp && checkedProp.indexOf(value) >= 0);
```
ericsoderberghp, line 73: So, if `checkedProp` doesn't have `value` but `option.checked` is true, we will call it `checked`? This doesn't seem right, as it won't align with what the form would submit because `checked` is initialized from `checkedProp`.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -0,0 +1,99 @@
...
+
+    // 'checked' is an array of checked valueKeys
+    const [checked, setChecked] = formContext.useFormContext(
+      name,
+      checkedProp,
+      [],
+    );
+
+    // Logic is necessary to maintain a proper data structure for Form logic
+    const onCheckBoxChange = (event, optionChecked, option) => {
```
ericsoderberghp, line 47: `optionChecked` -> `optionValue` might be clearer as I tend to think of "checked" as a boolean.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -0,0 +1,99 @@
+import React, { forwardRef, useContext, useMemo } from 'react';
+
+import { Box } from '../Box';
+import { CheckBox } from '../CheckBox';
+import { FormContext } from '../Form/FormContext';
+
+export const CheckBoxGroup = forwardRef(
+  (
+    {
+      checked: checkedProp,
```
ericsoderberghp, line 10: Now that I'm reading more detail here, I think this property should be `value`, to align more with Select and how the Form `value` will work.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
ShimiSun: I removed the individual CheckBox `checked` functionality and documented it as part of `options`.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
ShimiSun: Done.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
ShimiSun: "CheckBoxGroup just joined the value party"

src/js/components/CheckBoxGroup/CheckBoxGroup.js
ShimiSun: Done.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -0,0 +1,111 @@
...
+    const formContext = useContext(FormContext);
+
+    // In case option is a string, normalize it to be an object
+    const options = useMemo(
+      () =>
+        optionsProp.map(option => {
+          return typeof option === 'string'
+            ? {
+                disabled: disabledProp,
+                key: option,
```
ericsoderberghp, line 31: `value`? Seems odd to have a key named "key"

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -0,0 +1,111 @@
...
+
+    // 'value' is an array of checked valueKeys
+    const [value, setValue] = formContext.useFormContext(name, valueProp, []);
+
+    // Logic is necessary to maintain a proper data structure for Form logic
+    const onCheckBoxChange = (event, optionValue, option) => {
+      // deep copy of value
+      const nextValue = JSON.parse(JSON.stringify(value)) || [];
+      const optionIndex = nextValue.indexOf(optionValue);
+      // If the index of the value option isn't in the array add the option.
```
ericsoderberghp, line 47:
```suggestion
      // If the value option isn't in the array, add it.
```

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -0,0 +1,111 @@
...
+      // Otherwise, remove the option from the array to simulate a toggle action
```
ericsoderberghp, line 48:
```suggestion
      // Otherwise, remove it.
```

ericsoderberghp: We are actually toggling the value, not just "simulating" it.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -0,0 +1,111 @@
...
+      if (optionIndex < 0) nextValue.push(optionValue);
+      else nextValue.splice(optionIndex, 1);
+      setValue(nextValue);
+      // Same functionalities as Select onChange()
```
ericsoderberghp, line 52:
```suggestion
      // Similar functionality to Select onChange()
```

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -0,0 +1,111 @@
...
+      <Box ref={ref} gap={gap} {...rest}>
+        {options.map(
+          ({
+            checked: checkedOption,
+            disabled: disabledOption,
+            label: labelOption,
+            ...optionRest
+          }) => {
+            const label = labelKey
+              ? optionRest[labelKey] || labelOption
```
ericsoderberghp, line 72: If, for whatever reason, `optionRest[labelKey]` is 0, this will incorrectly set `label` to `labelOption`. Instead, I think we should just do something like:
```
options.map(option => {
  const label = labelKey ? option[labelKey] : option.label;
  const value = valueKey ? option[valueKey] : option.value;
  const checked = valueProp && valueProp.indexOf(valueOption) >= 0;
  ...
  return <CheckBox {...option} label={label} checked={checked} />;
```

src/js/components/CheckBoxGroup/doc.js
```diff
@@ -0,0 +1,59 @@
...
+        <CheckBoxGroup />`,
+    )
+    .intrinsicElement('div');
+
+  DocumentedCheckBoxGroup.propTypes = {
+    value: PropTypes.arrayOf(
+      PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+    ).description(
+      `An array of the values for the checked options. 
+      If options is provided as an object, the value array will be the strings
```
ericsoderberghp, line 20:
```suggestion
      If options is provided as an object, the value array will be the values 
```

src/js/components/CheckBoxGroup/doc.js
```diff
@@ -0,0 +1,59 @@
...
+    )
+    .intrinsicElement('div');
+
+  DocumentedCheckBoxGroup.propTypes = {
+    value: PropTypes.arrayOf(
+      PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+    ).description(
+      `An array of the values for the checked options. 
+      If options is provided as an object, the value array will be the strings 
+      of the valueKeys.`,
```
ericsoderberghp, line 21:
```suggestion
     that the valueKey maps to.`,
```

src/js/components/CheckBoxGroup/doc.js
```diff
@@ -0,0 +1,59 @@
...
+    ),
+    disabled: PropTypes.bool
+      .description(`Disables all options.`)
+      .defaultValue(undefined),
+    labelKey: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.func,
```
ericsoderberghp, line 28: I don't see the function version of `labelKey` in the code.

src/js/components/CheckBoxGroup/index.d.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export type CheckBoxType = Omit<CheckBoxProps & JSX.IntrinsicElements['input'], 'checked'>;
+export interface CheckBoxGroupProps {
+  value?: (number | string )[]
+  disabled?: boolean;
+  labelKey?: string | ((...args: any[]) => any);  
+  name?: string;
+  onChange?: ((event?: OnChangeEvent) => void);
+  options: (CheckBoxType)[];
+  valueKey?: string | ((...args: any[]) => any);
```
ericsoderberghp, line 19: The function flavor isn't in the code and doesn't match doc.js.

src/js/components/CheckBoxGroup/stories/ControlledInput.js
```diff
@@ -0,0 +1,38 @@
...
+            // whether any of the checkboxes had changed.
+            console.log('Submit', values, touched)
+          }
+        >
+          <FormField name="controlled">
+            <CheckBoxGroup
+              id="check-box-group-id"
+              name="controlled"
+              value={value}
+              onChange={() => setValue()}
```
ericsoderberghp, line 25:
```suggestion
              onChange={({ value: nextValue }) => setValue(nextValue)}
```

src/js/components/CheckBoxGroup/CheckBoxGroup.js
ShimiSun:
```
 ...
  return <CheckBox {...option} label={label} checked={checked} />;

Comment (ShimiSun) [reply src/js/components/CheckBoxGroup/CheckBoxGroup.js]: Done.

File: src/js/components/CheckBoxGroup/CheckBoxGroup.js [hunk 23]
```diff
@@ -0,0 +1,105 @@
...
+        onChange(adjustedEvent);
+      }
+    };
+
+    return (
+      <Box ref={ref} gap={gap} {...rest}>
+        {options.map(option => {
+          const label = labelKey ? option[labelKey] : option.label;
+          const valueOption = valueKey ? option[valueKey] : option.value;
+          const checked = valueProp && valueProp.indexOf(valueOption) >= 0;
```
ericsoderberghp, line 67:
```suggestion
          const checked = value && value.indexOf(valueOption) >= 0;
```

PR #4034, opened 2020-05-01
add margin to accordion

britt6612:
#### What does this PR do?
This PR add the option to change how the margin sizes are being applied to the accordion panel heading
accordionPanel.js
storybook
storybook
When a label was passed into accordionPanel there was no say for the users on what size the margin around it was. This was all done within the AccordionPanel.js starting at Line 70
There was really no way for the users to customize what they wanted for heading. 

The margin was being taken from the `Heading line size` for level 4
design system
auto
no
backwards compatible

### DIFF
--- a/src/js/components/Accordion/stories/Custom.js
+++ b/src/js/components/Accordion/stories/Custom.js
@@ -7,8 +7,10 @@ import { Accordion, AccordionPanel, Box, Grommet, Text } from 'grommet';
 
 const CustomAccordionTheme = {
   accordion: {
+    pad: '6px',
     heading: {
       level: 3,
+      margin: '12px',
     },
     hover: {
       color: 'accent-2',
--- a/src/js/components/Accordion/stories/Custom.js
+++ b/src/js/components/Accordion/stories/Custom.js
@@ -9,7 +9,7 @@ const CustomAccordionTheme = {
   accordion: {
     heading: {
       level: 3,
-      margin: { left: '6px', right: '6px', top: '24px', bottom: '24px' },
+      margin: { vertical: '6px', horizontal: '24px' },
     },
     hover: {
       color: 'accent-2',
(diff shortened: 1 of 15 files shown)


britt6612:
@ShimiSun moved to a new branch sorry last branch was a merge mess!

ShimiSun:
>still not sure where the snapshot changes are coming from in the theme I did see your PR though addressing this

No worries, we'll be taken a look at those on a separate PR. I'll ignore those files while reviewing.

src/js/components/AccordionPanel/doc.js
```diff
@@ -41,6 +46,11 @@ export const themeDoc = {
     type: 'React.Element',
     defaultValue: '<FormDown />',
   },
+  'accordion.pad': {
+    description: 'The pad size around the accordion.',
```
ShimiSun, line 50: The description here isn't accurate, I ran some tests and it looks like the placement of the pad is only wrapping the heading and not the pad of the Accordion itself.
From this description, I would expect the accordion to have a pad, but the pad only applies on the wrapper of the heading and not even on a Panel.

Was that the intention of this PR? I would think the design-system would want the pad to apply on the all Panel.

See an example I created:
image

britt6612: yes this is around the heading. It is set at 'horizontal: xsmall' which we are not able to change in theme and design system the pad around header is a little big bigger.

ShimiSun: Cleaned the pad out, resolving the conversation.

src/js/components/Accordion/stories/Custom.js
```diff
@@ -9,6 +9,7 @@ const CustomAccordionTheme = {
   accordion: {
     heading: {
       level: 3,
+      margin: { left: '6px', right: '6px', top: '24px', bottom: '24px' },
```
ShimiSun, line 12:
```suggestion
      margin: { vertical: '6px', horizontal: '24px' },
```

PR #4036, opened 2020-05-01
Fixed Tabs to be more responsive

### LINKED ISSUE #4025: margin aren’t scaling for Tabs as they do for Box.
I have been testing on Small and large screen, apparently on small screens margin='large' doesn't value the same on Tabs than on Box.

### Expected Behavior

I think the expected behaviour is that the value of margin="large" is the same on Tabs than on Box if both are at the same screen size/page.

### Actual Behavior

On small screens, comparing Tabs and Box components with margin "large" and being on the same page, box margin is 24px and Tabs margin is 48px.

https://codesandbox.io/s/grommet-v2-template-ec7on?file=/index.js

### Steps to Reproduce

Not sure if there is a specific scenario, but as simple as the codesandbox you can test. Also not sure if other components does the same or not.

### Your Environment

- Grommet version: 2.13.0
- Operating System and version (desktop or mobile): Mac

IanKBovard:
#### What does this PR do?
closes #4025 

Adds responsive prop to tabs.js

Tabs.js

Storybook examples, and the example provided in issue #4025 

Same as above

#4025 

No

No

yes


ShimiSun:
After a quick discussion, this approach isn't backward compatible, we will need to find the reason the responsive values are being overridden.

IanKBovard:
Updated. Responsive is now passed through tabs. Backwards compatible

src/js/components/Tabs/Tabs.js
```diff
@@ -11,7 +11,7 @@ import { defaultProps } from '../../default-props';
 
 import { Box } from '../Box';
 
-import { StyledTabPanel, StyledTabs, StyledTabsHeader } from './StyledTabs';
+import { StyledTabs, StyledTabPanel, StyledTabsHeader } from './StyledTabs';
```
ShimiSun, line 14: Why do we need this change? it is not alphabetical this way.

Issue #4024, opened 2020-04-29
Incorrent TS definition for Form onChange


nozzle-a:
### Expected Behavior

According to [the documentation](https://v2.grommet.io/form), The `Form` component `onChange` handler should implement an interface like:
```
(nextState: FormState) => void;
```
Where `FormState` is whatever I'm passing to `React.useState`.

### Actual Behavior

However, the TS definitions defined for `Form` have a different interface:
```
((event: React.ChangeEvent) => void);
```

In `Form.js` line 141 you can see that the documentation is correct, `onChange` is passed the next form state object, not a `React.ChangeEvent`.
```
      if (onChange) onChange(nextValue);
```

If I have strict settings on my compiler which disallow `any`, I can't compile due to this issue.

### URL, screen shot, or Codepen exhibiting the issue

This compiles
```
  function onChange(e: ChangeEvent): void {
    console.log('nextState', e);
  }
// elsewhere
      <Form onChange={onChange}>
```
But `e` is not a `ChangeEvent` object, it is a `FormState` object (see below)

This does not compile:
```
interface FormState {
  email?: string;
  password?: string;
  username?: string;
}

  function onChange(nextState: FormState): void {
    console.log('nextState', nextState);
  }
// elsewhere
      <Form onChange={onChange}>
```

It produces the error:
```
Type '(nextState: FormState) => void' is not assignable to type '((event: ChangeEvent<Element>) => void) & ((event: FormEvent<HTMLFormElement>) => void)'.
  Type '(nextState: FormState) => void' is not assignable to type '(event: ChangeEvent<Element>) => void'.
    Types of parameters 'nextState' and 'event' are incompatible.
      Type 'ChangeEvent<Element>' has no properties in common with type 'FormState'.
```

### Your Environment

- Grommet version: 2.12.0
- Typescript version: 3.8.3

ShimiSun:
Thanks for reporting, feel free to submit a fix for the issue, looks like you've already done some investigation here https://github.com/grommet/grommet/blob/8912aefbc78b041d4670e68b4ed2d346aedbef9f/src/js/components/Form/index.d.ts#L7

nozzle-a:
I've [opened a PR](https://github.com/grommet/grommet/pull/4037) for this.  Anything I need to do to get it reviewed and merged?

ShimiSun:
Thank you for the PR @nozzle-a , I dropped a couple of comments.

PR #4037, opened 2020-05-01
Fix TS declaration for <Form>

nozzle-a:
#### What does this PR do?

Fix TS declaration for `Form.onChange`.

Originally, in `src/js/components/Form/index.d.ts` the `onChange` signature incorrectly expected an `Event` object, rather than a state object.

I've added a new type definition, which allows anyone using Typescript to either use the `Form` export which implicitly sets the `onChange` payload to `any`, or else they can use the `TypedForm` type to explicitly set the `onChange` payload type, which will be preferred by developers who use stricter TS settings.

This still works -- 
```typescript
import { Form } from 'grommet';

function onChange(nextState: MyState): void {
  // ...
}

<Form onChange={onChange} />
```

However, if you want strict type checking on your `onChange` function, you can also now do this --
```typescript
import { Form, TypedForm} from 'grommet';

interface MyState {
  name: string;
  age: number;
}

const MyForm = Form as TypeForm<MyState>;

// changing the function signature will provoke a TS error
function onChange(nextState: MyState): void {
  // ...
}

<MyForm onChange={onChange} />
```

I `link`ed this locally and ran the TS compiler.

Same as above.

https://github.com/grommet/grommet/issues/4024

I think the ability to strongly type the expected `onChange` parameter would be important to document, but I'm not sure where TS specific documentation should go.

Backwards compatible.


ShimiSun:
@negreirosleo & @dallas can you help with the review on this PR.

src/js/components/Form/index.d.ts
```diff
@@ -1,16 +1,17 @@
 import * as React from "react";
 
-export interface FormProps {
+export interface FormProps<T> {
```
ShimiSun, line 3: Why do you think the <T> is essential for this solution?

src/js/components/Form/index.d.ts
```diff
@@ -1,16 +1,17 @@
...
   errors?: {};
   infos?: {};
   messages?: {invalid?: string,required?: string};
-  onChange?: ((event: React.ChangeEvent) => void);
+  onChange?: (nextState: T) => void;
```
ShimiSun, line 7: I'm wondering if this should be `value` instead of `nextState`.

src/js/components/Form/index.d.ts
nozzle-a: Being able to type your input makes it possible to type check your handler function -- consider:

```typescript
interface MyFormState {
  email: string;
}

function onChange(value: OtherFormState): void {
  const name = value.name; // no error here
}

// elsewhere
<Form onChange={onChange}>
  <FormField name="email" label="Email">
        <TextInput name="email" />
      </FormField>
</Form>
```
Without the generic, this code will not raise an error, however the `onChange` function is expecting the wrong shape.

Ideally we would also add type definitions such that if you tried to use the wrong `name` parameter in `FormField` or `TextInput`, this would also raise an error, but that is beyond the scope of this change, IMO.

src/js/components/Form/index.d.ts
nozzle-a: Agreed, I'll update.

ShimiSun:
I would like to test this example with the typescript storybook example.

nozzle-a:
> I would like to test this example with the typescript storybook example.

Is that something I can help with?

ShimiSun:
>Is that something I can help with?

Yes, thanks for asking, I tried to do it myself but it's difficult to commit to your branch when you are working on your master branch.

Please add an `onChange` verification function to src/js/components/Form/stories/typescript/Submit.tsx
Something like 
```
onChange={({ value }) => console.log('Submit', value)}
```
Just to verify this was failing in the past but no longer failing with your fix, and feel free to commit this change on your PR.

nozzle-a:
@ShimiSun I've added that to the story -- I've also added the mechanism for adding knowledge of your form state shape to the `onChange` function.

src/js/components/Form/index.d.ts
```diff
@@ -1,16 +1,17 @@
 import * as React from "react";
 
-export interface FormProps {
+export interface FormProps<T> {
   errors?: {};
   infos?: {};
   messages?: {invalid?: string,required?: string};
-  onChange?: ((event: React.ChangeEvent) => void);
+  onChange?: (value: T) => void;
```
ShimiSun, line 7: @ericsoderberghp , I think this PR is good to go, just wanted you to give your inputs on this change and let us know if there are any concerns.

ericsoderberghp: no concerns from me

ShimiSun:

PR #4042, opened 2020-05-05
Changed Accordion hover, border and focus

ShimiSun:
#### What does this PR do?
1. Set consistent focus behavior
2. Allow border on the Panel Box (vs the existing Panel Content)
3. Changed `hover.color` to `hover.heading.color` for accuracy.
4. Clean-up Typescript examples of the accordion.

doc.js to get familiar with the upcoming changes

added a test yo unit testing, added a Themed story with hpe content, and tested across all storybook examples.
Go over these storybook examples and compare them with storybook master branch deployment.
Compare with https://design-system.hpe.design/components/accordion
Closes #4021 , related to https://github.com/hpe-design/design-system/issues/512
Done.
yes
backwards compatible


src/js/components/AccordionPanel/AccordionPanel.js
```diff
@@ -101,7 +126,7 @@ const AccordionPanel = forwardRef(
             </Box>
           )}
         </Button>
-        <Box border={theme.accordion.border}>
+        <Box border={contentBorder}>
```
ShimiSun, line 129: Unfortunately, after applying the following code as discussed (in order to replace the Box in line 129):
```
    const { border: contentBorder } = theme.accordion;
    const content = active ? (
      <Box border={contentBorder}> {children} </Box>
    ) : (
      undefined
    );
    ...
        {animate ? (
          <Collapsible open={active}>{content}</Collapsible>
        ) : (
          { wrapper: content }
        )}
```
I've encountered a backward compatibility issue due to the default bottom border this is always being rendered whether or not children, animate or active exist.
The previous behavior was enforcing the render of the Box due to the out of the box border it provides, so it is not possible not to display it. we may want to think on a more sophisticated way to work around the current behavior, we can apply it on a separate PR.

src/js/components/Markdown/README.md
```diff
@@ -34,4 +30,4 @@ You can see all the options [here](https://github.com/probablyup/markdown-to-jsx
 
 ```
 div
-```
+```
```
ShimiSun, line 33: This file is fixed on https://github.com/grommet/grommet/pull/4043,  I'll merge with master once PR #4043 will be merged.

src/js/components/Accordion/__tests__/__snapshots__/Accordion-test.js.snap
```diff
@@ -224,6 +224,11 @@ exports[`Accordion AccordionPanel 1`] = `
         onMouseOut={[Function]}
         onMouseOver={[Function]}
         role="tab"
+        style={
```
ericsoderberghp, line 227: I think we should avoid passing `style` if we aren't setting `zIndex`.

src/js/components/Accordion/stories/Custom.js
```diff
@@ -3,9 +3,17 @@ import { storiesOf } from '@storybook/react';
...
+  Accordion,
+  AccordionPanel,
+  Box,
+  Grommet,
+  Text,
+} from 'grommet';
+import { deepMerge } from 'grommet/utils';
 
-const CustomAccordionTheme = {
+const CustomAccordionTheme = deepMerge(grommet, {
```
ericsoderberghp, line 16: Why do we need to deepMerge with grommet? Can't we just create our theme object as JSON?

Also, can we rename the variable `customTheme`, as upper case sounds more like a component.

src/js/components/AccordionPanel/AccordionPanel.js
```diff
@@ -40,35 +41,59 @@ const AccordionPanel = forwardRef(
       [active, theme.accordion.icons],
     );
 
+    // backward compatibility to the deprecated theme.accordion.hover.color
+    const headingColor =
+      theme.accordion.hover &&
+      theme.accordion.hover.color !== { dark: 'light-4', light: 'dark-3' }
```
ericsoderberghp, line 47: I don't think this comparison does what you think it does. It will always be false.

I think we just want to test for existence of `theme.accordion.hover.color`. If that exists, we use that.

src/js/components/AccordionPanel/AccordionPanel.js
```diff
@@ -40,35 +41,59 @@ const AccordionPanel = forwardRef(
...
+            setHover(headingColor);
+            setFocus(true);
             if (onFocus) onFocus(event);
           }}
           onBlur={event => {
             setHover(undefined);
+            setFocus(false);
             if (onBlur) onBlur(event);
           }}
+          style={{ zIndex: focus ? 1 : undefined }}
```
ericsoderberghp, line 96: `style={focus ? { zIndex: 1 } : undefined}`

src/js/components/AccordionPanel/AccordionPanel.js
```diff
@@ -40,35 +41,59 @@ const AccordionPanel = forwardRef(
...
 
+    // backward compatibility to the deprecated theme.accordion.hover.color
+    const headingColor =
+      theme.accordion.hover &&
+      theme.accordion.hover.color !== { dark: 'light-4', light: 'dark-3' }
+        ? theme.accordion.hover.color
+        : theme.accordion.hover.heading.color;
+
+    const { border: contentBorder } = theme.accordion;
+    const { border } = theme.accordion.panel;
```
ericsoderberghp, line 52: can we rename this too, to be clearer: "border" => "panelBorder"

src/js/components/AccordionPanel/AccordionPanel.js
```diff
@@ -40,35 +41,59 @@ const AccordionPanel = forwardRef(
...
+
+    let abutMargin;
+    if (border)
+      abutMargin = {
```
ericsoderberghp, line 56: Can we add a comment explaining what this is here for?

src/js/components/Accordion/stories/Custom.js
ShimiSun: >Why do we need to deepMerge with grommet? Can't we just create our theme object as JSON?

I like the font style of 'grommet', but I can add the font directly to the theme instead.

> Also, can we rename the variable customTheme, as upper case sounds more like a component.

agreed.

src/js/components/AccordionPanel/AccordionPanel.js
ShimiSun: At the moment the base.js deafult is `theme.accordion.hover.color: { dark: 'light-4', light: 'dark-3' }`.
Let's say users decide to override this behavior on their theme by setting  `theme.accordion.hover.color='undefined'`,  in that case, your suggestion wouldn't be backward compatible as their setting will be overridden by `theme.accordion.hover.heading.color` which is currently `{ dark: 'light-4', light: 'dark-3' }`.

So the intended logic is if the user has defined `theme.accordion.hover.color` that is different than the default -> pick that. Otherwise, pick `theme.accordion.hover.heading.color`.

src/js/components/AccordionPanel/AccordionPanel.js
ShimiSun: Done, thanks.

src/js/components/Accordion/__tests__/__snapshots__/Accordion-test.js.snap
ShimiSun: Good catch! fixed.

src/js/components/AccordionPanel/AccordionPanel.js
ShimiSun: Fixed.

PR #4048, opened 2020-05-05
Layer constantly updates fix

### LINKED ISSUE #3964: Issue with Layer component
When I try to show up two layers at a time, the Layer component is setting and unsetting `area-hidden` property in an infinite loop. It stops setting only when I dismiss the Layer (by closing).
### Expected Behavior
It shouldn't set and unset in an infinite loop. 

At some point in time browser may freeze.
### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

IanKBovard:
What does this PR do?
closes #3964 

In FocusedContainer.js bodyOverflowStyle was constantly changing from undefined to 'hidden' causing the useEffect to run over and over again. With the conditional, only runs on mount.

I have a question about LayerContainer.js. There are two checks for modal, wouldn't one override the other?

```javascript
// first modal check line 121
    if (modal) {
      content = (
        <StyledLayer
          ref={layerRef}
          id={id}
          targetBounds={targetBounds}
          plain={plain}
          position={position}
          responsive={responsive}
          tabIndex="-1"
          dir={theme.dir}
        >
          <StyledOverlay
            plain={plain}
            onMouseDown={onClickOutside}
            responsive={responsive}
          />
          {content}
        </StyledLayer>
      );
    }
```

```javascript
// second modal check line 157
    if (modal) {
      content = (
        <FocusedContainer hidden={position === 'hidden'} restrictScroll>
          {content}
        </FocusedContainer>
      );
    }
```

FocusedContainer.js

Storybook layer examples.

Same as above

No

Backwards compatible

### DIFF
--- a/src/js/components/FocusedContainer.js
+++ b/src/js/components/FocusedContainer.js
@@ -23,7 +23,6 @@ export const FocusedContainer = ({
       getBodyChildElements()
         .filter(isNotAncestorOf(child))
         .forEach(makeNodeFocusable);
-      // bodyOverflowStyle is causing the layer to constantly update
       if (restrictScroll) {
         document.body.style.overflow = bodyOverflowStyle;
       }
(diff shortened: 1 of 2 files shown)


ericsoderberghp:
> I have a question about LayerContainer.js. There are two checks for modal, wouldn't one override the other?

The second check wraps the `content` in a FocusedContainer. So, it doesn't override content, it wraps it further.

IanKBovard:
Ah! I gotcha. thanks for clarifying that.

PR #4049, opened 2020-05-05
Allow disabled styling to affect MaskedInput

taysea:
#### What does this PR do?
Adds disabledStyle to MaskedInput. Currently, disabled styling is not being applied to maskedinput like it is for other inputs such as TextInput.

src/js/components/MaskedInput/StyledMaskedInput.js

disabled was applied to a storybook to ensure the opacity changed the same way that TextInput opacity does by default. A jest test was added for future.

Add `disabled` to MaskedInput -- if using `grommet theme`, or base.js, the disabledStyle should default to opacity 0.3.

When working on MaskedInput examples for the HPE Design System, I noticed that the placeholder text of the MaskedInput was not being toggled in opacity the same that TextInput was.

Yes, they're updated here.

[Shimi] Yes - Added `disabled` theme styling to MaskedInput.

This will now toggle opacity to 0.3 for disabled maskedinputs by default, which wasn't in place before but is needed to align with other inputs. It wouldn't be backwards compatible -- suggestions on what to do here?

### DIFF
--- a/src/js/components/MaskedInput/StyledMaskedInput.js
+++ b/src/js/components/MaskedInput/StyledMaskedInput.js
@@ -1,6 +1,7 @@
 import styled, { css } from 'styled-components';
 
 import {
+  disabledStyle,
   focusStyle,
   getInputPadBySide,
   inputStyle,
@@ -39,6 +40,12 @@ export const StyledMaskedInput = styled.input`
   }
 
   ${props => props.focus && !props.plain && focusStyle()};
+  ${props =>
+    props.disabled &&
+    disabledStyle(
+      props.theme.maskedInput.disabled &&
+        props.theme.maskedInput.disabled.opacity,
+    )}
   ${props => props.theme.maskedInput && props.theme.maskedInput.extend};
 `;
 
--- a/src/js/components/MaskedInput/doc.js
+++ b/src/js/components/MaskedInput/doc.js
@@ -87,8 +87,9 @@ export const themeDoc = {
   'maskedInput.disabled.opacity': {
     description: 'The opacity when the MaskedInput is disabled.',
     type: 'number',
-    defaultValue: 0.3,
+    defaultValue: 'global.control.disabled.opacity',
   },
+  ...themeDocUtils.disabledStyle,
   ...themeDocUtils.focusStyle,
   ...themeDocUtils.placeholderStyle,
   ...themeDocUtils.inputStyle,
--- a/src/js/components/MaskedInput/doc.js
+++ b/src/js/components/MaskedInput/doc.js
@@ -86,8 +86,8 @@ export const themeDoc = {
   },
   'maskedInput.disabled.opacity': {
     description: 'The opacity when the MaskedInput is disabled.',
-    type: 'number',
-    defaultValue: 'global.control.disabled.opacity',
+    type: 'number | string',
+    defaultValue: undefined,
   },
   ...themeDocUtils.disabledStyle,
   ...themeDocUtils.focusStyle,
--- a/src/js/components/MaskedInput/doc.js
+++ b/src/js/components/MaskedInput/doc.js
@@ -84,6 +84,11 @@ export const themeDoc = {
     type: 'string',
     defaultValue: '18px',
   },
+  'maskedInput.disabled.opacity': {
+    description: 'The opacity when the MaskedInput is disabled.',
+    type: 'number',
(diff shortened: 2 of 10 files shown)


src/js/components/MaskedInput/StyledMaskedInput.js
```diff
@@ -39,6 +40,12 @@ export const StyledMaskedInput = styled.input`
   }
 
   ${props => props.focus && !props.plain && focusStyle()};
+  ${props =>
+    props.disabled &&
+    disabledStyle(
+      props.theme.maskedInput.disabled &&
+        props.theme.maskedInput.disabled.opacity,
+    )}
```
ShimiSun, line 48: It looks consistent with TextInput.

src/js/components/MaskedInput/doc.js
```diff
@@ -84,6 +84,11 @@ export const themeDoc = {
     type: 'string',
     defaultValue: '18px',
   },
+  'maskedInput.disabled.opacity': {
+    description: 'The opacity when the MaskedInput is disabled.',
+    type: 'number',
+    defaultValue: 0.3,
```
ShimiSun, line 90: According to your base.js definition, the defaultValue should be undefined.

src/js/components/MaskedInput/doc.js
```diff
@@ -84,6 +84,11 @@ export const themeDoc = {
...
+  },
```
ShimiSun, line 91: I think you are missing the property that will make the connection to the default 0.3 value
```
...themeDocUtils.disabledStyle,
```

taysea: You're right, it's falling back on global.control.disabled.opacity -- I will update the documentation so the link is clear.

src/js/components/MaskedInput/doc.js
taysea: it's falling back on global.control.disabled.opacity -- I will update the documentation so the link is clear.

Going to adjust TextInput docs as well since the same is occurring there.

src/js/components/MaskedInput/StyledMaskedInput.js
taysea: Yes, it is consistent with TextInput.

src/js/components/MaskedInput/doc.js
```diff
@@ -84,6 +84,12 @@ export const themeDoc = {
     type: 'string',
     defaultValue: '18px',
   },
+  'maskedInput.disabled.opacity': {
+    description: 'The opacity when the MaskedInput is disabled.',
+    type: 'number',
+    defaultValue: 'global.control.disabled.opacity',
```
ShimiSun, line 90: According to base.js, the `defaultValue` is `undefined`, and I think that is the value we should be explicit about.
```
...themeDocUtils.disabledStyle,
```
will be documenting the `global.control.disabled.opacity` property independently.

src/js/components/TextInput/doc.js
```diff
@@ -199,7 +199,7 @@ export const themeDoc = {
   'textInput.disabled.opacity': {
     description: 'The opacity when the textInput is disabled.',
     type: 'number',
-    defaultValue: 0.3,
+    defaultValue: 'global.control.disabled.opacity',
```
ShimiSun, line 202: Good catch, thanks for addressing it.
According to base.js, the defaultValue should be `undefined`
```
textInput: {
      // extend: undefined,
      // disabled: { opacity: undefined },
    },
```

src/js/components/MaskedInput/doc.js
```diff
@@ -84,6 +84,12 @@ export const themeDoc = {
     type: 'string',
     defaultValue: '18px',
   },
+  'maskedInput.disabled.opacity': {
+    description: 'The opacity when the MaskedInput is disabled.',
+    type: 'number',
```
ShimiSun, line 89: I'm thinking boolean and string are also acceptable values, can you confirm?

taysea: I will test it out and update the docs if needed.

ShimiSun: great, opacity should be able to accept t-shirt sizes ("small",...)

taysea: String is also accepted, however applying a boolean did not have the intended effect.

ShimiSun:
Thank you! 💟

PR #4050, opened 2020-05-05
Changed Button to add kind for default and secondary

ericsoderberghp:
#### What does this PR do?

Changed Button to add kind for default and secondary.

Broadly, this change attempts to preserve the prior behavior (StyledButton.js) and separately implement the default and secondary behavior (StyledButtonKind.js).

Button.js

unit tests
Storybook -> Button -> Themes

Storybook -> Button -> Themes

[Shimi] add fix description to #2820

yes, props automatically
but, we should add guidance on the various kinds of buttons in the documentation, like we did with Form

yes

backwards compatible


src/js/themes/base.d.ts
```diff
@@ -769,7 +778,6 @@ export interface ThemeType {
     extend?: ExtendType;
     icons?: {
       down?: any;
-      color?: ColorType;
```
ShimiSun, line 772: Are we not allowing the `color` theme style for an `icon`? how is that backward compatible?

src/js/themes/base.d.ts
```diff
@@ -77,6 +77,19 @@ type Colors = typeof colors & {
   [x: string]: ColorType;
 };
 
+interface ButtonKindType {
+  background?: BackgroundType;
+  border?: {
+    color?: ColorType;
+  };
+  color?: ColorType;
+  padding?: {
```
ShimiSun, line 86: I'm thinking we should call it `pad` for grommet consistency.

src/js/components/Button/doc.js
```diff
@@ -177,6 +181,46 @@ export const themeDoc = {
...
+    type: 'string | { dark: string, light: string }',
+    defaultValue: undefined,
+  },
+  'button.active.color': {
+    description: `Label color when the button is active.`,
+    type: 'string | { dark: string, light: string }',
+    defaultValue: undefined,
+  },
+  'button.active.extend': {
+    description: 'Any additional style for a active Button.',
```
ShimiSun, line 201:
```suggestion
    description: 'Any additional style for an active Button.',
```

src/js/components/Button/doc.js
```diff
@@ -177,6 +181,46 @@ export const themeDoc = {
...
+    type: 'string | (props) => {}',
+    defaultValue: undefined,
+  },
+  'button.active.default': {
+    description:
+      'Adjustments to the default Button style when the Button is active.',
+    type: '{}',
```
ShimiSun, line 208: In most cases of documentations, we use `object` vs `{}`, something like:
```suggestion
    type: 'object',
```

src/js/themes/base.d.ts
ericsoderberghp: I went back and forth on this. My current thinking is that we use `padding` where it is CSS and we use `pad` where it is being used as a grommet prop. Also, since we already had `padding` in theme.button, it seemed like consistency within the component would be helpful. All that to say, if you think `pad` would be better, I'm happy to switch.

src/js/components/Button/index.d.ts
```diff
@@ -31,6 +31,7 @@ export interface ButtonProps {
   plain?: boolean;
   primary?: boolean;
   reverse?: boolean;
+  secondary?: boolean;
```
ShimiSun, line 34: we should also add `kind`.

src/js/components/Button/doc.js
```diff
@@ -102,6 +102,9 @@ with plain Buttons.`,
               end of the anchor.`,
       )
       .defaultValue(false),
+    secondary: PropTypes.bool.description(
```
ShimiSun, line 105: `kind` is missing from doc.js.

src/js/themes/base.d.ts
ShimiSun: I'm ok either as long as we make a decision here.
Your statement makes sense to me, but let's keep in mind for consistency that components like DataTable, List, Select, Tab, Table, and even button.size is using `pad` vs `padding`. but I'm OK proceeding with the direction you suggested.

src/js/components/Button/doc.js
ericsoderberghp: I wasn't sure I was ready to commit fully to it. I was considering having it available for experimentation, part of this request: https://grommet.slack.com/archives/C09QQEG69/p1588591623025200?thread_ts=1588591623.025200
But, not documenting it officially until we are sure we want it. Too strange?

ShimiSun: > Too strange?

No, I actually agree with the experimental approach.
That just means we wouldn't want to mention the `kind` approach and functionality on the release-notes.

ericsoderberghp: completely agree

src/js/components/Button/StyledButtonKind.js
```diff
@@ -0,0 +1,221 @@
...
+};
+
+const fontStyle = props => {
+  const size = props.sizeProp || 'medium';
+  const data = props.theme.text[size];
+  return css`
+    font-size: ${data.size};
+    line-height: ${data.height};
+  `;
+};
```
ShimiSun, line 26: My only concern is that the function is duplicated on StyledButton.js, and any future enhancements for this function would need to be applied on both hence might be missed for either one.

ericsoderberghp: I thought about this. But, you'll notice that `plainStyle()` is slightly different between the two. I was more concerned that unintended side-effects would happen changing a single copy. I was trying to maintain separation between old and new approaches.

src/js/components/Button/StyledButtonKind.js
```diff
@@ -0,0 +1,221 @@
...
+const kindPartStyles = (obj, theme, colorValue) => {
+  const styles = [];
+  if (obj.padding) {
+    if (obj.padding.vertical)
+      styles.push(
+        `padding: ${theme.global.edgeSize[obj.padding.vertical] ||
+          obj.padding.vertical ||
+          0} ${theme.global.edgeSize[obj.padding.horizontal] ||
+          obj.padding.horizontal ||
+          0};`,
```
ShimiSun, line 70: I didn't get a chance to test it yet, but where does the parsing of `horizontal` and `vertical` (to `left`, `top` etc) is happening? should this logic be using `parseMetricToNum`?

src/js/components/Button/StyledButtonKind.js
```diff
@@ -0,0 +1,221 @@
...
+
+// CSS for this sub-object in the theme
+const kindPartStyles = (obj, theme, colorValue) => {
+  const styles = [];
+  if (obj.padding) {
+    if (obj.padding.vertical)
+      styles.push(
+        `padding: ${theme.global.edgeSize[obj.padding.vertical] ||
+          obj.padding.vertical ||
+          0} ${theme.global.edgeSize[obj.padding.horizontal] ||
```
ShimiSun, line 68: If the user defines `obj.padding.vertical` but not `obj.padding.horizontal` we are assuming the `horizontal` value is '0' rather than falling back to `button.padding.horizontal`? just confirming.

src/js/themes/base.d.ts
```diff
@@ -77,6 +77,19 @@ type Colors = typeof colors & {
   [x: string]: ColorType;
 };
 
+interface ButtonKindType {
+  background?: BackgroundType;
+  border?: {
+    color?: ColorType;
```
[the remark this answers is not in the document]
ericsoderberghp: we can add that later, if the demand arises.

src/js/components/Button/StyledButtonKind.js
ericsoderberghp: The CSS `padding: a b;` means top and bottom get "a" and left and right get "b". We don't need `parseMetricToNum()` because edgeSize wants a string, not a number, and the CSS will take whatever is passed through.

src/js/components/Button/StyledButtonKind.js
ericsoderberghp: Correct. I've updated the test such that they can specify either `horizontal` or `vertical`.

src/js/components/Button/StyledButtonKind.js
ShimiSun: Thanks for clarifying.

src/js/themes/base.d.ts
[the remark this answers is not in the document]
ShimiSun: We are using border.width on the storybook example, I think we should at least add that.

ShimiSun, line 83:
```suggestion
    color?: ColorType;
    width?: string;
```

PR #4060, opened 2020-05-07
Update docs for focusIndicator prop on Select

markjackson02:
#### What does this PR do?
Update documentation to reflect current functionality. Current documentation indicates that `focusIndicator` prop is only applicable if `plain={true}`. However this is only true of the `TextInput`. For `Button` and `Select`, `focusIndicator` works without the prerequisite of `plain={true}`.

#### Where should the reviewer start / How should this be manually tested?
You can confirm that `focusIndicator` works without `plain` by testing adding `focusIndicator={false}` on the interactive demos.
https://v2.grommet.io/button
https://v2.grommet.io/select

N/A - only documentation updates

I don't think so


src/js/components/Button/doc.js
```diff
@@ -47,7 +47,7 @@ export const doc = Button => {
       )
       .defaultValue(false),
     focusIndicator: PropTypes.bool
-      .description("Whether when 'plain' it should receive a focus outline.")
+      .description('Whether it should receive a focus outline.')
```
ShimiSun, line 50: I think this is only true for `plain` button.

markjackson02: Hmm you're right I must have tested on the wrong page. So it looks like Select is the only one whose docs need updating. Will update PR

ShimiSun: `  ${props =>
    props.focus && (!props.plain || props.focusIndicator) && focusStyle()}`

ShimiSun:
I'll be closing this PR since the `focusIndicator` should apply only on `plain` inputs or controls.
Linking to this issue https://codesandbox.io/s/grommet-v2-template-xe041?file=/index.js

PR #4062, opened 2020-05-08
Changed Button to fix an issue with plain + icon

ericsoderberghp:
#### What does this PR do?

Changed Button to fix an issue with plain + icon.
This only occurs when a Button has `plain`, `icon`, and the theme has `button.default`.

Button.js

whitebox

no

no

backwards compatible


RobertGary1:
What was the symptom of this PR?

ShimiSun:
@RobertGary1 I think we wanted to apply the icon color only if the button wasn't plain. and that is only for kind button when button.default is turned on on the theme.

PR #4063, opened 2020-05-08
Changed Select and TextInput to accept a11yTitle

ShimiSun:
#### What does this PR do?
Propagates the a11yTitle prop as `aria-label` prop to TextInput, hence fixes the aria-label issue for both Select and TextInput components.
TextInput.js
Added unit tests and verified snapshots are containing the `aria-label` prop.

snapshots
a request for this fix came from @MikeKingdom
https://codesandbox.io/s/restless-smoke-g3zjc?file=/src/App.js
no
yes
backwards compatible

### DIFF
--- a/src/js/components/Select/__tests__/__snapshots__/Select-test.js.snap
+++ b/src/js/components/Select/__tests__/__snapshots__/Select-test.js.snap
@@ -235,6 +235,241 @@ exports[`Select 0 value 1`] = `
 </button>
 `;
 
+exports[`Select a11yTitle 1`] = `
+.c8 {
+  display: inline-block;
+  -webkit-flex: 0 0 auto;
+  -ms-flex: 0 0 auto;
+  flex: 0 0 auto;
+  width: 24px;
+  height: 24px;
+  fill: #7D4CDB;
+  stroke: #7D4CDB;
+}
+
+.c8 g {
+  fill: inherit;
+  stroke: inherit;
+}
+
+.c8 *:not([stroke])[fill="none"] {
+  stroke-width: 0;
+}
+
+.c8 *[stroke*="#"],
+.c8 *[STROKE*="#"] {
+  stroke: inherit;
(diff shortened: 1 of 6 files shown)


MikeKingdom:
Thanks Shimi!

src/js/components/Select/__tests__/__snapshots__/Select-test.js.snap
```diff
@@ -235,6 +235,241 @@ exports[`Select 0 value 1`] = `
...
+    className="c2"
+  >
+    <div
+      className="c3"
+    >
+      <div
+        className="c4"
+      >
+        <input
+          aria-label="aria-test, "
```
ericsoderberghp, line 433: Why the ", " at the end?

ShimiSun: The way that Select is processing and assigning `a11yTitle` is a little funky https://github.com/grommet/grommet/blob/d691e84055eee8a60f13dee99c97b7e311e81fb7/src/js/components/Select/Select.js#L246

I didn't dive into this existing logic (it's been there for a couple of years), but I'm guessing it is to allow concatenation of strings.

I can take a deeper look into it if needed.

ericsoderberghp: In Select.js, I think we should make this change...
from:
```
a11yTitle={
  a11yTitle &&
  `${a11yTitle}${
    typeof value === 'string' ? `, ${value}` : ''
  }`
}
```
to:
```
a11yTitle={
  a11yTitle &&
  `${a11yTitle}${
    value && typeof value === 'string' ? `, ${value}` : ''
  }`
}
```

ShimiSun: Done. the snapshot looks as expected now.

PR #4069, opened 2020-05-11
TextInput onSelect and onSuggestionSelect handlers

### LINKED ISSUE #3118: TextInput makes it difficult to listen for select events on input element
`<TextInput />` has an `onSelect` prop used for the suggestions feature which overrides the `<input />`s `onSelect` prop for listening to [`select`](https://developer.mozilla.org/en-US/docs/Web/API/Element/select_event) events. This makes listening to `select` events on the input more difficult than it should be.

As a solution, I purpose changing the current `onSelect` prop to `onSuggestionSelect` and allow `onSelect` to resume it's default behaviour on the `<input />`.

### Expected Behavior

`<TextInput onSelect={/* ...  */} />` should add the provided function as an event handler for `select` events.

### Actual Behavior

Grommet overrides the `onSelect` prop for suggestion selection events.

https://codesandbox.io/s/grommet-v2-template-5rcuj (includes a workaround).

### Steps to Reproduce

1. Define an `onSelect` event handler on a `<TextInput />`
2. Notice that it does not listen for `select` events on the `<input />`.

### Your Environment

- Grommet version: 2.7.0

IanKBovard:
#### What does this PR do?
Allows the use of default onSelect event for textInput.

TextInput.js

Within storybook using this.

```javascript
import React from 'react';

import { Box, Grommet, TextInput } from 'grommet';

const suggestions = Array(100)
  .fill()
  .map((_, i) => `suggestion ${i + 1}`);

const App = () => {

  const onChange = event => setValue(event.target.value);

  const onSelect = event => setValue(event.suggestion);
  const onHightlight = event =>
    console.log(
      event.target.value.substring(
        event.target.selectionStart,
        event.target.selectionEnd,
      ),
    );
  return (
        <Box width="medium">
          <TextInput
            value={value}
            onChange={onChange}
           // i tested with a combination of onSelect and onSuggestionSelect being present and 
           absent
            onSelect={onHightlight}
            onSuggestionSelect={onSelect}
            suggestions={suggestions}
          />
        </Box>
      </Box>
    </Grommet>
  );
};

export default App;
```

Same as above. 

Grommets onSelect was overriding the html onSelect.

closes #3118 

Yes

Yes

Backwards compatible


IanKBovard:
Still need to add tests.

This pr also updates the select-test snapshots because onSelect was added as prop.

src/js/components/TextInput/TextInput.js
```diff
@@ -376,43 +391,88 @@ const TextInput = forwardRef(
...
-            {...rest}
-            defaultValue={renderLabel(defaultValue)}
-            value={renderLabel(value)}
-            readOnly={readOnly}
-            onFocus={event => {
-              setFocus(true);
-              if (suggestions && suggestions.length > 0) {
-                announce(messages.suggestionsExist);
-                openDrop();
+          {handleTextSelect.current ? (
```
ShimiSun, line 394: I don't think we should have a condition of 'handleTextSelect' to control the whole rendering of StyledTextInput. Lots of duplicated code here.

src/js/components/TextInput/TextInput.js
```diff
@@ -114,6 +114,21 @@ const TextInput = forwardRef(
...
+    const handleTextSelect = useRef();
+
+    // assigns select handlers
+    useEffect(() => {
+      if (onSelect && !onSuggestionSelect) {
+        handleSuggestionSelect.current = onSelect;
+      } else if (!onSelect && onSuggestionSelect) {
+        handleSuggestionSelect.current = onSuggestionSelect;
+      } else {
+        handleSuggestionSelect.current = onSuggestionSelect;
```
ShimiSun, line 127:
```suggestion
        // in case both onSelect & onSuggestionSelect are defined
        handleSuggestionSelect.current = onSuggestionSelect;
```
Adding descriptive comment.

ShimiSun:
@IanKBovard tests and adding the new function to doc.js is still missing.

IanKBovard:
added tests, some comments, and documentation updates. A few snaps were updated, they seem to me necessary. if they are not let me know and i can try to figure out a way around them.

src/js/components/TextInput/doc.js
```diff
@@ -71,7 +71,13 @@ export const doc = TextInput => {
     ),
     onSelect: PropTypes.func.description(
       `Function that will be called when the user selects a suggestion.
-The suggestion contains the object chosen from the supplied suggestions.`,
+      The suggestion contains the object chosen from the supplied suggestions.
```
ShimiSun, line 74: I would add at the beginning of the description that this function is deprecated and the user should use the onSuggestionSelect instead.

IanKBovard:
If we can get this reviewed, we can merge this in.

ShimiSun:
@IanKBovard can you please resolve the conflicts and update the branch with master.

src/js/components/TextInput/TextInput.js
```diff
@@ -115,6 +115,26 @@ const TextInput = forwardRef(
     const [focus, setFocus] = useState();
     const [showDrop, setShowDrop] = useState();
 
+    const handleSuggestionSelect = useRef();
```
ericsoderberghp, line 118: The way these are being used seems like they should be using `useMemo` instead of `useRef + useEffect`.

src/js/components/TextInput/TextInput.js
```diff
@@ -228,6 +248,9 @@ const TextInput = forwardRef(
       placeholder && typeof placeholder !== 'string' && !value;
 
     let drop;
+    const extraProps = {
+      onSelect: handleTextSelect.current && handleTextSelect.current,
```
ericsoderberghp, line 252: Why the `&&` for identical conditions?

IanKBovard: removed, just old logic that was never completely deleted.

src/js/components/TextInput/TextInput.js
IanKBovard: I'm a little confused about this. My understanding of useMemo is to optimize the performance of complex functions. How can I use it to replace useRef and useEffect?

IanKBovard:
Closing this pr. no idea why its linting typescript files like this.

new one here https://github.com/grommet/grommet/pull/4417

Issue #4019, opened 2020-04-28
Select component displays gaps on scroll with many items (e.g. 25)


jesseolsen:
### Expected Behavior

Normal display of items in a Select component, even if > 25 items

### Actual Behavior

Select item in the middle (e.g. 22), then I see a large gap (whitespace) after that item.  See screenshot below.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-sandbox-2r3qw?file=/src/Select.js

![image](https://user-images.githubusercontent.com/6043493/80508603-7bb5b780-8935-11ea-8d8e-86388e3a614f.png)

### Steps to Reproduce

1. Go to codesandbox forked above
2. Select the "Select" item
3. Select "22" from the list of numbers
4. Scroll up and down quickly, until you notice a big gap, like the one shown in the screenshot after 20.

### Your Environment

- Grommet version: 2.13.0
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): Windows 10

ericsoderberghp:
Hmm... I can't seem to reproduce this. I'm wondering if it is tied to timing differences that only surface in a Windows environment. Still exploring ...

jesseolsen:
I have reproduced it on Windows as well as Mac, both on Chrome.  Steps I used to reproduce this (more detailed this time):
Select Input | Select.
Select 22.  The drop down will close.
Open it back up.  Slowly scroll down until you see "20".  Continue slowly scrolling, and you should see a large gap (white-space) before you eventually see 22 at which time the white space disappears.  Scroll back up and the white space reappears.

IanKBovard:
here's a video of the issue.
![Kapture 2020-05-08 at 11 19 35](https://user-images.githubusercontent.com/21990914/81425788-da411980-911d-11ea-9222-847b831e3d85.gif)

ShimiSun:
Fixed and available on the [stable branch](https://github.com/grommet/grommet/wiki/What-is-grommet-stable-and-how-to-use-it%3F).
Thanks for reporting.

PR #4071, opened 2020-05-11
Changed SelectContainer to use hooks.

ericsoderberghp:
#### What does this PR do?

Changed SelectContainer to use hooks.

SelectContainer.js

unit and some storybook

storybook!

#4019

no

yes

backwards compatible

### DIFF
--- a/src/js/components/Select/SelectContainer.js
+++ b/src/js/components/Select/SelectContainer.js
@@ -1,12 +1,14 @@
-import React, { createRef, Component } from 'react';
-import styled, { withTheme } from 'styled-components';
+import React, {
+  forwardRef,
+  useCallback,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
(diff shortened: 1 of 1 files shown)


src/js/components/Select/SelectContainer.js
```diff
@@ -323,15 +309,15 @@ class SelectContainer extends Component {
               <SelectTextInput
                 focusIndicator={!customSearchInput}
                 size="small"
-                ref={this.searchRef}
+                ref={searchRef}
                 type="search"
-                value={search}
+                value={search || ''}
```
ShimiSun, line 314: Following our discussion of controlled/uncontrolled components, I thought we are trying to avoid setting initial values that will force components to be controlled by grommet.

ericsoderberghp: This is the embedded TextInput for searching inside the Select drop down. So, we want to control this one.

PR #4072, opened 2020-05-12
Decouple documentation modules from functional modules

jecassis:
Currently modules from `react-desc` and some smaller utilities are leaking into the production bundle as well as downstream consumers of Grommet. I see that the roadmap lists finding an alternative for `react-desc` in 2020, so it is possible this aids in easing that transition when the time comes. This change results in approximately `11 KiB` savings on the size of `grommet.min.js`.

#### What does this PR do?
Separate the documentation modules from the `utils` index and import them into the `doc.js` modules directly. In this way, in production builds, no documentation-related code will appear.

See what was removed in `src/js/utils/index.js` and then open a `doc.js` file; the Box component one seems like a good one to get an idea of the changes.

Build and test suites pass. Documentation is generated. On my downstream project I can consume the package with the changes and verified functionality, in addition to the disappearance of `react-desc` from that bundle.

Since there is no functional change expected and only documentation changes. Manual verification should be focused there, if any.

None.

No issue filed. Consider this pull request the issue.

Not necessary, unless you want to see console output.

They were, sort of :)

Yes. Branched off 2.13.0. The [build](https://app.circleci.com/pipelines/github/grommet/grommet/4060/workflows/3597d79b-98bc-410b-87bd-1362920d6a7f/jobs/18503) for that release shows a bundle size of `434 KiB` for `grommet.min.js`. This pull request reduces it to `423 KiB`. My downstream project, after compression, sees a net savings of `4 KiB` if that matters.

This change is backwards-compatible and non-breaking.


ShimiSun:
From a quick local analysis, I can verify the[ tree shaking](https://webpack.js.org/guides/tree-shaking/) approach suggested on this PR is making an impact on the grommet.min.js file size
image
But I can still see that `react-desc` is part of the `dist` production package 
image
How were you able to verify
>the disappearance of react-desc from that bundle.

?

jecassis:
I verified it using `webpack-bundle-analyzer`. The `react-desc` package is still included as a non-dev dependency. So that is why you might be seeing it in `node_modules/` inside the `dist/` package. However the files do not seem to be parsed as you can tell by the numbers reported in the build. I deemed removing `react-desc` completely from the `dependencies` key in `package.json` too risky given I do not know all of the possible build modes and variations Grommet might have and what assumptions are built into each.

jecassis:
Found the culprit. No `react-desc` in `grommet.min.js` for a further `0.43 KiB` reduction gzip'd. And also moved `react-desc` into `devDependencies`.

jecassis:
Caution was warranted on `react-desc` in `package.json`. It cannot be moved to `devDependencies` because when working in non-production mode the library is missing and builds fail. So left untouched.

MikeKingdom:
Replaced by PR #4503

PR #4073, opened 2020-05-12
Changed Collapsible to use hooks

ericsoderberghp:
#### What does this PR do?

Changed Collapsible to use hooks.

Collapsible.js

storybook, unit

storybook

no

yes

backwards compatible

### DIFF
--- a/src/js/components/Accordion/__tests__/__snapshots__/Accordion-test.js.snap
+++ b/src/js/components/Accordion/__tests__/__snapshots__/Accordion-test.js.snap
@@ -154,9 +154,11 @@ exports[`Accordion AccordionPanel 1`] = `
 }
 
 .c10 {
-  max-height: 0;
-  visibility: hidden;
+  -webkit-transition: max-height 200ms,opacity 200ms;
+  transition: max-height 200ms,opacity 200ms;
+  opacity: 0;
   overflow: hidden;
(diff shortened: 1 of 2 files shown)


src/js/components/Accordion/__tests__/__snapshots__/Accordion-test.js.snap
```diff
@@ -154,9 +154,11 @@ exports[`Accordion AccordionPanel 1`] = `
 }
 
 .c10 {
-  max-height: 0;
-  visibility: hidden;
+  -webkit-transition: max-height 200ms,opacity 200ms;
+  transition: max-height 200ms,opacity 200ms;
+  opacity: 0;
   overflow: hidden;
+  max-height: 0;
```
ShimiSun, line 161: what is the need for adding `opacity` and `max-height` to the snapshots?

ericsoderberghp: `max-height` is just moving in the snapshot, see previous line 157.
`opacity` is replacing `visibility`. The old code was attempting to animate `visibility` via `transition`. But, `visibility` isn't an animatable property, `opacity` is.

ShimiSun:
`interns` - we need to add tests for Collapsible.

Issue #4082, opened 2020-05-15
Fix deprecation warnings when installing Grommet


jecassis:
`npm install grommet` and `yarn add grommet` should not output deprecation warnings on install.

### Actual Behavior
Yarn example:
```
warning grommet > css > urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
warning grommet > css > source-map-resolve > urix@0.1.0: Please see https://github.com/lydell/urix#deprecated
warning grommet > css > source-map-resolve > resolve-url@0.2.1: https://github.com/lydell/resolve-url#deprecated
warning grommet > recompose > fbjs > core-js@1.2.7: core-js@<3 is no longer maintained and not recommended for usage due to the number of issues. Please, upgrade your dependencies to the actual version of core-js@3.
```

### URL, screen shot, or Codepen exhibiting the issue
See console output above.

### Steps to Reproduce
`yarn install grommet@2.13.0`

### Your Environment
- Grommet version: 2.13.0
- Browser Name and version: Not applicable
- Operating System and version (desktop or mobile): Windows 10

jecassis:
The last one is fixed by removing the `recompose` dependency in #4074.

For the first three may I propose the following solution (as submitted in pull request #4083):

1. Replace `css` with `postcss`. `css` appears to not be actively maintained.
2. Move it out of `dependencies` and into `devDependencies` as the `portal.js` file is only imported in `*-test.js`.

ShimiSun:
Thank you for your contribution to this issue buddy! I'll be closing this issue.

PR #4083, opened 2020-05-15
Replace `css` with PostCSS and move test-only package out of dependencies

jecassis:
#### What does this PR do?
Replaced `css` package (last commit Sep 4, 2018) with PostCSS and use it to parse and traverse the AST in `portal.js`. Also leverages Prettier to format the output for the snapshot. Finally, and most importantly, moving this test-only package into `devDependencies` means consumers of Grommet will not see deprecation warnings when installing.

`src/js/utils/portal.js` for the changes. Then also verify that the differences in the modified snapshot are only whitespace and simply due to Prettier's default CSS formatting.

All tests pass. Build completes with almost binary equivalence to 2.13.0.

Snapshots are the only things that should be affected and the diffs should show the change.

See above.

#4082 

N/A

No.

Details seem only relevant to developers.

Backwards compatible. No change in behavior or functionality for users.

### DIFF
--- a/src/js/components/Drop/__tests__/__snapshots__/Drop-test.js.snap
+++ b/src/js/components/Drop/__tests__/__snapshots__/Drop-test.js.snap
@@ -79,10 +79,9 @@ exports[`Drop align left left 2`] = `
   -ms-flex-direction: column;
   flex-direction: column;
   overflow: auto;
-  box-shadow: 0px 2px 4px rgba(0,0,0,0.20);
+  box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.2);
 }
-
-@media screen and (-ms-high-contrast:active),(-ms-high-contrast:none) {
+@media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
   .kQAKuZ {
     display: -webkit-box;
     display: -webkit-flex;
@@ -175,9 +174,8 @@ exports[`Drop align left right top bottom 2`] = `
   -ms-flex-direction: column;
   flex-direction: column;
   overflow: auto;
-  box-shadow: 0px 2px 4px rgba(0,0,0,0.20);
+  box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.2);
 }
-
 .kQAKuZ {
   font-size: 18px;
   line-height: 24px;
@@ -201,8 +199,7 @@ exports[`Drop align left right top bottom 2`] = `
   -webkit-animation-delay: 0.01s;
   animation-delay: 0.01s;
 }
-
-@media screen and (-ms-high-contrast:active),(-ms-high-contrast:none) {
+@media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
   .kQAKuZ {
     display: -webkit-box;
     display: -webkit-flex;
@@ -295,10 +292,9 @@ exports[`Drop align right left top top 2`] = `
   -ms-flex-direction: column;
   flex-direction: column;
   overflow: auto;
-  box-shadow: 0px 2px 4px rgba(0,0,0,0.20);
+  box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.2);
 }
-
-@media screen and (-ms-high-contrast:active),(-ms-high-contrast:none) {
+@media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
   .kQAKuZ {
     display: -webkit-box;
     display: -webkit-flex;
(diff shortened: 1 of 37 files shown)


ShimiSun:
@jecassis your PRs are awesome! thank you! I'll try to dive into this one soon

src/js/utils/portal.js
```diff
@@ -1,4 +1,5 @@
-import css from 'css';
+import css from 'postcss'; // eslint-disable-line max-len, import/no-extraneous-dependencies
+import prettier from 'prettier'; // eslint-disable-line max-len, import/no-extraneous-dependencies
```
ShimiSun, line 2: why do we need all the linter comments?

src/js/utils/portal.js
```diff
@@ -10,24 +11,32 @@ export const expectPortal = portalId => ({
...
-          return true;
+        if (rule.selectors) {
+          const selector = rule.selectors.join('');
+          if (
+            !Array.from(node.classList).some(
+              className => selector.indexOf(className) >= 0,
+            )
+          ) {
+            rule.remove();
+          }
```
ShimiSun, line 33: Can you explain what is the motivation for these changes? Is it related to the package changes?

src/js/utils/portal.js
jecassis: The `eslint` rules correctly disallow devDependecies from being imported in the regular source code. However because in `utils` all types of files are thrown in together, some of these are false positives. It is possible to change the global policy or override the rule itself in `.eslintrc` but I thought that was too heavy-handed and could lead to regressions. So I chose to waive the violations per line, however the waiver comments make the line go over the max-len limit so also waived that inline.

Options are:
- Leave as is.
- Waive the block, i.e. disable then enable with the comments above or below.
- Add an additional file to the override block in `.eslintrc` where more rules are also modified or create a new override block just for this file.
- Specify the lint rule globally "specifically" for Grommet throwing away the definition in `eslint-config-airbnb`. If this is what you'd prefer it is better to file it as a new issue. Example:
```json
    'import/no-extraneous-dependencies': ['error', {
      devDependencies: [
        '**/__tests__/**',
        '*.stories.js',
        '**/stories/**'
        '**/portal.js',
        '**/webpack.config.*.js'
      ],
      optionalDependencies: false,
    }],
```

src/js/utils/portal.js
jecassis: Besides the errors, the `css` package does not seem to be maintained any longer and it is not a standalone package so its dependencies are very stale. `postcss` seems to be actively maintained. And so, the motivation, is to migrate Grommet to devDependencies that are maintained with security and bug fixes. `postcss` also greatly expands the capabilities of what can be done given its ecosystem of plugins (though not relevant for this discussion).

This snippet is just reproducing the same behavior of traversing the AST once the CSS is parsed by the tool. Because the specifications for the ASTs are different between `css` and `postcss` the logic needed to be slightly changed. The huge advantage that `postcss` gives for this purpose, in my opinion, is provide OOP-like functions to manipulate the AST instead of interacting directly with its structure.

src/js/components/Menu/__tests__/__snapshots__/Menu-test.js.snap
```diff
@@ -575,11 +575,12 @@ exports[`Menu close by clicking outside 2`] = `
...
 
 exports[`Menu close by clicking outside 3`] = `
-"@media only screen and (max-width:768px) {
+"@media only screen and (max-width: 768px) {
   .hYQclW {
     padding: 6px;
   }
-}"
+}
+"
```
ShimiSun, line 583: Can you adjust the logic on portal.js to avoid this line break on the CSS snapshot style? I don't think it is semantically accurate nor necessary and it will fix many of the snapshot changes.

src/js/utils/portal.js
ShimiSun: I'm ok with leaving it as is. thanks for clarifying.

src/js/components/Menu/__tests__/__snapshots__/Menu-test.js.snap
jecassis: Sure. Rebased against `master` and pushed with the update.

src/js/components/Drop/__tests__/__snapshots__/Drop-test.js.snap
```diff
@@ -1383,7 +1364,7 @@ exports[`Drop plain renders 1`] = `
 `;
 
 exports[`Drop plain renders 2`] = `
-"@media screen and (-ms-high-contrast:active),(-ms-high-contrast:none) {
+"@media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
```
ShimiSun, line 1367: I've verified this additional space is the correct behavior, thanks.

src/js/components/Menu/__tests__/__snapshots__/Menu-test.js.snap
ShimiSun: Perfect, thanks buddy!

jecassis:
True, it is much larger. But it only adds cost to the development environment. The output bundle in my local build mirrors almost to the byte and module count that of the latest commit to master. Meaning that `css` is already a test-only dependency. Whether using `css` or `postcss`, moving either out of `dependencies` makes it so that neither `grommet.min.js` nor downstream consumers using es6 modules would see this increase.

jecassis:
Just to add a point to what I said before. PostCSS is a dependency of Storybook and, as such, Grommet was already including it anyway.

ShimiSun:

Issue #4077, opened 2020-05-13
Default button is causing issues on components that use Button 


britt6612:
With the new changes that were done to `Button` and default this is now effecting other components in which use `Button` under the hood to made up that component
### Expected Behavior
I would expect that if I had certain properties defined in theme for `default button` these would not effect the components that are should be using plain button under the hood.

### Actual Behavior
When you have in your theme 

```
  button: {
    default: {
      color: 'text',
      border: undefined,
      padding: {
        horizontal: '12px',
        vertical: '6px',
      },
    },
```
This will also effect the padding in `accordion` and `select` since these components are using `button` there is unnecessary pad now in which the user would want the plain button to still be used in `accordion`

### Steps to Reproduce

Currently there is a theme story for accordion if you make sure you have the latest master changes and `grommet-theme-hpe-next` changes you will see the hover behavior and the extra padding of 12px horizontal and 6px vertical. 

In `accordionpanel.js` `Button` is being used on line 90 and so if you set that `button` prop plain then that will fix the issue.

I would assume that we would want to ignore these default props given to button if they are being used inside other components. 

### Your Environment

- Grommet version: [Shimi] stable branch
- Browser Name and version: [Shimi] any
- Operating System and version (desktop or mobile): [Shimi] any

britt6612:
adding @halocline comment from Slack thread 

Basically, any components that use Button in their implementations are receiving excess padding (as defined by default in the theme), and also receiving the wrong hover background color (switching light/dark styling).

Menu is another example: https://design-system.hpe.design/components/menu

britt6612:
After some exploration: 

After looking at menu and accordion in the design system I found that with adding the plain prop now to these when default button is in the theme the text color in dark is not correct. 

https://design-system.hpe.design/components/accordion
![Screen Shot 2020-05-18 at 3 34 28 PM](https://user-images.githubusercontent.com/42451602/82261881-14c86480-991d-11ea-9928-c034fe824684.png)

https://design-system.hpe.design/components/menu

![Screen Shot 2020-05-18 at 3 34 50 PM](https://user-images.githubusercontent.com/42451602/82261904-2447ad80-991d-11ea-9d91-e4bddf2fd505.png)

* The text color on these should be white in dark mode

halocline:
Additional related issues: https://github.com/hpe-design/design-system/issues/770, https://github.com/hpe-design/design-system/issues/771

ShimiSun:
[...]

PR #4085, opened 2020-05-15
Changed core components that use Button to apply 'plain' for theme.button.default

ShimiSun:
#### What does this PR do?
With the new button.default theme assignment, the core components that use Button are being impacted by the new styles.
This PR is assigning the `plain` prop for those Buttons in case theme.button.default is defined.

Any changed file.
jest
with the design system
#4077 
nope
no
backwards compatible

### DIFF
--- a/src/js/components/AccordionPanel/AccordionPanel.js
+++ b/src/js/components/AccordionPanel/AccordionPanel.js
@@ -91,7 +91,7 @@ const AccordionPanel = forwardRef(
           role="tab"
           aria-selected={active}
           aria-expanded={active}
-          plain={theme.button.default}
+          plain={!!theme.button.default}
           onClick={onPanelChange}
           onMouseOver={event => {
             setHover(headingColor);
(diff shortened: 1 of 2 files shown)


src/js/components/AccordionPanel/AccordionPanel.js
```diff
@@ -91,6 +91,7 @@ const AccordionPanel = forwardRef(
           role="tab"
           aria-selected={active}
           aria-expanded={active}
+          plain={theme.button.default}
```
ericsoderberghp, line 94: Since plain is boolean, I think this should be `!!theme.button.default`

ShimiSun: Right. 
Though, in case `theme.button.default` is undefined, `!!theme.button.default` will be `false`, and having plain false wouldn't be the desired behavior, so I set it to 
`plain={theme.button.default ? true : undefined}`

ShimiSun: waiting on https://www.githubstatus.com/incidents/3q541nk49qmw to push the fix

src/js/components/Menu/Menu.js
```diff
@@ -217,6 +217,9 @@ const Menu = forwardRef((props, ref) => {
       </Button>
     </Box>
   );
+  console.log(theme.button.default);
```
ericsoderberghp, line 220: remove console.log :)

PR #4094, opened 2020-05-17
fix: header should include JSX div types

ncausey:
The header was set to use `BoxProps` for the type, but it should use `BoxTypes` so that it inherits the JSX `div` types.

#### What does this PR do?
Fixes an issue where the Header component did not extend the general `div` types

I made the type change on my machine and verified that the header props extended div props.

Use the Header component and verify it now extends `div` props.

I was trying to customize the Header and noticed it only had the Box props, without also having div props.

Probably not

Probably not necessary

Backwards compatible


ShimiSun:

PR #4086, opened 2020-05-15
Storybook - add dual color track story for RangeInput

halocline:
Signed-off-by: Matt Glissmann <mdglissmann@gmail.com>

#### What does this PR do?

Adds Storybook story to illustrate how to style `<RangeInput>`'s track so that track can have different colors on either side of the thumb.

#### Where should the reviewer start?
Storybook - `RangeInput/stories/TrackDualColor.js`

#### What testing has been done on this PR?
Storybook

#### How should this be manually tested?
Storybook

#### Any background context you want to provide?
Discussed adding this story with @ShimiSun 

#### What are the relevant issues?
None.

#### Screenshots (if appropriate)
Screen Shot 2020-05-15 at 12 25 45 PM

#### Do the grommet docs need to be updated?
No.

#### Is this change backwards compatible or is it a breaking change?
Backwards compatible

Issue #53, opened 2015-06-24
Submitting form by pressing Enter in a form field


Thylossus:
Hi,

I think there is an issue with submitting forms when pressing enter in a form field when there are two buttons (a submit button and a cancel button). I was able to reproduce this issue using the [FullForm](https://github.com/HewlettPackard/grommet/blob/master/docs/src/documentation/components/samples/FullForm.js) example from the docs.

If the user presses enter within a form, the form is not always submitted. Instead the `onClick` action associated to the first button defined within the form's DOM will be executed, i.e. if the submit button is defined before the cancel button, pressing enter will submit the form. However, if the cancel button is defined before the submit button, pressing enter will 'cancel' the form, i.e. execute the `onClick` action associated with the cancel button.

This issue can be observed by changing 

````
<Footer>
  <Menu direction="right">
    <Button label="OK" primary={true} strong={true} onClick={this.props.onSubmit} />
    <Button label="Cancel" onClick={this.props.onCancel} />
  </Menu>
</Footer>
````

to 

````
<Footer>
  <Menu direction="right">
    <Button label="Cancel" onClick={this.props.onCancel} />
    <Button label="OK" primary={true} strong={true} onClick={this.props.onSubmit} />
  </Menu>
</Footer>
````

in the [FullForm](https://github.com/HewlettPackard/grommet/blob/master/docs/src/documentation/components/samples/FullForm.js) sample. (I have also added log messages to [FormDoc](https://github.com/HewlettPackard/grommet/blob/master/docs/src/documentation/components/FormDoc.js) to see which function is executed).

I know this is not really a grommet specific issue, but maybe a hint in the documentation would be useful because in my opinion adding the `primary` attribute to a button suggests that the action associated with a primary button will be the default action when hitting Enter while filling out the form.

I suppose a solution with `<input type="reset">` and `<input type="submit">` as in [this fiddle](http://jsfiddle.net/ngdqko30/) is not a feasible solution for grommet because it would require dedicated "form buttons". That's why I think a hint in the documentation would be sufficient.

alansouzati:
Fixed

PR #4100, opened 2020-05-20
Changed RangeInput to support bounds

### LINKED ISSUE #2969: Interactive RangeInput
### Expected Behavior
RangeInput should allow a richer functionality by providing the context of value within the thumb progress along the track.

### Actual Behavior
the track has the same L&F before and after the thumb, so there is no indication of value.

drafts from brainstorming 
rangeInput2
rangInput1

### Your Environment
any

- Grommet version: 2.6.5

ShimiSun:
#### What does this PR do?
1. Supports color and opacity styling of upper and lower bounds.
1. Supports color and opacity styling of the regular track (splitting existing behavior where they both were applied under 'color'). The new behavior is backward compatible.
1. Fixes typescript issues, and cleaning up stories.
1. Inspired by https://github.com/grommet/grommet/pull/4086 @halocline 

base.d.ts and doc.js to see the theme API changes and additions
jest - confirming no snapshots broken, and storybook.
storybook - new story.

Closes #4086 
Fixes #2969

done.
Yes
 backwards compatible


src/js/themes/base.js
```diff
@@ -879,14 +877,24 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
     },
     rangeInput: {
+      // extend: undefined
       track: {
         height: '4px',
-        color: css`
-          ${props => rgba(normalizeColor('border', props.theme), 0.2)};
-        `,
+        color: 'border',
+        // opacity: undefined,
```
ericsoderberghp, line 884: Should this be 0.2 to preserve the prior style?

src/js/components/RangeInput/StyledRangeInput.js
```diff
@@ -1,14 +1,75 @@
...
+  // backward compatibility in case no bounds are defined
+  if (
+    props.theme.rangeInput &&
+    props.theme.rangeInput.track &&
+    !props.theme.rangeInput.track.lower &&
+    !props.theme.rangeInput.track.upper
+  ) {
+    const color = rgba(
+      normalizeColor(props.theme.rangeInput.track.color, props.theme),
+      0.2,
```
ericsoderberghp, line 46: I think we should be pulling the opacity from `props.theme.rangeInput.track.opacity`.

src/js/components/RangeInput/StyledRangeInput.js
```diff
@@ -1,14 +1,75 @@
...
+    );
+    if (color === 'rgba(0,0,0,0.2)') return `background: ${color}`;
```
ericsoderberghp, line 48: Why do we need to check for a specific value here? We already know there aren't upper and lower bounds to fall through to.

src/js/components/RangeInput/StyledRangeInput.js
```diff
@@ -1,14 +1,75 @@
...
+  }
+
+  // 'max' defaults to 100 in case not specified
+  const max = props.max || 100;
+  const thumbPosition = `${(props.value / max) * 100}%`;
```
ericsoderberghp, line 53: I think we also need to handle a `min` value.

src/js/components/RangeInput/StyledRangeInput.js
```diff
@@ -1,14 +1,75 @@
...
+};
+
+const getBoundColor = (props, bound) => {
+  if (
+    props.theme.rangeInput &&
+    props.theme.rangeInput.track &&
+    props.theme.rangeInput.track[bound] &&
+    props.theme.rangeInput.track[bound].color
+  ) {
+    return rgba(
```
ericsoderberghp, line 24: Why not pass the opacity into `normalizeColor()` and let it compose the `rgba()`?

src/js/themes/base.js
ShimiSun: I don't think so. 
Assume the user changed the color to `color: 'brand'`, and we will apply an opacity of 0.2 on that color, this will be a breaking change for users that have modified the theme color property.
On StyledRangeInput I added a backward compatibility section that will take the opacity into account. this is why the snapshots are still intact.

src/js/components/RangeInput/StyledRangeInput.js
ShimiSun: I agree, but not on this section, I added a block below that one that will use `props.theme.rangeInput.track.opacity` in case the user changed the color and wanted to use the theme opacity. 
This block is to maintain backward compatibility.

src/js/components/RangeInput/StyledRangeInput.js
ShimiSun: Since we changed the color on base.js from border with opacity to just `border` this condition is used to make sure we are applying the opacity correctly in case the color is indeed the `border` color.
Your comment could have been true if we defaulted the opacity to 0.2, but see my first comment above regarding the concern of adding a default opacity in base.js.

src/js/components/RangeInput/StyledRangeInput.js
ShimiSun: In what way? Not sure what you mean.

src/js/components/RangeInput/StyledRangeInput.js
ShimiSun: I haven't seen how `normalizedColor` can control the opacity, it doesn't look like it is using `rgba`.

ericsoderberghp: Oops, you are correct. My mistake.

src/js/components/RangeInput/StyledRangeInput.js
ericsoderberghp: If `min={10} max={20} value={15}` then `thumbPosition === '75%'` when it should be `'50%'.

src/js/components/RangeInput/StyledRangeInput.js
ericsoderberghp: But, from line 884 in base.js we have `// opacity = undefined`. If I set `theme.rangeInput.track.opacity = 0.4` in my theme, isn't this where it would be used?

src/js/components/RangeInput/StyledRangeInput.js
ericsoderberghp: Can we add a comment explaining this? It seems a bit obscure otherwise.

src/js/components/RangeInput/StyledRangeInput.js
ShimiSun: It will be used in line #53.

src/js/components/RangeInput/StyledRangeInput.js
ShimiSun: Good call! thank you

src/js/components/RangeInput/StyledRangeInput.js
ShimiSun: Done

src/js/components/RangeInput/StyledRangeInput.js
ShimiSun: Done.

PR #4106, opened 2020-05-21
NEW Card component

ShimiSun:
#### What does this PR do?
Adds Card component to the component library.
I'm open to feedback, especially regarding the assignment of default values on the theme.

Design system Card https://www.figma.com/file/WzYBJncJEbjA2JU9GqZMZB/HPE-Cards?node-id=0%3A1
doc.js for API
storybook + jest
storybook
Moving away from aries-core
#3489 

done.
Of course. bumping up the 'minor' version.
backwards compatible

### DIFF
--- a/src/js/components/Card/Card.js
+++ b/src/js/components/Card/Card.js
@@ -1,0 +1,41 @@
+import React, { forwardRef, useContext } from 'react';
+
+import { ThemeContext } from 'styled-components';
+import { defaultProps } from '../../default-props';
+
+import { Box } from '../Box';
+import { Header } from '../Header';
+import { Footer } from '../Footer';
+
+export const Card = forwardRef(({ children, header, footer, ...rest }) => {
+  const theme = useContext(ThemeContext) || defaultProps.theme;
+
+  return (
+    <Box align="start" justify="between" {...theme.card} {...rest}>
+      {header && (
+        <Header
+          id="shimi"
+          round={{ size: theme.card.round, corner: 'top' }}
+          fill="horizontal"
+          overflow="hidden"
+          {...theme.card.header}
+        >
+          {header}
+        </Header>
+      )}
+      <Box flex {...theme.card.content}>
+        {children}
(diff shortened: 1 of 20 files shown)


src/js/themes/base.js
```diff
@@ -511,6 +511,26 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       },
       heading: { level: '4' }, // level ranges from 1-6
     },
+    // add to index.d.ts
```
ShimiSun, line 514:
```suggestion

```

src/js/components/Card/Card.js
```diff
@@ -0,0 +1,53 @@
...
+import { Header } from '../Header';
+import { Footer } from '../Footer';
+
+export const Card = forwardRef(({ children, header, footer, ...rest }) => {
+  const theme = useContext(ThemeContext) || defaultProps.theme;
+
+  return (
+    <Box align="start" justify="between" {...theme.card} {...rest}>
+      {header && (
+        <Header
```
ericsoderberghp, line 17: Hmm... something feels a bit odd here. What if the caller wants to change the `justify` on the Header? And, passing multiple contents via something like `<Card header={[<Avatar />, <Button />]} />` seems heavy

I think we are wanting Card to have specific default Header and Footer styles but allow them to be customized where needed. 

What about treating Card more like Table?:
```
<Card>
  <CardHeader />
  <CardBody />
  <CardFooter />
 </Card>
```
Where CardBody is only needed if you have CardHeader or CardFooter.
Just thinking.

erickriva:
I'm trying to develop some ready-to-use components using Grommet, one of them is a Card.
Maybe my implementation can help to think in different ways.

[Link to Codesandbox](https://codesandbox.io/s/grommet-test-4qgos?file=/src/components/general/Card/index.tsx)

ShimiSun:
Great feedback!! I filed a new PR that adapts the suggested API. Thank you!

PR #4120, opened 2020-05-27
Changed Grid to restrict repeated column minimum size to not exceed total width

ericsoderberghp:
#### What does this PR do?

Changed Grid to restrict repeated column minimum size to not exceed total width.

A frequent design goal is to have a Grid of cards that responds to the available width. The problem comes when the available width is less than the minimum column size specified. You could reduce the minimum column size, but then that doesn't always look right for larger displays. This resolves that.

StyledGrid.js

storybook, grommet-designer

storybook

no

yes

backwards compatible


ericsoderberghp:
I can't figure out why Chromatic is calling out changes. When I run those stories locally, they aren't changed.

ShimiSun:
> I can't figure out why Chromatic is calling out changes. When I run those stories locally, they aren't changed.

Try to see if it changes on IE, FF or Safari.

Regardless, I think the Animation story should be excluded from Chromatic since Chromatic doesn't capture animations anyway.

ericsoderberghp:
> > I can't figure out why Chromatic is calling out changes. When I run those stories locally, they aren't changed.
> 
> Try to see if it changes on IE, FF or Safari.
> 
> Regardless, I think the Animation story should be excluded from Chromatic since Chromatic doesn't capture animations anyway.

I tested the stories Chromatic complained about locally with Safari, Firefox, and Chrome and on Windows 10 with Edge and Chrome via browserstack.com. All appeared as expected. So, not sure what's up with Chromatic.

ShimiSun:
From taking a quick look, I was able to reproduce the Grid story layout issue on Chrome via the netlify deployment. I haven't tried the other examples.

image

ericsoderberghp:
> From taking a quick look, I was able to reproduce the Grid story layout issue on Chrome via the netlify deployment. I haven't tried the other examples.

I was able to reproduce this one, and have addressed it with the latest commit. However, I still can't reproduce the other Chromatic issues.

taysea:
> > From taking a quick look, I was able to reproduce the Grid story layout issue on Chrome via the netlify deployment. I haven't tried the other examples.
> 
> I was able to reproduce this one, and have addressed it with the latest commit. However, I still can't reproduce the other Chromatic issues.

Yes this is strange. I'm not able to reproduce either. The changes look like `Grid` is not being applied at all

ShimiSun:
It is not typical, this is why we hesitated to merge this PR.
I'm thinking that maybe if the branch will be up to date with master and chromatic will trigger a new round of testing it may have an impact on the rendering, will you care to give it a try?
Also, check on other browsers than Chrome to verify it displays as expected.
I'm thinking maybe Chromatic has a delay and doesn't capture the final layout, which may be worth the investigation as well.

PR #4124, opened 2020-05-28
Updated Build script

JoeButler7:
Updated build script to exclude test files and portal.js from distribution folder

#### What does this PR do?
 Removes test files and portal.js from dist folder

`package.json` line 32

I have manually tested it by running `yarn build` and checking dist folder to confirm that the right files were excluded
Run `yarn build` and then browse through the `grommet/dist` directory to confirm that there are no test files included

#4118 and #4119 

### Relevant Screenshots
`grommet/dist` directory before updating the script 
Screen Shot 2020-05-28 at 3 54 19 PM
`grommet/dist` directory  after updating script
Screen Shot 2020-06-01 at 4 19 34 PM

`grommet/dist/utils` directory before updating script
Screen Shot 2020-05-28 at 3 54 34 PM
`grommet/dist/utils` directory after updating script
Screen Shot 2020-06-01 at 4 19 57 PM

No
No
It is backwards compatible


ShimiSun:
Can you please attach a screenshot of the dist directory before and after this change?

JoeButler7:
@ShimiSun Yup just added them

package.json
```diff
@@ -29,7 +29,7 @@
   },
   "scripts": {
     "generate-readme": "cross-env NODE_ENV=production babel-node ./tools/generate-readme",
-    "build": "webpack --mode production && babel ./src/js/ --ignore '__tests__' --out-dir ./dist --copy-files && cross-env BABEL_ENV=es6 babel ./src/js/ --ignore '__tests__' --out-dir ./dist/es6 --copy-files",
+    "build": "webpack --mode production && babel ./src/js/ --out-dir ./dist --copy-files --ignore '**/**/__tests__' --ignore 'src/**/portal.js' --no-copy-ignored && cross-env BABEL_ENV=es6 babel ./src/js/ --out-dir ./dist/es6 --copy-files --ignore '**/**/__tests__' --ignore 'src/**/portal.js' --no-copy-ignored && rm -r ./dist/__tests__/ ./dist/**/__tests__/ ./dist/**/**/__tests__",
```
ShimiSun, line 32: why do we need 
```
rm -r ./dist/__tests__/ ./dist/**/__tests__/ ./dist/**/**/__tests__",
```
If the tests shouldn't be included in `dist` from the first place.

JoeButler7: Not sure if it was a bug or something that we were missing but there was no way to ignore/exclude the snapshot files effectively. The `__tests__` directories were still being included but all they had in them were the snapshots so we figured this was the best way to get rid of the snapshot files from the dist folder

ShimiSun: We shouldn't use the `rm` command.
We should ignore the directory and the files without adding and removing it on the distribution package.

ShimiSun: Coping our conversation thread for more transparncy:
>The issue seems to be that the --no-copy-ignore flag only applies to js files so the snapshot files are being ignored but the --copy-files flag is still copying them to the dist folder since they're of the form *.js.snap. The workarounds I have found have either used a different package to do it (usually jest-gulp) or use a rm/del command to get rid of the snap files
>Yeah the babel CLI is pretty limited with alot of its options and it doesn't seem like theres any other flags that would be applicable
the docs for the CLI are here https://babeljs.io/docs/en/babel-cli

package.json
```diff
@@ -29,7 +29,7 @@
   },
   "scripts": {
     "generate-readme": "cross-env NODE_ENV=production babel-node ./tools/generate-readme",
-    "build": "webpack --mode production && babel ./src/js/ --ignore '__tests__' --out-dir ./dist --copy-files && cross-env BABEL_ENV=es6 babel ./src/js/ --ignore '__tests__' --out-dir ./dist/es6 --copy-files",
+    "build": "webpack --mode production && babel ./src/js/ --out-dir ./dist --copy-files  && cross-env BABEL_ENV=es6 babel ./src/js/ --out-dir ./dist/es6 --copy-files && rm -r ./dist/utils/portal.js ./dist/es6/utils/portal.js ./dist/__tests__/ ./dist/**/__tests__/ ./dist/**/**/__tests__",
```
ShimiSun, line 32:
```suggestion
    "build": "webpack --mode production && babel ./src/js/ --out-dir ./dist --copy-files && cross-env BABEL_ENV=es6 babel ./src/js/ --out-dir ./dist/es6 --copy-files && rm -r ./dist/utils/portal.js ./dist/es6/utils/portal.js ./dist/__tests__/ ./dist/**/__tests__/ ./dist/**/**/__tests__",
```

Removing redundant spacing.

ShimiSun:
Thinking it through, I still don't feel like the `rm -r` approach is the right one, and that those files shouldn't be included in from the first place. 
I would like us to explore other packages, and whether they provide commands that can help us achieve the goal without using `rm`.

JoeButler7:
We have decided to use the `copyfiles` node package instead of `rm -r` to achieve the right files in the `dist` directory so we are closing this pull request

PR #4135, opened 2020-06-01
Add alignControls and tabs.header.border theming options to Tabs

taysea:
#### What does this PR do?
This PR adds `tabs.header.border` to theming options and a prop of `alignControls` to the Tabs component. Both would allow for broader control for all grommet users, but was prompted specifically by the detail of this design system issue: https://github.com/hpe-design/design-system/issues/789

The additional theme control allows and border to be applied on the overall Tabs container which allows, in the case of the above noted issue, a border to extend across the entire width of the tabs. The addition of `alignControls` is way to give users control of the tabs alignment without restricting the alignment/width of the tabs content.

src/js/components/Tabs/Tabs.js

A new storybook has been added demonstrating the effects and a jest snapshot for `alignControls` has also been added.

Apply a theme value such as:
```
tabs: {
   header: {
      border: {
         side: 'bottom',
         color: 'red',
         size: 'small',
      }
   }
}
```

Apply `alignControls="center"` to tabs.

Needed to solve this design system issue: https://github.com/hpe-design/design-system/issues/789

https://github.com/hpe-design/design-system/issues/789

Sure, if helpful!

Yes, backwards compatible.


src/js/components/Tabs/doc.js
```diff
@@ -90,6 +96,15 @@ export const themeDoc = {
     type: 'string | { dark: string, light: string }',
     defaultValue: undefined,
   },
+  'tabs.header.border': {
```
ShimiSun, line 99: For consistency with Box border, any reason we don't allow the `style` of the border (dash, etc...)

src/js/components/Tabs/Tabs.js
```diff
@@ -75,6 +77,19 @@ const Tabs = forwardRef(
...
 
+    const tabsHeaderStyles = {};
+    if (theme.tabs.header && theme.tabs.header.border) {
+      let borderColor =
+        theme.tabs.header.border.color || theme.global.control.border.color;
+      borderColor = normalizeColor(borderColor, theme);
+
+      tabsHeaderStyles.border = {
+        side: theme.tabs.header.border.side,
+        size: theme.tabs.header.border.size,
```
ShimiSun, line 88: You may have a case where `theme.tabs.header.border` exist, but `theme.tabs.header.border.size` doesn't. Same for `side`.

src/js/components/Tabs/doc.js
taysea: No reason to exclude. I had pulled the logic from `Tab` which wasn't including style, but I think it's a valid thing to have since it's included on Box. I will add it here.

src/js/components/Tabs/Tabs.js
taysea: Yes, in these cases the values would just become undefined. which is fine right?

src/js/components/Tabs/doc.js
taysea: Style has been added.

src/js/components/Tabs/Tabs.js
ShimiSun: Not sure how a border that has color but no size/side will react, it doesn't sound right to me. Can you compare this use case with Box border and make sure it is consistent?

taysea: Yes, let me check it out

taysea: If any of these values is undefined, then it will fallback to the defaults defined in `borderStyle` in `border.js` (consistent with Box):
```
const color = normalizeColor(data.color || 'border', theme);
const borderSize = data.size || 'xsmall';
const style = data.style || 'solid';
const side = typeof data === 'string' ? data : data.side || 'all';
```
https://github.com/grommet/grommet/blob/master/src/js/utils/border.js#L7

This is because `StyledTabsHeader` is a `Box`, so the behavior matches that of Box.

src/js/components/Tabs/doc.js
```diff
@@ -22,6 +22,12 @@ export const doc = Tabs => {
 This means that future tab changes will not work unless you subscribe to
 onActive function and update activeIndex accordingly.`,
     ),
+    alignTabs: PropTypes.oneOf([
```
ShimiSun, line 25: what about `alignControls` or something in that direction?

taysea: I am fine with that.

src/js/components/Tabs/stories/AlignControls.js
```diff
@@ -0,0 +1,71 @@
...
+        <Box fill pad="large" align="center">
+          <TreeOption size="xlarge" />
+        </Box>
+      </Tab>
+      <Tab title="Tab 3">
+        <Box fill pad="large" align="center">
+          <Car size="xlarge" />
+        </Box>
+      </Tab>
+    </Tabs>
```
ShimiSun, line 49: I think we can remove this portion of the example, looks awkward 
image

src/js/components/Tabs/stories/AlignControls.js
```diff
@@ -0,0 +1,71 @@
...
+  </Grommet>
+);
+
+storiesOf('Tabs', module).add('alignControls', () => <AlignControls />);
```
ShimiSun, line 71:
```suggestion
storiesOf('Tabs', module).add('Align Controls', () => <AlignControls />);
```

src/js/components/Tabs/stories/AlignControls.js
taysea: Sounds good. Removed it.

ShimiSun:
Thank you dear! 💟

PR #4145, opened 2020-06-04
Changed DataChart to improve axes

ericsoderberghp:
#### What does this PR do?

Changed DataChart to improve axes.

- truncate y-axis values to avoid long fractions
- added prefix and suffix for axis values, to avoid the need for a custom renderer in typical situations
- y-axis with key will check for a date format and use “right” default date format for the span of dates
- align y-axis labels better with guide lines

DataChart.js

enhanced unit tests, enhanced stories

storybook

no

yes

This has a breaking change for the `yAxis.render()` signature. It is now called with `(content, data, dataIndex, axisIndex)` instead of `(dataIndex, axisIndex)`.

### DIFF
--- a/src/js/components/Chart/__tests__/__snapshots__/Chart-test.js.snap
+++ b/src/js/components/Chart/__tests__/__snapshots__/Chart-test.js.snap
@@ -9,6 +9,7 @@ Object {
     ],
     Array [
       3.2,
+      0.8,
     ],
   ],
   "bounds": Array [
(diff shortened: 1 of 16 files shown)


ShimiSun:
>This has a breaking change for the yAxis.render() signature. It is now called with (content, data, dataIndex, axisIndex) instead of (dataIndex, axisIndex).

The component was never officially released yet, so I won't put it on the release notes but I will communicate it in person.

Not sure about the test failures on this PR.

ShimiSun:
Looking at this example:
image

I think the values of the bars are not aligned with the x-axis lines, the bar is overflowing the x-axis on both the min and max values, makes me think there is an accuracy issue here.

PR #4151, opened 2020-06-05
TS: Add extend type to text

felipearosemena:
`text.extend` is not defined in the `ThemeType` definition, causing TS compile to throw an error.

#### What does this PR do?

Adds type definition for the themes `text.extend` 

`src/js/themes/base.d.ts` -> `ThemeType`

Tested locally

Using typescript, attempt to add the following theme:

```
 {
  text: {
    extend: (props: any) => {
      const { value } = props.theme.global.breakpoints.small;
      return `
        @media (max-width: ${value}px) {
          font-size: 16px
        }
      `;
    },
  },
};
```

they already reference this property, so no

https://v2.grommet.io/text#text.extend

### DIFF
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -984,6 +984,7 @@ export interface ThemeType {
     };
   };
   text?: {
+    extend?: ExtendType;
     xsmall?: {
       size?: string;
       height?: string;


felipearosemena:
Thanks! Loving the library ❤️

PR #4152, opened 2020-06-05
Changed RadioButtonGroup to support numeric values

ericsoderberghp:
#### What does this PR do?

Changed RadioButtonGroup to support numeric values.

This allows setting something to `5` instead of `"5"`.

doc.js

added unit test

see unit test

automatic

yes

backwards compatible

### DIFF
--- a/src/js/components/RadioButtonGroup/index.d.ts
+++ b/src/js/components/RadioButtonGroup/index.d.ts
@@ -5,7 +5,7 @@ export interface RadioButtonGroupProps {
   disabled?: boolean;
   name: string;
   onChange?: ((event: React.ChangeEvent<HTMLInputElement>) => void);
-  options: (string | { disabled?: boolean, id?: string, label?: (string | React.ReactNode), value: string | number })[];
+  options: (string | number | { disabled?: boolean, id?: string, label?: (string | React.ReactNode), value: string | number })[];
   value?: string | number | object;
 }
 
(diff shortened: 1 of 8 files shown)


src/js/components/RadioButtonGroup/index.d.ts
```diff
@@ -5,8 +5,8 @@ export interface RadioButtonGroupProps {
   disabled?: boolean;
   name: string;
   onChange?: ((event: React.ChangeEvent<HTMLInputElement>) => void);
-  options: (string | { disabled?: boolean, id?: string, label?: (string | React.ReactNode), value: string})[];
-  value?: string | object;
+  options: (string | { disabled?: boolean, id?: string, label?: (string | React.ReactNode), value: string | number })[];
+  value?: string | number | object;
```
ShimiSun, line 9: Should we remove the `object` option?

ericsoderberghp: nope, we still allow options to contain objects. But, I reviewed the typescript types again and found another issue, fixed.

ShimiSun: In this case, should it be added to PropTypes in doc.js?

ericsoderberghp: update to add value as number to prop types

PR #4163, opened 2020-06-09
Changed Button to add option kind for Menu, Select, TextInput, etc. - DO NOT MERGE

ericsoderberghp:
#### What does this PR do?

Changed Button to add option kind for Menu, Select, TextInput, etc.

Added `option` kind of Button. Use it in drops.

Menu.js

Added internal `hpe` theme to AllComponents story.

All Components story with `hpe` theme.

yes

yes

backwards compatible

### DIFF
--- a/src/js/components/Button/StyledButtonKind.js
+++ b/src/js/components/Button/StyledButtonKind.js
@@ -173,12 +173,13 @@ const kindStyle = ({ colorValue, themePaths, theme }) => {
 const hoverIndicatorStyle = ({ hoverIndicator, theme }) => {
   const themishObj = {};
   if (hoverIndicator === true || hoverIndicator === 'background')
-    themishObj.background = theme.global.hover;
+    themishObj.background = theme.global.hover.background;
   else themishObj.background = hoverIndicator;
   const styles = kindPartStyles(themishObj, theme);
   if (styles.length > 0)
     return css`
       &:hover {
+        overflow: visible;
         ${styles}
       }
     `;
@@ -226,6 +227,12 @@ const StyledButtonKind = styled.button`
   ${props => props.plain && plainStyle(props)}
   ${props => !props.plain && basicStyle(props)}
   ${props => !props.plain && kindStyle(props)}
+  ${props =>
+    !props.plain &&
+    props.align &&
+    `
+  text-align: ${props.align};
+  `}
   ${props => props.hoverIndicator && hoverIndicatorStyle(props)}
   ${props =>
     props.focus && (!props.plain || props.focusIndicator) && focusStyle()}
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -188,7 +188,7 @@ const Menu = forwardRef((props, ref) => {
       direction="row"
       justify={justifyContent}
       align="center"
-      pad="small"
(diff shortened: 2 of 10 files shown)


taysea:
A couple things I noticed is that the keyboard styling is resolved for textinput suggestions, but there is still no styling for keyboard on Select as I use arrow keys to move down. Also, Select dropdown items need to be 36px tall, but they're only 34px. However, Menu and TextInput suggestions are the correct dimensions.

src/js/components/Button/StyledButtonKind.js
```diff
@@ -226,6 +227,12 @@ const StyledButtonKind = styled.button`
   ${props => props.plain && plainStyle(props)}
   ${props => !props.plain && basicStyle(props)}
   ${props => !props.plain && kindStyle(props)}
+  ${props =>
+    !props.plain &&
+    props.align &&
+    `
+  text-align: ${props.align};
```
ShimiSun, line 234: Since `align` underlying is resolved to `text-align` I'm thinking we might want to communicate better the name of the prop as `textAlign` rater than `align`.

ericsoderberghp:
> In case we will proceed with the direction of this PR, we need to add:
> 
> 1. `align` and and `options` to docs props and theme correspondingly.
> 2. `align` to the typescript interface.

Not necessarily, we could leave them as internal-only properties for the time being, like `kind`.

src/js/components/Button/StyledButtonKind.js
ericsoderberghp: So, the trick here is that when the Button contains only text, we need `text-align` but when there is an icon and a label, we actually set `justify` on the Box that contains them. So, `align` here is more about the visual manifestation and less about the underlying implementation, which varies based on context.

ShimiSun:
> Not necessarily, we could leave them as internal-only properties for the time being, like kind.

In that case, it should be consistent and `option` should be removed from base.d.ts.

src/js/components/Menu/Menu.js
```diff
@@ -278,16 +281,18 @@ const Menu = forwardRef((props, ref) => {
...
-                        gap={item.gap}
-                      >
-                        {item.reverse && item.label}
-                        {item.icon}
-                        {!item.reverse && item.label}
-                      </Box>
+                      {!theme.button.default && (
+                        <Box
+                          align="start"
+                          pad={theme.button.default ? undefined : 'small'}
```
ShimiSun, line 287:
```suggestion
                          pad="small"
```

src/js/components/Button/StyledButtonKind.js
```diff
@@ -173,12 +173,13 @@ const kindStyle = ({ colorValue, themePaths, theme }) => {
...
   const themishObj = {};
   if (hoverIndicator === true || hoverIndicator === 'background')
-    themishObj.background = theme.global.hover;
+    themishObj.background = theme.global.hover.background;
   else themishObj.background = hoverIndicator;
   const styles = kindPartStyles(themishObj, theme);
   if (styles.length > 0)
     return css`
       &:hover {
+        overflow: visible;
```
ShimiSun, line 182: Should be deleted.

ericsoderberghp: agreed

PR #4170, opened 2020-06-10
Fixed video src on stories

JoeButler7:
#### What does this PR do?
Updates src on  video stories to fix chrome bug with mixed content. It changes the src on the video in stories to HTTPS site instead of HTTP,  which removes the mixed content error from storybook in chrome. Error was happening because the site was loaded with HTTPS but video source was requesting HTTP, and the site that the videos were hosted on did not support upgrading the connection to HTTPS.
`src/js/components/videos/stories/Simple.js` and `src/js/components/videos/stories/typescript/Simple.tsx`
Ran `yarn storybook` with updated video src and used the video component in storybook. The video played fine, no mixed content error. 
See above
Previously in storybook on chrome the videos were not loading because the storybook site is loaded through HTTPS but the video src was requesting from HTTP. This was causing mixed content, so chrome was automatically upgrading the video request to HTTPS but the site that the videos were hosted on did not support HTTPS so the videos would not load as the browser could not connect to the site. 
No
No
Backwards compatible


JoeButler7:
### Update:
We have decided to use a local file `small.mp4`, which was added to the directory `grommet/storybook/public`, as the source for the video in storybook

src/js/components/Video/stories/Simple.js
```diff
@@ -8,22 +8,7 @@ const SimpleVideo = props => (
...
-        />
-        <source
-          src="http://techslides.com/demos/sample-videos/small.mp4"
-          type="video/mp4"
-        />
-        <source
-          src="http://techslides.com/demos/sample-videos/small.3gp"
-          type="video/3gp"
-        />
+        <source src="small.mp4" type="video/mp4" />
```
ShimiSun, line 11: I'm wondering if we should keep the other source tags since it's optional anyway and it will demonstrate better how the tag could be used.
```
The <source> element allows you to specify alternative video files which the browser may choose from. The browser will use the first recognized format.
```
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source

JoeButler7: Yeah I agree, I will add it in the next commit

PR #4174, opened 2020-06-10
Changed Button so kind styling can have radius and base border can be omitted

ericsoderberghp:
#### What does this PR do?

Changed Button so kind styling can have radius and base border can be omitted.

StyledButtonKind.js

custom app

no

yes

backwards compatible

### DIFF
--- a/src/js/components/Button/StyledButtonKind.js
+++ b/src/js/components/Button/StyledButtonKind.js
@@ -15,7 +15,9 @@ const radiusStyle = props => {
       border-radius: ${props.theme.button.size[size].border.radius};
     `;
   if (props.theme.button.border && props.theme.button.border.radius)
-    return css`border-radius: props.theme.button.border.radius};`;
+    return css`
+      border-radius: ${props.theme.button.border.radius};
+    `;
   return '';
 };
 
--- a/src/js/components/Button/StyledButtonKind.js
+++ b/src/js/components/Button/StyledButtonKind.js
@@ -10,10 +10,13 @@ import { defaultProps } from '../../default-props';
 
 const radiusStyle = props => {
   const size = props.sizeProp;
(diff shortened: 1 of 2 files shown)


src/js/components/Select/Select.js
```diff
@@ -231,7 +231,9 @@ const Select = forwardRef(
               {children}
             </SelectContainer>
           }
-          plain={plain}
+          // StyledDropButton needs to know if the border should be shown
+          callerPlain={plain}
```
ShimiSun, line 235: I'm probably missing something, but why do we need `callerPlain` if it performs the same as `plain`.

src/js/components/Button/StyledButtonKind.js
```diff
@@ -106,6 +111,10 @@ const kindPartStyles = (obj, theme, colorValue) => {
           theme,
         )};
       `);
+    if (obj.border.radius)
+      styles.push(css`
+        border-radius: ${obj.border.radius};
+      `);
```
ShimiSun, line 117: If we are already providing `border-radius` styling from `basicStyles`, why is this needed in addition? we might end up with two css values of `border-radius`?

src/js/components/Select/Select.js
ericsoderberghp: Because StyledDropButton and Button both see the same properties. We want Button to do one thing but StyledDropButton to do something else.

src/js/components/Button/StyledButtonKind.js
ericsoderberghp: We often have two CSS rules, last one wins. This allows each kind to have a different radius.

src/js/components/Select/Select.js
ShimiSun: I think I got it. tricky.

Issue #4178, opened 2020-06-11
Selection visualization within Select with searchPlaceholder is wrong


ecircenis:
### Expected Behavior

When opening a select with a searchPlaceholder, I should clearly see the selected item.

### Actual Behavior
This is showing an expanded Select with option 1 selected:
![image](https://user-images.githubusercontent.com/17883783/84438881-7ade9980-abf4-11ea-8d60-49822f0afc97.png)

This is showing an expanded Select (with searchPlaceholder) and option 1 selected:
![image](https://user-images.githubusercontent.com/17883783/84438939-92b61d80-abf4-11ea-8faa-5160d4ca9021.png)

The selection highlighting is incorrect in the second case.  As soon as I hover over any of the items in the second case, the selection highlighting changes to be correct.

ShimiSun:
I think this issue is fixed on the Recent release (2.14.0), can you please verify?

ShimiSun:
Confirmed this was fixed on the recent release, closing.

PR #4179, opened 2020-06-12
Changed Select to fix an issue with onSearch being called multiple times

ericsoderberghp:
#### What does this PR do?

Changed Select to fix an issue with onSearch being called multiple times.

The management of `activeIndex` was too convoluted and too different between onSearch and no onSearch cases. Previously, if onSearch was specified, an effect would initialize activeIndex differently than if it wasn't. So, keyboard navigation was inconsistent. Now, we only set activeIndex based on any `search` text, which aligns the keyboard behavior better.

Also, uncovered that `onSearch` was being called repeatedly due to the effect with a dependency on it. So, callers that didn't use `useCallback()` would see repeated calls. Now, we only call `onSearch` directly in the onChange handler.

Also fixed a linter warning in Button.

Button.js, for a lightweight warm up ;)

whitebox by adding onSearch to Select in All Components story
Also ran through Select stories.

Storybook

#4178 

no

yes

backwards compatible

### DIFF
--- a/src/js/components/Button/Button.js
+++ b/src/js/components/Button/Button.js
@@ -173,7 +173,7 @@ const Button = forwardRef(
         }
       }
       return result;
-    }, [active, disabled, kind, plain]);
+    }, [active, disabled, kind, plain, selected]);
 
     // only used when theme does not have button.default
     const isDarkBackground = () => {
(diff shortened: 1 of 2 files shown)


ShimiSun:
Labeled `interns` since we need more testing for Select active index.

Issue #4177, opened 2020-06-11
Clean up all dev warnings and errors on running`yarn test`


ShimiSun:
### Expected Behavior
No warnings and errors

### Actual Behavior
Existing errors and warnings that need to be addressed.

### Steps to Reproduce

1. run `yarn test` 
2. view the warnings and errors

### Your Environment
Any

ShimiSun:
Excellent work on this one! Can we close it?

jcfilben:
Yes this issue can be closed

PR #4201, opened 2020-06-17
fixed warnings in Anchor test

jcfilben:
### What does this PR do?
Fixes warnings on `yarn test` for Anchor-test.js. This PR has no effect on the overall testing coverage. Also updated the tests to use render instead of renderer.create which is what causes the changes in the snapshots.

Look at the warnings that were previously being generated from running yarn test. Run this command again on this PR and there won't be warnings for Anchor-test.js.

Running yarn test, looking at the snapshots for 'warns about invalid icon' and 'warns about invalid label', and checking that the coverage stayed the same.

^ see above

#4177

Screen Shot 2020-06-17 at 11 33 54 AM

No

No

Backwards compatible

### DIFF
--- a/src/js/components/Anchor/__tests__/Anchor-test.js
+++ b/src/js/components/Anchor/__tests__/Anchor-test.js
@@ -32,41 +32,43 @@ describe('Anchor', () => {
   });
 
   test('warns about invalid label render', () => {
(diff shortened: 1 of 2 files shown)


ShimiSun:
The logs look good, thank you!

PR #4205, opened 2020-06-17
Fixed warnings in Form test

### LINKED ISSUE #4166: Form Testing for controlled and uncontrolled states.

jcfilben:
#### What does this PR do?
Fixes warnings on `yarn test` for Form-test.js. This PR has no effect on the overall testing coverage.

Look at the warnings that were previously being generated from running yarn test. Run this command again on this PR and there won't be warnings for Form-test.js.

Explaination and solution for where the errors were coming from on this component
https://stackoverflow.com/questions/47012169/a-component-is-changing-an-uncontrolled-input-of-type-text-to-be-controlled-erro

Running `yarn test` to see if the warnings were there for Form-test.js

#4177 

Screen Shot 2020-06-17 at 2 50 10 PM

No

No

Backwards compatible


ShimiSun:
@jcfilben I think this PR shouldn't close #4166 
There should be a separate PR to handle the use cases of adding tests on Form with controlled/uncontrolled cases.

jcfilben:
Updated so that the PR no longer closes issue #4166

src/js/components/Form/__tests__/Form-test.js
```diff
@@ -96,7 +96,7 @@ describe('Form', () => {
   test('controlled lazy', () => {
     const onSubmit = jest.fn();
     const Test = () => {
-      const [value, setValue] = React.useState({});
+      const [value, setValue] = React.useState({ test: '' });
```
ShimiSun, line 99: Few words why this change fixes the warnings?

jcfilben: If the value is undefined the input is uncontrolled. Then in line 100 when we try to assign a value to it the input becomes controlled which causes the warning: `A component is changing an uncontrolled input of type undefined to be controlled.` Setting test to an empty string initially makes the input controlled instead of uncontrolled.

PR #4214, opened 2020-06-18
Video - added `messages` prop to support accessibility standards

### LINKED ISSUE #4176: Video to support accessibility standards
### Expected Behavior
In `Video-test.js` the `Video` component should have no accessibility violations and support a11yTitle

### Actual Behavior
The `Video` component does have accessibility violations due to a messing `aria-label`. When you try and use `a11yTitle` instead of `aria-label` there is still violations because `Video` does not support `a11yTitle`

Screen Shot 2020-06-11 at 4 46 49 PM
Screen Shot 2020-06-11 at 4 46 14 PM

### Steps to Reproduce
Run `yarn Video-test`, add a11yTitle to component with accessibility violations and rerun `yarn test Video-test`
### Your Environment

- Grommet version: 2.13.0
- Operating System and version (desktop or mobile): macOS catalina 10.15.4

JoeButler7:
#### What does this PR do?
Adds accessibility to the video component. Adds a `messages` prop that contains labels for each input element in the video. I did not add a `a11yTitle` to the video itself because I did not think it was necessary as its generally not done (https://developer.mozilla.org/en-US/docs/Learn/Accessibility/Multimedia) and not listed as something that is needed when making a video accessible (https://www.washington.edu/accessibility/videos/). Also it looked like generally `aria-label` (which the `a11yTitle` sets) do not work well with videos https://www.powermapper.com/tests/screen-readers/media/video-aria-label/
`src/js/components/Video/Video.js`
I ran `yarn test Video-test` to ensure that the video still passes all tests and is accessible. 
See above
Closes #4176 
Yes
Yes
backwards compatible


src/js/components/Video/__tests__/Video-test.js
```diff
@@ -21,6 +24,12 @@ describe('Video', () => {
 
   afterEach(cleanup);
 
+  test('Video has not accessibility violations', async () => {
```
jcfilben, line 27:
```suggestion
  test('should have no accessibility violations', async () => {
```

src/js/components/Video/Video.js
```diff
@@ -413,6 +441,33 @@ const Video = forwardRef(
...
+    closeMenu: PropTypes.string,
+    playButton: PropTypes.string,
+    pauseButton: PropTypes.string,
+    volumeUp: PropTypes.string,
+    volumeDown: PropTypes.string,
+    fullScreen: PropTypes.string,
+  }),
+};
+Video.defaultProps = {
+  messages: {
```
ShimiSun, line 458: The messages will be translated to `aria-label` which usually uses a lower case syntax.

src/js/components/Video/Video.js
```diff
@@ -7,8 +7,9 @@ import React, {
   useRef,
   useState,
 } from 'react';
-import { ThemeContext } from 'styled-components';
 
+import PropTypes from 'react-desc/lib/PropTypes';
```
ShimiSun, line 11:
```suggestion
import { PropTypes } from 'react-desc';
```

src/js/components/Video/Video.js
```diff
@@ -413,6 +441,33 @@ const Video = forwardRef(
   },
 );
 
+Video.propTypes = {
```
ShimiSun, line 444: Can we move the propTypes to the doc.js file in order to be consistent with the rest of the components?

src/js/components/Video/Video.js
```diff
@@ -52,6 +53,7 @@ const Video = forwardRef(
       loop,
       margin,
       mute,
+      messages,
```
ShimiSun, line 56: Move to line 55 for alphabetical order?

JoeButler7:
### Update:
Moved `PropTypes` to `doc.js` and `index.d.ts`. Also changed default messages to labels

src/js/components/Video/index.d.ts
```diff
@@ -10,6 +10,18 @@ export interface VideoProps {
...
+    closeMenu?: string,
+    fullScreen?: string,
+    meter?: string,
+    openMenu?: string};
+    scrubber?: string,
+    pauseButton?: string,
+    playButton?: string,
+    volumeUp?: string,
+    volueDown?: string,
+
```
ShimiSun, line 23: are we missing a close for the curly braces that are opened on line 13? I think the closes on line 17 as a mistake.

src/js/components/Video/Video.js
```diff
@@ -413,6 +440,20 @@ const Video = forwardRef(
   },
 );
 
+Video.defaultProps = {
+  messages: {
+    closeMenu: 'close menu',
+    fullScreen: 'full screen',
+    meter: 'video progress',
```
ShimiSun, line 447: Any reason we picked the name `meter`? would it make more sense to call it progress?

JoeButler7: No reason really, I agree that it would make more sense to call it `progressMeter`

JoeButler7:
### Update:
Changed `meter` message to `progressMeter` to improve semantic meaning

src/js/components/Video/Video.js
```diff
@@ -282,7 +289,7 @@ const Video = forwardRef(
               <Box flex>
                 <Stack>
                   <Meter
-                    aria-label="Video progress"
+                    aria-label={messages.progressMeter || 'Video progress'}
```
ShimiSun, line 292: Since we are providing a default value to `messages.progressMeter`, I don't think there is a point using the OR operator here.

src/js/components/Video/Video.js
```diff
@@ -296,6 +303,7 @@ const Video = forwardRef(
                     values={[{ value: percentagePlayed || 0 }]}
                   />
                   <StyledVideoScrubber
+                    aria-label={messages.scrubber || 'Scrubber'}
```
ShimiSun, line 306: Since we are providing a default value to messages.scrubber, I don't think there is a point using the OR operator here.

src/js/components/Video/Video.js
JoeButler7: good catch!

src/js/components/Video/Video.js
```diff
@@ -7,8 +7,8 @@ import React, {
   useRef,
   useState,
 } from 'react';
-import { ThemeContext } from 'styled-components';
 
+import { ThemeContext } from 'styled-components';
```
ShimiSun, line 11: 'styled-components' is an external library so it should be on the upper import section. Just revert to how it was.

src/js/components/Video/__tests__/Video-test.js
```diff
@@ -1,6 +1,9 @@
 import React from 'react';
 import 'jest-styled-components';
 import { cleanup, render, fireEvent } from '@testing-library/react';
+import { axe } from 'jest-axe';
+import 'jest-axe/extend-expect';
+import 'regenerator-runtime/runtime';
```
ShimiSun, line 6:
```suggestion
import 'regenerator-runtime/runtime';

```

src/js/components/Video/index.d.ts
```diff
@@ -10,6 +10,19 @@ export interface VideoProps {
...
+  messages?: {
+    closeMenu?: string,
+    fullScreen?: string,
+    progressMeter?: string,
+    openMenu?: string;
+    scrubber?: string,
+    pauseButton?: string,
+    playButton?: string,
+    volumeUp?: string,
+    volueDown?: string,
```
ShimiSun, line 22: Each message type should end with `;`

ShimiSun:

Issue #3161, opened 2019-06-14
Heading doesn't respect "small" responsive theme settings


sandersoeteman:
I'm trying to have a heading that is smaller on phones (small) than on tablets and desktop (medium & large). The size and height specified in the theme are respected for medium and large breakpoints, but not for the small breakpoint. It seems to always default to a global setting.
Tested on FF and Chrome.

See codesandbox for a demonstration: https://codesandbox.io/s/92rzl96r9y

oorestisime:
Hey so i had a look at the code. it seems the idea is to not to reduce the size property from medium to small etc depending on the screen size but the heading level. which seems to be working.
Maybe we should mention this in the docs?

sandersoeteman:
Disagree. The heading level communicates hierarchy in the page. It is a constant for the instance and should not be changed depending on the screen width.

oorestisime:
Yeah true. I am trying to find the rationale behind this. i might be even wrong actually and the level doesn't change. I read the code a week ago and forgot to answer here so i ll need to recheck.

jackielfu:
any update on this? i'm getting the same issue in version 2.7.8.

sandersoeteman:
Sorry, no. I've decided to implement another system instead.

ShimiSun:
The `level` prop is communicating the hierarchy from the responsive aspect, each `level` has different `size` values and the value can be determined using the size `prop` (defaults to 'medium').
The `size` prop doesn't control the responsive behavior; if the user assigns `size="small" ` and using responsive behavior, the size will remain `small` though the `level` will change according to the global breakpoints definitions from the theme. https://github.com/grommet/grommet/blob/7efd28655787d4f8aefc2ffaa584227db3c7855a/src/js/themes/base.js#L131

For further customizations on `size`, you could always create your own responsive behavior by using `<ResponsiveContext />`.

For now, I'm putting an action item to improve documentation and maybe state on the storybook example that the Headings(s) displayed are according to the screen size, and `level` values may vary according to the responsive behavior.
One more action item is to explore the `Math.min(props.level + 1, 4)` behavior when the user defines level 5 and 6. https://github.com/grommet/grommet/blob/7efd28655787d4f8aefc2ffaa584227db3c7855a/src/js/components/Heading/StyledHeading.js#L27 

@jackielfu would you like to share your issue so we can see what else is missing in this area?

jackielfu:
hey shimi!

[...]

PR #4217, opened 2020-06-19
Changed responsive to address breakpoint fix

### LINKED ISSUE #4196: Overwriting breakpoints in theme leads to malfunctions
WHEN adding custom breakpoints in `theme.global.breakpoints`,
THEN the two objects should be merged, with the custom ones overwriting the default values;
INSTEAD the `medium` breakpoint is NOT being overwritten
AND is still being resolved as the default, instead of the overwritten value.

### Expected Behavior
Custom breakpoints are resolved with both the provided value and the configuration, and the default values are being overwritten;

### Actual Behavior
`ResponsiveContext` resolves to `medium` on screen sizes that should provide custom breakpoint values; 

https://codesandbox.io/s/grommet-breakpoints-14zbx

### Steps to Reproduce
1. Provide Grommet a theme with `global.breakpoints` that creates new breakpoints and overwrites existing ones (e.g.: `xsmall`, `small`, `medium`, `large`, `xlarge`)
2. Wrap your component in ResponsiveContext and log its result
3. Start resizing the window and see the logs

### Your Environment
- Grommet version: 2.14.0
- Browser Name and version: Chrome 83.0.4103.97
- Operating System and version (desktop or mobile): OSX Catalina 10.15.5, desktop

ShimiSun:
#### What does this PR do?
Changed responsive to address breakpoint fix along with splitting the stories of Responsive Context.

The current logic to resolve the size of the breakpoint is NOT because of a bad theme merge. The size defaults to medium once the value of the breakpoint is bigger than the window size (returned as undefined).
The last size i.e. the largest breakpoint should have no boundary of windowWidth, and the breakpoint should default to it.

responsive.js
Storybook -> Responsive -> Custime Breakpoint story.
Jest.
Storybook -> Responsive -> Custime Breakpoint story.
#3212
Fixes #4196 
Might be related to #3161 
No
Yes
backwards compatible.


src/js/components/Grommet/__tests__/__snapshots__/Grommet-test.js.snap
```diff
@@ -265,7 +265,7 @@ exports[`Grommet ssr rendering Mozilla/5.0 (iPad; CPU OS 1`] = `
   <h1
     className="c1"
   >
-    Received size medium for Mozilla/5.0 (iPad; CPU OS 11_2_1 like Mac OS X) AppleWebKit/604.4.7 (KHTML, like Gecko) Mobile/15C153 [FBAN/FBIOS;FBAV/156.0.0.41.97;FBBV/89172188;FBDV/iPad5,3;FBMD/iPad;FBSN/iOS;FBSV/11.2.1;FBSS/2;FBCR/;FBID/tablet;FBLC/en_GB;FBOP/5;FBRV/0]
+    Received size large for Mozilla/5.0 (iPad; CPU OS 11_2_1 like Mac OS X) AppleWebKit/604.4.7 (KHTML, like Gecko) Mobile/15C153 [FBAN/FBIOS;FBAV/156.0.0.41.97;FBBV/89172188;FBDV/iPad5,3;FBMD/iPad;FBSN/iOS;FBSV/11.2.1;FBSS/2;FBCR/;FBID/tablet;FBLC/en_GB;FBOP/5;FBRV/0]
```
ericsoderberghp, line 268: Should these have changed?

ShimiSun: I agree it doesn't look right for the iPad to receive the `large` vs `medium` breakpoint. I'll have to take a deeper look into the SSR role and its impact on the logic. 
@oorestisime, any insights could be helpful here.

src/js/utils/responsive.js
```diff
@@ -1,25 +1,32 @@
...
-        if (!breakpoint.value || breakpoint.value >= windowWidth) {
-          result = name;
-          return true;
-        }
+    },
+  );
+  const lastBreakpoint = sortedBreakpoints[sortedBreakpoints.length - 1];
+
+  sortedBreakpoints.some(name => {
+    const breakpoint = theme.global.breakpoints[name];
```
oorestisime, line 17: I am not sure i understand what is going on here :/
what is lastBreakpoint and why we use this. Probably this is what lead to the snapshot changes?

I think this whole part is a bit confusing because we use array.some to not do anything here just set the result. maybe this was better as a filter? Do we need the last value?

ShimiSun: The condition in line 22 `breakpoint.value >= windowWidth` requests the value of the breakpoint to be larger than the windowWidth.

Let's say the last breakpoint size is `xlarge: { value: 1000 }`, and the client is using a monitor that has 2000 of windowWidth, our logic, in that case, should be to return the `xlarge` size, but instead we are returning `undefined` (due to `breakpoint.value >= windowWidth` returns `false`) that eventually default the breakpoint to be `medium`.
I think the last breakpoint should have no boundary with its value, and that is what this PR is suggesting. I'm not sure how it impacts the SSR tests and wanted your inputs.

oorestisime:
So i might have understood why snapshots are failing. i haven't really tested my assumption but here it is:

we were relying on getting an undefined on the first run of https://github.com/grommet/grommet/blob/master/src/js/components/Grommet/Grommet.js#L90

because at that point we were enterring this part https://github.com/grommet/grommet/blob/master/src/js/components/Grommet/Grommet.js#L138 and hence taking the user agent endpoint.

Now that getBreakpoint was changed to always return something then on mount we already get the large endpoint and we don't really execute the get device breakpoint with the user agent anymore.

What we could do to fix is call the deviceBreakpoint here https://github.com/grommet/grommet/blob/master/src/js/components/Grommet/Grommet.js#L90 if we don't have the responsive state variable set. that way on mount (on SSR) it will take the user agent breakpoint. 

 Hope this helps

ShimiSun:
Thank you for the findings @oorestisime ! 
I've incorporated your feedback along with additional fine-tuning for breakpoint calculation on the initial first load, and both the jest-snapshots and Chromatic are happy with the updates.

PR #4232, opened 2020-06-26
Fixed Tab disabled styles of border and color

halocline:

#### What does this PR do?

Allows for the disabled state of Tab to be styled via the theme. Initially inspired by https://github.com/hpe-design/design-system/issues/920 which surfaced the need for being able to define disabled styling when `theme.button.default` is defined.

Tab.js, StyledTab.js

Storybook - Tab 'States' story added
Additional Jest tests

Storybook

Original issue surfaced: https://github.com/hpe-design/design-system/issues/920

An additional PR will be submitted to apply `utils/styles/disableStyle()` to `<StyledButtonKind />` which patches backwards compatibility for folks using `theme.button.default`, applying opacity and cursor styles.

https://github.com/hpe-design/design-system/issues/920

**Before**
Screen Shot 2020-06-25 at 6 57 50 PM

**After**
Screen Shot 2020-06-25 at 6 58 15 PM

Yes.

??? [Shimi] Yes. as a bug fix for disabled Tab state.

Backwards compatible.


src/js/components/Tab/StyledTab.js
```diff
@@ -22,9 +22,17 @@ const tabHoverStyle = css`
   }
 `;
 
+// const StyledTab = styled.div`
+//   ${genericStyles}
+//   ${props => !props.plain && props.theme.tab.hover && tabHoverStyle}
+//   ${props => props.theme.tab.extend}
+// `;
+
```
ShimiSun, line 30: Can we remove the commented code?

src/js/components/Tab/doc.js
```diff
@@ -56,6 +56,9 @@ export const themeDoc = {
           light: 'black',
         },
       },
+      disabled: {
+        // color: undefined,
+      },
```
ShimiSun, line 61: We should actually stay away from nested objects on the default value, we should break the object down to separate props and that will probably take place on a separate PR. For now, can you please create a new doc entry for `tab.border.disabled.color`?

src/js/components/Tab/Tab.js
```diff
@@ -70,6 +73,8 @@ const Tab = forwardRef(
           theme.tab.border.color || theme.global.control.border.color;
         if (active) {
           borderColor = theme.tab.border.active.color || borderColor;
+        } else if (disabled && theme.tab.border.disabled) {
+          borderColor = theme.tab.border.disabled.color || borderColor;
```
ShimiSun, line 77: Going over the flow of resolving the border color, if somehow the tab is defined as `active` while `disabled`, the `disabled` will be ignored -- so `active` will triumph. I wanted to hear your thoughts on maybe put a warning message for the combined usage of both ` active` & `disabled`, and I'm thinking maybe `plain` & `disabled` as well. let me know if you think it will be useful.

src/js/components/Tabs/__tests__/Tabs-test.js
```diff
@@ -115,4 +115,93 @@ describe('Tabs', () => {
...
+
+    expect(container.firstChild).toMatchSnapshot();
+
+    const disabledTab = getByText('Disabled Tab').parentElement;
+    const disabledTabStyle = window.getComputedStyle(disabledTab);
+    expect(disabledTabStyle.color).toBe(disabledTextColor);
+    expect(disabledTabStyle.borderBottomColor).toBe(disabledBorderBottomColor);
+  });
+
+  test(`should apply custom theme disabled style when theme.button.default is
```
jcfilben, line 169: Looks like the snapshot needs to be updated for this test, right now it is failing because the snapshot doesn't match.

src/js/components/Tab/Tab.js
halocline: I think implementing warning for `active && disabled` would beneficial. I will add.

For `plain && disabled`, I think this is a valid combination. I could see a use case where I have a `plain` Tab for my style reasons, and would want to apply `disabled` in addition to `plain`.

src/js/components/Tab/StyledTab.js
halocline: Done.

src/js/components/Tab/doc.js
halocline: Done.

src/js/components/Tab/Tab.js
halocline: Added warning for the `active && disabled` case as they are conflicting states.

Opted not to add warning for the `plain && disabled` as having a `disabled` state on `plain` styled Tab is reasonable scenario.

src/js/components/Tabs/__tests__/Tabs-test.js
halocline: Snapshot updated.

src/js/components/Tab/Tab.js
ShimiSun: Thanks! I believe the users will find it helpful.

Regarding `plain & disabled`, I noticed that on the changes you've made in `StyledTab.js`, the `disabled` style will only apply on a button that isn't `plain`, so I might be missing something but from the code on this PR I don't see a way for `disabled` style to be applied on a `plain` button. Is that the intention?

halocline: That was the intention originally, but based on the above, I have come around and believe that applying disabled state to a plain style is reasonable. It would also be consistent with the behavior for `<Button label="Button" plain disabled />`. Good catch. I will modify.

ShimiSun: Great, thanks.

halocline: FYI, confirming that the `plain` restriction on StyledTab was removed.

src/js/components/Tab/doc.js
```diff
@@ -65,11 +65,21 @@ export const themeDoc = {
       },
     }`,
   },
+  'tab.border.disabled.color': {
+    description: 'border color of the Tab when disabled',
+    type: 'string | object',
```
ShimiSun, line 70:
```suggestion
    type: 'string | {dark: string, light: sting}',
```

src/js/components/Tab/doc.js
```diff
@@ -65,11 +65,21 @@ export const themeDoc = {
...
+    defaultValue: undefined,
+  },
   'tab.color': {
     description: 'text color for the Tab.',
     type: 'string',
     defaultValue: 'control',
   },
+  'tab.disabled.color': {
+    description: 'text color of the Tab when disabled.',
+    type: 'string | object',
```
ShimiSun, line 80:
```suggestion
    type: 'string | {dark: string, light: sting}',
```

src/js/components/Tab/doc.js
```diff
@@ -65,11 +65,21 @@ export const themeDoc = {
...
     }`,
   },
+  'tab.border.disabled.color': {
+    description: 'border color of the Tab when disabled',
+    type: 'string | object',
+    defaultValue: undefined,
+  },
   'tab.color': {
     description: 'text color for the Tab.',
     type: 'string',
```
ShimiSun, line 75: while you are already here :)
```suggestion
    type: 'string | {dark: string, light: sting}',
```

src/js/components/Tab/doc.js
halocline:
```suggestion
    type: 'string | {dark: string, light: string}',
```

halocline:
> LGTM, thank you! Can you please resolve the conflicts so this PR will be good to go?

Done.

Issue #4097, opened 2020-05-18
Button target property complains about named targets


MikeKingdom:
The "target" property to Button complains if the target is a string other than _self, _blank, _parent or _top.

### Expected Behavior

String values for the "target" property other than the 4 builtin values shouldn't result in a complaint from prop types. It's actually not unusual to specify a named window as a target so that links from that anchor always end up in that tab/window rather than creating a new one. For example "quakehelp" for all help links in quake.

### Actual Behavior

Prop types complains in the browser console.log:
Warning: Failed prop type: Invalid prop `target` of value `quakehelp` supplied to `Button`, expected one of ["_self","_blank","_parent","_top"].

### URL, screen shot, or Codepen exhibiting the issue
![image](https://user-images.githubusercontent.com/12054362/82261869-109c4700-991d-11ea-8e87-25a946d62c05.png)

### Steps to Reproduce

1. Create a grommet app with a `<Button target="quakehelp"/>`
2. Run the app with a dev server so it logs prop type complaints
3. Look in the F12 debug console log.

### Your Environment
- Grommet version: 2.13.0 
- Browser Name and version: Chrome Version 81.0.4044.138 (Official Build) (64-bit
- Operating System and version (desktop or mobile): Windows 10

PR #4236, opened 2020-06-26
Fix Button propTypes for the target property to allow string

MikeKingdom:
The Button propTypes currently complain about specifying
something other than _self, _blank, _parent, _top. In
html the anchor tag actually allows any string in target
to create a named tab/window. The names above that start
with _ are special names but it is perfectly fine to
create a named window.

#### What does this PR do?
Changes Button propTypes for the target prop to allow any string
and not just _self, _blank, _parent, _top

Manually tested with a temporary tweak in the storybook for Button.
Ran jest tests.

Just add a target="somename" to a Button with an href.

Setting target to some string like quakehelp is something we often do for help links. That way, when the user selects any links with the same target window name, the help always ends up in the same window rather than continually creating new windows/tabs.

#4097 

Done

Don't think this needs to be mentioned

It is backwards compatible

### DIFF
--- a/src/js/components/Button/README.md
+++ b/src/js/components/Button/README.md
@@ -327,6 +327,7 @@ _self
 _blank
 _parent
 _top
+string
 ```
 
 **type**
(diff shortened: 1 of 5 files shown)


ShimiSun:
Thanks for the PR, I read that before HTML5 the standard was to use the explicit `target` keywords, but on HTML5, where `target` can also be a tab/window string this PR makes sense. Thank you for your contribution!

PR #4238, opened 2020-06-27
RadioButton label fontWeight from theme (fixes #4186)

### LINKED ISSUE #4186: Would like ability for RadioButton label font-weight to be specified via theme
### Expected Behavior

Able to specify font-weight for RadioButton labels in my theme.

### Actual Behavior

I believe the label is inheriting the global font-weight.

Use Case: HPE Design System designs call for RadioButton label to be font-weight: medium/500, however the default label is [currently implemented](https://github.com/grommet/grommet/blob/6219ed0086c8673692ff78ca6a31d710f550ed79/src/js/components/RadioButton/RadioButton.js#L23) as a `span` inheriting global font weight

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

jackjwilliams:
#### What does this PR do?
Adds the ability to set the font-weight of the RadioButton label using theme overrides.

themes\base.js

Storybook and added a new test for the fontWeight modified label

Run storybook, check out the RadioButton\Theme story

#4186 

Yes

Sure, if you want

It is backwards compatible

### General Comments
I see some components using `weight: ###` and some using `fontWeight: ###` in the theme. Wasn't sure which one to use so I opted for more descriptive. Might consider standardizing for less confusion.

Additionally, I wasn't sure if I should commit all the README*.md changes (I figured they were automatically generated in some build script) - I left them out.


src/js/components/RadioButtonGroup/__tests__/__snapshots__/RadioButtonGroup-test.js.snap
```diff
@@ -408,7 +408,12 @@ exports[`RadioButtonGroup boolean options 1`] = `
           </svg>
         </div>
       </div>
-      <span>
+      <span
+        className=""
+        onBlur={[Function]}
```
jackjwilliams, line 413: This is probably from passing ...rest to the StyledRadioButtonLabel. Looking to see how other components handle this...

src/js/components/RadioButton/doc.js
```diff
@@ -97,6 +97,11 @@ export const themeDoc = {
     type: 'string',
     defaultValue: 'small',
   },
+  'radioButton.fontWeight': {
```
ShimiSun, line 100:
```suggestion
  'radioButton.font.weight': {
```
I'm inclining to support the structure of  `radioButton.font.weight` instead of `fontWeight`, and the reason is that if we would like to scale in the future and enhance the font object (e.g. adding font family) it will be cleaner and more straight forward. thoughts? @ericsoderberghp I would like your quick feedback here as well.

src/js/components/RadioButtonGroup/__tests__/__snapshots__/RadioButtonGroup-test.js.snap
```diff
@@ -408,7 +408,9 @@ exports[`RadioButtonGroup boolean options 1`] = `
           </svg>
         </div>
       </div>
-      <span>
+      <span
+        className=""
+      >
```
ShimiSun, line 413: Existing snap tests should not change, so the snapshot should stay intact with only code added and not code removed. One reason for it may be that the test you added wasn't being added to the end of the test file, so try that,  if that doesn't fix the snapshot, then this PR may not be backward compatible and we'll need to find the root cause for it and fix it.

src/js/themes/base.js
```diff
@@ -886,6 +886,7 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       },
       gap: 'small',
       size: `${baseSpacing}px`,
+      // fontWeight: undefined,
```
ShimiSun, line 889: Great, thanks for adding this line.
Could you also add the prop to[ base.d.ts](https://github.com/grommet/grommet/blob/7ed818375486b580b9585a2bd9d5ad90037a1a52/src/js/themes/base.d.ts#L836), similarly to how other font weights are defined?

src/js/components/RadioButtonGroup/__tests__/__snapshots__/RadioButtonGroup-test.js.snap
jackjwilliams: Hrmmm, from the research I've done (I'm new to styled-components) - it looks like empty classes are pretty common when moving from non-styled components to styled components (in the event the logic dictates that no css is written). Here, we have a new styled.span that only renders css if radioButton.fontWeight is specified. So the default case (radioButton.fontWeight = undefined) renders no css, and thus an empty className="" (I'm guessing, based on my new understanding of styled-components).

The reason RadioButtonGroup-test.js.snap changed is because RadioButtonGroup uses RadioButton under the hood.

I did try moving it to the end of the tests, but that doesn't fix it.

I guess I could add theme checking logic in RadioButton.js (if theme.radioButton.fontWeight { render new StyledLabel } else render old span with no empty className, but not sure if it's just overcomplicating things.

ShimiSun: Thanks for sharing these great insights. I think I'm OK with how it works now.

src/js/components/RadioButton/doc.js
jackjwilliams: I do like that better - @ericsoderberghp let me know your thoughts.

src/js/themes/base.js
jackjwilliams: Will do!

src/js/components/RadioButton/doc.js
ShimiSun: @ericsoderberghp  approves :), @jackjwilliams let's head this direction, thank you!

jackjwilliams:
@ShimiSun I've pushed up changes based on your review.

One thought: This might be a good time (since we are adding new theme elements) to introduce a FontWeightType to index.d.ts / base.d.ts. Keeping with the pattern I see of abstracting some css properties. Later, other components could be updated to use this ts def behind the scenes (instead of number | string).

ShimiSun:
@jackjwilliams 
>One thought: This might be a good time (since we are adding new theme elements) to introduce a FontWeightType to index.d.ts / base.d.ts. Keeping with the pattern I see of abstracting some css properties. Later, other components could be updated to use this ts def behind the scenes (instead of number | string).

If you'd like this let me know and I'll update it, otherwise this PR should be done.

I love it! but I don't think it should hold back this PR, feel free to file it on a separate PR.

src/js/components/RadioButton/stories/Theme.js
```diff
@@ -22,6 +22,9 @@ const theme = deepMerge(grommet, {
     icon: {
       size: '10px',
     },
+    font: {
+      weight: 500,
+    },
```
ShimiSun, line 27: Thank you for adding that!

jackjwilliams: No problem, the codebase is very clean and easy to follow 🙂

PR #4241, opened 2020-06-29
Support clicking inside drops in a shadow root

brentropy:
If a drop is rendered inside a shadow root, clicks in the drop were still triggering `onClickOutside`. This is is because the host element for the shadow root is the target on the document `mousedown` event making it appear to be outside of the drop.

#### What does this PR do?

When a shadow root is attached in open mode, it is possible to get the real target by using `event.composedPath()`. This change uses `composedPath` to get the target if it is supported, and gracefully falls back to `event.target` when it isn't supported.

This has been tested in storybook with components that use a drop with `onClickOutside` and in my application that uses drops in a shadow root. This commit also includes a new test to cover this. 

The biggest thing is ensuring no existing functionality with `onClickOutside` is broken. If you would like to test with a shadow root you can set up a `containerTarget` that is in a shadow root and render a `Menu` making sure `onClick` for the items is still triggering.

No

Maybe?

Yes

### DIFF
--- a/src/js/components/Drop/DropContainer.js
+++ b/src/js/components/Drop/DropContainer.js
@@ -242,11 +242,21 @@ const DropContainer = forwardRef(
       const onClickDocument = event => {
         // determine which portal id the target is in, if any
         let clickedPortalId = null;
-        let node = event.target;
+        // using the first node in composedPath (if available) allows getting
+        // the real target if it is in a shadow root (as long as the shadow
+        // root mode is set to open)
+        let node = event.composedPath ? event.composedPath()[0] : event.target;
         while (clickedPortalId === null && node !== document) {
-          const attr = node.getAttribute('data-g-portal-id');
-          if (attr !== null) clickedPortalId = parseInt(attr, 10);
-          node = node.parentNode;
+          if (!node.getAttribute) {
+            // node is a DocumentFragment, continue to host element
+            // DocumentFragment is not defined in IE which prevents using
+            // an instanceof check
+            node = node.host;
+          } else {
+            const attr = node.getAttribute('data-g-portal-id');
+            if (attr !== null) clickedPortalId = parseInt(attr, 10);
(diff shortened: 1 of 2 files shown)


brentropy:
It looks like chromatic picked up some imperceptible text rendering changes with my latest merge of master. When that happens do you usually just accept them or do you re-trigger it?

ShimiSun:
Thanks buddy! The branch wasn't up to date with the master branch, once I've merged it with master, it resolved the Chromatic changes.

brentropy:
Hi @ShimiSun. Thanks for the review. I definitely understand the need to maintain support for IE. In this case I handled that by falling back to `event.target` if `event.composedPath()` is not supported. IE would continue working exactly as it is right now. Since IE doesn't support shadow dom anyway, the original bug is a non-issue for applications that need to support IE 11.

ShimiSun:
We are trying to learn about root shadow and see how it would be used in the context of grommet. 
Can you please share a codesandbox that reproduces the problem you are trying to solve for us to learn how one might use the library?

brentropy:
@ShimiSun sorry for taking a while to get back to you. I will get something put together.

brentropy:
Here is an example that demonstrates this specific bug, and the style isolation advantage of using shadow dom.

[Example](https://codesandbox.io/s/determined-jepsen-c1md0?file=/src/App.js)

I am working on a browser extension that injects UI in to any page. Using shadow dom has made it easy to not have to deal with trying to reset everything and fight specificity wars.

brentropy:
Hi @ShimiSun, is there anything else I can help with this?

ShimiSun:
Thanks for the reminder @brentropy , and for your patience. I believe it is on us at this point.

jcfilben:
Hey @brentropy sorry this pull request has been sticking around for a while, just wanted to check in and see if there was still interest in getting this merged

PR #4242, opened 2020-06-29
NEW Card component

ShimiSun:
#### What does this PR do?
This PR is a follow up from the great feedback on https://github.com/grommet/grommet/pull/4106
Design system Card https://www.figma.com/file/WzYBJncJEbjA2JU9GqZMZB/HPE-Cards?node-id=0%3A1

The PR is not fully complete yet, I would like to enhance the documentation, tests and the code examples.

Card.js
jest + storybook
100% coverage to the component 
image

storybook
#3489 
Yes, waiting for feedback for completing it.
Update: Done!
YES
 backwards compatible


src/js/components/Card/Card.js
```diff
@@ -0,0 +1,18 @@
...
+import { ThemeContext } from 'styled-components';
+
+import { defaultProps } from '../../default-props';
+import { Box } from '../Box';
+
+export const Card = ({ ...rest }) => {
+  const theme = useContext(ThemeContext) || defaultProps.theme;
+  return (
+    <Box
+      round="small"
```
ericsoderberghp, line 11: Can we put `round` and `elevation` in the base theme instead?

src/js/components/Card/__tests__/Card-test.js
```diff
@@ -0,0 +1,67 @@
...
+
+const src = '';
+
+describe('Card', () => {
+  afterEach(cleanup);
+
+  test('renders', () => {
+    const component = renderer.create(
+      <Grommet>
+        <Card id="test id" name="test name" />
```
ericsoderberghp, line 18: I don't think Card should need `id` or `name`. Can we prune this down to just `<Card />`?

src/js/components/Card/stories/Collapsible.js
```diff
@@ -0,0 +1,123 @@
...
+              </Box>
+            </Box>
+            <CardFooter>
+              <Box direction="row">
+                <Button icon={<Favorite color="red" />} hoverIndicator />
+                <Button icon={<ShareOption color="plain" />} hoverIndicator />
+              </Box>
+              <ExpandButton onClick={() => setOpen(!open)} />
+            </CardFooter>
+            <Collapsible open={open}>
```
ericsoderberghp, line 113: This doesn't seem like a behavior we want to promote in Card. I think we should keep the in-Card interaction to a minimum. Can we remove this story?

src/js/components/Card/stories/Media.js
```diff
@@ -0,0 +1,65 @@
...
+    <Box pad="medium" align="start">
+      <Card elevation="large" width="medium">
+        <CardBody height="small">
+          <Image
+            fit="cover"
+            src="//v2.grommet.io/assets/IMG_4245.jpg"
+            a11yTitle="bridge"
+          />
+        </CardBody>
+        <Box pad={{ horizontal: 'medium' }} responsive={false}>
```
ericsoderberghp, line 45: Why isn't this inside CardBody?

src/js/components/CardFooter/CardFooter.js
```diff
@@ -0,0 +1,13 @@
...
+import { ThemeContext } from 'styled-components';
+
+import { defaultProps } from '../../default-props';
+import { Footer } from '../Footer';
+
+// Needs to have a CardBody or a flex container when Card uses a fixed height.
+export const CardFooter = ({ ...rest }) => {
+  const theme = useContext(ThemeContext) || defaultProps.theme;
+  return (
+    <Footer background="background-contrast" {...theme.card.footer} {...rest} />
```
ericsoderberghp, line 11: Can we put `background` inside the base theme instead?

src/js/components/Card/stories/Media.js
ShimiSun: Good question, I read about the [body tag best practice in HTML](https://stackoverflow.com/questions/2035462/multiple-htmlbody-html-body-in-same-file), and its recommended to use only one body tag, I know this is not html but since we are using html tags concept (header, body, footer) it might be confusing to contradict that with multiple body sections. 
I'm thinking we should rename CardBody to CardFlex or CardContent or something along those lines... thoughts?

src/js/components/Card/stories/Collapsible.js
ShimiSun: I'm curious to know why,  I agree the Collapsible action may be too much since we don't recommend the changes it will cause on the overall layout, but I think the Share option and Favorite are on point examples which I was planning to enhance on this example.

src/js/components/Card/stories/Media.js
ericsoderberghp: I think there should still be one `CardBody`. I just think both children should be inside it.

src/js/components/Card/stories/Collapsible.js
ericsoderberghp: I'm basing it off of the guidance we're working on for Card in the design system, where a Card should be fairly focused, typically with at most one action. Just wanting to reinforce that type of behavior in our examples.

Issue #4173, opened 2020-06-10
Pressing `Enter` in List results in 2 onClickItem calls


jcfilben:
### Expected Behavior

When the user presses `Enter` on a list item one call should be made to `onClickItem`

### Actual Behavior

Instead when a user presses `Enter` two calls are made to `onClickItem`

### URL, screen shot, or Codepen exhibiting the issue

<img width="560" alt="Screen Shot 2020-06-10 at 3 11 10 PM" src="https://user-images.githubusercontent.com/54560994/84319148-c75ea200-ab2c-11ea-8eb8-4e754356e4f3.png">

### Steps to Reproduce
This issue is shown using the following test:
```
let onClickItem;
let App;

  beforeEach(() => {
    onClickItem = jest.fn();
    App = () => {
      return (
        <Grommet>
          <List
            data={[{ a: 'alpha' }, { a: 'beta' }]}
            onClickItem={onClickItem}
          />
        </Grommet>
      );
    };
  });

  afterEach(cleanup);

  test('Enter key', () => {
    const { container, getByText } = render(<App />);

    expect(container.firstChild).toMatchSnapshot();
    fireEvent.click(getByText('beta'));
    fireEvent.mouseOver(getByText('beta'));
    fireEvent.keyDown(getByText('beta'), {
      key: 'Enter',
      keyCode: 13,
      which: 13,
    });
    expect(onClickItem).toHaveBeenCalledTimes(2);
    expect(container.firstChild).toMatchSnapshot();
  });
```
This test will fail because `onClickItem` is called is 3 times instead of 2 (one from the fireEvent.click and 1 from the fireEvent.keydown on the Enter key).

ShimiSun:
@halocline for some reason I can't assign you this issue (@MikeWalrath ?) see if you can assign yourself.
In the spirit of approaching List on the design-system, this could be a nice warm up for next week's task force work.

halocline:
@MikeWalrath @ShimiSun I am unable to assign myself or take any administrative actions with issues in this repo.

MikeWalrath:
Thanks @halocline can you try again?

halocline:
Bingo! 🎉

PR #4248, opened 2020-06-30
List - prevent misfire of onClickItem handler

halocline:

#### What does this PR do?

Prevents extraneous call to `List`'s `onClickItem` handler.

Closes: https://github.com/grommet/grommet/issues/4173

**Steps to replicate**
Occurs in the following scenario where both mouse and keyboard are used:
- Go to Storybook --> List --> onClickItem https://storybook.grommet.io/?path=/story/list--onclickitem
  - To see the multiple calls, modify the story's `onClickItem` handler to log the event to console
- Mouse over an item, then click. That item is the active item index and now has focus.
- Leave mouse cursor over item (if cursor leaves that item, the onMouseLeave function will set active to undefined and the multiple calls will not occur)
- Use keyboard. Press Enter. Two calls to `onClickItem` will occur.
  - Optionally, use arrow keys to move active item, then press enter. Two calls to `onClickItem` will occur.

In this scenario, a single enter keydown event is triggering calls from both `List`'s `<Keyboard onEnter />` and `<StyledItem onClick />`.

~The approach to this fix is to check that the `event.type` is appropriate.~

Modified approach to remove onEnter function, then needed to update onClick function to account for active item being set by keyboard interaction in addition to mouse events.

- Review replication scenario
- List.js 

**Jest test needed to be modified to include setting an active item with an onMouseOver event prior to firing the click event. Snapshot: Including the mouseOver interaction triggers active state styling of the item which results in a change to the snapshot.**

Storybook, plus passes 'Enter key' Jest test which was previously failing.

Storybook

https://github.com/grommet/grommet/issues/4173 & Replication steps described above

https://github.com/grommet/grommet/issues/4173

No

Yes

Backwards compatible

### DIFF
--- a/src/js/components/List/List.js
+++ b/src/js/components/List/List.js
@@ -186,12 +186,14 @@ const List = React.forwardRef(
                   tabIndex: -1,
                   active: active === index,
                   onClick: event => {
-                    // extract from React's synthetic event pool
-                    event.persist();
-                    const adjustedEvent = event;
-                    adjustedEvent.item = item;
-                    adjustedEvent.index = index;
-                    onClickItem(adjustedEvent);
+                    if (event.type === 'click') {
+                      // extract from React's synthetic event pool
+                      event.persist();
+                      const adjustedEvent = event;
+                      adjustedEvent.item = item;
+                      adjustedEvent.index = index;
+                      onClickItem(adjustedEvent);
+                    }
                   },
                   onMouseOver: () => setActive(index),
                   onMouseOut: () => setActive(undefined),
--- a/src/js/components/List/List.js
+++ b/src/js/components/List/List.js
@@ -59,17 +59,6 @@ const List = React.forwardRef(
 
     return (
       <Keyboard
-        onEnter={
-          onClickItem && active >= 0
-            ? event => {
-                event.persist();
-                const adjustedEvent = event;
-                adjustedEvent.item = data[active];
-                adjustedEvent.index = active;
-                onClickItem(adjustedEvent);
-              }
-            : undefined
-        }
         onUp={
           onClickItem && active
             ? () => {
@@ -186,14 +175,12 @@ const List = React.forwardRef(
                   tabIndex: -1,
                   active: active === index,
                   onClick: event => {
-                    if (event.type === 'click') {
-                      // extract from React's synthetic event pool
-                      event.persist();
-                      const adjustedEvent = event;
-                      adjustedEvent.item = item;
-                      adjustedEvent.index = index;
-                      onClickItem(adjustedEvent);
-                    }
+                    // extract from React's synthetic event pool
+                    event.persist();
+                    const adjustedEvent = event;
+                    adjustedEvent.item = data[active];
+                    adjustedEvent.index = active;
+                    onClickItem(adjustedEvent);
                   },
                   onMouseOver: () => setActive(index),
                   onMouseOut: () => setActive(undefined),
--- a/src/js/components/List/__tests__/List-test.js
+++ b/src/js/components/List/__tests__/List-test.js
@@ -70,7 +70,9 @@ describe('List', () => {
       </Grommet>,
     );
     expect(container.firstChild).toMatchSnapshot();
-    fireEvent.click(getByText('beta'));
+    const item = getByText('beta');
+    fireEvent.mouseOver(item);
+    fireEvent.click(item);
     expect(onClickItem).toBeCalledWith(
       expect.objectContaining({ item: { a: 'beta' } }),
     );
--- a/src/js/components/List/__tests__/List-test.js
+++ b/src/js/components/List/__tests__/List-test.js
@@ -256,10 +256,7 @@ describe('List events', () => {
       keyCode: 13,
       which: 13,
     });
-    // Reported bug: onEnter calls onClickItem twice instead of once.
-    // Issue #4173. Once fixed it should be
-    // `expect(onClickItem).toHaveBeenCalledTimes(2);`
-    expect(onClickItem).toHaveBeenCalledTimes(3);
+    expect(onClickItem).toHaveBeenCalledTimes(2);
     // Both focus and active should be placed on 'beta'
     expect(container.firstChild).toMatchSnapshot();
   });
--- a/src/js/components/List/__tests__/__snapshots__/List-test.js.snap
+++ b/src/js/components/List/__tests__/__snapshots__/List-test.js.snap
@@ -2454,7 +2454,7 @@ exports[`List onClickItem 2`] = `
       alpha
     </li>
     <li
-      class="StyledBox-sc-13pk1d4-0 fbIRTP List__StyledItem-sc-130gdqg-1 fBniOQ"
+      class="StyledBox-sc-13pk1d4-0 hEstNb List__StyledItem-sc-130gdqg-1 fBniOQ"
       tabindex="-1"
     >
       beta


src/js/components/List/List.js
```diff
@@ -186,12 +186,14 @@ const List = React.forwardRef(
                   tabIndex: -1,
                   active: active === index,
                   onClick: event => {
-                    // extract from React's synthetic event pool
-                    event.persist();
-                    const adjustedEvent = event;
-                    adjustedEvent.item = item;
-                    adjustedEvent.index = index;
-                    onClickItem(adjustedEvent);
+                    if (event.type === 'click') {
```
ericsoderberghp, line 189: This looks a little awkward and slightly obscure, that the `onClick` would be called for an event other than `click`.

Would it produce the same effect if we just removed `onEnter` from the use of `Keyboard`? That feels a bit cleaner to me, if possible. As is, having the Keyboard onEnter here feels a bit like adding Keyboard with onEnter in a Form context and then preventing double submit by checking for the event.type on the submit button. Better just to not listen for Enter with Keyboard.

taysea: I would agree with Eric's comment in terms of readability. Just to confirm, a "click" only occurs as an effect of mousedown/mouseup but not keyboard and that is why this specification solves the issue?

halocline: @ericsoderberghp I agree that it feels awkward. It was feeling weird to me that the onClick was accepting other event types which was I opted to place the check there. I believe the ListItem `onClick` is using `<Box onClick />`.  I will check out your suggested approach.

@taysea - Confirming keyboard actions result in event types of `keydown`, `keypress`, and `keyup`. A mouse 'left click' results in event types of `mousedown`, `mouseup`, and `click`. A mouse 'right click' typically will fire an event type of `mousedown`, often opening a menu awaiting for other events to occur.

So yes, I was thinking of restricting the `onClick` to click events, but will check out alternatives.

halocline: @ericsoderberghp Thanks for your question and suggested investigation. I modified approach to remove onEnter function, then needed to update onClick function to account for active item being set by keyboard interaction in addition to mouse events. This feels much better.

Note: I did need to slightly modify one Jest test to include a mouseOver event prior to the click event so that an active item had been set (models how a user would interact). Including the mouseOver did result in a snapshot change for that test because it changes the active state styling.

This should be ready for re-review.

src/js/components/List/List.js
```diff
@@ -189,8 +178,8 @@ const List = React.forwardRef(
                     // extract from React's synthetic event pool
                     event.persist();
                     const adjustedEvent = event;
-                    adjustedEvent.item = item;
-                    adjustedEvent.index = index;
+                    adjustedEvent.item = data[active];
```
taysea, line 181: Noticing that this was modified to match what have been in `Keyboard onEnter`, and just confirming that this had no effect on mouse clicks?

ericsoderberghp: I wouldn't have thought these changes were needed, since clicking Enter while this item is active would result in this onClick being called, which already has `item` and `index` in scope.

halocline: The issue was that without updating them, they would pass the incorrect item value in certain scenarios (when mixing mouse and keyboard interactions). In "happy path" scenarios they would not have been needed. 

However, in the scenario where focus is initiated by the mouse, then up/down keyboard actions, then Enter. The item passed to the `onClickItem()` still references the original mouse click item and index. The following screenshots illustrate this scenario w/out the change. 

**Scenario - onEnter removed, onClick w/out change - onClickItem story**
Config:
```
onClick: event => {
  // extract from React's synthetic event pool
  event.persist();
  const adjustedEvent = event;
  adjustedEvent.item = item;
  // adjustedEvent.item = data[active];
  adjustedEvent.index = index;
  // adjustedEvent.index = active;
  onClickItem(adjustedEvent);
},
```
Step 1 
- Load
Screen Shot 2020-07-08 at 10 19 13 AM

Step 2
- Click on 'entry-3'
- Expected entry-3 object
- Result entry-3 object
Screen Shot 2020-07-08 at 10 19 34 AM

Step 3
- Arrow down to entry-7
- Press Enter
- Expected entry-7 object
- Result entry-3 object
Screen Shot 2020-07-08 at 10 19 59 AM

Fail.

----------------------------------------------------

**Scenario - onEnter removed, onClick w/ change - onClickItem story**
Config:
```
onClick: event => {
  // extract from React's synthetic event pool
  event.persist();
  const adjustedEvent = event;
  // adjustedEvent.item = item;
  adjustedEvent.item = data[active];
  // adjustedEvent.index = index;
  adjustedEvent.index = active;
  onClickItem(adjustedEvent);
},
```
Step 1 
- Load
Screen Shot 2020-07-08 at 10 23 17 AM


Step 2
- Click on 'entry-4'
- Expected entry-4 object
- Result entry-4 object
Screen Shot 2020-07-08 at 10 23 37 AM


Step 3
- Arrow down to entry-8
- Press Enter
- Expected entry-8 object
- Result entry-8 object
Screen Shot 2020-07-08 at 10 23 53 AM


Pass.

-------------------------------------------------

halocline: That said, I just uncovered an issue with this setup...  by removing the onEnter altogether, we lost keyboard only functionality... so I need to revisit.

taysea: Thanks for the detailed examples here and checking out these different cases. By "lost keyboard only functionality", you mean that if you hadn't clicked on an item first with mouse, then you would be able to arrow through the options?

halocline: > Thanks for the detailed examples here and checking out these different cases. By "lost keyboard only functionality", you mean that if you hadn't clicked on an item first with mouse, then you would be able to arrow through the options?

Arrow through works fine, but arrowing through then pressing Enter does not. 

There is a test written for 'Enter key' and this passed, so I missed manually testing: Tab (for focus on list) --> arrow up/down --> Enter. However, looking closer at the test's implementation, it is initiated with a click, then later presses Enter. I think we need to revisit how these tests are written and ensure they include keyboard only scenarios.

ShimiSun:
Where are we with this PR?

halocline:
> Where are we with this PR? @halocline

The original approach solved the issue, however there was a question about the implementation and an ask to explore an alternate approach.

The alternate approach seemed promising, solved the issue, passed all Jest tests, but during manual testing discovered that the alternate approach completely broke ability to use keyboard (by itself, no mouse interaction) to select an item with List - specifically, there is a test written for 'keyboard onEnter' which passed, however in real world this functionality is broken without the use of a mouse first.

So, I believe next steps from here are:
- Re-visit and re-write keyboard tests for this component, ensuring the keyboard tests aren't reliant on mouse interaction ([example](https://github.com/grommet/grommet/blob/ea8a37da89eb1e985a274c535c8dc934dd1a645e/src/js/components/List/__tests__/List-test.js#L252))
- Either:
  - Re-implement the original fix, OR
  - Open up further exploration for alternate fix approaches.

ShimiSun:
Would you like any help to keep this PR going? Reach out to @jcfilben if you need help with the tests.

halocline:
> Would you like any help to keep this PR going? Reach out to @jcfilben if you need help with the tests.

Happy to keep this going, but am currently prioritizing (https://github.com/grommet/grommet/issues/4334) as it seems to have broader and more significant implications. Can redirect focus if you'd like.

ShimiSun:
@halocline, feel free to prioritize how you see fit, I'm just making sure this PR is still under your radar.

taysea:
I am revisiting this issue. It seems like the initial double fire of onClick was handled by https://github.com/grommet/grommet/pull/4364. I followed the steps described above to reproduce and was unable. When I reverted the changes locally, I saw the double onClick again.

We may be able to close this issue, however I am currently resolving https://github.com/grommet/grommet/issues/4441, and will confirm that fixes there do not reintroduce this issue.

ShimiSun:
@taysea Fantastic, keep us posted! 
(A proper hebrew phrase for the situation, roughly translated to English "the work of the righteous is done by others")

ShimiSun:
@taysea approved this functionality was taken care of on later PRs. PR will be closed.

PR #4253, opened 2020-07-01
Changed Chart to add point flavors

ericsoderberghp:
#### What does this PR do?

Changed Chart to add point flavors

doc.js

Added Chart -> Point story
Added unit test

Chart -> Point story

Driven off of the great accessibility research Matt did.

automatic

yes

backwards compatible

### DIFF
--- a/src/js/components/Chart/__tests__/__snapshots__/Chart-test.js.snap
+++ b/src/js/components/Chart/__tests__/__snapshots__/Chart-test.js.snap
@@ -701,6 +701,243 @@ exports[`Chart pad 1`] = `
 </div>
 `;
 
+exports[`Chart point 1`] = `
+.c0 {
+  font-size: 18px;
+  line-height: 24px;
+  box-sizing: border-box;
+  -webkit-text-size-adjust: 100%;
+  -ms-text-size-adjust: 100%;
+  -moz-osx-font-smoothing: grayscale;
+  -webkit-font-smoothing: antialiased;
+}
+
+.c1 {
+  display: block;
+  max-width: 100%;
+  overflow: visible;
+}
+
+<div
+  className="c0"
+>
+  <svg
+    className="c1"
+    height={192}
+    preserveAspectRatio="none"
+    viewBox="-12 -12 408 216"
+    width={384}
+  >
+    <g
+      fill="#6FFFB0"
+      stroke="none"
+      strokeLinecap="butt"
+      strokeLinejoin="miter"
+    >
+      <g
+        stroke="none"
+      >
+        <title>
+          sixty
+        </title>
+        <circle
+          cx={384}
+          cy={0}
+          r={12}
+        />
+      </g>
+      <g
+        stroke="none"
+      >
+        <title>
+          zero
+        </title>
+        <circle
+          cx={0}
+          cy={192}
+          r={12}
+        />
+      </g>
+    </g>
+  </svg>
+  <svg
+    className="c1"
+    height={192}
+    preserveAspectRatio="none"
+    viewBox="-12 -12 408 216"
+    width={384}
+  >
+    <g
+      fill="#6FFFB0"
+      stroke="none"
+      strokeLinecap="butt"
+      strokeLinejoin="miter"
+    >
+      <g
+        stroke="none"
+      >
+        <title>
+          sixty
+        </title>
+        <path
+          d="M 384 -12 L 396 0 L 384 12 L 372 0 Z"
+        />
+      </g>
+      <g
+        stroke="none"
+      >
+        <title>
+          zero
+        </title>
+        <path
+          d="M 0 180 L 12 192 L 0 204 L -12 192 Z"
+        />
+      </g>
+    </g>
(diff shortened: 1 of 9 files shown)


src/js/components/Chart/doc.js
```diff
@@ -80,6 +80,18 @@ export const doc = Chart => {
...
+      'triangle',
+      'triangleDown',
+    ]).description(
+      `When using a 'point' type, what shape the points should use.
+      If not specified, 'square' will be used if not 'round',
+      otherwise 'circle' will be used if 'round' is true.`,
+    ),
     round: PropTypes.bool
       .description('Whether to round the line ends.')
       .defaultValue(false),
```
ShimiSun, line 97: It is not clear that the `round` prop also affect the `point` type, the description should mention that round also "rounds" the default square shape of point to a circle, or even better, be explicit that `round` triumph any explicit `point` value with returning a `circle` point, so users won't wonder how `round` should behave when used in congestion with `star`, `triangle`(s) or `square`.

src/js/components/Chart/doc.js
```diff
@@ -80,6 +80,18 @@ export const doc = Chart => {
...
     ),
+    point: PropTypes.oneOf([
+      'circle',
+      'diamond',
+      'square',
+      'star',
+      'triangle',
+      'triangleDown',
+    ]).description(
+      `When using a 'point' type, what shape the points should use.
```
ShimiSun, line 91:
```suggestion
      `When using a 'point' type and not using 'round' prop, what shape the points should use.
```
Just an idea of how to communicate the dependency on the `round` prop (see my other comment in this file)...

src/js/components/Chart/__tests__/__snapshots__/Chart-test.js.snap
```diff
@@ -1367,11 +1601,8 @@ exports[`Chart type 1`] = `
         <title>
           zero
         </title>
-        <rect
-          height={24}
-          width={24}
-          x={-12}
-          y={180}
+        <path
```
ShimiSun, line 1604: I tried to use the following code snippet to compare the visualization for backward compatibilty purposes:
```
<svg viewBox="0 0 220 100" xmlns="http://www.w3.org/2000/svg">
  <!-- Simple rectangle test renders as expected -->
  <rect width="24" height="24" />

  <!-- 2.14.0 version for rendering point -->
  <rect height={24} width={24} x={-12} y={180} />
  <!-- this PR replaces rect with path -->
  <path d="M -12 180 L 12 180 L 12 204 L -12 204 Z"/>
</svg>
```
But nothing was rendered for me on either test, I was hoping you'll have insights into what I'm doing wrong.
Also, I'm calling out that users that may have specific styles for `<rect />` would loose their styles due to this change.

src/js/components/Chart/doc.js
ericsoderberghp: I will attempt to clarify more in the documentation. Just so you know here, `point` trumps `round`. `round` only comes into play if `point` is specified.

src/js/components/Chart/doc.js
ericsoderberghp: Actually, it's the other way around. If `point` isn't specified, then `round` affects it, to align with current behavior.

src/js/components/Chart/__tests__/__snapshots__/Chart-test.js.snap
ericsoderberghp: I think the key issue is that you can't use `{24}` in vanilla SVG. I did this to illustrate:
```
<svg width="220" height="100" viewBox="0 0 220 100" xmlns="http://www.w3.org/2000/svg">
  <rect width="24" height="24" fill="green" />
  <rect height="24" width="24" x="24" y="48" fill="blue" />
  <path d="M 48 24 L 72 24 L 72 48 L 48 48 Z" fill="red" />
</svg>
```

ericsoderberghp: I'm not concerned about the internal being `rect` as we don't guarantee DOM element consistency of internal elements.

src/js/components/Chart/doc.js
ericsoderberghp: documentation updated

ericsoderberghp:
> Overall looks good, I personally think that allowing the user to provide a custom path or an SVG will be more scalable then providing out of the box shapes.

We can head that direction eventually. This seemed to hit the primary usage and to be easier than having to create a good small graphic.

src/js/components/Chart/__tests__/__snapshots__/Chart-test.js.snap
ShimiSun: Agreed. I'm also guessing that Chromatic would have alerted us in case there was a visual impact on the svg when `<rect />` got replaced with `<path />`

ericsoderberghp: correct

PR #4255, opened 2020-07-01
Create TabsContext to fix Tab issue

### LINKED ISSUE #4224: Props not being passed from Tabs to Tab when using custom Tab component
I have a component that uses Tab at the top level:
```
const MyTabComponent = props => (
  <Tab title="Tab 2" {...props}>
      <Text>Content for the Second tab</Text>
  </Tab>
);
```

When I render it within tabs like this:
```
<Tabs>
      <Tab title="Tab 1">
         <Text>Content for the First tab</Text>
      </Tab>

      <MyTabComponent />
</Tabs>
```

"Content for the First tab" is displayed in Tab 1 but "Content for the Second tab" is not displayed within Tab 2.

Code: https://codesandbox.io/s/grommet-sandbox-tab-composition-issue-447vm?file=/src/Tabs.js
Demo: https://447vm.csb.app/tabs

### Your Environment

- Grommet version: 2.14.0

taysea:
#### What does this PR do?
Creates `TabsContext` which stores the `activeIndex` (needed to properly assign `activeTitle`), if tab is `active` (previously passed to Tab as a prop via clone child), `onActivate` (previously passed to individual Tab as a prop via clone child), `setActiveContent` (allows Tab to assign its children as the active content to be rendered by Tabs, and `setActiveTitle` (for screen reader accessibility).

Previously, Tabs was looking into its children to find what should be the active content displayed. If `Tab` was not the direct descendant of `Tabs`, it wouldn't render the child properly. Now, `Tab` takes more control by assigning its children to the activeContent. This hand-off is done using `TabsContext`.

Closes #4224 

src/js/components/Tabs/Tabs.js

In local storybook, I recreated the setup causing the bug in the filed issue:

```
const MyTabComponent = props => (
  <Tab title="Tab 2" {...props}>
      <Text>Content for the Second tab</Text>
  </Tab>
);
...
...
<Tabs>
      <Tab title="Tab 1">
         <Text>Content for the First tab</Text>
      </Tab>
      <MyTabComponent />
</Tabs>
```

Prior to this change, the content inside of `MyTabComponent` wasn't rendering, but after this change it renders properly.

Same as above.

Solves [this issue](https://github.com/grommet/grommet/issues/4224), when users created custom Tab component, the tab itself would render but its child content wouldn't.

https://github.com/grommet/grommet/issues/4224

No.

No.

### DIFF
--- a/src/js/components/Tab/Tab.js
+++ b/src/js/components/Tab/Tab.js
@@ -32,12 +32,7 @@ const Tab = forwardRef(
     useEffect(() => {
       if (active) {
         setActiveContent(children);
-        let activeTitle;
-        if (typeof title === 'string') {
-          activeTitle = title;
-        } else {
-          activeTitle = activeIndex + 1;
-        }
+       const activeTitle = typeof title === 'string' ? title : activeIndex + 1;
         setActiveTitle(activeTitle);
       }
     }, [
--- a/src/js/components/Tab/Tab.js
+++ b/src/js/components/Tab/Tab.js
@@ -1,4 +1,4 @@
-import React, { forwardRef, useContext, useState } from 'react';
+import React, { forwardRef, useContext, useEffect, useState } from 'react';
 import { ThemeContext } from 'styled-components';
 
 import { defaultProps } from '../../default-props';
@@ -7,29 +7,47 @@ import { Box } from '../Box';
 import { Button } from '../Button';
 import { Text } from '../Text';
 import { normalizeColor } from '../../utils';
+import { TabsContext } from '../Tabs/TabsContext';
 
 import { StyledTab } from './StyledTab';
 
 const Tab = forwardRef(
   (
-    {
-      active,
-      icon,
-      plain,
-      title,
-      onActivate,
-      onMouseOver,
-      onMouseOut,
-      reverse,
-      ...rest
-    },
+    { children, icon, plain, title, onMouseOver, onMouseOut, reverse, ...rest },
     ref,
   ) => {
+    const {
+      active,
+      activeIndex,
+      setActiveContent,
+      setActiveTitle,
+      onActivate,
+    } = useContext(TabsContext);
     const theme = useContext(ThemeContext) || defaultProps.theme;
     const [over, setOver] = useState(undefined);
     const [focus, setFocus] = useState(undefined);
     let normalizedTitle = title;
     const tabStyles = {};
+
+    useEffect(() => {
+      if (active) {
+        setActiveContent(children);
+        let activeTitle;
+        if (typeof title === 'string') {
+          activeTitle = title;
+        } else {
+          activeTitle = activeIndex + 1;
+        }
+        setActiveTitle(activeTitle);
+      }
+    }, [
+      active,
+      activeIndex,
+      children,
+      setActiveContent,
+      setActiveTitle,
+      title,
+    ]);
 
     const onMouseOverTab = event => {
       setOver(true);
--- a/src/js/components/Tab/Tab.js
+++ b/src/js/components/Tab/Tab.js
@@ -19,9 +19,9 @@ const Tab = forwardRef(
     const {
       active,
       activeIndex,
+      onActivate,
       setActiveContent,
       setActiveTitle,
-      onActivate,
     } = useContext(TabsContext);
     const theme = useContext(ThemeContext) || defaultProps.theme;
     const [over, setOver] = useState(undefined);
--- a/src/js/components/Tab/Tab.js
+++ b/src/js/components/Tab/Tab.js
@@ -6,8 +6,8 @@ import { defaultProps } from '../../default-props';
 import { Box } from '../Box';
 import { Button } from '../Button';
 import { Text } from '../Text';
-import { normalizeColor } from '../../utils';
 import { TabsContext } from '../Tabs/TabsContext';
+import { normalizeColor } from '../../utils';
 
 import { StyledTab } from './StyledTab';
 
--- a/src/js/components/Tabs/Tabs.js
+++ b/src/js/components/Tabs/Tabs.js
@@ -1,16 +1,10 @@
-import React, {
-  forwardRef,
-  cloneElement,
-  Children,
-  useContext,
-  useState,
-} from 'react';
+import React, { forwardRef, useContext, useState } from 'react';
 import { ThemeContext } from 'styled-components';
 
 import { defaultProps } from '../../default-props';
 
 import { Box } from '../Box';
-
+import { TabsContext } from './TabsContext';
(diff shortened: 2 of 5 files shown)


src/js/components/Tab/Tab.js
```diff
@@ -7,30 +7,48 @@ import { Box } from '../Box';
 import { Button } from '../Button';
 import { Text } from '../Text';
 import { normalizeColor } from '../../utils';
+import { TabsContext } from '../Tabs/TabsContext';
```
halocline, line 10: Nit: Recommend keeping Grommet's component imports grouped alphabetically and util functions can follow.

```suggestion
import { TabsContext } from '../Tabs/TabsContext';
import { normalizeColor } from '../../utils';
```

src/js/components/Tab/Tab.js
```diff
@@ -7,30 +7,48 @@ import { Box } from '../Box';
...
-    },
+    { children, icon, plain, title, onMouseOver, onMouseOut, reverse, ...rest },
     ref,
   ) => {
+    const {
+      active,
+      activeIndex,
+      setActiveContent,
+      setActiveTitle,
+      onActivate,
```
halocline, line 24: Nit: recommend alphabetical order

```suggestion
      active,
      activeIndex,
      onActivate,
      setActiveContent,
      setActiveTitle,
```

src/js/components/Tab/Tab.js
```diff
@@ -7,30 +7,48 @@ import { Box } from '../Box';
...
 
+    useEffect(() => {
+      if (active) {
+        setActiveContent(children);
+        let activeTitle;
+        if (typeof title === 'string') {
+          activeTitle = title;
+        } else {
+          activeTitle = activeIndex + 1;
+        }
```
halocline, line 40:
```suggestion
        const activeTitle = typeof title === 'string' ? title : activeIndex + 1;
```

src/js/components/Tabs/Tabs.js
```diff
@@ -49,33 +44,7 @@ const Tabs = forwardRef(
...
-        }
-
-        return cloneElement(tab, {
-          active: isTabActive,
-          onActivate: () => activateTab(index),
-        });
-      },
-      this,
-    );
+    const [activeTitle, setActiveTitle] = useState();
```
halocline, line 47: This seems disconnected from its usage. Recommend either placing together with together rest of state declarations, or directly ahead of its usage.

src/js/components/Tabs/Tabs.js
```diff
@@ -114,7 +83,21 @@ const Tabs = forwardRef(
           gap={theme.tabs.gap}
           {...tabsHeaderStyles}
         >
-          {tabs}
```
halocline, line 117: For readability, I kind of liked having `tabs` defined above.

Consider `const tabs = React.Children.map( ...`

src/js/components/Tabs/Tabs.js
```diff
@@ -114,7 +83,21 @@ const Tabs = forwardRef(
...
+          {React.Children.map(children, (child, index) => (
+            <TabsContext.Provider
+              // eslint-disable-next-line react/no-array-index-key
+              key={index}
```
halocline, line 89: `key` does not need to be specified here. `React.Children.map()` is already generating unique keys. So `key` can be removed.
```suggestion
```

src/js/components/Tabs/Tabs.js
```diff
@@ -114,7 +83,21 @@ const Tabs = forwardRef(
           gap={theme.tabs.gap}
           {...tabsHeaderStyles}
         >
-          {tabs}
+          {React.Children.map(children, (child, index) => (
+            <TabsContext.Provider
+              // eslint-disable-next-line react/no-array-index-key
```
halocline, line 88: This could be removed as well if the key={index} is removed. Missed this before.

```suggestion
```

src/js/components/Tabs/Tabs.js
taysea: Good clean up idea.

src/js/components/Tabs/Tabs.js
taysea: Good clean up, thanks!

ericsoderberghp:
> Thoughts on tests? A snapshot test for supplying a custom component as a child of Tabs might be beneficial.

+1

src/js/components/Tabs/Tabs.js
taysea: Agreed, will group it with the other declarations.

taysea:
> Nice improvements @taysea! Minor comments / suggestions.
> 
> Thoughts on tests? A snapshot test for supplying a custom component as a child of Tabs might be beneficial.
> 
> Lastly, does the custom component need to be a `<Tab />`? What happens if `<Tab />` isn't supplied as the custom component?

The custom component still needs to be a tab. If you supply a box or other component, it will place that in the Tab header. However, I think we are okay with this because this is being placed inside `Tabs`, so if you never supply a `Tab` it seems like it would be expected not to work.

ShimiSun:
>The custom component still needs to be a tab. If you supply a box or other component, it will place that in the Tab header. However, I think we are okay with this because this is being placed inside Tabs, so if you never supply a Tab it seems like it would be expected not to work.

Is there a way to document it? maybe add a warning? just throwing ideas.

taysea:
> > The custom component still needs to be a tab. If you supply a box or other component, it will place that in the Tab header. However, I think we are okay with this because this is being placed inside Tabs, so if you never supply a Tab it seems like it would be expected not to work.
> 
> Is there a way to document it? maybe add a warning? just throwing ideas.

I like the idea of adding a console warning. In addition, something could be added to Tabs documentation?

taysea:
> Nice work! The testing coverage didn't change and overall it looks good to be. Also, I agree that we should add a warning for when the custom component is not a `Tab`, and perhaps a test to go with it too.

Thanks, @JoeButler7! Do you have any suggestions re: the console warning? In my above comment I just added a description of what I tried but I'm not sure it's the best approach.

JoeButler7:
> > Nice work! The testing coverage didn't change and overall it looks good to be. Also, I agree that we should add a warning for when the custom component is not a `Tab`, and perhaps a test to go with it too.
> 
> Thanks, @JoeButler7! Do you have any suggestions re: the console warning? In my above comment I just added a description of what I tried but I'm not sure it's the best approach.

If we can't access the children of a custom component then it might be better to just put a note in the documentation about it. Checking the children was the only approach that I could think of and I got the same result that you did

ericsoderberghp:
I don't think we need a warning. It's problematic for a parent to assert something about a child.

PR #4256, opened 2020-07-01
Pre-commit fails on warning

### LINKED ISSUE #4218: Add no warnings for tests on pre-commit

JoeButler7:
#### What does this PR do?
Makes jest tests fail on `console.warn`, which in turn makes the pre-commit fail on warnings. Does this by adding a `setupTests.js` file and including this file as a config file for jest in `package.json`. Relevant documentation  https://create-react-app.dev/docs/running-tests/#initializing-test-environment
`src/setupTests.js` and `package.json` line 140
I added an example `console.warn` in one of the tests and then ran `yarn test` as well as `git commit` to ensure that the test will fail which will cause the pre-commit checks to fail
See above
Closes #4218 
Commit failing because of warning
Screen Shot 2020-07-01 at 3 26 57 PM

No
No
Backwards compatible


src/setupTests.js
```diff
@@ -0,0 +1,7 @@
+global.console.warn = message => {
+  throw message;
+};
+
+global.console.error = message => {
+  throw message;
+};
```
ShimiSun, line 7: Pre-commit already used to fail on error, why is this needed?

JoeButler7: I just included it because of the documentation I was looking at but you're right it is redundant. It has been removed on the most recent push

JoeButler7: I just included it because of the documentation I was looking at but you're right it is redundant. It has been removed on the most recent push

ShimiSun: Thanks. without this definition, how does the pre-commit knew until now to block the commits with errors?

JoeButler7: Its because jest automatically fails tests if they throw errors and the pre-commit blocks the commit if any tests fail

JoeButler7:
### Update:
Added back the code that will cause tests to fail on `console.error`. This was not previously implemented which can be shown by the following screen shot showing a test passing on pre-commit despite an error. 
Screen Shot 2020-07-07 at 3 03 07 PM
Jest will automatically fail on an unhandled exception but not `console.error` or `console.warn`. Adding this code in will turn `console.error` and `console.warn` into thrown exceptions which will thus make the test fail and cause the precommit to fail. Below is a screenshot of the same error from the previous screenshot but it fails due to the new code added
Screen Shot 2020-07-07 at 3 22 37 PM

src/setupTests.js
ShimiSun: Post discussion on this matter:
Currently, pre-commit fails because jest is throwing an exception with exit code 1 in case a test fail. 
This PR is enhancing this behavior so the pre-commit will fail not just for exceptions but for `console.error` and `console.warnings` messages.

ShimiSun:
Is the new config file also published on the production files?

JoeButler7:
> Is the new config file also published on the production files?

No it is not

ShimiSun:
Thank you for addressing this issue!

Issue #4258, opened 2020-07-01
TextArea does not support a11yTitle


JoeButler7:
### Expected Behavior
TextArea should support the `a11yTitle` prop for accesibility

### Actual Behavior
TextArea does not support the `a11yTitle` prop. The only way to have a label to to set the `aria-label` directly

### URL, screen shot, or Codepen exhibiting the issue
Accessibility test failing because a11yTitle is not implemented
<img width="979" alt="Screen Shot 2020-07-01 at 5 48 11 PM" src="https://user-images.githubusercontent.com/49867133/86294691-15624500-bbc3-11ea-9f77-335fce2992d8.png">

### Your Environment

- Grommet version: 2.14.0

PR #4262, opened 2020-07-02
Add a11yTitle support to TextArea component

JoeButler7:
#### What does this PR do?
Add support for the `a11yTitle` prop to the `textArea` component to comply with accessibility standards. Also adds accessibility test to `TextArea-test.js`
`src/js/components/TextArea/TextArea.js`
ran `yarn test TextArea-test` to ensure that the accessibility test passes and the `a11yTitle` is properly working
see above
#4258 
Yes
Yes
Backwards compatible


src/js/components/TextArea/doc.js
```diff
@@ -13,6 +13,9 @@ export const doc = TextArea => {
     .intrinsicElement('textarea');
 
   DocumentedTextArea.propTypes = {
+    a11yTitle: PropTypes.string.description(
+      'Custom title to be used by screen readers.',
```
ShimiSun, line 17:
```suggestion
      'Custom label to be used by screen readers. When provided, an aria-label will be added to the element.',
```
A consistent message with other components.

src/js/components/TextArea/index.d.ts
```diff
@@ -1,6 +1,8 @@
 import * as React from "react";
+import {A11yTitleType} from "../../utils";
```
ShimiSun, line 2:
```suggestion
import { A11yTitleType } from "../../utils";
```
I've filed this issue https://github.com/grommet/grommet/issues/4263

Issue #4264, opened 2020-07-03
[Typescript] Form onSubmit wrong event type expected


skad0:
### Expected Behavior

Form property `onSubmit` type should expect `event` argument as React.FormEvent extended with `value` and `touched` object properties

### Actual Behavior

Currently type for onSubmit expects `event` as React.FormEvent and doesn't know about `value` and `touched` props

### URL, screen shot, or Codepen exhibiting the issue

Codesandbox with working example: https://codesandbox.io/s/zen-curie-4shqk?file=/src/App.tsx

### Steps to Reproduce

1. add Form component
2. add onSubmit handler for Form
3. try to get `value` or `touched` property

### Your Environment

- Grommet version: 2.14.0
- Browser Name and version: -
- Operating System and version (desktop or mobile): -

PR #4265, opened 2020-07-03
Fix/form on submit property event type

skad0:
#### What does this PR do?

Fix form component onSubmit property event type to enriched with value and touched properties.

Besides small fix for lint errors of usage PropTypes.any and PropTypes.object in docs

Reviewer should start from index.d.ts and then follow to submit.tsx story, after they should check both doc.js files. Other was generated. 

Storybook

Storybook, typescript - form - all form controls

Second commit with fixes in doc.js I made just because pre-commit hook failed on this, so somehow it was in master.

Fixes https://github.com/grommet/grommet/issues/4264

I don't think so

I think yes, this should be very common issue for all typescript and Form users

It is backwards compatible, because it is a fix

### DIFF
--- a/src/js/components/Form/index.d.ts
+++ b/src/js/components/Form/index.d.ts
@@ -1,11 +1,16 @@
 import * as React from "react";
 
+export interface FormExtendedEvent<T = Element> extends React.FormEvent<T> {
+  value: Record<string, any>;
+  touched: Record<string, boolean>;
+}
+
 export interface FormProps<T> {
   errors?: {};
   infos?: {};
   messages?: {invalid?: string,required?: string};
   onChange?: (value: T) => void;
-  onSubmit?: ((event: React.FormEvent) => void);
+  onSubmit?: ((event: FormExtendedEvent) => void);
   onReset?: ((event: React.SyntheticEvent) => any);
   validate?: 'blur' | 'submit';
   value?: {};
--- a/src/js/components/TextInput/doc.js
+++ b/src/js/components/TextInput/doc.js
@@ -55,7 +55,7 @@ export const doc = TextInput => {
       suggestionIsOpen: PropTypes.string,
     })
       .description(
-        `Custom messages for TextInput. Used for accessibility by screen 
+        `Custom messages for TextInput. Used for accessibility by screen
         readers.`,
       )
       .defaultValue({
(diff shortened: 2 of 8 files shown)


src/js/components/Form/index.d.ts
```diff
@@ -1,11 +1,16 @@
 import * as React from "react";
 
+export interface FormExtendedEvent extends React.FormEvent {
+  value: Record<string, unknown>;
```
ShimiSun, line 4: A Record<K, T> is an object type whose property keys are K and whose property values are T. Doesn't it imply that the `unknown` in this case should be `any` instead?

skad0: Hi!
As far as I can understand `unknown` is a type-safe version of `any`. It says that “you can receive anything here and you should consider checking the type”.
But thinking about use-cases, I think you are actually right. It should be or `any` or the actual types. I will set it to any for now.

ShimiSun:
@jasnross this PR looks good to me, but I'll appreciate another pair of eyes here.

src/js/components/Form/index.d.ts
```diff
@@ -1,11 +1,16 @@
 import * as React from "react";
 
+export interface FormExtendedEvent extends React.FormEvent {
```
carolineartz, line 3: does it make sense that this take a type param?

i.e.,

```ts
export interface FormExtendedEvent<T = Element> extends React.FormEvent<T> {
  value: Record<string, any>;
  touched: Record<string, boolean>;
}
```

since the react type is

```ts
interface FormEvent<T = Element> extends SyntheticEvent<T> {
}
```

src/js/components/TextInput/doc.js
```diff
@@ -101,7 +101,8 @@ Only use this when the containing context provides sufficient affordance`,
       PropTypes.oneOfType([
         PropTypes.shape({
           label: PropTypes.node,
-          value: PropTypes.any,
+          // eslint-disable-next-line
+          value: PropTypes.any, // this is actually any
```
carolineartz, line 105: this isn't immediately clear to me... if this is saying that we have to disable eslint because the violation is that we can't have an `any` type, because this is intentionally type of `any`, then could we add that rule to the the disable comment? At first i was thinking that the violation was having inline comments lol Maybe could use "intentional" vs "actually"?

src/js/components/Form/index.d.ts
skad0: Actually, you are right, wasn't attentive

skad0:
🆙 
@carolineartz 
I pushed fixes, thanks.

ps: can't re-request it via github somehow

carolineartz:
@skad0 awesome thank you. LGTM! @ShimiSun i feel good about this PR :) :shipit:

ShimiSun:
Thank you guys for collaborating on this PR! 
And thank you for your contribution @skad0 , looking forward to seeing more from you!

PR #4266, opened 2020-07-04
RadioButton - add backgroundColor option in theme

### LINKED ISSUE #4187: Would like RadioButton input background color to be specified via theme
### Expected Behavior

Able to set the background color of the check input in my theme.

### Actual Behavior

Currently need to specify this via `extend`

Background: https://github.com/grommet/grommet-theme-hpe/pull/75#discussion_r437616833

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

netochaves:
#### What does this PR do?
Adds backgroundColor option in radioInput theme.
src/js/components/RadioButton/StyledRadioButton.js
Unit
change theme on src/js/components/RadioButton/stories/Theme.js
Fixes #4187 
Yes
Yes
It is backward compatible


src/js/components/RadioButton/doc.js
```diff
@@ -68,6 +68,11 @@ export const themeDoc = {
     type: 'string',
     defaultValue: '2px',
   },
+  'radioButton.check.backgroundColor': {
```
ShimiSun, line 71:
```suggestion
  'radioButton.check.background.color': {
```
For consistency, we usually name background with simply just `background` vs `backgroundColor`. This change help us scale under the umbrella of background. so ultimately, you'll have:
```
 'radioButton.check.background': {...}
 'radioButton.check.background.color': ''

radioButton: {
   check: {
      background: {
         color: ' ',
      }
   }
}
``` 

This change may affect the rest of the PR, but it should be an easy fix.

src/js/components/RadioButton/doc.js
```diff
@@ -68,6 +68,11 @@ export const themeDoc = {
...
+    description: 'The background color of the checked icon in the RadioButton.',
+    type: 'string',
```
ShimiSun, line 73:
```suggestion
    type: 'string | {dark: string, light: string}',
```
The color could also be a dark/light object.

src/js/components/RadioButton/__tests__/RadioButton-test.js
```diff
@@ -67,4 +67,23 @@ describe('RadioButton', () => {
     const tree = component.toJSON();
     expect(tree).toMatchSnapshot();
   });
+
+  test('bgColor', () => {
+    const customTheme = {
+      radioButton: {
+        check: {
+          backgroundColor: 'red',
```
ShimiSun, line 75:
```suggestion
          background: {
             color: 'red'
          }
```

src/js/components/RadioButton/doc.js
netochaves: Thanks, I'll fix it.

netochaves:
Hey @ShimiSun Thanks for your review, I think I addressed all your comments.

ShimiSun:
@halocline, since you filed the original issue, I'd love your feedback and confirmation whether this PR approach solves the use case for your design-system requirements. Feel free to leave a review comment.

halocline:
> @halocline, since you filed the original issue, I'd love your feedback and confirmation whether this PR approach solves the use case for your design-system requirements. Feel free to leave a review comment.

I'd be happy to. Will do.

src/js/components/RadioButton/doc.js
```diff
@@ -65,9 +65,14 @@ export const themeDoc = {
   },
   'radioButton.border.width': {
     description: 'The width size of the border of the RadioButton.',
-    type: 'string',
+    type: 'string | {dark: string, light: string}',
```
halocline, line 68: This looks like the type modification is being applied to the wrong property here. I think we want to leave `radioButton.border.width` as it was and instead apply the ` | {dark: string, light: string}` to `radioButton.check.background.color`

```suggestion
    type: 'string',
```

src/js/components/RadioButton/doc.js
```diff
@@ -65,9 +65,14 @@ export const themeDoc = {
...
     defaultValue: '2px',
   },
+  'radioButton.check.background.color': {
+    description: 'The background color of the checked icon in the RadioButton.',
+    type: 'string',
```
halocline, line 73: Per comment above, I think this color object was applied in the wrong location.

```suggestion
    type: 'string | {dark: string, light: string}',
```

src/js/themes/base.js
```diff
@@ -880,6 +880,9 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       },
       check: {
         radius: '100%',
+        background: {
+          // color: undefined,
+        },
```
halocline, line 885: @ShimiSun from a convention standpoint, do you want the entire `background` property commented out since none of the object properties are defined?

```
        // background: {
          // color: undefined,
        // },
```

ShimiSun: Yes, I agree it should be commented out.

src/js/components/RadioButton/doc.js
ShimiSun: Good catch.

netochaves: Good catch @halocline, thanks.

src/js/themes/base.js
netochaves: Makes sense, thanks.

netochaves: If we comment the entire `background` this line will throw an error because background will be `undefined`
https://github.com/grommet/grommet/blob/1a7e69e8406342456152dc4e840e9a20faea7959/src/js/components/RadioButton/StyledRadioButton.js#L63

So, we have to add some condition, e.g.:
```js
 background-color: ${props =>
    props.theme.radioButton.check.background &&
    props.theme.radioButton.check.background.color};
```

Do you want to follow this path?

halocline: Good call. That seems reasonable to me and would be consistent with approaches elsewhere in Grommet.

src/js/themes/base.d.ts
```diff
@@ -852,6 +852,7 @@ export interface ThemeType {
     };
     check?: {
       radius?: string;
+      background?: Object;
```
halocline, line 855: Small adjustment for additional type specificity. 

```suggestion
      background?: BackgroundType;
```

ShimiSun: I had a similar thought at the beginning, but I don't think this suggestion is accurate.
In this case, the background is not treated as a BackgroundType, we are not handling any other background props such as `image`, `repeat` `opacity`, or anything other than the background `color`.
I would agree that the type should be more explicit to the color prop as follows:
```suggestion
      background?: {
          color?: ColorType;
      }
```

PR #4271, opened 2020-07-07
Chart - changed `bound` doc description

ShimiSun:
Small refinement on the description of the `bounds` in Chart.


src/js/components/Chart/doc.js
```diff
@@ -13,7 +13,8 @@ export const doc = Chart => {
   DocumentedChart.propTypes = {
     ...genericProps,
     bounds: PropTypes.arrayOf(PropTypes.arrayOf(PropTypes.number)).description(
-      `The limits for the values, specified as a two dimensional array.
+      `The limits for the values, specified as a two dimensional array for each
```
ericsoderberghp, line 16: I think this reads as though each axis has a two dimensional array. Perhaps:
"The limits for the values, specified as a two dimensional array. The first array specifies the limits of the x-axis. The second array specifies the limits of the y-axis. For example: [[x-min, x-max], [y-min, y-max]].

PR #4279, opened 2020-07-08
Add list.hover.item.background to theme

taysea:
#### What does this PR do?

Adds `list.hover.item.background` to theme capabilities. This allows the user to specify the hover background color of a list item. Currently hover background is only possible with `onClickItem` applied to list. However, designs for the Design System call for a hover background regardless of `onClickItem` or not. This theme style would override the onClickItem background color.

Added custom theme with `list.hover.item.background` and then hovered over list item.

Add custom theme such as 
```
list: {
   hover: {
        item: {
           background: 'red',
        }
   }
}
```

There may be instances when a user wants a hover background color even when the list item is not clickable. This is helpful for lists with no borders that still allow the user to understand the boundaries of the item they are looking at.

Needed to complete https://github.com/hpe-design/design-system/pull/956

Yes, they have been.

Yes.

Backwards compatible.


src/js/components/List/List.js
```diff
@@ -24,6 +24,13 @@ const StyledList = styled.ul`
 
 const StyledItem = styled(Box)`
   ${props => props.onClick && `cursor: pointer;`}
+  &:hover {
+    ${props =>
+      props.theme.list.hover &&
+      props.theme.list.hover.item &&
+      props.theme.list.hover.item.background &&
+      backgroundStyle(props.theme.list.hover.item.background, props.theme)}
```
ericsoderberghp, line 32: I think this implies that any List will have per-item hover styling, whether or not that item is interactive in any way. Something seems a little off here. Perhaps I am missing something?

src/js/components/List/doc.js
```diff
@@ -149,6 +149,12 @@ export const themeDoc = {
     type: 'boolean | string | object',
     defaultValue: 'horizontal',
   },
+  'list.hover.item.background': {
+    description:
+      'Background of a list item on hover. Overrides global.hover.background.',
```
ShimiSun, line 154:
```suggestion
      'The background of a list item on hover. Overrides global.hover.background.',
```

src/js/components/List/List.js
taysea: The intention of this is to allow a list to have per-item hover styling, specifically when it doesn't have `onClick`. The motivation for this was because of HPE design system designs that called for a list item hover state regardless of if it has an onClick or not.

Some list items may have an action button, in which case the entire list item will not have an `onClick`. However, having the the hover behavior helps with the visual experience of differentiating list items. If we don't feel like this is a behavior we want to support though, we would want to revisit the List designs to modify what is currently noted as hover behavior.

cliffhall: Will this work on Menu items?

taysea: Hey, @cliffhall! This will specifically be affecting the List component. Menu is not built with List, so this change will not be affecting Menu. 

If you're looking for styling customization on hover for Menu that you're unable to achieve, feel free to file a new issue!

ShimiSun:
Should this PR be a Draft? Is it still WIP?

taysea:
> Should this PR be a Draft? Is it still WIP?

I left it as a draft for now because the motivation to add it has changed. As of now for the design system, we won't be using hover styling on list items that don't have `onClickItem`. However, if we think this theming behavior would be beneficial for general grommet users, I can convert it back to an open PR.

taysea:
Closing this for now since the immediate need is no longer present. We can reopen in future if necessary.

PR #4280, opened 2020-07-08
Allow opacity to affect background image

### LINKED ISSUE #4121: Opacity not working for background image in Box
"opacity" is not working when used in the background property of the Box component if "image" is set. However, it does work when "color" is used instead of "image" for the background.

### Expected Behavior

Opacity works when using "image" inside the background property of the Box component.

### Actual Behavior

Opacity only works with "color".

https://codesandbox.io/s/box-opacity-4t6ym?file=/src/App.js

### Steps to Reproduce

1. Use the Box component.
2. Specify the background property.
3. Use the object {"image": "url(...)", "opacity": 0.5} in the background.

### Your Environment

Webstorm IDE

- Grommet version: 2.13.0
- Operating System and version: Windows 10

taysea:
#### What does this PR do?

Allows Box background opacity to have effect when image is used. Some context on the approach can be found in these discussions:

1. When background image is applied to a div, we can't directly adjust the image's opacity without also adjusting the div content's opacity. Therefore, we need to use a psuedo :before selector to just affect the image: https://stackoverflow.com/questions/32490798/apply-opacity-to-background-image-but-not-text/32491043

2. A z-index of -1 needs to be applied to the psuedo selector and a z-index of 0 needs to be applied to the div to ensure that the image is not appearing on top of the div content: https://stackoverflow.com/questions/23046368/css-pseudo-element-appearing-over-top-of-content

src/js/utils/background.js

Added snapshot and added opacity to Box background storybook with image.

Create a Box with this code, swap out opacity for any value you want ('weak', 0.5, etc.):
```
<Box
        pad="small"
        background={{
          image:
            "url(https://images.unsplash.com/photo-1487088678257-3a541e6e3922?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=2167&q=80)",
          opacity: "strong"
        }}
      >
        image
</Box>
```

Opacity was not having effect when images where used as background. It only had effect on background colors.

Closes #4121 

No.

Yes.

Backwards compatible.


src/js/utils/background.js
```diff
@@ -140,12 +140,39 @@ export const backgroundStyle = (backgroundArg, theme, textColorArg) => {
...
+          ? `
+        position: relative;
+        z-index: 0;
+        &:before {
+        content: '';
+        position: absolute;
+        background-image: ${background.image};
+        background-repeat: ${background.repeat || 'no-repeat'};
+        background-position: ${background.position || 'center center'};
+        background-size: ${background.size || 'cover'};
```
ShimiSun, line 162: Whether or not opacity is set all these styles are in common for the two cases, can we optimize the structure of the background a little better so there will be less code? maybe have a function for opacityStyle?

taysea: Good thought, let me look into optimizing this.

src/js/utils/background.js
```diff
@@ -138,14 +138,39 @@ export const backgroundStyle = (backgroundArg, theme, textColorArg) => {
...
-      ${backgroundColor ? `background-color: ${backgroundColor};` : ''}
+    `;
+
+    // allow both background color and image, in case the image doesn't fill
+    // when image and opacity are used together, we need to use pseudo :before
+    // to ensure that only image and background color are affected by opacity
+    // but not the container contents
+    return css`
       ${textColor ? `color: ${textColor};` : ''}
+      ${!background.opacity
```
ShimiSun, line 155: Can we test the scenario of having background-image, background-color, and background-opacity? And make sure the opacity applies to both the image and the color.

src/js/utils/background.js
```diff
@@ -138,14 +138,39 @@ export const backgroundStyle = (backgroundArg, theme, textColorArg) => {
...
+        ? backgroundStyles
+        : `position: relative;
+        z-index: 0;
```
ShimiSun, line 158: If the caller sets the `z-index` of the container to '-2' will this still gonna work?

src/js/utils/background.js
taysea: Yes, I will test it.

src/js/utils/background.js
taysea: Good comment, let me test that out and confirm.

src/js/utils/background.js
taysea: I checked this and it applies to both the image and background color : )

src/js/utils/background.js
taysea: With psuedo elements, when we create the z-index with absolute positioning, it puts it relative to the container. So in this case the image is stacked underneath the div content. If the caller sets the `z-index` of the container to `-2`, the Box itself may not be visible on the screen because it's falling underneath the general grommet wrapper, but the relative stacking between the box content and the image holds.

Reference: https://stackoverflow.com/questions/3032856/is-it-possible-to-set-the-stacking-order-of-pseudo-elements-below-their-parent-e

PR #4300, opened 2020-07-10
Filter out 'size' prop from leaking through to the DOM in <Text>

### LINKED ISSUE #4003: FormField is leaking `size` into the DOM.
When applying FormField label styles (see #3998), for example`size`, the prop is being leaked to the DOM while it shouldn't. 

_Originally posted by @ericsoderberghp in https://github.com/grommet/grommet/pull/3998_
```
Not part of this review per-se, but we shouldn't be leaking `size` into the DOM.
```

MikeKingdom:
Fixes #4003

#### What does this PR do?
Filter out 'size' prop from leaking through to the DOM in <Text>

StyledText.js

Run JEST tests and storybook

Use storybook to check that components which use <Text> look correct and have the correct text size

&lt;Text&gt; passes 'size' to StyledText which uses 'styled' to convert it into CSS for the font. However, styled-components silently passes any props which are valid HTML props to the underlying target element. Since size is a valid HTML prop for inputs and selects, it gets passed onto the target element (no matter what the tag actually is.)

styled-components 5.1 introduced a config option for 'styled' which lets you pass a shouldForwardProp function to filter out props you don't want to make it to the element. Grommet had a peer dependency on styled-components >= 5.X so this was upgraded to 5.1
 
#4003 

No

No


src/js/components/FormField/__tests__/__snapshots__/FormField-test.js.snap
```diff
@@ -761,7 +760,6 @@ exports[`FormField disabled with custom label 1`] = `
     >
       <label
         className="c2"
-        size="small"
```
ShimiSun, line 764: which node version are you using? make sure you are using ^12.18.

MikeKingdom: Thanks. Updated node to 12.18.2
What's the issue with the review?

ShimiSun: The snapshots should not change, running `yarn test-update` on node 12.18 should fix the snapshots, can you update and push so the diff would point only to the relevant files?

MikeKingdom: I'm confused. Don't these snapshots need to change since the whole point is FormField will no longer end up with size="small in it's DOM?

ShimiSun: From a quick look I've taken, I don't think that components such as Avatar, Sidebar, and Calendar should change their snapshots, unless you have a deeper understanding to share on those changes.

MikeKingdom: Like FormField does, I believe they each use &lt;Text&gt; with a size prop. As a result, the DOM they produce will include a leaked size attribute. Therefore, their snapshots would also have size properties that need to go away as with FormField.

ShimiSun: Thanks for the insight, I'll need to dive a little into it to ensure that is what's happening.

ShimiSun:
One question regarding this comment:
>Since size is a valid HTML prop for inputs and selects, it gets passed onto the target element (no matter what the tag actually is.)

I agree `size` is a valid HTML prop for inputs and selects, in this case, StyledText is using a span though (and not an input tag), is it correct to say that `size` also goes through span and not just input and selects tags?

MikeKingdom:
@ShimiSun Yes. Even when it's a span it still leaks through. styled() doesn't care what the element type is when it passes along props. It simply cares that the props are valid HTML props in general. In the FormField case it's creating a label with &lt;Text as="label" &gt;

PR #4302, opened 2020-07-10
Increases testing coverage on `Chart`

JoeButler7:
#### What does this PR do?
Increases the testing coverage on `Chart` component and also adds an accessibility test. Coverage changes are shown in the table below
|            | Before           | After            |
|------------|------------------|------------------|
| Statements | 74.29% (234/315) | 81.59% (257/315) |
| Branches   | 63.8% (178/279)  | 70.97% (198/279) |
| Functions  | 70.91% (39/55)   | 70.91% (39/55)   |
| Lines      | 77.74% (227/292) | 84.93% (248/292) |
`src/js/components/Chart/__tests__/Chart-test.js`
Ran `yarn test Chart-test` to ensure that all the tests passed and that coverage was increased
See above
Coverage before the new tests were added
Screen Shot 2020-07-10 at 3 03 02 PM
Coverage after the new tests were added
Screen Shot 2020-07-10 at 2 46 55 PM

No
No
Backwards compatible


src/js/components/Chart/__tests__/Chart-test.js
```diff
@@ -171,8 +188,84 @@ describe('Chart', () => {
...
+  });
+
+  test('calcs large thickness', () => {
+    const vals = [];
+    for (let i = 0; i < 8; i += 1) {
+      vals[i] = [];
+      vals[i].push(1);
+      vals[i].push(2);
+      vals[i].push(3);
+    }
```
jcfilben, line 219: Not sure why vals[i] is being pushed to 3 times. Also you do something like this instead so you don't have to loop:
> let vals = Array(8).fill(1);

JoeButler7: `vals[i]` is being pushed three times because `vals` needs to be a 2D array with each element being an array of 3 numbers, so we need to push three numbers into each index. Also I tried to use `Array.fill` but I wasn't able to get it to work with filling 2D arrays

MikeKingdom: You could probably do
let vals = Array(8).fill([1, 2, 3])

JoeButler7: @MikeKingdom Thanks that worked!

src/js/components/Chart/__tests__/Chart-test.js
```diff
@@ -171,8 +188,54 @@ describe('Chart', () => {
...
+
+  test('calcs with negative min', () => {
+    const result = calcs([
+      [1, -2, -2],
+      [2, 2, 2],
+    ]);
+    expect(result).toMatchSnapshot();
+  });
+
+  test('calcs large thickness', () => {
```
ShimiSun, line 212: Why did you choose the term thickness? Chart has a `thickness` prop, so it might be a little confusing.

Issue #4275, opened 2020-07-07
[TS] RangeInput onChange type is missing generic param T


ShimiSun:
### Expected Behavior
RangeInput onChange should support the onChange inputs.

### Actual Behavior
The issue was taken from StackOverflow, read this for more details:
https://stackoverflow.com/questions/62124013/what-is-the-right-type-definition-to-use-for-the-onchange-event-for-grommets-ra

PR #4304, opened 2020-07-11
Bug fix: RangeInput onChange

StheffanyHadlich:
#### What does this PR do?

This PR solves #4275 adding a type to `RangeInput onChange`

### DIFF
--- a/src/js/components/RangeInput/index.d.ts
+++ b/src/js/components/RangeInput/index.d.ts
@@ -7,7 +7,7 @@ export interface RangeInputProps {
   min?: number | string;
   max?: number | string;
   name?: string;
-  onChange?: ((event: React.ChangeEvent) => void);
+  onChange?: ((event: React.ChangeEvent<HTMLInputElement>) => void);
   step?: number;
   value?: number | string;
 }
--- a/src/js/all/stories/typescript/AllComponents.tsx
+++ b/src/js/all/stories/typescript/AllComponents.tsx
@@ -163,7 +163,7 @@ const Components = () => {
       />
       <RangeInput
         value={rangeInputValue}
-        onChange={event => setRangeInputValue(event.target.value)}
+        onChange={event => setRangeInputValue(parseInt(event.target.value, 10))}
       />
       <Stack>
         <Box direction="row" justify="between">


src/js/components/RangeInput/index.d.ts
```diff
@@ -7,7 +7,7 @@ export interface RangeInputProps {
   min?: number | string;
   max?: number | string;
   name?: string;
-  onChange?: ((event: React.ChangeEvent) => void);
+  onChange?: ((event: React.ChangeEvent<HTMLInputElement>) => void);
```
ShimiSun, line 10: Thank you for this PR! 
Can you please add a few words why you think this fixes the problem?

ShimiSun:
@carolineartz I'd love your insight here.

carolineartz:
This will fix the problem...but I'm wondering why we're re-defining the types at all.... the `onChange` for an input has this type already defined internally, with the properly typed generic.  On first glance all of the props in `RangeProps` are already on `input` element and the `RangeInput` only wraps `input` to intercept the values for `formContext` -- doesn't add anything custom. 

Would this just be fixed by getting rid of `RangeInputProps` all together and instead:

```ts
declare const RangeInput: React.FC<JSX.IntrinsicElements['input']>;

const RangeInput = forwardRef<HTMLInputElement, JSX.IntrinsicElements['input']>(
  ({ name, onChange, onFocus, onBlur, value: valueProp, ...rest }, ref) => {
  ...
```

cc @ShimiSun

ShimiSun:
How will this approach work with props that are not part of the input IntrinsicElements? Such as `a11yTitle` prop?

src/js/components/RangeInput/index.d.ts
taysea: Found a resource here that seems to mirror this solution: https://blaipratdesaba.com/react-typescript-cheatsheet-form-elements-and-onchange-event-types-8c2baf03230c

taysea:
> @carolineartz, How will this approach work with props that are not part of the input IntrinsicElements? Such as `a11yTitle` prop?

When I comment out the `onChange` from RangeInputProps, I do see the intrinsic property and it matches the proposed fix. Since we are not making any modifications to the handler should we remove it from the interface? or is the convention that we want to display all handlers regardless of if we are modifying them or not (if this is the case we can move forward with the proposed change).

ShimiSun:
@taysea good question, since we are not making any modifications on the handler, we should fall back to its default intrinsic definition, so unless we are adding more functionalities to it, I think we can remove it.

taysea:
> @taysea good question, since we are not making any modifications on the handler, we should fall back to its default intrinsic definition, so unless we are adding more functionalities to it, I think we can remove it.

Great, I've opened https://github.com/grommet/grommet/pull/4429 which implements these changes.

ShimiSun:
Thank you @carolineartz, @taysea & @StheffanyHadlich for working on this PR and getting it to the finish line!

PR #4309, opened 2020-07-14
Add opacity for placeholder on FF

taysea:
#### What does this PR do?
Adds `opacity: 1` for input placeholders for FF. By default, FF applies a lowered opacity to placeholder text causing it to render differently than it does on Safari, Chrome, etc. This allows the cross browser behavior to match.

Resources:
- https://ilikekillnerds.com/2014/10/firefox-placeholder-text-looking-lighter-browsers/
- https://stackoverflow.com/questions/18415765/font-color-of-placeholder-difference-between-firefox-23-0-1-chrome-23-0-1271-64

src/js/components/TextInput/StyledTextInput.js

Opened Simple TextInput storybook and added a placeholder locally. Compared in Chrome and FF before and after change. 

Create a TextInput with placeholder text. Open it in Chrome, Safari, and FF. If these changes are removed, the FF placeholder will be a lighter color. With these changes, all will be the same.

Closes https://github.com/hpe-design/design-system/issues/972

https://github.com/hpe-design/design-system/issues/972

**Before** (top: Chrome, bottom: FF):
Screen Shot 2020-07-14 at 2 53 58 PM

**After**(top: Chrome, bottom: FF):
Screen Shot 2020-07-14 at 2 54 15 PM

No.

Yes. 

In order to provide cross-browser consistency, an opacity of 1 has been added to Firefox’s rendering of placeholder text to override its default placeholder opacity of 0.54, which other browsers do not apply. This change ensures the visual appearance of placeholder text is consistent across browsers.

Backwards compatible.


halocline:
On the surface the direction here looks appropriate. I will review closer tomorrow, however two things come to mind immediately:

1) Wondering about the discreet implementations in both TextInput and MaskedInput vs. implementing in a 'placeholderStyle' in 'utils' which is then referenced by both?

2) I do think this change should be documented in the release notes. This will result in a visual regression and will appear for any users testing for them, so we should call this out in the notes if one were to investigate.

I am not familiar enough for how Grommet has handled equivalent CSS resets overriding User Agent defaults in the past and would like to look into that more.

In general, as a user, my desire is that if I specify a color, I'd want that color to "just work" and not have to worry about browser edge cases for which the library does not account, resulting in me having to cover in my implementations. This PR is definitely in the "just works" spirit, but want to check in with the precedents set up to this point.

taysea:
> On the surface the direction here looks appropriate. I will review closer tomorrow, however two things come to mind immediately:
> 
> 1. Wondering about the discreet implementations in both TextInput and MaskedInput vs. implementing in a 'placeholderStyle' in 'utils' which is then referenced by both?

I think this is a great idea. I will confirm that they are the same and then pull it into utils.

> 2. I do think this change should be documented in the release notes. This will result in a visual regression and will appear for any users testing for them, so we should call this out in the notes if one were to investigate.
> 

That's a good point. Since it's aligning to other browser behavior, I hadn't thought of it as a regression but you are right for those that are doing testing. 

> I am not familiar enough for how Grommet has handled equivalent CSS resets overriding User Agent defaults in the past and would like to look into that more.
> 

From what I've seen, they typically take place like this in a styled component. You can see other examples in `StyledRangeInput`, `StyledTextArea`, `baseStyle` in `styles.js` (as well as others but hopefully these are helpful starting points).

> In general, as a user, my desire is that if I specify a color, I'd want that color to "just work" and not have to worry about browser edge cases for which the library does not account, resulting in me having to cover in my implementations. This PR is definitely in the "just works" spirit, but want to check in with the precedents set up to this point.

Agreed on the "just work" mentality and think this change helps to ensure cross browser behavior matches.

src/js/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
```diff
@@ -5962,6 +5962,11 @@ exports[`DataTable search 2`] = `
   outline: none;
 }
 
+.c2:-moz-placeholder,
+.c2::-moz-placeholder {
+  opacity: 1;
+}
+
```
ShimiSun, line 5969: All the snapshot changes with this CSS change make sense to me, though we would need to communicate it well on the release notes to call out this is not a backward compatibility issue but an educated move by grommet for cross-browser consistency. Please come up with 1-2 lines for the release notes that explain the impact this PR might have on existing applications to call it out in advance.

src/js/components/MaskedInput/StyledMaskedInput.js
```diff
@@ -39,6 +39,11 @@ export const StyledMaskedInput = styled.input`
     outline: none;
   }
 
+  &:-moz-placeholder, // FF 18-
+  &::-moz-placeholder { // FF 19+
+    opacity: 1;
+  }
```
ShimiSun, line 45: Code duplication with StyledTextInput, we should extract it to utils.

src/js/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
taysea: Sounds good, I will do that!

src/js/components/MaskedInput/StyledMaskedInput.js
taysea: @ShimiSun I've extract the duplication to utils and added it to `inputStyle`. The snapshot changes are not truly changes but rather just the css being a different order from what it previously was.

src/js/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
taysea: In order to provide cross-browser consistency, an opacity of 1 has been added to Firefox’s rendering of placeholder text to override its default placeholder opacity of 0.54, which other browsers do not apply. This change ensures the visual appearance of placeholder text is consistent across browsers.

PR #4311, opened 2020-07-15
Fix storybook theme that was mistakenly deleted

negreirosleo:
#### What does this PR do?
Fix Storybook Theme
Start with `storybook/theme.js`
Checking that its running properly with `yarn build-storybook`
Just run `yarn build-storybook` and check that its working as it should
It was remove in a previous PR due to a small confusion caused by cache. This PR recovers it
image

No
No
No

### DIFF
--- a/storybook/manager.js
+++ b/storybook/manager.js
@@ -1,6 +1,8 @@
 import { addons } from '@storybook/addons';
+import theme from './theme';
 
 addons.setConfig({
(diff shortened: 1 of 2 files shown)


ShimiSun:
The colors on the screen shot above may look a little off, but I've confirmed the deployment is using the correct log and colors. 
image

PR #4320, opened 2020-07-17
Calendar - fix propagation of 'disabled' prop to Button

### LINKED ISSUE #4146: [TS] Calendar: disabled dates (buttons) receive number instead of boolean for disabled prop
### Expected Behavior
Date buttons receive boolean value if disabled instead of number.

### Actual Behavior
If you define disabled dates via prop you get a console error/warning ```Warning: Failed prop type: Invalid prop `disabled` of type `number` supplied to `Button`, expected `boolean`.```

Have a look a this sandbox: https://codesandbox.io/s/react-typescript-0v6jy?file=/src/index.tsx and watch the console (see screenshot)

Bildschirmfoto 2020-06-04 um 18 38 41

### Your Environment

See code sandbox

- Grommet version: latest
- Browser Name and version: all

britt6612:
Initially I thought this was a TS issue but it really was just PropTypes warning
I originally made the TS story for Calendar but this isnt really needed but good as an example to 
look in the console...
#### What does this PR do?
This PR adds to the `PropTypes` in `button`
button/doc.js
storybook with checking console
I added a new TS story to display that this issue has been solved in the console is opened there should not be anything in there..

The before error can be found here: 
https://codesandbox.io/s/react-typescript-2f3dh
Calendar under the hood is using `button` is which it is passing `disabled`
When the disabled prop was being used in Calendar it was checking the definition that was
defined in `button` which was just a boolean.. Therefore the error was 
```
Warning: Failed prop type: Invalid prop `disabled` of type `number` supplied to `Button`, expected `boolean`.
```
I added to the prop types of `button` to also accept a number since that is what is happening in `Calendar`
closes #4146 
auto
no
backwards compatible

### DIFF
--- a/src/js/components/Button/doc.js
+++ b/src/js/components/Button/doc.js
@@ -46,7 +46,7 @@ export const doc = Button => {
     color: colorPropType.description(
       'Fill color for primary, label color for plain, border color otherwise.',
     ),
-    disabled: PropTypes.oneOfType([PropTypes.bool, PropTypes.number])
+    disabled: PropTypes.bool
       .description('Whether the button is disabled.')
       .defaultValue(false),
     fill: PropTypes.oneOfType([
--- a/src/js/components/Button/doc.js
+++ b/src/js/components/Button/doc.js
@@ -34,7 +34,7 @@ export const doc = Button => {
     color: colorPropType.description(
       'Fill color for primary, label color for plain, border color otherwise.',
     ),
-    disabled: PropTypes.oneOfType([PropTypes.number, PropTypes.bool])
+    disabled: PropTypes.bool
       .description('Whether the button is disabled.')
       .defaultValue(false),
     fill: PropTypes.oneOfType([
--- a/src/js/components/Button/doc.js
+++ b/src/js/components/Button/doc.js
@@ -25,6 +25,18 @@ export const doc = Button => {
 
   DocumentedButton.propTypes = {
     ...genericProps,
+    children: PropTypes.oneOfType([
+      PropTypes.func,
+      PropTypes.object,
+      PropTypes.node,
+    ]).description(
+      `Function that can be called to render the visual representation.
+      Button can take in Children as a function, node, or object. 
+      For example hover can be passed as an object that would 
+      then return a react element.
+      \`children={({ hover }) => <Box...>{...}</Box>}\`
+      `,
+    ),
     active: PropTypes.bool
       .description('Whether the button is active.')
       .defaultValue(false),
--- a/src/js/components/Button/doc.js
+++ b/src/js/components/Button/doc.js
@@ -46,7 +46,7 @@ export const doc = Button => {
     color: colorPropType.description(
       'Fill color for primary, label color for plain, border color otherwise.',
     ),
-    disabled: PropTypes.bool
+    disabled: PropTypes.oneOfType([PropTypes.bool, PropTypes.number])
       .description('Whether the button is disabled.')
       .defaultValue(false),
     fill: PropTypes.oneOfType([
--- a/src/js/components/Button/doc.js
+++ b/src/js/components/Button/doc.js
@@ -34,7 +34,7 @@ export const doc = Button => {
     color: colorPropType.description(
       'Fill color for primary, label color for plain, border color otherwise.',
     ),
-    disabled: PropTypes.bool
+    disabled: PropTypes.oneOfType([PropTypes.number, PropTypes.bool])
       .description('Whether the button is disabled.')
       .defaultValue(false),
     fill: PropTypes.oneOfType([
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -405,7 +405,7 @@ const Calendar = forwardRef(
               plain
               tabIndex={-1}
               active={active && active.getTime() === day.getTime()}
-              disabled={!!dayDisabled}
+              disabled={dayDisabled}
               onClick={() => {
                 selectDate(dateString);
                 // Chrome moves the focus indicator to this button. Set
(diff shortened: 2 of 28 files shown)


src/js/components/Calendar/stories/typescript/Simple.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const onSelect = nextDate => {
+    setDate(nextDate !== date ? nextDate : undefined);
+  };
+
+  return (
+    <Grommet theme={grommet}>
+      <Box align="center" pad="large">
+        <Calendar
+          bounds={['2020-06-05', '2020-06-12']}
+          disabled={['2020-08-07']}
```
taysea, line 20: Would this be an array of strings as opposed to a number?

britt6612: Good question! 😄 

In `Calendar` it is passing `daydisabled` to the disabled prop in `button` which is actually a number. 

https://github.com/grommet/grommet/blob/master/src/js/components/Calendar/Calendar.js#L390
so if you were to console what is being passed to disabled you will be able to see that there is a `number` as well as the `boolean`.  

This is why having a number in `button` for props should be sufficient. Then the `Calendar` `proptypes` will check if it is an array of strings

taysea: Thanks for this explanation, I see how it all connects now. This line in the Calendar utils is helpful for an explanation of why a certain number is returned:

https://github.com/grommet/grommet/blob/master/src/js/components/Calendar/utils.js#L93

ShimiSun:
I have been thinking about this PR, and I think it wouldn't make sense for users that are using Button (without Calendar) to guess what the number type of `disabled` means. 
I think the more right approach for this issue is to fix the root cause of the problem where Calendar shouldn't send a number type to Button `disabled` and use `!!` to transform the number value to a boolean value, and hence the error wouldn't occur because button `disabled` will only receive boolean value. Let me know what you think.

britt6612:
> I have been thinking about this PR, and I think it wouldn't make sense for users that are using Button (without Calendar) to guess what the number type of disabled means.
> I think the more right approach for this issue is to fix the root cause of the problem where Calendar shouldn't send a number type to Button disabled and use !! to transform the number value to a boolean value, and hence the error wouldn't occur because button disabled will only receive boolean value. Let me know what you think.

@ShimiSun this sounds like a better solution its more towards to root of the issue rather then me just adding to props.. I will investigate more on the number its sending and where all that is cover and then convert that to a boolean

britt6612:
need to investigate further as to why it is changing the snap shots

src/js/components/DateInput/__tests__/__snapshots__/DateInput-test.js.snap
```diff
@@ -2126,6 +2126,7 @@ exports[`DateInput format inline 1`] = `
                 <button
                   aria-label="Sun Jun 28 2020"
                   className="c16"
+                  disabled={false}
```
ShimiSun, line 2129: What triggered all the changes in the snapshot?

britt6612: Snap Shots are updated to add disabled={false} since we are now passing a boolean which is going to be true or false
Since all of the previous tests for calendar were not passing in disabled as a prop this is why disabled={false} is now being adding to the snapshots
Similar behavior is seen in snapshots for Carousel and DateInput.

britt6612: I commented this above if it doesnt make sense I can elaborate

britt6612: oh wait just noticed the new test failed..

src/js/components/Button/doc.js
```diff
@@ -46,7 +46,7 @@ export const doc = Button => {
     color: colorPropType.description(
       'Fill color for primary, label color for plain, border color otherwise.',
     ),
-    disabled: PropTypes.bool
+    disabled: PropTypes.oneOfType([PropTypes.bool, PropTypes.number])
```
taysea, line 49: Think we can remove this now that only a bool is going to get passed to Button

britt6612: yeah I was doing this just to revert back when adding a `disable test`

src/js/components/Calendar/Calendar.js
```diff
@@ -405,7 +405,7 @@ const Calendar = forwardRef(
               plain
               tabIndex={-1}
               active={active && active.getTime() === day.getTime()}
-              disabled={dayDisabled}
+              disabled={!!dayDisabled}
```
ShimiSun, line 408:
```suggestion
              disabled={dayDisabled && !!dayDisabled}
```
I'm hoping this change will fix the changes on the snapshots. Let's give it a try.

britt6612: awe yes so it will only put `disabled={true}` on snapshots!

src/js/components/Calendar/stories/typescript/Simple.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const onSelect = nextDate => {
+    setDate(nextDate !== date ? nextDate : undefined);
+  };
+
+  return (
+    <Grommet theme={grommet}>
+      <Box align="center" pad="large">
+        <Calendar
+          bounds={['2020-08-05', '2020-08-12']}
+          disabled={['2020-08-07']}
```
taysea, line 20:
```suggestion
          disabled={['2020-08-07T00:00:00-08:00']}
```

adding a time stamp to mirror what is done in tests and ensure proper day is disabeld.

britt6612:
> Everything code-wise looks good here, but for the Calendar storybook example we're setting the bounds to august but the Calendar is showing July by default, so it's a little strange. I think we should extend the bounds and make sure the default month being disabled matches the month of the disabled date we are showing (in this case that would be August)

I took out the bounds in the storybook example. Im also not sure why it would not let you move the `calendar` to August this same behavior is also here https://storybook.grommet.io/?path=/story/calendar--custom-header 
unless I am misunderstanding what the bounds mean.

ShimiSun:

PR #4321, opened 2020-07-17
Pass overflow prop into FormField content box

### LINKED ISSUE #4229: Forced "overflow: hidden" in FormField component (FormFieldContentBox)
# Expected Behavior

I should be able to have a custom input child component that overflows the bounds of the FormField if desired.
My use case: https://github.com/ubilabs/react-geosuggest

### Actual Behavior

The `<FormFieldContentBox>` of FormField is hardcoded with `overflow: hidden`.
See: https://github.com/grommet/grommet/blob/master/src/js/components/FormField/FormField.js#L235

 It'd be nice to be able to override this behavior.

Similar issue: https://github.com/grommet/grommet/issues/2018

- Grommet version: 2.14.0^

MikeKingdom:
#### What does this PR do?
Today, FormField passes a hardcoded overflow="hidden" to its content Box for the FormField body.
This change removes the hardcoded overflow value and adds a contentProps property which allows properties to be directly set on the content box. These contentProps can override any of the same properties coming from the theme or the pad property

FormField.js

Jest tests ran (passed) and test case for specifying overflow added.
Storybook.

Storybook.

Fixes #4229 

Done

Can be as an added feature

This no longer passes overflow="hidden" but this should have minimal effect on the content.


MikeKingdom:
When making the change I made sure the value from the theme would still be honored.

Honestly the documentation for grommet components is a bit lacking and I often have to look at the source to see what is possible and what grommet will do with it. This is certainly an example of that.

While valuable to be able to set these props via the theme, this need seems more likely to be used in a few certain cases and therefore you wouldn't want it generalized in your theme. It seems like in this case a property would be useful so you don't have to affect all FormFields in the app or resort to a one-of theme applied to just a single component. I've run into this decision a few times while working on an app that uses grommet. It seems useful and consistent for the grommet FormField to support this as a property.

ShimiSun:
>Honestly the documentation for grommet components is a bit lacking and I often have to look at the source to see what is possible and what grommet will do with it. This is certainly an example of that.

Our goal is to find the places that the documentation is lacking and improve them, so feel free to file an issue or submit documentation fixes to improve the overall user experience. This PR is certainly the right place to add docs for `theme.formField.content`, or a separate PR will be fine too.

>While valuable to be able to set these props via the theme, this need seems more likely to be used in a few certain cases, and therefore you wouldn't want it generalized in your theme. It seems like in this case a property would be useful so you don't have to affect all FormFields in the app or resort to a one-of theme applied to just a single component. I've run into this decision a few times while working on an app that uses grommet. It seems useful and consistent for the grommet FormField to support this as a property.

Interesting perspective on the usage of `overflow`, thank you for sharing. I'm personally inclining to keep the FormField API more simple, but if other users agree to this approach I'm willing to take it. Waiting to see what @ericsoderberghp or @halocline thinks.

ericsoderberghp:
A couple of quick thoughts on this:

- I'm wondering if we should remove the setting of `overflow` on the content box entirely. That would allow child elements to do whatever they wanted.
- A pattern that's emerging for grommet components that are compositional is to be more flexible and have something like `contentProps` as a way to pass any arbitrary props into the content Box. We do this with `dropProps` in several components and this is the direction we are heading for DataChart as well. The current `pad` would need to remain for backwards compatibility but `contentProps={{ pad: ... }}` would be equivalent.

MikeKingdom:
I like the idea of removing the hardcoded property

src/js/components/FormField/doc.js
```diff
@@ -34,6 +34,7 @@ export const doc = FormField => {
       of name, value, onChange (event => {}), while event has either event.value
       or event.target.value.`,
     ),
+    contentProps: PropTypes.object.description('Any valid Box prop.'),
```
ShimiSun, line 37: We should add on the description where would that be applied.

src/js/components/FormField/FormField.js
```diff
@@ -179,26 +180,30 @@ const FormField = forwardRef(
...
+        fieldContentProps.background = formFieldTheme.disabled.background;
       }
     }
-    // Removing margin from contentProps because margin should be applied
+    // Removing margin from fieldContentProps because margin should be applied
     // on containing <FormFieldContentBox>
-    // contentProps.margin = undefined;
-    // contents = <Box {...contentProps}>{contents}</Box>;
+    // fieldContentProps.margin = undefined;
+    // contents = <Box {...fieldContentProps}>{contents}</Box>;
```
halocline, line 199: @MikeKingdom I believe these should be removed altogether. I missed removing these when resolving a merge conflict two weeks back. I can handle in a separate PR if you prefer.

src/js/components/FormField/FormField.js
```diff
@@ -250,9 +255,9 @@ const FormField = forwardRef(
           : {};
       contents = (
         <FormFieldContentBox
-          overflow="hidden"
-          {...contentProps}
+          {...fieldContentProps}
           {...innerProps}
+          {...contentProps}
```
halocline, line 260: In terms of order of operations, in this case whatever is specified in `contentProps` will trump all other styling. I am wondering about states where a border or background is specified for errors, disabled, focus, etc. states. This would override the styling for those states, correct? Do we want the styling for those states to be overridden? Alternative approach suggested below...

Also, I was struggling to make sense for what the differences are between `fieldContentProps` and `contentProps`. I think where I landed is that they are effectively the same thing, but originating from two locations (theme, prop).

Should they merge into a single object where we can better account for the manipulation of that object's values? This could help with readability as well as allowing for better control flow of object.

Line 183 could initiate a `adjustedContentProps` with definition from theme. Then apply any modifications specified in `contentProps`, then execute the rest of the object's manipulation based on the existing state based logic. The resulting hierarchy would be: theme.formField.content --> contentProps --> theme.formField.[state].

MikeKingdom: I did think about disabled and such, but I decided I this was the most flexibility. If I'm bothering to specify separate contentProps rather than define it in the theme, this is more of a special case situation and I don't want other things to affect it. If you let disabled and such override what I pass in contentProps then I don't really have a way to do what I want. I'm at the mercy of existing disabled modifiers and such. And I specifically kept contentProps as the passed in user-defined prop. Other props come from the theme for that particular field type. At least that's what was in my brain at the time.

halocline: Makes sense. I'd need to ponder this a bit more. If we do go the `contentProps` trump all direction, I do think it would be beneficial to explicitly say so / provide guidance for its use in the docs.

src/js/components/FormField/FormField.js
```diff
@@ -179,26 +180,30 @@ const FormField = forwardRef(
       );
     }
 
-    const contentProps = { ...formFieldTheme.content };
+    const fieldContentProps = { ...formFieldTheme.content };
```
halocline, line 183: Would naming this `themeContentProps` help with clarity?

MikeKingdom: Great idea

src/js/components/FormField/FormField.js
MikeKingdom: Added a little more detail in the doc for contentProps

src/js/components/FormField/doc.js
MikeKingdom: Added more detail in the description about what they are applied to and may override

src/js/components/FormField/doc.js
```diff
@@ -34,6 +34,9 @@ export const doc = FormField => {
       of name, value, onChange (event => {}), while event has either event.value
       or event.target.value.`,
     ),
+    contentProps: PropTypes.object.description(`Any valid Box property. These
+     properties are applied to the FormField contents container and can override
```
halocline, line 38:
```suggestion
     properties are applied to the FormField contents container and will override
```

src/js/components/FormField/FormField.js
halocline: > @MikeKingdom I believe these should be removed altogether. I missed removing these when resolving a merge conflict two weeks back. I can handle in a separate PR if you prefer.

@MikeKingdom are you okay with removing these commented lines? Or would you like me to in a separate PR?

src/js/components/FormField/FormField.js
halocline: I like this. Helps me out.

halocline:
@MikeKingdom I am thinking we could benefit from adding a Storybook example for visual regression testing purposes. Would you consider adding the following?

```
import React from 'react';

import {
  grommet,
  Box,
  Form,
  FormField,
  Grommet,
  Heading,
  Text,
  TextInput,
} from 'grommet';
import { deepMerge } from '../../../utils';

  formField: {
    border: {
      side: 'all',
    },
    content: {
      pad: 'large',
      margin: 'medium',
    },
    error: {
      size: 'xsmall',
    },
    help: {
      size: 'xsmall',
    },
    info: {
      size: 'xsmall',
    },
    label: {
      size: 'small',
    },
    round: '4px',
  },
  global: { font: { size: 'small' } },
});

const ContentProps = () => {
  return (
    <Grommet theme={customTheme}>
      <Box gap="medium">
        <Box border pad={{ horizontal: 'medium' }} width="medium">
          <Form>
            <Heading level={2} size="xsmall">
              Styling FormField content container with ContentProps
            </Heading>
            <Text>
              ContentProps will override any settings applied in your theme.
            </Text>
            <FormField
              htmlFor="example1-id"
              name="example1"
              label="Field Label"
              help="Some helpful descriptive text"
              error="Message to show on error"
              info="Additional contextual information"
              contentProps={{
                background: 'lightblue',
                border: { color: 'blue', size: 'small' },
                pad: 'medium',
                overflow: 'auto',
              }}
            >
              <TextInput
                id="example1-id"
                name="example1"
                placeholder="Placeholder input prompt"
              />
            </FormField>
          </Form>
        </Box>
        <Box border pad={{ horizontal: 'medium' }} width="medium">
          <Form>
            <Heading level={2} size="xsmall">
              ContentProps + Disabled State
            </Heading>
            <FormField
              htmlFor="example2-id"
              name="example2"
              label="Field Label"
              help="Some helpful descriptive text"
              contentProps={{
                background: 'lightblue',
                border: { color: 'blue', size: 'small' },
                pad: 'medium',
                overflow: 'auto',
              }}
            >
              <TextInput
                id="example1-id"
                name="example1"
                placeholder="Placeholder input prompt"
                disabled
              />
            </FormField>
          </Form>
        </Box>
      </Box>
    </Grommet>
  );
};

storiesOf('Form', module).add('FormField ContentProps', () => <ContentProps />);

```

halocline:
Everything else is looking good to me. I have come around and am comfortable with having `contentProps` trump other styling.

src/js/components/FormField/FormField.js
MikeKingdom: I can, especially since I'm making other tweaks based on feedback. Sorry I missed this in the most recent changes

MikeKingdom: removed

src/js/components/FormField/doc.js
MikeKingdom: Done

ShimiSun:
@MikeKingdom can you add the example you shared under the story name of Themed?

src/js/components/FormField/stories/ContentProps.js
```diff
@@ -0,0 +1,107 @@
...
+              />
+            </FormField>
+          </Form>
+        </Box>
+      </Box>
+    </Grommet>
+  );
+};
+
+storiesOf('Form', module).add('FormField ContentProps', () => <ContentProps />);
```
ShimiSun, line 107:
```suggestion
storiesOf('Form', module).add('Themed', () => <ContentProps />);
```

src/js/components/FormField/stories/ContentProps.js
```diff
@@ -0,0 +1,107 @@
+import React from 'react';
```
ShimiSun, line 1: Can you please rename the file to Themed.js, it will be more straight forward to scale it in the future in case we'd like to demonstrate more theme changes of Form.

src/js/components/FormField/stories/ContentProps.js
MikeKingdom: Done

src/js/components/FormField/stories/ContentProps.js
MikeKingdom: Done

ShimiSun:
Thank you for your contribution @MikeKingdom !

ptsteadman:
Thanks!

Issue #4244, opened 2020-06-29
FormField accessibility insights


jcfilben:
Action Items:
1. FormField documentation should show that if FormField doesn't have a child the `a11yTitle` prop should be on FormField.
2. CheckBox documentation needs to be updated to show that `label` is being used for `aria-label`.
3. TextInput documentation needs to be updated to show that it has the `a11yTitle` prop.

ShimiSun:
For the task force assignment, take care of bullet #2 and add `a11yTitle` to CheckBox. while falling back to label if it doesn't exist for backward compatibility.

ShimiSun:
@jcfilben I think we can close this issue now, thoughts?

jcfilben:
Yes, closing issue

PR #4326, opened 2020-07-20
Updated FormField docs to include a11yTitle

JoeButler7:
#### What does this PR do?
Updates `FormField` docs to include a11yTitle
`src/js/components/FormField/doc.js`
#4244 
Updated by this PR
No
Backwards compatible


src/js/components/FormField/doc.js
```diff
@@ -17,6 +17,11 @@ export const doc = FormField => {
     .intrinsicElement('div');
 
   DocumentedFormField.propTypes = {
+    a11yTitle: PropTypes.string.description(
+      `Custom label to be used by screen readers. When provided,
+        an aria-label will be added to the element. Should be provided
+        when FormField does not have a child with a11yTitle property.`,
+    ),
```
ShimiSun, line 24: This sentence needs refinement for FormField, It's not accurate to say `When provided,
        an aria-label will be added to the element`, It is only true to when there is no child. try to refine the sentence.

src/js/components/FormField/doc.js
```diff
@@ -17,6 +17,11 @@ export const doc = FormField => {
     .intrinsicElement('div');
 
   DocumentedFormField.propTypes = {
+    a11yTitle: PropTypes.string.description(
+      `Custom label to be used by screen readers. When provided and FormField
+        has no child component an aria-label will be added to the element.
+        Should not be provided if FormField has child.`,
```
ShimiSun, line 23: should it be 'children' instead of 'child'? The grammar doesn't sound right

PR #4327, opened 2020-07-21
RadioButton - add option to theme for background color on hover

### LINKED ISSUE #4188: Would like the ability to set background-color on RadioButton bounding Box via theme
### Expected Behavior

Able to specify RadioButton's background color via theme. (useful for hover state styling)

### Actual Behavior

Currently this can only be accomplished via `extend` in the theme.

Background: https://github.com/grommet/grommet-theme-hpe/pull/75/files/72e828b82f902dd7b1477e7951e56fb0debf6851#r437617242

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

britt6612:
#### What does this PR do?
This PR adds the option to RadioButton for a background color when hovered
RadioButton/StyledRadioButton.js
Storybook testing 
I added to the theme storybook for RadioButtonGroup
I will admit the colors probably are not the best so if you want 
me to take out I will
```
 hover:{
     background: {
        color: 'accent-4',
      },
 }
```

storybook
This goes back to the design-system
There was some hover styling needed with `RadioButtonGroup`
The way it was implemented was with css in the theme using the `extend`
It brought up the fact that `grommet` can provide this option to users
closes #4188 
auto
no
backwards compatible


britt6612:
One comment I was a little confused on is since `RadioButtonGroup` is built from `RadioButton` and all of the theme options that are in `RadioButton` can also be used for `RadioButtonGroup` but this is not in the docs for `RadioButtonGroup` ? wasn't sure if there was a reason behind that ?

src/js/components/RadioButton/doc.js
```diff
@@ -116,6 +116,11 @@ export const themeDoc = {
     type: 'string',
     defaultValue: '24px',
   },
+  'radioButton.hover.background.color': {
+    description: `The color of the RadioButton border when hovered over.`,
```
taysea, line 120: Think this should say `background` -- but I feel like we need to make this a little more clear that it's going to be the background of the entire div (including the label).

src/js/components/RadioButton/StyledRadioButton.js
```diff
@@ -15,14 +15,23 @@ const StyledRadioButtonContainer = styled.label`
...
-    !props.disabled &&
-    'cursor: pointer;'}
+  !props.disabled && 'cursor: pointer;'}
 
   :hover input:not([disabled]) + div,
   :hover input:not([disabled]) + span {
     border-color: ${props =>
       normalizeColor(props.theme.radioButton.hover.border.color, props.theme)};
   }
+  :not(div):hover {
```
taysea, line 25: Trying to understand this selector a little more. How come we're saying `:not(div)`?

src/js/components/RadioButton/doc.js
britt6612: I totally forgot to edit that last night just copied from the border Ill fix

src/js/components/RadioButton/README.md
```diff
@@ -217,6 +217,17 @@ Defaults to
 24px
 ```
 
+**radioButton.hover.background.color**
+
+The color of the Box surronding the RadioButton
```
taysea, line 222:
```suggestion
The background color of the Box surrounding the RadioButton
```

src/js/components/RadioButton/doc.js
```diff
@@ -116,6 +116,12 @@ export const themeDoc = {
     type: 'string',
     defaultValue: '24px',
   },
+  'radioButton.hover.background.color': {
+    description: `The color of the Box surronding the RadioButton
```
taysea, line 120:
```suggestion
    description: `The background color of the Box surrounding the RadioButton
```

taysea: Think you'll have to pull this down locally to update the snapshots properly.

britt6612: yes I believe so then Ill have to run yarn update

src/js/components/RadioButton/StyledRadioButton.js
```diff
@@ -15,14 +15,23 @@ const StyledRadioButtonContainer = styled.label`
...
-    !props.disabled &&
-    'cursor: pointer;'}
+  !props.disabled && 'cursor: pointer;'}
 
   :hover input:not([disabled]) + div,
   :hover input:not([disabled]) + span {
     border-color: ${props =>
       normalizeColor(props.theme.radioButton.hover.border.color, props.theme)};
   }
+  :hover {
```
taysea, line 25: I think this hover should also only happen when the input is not disabled, similar to the ones above. Not sure off the top of my head what the selector following it would be (div, span, maybe nothing) but could you try that locally?

britt6612: I wasnt sure about that because the behavior in the design-system has hover on disabled so Im not sure if that was what we wanted that was why I left it

taysea: That might be something we need to fix in the design system theme. I think most grommet components follow the behavior that if a component is disabled, it doesn't receive hover styling.

britt6612: That makes more sense I think your right with that just being an issue on the design-system..

src/js/components/RadioButton/StyledRadioButton.js
```diff
@@ -25,14 +26,13 @@ const StyledRadioButtonContainer = styled.label`
   :hover {
     background-color: ${props =>
       normalizeColor(
-        props.theme.radioButton.hover &&
+        !props.disabled &&
```
taysea, line 29: It seems like we don't even want to get into this block of code if the radio button is disabled -- any way we can utilize the `input:not([disabled])` being used in lines 21 and 22? Just thinking about how to keep it as clean as possible.

britt6612: This block of code with not run if the `disabled` prop is used within the `calendar`

src/js/themes/base.d.ts
```diff
@@ -81,10 +81,12 @@ type Colors = typeof colors & {
 
 interface ButtonKindType {
   background?: BackgroundType;
-  border?: {
-    color?: ColorType;
-    width?: string;
-  } | boolean;
+  border?:
+    | {
```
taysea, line 85: Is the pipe in the front a new convention? noticing it changed on all of these

britt6612: I think this was the prettier I have on my local I will change it back

src/js/components/RadioButton/doc.js
```diff
@@ -116,6 +116,12 @@ export const themeDoc = {
     type: 'string',
     defaultValue: '24px',
   },
+  'radioButton.hover.background.color': {
+    description: `The background color of the Box surrounding the RadioButton
+    when hovered over.`,
+    type: 'string | { dark: string, light: string }',
+    defaultValue: '{dark: undefined, light: undefined}',
```
ShimiSun, line 123: This actually defaults to an object (with undefined values for its keys), we should default to undefined.
```suggestion
    defaultValue: 'undefined',
```

src/js/components/RadioButton/StyledRadioButton.js
```diff
@@ -23,7 +23,16 @@ const StyledRadioButtonContainer = styled.label`
...
+  :hover {
+    background-color: ${props =>
+      normalizeColor(
+        !props.disabled &&
+          props.theme.radioButton.hover &&
+          props.theme.radioButton.hover.background &&
+          props.theme.radioButton.hover.background.color,
+        props.theme,
+      )};
+  }
```
ShimiSun, line 35: Why isn't this block of code part of line 23? why are we opening a new hover block instead of leveraging an existing one?

src/js/themes/base.js
```diff
@@ -889,6 +889,9 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       },
       // color: undefined,
       hover: {
+        // background: {
+        //   color: { dark: undefined, light: undefined },
```
ShimiSun, line 893:
```suggestion
        //   color: undefined,
```

src/js/components/RadioButton/StyledRadioButton.js
britt6612: Line 23 is adding a border color on hover to the child div that is contained inside.
we want the background color for hover to go on the parent div not the child. Therefore we need a new block of code.

britt6612: If you look at the 
https://deploy-preview-4327--sad-tereshkova-173d07.netlify.app/?path=/story/radiobuttongroup--custom-theme
the difference between child div getting the border color on hover and the background is. easier to see

ShimiSun: Thank you for the clarification.

PR #4336, opened 2020-07-24
add calendar test DO NOT MERGE 

britt6612:
Testing with adding new snapshot

### DIFF
--- a/src/js/components/Button/doc.js
+++ b/src/js/components/Button/doc.js
@@ -46,7 +46,7 @@ export const doc = Button => {
     color: colorPropType.description(
       'Fill color for primary, label color for plain, border color otherwise.',
     ),
-    disabled: PropTypes.bool
+    disabled: PropTypes.oneOfType([PropTypes.bool, PropTypes.number])
       .description('Whether the button is disabled.')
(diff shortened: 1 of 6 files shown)


src/js/components/Button/doc.js
```diff
@@ -46,7 +46,7 @@ export const doc = Button => {
     color: colorPropType.description(
       'Fill color for primary, label color for plain, border color otherwise.',
     ),
-    disabled: PropTypes.bool
+    disabled: PropTypes.oneOfType([PropTypes.bool, PropTypes.number])
```
ShimiSun, line 49: I don't think we want to add the 'number' type to button as discussed on previous threads.

britt6612: I know Im only doing this for testing purposes

britt6612: It wont allow me to add disabled test without it (if only we had it before we would of saw this error)

britt6612:
Closing this PR:

I was testing why snapshots were failing when passing in `dates` to `disabled` 
In order for `disabled` prop to pass it needed to be passed the `date` with timestamp

Issue #4076, opened 2020-05-13
Box children don't get gap if wrapped in a fragment


arthurRbr:
If you have a component that returns an array mapped into a list:

```
// ListComponent
return (
   <>
        {[1, 2, 3, 4].map(n => (
          <Box>child</Box>
        ))}
   </>
)
```

A component that uses that list cannot set a gap between elements in the list.
```
<Box gap="large">
    <ListComponent />
<Box>
```

MikeKingdom:
This issue would be difficult to resolve given how Box currently puts a div spacer between children when it renders. The problem is when Box renders, ListComponent is a single child and Box can't tell it renders a Fragment until after ListComponent renders which is way too late. If Box contained a direct Fragment child (rather than the Fragment being rendered by a child component) it could tell that and react accordingly, but that is much less useful than the example in this issue.

ShimiSun:
@MikeKingdom, since I don't think we'll attempt to fix it or change how it behaves at this point, can you add a comment to the documentation of `gap` on fragment children?

MikeKingdom:
I changed the Box gap documentation in #4338 to specifically mention it won't put gaps between children in a Fragment.

PR #4338, opened 2020-07-24
Fix/box gap doc

MikeKingdom:
#### What does this PR do?
Updates the Box documentation for the gap property to note that gap won't be added between any elements that are contained in a fragment.

doc.js

jest tests updated

#4076 

Done

No

It's backwards compatible

### DIFF
--- a/src/js/components/Box/doc.js
+++ b/src/js/components/Box/doc.js
@@ -212,9 +212,8 @@ export const doc = Box => {
       PropTypes.string,
     ]).description(`The amount of spacing between child elements. This
         should not be used in conjunction with 'wrap' as the gap elements
-        will not wrap gracefully. If a child component renders a fragment,
-        Box will not add a gap between the individual elements in the
-        fragment.`),
+        will not wrap gracefully. If a child is a Fragment,
+        Box will not add a gap between the choldren of the Fragment.`),
     height: PropTypes.oneOfType([
       PropTypes.oneOf([
         'xxsmall',
--- a/src/js/components/Box/doc.js
+++ b/src/js/components/Box/doc.js
@@ -212,7 +212,9 @@ export const doc = Box => {
       PropTypes.string,
     ]).description(`The amount of spacing between child elements. This
         should not be used in conjunction with 'wrap' as the gap elements
-        will not wrap gracefully.`),
+        will not wrap gracefully. If a child component is or renders a fragment,
+        Box will not add a gap between the individual elements in the
+        fragment`),
     height: PropTypes.oneOfType([
       PropTypes.oneOf([
         'xxsmall',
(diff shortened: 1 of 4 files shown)


src/js/components/Box/doc.js
```diff
@@ -212,7 +212,9 @@ export const doc = Box => {
       PropTypes.string,
     ]).description(`The amount of spacing between child elements. This
         should not be used in conjunction with 'wrap' as the gap elements
-        will not wrap gracefully.`),
+        will not wrap gracefully. If a child component is or renders a fragment,
```
ShimiSun, line 215:
```suggestion
        will not wrap gracefully. If a child component renders a fragment,
```

src/js/components/Box/doc.js
```diff
@@ -212,7 +212,9 @@ export const doc = Box => {
...
+        Box will not add a gap between the individual elements in the
+        fragment`),
```
ShimiSun, line 217:
```suggestion
        fragment.`),
```

MikeKingdom: I can word it that way if you prefer. I did specifically choose that phrasing because there is a difference between a functional component that returns a fragment and a literal fragment as a child. The latter we could actually detect and iterate over it's children from what I can tell

src/js/components/Box/doc.js
MikeKingdom: updated

src/js/components/Box/doc.js
MikeKingdom: updated

src/js/components/Box/doc.js
```diff
@@ -212,7 +212,9 @@ export const doc = Box => {
       PropTypes.string,
     ]).description(`The amount of spacing between child elements. This
         should not be used in conjunction with 'wrap' as the gap elements
-        will not wrap gracefully.`),
+        will not wrap gracefully. If a child component renders a fragment,
+        Box will not add a gap between the individual elements in the
+        fragment.`),
```
ShimiSun, line 217:
```suggestion
        will not wrap gracefully. If a child is a Fragment,
        Box will not add a gap between the children of the Fragment.`),
```

MikeKingdom: This might get the point across but it's not the situation in issue that was submitted. In this issue the child is not a fragment. It's a functional component that renders a fragment. It actually produces a different component tree, at least from what Box can figure out when it's rendered.

ShimiSun: Thanks for the quick call, I think we can incorporate the feedback and wrap up this issue.

MikeKingdom: Simplified the wording

PR #4341, opened 2020-07-25
Changed Chart to add per-value color and thickness for bar and point types

ericsoderberghp:
#### What does this PR do?

Changed Chart to add per-value color and thickness for bar and point types.

#### Where should the reviewer start?

Take a look at the new story to see how they are used.

#### What testing has been done on this PR?

Added unit tests and a storybook story.

#### How should this be manually tested?

storybook

#### Any background context you want to provide?

This was asked for by a group doing data science work as a way to represent four values in one Chart: x, y, thickness, and color.

#### Do the grommet docs need to be updated?

automatic

#### Should this PR be mentioned in the release notes?

yes

#### Is this change backwards compatible or is it a breaking change?

backwards compatible

PR #4342, opened 2020-07-27
box props for radiobuttongroup added in theme

### LINKED ISSUE #4184: Would like ability for RadioButtonGroup styling to be set in theme
### Expected Behavior

I'd like to be able to specify styling for RadioButtonGroup in a theme, specifically the options `gap`.

### Actual Behavior

Theme not currently referenced.
RadioButtonGroup styling can not be set in theme (e.g. desired gap size)
  - Same applies to CheckBoxGroup

Use case. HPE Design System designs desire a gap of none between RadioButtonGroup and CheckBoxGroup options. Implementers can achieve this by setting the `gap` property for each instance, however for the Design System, this specification would make sense to live in the theme.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

IanKBovard:
#### What does this PR do?
closes #4184 
Adds a gap to RadioButtonGroup in theme

RadioButtonGroup.js
base.js
base.d.ts
RadioButtonGroup Stories - CustomTheme.js
RadioButtonGroup doc.js

Storybook, tested with and without custom theme.
Used none, xxsmall, small, medium, large, xlarge, and px values to test gap within custom theme.

same as above

RadioButtonGroup did not have the ability to set gap within theme.

#4341 

Yes, they are in this pr

backwards compatible


src/js/components/RadioButtonGroup/RadioButtonGroup.js
```diff
@@ -27,6 +28,10 @@ const RadioButtonGroup = forwardRef(
     ref,
   ) => {
     const formContext = useContext(FormContext);
+    const theme = useContext(ThemeContext) || defaultProps.theme;
+    const gap = theme.global.radioButtonGroup
+      ? theme.global.radioButtonGroup.gap
+      : 'small';
```
ShimiSun, line 34: Would setting a default value here will be backward compatible?

src/js/components/RadioButtonGroup/doc.js
```diff
@@ -61,3 +61,22 @@ export const doc = RadioButtonGroup => {
 
   return DocumentedRadioButtonGroup;
 };
+
+export const themeDoc = {
+  'radioButtonGroup.gap': {
+    description: `The amount of spacing between child elements. This
+    should not be used in conjunction with 'wrap' as the gap elements
+    will not wrap gracefully.`,
```
ShimiSun, line 69: Not sure this level of description is relevant for a theme prop, thoughts?

src/js/themes/base.d.ts
```diff
@@ -874,7 +887,10 @@ export interface ThemeType {
     size?: string;
     font?: {
       weight?: number | string;
-    }
+    };
+  };
+  radioButtonGroup?: {
+    gap?: string;
```
ShimiSun, line 893: I think it should be GapType instead of string. import it from utils.

src/js/components/RadioButtonGroup/RadioButtonGroup.js
IanKBovard: in the original code radiobuttongroup had a default value of small.
```
const RadioButtonGroup = forwardRef(
  (
    {
      children,
      disabled,
      gap = 'small',
      name,
      onChange,
      options: optionsProp,
      value: valueProp,
      ...rest
    },
    ref,
  ) => {
```

src/js/components/RadioButtonGroup/doc.js
IanKBovard: its just copied from box, i can remove that section

src/js/themes/base.d.ts
IanKBovard: done

src/js/components/RadioButtonGroup/doc.js
IanKBovard: removed that section

src/js/components/RadioButtonGroup/RadioButtonGroup.js
ShimiSun: We should keep the behavior of the original code and add the theme option for the `gap` without setting a default value on it. That will be more backward compatible in case a user is overriding the theme.

britt6612:
This code looks fine 
However I think the original issue was for the vertical gap between the icon on radio button and the label.
this can be done with gap=“none”, but would require developers to always set that prop. So the intent was for the gap to be specified in the theme. This was so in the design-system we can control this in the theme

```
<RadioButtonGroup
          id="power-stzw"
          name="power-regulation"
          options={powerRegulationOptions}
          value={powerRegulation}
          onChange={event => setPowerRegulation(event.target.value)}
          gap="none"
        />
```

src/js/components/RadioButtonGroup/RadioButtonGroup.js
IanKBovard: changes made

src/js/themes/base.d.ts
```diff
@@ -81,10 +81,12 @@ type Colors = typeof colors & {
 
 interface ButtonKindType {
   background?: BackgroundType;
-  border?: {
-    color?: ColorType;
-    width?: string;
-  } | boolean;
+  border?:
```
britt6612, line 84: Maybe change linter back to how it is

britt6612: This add extra pipe before

IanKBovard: when i commit it auto lints it for me. is there a config im missing?

britt6612: hmm maybe that got changed ? @ShimiSun is this the right linting ?

ShimiSun: We recently added linter & prettier for typescript files. So I think it might be a side effect, we'll need to take a look whether that is the default behavior that is a best practice.

IanKBovard: do i need a new config file, or to adjust the current config to account for this?

ShimiSun: It will probably be a separate task from this PR, leave it be for now if its complaint with lint.

src/js/components/RadioButtonGroup/RadioButtonGroup.js
```diff
@@ -27,6 +29,10 @@ const RadioButtonGroup = forwardRef(
     ref,
   ) => {
     const formContext = useContext(FormContext);
+    const theme = useContext(ThemeContext) || defaultProps.theme;
+    const radioButtonGap = theme.global.radioButtonGroup
+      ? theme.global.radioButtonGroup.gap
+      : gap;
```
ShimiSun, line 35: I think that instead of fetching only the `gap` style from the theme, the behavior here should be more generic and should fetch all the Box props from the theme.
So essentially this const will be removed, and see my changes on line 110.

src/js/components/RadioButtonGroup/RadioButtonGroup.js
```diff
@@ -101,7 +107,7 @@ const RadioButtonGroup = forwardRef(
         onLeft={focus ? onPrevious : undefined}
         onRight={focus ? onNext : undefined}
       >
-        <Box ref={ref} gap={gap} {...rest}>
+        <Box ref={ref} gap={radioButtonGap} {...rest}>
```
ShimiSun, line 110:
```suggestion
        <Box ref={ref} {...theme.radioButtonGroup} {...rest}>
```
This change will enable all Box props and not just `gap`.

IanKBovard: changes made

src/js/components/RadioButtonGroup/doc.js
```diff
@@ -61,3 +61,20 @@ export const doc = RadioButtonGroup => {
...
+    defaultValue: `{
+      none,
+      xxsmall,
+      xsmall,
+      small,
+      medium,
+      large,
+      xlarge,
+      any css size,
+    }`,
```
ShimiSun, line 78: I don't think the default value is an object, it should be `undefined`.

ShimiSun: And regardless, the docs need to accommodate your recent changes where all the Box props are included and not a single doc entry just for gap prop.

src/js/themes/base.d.ts
```diff
@@ -874,7 +887,10 @@ export interface ThemeType {
     size?: string;
     font?: {
       weight?: number | string;
-    }
+    };
+  };
+  radioButtonGroup?: {
+    gap?: GapType;
```
ShimiSun, line 893: This should accommodate your recent changes.

src/js/themes/base.js
```diff
@@ -909,6 +909,9 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         // weight: undefined,
       },
     },
+    radioButtonGroup: {
+      // gap: undefined,
```
ShimiSun, line 913: should include your recent changes.

src/js/themes/base.d.ts
IanKBovard: gotcha, just to clarify, something like this:

```
radioButtonGroup?: {
  a11yTitle: string;
  align: string;
  alignContent: string;
  ...
  ...
}
```

src/js/components/RadioButtonGroup/doc.js
IanKBovard: something like this:
```
export const themeDoc = {
  'radioButtonGroup.a11yTitle': {
    description: `a user friendly label for screen readers`,
    type: 'string',
    defaultValue: `undefined`,
  },
'radioButtonGroup.align': {
    description: `How to align the contents along the cross axis.`,
    type: 'string',
    defaultValue: `undefined`,
  }
  ...
  ...
};
```

am i correct to assume that, if there is no bolded value in the docs it defaults to undefined?

ShimiSun: Similar to https://v2.grommet.io/select#select.options.container

src/js/themes/base.d.ts
ShimiSun: Similar to: 
```
  card?: {
    container?: BoxProps;
  },
```

IanKBovard:
changed made

src/js/components/RadioButtonGroup/doc.js
```diff
@@ -61,3 +61,11 @@ export const doc = RadioButtonGroup => {
 
   return DocumentedRadioButtonGroup;
 };
+
+export const themeDoc = {
+  'radioButtonGroup.container': {
+    description: 'Any valid Box props for the RadioButtonGroup container.',
+    type: 'object',
+    defaultValue: "{ align: 'start', pad: 'small' , ...}",
```
ShimiSun, line 69: Are you sure this is the default value? I'm not sure it is.

src/js/themes/base.js
```diff
@@ -918,6 +918,11 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         // weight: undefined,
       },
     },
+    radioButtonGroup: {
+      container: {
+        // gap: undefined
```
ShimiSun, line 923: Should be commented out as well.

src/js/utils/index.d.ts
```diff
@@ -50,7 +50,11 @@ export { isObject, deepFreeze, deepMerge, removeUndefined };
  * type SomeComponentProps = PropsOf<typeof SomeComponent>;
  * ```
  */
-export type PropsOf<TComponent> = TComponent extends React.ComponentType<infer P> ? P : never;
+export type PropsOf<TComponent> = TComponent extends React.ComponentType<
+  infer P
+>
+  ? P
+  : never;
```
ShimiSun, line 57: Please revert this change unless you changed the code's meaning.

src/js/components/RadioButtonGroup/doc.js
IanKBovard: used the same description for select 
https://v2.grommet.io/select#select.options.container

```
  'select.options.container': {
    description: 'Any valid Box prop for the options container.',
    type: 'object',
    defaultValue: "{ align: 'start', pad: 'small' }",
  },
```

my assumption is that the description on the select page is not default values either, just showing values that you can pass into select.options.container.

src/js/utils/index.d.ts
IanKBovard: I believe this is because of the linter.

src/js/themes/base.js
IanKBovard: done

src/js/components/RadioButtonGroup/doc.js
ShimiSun: Please put the correct default values for the RBG, and take a look at how the Select entry could be fixed as well.

src/js/utils/index.d.ts
ShimiSun: Please revert it, my linter doesn't apply it, might be something with your IDEA.

src/js/themes/base.js
```diff
@@ -918,6 +918,11 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         // weight: undefined,
       },
     },
+    radioButtonGroup: {
+      // container: {
+      // gap: undefined
+      // },
```
ShimiSun, line 924:
```suggestion
      // container: {}, // any box props
```

src/js/components/RadioButtonGroup/doc.js
IanKBovard: changed default to an empty object

src/js/themes/base.js
IanKBovard: made changes

src/js/components/RadioButtonGroup/doc.js
```diff
@@ -61,3 +61,11 @@ export const doc = RadioButtonGroup => {
 
   return DocumentedRadioButtonGroup;
 };
+
+export const themeDoc = {
+  'radioButtonGroup.container': {
+    description: 'Any valid Box props for the RadioButtonGroup container.',
+    type: 'object',
+    defaultValue: '{}',
```
ShimiSun, line 69:
```suggestion
    defaultValue: 'undefined',
```

src/js/utils/index.d.ts
```diff
@@ -1,249 +0,0 @@
-// colors.js
```
ShimiSun, line 1: Why did you delete this file?

IanKBovard:
created a new pr that fixes the linting issues on this one https://github.com/grommet/grommet/pull/4383

Issue #4340, opened 2020-07-25
Cannot assign to read only property 'target' of object '#<MouseEvent>'


AlexBHarley:
Using the `Select` component with Preact fails with `Cannot assign to read only property 'target' of object '#<MouseEvent>'
 `.

I'd very much like to use Preact in my project but this is blocking me.

### Expected Behavior

`Select` component should work as it does with React.

### Actual Behavior

Crashes with `Cannot assign to read only property 'target' of object '#<MouseEvent>'`.

### URL, screen shot, or Codepen exhibiting the issue

![Screen Shot 2020-07-25 at 16 18 31](https://user-images.githubusercontent.com/11168524/88459055-8af3c500-ce92-11ea-9893-f5fb30cb9ab7.png)

### Steps to Reproduce

1. Scaffold a create-react-app project and add grommet
2. Create a basic select component
3. Add a `config-overrides.js` file 

```
module.exports = function override(config, env) {
  config.resolve.alias['react'] = 'preact/compat';
  config.resolve.alias['react-dom'] = 'preact/compat';
  config.resolve.alias['create-react-class'] =
    'preact/compat/lib/create-react-class';
  config.resolve.alias['react-dom-factories'] =
    'preact/compat/lib/react-dom-factories';

  return config;
};
```
4. Use `react-app-rewired start` to start app

### Your Environment

- Grommet version: stable
- Browser Name and version: Chrome 84.0.4147.89
- Operating System and version (desktop or mobile): MacOS 10.13.2 (Desktop)

ShimiSun:
Thanks for reporting! we'll be happy to receive a contribution from the community for this issue.

PR #4346, opened 2020-07-28
#18027 Fix target is not writable on the native event

teimurjan:
#### What does this PR do?

Fixing select event behavior in Preact by changing simple assignment to `Object.defineProperty`.

1. Understand the bug's context from [the issue](https://github.com/grommet/grommet/issues/4340).
2. Check [the sandbox preview](https://codesandbox.io/s/grommet-preact-select-fix-4zmlx).
3. Go through the code.

- Open a select.
- Click on a value.
- Be sure that the value is set.

Go to the sandbox https://codesandbox.io/s/grommet-preact-select-fix-4zmlx and check that the select value is changing without errors.
This sandbox patching the grommet package with the change from this PR.

Inspired by a similar solution for the `material-ui` package: https://github.com/mui-org/material-ui/pull/18027/files.

https://github.com/grommet/grommet/issues/4340

---

No.

No.

It's backward compatible.


AlexBHarley:
I raised #4340 so I'll give this a test later today. Quick turn around!

ShimiSun:
Thank you @AlexBHarley, keep us posted!

AlexBHarley:
I guess I was overconfident with this one. I've built and linked this branch locally, but my app never builds. Any ideas?

ShimiSun:
Have you tried this https://github.com/grommet/grommet/wiki/How-to-Apply-Your-Grommet-Development-Branch-to-a-Local-Project?

AlexBHarley:
OK yeah I needed to `yarn link` from the dist folder, thanks. 

Unfortunately there's a new problem so I'm unable to test because my whole Preact branch fails.

I have a StyledComponent that looks like this:

```
export default styled(NavTab)`
  background: ${props =>
    console.log(props) || props.theme.global.colors['background']};
  display: inline-block;
  ...
`;
```

`props.theme` is an empty object and as such I get `Unhandled Rejection (TypeError): Cannot read property 'colors' of undefined`. 

I inject theme as normal in the Grommet component

```
<Grommet
  full
  theme={themes[theme]}
  ...
/>
```

To be clear this also happens on `https://github.com/grommet/grommet/tarball/stable` so it's probably unrelated to this PR.

My branch using React doesn't have these issues.

teimurjan:
@AlexBHarley you can test it out on the CodeSandbox link I've posted in the description. It's using Preact and there is a folder `patches` which applies the changes from this PR.

teimurjan:
@AlexBHarley did you have time to test?

AlexBHarley:
👍 from me! Will have to sort out the other issue another time

halocline:
@ShimiSun I have gone through the existing Storybook tests and compared the resulting adjustedEvent objects. I did not spot any regressions.

I also searched the rest of Grommet to see if we'd be running into an unwritable event target anywhere else. This surfaced in TextInput in two places:
- https://github.com/grommet/grommet/blob/5d97577f2bc5ac7e5af6ab5fc4196fa9ff1d2256/src/js/components/TextInput/TextInput.js#L315
- https://github.com/grommet/grommet/blob/5d97577f2bc5ac7e5af6ab5fc4196fa9ff1d2256/src/js/components/TextInput/TextInput.js#L355

If we like this approach, it may make sense to apply to TextInput instances as well.

Let me know if you need anything else from me on this issue.

ShimiSun:
Thanks @halocline, this is helpful.
I think our concern is that we are not sure what `Object.defineProperty` is doing, and we'll need to research its potential side effects, can you articulate how does it work and how it may impact the users of React & Preact?

halocline:
> Thanks @halocline, this is helpful.
> I think our concern is that we are not sure what `Object.defineProperty` is doing, and we'll need to research its potential side effects, can you articulate how does it work and how it may impact the users of React & Preact?

In this case, `[Object.defineProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty)` is modifying the existing `target` property on the native [`Event`](https://developer.mozilla.org/en-US/docs/Web/API/Event). Specifically, it is changing the `writable` property of `target` from `false` to `true`, effectively changing it from a ReadOnly object to a writable object. The `value` is assigned `inputRef.current` same as using `=`.

Preact does not have the concept of Synthetic Events like React, instead it is relying on native JS events.

This seems valid from a pure JS standpoint, but it'd be worthwhile to see how it plays with SyntheticEvents. That said, I think it is worthwhile to have a discussion if we should be supporting Preact at this time, considering adoption, market share, and available resources.

I am happy to do more research. How would you feel about a time boxed research spike?

ShimiSun:
@halocline not sure we can afford spending resources on the research at this point as we haven't seen high demand in Preact users, and supporting this change will mean supporting more areas that need Preact attention. Let me think about it a little more and we can circle back to it if needed.

ShimiSun:
Closing this PR in favor of #4461

PR #4359, opened 2020-07-29
Fix MaskedInput 'fixed' segments which are longer than 1 character

### LINKED ISSUE #4122: Backspacing in maskedInput has odd side-effects
I am trying to use a maskedInput to force a URL to begin with "https://".

It seems to work well and allows for both pasting in a full URL as well as typing in the part after the fixed portion.  But, if you backspace all the way to the beginning, it is adding unexpected text into the input.

I should be able to backspace without having unwanted text inserted.

Try it here:
https://designer.grommet.io/screen?id=MaskedInput-edgar-circenis-hpe-com

Use cases (you will have to refresh browser between use cases to get back to starting conditions):
1. Paste in "https://foo.com" (works)
2. Paste in "http://foo.com" (fails; the input ends up containing "https://http").  I would expect this paste should not be allowed.
3. Type in foo.com (works), then backspace all the way back (fails; Eventually, input contains "https://http")

MikeKingdom:
Anytime a mask segment is a 'fixed' string of more than 1
character, it always adds the full fixed segment in an input value.
However when the user attempts to backspace over that fixed
value, once they back over the last character in that segment,
MaskedInput puts back in the entire fixed string twice (minus the last
character of the second copy.)

The change allows you to have just part of the fixed segment there
until you add something that doesn't match the the first part of the
fixed string at which point it fills in the rest of the fixed part for
you. This works much more naturally with general typing and backspace.
It also still works the same as it used to for the more typical
single character fixed portions (like the phone number or IP Address
examples).

I added a storybook example of a multi-character fixed string (URL)
since it seemed like a useful example either way.

Fixes #4122

#### What does this PR do?
Allows you to backspace over a multi-character fixed region in the mask
Storybook example, added a JEST test case for how the fixed portion behaves

Look at the MaskedInput URL example in Storybook
Try typing the for URL and backspacing over it

#4122 

No

No

This is backwards compatible


ericsoderberghp:
Looks like a MaskedInput snapshot merge issue.

PR #4363, opened 2020-07-30
Add option for icon on error message in formfield

### LINKED ISSUE #4362: Add object to prop types for messages in Form
Currently messages prop in `form` only takes a string
https://v2.grommet.io/form#messages
If a user wanted to have a custom text they would pass in an 
object but then get a prop type error
### Expected Behavior
no prop type error

### Actual Behavior
Screen Shot 2020-07-29 at 5 05 16 PM

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:latest
- Browser Name and version:

britt6612:
#### What does this PR do?
This PR add the option for users to add a icon next to a required message in the `formfield` on error

### Where should the reviewer start?
FormField/FormField.js 
testing in. storybook with the following
```
const customFormFieldTheme = {
  formField: {
    error: {
      iconContainer: {
        margin: 'hair',
        pad: 'xxsmall',
      },
      size: 'xsmall',
      icon: <Add size="small" />,
    },
  },
};



```
storybook
The design-system calls for a icon to be used within the required message 
so when implementing this came across the PropType error
closes #4362
already done
no
backwards compatible


britt6612:
add @taysea for review

ericsoderberghp:
The `messages` property should be strictly JSON and accept only strings. This is to keep it aligned with localization practices where each language will have their own translation. We don't want to embed styling or layout aspects inside the messages. Instead, I think we should explore another way to add this capability into the theme to add something like `theme.formField.error.icon`.

britt6612:
> The `messages` property should be strictly JSON and accept only strings. This is to keep it aligned with localization practices where each language will have their own translation. We don't want to embed styling or layout aspects inside the messages. Instead, I think we should explore another way to add this capability into the theme to add something like `theme.formField.error.icon`.

@ericsoderberghp maybe I misunderstood how the localization practices work. I had pulled down the i18n project that was up on grommet to play around with it and saw that you still had to provide your own translations within the public folder. So with the `form` example I saw in the translation it was something like 

```
      <Form messages={{ required: t('form.errorMsg') }} validate="blur">
```
so after wrapping that in a box with text and not touching the `form.errorMsg` the translation was still working for me 
for instance I had something like: 

```
        messages={{
          required: (
            <Box direction="row">
              <strong>!</strong>
              {t('form.errorMsg')}
            </Box>
          ),
        }}

```

where the `form.errorMsg` is still being translated

britt6612:
I can explore adding a prop but I want to also understand the whole translation

ericsoderberghp:
Don't we want this to be themeable though, so that all HPE form users don't have to set a property to get this behavior?

britt6612:
> Don't we want this to be themeable though, so that all HPE form users don't have to set a property to get this behavior?

you are correct! I didnt look at it that way lol 
In the design system for now I have it as a standalone component that contains the text with icon but yes preferably we would want this in theme.

ShimiSun:
Where are we with this PR?

britt6612:
By adding the option to have an icon in the theme with the error message 
I found that we needed to wrap that in a `<Box/>` 
To provide more options for the user they can set their own margin and pad to align the icon how they prefer next to the text

ericsoderberghp:
Can we update the description at the top of this pull request to align more with the actual changes?

src/js/components/FormField/doc.js
```diff
@@ -146,6 +146,21 @@ export const themeDoc = {
     type: "string | {'dark': string, 'light': string}",
     defaultValue: undefined,
   },
+  'formField.error.icon': {
+    description: 'An icon that can be placed next to the error message.',
+    type: 'React.Element',
+    defaultValue: undefined,
+  },
+  'formField.error.iconContainer.margin': {
```
ericsoderberghp, line 154: I'm wondering if we should do something like: `formField.error.container` instead, which would take all Box props. This would allow adding `gap`, `margin`, `pad`, and even allow `direction: "row-reverse"` in case the icon should be placed at the end.

britt6612: yes this is good plus if the user wanted to center the icon they could also use `align`

src/js/themes/base.d.ts
```diff
@@ -663,6 +663,11 @@ export interface ThemeType {
       background?: BackgroundType;
       color?: ColorType;
       margin?: MarginType;
+      icon?: any;
+      iconContainer: {
```
taysea, line 667: Should this be updated to have any box props?

taysea: Or we will just highlight these main ones, but all are accepted

britt6612: oh yes let me check how other components are addressing this when you accept all props

src/js/themes/base.js
```diff
@@ -745,6 +745,8 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       error: {
         color: 'status-critical',
         margin: { vertical: 'xsmall', horizontal: 'small' },
+        // icon: undefined
+        // iconContainer: {}, // any box props
```
ericsoderberghp, line 749: Can we have the error `container` mean the Box with `direction="row"` that contains the error message and any icon, rather than a Box just containing the icon? Spacing between the icon and message would be controlled with `gap`, and you could still use `align`.

src/js/components/FormField/FormField.js
```diff
@@ -45,8 +45,23 @@ const FormFieldContentBox = styled(Box)`
...
+  const theme = useContext(ThemeContext) || defaultProps.theme;
+  const icon = theme.formField.error && theme.formField.error.icon;
+  const iconContainer =
+    theme.formField.error && theme.formField.error.iconContainer;
   if (message) {
+    if (error) {
+      if (icon && typeof message === 'string')
+        return (
+          <Box direction="row" align="start" gap="xsmall">
+            <Box flex={false} {...iconContainer}>
```
ericsoderberghp, line 58: I don't think we should be including this Box if there isn't and `icon`.

ShimiSun:
@taysea can you please take over the work that is left for this PR? If you can't push to this branch, feel free to create a new PR.

src/js/themes/base.js
taysea: The reason we have put a container directly around the icon was to solve for a case where some padding/margin needed to be applied to an icon to align properly with text (depending on the text line height) -- this came up in [this design system example](https://github.com/hpe-design/design-system/blob/master/aries-site/src/examples/templates/forms/RequiredFieldsExample.js#L47). You can see the visual [at the bottom of this form](https://design-system.hpe.design/templates/forms#required-and-optional-fields). So the intent is to be able to style just an icon container and not necessary the message and icon

src/js/components/FormField/FormField.js
taysea: Good point, will make an adjustment

src/js/themes/base.js
taysea: Having the container around the both of them may have benefit as well -- but I wouldn't say they're solving the same thing

ShimiSun:
The full fix for this PR is delivered on #4473. closing this PR

Issue #4278, opened 2020-07-08
List item onClick displaying UserAgent outline


taysea:
### Expected Behavior

The outline should not appear onClick on list item.

### Actual Behavior

The default UserAgent outline is outlining the List item when onClick is given.

### URL, screen shot, or Codepen exhibiting the issue

![image](https://user-images.githubusercontent.com/12522275/86940974-3ca39f80-c0f8-11ea-9472-f76ab921fdc4.png)

Also visible when clicking on a list item in this storybook example: 
https://storybook.grommet.io/?path=/story/list--onclickitem

### Steps to Reproduce

1. Create a list with onClickItem function

2. Click on an item

### Your Environment

- Grommet version: Stable
- Browser Name and version: Chrome Version 83.0.4103.116 (Official Build) (64-bit)
- Operating System and version (desktop or mobile): MacOS 10.14.6

taysea:
From preliminary research, what we are seeing is the updated Chrome focus outline. Resources here on this updated focus outline:
- https://blog.chromium.org/2020/03/updates-to-form-controls-and-focus.html
- https://github.com/octobercms/october/issues/4892

To align with how other Grommet clickable elements work, we would want to override this with grommet theme focus indicator.

taysea:
As I tried to address the above issue, I came into some issues regarding focus hand off between the `ul` and the click of an individual item. It seems that the `ul` should hold focus the entire time (even when an individual item is clicked) but I can't avoid having the focus indicator go onto the list item when clicked. I think this is based on how our `Box` handles onClick.

PR #4364, opened 2020-07-30
Improve List focus behavior

taysea:
#### What does this PR do?

Aligns List focus and keyboard behavior with WCAG guidelines that state the `ul` should remain in focus always. When a list item is clicked, the focus still is placed on the overall `ul`. 

Resource: https://www.w3.org/TR/2017/WD-wai-aria-practices-1.1-20170628/examples/listbox/listbox.html

Engaged with List on Chrome, Safari, and FF using mouse click and keyboard to navigate and confirmed that the `ul` maintains focus throughout these interactions but that the click on an individual list item is still registered. Use `onClickItem` storybook for testing.

Engage with List by using mouse click and keyboard.

Closes https://github.com/grommet/grommet/issues/4278
Closes https://github.com/hpe-design/design-system/issues/965
No.

Maybe.

Yes, backwards compatible.


src/js/components/List/List.js
```diff
@@ -88,9 +82,16 @@ const List = React.forwardRef(
         }
       >
         <StyledList
-          ref={ref}
+          ref={ref || listRef}
```
ericsoderberghp, line 85: We should use `useForwardedRef` instead.

src/js/components/List/List.js
```diff
@@ -56,6 +48,8 @@ const List = React.forwardRef(
   ) => {
     const theme = useContext(ThemeContext);
     const [active, setActive] = useState();
+    const [focus, setFocus] = useState();
```
ericsoderberghp, line 51: We're trying to move away from this pattern of tracking focus in javascript and instead relying on having `&:focus` in the CSS.

src/js/components/List/List.js
```diff
@@ -192,7 +193,14 @@ const List = React.forwardRef(
...
                     adjustedEvent.index = index;
                     onClickItem(adjustedEvent);
+                    // put focus back on the List container to meet
+                    // WCAG accessibility guidelines that focus remains
+                    // on `ul`
+                    (ref || listRef).current.focus();
                   },
+                  // prevent default focus outline from being shown
+                  // onMouseDown since `ul` is maintaining focus
+                  onMouseDown: event => event.preventDefault(),
```
ericsoderberghp, line 203: Do we need this? It feels like we might be preventing behavior that might be wanted somewhere.

src/js/components/List/List.js
```diff
@@ -192,7 +193,14 @@ const List = React.forwardRef(
...
                   onMouseOver: () => setActive(index),
                   onMouseOut: () => setActive(undefined),
                   onFocus: () => setActive(index),
```
ericsoderberghp, line 206: I wonder if we should do something here, so that focus can't go on the item at all.

src/js/components/List/List.js
taysea: Thanks for the reminder, I've updated it back to use CSS.

src/js/components/List/List.js
taysea: If this is not here, the `li` tries to grab focus, I wasn't able to find another way (to your below comment) to intercept this element from receiving focus. However, this does not have an effect on the `onClick`.

focus

src/js/components/List/List.js
taysea: Agreed, that was what I was attempting to do with line 203. I couldn't figure out another way to avoid focus being put on it. Do you have any immediate thoughts?

ericsoderberghp: I was suggesting that we don't prevent it from getting focus, but we steal it back as soon as it takes it.

src/js/components/List/List.js
taysea: Thanks! I wasn't aware of this hook.

src/js/components/List/List.js
taysea: I see -- I tried applying the same logic as was applied with `onClick` which is `listRef.current.focus();` but it doesn't seem to be having the same effect on this event. Will need to dive in a little more.

taysea: This is resolved by instead grabbing focus back in the `onFocus` handler instead of touching `onMouseDown`. thank you, @ericsoderberghp!

src/js/components/List/List.js
```diff
@@ -5,21 +5,17 @@ import { Box } from '../Box';
...
 
-  ${props =>
-    props.onClickItem &&
-    `
-    &:focus {
-      ${focusStyle({ forceOutline: true, skipSvgChildren: true })}
-    }
-  `}
+  &:focus {
+    ${props => props.tabIndex >= 0 && focusStyle({ skipSvgChildren: true })}
```
ericsoderberghp, line 17: Are you sure we don't want `forceOutline: true`?

src/js/components/List/List.js
```diff
@@ -5,21 +5,17 @@ import { Box } from '../Box';
...
   ${genericStyles}
 
-  ${props =>
-    props.onClickItem &&
-    `
-    &:focus {
-      ${focusStyle({ forceOutline: true, skipSvgChildren: true })}
-    }
-  `}
+  &:focus {
```
ericsoderberghp, line 16: Can we put `&:focus` inside the check for `tabIndex`? That will keep the CSS a bit cleaner.

taysea: When I place the &:focus instead of the check, it does not apply the focus style. This is the code I was using:
```
  ${props =>
    props.tabIndex >= 0 &&
    `
    &:focus {
      ${focusStyle({ forceOutline: true, skipSvgChildren: true })}
    }
  `}
```

src/js/components/List/List.js
taysea: What is the difference of using forceOutline vs not? What I noticed was that it wasn't applying the same box-shadow styling as the typical focus indicator. instead, it was a solid border with no fuzziness around the edges.

taysea:
> Also, not sure why there are theme changes in this review.

I think my yarn.lock was out of date potentially -- pushing a commit that reverts these changes.

src/js/components/List/List.js
taysea: I'm looking at `focusStyles` and seeing that by adding `forceOutline` we are entering this block of code, but I am not fully understanding the rationale for using it vs not using it: https://github.com/grommet/grommet/blob/master/src/js/utils/styles.js#L217

ericsoderberghp: If the theme wants the focus to be drawn via a border, this flag indicates that for this situation, we should use an `outline` instead, which is what I think we want here, since the `<ul />` doesn't have a border itself.

taysea: Thanks for the clarification. I see this better in the code now.

src/js/components/List/List.js
taysea: I don't believe any CSS is being added for List without an onClickItem because the browser is not interpreting it as focusable

PR #4367, opened 2020-07-30
Adds type for `theme.button.disabled.opacity`

### LINKED ISSUE #4339: [TS] Missing type for `theme.button.disabled.opacity`
# Problem

`theme.button.disabled` is declared here:
https://github.com/grommet/grommet/blob/76ed3c4f733049aef924a90e10bf25907babf094/src/js/themes/base.d.ts#L330

`ButtonKindType` is declared here without an `opacity` prop:
https://github.com/grommet/grommet/blob/76ed3c4f733049aef924a90e10bf25907babf094/src/js/themes/base.d.ts#L82-L94

`theme.button.disabled.opacity` is being used here:
https://github.com/grommet/grommet/blob/76ed3c4f733049aef924a90e10bf25907babf094/src/js/components/Button/StyledButton.js#L162

...and documented here: https://v2.grommet.io/button#button.disabled.opacity

# Expectation

`theme.button.disabled` should probably be:
```TypeScript
disabled?: ButtonKindType & { opacity?: OpacityType }; 
```

JoeButler7:
#### What does this PR do?
Adds type for `theme.button.disabled.opacity`

`src/js/themes/base.d.ts`

Closes #4339 
No
No
Backwards compatible


src/js/themes/base.d.ts
```diff
@@ -384,7 +384,7 @@ export interface ThemeType {
       primary?: ButtonKindType;
       secondary?: ButtonKindType;
     };
-    disabled?: ButtonKindType;
+    disabled?: ButtonKindType & { opacity?: OpacityType };
```
ShimiSun, line 387: Withdrawing the approval for now, should this be '|' instead of '&'?
can we add a test for it on the Button src/js/components/Button/stories/typescript/Custom.tsx story to make sure it works as expected?

JoeButler7: I think that it should be the `&`. An intersection makes more sense to me than a Union here because the button disabled will have button type and opacity, not one or the other. A test has been added in the most recent push

ShimiSun: In your example, you are not using ButtonKindType and you are only using `opacity`.

JoeButler7: Fixed in the most recent push

JoeButler7:
### Update:
We have decided to keep the intersection type (`&`) instead of the union (`|`), relevant stack-overflow issue https://stackoverflow.com/questions/53003158/whats-the-differences-between-and-in-typescript

Issue #4351, opened 2020-07-28
DateInput Range crashes after click same date three times 


maryjufang:
DateInput Range crashes after click same date three times 

### Expected Behavior

same date selected as range

ShimiSun:
https://storybook.grommet.io/?path=/story/dateinput--range

ShimiSun:
This issue is fixed on the [stable branch ](https://github.com/grommet/grommet/wiki/What-is-grommet-stable-and-how-to-use-it%3F), and will be available on our next release.

PR #4370, opened 2020-08-02
Fix: calendar crashing with three clicks

netochaves:
#### What does this PR do?
Fix calendar crashing
`src/js/components/Calendar/Calendar.js`

With storybook in range component of DateInput

#4351
I don't think so
I don't think so
backwards compatible


ericsoderberghp:
Thanks for your contribution!

ShimiSun:

PR #4378, opened 2020-08-05
New prop trapFocus added to Drop and DropButton

### LINKED ISSUE #2660: Target element of Drop and DropButton get tabindex of -1
### Expected Behavior
The target element should keep a tab index so that it is accessible by keyboard interaction.

### Actual Behavior
The target element, or in the case of the DropButton, the button gets a tabindex of -1 when the drop is in an open state. 

Here's an example of trying to build a reusable button with a tooltip on hover/focus using DropButton:

https://codesandbox.io/s/jv3nwx1jrv

User should be able to tab between the buttons.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

IanKBovard:
#### What does this PR do?
Adds a new prop called trapFocus to Drop and DropButton. 
Optional prop allows users to specify if they want focus to be trapped when the drop opens.
Accepts true or false, defaults to true(to preserve original functionality).

Drop.js, DropContainer.js, DropButton.js, FocusedContainer.js

Tested locally in storybook using this.
https://codesandbox.io/s/strange-williams-f7vmd?file=/src/App.js

Same as above. Should preserve original drop functionality.

The FocusContainer that wrapped the Drop component trapped focus whenever it was open. This prevented tabbing through other buttons.
Kapture 2020-08-05 at 15 14 28

This functionality makes sense with DropButton, when you use Drop with some inputs.
Kapture 2020-08-05 at 15 17 09

closes #2660 
trapFocus set to false

Kapture 2020-08-05 at 14 30 24

Kapture 2020-08-05 at 14 27 50

Yes, will update in this pr if trapFocus is accepted as a prop name.

Yes
Default value for trapFocus is true to keep backwards compatibility.


ShimiSun:
@srallen, since you have the most background on the addressed issue, will appreciate your feedback on this PR approach.

srallen:
@ShimiSun I think this will enable me to build a series of buttons with tooltips now that use `Drop`. The only thing I'd recommend is updating documentation for this prop and adding a story demonstrating this and why you'd use it. I believe there's an existing story showing how to build a tooltip using `Drop`, so perhaps another showing a series of components with tooltips?

IanKBovard:
added documentation, and modified the current tooltip story

src/js/components/DropButton/DropButton.js
```diff
@@ -20,6 +20,7 @@ const DropButton = forwardRef(
       onClick,
       onClose,
       onOpen,
+      trapFocus = true,
```
ericsoderberghp, line 23: Do we need this? Since Drop already defaults `trapFocus` to `true`, can we let `dropProps` take care of overriding it if needed?

IanKBovard: yeah, i think you're right. just tested without it, works the same.

src/js/components/Drop/Drop.js
```diff
@@ -12,6 +12,7 @@ const Drop = forwardRef(
     {
       restrictFocus,
       target: dropTarget, // avoid DOM leakage
+      trapFocus = true,
```
taysea, line 15: Does this have the same intention of `modal` prop in Layer? if so, would we want to match the terms. If not, ignore this comment :)

https://v2.grommet.io/layer#modal

IanKBovard: I wouldn't say the intention is the same. Modal prevents all interaction below the layer. This just locks keyboard interaction between the modal and the window. The mouse can still interact with the page when a drop is rendered.

taysea: Gotcha, thanks for clarifying!

ShimiSun:
Did you notice this change in Chromatic @IanKBovard https://www.chromatic.com/snapshot?appId=5d9774839a6eff00203f5cbf&id=5f3590dec1c8d00022d1b01e?

IanKBovard:
> Did you notice this change in Chromatic @IanKBovard https://www.chromatic.com/snapshot?appId=5d9774839a6eff00203f5cbf&id=5f3590dec1c8d00022d1b01e?

Yes, i changed the tooltip story to cover the changes in this pr.

src/js/components/Drop/stories/Tooltip.js
```diff
@@ -1,40 +1,188 @@
...
+            <Text color="white">{name}</Text>
+          </Box>
+        </Drop>
+      )}
+    </Box>
+  );
+};
+const Sidebar = () => {
+  return (
+    <Box direction="row" fill>
```
ShimiSun, line 140: If it's a Sidebar it should use the Sidebar component.

src/js/components/Drop/stories/Tooltip.js
```diff
@@ -1,40 +1,188 @@
...
+const Tooltips = () => {
+  return (
+    <Grommet theme={grommet} full>
+      <Sidebar />
     </Grommet>
   );
 };
 
-storiesOf('Drop', module).add('Tooltip', () => <TooltipDrop />);
+storiesOf('Drop', module).add('Tooltips', () => <Tooltips />);
```
ShimiSun, line 188: I think we still need a single Tooltip story so the users can find an easy way to come up with tooltips. So separating Tooltips to a separate story would be a good approach here, unless we can cover it with a unit test and then we won't need the story, thoughts?

src/js/components/Drop/stories/Tooltip.js
```diff
@@ -1,40 +1,188 @@
...
+                  icon={icon}
+                  index={index}
+                  name={name}
+                />
+              );
+            })}
+          </Nav>
+        </Box>
+        <Box>
+          <SidebarButton icon={<Action />} index="Action" name="Action" />
```
ShimiSun, line 173: The last icon on theSidebar is looking funky
image

src/js/components/Drop/stories/Tooltip.js
```diff
@@ -1,40 +1,188 @@
...
+  );
+};
+const Sidebar = () => {
+  return (
+    <Box direction="row" fill>
+      <Box
+        direction="column"
+        basis="15%"
+        justify="between"
+        style={{ border: 'blue solid 4px' }}
```
ShimiSun, line 145: We shouldn't use `style` in the story examples, why is that needed?

src/js/components/Drop/stories/Tooltip.js
IanKBovard: i gotcha, makes sense. ill have a new, simpler example up soon.

IanKBovard:
updated the storybook example

PR #4381, opened 2020-08-06
Changed SkipLinks 

### LINKED ISSUE #4360: SkipLink example broken in docs site
### Expected Behavior
When using the examples given on the grommet docs site, I'd expect the skiplink to be working.

### Actual Behavior
The skiplinks don't work in the examples on the grommet site

https://v2.grommet.io/skiplinks

### Steps to Reproduce
Try any of the skip link examples on the grommet site.

### Your Environment

- Grommet version: any since v2
- Operating System and version (desktop or mobile): Mac OSX

ShimiSun:
#### What does this PR do?
1. Allows the SkipLinks layer to open by entering Tab on the keyboard once the page loaded.
1. Allows skipping the SkipLinks layer content and tabbing to the header.
1. Allows editing the SkipLinks message and styling the container, layer, and the text message.
1. Allows cyclic navigation of the page that includes the layer using the keyboard.
1. Enabling access to the rest of the interactive elements while the layer is shown.
1. Changing the DOM behavior to prepend the Layer to the DOM so tabIndex order will behave correctly by default.
1. Provides clear examples of how to use the component on Storybook.
1. Tested with typescript.
1. Change initial focus behavior to the first focused link instead of the container.

SkipLinks.js
Manual testing - many.
Typescript local testing - all passed.
Go to the storybook examples of SkipLinks, enter Tab on the keyboard, and follow directions.
I would like to add one more enhancement on a separate PR, that the first focusable element will be the first SkipLink element and not the container as its behaving now (and always behaved that way). I can file it on a separate PR if agreed.
Edit: this functionality is already done on this PR.
closes #4360 
Done
Yes
backwards compatible as this PR is a bug fix.

### DIFF
--- a/src/js/components/Layer/Layer.js
+++ b/src/js/components/Layer/Layer.js
@@ -8,14 +8,15 @@ import { animationDuration } from './StyledLayer';
 import { ContainerTargetContext } from '../../contexts/ContainerTargetContext';
 
 const Layer = forwardRef((props, ref) => {
-  const { animate, animation, bodyTarget } = props;
+  const { animate, animation, targetChildPosition } = props;
   const [originalFocusedElement, setOriginalFocusedElement] = useState();
   useEffect(() => setOriginalFocusedElement(document.activeElement), []);
   const [layerContainer, setLayerContainer] = useState();
   const containerTarget = useContext(ContainerTargetContext);
   useEffect(
-    () => setLayerContainer(getNewContainer(containerTarget, bodyTarget)),
-    [containerTarget, bodyTarget],
+    () =>
+      setLayerContainer(getNewContainer(containerTarget, targetChildPosition)),
+    [containerTarget, targetChildPosition],
   );
 
   // just a few things to clean up when the Layer is unmounted
--- a/src/js/components/Layer/Layer.js
+++ b/src/js/components/Layer/Layer.js
@@ -8,14 +8,15 @@ import { animationDuration } from './StyledLayer';
 import { ContainerTargetContext } from '../../contexts/ContainerTargetContext';
 
 const Layer = forwardRef((props, ref) => {
-  const { animate, animation } = props;
+  const { animate, animation, bodyTarget } = props;
   const [originalFocusedElement, setOriginalFocusedElement] = useState();
   useEffect(() => setOriginalFocusedElement(document.activeElement), []);
   const [layerContainer, setLayerContainer] = useState();
   const containerTarget = useContext(ContainerTargetContext);
-  useEffect(() => setLayerContainer(getNewContainer(containerTarget)), [
-    containerTarget,
-  ]);
+  useEffect(
+    () => setLayerContainer(getNewContainer(containerTarget, bodyTarget)),
+    [containerTarget, bodyTarget],
+  );
 
   // just a few things to clean up when the Layer is unmounted
   useEffect(
--- a/src/js/components/SkipLink/SkipLink.js
+++ b/src/js/components/SkipLink/SkipLink.js
@@ -1,10 +1,10 @@
-import React from 'react';
+import React, { forwardRef } from 'react';
 
 import { Anchor } from '../Anchor';
 import { Box } from '../Box';
 
-export const SkipLink = ({ id, label, ...rest }) => (
+export const SkipLink = forwardRef(({ id, label, ...rest }, ref) => (
   <Box margin="small">
-    <Anchor href={`#${id}`} label={label} {...rest} />
+    <Anchor href={`#${id}`} ref={ref} label={label} {...rest} />
   </Box>
-);
+));
--- a/src/js/components/SkipLink/SkipLink.js
+++ b/src/js/components/SkipLink/SkipLink.js
@@ -1,10 +1,7 @@
 import React, { forwardRef } from 'react';
 
 import { Anchor } from '../Anchor';
-import { Box } from '../Box';
 
 export const SkipLink = forwardRef(({ id, label, ...rest }, ref) => (
-  <Box margin="small">
-    <Anchor href={`#${id}`} ref={ref} label={label} {...rest} />
-  </Box>
+  <Anchor href={`#${id}`} ref={ref} label={label} {...rest} />
 ));
--- a/src/js/components/SkipLinks/README.md
+++ b/src/js/components/SkipLinks/README.md
@@ -94,7 +94,7 @@ Defaults to
 { margin: 'small', size: 'medium' }
 ```
 
-**skipLinks.text.margin**
+**skipLinks.label.margin**
 
 The margin size around the text message. Expects `string`.
 
@@ -104,7 +104,7 @@ Defaults to
 medium
 ```
 
-**skipLinks.text.size**
+**skipLinks.label.size**
 
 The font size of the text label. Expects `string`.
 
--- a/src/js/components/SkipLinks/README.md
+++ b/src/js/components/SkipLinks/README.md
@@ -23,7 +23,7 @@ node
 
 Custom messages for SkipLinks. Used for accessibility by screen 
 readers. Defaults to `{
-  "skipTo": "Skip To"
+  "skipTo": "Skip To:"
 }`.
 
 ```
--- a/src/js/components/SkipLinks/README.md
+++ b/src/js/components/SkipLinks/README.md
@@ -32,3 +32,94 @@ readers. Defaults to `{
 }
 ```
   
+## Theme
+  
+**skipLinks.position**
+
+Position of the layer content once opened. Expects `string`.
+
+Defaults to
+
+```
+top
+```
+
+**skipLinks.container**
+
+Any valid Box prop for the SkipLinks container. Expects `object`.
+
+Defaults to
+
+```
+{ elevation: 'large', pad: 'medium', round: 'small' }
+```
+
+**skipLinks.container.elevation**
+
+The container shadow. Expects `string`.
+
+Defaults to
+
+```
+large
+```
+
+**skipLinks.container.pad**
+
+The pad used for the layer container. Expects `string | object`.
+
+Defaults to
+
+```
+medium
+```
+
+**skipLinks.container.round**
+
+The rounding of the later container. Expects `boolean | string | object`.
+
+Defaults to
+
+```
+small
+```
+
+**skipLinks.text**
+
+Any valid Text prop for the text message. Expects `object`.
+
+Defaults to
+
+```
+{ margin: 'small', size: 'medium' }
+```
+
+**skipLinks.text.margin**
+
+The margin size around the text message. Expects `string`.
+
+Defaults to
+
+```
+medium
+```
+
+**skipLinks.text.size**
+
+The font size of the text label. Expects `string`.
+
+Defaults to
+
+```
+medium
+```
+
+**text.medium.size**
+
+The font size of the text label. Expects `string`.
+
+Defaults to
+
+```
+18px
+```
--- a/src/js/components/SkipLinks/README.md
+++ b/src/js/components/SkipLinks/README.md
@@ -84,7 +84,7 @@ Defaults to
 small
 ```
 
-**skipLinks.text**
+**skipLinks.label**
 
 Any valid Text prop for the text message. Expects `object`.
 
--- a/src/js/components/SkipLinks/README.md
+++ b/src/js/components/SkipLinks/README.md
@@ -101,7 +101,7 @@ The margin size around the text message. Expects `string`.
 Defaults to
 
 ```
-medium
+{ bottom: medium }
 ```
 
 **skipLinks.label.size**
--- a/src/js/components/SkipLinks/SkipLinks.js
+++ b/src/js/components/SkipLinks/SkipLinks.js
@@ -10,10 +10,24 @@ const SkipLinks = ({ children, id, messages }) => {
   const theme = useContext(ThemeContext) || defaultProps.theme;
   const [showLayer, setShowLayer] = useState(false);
 
+  const firstAnchorRef = useRef(null);
   const layerRef = useRef(null);
 
   const onFocus = () => {
     setShowLayer(true);
+    // timeout needed so it gives enough time for activeElement to be updated
+    setTimeout(() => {
+      // The layer container is receiving the focus by default since the layer
+      // isn't on the DOM yet and its children are not accessible.
+      // Once the layer is focused and the children of the layer are
+      // accessible on the DOM, the first anchor of the skiplink list will
+      // steal the focus to itself.
+      const layerNode = layerRef.current;
+      // change the focus to the first SkipLink anchor
+      if (layerNode === document.activeElement) {
+        firstAnchorRef.current.focus();
+      }
+    }, 0);
   };
 
   const removeLayer = () => {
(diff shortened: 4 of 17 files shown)


src/js/components/SkipLinks/SkipLinks.js
```diff
@@ -35,29 +37,38 @@ const SkipLinks = ({ children, id, messages }) => {
...
-            cloneElement(element, {
-              key: `skip-link-${index}`,
-              onClick: removeLayer,
-            }),
+      {showLayer && (
+        <Box {...theme.skipLinks.container}>
+          {messages.skipTo && (
+            <Text {...theme.skipLinks.text}>{messages.skipTo}</Text>
           )}
+          <Box align="center">
```
ericsoderberghp, line 55: What if the theme wants a different alignment?

I'm getting nervous that we're putting more and more theming layout and styling in the JSON theme when they would be better handled via JSX markup. Just thinking out loud.

src/js/components/SkipLinks/SkipLinks.js
```diff
@@ -35,29 +37,38 @@ const SkipLinks = ({ children, id, messages }) => {
...
+            {children.map((element, index) =>
+              cloneElement(element, {
+                key: `skip-link-${index}`,
+                onClick: removeLayer,
```
ericsoderberghp, line 59: What if the child already has an `onClick` handler?

src/js/utils/DOM.js
```diff
@@ -85,10 +85,15 @@ export const getBodyChildElements = () => {
   return children;
 };
 
-export const getNewContainer = (rootNode = document.body) => {
+export const getNewContainer = (rootNode = document.body, bodyTarget) => {
```
ericsoderberghp, line 88: I think `rootTarget` would be more aligned with `rootNode`. Although, I think `rootChildPosition` would be even more accurate.

src/js/components/SkipLinks/SkipLinks.js
```diff
@@ -35,29 +37,38 @@ const SkipLinks = ({ children, id, messages }) => {
   return (
     <Layer
       id={id}
-      position={showLayer ? 'top' : 'hidden'}
+      position={showLayer ? theme.skipLinks.position : 'hidden'}
       ref={layerRef}
       onFocus={onFocus}
       onBlur={onBlur}
+      tabIndex={0}
```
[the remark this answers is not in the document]
ShimiSun: I agree. 
Until now, that is how it used to behave since the implementation used to have a `modal`.
I've put a comment on the PR description:
> I would like to add one more enhancement on a separate PR, that the first focusable element will be the first SkipLink element and not the container as its behaving now (and always behaved that way). I can file it on a separate PR if agreed.

Let me know what you think.

src/js/components/Layer/Layer.js
```diff
@@ -8,14 +8,15 @@ import { animationDuration } from './StyledLayer';
 import { ContainerTargetContext } from '../../contexts/ContainerTargetContext';
 
 const Layer = forwardRef((props, ref) => {
-  const { animate, animation } = props;
+  const { animate, animation, bodyTarget } = props;
```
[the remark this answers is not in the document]
ShimiSun: I'm pretty sure we closed on `bodyTarget`, feel free to make any other suggestions, it's an internal prop so will be flexible for editing.

src/js/components/SkipLinks/SkipLinks.js
```diff
@@ -35,29 +37,38 @@ const SkipLinks = ({ children, id, messages }) => {
...
+      modal={false}
+      bodyTarget="first" // prepend the Layer on the body container
+      // modal={false} will take the full screen a small breakpoint
```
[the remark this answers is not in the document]
ShimiSun: I added this comment because I thought it is valuable to understand why I've assigned `responsive={false}`, do you disagree?

src/js/components/SkipLinks/SkipLinks.js
```diff
@@ -35,29 +37,38 @@ const SkipLinks = ({ children, id, messages }) => {
...
+      responsive={false}
```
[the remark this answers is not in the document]
ShimiSun: See my previous comment :)

src/js/components/SkipLinks/SkipLinks.js
```diff
@@ -35,29 +37,38 @@ const SkipLinks = ({ children, id, messages }) => {
...
-        <Box direction="row" align="center" pad={{ bottom: 'medium' }}>
-          {children.map((element, index) =>
-            cloneElement(element, {
-              key: `skip-link-${index}`,
-              onClick: removeLayer,
-            }),
+      {showLayer && (
+        <Box {...theme.skipLinks.container}>
+          {messages.skipTo && (
+            <Text {...theme.skipLinks.text}>{messages.skipTo}</Text>
```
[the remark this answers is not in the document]
ShimiSun: Hmmm... it is the text in the Layer, but just the message text.
Would `theme.skipLinks.message.text` be clearer?

src/js/components/SkipLinks/SkipLinks.js
ShimiSun: Not sure, I didn't delve to this part of the functionality, it was always there. I can take a look though.

src/js/utils/DOM.js
ShimiSun: I like `rootChildPosition`. I'll use that.

src/js/components/SkipLinks/SkipLinks.js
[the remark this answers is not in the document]
ericsoderberghp: Hmm... is that how skip links work on other sites? I think it's more important to be compatible with the general use of skip links than with prior grommet behavior in this case.

src/js/components/Layer/Layer.js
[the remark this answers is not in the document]
ericsoderberghp: `targetChildPosition`?

src/js/components/SkipLinks/SkipLinks.js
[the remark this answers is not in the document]
ericsoderberghp: Ah, it looked more like you had commented out `model={false}` to me. How about something like:

"Non-modal Layer's will take the full screen at small breakpoints by default, which isn't what we want."

src/js/components/SkipLinks/SkipLinks.js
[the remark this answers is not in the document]
ericsoderberghp: What about `theme.skipLink.label`?

src/js/components/SkipLinks/SkipLinks.js
ericsoderberghp: We can wait until a need arises.

src/js/utils/DOM.js
ShimiSun: I actually liked your proposal above to use `targetChildPosition` and used that.

src/js/components/SkipLinks/SkipLinks.js
```diff
@@ -20,12 +23,11 @@ const SkipLinks = ({ children, id, messages }) => {
   const onBlur = () => {
     // timeout needed so it gives enough time for activeElement to be updated
     setTimeout(() => {
-      const layerNode = layerRef.current;
+      // close the layer in case the activeElement isn't an anchor
```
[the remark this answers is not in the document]
ShimiSun: I agree. Fixed on the recent commit.

src/js/components/Layer/Layer.js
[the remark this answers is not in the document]
ShimiSun: Yes. done.

src/js/components/SkipLinks/SkipLinks.js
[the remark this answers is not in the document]
ShimiSun: Done.

src/js/components/SkipLinks/SkipLinks.js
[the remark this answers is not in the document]
ShimiSun: Done.

src/js/components/SkipLinks/SkipLinks.js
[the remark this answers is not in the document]
ShimiSun: Done on the recent commit, and I've updated the PR description.

src/js/components/SkipLink/SkipLink.js
```diff
@@ -1,10 +1,10 @@
-import React from 'react';
+import React, { forwardRef } from 'react';
 
 import { Anchor } from '../Anchor';
 import { Box } from '../Box';
 
-export const SkipLink = ({ id, label, ...rest }) => (
+export const SkipLink = forwardRef(({ id, label, ...rest }, ref) => (
   <Box margin="small">
```
ericsoderberghp, line 7: I'm wondering if `SkipLinks` should control all of the layout and use `gap` and `margin` and leave `SkipLink` to be just the `Anchor`?

src/js/components/SkipLinks/README.md
```diff
@@ -23,12 +23,103 @@ node
...
+
+Any valid Text prop for the text message. Expects `object`.
+
+Defaults to
+
+```
+{ margin: 'small', size: 'medium' }
+```
+
+**skipLinks.text.margin**
```
ericsoderberghp, line 97: `skipLinks.label.margin`?

Or, maybe I'm misunderstanding something. How does `skipLinks.label.margin` relate to `skipLinks.text.margin`?

src/js/components/SkipLinks/SkipLinks.js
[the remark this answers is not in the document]
ericsoderberghp: I'm still not sure why the Layer needs `tabIndex={0}`. I was thinking that the first `SkipLink` would be what receives the focus already.

src/js/components/SkipLinks/SkipLinks.js
```diff
@@ -35,29 +48,40 @@ const SkipLinks = ({ children, id, messages }) => {
...
     >
-      <Box pad={{ horizontal: 'medium' }}>
-        <Heading level={2}>{messages.skipTo}:</Heading>
-        <Box direction="row" align="center" pad={{ bottom: 'medium' }}>
-          {children.map((element, index) =>
-            cloneElement(element, {
-              key: `skip-link-${index}`,
-              onClick: removeLayer,
-            }),
+      {showLayer && (
```
ericsoderberghp, line 62: I would have thought we'd always render each `SkipLink` and let the user's tab take care of giving it focus and causing the Layer to remove 'hidden'.

src/js/utils/DOM.js
```diff
@@ -85,10 +85,18 @@ export const getBodyChildElements = () => {
   return children;
 };
 
-export const getNewContainer = (rootNode = document.body) => {
+export const getNewContainer = (
+  rootNode = document.body,
```
ericsoderberghp, line 89: since we're calling the new argument `targetChildPosition`, I think we should rename `rootNode` to `target` here.

src/js/components/SkipLink/SkipLink.js
ShimiSun: You read my mind. 
I actually have code in place that does just that, but then I thought maybe we should keep the Box and SkipLink will have a theme to style this Box container and the user will be able to control margin, padding, or any other Box property for an individual SkipLink. 
Two of the approaches sounded like they could work for me, so I said I should leave it as is and see which need arises :)
If you have a strong preference I'm good either way.

src/js/components/SkipLinks/README.md
ShimiSun: Oopsy, good catch.

src/js/utils/DOM.js
ShimiSun: Done.

src/js/components/SkipLinks/README.md
ShimiSun: Fixed.

src/js/components/SkipLinks/SkipLinks.js
ShimiSun: I suspect there might be an issue with how Layer is interpolating `position="hidden"` with `modal={false}` that drove this condition.
Even when the Layer is set to the hidden position, it's still being displayed, though removing modal={false} makes the Layer hidden as expected. 
I'll delve into Layer to find the root cause of this issue -- unless you think there should be a good reason for not letting hidden work on `modal={false}`.

src/js/components/SkipLinks/SkipLinks.js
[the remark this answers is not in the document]
ShimiSun: There are comments in SkipLinks `onFocus` that explains how the focus is being addressed, and how this tabIndex is playing an important part in it.
That being said, as part of investigating how to let `position='hidden'` do its thing and addressing the Layer issue that doesn't hide the Layer on `modal={false}`, I might be able to simplify this tabIndex behavior and remove it altogether. 
I'll keep this comment open and circle back to it once I'll have some resolutions.

src/js/components/SkipLinks/SkipLinks.js
ShimiSun: Update: 
I pushed a fix for Layer that addresses the issue of hiding the Layer when `position="hidden" && modal={false}` that fixes the issue for SkipLinks, and thanks to this fix I was able to remove this condition and the `tabIndex=0` from the SkipLinks logic.

src/js/components/SkipLinks/SkipLinks.js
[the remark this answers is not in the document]
ShimiSun: Update: 
I pushed a fix for Layer that addresses the issue of hiding the Layer when `position="hidden" && modal={false}` and that fixes the issue for SkipLinks. Thanks to this fix I was able to remove `tabIndex=0` so this concern should be out of our way now.

src/js/components/SkipLink/SkipLink.js
ericsoderberghp: Since it's a "Link", I think it should just be an "Anchor". I don't think we need to do too much styling for it, since it's designed more for users with sight issues.

src/js/components/SkipLinks/SkipLinks.js
```diff
@@ -35,14 +34,21 @@ const SkipLinks = ({ children, id, messages }) => {
   return (
     <Layer
       id={id}
-      position={showLayer ? 'top' : 'hidden'}
+      position={showLayer ? theme.skipLinks.position : 'hidden'}
       ref={layerRef}
       onFocus={onFocus}
       onBlur={onBlur}
+      modal={false}
+      targetChildPosition="first" // prepend the Layer on the body container
```
ericsoderberghp, line 42: Can we add just a bit more to this comment to indicate why we are doing this? For example:

"Prepend the Layer so any SkipLink will be the first element that pressing the Tab key reaches"

src/js/components/SkipLink/SkipLink.js
ericsoderberghp: I think SkipLink should just be Anchor. Since this is for users with sight issues, I don't think we need to invest a lot for styling it.

ShimiSun: Ok, I'll clean it up.

src/js/components/SkipLinks/SkipLinks.js
ShimiSun: Absolutely. good comment.

src/js/components/SkipLink/SkipLink.js
ShimiSun: Done.

PR #4382, opened 2020-08-06
add margin so focus is not getting cut off

britt6612:
#### What does this PR do?
Proof of concept providing margin to calendar so that the focus does not get cut off 
calendar.js
storybook
Screen Shot 2020-08-06 at 12 38 23 AM
currently the focus is being cut off in dateInput 

This is POC just to show that adding some margin will not cut off the focus.
This may be something we can add into theme for users to provide themselves 
or we can look at other approaches as well! 
design-system
no
no


ShimiSun:
Thank you for this PR and for surfacing the issue, I'm OK with the fix, but I'm wondering whether there is another elegant fix to handle the issue without adding a margin. I need to think about it for a bit or maybe open it for discussion.

britt6612:
I was also looking at other ways.. Since `Calendar` does not have pad as a prop I went ahead and went with margin. I also think if there is a way I can also fix this with some overflow then that may give us another option to go with.

britt6612:
I went ahead and added pad to the box containing just the date and the icons as something else for us to look at. I was trying to mess with the overflow but couldnt get anything working

ericsoderberghp:
I think it's more of an `overflow` issue. Have you tried `<Drop overflow="visible" />` within DateInput?

britt6612:
> I think it's more of an `overflow` issue. Have you tried `<Drop overflow="visible" />` within DateInput?



> I think it's more of an `overflow` issue. Have you tried `<Drop overflow="visible" />` within DateInput?

Sorry @ericsoderberghp must of missed this!
But yes the overflow on the drop worked correctly

Issue #4334, opened 2020-07-22
InfiniteScroll props.show only works when props.show < props.step IF replace = true.


valkenburg:
### Expected Behavior
```
<InfiniteScroll
                    replace
                    show={20}
                    step={10}
                    items={items}
                >
```
should scroll down to item at index 20 upon the first rendering.

### Actual Behavior
doesn't scroll, unless we set props.step to a value > 20, such that the item is anyway rendered, **or** we remove the flag `replace`. Only then, InfiniteScroll will scroll to it.

### URL, screen shot, or Codepen exhibiting the issue

[https://codesandbox.io/s/grommet-v2-template-zb62p?file=/index.js](https://codesandbox.io/s/grommet-v2-template-zb62p?file=/index.js)

### Steps to Reproduce

enable / disable 'replace' to see the effect.

### Your Environment

- Grommet version: 2.14.0
- Browser Name and version: Brave Version 1.10.97 Chromium: 83.0.4103.116 (Official Build) (64-bit)
- Operating System and version (desktop or mobile): OS X 10.15.4

halocline:
Upon initial investigation, I am finding a few oddities. Documenting for context:
- `show` behavior does not scroll to specified index: [Storybook Example](https://storybook.grommet.io/?path=/story/infinitescroll--show-118th-item) --> should contain index 117
- when `show` is specified, the result set does not render items past the show indexes page: [Storybook Example](https://storybook.grommet.io/?path=/story/infinitescroll--show-118th-item) --> replicate by attempting to scroll past index 150
- when `show` is specified, result set displays duplicated items at `step` index (default 50): [Storybook Example](https://storybook.grommet.io/?path=/story/infinitescroll--show-118th-item) --> replicate by scrolling to 'item 50', you will see two 'item 50' in the result set. Also result in a console warning for duplicate key of 'last'
- Was able to replicate case where `replace && show > step` only displays first page of results, does not display results page containing show index, replace page does not execute.

Because of these discreet behaviors, I will likely submit a PR/PRs addressing the first 3 scenarios. Then submit PR for the more complex scenario where `replace`, `show`, and `step` are potential factors.

PR #4390, opened 2020-08-07
Infinite Scroll test number of items generated

jcfilben:
#### What does this PR do?
Adds testing to infinite scroll `step` prop. Tests small, medium, and large step sizes. Ensures that the number of elements being rendered is correct.

NOTE: this also tests that only unique items are being rendered, (no duplicates). On the `length when step < array` test if you uncomment `show={117}` the test will generate a warning and fail because of a bug causing there to be duplicate items when `show` is set and it is a higher value than `step`.

`yarn test` before and after uncommenting line 89

Screen Shot 2020-08-07 at 4 49 44 PM

No

No

Backwards compatible

### DIFF
--- a/src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
+++ b/src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
@@ -83,8 +83,11 @@ describe('Number of Items Rendered', () => {
     Array(value)
       .fill()
       .map((_, i) => `item ${i + 1}`);
-  function filterRefs(childItem) {
-    return childItem.outerHTML.includes('item');
+
+  function createPageItems(allChildren) {
+    const unfiltered = Array.from(allChildren);
+    // Removing any children which are serving as refs
+    return unfiltered.filter(childItem => childItem.outerHTML.includes('item'));
   }
 
   test(`Should render items equal to the length of 
@@ -102,9 +105,7 @@ describe('Number of Items Rendered', () => {
       </Grommet>,
     );
 
-    const allChildren = Array.from(container.firstChild.children);
-    // Removing any children which are serving as refs
-    const pageItems = allChildren.filter(filterRefs);
+    const pageItems = createPageItems(container.firstChild.children);
     const expectedItems = step;
     expect(pageItems.length).toEqual(expectedItems);
   });
@@ -120,9 +121,7 @@ describe('Number of Items Rendered', () => {
       </Grommet>,
     );
 
-    const allChildren = Array.from(container.firstChild.children);
-    // Removing any children which are serving as refs
-    const pageItems = allChildren.filter(filterRefs);
+    const pageItems = createPageItems(container.firstChild.children);
     const expectedItems = step;
     expect(pageItems.length).toEqual(expectedItems);
   });
@@ -138,9 +137,7 @@ describe('Number of Items Rendered', () => {
       </Grommet>,
     );
 
-    const allChildren = Array.from(container.firstChild.children);
-    // Removing any children which are serving as refs
-    const pageItems = allChildren.filter(filterRefs);
+    const pageItems = createPageItems(container.firstChild.children);
     const expectedItems = numItems;
     expect(pageItems.length).toEqual(expectedItems);
   });
--- a/src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
+++ b/src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
@@ -98,8 +98,8 @@ describe('InfiniteScroll', () => {
 
     const allChildren = Array.from(container.firstChild.children);
     // Removing any children which are serving as refs
-    const pageItems = allChildren.filter(
-      childItem => !!childItem.outerHTML.includes('item'),
+    const pageItems = allChildren.filter(childItem =>
+      childItem.outerHTML.includes('item'),
     );
     const expectedItems = step;
     expect(pageItems.length).toEqual(expectedItems);
@@ -118,8 +118,8 @@ describe('InfiniteScroll', () => {
 
     const allChildren = Array.from(container.firstChild.children);
     // Removing any children which are serving as refs
-    const pageItems = allChildren.filter(
-      childItem => !!childItem.outerHTML.includes('item'),
+    const pageItems = allChildren.filter(childItem =>
+      childItem.outerHTML.includes('item'),
     );
     const expectedItems = step;
     expect(pageItems.length).toEqual(expectedItems);
@@ -138,8 +138,8 @@ describe('InfiniteScroll', () => {
 
     const allChildren = Array.from(container.firstChild.children);
     // Removing any children which are serving as refs
-    const pageItems = allChildren.filter(
-      childItem => !!childItem.outerHTML.includes('item'),
+    const pageItems = allChildren.filter(childItem =>
+      childItem.outerHTML.includes('item'),
     );
     const expectedItems = numItems;
     expect(pageItems.length).toEqual(expectedItems);
--- a/src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
+++ b/src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
@@ -4,6 +4,7 @@ import 'jest-styled-components';
 
 import { Grommet } from '../../Grommet';
 import { InfiniteScroll } from '..';
+import { Box } from '../..';
 
 describe('InfiniteScroll', () => {
   const items = [];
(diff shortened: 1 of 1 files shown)


ShimiSun:
Relevant issues #4369 , #4334

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
```diff
@@ -75,4 +76,52 @@ describe('InfiniteScroll', () => {
...
+          step={step}
+        >
+          {item => <Box key={item}>{item}</Box>}
+        </InfiniteScroll>
+      </Grommet>,
+    );
+    expect(container.firstChild.children.length).toEqual(step + 1);
+  });
+
+  test('length when step = array', () => {
```
halocline, line 99: Nice. Was going to suggest adding this case.

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
```diff
@@ -75,4 +76,52 @@ describe('InfiniteScroll', () => {
     );
     expect(container.firstChild).toMatchSnapshot();
   });
+
+  test('length when step < array', () => {
+    const allItems = Array(1000)
+      .fill()
+      .map((_, i) => `item ${i + 1}`);
```
halocline, line 83: Since this seems like we are repeating this logic across several tests, does it make sense for us to abstract this and call it something like `simpleItems`?

Or is the direction to keep the test completely self contained?

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
```diff
@@ -75,4 +76,52 @@ describe('InfiniteScroll', () => {
     );
     expect(container.firstChild).toMatchSnapshot();
   });
+
+  test('length when step < array', () => {
```
halocline, line 80: Reading the descriptions/names of the tests, I am able to understand the test scenario, but it is not immediately clear what the expected results or outcome should be.

Could these be written in the form of "Should [expectation / outcome] when [scenario]?" For example, "Should render items equal to the length of step when step < items.length" or something similar. I can reason with the code to try to interpret the intent, but naming in this form might be helpful. Thoughts?

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
```diff
@@ -75,4 +76,52 @@ describe('InfiniteScroll', () => {
...
+        <InfiniteScroll
+          items={allItems}
+          // show={117}
+          step={step}
+        >
+          {item => <Box key={item}>{item}</Box>}
+        </InfiniteScroll>
+      </Grommet>,
+    );
+    expect(container.firstChild.children.length).toEqual(step + 1);
```
halocline, line 96: Why `step + 1`? If I am unfamiliar with how this component is implemented, then I may not have context for why adding 1 to the step. A comment could be helpful here.

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
IanKBovard: I agree, I think the `should ---> when` format provides a good amount of description.

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
halocline, line 96: This is a style suggestion which I think helps readability and intent for those who follow afterward. It's a bit more verbose, but helps communicate what each of these values represent.

```suggestion
    const renderedItems = container.firstChild.children.length;
    const expectedItems = step + 1 // comment explaining why + 1
    expect(renderedItems).toEqual(expectedItems);
```

IanKBovard: i could be wrong, but i believe is because it'll render items 0(inclusive)-50(inclusive)? making it 51 items rendered.

halocline: I think in this case, the 51st element is actually a 0 height ref to mark the end of the page of results.

IanKBovard: ah, yes. that makes more sense

jcfilben: I added comments about `step + 1`. Let me know if they are too long or could be phrased better

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
jcfilben: Yes I agree it makes sense to get rid of the repeating logic. I added `simpleItems` at the beginning of the test suite so it can be used by all 3 of the tests instead of being duplicated

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
jcfilben: I renamed the tests to variations of `Should render items equal to the length of 
  step + 1 when step < items.length`. I agree it is more descriptive than what the tests were named previously.

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
jcfilben: Updated tests to include this style suggestion

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
```diff
@@ -75,4 +80,62 @@ describe('InfiniteScroll', () => {
...
+          {item => <Box key={item}>{item}</Box>}
+        </InfiniteScroll>
+      </Grommet>,
+    );
+    // When step > the length of the number of items,
+    // an additional item is NOT rendered because the items
+    // don't approach the end of the page. This is why the total
+    // number of items is the length of simpleItems, 1000.
+    const renderedItems = container.firstChild.children.length;
+    const expectedItems = 1000;
```
halocline, line 138: What would happen if simpleItems was redefined at some point in the future?

```suggestion
    const expectedItems = simpleItems.length;
```

halocline, line 138: Rather than hardcoding 1000 twice, can we `const numItems = 1000;` and then have `simpleItems(numItems)` and expectedItems = numItems?

jcfilben: Yes, fixed this in latest push

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
```diff
@@ -75,4 +80,62 @@ describe('InfiniteScroll', () => {
     );
     expect(container.firstChild).toMatchSnapshot();
   });
+
+  test(`Should render items equal to the length of 
+  step + 1 when step < items.length`, () => {
+    const step = 50;
```
halocline, line 86: Just calling this out to set stage for future context:

As our test scenarios get more complex, we will likely see a couple of other marker refs being included in the array. We will also likely want to find a way to re-use and reference these three tests by other tests.

As we write those scenarios, these tests may get refactored to accommodate. That said, I think the current state helps us move forward and we can refactor as needed.

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
halocline: I think what we have here is good for now. We may revisit as we introduce additional tests and try to repurpose what is being done here.

For context, what '1' is representing here is an additional ref to serve as an anchor/trigger for additional functionality. As soon as we add a prop such as `show` there will be an additional ref marking the show index position in the middle of the resulting items and so the children in the array would then be step + 2.

@IanKBovard can you think of a way where we can test for `renderedItems.length === step`? Perhaps we do something like:
```
    const allChildren = container.firstChild.children;
    // Removing any children which are serving as refs
    const refKeys = ["first", "last", "show"];
    const pageItems = allChildren.filter( /* some sort of filter logic which returns children where there key is not in refKeys */ )
    const expectedItems = step;
    expect(pageItems.length).toEqual(expectedItems);
```

IanKBovard: I agree. Filter is probably the best approach.

halocline: @jcfilben Can you take a look at the suggested approach above. If it makes sense, would you mind taking a stab at implementing?

IanKBovard: If possible, i think we can move all creating and filtering of children logic in a before block when we start to refactor these in future prs.

jcfilben: Yup, just updated the PR using the approach outlined above

Issue #4375, opened 2020-08-05
Formfield Validation does not show up 


ypinneka:
I have a use case where we need to show the error message returned from validation function of formfield. But when the error is returned via promise (backend API), it doesn't show up.

Created a codeSandBox replicating the same issue, https://codesandbox.io/s/fervent-ride-z1fdw?file=/src/App.js

Used setTimeout for simulating promise.

novykh:
Hello @ypinneka , there are two reasons for this:
1. You are not returning the actual error string.
2. FormField validations do not work in async.

I kind of [updated](https://codesandbox.io/s/fast-mountain-pe999?file=/src/App.js) your codesandbox snippet, to validate on form change.

You can do the validation on other events (of the form or the formfield) as well, but in order to display them, you need to keep and update their state.

ypinneka:
Hi @novykh 

I tried this solution prior, but it does nto cater to all use cases, please see the sndBox, https://codesandbox.io/s/boring-bird-kbkzt?file=/src/App.js
In here, if we pass errors via Form component & the specific formfield has a required already, then the error from Form does not show up.

Note: we have a very specific case where we are wrapping the validation function to create a custom error component.

novykh:
`required` takes precedence over form `errors`. In such cases, pass the custom error directly to the formField. Hope it helps.
https://codesandbox.io/s/elastic-snow-khi2x?file=/src/App.js

ypinneka:
Oh, I missed that there is an error prop for formfield as well. Will try and let you know ! Thanks !

ShimiSun:
In addition to the great suggestion coming from @novykh, you can also use the new `onValidate` function provided on the stable branch. I'll be closing this issue. feel free to reopen or file a new one if needed.

PR #4396, opened 2020-08-10
Add onValidate to Form

### LINKED ISSUE #3921: Access Validation Status of Form
Is there a way to access form validation state? If not, something like an `onValidate` event would be nice.

taysea:
#### What does this PR do?

This adds onValidate to form which allows users to access `error` and `info` when a form validates. This would allow users to access which formfields have errors prior to form submission. 

Additionally, `errors` and `infos` had been stored separately, but are now combined into a single `validationResults` state object that contains them both since validation runs on both.

This helps address issues noted in these use cases:
https://github.com/hpe-design/design-system/pull/998
https://github.com/grommet/grommet/issues/4375

Added tests in jest

Added example in `AllChildren` Form storybook.

add this to a form:
`onValidate={({error, info}) => console.log(error, info)}`

Users didn't have access to the error and info data that Form holds in state.

Closes https://github.com/hpe-design/design-system/issues/1047
Closes #3921 
Yes, updated in this PR.

Yes.

Yes, backwards compatible.


src/js/components/Form/Form.js
```diff
@@ -233,6 +234,14 @@ const Form = forwardRef(
                 // which probably came from a submit
                 setErrors(prevErrors => ({ ...prevErrors, ...nextErrors }));
                 setInfos(prevInfos => ({ ...prevInfos, ...nextInfos }));
+
+                // give user access to errors that have occurred on validation
+                if (onValidate) {
+                  const validation = {};
+                  validation.error = { ...errors, ...nextErrors };
```
ericsoderberghp, line 241: It feels slightly odd that we are using `errors` here but `prevErrors` above on line 235. I wonder if we should merge `errors` and `infos` into the same state. For example:
```
const [validationResults, setValidationResults] = useState(defaultValidationResults);
...
setValidationResults(prevValidationResults => {
  const nextValidationResults = { errors: { ...prevValidationResults.errors, ...nextErrors }, infos: ... };
  if (onValidate) onValidate(nextValidationResults);
  return nextValidationResults;
});
```

src/js/components/Form/Form.js
```diff
@@ -233,6 +234,14 @@ const Form = forwardRef(
...
+                  validation.info = { ...infos, ...nextInfos };
+                  onValidate(validation);
```
ericsoderberghp, line 243: It seems a bit inconsistent to use the `event` for `onSubmit` but not for `onBlur`.

src/js/components/Form/doc.js
```diff
@@ -45,6 +45,11 @@ export const doc = Form => {
       `Function that will be called when the form is reset. The
       single argument is the event provided by react.`,
     ),
+    onValidate: PropTypes.func.description(
+      `Function that will be called when the form is validated. The
+      single argument is an event containing the latest error object
+      via \`event.error\` and info object via \`event.info\`.`,
```
ericsoderberghp, line 51: I think these should be `event.errors` and `event.infos` to align with the property names.

taysea: I think that's a good idea

src/js/components/Form/Form.js
taysea: I would agree. Updating `onBlur` to use the event as well.

src/js/components/Form/Form.js
taysea: This is a nice idea. Let me try that

ShimiSun:
Can you take a look at the Chormatic file change and confirm it is expected https://www.chromatic.com/snapshot?appId=5d9774839a6eff00203f5cbf&id=5f33215e912024002272fa83?

taysea:
> Can you take a look at the Chormatic file change and confirm it is expected https://www.chromatic.com/snapshot?appId=5d9774839a6eff00203f5cbf&id=5f33215e912024002272fa83?

Yes, this is expected. I had added one formfield into that story to test an error message, so the visual change is due to a field being added to the form.

ShimiSun:
Were you able to test this solution in the long-form context on the design-system? I'm curious whether it answered all the expected requirements.

taysea:
> Were you able to test this solution in the long-form context on the design-system? I'm curious whether it answered all the expected requirements.

I have not implemented it, but I can try it out in this branch locally. @britt6612 had two good proposed approaches, so let me test them out here.

src/js/components/Form/Form.js
```diff
@@ -56,21 +59,29 @@ const Form = forwardRef(
...
+      () => setValidationResults({ errors: errorsProp, infos: infosProp }),
+      [errorsProp, infosProp],
+    );
     const validations = useRef({});
 
     // clear any errors when value changes
     useEffect(() => {
-      setErrors(prevErrors => {
+      setValidationResults(prevValidationResults => {
         const [nextErrors] = validate(
```
ericsoderberghp, line 75: Should this be `[nextErrors, nextInfos]` and then use `nextInfos` at line 83? I would think the `validate()` could return some fresh infos.

taysea: I think you're right -- we didn't have that prior but I would also think that there could be some new infos coming in after validation

PR #4397, opened 2020-08-10
Adds test for items with mixed content to Infinite Scroll

JoeButler7:
#### What does this PR do?
Adds a test to `src/js/coponents/InfiniteScroll/__tests__/InfiniteScroll-test.js` where the `items` array has mixed types of content
 `src/js/coponents/InfiniteScroll/__tests__/InfiniteScroll-test.js`
ran `yarn test` to ensure that the tests still pass
see above
No
No
Backwards Compatible


ShimiSun:
Adding @IanKBovard & @halocline for review.
This PR should resolve the conflicts to enable the merge.

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
```diff
@@ -81,6 +80,34 @@ describe('InfiniteScroll', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
+  test('mixed items', () => {
+    const tempItems = [
```
halocline, line 84: This is a good start for the types of varied content we want to be able to support.

As it currently stands, we have 5 items here which would most likely fit on a single results page. With InfiniteScroll, users are presumably reaching for the component because they are likely presenting many pages of results.

How can we programmatically take what is here and populate a larger array which would span multiple pages of results? Loop through, splicing string lengths, vary numbers by iteration, vary strings using template literals and iterations, ... ?

Goal is to generate an array that spans multiple pages of results, as well as have enough variation so that resulting pageheights will vary while the items.length per page is consistent.

Let me know if a quick call would help to clarify, or if this is enough direction on which to iterate.

JoeButler7: Made some changes in the most recent push. Let me know what you think

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
```diff
@@ -81,26 +81,59 @@ describe('InfiniteScroll', () => {
...
+    const lorem = `Lorem ipsum dolor sit amet, consectetur adipisicing elit,
+    sed do eiusmod temporincididunt ut labore et dolore magna
+    aliqua. Ut enim ad minim veniam, quis nostrud exercitation
+    ullamco laboris nisi ut aliquip ex ea commodo consequat.
+    Duis aute irure dolor in reprehenderit in voluptate velit
+    esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
+    occaecat cupidatat non proident, sunt in culpa qui officia
+    deserunt mollit anim id est laborum.`;
+    const mixedItems = [];
+    for (let i = 0; i < 1000; i += 1) {
```
halocline, line 93: Probably can nudge this down a bit.
```suggestion
    for (let i = 0; i < 200; i += 1) {
```

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
```diff
@@ -81,26 +81,59 @@ describe('InfiniteScroll', () => {
...
+              mixedItems.push(lorem);
+              break;
+            case 1:
+              mixedItems.push(lorem.slice(140));
+              break;
+            case 2:
+              mixedItems.push(lorem + lorem);
+              break;
+            case 3:
+              mixedItems.push(lorem.slice(Math.random()));
```
halocline: If we are relying on a snapshot, how will this `Math.random()` affect the snapshot validation?

I don't believe this Javascript method allows us to set a seed in random, does it?
I was thinking about something like `lorem.slice(i, Math.min(i * 3, lorem.length))`
might be an option?

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
halocline, line 93: A comment explaining the purpose here might be helpful. e.g. "Generating an array to simulate varying data elements to generate differing page heights" or something like that.

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
```diff
@@ -81,26 +81,59 @@ describe('InfiniteScroll', () => {
   });
 
   test('mixed items', () => {
```
halocline, line 83: Trying to move tests to explicitly state what the purpose of the test is by supplying in the format:
"should [expectation] when [scenario]".

[Related discussion here](https://github.com/grommet/grommet/pull/4390#discussion_r468112613)

```suggestion
  test('should render expected items when supplied assortment of mixed items', () => {
```

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
halocline, line 93: I think this will help out quite a bit. Thank you for adding!

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
halocline: Nice. Thank you. Added a few more refining comments, then should be pretty close.

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
JoeButler7: Yeah I realized that and have changed this

src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
```diff
@@ -81,6 +80,69 @@ describe('InfiniteScroll', () => {
...
+              mixedItems.push(lorem);
+              break;
+            case 1:
+              mixedItems.push(lorem.slice(140));
+              break;
+            case 2:
+              mixedItems.push(lorem + lorem);
+              break;
+            case 3:
+              mixedItems.push(lorem.slice(i, Math.min(i * 3, lorem.length)));
```
halocline, line 123: Looks like this worked out well!

Issue #4093, opened 2020-05-17
Text doesnt apply vertical margin as shown on documentation


jjalonso:
Apparently Vertical margin doesnt work on Text component.

### Expected Behavior

expect to see vertical margin

### Actual Behavior

Only work with horizontal margin

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/text-margin-12ois?file=/index.js

ShimiSun:
Hi @jjalonso, thanks for reporting, a quick solution for the sandbox you described is to have a joint parent container for the two Text components https://codesandbox.io/s/text-margin-2siqs?file=/index.js.

jjalonso:
Yeah it work but I though it was a bug since the doc talk about you can apply vertical margin to text.

Is that a bug or only I'm using it wrong?
 cc @ShimiSun

ShimiSun:
we'll need to see why doesn't work without the Box parent. I'll keep this issue open for now.

taysea:
The reason this is happening is because Text is rendered as `span` in the DOM and `span` by default has display inline which only takes up horizontal margins. That is why the vertical margin is having no effect. However, once Text is wrapped in a Box, it has display block, allowing the vertical margin to take effect. 

Applying `display: inline-block;` to StyledText seems to remedy this issue, but was there reasoning prior to leave Text as inline display by default?

jjalonso:
In case it will not accept vertical margin.
Doc should be changed.

ShimiSun:
An idea is to improve the documentation that the vertical `fill` will be applied when using the component within a Layout component such as Box.

jjalonso:
It definitely need to be annotated on the margin section in the doc.

PR #4398, opened 2020-08-10
Add clarification to text margin docs

taysea:
#### What does this PR do?

Clarifies docs for Text margin by explaining that Text needs to be wrapped in a Box for margin to take effect. Uses the `genericStyles` as a base for wording, then adds clarification sentence.

src/js/components/Text/doc.js

Updated snapshots.

Reasoning can be found here: https://github.com/grommet/grommet/issues/4093#issuecomment-631759230

Closes https://github.com/grommet/grommet/issues/4093

Updated in this PR.

No.

Yes, backwards compatible.

### DIFF
--- a/src/js/components/Text/doc.js
+++ b/src/js/components/Text/doc.js
@@ -60,10 +60,12 @@ export const doc = Text => {
         ]),
       }),
       PropTypes.string,
-    ]).description(`The amount of margin around the component. For margin to be 
-      applied, Text needs to be wrapped in a containing Box. An object can be 
-      specified to distinguish horizontal margin, vertical margin, and margin 
-      on a particular side.`),
+    ]).description(`The amount of margin around the component. An object can be 
+    specified to distinguish horizontal margin, vertical margin, and margin on 
+    a particular side. For vertical margin to be applied, Text needs to be 
+    contained within a layout component (such as Box or a generic div) or 
+    behave as a div (by applying as="div" or a display style of 
+    inline-block).`),
     size: PropTypes.oneOfType([
       PropTypes.oneOf([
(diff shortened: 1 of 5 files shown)


src/js/components/Text/doc.js
```diff
@@ -22,6 +23,47 @@ export const doc = Text => {
...
+          PropTypes.string,
+        ]),
+        vertical: PropTypes.oneOfType([
+          PropTypes.oneOf(MARGIN_SIZES),
+          PropTypes.string,
+        ]),
+      }),
+      PropTypes.string,
+    ]).description(`The amount of margin around the component. For margin to be 
+      applied, Text needs to be wrapped in a containing Box. An object can be
```
ShimiSun, line 64: Do you think that explicitly specifying Box is a clear way to communicate it? I'm thinking it can be a div, or any other layout component.

ericsoderberghp, line 64: Actually, what is driving this is that Text ends up as a `<span />` with `display: inline`. I think this means that horizontal margin will work but vertical margin will be ignored. So, it would be more accurate to include that detail.

ericsoderberghp: I think another way to get the margin to take effect is to do something like `as="div"`.

taysea: Yes, you're right. Let me specify the wording. This is the conclusion I found when first investigating: https://github.com/grommet/grommet/issues/4093#issuecomment-631759230

taysea: You're right, let me clarify the wording here so it is less restricted to "Box"

PR #4399, opened 2020-08-10
InfiniteScroll Test `show` item should be visible

jcfilben:
#### What does this PR do?
Adds a test to infinite scroll that checks if the `show` item is being rendered when show > step

Currently this test is passing, but once the show prop is fixed lines 96 and 97 need to be deleted and lines 89, 98, and 99 uncommented for the test to work correctly.

#4334 #4369 

No

No

Backwards Compatible

### DIFF
--- a/src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
+++ b/src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
@@ -4,10 +4,15 @@ import 'jest-styled-components';
 
 import { Grommet } from '../../Grommet';
 import { InfiniteScroll } from '..';
+import { Box } from '../..';
 
 describe('InfiniteScroll', () => {
   const items = [];
   while (items.length < 4) items.push(items.length);
+  const simpleItems = value =>
+    Array(value)
+      .fill()
+      .map((_, i) => `item ${i + 1}`);
 
   test('basic', () => {
     const { container } = render(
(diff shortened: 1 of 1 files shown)


src/js/components/InfiniteScroll/__tests__/InfiniteScroll-test.js
```diff
@@ -75,4 +80,22 @@ describe('InfiniteScroll', () => {
...
+        >
+          {item => <Box key={item}>{item}</Box>}
+        </InfiniteScroll>
+      </Grommet>,
+    );
+    // item(104) = 'item 105' because indexing starts at 0
+    const renderedItems = container.firstChild.children.item(10).outerHTML;
+    expect(renderedItems).toContain('item 11');
+    // const renderedItems = container.firstChild.children.item(104).outerHTML;
+    // expect(renderedItems).toContain('item 105');
```
halocline, line 99: You have this "Currently this test is passing, but once the show prop is fixed lines 96 and 97 need to be deleted and lines 89, 98, and 99 uncommented for the test to work correctly." called out in the PR. Might be helpful to document inline as code comments as well so we don't forget.

```suggestion
          /* Uncomment the next 2 lines once fix for show is in place */
          // show={105}
        >
          {item => <Box key={item}>{item}</Box>}
        </InfiniteScroll>
      </Grommet>,
    );
    // item(104) = 'item 105' because indexing starts at 0.
    /* Delete the next 2 lines once fix for show is in place */
    const renderedItems = container.firstChild.children.item(10).outerHTML;
    expect(renderedItems).toContain('item 11');
    /* Uncomment the next 2 lines once fix for show is in place */
    // const renderedItems = container.firstChild.children.item(104).outerHTML;
    // expect(renderedItems).toContain('item 105');
```

PR #4400, opened 2020-08-10
Add listdocs issue4356

### LINKED ISSUE #4356: List - missing action prop on docs and filling 'data' on the stories
### Expected Behavior
`action` prop of List should be documented.
The `data` on the story examples should appear in the story example and not imported.

https://v2.grommet.io/list

`data` in line 8 should be explicit:
https://storybook.grommet.io/?path=/story/list--action 

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: stable
- Operating System and version (desktop or mobile): N/A

britt6612:
#### What does this PR do?
This PR adds `action` to props for `list` in documentation 
This PR also takes out imports of `{data}` and uses the actual code for users to see
list/docs.js
storybook
storybook
issue #4356
closes #4356 
auto
no
backwards compatible


britt6612:
action can take any `jsx` and return it. It doesn't have to be necessary be an on-click. 
@ShimiSun do you have any wording in mind for the description ? 
You can pass in 

```
        action={(item, index) => {
          return (
            <Text>hello</Text>
          );
        }}
```

for instance. As well as the menu button that is in the storybook example.

ericsoderberghp:
`action` isn't really the right name for this property. Alas.

britt6612:
> `action` isn't really the right name for this property. Alas.

Those were my thoughts as well that name isnt the best

src/js/components/List/stories/action.js
```diff
@@ -5,7 +5,25 @@ import { Grommet, Box, List, Menu } from 'grommet';
...
+
+export const data = [];
+
+for (let i = 0; i < 40; i += 1) {
+  data.push({
+    entry: `entry-${i + 1}`,
+    location: locations[i % locations.length],
+    date: `2018-07-${(i % 30) + 1}`,
+    percent: (i % 11) * 10,
+    paid: ((i + 1) * 17) % 1000,
```
ShimiSun, line 24: I'm not sure the story uses all these fields, can you please clean up the fields that are not relevant or doesn't impact the story? Same for the other stories.

src/js/components/List/doc.js
```diff
@@ -47,6 +47,10 @@ export const doc = List => {
 
   DocumentedList.propTypes = {
     ...genericProps,
+    action: PropTypes.func.description(
+      `Action will be something for the user to click on and interact
+       with within the list component`,
```
ShimiSun, line 52: Needs to be a little more clear as @IanKBovard suggested.
what about a different opening, "The interactive element property for the user to..."? thoughts?

IanKBovard:
> `action` isn't really the right name for this property. Alas.

I agree with this. Action doesn't describe what this prop does, as it can be used to render any component. It behaves like the render function in DataTable's columns prop. 

`render: (datum) => <Content />`
https://v2.grommet.io/datatable#columns

I think renaming this to render, or something along those lines would be more descriptive, and influence the documentation.

src/js/components/List/stories/onClickItem.js
```diff
@@ -4,7 +4,25 @@ import { storiesOf } from '@storybook/react';
...
+
+const data = [];
+
+for (let i = 0; i < 40; i += 1) {
+  data.push({
+    entry: `entry-${i + 1}`,
+    location: locations[i % locations.length],
+    date: `2018-07-${(i % 30) + 1}`,
+    percent: (i % 11) * 10,
+    paid: ((i + 1) * 17) % 1000,
```
ShimiSun, line 23: Does the onClickItem example require all these fields? It doesn't look like it's using any other field then `location` and unused fields should be removed.
A similar clean up is required for the other List stories that use `data`.

src/js/components/List/doc.js
britt6612: I wonder if we should change the Action prop name like suggested above then..

Even from looking at the code I was not clear on the `action` prop

src/js/components/List/stories/onClickItem.js
britt6612: `onclick` example prints out all of these fields so unless you do not want that behavior anymore ? 

Screen Shot 2020-08-12 at 9 51 04 AM

ShimiSun: Correct, thank you for drawing my attention to the print.

src/js/components/List/doc.js
```diff
@@ -47,6 +47,10 @@ export const doc = List => {
 
   DocumentedList.propTypes = {
     ...genericProps,
+    action: PropTypes.func.description(
+      `The interactive element property for the user to be able to pass a 
+      component within list.`,
```
ShimiSun, line 52: I think we should also add a few words about the arguments that action is receiving and what they mean ` action={(item, index) => {...}}`, similarly to how you documented it for Menu children few PRs ago.

britt6612: okay yeah that sounds good what were your thought about changing name of `action` ?

ShimiSun: I think renaming the `action` is a valid concern, and I also think it should go on a separate PR since it will require backward compatibility handling and I wouldn't want it to block this current PR.

britt6612: good point! I will fix the definition to finish this PR

src/js/components/List/doc.js
```diff
@@ -47,6 +47,13 @@ export const doc = List => {
 
   DocumentedList.propTypes = {
     ...genericProps,
+    action: PropTypes.func.description(
+      `The interactive element property for the user to be able to pass a 
+      component within list. It should be passed with an item and index of an
+      array and return a react element.
+      \`action = ({item, index}) => <Content />\`
+    />`,
```
ShimiSun, line 55: Thinking about this description some more, and the element doesn't have to be interactive, it could just be a component rendered on the cell. So maybe a different phrasing such as:
"Accepts a function that allows for a custom rendering of a component, it should be passed with an item and index of an array and return a react element."
Thoughts? Do you think it will be more accurate?

britt6612: yes it doesnt have to be interactive it can be a <box> for example or something else yes I think that sounds better

ShimiSun: Thank you for your patience and for brainstorming it through together...

britt6612: no problem it helped me better understand when I first looked I was also unsure about the action prop

PR #4401, opened 2020-08-11
Update type definitions for Form payloads

nozzle-a:
#### What does this PR do?
Slight tweak to type definitions around form payloads.

This change makes it so that we can strongly type the expected payload in the `onSubmit` event, defaulting to `Record<string, unknown>` if the user doesn't supply a type definition for the form payload.  Using `unknown` will force proper validation when the payload is consumed (as opposed to `any` which won't, and which the standard TS linter disallows).

This PR also alters the `onSubmit` definition to pass that type through, so that the following works as expected.

```
interface MyFormData {
  name: string;
  age: number;
}

function onSubmit({ value }: FormExtendedEvent<MyFormData>) {
  // ...
}

const MyForm = Form as TypedForm<MyFormData>;

<MyForm onSubmit={onSubmit}> 
  // ...
</MyForm>
```

Working on my local project working off my fork.

Create a TypedForm using a defined form payload generic.


src/js/components/Form/index.d.ts
```diff
@@ -10,13 +10,13 @@ export interface FormProps<T> {
   infos?: {};
   messages?: {invalid?: string,required?: string};
   onChange?: (value: T) => void;
-  onSubmit?: ((event: FormExtendedEvent) => void);
+  onSubmit?: ((event: FormExtendedEvent<T>) => void);
```
carolineartz, line 13: this looks good to me. It's maybe a little weird since the type param naming here T is referring to R above...but maybe it isn't that weird and its nbd either way. 

Good addition!

ShimiSun:
Thank you for your help reviewing @carolineartz & @JoeButler7 ❤️

PR #4402, opened 2020-08-11
Heading responsiveData logic updated

### LINKED ISSUE #4332: Shrink width of Heading level 6 causes it to jump up in size
See https://grommet.slack.com/archives/C09QQEG69/p1595361593434100?thread_ts=1595361593.434100&cid=C09QQEG69

For h6 elements, shrinking page width should not cause the font size to increase.

Thanks to Taylor Seamans for this explanation... (paraphrased)

...this is likely a boundary condition.  As the pages shrinks and crosses breakpoints, the theme's heading sizes are reduced.  Since "6" is the smallest, it can't go any smaller and I noted finds some defa t size.

### Actual Behavior

Shrinking page width causes font size to increase.

### Steps to Reproduce

Start with component `<Heading level="6">Hello Grommet</Heading>`.  Now shrink the page below the breakpoint of 768 and the font size increases.

### Your Environment

- Grommet version: 2.14
- Operating System and version (desktop or mobile): Windows

IanKBovard:
#### What does this PR do?
changes
`const responsiveData = headingTheme.level[Math.min(props.level + 1, 4)][size]`
to
`const responsiveData = headingTheme.level[Math.min(
    props.level + 1, Math.max(...Object.keys(headingTheme.level)),)][size]`
StyledHeading.js
storybook
same as above
The current logic only supported heading levels up to 4.
closes #4332 
Kapture 2020-08-11 at 9 43 43
no
no
Yes, it does change the behavior of heading levels 5 and 6, since they sized up to level 4.


src/js/components/Heading/StyledHeading.js
```diff
@@ -24,8 +23,13 @@ const sizeStyle = props => {
         props.theme.global.breakpoints[headingTheme.responsiveBreakpoint];
       if (breakpoint) {
         const responsiveData =
-          headingTheme.level[Math.min(props.level + 1, 4)][size];
-        if(responsiveData) {
+          headingTheme.level[
+            Math.min(
+              props.level + 1,
+              Math.max(...Object.keys(headingTheme.level)),
```
ShimiSun, line 29: I think this will always be 6 since the base theme will always have the level 6 defined, is that we want?

IanKBovard: Yes, it will always be 6. Going off of @ericsoderberghp suggestion. I assume this will support any future heading levels you might create later.

ShimiSun: Maybe I'm going a little too far here, but we might want to make sure the users didn't override the theme with empty objects of 5,6, just so they won't be considered.
Should we reach inside the level to see if the 'medium' size (default) or the required size is even there?

ericsoderberghp: It would be friendlier to be defensive. Use the next level down, until we get to the last level we have, which could be different for different themes.

IanKBovard: The way the code currently is
```
Math.min(props.level + 1, Math.max(...Object.keys(headingTheme.level))
```
will return the lowest level value(largest number) that isn't an empty object.

ShimiSun: Can you please add a test that proves and enforces it?

ShimiSun: Interesting, you can assign object to undefined instead of undefined.

IanKBovard: im not sure what you mean, could you elaborate?

IanKBovard: passing in undefined in the level value will cause an error. Although, I'm not sure what we should do about this. To me, this seems like a problem with user error. I could go to the next lowest level in heading that isn't empty, but if undefined is passed through any key in level(like font, or size), I will also receive an error. 

Everything that causes an error:
```
'6': undefined,
```

```
'6': {
  font: undefined,
  small: {
    size: '140px',
    height: '20px',
    maxWidth: '336px',
  },
  medium: {
    size: '140px',
    height: '20px',
    maxWidth: '336px',
  },
  large: {
    size: '140px'
    height: '20px',
    maxWidth: '336px',
  },
  xlarge: {
    size: '140px',
    height: '20px',
    maxWidth: '336px',
  },
},
```

```
'6': {
  font: {},
  small: {
    size: '140px',
    height: '20px',
    maxWidth: '336px',
  },
  medium: undefined ,
  large: {
    size: '140px'
    height: '20px',
    maxWidth: '336px',
  },
  xlarge: {
    size: '140px',
    height: '20px',
    maxWidth: '336px',
  },
},
```
doesn't error - uses default theme:
```
'6': {},
```

I could be wrong, but I don't think there's a way to enforce the correct theme shape in this instance(there are a lot of ways that i can make this fail). There's a lot that a user can do wrong, and I think this would be better addressed with complete and clear documentation.

ShimiSun: @halocline can you please advise Ian on how to apply a test or a fix for this issue and move this PR forward.

halocline: @IanKBovard What do you think about an approach where we probe for the next level down, if it is defined, use it; if undefined, use current level?

Could look something like: 
```
const responsiveData = headingTheme.level[props.level + 1]
          ? headingTheme.level[props.level + 1][size]
          : headingTheme.level[props.level][size];
```

From a test standpoint, I think we could add Jest tests for theme where we pass in each of your scenarios above, perhaps at varying lower levels as well.

Regarding the font definitions, I think we may need to place an additional check for heading.level existing at where the fontFamily is being defined on line 53.

IanKBovard:
Added matts suggestion. font defaults to an empty object if none is available. Added a test to check if a warning is thrown when heading.level is undefined.

src/js/components/Heading/__tests__/Heading-test.js
```diff
@@ -205,3 +205,26 @@ test('Theme based font weight renders', () => {
   const tree = component.toJSON();
   expect(tree).toMatchSnapshot();
 });
+
+test('Throws a warning when heading.level is undefinedin the theme.', () => {
```
ShimiSun, line 209:
```suggestion
test('Throws a warning when heading.level is undefined in the theme.', () => {
```

src/js/components/Accordion/__tests__/__snapshots__/Accordion-test.js.snap
```diff
@@ -220,9 +220,9 @@ exports[`Accordion AccordionPanel 1`] = `
 
 @media only screen and (max-width:768px) {
   .c6 {
-    font-size: 18px;
-    line-height: 24px;
-    max-width: 432px;
+    font-size: 16px;
+    line-height: 22px;
+    max-width: 384px;
```
ShimiSun, line 225: Can you add a few words from your analysis why the snapshots had changed and whether that's a backward compatibility concern?

src/js/components/Heading/__tests__/Heading-test.js
IanKBovard: fixed the typo

src/js/components/Accordion/__tests__/__snapshots__/Accordion-test.js.snap
IanKBovard: This is a heading level 4 that resized to a level 5. Before this pr heading level 4 never resized. I don't see this as a backwards compatibility issue, as the heading not resizing above level 4 was a bug.

ShimiSun: I agree, thanks for the inputs.

PR #4406, opened 2020-08-11
Allow plain button to receive active styling

taysea:
#### What does this PR do?

Allows plain buttons (that exist in applications where the theme has button.default defined) to receive activeStyle. Currently, there are two paths that determine the styling of a Button. If a theme has `button.default` defined, the styling comes from `StyledButtonKind`. If not, the styling comes from `StyledButton`. Plain buttons in the default button theme were not receiving the active styling that plain buttons in the existing button styling could.

Now, if a button is plain and active, it will receive `activeStyle`. This remedies issues being seen on grommet-theme-hpe regarding DateInput where the active day was not showing an active background on mouse over or keyboard navigation.

Set up grommet-theme-hpe NEXT in local storybook and applied to DateInput.

No.

No.

This will introduce an active background to plain buttons (when they are active) in themes that define button.default. This is a visual change, but it is desired since this kind of behavior already exists on buttons that are relying on StyledButton.js


src/js/components/Button/StyledButtonKind.js
```diff
@@ -1,6 +1,7 @@
 import styled, { css } from 'styled-components';
 
 import {
+  activeStyle,
```
britt6612, line 4: We found that ActiveStyle was not pulled over from StyledButton.js so it was not getting the same behavior.

src/js/components/Button/StyledButtonKind.js
```diff
@@ -246,9 +247,13 @@ const StyledButtonKind = styled.button`
...
-  text-align: ${props.align};
-  `}
-  ${props => props.hoverIndicator && hoverIndicatorStyle(props)}
+    text-align: ${props.align};
+    `}
+    ${props => props.hoverIndicator && hoverIndicatorStyle(props)}
+  // plain buttons should still be able to receive active styling even
+  // though they do not have a kindStyle.
+  // other button types are receiving active styling via their kindStyle.
+  ${props => !props.disabled && props.active && props.plain && activeStyle}
```
ShimiSun, line 256: Is the regular button that isn't using `button.default` applying active styling on `plain`?

src/js/components/Button/Button.js
```diff
@@ -253,6 +253,7 @@ const Button = forwardRef(
           {...rest}
           as={domTag}
           ref={ref}
+          active={active}
```
britt6612, line 256: can we just leave this as active prop ? without passing active ?

src/js/components/Button/StyledButtonKind.js
britt6612: no its not using plain 
I think we can change to 
` ${props => !props.disabled && props.active && activeStyle}`

this still worked for me without the plain

src/js/components/Button/Button.js
britt6612: otherwise some examples are still showing active without being moused over on calendar

src/js/components/Button/StyledButtonKind.js
taysea: Yes, the regular button does apply active styling on `plain`: https://github.com/grommet/grommet/blob/master/src/js/components/Button/StyledButton.js#L203
```
${props => !props.disabled && props.active && activeButtonStyle(props)}
```

ShimiSun:
What was the reason behind the recent commit 'Move placement of active style'?

src/js/components/Button/Button.js
britt6612: opps I put this change on line 96 when testing and didnt realize it was under  `<StyledButtonKind/>`
so you can ignore this comment

taysea:
> What was the reason behind the recent commit 'Move placement of active style'?

I just updated the comment to hopefully clarify more, but essentially this will set the activeStyle on all button kinds by default (which will include plain) -- similar to how `StyledButton` does. Then, `kindStyle` follows it in the code so specific button kinds (such as default, primary, etc) will be able to override the generic activeStyle with any of their specific styles.

Issue #4385, opened 2020-08-07
Storybook/Chromatic clean-ups


ShimiSun:
1. Clean up the console warning:
![image](https://user-images.githubusercontent.com/6320236/89604045-ae6e2680-d827-11ea-8f4d-75709bc21612.png)

1. Disable all Chromatic testing for DropButton component.

1. Disable Chromatic on DateInput -> Range and Simple stories. 

1. Remove Chromatic from all these 4 stories of Layer 
![image](https://user-images.githubusercontent.com/6320236/89604593-d27e3780-d828-11ea-9eab-1df36540fc9c.png)

1. Add Carousel story example for the use case of #3850.

1. Add the Grommet Tools Header to Storybook Header component.

ShimiSun:
More relevant content https://www.chromatic.com/docs/ischromatic

jcfilben:
The chromatic snapshots for
Menu: reverse, simple, and themed
seem like they could be removed. Also not sure if we need chromatic testing on the Collapsible stories.

ShimiSun:
Example code for Layer-> Form, should use Form and not just FomFields inside of Box.https://storybook.grommet.io/?path=/story/layer--form

ShimiSun:
Can we also exclude Chart -> Scan & Zoom stories from Chormatic.

jcfilben:
Removed: 19 chromatic snapshots
Added: 2 new stories added to carousel and header

Total difference is chromatic snapshots: -17

ShimiSun:
Fantastic work team, I'll be closing this issue!!

PR #4408, opened 2020-08-12
Adds Multiple view to Carousel Story

JoeButler7:
#### What does this PR do?
Adds storybook example to Carousel where multiple images are in view at once
`src/js/components/Carousel/stories/Multiple.js`
Local storybook
Storybook
#4385
Screen Shot 2020-08-12 at 12 20 06 PM

No
No
Backwards compatible


src/js/components/Carousel/stories/Multiple.js
```diff
@@ -0,0 +1,42 @@
...
+  // 3 Carousel Views with up to 3 images in each view
+  const getCarouselView = () => {
+    const carouselView = new Array(3)
+      .fill([])
+      .map((_, i) => data.slice(i * 3, i * 3 + 3));
+
+    // add images to each view
+    return carouselView.map(view => {
+      const getCarouselItem = () =>
+        view.map(img => <Image src={img} fit="contain" />);
```
ShimiSun, line 28: I don't think this function is needed, so line 28 can be embedded in line 30.

src/js/components/Carousel/stories/Multiple.js
```diff
@@ -0,0 +1,42 @@
...
+      // return view
+      return <Box direction="row">{getCarouselItem()}</Box>;
+    });
+  };
+  return (
+    <Grommet>
+      <Carousel controls="arrows" play={0}>
```
ShimiSun, line 35: why does `play={0}` needed?

JoeButler7: I originally had it so that the carousel would not auto transition but now I see that if you don't include the prop it won't auto transition. Removed in most recent push

src/js/components/Carousel/stories/Multiple.js
```diff
@@ -0,0 +1,44 @@
...
+          {view.map(img => (
+            <Image src={img} fit="contain" />
+          ))}
+        </Box>
+      );
+    });
+  };
+  return (
+    <Grommet>
+      <Carousel controls="arrows">{getCarouselView()}</Carousel>
```
ShimiSun, line 39: Shouldn't the function just be a functional component?
So something like:
```
<Carousel controls="arrows"><View /></Carousel>
```
Instead of calling a function?

JoeButler7: Not Sure if I was missing something but when I tried to use it as a functional component I wasn't able to get the carousel to actually work with just 3 images, instead all the images were rendered 
Screen Shot 2020-08-12 at 2 17 47 PM

src/js/components/Carousel/stories/Multiple.js
```diff
@@ -0,0 +1,44 @@
...
+  // 3 Carousel Views with up to 3 images in each view
+  const getCarouselView = () => {
+    const carouselView = new Array(3)
+      .fill([])
+      .map((_, i) => data.slice(i * 3, i * 3 + 3));
+
+    // add images to each view
+    return carouselView.map(view => {
+      // return each view with three images
+      return (
```
ShimiSun, line 28: These two nested returns look like we can keep just one of them and clean up the example.

IanKBovard:
This is just me, but I find this example hard to read. There is some clever abstraction here, but I think it comes at the cost of readability; which, with educational tools such as this one, is a priority.

This isn't a perfect example, but I think the code here doesn't get in the way of the example.
```
const data = [
  '//v2.grommet.io/assets/Wilderpeople_Ricky.jpg',
  '//v2.grommet.io/assets/IMG_4245.jpg',
  '//v2.grommet.io/assets/IMG_4210.jpg',
  'https://avatars1.githubusercontent.com/u/14203820?s=280&v=4',
  'https://developer.hpe.com/img/hpe-dev-grommet-gremlin-rockin-static.svg',
  '//v2.grommet.io/assets/Wilderpeople_Ricky.jpg',
  '//v2.grommet.io/assets/IMG_4245.jpg',
  '//v2.grommet.io/assets/IMG_4210.jpg',
  '//v2.grommet.io/assets/Wilderpeople_Ricky.jpg',
];

const Page1 = () => {
  const images = data.slice(0, 3);
  return (
    <Box direction="row">
      {images.map(img => (
        <Image src={img} fit="contain" key={img} />
      ))}
    </Box>
  );
};

const Page2 = () => {
  const images = data.slice(3, 6);
  return (
    <Box direction="row">
      {images.map(img => (
        <Image src={img} fit="contain" key={img} />
      ))}
    </Box>
  );
};

const Page3 = () => {
  const images = data.slice(6, 9);
  return (
    <Box direction="row">
      {images.map(img => (
        <Image src={img} fit="contain" key={img} />
      ))}
    </Box>
  );
};

const CarouselMulti = () => (
  <Grommet>
    <Carousel>
      <Page1 />
      <Page2 />
      <Page3 />
    </Carousel>
  </Grommet>
);
```

JoeButler7:
### Update:
The latest push refactors the code with @IanKBovard suggestions in mind. I agree that the code will be more readable this way and would be a better example for people looking on storybook

PR #4409, opened 2020-08-12
Removed Chart (Scan and Zoom) stories from Chromatic

jcfilben:
#### What does this PR do?
Removed Chart -> scan and zoom stories from chromatic testing

Number of chromatic snapshots should go down by 2

#4385 

No

No

Backwards compatible

### DIFF
--- a/src/js/components/Chart/stories/Scan.js
+++ b/src/js/components/Chart/stories/Scan.js
@@ -122,6 +122,8 @@ const ScanChart = props => {
   );
 };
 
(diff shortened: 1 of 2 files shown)


ericsoderberghp:
Darn, I was thinking we should change these to have fixed dates rather than removing them from Chromatic.

ShimiSun:
@ericsoderberghp this change was intentional. I can bring you up speed with the mindset I had for the reasons behind removing it offline.

PR #4416, opened 2020-08-13
change value is set in dateInput

britt6612:
#### What does this PR do?
This PR takes out a statement that only sets value if it is valid 
DateInput.js
storybook 

I wanted to be able to access the value in `onChange`
the value was only being passed if it was valid
so If the field was empty I was not able to set an error 

```

  const onChange = event => {
    setValue(nextValue);
    if (!nextValue) {
      console.log('no value error message ');
    } else {
      console.log('value no error message');
    }
  };


```
above code 
I want to be able to access the event.value even if it is undefined 
design-system 
showing an example of validation 
no
no
backwards compatible

### DIFF
--- a/src/js/components/DateInput/DateInput.js
+++ b/src/js/components/DateInput/DateInput.js
@@ -133,15 +133,13 @@ const DateInput = forwardRef(
               const nextTextValue = event.target.value;
               setTextValue(nextTextValue);
               const nextValue = textToValue(nextTextValue, schema);
-              if (nextValue) {
-                // valid value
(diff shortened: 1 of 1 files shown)


britt6612:
Im putting this PR more of a discussion as well to understand the reasoning behind only passing value if it is valid.

britt6612:
> LGTM, I'm wondering if there was the intentional reason for the condition that got removed. @ericsoderberghp ?

Also was wondering the same!

ShimiSun:
Thank you @britt6612 !

PR #4417, opened 2020-08-13
TextInput onSelect and onSuggestionSelect handlers

### LINKED ISSUE #3118: TextInput makes it difficult to listen for select events on input element

IanKBovard:
#### What does this PR do?
Allows the use of default onSelect event for textInput.

TextInput.js

Within storybook using this.

```javascript
import React from 'react';

import { Box, Grommet, TextInput } from 'grommet';

const suggestions = Array(100)
  .fill()
  .map((_, i) => `suggestion ${i + 1}`);

const App = () => {

  const onChange = event => setValue(event.target.value);

  const onSelect = event => setValue(event.suggestion);
  const onHightlight = event =>
    console.log(
      event.target.value.substring(
        event.target.selectionStart,
        event.target.selectionEnd,
      ),
    );
  return (
        <Box width="medium">
          <TextInput
            value={value}
            onChange={onChange}
           // i tested with a combination of onSelect and onSuggestionSelect being present and 
           absent
            onSelect={onHightlight}
            onSuggestionSelect={onSelect}
            suggestions={suggestions}
          />
        </Box>
      </Box>
    </Grommet>
  );
};

export default App;
```
Same as above. 

Grommets onSelect was overriding the html onSelect.

closes #3118 

Yes

Yes

Backwards compatible


IanKBovard:
this issue has not been addressed yet

ericsoderberghp on May 26 Member
The way these are being used seems like they should be using useMemo instead of useRef + useEffect.

IanKBovard 17 days ago Author Member
I'm a little confused about this. My understanding of useMemo is to optimize the performance of complex functions. How can I use it to replace useRef and useEffect?

src/js/components/TextInput/TextInput.js
```diff
@@ -119,6 +119,26 @@ const TextInput = forwardRef(
     const [focus, setFocus] = useState();
     const [showDrop, setShowDrop] = useState();
 
+    const handleTextSelect = useRef();
+    const handleSuggestionSelect = useRef();
+
+    // assigns select handlers
+    useEffect(() => {
```
ericsoderberghp, line 126: Would memo hooks be simpler than ref hooks?
```
const handleSuggestionSelect = useMemo(onSelect && !onSuggestionSelect ? onSelect : onSuggestionSelect, [onSelect, onSuggestionsSelect]);
const handleTextSelect = useMemo(onSelect && onSuggestionSelect ? onSelect : undefined, [onSelect, onSuggestionsSelect]);
```
just wondering

src/js/components/TextInput/TextInput.js
```diff
@@ -308,12 +331,12 @@ const TextInput = forwardRef(
...
                             inputRef.current.focus();
                             closeDrop();
-                            if (onSelect) {
+                            if (handleSuggestionSelect.current) {
                               event.persist();
                               const adjustedEvent = event;
                               adjustedEvent.suggestion = suggestion;
-                              adjustedEvent.target = inputRef.current;
-                              onSelect(adjustedEvent);
+                              adjustedEvent.target = (ref || inputRef).current;
```
ericsoderberghp, line 338: Don't need `ref ||`, `useForwardedRef` already does the right thing

src/js/components/TextInput/TextInput.js
```diff
@@ -232,6 +252,9 @@ const TextInput = forwardRef(
       placeholder && typeof placeholder !== 'string' && !value;
 
     let drop;
+    const extraProps = {
```
ericsoderberghp, line 255: Why do we need this variable, since it's only used once, it would be cleaner to just have it inline: `onSelect={handleTextSelect.current}`

ericsoderberghp:
> this issue has not been addressed yet
> 
> ericsoderberghp on May 26 Member
> The way these are being used seems like they should be using useMemo instead of useRef + useEffect.
> 
> IanKBovard 17 days ago Author Member
> I'm a little confused about this. My understanding of useMemo is to optimize the performance of complex functions. How can I use it to replace useRef and useEffect?

`useMemo` provides a simple way to create variables derived from props and state without having to repeat the logic to create them if they are used in multiple places. It's also similar to `useCallback`. I suggest it as we don't want the logic inline, as that will create new callbacks on each render, and the `useRef` approach seems like it takes more code.

IanKBovard:
> `useMemo` provides a simple way to create variables derived from props and state without having to repeat the logic to create them if they are used in multiple places. It's also similar to `useCallback`. I suggest it as we don't want the logic inline, as that will create new callbacks on each render, and the `useRef` approach seems like it takes more code.

Thanks for that! Very clear, and easy to understand.

src/js/components/TextInput/TextInput.js
IanKBovard: changes made

src/js/components/TextInput/TextInput.js
IanKBovard: changes made

src/js/components/TextInput/TextInput.js
IanKBovard: I had that initially, but this was added to keep the tests from updating.

src/js/components/TextInput/TextInput.js
```diff
@@ -346,14 +359,14 @@ const TextInput = forwardRef(
...
-            if (activeSuggestionIndex >= 0 && onSelect) {
+            if (activeSuggestionIndex >= 0 && handleSuggestionSelect) {
               // prevent submitting forms when choosing a suggestion
               event.preventDefault();
               event.persist();
               const adjustedEvent = event;
               adjustedEvent.suggestion = suggestions[activeSuggestionIndex];
-              adjustedEvent.target = inputRef.current;
-              onSelect(adjustedEvent);
+              adjustedEvent.target = (ref || inputRef).current;
```
ericsoderberghp, line 368: This should just be `inputRef.current`.

IanKBovard: changes made

PR #3800, opened 2020-02-28
[TS] InifiniteScrollProps TS handler improve #3165

fredhase:
#### What does this PR do?
Fixes the types in src/js/components/InfiniteScroll/index.d.ts

#### Where should the reviewer start?
src/js/components/InfiniteScroll/

#### What testing has been done on this PR?
Tested using the test that already exists in the component and using 'yarn storybook'.

#### How should this be manually tested?
Testing InfiniteScroll

#### What are the relevant issues?
#3165 

#### Do the grommet docs need to be updated?
I believe not

#### Should this PR be mentioned in the release notes?
I believe not

#### Is this change backwards compatible or is it a breaking change?
I believe it's compatible

PR #4422, opened 2020-08-17
Update InfiniteScroll typescript

taysea:
#### What does this PR do?

This PR updates typescript types for InfiniteScroll based on work started in [this PR](https://github.com/grommet/grommet/pull/3800).

src/js/components/InfiniteScroll/index.d.ts

Implemented new TS story and tested locally with `items` of type array of object and React Element and number. such as:
```
const items = [
   { name: "John", age: 22 },
   { name: "Jenny", age: 23 },
]
```
Run in storybook and ensure no flagged type warnings.

https://github.com/grommet/grommet/issues/3165

Closes #3800 
No.

No.

Yes, backwards compatible.

### DIFF
--- a/src/js/components/InfiniteScroll/index.d.ts
+++ b/src/js/components/InfiniteScroll/index.d.ts
@@ -2,7 +2,7 @@ import * as React from 'react';
 
 export interface InfiniteScrollProps {
   children?: (...args: [string, number]) => React.ReactElement;
-  items?: (string | Record<string, any>)[];
+  items?: (string | number | React.ReactElement | Record<string, any>)[];
   onMore?: () => void;
   renderMarker?: (marker: React.ReactElement) => React.ReactElement;
   replace?: boolean;
--- a/src/js/components/InfiniteScroll/index.d.ts
+++ b/src/js/components/InfiniteScroll/index.d.ts
@@ -1,7 +1,6 @@
 import * as React from 'react';
 
 export interface InfiniteScrollProps {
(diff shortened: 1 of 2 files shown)


src/js/components/InfiniteScroll/index.d.ts
```diff
@@ -1,12 +1,12 @@
-import * as React from "react";
+import * as React from 'react';
 
 export interface InfiniteScrollProps {
-  children?: ((...args: any[]) => any);
-  items?: any[];
-  onMore?: ((...args: any[]) => any);
-  renderMarker?: ((...args: any[]) => any);
+  children?: (...args: [string, number]) => React.ReactElement;
```
ShimiSun, line 4: Is the number optional?

taysea: I'm actually going to change children back to any now. I think that `children` should technically match the types for `items` since infinite scroll. however, copying the types from line 5 into the children args cause an error for the `key` prop which was looking for a string or number. I think for now it is safest to leave the children arg as `any`

ShimiSun: Can we remove `children` and just fall back to the default children, or do we need it because of the function?

taysea: Good call. Let me look into that.

taysea: It seems we can remove it, this is behaving similar to how children is used in Menu or Button, and we don't define types there. Updating PR now.

PR #4423, opened 2020-08-17
Changed Chart to add `animate`.

ericsoderberghp:
#### What does this PR do?

Changed Chart to add an `animate` property.
This creates and an animation appropriate to the type of chart. Line draws, bar and area grow up, points fade in.

StyledChart.js

Added unit test and Animate story.

Animate story

automatic

yes

backwards compatible


ShimiSun:
I think we should exclude animated stories from chromatic, the snapshot isn't being captured on time to take a snapshot of the chart, hence it's not needed.
Here is how the snapshot is taken (without the Chart):
image

src/js/components/Chart/stories/Animate.js
```diff
@@ -0,0 +1,24 @@
...
+            {type}
+          </Heading>
+          <Chart type={type} values={values} animate />
+        </Box>
+      ))}
+    </Box>
+  </Grommet>
+);
+
+storiesOf('Chart', module).add('Animate', () => <Example />);
```
ShimiSun, line 24: I think we should exclude animated stories from chromatic, the snapshot isn't being captured on time and the Chart isn't included in the snapshot.
Here is how the snapshot is taken (without the Chart):
image

ericsoderberghp: fixed

Issue #4005, opened 2020-04-23
Proposal: Move TypeScript stories into component sub-folders


avin-kavish:
At the moment, all TypeScript stories are under a root level TypeScript folder. This is inconvenient because usually since the TypeScript stories are incomplete compared to JavaScript ones, users have to switch back and forth between two mega folder structures.

![image](https://user-images.githubusercontent.com/48435155/80121888-73aeee00-85aa-11ea-8372-ba4da4939b62.png)

I propose having a typescript sub-folder for each component, so that all stories related to a component are kept locally and easily accessed for reference. As follows,

![image](https://user-images.githubusercontent.com/48435155/80123167-259aea00-85ac-11ea-817a-155b1c85389d.png)

ShimiSun:
Thanks for reporting, we will be heading this direction and @negreirosleo will help us get there!

PR #4425, opened 2020-08-17
Typescript story organization (Button, Clock, Collapsible, DataTable)

jcfilben:
#### What does this PR do?
Removes the following typescript stories because they are duplicates:
Button -> Basic
Clock -> Analog
Collapsible -> Simple
DataTable -> Clickable, Styled

Removed the following `.js` stories because the typescript version included a custom theme. Also moved the ts stories into their component folder
Clock -> CustomAnalog, Digital

^ wasn't sure if these typescript stories needed a comment added above the theme or what to put there

storybook

storybook

#4005 

No

No

Backwards compatible


ShimiSun:
>^ wasn't sure if these typescript stories needed a comment added above the theme or what to put there

The comment needs to be added every time there is a specific typescript syntax, so it can appear more than once or not appear at all, depends on the stories.

src/js/components/Clock/stories/typescript/CustomAnalog.tsx
```diff
@@ -38,8 +37,4 @@ const CustomAnalog = () => (
   </Grommet>
 );
 
-if (!isChromatic()) {
-  storiesOf('TypeScript/Clock', module).add('Custom Analog', () => (
-    <CustomAnalog />
-  ));
-}
+storiesOf('Clock', module).add('Custom Analog', () => <CustomAnalog />);
```
ShimiSun, line 40: This should actually be disabled on Chormatic, since the time is changing every second and we'll always have chromatic changes to approve if this story is not disabled.

jcfilben: Disabled CustomAnalog and Digital stories on chromatic in the latest update

src/js/components/DataChart/XAxis.js
```diff
@@ -1,32 +1,30 @@
 import React, { forwardRef } from 'react';
 import { Box } from '../Box';
 
-const XAxis = forwardRef(
```
ShimiSun, line 4: Shouldn't have any file changes.

src/js/components/Button/stories/typescript/Basic.tsx
```diff
@@ -1,26 +0,0 @@
-import React from 'react';
-import { storiesOf } from '@storybook/react';
-import isChromatic from 'chromatic/isChromatic';
-
-import { grommet, Box, Button, Grommet } from 'grommet';
-
-const BasicButtons = props => (
```
ShimiSun, line 7: we should also:
1. Remove CustomColor.tsx.
2. Add ThemeType and comments to the story we're keeping

src/js/components/DataTable/stories/typescript/Clickable.tsx
```diff
@@ -1,156 +0,0 @@
...
-        step={10}
-        onClickRow={event => console.log(event.datum)}
-      />
-    </Box>
-  </Grommet>
-);
-
-if (!isChromatic()) {
-  storiesOf('TypeScript/DataTable', module).add('Clickable', () => (
-    <ClickableDataTable />
```
ShimiSun, line 154: Let's keep this story

src/js/components/Button/stories/typescript/Basic.tsx
jcfilben: Removed `CustomColor.tsx`. For Button `Custom.tsx` I realized that it is different from `Custom.js` so I kept both and renamed the tsx story to TS-Custom and added comments about Typescript. Let me know if I should have done something else here instead.

src/js/components/DataTable/stories/typescript/Clickable.tsx
jcfilben: Done, also added it to chromatic testing

PR #4412, opened 2020-08-13
Fix 4334 InfiniteScroll remove duplicate items and duplicate "last" keys

halocline:
#### What does this PR do?

Fixes a scenario where when `show` is specified, a duplicate item results at the `step` index and also generates duplicate nodes with a key of "last".

This screenshot was taken on 8/12/2020 on the "Show 118th item" story. Note the duplicate rows of "item 50."

Screen Shot 2020-08-12 at 6 06 21 PM 

#### Where should the reviewer start?

InfiniteScroll.js

Specifically, the duplicates arise from this line: https://github.com/grommet/grommet/blob/b2a570efd87978cca46886670b581ea3ef0edbb8/src/js/components/InfiniteScroll/InfiniteScroll.js#L202

When the `show` index is > `step`, it will satisfy the condition twice.

Changes here also refactored to attempt to aid readability and better ability to reason about the code. Not sure if this was successful.

Lastly, DataTable-test.js.snap had one snapshot change. I verified this change manually, comparing the snapshot classes to the CSS values resulting from the React generated class names. I did not see any differences.

#### What testing has been done on this PR?

Jest, plus Storybook.

Added additional test for unique items.

#### How should this be manually tested?

Storybook.

When in Storybook, I found it helpful to `console.log(child)` prior to `result.push(child);` in InfiniteScroll.js for manual inspection of the nodes.

#### Any background context you want to provide?

Summary of findings when researching issue: https://github.com/grommet/grommet/issues/4334#issuecomment-665994857

https://github.com/grommet/grommet/issues/4334

**This is not a complete fix for #4334** -- it is a step towards solving.

#### What are the relevant issues?

https://github.com/grommet/grommet/issues/4334 & https://github.com/grommet/grommet/issues/3106

#### Do the grommet docs need to be updated?

No.

#### Should this PR be mentioned in the release notes?

Yes.

#### Is this change backwards compatible or is it a breaking change?

Fixes bug. Should be backwards compatible otherwise.

PR #4426, opened 2020-08-18
Fix 4334 InfiniteScroll show behavior

halocline:
#### What does this PR do?

- Fixes behavior to scroll to the specified showIndex (scrollTo was not happening)
- Fixes pageHeight calculation for when show > step (was being calculated too large)
- Expands testing coverage for "show scenarios"
- Expands Storybook stories for manual testing of show cases

Review current Storybook stories
- https://storybook.grommet.io/?path=/story/infinitescroll--show-118th-item

Manually review and compare Storybook stories with these changes 

InfiniteScroll.js

InfiniteScroll-test.js

Storybook, plus expanded Storybook and Jest coverage

Storybook.

https://github.com/grommet/grommet/issues/4334#issuecomment-665994857

This is not a complete fix for https://github.com/grommet/grommet/issues/4334 -- it is a step towards solving. --> A test is written for the following scenario when `replace && show > step` --> The beginPage is not being set correctly (being set to page 0) and a firstPageIndexRef is not being set for the scenario when the initial beginPage is > 0.

https://github.com/grommet/grommet/issues/4334 & https://github.com/grommet/grommet/pull/4412

No.

Yes.

Fixes bugs. Should be backwards compatible otherwise.


IanKBovard:
I might be misunderstanding how isBeforeScroll and isAfterScroll works, but they look inherently broken to me. If the scrollparent return from findScrollParent is the document these two functions always return false.

original
```
export const isNodeAfterScroll = (node, target) => {
  const { bottom } = node.getBoundingClientRect();
  // this seems to be the issue with show to me.
  // destructuring height and top from 0 would be undefined
  // this would be the case every time target it document
  const { height, top } = target.getBoundingClientRect
    ? target.getBoundingClientRect()
    : 0;
  // top and height do not === 0 but undefined
  return bottom >= top + height;
};

// same issues as above
export const isNodeBeforeScroll = (node, target) => {
  const { top } = node.getBoundingClientRect();
  // target will be the document from findScrollParent()
  const { top: targetTop } = target.getBoundingClientRect
    ? target.getBoundingClientRect()
    : 0;
  return top <= targetTop;
};
```

fix
```
export const isNodeAfterScroll = (node, target) => {
  const { bottom } = node.getBoundingClientRect();
  // this seems to me how these functions are intended to work.
  // if target.getBoundingClientRect then destructure, else height and top = 0 
  if (target.getBoundingClientRect) {
    const { height, top } = target.getBoundingClientRect();
    return bottom >= top + height;
  }
  return bottom >= 0;
};

export const isNodeBeforeScroll = (node, target) => {
  const { top } = node.getBoundingClientRect();
  if (target.getBoundingClientRect) {
    const { top: targetTop } = target.getBoundingClientRect();
    return top <= targetTop;
  }
  return top <= 0;
};
```

what do you think?

IanKBovard:
Currently this section never runs if the scrollParent is the document or body
```
always false
if (isNodeBeforeScroll(showNode, scrollParent)) {
  showNode.scrollIntoView(true);
  } else if (isNodeAfterScroll(showNode, scrollParent)) {
    showNode.scrollIntoView(false);
   }
}
```

halocline:
> what do you think?

Good call.

src/js/components/DataChart/XAxis.js
```diff
@@ -1,32 +1,30 @@
...
 import { Box } from '../Box';
 
-const XAxis = forwardRef(
-  ({ chartProps, data, renderValue, serie }, ref) => {
-    // pull the x-axis values from the first chart, all should have the same
-    const [axisValues] = (Array.isArray(chartProps[0])
-      ? chartProps[0][0]
-      : chartProps[0]
-    ).axis;
+const XAxis = forwardRef(({ chartProps, data, renderValue, serie }, ref) => {
```
ShimiSun, line 4: Why did this file change?

src/js/components/Chart/index.d.ts
```diff
@@ -19,7 +19,7 @@ export interface ChartProps {
   color?:
     | ColorType
     | { color: ColorType; value: number | number[] }[]
-    // deprecated 
+    // deprecated
```
ShimiSun, line 22: Can we try to avoid the distraction of file changes that are not relevant for this PR?

src/js/components/InfiniteScroll/stories/Replace.js
```diff
@@ -22,7 +22,12 @@ const InfiniteScrollReplace = props => (
...
-    <InfiniteScrollReplace replace />
-  ));
+  storiesOf('InfiniteScroll', module)
+    .add('Replace', () => <InfiniteScrollReplace replace />)
+    .add('Replace with Show 28th item', () => (
+      <InfiniteScrollReplace replace show={27} />
+    ))
+    .add('Replace with Show 88th item', () => (
+      <InfiniteScrollReplace replace show={87} />
+    ));
```
ShimiSun, line 32: Why do we need two stories that point to a specific item with `replace`? I'm asking because we want to keep the stories concise.

halocline: Since we have limited tools to test behaviors, Storybook is our best current option. I am fine removing the stories once this work is merged in or even relocating them, but they are valuable for QA effort.

In this case in particular, behaviors are different when show > step vs. show < step. We need both circumstances to work. When we only tested for one variant, we got false positives and scenarios where the component did not function properly.

Again happy to remove once we feel good about all the fixes on this component.

src/js/components/DataChart/XAxis.js
halocline: I believe it is a formatting change implemented by the linter on commit. I did not touch the file.

src/js/components/Chart/index.d.ts
halocline: I believe it is a formatting change implemented by the linter on commit. I did not touch the file.

src/js/components/InfiniteScroll/stories/Replace.js
ShimiSun: I agree we should keep it, but maybe adding a comment on the stories regarding show > step vs. show < step, as a user at a first glance it might not be obvious the difference between the two stories.

halocline: I could put both examples into one story, each labeled with the scenario which would demonstrate the functionality. That way we could reduce the clutter, but maintain QA scenarios.

The monorepo work may also be an opportunity to provide a place for these sorts of QA items, while leaving Storybook clean for users and documentation.

src/js/components/Chart/index.d.ts
halocline: It looks like these changes came in on this commit which was a merge of master: https://github.com/grommet/grommet/pull/4426/commits/8873ff0c64f3ab9836effa8b71540f6762d6b114

src/js/components/DataChart/XAxis.js
halocline: It looks like these changes came in on this commit which was a merge of master: https://github.com/grommet/grommet/pull/4426/commits/8873ff0c64f3ab9836effa8b71540f6762d6b114

ShimiSun: Can you merge master properly so we won't see those changes? If it was already done on a master's commit it shouldn't appear on this PR.

src/js/components/InfiniteScroll/stories/Replace.js
ShimiSun: >I could put both examples into one story, each labeled with the scenario which would demonstrate the functionality. That way we could reduce the clutter, but maintain QA scenarios.

I think I like that.

src/js/components/DataChart/XAxis.js
halocline: Will give it another go.

ShimiSun:
Thank you @halocline @halocline @jecassis & @JoeButler7 for collaborating on this and getting infinite scroll fixes to the finish line.

PR #4427, opened 2020-08-18
Update grid column logic

taysea:
#### What does this PR do?

Mirroring work done in https://github.com/grommet/grommet/pull/4120 to see if chromatic build has any snapshot changes


ShimiSun:
Ouch, I suspect even more components are affected... there is definitely something here...

image

taysea:
> Ouch, I suspect even more components are affected... there is definitely something here...
> 
> image

Hmm, will do some investigation

ShimiSun:
Do you have any explanation into 'why' it shows changes though? Can you maybe read about chromatic QA/ stack overflow what may be the root cause of these issues (maybe it's a well-known issue of Grid? or maybe there is a hidden animation?)? any explanation would help us feel more comfortable with changing the baseline. 
Also, did any of the snapshots changed on the PR? if it did it might be wort to start the investigation from there.

taysea:
> Do you have any explanation into 'why' it shows changes though? Can you maybe read about chromatic QA/ stack overflow what may be the root cause of these issues (maybe it's a well-known issue of Grid? or maybe there is a hidden animation?)? any explanation would help us feel more comfortable with changing the baseline.

I hadn't found anything on my initial investigation about delays/animations related to Grid or Chromatic specific issues, but I will do another investigation.

> Also, did any of the snapshots changed on the PR? if it did it might be wort to start the investigation from there.

The jest snapshot changes seem to align with the new logic of the code block, but I will look at those tests to see if they seem to overlap with the chromatic snapshot changes.

ShimiSun:
Any updates on this beauty?

taysea:
> Any updates on this beauty?

I have not had a chance to revisit it this week, but going to try diving back into it this afternoon.

taysea:
I think I've concluded the issue here, but I'm still trying to understand why it only replicates on some of these screenshots and not when we deploy storybook or run locally.

[This StackOverflow response](https://stackoverflow.com/a/51597753) explains that one of the `minmax()` arguments needs to be a fixed length. However, in our case both values can be variable (the first is the resolution of `min([user size], 100%)` and the second is `1fr` which is defined as a [flexible length](https://www.w3.org/TR/css3-grid-layout/#fr-unit). Because both are a flexible length, it's causing issues. It seems like the `min([user size], 100%)` needs to be resolved before being plugged into the `minxmax()` so that it is a fixed value.

ShimiSun:
@taysea Great progress, keep us posted.

taysea:
Some follow on findings. Based on these articles ([article 1](https://evanminto.com/blog/intrinsically-responsive-css-grid-minmax-min/) and [article 2](https://css-tricks.com/intrinsically-responsive-css-grid-with-minmax-and-min/)) discussing intrinsically responsive grid, it seems like the logic we've employed is valid given the newer CSS calculations of min(), etc. but that browser support is still growing. The support of these functions can be tracked on [caniuse](https://caniuse.com/css-math-functions). I am trying to see what version of Chrome our chromatic tests are running on and I'm wondering if it's running <78 or a version that does not yet support the min() function we're relying on.

ShimiSun:
@taysea Thanks! I sent Chromatic's support a message request for more information on the Chrome version the snapshots are being taken on.

taysea:
Great, thanks for confirming that @ShimiSun. So with that, I would say that the reason Chromatic is not respecting the grid is because version 73 of chrome doesn't handle use of `min()` which we are relying on for those. Locally, I am using version 85 (and I assume you and Eric were also on recent versions).

So, if we move forward with this change, we just have to be aware that this functionality will work only on what is displayed on [caniuse for css math functions](https://caniuse.com/css-math-functions).

ShimiSun:
@taysea fantastic resolution. I reached out to Chroma with a request to upgrade the chrome browser version and verify that the issue is resolved on a. newer browser.

taysea:
> @taysea fantastic resolution. I reached out to Chroma with a request to upgrade the chrome browser version and verify that the issue is resolved on a. newer browser.
> 
> Update:
> Response from Chromatic regarding upgrading the chrome version for us:
> 
> ```
> We unfortunately are not able to do that. 
> As far as the upgrade, there is no set timeline, but it is something we are actively working on. So, my hope is that it will happen sooner rather than later. Sorry for not having more specifics right now.
> ```

Gotcha, do we want to wait on merging these changes until we can confirm that no snapshots will be affected? Or do we want to move forward knowing we've tested on BrowserStack and all looks fine?

ShimiSun:
Good question, I was told the upgrade on chromatic will happen soon, we can check if the snapshots are fixed every few days before the merge.

ShimiSun:
CONGRATS!!! 
Finally, Chromatic upgraded their chrome version and no snapshots are marked as changed.

taysea:
> CONGRATS!!!
> Finally, Chromatic upgraded their chrome version and no snapshots are marked as changed.

Woohooo!

PR #4428, opened 2020-08-18
clean up duplicate stories

britt6612:
#### What does this PR do?
This PR cleans up TS stories that were duplicate of the regular TS stories
(G-Z)

Delete dups of `grid` only keep App.jsx for Typescript story
Delete `Basic.tsx` for `infinite scroll` 
Keep only `theme.jsx` for `List`
Delete `custom` for `Menu` only keep theme
Keep `MultipleValues` for Meter TS
keep `simple` for sidebar
keep `controlled` for `tabs`

none
storybook they should be gone
#4005
The idea here is to remove any typescript stories that are duplicated in regular js stories.
Then go back and organize the TS stories in their own folder within the component 
This is the first step of just. removing duplicate stories went through G-Z
#4005
no
no
backwards compatible


britt6612:
@jcfilben

britt6612:
> For the InfiniteScroll OnMore story there is a `.tsx` and a `.js` version but the code for the two versions is different. The `.tsx` version doesn't include a custom theme. This PR is removing the `.js` version. Not sure which one we should keep or if we should have both examples since the code is different.
> 
> Also List `selected.tsx` isn't duplicated however it is similar to `selection.js` so maybe we should only have one of these stories?
> 
> Looks like there is a Menu `Themed.tsx` and `Themed.js` but they are slightly different. We should either get rid of one or keep both and rename

I think initially I thought they were the same but yes since they differ I think we should keep both versions and just rename. That way we have more examples then getting rid of useful content

britt6612:
> > For the InfiniteScroll OnMore story there is a `.tsx` and a `.js` version but the code for the two versions is different. The `.tsx` version doesn't include a custom theme. This PR is removing the `.js` version. Not sure which one we should keep or if we should have both examples since the code is different.
> > Also List `selected.tsx` isn't duplicated however it is similar to `selection.js` so maybe we should only have one of these stories?
> > Looks like there is a Menu `Themed.tsx` and `Themed.js` but they are slightly different. We should either get rid of one or keep both and rename
> 
> I think initially I thought they were the same but yes since they differ I think we should keep both versions and just rename. That way we have more examples then getting rid of useful content

After looking 
> > Menu only had 3 lines of the custom theme so I moved those lines. to the tsx file and deleted the js

jcfilben:
This PR can be closed

PR #3130, opened 2019-05-29
implement same tag inheritance for gaps in box component

buoyantair:
#### What does this PR do?
Attempts to fix #3046 
#### Where should the reviewer start?
Start with reviewing the Box component :)

#### What testing has been done on this PR?
Snapshot testing has been done for normal and nested use cases.
#### How should this be manually tested?
Maybe
#### What are the relevant issues?
#3046 
#### Do the grommet docs need to be updated?
Yes, this can effect workflows, the distinction needs to be documented
#### Should this PR be mentioned in the release notes?
Sure 😬 

#### Is this change backwards compatible or is it a breaking change?
Likely a breaking change.

Issue #3046, opened 2019-04-23
Should Box gap use the same tag as its parent Box?


dallas:
### Expected Behavior

I would expect the gap elements within a `Box` component to have the same tag as the `Box` itself (e.g. `<Box as="span" gap="small">…</Box>` would also use `<span>` for each of the gaps).

### Actual Behavior

Instead, `Box` gaps are always `div` elements. This isn’t normally an issue, but since an `Anchor` with an icon uses an `"inline-flex"` `Box` component with a gap, it becomes an issue when that `Anchor` is within a `Paragraph`.

### Code Sandbox exhibiting the issue

Notice the console warnings when you first load this Code Sandbox: https://codesandbox.io/s/5yr109m8k4

### Steps to Reproduce

1. Create a `Paragraph`
2. Add an `Anchor` with an `icon` prop
3. Watch console for errors/warnings

### Your Environment

- Grommet version: 2.6.5
- Browser Name and version: Google Chrome 73.0.3683.103
- Operating System and version (desktop or mobile): MacOS Mojave 10.14.4

### Suggested Solution

Pass along the same `as`/`tag` props to the `StyledBoxGap` that are currently being passed to the `StyledBox` component.

buoyantair:
@ShimiSun Can I take this? :)

ShimiSun:
@buoyantair you got it buddy!

jcfilben:
I think this issue can be closed since #4464 was merged

PR #4436, opened 2020-08-20
Added 'as' option to the gap property

MikeKingdom:
#### What does this PR do?
This is a backward compatible way to make the tag used for the gap be
something other than a div. This is useful in cases where the box
'as' property is used to make the box something other than a div like
a span or ul, etc.

Box.js

JEST, storybook

See #3046 and #3130 for background.

#3046

Done

Can be

Backwards compatible

### DIFF
--- a/src/js/components/Box/Box.js
+++ b/src/js/components/Box/Box.js
@@ -95,8 +95,8 @@ const Box = forwardRef(
           } else {
             contents.push(
               <StyledBoxGap
-                // eslint-disable-next-line react/no-array-index-key
                 as={gap && gap.as ? gap.as : undefined}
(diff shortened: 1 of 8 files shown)


ericsoderberghp:
This feels a bit too over-engineered to me. I think we would address the primary issue by just detecting `<Box as="span" />` and then using `span` for the gap element. This will keep our API surface from growing unnecessarily. I don't think the example of `<Box gap={{ as: 'li' }} />` make sense as the gap wouldn't be an item in the list. For example, a screen reader shouldn't read it.

MikeKingdom:
Replaced with PR #4464

PR #3736, opened 2020-02-08
Feat 2973 2

markjackson02:
Picking up from @oorestisime 's work and @ShimiSun 's feedback on #3695 .

#### What does this PR do?
Implement #2973 

#### Where should the reviewer start?
storybook

#### What testing has been done on this PR?
none yet

#### How should this be manually tested?
storybook

#### What are the relevant issues?
#2973 

#### Do the grommet docs need to be updated?
done

#### Should this PR be mentioned in the release notes?
yes

#### Is this change backwards compatible or is it a breaking change?
compatible

PR #4440, opened 2020-08-21
Clear prop added to Select

### LINKED ISSUE #2973: Ability to unselect / clear selected value with <Select>

IanKBovard:
#### What does this PR do?
Adds a clear prop to the select component. When enabled, it adds a clear button to the top or bottom of the select dropdown that clears all selections when clicked.

Select component

Storybook

Same as above

#3736 

closes #2973 

Yes

Yes

Backwards compatible


src/js/components/Select/SelectContainer.js
```diff
@@ -39,9 +39,27 @@ const SelectOption = styled(Button)`
...
+  const { label, position, render } = clear;
+  const align = position !== 'bottom' ? 'start' : 'center';
+  if (render) {
+    return render({ onClear });
+  }
+  return (
+    <Box flex={false}>
+      <Button tabIndex="-1" onClick={onClear}>
+        <Box {...theme.select.clear.container} align={align}>
+          <Text {...theme.select.clear.text}>{label || 'Clear selection'}</Text>
```
ShimiSun, line 52: How would this work with I18N strings? Is it generic enough? 
I think we should be consistent using the `messages` prop for I18N supported strings.

src/js/components/Select/doc.js
```diff
@@ -23,6 +23,27 @@ export const doc = Select => {
...
+          .description('Label for the clear selection item')
+          .defaultValue('Clear selection'),
+        render: PropTypes.func.description(
+          `Render function to customize clear value item.
+          It receives as a parameter the callback to call
+          when clicking on the item`,
+        ),
+      }),
+    ])
+      .description(`Whether to provide button to clear selection.`)
```
ShimiSun, line 45:
```suggestion
      .description(`Whether to provide a button option to clear selections.`)
```
Does that make more sense?

src/js/components/Select/SelectContainer.js
IanKBovard: Only clear selection wouldn't be able to be translated. But anything they put into label will work. 

The reason this is named render is because it works like the render prop for other grommet components. I believe this operates much differently than form's messages prop.

ShimiSun: We should use the `messages` prop to every string we are providing. 'Clear selection' is one of them.

IanKBovard: label would handle that if they need to translate the string. Clear selection is just a default label, if no label is passed in. Same as required is the default label for the error message. I could rename the label prop to messages, but I think that would make the prop unclear as to its use. Since, it operates the same same Button's label prop.

ShimiSun: Can you take a look at how it's done on other components? I'm pretty sure it should be under `messages`.

IanKBovard: https://v2.grommet.io/button#label
https://v2.grommet.io/menu#label
https://v2.grommet.io/textarea#placeholder

label makes the most sense, since it's purpose is to work the similarly to Button's label prop. If you're concerned about i18n, any string that I can pass in can be translated it doesn't have to be in the messages prop. The reason messages was necessary when using form was because there was no other way to modify the default required and invalid error responses otherwise. 

here are examples of i18n working with different grommet props.
https://github.com/grommet/grommet-i18n/tree/master/src/js/pages/Reacti18next
The select example has buttons with their labels being translated
https://github.com/grommet/grommet-i18n/blob/master/src/js/pages/Reacti18next/SelectI18next/index.js

src/js/components/Select/SelectContainer.js
```diff
@@ -39,9 +39,27 @@ const SelectOption = styled(Button)`
...
 `;
 
+const ClearButton = ({ clear, onClear, theme }) => {
+  const { label, position, render } = clear;
+  const align = position !== 'bottom' ? 'start' : 'center';
+  if (render) {
+    return render({ onClear });
+  }
+  return (
+    <Box flex={false}>
```
ericsoderberghp, line 49: Do we need this Box? I'm just comparing this to how SelectOption and OptionBox work for options.

src/js/components/Select/SelectContainer.js
```diff
@@ -207,6 +225,13 @@ const SelectContainer = forwardRef(
       [multiple, onChange, optionIndexesInValue, options, valueKey],
     );
 
+    const onClear = useCallback(
+      event => {
+        onChange(event, { option: '', value: '', selected: '' });
```
ericsoderberghp, line 230: I wonder if `option` should be `undefined` here?

src/js/components/Select/doc.js
```diff
@@ -23,6 +23,27 @@ export const doc = Select => {
...
+        position: PropTypes.oneOf(['top', 'bottom'])
+          .description(
+            `Add a clear value to the top or at the bottom of the
+            container. By default no clear option is present.`,
+          )
+          .defaultValue('top'),
+        label: PropTypes.string
+          .description('Label for the clear selection item')
+          .defaultValue('Clear selection'),
+        render: PropTypes.func.description(
```
ericsoderberghp, line 38: Has someone asked for `render` or is this something we're anticipating? I ask because I'd like to keep the API surface as small as we can until a need arises.

src/js/components/Select/doc.js
```diff
@@ -23,6 +23,27 @@ export const doc = Select => {
       { active, disabled, selected } keys indicating the current state
       of the option.`,
     ),
+    clear: PropTypes.oneOfType([
+      PropTypes.bool,
+      PropTypes.shape({
+        position: PropTypes.oneOf(['top', 'bottom'])
+          .description(
+            `Add a clear value to the top or at the bottom of the
```
ericsoderberghp, line 31: I think we should use the term `option` instead of `value` or `button` in our descriptions. I think that aligns a bit better with `options` being what we show in the drop.

src/js/components/Select/SelectContainer.js
ericsoderberghp: I think `label` makes sense here. I can see cases where the clear label text should align with the context of the Select. For example, if the Select is in a FormField with `label="Cities"`, the clear option label could be `Clear cities`. In other words, the label could be aligned with the field label, and different for each field.

src/js/components/Select/doc.js
IanKBovard: I don't think anyone has asked for it, i took it from this pr
#3736

src/js/components/Select/SelectContainer.js
IanKBovard: changes made

src/js/components/Select/doc.js
IanKBovard: removed render

src/js/components/Select/doc.js
IanKBovard: changes made

src/js/components/Select/SelectContainer.js
IanKBovard: removed the box

src/js/components/Select/SelectContainer.js
IanKBovard: the button label is now either label, 'clear selection', or 'clear {name of formfield}' if its passed in.

src/js/themes/base.js
```diff
@@ -962,6 +962,15 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
     select: {
       // background: undefined,
+      clear: {
+        container: {
+          pad: 'small',
+          background: 'light-2',
+        }, // any box props
+        text: {
+          color: 'dark-3',
+        }, // any text props
```
ShimiSun, line 972: Let's cross these colors with Jet's designs and see whether we can make the theme not default unless a must.

src/js/components/Select/stories/Clear.js
```diff
@@ -0,0 +1,63 @@
...
+        value={value}
+        options={options}
+        onChange={({ option }) => setValue(option)}
+        clear={{ label: 'Click me!' }}
+      />
+    </Box>
+  );
+};
+
+const ClearExamples = () => (
```
ShimiSun, line 53: Each example should use multiple.
We should add a FormField use case and confirm the feature works as expected in Forms.

IanKBovard:
Modified stories to include multiple. Added Select wrapped in formfield and with a custom theme to stories.
Tests added:
clear button renders on top of selection list
clear button renders on bottom of selection list
clear button renders custom label
clear button renders with correct label when wrapped by formfield
clear button clears value

src/js/components/Select/stories/Clear.js
IanKBovard: formfield and multiple examples have been added with the latest commit

src/js/themes/base.js
IanKBovard: I used these designs for the default colors. 

65728693-eee20d00-e078-11e9-829e-cce013916dfb

ShimiSun: I've been discussing the usage of the colors with eric, and we should avoid the usage of `dark-*` and `light-*` colors. We should be using generic coloring like`background-contrast` and `text`/`text-weak` instead.
It might need some exploration of what are the right colors to pick, and it would be OK if we wouldn't have the exact coloring as the designs show, as long as we are maintaining the standard of using generic colors on our default theme.
The light /dark colors are intended to be deprecated at one point, so less work for us later if we remove it from here now.

ShimiSun: Interesting, it should work. have you tried other colors, maybe `background-front`? I'm suspecting background-contrast might not work on a white background, see if that is the reason.

IanKBovard: that doesnt work either. just to make sure, this is what i have.

```
    select: {
      // background: undefined,
      clear: {
        container: {
          pad: 'small',
          background: 'background-front',
        }, // any box props
        text: {
          color: 'text-weak',
        }, // any text props
      },
```

ShimiSun: Yea. that should work, take a look at what colors are 'text-weak' and  'background-*' are being resolved to.

IanKBovard: text-weak : #555555
background-front: #FFFFFF08

ShimiSun: @taysea can you help with figuring the color issue?

taysea: Yes, let me have a look.

taysea: Hey @IanKBovard, it looks like for the background-contrast adds, you need to invert light and dark values, I used this local on your branch and things worked as expected:
```
'background-contrast': {
    dark: '#FFFFFF08',
    light: '#33333308',
  },
```
Screen Shot 2020-09-03 at 2 54 01 PM

taysea:
Right now there is no way to access the "Clear" button via keyboard. Off first glance / brief discussion with Ian, it feels like the "clear" button should be accessible by Tab (so users could clear options without having to arrow through all options)

IanKBovard:
button can now be tabbed into

IanKBovard:
added onblur and onfocus to the clearbutton. if its focused it wont run onenter for the selection, and will clear the selections.

taysea (approved):
Nice addition. This LGTM, I think it would be good to get a final look from @ericsoderberghp 

Also, it seems unrelated but I'm wondering why there are some many chromatic snapshots that need to be added.

IanKBovard:
> Also, it seems unrelated but I'm wondering why there are some many chromatic snapshots that need to be added.

just merged with master, fixed the chromatic snapshots.

ShimiSun:
@IanKBovard, overall looking good and I would like to add this functionality to the next release.
The last action item we should take is to show "Clear" only when there is at least one selected item, there is no need to display the Clear option when no item is selected. Can we update this functionality so it will reflect on the stories?

src/js/components/Select/stories/Clear.js
```diff
@@ -0,0 +1,130 @@
...
+      </Box>
+      <Box direction="row">
+        <ClearForm />
+        <ClearCustomTheme />
+      </Box>
+    </Box>
+  </Grommet>
+);
+
+storiesOf('Select', module).add('Clear', () => <ClearExamples />);
```
ShimiSun, line 130: Can we exclude this story from Chromatic testing?

IanKBovard: changes made

taysea (commented):
This looks like it's really close! just noted one more item. If I tab to the "Clear" button, then realize I actually would like to go back to selecting an option there's no way for me to tab backwards to get back into that mode. I have to close out the whole Select and start the flow over. Would we want to support the ability to go back to selecting options after they've already tabbed to the "clear" button?

IanKBovard:
That sounds reasonable to me. Working on that now.

IanKBovard:
> This looks like it's really close! just noted one more item. If I tab to the "Clear" button, then realize I actually would like to go back to selecting an option there's no way for me to tab backwards to get back into that mode. I have to close out the whole Select and start the flow over. Would we want to support the ability to go back to selecting options after they've already tabbed to the "clear" button?

I've been doing some digging on this, and it seems to be an already existing issue with the Select dropdown component. 
Kapture 2020-10-26 at 10 11 51
(here im tabbing into and out of the address bar, and mashing the arrow keys)

I can tab into the component, but once I open it and tab out im unable to use the arrow keys to select a dropdown option. I can close the menu with esc, however.

Here are a few things to consider:
 - I don't think multi-select dropdowns are very accessible to begin with. Checkboxes would serve that purpose better.
 - The tab behavior of Select is strange to me. Either I'm locked in the menu, or tab closes the dropdown and I focus the 
    next focusable element.
 - A button within the select doesn't seem accessible either. This makes the most sense for accessiblity 
   https://storybook.grommet.io/?path=/story/select--unselect

taysea:
> > This looks like it's really close! just noted one more item. If I tab to the "Clear" button, then realize I actually would like to go back to selecting an option there's no way for me to tab backwards to get back into that mode. I have to close out the whole Select and start the flow over. Would we want to support the ability to go back to selecting options after they've already tabbed to the "clear" button?
> 
> I've been doing some digging on this, and it seems to be an already existing issue with the Select dropdown component.
> Kapture 2020-10-26 at 10 11 51
> (here im tabbing into and out of the address bar, and mashing the arrow keys)
> 
> I can tab into the component, but once I open it and tab out im unable to use the arrow keys to select a dropdown option. I can close the menu with esc, however.
> 
> Here are a few things to consider:
> 
> * I don't think multi-select dropdowns are very accessible to begin with. Checkboxes would serve that purpose better.
> * The tab behavior of Select is strange to me. Either I'm locked in the menu, or tab closes the dropdown and I focus the
>   next focusable element.
> * A button within the select doesn't seem accessible either. This makes the most sense for accessiblity
>   https://storybook.grommet.io/?path=/story/select--unselect

I think that given your analysis of the existing issues with Select tab behavior, your implementation here is sufficient and provides for an improvement to Select. I think the other issues you noted should be filed as separate Select accessibility enhancements. Once those are resolved, we can revisit how to support tabbing back to the list of options. Given that, I think the state of this PR looks good to me. But I'd like to hear @ShimiSun or @ericsoderberghp thoughts on that.

src/js/components/Select/stories/Clear.js
jcfilben, line 130: Chromatic was disabled for Select/stories/Basics.js but it doesn't look like it was disabled for Select/stories/Clear.js. Other than that the PR looks good to me

ShimiSun: @IanKBovard can you please disable chormatic for this story?

IanKBovard: changes made

taysea (approved):
LGTM (pending jest tests being fixed). I think there are some items to re-address down the road as Ian noted in his final comment but these are existing behaviors, not newly introduced by this PR.

ShimiSun:
The new story is not added to the Select component.

PR #4452, opened 2020-08-25
onEsc and onClickOutside for non-modal layers

IanKBovard:
#### What does this PR do?
Adds a new condition to allow onEsc and onClickOutside for non-modal layers.
Layers with modal={false} will not allow onEsc and onClickOutside by default, but layers with modal={false} and onEsc and onClickOutside handlers will work

Keystrokes will not be taken
```
<Layer modal={false}>...</Layer>
```

Keystrokes will be taken
```
<Layer modal={false} onEsc={onEsc} onClickOutside={onClickOutside}>...</Layer>
```

Keystrokes will be taken
```
<Layer modal={false} onClickOutside={onClickOutside}>...</Layer>
```

Keystrokes will be taken
```
<Layer modal={false} onEsc={onEsc}>...</Layer>
```

LayerContainer.js

None yet. Will test with storybook.

Same as above

onEsc functionality of layers with modal set to false was removed with this PR #3632.

#3771 

No

No

Yes

### DIFF
--- a/src/js/components/Layer/LayerContainer.js
+++ b/src/js/components/Layer/LayerContainer.js
@@ -66,8 +66,16 @@ const LayerContainer = forwardRef(
         if (modal && !element && anchorRef.current) {
           anchorRef.current.focus();
         }
+        if (
+          !modal &&
+          !element &&
+          anchorRef.current &&
+          (onEsc || onClickOutside)
+        ) {
+          anchorRef.current.focus();
+        }
       }
-    }, [modal, position, ref]);
+    }, [modal, position, ref, onEsc, onClickOutside]);
 
     useEffect(() => {
       if (position !== 'hidden') {


src/js/components/Layer/LayerContainer.js
```diff
@@ -66,8 +66,16 @@ const LayerContainer = forwardRef(
         if (modal && !element && anchorRef.current) {
           anchorRef.current.focus();
         }
+        if (
+          !modal &&
+          !element &&
+          anchorRef.current &&
+          (onEsc || onClickOutside)
```
ericsoderberghp, line 73: This doesn't seem to address the issue. If it's not a modal Layer, I don't think `onClickOutside` is connected up to anything.

IanKBovard: I gotcha. I was mistaken when testing. onEsc will work, however.

IanKBovard: ill wait on a response in #3771 before I make changes. Maybe we should just update the docs to be explicit about the purpose of non-modal layers instead.

ShimiSun:
where are we with this PR @IanKBovard ?

IanKBovard:
Closing this pr because onclickoutside and onesc would break functionality of a non-modal layer

PR #4464, opened 2020-08-27
Changed the gap to be a span if the box is a span

MikeKingdom:
## What does this PR do?
Implements the solution recommended by Eric. It makes the gap element be a span rather than a div if the box is a span. Browsers produce warnings in the console if a div is a child of a span.

One example of this is the implementation of Anchor with an icon and label. This change fixes this situation.

This PR replaces #4436 and #3130 with a different implementation

## Where should the reviewer start?
Box.js

## What testing has been done on this PR?
JEST, storybook

## How should this be manually tested?
See #3046 and #3130 for background.

## What are the relevant issues?
#3046

## Do the grommet docs need to be updated?
Done

## Should this PR be mentioned in the release notes?
No need.

## Is this change backwards compatible or is it a breaking change?
Backwards compatible

### DIFF
--- a/src/js/components/Anchor/__tests__/__snapshots__/Anchor-test.js.snap
+++ b/src/js/components/Anchor/__tests__/__snapshots__/Anchor-test.js.snap
@@ -169,7 +169,7 @@ exports[`Anchor icon label renders 1`] = `
       <svg
         color="#7D4CDB"
       />
-      <div
+      <span
         className="c3"
       />
       Test
@@ -412,7 +412,7 @@ exports[`Anchor reverse icon label renders 1`] = `
       }
     >
       Test
-      <div
+      <span
         className="c3"
       />
       <svg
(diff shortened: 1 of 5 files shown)


src/js/components/Anchor/__tests__/__snapshots__/Anchor-test.js.snap
```diff
@@ -169,7 +169,7 @@ exports[`Anchor icon label renders 1`] = `
       <svg
         color="#7D4CDB"
       />
-      <div
+      <span
```
ShimiSun, line 172: The fact that the div is not a child of span anymore is obviously the right approach, but I was wondering if you have more insights into why this change happened. 
I can see why Box will change its snapshots due to the new test, but I'm not sure why did this snapshot of Anchor change. I'll appreciate the insight.

MikeKingdom: That's what I was alluding to in the second paragraph of the description. If you look at the original issue it was found because Anchor does this internally and that was what the submitter used as the example problem. This change fixes that and it shows up as the change in Anchor's output when you have an icon and label.

ShimiSun:

Issue #4460, opened 2020-08-26
Tab - active props is no longer working


deniselashlley:
We created a general tab style in the main theme, however want to override the active style in a specific component using props.active. It works fine in version 2.14.0. Updated to the latest version and no longer works with props.active

### Expected Behavior

The tab button should show the overriding background-color when active=true/false

### Actual Behavior

the background color uses the default theme color instead active=undefined

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/brave-tdd-ussjb?file=/src/component/buttonTab.js

### Your Environment

- Grommet version: @2.15.0
- Browser Name and version: chrome
- Operating System and version (desktop or mobile): both

ShimiSun:
Adding @jcfilben comment from [Slack](https://grommet.slack.com/archives/C04LMHN59/p1598483127016400?thread_ts=1598439935.015400&cid=C04LMHN59):

>Jessica Filben  2 hours ago
Hey Denise, I found this PR that was merged in the latest release. It mentions that previously active was passed to Tab as a prop via clone child. Now active is passed via TabsContext . I think that this is why props.active is not working. I think that the active prop can still be accessed from Tab just might have to go about it differently than before. I'm still trying to figure out exactly how this can be done or if something needs to be changed in the code. Let me know if you have any insights
https://github.com/grommet/grommet/pull/4255

ericsoderberghp:
It looks like we should have passed `active` onto StyledTab: `<StyledTab active={active} />`.

taysea:
I'm taking a look at where this isn't getting passed through. I tried out adding active to styled tab and didn't have success but will try again.

ShimiSun:
Thank you for reporting, the fix is now available on the [stable branch](https://github.com/grommet/grommet/wiki/What-is-grommet-stable-and-how-to-use-it%3F). And it will be available on our next official release (^2.15.1)

PR #4466, opened 2020-08-28
Allow active to pass as prop

taysea:
#### What does this PR do?
Passes `active` as a prop to Tab to allow custom styled component css to apply based on active state. The TabContext still keeps track of which tab is active, active title, etc. but this needs to be passed down to the individual Tab to allow users to styled via styled components.

src/js/components/Tab/Tab.js

Test this styled component locally in storybook: https://codesandbox.io/s/jolly-morse-pwte4?file=/src/component/buttonTab.js:296-839

Using the above code

Closes https://github.com/grommet/grommet/issues/4460#issuecomment-682247331

No.

Yes.

No, backwards compatible.


jcfilben:
Looks good! I tested in storybook with the example you have and it worked for me. Not sure if it is necessary to have but here is a quick test that we could add to `Tabs-test.js` that would test this in the future
```javascript
const ButtonTab = styled(Tab)`
  ${(props) => {
    return css`
      background: ${props.active ? "blue" : "green"};
    `;
  }}
`;

  test('styled component should change tab color when active', () => {
    const { container, getByText } = render(
      <Grommet>
        <Tabs justify="start" margin={{ vertical: "medium" }}>
            <ButtonTab title="About" />
            <ButtonTab title="Activity" />
            <ButtonTab title="Stickers" />
        </Tabs>
      </Grommet>
    );
    fireEvent.click(getByText('Activity'));
  });
```

src/js/components/Tabs/Tabs.js
```diff
@@ -55,7 +55,13 @@ const Tabs = forwardRef(
           setActiveTitle,
         }}
       >
-        {child}
+        {/* possible to have undefined child. in that case, you can't
+        do cloneElement */}
+        {child
+          ? // need active to be passed as prop for custom styled component
+            // that uses props.active
+            React.cloneElement(child, { active: activeIndex === index })
```
ericsoderberghp, line 63: I don't think we need to `cloneElement`. I think we can just add `<StyledTab active={active} />`. At least, I tried that on the custom theme story with:
```
extend: ({ active, theme }) => css`
      ${active && 'background: #f00;'}
`
```
and it worked fine.

taysea: In the case of this user, they weren't applying it in the theme but instead were using a styledcomponent: https://codesandbox.io/s/brave-tdd-ussjb?file=/src/component/buttonTab.js

I tried their code locally with passing active to `StyledTab` and wasn't able to see the proper results.

ShimiSun:
Can you please address the test failure?

src/js/components/Tabs/Tabs.js
ShimiSun: Can we add a comment:
// cloneElement is needed for backward compatibility. We should remove that in our next major release.

taysea: Wouldn't we want to keep it in the future as well if it's allowing the styled component to be styled properly?

ShimiSun: It's more of a discovery comment for the purpose of cleaning up the `cloneElement` usage, not sure which direction it will play.

taysea: Sounds good I will add the comment in.

PR #4468, opened 2020-08-28
Add internal defaultCursor to fix Select disabled cursor

taysea:
#### What does this PR do?
Passes `defaultCursor={disabled}` an internal prop to the TextInput that makes up the Select button. When Select is disabled, we don't want this TextInput to apply its own disabled styling but we do want the cursor to remain default as opposed to the pointer hand which alludes to the ability to click.

Applied `disabled` in storybook to Simple Select and saw that the cursor remained as the default as opposed to the pointer.

Same as above.

Closes https://github.com/grommet/grommet-theme-hpe/issues/96

No.

No.

Yes, backwards compatible.


src/js/components/Select/Select.js
```diff
@@ -23,7 +23,7 @@ import { SelectContainer } from './SelectContainer';
 import { applyKey } from './utils';
 
 const SelectTextInput = styled(TextInput)`
-  cursor: pointer;
+  cursor: ${props => (props.defaultCursor ? 'default' : 'pointer')};
```
ShimiSun, line 26: why not using `props.disabled` directly and remove the `defaultCursor` prop?

taysea: If we pass as `disabled` then `opacity` will be affected as well. We don't want opacity to be lowered by both Select and the underlying TextInput. We just want to control the cursor.

src/js/components/Select/Select.js
```diff
@@ -255,7 +255,7 @@ const Select = forwardRef(
                   }
                   // When Select is disabled, we want to show a default cursor
                   // but not have disabled styling come from TextInput
-                  defaultCursor={disabled}
+                  defaultCursor={disabled === true || undefined}
```
ShimiSun, line 258:
```suggestion
                  defaultCursor={disabled || undefined}
```
Should work as well

taysea: disabled can also take an array that will disabled certain options. I think we only want to disable the TextInput if the control button should be disabled. this would occur only when disabled is true, rather than when an array of disabled options is defined.

ShimiSun: Good observation. would be helpful to add a comment there.

taysea: sounds good, I will add it in

PR #4470, opened 2020-08-31
Fix "min" aggregate function in DataTable

parth611git:
#### What does this PR do?
It solves problem when use has applied aggregate function as "min" in DataTable
Issue can be duplicated by having negative value in data and apply aggregate function as min on that data, it will provide value 0 instead of negative value which is lowest
Tested with local data setup by running buildState.js file by providing sample data and columns
No
Yes
Yes backward compatible.


ShimiSun:
closes #4471

Issue #4467, opened 2020-08-28
CheckBoxGroup options' key attributes should be value instead of label


phinch:
CheckBoxGroup appears to use the `label` of each option as their React `key`s instead of using the `value`. By contrast, RadioButtonGroup options use `value` for their `key`s. 

I don't know of any material bug that this issue causes; it's more to address a React warning that appears in the browser. It's a duplicate React key issue, so any potential bugs would probably be around trying to render updates to specific CheckBoxGroup labels.

### Expected Behavior

When a CheckBoxGroup or RadioButtonGroup has options that have the same label but different values, the options of the CheckBoxGroup or RadioButtonGroup should have unique keys, since the options have different values

### Actual Behavior

When passed the same set of options, React warns that the CheckBoxGroup has multiple options with identical keys, while not throwing the same warning for RadioButtonGroup.

Broadly, I would argue that there should be consistency between these two components' usage of `key`. [CheckBoxGroup uses the label for the key](https://github.com/grommet/grommet/blob/master/src/js/components/CheckBoxGroup/CheckBoxGroup.js#L80) while [RadioButtonGroup uses the value](https://github.com/grommet/grommet/blob/master/src/js/components/RadioButtonGroup/RadioButtonGroup.js#L127). I would posit that it's more canonical to use `value` as the `key` attribute instead of `label`, because `label` is just a display prop that shouldn't be guaranteed to be unique, whereas `value`, which is what would be stored as data, is more expected to be unique, making for a better candidate for `key`. That said, I'm not familiar with how these decisions have been made in the past for Grommet, and may be speaking too specifically to my context.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/proud-worker-g8h23?file=/src/App.js

### Steps to Reproduce

1. Load the page and open the console - you should see the React warning applied to the `CheckBoxGroup`, but not to the `RadioButtonGroup`:
```
Encountered two children with the same key, `Yes!`. Keys should be unique so that components maintain their identity across updates.
```

### Your Environment

- Grommet version: 2.14.0
- Browser Name and version: Chrome 84.0.4147.135
- Operating System and version (desktop or mobile): OSX Catalina 10.15.6

ShimiSun:
Hi buddy, sounds like you have delved in it enough to understand the is, care to contribute a fix to the open source?
https://github.com/grommet/grommet/blob/master/src/js/components/CheckBoxGroup/CheckBoxGroup.js#L79

[...]

PR #4474, opened 2020-09-01
Fix duplicate keys on CheckBoxGroup/Checkbox

abnersajr:
Signed-off-by: Abner Soares Alves Junior <abner.alves@codeminer42.com>

#### What does this PR do?

This PR fixes the problem of `CheckBoxGroup`, where `CheckBox` doesn't have a unique key for each `CheckBox` item.

`components/CheckBoxGroup/CheckBoxGroup.js`

Manually checking and existents unit tests.

Create a `CheckBoxGroup` with two or more Items.

No.

This fixes the issue #4467.

No.

Not necessary.

No.


ShimiSun:
Thank you for this PR and for your contribution.
To be honest, I do not see the use case of a valid UX scenario where two labels are named the same. (i.e. in your example two Yes options on separate CheckBox items), and I think the value will be more unique than the label. Please share with me a valid scenario of identical labels of CheckBox items on a CheckBoxGroup component (and out of curiosity for RadioButtonGroup as well).
RadioButtonGroup vs CheckBoxGroup do not operate exactly the same, CheckBox has the extra mechanism layer that knows how to select more than an item on the group.

Regardless, I'm thinking it will be best if the key will be the combination of both label+value, that should guarantee a unique key, thoughts?

src/js/components/CheckBoxGroup/__tests__/CheckBoxGroup-test.js
```diff
@@ -64,8 +64,12 @@ describe('CheckBoxGroup', () => {
     const component = renderer.create(
       <Grommet>
         <CheckBoxGroup disabled options={['First', 'Second']} />
-        <CheckBoxGroup options={[{ label: 'First', disabled: true }]} />
-        <CheckBoxGroup options={[{ label: 'First', disabled: true }]} />
+        <CheckBoxGroup
+          options={[{ label: 'First', value: 'First', disabled: true }]}
```
ShimiSun, line 68: If you change the test, you need to update the snapshots `yarn test-update`.
And according to the issue, this PR is trying to solve, I think you want to add a new test with similar labels on the same CheckBoxGroup, to verify that the keys are different.

abnersajr:
@ShimiSun definitely a checkbox group with two checkboxes with same label isn't a common idea. One scenario that come in mind is if you for example decide to Have a list and you told the user that you must agree with all the checkboxes below, and for some reason they decide to have same label for those this might cause issue.

Definitely the label+value is a better combo.
I'm gonna add this instead only value.

src/js/components/CheckBoxGroup/__tests__/CheckBoxGroup-test.js
abnersajr: Key aren't rendered in snapshots and I didn't find a correct way to test the duplication. But It's a simple concatenation of two variables

ShimiSun: I think we shouldn't change this test since you added a new test below to cover your use case. Let's keep this test intact.

ShimiSun (approved):

PR #4475, opened 2020-09-01
Add custom id to AnnounceContext

### LINKED ISSUE #4384: Unintentional removal of DOM elements
Grommet removes DOM elements that were built by external libraries (and live outside of the application).

### Expected Behavior

Grommet doesn't remove DOM elements that are unrelated to the application and the design system.

### Actual Behavior

In the AnnounceContext component, `document.body.querySelector('[aria-live]')` is used to target elements, but because of the broadness of the selector it also targets elements it doesn't intend to target. A few lines later it replaces the innerHTML of the wrong element with content (or a blank string).

https://codesandbox.io/s/grommet-v2-template-rl60j

### Steps to Reproduce

1. Go to https://codesandbox.io/s/grommet-v2-template-rl60j
2. Make note of the text at the bottom.
3. Start typing something in the TextInput. The content is replaced with one of the suggestions or an empty string.

### Your Environment

- Grommet version: 2.14.0
- Browser Name and version: Saw the issue on Chrome and Firefox. Haven't checked others.
- Operating System and version (desktop or mobile): MacOS Catalina (10.15.6)

abnersajr:
Signed-off-by: Abner Soares Alves Junior <abner.alves@codeminer42.com>

#### What does this PR do?

Adds an ID to `AnnounceContext`  with a default id and allow pass a custom ID.
This fixes one issue where any `aria-live`  element was being cleaned.
Grommet's Announcer has a default id: `grommet-announcer`.

`AnnounceContext.js` 

StoryBook with a custom ID.

Go to StoryBook and Inspect the `AnnounceContext` code with developerTools, inspecting the correct frame.

No.

It fixes #4384.

Yes. `doc.js`  file was edited

IMO yes. Accessibility is something important.

Backwards Compatible. Has a default value.


src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -13,11 +14,11 @@ const createAnnouncer = () => {
 };
 
 export const AnnounceContext = React.createContext(
-  (message, mode = 'polite', timeout = 500) => {
+  (message, mode = 'polite', timeout = 500, id = 'grommet-announcer') => {
```
ShimiSun, line 17: I don't think we should be setting a new default value. can we keep the default behavior as it is now?

abnersajr: Letting default behavior as it is now won't fix the issue if people don't explicit add one id.
The addition of default value doesn't cause any side and negative effect. Although, it may prevent further issues.
Please let me know if you still wanting remove it.

ericsoderberghp: I think it's OK to add this as the default.

src/js/contexts/AnnounceContext/doc.js
```diff
@@ -19,9 +19,16 @@ export const doc = AnnounceContext => {
...
+      'id' used to query a specific Announce Context. 
+      Default is 'grommet-announcer'
       Example:  
       {announce => 
-        <Button onClick={() => announce("Button was clicked", "polite", 1000)
+        <Button onClick={() => announce(
+          "Button was clicked", 
+          "polite", 
+          1000, 
+          "custom-announcer"
```
ericsoderberghp, line 30: Do we really need this level of flexibility? Can Grommet's AnnounceContext be thought of as a singleton and just use its embedded id automatically?

abnersajr: My idea in add this flexibility, instead of let with a fixed ID, is to give user the power to change if is necessary. We never know what user will need.
This doesn't add a lot of code and complexity.
In case you think isn't necessary this flexibility we can let with the default  id only.

ericsoderberghp: Understood. We try to keep the API surface of grommet as simple as possible, resisting the urge to build things out because they might be useful someday to someone in the future. Instead, we let practical needs of specific situations drive additions. In this regard, I think we should just fix the issue by adding and using the `grommet-announcer` id. If someone encounters a specific issue that requires further customization in the future, we can add it then.

abnersajr: Gotcha. Good to know it. I'm gonna remove it.

src/js/contexts/AnnounceContext/index.d.ts
```diff
@@ -3,10 +3,12 @@ import * as React from 'react';
 export type AnnounceMessage = string;
 export type AnnounceMode = 'off' | 'polite' | 'assertive';
 export type Timeout = number;
+export type Id = string;
 export type AnnounceValue = (
   message: AnnounceMessage,
   mode: AnnounceMode,
   timeout: Timeout,
+  id: Id,
```
ericsoderberghp, line 11: I think we removed this.

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -16,8 +17,9 @@ export const AnnounceContext = React.createContext(
   (message, mode = 'polite', timeout = 500) => {
     // we only create a new container if we don't have one already
     // we create a separate node so that grommet does not set aria-hidden to it
+    const id = 'grommet-announcer';
```
ericsoderberghp, line 20: Seems inconsistent to take `id` as an argument but then hard code it here. Should we remove the argument and move line 20 up to line 4?

src/js/components/Grommet/__tests__/Grommet-test.js
```diff
@@ -112,6 +117,29 @@ describe('Grommet', () => {
     }, 600); // wait the aria-live container to clear
   });
 
+  test('custom id announce', done => {
+    const id = 'custom-announcer';
```
ericsoderberghp, line 121: I don't think we need the CustomAnnouncer test anymore since we aren't taking `id`.

ericsoderberghp (approved):

src/js/contexts/AnnounceContext/doc.js
```diff
@@ -14,14 +14,19 @@ export const doc = AnnounceContext => {
...
+      can be called when something should be announced.
       'announce' function accepts 'message', 'mode' and 'timeout' as arguments
       and these arguments can be passed as 'props' to the return component.
-      'mode' can be one of 'polite', 'assertive' or 'off'. 
+      'mode' can be one of 'polite', 'assertive' or 'off'.
       'timeout' is measured in milliseconds.
-      Example:  
-      {announce => 
-        <Button onClick={() => announce("Button was clicked", "polite", 1000)
+      Default is 'grommet-announcer'
```
ShimiSun, line 22: I don't think the context of this line makes sense since it's not connecting to the 'id' value. 
Regardless, I'm not sure it's relevant now that `id` isn't a prop.

IanKBovard: I agree with shimi on this on. id isn't a prop, so I don't think it needs to be in the docs here. If it is important, maybe we could think about accessibility documentation for grommet in the future. Besides that, pretty simple changes LGTM.

IanKBovard:
Screen Shot 2020-10-13 at 2 55 37 PM

just an image showing this pr is working

src/js/contexts/AnnounceContext/doc.js
ShimiSun: @IanKBovard can you push a fix to this PR to fix the docs, it will speed up the reviewing.

abnersajr: @ShimiSun @IanKBovard thanks for the discussion. I just pushed removing this line from the docs.

ShimiSun (approved):

ShimiSun:
Thank you for your contribution @abnersajr! Hope you are participating in [Hacktoberfest](https://hacktoberfest.digitalocean.com/) and we'll see more PRs from you!

Issue #4446, opened 2020-08-24
Customizable sort icons for DataTable


bhavin-r-shah:
Can DataTable sort icons be customized?

Requirements:
1. Allow user to provide sort icon. 3 options - sort ascending, sort descending and icon to show column is sortable but not currently sorted upon.
2. Allow user to provide background color for the sorted column

Grommet version used - v2.15.0

halocline:
@bhavin-r-shah Sort icons may be defined via the theme at `theme.dataTable.icons`.

https://github.com/grommet/grommet/blob/03136c5e1a5fc55b20b41ddeb3162d2c096d25a9/src/js/themes/base.js#L683

This will take care of the "sort ascending" and "sort descending".

----------------------------------------

For "sortable, but not currently sorted upon," this would be a feature enhancement. The sort Icon is being defined here: https://github.com/grommet/grommet/blob/dc3ad5cbaf99c64c01a6c987c472f8bf2141e6d7/src/js/components/DataTable/Header.js#L69
This definition would need to be expanded to consider the scenario for when `sort.property !== property`. Additional considerations would be:
- Does the implementer have the ability to choose whether or not this "sortable" icon is shown?
- Does the implementer have the ability to specify when this "sortable" icon is shown? (e.g. always, on hover, etc.)
@ShimiSun I can create a feature request issue for this if you would like.

------------------------------------

For "allow user to provide background color for the sorted column," this would also be a feature enhancement. The DataTable's sort state would need to be passed to its Body / Body's Cell (Header already receives this) and then styling for the Cell's would need be added based on the `sort.property` state.
@ShimiSun I can create a feature request issue for this if you would like.

ShimiSun:
Thank you @halocline, I think your suggestion of the existing icons API should work 
https://github.com/grommet/grommet/blob/03136c5e1a5fc55b20b41ddeb3162d2c096d25a9/src/js/themes/base.js#L683

Regarding the 3rd option icon: 
>...to show column is sortable but not currently sorted upon.
We would stick with the simple design of not showing an icon at all.
@bhavin-r-shah feel free to share screenshots that your team/designer thinks it will be useful to use a sortable icon before we'll add enhancements to the code base.

Regarding:
> Allow user to provide background color for the sorted column

We would like to see a design for that, that demonstrates the required behavior.

PR #4480, opened 2020-09-01
Changed DataTable to add theme dataTable.icons.sortable

ericsoderberghp:
#### What does this PR do?

Changed DataTable to add theme dataTable.icons.sortable

Header.js

Added story

story

https://github.com/grommet/grommet/issues/4446

automatic

yes

backwards compatible


src/js/components/DataTable/Header.js
```diff
@@ -66,14 +66,17 @@ const Header = ({
...
-              const Icon =
-                onSort &&
-                sortable !== false &&
-                sort &&
-                sort.property === property &&
-                theme.dataTable.icons[
-                  sort.direction !== 'asc' ? 'ascending' : 'descending'
-                ];
+              let Icon;
+              if (onSort && sortable !== false) {
```
ShimiSun, line 70: I think the icon should still work if `sortable` isn't false and `onSort` wasn't defined.

src/js/components/DataTable/doc.js
```diff
@@ -329,6 +329,11 @@ export const themeDoc = {
     type: 'React.Element',
     defaultValue: '<FormDown />',
   },
+  'dataTable.icons.sortable': {
+    description: 'The icon indicating a column can be sorted.',
```
ShimiSun, line 333: I'm a little confused about whether the icon will be placed in addition to the up/down icon.
Edit: after looking on storybook, we should add a comment that this is an interim icon until the sort action will take place.

src/js/components/DataTable/stories/Custom.js
```diff
@@ -0,0 +1,44 @@
...
+
+const customTheme = {
+  global: {
+    font: {
+      family: 'Helvetica',
+    },
+  },
+  dataTable: {
+    icons: {
+      sortable: FormNext,
```
ShimiSun, line 19: I personally found this icon to be confusing and too similar to the Next/Prev icons. can we use a different icon that will stand out?

src/js/components/DataTable/Header.js
ericsoderberghp: The point of the icon is to indicate that the user could sort on that column. We don't want to show the icon if the user cannot sort on the column.

src/js/components/DataTable/doc.js
ericsoderberghp: The icon is intended as an indicator that the column could be sorted. So, it's not exactly "interim". The user might never sort, and that's just fine.

src/js/components/DataTable/stories/Custom.js
ericsoderberghp: I scanned through grommet-icons looking for something that might work, to no avail. :(
Perhaps we should add an icon to grommet-icons?

src/js/components/DataTable/Header.js
ShimiSun: I think my point is that one can use `sortable` prop without `onSort` and still expect the new icon functionality to work as expected. For example, this story https://storybook.grommet.io/?path=/story/datatable--controlled-grouped
uses only `sortable` prop, without `onSort` and I can see a use case of it using the new sortable icon.

src/js/components/DataTable/stories/Custom.js
ShimiSun: https://github.com/grommet/grommet-icons/issues/207

src/js/components/DataTable/doc.js
ShimiSun:
```suggestion
    description: 'The icon indicating a column can be sorted. The icon will be replaced by the ascending / descending sort icons once clicked upon.',
```

ericsoderberghp: But, then I think we should also say something about it being restored when another column is sorted. We don't currently describe that level of detail for `dataTable.icons.ascending`, where we just say "The ascending icon.". Are you thinking we should document when the icons are used and then removed for all of them?

src/js/components/DataTable/stories/Custom.js
ericsoderberghp: In the meantime, I've built a custom icon into the story.

src/js/components/DataTable/Header.js
ShimiSun: Thank you for clarifying.

src/js/components/DataTable/doc.js
ShimiSun: >But, then I think we should also say something about it being restored when another column is sorted

I agree. This level of detail might end up confusing.

src/js/components/DataTable/stories/Custom.js
ShimiSun: And it looks great!

ShimiSun (approved):

PR #4490, opened 2020-09-02
Align option added to Drop's stretch prop

### LINKED ISSUE #2428: v2 Select: optionally tell Drop within select to respect the width of the main Select target
Today if the content inside the Drop is bigger than the Target, the drop will expand. it would be nice to have a way to tell the Drop to wrap and fix position so that the Select options are always aligned to the end of the target.

IanKBovard:
#### What does this PR do?
Adds a constrain prop to select component that restricts the width of the drop to the width of the select.

DropContainer.js
DropButton.js
Select.js

Manual testing. Storybook

Storybook

closes #2428 

Default behavior - w/o constrain :
Screen Shot 2020-09-02 at 5 39 07 PM

with constrain :
Screen Shot 2020-09-02 at 5 39 13 PM

Yes

No

backwards compatible


IanKBovard:
I will update the docs, add a story, and a test once this solution is approved. Just not sure if we want to take this approach.

ShimiSun:
Does the terminology `fit` would work better? It will be more grommet consistent.

ericsoderberghp:
Two thoughts:
- I think we should connect this up to the `stretch` prop in Drop somehow, maybe `stretch="align"`? Open to exploring other values.
- I think we should do this on Drop and then rely on `dropProps` of Select, and Menu, rather than making it a formal property of Select.

IanKBovard:
> * stretch

Ill make those changes

IanKBovard:
changes made.
stretch prop now has align.
docs updated.

ericsoderberghp (dismissed):

ericsoderberghp:
Can we also add `stretch="align"` to the unit tests?

ericsoderberghp:
I had been thinking we'd update Drop-test.js, where we have a test for the `stretch` prop.

ericsoderberghp (approved):

ShimiSun (approved):

ShimiSun:
Thank you @IanKBovard !

PR #4494, opened 2020-09-03
fix: remove hover indicator of disabled Styled Button

abnersajr:
#### What does this PR do?

This PR disables hover styles of Styled Button when it is disabled.

#### Where should the reviewer start?

The only one file in the PR

#### What testing has been done on this PR?

Local environment using a  ` theme.button.default` 

#### How should this be manually tested?

Manually test adding a ` theme.button.default`  and add a button with kind

#### Any background context you want to provide?

https://github.com/grommet/grommet-theme-hpe/pull/110

#### What are the relevant issues?

#4492 
https://github.com/grommet/grommet-theme-hpe/pull/110

#### Do the grommet docs need to be updated?
No

#### Should this PR be mentioned in the release notes?
No

#### Is this change backwards compatible or is it a breaking change?
No

Issue #4492, opened 2020-09-03
Button - kind and disabled button doesn't apply when hover and plain


ShimiSun:
### Expected Behavior
`disabled` styles should apply when `hoverindicator` or `plain` is set on a grommet button.

### Actual Behavior
No disabled styles applied.

### URL, screen shot, or Codepen exhibiting the issue
Use this code when using the kind button, and see how disabled styles don't apply.
```        
<Button
          label="Disabled"
          disabled
          hoverIndicator
          plain
          icon={<FormNext />}
        />
```
Related to https://github.com/grommet/grommet-theme-hpe/pull/110

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.15.0
- Browser Name and version:
- Operating System and version (desktop or mobile):

abnersajr:
@ShimiSun This PR can be closed. 🎉

ShimiSun:
Thanks @abnersajr ❤️ I think there is still an issue with how `plain` Button (kind) doesn't apply disabled properly, would you like to tackle it?

leossantiago:
@ShimiSun did a quick overview and it looks like plain Button is applying disabled properly aswell.

ShimiSun:
@leonardodino thanks buddy, from what I've seen, Button + plain + disabled + kind (kind is when button.default is defined on the theme), the plain button doesn't get the text opacity or the `text-weak` property, so the Button `disabled` text isn't reflecting the disabled property well.

taysea:
After investigating this before, I've come to a couple conclusions.
1. When a Button is `plain` and `disabled`, the only disabled styling comes from [this line](https://github.com/grommet/grommet/blob/master/src/js/components/Button/StyledButtonKind.js#L183):
```
${props => props.disabled && disabledStyle(props.theme.button.disabled.opacity)}
```
where `disabledStyle` only controls `opacity`.
2. When a theme defines disabled opacity of `1.0` via `props.theme.button.disabled.opacity`, then the disabled, plain button will have an opacity of 1.0 (meaning that it will appear that same as it does when it's enabled. 
3. For themes that defined disabled opacity as 1.0, they are typically using other means to indicate disabled state such as shifting the text color from `text` to `text-weak`. However, this shift in styling does not get applied to the plain buttons.

Approaches to fix:
1. Allow `plain` to be added to button theming with an opacity option (`props.theme.button.disabled.plain.opacity`) where this would override the general button disabled opacity.
[...]

PR #4497, opened 2020-09-03
Added test for disabled button with hover indication

ShimiSun:
Complimentary test to https://github.com/grommet/grommet/pull/4494
Issue #4492

On the unit test, I've verified that after the fix the hover indication background color is being removed.


src/js/components/Button/__tests__/__snapshots__/Button-kind-test.js.snap
```diff
@@ -311,6 +311,98 @@ exports[`Button kind custom theme secondary 1`] = `
...
+exports[`Button kind disabled with hoverIndicator should not hover 1`] = `
+.c1 {
+  display: inline-block;
+  box-sizing: border-box;
+  cursor: pointer;
+  font: inherit;
+  -webkit-text-decoration: none;
+  text-decoration: none;
+  margin: 0;
+  background: transparent;
```
taysea, line 323: Just to confirm if this test had existed, without the fix for grommet, this would have received a background color and now it's demonstrating that no background is applied on hover?

ShimiSun: Good question, the test didn't exist before, I added the test locally before applying the fix, and after applying the fix I've verified that the background color was removed as it should.

taysea: Great. Thanks!

taysea (approved):
Thanks for adding this. LGTM.

PR #4502, opened 2020-09-04
Removed some stories from chromatic testing

jcfilben:
#### What does this PR do?
Removes the following stories because snapshots weren't valuable & trying to decrease chromatic numbers

Original list:
**Accordion**
Simple - decided to keep this story because previously we removed 'Dark no animation' and 'Multiple' because they were similar to this story. This story is covered by jest testing but I think that it might be good to have a chromatic test.

**Anchor**
Default

**Carousel**
Simple

**Form**
Controlled input
Controlled lazy

**List**
onClickItem
Selection - kept selection instead of onClickItem

**MaskedInput**
Date
Filtered
IPv4 address
Size + units
URL

**Select**
Multiple

**TextInput**
onSelect and onSuggestionSearch
Simple

**WorldMap**
Simple

Number of chromatic snapshots should decrease by 14.

No

No

Backwards compatible

### DIFF
--- a/src/js/components/Anchor/stories/Default.js
+++ b/src/js/components/Anchor/stories/Default.js
@@ -14,4 +14,6 @@ const Default = () => {
   );
 };
 
-storiesOf('Anchor', module).add('Default', () => <Default />);
+storiesOf('Anchor', module).add('Default', () => <Default />, {
+  chromatic: { disable: true },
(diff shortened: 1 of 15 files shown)


ShimiSun (approved):
GREAT!! thank you!

PR #4503, opened 2020-09-04
Decouple documentation modules from functional modules (rework for current master)

MikeKingdom:
Update of PR #4072 to work with current master.
I created a new PR since the previous one wasn't associated with a branch.

From the original:

Currently modules from react-desc and some smaller utilities are leaking into the production bundle as well as downstream consumers of Grommet. I see that the roadmap lists finding an alternative for react-desc in 2020, so it is possible this aids in easing that transition when the time comes. This change results in approximately 11 KiB savings on the size of grommet.min.js.

## What does this PR do?
Separate the documentation modules from the utils index and import them into the doc.js modules directly. In this way, in production builds, no documentation-related code will appear.

## Where should the reviewer start?
See what was removed in src/js/utils/index.js and then open a doc.js file; the Box component one seems like a good one to get an idea of the changes.

## What testing has been done on this PR?
Build and test suites pass. Documentation is generated. On my downstream project I can consume the package with the changes and verified functionality, in addition to the disappearance of react-desc from that bundle.

## How should this be manually tested?
Since there is no functional change expected and only documentation changes. Manual verification should be focused there, if any.

## Any background context you want to provide?
None.

## What are the relevant issues?
No issue filed. Consider this pull request the issue.

## Screenshots (if appropriate)
Not necessary, unless you want to see console output.

## Do the grommet docs need to be updated?
They were, sort of :)

## Should this PR be mentioned in the release notes?
Yes. Branched off 2.13.0. The build for that release shows a bundle size of 434 KiB for grommet.min.js. This pull request reduces it to 423 KiB. My downstream project, after compression, sees a net savings of 4 KiB if that matters.

## Is this change backwards compatible or is it a breaking change?
This change is backwards-compatible and non-breaking.


ShimiSun:
Thank you for getting it to the finish line, and thank you @jecassis for the inspiration and initiative, fantastic PR!

ShimiSun (approved):

PR #4504, opened 2020-09-04
Fixes bug (#4334) when replace && show > step

halocline:

#### What does this PR do?

Fixes & closes https://github.com/grommet/grommet/issues/4334

InfiniteScroll.js

- In the scenario cited in the original issue, the `nextEndPage` calculation was setting this page to a value which did not contain the show index.
- The show index is now being considered when determining `nextEndPage`.

All InfiniteScroll Storybook stories, specifically "Replace with Show"

Current Jest tests need additional work to test for "replace with show" scenarios. Because they rely on JSDOM, the onScroll event is not available and difficult to emulate scroll behavior. Needs further investigation.

Storybook stories, specifically InfiniteScroll --> "Replace with Show"

https://github.com/grommet/grommet/issues/4334

No.

Yes.

Fixes bug. Restores intended functionality.


src/js/components/InfiniteScroll/stories/Replace.js
```diff
@@ -24,10 +24,20 @@ const InfiniteScrollReplace = props => (
 if (!isChromatic()) {
   storiesOf('InfiniteScroll', module)
     .add('Replace', () => <InfiniteScrollReplace replace />)
-    .add('Replace with show 28th item', () => (
-      <InfiniteScrollReplace replace show={27} />
-    ))
-    .add('Replace with show 88th item', () => (
-      <InfiniteScrollReplace replace show={87} />
+    .add('Replace with Show', () => (
```
ShimiSun, line 27:
```suggestion
    .add('Replace with show', () => (
```
We try to use sentence case on stories

ShimiSun (approved):
LGTM, Thank you! 
I'm guessing our tests for this specific scenario will only take place on storybook and not on jest? @jcfilben can you check if the chromatic testing is valuable for this storybook change as we try to optimize our chromatic utilization.

ericsoderberghp (approved):

jcfilben:
@ShimiSun I'm thinking the snapshot would trigger too early and not capture the 'show' behavior so this snapshot would not be valuable on chromatic. Currently chromatic is disabled for this story so I think the only way to actually test this would be to create another branch with it enabled.

Issue #4465, opened 2020-08-27
Ability to Specify RadioButton Check Size


halocline:
### Expected Behavior

Would like to be able to specify the check size of a RadioButton in my theme.

### Actual Behavior

My [design](https://www.figma.com/file/aZyY606ENQedz4FXugdzgS/HPE-Radio-Button-Group-Component?node-id=77%3A323) calls for a 16x16 check (r=8), but the default icon is 12x12 (r=6).

Related Issue: https://github.com/hpe-design/design-system/issues/851

[Proposal](https://github.com/hpe-design/design-system/issues/851#issuecomment-682100097) for something like `radioButton.check.size`

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.15.0 / Stable
- Browser Name and version:
- Operating System and version (desktop or mobile):

netochaves:
I open a PR for this #4505

ShimiSun:
Can we try leveraging the existing theme prop `theme.radioButton.icons.circle` instead of adjusting the radius on the code?

halocline:
> Can we try leveraging the existing theme prop `theme.radioButton.icons.circle` instead of adjusting the radius on the code?

@ShimiSun Ah, I remember why I abandoned this path initially. Initially, I had wanted to add 
```
icons: {
  circle: () => (
    <Blank color="selected-background">
      <circle cx="12" cy="12" r="8" />
    </Blank>
  ),
}
```
to the theme file. That worked great in a local theme, but did not translate well to `grommet-theme-hpe`. When trying to apply to that project, the following changes would need to be made to enable this:
- import React to the theme file
- 'react' then must be added as a dependency, most likely a peer dependency in this case.
- Additionally, we would then need to update our Babel configs, adding `@babel/preset-react` as a preset, so that Babel knows how to parse `index.js` as JSX.
- And then there was something else, that I can't remember.

At the time, this felt like a lot of overhead to get at what we needed to accomplish. Plus, felt that others might like this ability as well.

Thoughts on whether we want to pursue these types of changes on `grommet-theme-hpe`?

Another approach would be to add a larger circle to `grommet-icons`, but that seemed like a pretty acute need.

ShimiSun:
@halocline why not simply using SVG instead of the Blank react component?

halocline:
> @halocline why not simply using SVG instead of the Blank react component?

[...]

PR #4505, opened 2020-09-05
feat(RadioButton): add check size option to theme

netochaves:
#### What does this PR do?
Adds a new option to change the check size of RadioButton
`src/js/components/RadioButton/RadioButton.js`
Unit testing
With storybook
No
#4465 
Yes
Yes
It is backward compatible


src/js/components/RadioButton/__tests__/RadioButton-test.js
```diff
@@ -123,4 +124,38 @@ describe('RadioButton', () => {
...
+      };
+
+      const { container } = render(
+        <Grommet theme={customTheme}>
+          <RadioButton name="test" checked onChange={jest.fn()} />
+        </Grommet>,
+      );
+
+      expect(
+        container.querySelector('circle').hasAttribute('r', '6'),
```
jcfilben, line 157: On this second test should this be 8 instead of 6 since the check size is being set to 8 in the customTheme?
```
        container.querySelector('circle').hasAttribute('r', '8'),
```

src/js/components/RadioButton/README.md
```diff
@@ -166,6 +166,16 @@ Defaults to
 100%
 ```
 
+**radioButton.check.size**
+
+The check size of the RadioButton. Expects `string`.
```
jcfilben, line 171: Should this be `number` instead of `string`?

jcfilben (commented):
Great PR! just had a few questions

src/js/components/RadioButton/__tests__/__snapshots__/RadioButton-test.js.snap
```diff
@@ -402,7 +402,7 @@ exports[`RadioButton checked 1`] = `
           <circle
             cx={12}
             cy={12}
-            r={6}
+            r="6"
```
ShimiSun, line 405: we should be backward compatible with the types.

src/js/components/RadioButton/doc.js
```diff
@@ -91,6 +91,11 @@ export const themeDoc = {
     type: 'string',
     defaultValue: '100%',
   },
+  'radioButton.check.size': {
+    description: 'The check size of the RadioButton.',
+    type: 'string',
```
ShimiSun, line 96: would this support t-shirt sizing of small, medium, large?

src/js/components/RadioButton/__tests__/RadioButton-test.js
netochaves: Good catch, also this should be failing because of that. Thanks

src/js/components/RadioButton/README.md
netochaves: Yeah, `number` makes more sense than `string`. thanks.

src/js/components/RadioButton/doc.js
netochaves: I changed the type to `number` so isn't being supported, but if we change the type to  `number | string` this can be implemented. What do you think?

src/js/components/RadioButton/__tests__/RadioButton-test.js
ShimiSun: Sounds fishy that the test passed when it was size, what am I missing?

src/js/components/RadioButton/RadioButton.js
```diff
@@ -102,7 +102,7 @@ const RadioButton = forwardRef(
                     viewBox="0 0 24 24"
                     preserveAspectRatio="xMidYMid meet"
                   >
-                    <circle cx={12} cy={12} r={6} />
+                    <circle cx={12} cy={12} r={theme.radioButton.check.size} />
```
ericsoderberghp, line 105: I know this is being picky, but I wonder if `radius` would be more appropriate here? We aren't using any of our string T-shirt sizes and I think pretty much everywhere else, `size` would be used for the width and height.

ericsoderberghp (commented):
Just one small terminology comment. Looking good otherwise.

src/js/components/RadioButton/RadioButton.js
netochaves: I agree with @ericsoderberghp, we're talking about a circle element so it makes sense to be `radius` if @ShimiSun or @jcfilben don't have a different thought I will change it.

src/js/components/RadioButton/doc.js
ShimiSun: I think that if we stick with eric's comment below it will naturally take care of it.

src/js/components/RadioButton/RadioButton.js
netochaves: Hey, I found a problem with this rename, we already have a `radioButton.check.radius` that it's for `border-radius` of `check`, So how do you guys want to proceed with this?  
https://github.com/grommet/grommet/blob/85e3d85896327d9e6bf96ca5e6e307bd4793fccc/src/js/components/RadioButton/doc.js#L89

ShimiSun: Good catch. I agree it's a little tricky, after putting some thoughts into this issue, I wish we had defined the current api with `radioButton.check.border.radius` instead of `radioButton.check.radius`. but from backward compatibility reasons, it will be tricky to override it now while keeping a backward compatibility behavior.

I think the best way to enhance the api while keeping a clean implementation and backward compatible behavior is to use `radioButton.check.size.radius`.

@jcfilben & @ericsoderberghp any thoughts on the new api suggestions?

jcfilben: I agree with Shimi that the best thing to do here is to use `radioButton.check.size.radius`

jcfilben:
Hi @netochaves great progress so far, any updates on this PR?

netochaves:
Hi @jcfilben I changed the name for `theme.check.size.radius` as we discussed.

src/js/components/Calendar/__tests__/Calendar-test.js
```diff
@@ -10,7 +10,7 @@ import { cleanup, fireEvent, render, act } from '@testing-library/react';
 import { FormNextLink, FormPreviousLink } from 'grommet-icons';
 import { Box, Button, Calendar, Grommet, Text } from '../..';
 
-const DATE = '2020-01-15T00:00:00-08:00';
+const   DATE = '2020-01-15T00:00:00-08:00';
```
ShimiSun, line 13: Where is this space is coming from?

src/js/components/Calendar/__tests__/__snapshots__/Calendar-test.js.snap
```diff
@@ -2487,14 +2487,14 @@ exports[`Calendar children 1`] = `
             className="c5"
             size="medium"
           >
-            September 2020
+            October 2020
```
ShimiSun, line 2490: This change will be fixed once merging in with the master branch.

src/js/themes/base.d.ts
```diff
@@ -954,6 +954,9 @@ export interface ThemeType {
       background?: {
         color?: ColorType;
       };
+      size?: {
+        radius: number;
```
ShimiSun, line 958:
```suggestion
        radius?: number;
```

ShimiSun (commented):
Looking good, dropped a couple of small comments.

src/js/themes/base.d.ts
netochaves: Thanks

src/js/components/Calendar/__tests__/__snapshots__/Calendar-test.js.snap
netochaves: Ok, I'll remove it.

src/js/components/Calendar/__tests__/Calendar-test.js
netochaves: I'll remove it.

jcfilben:
Try merging with master and then pushing that commit, I think that should solve the failing checks on this PR

ShimiSun:
The need for this PR is discussed in #4465. Stay tuned.

src/js/components/RadioButton/stories/Theme.js
```diff
@@ -18,8 +18,8 @@ const theme = deepMerge(grommet, {
       color: {
         light: 'neutral-1',
       },
-      background: {
-        color: 'red',
```
jcfilben, line 22: Why is the background color removed from the theme in this story?

jcfilben (commented):
I think this PR is very close to being done, just added a small comment below

ShimiSun:
I'll be closing this PR since we found an alternative way to approach a fix for the original issue, thank you for helping the discussion with this PR 💟

PR #4512, opened 2020-09-09
Add plain='full' option to TextInput

### LINKED ISSUE #4078: TextInput plain doesnt work?
TextInput plain doesn't appear to do anything.

### Expected Behavior

Remove padding

### Actual Behavior

Doesn't remove the padding

https://codesandbox.io/s/textinput-plain-issue-jexeu

### Your Environment

- Grommet version: 2.13.0
- Browser Name and version: Latest Chrome
- Operating System and version (desktop or mobile): both

jcfilben:
#### What does this PR do?
Previously on TextInput when the plain prop was set there was still 11px of padding. Demonstrated in this codesandbox: https://codesandbox.io/s/textinput-plain-issue-forked-w4t6q?file=/index.js:285-333

Changing plain to have `padding: 0` would cause backwards compatibility issues. The most notable ones being on Form when there is a textInput. Notice in the screenshot below how the padding was affected around some of the fields ('initial', 'initial@my.com', and 'large') and the Form doesn't look right with these changes. When TextInput is contained in a form the plain prop is set to true, which is why there are so many snapshot changes for Form with this change.
Screen Shot 2020-09-09 at 11 07 09 AM

**previously the form looked like this:**

Screen Shot 2020-09-09 at 11 18 44 AM

To prevent backwards compatibility issues, plain='full' was added as an option to remove the padding. Plain={true} maintains its original behavior.

storybook testing with the following code:
```javascript
import React from 'react';

import { Grommet, TextInput } from 'grommet';

const SimpleTextInput = () => {
  return (
          <TextInput
            plain='full'
            placeholder="Its plain, should not have padding"
            name="name"
            type="text"
          />
    </Grommet>
  );
};

storiesOf('TextInput', module).add('TEST plain', () => <SimpleTextInput />, {
  chromatic: { disable: true },
});

```

Closes #4078 

Yes, updated in this PR

Yes

Backwards compatible

### DIFF
--- a/src/js/components/TextInput/README.md
+++ b/src/js/components/TextInput/README.md
@@ -183,8 +183,9 @@ node
 **plain**
 
 Whether this is a plain input with no border or outline.
-      Only use this when the containing context provides sufficient affordance.
       Use "full" to remove padding in addition to removing border and outline.
+      Only use this when the containing context provides sufficient 
+      affordance.
 
 ```
 boolean
--- a/src/js/components/TextInput/README.md
+++ b/src/js/components/TextInput/README.md
@@ -183,10 +183,12 @@ node
 **plain**
 
 Whether this is a plain input with no border or padding.
-      Only use this when the containing context provides sufficient affordance
+      Only use this when the containing context provides sufficient affordance.
+      Use "full" to remove padding in addition to removing border and outline.
 
 ```
 boolean
+full
 ```
 
 **reverse**
--- a/src/js/components/TextInput/README.md
+++ b/src/js/components/TextInput/README.md
@@ -182,7 +182,7 @@ node
 
 **plain**
 
-Whether this is a plain input with no border or padding.
+Whether this is a plain input with no border or outline.
       Only use this when the containing context provides sufficient affordance.
       Use "full" to remove padding in addition to removing border and outline.
 
--- a/src/js/components/TextInput/StyledTextInput.js
+++ b/src/js/components/TextInput/StyledTextInput.js
@@ -13,9 +13,7 @@ import { defaultProps } from '../../default-props';
 const getPlainStyle = plain => {
   if (plain === 'full') {
     return css`
-      outline: none;
-      border: none;
-      padding: 0;
+      ${plainInputStyle} padding: 0;
     `;
   }
   return plain && plainInputStyle;
--- a/src/js/components/TextInput/StyledTextInput.js
+++ b/src/js/components/TextInput/StyledTextInput.js
@@ -5,13 +5,13 @@ import {
   getInputPadBySide,
   inputStyle,
   parseMetricToNum,
-  plainInputStyle,
+  getPlainStyle,
 } from '../../utils';
 import { defaultProps } from '../../default-props';
 
 const StyledTextInput = styled.input`
   ${inputStyle}
-  ${props => props.plain && plainInputStyle}
+  ${props => getPlainStyle(props.plain)}
   ${props =>
     props.icon &&
     (props.reverse
--- a/src/js/components/TextInput/StyledTextInput.js
+++ b/src/js/components/TextInput/StyledTextInput.js
@@ -1,6 +1,5 @@
 import styled, { css } from 'styled-components';
 
-import { PropTypes } from 'react-desc';
 import {
   disabledStyle,
   getInputPadBySide,
@@ -18,11 +17,6 @@ const getPlainStyle = plain => {
   }
   return plain && plainInputStyle;
 };
-
-const plainPropType = PropTypes.oneOfType([
-  PropTypes.bool,
-  PropTypes.oneOf(['full']),
-]);
 
 const StyledTextInput = styled.input`
   ${inputStyle}
@@ -111,5 +105,4 @@ export {
   StyledPlaceholder,
   StyledIcon,
   StyledSuggestions,
-  plainPropType,
 };
--- a/src/js/components/TextInput/StyledTextInput.js
+++ b/src/js/components/TextInput/StyledTextInput.js
@@ -1,13 +1,30 @@
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 
+import { PropTypes } from 'react-desc';
 import {
   disabledStyle,
   getInputPadBySide,
   inputStyle,
   parseMetricToNum,
-  getPlainStyle,
+  plainInputStyle,
 } from '../../utils';
 import { defaultProps } from '../../default-props';
+
+const getPlainStyle = plain => {
+  if (plain === 'full') {
+    return css`
+      outline: none;
+      border: none;
+      padding: 0;
+    `;
+  }
+  return plain && plainInputStyle;
+};
+
+const plainPropType = PropTypes.oneOfType([
+  PropTypes.bool,
+  PropTypes.oneOf(['full']),
+]);
 
 const StyledTextInput = styled.input`
   ${inputStyle}
@@ -96,4 +113,5 @@ export {
   StyledPlaceholder,
   StyledIcon,
   StyledSuggestions,
+  plainPropType,
 };
--- a/src/js/components/TextInput/__tests__/__snapshots__/TextInput-test.js.snap
+++ b/src/js/components/TextInput/__tests__/__snapshots__/TextInput-test.js.snap
@@ -2214,6 +2214,7 @@ exports[`TextInput select suggestion 1`] = `
   border-radius: 4px;
   outline: none;
   border: none;
+  padding: 0px;
 }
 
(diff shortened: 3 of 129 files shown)


src/js/utils/styles.js
```diff
@@ -458,4 +458,5 @@ export const sizeStyle = (name, value, theme) => css`
 export const plainInputStyle = css`
   outline: none;
   border: none;
+  padding: 0px;
```
ShimiSun, line 461:
```suggestion
  padding: 0;
```

ShimiSun (commented):
I understand your concerns with this PR, and I need to think about whether we should apply it and how to communicate it on the release notes.

taysea:
To avoid these backwards compatibility issues, would it make sense in FormField to retroactively add back the padding to TextInput that is being taken away by `plain`?

jcfilben:
@taysea I think that is definitely a possible solution. The person who originally filed an issue for this #4078 was using textInput within a FormField and they wanted to be able to have the plain option for textInput. So maybe when you do
```javascript
<FormField>
  <TextInput>
</FormField>
```
It would add the padding like you mentioned but if you did
```javascript
<FormField>
  <TextInput plain>
</FormField>
```
It would not have the padding

taysea:
@jcfilben Gotcha. Yeah this is tricky because it seems like Grommet automatically applies `plain` to input components when they're in a form (https://github.com/grommet/grommet/blob/master/src/js/components/FormField/FormField.js#L98). Now that we're stripping out the padding with your changes (which I think makes sense conceptually to align with the docs definition), we're seeing a misalignment with the rest of the form text / inputs. The issue of aligning with the docs here is that it feels like we are going to be introducing lots of visual regressions for users.

I'm trying to think about all the changes that might result for people who are currently using `<TextInput plain />` based on this, and it feels difficult to be able to anticipate all the ways grommet users may be using the component.

Another thought around these changes:
- What if I just want to remove the border for a single TextInput (not style my entire theme to have no border) -- how would I do this via grommet or would I need to rely on `style={{ ... }}` which we typically try to avoid?

I know, for example, we are using a `plain` TextInput for the search input on https://design-system.hpe.design/ and we wouldn't want to lost the padding here. Just something to consider around not taking away functionality from users.

ericsoderberghp:
I agree with the comments that we can't make this change due to the compatibility issues. I can see a couple of ways to address the issue:
- connect it to something in the theme such that the `plain` prop removes padding, such as `theme.textInput.plain = 'full';`
- add another value, such as `plain='full'`

ShimiSun:
I'm OK with proceeding with the suggestion above of
> add another value, such as plain='full'

jcfilben:
**Update:** I went ahead with Eric's suggestion and added plain='full'. I added snapshot tests to TextInput to test that plain='full' removes padding and plain={true} still has padding. I updated the docs to included explanation for plain='full'.

This change fixes the compatibility issue, the snapshots are no longer being affected and I think it is now safe to merge

src/js/components/TextInput/README.md
```diff
@@ -183,10 +183,12 @@ node
 **plain**
 
 Whether this is a plain input with no border or padding.
-      Only use this when the containing context provides sufficient affordance
+      Only use this when the containing context provides sufficient affordance.
```
ShimiSun, line 186: I think the first line isn't accurate if it's referring to the boolean value
> Whether this is a plain input with no border or padding.
Should it be something like:
```
Whether this is a plain input with no border or outline.
Only use this when the containing context provides sufficient affordance.
Use "full" to remove the padding in addition to removing the border and outline.
```
Thoughts?

src/js/components/TextInput/doc.js
```diff
@@ -1,5 +1,6 @@
 import { describe, PropTypes } from 'react-desc';
 
+import { plainPropType } from '../../utils/prop-types';
```
ShimiSun, line 3: The `plain="full"` is very specific for TextInput and not something we would like to encourage in other components (with the thought of keeping the API simple), hence there is no need to extract the logic to the utils file.

src/js/utils/styles.js
```diff
@@ -459,3 +459,14 @@ export const plainInputStyle = css`
   outline: none;
   border: none;
 `;
+
+export const getPlainStyle = plain => {
+  if (plain === 'full') {
```
ShimiSun, line 464: Along the same lines, this behavior is very specific to TextInput hence the logic should be placed TextInput.

src/js/components/TextInput/README.md
ShimiSun: @jcfilben where are we with this comment?

jcfilben: I updated the readme to say 'no border or outline.' instead of 'no border or padding'

src/js/components/TextInput/doc.js
jcfilben: Moved plainPropType to just be in TextInput

src/js/utils/styles.js
jcfilben: I moved getPlainStyle to just be in TextInput

src/js/components/TextInput/StyledTextInput.js
```diff
@@ -9,9 +10,25 @@ import {
 } from '../../utils';
 import { defaultProps } from '../../default-props';
 
+const getPlainStyle = plain => {
+  if (plain === 'full') {
+    return css`
+      outline: none;
+      border: none;
+      padding: 0;
+    `;
```
ShimiSun, line 19: we should leverage the existing `plainInputStyle` instead of specifying a separate css.

ShimiSun (commented):
We should update the *.d.ts (typescript file of each component) to include the `full` value of the plain prop.

jcfilben:
Update: added plain: full option to index.d.ts. Github is showing lots of changes for this file, this is because of prettier. Only real change was line 32 `plain?: boolean | 'full';`

src/js/components/TextInput/StyledTextInput.js
jcfilben: Updated to use `plainInputStyle`
```javascript
return css`
      ${plainInputStyle} padding: 0;
    `;
```

src/js/components/TextInput/StyledTextInput.js
```diff
@@ -9,9 +10,23 @@ import {
...
+  }
+  return plain && plainInputStyle;
+};
+
+const plainPropType = PropTypes.oneOfType([
+  PropTypes.bool,
+  PropTypes.oneOf(['full']),
+]);
+
```
ShimiSun, line 26: Why are propTypes defined on the Styled file? this is not consistent with other components, it is usually kept on the doc.js file.

src/js/components/TextInput/doc.js
```diff
@@ -93,9 +94,10 @@ export const doc = TextInput => {
     placeholder: PropTypes.node.description(
       'Placeholder to use when no value is provided.',
     ),
-    plain: PropTypes.bool.description(
-      `Whether this is a plain input with no border or padding.
-      Only use this when the containing context provides sufficient affordance`,
+    plain: plainPropType.description(
+      `Whether this is a plain input with no border or outline.
+      Only use this when the containing context provides sufficient affordance.
+      Use "full" to remove padding in addition to removing border and outline.`,
```
ShimiSun, line 100: I think the context will make more sense if lines 2 & 3 could be flipped.

src/js/components/TextInput/StyledTextInput.js
jcfilben: Moved propTypes to doc.js in latest push

src/js/components/TextInput/doc.js
jcfilben: Switched these lines in latest push

src/js/components/TextInput/__tests__/__snapshots__/TextInput-test.js.snap
```diff
@@ -2640,6 +2717,84 @@ exports[`TextInput should not have accessibility violations 1`] = `
...
+.c2 {
+  box-sizing: border-box;
+  font-size: inherit;
+  font-family: inherit;
+  border: none;
+  -webkit-appearance: none;
+  background: transparent;
+  color: inherit;
+  width: 100%;
+  padding: 11px;
```
ShimiSun, line 2740: How come the snapshot shows the TextInput still have padding when plain is full.

jcfilben: The padding is originally set to 11px but then modified to be 0. Look at line 2745 in the snapshot: `padding: 0;`

ShimiSun (approved):

PR #4513, opened 2020-09-09
Added font to theme for Paragraph and Text

### LINKED ISSUE #4225: More theme support for customization of font family per component
### Expected Behavior

More components in the theme object should support customization by specifying a `family` property for those components. I was attempting to customize `List` while a coworker is trying to customize `Anchor`. Using the extend function is a workaround for now.

### Actual Behavior

Only `Heading` and `Button` support `family` on the theme object.

### Steps to Reproduce

1. 2. 3. N/A

### Your Environment

- Grommet version: v2.x
- Operating System and version (desktop or mobile): N/A

jcfilben:
#### What does this PR do?
Adds font to the theme for both Paragraph and Text

Storybook
```javascript
import React from 'react';

import { Grommet, Text, Paragraph } from 'grommet';

    text: {
        font: {
            family: "Comic Sans MS"
        },
    },
    paragraph: {
        font: {
            family: "Comic Sans MS"
        },
    },
  });

const All = () => (
  <Grommet theme={customTheme}>
    <Text>text text text</Text>
    <Paragraph>paragraph paragraph paragraph</Paragraph>
  </Grommet>
);

storiesOf('Text', module).add('TEST', () => <All />);
```

Closes #4225 

Updated in this PR

Yes

Backwards compatible


src/js/components/Paragraph/doc.js
```diff
@@ -72,6 +72,11 @@ export const themeDoc = {
       },
     }`,
   },
+  'paragraph.font': {
```
ShimiSun, line 75: Should be `paragraph.font.family`

src/js/components/Text/StyledText.js
```diff
@@ -46,6 +46,14 @@ const wordBreakStyle = css`
   word-break: ${props => props.wordBreak};
 `;
 
+const fontFamily = props => {
+  return props.theme.text.font
+    ? css`
+        font-family: ${props.theme.text.font.family};
+      `
+    : '';
```
ShimiSun, line 54: This empty string will change our default behavior. 
We should default to the existing font family in case the font object doesn't exist instead of setting in it a new default.

src/js/components/Collapsible/__tests__/__snapshots__/Collapsible-test.js.snap
```diff
@@ -65,7 +65,7 @@ exports[`Collapsible onClick open default 2`] = `
     style="max-height: 0px;"
   >
     <span
-      class="StyledText-sc-1sadyjn-0 hUokoe"
+      class="StyledText-sc-1sadyjn-0 dqKKpZ"
```
ShimiSun, line 68: This change usually means we changed a default behavior (which will be a potential backward compatibility issue). I believe that if you address my comment below on maintaining the default value it will fix the snapshot changes.

src/js/components/Text/StyledText.js
jcfilben: I am still seeing the same snapshot changes when I remove all behavior from
`const fontFamil = props => {` (function at line 49 of StyledText.js). I don't think the snapshot changes are happening because of this. I think that the snapshot class name changes when `$props => fontFamily(props)` is added at line 78. So I think the snapshot changes are unavoidable.

As far as changing the default behavior goes I modeled this code off of what was already done in StyledHeading.js. I don't think this code is changing the default behavior. If `props.theme.text.font` is not specified the default font is still showing up.

I could change the function to be:
```javascript
const fontFamily = props => {
    return props.theme.text.font;
}   
...
${props => {if(props.theme.text.font.family){fontFamily(props)}}}
```

ShimiSun: I think something close to
```
${props => { props.theme.text.font && props.theme.text.font.family && fontFamily(props) }
```
Would be cleanest

jcfilben: Changed to:

```javascript
const fontFamily = css`
  font-family: ${props => props.theme.text.font.family};
`;
...
${props =>
    props.theme.text.font &&
    props.theme.text.font.family &&
    fontFamily}
```

jcfilben:
Update: changed docs to say 'font family' instead of just 'font'. Also changed type t 'string' instead of 'object'. Added 'Theme' storybook story for Paragraph.

@ShimiSun In base.js should font be commented out for Paragraph and Text? (lines 887 and 1111)

ShimiSun:
>@ShimiSun In base.js should font be commented out for Paragraph and Text? (lines 887 and 1111)

I think we can keep it  as is (uncommented)

IanKBovard:
Just checking in on this pr. what was the conclusion about all the snapshot changes? Is it changing the default font-family? I think the yarn lock needs to be updated as well.

jcfilben:
The snapshot changes are being caused by adding `${props=>}` to StyledParagraph and StyledText
```javascript
${props =>
    props.theme.paragraph.font &&
    props.theme.paragraph.font.family &&
    fontFamily}
```
The snapshots are changed even when $props isn't making changes. For example:
```javascript
${props => console.log('test')}
```
still changes the snapshots. The snapshot changes are not happening because of the font.family

src/js/components/Paragraph/stories/Theme.js
```diff
@@ -0,0 +1,26 @@
...
+
+const All = () => (
+  <Grommet theme={customTheme}>
+    <Paragraph>
+      The font family for this paragraph is being defined by a custom theme.
+    </Paragraph>
+  </Grommet>
+);
+
+storiesOf('Paragraph', module).add('Theme', () => <All />, {
```
ShimiSun, line 24:
```suggestion
storiesOf('Paragraph', module).add('Themed', () => <All />, {
```
Please also change the name of the file.

src/js/components/Paragraph/stories/Theme.js
```diff
@@ -0,0 +1,26 @@
...
+  chromatic: { disable: true },
```
ShimiSun, line 25: why do you think it is OK to exclude the test from chromatic?

ShimiSun (commented):
Looking good, dropped small comments.

src/js/components/Paragraph/stories/Theme.js
jcfilben: Done, changed from Theme -> Themed

src/js/components/Paragraph/stories/Themed.js
```diff
@@ -0,0 +1,27 @@
...
+
+const All = () => (
+  <Grommet theme={customTheme}>
+    <Paragraph>
+      The font family for this paragraph is being defined by a custom theme.
+    </Paragraph>
+  </Grommet>
+);
+
+if (isChromatic && document.fonts) {
```
ShimiSun, line 25: What is `document.fonts` and why it is needed?

src/js/components/Paragraph/stories/Theme.js
jcfilben: The customTheme font is showing up in storybook but the chromatic snapshots are not capturing the font.

> Browsers can decide to render HTML in multiple passes when custom fonts are used. They do this to speed up the time-to-first-meaningful-paint.
> Unfortunately, this behavior can cause your story to render without the custom font.

found information about this here: https://www.chromatic.com/docs/resource-loading#loading-custom-fonts

src/js/components/Paragraph/stories/Themed.js
jcfilben: I added `document.fonts` as a workaround suggested[ here ](https://www.chromatic.com/docs/resource-loading#loading-custom-fonts) to try and get the chromatic snapshots to show the font but the snapshot still is not capturing the font

src/js/components/Paragraph/stories/Theme.js
ShimiSun: Hello

ShimiSun: Test

jcfilben:
Update: ended up disabling chromatic on the story because font wasn't being captured in the snapshot

ShimiSun (approved):

PR #4521, opened 2020-09-15
Fix DataTable `background` prop + theme incompatibilities and enhance theme options

taysea:
#### What does this PR do?

This PR does two things:
1) solves a bug where DataTable background prop of `pinned` was incompatible with theme defined values
2) expands this fix to allow `pinned` to be styled via the theme as well

With regards to the first point noted above, the background prop of DataTable accepts an object where the user can define:
```
background={{
   pinned: ...,
   body: ...,
   footer: ...,
   header: ...
}}
```

When just `background.pinned` was defined on the props, then any theme values for `table.background.header` or `table.background.footer` were being ignored because DataTable was improperly passing `background.pinned` to these cells and they did not know how to resolve the value.

In resolving this issue, it became apparent that `pinned` styling for different DataTable cells could be a theme enhancement. After discussion with @ericsoderberghp, we came to a conclusion that the `pinned` object would living under the `dataTable` object of the theme. The API is as follows (where each pinned area can have a background defined or extend):
```
dataTable: {
   pinned: {
      body: {},
      header: {},
      footer: {},
   },
}
```

src/js/components/DataTable/DataTable.js

Added a jest test and adjusted the Fill and pin story.

Define background colors in theme for `table.header.background.color`/ `dataTable.pinned.header.background.color` or `table.footer.background.color`/`dataTable.pinned.footer.background.color` and then define a prop on background with `background={{ pinned: [COLOR VALUE] }}`.

Prior to these changes, having `background={{ pinned: [COLOR VALUE] }}` defined on the `DataTable` would lead to any theme values being ignored for header or footer (causing these table sections to have undefined backgrounds where, on scroll of the table, the data was visible through the header).

Closes https://github.com/hpe-design/design-system/issues/1121

Yes, they are updated here.

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/DataTable/Cell.js
+++ b/src/js/components/DataTable/Cell.js
@@ -14,7 +14,7 @@ const normalizeProp = (name, rowProp, prop) => {
 };
 
 const Cell = ({
-  background,
+  background: backgroundProp,
   border,
   column: { align, pin: columnPin, property, render, verticalAlign, size },
   datum,
@@ -46,6 +46,11 @@ const Cell = ({
   if (cellPin) pin = cellPin;
   else if (columnPin) pin = ['left'];
 
+  let background;
+  if (pin && theme.dataTable.pinned && theme.dataTable.pinned[context]) {
+    background = theme.dataTable.pinned[context].background;
+  } else background = undefined;
+
   return (
     <StyledDataTableCell
       scope={scope}
@@ -54,13 +59,15 @@ const Cell = ({
       context={context}
       verticalAlign={verticalAlign}
       size={size}
-      background={normalizeProp(
-        'background',
-        rowProp,
-        Array.isArray(background)
-          ? background[index % background.length]
-          : background,
-      )}
+      background={
+        normalizeProp(
+          'background',
+          rowProp,
+          Array.isArray(backgroundProp)
+            ? backgroundProp[index % backgroundProp.length]
+            : backgroundProp,
+        ) || background
+      }
       border={normalizeProp('border', rowProp, border)}
       pad={normalizeProp('pad', rowProp, pad)}
       pin={pin}
--- a/src/js/components/DataTable/Cell.js
+++ b/src/js/components/DataTable/Cell.js
@@ -6,6 +6,7 @@ import { defaultProps } from '../../default-props';
 import { Text } from '../Text';
 import { StyledDataTableCell } from './StyledDataTable';
 import { datumValue } from './buildState';
+import { TableContext } from '../Table/TableContext';
 
 const normalizeProp = (name, rowProp, prop) => {
   if (rowProp && rowProp[name]) return rowProp[name];
@@ -16,7 +17,6 @@ const Cell = ({
   background,
   border,
   column: { align, pin: columnPin, property, render, verticalAlign, size },
-  context,
   datum,
   index,
   pad,
@@ -27,6 +27,8 @@ const Cell = ({
 }) => {
   const theme = useContext(ThemeContext) || defaultProps.theme;
   const value = datumValue(datum, property);
+  const context = useContext(TableContext);
+
   let content;
   if (render) {
     content = render(datum);
--- a/src/js/components/DataTable/Cell.js
+++ b/src/js/components/DataTable/Cell.js
@@ -47,7 +47,7 @@ const Cell = ({
   return (
     <StyledDataTableCell
       scope={scope}
-      // columnPin={columnPin}
+      columnPin={columnPin}
       {...theme.dataTable[context]}
       align={align}
       verticalAlign={verticalAlign}
--- a/src/js/components/DataTable/Cell.js
+++ b/src/js/components/DataTable/Cell.js
@@ -47,9 +47,9 @@ const Cell = ({
   return (
     <StyledDataTableCell
       scope={scope}
-      columnPin={columnPin}
       {...theme.dataTable[context]}
       align={align}
+      context={context}
       verticalAlign={verticalAlign}
       size={size}
       background={normalizeProp(
--- a/src/js/components/DataTable/Cell.js
+++ b/src/js/components/DataTable/Cell.js
@@ -47,6 +47,7 @@ const Cell = ({
   return (
     <StyledDataTableCell
       scope={scope}
+      // columnPin={columnPin}
       {...theme.dataTable[context]}
       align={align}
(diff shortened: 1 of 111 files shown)


src/js/themes/base.js
```diff
@@ -668,6 +668,17 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
+      //   body: {
+      //      extend: undefined,
+      //   },
+      //   header: {
+      //      extend: undefined,
+      //   },
+      //   footer: {
+      //      extend: undefined,
+      //   },
+      // },
```
halocline, line 681: Do these need background properties?

```suggestion
      // pinned: {
      //   body: {
      //      background: undefined,
      //      extend: undefined,
      //   },
      //   header: {
      //      background: undefined,
      //      extend: undefined,
      //   },
      //   footer: {
      //      background: undefined,
      //      extend: undefined,
      //   },
      // },
```

src/js/components/DataTable/doc.js
```diff
@@ -352,6 +352,36 @@ export const themeDoc = {
     type: 'React.Element',
     defaultValue: undefined,
   },
+  'dataTable.pinned.body.background.color': {
```
halocline, line 355: Is it only `background.color` which can be set? What about `opacity`, `image`, etc.? Should the docs be speaking to `dataTable.pinned.body.background` and type would be also include a background type object?

I think the way this is implemented the other background type properties such as image, opacity, etc. all are supported, correct?

halocline (commented):
Looking good. Only question is regarding documentation. Are we supporting only background.color? Or all backgroundType properties?

src/js/components/DataTable/doc.js
taysea: All background props apply. I will update the site docs to reflect this. Thanks for noting.

taysea: Updated in recent commits.

src/js/components/DataTable/doc.js
```diff
@@ -352,6 +352,45 @@ export const themeDoc = {
     type: 'React.Element',
     defaultValue: undefined,
   },
+  'dataTable.pinned.body.background': {
+    description: `The background color for pinned footer cells. Any valid Box 
+    background options apply.`,
+    type: `string | { color: { dark: string, light: string }, dark: bool | 
+    string, image: string, position: string, opacity: string | { dark: string, 
+    light: string }, repeat: no-repeat | repeat, size: cover | contain }`,
```
halocline, line 360: Can we double check the types here? Something looks jumbled up here. Bracket or pipe out of place?

I feel like the type looks something more like this, but could be wrong:

```suggestion
    type: `string | 
               { dark: string, light: string } |
               { 
                   color: { dark: string, light: string }, 
                   dark: bool, 
                   image: string, 
                   position: string, 
                   opacity: bool | string, 
                   repeat: no-repeat | repeat, 
                   size: cover | contain | string
               }`,
```

halocline (commented):
Question re: background type definition on doc.js

src/js/components/DataTable/doc.js
taysea: Thank you, good catch I had not included the `{ dark: string, light: string }` option on its own.

taysea:
> Question re: background type definition on doc.js

Resolved in recent commit. Thank you!

halocline (approved):
Looks good. I'm happy. 😊

ShimiSun (commented):
LGTM. I  dropped quick comments about the docs, the comments made on `body` but should be duplicated for header+footer.

ericsoderberghp:
Would it be possible to do this without changing TableCell? My rationale is that pinning isn't a part of Table, just DataTable. I am wondering if TableCell can use the styling as is, based on the table context, then DataTable puts the pinning styling in StyledDataTableCell, with the pinning styling in the theme being a part of DataTable and not Table. Perhaps I am missing something though?

taysea:
> Would it be possible to do this without changing TableCell? My rationale is that pinning isn't a part of Table, just DataTable. I am wondering if TableCell can use the styling as is, based on the table context, then DataTable puts the pinning styling in StyledDataTableCell, with the pinning styling in the theme being a part of DataTable and not Table. Perhaps I am missing something though?

I think that that mindset makes sense. I had not tried putting the styling onto `StyledDataTableCell` because I was following all the way down to where the styling logic of props vs theme was being sorted out. Let me give your approach a try.

src/js/components/DataTable/doc.js
```diff
@@ -352,6 +352,69 @@ export const themeDoc = {
     type: 'React.Element',
     defaultValue: undefined,
   },
+  'dataTable.pinned.body.background': {
+    description: `The background color for pinned footer cells. Any valid Box
```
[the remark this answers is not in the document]
taysea: Good comment. Yes, I will remove that from all of those.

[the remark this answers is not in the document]
taysea: Resolved in recent commit.

src/js/components/DataTable/Cell.js
```diff
@@ -49,6 +49,7 @@ const Cell = ({
       scope={scope}
       {...theme.dataTable[context]}
       align={align}
+      context={context}
```
ericsoderberghp, line 52: Can Cell.js use TableContext? It seems odd to have TableContext but then also use a prop.

src/js/components/Table/StyledTable.js
```diff
@@ -47,6 +47,7 @@ const StyledTableCell = styled.td`
       props.theme,
     )}
   ${props => props.tableContextTheme && props.tableContextTheme.extend}
+  ${props => props.pinnedDataTableTheme && props.pinnedDataTableTheme.extend}
```
ericsoderberghp, line 50: Doesn't seem right to have a data table theme element inside StyledTable.js

taysea: Whoops this should be removed now. Thanks for the catch.

src/js/components/DataTable/Cell.js
taysea: Updated Cell to use `TableContext` instead of receiving a prop of `context`, but we still need to pass it as a prop to `StyledDataTableCell` to be able to utilize it in the styled component logic. Unless there is a way to access context in a styled component without props?

ericsoderberghp (approved):

Issue #4522, opened 2020-09-15
[TS] Text Input Prop is missing type prop


babbarankit:
### Expected Behavior

Text Input Prop must have type prop

ShimiSun:
As I commented on the PR, we try not to document props that are not part of the grommet API and keep the default of the HTML documentation and types. type isn't part of the grommet API hence we do not document it or add typing to it. Any reason you think this addition is useful? Is it currently not working?

babbarankit:
Hi @ShimiSun 
Thanks for reaching out so quickly. The use case is very  simple, I have constructed new Form Components using the TextInput, extended as PasswordInput or EmailInput. 
Unless I pass the prop to TextInput for type, how will it behave as HTML5 element. HTML5 prop must be passed to TextInput.
Currently I am using Patch Package otherwise Typescript complaints on passing extra type prop to TextInput.

ShimiSun:
@babbarankit I've built a quick markup that uses TypeScript+Grommet+TextInput with the `type` prop and it doesn't reproduce the issue, https://codesandbox.io/s/grommet-ts-textinput-type-nez6b?file=/src/index.tsx.
Feel free to tweak the example as you see fit to reproduce the problem so we can reproduce it reliably on the grommet side before attempting to fix it.

ShimiSun:
Interesting discussion on the PR #4523

jcfilben:
Closing since we were unable to reproduce the issue. Feel free to reopen with a code example.

PR #4523, opened 2020-09-15
Add Type on Grommet Text Input

babbarankit:
Signed-off-by: Ankit Babbar <babbar.ankit@gmail.com>

#### What does this PR do?
Fixes https://github.com/grommet/grommet/issues/4522

### DIFF
--- a/src/js/components/Text/index.d.ts
+++ b/src/js/components/Text/index.d.ts
@@ -23,6 +23,7 @@ export interface TextProps {
   truncate?: boolean;
   weight?: "normal" | "bold" | number;
   wordBreak?: "normal" | "break-all" | "keep-all" | "break-word";
+  type?: "text" | "email" | "hidden" | "password" | "tel" | "time" | "date" | "datetime-local" | "url" | "week"; 
 }
 
 declare const Text: React.FC<TextProps & Omit<JSX.IntrinsicElements['span'], 'color'>>;


src/js/components/Text/index.d.ts
```diff
@@ -23,6 +23,7 @@ export interface TextProps {
   truncate?: boolean;
   weight?: "normal" | "bold" | number;
   wordBreak?: "normal" | "break-all" | "keep-all" | "break-word";
+  type?: "text" | "email" | "hidden" | "password" | "tel" | "time" | "date" | "datetime-local" | "url" | "week";
```
ShimiSun, line 26: We try not to document props that are not part of the grommet API, and default to the HTML documentation and types. `type` isn't part of the grommet API hence we do not document it. Any reason you think this addition is useful?

babbarankit: I see my mistake, I understand TextInputProps is not complete, but I feel that since the component TextInput takes all props including that of JSX, write method must be
```
export interface TextInputProps extends Omit<JSX.IntrinsicElements['input'], 'onSelect' | 'size' | 'placeholder'>> {
  //the extra props.
 }
```
We need not document them but since the component takes these props, I must be able to extend it in my Component Directly instead of still doing like:

```
export interface PasswordInputProps extends TextInputProps, Omit<JSX.IntrinsicElements['input'], 'onSelect' | 'size' | 'placeholder'>> {
  //the extra props.
 }
```
This is standard practice, you can also check [AntDesign Input](https://github.com/ant-design/ant-design/blob/master/components/input/Input.tsx)
Therefore, props a component can take (whether they are of Grommet or JSX) must be exported as one. 
I can certainly create a PR for this and other Input Elements, once you approve.

ShimiSun: >I can certainly update the PR for this and other Input Elements, once you approve.
Feel free to update your PR. @carolineartz your inputs & insights could be helpful here.

carolineartz: ❤️ 🔥 topic! 🌶️  ❤️ 

This is an important point and something that has forced me to work around a few times. A related convo here: https://github.com/grommet/grommet/pull/3647%23discussion_r451646653

I use @jasnross's lovely utility addition, `PropsOf`, all the time to get the actual computed props for the grommet components so i can omit and extend them. And there likely exists other places where intrinsic props have definitions that don't actually match the intrinsic definition, which would be the case if these changes were merged, but correct me if i'm mistaken.

Inspecting the type of `JSX.IntrinsicElements['input']["type"]`, it looks like TS doesn't actually type the `type` attribute beyond an optional `string`, which sucks but thats because values which are not permitted are not actually _invalid_ for some reason in the HTML api 🤷‍♀️ 

So....what to do?

One idea that I have not fully thought through but might avoid breaking changes with existing definitions and allow a period of deprecation could be to create and expose a new prop type for each of the components' exhaustive props. Currently these are all hidden in their implementation e.g, `declare const Text: React.FC<TextProps & Omit<JSX.IntrinsicElements['span'], 'color'>>`.

so like just adding something with a naming convention 

```ts
export type GrommetTextProps = TextProps & Omit<JSX.IntrinsicElements['span'], 'color'>
declare const Text: React.FC<GrommetTextProps>`
```

I don't know...thoughts? @jasnross any ideas?

jasnross: I don't mind that idea. Although, it got me wondering how big of a breaking change it would really be to just start including all the intrinsic props in the props types. I worry a little bit that introducing another props type could add to the confusion by adding one more thing for new users to learn about, and doing a quicker change, even if it has some (hopefully relatively minor) breaking changes, seems like it could be worth considering.

Anecdotally, I can say somewhat confidently that it would not break much if anything, and even if it did break something it would be an easy fix since the only failures would be at build time, making them simple to resolve in any cases I can imagine right now.

I do feel like the most common expectation among TS developers would be that types like `TextProps` should represent _all_ the props that can be accepted by the `Text` component, not just the non-intrinsic props. This would probably alleviate some initial confusion with new Grommet/TypeScript users, as I've seen new devs on my team stumble over this a few times as well.

My main hesitation would really just be that I don't fully understand why the exported props types are not already including intrinsic element props. Maybe there isn't really a strong reason other than "that's the way it is" (it happens 😄). But if there is a reason behind it I'd like to consider that as well.

Would it be appropriate to open a separate issue to talk about this further (either the incremental approach you proposed or just updating the types in general)?

jasnross: @babbarankit Since, as @ShimiSun mentioned, Grommet does not typically define its own versions of the HTML props in the types, for now a good workaround could be to use the `PropsOf` utility type that @carolineartz mentioned.

Something like:

```typescript
import { PropsOf } from "grommet/utils";

export interface PasswordInputProps extends PropsOf<typeof TextInput> {
   // the extra custom props for password input (if any).
 }
```

ShimiSun:
@carolineartz & @jasnross how do you think we should proceed with this PR? we should thrive for consistency across the typed components, so I'm not sure whether that will be something we should document, fix or just leave it as is.

src/js/components/Text/index.d.ts
babbarankit: @jasnross thanks for the help I applied this and it worked.
On the other end @ShimiSun if we export complete Props , the chances of Breaking Code will be very less as we are adding additional Props which the element already has.

carolineartz: glad that worked for you @babbarankit! @ShimiSun we should close this PR as it stands right now the proposed changes aren't actually correct because`type?` on input is not actually one of `"text" | "email" | "hidden" | "password" | "tel" | "time" | "date" | "datetime-local" | "url" | "week"; `; it is `string | undefined`. 

image

I think discussion of transitioning to complete props for `<Component>Props` types should be continued elsewhere. I'm in agreement with @jasnross and @babbarankit in favor but not for this PR 👍

ericsoderberghp:
Closed due to inactivity

PR #4525, opened 2020-09-15
InfiniteScroll items with variable height fix

IanKBovard:
Related to https://github.com/grommet/grommet/pull/4369
Taken from pr's description.
> The pageHeight is calculated at the beginning of the rendered component but pageHeight is kept constant. A page can have varying content as a user scrolls and pageHeight is still kept the same and can cause the InfiniteScroll to miscalculate the nextEndPage. Rather than using pageHeight, I am proposing the usage of scrollHeight to determine when to call the onMore() function.
Although this works, the scrollHeight doesn't play well with the replace feature, I left a conditional statement for nextEndPage to use the original pageHeight if replace is set to true.
Feedback is very much appreciated.
What does this PR do?
Introduce using scrollHeight rather than using the pageHeight as a user scrolls.
src/js/components/InfiniteScroll/InfiniteScroll.js
Before using InfiniteScroll from master: https://codesandbox.io/s/green-water-n7e21?file=/src/App.js
After using InfiniteScroll from this PR: https://codesandbox.io/s/blue-tree-q4d1o?file=/src/App.js
Scroll through the contents of the table until you are no longer able to. The maximum index number to be reached at is 999.
What are the relevant issues?
#4337 & #3106
No


nnbrandon:
@IanKBovard @halocline Just ran this on my laptop and it looks great with the additional tests. Thanks for addressing the issues in this PR. 👍

halocline (commented):
Nice progress! I think things are looking good, but would like to take another review pass Monday morning with fresh eyes.

src/js/components/InfiniteScroll/InfiniteScroll.js
```diff
@@ -124,14 +147,36 @@ const InfiniteScroll = ({
...
+            ? Math.ceil((scrollBufferBottom * width) / avgPageArea)
+            : Math.floor(scrollBufferBottom / avgPageHeight),
           show ? Math.floor(show / step) : 0,
         ),
       );
-
       if (nextBeginPage !== beginPage) setBeginPage(nextBeginPage);
       if (nextEndPage !== endPage) setEndPage(nextEndPage);
+
+      // Keep track of page heights as begin/end pages increment/decrement
```
halocline, line 158: So the one edge case that still isn't behaving ideally is where we have the following scenario:
- `replace` && `onMore` && large variation pageHeight (pageHeights differing >20% +/-) --> specifically when scrolling past a large page of results, then "back up".  This is illustrated in the "Variable Item Height with Replace" story.

All other behaviors are solid.

halocline, line 158: Would it be better to keep all page heights in this array instead of removing? Then we could reference the avgPageHeight calculation by beginPage and endPage indexes in which we are interested?

src/js/components/InfiniteScroll/InfiniteScroll.js
```diff
@@ -196,9 +243,15 @@ const InfiniteScroll = ({
 
   const result = [];
 
+  const avgPageHeight =
```
halocline, line 246: We are duplicating this calculation. Does it make sense to pull into a single calc?

src/js/components/InfiniteScroll/InfiniteScroll.js
```diff
@@ -270,7 +323,11 @@ const InfiniteScroll = ({
         key="below"
         ref={belowMarkerRef}
         flex={false}
-        height={`${replace ? (lastPage - endPage) * pageHeight : 0}px`}
+        height={`${
+          replace && pageHeight && !onMore
```
halocline, line 327: Would a comment as to why `!onMore` be helpful here?

src/js/components/InfiniteScroll/stories/ItemHeights.js
```diff
@@ -0,0 +1,47 @@
...
+        </InfiniteScroll>
+      </Box>
+    </Grommet>
+  );
+};
+
+if (!isChromatic()) {
+  storiesOf('InfiniteScroll', module)
+    .add('Variable Item Height', () => <ItemHeights />)
+    .add('Variable Item Height w/replace', () => <ItemHeights replace />);
```
halocline, line 46: I believe the story naming convention is moving towards sentence casing.

Also, these stories have been useful from a QA standpoint, but I don't know if they add much for end users. I am bit torn on the balance for need for behavioral QA vs. instructive benefit.

IanKBovard: I dont think they add anything for end users. I'm ok with removing these, but I do agree that it does help with manual testing.

src/js/components/InfiniteScroll/InfiniteScroll.js
IanKBovard: ill add comment above. but !onmore will prevent the height of the bottom marker from being calculated when onmore and replace are both available. It's not possible, at least to me, for us to know the correct amount of pages needed for this calculation. I think its safe to assume that if a user passes onMore, that behavior should take precedence over replace.

src/js/components/InfiniteScroll/InfiniteScroll.js
IanKBovard: i can do that

src/js/components/InfiniteScroll/InfiniteScroll.js
halocline: I was thinking that sort of rationale would be worthwhile to leave for future contributors.

halocline:
> Related to #4369
> Taken from pr's description.

**For reviewers:** 
- **Context:** The problem being addressed with this PR has to due with when "steps of results" occupy significantly different heights in the DOM. InfiniteScroll's implementation up to this point calculated a "page height" based on the first "step of results" and did not revisit this assumption.
- **Problem:** When a page height (i.e. "step of results") is assumed to be static, situations arise where retrieving the next step of results, such as with `onMore` or `replace`, the mechanism for triggering retrieval of the next step can break down if the height of result sets differs significantly from one step to another.
- **Examples of Differing Heights for Steps of Results:**
  - A table with a column which may contain multiple lines of text, such as a comment. In this case, some table rows might occupy several "lines of height" where other only a single line.
  - Feed of Tweets. Some might be one liners, while others might have an image, link to referenced article etc. Because of this, each "step" of tweets is unknowable.
- **Solution Approach:** Instead of using a static `pageHeight` based on initial render, the approach taken here recalculates `pageHeight` as steps of results are added/removed. An average page height is calculated from the included result sets and then serves as the new trigger point for adding/removing pages.

IanKBovard:
Coming across a new issue here.
Kapture 2020-09-22 at 10 35 57

It looks like the above marker is being rendered and removed in between the first two pages only. This behavior isn't present in other pages.

src/js/components/InfiniteScroll/InfiniteScroll.js
IanKBovard: just double checking the only piece of this code block you want to keep is 
```
    if (nextEndPage > endPage) {
        nextPageHeights.push(pageHeight);
        nextPageAreas.push(pageArea);
      }
```
?

IanKBovard:
Changes made. The mentioned issue was because if nextBeginPage being constantly recalculated. Added !onmore to that calculation. Anytime, it seems to me, last page is required for a calculation it can't work with onmore.

halocline:
> Changes made. The mentioned issue was because if nextBeginPage being constantly recalculated. Added !onmore to that calculation. Anytime, it seems to me, last page is required for a calculation it can't work with onmore.

Cool. I'll check out the latest.

halocline (approved):
I think it's there. Nice work!

halocline:
**For reviewers:**
The following stories were useful for reproducing the original issue and testing the resolution. While valuable for QA, they were not included in the PR as we didn't see these stories as being valuable for end users.

```
import isChromatic from 'chromatic/isChromatic';

import { Grommet, Box, InfiniteScroll, Text } from 'grommet';

// import { allItems } from './Basics';
const allItems = Array(240)
  .fill()
  .map((_, i) => i + 1);

const ItemHeights = props => {
  const [items, setItems] = useState(allItems.slice(0, 50));

  const onMore = () => {
    setTimeout(() => {
      setItems(allItems.slice(0, items.length + 50));
    }, 1000);
  };

  return (
    <Grommet theme={grommet}>
      <Box>
        <InfiniteScroll items={items} onMore={onMore} {...props}>
          {item => (
            <Box
              key={item}
              height={item <= 25 ? 'xsmall' : 'xxsmall'}
              pad="medium"
              border={{ side: 'bottom' }}
              align="center"
            >
              <Text>item {item}</Text>
            </Box>
          )}
        </InfiniteScroll>
      </Box>
    </Grommet>
  );
};

if (!isChromatic()) {
    .add('Variable Item Height', () => <ItemHeights />)
    .add('Variable Item Height w/replace', () => <ItemHeights replace />);
}

```

ericsoderberghp:
Relying on the average page height feels a little brittle to me. If the page sizes are variable enough, and depending on their distribution, it seems like we might not always be calculating things right. This is just a theoretical reflection. I haven't delved into the current code enough to come with specific scenarios that will trouble it.

src/js/components/InfiniteScroll/InfiniteScroll.js
```diff
@@ -101,21 +105,43 @@ const InfiniteScroll = ({
...
-            Math.max(
-              0,
-              multiColumn
-                ? Math.floor((Math.max(0, top - offset) * width) / pageArea)
-                : Math.floor(Math.max(0, top - offset) / pageHeight),
-            ),
-          )
-        : 0;
+      // If onMore is present nextBeginPage should be 0,
+      // as there is no way of knowing lastPage.
```
ericsoderberghp, line 130: I think `lastPage` was meant to be the last page of the `data` we have, not the eventual last page when onMore is no longer provided.

ericsoderberghp (commented):
Here are a few comments based on a quick skim. I'm still looking further though.

src/js/components/InfiniteScroll/InfiniteScroll.js
```diff
@@ -76,7 +80,7 @@ const InfiniteScroll = ({
       setPageArea(nextPageArea);
       setMultiColumn(nextMultiColumn);
     }
-  }, [pageHeight, step, show]);
+  }, [beginPage, endPage, items, step, show]);
```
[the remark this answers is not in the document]
halocline: The intent is to recalculate pageHeight whenever begin and end pages are incremented/decremented. In this case they are triggers, rather than dependencies.

re: items - @IanKBovard, I can't recall, was items added here also as a trigger for when onMore collects additional items?

halocline:
> Relying on the average page height feels a little brittle to me. If the page sizes are variable enough, and depending on their distribution, it seems like we might not always be calculating things right. This is just a theoretical reflection. I haven't delved into the current code enough to come with specific scenarios that will trouble it.

This concern is valid and worth working/thinking through. I believe the calculation resolves correctly, but in the theoretical realm could cause "jumpiness" in re-renders. The effect would be most exaggerated when one has minimal pages (2 pages) and the heights of those pages varies significantly (i.e. this scenario has the least amount of diversification and would thus feel the most volatility of the average; more pages equates to greater diversification and less volatility). That said, the current implementation was tested against page heights which reduced height by 50%, which probably lives at the extreme of real world data (unless the data had a power curve like distribution).

An alternative is to always use the minimum pageHeight when adding result pages (max pageHeight when removing result pages). Pluses are 1) that using a min pageHeight would always have the scroll window buffers triggering nextBegin/EndPage calcs ahead of when they are needed to add items (max pageHeight would ensure results pages aren't removed prematurely), and 2) would not rely on the average approach. The downside is that a min/max pageHeight would be holding an excess of items in the DOM which goes against the purpose of replace, but the cost

src/js/components/InfiniteScroll/InfiniteScroll.js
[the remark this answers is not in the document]
IanKBovard: that is correct. items is needed to recalculate pageheight.

src/js/components/InfiniteScroll/InfiniteScroll.js
IanKBovard: The main reasoning behind this is to prevent some issues with variable heights when onmore and replace are present.

This is when onmore and replace are present:

(nextEndPage fluctuates between 0 and 1)
Kapture 2020-09-24 at 10 38 52

(bottom marker is too tall and causes scrollbar to snap)
Kapture 2020-09-24 at 10 43 49

nnbrandon:
Any updates on this PR?

ericsoderberghp:
> Any updates on this PR?

There were a few parts that felt a little fragile to me. I have been off exploring other ways to approach this, but have been slow due to vacations and other work. I am hoping to either have something good or to just take these changes here as an improvement within the next week. Sorry for the delay.

nnbrandon:
> > Any updates on this PR?
> 
> There were a few parts that felt a little fragile to me. I have been off exploring other ways to approach this, but have been slow due to vacations and other work. I am hoping to either have something good or to just take these changes here as an improvement within the next week. Sorry for the delay.

Thanks Eric!

ShimiSun:
I think we can close this PR due to the recent PR merge of #4642 
Thank you!

Issue #4441, opened 2020-08-24
List onClickItem not raised


matthieu-stacklabs:
Since version 2.15 (it used to work on 2.14)

In a List component with onClickItem handler, the expected event is not raised when the item is clicked.

### Expected Behavior

It should behave as documented: "When supplied, this function will be called with an event object that include a 'item' property containing the data value associated with the clicked item and an 'index' property containing the index in 'data' of the clicked item."

### Actual Behavior

The whole List component takes focus and, I guess, prevents the click event from bubbling

### URL, screen shot, or Codepen exhibiting the issue

See official codesandbox, https://storybook.grommet.io/?path=/story/list--onclickitem

### Your Environment

- Grommet 2.15
- Chrome browser v84.0.4147.135
- MacOS Catalina (Desktop)

taysea:
Hey @matthieu-stacklabs, could you provide a sandbox demonstrating the bug? 

I tried the storybook example locally and adjusted the onClickItem to be: `onClickItem={event => console.log(event)}` and the event properly logged out and contained the additional `event.item` as documented.

ericsoderberghp:
FYI, I am seeing this behavior on an app as well. The first click causes the List to receive focus but the `onClickItem` handler isn't called. The second click does call `onClickItem`.

ericsoderberghp:
If I insert a `<Box height="medium" flex={false} />` above the List in the onClickItem story, the problem reproduces. It looks like if clicking will cause the List to scroll in some way, the onClickItem function isn't called.

matthieu-stacklabs:
Hi Taysea, you're right it works well on simple example, but find below a little more complex one which raises the issue:

```
import React from "react";

import { Grommet, List } from "grommet";

const Demo = (props) => {
    const data = [
        { label: "a" },
        { label: "b" },
        { label: "c" }];

    return (
        <Grommet><List data={data} primaryKey="label"
            onClickItem={(e) => { console.log(e.item.label) }} />
        </Grommet>
    )
}

export default Demo;
```

Actually, it's when the "primaryKey=..." attribute is present that the behaviour is odd: it seems that only the clicks on "b" are well received, many clicks on "a" or "c" go missing

Regards

ericsoderberghp:
I'm wondering if the problem is tied to `listRef.current.focus();` inside `onFocus`. If I remove that, it seems to behave correctly.

taysea:
[...]

PR #4529, opened 2020-09-17
Fix onClick event for List and handle focus

taysea:
#### What does this PR do?

Updates the way focus is handled within List. In order for the click event to properly propagate, we have to restore focus on the List container after the onClick of the list item has completed. However, for a brief interim (when the user has the mousedown, but has not completed the click event), the individual list item has focus in the DOM. At this stage, the focus indicator was being pulled away from the list container to wrap around an individual list item— creating a jumpy behavior of the focus indicator. Since the focus is restored to the list container as soon as the onClick finishes, we wanted to eliminate this visual jump of the focus before wrapped briefly around an individual list item.

@ericsoderberghp and I worked through a couple variations to this approach. We noted that this solution isn't perfect, but it seemed like the best fix for now to ensure that onClick works properly with mouse and keyboard and that List was most closely aligned with accessibility guidelines as possible. If we can think of a more seamless solution in the future to the focus handling, then we can enhance the component.

Test with `onClickItem` List storybook example and observed the existing onClick List test changes.

With `onClickItem` List storybook example.

Closes https://github.com/grommet/grommet/issues/4441
Closes https://github.com/grommet/grommet/issues/4173
No.

Yes.

Yes, backwards compatible.

### DIFF
--- a/src/js/components/List/List.js
+++ b/src/js/components/List/List.js
@@ -5,7 +5,12 @@ import { Box } from '../Box';
 import { InfiniteScroll } from '../InfiniteScroll';
 import { Keyboard } from '../Keyboard';
 import { Text } from '../Text';
-import { focusStyle, genericStyles, useForwardedRef } from '../../utils';
+import {
+  focusStyle,
+  genericStyles,
+  unfocusStyle,
+  useForwardedRef,
+} from '../../utils';
 
 const StyledList = styled.ul`
   list-style: none;
@@ -36,7 +41,7 @@ const StyledItem = styled(Box)`
   // for visual consistency, we are showing focus on the list container
   // as opposed to the item itself.
   &:focus {
-    outline: none;
+    ${unfocusStyle({ forceOutline: true, skipSvgChildren: true })}
   }
 `;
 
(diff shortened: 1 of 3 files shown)


src/js/components/List/List.js
```diff
@@ -18,10 +18,26 @@ const StyledList = styled.ul`
...
 
 const StyledItem = styled(Box)`
   ${props => props.onClick && `cursor: pointer;`}
+  // during the interim state when a user is holding down a click,
+  // the individual list item has focus in the DOM until the click
+  // completes and focus is placed back on the list container.
+  // for visual consistency, we are showing focus on the list container
+  // as opposed to the item itself.
+  &:focus {
+    outline: none;
```
ericsoderberghp, line 39: There are other styles of focus possible, based on how the theme works. I think we should consider adding an `unfocusStyle()` that parallels `focusStyle()` to handle that.

taysea: I've created an `unfocusStyle()` function next to `focusStyle()` -- I added a comment in the file noting that it is being used in List for a specific case and that the general recommendation would be to not integrate this into other components.

ericsoderberghp (approved):

PR #4531, opened 2020-09-17
Updated dependencies - styled-components excluded

ShimiSun:
We need to update the grommet dependencies. This PR is upgrading all dependencies except the `styled-components` one due to the breaking changes on the tests, see https://github.com/styled-components/styled-components/releases/tag/v5.2.0  for more insight.
The styled-components upgrade will be isolated to a separate and more thought-through PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,7 +129,7 @@
     "regenerator-runtime": "^0.13.1",
     "require-reload": "^0.2.2",
     "simple-git": "^1.95.0",
-    "styled-components": "^5.1.0",
+    "styled-components": "5.1.1",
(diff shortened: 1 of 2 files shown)


ericsoderberghp (commented):
Do we need to update the `peerDependencies` too?

ShimiSun:
Updating the peer-dependency will force the users to use a  certain specific version of `styled-components` in their applications. I don't think we should limit our users to do so since the failures we currently see with styled-components are only on the snapshots and not on the application itself.

ericsoderberghp (approved):

PR #4538, opened 2020-09-21
Fix Avatar text size customizability

### LINKED ISSUE #4537: Customizable Avatar text size
### Expected Behavior

I'd expect the size of the text in the Avatar component to be customizable to suit the different sizes of the Avatar component.

### Actual Behavior

The text size is fixed at `"large"`, per https://github.com/grommet/grommet/blob/85e3d85896327d9e6bf96ca5e6e307bd4793fccc/src/js/components/Avatar/Avatar.js#L50.

Looks ridiculous when using `"xsmall"` for the size of the Avatar component, while the text is still `"large"`. The problem is there is no way to change the text size without changing the global size of `"large"` Text.
image

### Steps to Reproduce

```
<Avatar size="xsmall">T1</Avatar>
```

### Your Environment

- Grommet version: 2.14.0
- Browser Name and version: Irrelevant
- Operating System and version (desktop or mobile): Irrelevant

leossantiago:
Signed-off-by: Leonardo Santiago <leossantiago1@gmail.com>

#### What does this PR do?
Fixes issue #4537 . When using a string as children of an Avatar component, the text size would always be "large", despite property "size" being applied. With this PR, text size is now customizable and follows the property (same as the Avatar size), and both default to medium.

src/js/components/Avatar/Avatar.js
Manual testing using the component and setting the property "size". Also updated the snapshot test of component Sidebar, that uses Avatar and was expecting a large text, but it now defaults to medium.

Create component  Avatar, put some text as children and apply the property size ("xsmall", "small", "medium", "large" ...) over Avatar.

Closes #4537 
No
Backwards compatible (bug fix)


Zodiase:
Thanks for working on this!

I like the simplicity of this approach: medium avatar with medium text. However this would be a breaking change for all current users of the Avatar component, and I'm still slightly concerned about the flexibility of the text sizes.

On one hand, I'd love for all text sizes to fit into the few levels defined in Text. One the other hand, IMO Avatar text is more like a decoration/placeholder for the image, so it doesn't feel like Avatar text size should be confined by Text size options.

For not knowing a better way, I'd propose a new prop `textSize` on `Avatar` and have it passed directly to the `size` prop of `StyledAvatarText`. Default `textSize` to `"large"` to be compatible to the current behavior.

leossantiago:
I like how text fits nicely within the Avatar with this approach, but I totally understand your concern. Mind if I try to implement this?
Guess it'll also require some changes to doc and storybook, am I right?

netochaves:
I think a better approach would be to add a new theme option for `StyledAvatarText`, something like` theme.avatar.text.size` what do you think?

leossantiago:
@netochaves text size can be themed using extend, as it receives any CSS styling. It might be a nice shortcut to have, along with the property in Avatar itself. That way, text size can be customized from both theme and the component props.
image

leossantiago:
Pushed some changes. Added a new property to Avatar (textSize), updated docs and snapshots.

ShimiSun (approved):

ShimiSun:
Thank you for submitting this PR and making a contribution 🎉 

Here are some thoughts on the overall approach:
1. Text size should automatically & proportionally respond to the Avatar size.
2. `theme.avatar.text` should support `size` prop that will map the size of the avatar to text size.
3. We will keep the component API clean from new props and put the load of the logic on the theme, so the users can customize the correlation of textSize and the avatar size generically on the theme.

So the theme would include `theme.avatar.text.size` as follows:
```
avatar: {
      // extend: undefined,
      size: {
        xsmall: `${baseSpacing * 0.75}px`,
        small: `${baseSpacing}px`,
        medium: `${baseSpacing * 2}px`, // default 48
        large: `${baseSpacing * 3}px`,
        xlarge: `${baseSpacing * 4}px`,
      },
      text: {
        size: {
          xsmall: 'xsmall',
          small: ''medium,
          medium: 'large',
          large: 'xlarge',
          xlarge: 'xxlarge',
        },
        // fontWeight: undefined,
        // extend: undefined
      },
    },
```

Feel free to let me know what you think! Thank you for working on this.

leossantiago:
I like this approach. Text size follows Avatar size, and the only way to use different sizes between them would the theming. I'll work on the `theme.avatar.text.size`. Regarding dropping the new prop `textSize`, my only concern is it might be a breaking change, as noted by @Zodiase . But I can change it back too (it was the first approach). What do you guys think?

Zodiase:
@ShimiSun I definitely prefer using themes to make it both flexible and clean. This may go against the best practices but I'd suggest considering the ease of use of the Avatar component. For example right now in our code we are using an oddly sized Avatar that's not of any size currently defined in the theme. I know it's bad but we need that size and haven't really consolidated it with the sizes in the theme. The `size` prop from `Avatar` *can* take an arbitrary value to make it fit into scenarios like the one I just mentioned. So far the missing piece is the ability to also change the text size to fit. It's nice to have the text-avatar size correlation in the theme but when `Avatar` is set to an arbitrary value, there is no way to also specify the text size.

ShimiSun:
@leossantiago awesome, thanks. Let's look at the code once it's ready, and see how we can mitigate potential breaking changes. The discussion will be more productive once there will be code/tests.

@Zodiase, thank you for the insights, we appreciate bringing up use cases to our attention. Regarding your concern of:
> The size prop from Avatar can take an arbitrary value to make it fit into scenarios like the one I just mentioned. So far the missing piece is the ability to also change the text size to fit.

I definitely agree, and I would like to see how we can integrate this behavior to the PR.

leossantiago:
Taking everything discussed in consideration, I did some changes to the approach and just pushed some changes.

Changed base theme to (slightly changed: xsmall > small instead of xsmall > xsmall)
```
avatar: {
      // extend: undefined,
      size: {
        xsmall: `${baseSpacing * 0.75}px`,
        small: `${baseSpacing}px`,
        medium: `${baseSpacing * 2}px`, // default 48
        large: `${baseSpacing * 3}px`,
        xlarge: `${baseSpacing * 4}px`,
      },
      text: {
        size: {
          xsmall: 'small',
          small: 'medium',
          medium: 'large',
          large: 'xlarge',
          xlarge: 'xxlarge',
        },
        // fontWeight: undefined,
        // extend: undefined
      },
    },
```

And size of text is determined by `const avatarTextSize = textSize || theme.avatar.text.size[size];`, and textSize doesn't default to any value. That way, if `textSize`is passed down, it controls text size. Otherwise, it follows theme, that behaves according Avatar `size`.

`size`has 'medium' as default value. That means that, if no props are passed, text size will default 'large'.

Also updated docs and snapshots. And I think the test created before is still valid for this approach.

What do you guys think?

src/js/components/Avatar/Avatar.js
```diff
@@ -12,11 +12,13 @@ const Avatar = ({
   round = 'full',
   size = 'medium',
   src,
+  textSize,
```
ShimiSun, line 15: I don't think we need to extend the component API if we are providing an extensive theming option.

src/js/components/Avatar/__tests__/Avatar-test.js
```diff
@@ -101,4 +101,19 @@ describe('Avatar', () => {
     const tree = component.toJSON();
     expect(tree).toMatchSnapshot();
   });
+
+  test('text renders with different sizes', () => {
+    const component = renderer.create(
+      <Grommet>
+        <Avatar background="dark-2" textSize="small">
```
ShimiSun, line 108: I think we should transfer the test to testing the theme value size and its corresponding textSize. 
One request, if you can separate this test to a separate PR, so we can make sure no changes in the snapshot are being made with this PR it will be fantastic.

leossantiago: Should I move docs update aswell? As it updates README snapshot.

ShimiSun: Yep.

leossantiago: No problem. I'll let you know when those things are done. I'm just not sure if I understood the 

> And on this PR, you can just add the reference to avatar.text.size once it exists.

part.Are you talking about the docs edit or tests? Sorry for that.
Do you mind elaborating a little more? Thanks.

leossantiago:
Updating. Latest changes:

- Dropped the addition of new props to the component API
- Removed changes to test snapshots of Avatar

What this PR do:

- Avatar > Text size now changes according to Avatar's prop `size`. That's defined by theming, and can be customized there to fulfill user needs.
- When Avatar `size` is set to an arbitrary value and this case is not defined in theme, text `size` defaults to `large`
- Update docs according to the changes.

@ShimiSun as I said before in code comments, I'm not sure if I understood all your suggestions. Could you verify if there's anything else I could do to improve this PR? (Other than sending a new PR with tests).

src/js/components/Avatar/__tests__/Avatar-test.js
ShimiSun: Yes, sorry for not being clear, and thanks for your patience. 
Just keep the tests on this PR, no need for the extra complexity.

src/js/components/Avatar/doc.js
```diff
@@ -68,4 +68,34 @@ export const themeDoc = {
     type: 'number',
     defaultValue: undefined,
   },
+  'avatar.text.size.xsmall': {
+    description:
+      'The xsmall size of the text in Avatar (Mapped according Avatar size).',
```
ShimiSun, line 73:
```suggestion
      "The size of the text that is mapped according to 'avatar.size.xsmall'". ,
```
What do you think of the proposed structure sentence? Should we add to it anything so it will be more clear?

src/js/themes/base.d.ts
```diff
@@ -353,6 +353,14 @@ export interface ThemeType {
       [x: string]: string | undefined;
     };
     text?: {
+      size?: {
```
ShimiSun, line 356: Thank you!

ShimiSun (commented):
Looking great, dropped a small comment. The suggested test looks great, feel free to add it to the PR. Thank You!

src/js/components/Avatar/doc.js
leossantiago: At first I was a little bit worried about people getting confused about the relation being medium to large, large to xlarge and so on. But considering the docs show the default value it's all good. And I like the suggested change. Will apply it in the next commit! Thanks for that.

src/js/themes/base.d.ts
leossantiago: You're welcome :) :tada:

leossantiago:
Added testing and updated docs descriptions to your suggestion. What do you think? Any way we could still improve this PR?

ShimiSun (approved):

ShimiSun:
Looking great, thank you for your work & patience with this PR! 💟  Pleasure working with you.

PR #4541, opened 2020-09-22
Fix calendar update when changing data manually on DateInput

### LINKED ISSUE #4535: DateInput change year after UI dropdown displays results in incorrect value
### Expected Behavior
When the UI date input dropdown is showing, changing the year manually should result in the calendar scrolling and date changing.

### Actual Behavior
The UI appears to change, but upon selecting a new date from that new year the value changes to that date but the old year.

https://storybook.grommet.io/?path=/story/dateinput--form

Steps to reproduce:

1) Pick any date from UI drop down
2) Open dropdown again and change year manually 
3) 
      a) If the UI scroll-animates its change, the new year will work
      b) If not, the stale year issue will occur

See link from grommet slack
https://slack-files.com/T04LMHMUT-F01BCJSTC5P-7792c8c3b6

### Your Environment
- Grommet version: latest
- Operating System and version:  desktop and mobile

leossantiago:
#### What does this PR do?
Fixes issue #4535 , where Calendar date range wouldn't update properly when changing date using an input (specially DateInput component). It also updates a typo in Calendar tests, where a range date wasn't identified correctly by the component (missing a zero).

Calendar component has a prop *animate* that defaults to true. There's a hook that triggers when reference (date) changes. If animate is true, it sets a new value for calendar bounds using TargetDisplayBounds, that is related to animated transitions.
However, animations only occur if the difference between the new and old references is below a year. That's what triggers the issue.

If prop animate is set to false, it works fine. animate can be passed down to Calendar even if you're using DateInput, with the property calendarProps.

Manual testing using Calendar and DateInput components.

```
import React from "react";

import { Grommet, Box, Calendar, DateInput } from "grommet";
import { grommet } from "grommet/themes";

const DATE = "2020-01-15T00:00:00-08:00";

const App = () => {
  const [value, setValue] = React.useState("");
  const onChange = (event) => {
    console.log("onChange", nextValue);
    setValue(nextValue);
  };
  return (
    <Grommet theme={grommet}>
        <Box width="medium">
          <DateInput
            format="mm/dd/yyyy"
            value={value}
            onChange={onChange}
          />
          <Calendar date={DATE} />
        </Box>
      </Box>
    </Grommet>
  );
};

export default App;
```

Closes #4535 
No
Backwards compatible


netochaves:
I think some tests would be good here

leossantiago:
Thanks for the suggestion @netochaves . I'll take a look at existing tests and see if we can cover this solution.

jcfilben:
I tested this in storybook, but I am still seeing the wrong year displayed when I go from 2010 to 2011 and select a date in 2011
Sep-25-2020 16-07-44

leossantiago:
@jcfilben thanks for pointing that. I did some quick manual tests here and it looks like the change doesn't occur properly when the time gap is less than a year. I'll try to fix that. I guess testing will be the ultimate way to tell this is working as it should, as pointed by @netochaves .

leossantiago:
@jcfilben, I spent the last couple days doing some tests using both storybook and an external little project. 
I identified a problem related to quick manual changes to date when the user tries to change it before the timer for animating ends.
It looks like this:
> A date change smaller than a year (as in from 2020 to 2021) triggers the animation (2010 to 2011 wouldn't be an option because deleting the zero makes the date go to default value, 2020). After that, quickly changing from 2021 to 2022 makes the reference update before the end of the transition from 2021 to default, and thus the new date is not set properly.

I tried your example, even using the same dates, but I couldn't reproduce it. I don't want to be troublesome, but could you please try it again and verify if there's any reproducible pattern?

jcfilben:
Hi @leossantiago I am still able to reproduce the issue. If you select a date with day X and year Y and then change the year to Y+1 the select a day that is >= X+1 it shows the date with the incorrect year. Let me know if you are able to reproduce this behavior as well.

Sep-30-2020 10-33-47

leossantiago:
I'll try to reproduce it, and hopefully, fix it.
One thing I noticed is that both your recordings show a transition from 2010 to 2011 and from 2016 to 2017 that were not supposed to happen, since the calendar goes back to 2020 whenever `Input` contains an invalid date. My guess is that the issue is related to that. I'll let you know if I discover something else. Thanks for helping with that. :smile:

leossantiago:
Pushed some changes to the code. The react hooks trigger a couple of times every time the Calendar is opened and tries updating the displayed bounds even if there's no change to it. And trying to move from one DisplayBound to itself was causing the code to try updating using transitions (`setTargetDisplayBounds`). I no longer can reproduce the error. Mind testing again and see if it works for you too?
I did some testing using storybook.

jcfilben:
I tested on storybook and was not able to reproduce the error, nice job!

src/js/components/Calendar/__tests__/Calendar-test.js
```diff
@@ -13,7 +13,7 @@ import { Box, Button, Calendar, Grommet, Text } from '../..';
 const DATE = '2020-01-15T00:00:00-08:00';
 const DATES = [
   '2020-01-12T00:00:00-08:00',
-  ['2020-01-8T00:00:00-08:00', '2020-01-10T00:00:00-08:00'],
+  ['2020-01-08T00:00:00-08:00', '2020-01-10T00:00:00-08:00'],
```
jcfilben, line 16: Why was this date changed? I think this is causing the snapshot changes

leossantiago: Yes.
I'm not sure if I'm supposed to change that, but as I described in the PR, I considered that a typo.
`DATES` is an Array that contains another in the second position. That means that Calendar should render two selections:
1. Single Date (Jan 12)
2. Range Date (Jan 8 - Jan 10)

Value `2020-01-8T00:00:00-08:00` doesn't seem to be recognized as valid, making Calendar render this:
image

instead of:

image


**Code used is the same as DATES Test**
```
import React from "react";
import { Grommet, Calendar } from "grommet";

const DATES = [
  "2020-01-12T00:00:00-08:00",
  ["2020-01-8T00:00:00-08:00", "2020-01-10T00:00:00-08:00"],
];

const App = () => {
  return (
    <Grommet>
      <Calendar dates={DATES} animate={false} />
    </Grommet>
  );
};

export default App;
```

leossantiago: Am I supposed to turn that into a separate PR?

jcfilben: Ah okay I see, I think it is fine to leave it in this PR instead of a separate one, what do you think @ShimiSun?

ShimiSun: I agree we can keep it.

src/js/components/Calendar/Calendar.js
```diff
@@ -189,7 +189,34 @@ const Calendar = forwardRef(
     // clear the slide and targetDisplayBounds.
     useEffect(() => {
       const nextDisplayBounds = buildDisplayBounds(reference, firstDayOfWeek);
-      if (!animate) {
+
+      // Checks if the difference between the current and next DisplayBounds is
+      // greater than a year. If that's the case, calendar should update without
+      // animation.
```
ShimiSun, line 195: Thank you for adding this helpful comment and making the open-source code more readable.

ShimiSun (approved):

Issue #4454, opened 2020-08-26
DataTable - Ability to style table column header cell when behaving as a button


halocline:
### Expected Behavior

DataTable column header cells can have button-like actions such as sorting. For UI consistency, I would like to set the style of my column header cells to resemble my button treatments.

### Actual Behavior

Column header cells currently have default stylings, but somewhat limited ability to customize.

Specifically, I would like to achieve the following:
- Specify a cell background color on hover
- Have my click target span/flex the width and/or height of the cell (currently fill is set to "vertical" but does not allow for "horizontal" or true)
- Have my click target have affordance/padding above and below cell title text (currently click target is limited to the text itself)
- Affect the default cell padding
- Specify font-weight
- Consideration for text alignment inside button when `column.align = 'end'` is specified.

Ideally, these would be properties I could specify in my theme.

Related Issue: https://github.com/hpe-design/design-system/issues/1078

### URL, screen shot, or Codepen exhibiting the issue
Have been able to achieve the follow using `table.header.extend` but has limitations and is a bit clunky to implement/

**Default state:**
![Screen Shot 2020-08-26 at 9 51 01 AM](https://user-images.githubusercontent.com/1756948/91326661-d6470f00-e781-11ea-87fb-bef8bc984ac4.png)

**Hover state** (able to affect button width, padding, background color, but cell has undesired padding on hover):
![Screen Shot 2020-08-26 at 9 51 16 AM](https://user-images.githubusercontent.com/1756948/91326675-d9da9600-e781-11ea-98a5-fa33fb620d5c.png)

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.15 / Stable
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #4542, opened 2020-09-22
Update theme styling options for DataTable header

taysea:
#### What does this PR do?
This PR extends theming capabilities of data table header cells. The contents of this cell can include a label (clickable if column is sortable, otherwise just text/custom render), a search button (not encouraged but already exists and left untouched), and a resizer. The `theme.dataTable.header` structure now supports:
```
         background: undefined,
         border: undefined,
         font: {
           weight: undefined,
           size: undefined,
         },
        gap: 'small', // previously defined inline, but now giving users to style via theme and setting inline value as default
         hover: {
           background: undefined,
        },
        pad: undefined,
```
`background` and `border` will apply to the overall `th`, but the remaining will affect the styling of the content instead. In the case that the content is just a Text, then it will be applied to the Box surrounding the Text. In the case that the content is sortable and becomes a Button, then the theming will be applied to that Button. Additionally, the column label (whether Text or Button) now has `flex="grow"` which allows it to take up any remaining space in the column.

Added a jest test. 

In storybook for DataTable Sort, apply the following theme locally:
```
const customTheme = {
      dataTable: {
        header: {
          background: 'skyblue',
          border: {
            color: 'green',
            size: 'medium',
          },
          gap: 'none',
          font: {
            weight: 'bold',
          },
          hover: {
            background: {
              color: 'light-2',
            },
          },
        },
      },
    };
```
You should see the following as a result (the grey background was where the mouse was hovered over):
Screen Shot 2020-09-22 at 4 01 35 PM

Closes https://github.com/grommet/grommet/issues/4454

Yes, updated here.

Yes.

Yes, backwards compatible.

### DIFF
--- a/src/js/components/Button/StyledButtonKind.js
+++ b/src/js/components/Button/StyledButtonKind.js
@@ -70,21 +70,18 @@ const basicStyle = props => css`
 `;
 
 // CSS for this sub-object in the theme
-const kindPartStyles = (obj, theme, colorValue) => {
+export const kindPartStyles = (obj, theme, colorValue) => {
   const styles = [];
-  if (obj.padding) {
-    if (obj.padding.vertical || obj.padding.horizontal)
+  if (obj.padding || obj.pad) {
+    // button uses `padding` but other components use Grommet `pad`
+    const pad = obj.padding || obj.pad;
+    if (pad.vertical || pad.horizontal)
       styles.push(
-        `padding: ${theme.global.edgeSize[obj.padding.vertical] ||
-          obj.padding.vertical ||
-          0} ${theme.global.edgeSize[obj.padding.horizontal] ||
-          obj.padding.horizontal ||
-          0};`,
+        `padding: ${theme.global.edgeSize[pad.vertical] ||
+          pad.vertical ||
+          0} ${theme.global.edgeSize[pad.horizontal] || pad.horizontal || 0};`,
       );
-    else
-      styles.push(
-        `padding: ${theme.global.edgeSize[obj.padding] || obj.padding || 0};`,
-      );
+    else styles.push(`padding: ${theme.global.edgeSize[pad] || pad || 0};`);
   }
   if (obj.background)
     styles.push(
--- a/src/js/components/Button/StyledButtonKind.js
+++ b/src/js/components/Button/StyledButtonKind.js
@@ -2,11 +2,10 @@ import styled, { css } from 'styled-components';
 
 import {
   activeStyle,
-  backgroundStyle,
   disabledStyle,
   focusStyle,
   genericStyles,
-  normalizeColor,
+  kindPartStyles,
 } from '../../utils';
 import { defaultProps } from '../../default-props';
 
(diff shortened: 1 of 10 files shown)


src/js/components/Button/StyledButtonKind.js
```diff
@@ -70,21 +70,18 @@ const basicStyle = props => css`
 `;
 
 // CSS for this sub-object in the theme
-const kindPartStyles = (obj, theme, colorValue) => {
+export const kindPartStyles = (obj, theme, colorValue) => {
```
ericsoderberghp, line 73: should we move this to utils?

src/js/components/DataTable/doc.js
```diff
@@ -327,6 +327,63 @@ export const themeDoc = {
...
+      size: cover | contain | string
+    }`,
+    defaultValue: undefined,
+  },
+  'dataTable.header.border': {
+    description: 'Any valid Box border value.',
+    type: 'string | object',
+    defaultValue: undefined,
+  },
+  'dataTable.header.font.weight': {
```
ericsoderberghp, line 350: I wonder if this should be `text` instead of `font`. This would allow us to use `color` here as well.

ericsoderberghp (commented):
Initial comments. I'd still like to delve into the meat of Text, Box, Button stuff.

src/js/components/Button/StyledButtonKind.js
taysea: Yes, I can move it to utils -- possible that in future we would leverage it for other component styles.

src/js/components/DataTable/doc.js
taysea: That's a good point. I was aiming to mirror other theme areas that use the `font` and weight/size convention, but those might not support color. Let me see if any of those other `font` theme objects support color or not.

src/js/components/Button/StyledButtonKind.js
taysea: moved to utils in recent commit

src/js/components/DataTable/doc.js
taysea: A follow-up here @ericsoderberghp. Applying `theme.dataTable.header.color` will affect the text color -- as is inline with the convention of other components. the `font` object would purely support weight and size. I can add `header.color` to the docs for clarity here if you are fine with that convention?

ericsoderberghp: Thanks for looking into it. I'm OK with what you've already got.

src/js/components/DataTable/Header.js
```diff
@@ -18,6 +18,42 @@ import {
...
+
+// build up CSS from basic to specific based on the supplied sub-object paths.
+// adapted from StyledButtonKind to only include parts relevant for DataTable
+const buttonStyle = ({ theme }) => {
+  const styles = [];
+  // don't apply background or border on button. these
+  // should only apply to the `th`
+  const contentStyles = { ...theme.dataTable.header };
+  delete contentStyles.background;
+  delete contentStyles.border;
```
ericsoderberghp, line 31: Should we also be omitting `color` and `font`, since those are button-y?

src/js/components/DataTable/Header.js
```diff
@@ -91,8 +128,21 @@ const Header = ({
...
-            let content =
-              typeof header === 'string' ? <Text>{header}</Text> : header;
+            // don't apply background or border on other contents of the cell
+            const contentStyles = { ...theme.dataTable.header };
+            delete contentStyles.background;
+            delete contentStyles.border;
+
+            let content;
+            if (typeof header === 'string') {
+              content = <Text {...contentStyles.font}>{header}</Text>;
```
ericsoderberghp, line 138: Shouldn't `color` should be passed here?

src/js/components/DataTable/Header.js
```diff
@@ -91,8 +128,21 @@ const Header = ({
             verticalAlign,
             size,
           }) => {
-            let content =
-              typeof header === 'string' ? <Text>{header}</Text> : header;
+            // don't apply background or border on other contents of the cell
+            const contentStyles = { ...theme.dataTable.header };
+            delete contentStyles.background;
+            delete contentStyles.border;
```
ericsoderberghp, line 134: Shouldn't we remove `font` and `color` here too?

src/js/components/DataTable/Header.js
```diff
@@ -18,6 +18,42 @@ import {
...
+// adapted from StyledButtonKind to only include parts relevant for DataTable
+const buttonStyle = ({ theme }) => {
+  const styles = [];
+  // don't apply background or border on button. these
+  // should only apply to the `th`
+  const contentStyles = { ...theme.dataTable.header };
+  delete contentStyles.background;
+  delete contentStyles.border;
+
+  const obj = contentStyles;
```
ericsoderberghp, line 33: Why do we need to create a const object which is set to another const object?

src/js/components/DataTable/Header.js
ericsoderberghp: I'm wondering if we should have a utility function inside this file that separates the `theme.dataTable.header` object into three objects: ` const [cellProps, layoutProps, textProps] = extricateThemeProps(theme);`? Then, `StyledHeaderCellButton` could use `layoutProps` and be sure it is using the same as the content `Box`. Just a thought.

taysea: I like the idea of a utility function in this file. let me work on that

src/js/components/DataTable/Header.js
taysea: If color is defined, then it would we placed on the box surrounding this and propagate down to the Text

src/js/components/DataTable/Header.js
taysea: Good point, this is unnecessary. Will simplify in upcoming commit.

src/js/components/DataTable/Header.js
taysea: For clarity, I've included `color` in the `textProps` section of the utility function added in a recent commit.

src/js/components/DataTable/Header.js
taysea: Added `separateThemeProps` utility function in a recent commit that returns an array `[cellProps, layoutProps, textProps]` -- I felt that this function name would be more intuitive, but I liked the structure you proposed.

src/js/components/DataTable/Header.js
taysea: Handled this with new utility function `separateThemeProps`

ericsoderberghp (approved):

PR #4549, opened 2020-09-26
Added RTL support for the meter bar component (Signed-off-by: Nizar Jubran <Nizaretto@gmail.com> )

Net301:
#### What does this PR do?

Adds RTL support to the bar Meter.

Meter/utils.js getPathByDirection() -> creates the path according to the direction proprty.

Manual testing in codesandbox.

Checking the sandbox of the usual bar and the RTL bar and ensuring there is a consistency.

The Meter lacks RTL support, where you want to display its progress from right to left rather than left to right (this is a feature that I personally needed in one of my apps).

image

Yes, the docs needs to be updated to reflect the new property (direction) in the grommet meter.

Yes,

backward compatible.

### DIFF
--- a/src/js/components/DataTable/__tests__/DataTable-test.js
+++ b/src/js/components/DataTable/__tests__/DataTable-test.js
@@ -351,15 +351,10 @@ describe('DataTable', () => {
               aggregate: 'sum',
               footer: { aggregate: true },
             },
-            {
-              property: 'obj2.value',
-              header: 'object 2',
-              render: datum => datum.obj2.value,
-            },
           ]}
           data={[
-            { a: 'one', obj: { value: 1 }, obj2: { value: 10 } },
-            { a: 'two', obj: { value: 2 }, obj2: { value: 20 } },
+            { a: 'one', obj: { value: 1 } },
+            { a: 'two', obj: { value: 2 } },
           ]}
         />
       </Grommet>,
--- a/src/js/components/DataTable/__tests__/DataTable-test.js
+++ b/src/js/components/DataTable/__tests__/DataTable-test.js
@@ -351,10 +351,15 @@ describe('DataTable', () => {
               aggregate: 'sum',
               footer: { aggregate: true },
             },
+            {
+              property: 'obj2.value',
+              header: 'object 2',
+              render: datum => datum.obj2.value,
+            },
           ]}
           data={[
-            { a: 'one', obj: { value: 1 } },
-            { a: 'two', obj: { value: 2 } },
+            { a: 'one', obj: { value: 1 }, obj2: { value: 10 } },
+            { a: 'two', obj: { value: 2 }, obj2: { value: 20 } },
           ]}
         />
       </Grommet>,
--- a/src/js/components/Meter/Bar.js
+++ b/src/js/components/Meter/Bar.js
@@ -5,7 +5,11 @@ import { defaultProps } from '../../default-props';
 import { parseMetricToNum } from '../../utils';
 
 import { StyledMeter } from './StyledMeter';
-import { strokeProps, defaultColor, getPathByDirection } from './utils';
+import {
+  strokeProps,
+  defaultColor,
+  getPathByDirection as getPathByMargin,
+} from './utils';
 
 const Bar = forwardRef((props, ref) => {
   const {
@@ -15,7 +19,7 @@ const Bar = forwardRef((props, ref) => {
     size,
     thickness,
     values,
-    direction,
+    margin,
     ...rest
   } = props;
   const theme = useContext(ThemeContext) || defaultProps.theme;
@@ -37,7 +41,7 @@ const Bar = forwardRef((props, ref) => {
 
       const key = `p-${index}`;
       const delta = (value * (width - 2 * capOffset)) / max;
-      const d = getPathByDirection(direction, { start, mid, delta, width });
+      const d = getPathByMargin(margin, { start, mid, delta, width });
       const colorName =
         color || defaultColor(index, theme, values ? values.length : 0);
       let hoverProps;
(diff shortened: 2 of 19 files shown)


Net301:
The integrity check fails on "grommet-icons" with yarn install (I think it works with npm).

error https://github.com/grommet/grommet-icons/tarball/stable: Integrity check failed for "grommet-icons" (computed integrity doesn't match our records, got "sha512-6QAqH//uaOVrkbTkA7HhtVev52v7mXTLj3MWtYYq41rm3RT1/kHEaTPwVC9EVw6ZwMxr+DvBaCi+XH1AR0Mh3Q== sha1-8WQd6Fp3vuB+jbBVbJM4wojPQss=")
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.

Net301:
I have also found that some JEST tests (that are not related to the files edited here) failed.
these tests, if am not mistaken use snapshots, and apparently there was a change that broke them. this needs to be checked in the future.

ShimiSun:
Make sure you are using node version > 14.4.

src/js/components/Meter/Bar.js
```diff
@@ -5,10 +5,19 @@ import { defaultProps } from '../../default-props';
...
 const Bar = forwardRef((props, ref) => {
-  const { background, max, round, size, thickness, values, ...rest } = props;
+  const {
+    background,
+    max,
+    round,
+    size,
+    thickness,
+    values,
+    direction,
```
ShimiSun, line 18: Using `direction` for 'rtl' wouldn't be consistent with the current Box `direction` prop.
I'm thinking we should leverage the existing `margin` prop with the value `start` or `end` for the RTL behavior, consistently with how it is done for other components with rtl functionality.
Rleated to #3629

ShimiSun (commented):
Thanks for this PR! 
Once the PR direction is approved, the doc.js and index.d.ts should be completed as well.

IanKBovard:
> The integrity check fails on "grommet-icons" with yarn install (I think it works with npm).
> 
> error https://github.com/grommet/grommet-icons/tarball/stable: Integrity check failed for "grommet-icons" (computed integrity doesn't match our records, got "sha512-6QAqH//uaOVrkbTkA7HhtVev52v7mXTLj3MWtYYq41rm3RT1/kHEaTPwVC9EVw6ZwMxr+DvBaCi+XH1AR0Mh3Q== sha1-8WQd6Fp3vuB+jbBVbJM4wojPQss=")
> info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.

You need to update your yarn.lock files. grommet-icons was just updated recently.

```
rm -rf yarn.lock
rm -rf package-lock.json
rm -rf node_modules
yarn cache clean
yarn
```

src/js/components/Meter/Bar.js
Net301: * I have changed the prop controlling the meter direction from "direction" to "margin" where if the meter has a margin.start  (for example <meter margin={{start:"xsmall"}} ... />) then it behaves as an rtl element.
* Now there's no need to redefine the types as the margin is already defined there.
* the calendar test is failing - although i didn't touch that area in my PR.

Net301: @ShimiSun some UI tests are failing, i think because its yet to be reviewed by someone. I haven't used chromatic before, so It is a bit unclear for me.

halocline: > @ShimiSun some UI tests are failing, i think because its yet to be reviewed by someone. I haven't used chromatic before, so It is a bit unclear for me.

@Net301 - You are correct that the Chromatic snapshot just needs to be reviewed since it is a newly added Storybook story. So, looking good.

Also, per:
> Thanks for this PR!
> Once the PR direction is approved, the doc.js and index.d.ts should be completed as well.

It looks like next steps are for the direction to be approved and then some documentation and type definitions.

Net301: @halocline
- I have used the existing types where the margin.start type already exists in index.d.ts . so I don't think the types need extra modification.
- Regarding the documentaiton: As I understand from the code I should update the DocumentedMeter.propTypes object, override the margin child and explain that if a start is added it transforms the bar meter to "rtl". my questions here is: how do I view the documentation in order to manually test it before pushing?
- Regarding the chromatic review, how is the process done ?  by the core team ?

halocline: Hi @Net301 -

Regarding documentation, yes in `Meter/doc.js` modify the margin property. To test, you can run `yarn generate-readme` which is a script that generates all of the README markdown files for each component. After running, you will see that `Meter/README.md` will have been modified and can be manually inspected there. As FYI, the `generate-readme` script is also run when `yarn test-update` script is executed.

Regarding Chromatic review, yes, someone from the core team will review.

src/js/components/DataTable/__tests__/DataTable-test.js
```diff
@@ -351,15 +351,10 @@ describe('DataTable', () => {
               aggregate: 'sum',
               footer: { aggregate: true },
             },
-            {
```
ericsoderberghp, line 354: Why are we removing this as part of this change?

ericsoderberghp (commented):
A couple of thoughts on this ...
- `margin.start` seems a bit unintuitive as a property name for this behavior. It doesn't seem to align with how we use `margin` elsewhere. I'm wondering if we should add a `reverse` property. This would work for both circle and bar types.
- I'm wondering if we should automatically behave this way if `<Grommet dir="rtl" />` is specified.

ShimiSun:
Hi @Net301, thanks for this PR, can you please follow up with the comments so we can target it for the next release.

ShimiSun:
@Net301 hi buddy, any updates on this PR? The release is coming and we'll appreciate your response.

jcfilben:
Closing due to inactivity. I opened a new issue for this: https://github.com/grommet/grommet/issues/6358

Issue #4191, opened 2020-06-14
How to get touched object for forms on onChange not just on onSubmit


ImSolitude:
I am trying to add a prompt for Unsaved changes  in my Form, So I want to check If the input is dirty or not,

I think it would be a good idea to get the `touched` object on `onChange` for forms, Like we have it right now on `onSubmit`:

`onSubmit={({ value, touched }) => {}};`
vs
`onChange={({ value}) => {}};`
:(

netochaves:
Hey @ShimiSun, correct me if I'm wrong but this will not be a breaking change? Right now `onChange` is receiving as parameter the `form value` but if we add another property like `touched` it will receiving an object and you will have to access the object to get the `form value` so this can break some implementations. 

I think this will be an awesome feature and I want to implement it, my worries are with the backwards compatible.

ShimiSun:
@netochaves good observation. I don't think this will be a breaking a change, we'll know better once we'll have a concrete PR 
 too look at, would you like to take a stab at it and see if you are running into potential changes and submit a draft with your observations? It will help us keep the conversation running.

ShimiSun:
Thank you @netochaves for adding this enhancement! It will be available on our next official release.

PR #4550, opened 2020-09-27
Feat(Form): add touched for onChange

netochaves:
#### What does this PR do?
Adds `touched` option for onChange in `Form`.

`src/js/components/Form/Form.js`
Unit testing
With storybook
no
#4191
Yes
I think so
I think this is a `breaking change` because before we were sending just the `form value` at onChange but now we're returning an object that contains the `touched` object and the `form value` and this can break some implementations. See the tests and the stories that I had to change.

### DIFF
--- a/src/js/components/Form/Form.js
+++ b/src/js/components/Form/Form.js
@@ -153,19 +153,18 @@ const Form = forwardRef(
         nextComponentValue => {
           if (name) {
             // we have somewhere to put this
+            const nextTouched = { ...touched };
+            nextTouched[name] = true;
+
             if (!touched[name]) {
               // don't update if not needed
-              setTouched(prevTouched => {
-                const nextTouched = { ...prevTouched };
-                nextTouched[name] = true;
-                return nextTouched;
-              });
+              setTouched(nextTouched);
             }
 
             const nextValue = { ...value };
             nextValue[name] = nextComponentValue;
             setValueState(nextValue);
-            if (onChange) onChange(nextValue);
+            if (onChange) onChange({ value: nextValue, touched: nextTouched });
           }
           if (initialValue !== undefined) setInputValue(nextComponentValue);
         },
@@ -264,7 +263,8 @@ const Form = forwardRef(
         onReset={event => {
           if (!valueProp) {
             setValueState(defaultValue);
-            if (onChange) onChange(defaultValue);
+            if (onChange)
+              onChange({ value: defaultValue, touched: defaultTouched });
           }
           setTouched(defaultTouched);
           setValidationResults(defaultValidationResults);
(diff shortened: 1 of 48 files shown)


ericsoderberghp:
I like the idea of providing `touched` updates. But, I don't think we can take this change as is, since as you mention the `onChange()` arguments are a breaking change. How about something like this instead: `onChange=((value, { touched }) => {}}`?

netochaves:
I liked this approach @ericsoderberghp , I try to replicate the same API of `onSubmit`, but I agree that handle withe the break change it's a bigger problem. I'll change it, thanks.

ShimiSun:
@netochaves thank you, any plans to update this PR so we can plan it to the next release?

netochaves:
Hi @ShimiSun I updated the PR, sorry for the delay, I got a little busy at work.

src/js/components/Form/Form.js
```diff
@@ -153,19 +153,18 @@ const Form = forwardRef(
...
+            nextTouched[name] = true;
+
             if (!touched[name]) {
               // don't update if not needed
-              setTouched(prevTouched => {
-                const nextTouched = { ...prevTouched };
-                nextTouched[name] = true;
-                return nextTouched;
-              });
+              setTouched(nextTouched);
```
ericsoderberghp, line 161: I seem to recall that we had to use the longer form of `setTouched()` to deal with concurrent updates to values, which I think could come programmatically? I don't remember the details.

But, based on how we are changing `valueState`, I can't see how we couldn't do what you have. Just thought I'd add this comment for reference.

netochaves: I had to make this change because `touch` was not being set in the state on the first input change only on the second.

src/js/components/Form/index.d.ts
```diff
@@ -10,7 +10,7 @@ export interface FormProps<T> {
   errors?: {};
   infos?: {};
   messages?: { invalid?: string; required?: string };
-  onChange?: (value: T) => void;
+  onChange?: (event: FormExtendedEvent<T>) => void;
```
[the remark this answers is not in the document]
netochaves: That's right. I forgot to change that. Thanks.

src/js/components/Form/index.d.ts
```diff
@@ -10,7 +10,7 @@ export interface FormProps<T> {
   errors?: {};
   infos?: {};
   messages?: { invalid?: string; required?: string };
-  onChange?: (value: T) => void;
+  onChange?: (value: T, options: { touched?: Record<string, boolean> }) => void;
```
netochaves, line 13: I had to give this object a name and to be honest, I don't know if 'options' is a good name, but I can't think of a better name.

ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution!

ShimiSun:
Thank you @netochaves ! 💟

PR #4555, opened 2020-09-28
Updating yarn.lock with latest grommet-icons

halocline:
#### What does this PR do?

Updates yarn.lock integrity check for grommet-icons stable branch.

yarn.lock

At first I thought including the stable branch of grommet-icons was a mistake because the package.json specifies "^4.2.0", but then realized that it is being included because of grommet-theme-hpe.
Screen Shot 2020-09-28 at 3 35 19 PM

No.

No.

Backwards compatible.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6517,8 +6517,8 @@ grommet-icons@^4.2.0:
     grommet-styles "^0.2.0"
 
 "grommet-icons@https://github.com/grommet/grommet-icons/tarball/stable":
-  version "4.4.0"
-  resolved "https://github.com/grommet/grommet-icons/tarball/stable#61edce8fd17d26c15db3c1826d011b4c62950dd4"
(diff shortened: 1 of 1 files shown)


ericsoderberghp (approved):

ShimiSun:
I thought we removed the icons dependency from grommet-theme-hpe, and we at least changed it to not point to stable. we'll need to take a look at why this issue is still occurring.

halocline:
> I thought we removed the icons dependency from grommet-theme-hpe, and we at least changed it to not point to stable. we'll need to take a look at why this issue is still occurring.

It does look like grommet-theme-hpe was moved to a peer-dependency, but perhaps is in need of being released after the update?

ShimiSun:
@halocline I think that in case we are not pointing to the theme stable it should be released again to apply

Issue #4487, opened 2020-09-02
DataTable - Aggregation and custom render of nested property don't like each other


halocline:
### Expected Behavior

If I have a data set in the shape of:
```
[
    {
        name: 'Johannes',
        amountOfMoneyInWallet: 2.73,
        favorites: {
            color: 'blue',
            movie: 'Butch Cassidy and the Sundance Kid',
            number: 3,
        },
    },
    ...
]
```

I should be able to aggregate `amountOfMoneyInWallet` as well as have a custom render function like `render: (datum) => <Text color={datum.favorites.color}>{datum.favorites.color}</Text>` for another column.

### Actual Behavior

An error is thrown:
![Screen Shot 2020-09-02 at 12 22 27 PM](https://user-images.githubusercontent.com/1756948/92021432-fd1fbb00-ed16-11ea-8af5-bcbf0dca038a.png)
 

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/aggregating-on-property-of-nested-object-rs27r?file=/src/App.js

[Related Issue](https://github.com/hpe-design/design-system/issues/1099)

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.15 / Stable
- Browser Name and version:
- Operating System and version (desktop or mobile):

halocline:
@MikeWalrath I think we might of lost track of an issue (which had similar subject) and doubled up on this one (https://github.com/grommet/hpe-design-system/issues/1099). Will have to do some digging.

PR #4556, opened 2020-09-29
fix-4487 nested obj + aggregate with custom render

halocline:
#### What does this PR do?

Fixes data + column configuration which causes app crash.

Closes https://github.com/grommet/grommet/issues/4487
Closes https://github.com/grommet/hpe-design-system/issues/1099

Review error reproduction: https://codesandbox.io/s/aggregating-on-property-of-nested-object-rs27r?file=/src/App.js

Modified unit test to reproduce the error, then placed fix.

Recreate https://codesandbox.io/s/aggregating-on-property-of-nested-object-rs27r?file=/src/App.js

https://github.com/grommet/grommet/issues/4487
https://github.com/grommet/hpe-design-system/issues/1099

No.

Sure.

Backwards compatible.

### DIFF
--- a/src/js/components/DataTable/Cell.js
+++ b/src/js/components/DataTable/Cell.js
@@ -30,7 +30,7 @@ const Cell = ({
   const context = useContext(TableContext);
 
   let content;
-  if (render) {
+  if (render && datum[property]) {
     content = render(datum);
   } else if (value !== undefined) {
     content = value;
--- a/src/js/components/DataTable/Cell.js
+++ b/src/js/components/DataTable/Cell.js
@@ -16,7 +16,15 @@ const normalizeProp = (name, rowProp, prop) => {
 const Cell = ({
   background: backgroundProp,
   border,
-  column: { align, pin: columnPin, property, render, verticalAlign, size },
+  column: {
+    align,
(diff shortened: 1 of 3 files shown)


halocline:
Need to investigate Chromatic snapshot changes.

src/js/components/DataTable/Cell.js
```diff
@@ -30,7 +30,7 @@ const Cell = ({
   const context = useContext(TableContext);
 
   let content;
-  if (render) {
+  if (render && datum[property]) {
```
ericsoderberghp, line 33: What if the caller wants their custom renderer to provide some indication that there is no data in this cell? I'm wondering if detecting a lack of the property should be the purview of the render function.

halocline: Good point. I can see use cases where identifying missing data would be useful. In that case, need to revisit how the footer values are being assembled.

halocline:
Chromatic test changes show that in examples using `<Meter />` were previously displaying an empty Meter in the footer even though column definitions did not specify aggregation for the metered column.

ericsoderberghp (approved):

Issue #4372, opened 2020-08-03
Stack anchor is offset


LiveFreeAndRoam:
### Expected Behavior

When setting the `anchor="top-right"`, the expectation is that the stacked element will sit in the top-right position.

### Actual Behavior

When setting the `anchor="top-right"`, the stacked element is offset from the top-right position.  The workaround is to wrap the stacked element with a `Box`.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/stack-anchor-needs-a-box-orfyy

![image](https://user-images.githubusercontent.com/8231064/89230293-5c11e780-d626-11ea-83e3-f6d10dbffb15.png)

Here is the snippet:

```
      <Heading level="2">Top-Right: Not working</Heading>
      <Button>
        <Stack anchor="top-right">
          <Notification />
          <StatusGoodSmall color="red" size="small" />
        </Stack>
      </Button>
      <Heading level="2">Top-Right: Works when wrapped</Heading>
      <Button>
        <Stack anchor="top-right">
          <Notification />
          <Box>
            <StatusGoodSmall color="red" size="small" />
          </Box>
        </Stack>
      </Button>
```

### Steps to Reproduce

See codesandbox.

### Your Environment

- Grommet version: 2.14.0
- Browser Name and version: Chrome and Firefox
- Operating System and version (desktop or mobile): Windows 10

LiveFreeAndRoam:
Some more data points after a chat with Shimi...

> One thought regarding this issue is that the default props of icon may cause it. If you'll replace the usage of SmallStatusIcon with simply a <Box pad='xsmall' round='full' background='green' /> that should work without the wrapper, can you give it a try?

I tested it and here are my findings...

1. The stacked element is a `Box`, so it seems to be the same solution as wrapping the stacked icon in a `Box`.  Perhaps this should be a mandatory requirement to wrap stacked elements in `Box`?

2. When using a `Box` without any children, you need `pad='xsmall'` , to give the box a dimension that can be filled.  This is not immediately intuitive.

3. The dimension of `<Box pad="small">` is larger than the dimension of `<Icon size="small">`.  While that is not a big deal here, it may be a factor in stacks that need more precise positioning.

![image](https://user-images.githubusercontent.com/8231064/89239338-3a235f80-d63c-11ea-8cd4-0a64d0b59805.png)

ShimiSun:
@britt6612, can you confirm this issue is solved with your recent PR as well?

[...]

PR #4560, opened 2020-10-01
Stack anchor offset Resolve Issue #4372

jmsunseri:
#### What does this PR do?

The purpose of this PR was to help resolve Issue #4372 Icons when placed directly in the stack seem to be offset

Only one file has been manually changed Stack.js

I ran storybook to try and make sure the position of the icon was as expected in the Stack section.  

There are several components where the snapshot had to be updated
DataChart
Carousel
Avatar
Stack
Diagram
Video

I would verify that these have not changed visually

[#4372](https://github.com/grommet/grommet/issues/4372)

Stacked Avatars

image

Stacked Demo

image

Probably not

I think It should not be breaking


halocline:
The number of snapshot changes + the number of items flagged as visual baseline changes in the Chromatic UI Tests give me pause. Also, since so much time has passed between submittal and current state, I think a first next step is to update this branch with the latest master.

@jmsunseri, I know review on this has been slow. Is this something on which you'd still be interested working?

Secondly, when adding the wrapping `Box` around the `{child}` we're effectively adding an additional `div` in DOM, ensuring that the first child is a div. I wonder if we might be missing the root issue which is "why StyledStackLayer behaves differently when the child is an `svg`?"

ericsoderberghp:
closed due to inactivity

Issue #4353, opened 2020-07-28
DateInput - Calendar not to show 6th row if all in next month


maryjufang:
Calendar not to show 6th row if all in next month

### URL, screen shot, or Codepen exhibiting the issue
![image](https://user-images.githubusercontent.com/9951106/88704118-a69ddc00-d0db-11ea-84bc-9b41a37ee102.png)

ShimiSun:
The example is taken from DateInput https://storybook.grommet.io/?path=/story/dateinput--range

halocline:
@ShimiSun Confirming we want to treat this as a feature enhancement, correct?

ShimiSun:
Yea, otherwise it might be considered as a breaking change.

halocline:
Looking into the work here, relates to "showAdjacentDays" functionality, I am noticing a few buggy behaviors which should be addressed. Documenting them here, then can broken out into separate issues if needed:
- Navigating to previous months, then forward does not advance the calendar dates and also results in incorrect styling of the "adjacent days"
- Navigating to previous months, then forward eventually results in the calendar dates getting "stuck" and not advancing
- Navigating forward a single month increments correctly, displaying the 1st of the month in the first row. Navigating forward an additional month adds a week from the previous month to the first row and does not advance the calendar. The calendar gets "stuck" here. Additional forward navigation ends up adding and removing a week from the view, but does not advance the calendar (always displaying the last week of the original month).

Steps to reproduce to be added to this comment.

ShimiSun:
@jcfilben check out the above analysis for testing effort.

halocline:
**Update**. Have proposed solution which would prevent the 6th row from displaying if all dates in next month. Need to add tests + question re: API.

@ShimiSun What are your thoughts of this being viewed as a bug vs. an enhancement? If enhancement, for backwards compatibility, I think we need to expose a new property. If viewed as bug, then we can implement without a new prop.

PR #4579, opened 2020-10-06
add support for trimming adjacent days in 6th row

halocline:

#### What does this PR do?

Closes https://github.com/grommet/grommet/issues/4353

Allows for a caller to `showAdjacentDays` but choose to not render the calendar's 6th row if all days fall into the next month.

To allow for backwards compatibility, this approach extends the types/values accepted by `showAdjacentDays`.

Calendar.js

Storybook, Jest

Includes new test in both Jest and Storybook

Storybook - Calendar/'Show adjacent days' story

https://github.com/grommet/grommet/issues/4353

Yes. Included.

Yes.

Backwards compatibile.


src/js/components/Calendar/Calendar.js
```diff
@@ -474,6 +474,23 @@ const Calendar = forwardRef(
             <StyledDay sizeProp={size} fillContainer={fill} />
           </StyledDayContainer>,
         );
+      } else if (
+        /* Do not show adjacent days in 6th row if all days 
+        fall in the next month */
+        showAdjacentDays === 'trim' &&
+        otherMonth &&
+        weeks.length === 5 &&
+        day.getDate() >= days.length + 1
```
taysea, line 483: Just trying to understand the logic here -- I'm not exactly following this last comparison. Could you elaborate?

src/js/components/Calendar/Calendar.js
```diff
@@ -474,6 +474,23 @@ const Calendar = forwardRef(
             <StyledDay sizeProp={size} fillContainer={fill} />
           </StyledDayContainer>,
         );
+      } else if (
+        /* Do not show adjacent days in 6th row if all days 
+        fall in the next month */
+        showAdjacentDays === 'trim' &&
```
taysea, line 480: Nice, I like `trim`

taysea (commented):
Just a small question

src/js/components/Calendar/Calendar.js
halocline: Good question.

So, the loop iterates through a range of dates, building up a week's worth of days in a `days` array. Whenever, the day is the first day of a week, it pushes that week of days into a `weeks` array and then resets `days` to be an empty array.

This condition comes into play specifically when building up the 6th week (i.e. last calendar row).

In the 6th week we know that 2 scenarios could occur:
- a) The beginning of the row will be populated with the end dates of the current month and the remaining days of the row filled in by dates in the next month (we want to populate row 6 here)
- b) The current month's dates were exhausted in row 5 and the minimum value for the first day in row 6 is the 1st of next month (we do not want to populate row 6 here)

If scenario **a)**,  by the time the `otherMonth` condition is satisfied the `days` will be populated with at least one date of the current month and the array length will be greater than 1 (i.e. `getDate()` of the first day of the next month) and the condition will not be satisfied.

If scenario **b)**, the `otherMonth` condition is satisfied on the first pass when days.length will be 0, and the `getDate()` will be greater or equal to 1 (i.e. the first of the next month). In this case, the condition is met and we add blank placeholder days instead of "visible" days.

Phew!!! That was a lot!

Hopefully you were able to follow that, if not, I can try again. 😬

taysea: Gotcha, thanks for the detailed explanation. It took me better understanding how the `days` and `weeks` arrays are built up to grasp the logic fully, but now it makes sense. I think that will be the case for others that may come into this file in the future but hoping the comment that you are adding will help serve as a high-level explanation.

halocline: @taysea I included the comment we discussed. Appreciate your feedback and suggestions.

taysea (approved):
LGTM. thanks for adding that comment.

IanKBovard:
This LGTM too. Like the changes and the prop, clean and simple.

ericsoderberghp (approved):

Issue #4092, opened 2020-05-17
TextInput component lose's ref 


ashcherbyak:
The use of ref on TextInput does not work as expected. When I try to focus input on first render I get error, because of ref is undefined

```
 const StyledTextInput = styled(TextInput)`
  border: ${({ theme }) => `1px solid ${theme.global.colors.paletteGray10}`};
  padding-top: ${({ theme }) => theme.global.edgeSize.xsmall};
  padding-bottom: ${({ theme }) => theme.global.edgeSize.xsmall};
  border-radius: 0.3rem;
  width: 100%;
`;

const searchRef = useRef();

  useEffect(() => {
    searchRef.current.focus();
  }, []);

    <StyledTextInput
          ref={searchRef}
          type="text"
          onChange={onChangeTextInput}
          value={searchText}
     />
```
When I tryed to use simple ```<input>``` all works as expected 
```
 <input
          ref={searchRef}
          type="text"
          onChange={onChangeTextInput}
          value={searchText}
          placeholder={searchPlaceholder}
          data-automation-id={`${dataAutomationId}-search`}
        />
```
### Expected Behavior
ref should point on TextInput

### Actual Behavior

![image](https://user-images.githubusercontent.com/62097255/82148443-d7040700-985c-11ea-83dc-ca8f49144bc5.png)

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: v2.10.0
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): macOs

coltonw:
This exact issue didn't happen for me in v2.13.0 but now DOES happen in v2.14.0.  It seems like the PR above actually caused this issue to crop up in certain circumstances.

ShimiSun:
@coltonw Can you share a codesandbox that reproduces the issue so we could investigate it further?

coltonw:
https://codesandbox.io/s/cannot-read-property-focus-of-undefined-j0cjp

Seems like it is just failing in our unit tests because our unit tests didn't have a theme wrapper before.  If you wrap it in a Grommet component with a theme, it works fine.  Not sure how defensive you want to code that but figured it was worth noting at least for future developers.

IanKBovard:
I think this issue can be closed. This looks like, to me, an implementation issue. In my opinion, I don't think grommet needs to be defensive about making sure users wrap their app in `<Grommet></Grommet>`, as it is stated pretty clearly on the getting started guide.

[...]

PR #4583, opened 2020-10-07
Class components don't receive updated refs - Don't merge

IanKBovard:
#### What does this PR do?
Mostly addresses issue #4092 to start a conversation about fixes

I have two storybook examples in TextInput, Test.js, and Test2.js, with comments.
Test.js shows the issue, and Test2.js shows a workaround.
There are also, some comments in TextInput.js on line 110
```
    // custom hook useForwardedRef seems to cause issues with class components.
    // createRef always returns null, even with componentDidUpdate.
    const inputRef = useForwardedRef(ref);
```

No testing, as no changes have been made yet.

Same as above

Class based components will not receive updated refs, at least when it comes to TextInput. 
TextInput refs are passed through a custom hook that evaluates the ref and returns it.
#4092 

No

No

No changes yet

### DIFF
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -106,6 +106,9 @@ const TextInput = forwardRef(
     const theme = useContext(ThemeContext) || defaultProps.theme;
     const announce = useContext(AnnounceContext);
     const formContext = useContext(FormContext);
+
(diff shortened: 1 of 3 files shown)


ShimiSun:
I think the suggested stories are a reasonable way to send refs for class components. Thank you for sharing those examples. I don't think there is something we should necessarily add to grommet core at this point, so I'll be closing this PR.

PR #4584, opened 2020-10-07
Update DataTable hover theme convention

taysea:
#### What does this PR do?

Swaps the DataTable hover convention from `dataTable.header.hover.background` to `dataTable.hover.header.background`. This aligns with the newer direction of defining themes via `component.state.[thing to be styled]`.

Test locally in storybook with HPE Design System theme and updated jest test.

In storybook, create a custom theme.

Backwards compatible.


taysea:
closing this PR because Eric and I realized that the initial convention was proper. We want the `hover` to be nested inside of the element it will be styling. In this case, `dataTable.header.hover` would be correct because we are hovering over the header.

PR #4585, opened 2020-10-07
Add DataTable resizer hover styling and increase grab target

### LINKED ISSUE #4582: Implement DataTable column resizing styling
### Expected Behavior

DataTable resizer should allow theming for hover styling and ability to specify a larger hit target (even if the visual is unchanged).

### Actual Behavior

DataTable resizer is only 1px wide and while this visually is nice, it doesn't leave a large enough area for the user to grab. If they are just slightly off from the resizer, they can't grab it.

https://storybook.grommet.io/?path=/story/datatable--sort

### Steps to Reproduce

DataTable Sort story, try to resize the column. The resizer is very difficult to grab.

### Your Environment

- Grommet version: 2.15.0
- Browser Name and version: Any

taysea:
#### What does this PR do?

This PR increases the grab area of the resizer by introducing an "invisible" box about the resizer that has some extra dimension. It does not affect the appearance of the resizer, but it creates a larger target area for the user to grab to resize a column. Previously, the resizer was only 1px wide, so grabbing it could be very difficult unless to were very precise. 

This PR also introduces the ability to style the resizer on hover by adjusting the border color, side, or size. These stylings affect the newly introduced box, which on hover, becomes visible.

src/js/components/DataTable/Resizer.js

Test locally with storybook and a jest test has been updated.

Apply custom theme in story book such as:
```
const myTheme = {
   dataTable: {
      hover: {
          resize: {
            border: {
              color: 'red',
              side: 'end',
              size: 'xsmall',
            },
          },
        },
   },
};
```

Closes #4582 

Yes, updated here.

Yes. Users can now introduce hover styling for DataTable resizer.

Backwards compatible.


src/js/components/DataTable/doc.js
```diff
@@ -378,12 +378,26 @@ export const themeDoc = {
...
     description: 'The pad around the contents of the header cell.',
     type: 'string | object',
     defaultValue: undefined,
   },
+  'dataTable.resize.hover.color': {
+    description: 'The color of the resizer when hovered over.',
+    type: 'string | object',
+    defaultValue: undefined,
+  },
+  'dataTable.resize.hover.side': {
```
ericsoderberghp, line 391: I'm trying to think of a case where a theme would want a different value for this. If we don't think it will be used. I think we should remove it.

src/js/components/DataTable/doc.js
```diff
@@ -378,12 +378,26 @@ export const themeDoc = {
...
+    description: 'The side of the resizer when hovered over.',
+    type: 'string',
+    defaultValue: undefined,
+  },
+  'dataTable.resize.hover.size': {
+    description: 'The size of the resizer when hovered over.',
```
ericsoderberghp, line 397: I think we should mention that the values for this correspond to `Box.border.size` values.

src/js/components/DataTable/doc.js
taysea: Right, since this is wrapping around an empty box with padding, putting any side except for `end` is just wrapping around something somewhat invisible. The default resize theme option sets `end` as default, but it can still be overridden. However, I don't think there is any loss from removing it here. Should I remove it only from the docs or from the theme structure as well?

ericsoderberghp: Hmm... my inclination is to automatically apply it if `resizer.hover.border` exists in the theme. On the other hand, I could see just being more explicit about this driving a Box border and that the theme probably wants this to be 'end'.

src/js/components/DataTable/doc.js
taysea: Good call.

src/js/components/DataTable/doc.js
taysea: Gotcha -- it feels like to be consistent with default resize, we should set this to `end` by default and then also be explicit in the docs that this drives a box border but that `end` is the recommended value.

ericsoderberghp: Are you thinking we have the default resize hover be `resize.hover.border = { side: 'end', color: 'text-strong' }`? It would seem odd to have only `side`, which wouldn't really do anything.

taysea: I guess it's a question of if we want to introduce a new default visual change for existing grommet users. The reason I was just thinking of defining `end` (or `end` and color: `border`) was so that there were no new visual behaviors on grommet by default. 

However, if we are not introducing any visual changes on grommet base.js, then we could just leave the values undefined in the theme and do:
1. add a check in the component where if `color` or `size` is specified for border, but side isn't that `side=end` gets applied automatically. 
2. recommend in the docs that `end` is used for this value

ericsoderberghp: I lean towards #1

taysea: gotcha, I would agree. Let me implement that.

ericsoderberghp (approved):

PR #4580, opened 2020-10-06
Improve Testing Coverage for Menu

esilverm:
Signed-off-by: Evan Silverman <es4753@nyu.edu>

#### What does this PR do?

`onPreviousMenuItem` had little coverage so I introduced tests to open the Menu with the `up` key and to navigate backwards through the Menu with `shift + tab` keys. Also added a test for bottom aligned drop.

#### Where should the reviewer start?

`/src/js/components/Anchor/__tests__/Menu-test.js`

#### What testing has been done on this PR?

Run `yarn test` to ensure that all tests pass and view coverage

#### What are the relevant issues?

#4539

#### Screenshots (if appropriate)
Before (80.52%)
Screen Shot 2020-10-06 at 3 26 56 PM

After (97.40%)
Screen Shot 2020-10-06 at 4 00 07 PM

#### Do the grommet docs need to be updated?

No

#### Is this change backwards compatible or is it a breaking change?

Backwards Compatible

PR #4586, opened 2020-10-07
Restore arrow key functionality to Menu component

esilverm:
#### What does this PR do?

This PR addresses missing [WAI-ARIA keyboard interaction](https://www.w3.org/TR/wai-aria-practices/#menu) handling which was removed in #3464 to fix onClick handling in FireFox. Up and Down arrow keys can be used to navigate between menu-items after they are in focus. This also raises the testing coverage for the Menu component by ~1.5%.

`src/js/components/Menu/Menu.js`

I implemented test cases to assert that the arrow keys function properly when being used. I also tested on Chrome, Firefox, Safari, Opera, and Edge to make sure that the bug which was fixed in #3464 had not returned when this code was restored. All browsers I tested with handled onClick properly.

 Run `yarn test` to ensure that all tests pass and view coverage. Run `yarn storybook` to ensure that keypress handling works properly.

See https://github.com/grommet/grommet/pull/4580#issuecomment-705089240
In the future, toggling the menu should automatically set the focus to the first menu element.

Improved test coverage (from #4580)
Screen Shot 2020-10-07 at 2 44 38 PM

No

Backwards Compatible

### DIFF
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -252,6 +252,8 @@ const Menu = forwardRef((props, ref) => {
             onTab={event =>
               event.shiftKey ? onPreviousMenuItem(event) : onNextMenuItem(event)
             }
+            onDown={onNextMenuItem}
+            onUp={onPreviousMenuItem}
             onEnter={onSelectMenuItem}
           >
             <ContainerBox background={dropBackground || theme.menu.background}>
--- a/src/js/components/Menu/__tests__/Menu-test.js
+++ b/src/js/components/Menu/__tests__/Menu-test.js
@@ -277,7 +277,7 @@ describe('Menu', () => {
   });
 
   test('shift + tab through menu until it closes', () => {
-    const { getByLabelText, container } = render(
+    const { getByLabelText, getByText, container } = render(
       <Grommet>
         <Menu
           id="test-menu"
@@ -297,28 +297,37 @@ describe('Menu', () => {
       keyCode: 32,
       which: 32,
     });
+
     fireEvent.keyDown(document.activeElement.firstChild, {
       key: 'Tab',
       keyCode: 9,
       which: 9,
     });
+    expect(getByText('Item 1').parentElement).toHaveFocus();
+
     fireEvent.keyDown(document.activeElement, {
       key: 'Tab',
       keyCode: 9,
       which: 9,
     });
+    expect(getByText('Item 2').parentElement).toHaveFocus();
+
     fireEvent.keyDown(document.activeElement, {
       key: 'Tab',
       keyCode: 9,
       which: 9,
       shiftKey: true,
     });
+    expect(getByText('Item 1').parentElement).toHaveFocus();
+
     fireEvent.keyDown(document.activeElement, {
       key: 'Tab',
       keyCode: 9,
       which: 9,
       shiftKey: true,
     });
+    expect(getByLabelText('Close Menu')).toHaveFocus();
+
     fireEvent.keyDown(document.activeElement, {
       key: 'Tab',
       keyCode: 9,
--- a/src/js/components/Menu/__tests__/Menu-test.js
+++ b/src/js/components/Menu/__tests__/Menu-test.js
@@ -276,7 +276,7 @@ describe('Menu', () => {
     expect(document.getElementById('test-menu__drop')).toBeNull();
   });
 
-  test('shift + tab through menu until it closes', () => {
+  test('shift + tab through menu until it closes (top control)', () => {
     const { getByLabelText, getByText, container } = render(
       <Grommet>
         <Menu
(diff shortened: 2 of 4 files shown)


ShimiSun:
@taysea , can you please elaborate on why the functionality was removed on #3464?
>This PR addresses missing WAI-ARIA keyboard interaction handling which was removed in #3464 to fix onClick handling in FireFox. Up and Down arrow keys can be used to navigate between menu-items after they are in focus. This also raises the testing coverage for the Menu component by ~1.5%.

taysea:
> @taysea , can you please elaborate on why the functionality was removed on #3464?
> 
> > This PR addresses missing WAI-ARIA keyboard interaction handling which was removed in #3464 to fix onClick handling in FireFox. Up and Down arrow keys can be used to navigate between menu-items after they are in focus. This also raises the testing coverage for the Menu component by ~1.5%.

Yes, we had temporarily removed the arrow key functionality that I had added in a previous PR because it had causes issues with onClick via mouse to propagate properly. The intention was for the functionality of arrow keys to be added back in, but it had not been implemented yet. Will review this PR now!

taysea (commented):
@esilverm  Looks like a good start here for reintroducing the arrow keys! However, the user needs to first `tab` to the items before they can use arrows. I think the ideal behavior would be that when the menu opens, the first menu item is active so as a user I could immediately start to use the arrow keys without needing to tab first. This is the common behavior I saw when exploring best practices.

esilverm:
> Looks like a good start here for reintroducing the arrow keys! However, the user needs to first `tab` to the items before they can use arrows. I think the ideal behavior would be that when the menu opens, the first menu item is active so as a user I could immediately start to use the arrow keys without needing to tab first. This is the common behavior I saw when exploring best practices.

Noted. I was working on implementing this earlier, but was having trouble setting the focus of the first menu item when the drop opened since the refs weren't accessible at that point in time. I can continue looking into a solution for this. Do you have any ideas as for immediately setting focus to the first menu item when the drop opens up?

I noticed that by using the `setActiveItemIndex` hook I was able to color the item as if it were in focus, but that only is half of the solution.

taysea:
> > Looks like a good start here for reintroducing the arrow keys! However, the user needs to first `tab` to the items before they can use arrows. I think the ideal behavior would be that when the menu opens, the first menu item is active so as a user I could immediately start to use the arrow keys without needing to tab first. This is the common behavior I saw when exploring best practices.
> 
> Noted. I was working on implementing this earlier, but was having trouble setting the focus of the first menu item when the drop opened since the refs weren't accessible at that point in time. I can continue looking into a solution for this. Do you have any ideas as for immediately setting focus to the first menu item when the drop opens up?
> 
> I noticed that by using the `setActiveItemIndex` hook I was able to color the item as if it were in focus, but that only is half of the solution.

Right, yeah that is understandable because that is the hurdle I had attempted to overcome in [this PR](https://github.com/grommet/grommet/pull/3334), but introduced the click issue. Maybe this could serve as a starting point but just keep in mind not wanting to reintroduce the issue.

taysea:
> @esilverm Looks like a good start here for reintroducing the arrow keys! However, the user needs to first `tab` to the items before they can use arrows. I think the ideal behavior would be that when the menu opens, the first menu item is active so as a user I could immediately start to use the arrow keys without needing to tab first. This is the common behavior I saw when exploring best practices.

Following up to my above note, the user still needs to `tab` first to the items before using arrow keys. This PR successfully allows users to navigate via arrow keys once they tab (and maintains the mouse click functionality) which is great. But, it doesn't solve for the general best practice behavior I have seen where the first menu item receives focus upon opening with keyboard. I had attempted to solve for that issue in [this PR](https://github.com/grommet/grommet/pull/3334) but it caused some backwards compatibility issues with mouse click.

I have verified locally that this PR allows both mouse and keyboard selection of Menu items, so I think it is good to go in that regard. I think it may be separate work though to fully achieve the best practice of having the first menu item become active when the menu opens.

esilverm:
> Following up to my above note, the user still needs to `tab` first to the items before using arrow keys. This PR successfully allows users to navigate via arrow keys once they tab (and maintains the mouse click functionality) which is great. But, it doesn't solve for the general best practice behavior I have seen where the first menu item receives focus upon opening with keyboard. I had attempted to solve for that issue in [this PR](https://github.com/grommet/grommet/pull/3334) but it caused some backwards compatibility issues with mouse click.
> 
> I have verified locally that this PR allows both mouse and keyboard selection of Menu items, so I think it is good to go in that regard. I think it may be separate work though to fully achieve the best practice of having the first menu item become active when the menu opens.

I haven't had much time to work on this recently since I have been focusing on my studies but when I last was working on this last I was struggling with the ability of having the first menu item receive focus upon opening without completely rolling back the changes. The main problem is getting access to the DOM element that should receive focus on open, since it can't be accessed through the buttonRefs list from within the onOpen method. Based on some debugging these values are only available when there is focus on the menu and not immediately when the menu is open. If there is a way I can get access to these elements as soon as the menu is opened then the fix shouldn't be too hard.

jcfilben:
Superseded by #5944

Issue #4539, opened 2020-09-21
Tracking the test coverage improvements effort


ShimiSun:
### Expected Behavior
Our goal is to improve the test coverage of the grommet core.
Components that need attention:
1-10 top components with the lowest coverage score.
1. Video - 60.53%
2. RadioButtonGroup - 61.22%
3. InfiniteScroll - 66.67%
4. RangeInput - 68.85%
5. Layer - 70.12%
6. DataChart - 74.76%
7. WorldMap - 78.57%
8. Chart - 79.72%
9. Main - 80%
10. Menu 80.52%

### Actual Behavior
Overall coverage:
85.15% Statements 5552/6520
75.58% Branches 4627/6122
82.59% Functions 1200/1453
86.18% Lines 5308/6159

### URL, screenshot, or Codepen exhibiting the issue
Screenshots of baselines
<img width="799" alt="Screen Shot 2020-09-28 at 11 46 08 AM" src="https://user-images.githubusercontent.com/54560994/94467439-3580b080-0180-11eb-8355-4c3dc62e9398.png">

<img width="1777" alt="Screen Shot 2020-09-28 at 11 46 25 AM" src="https://user-images.githubusercontent.com/54560994/94467465-3fa2af00-0180-11eb-83c5-1524528e8e7c.png">

### Steps to Reproduce on tracking coverage (running local coverage)
1. Run `yarn test --coverage`
2. Open grommet/coverage/icov-report/index.html
3. Click on the component for more detail. Lines of code that aren't being covered by tests are marked in red. 

### Additional Info
The top ten components are just a suggestion of what components currently have the lowest coverage. If you want to improve the coverage for another component not on that list feel free to do so.

Please share the coverage improvements (before and after) on your PRs.

Example PRs:
https://github.com/grommet/grommet/pull/4216/files
https://github.com/grommet/grommet/pull/4209/files

brendoncaulkins:
Would you be interested in some help integrating [coveralls](https://coveralls.io/) into your pipeline?  It's a code coverage tracking and enforcement tool that's free for open-source repos. It can deny PRs that drop global coverage below a certain threshold (which you set), or that drop coverage by a certain percentage (so if a PR drops you by, say, 2%, even it's still above the minimum it'll get denied).  This would give you code coverage enforcement on all future PRs, and you can tune the thresholds within coveralls, no code change necessary.

[...]

PR #4590, opened 2020-10-08
Add test coverage enforcement via Jest config

brendoncaulkins:
Signed-off-by: brendon.caulkins <brendon.caulkins@gmail.com>

#### What does this PR do?

This small change to `package.json` enables testing coverage enforcement via Jest, as a way to ward off future needs to swarm test coverage, like #4539 .  This approach is just one solution, and has the following benefits:

* Coverage minimums as code, so they are public, and can only be updated via PRs
* Requires no extra packages or integrations
* Works locally, including in pre-commit hooks
* Supports many configuration options beyond what's done in this PR: https://jestjs.io/docs/en/configuration#coveragethreshold-object

`package.json` and the Jest documentation: https://jestjs.io/docs/en/configuration#coveragethreshold-object

I set one of the thresholds below the current amount and the test run failed.  This is shown below in the included screenshot

You could alter the thresholds and see the failure due to being below, or add enough code somewhere to cause the coverage numbers to drop below the configured amount.

Enforcing code coverage minimums ensures ground isn't lost while development moves forward.  It's largely a safeguard and way to establish a standard, but it's no replacement for active pursuit of comprehensive coverage.

See #4539 

Example of a failure, locally (induced by artificially inflating the minimum).  This will also happen in the CI Pipeline, enforcing these minimums for PRs.
grommet-jest-config-threshold-example

Not for consumers.  The contributing guidelines may need to be updated, if you wish to call out testing specifically.  testing isn't mentioned in the contributing docs, but the current pre-commit hooks are a solid safeguard.

No.

This should be 100% backwards compatible and non-breaking as there is no functional code affected.  

**However**, if there are large PRs sitting around when this goes through, they could be surprised if they don't have adequate coverage.  Depending on repo configuration (whether or not PRs must be up-to-date with master before merging), a large PR that has already passed the pipeline checks and is approved could push master below these thresholds and cause master to break.  It may be worth holding off on this until any such large pieces are in, or lowering the values to an agreed upon minimum before merging.


jcfilben (commented):
I tested as you suggested and confirmed that this does work. The commit fails if the coverage is below the set coverage threshold.

I like the idea of encouraging test coverage numbers to remain high. My concern is that if someone from the community creates a PR that isn't passing due to test coverage it may discourage them from contributing. Also if a new feature is added or a component has major changes, sometimes the associated tests are added in a separate PR so that if there are snapshot changes they can be easily associated with the tests and not the actual code changes. If the tests are in a separate PR the original PR may not pass checks because it doesn't have high enough testing coverage.

I agree that this would be helpful in preventing the need for future swarm test coverage as you said. I am not against this idea, just pointing out a few things to consider.

ShimiSun:
@jcfilben good observations, thank you for sharing.
I think the way that @brendoncaulkins suggested the enforcement is easy enough to plug in and out in case we decide it's not contributing to the framework as we expected. 
I'm willing to give this PR a try to see how it works during Hacktoberfest and then we can measure whether it was helpful for us and something we'd like to continue with.

brendoncaulkins:
> I tested as you suggested and confirmed that this does work. The commit fails if the coverage is below the set coverage threshold.
> 
> I like the idea of encouraging test coverage numbers to remain high. My concern is that if someone from the community creates a PR that isn't passing due to test coverage it may discourage them from contributing. Also if a new feature is added or a component has major changes, sometimes the associated tests are added in a separate PR so that if there are snapshot changes they can be easily associated with the tests and not the actual code changes. If the tests are in a separate PR the original PR may not pass checks because it doesn't have high enough testing coverage.
> 
> I agree that this would be helpful in preventing the need for future swarm test coverage as you said. I am not against this idea, just pointing out a few things to consider.

Very valid points and things I've seen raised before.  This needs to fit into your normal workflow, after all.  Here are some work-arounds / solutions that may be worth considering.

**Coverage being discouraging to new contributors** - If your current coverage is 3-5%+ higher than the thresholds you configure, you would need a massive chunk of new code with very few tests to drop the coverage enough to trigger the blocking of the PR.  Example: Threshold is set at 80%, and your code coverage is around 85%, you need several dozens of lines of code to create enough debt to drop the entire project below 80%.  

I've set the values here very close to your current amounts in order to encourage growth away from these lower amounts, but you could absolutely lower them some in order to give yourself some breathing room, so to speak.  As you work on increasing the amounts, you could target increasing the thresholds whenever you feel comfortable doing so.

**Separating tests/snapshots from code changes** - On past projects, one of the ways around this issue was to push all the snapshot changes into a single commit at the end of the commit history of the PR.  So, after code changes and test updates and finished and committed, the snapshots go into one final commit. That lets reviewers ignore them while looking at code changes (either wholesale, or commit-by-commit), and then focus on the snapshots separately at the end.

You could also look at making snapshot breaks non-blocking, separating out snapshots into a separate test suite that's run in a different, non-breaking stage of the CI pipeline, or ditching snapshots all-together.

PR #4596, opened 2020-10-11
Fix DateInput date update before transition ends

### LINKED ISSUE #4535: DateInput change year after UI dropdown displays results in incorrect value

leossantiago:
#### What does this PR do?

Fix #4535 complimentary to #4541

Created a new hook that triggers with animations to Calendar, but acts only when the transition is finished. 
It checks if the current bounds correspond to the current date reference. If not, updates them.

Also did some code refactoring, for readability. Refactored what was previously added by me in #4541 . This refactor also prevents updating `displayBounds` to itself when no changes were made.

`Calendar/Calendar.js`

Storybook test, and trying to reproduce steps described in #4535 by @trumanpurnell 

Steps to reproduce:

    1.Head over to the trusty storybook: https://storybook.grommet.io/?path=/story/dateinput--form
    2.Select (not type) the date 11/04/2020
    3.Change (by typing) the year to 2012. (Oddly you have to type fast, if you type slow it works)
    4.Select what would be 11/04/2012
    5.The date returns to 11/04/2020

After the fix, it isn't reproducible anymore.

Storybook, steps described above. There shouldn't be any issue when selecting a date.

#4535 

No.

Backwards compatible.


src/js/components/Calendar/Calendar.js
```diff
@@ -194,32 +194,30 @@ const Calendar = forwardRef(
...
-          displayBounds[0].getTime() - nextDisplayBounds[0].getTime() >
-          millisecondsPerYear
-        ) {
-          diffBoundsAboveYearFlag = true;
-        }
-      } else if (nextDisplayBounds[1].getTime() > displayBounds[1].getTime()) {
-        if (
-          nextDisplayBounds[1].getTime() - displayBounds[1].getTime() >
-          millisecondsPerYear
+      if (nextDisplayBounds !== displayBounds) {
```
ericsoderberghp, line 197: Aren't these guaranteed to be different, since `buildDisplayBounds()` returns a new array.

src/js/components/Calendar/Calendar.js
```diff
@@ -194,32 +194,30 @@ const Calendar = forwardRef(
       // greater than a year. If that's the case, calendar should update without
       // animation.
       let diffBoundsAboveYearFlag = false;
```
ericsoderberghp, line 196: small naming question, could this be just `diffBoundsAboveYear`? Or, what does `Flag` convey?

src/js/components/Calendar/Calendar.js
```diff
@@ -279,6 +277,22 @@ const Calendar = forwardRef(
...
+      // Intended for when date changes before the Timeout of a transition ends.
+      // It triggers with changes to `animating` indicating a transition, but it
+      // acts only when `animating` is false, indicating the end of transition.
+      // It verifies the current displayBounds based on date `reference`. If
+      // it doesn't correspond to the wanted bonds, it updates them.
+
+      if (animating === false) {
+        const nextDisplayBounds = buildDisplayBounds(reference, firstDayOfWeek);
+
+        if (nextDisplayBounds !== displayBounds) {
```
ericsoderberghp, line 290: Won't this always be true, since `buildDisplayBounds()` returns a new array?

src/js/components/Calendar/Calendar.js
leossantiago: No problem. I'm just bad with names and tend to put 'flag' in every variable that controls something haha.
I'll change that in the next commit.

src/js/components/Calendar/Calendar.js
leossantiago: The thing is: this hook is triggered twice whenever Calendar is showed (and with inline Calendar, it triggers twice in the first render). The idea was to prevent the whole process of updating bounds being executed in this scenario (and any unknown situation this might happen).

Also, thanks for asking that. I did some testing to explain what I did, and discovered that 
`displayBounds === nextDisplayBounds` does not behave as
`nextDisplayBounds[0].getTime() === displayBounds[0].getTime() &&
          nextDisplayBounds[1].getTime() === displayBounds[1].getTime()`

So I'll update that in the refactor. Let me know if you think we should keep this or drop it. :smile:

src/js/components/Calendar/Calendar.js
leossantiago: As mentioned in your last comment, I'll also adapt this to compare each bound as time separately.
The goal is to compare both bounds. If they are not equal, meaning our bounds do not correspond to the current reference, a displayBounds update should be forced.
Comparing arrays is indeed always returning the same thing. I appreciate your help with that :tada:

leossantiago:
@ericsoderberghp pushed some changes. I'd appreciate your review. Also responded to your comments.
Let me know if there's anything I should/could fix!

jcfilben (approved):
Tested with storybook and it looks to be working. I also ran yarn test and yarn lint. Warning is no longer happening on yarn lint

ericsoderberghp (approved):

Issue #4591, opened 2020-10-09
Support width along with max-width / min-width


b1f6c1c4:
### Expected Behavior

For a `Box` I'd like to specify its desired width by CSS property `width` *and* its maximum width `max-width` at the same time. For example, I want my page container to be no larger than `800px` (so that things don't get too wide spread), but don't by smaller unless the user is using a mobile devices. So I want to put `width: 100%; max-width: 800px` as the style for the page container `Box`.

### Actual Behavior

I have to use `style={{ width: '100%'; max-width: '800px' }}`, instead of the convenient `width={{ default: '100%', max: '100px' }}` like in #3136 and #3313 .

### URL, screen shot, or Codepen exhibiting the issue

```jsx
const Test = () => {
  <Grommet full>
    <Box width={{ default: '100%', max: '300px' }} background="red">text</Box>
  </Grommet>
};
```

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.15.1
- Browser Name and version: Tried on Chromium on Linux desktop, Safari on iPad, Chrome on iPhone
- Operating System and version (desktop or mobile):

ericsoderberghp:
good idea. I might suggest: `width={{ width: '100%', max: '100px' }}`

ddaniss:
Can I give this a try?

jcfilben:
I think that PR #4597 solves this so this issue can be closed

PR #4597, opened 2020-10-11
Feat(Box): support width value on width object prop

netochaves:
#### What does this PR do?A
Adds support to set width in width object prop.
`src/js/components/Box/StyledBox.js`
Unit
Storybook
#4591
I don't think so
I don't think so
backwards compatible


ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution!

PR #4600, opened 2020-10-12
Add more type definitions to Select component

jeferson-sb:
#### What does this PR do?
Update Select existing typings for event handlers

`src/js/components/Select/index.d.ts`

Testing all stories through a local TS project

For some of the events and props I followed the description of the props in `src/js/components/Select/doc.js`

Notes:
`value` and `option` in `onChangeEvent` are set to any for the time being since we don't have access to a property that is not shared across all types.

#3165 

No

backwards compatible


jcfilben:
Tested with storybook. For the stories Multiple and ObjectMultiple I got the following error for onChange: `Argument of type 'string' is not assignable to parameter of type 'SetStateAction<number[]>'.`

ShimiSun:
@jeferson-sb can you please confirm the functionality works as expected without warnings to all of the Select stories?

jeferson-sb:
@ShimiSun Sure. So after my last update, almost stories are fine except `CustomSearch` onChange event which each option is an object with `name` and `id` properties.
I think we can change `option: string` to `option: string | object` 
What do you think?

jcfilben:
It looks like changing option to `option: string | object` still causes an error. I will have to look into this more to find what the right thing to put here for onChange is

jcfilben (commented):
The reason `option: string | object` is still causing an error is because when you try to access a property that is not shared across all types it causes an error. This is what is happening with line 176 `.indexOf(option.name);` in CustomSearch. I think the best option is to set the type for option to `option: any;`

jcfilben (approved):

src/js/components/Select/index.d.ts
```diff
@@ -8,14 +8,27 @@ import {
...
+    index: number,
+    options: object[],
+    state: { active: boolean; disabled: boolean; selected: boolean },
+  ) => React.ReactNode;
   closeOnChange?: boolean;
   disabled?: boolean | (number | string | object)[];
-  disabledKey?: string | ((...args: any[]) => any);
+  disabledKey?:
+    | string
+    | ((option: { label: string; value: number }) => boolean);
```
ShimiSun, line 31: Can `value` also be a 'string'?

src/js/components/Select/index.d.ts
```diff
@@ -51,8 +64,9 @@ export interface SelectProps {
   valueKey?:
     | string
     | { key: string; reduce?: boolean }
-    | ((...args: any[]) => any);
+    | ((option: { label: string; value: number }) => string);
```
ShimiSun, line 67: we might still need the `any` option @jcfilben can you please confirm?

src/js/components/Select/index.d.ts
```diff
@@ -8,14 +8,27 @@ import {
   PlaceHolderType,
 } from '../../utils';
 
+interface onChangeEvent extends React.ChangeEvent<HTMLSelectElement> {
+  value: any;
+  option: any;
```
ShimiSun, line 13: Can we please add comments on why this is using `any`?

src/js/components/Select/index.d.ts
jcfilben: Yes, I think the way it is setup right now the option is too restrictive. According to documentation: "If a function is provided, it is called with the option and the return value indicates the value. If reduce is true, the value coming via the key will be used for the onChange value and the value property is expected to be reduced to align."

src/js/components/Select/index.d.ts
jcfilben: Maybe this should be
``` typescript
  disabledKey?:
    | string
    | ((...args: any[]) => boolean);
```

src/js/components/Select/index.d.ts
jeferson-sb: Added in the PR description

jcfilben (approved):

src/js/components/Select/index.d.ts
```diff
@@ -8,14 +8,27 @@ import {
...
   a11yTitle?: A11yTitleType;
   alignSelf?: AlignSelfType;
   gridArea?: GridAreaType;
-  children?: (...args: any[]) => any;
+  children?: (
+    option: object,
+    index: number,
+    options: object[],
+    state: { active: boolean; disabled: boolean; selected: boolean },
+  ) => React.ReactNode;
```
ShimiSun, line 26: I'm not sure this is completely accurate. Why not just remove the children from the Interface and fall back to React children type? 
From what you have specified, all the fields are required and not optional, which may cause backward compatibility issues for users who don't necessarily use them. 
i.e.
```suggestion
  children?: (
    option?: object,
    index?: number,
    options?: object[],
    state?: { active: boolean; disabled: boolean; selected: boolean },
  ) => React.ReactNode;
```

I think we should remove the `children` prop from the interface unless you think otherwise.

src/js/components/Select/index.d.ts
```diff
@@ -51,8 +64,9 @@ export interface SelectProps {
   valueKey?:
     | string
     | { key: string; reduce?: boolean }
-    | ((...args: any[]) => any);
+    | ((...args: any[]) => string);
```
ShimiSun, line 67: Could this also return a number?

src/js/components/Select/index.d.ts
```diff
@@ -26,17 +39,17 @@ export interface SelectProps {
...
   multiple?: boolean;
   name?: string;
-  onChange?: (...args: any[]) => void;
-  onClose?: (...args: any[]) => any;
-  onMore?: (...args: any[]) => any;
-  onOpen?: (...args: any[]) => any;
+  onChange?: (event: onChangeEvent) => void;
+  onClose?: () => void;
+  onMore?: () => void;
+  onOpen?: () => void;
```
ShimiSun, line 52: This change needs to be tested using those props, I think we need to keep the args.

ShimiSun (commented):
I feel like there needs to be done more testing on this PR so we can improve the confidence level in these changes.

src/js/components/Select/index.d.ts
jeferson-sb: hmm, from I've seen in the component source code, it depends on the value prop,
So I think it's better to leave as any

src/js/components/Select/index.d.ts
```diff
@@ -26,17 +33,17 @@ export interface SelectProps {
   dropTarget?: object;
   dropProps?: DropProps;
   focusIndicator?: boolean;
-  icon?: boolean | ((...args: any[]) => any) | React.ReactNode;
+  icon?: boolean | (() => React.ReactElement) | React.ReactNode;
```
ShimiSun, line 36: Does this actually works without the `args`?

src/js/components/Select/index.d.ts
```diff
@@ -53,6 +60,7 @@ export interface SelectProps {
     | { key: string; reduce?: boolean }
     | ((...args: any[]) => any);
   emptySearchMessage?: string;
+  searching?: boolean;
```
ShimiSun, line 63: I  do not see a prop called `searching` on Select.

jeferson-sb: it's set as boolean in the [CustomSearch](https://github.com/grommet/grommet/blob/master/src/js/components/Select/stories/CustomSearch.js) story

src/js/components/Select/index.d.ts
jeferson-sb: yes

jimenezff: Please see this line: https://github.com/grommet/grommet/blob/f4003357390dd83e250c535f90c20e19368043da/src/js/components/Select/Select.js#L287 The type definition should include those 2 props (color and size). I've actually been working on some changes to the Select component and this is something I noticed.

jimenezff:
We may need to come to an agreement on some of the type definition changes I made on #4694

jeferson-sb:
@jimenezff yeah, I think I'm gonna close this, since it's already defined and tested in your PR

PR #4603, opened 2020-10-13
Add drop customization to MaskedInput

### LINKED ISSUE #2611: MaskedInput does not take a dropTarget (like TextInput and Select do)
The TextInput and Select components accept a custom ref to be provided so that the drop (for suggestions or options) is rendered based on a different component other than the input box that's associated with the drop. This allows further customization of the inputs such as adding an active bottom border to a box that is wrapping the input and when the drop is open.

The current implementation of the MaskedInput doesn't accept a dropTarget prop creating the following issue:
https://codesandbox.io/s/kknpk99yx5

### Expected Behavior

MaskedInput such accept a dropTarget prop.

### Steps to Reproduce

1. use the codebox link provided above

### Your Environment
- Grommet version: 2.2.1
- Operating System and version (desktop or mobile): desktop

emilpalsson:
#### What does this PR do?
Affects `MaskedInput` only, implementation matches the one in `TextInput`.
- Add `dropHeight` prop, sets `max-height` of drop container only if passed.
- Add `dropProps` prop, props passed down to `Drop`.
- Add new props to `doc.js`.
- Add new props to typescript typing (`index.d.ts`).
- Add storybook story showcasing the use of drop prop `target`.
- `README.md` and `__snapshots__` automatically updated.

`MaskedInput.js`

Storybook and manual (+ jest for regression), se below for manual steps tested.

- Any `dropProps` are passed down to the `Drop`, e.g. setting `{ stretch: false }` or passing a `target` (see new storybook story).
- Passing `dropHeight` sets `max-height` of drop, not passing sets `inherit`.

Closes #2611.

Yes, PR includes updates.

Yes.

Backwards compatible. The only change if none of the new props are passed is that the following CSS is added to the drop box of `MaskedInput` (just like it is in `TextInput`):

```css
max-height: inherit;
@media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
  width: 100%;
}
```


ericsoderberghp:
Agreed on adding `dropProps` to MaskedInput to align with TextInput. I don't think we should include the other `drop*` properties though. The reason TextInput has those is that we originally had one of them, started adding more, and then realized we didn't want to keep adding `drop*` for them. So, we switched to just having `dropProps` and taking all props for Drop there. We left the original ones to not break compatibility.

emilpalsson:
PR and PR description updated, `dropAlign` and `dropTarget` removed.

Issue #4095, opened 2020-05-18
Make years and months for the calendar editable


cliffpyles:
On the calendar component a user has to click the arrows to navigate through each month before they can switch years. So for example if I am on December 2020, and I want to be on December 2000, I have to click the back arrow 240 times.

### Expected Behavior

A user would have an efficient way to navigate/edit the year and month of the calendar.

### Actual Behavior

A user has to click the arrow to navigate through each month and each year.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/github/grommet/grommet-sandbox?initialpath=calendar&module=%2Fsrc%2FCalendar.js

### Your Environment

- Grommet version: All
- Browser Name and version: All
- Operating System and version (desktop or mobile): All

ericsoderberghp:
Before doing any work on this, we should have some design work exploring how to go about this, especially accessibility aspects.

PR #4604, opened 2020-10-13
Calendar: allow user to select arbitrary month and year

theerapatcha:
#### What does this PR do?
Enhance `Calendar` to allow users to select month and year. When a user accesses the `Calendar` component, he/she can click on the title, which contains the month and year info, to change the month and year with ease.
`Calendar` component
Test by storybook
Test by storybook
Customized month component not yet implemented :(
Issue: #4095 
-
Nope
Nope
Backward compatible


theerapatcha:
Hi, @ShimiSun 

This PR fixes #4095  If you think that there is something that needs to be changed, please feel free to say so.

Cheers

theerapatcha:
> @theerapatchasuper super interesting PR, thank you for starting the conversation.
> Since this PR is offering design enhancements, we need to have some design discussions around this functionality in order to make sure it is WCAG compatible and aligned with the grommet overall behavior.
> If I may ask, where did you get the inspiration for this functionality/design?
@ShimiSun 

My inspiration is based from what generally DatePicker component functionalities should has, `A user should be able to jump between month and year with ease`, with a combination of the UI/UX from another renowned DatePicker component from [Angular Material DatePicker](https://material.angular.io/components/datepicker/overview) 

However, the UX is not exactly the same as followings:
In Angular Material DatePicker: If user want to select a month -> he/she first has to select a year then follows by a month. 
In this PR: User can select a month first, then if he/she wants to select another year then he can do so.

This design decision is based on what I feel when using Angular DatePicker which required multiple interactions before I can choose a month and its level of granularity that user can edit of `Date` > `Year` > `Month` (which I prefer to be `Date` > `Month` > `Year`)

ericsoderberghp:
Here are some design approaches for consideration:

https://designer.grommet.io/?id=Calendar-updates-eric-soderberg-hpe-com

I didn't fill out everything but intended to give a flavor.

I think we could increase the font size of the selectors to be closer to what Calendar already has.

I think the months could be fully populated, unless the `bounds` is more restrictive. For the years, this approach provides some flexibility. For instance, if the Calendar has `bounds` and they are within a few years, we would just show those years. If there are no `bounds`, we could show a few years ahead and behind but then allow the user to type an arbitrary year, which provides a faster way to jump.

jcfilben:
Closing due to inactivity

PR #4605, opened 2020-10-13
Support dataTable.header.extend in theme and expose prop

halocline:

#### What does this PR do?

Enables a caller to specify styling via `dataTable.header.extend` and passes props to a header cell's button.

StyledDataTable.js, base.js, and /DataTable/Header.js

Jest + modified Storybook story DataTable/custom

Storybook story DataTable/custom

[Design File](https://www.figma.com/file/XEoIWIlLfcARRIFhkpFCch/HPE-Table?node-id=2390%3A13951)

Addressing a DataTable design which calls for a Header Cell's `icons.sortable` to only be shown upon hover, while an icons indicated a sort has been applied to persist. This behavior seemed opinionated and particular to this specific design, so this PR opts for a more generalized solution allowing for that customization to be done via `extend`.

https://github.com/grommet/hpe-design-system/issues/1180

Yes. Included.

Pretty minor.

Backwards compatible.


src/js/components/DataTable/StyledDataTable.js
```diff
@@ -105,6 +105,10 @@ const StyledDataTableHeader = styled(TableHeader)`
     width: 100%;
     table-layout: fixed;
   `}
+  ${props =>
+    props.theme.dataTable &&
+    props.theme.dataTable.header &&
+    props.theme.dataTable.header.extend};
```
ericsoderberghp, line 111: Shouldn't this be applied to the `<th />` cells instead of the `<thead />` header element?

taysea (commented):
Aside from Eric's comment, I think the rest looks appropriate

src/js/components/DataTable/StyledDataTable.js
halocline: Done.

ericsoderberghp (approved):

ericsoderberghp:
Can you look into why the header cells are being styled differently in the DataTable Custom story? Chromatic reported this and it seems like it might be a breaking change.

halocline:
> Can you look into why the header cells are being styled differently in the DataTable Custom story? Chromatic reported this and it seems like it might be a breaking change.

I believe it is because I modified the table header cell text to be bold, which caused the overall table width to grow, slightly displacing the rest of the table content. I will remove that styling change, and re-run Chromatic to confirm.

halocline:
> > Can you look into why the header cells are being styled differently in the DataTable Custom story? Chromatic reported this and it seems like it might be a breaking change.
> 
> I believe it is because I modified the table header cell text to be bold, which caused the overall table width to grow, slightly displacing the rest of the table content. I will remove that styling change, and re-run Chromatic to confirm.

@ericsoderberghp I removed some of the additional styling I had added to the Custom story (just left in the hover styling) and the Chromatic snapshots no longer show a diff.

ericsoderberghp (approved):

PR #4611, opened 2020-10-14
Updated Markdown docs to include available options for components

### LINKED ISSUE #4000: faulty documentation for <Markdown />
Documentation for `Markdown` gives this as the example code for customizing a `<p></p>`  element in the output:
```json
{
  "p": {
    "component": "Paragraph",
    "props": {"size": "medium"}
  }
}
```
but actually that results in the following console warning:
```
Warning: <Paragraph /> is using incorrect casing. Use PascalCase for React components, or lowercase for HTML elements.
    in Paragraph (created by Markdown)
    ...(component stack trace follows)
```
and the following markup:
```html
<paragraph><strong>Who I am</strong></paragraph>
```
When you replace `"Paragraph"` with `Paragraph` (no quotes) it works as intended. Not sure if this is an issue with the documentation or with the code?

jcfilben:
#### What does this PR do?
Included the available options for components in the markdown documentation
```
a: { component: Anchor },
img: { component: Image },
p: { component: Paragraph },
table: { component: Table },
td: { component: TableCell },
tbody: { component: TableBody },
tfoot: { component: TableFooter },
th: { component: TableCell },
thead: { component: TableHeader },
tr: { component: TableRow },
```
Previously the documentation didn't show the options and it is discussed in issue #4000 that this can be confusing to users
Closes #4000 

Done

No

Backwards compatible


src/js/components/Markdown/doc.js
```diff
@@ -20,7 +20,18 @@ export const doc = Markdown => {
     components: PropTypes.shape({}).description(
       `Custom components and props to override html elements such as 'img'
       or 'pre'. By default 'a', 'p', 'img', and table elements are overridden
-      with grommet components`,
+      with grommet components.
+      Available options:
```
ShimiSun, line 24: I don't think this is fixing the original warning that is shown on the issue.

IanKBovard:
I kind of see where you're going here. Trying to be explicit to avoid user error. I think the main issue here is that the example is wrapped in quotes when it should not be because of how the docs are parsed.

```
{
  "p": {
    "component": "Paragraph",
    "props": {"size": "medium"}
  }
}
```

should be
```
{
  "p": {
    "component": Paragraph, // no quotes here
    "props": {"size": "medium"}
  }
}
```

It's hard to say what the right approach is here. For me the docs saying it takes a component is enough for me to know that passing in a string won't work. I think being explicit is a good approach, but as long as the example is there people will just copy and paste it, and get the error.

jcfilben:
I'm not sure if changing components `PropTypes.shape({})` to `PropTypes.object` will display the example correctly or not. I don't know enough about how the docs examples are generated. Other than this I am not sure what else to try to get the example to display without quotes.

IanKBovard:
@jcfilben ill take a look at the doc generator again when I have the time. I do remember trying to mess with it sometime ago, and having a lot of trouble. Another solution would to give a warning that is more descriptive of the issue. 

eg.
`this prop expects react components - received string 'Paragraph'`

ShimiSun (approved):

Issue #3654, opened 2020-01-10
Grid - adding props from Box (`border` and `pad`)


skoshy:
### Expected Behavior

`Grid`s would have similar props to `Box`

### Actual Behavior

`Box` has many more stylistic props than `Grid`

### Reasoning

As `Grid`s and `Box`es are both used for basic layouting often, it could be beneficial to add similar stylistic props to `Grid`.

Currently focusing on `border` and `pad` before considering other props

See this Slack thread for more info - https://grommet.slack.com/archives/C04LMHN59/p1578672831048600

ShimiSun:
Linking to #3750

ShimiSun:
@skoshy can you please clarify what `border` would do? would you want it only on the outer Grid or in between rows and columns as well?

skoshy:
I was thinking the outer border only, to keep it consistent with how border would work if applied with css to a grid element and also with how border works in `Box`. Inner borders could be a separate prop if deemed necessary?

srallen:
I just ran into this trying to apply an outer border to a `Grid` and discovering that it is not supported at the moment. I would also expect that `border` when applied to `Grid` functioned like it does with `Box`

ShimiSun:
@srallen I'm supportive, feel free to work on a contribution.

srallen:
@ShimiSun I'm hoping to participate in Hacktoberfest more broadly this year, so I'll have a look! 😄

jcfilben:
I think this issue can be closed, Grid now has both `border` and `pad` props available

ShimiSun:
Thank you @jcfilben !

PR #4612, opened 2020-10-14
Add support for border for Grid similar to Box

srallen:
Signed-off-by: Sarah Allen <sarah@zooniverse.org>

#### What does this PR do?

Add support for border to Grid with similar function as Box. Toward #3654

The updates to Grid.js and StyledGrid.js for border prop support.

Added to the Grid spec file snapshot tests for border. Updating the snapshots updated several unrelated snapshots; I'm not sure how to prevent that from happening.

A storybook story has been added for Grid with border.

This is for #3654 to expand the styling configuration by prop for Grid. This does not close it as there may be others that are wanted, like pad support.

#3654 

Yes, to note the support for border for Grid.

Yes.

Backwards compatible.


srallen:
I'll have a look at that CI failure hopefully today or tomorrow. Tests passed when run locally after the snapshot updates.

ericsoderberghp:
I'm good with the code changes. It looks like there are some snapshot issues though.

srallen:
@ericsoderberghp this is ready for review again. I reverted the unrelated snapshots that had gotten updated and causing the CI failure. I'm presuming there's a way to run the jest CLI snapshot update for a single component, but I'm not very familiar with this as jest isn't used with the app I work on. Hope to at least get another PR in during hacktoberfest, so I'll have a look then. Thanks!

ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution!

Issue #4599, opened 2020-10-12
Tooltip component


ericsoderberghp:
Create a Tooltip component for use on it's own but that can also be embedded into Button, Menu, etc.

Issue #4588, opened 2020-10-08
Add caret/arrow on tooltip


padmacnu:
Dynamically change position of tooltip arrow based on position on content.
![image](https://user-images.githubusercontent.com/6428437/95487230-e76e6880-0948-11eb-9712-f6a4ca53d563.png)

Current:
![image](https://user-images.githubusercontent.com/6428437/95487265-f5bc8480-0948-11eb-82bb-61cd00f2f981.png)

ShimiSun:
Can you please add the full description of the issue?
I’m not sure I understand the expected behavior. We currently don’t have a tooltip component, so how do you think the behavior should work?

IanKBovard:
@ShimiSun is correct, there is no tooltip component in grommet. This is an example of using drop to replicate a tooltip. As it stands, it's up to the user to design the tooltip box. For me, I would prefer to styled the box myself using CSS, is this something we want to eventually implement?

padmacnu:
@ShimiSun I am talking about Drop component, we use it in CURO as tooltip component. Could you please confirm if you're planning to implement it in Grommet? Else we can try to style it our self.

ShimiSun:
I'm currently working on #4599, and I'm not sure how smoothly it will operate with a caret as we are not planning to provide the caret behavior out of the box in grommet. I'll keep you posted on how it goes early next week.

ShimiSun:
@padmacnu, it is still under review, but it is very close 😃  
https://deploy-preview-4613--sad-tereshkova-173d07.netlify.app/?path=/story/tip--caret 
Let me know what you think. Mike supposed to schedule us a call.

ShimiSun:
The solution for the Caret is now demonstrated in our storybook. Best of luck! https://storybook.grommet.io/?path=/story/visualizations-tip--caret

Issue #4128, opened 2020-05-30
No tool tooltip component in grommet v1


naveenrj98:
### Expected Behavior
When we click on button or hover on component, we need to show a component which has tip on it to referring component we use in the same thing.

### Actual Behavior
we do not have any component as like to show with tip on any side to reffer the component we use 

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: v1
- Browser Name and version: chrome
- Operating System and version (desktop or mobile): desktop

ShimiSun:
Hi @naveenrj98 thanks for reaching out.

We currently do not have the resources to address v1 enhancement requests or overall maintenance.

The solution in v2 for tooltip is using Drop component, https://v2.grommet.io/drop

PR #4613, opened 2020-10-14
NEW Tip (Tooltip) component 

ShimiSun:
#### What does this PR do?
Adapting the new API convention we have discussed in our last Tooltip discussion.
```
 <Tip content={<Box>My Tip</Box>}>
      <Button background="purple" label="Button" onClick={() => {}} />
 </Tip>
```
The docs, stories, and TS files are not adapted to the new API. I'll complete it once this approach and the final API is approved on this draft PR.

Tip.js.
And then only look at the Sidebar story example. The other examples need to be polished after this PR will be approved.

Storybook.
Review + Storybook

#4599 #3489 #4588 #4128
Yes, will add it after the overall approach is approved.
Yes!
Backwards compatible as it offers only new functionality.


src/js/components/Tip/Tip.js
```diff
@@ -0,0 +1,50 @@
...
+export const Tip = forwardRef(
+  ({ children, content, dropProps, ...rest }, tipRef) => {
+    const theme = useContext(ThemeContext) || defaultProps.theme;
+    const [over, setOver] = useState(false);
+    const ref = useRef();
+
+    return (
+      // Not sure about the extra Box, I'm still thinking of ways to simplify
+      <Box ref={tipRef} {...rest}>
+        <Box
```
ericsoderberghp, line 18: I think we should try to have less Boxes. I think we can do it with zero Boxes. Some ideas:
- use `cloneElement()` to add the event handlers and ref, rather than wrapping with a Box.
- return `[{clonedChild}, <Drop>]` when we render the Drop.

src/js/components/Tip/Tip.js
```diff
@@ -0,0 +1,81 @@
...
+        }
+      },
+    }),
+  );
+
+  return [
+    clonedChild,
+    over && (
+      <Drop
+        align={{ left: 'right' }} // most common use case is a sidebar?!
```
ShimiSun, line 58: I'm wondering if we should leave it to default to the drop behavior.

src/js/themes/base.js
```diff
@@ -1169,6 +1169,11 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       // extend: undefined,
       // disabled: { opacity: undefined },
     },
+    tip: {
+      // container: {}, // any Box props
+      content: {}, // any Box props
+      drop: {}, // any props for the drop
+    },
```
ShimiSun, line 1176: I'm still thinking whether there is a need for `extend`, I don't see a need at this point hence I'll wait to see if the need arises.

src/js/components/Tip/stories/Guide.js
```diff
@@ -0,0 +1,40 @@
+// import React, { createRef, useState, useRef } from 'react';
```
ShimiSun, line 1: I'll remove this file and create tests.

src/js/components/Tip/Tip.js
ericsoderberghp: I think the most common use case will be a Header containing Buttons, where I think we'd want the default to be `{ top: 'bottom' }`

src/js/components/Tip/stories/ChartTip.js
```diff
@@ -0,0 +1,72 @@
...
+      <List data={listData} primaryKey="action" secondaryKey="entry" />
+    </Box>
+  </Box>
+);
+
+const Example = () => (
+  <Grommet theme={theme}>
+    <Box align="center" justify="start" pad="large" gap="medium">
+      <Heading size="small"> Hover the Chart </Heading>
+      <Tip content={<TipContent />}>
```
ericsoderberghp, line 63: I don't think this isn't quite what we'd want. I'd expect a Tip per bar showing the details for that bar. I'm contemplating enhancing Chart to support `values={[{ value: ..., tipProps: { ... } }, ...]}`

src/js/components/Tip/stories/Table.js
```diff
@@ -0,0 +1,102 @@
...
+            step={10}
+            onClickRow={event => setSelected(event.datum)}
+          />
+        </Box>
+      </Box>
+    </Grommet>
+  );
+};
+
+storiesOf('Tip', module).add('Table Header', () => <Example />, {
```
ericsoderberghp, line 100: I'd probably call this one "Header actions", since it's not inside the `<thead />`.

src/js/components/Tip/stories/Table.js
```diff
@@ -0,0 +1,102 @@
...
+
+  const removeRow = () => {
+    const filteredData = data.filter(item => item.name !== selected.name);
+    setData(filteredData);
+    setSelected(undefined);
+  };
+
+  return (
+    <Grommet full theme={grommet}>
+      <Box align="center" justify="center" background="dark-1" fill gap="large">
```
ericsoderberghp, line 46: Any reason so many have dark backgrounds? It seems inconsistent with most of the other stories.

src/js/components/Tip/stories/typescript/Simple.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { Tip } from '../../Tip';
+
+const Example = () => {
+  const ref = useRef();
+  return (
+    <Grommet full theme={grommet}>
+      <Box align="center" justify="center" background="dark-1" fill>
+        <Tip
+          content={
+            <Box
```
ericsoderberghp, line 14: Can we make the "Simple" story even simpler?
```
<Tip content="something helpful"><Button label="action" /></Tip>
```

src/js/components/Sidebar/stories/Tooltips.js
```diff
@@ -20,54 +21,44 @@ import {
...
 
 const src = '//s.gravatar.com/avatar/b7fb138d53ba0f573212ccce38a7c43b?s=80';
 
-const NotificationIcon = () => (
-  <Stack anchor="top-right">
-    <Notification />
-    <Box background="accent-1" pad="xsmall" round responsive={false} />
-  </Stack>
-);
+const customTip = deepMerge(grommet, {
```
ericsoderberghp, line 28: I'm not sure this is a good type of guidance as it seems specific to the Sidebar scenario. Are you thinking all uses of Tip would be shown in this style?

src/js/components/Tip/Tip.js
```diff
@@ -0,0 +1,81 @@
...
+import { Box } from '../Box';
+import { Drop } from '../Drop';
+
+const Tip = forwardRef(({ children, content, dropProps }, tipRef) => {
+  const theme = useContext(ThemeContext) || defaultProps.theme;
+  const [over, setOver] = useState(false);
+
+  // console.log(tipRef);
+  const ref = useRef();
+  const componentRef = tipRef || ref;
```
ericsoderberghp, line 20: We should use our `useForwardedRef()` instead of `useRef()` and `||`.

src/js/components/Tip/Tip.js
```diff
@@ -0,0 +1,81 @@
...
+
+  const wrapChildIfString = child => {
+    // Handle the use case of a primitive string child
+    // so we'll be able to assign ref and events on the child.
+    // console.log(child);
+    return typeof child === 'string' ? <span>{child}</span> : child;
+  };
+
+  const clonedChild = React.Children.map(children, child =>
```
ericsoderberghp, line 29: Can we enforce that we only have a single child, like Keyboard.js does? See: `Children.only(children)`

src/js/components/Tip/doc.js
```diff
@@ -0,0 +1,40 @@
...
+    .description(
+      `Tooltip or a hint when hovering over a UI item. 
+    The tooltip will render when hovering on top of the 
+    Tip child node or string.`,
+    )
+    .usage(
+      `import { Tip } from 'grommet';
+<Tip />`,
+    )
+    .intrinsicElement('span'); // TODO after resolving the Box wrapper issue
```
ericsoderberghp, line 17: I think we can omit `intrinsicElement()`. Keyboard doesn't use it.

src/js/components/Tip/Tip.js
```diff
@@ -0,0 +1,81 @@
...
+      <Drop
+        align={{ left: 'right' }} // most common use case is a sidebar?!
+        target={componentRef.current}
+        trapFocus={false}
+        key="tip-drop"
+        plain
+        {...theme.tip.drop}
+        {...dropProps}
+      >
+        <Box {...theme.tip.content}>{content}</Box>
```
ericsoderberghp, line 66: Hmm...
This means the caller can't customize this for a particular Tip. I'm wondering if we'll end up wanting `plain` at some point. No need to add it yet.

src/js/components/Tip/index.d.ts
```diff
@@ -0,0 +1,14 @@
...
+// import { Omit } from "../../utils";
+import { DropProps } from '../Drop';
+
+export interface TipProps {
+  content?: React.ReactNode;
+  dropProps?: DropProps;
+}
+
+// TODO polish after resolving Box wrapper
+// declare const Tip: React.ComponentClass<TipProps & Omit<JSX.IntrinsicElements['TODO'], 'title'>>;
```
ericsoderberghp, line 11: See Keyboard/index.d.ts as a reference

src/js/components/Tip/stories/ChartTip.js
ShimiSun: I agree, the same approach will go towards:
1. WorldMap `places` - tooltip per place
2. List `item` - tooltip per item.
This example was more to demonstrate Tooltip on visual components that have extended data that uses a List, I've seen it used in a few places. I can replace the DataChart to be one stacked bar with three metrics, showing the same tooltip as it showed now on the DataChart.

src/js/components/Tip/stories/Table.js
ShimiSun: I just liked the contrast of the tooltip background in the dark mode. 
I'll flip a few examples to use light mode.

src/js/components/Tip/Tip.js
ShimiSun: Good comment. I was just having the exact same thought on my walk. The problem with more than one child is that the tip will only be shown on the latter.

src/js/components/Tip/Tip.js
ShimiSun: Please - no plain.
We can easily allow `contentProps` similarly to `dropProps` and that could help with specific customizations.

ShimiSun:
Merged with master to get the fix of List `extend`.

src/js/components/Tip/stories/Table.js
ShimiSun: I've removed the dark background from a few stories.

src/js/components/Tip/stories/typescript/Simple.tsx
ShimiSun: Done.

src/js/components/Sidebar/stories/Tooltips.js
ShimiSun: I've removed the `deepMerge`  and the `grommet` theme usage for better guidance.

src/js/components/Sidebar/stories/Tooltips.js
```diff
@@ -20,54 +20,50 @@ import {
...
-
 const src = '//s.gravatar.com/avatar/b7fb138d53ba0f573212ccce38a7c43b?s=80';
 
-const NotificationIcon = () => (
-  <Stack anchor="top-right">
-    <Notification />
-    <Box background="accent-1" pad="xsmall" round responsive={false} />
-  </Stack>
-);
+const customTip = {
```
ericsoderberghp, line 25: Can we call this `customTheme`?

src/js/components/Tip/Tip.js
```diff
@@ -0,0 +1,88 @@
...
+import { useForwardedRef } from '../../utils/refs';
+
+const Tip = forwardRef(({ children, content, dropProps }, tipRef) => {
+  const theme = useContext(ThemeContext);
+  const [over, setOver] = useState(false);
+
+  const componentRef = useForwardedRef(tipRef);
+
+  // In cases the child is a primitive
+  const wrapInvalidElement = () => {
```
ericsoderberghp, line 21: Maybe call this `wrapPrimitiveElement`?

src/js/components/Tip/Tip.js
```diff
@@ -0,0 +1,88 @@
...
+    // so we'll be able to assign ref and events on the child.
+    return !React.isValidElement(children) ? <span>{children}</span> : children;
+  };
+
+  /* Three major use case 
+    1. Tip has a single child + it is a React Element => Great!
+    2. Tip has a single child +  not React Element => span to wrap the child.
+    3. Tip has more than one child => Abort, display Children.only error 
+  */
+  const child =
```
ericsoderberghp, line 32: I think this can be simplified to `const child = wrapInvalidElement(Children.only(children));`

src/js/components/Tip/stories/ChartTip.js
```diff
@@ -0,0 +1,72 @@
...
+      <List data={listData} primaryKey="action" secondaryKey="entry" />
+    </Box>
+  </Box>
+);
+
+const Example = () => (
+  <Grommet theme={theme}>
+    <Box align="center" justify="start" pad="large" gap="medium">
+      <Heading size="small"> Hover the Chart </Heading>
+      <Tip content={<TipContent />} dropProps={{ align: { left: 'right' } }}>
```
ericsoderberghp, line 63: I don't think this is a realistic example. I think the expectation is that the Tip would change content as the user hovers over different bars in the Chart. I think we should remove this story and wait for Chart to have support for `tip` built in and show it there.

src/js/components/Tip/stories/typescript/Simple.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import { grommet, Box, Button, Grommet, Text } from 'grommet';
+import { Tip } from '../../Tip';
+
+const Example = () => {
+  const ref = useRef();
+  return (
+    <Grommet full theme={grommet}>
+      <Box align="center" justify="center" fill>
+        <Tip content="tooltip">
```
ericsoderberghp, line 12: I think the grommet theme should have a Tip content styled to look a bit more grommet-y. This story, which I think should work in the 80% case, seems more "plain".

src/js/components/Tip/Tip.js
ShimiSun: I'm not sure if there is 1:1 correlation between primitives and invalidReactElements. Semantically, the function is wrapping invalid React elements.

src/js/components/Tip/Tip.js
ShimiSun: It would be cleaner, wouldn't it 😉 
`Children.only(children)` is failing when the children are not valid to React element, so it is required to wrap the child to be a React Element (i.e. `warpInvalidElement`) before calling `Children.only(children)`.

src/js/components/Tip/Tip.js
ericsoderberghp: ok

src/js/components/Tip/stories/ChartTip.js
ShimiSun: Thanks for the feedback, I replaced the example to use a Meter instead of a Chart and added a legend to the List items on the tooltip.

src/js/components/Tip/stories/typescript/Simple.tsx
ShimiSun: Check out base.js to see the default values, and let me know if you'd like to tweak any styles.
I've also added a separate `plain` option.

src/js/components/Tip/stories/ChartTip.js
ShimiSun: I've modified the existing Children's story of List to display tips for each item.

src/js/components/Tip/Tip.js
ShimiSun: Update: after adding default content styles, I saw the need to have a `plain` option so those styles could be removed when the users build their own styled contents.

You noticed the need for it sooner than me 😉

ericsoderberghp:
This story needs some responsive fixes.

Screen Shot 2020-10-28 at 11 06 44 AM

ericsoderberghp:
This looks a bit odd with the caret at this resolution.

Screen Shot 2020-10-28 at 11 08 02 AM

ericsoderberghp:
I think this needs a bit more horizontal padding for the 80% case.

Screen Shot 2020-10-28 at 11 08 58 AM

src/js/components/List/stories/children.js
```diff
@@ -20,7 +25,9 @@ export const ChildrenExample = () => (
             align="center"
           >
             <Gremlin size="large" />
-            <Text weight="bold">{datum}</Text>
+            <Tip content={datum.state} dropProps={{ align: { left: 'right' } }}>
```
ericsoderberghp, line 28: Can we put the Tip containing the Box, at line 20?

I'm also thinking that this story doesn't feel very "real". What about shrinking the List to feel more like it's inside a small panel, such that there isn't room for more than the icon + name. As it stand, the List is so large that there's plenty of room for the city to be shown. The Tip content feels a bit underwhelming currently.

ericsoderberghp:
This one needs some responsive help as well.

Screen Shot 2020-10-28 at 11 14 59 AM

ShimiSun:
Screen shots after the fixes:

image

Still need some work:
image

image

image

src/js/components/List/stories/children.js
ShimiSun: Fixed in #4705.

Changed the panel to be smaller, so Tooltip will make more sense.

src/js/components/Tip/stories/typescript/Simple.tsx
ShimiSun: One thing to consider is that "background-contrast" has opacity, so the default behavior will be to have opacity on the tooltip.

src/js/components/Drop/stories/TrapFocus.js
```diff
@@ -23,13 +23,14 @@ const TooltipButton = ({ icon, name }) => {
       </Button>
       {ref.current && over && (
```
ericsoderberghp, line 24: Any reason we aren't using `<Tip />`?

src/js/components/List/stories/children.js
```diff
@@ -3,25 +3,31 @@ import { storiesOf } from '@storybook/react';
...
-            gap="large"
-            size="xsmall"
-            align="center"
-          >
-            <Gremlin size="large" />
-            <Text weight="bold">{datum}</Text>
-          </Box>
+          <Tip content={datum.state} dropProps={{ align: { left: 'right' } }}>
+            <Box
+              key={index}
```
ericsoderberghp, line 22: Don't need `key` here. Should have it on `Tip`.

src/js/components/Drop/stories/TrapFocus.js
ShimiSun: Good catch, I think I've lost it while merging conflicts on this story. I'll roll it back

ShimiSun: I added a comment. Let me know what you think.

ericsoderberghp (approved):

Issue #4609, opened 2020-10-14
backgroundIsDark (?) fails to detect values in linear-gradient background


oskari:
My experience with Grommet is still limited but I am under assumption that it does try to detect component background and adjust light or dark text usage based on it. If this is not the case, then my assumption on this issue is probably wrong.

### Expected Behavior

Text to have good contrast in dark linear-gradient background

### Actual Behavior

Text has bad contrast in dark linear-gradient background. This is probably almost impossible to make fool-proof due to the nature of linear-gradients, but maybe there is some other way to control the text in these cases?

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-background-linear-gradient-rgba-u3oi7

### Your Environment

Dependencies as listed by codesandbox: 
Package | Version
-|-
Grommet | ^2.12.0 (2.15.1)
grommet-icons | ^4.4.0 (4.5.0)
react | ^16.6.1 (16.13.1)
react-dom | ^16.6.1 (16.13.1)
styled-components | 5.0.1 
- Grommet version: 2.15.1
- Browser Name and version: FF Dev 82.0b
- Operating System and version (desktop or mobile): MacOS Catalina

ericsoderberghp:
You are correct that grommet attempts to calculate the luminosity of a background to determine whether dark or light text color would provide better contrast. There are some situations that grommet is unable to detect though, such as images or gradients. The design intent is that you should be able to explicitly indicate via something like: `background={{ color: 'linear-gradient()', dark: true }}`. But, I tried that and it doesn't appear to be working. I think the starting place for this issue is to address using linear-gradient() via `background.color`.

oskari:
I didn't notice the `dark: boolean` in `background` props, that seems logical option to solve this, just have to handle linear(&radial?)-gradient usage in color so the css property is formed correctly.

ShimiSun:
Thank you for reporting, issue is fixed and available on the stable branch and will be available officially on the next release.

Issue #4608, opened 2020-10-14
Box background property doesn't work properly with linear-gradient that has RGBA values


oskari:
### Expected Behavior

Form correct css style for Box component background when using linear-gradient that has rgba values as stops.

### Actual Behavior

Box component ends up using the linear gradient in `background-color` css property.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-background-linear-gradient-rgba-u3oi7

### Your Environment

Dependencies as listed by codesandbox: 
Package | Version
-|-
Grommet | ^2.12.0 (2.15.1)
grommet-icons | ^4.4.0 (4.5.0)
react | ^16.6.1 (16.13.1)
react-dom | ^16.6.1 (16.13.1)
styled-components | 5.0.1 
- Grommet version: 2.15.1
- Browser Name and version: FF Dev 82.0b
- Operating System and version (desktop or mobile): MacOS Catalina

ShimiSun:
Thank you for reporting, issue is fixed and available on the [stable branch](https://github.com/grommet/grommet/wiki/What-is-grommet-stable-and-how-to-use-it%3F) and will be available officially on the next release.

PR #4616, opened 2020-10-15
Fixed issues with linear-gradient + rgb() and rgba() backgrounds

ericsoderberghp:
#### What does this PR do?

We were detecting rgb(), rgba, and hsl() color strings without a preceding `^` in the regular expressions. This meant that linear-gradients using those colors were incorrectly being evaluated as simple colors, not gradients. The change is to adjust the regular expressions to match on the beginning of the string, like we already did for hex colors like `#f2f2f2`.

White-box testing with various gradient colors.

#4608 #4609

no

yes

backwards compatible


ShimiSun (approved):

jcfilben:
Screen Shot 2020-10-16 at 2 28 54 PM
If you take a look at the circle CI output 16 of the jest tests are being skipped in this PR. I looked at earlier PR's and I am not seeing the tests being skipped in the earlier ones. I think this PR is the one that is causing them to be skipped. I will try to look into this more and see which tests are being skipped and why

PR #4619, opened 2020-10-16
Select does not display initial option if key is present and option is not an object

### LINKED ISSUE #4335: Select does not display the selected value
I updated my packages today and noticed that my selects are not showing the selected value. I am using the `valueKey` and `labelKey` properties as shown in the examples, but to no avail.

I performed some tests and the problem occurs from the version v2.13.0.

### Expected Behavior

Display the selected value

### Actual Behavior

Not applicable

Here's a Codesandbox template that serves as a nice starting point for demonstrating an issue:
https://codesandbox.io/s/jovial-rgb-eszoe?file=/src/App.js

### Steps to Reproduce

Not applicable

### Your Environment

- Grommet version: 2.14.0
- Browser Name and version: Versão 84.0.4147.89 (Versão oficial) 64 bits

IanKBovard:
#### What does this PR do?
closes #4335 

Select/utils.js

Storybook using
```
import { Grommet, Select, Form, FormField } from 'grommet';

const OPTIONS = [
  { id: '1', name: 'Option 1' },
  { id: '2', name: 'Option 2' },
  { id: '3', name: 'Option 3' },
];

const SimpleSelect = () => {
  const [value, setValue] = useState('1'); //2.13 and above requires useState({.id: '1  }) to display an initial value
  const [options] = useState(OPTIONS);
  return (
    <Grommet>
      <Select
        value={value}
        valueKey="id"
        labelKey="name"
        placeholder="Please select a option"
        onChange={event => setValue(event.value)}
        options={options}
      />
    </Grommet>
  );
};

storiesOf('Select', module).add('Simple', () => <SimpleSelect />);

```

Storybook

Before 2.13 when valueKey was present users could pass down a value that was not an object through to select.

current scenario using code example from above:
the initial value of option is '2' and the value of key is 'id'
```
export const applyKey = (option, key) => {
  if (option === undefined) return undefined;
  if (typeof key === 'object') return applyKey(option, key.key);
  if (typeof key === 'function') return key(option);
  if (key !== undefined) return option[key]; // reaches this codeblock evaluates ['2']['id'] and returns undefined
  return option;
};
```

my fix
```
export const applyKey = (option, key) => {
  if (option === undefined) return undefined;
  if (typeof key === 'object') return applyKey(option, key.key);
  if (typeof key === 'function') return key(option);
  if (key !== undefined typeof option === 'object' return option[key]; // checks if option is an object
  return option; // reaches this instead
};
```

My only issue with this is that, this fix really only applies to the initial render. onChange will make the value { id: 1 }. I think allowing incorrect option shapes in value promotes bad practices with grommet.

This is mostly to make users coming from pre 2.13 not experience any bugs if they upgrade to 2.15 and have this syntax already implemented. 

#4335 

No
No
Backward compatible

### DIFF
--- a/src/js/components/Select/utils.js
+++ b/src/js/components/Select/utils.js
@@ -2,6 +2,6 @@ export const applyKey = (option, key) => {
   if (option === undefined) return undefined;
   if (typeof key === 'object') return applyKey(option, key.key);
   if (typeof key === 'function') return key(option);
-  if (key !== undefined) return option[key];
+  if (key !== undefined && typeof option === 'object') return option[key];
   return option;
 };


ericsoderberghp:
See my comment on the issue: https://github.com/grommet/grommet/issues/4335#issuecomment-713041289

IanKBovard:
Saw the comment. Was unaware about reduce.

PR #4630, opened 2020-10-19
Added tests for Chart utils

shabtaisharon:
#### What does this PR do?
Added tests for Chart utils
`src/js/components/Chart/__tests__/utils-test.js`
`yarn test`
`yarn test`
No
#4539
Screen Shot 2020-10-19 at 4 16 59 PM
Screen Shot 2020-10-19 at 4 15 46 PM
No
No
backwards compatible

### DIFF
--- a/src/js/components/Chart/__tests__/utils-test.js
+++ b/src/js/components/Chart/__tests__/utils-test.js
@@ -1,0 +1,63 @@
+import { areNormalizedBoundsEquals, areNormalizedValuesEquals } from '../utils';
+
+describe('utils', () => {
+  test('areNormalizedValuesEquals', () => {
+    expect(
+      areNormalizedValuesEquals([{ value: [1] }], [{ value: [1] }]),
+    ).toBeTruthy();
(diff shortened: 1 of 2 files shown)


ShimiSun:
Those two functions are not being used by grommet-core, I'm wondering if the better approach to increase the coverage and add simplicity will be to simply remove those functions without needing the tests. @ericsoderberghp, any chance users will be using those functions? we do not publish those externally and I'm inclining to simply remove them.

ericsoderberghp:
Agreed that we aren't using them internally. And, we don't document them. However, we are exporting them. So, someone could be using them. I'm wondering if we should add a `console.warn()` to say that they will be removed and ask people to get in touch with us if they have concerns.

shabtaisharon:
> Agreed that we aren't using them internally. And, we don't document them. However, we are exporting them. So, someone could be using them. I'm wondering if we should add a `console.warn()` to say that they will be removed and ask people to get in touch with us if they have concerns.

@ericsoderberghp I've updated my PR

ShimiSun (approved):
Thanks! 🎉

PR #4636, opened 2020-10-20
Update snapshots from master

taysea:
#### What does this PR do?
These snapshots were updating when I ran `yarn test-update` directly after pulling from master. No local changes made. Just pushing these up to master so they don't clutter future PRs.


ShimiSun:
The snapshot changes are due to your yarn.lock file update, we need to explore why those styles are changing before comfortably upgrading the dependencies.

taysea:
Thanks for the quick reply. You're right, it's because I had updated the yarn.lock locally. Will close this currently and revert my local branch to the published yarn.lock.

PR #4644, opened 2020-10-23
Fix responsive breakpoint values for Box

### LINKED ISSUE #4246: Style ResponsiveBreakpoint values other than "small" generate an exception
see https://codesandbox.io/s/grommet-scroll-works-yhm1o?file=/index.js

StyledBox gapStyle() is looking for theme.global.breakpoints.medium.edgeSize which is undefined

jcfilben:
#### What does this PR do?
Previously when responsiveBreakpoint was used for Box and set to a value other than "small" the code would generate an exception because on base.js the medium and large breakpoints are undefined.

Issue demonstrated here: https://codesandbox.io/s/grommet-scroll-works-yhm1o?file=/index.js

In this PR if the breakpoints are not defined by the user, responsiveBreakpoint will use the global breakpoints for edgeSize, borderSize, and size. This fixes the issue with Box and the breakpoints. 

**Note:** there may still be other components that need to be fixed to work with responsiveBreakpoint. I am currently looking at Layer to see if it needs to be fixed as well. If there are other components that need fixes I can open a separate PR for them.

tested with the following story:
```javascript
import React from "react";
import { Grommet, ThemeContext, Box, Text } from "grommet";

const SimpleBox = () => (
    <Grommet>
        <ThemeContext.Extend
        //   value={{ box: { responsiveBreakpoint: "small" } }}
          value={{ box: { responsiveBreakpoint: "medium" } }}
        //   value={{ box: { responsiveBreakpoint: "large" } }}
        >
          <Box
            direction={"row-responsive"}
            gap={"small"}
            border={{
                color: '#000000',
                size: "medium",
                style: "dashed",
                side: "all"
              }}
            margin={"medium"}
            pad={"medium"}
            round={"small"}
          >
            <Text>hello</Text>
            <Text>word</Text>
          </Box>
        </ThemeContext.Extend>
    </Grommet>
);

storiesOf('Box', module).add('TEST', () => <SimpleBox />);

```

Closes #4246 

No

Maybe

Backwards Compatible


src/js/components/Box/StyledBox.js
```diff
@@ -73,6 +73,9 @@ const directionStyle = (direction, theme) => {
   ];
   if (direction === 'row-responsive' && theme.box.responsiveBreakpoint) {
     const breakpoint = theme.global.breakpoints[theme.box.responsiveBreakpoint];
+    if (!breakpoint.edgeSize) breakpoint.edgeSize = theme.global.edgeSize;
+    if (!breakpoint.borderSize) breakpoint.borderSize = theme.global.borderSize;
+    if (!breakpoint.size) breakpoint.size = theme.global.size;
```
halocline, line 78: @jcfilben It seems like we are repeating this logic in multiple places which may or may not be we want, but it at least gives me enough pause to wonder if there might be an opportunity to handle this uniformly across the project in a utils function perhaps?

You also mentioned in your write up needing to investigate Layer. I peeked and yes it does look like in addition to StyledBox.js and border.js (which you have identified), StyledLayer.js and StyledHeading.js will need to be updated, as well as possibly responsive.js and styles.js which might be tangentially related.

Did you happen to explore any solutions that might be applied more universally? 

I haven't vetted these yet, but perhaps an approach which instead of calling 
```
const breakpoint =
    theme.box.responsiveBreakpoint &&
    theme.global.breakpoints[theme.box.responsiveBreakpoint];
```
we could instead call a utility function which returns a normalized breakpoint style object?
```
const breakpoint =
    theme.box.responsiveBreakpoint &&
    getBreakpointStyle(theme.box.responsiveBreakpoint);
```

The function might live in `utils/responsive.js` and would take in a t-shirt size string as a parameter and would return an object in the shape of:
```
{
  borderSize: { ... },
  edgeSize: { ... },
  size: { ... },
  value: ...
}
```

Again, haven't vetted this approach, but curious if you have explored or would be willing to explore something along these lines?

@ShimiSun or @ericsoderberghp might have some thoughts before pursuing this angle.

jcfilben: I agree that this would be a solution worth exploring because it could solve the duplicated code issue as well as solving the breakpoint issue in other components. I will explore this option and update this PR as I go

halocline (approved):
@jcfilben I think this is looking good.

src/js/utils/responsive.js
```diff
@@ -29,3 +29,11 @@ export const getBreakpoint = (windowWidth, theme) => {
 
 export const getDeviceBreakpoint = (type, theme) =>
   theme.global.deviceBreakpoints[type];
+
+export const getBreakpointStyle = (theme, breakpointSize) => {
+  const breakpoint = breakpointSize && theme.global.breakpoints[breakpointSize];
```
ericsoderberghp, line 34:
```suggestion
  const breakpoint = (breakpointSize && theme.global.breakpoints[breakpointSize]) || {};
```

ericsoderberghp: This will be a little more defensive in case `breakpointSize` isn't truth-y or there isn't a breakpoint in the theme at that size.

jcfilben: Done

ericsoderberghp (approved):

Issue #4617, opened 2020-10-15
Pagination component API ideation


MikeWalrath:
Pagination component that can also integrate with other components (ex: table, list) 
Design thinking: https://docs.google.com/document/d/1rBlrQPNiV9Mvl9mjea3rWb4xq5tVshWECi0AWGYpOvk/edit
https://www.figma.com/file/slRqNvhJub0QlgN46zSXoY/HPE-Pagination-Component?node-id=1%3A9

Construct a variety of code examples of how this could be done. For instance,
- embedded within DataTable/List or as a separate component
- just the "< 1 2 3 4 >" stuff or also including the count summary

We don't need to implement pagination, just explore what API surface we should target.

halocline:
In progress draft of potential functionality: https://docs.google.com/document/d/1rBlrQPNiV9Mvl9mjea3rWb4xq5tVshWECi0AWGYpOvk/edit#heading=h.lg9p3h5pqpgw

ShimiSun:
Thanks @taysea @halocline! Feel free to share the API for the component with the community to start a NEW component discussions

halocline:
> Thanks @taysea @halocline! Feel free to share the API for the component with the community to start a NEW component discussions

https://github.com/grommet/grommet/pull/4646

ShimiSun:
Let's track the pagination effort under #4647 or on a design-system issue. Feel free to link between relevant issues.

PR #4646, opened 2020-10-23
NEW Pagination component

halocline:
[Storybook preview](https://deploy-preview-4646--sad-tereshkova-173d07.netlify.app/)

#### What does this PR do?

Provides a first pass proposal and prototype examples for `<Pagination />`.

**Demonstrates proof of concept:**
- Pagination applied to `<List />` and `<DataTable />`

```
<List
  data={data}
  paginationProps={{
    border: { side: 'top', color: 'border' },
    show: 6,
    step: 15,
    pad: { top: 'small' },
  }}
/>
```
```
<DataTable
  columns={columns}
  data={[...DATA]}
  step={10}
  paginationProps={{
    margin: { top: 'small' },
    pad: { vertical: 'small' },
    round: 'xsmall',
    step: 4,
  }}
 />
```
- Pagination as a parent which renders its children
```
<Pagination items={data}>
  {item => <Text key={item.entry}>{item.entry}</Text>}
</Pagination>
```

**Proposed API Surface:**
```
export interface PaginationProps {
  a11yTitle?: string,
  items: number | (string | number | React.ReactElement)[ ], /* Either array of items if called directly, or number of items if called indirectly via another component */
  numEdgePages?: number, /* number page indices to display at indices bounds */
  numMiddlePages?: number, /* number page indices to display pre/post current page index */
  onChange?: function,
  page?: number,        /* current page */
  show?: number,        /* initial page index to show, defaultPage? */
  showFirst?: boolean,  /* show button for jump to first page index */
  showLast?: boolean,   /* show button for jump to last page index */
  step?: number,        /* number of items to show per page, itemsPerPage? */
  children?: function,    /* function to be called when each item is rendered */
  boxProps?: BoxTypes,  /* any box props */
}
```
**Functionality brainstorm:**
For users:

- Should allow user to progress to next step of results
- Should allow user to return to the previous step of results
- Should provide an easy mechanism for the user to jump to the first page of results
- Should provide an easy mechanism for the user to jump to the last page of results
- Should provide an indicator for the total number of pages
- Should provide an indicator for the current page

For callers:

- Should provide simple/minimalistic implementation, but allow for opt-in added functionality
- Should be able to paginate the following layouts / visualizations:
  - Grid of results
  - List of results
  - Rows in a Table (?)
  - Rows in a DataTable
  - Others? (e.g. Drop?)
- Should allow caller to specify the number of items per page
- Should automatically calculate the number of pages so the caller does not have to
- Should allow the caller to call directly (?)
  - Should allow caller to provide an array of items
  - Should allow caller to specify how each item is rendered
- Should allow the caller to use Pagination indirectly via other components such as List, Table (?), DataTable, and Grid.
  - Should allow the caller to specify Pagination props via the calling component

**TODOs:**
- Gather feedback on interface and approach
- Documentation
- Unit tests
- Stories

Pagination.js & Pagination stories

[Storybook](https://deploy-preview-4646--sad-tereshkova-173d07.netlify.app/?path=/story/pagination--list)

TODOS:
- Unit tests & snapshots 
- Clean up and elaborate stories

Storybook

https://github.com/grommet/grommet/issues/4617
https://github.com/grommet/grommet/issues/4647

Eventually.

Eventually.


ericsoderberghp:
A couple of initial thoughts ...

I think we should provide a reasonable default experience. For example:
`<List paginate />` and `<DataTable paginate />` would indicate to use pagination instead of infinite scroll. Everything would happen automatically.

DataTable and List already have `step`. Can we leverage that?

I think we should initially focus the props on behavior and the theme on styling. That way, the theme controls the visual look the vast majority of the time. We can add prop overrides for one-off styling as needed.

I think a `children` render function approach would need some state passed. For example: `(item, { active }) => {}`.

I'm on the fence whether the property should be `items`, like InfiniteScroll or `data`, like DataTable and List.

I think we should consider adding `show` to DataTable and List, like we have it in InfiniteScroll. This would allow pre-positioning the scroll/page to a particular data item. I'm thinking that if `primaryKey` or `primary` is defined, the `show={}` would match that. We could also do `show={{ page: # }}` as a way to pre-load a particular page.

halocline:
ToDos: 
- [ ] Children render func (Matt)
- [ ] Extend show prop to allow for specified item to show (Taylor)
- [ ] Document props, types, theme
- [ ] Generate list of test scenarios
- [ ] Generate list of Storybook examples
- [ ] Add tests
- [ ] Add stories

taysea:
> ToDos:
> 
> * [ ]  Children render func
> * [ ]  Extend show prop to allow for specified item to show
> * [ ]  Document props, types, theme
> * [ ]  Generate list of test scenarios
> * [ ]  Generate list of Storybook examples
> * [ ]  Add tests
> * [ ]  Add stories

Getting started on documentation updates for what we have so far. If we change conventions slightly before merging, it won't cause much rework.

taysea:
**Updated Pagination API Surface:**

```
export interface PaginationProps {
  a11yTitle?: string,
  children?: function,  /* function to be called when each item is rendered */
  items: number | (string | number | React.ReactElement)[ ], /* Either array of items if called directly, or number of items if called indirectly via another component */
  numEdgePages?: number, /* number of page indices to display at indices bounds */
  numMiddlePages?: number, /* number of page indices to display pre/post current page index */
  onChange?: function,
  page?: number,  /* current page */
  show?: number | { index: number },  /* initial page to show, or if provided with children, initial child index to show*/
  showFirst?: boolean,  /* show button for jump to first page index */
  showLast?: boolean,   /* show button for jump to last page index */
  step?: number,  /* number of items to show per page */
  boxProps?: BoxTypes,  /* any box props */
}
```

**Additions to DataTable:**

```
  paginate?: bool,
  paginationProps?: object,    /* accepts any Box props or Pagination props */
  show?: number | { page: number },  /* by default, initial item index to show. if provided {page: #}, initial page to show */
```

**Additions to List:**

```
  paginate?: bool,
  paginationProps?: object,    /* accepts any Box props or Pagination props */
  show?: number | { page: number },  /* by default, initial item index to show. if provided {page: #}, initial page to show */
```

ericsoderberghp:
Some thoughts on this...

> **Updated Pagination API Surface:**
> 
> ```
> export interface PaginationProps {
>   a11yTitle?: string,
>   children?: function,  /* function to be called when each item is rendered */
>   items: number | (string | number | React.ReactElement)[ ], /* Either array of items if called directly, or number of items if called indirectly via another component */

Is an "item" a "page"? Or is it more like an InfiniteScroll item, where there is a 1-1 with the data? I'm guessing an "item" corresponds to a "page".  See comments below. I'm wondering if this should be "pages".

Why could this be a React.ReactElement?

I'm wondering if this scenario would be helpful: `<Pagination data={data} step={50} onChange={(page, beginIndex, endIndex) => {setListData(data.slice(beginIndex, endIndex+1)}} />`
I think I'd like to explore an example using it "manually" with a List and/or DataTable.

>   numEdgePages?: number, /* number of page indices to display at indices bounds */

I think the mix of "item" and "page" terms is confusing me a bit. Given "items" above, I would have expected this to maybe be "numEdgeItems".

>   numMiddlePages?: number, /* number of page indices to display pre/post current page index */
>   onChange?: function,

What is the function signature, arguments?

>   page?: number,  /* current page */
>   show?: number | { index: number },  /* initial page to show, or if provided with children, initial child index to show*/

How is "page" different from "show"? For instance, what if I set `page={3} show={20}`? How do I show both?

>   showFirst?: boolean,  /* show button for jump to first page index */

Is this essentially equivalent to "numEdgePages={1}"? Or, when would I use both "showFirst" and "numEdgePages"?

>   showLast?: boolean,   /* show button for jump to last page index */
>   step?: number,  /* number of items to show per page */

Why does Pagination need to know this, if all other properties above deal only with page units? Perhaps this is related to my question about what an "item" is.

> }
> ```
> 
> **Additions to DataTable:**
> 
> ```
>   paginate?: bool,
>   paginationProps?: object,    /* accepts any Box props or Pagination props */
>   show?: number | { page: number },  /* by default, initial item index to show. if provided {page: #}, initial page to show */
> ```
> 
> **Additions to List:**
> 
> ```
>   paginate?: bool,
>   paginationProps?: object,    /* accepts any Box props or Pagination props */
>   show?: number | { page: number },  /* by default, initial item index to show. if provided {page: #}, initial page to show */
> ```

taysea:
> Some thoughts on this...
> 
> > **Updated Pagination API Surface:**
> > ```
> > export interface PaginationProps {
> >   a11yTitle?: string,
> >   children?: function,  /* function to be called when each item is rendered */
> >   items: number | (string | number | React.ReactElement)[ ], /* Either array of items if called directly, or number of items if called indirectly via another component */
> > ```
> 
> Is an "item" a "page"? Or is it more like an InfiniteScroll item, where there is a 1-1 with the data? I'm guessing an "item" corresponds to a "page". See comments below. I'm wondering if this should be "pages".
> 
> Why could this be a React.ReactElement?
> 
> I'm wondering if this scenario would be helpful: `<Pagination data={data} step={50} onChange={(page, beginIndex, endIndex) => {setListData(data.slice(beginIndex, endIndex+1)}} />`
> I think I'd like to explore an example using it "manually" with a List and/or DataTable.
> 
> > numEdgePages?: number, /* number of page indices to display at indices bounds */
> 
> I think the mix of "item" and "page" terms is confusing me a bit. Given "items" above, I would have expected this to maybe be "numEdgeItems".
> 
> > numMiddlePages?: number, /* number of page indices to display pre/post current page index */
> > onChange?: function,
> 
> What is the function signature, arguments?
> 
> > page?: number,  /* current page _/
> > show?: number | { index: number },  /_ initial page to show, or if provided with children, initial child index to show*/
> 
> How is "page" different from "show"? For instance, what if I set `page={3} show={20}`? How do I show both?
> 
> > showFirst?: boolean,  /* show button for jump to first page index */
> 
> Is this essentially equivalent to "numEdgePages={1}"? Or, when would I use both "showFirst" and "numEdgePages"?
> 
> > showLast?: boolean,   /* show button for jump to last page index _/
> > step?: number,  /_ number of items to show per page */
> 
> Why does Pagination need to know this, if all other properties above deal only with page units? Perhaps this is related to my question about what an "item" is.
> 
> > }

Thanks for the feedback, Eric.

- `items` refers to either the total # of items the pagination component needs to paginate OR an array of items that will be rendered via child render function. 

- `onChange` is triggered any time the page changes. It returns an event where `event.page` is the new active page.

- Good point re: `page` vs `show` -- The thought was that `show` could take a number and would represent the page to show by default OR an object of form { index: # } where index would refer to the index of a child to show (if provided with children). We should consider the conflict between `page` and `show` and could likely consolidate them into one.

- Regarding your other questions, maybe this screenshot from the "Simple Pagination" storybook starter example will be helpful:
Screen Shot 2020-12-09 at 1 15 56 PM

`showFirst` and `showLast` refer to if there should be a "jump to first page" / "jump to last page" button available to the user.

`numEdgePages` and `numMiddlePages` refers to how many page buttons are visible in the pagination component, it does not have to do with the items.

Open to reworking the prop names to have more clarity, but let me know if that explanation helps clear up some of the initial questions.

halocline:
> I'm on the fence whether the property should be `items`, like InfiniteScroll or `data`, like DataTable and List.
> 

Based on the subsequent terminology questions, I am starting to lean toward `data` as the property. It seems that "pages" and "items" may be too closely related and lead to confusion, whereas "data" is distinctly different than "page."

taysea:
Change `items` to `numItems`
Remove `show`, just use `page`
onChange -- begin / end index
showFirst, showLast -- not start with this
numMiddlePages >= 1, refers to the entire middle set. bias towards moving forward if provided with even #

taysea:
Appreciate all the feedback yesterday, @ericsoderberghp. We have incorporated it and here is a summary of the updated API:

**Updated Pagination API Surface:**

```
export interface PaginationProps {
  a11yTitle?: string,
  numItems: number, /* number of items to paginate */
  numEdgePages?: number, /* number of page buttons to display at each end of control bounds */
  numMiddlePages?: number, /* number of page buttons to display in the middle group of page controls (between the "..."'s) */
  onChange?: ({ page, startIndex, endIndex }) => {},
  page?: number,  /* default page to show, or can be used with onChange as a controlled component */
  step?: number,  /* number of items to show per page */
}
```

**Additions to DataTable:**

```
  paginate?: bool,
  paginationProps?: object,    /* accepts any Box props or Pagination props */
  show?: number | { page: number },  /* by default, initial item index to show. if provided {page: #}, initial page to show */
```

**Additions to List:**

```
  paginate?: bool,
  paginationProps?: object,    /* accepts any Box props or Pagination props */
  show?: number | { page: number },  /* by default, initial item index to show. if provided {page: #}, initial page to show */
```

ericsoderberghp:
I'm good with the props proposed for the Pagination component. One suggestion for DataTable and List is to just have `paginate?: boolean | PaginationProps`

taysea:
> I'm good with the props proposed for the Pagination component. One suggestion for DataTable and List is to just have `paginate?: boolean | PaginationProps`

Nice, I like this.

src/js/components/DataTable/README.md
```diff
@@ -616,6 +626,22 @@ When supplied, causes checkboxes to be added to each row to indicate
...
+        the index of an item to show. If provided as an object in the format of 
+        show={{ page: 2 }}, the default page to show.
+
+```
+number
+{
+  page: number
+}
+{
+  index: number
```
taysea, line 641: Can we get rid of this now that pagination doesn't have a `show index` ?

src/js/components/DataTable/DataTable.js
```diff
@@ -219,62 +223,35 @@ const DataTable = ({
     console.warn('DataTable cannot combine "size" and "resizeble".');
   }
 
+  const [page, showItem] = normalizeShow(showProp, 'dataTable');
```
ericsoderberghp, line 226: Something feels a bit odd, to have to pass "dataTable" to this function, like every time we call this function from a new component, we have to make the function aware of that component. Can we come up with something that makes more sense from the context of `normalizeShow`?

src/js/components/DataTable/DataTable.js
```diff
@@ -219,62 +223,35 @@ const DataTable = ({
...
+  const [setPage, currentItems, currentPage] = usePagination({
```
ericsoderberghp, line 227: I think a return value like `[items, page, setPage]` would be more consistent with the other hooks.

src/js/components/DataTable/DataTable.js
```diff
@@ -219,62 +223,35 @@ const DataTable = ({
...
+    data: adjustedData,
+    paginationProps: { page, showItem, step, ...paginate },
```
ericsoderberghp, line 229: Since we're already in "usePagination" context, can we simplify this down to `{data: adjustedData, show, step, ...paginate}`?

src/js/components/DataTable/DataTable.js
```diff
@@ -219,62 +223,35 @@ const DataTable = ({
...
-        }
-        onSort={sortable || sortProp || onSortProp ? onSort : undefined}
-        onToggle={onToggleGroups}
-        primaryProperty={primaryProperty}
-      />
-      {groups ? (
-        <GroupedBody
-          background={normalizeProp(background, 'body')}
-          border={normalizeProp(border, 'body')}
+    <>
```
ericsoderberghp, line 233: I think when we have `paginate` we should wrap the table in a Box that contains the table and the pagination, that will provide a place to affect the layout, such as gap and alignment.

src/js/components/DataTable/DataTable.js
```diff
@@ -283,50 +260,98 @@ const DataTable = ({
...
+              placeholder
+            )}
+          </StyledPlaceholder>
+        )}
+      </StyledDataTable>
+      {paginate && currentItems && (
+        <Pagination
+          direction="row"
+          numItems={data.length}
+          justify={paginate.justify || 'end'}
```
ericsoderberghp, line 345: I'm wondering whether `justify` should be the responsibility of Pagination or of the caller. If it's up to the caller, then the caller can use a Footer with the custom controls on one side and Pagination on the other.

src/js/components/DataTable/DataTable.js
ericsoderberghp, line 227: Something feels a bit confusing to have both `page` and `currentPage`.

src/js/utils/pagination.js
```diff
@@ -0,0 +1,53 @@
...
+  if (['dataTable', 'list'].includes(context)) {
+    // by default, show refers to the index of an item,
+    // but if using pagination, show can take the form of { page: # },
+    // where page refers to the page # to show
+    if (typeof showProp === 'number') showItem = showProp;
+    else if (typeof showProp === 'object' && 'page' in showProp)
+      page = showProp.page;
+  }
+
+  return [page, showItem];
```
ericsoderberghp, line 16: I'm wondering if it would make things cleaner through the calls to return `{ page, item }`. This would allow us to passing this around as a `show` variable/property.

src/js/utils/pagination.js
```diff
@@ -0,0 +1,53 @@
+import { useCallback, useEffect, useState } from 'react';
+
+export const normalizeShow = (showProp, context) => {
+  let page;
+  let showItem;
+
+  if (['dataTable', 'list'].includes(context)) {
```
ericsoderberghp, line 7: Since we only call this from DataTable and List, can we just remove `context`?

src/js/components/DataTable/DataTable.js
```diff
@@ -283,50 +260,98 @@ const DataTable = ({
...
+          onChange={event => {
```
ericsoderberghp, line 346: We aren't using the indices in the event here. Can we?

src/js/components/DataTable/DataTable.js
```diff
@@ -283,50 +260,98 @@ const DataTable = ({
...
+              </Box>
+            ) : (
+              placeholder
+            )}
+          </StyledPlaceholder>
+        )}
+      </StyledDataTable>
+      {paginate && currentItems && (
+        <Pagination
+          direction="row"
```
ericsoderberghp, line 343: Can `direction="row"` be the default?

src/js/components/DataTable/DataTable.js
ericsoderberghp: I'm wondering if `usePagination` could take on more of this via:
`const [items, paginationProps] = usePagination({ data, show, step, ...paginate })`
and then use Pagination via:
`<Pagination {...paginationProps} />`
All DataTable and List really care about is the list of `items`.

src/js/components/DataTable/doc.js
```diff
@@ -259,6 +265,16 @@ export const doc = DataTable => {
       object. If 'onSelect' is provided, the CheckBoxes are enabled
       and this function can be used to track select changes.`,
     ),
+    show: PropTypes.oneOfType([
+      PropTypes.number,
+      PropTypes.shape({ page: PropTypes.number }),
+    ])
+      .description(
+        `To be used in conjunction with paginate. If provided as a number,
```
ericsoderberghp, line 273: I'm not sure we need to make this only for when using `paginate`. Even with InfiniteScroll it should be able to work.

src/js/components/Pagination/PageIndex.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Box } from '../Box';
+import { Button } from '../Button';
+import { Text } from '../Text';
+
+export const PageIndex = ({ page, ...rest }) => {
+  return (
+    <Box as="li">
+      {page === 'more-prev' || page === 'more-next' ? (
+        <Box pad="small">
```
ericsoderberghp, line 10: To keep the DOM simpler, could we remove the Box and just do `<Text margin="small" weight="bold">&#8230;</Text>`?

src/js/components/Pagination/PageIndex.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Box } from '../Box';
+import { Button } from '../Button';
+import { Text } from '../Text';
+
+export const PageIndex = ({ page, ...rest }) => {
+  return (
+    <Box as="li">
+      {page === 'more-prev' || page === 'more-next' ? (
```
ericsoderberghp, line 9: since the ellipsis isn't really a "page", maybe we could add something like a `separator` property. Or, maybe even make that a separate component, like "PageSeparator"?

src/js/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,207 @@
...
+      onChange,
+      page: pageProp,
+      step: stepProp,
+      ...rest
+    },
+    ref,
+  ) => {
+    const theme = useContext(ThemeContext) || defaultProps.theme;
+
+    const [, , defaultPage, step] = usePagination({
```
ericsoderberghp, line 30: This looks like we're trying to put too much into `usePagination`. We already have the `step`. Do we even need `defaultPage` via this method?

src/js/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,207 @@
...
+    if (middlePagesBegin > numEdgePages + 2) beginFlex = ['more-prev'];
+    else if (numEdgePages + 1 < totalPages - numEdgePages)
+      beginFlex = [numEdgePages + 1];
+
+    let endFlex = [];
+    if (middlePagesEnd < totalPages - numEdgePages - 1) endFlex = ['more-next'];
+    else if (totalPages - numEdgePages > numEdgePages)
+      endFlex = [totalPages - numEdgePages];
+
+    const navPages = [
```
ericsoderberghp, line 99: Something is feeling a bit off to use the term "pages" here when it contains things other than pages. Maybe just "controls", since some are to direct pages and other are more generic?

ericsoderberghp (commented):
Here are some comments to get the discussion started. I didn't look at everything, figuring we could start with these first.

Nice to see a lot of things covered in all of this. It definitely helps to see it in use.

src/js/components/DataTable/DataTable.js
taysea: I think we can probably simplify this function now. Previously we were having a different behavior to normalize "show" for Pagination, so context was helping with clarity, but I think we can omit. I'll update.

src/js/components/DataTable/DataTable.js
taysea: Good idea. I like that. From a props / theming perspective we can discuss what we'd want the naming structure to be.

src/js/components/DataTable/DataTable.js
taysea: I think we can clean up a bit here by adjusting variable naming. I'll do an iteration and see if it's more clear.

src/js/utils/pagination.js
taysea: Yes, will update. Context was lingering from a previous iteration.

src/js/components/Pagination/Pagination.js
taysea: Thanks, I've cleaned up.

src/js/utils/pagination.js
taysea: Done.

src/js/utils/pagination.js
taysea: I think we can actually simplify this so it handles the normalizing internally and just returns the expected active page. I've udpated.

src/js/components/DataTable/DataTable.js
taysea: Cleaned this up.

ericsoderberghp:
Screen Shot 2020-12-17 at 8 22 55 AM

Visually, as I interact with the controls, the buttons shift frequently, based on the font widths. I'm wondering if we can structure it in a way that the previous/next arrows are more fixed.

Also, the button feel a bit too prominent in the grommet theme. I think we should consider something closer to how Calendar dates are styled.

src/js/components/DataTable/DataTable.js
taysea: Removed in recent changes, user controls this via `paginate` prop if they want.

src/js/components/DataTable/DataTable.js
taysea: Adjusted in recent commit.

src/js/components/Pagination/PageIndex.js
taysea: Agreed. Adjusted.

src/js/components/DataTable/DataTable.js
```diff
@@ -283,50 +261,92 @@ const DataTable = ({
...
+            ) : (
+              placeholder
+            )}
+          </StyledPlaceholder>
+        )}
+      </StyledDataTable>
+      {paginate && items && (
+        <Pagination
+          alignSelf="end"
+          numItems={data.length}
```
ericsoderberghp, line 345: Can `numItems` come from `paginationProps`?

src/js/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,202 @@
...
+      const adjustedEvent = event;
+      adjustedEvent.page = nextPage;
+
+      // for controlled use cases, provide user with info on
+      // what range of indices should be displayed given the active page
+      const { startIndex, endIndex } = getItemIndices(nextPage);
+      adjustedEvent.startIndex = startIndex;
+      adjustedEvent.endIndex = endIndex;
+
+      setActivePage(event.page);
```
ericsoderberghp, line 111: event -> adjustedEvent

src/js/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,202 @@
...
+
+      if (onChange) {
```
ericsoderberghp, line 113: I think we can put the `event.persist()` and almost everything else within this `if (onChange)` and just call `setActivePage(nextPage)` otherwise.

src/js/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,202 @@
...
+        onClick: event => {
+          handleClick(event, control);
+        },
+        separator: control === 'more-prev' || control === 'more-next',
+        ...navProps[control],
+      };
+    });
+
+    return (
+      <StyledPaginationContainer {...theme.pagination.container} {...rest}>
```
ericsoderberghp, line 180: Do we even need the container? Can we just have the Nav be the outermost component?

src/js/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,202 @@
...
+        <Nav a11yTitle={a11yTitle || 'Pagination Navigation'} ref={ref}>
+          <Box as="ul" {...theme.pagination.controls}>
```
ericsoderberghp, line 182: Can we remove this Box and merge it with the Nav above? `<Nav as="ul" .../>`

src/js/components/Pagination/__tests__/Pagination-test.js
```diff
@@ -0,0 +1,262 @@
...
+for (let i = 0; i < 95; i += 1) {
+  data.push(`entry-${i}`);
+}
+
+const TestPagination = ({ step, ...rest }) => {
+  const [currentData] = useState(data.slice(0, step));
+
+  return (
+    <>
+      {currentData.map(datum => (
```
ericsoderberghp, line 26: Do we even need to have any data elements when testing Pagination?

src/js/components/Pagination/__tests__/Pagination-test.js
```diff
@@ -0,0 +1,262 @@
...
+    );
+
+    // default step is 10
+    const expectedPageCount = Math.ceil(NUM_ITEMS / 10);
+    const lastPageButton = getByText(expectedPageCount.toString());
+
+    expect(lastPageButton).toBeTruthy();
+    expect(container.firstChild).toMatchSnapshot();
+  });
+
```
ericsoderberghp, line 63: How about testing more edge cases? For example, `numItems` in [0, 1, < step, == step, > step].

src/js/components/Pagination/stories/DataTable.js
```diff
@@ -0,0 +1,26 @@
...
+          sortable
+          step={3}
+          paginate
+        />
+      </Box>
+    </Grommet>
+  );
+};
+
+export { PaginatedDataTable as DataTable };
```
ericsoderberghp, line 26: I think this story should be under DataTable instead of Pagination, since that's the outer context.

ericsoderberghp (commented):
Great progress!

src/js/components/Pagination/Pagination.js
taysea: The reason I had added it now was thinking if we'd ever want Pagination to have ability to internally control some sort of "showSummary" where user could pass in a summary message like "Showing 1-10 of 95 results". Mostly thinking of it from the perspective of use w DataTable/List where Pagination is controlled internally by those components. Felt that if that was something we may want to add in future, setting up the theming in this way would leave us flexibility for that.

src/js/components/Pagination/Pagination.js
taysea: I think we want Pagination to appear as a `nav` in the DOM with the `ul` is its child. I will confirm.

src/js/components/Pagination/stories/DataTable.js
taysea: That works. I'd assume the same for the List story as well then. I can move them both.

src/js/components/DataTable/DataTable.js
taysea: Yes, good call.

src/js/components/Pagination/Pagination.js
taysea: Done.

src/js/components/Pagination/Pagination.js
taysea: To confirm, if user provides onChange then we won't increment the page for them?

taysea: or we increment the page, and then apply whatever additional onChange functionality they provide.

src/js/components/Pagination/stories/DataTable.js
taysea: For the "Grid" story -- it's really an example of using Pagination as a controlled component vs integrated into DataTable/List. Do you feel like Grid is an appropriate title or should it be "Controlled"?

src/js/components/Pagination/Pagination.js
taysea: Yes, we want nav with ul child. Here's a few links for reference:
https://www.a11ymatters.com/pattern/pagination/
https://a11y-style-guide.com/style-guide/section-navigation.html

src/js/components/Pagination/stories/DataTable.js
ericsoderberghp: I think Grid is fine

src/js/components/Pagination/Pagination.js
ericsoderberghp: Thanks for investigating.

src/js/components/Pagination/Pagination.js
ericsoderberghp: I'm generally reluctant to plan for potential future functionality too much. But, I think I can overcome that in this instance.

src/js/components/Pagination/Pagination.js
ericsoderberghp: > or we increment the page, and then apply whatever additional onChange functionality they provide.

this one ^^^

src/js/components/Pagination/__tests__/Pagination-test.js
taysea: We don't have to. This was testing the case where the caller is controlling the page via state and confirming that the proper number of results displays. However, I'm realizing this is more a test of the caller's implementation vs something that the component is doing. I am going to remove this test.

src/js/components/Pagination/Pagination.js
taysea: Sounds good. I understand reluctance to do that in general.

src/js/components/DataTable/doc.js
taysea: Agreed. Adjusted docs and made sure functionality is working.

src/js/components/DataTable/DataTable.js
taysea: Updated in recent changes, the caller decides this

src/js/components/DataTable/DataTable.js
taysea: Simplified in recent updates.

src/js/components/DataTable/DataTable.js
taysea: Done, returns [items, paginationProps].

src/js/components/DataTable/DataTable.js
taysea: Done.

src/js/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,207 @@
...
+  ) => {
+    const theme = useContext(ThemeContext) || defaultProps.theme;
+
+    const [, , defaultPage, step] = usePagination({
+      data: numItems,
+      paginationProps: { page: pageProp, step: stepProp },
+    });
+    const [activePage, setActivePage] = useState(defaultPage);
+
+    const getPageIndices = (begin, end) => {
```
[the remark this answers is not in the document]
taysea: Done.

src/js/components/DataTable/DataTable.js
taysea: Added `dataTable.container` and `StyledContainer`. Inside DataTable, if `paginate` is used, then the DataTable and Pagination cmponents are wrapped in `StyledContainer`, otherwise it behaves as is.

taysea:
An update on where we are with this component:

With regards to behavior/function, we are in a good spot. I'd expected only minor cleanups.

What we need to do still is centered around styling of the component:
- [ ] What is the theming structure?
   - Direction 1: Adding a button "kind" called `pagination`. However, this only would work for themes with button default defined. Maybe we can find a way to make it work with the old button behavior as well?
   - Direction 2: `theme.pagination.button` is used to define the theme structure. -- would have to find a way for Button to know to use that versus its go-to styling
   - Another consideration for either direction. It should be able to accept an object or string where string would reference an existing kind. For example `theme.pagination.button` could have value `secondary` meaning it would apply secondary button styling to the pagination component.
- [ ] Make pagination buttons have a more consistent width (right now the width varies often)

halocline:
@taysea Tag. You're it.

ShimiSun:
@taysea Thanks for summing it up!! Can we also see an example of how pagination behaves with the kind button (HPE)?

halocline:
[Storybook preview](https://deploy-preview-4646--sad-tereshkova-173d07.netlify.app/)

+ @ericsoderberghp and @ShimiSun this is also in a state where we could use others' eyes to create POCs and collect feedback against the work that is here.

src/js/components/Pagination/StyledPageControl.js
```diff
@@ -0,0 +1,104 @@
...
+  
+  border-radius: ${defaultStyles.border.radius};
+  ${props => sizeStyle(props).content};
+`;
+
+export const StyledContainer = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  height: 32px; // default line-height + default padding
```
ericsoderberghp, line 98: We should pull using theme values, since line-height can change.

src/js/themes/base.d.ts
```diff
@@ -946,6 +950,30 @@ export interface ThemeType {
     colors?: GraphColorsType;
     extend?: ExtendType;
   };
+  pagination?: {
+    container?: {
+      extend: ExtendType;
+    };
+    controls?: {
+      align?: AlignContentType;
+      direction?:
```
ericsoderberghp, line 959: Can we have a `DirectionType` and re-use with Box?

ericsoderberghp (commented):
Looking great, just a few minor comments.

src/js/components/Pagination/StyledPageControl.js
```diff
@@ -0,0 +1,104 @@
...
+// Assemble button theme object for each button state
+const buildButtonStyle = (buttonTheme, kind) => {
+  const style = buttonTheme[kind];
+  BUTTON_STATES.forEach(state => {
+    if (state !== 'normal') style[state] = buttonTheme[state][kind];
+  });
+  return style;
+};
+
+// Allow padding to flex when border is present. Calculated adjusted
```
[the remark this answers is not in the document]
taysea: Yes, I reviewed his PR a little today and it looks like we're doing very similar things. Will see if his work would alleviate the need for these calculations.

src/js/themes/base.d.ts
taysea: Done.

src/js/components/List/stories/PaginatedList.js
```diff
@@ -0,0 +1,40 @@
...
+              key={index}
+              icon={<More />}
+              hoverIndicator
+              items={[{ label: 'one' }]}
+            />
+          );
+        }}
+        step={3}
+        show={{ page: 7 }}
+        paginate={{ margin: { top: 'medium' } }}
```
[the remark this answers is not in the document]
taysea: Good callout to remove custom margin. It revealed that this would be the default List behavior. It feels like we should put some space between the List and pagination by default, but that the user could override this if desired. What do you think?

Screen Shot 2021-01-20 at 4 32 39 PM

src/js/components/Button/Button.js
```diff
@@ -134,16 +136,20 @@ const Button = forwardRef(
       );
     }
 
+    // kindArg is object if we are referencing a theme object
+    // outside of theme.button
+    const kindObj = typeof kindArg === 'object';
```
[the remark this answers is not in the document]
taysea: Good call, I'll take a look at that.

src/js/components/List/doc.js
```diff
@@ -123,6 +129,16 @@ export const doc = List => {
...
+    ])
+      .description(
+        `If provided as a number, the index of an item to show. If using 
+        paginate and provided as an object in the format of show={{ page: 2 }}, 
+        the default page to show.`,
+      )
+      .defaultValue(undefined),
     step: PropTypes.number
       .description('How many items to render at a time.')
       .defaultValue(50),
```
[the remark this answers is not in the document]
taysea: Thanks, I've updated to use 50.

src/js/components/List/List.js
```diff
@@ -67,9 +70,11 @@ const List = React.forwardRef(
       focus,
       itemProps,
       pad,
+      paginate,
       primaryKey,
       secondaryKey,
-      step,
+      show: showProp,
+      step = paginate ? 10 : undefined,
```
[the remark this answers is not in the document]
taysea: Thanks. I've updated to use 50. For backwards compatibility, we still need to check if user is paginating or not since previously had no default step set.

src/js/components/Pagination/StyledPageControl.js
```diff
@@ -0,0 +1,104 @@
...
+      '0px';
+
+    const adjustedPad = `${pad.replace('px', '') - border.replace('px', '')}px`;
+    adjustedPadding[state] = { pad: undefined };
+    adjustedPadding[state].pad = adjustedPad;
+  });
+  return adjustedPadding;
+};
+
+// sizeStyle is exploratory, but would allow user to apply a size
```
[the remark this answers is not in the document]
taysea: I'll leave that one up to @halocline who did the exploratory work there.

src/js/components/List/stories/PaginatedList.js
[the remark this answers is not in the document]
ericsoderberghp: The default case should look reasonable. This sounds like some default margin in the base theme might be in order.

src/js/components/DataTable/doc.js
```diff
@@ -210,6 +210,12 @@ export const doc = DataTable => {
       `Cell padding. You can set the padding per context by passing an
       object with keys for 'heading', 'body', and/or 'footer'.`,
     ),
+    paginate: PropTypes.oneOfType([PropTypes.bool, PropTypes.object])
+      .description(
+        `Whether to paginate the data. If providing an object, any Box props or 
+      Pagination props are valid.`,
+      )
```
ShimiSun, line 217: It is not clear to me how accepting an object will affect the component.  Can we please elaborate?

src/js/themes/base.d.ts
```diff
@@ -612,6 +614,9 @@ export interface ThemeType {
     body?: {
       extend?: ExtendType;
     };
+    container?: {
```
ShimiSun, line 617: The container should also be able to accept Box props if I understood correctly.

src/js/themes/base.d.ts
```diff
@@ -946,6 +951,25 @@ export interface ThemeType {
...
+      gap?: GapType;
+      pad?: PadType;
+      margin?: MarginType;
+    };
+    icons?: {
+      color: ColorType;
+      first: React.ReactNode;
+      last: React.ReactNode;
+      next: React.ReactNode;
+      previous: React.ReactNode;
```
ShimiSun, line 970: I don't think all these icons have to be defined. You might be missing the question marks symbol to indicate those are optional.

src/js/themes/base.js
```diff
@@ -698,6 +698,10 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       //      extend: undefined,
       //   },
       // },
+      // container: {
+      //   // any box props
+      //   // extend: undefined,
```
ShimiSun, line 703:
```suggestion
      //   extend: undefined,
```

src/js/themes/base.js
```diff
@@ -945,6 +949,70 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       // colors: [] || colors: ['graph-0', 'graph-1', 'graph-2', 'graph-3'],
       // extend: undefined,
     },
+    pagination: {
+      // container: {
+      //   // any box props,
+      //   // extend: undefined,
```
ShimiSun, line 955:
```suggestion
      //   extend: undefined,
```

src/js/themes/base.js
```diff
@@ -945,6 +949,70 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
+      //       font: { ...fontSizing(1) },
+      //       height: `${baseSpacing * 2}px`,
+      //       width: `${baseSpacing * 2}px`,
+      //       border: {
+      //         radius: `${baseSpacing / 2}px`, // 12
+      //         width: '2px',
+      //       },
+      //     },
+      //   },
+      // },
```
ShimiSun, line 1002: Why is this block needed? Is it still relevant?

src/js/components/Pagination/doc.js
```diff
@@ -0,0 +1,98 @@
...
+      )
+      .defaultValue(1),
+    numItems: PropTypes.number
+      .description('The number of items to paginate.')
+      .defaultValue(undefined),
+    numMiddlePages: PropTypes.number
+      .description(
+        `The number of page buttons visible in the middle of the controls.`,
+      )
+      .defaultValue(3),
```
ShimiSun, line 35: We usually try not to use shortcuts on the API  prop names (i.e. num vs number). Did we try to think about better naming?

src/js/components/Pagination/doc.js
```diff
@@ -0,0 +1,98 @@
...
+      .defaultValue(10),
+  };
+
+  return DocumentedPagination;
+};
+
+export const themeDoc = {
+  pagination: {
+    description: `The possible sizes of the Pagination in terms of its 
+      max-width, font-size and line-height.`,
```
ShimiSun, line 62: according to the description, any reason not to use `pagination.size` instead of `pagination`? to be more consistent with theme props.

src/js/components/Pagination/stories/Custom.js
```diff
@@ -0,0 +1,91 @@
...
+  pagination: {
+    button: 'secondary',
+  },
+});
+
+const primaryTheme = deepMerge(hpe, {
+  pagination: {
+    button: 'primary',
+  },
+});
```
ShimiSun, line 47: This is nice for the POC, but we should potentially remove it from the grommet's storybook.

src/js/components/Pagination/stories/Grid.js
```diff
@@ -0,0 +1,52 @@
+import React, { useState } from 'react';
+
+import { Box, Card, Grid, Grommet, Text } from 'grommet';
+import { grommet } from 'grommet/themes';
+import { Pagination } from '..';
```
ShimiSun, line 5: Can we move this import to line 3, to make sure the component is imported well from the overall package and not just locally? (Relevant for all story examples)

src/js/themes/base.d.ts
```diff
@@ -946,6 +951,25 @@ export interface ThemeType {
     colors?: GraphColorsType;
     extend?: ExtendType;
   };
+  pagination?: {
+    container?: {
+      extend: ExtendType;
+    };
```
ShimiSun, line 957: The container should also be able to accept Box props if I understood correctly.

src/js/components/Pagination/stories/Custom.js
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
+
+import { Box, Grommet, Text } from 'grommet';
+import { grommet } from 'grommet/themes';
+import { deepMerge } from 'grommet/utils';
+import { FormPreviousLink, FormNextLink } from 'grommet-icons';
```
ShimiSun, line 6:
```suggestion
import { FormPreviousLink } from 'grommet-icons/icons/FormPreviousLink';
import { FormNextLink } from 'grommet-icons/icons/FormNextLink';
```

ShimiSun: Not a must a better a better practice.

src/js/components/Pagination/Pagination.js
```diff
@@ -0,0 +1,202 @@
...
+      totalPages,
+    );
+
+    let numMiddlePages;
+    if (numMiddlePagesProp < 1) {
+      numMiddlePages = 1;
+      console.warn(
+        // eslint-disable-next-line max-len
+        `Property "numMiddlePages" should not be < 1. One middle page button will be shown. Set "numMiddlePages" >= 1 to remove this warning.`,
+      );
```
ShimiSun, line 56: Is there a way to control the input within propTypes and typescript instead of us placing the warning?

ShimiSun (commented):
The first phase of API review.

src/js/components/Pagination/stories/Custom.js
taysea: Agreed, using this for demo purposes but will adjust so it's not hpe specific.

src/js/themes/base.js
taysea: If we choose to implement the `size` prop, yes (https://github.com/grommet/grommet/pull/4646#discussion_r561400324). If we choose not to include it in the first pass, then we will remove it.

src/js/components/Pagination/doc.js
taysea: We did not consider using `number`. I think we were aiming to keep the prop names as consolidated as possible, but open to switching to `numberMiddlePages` to align with grommet conventions. Thoughts here @halocline ?

src/js/components/Pagination/Pagination.js
taysea: I will have a look and see.

src/js/components/DataTable/doc.js
taysea: Yes we can elaborate. If desired, the user can provide styling to the pagination here, such as customizing the margin, alignment, background color, etc. Since the pagination component is built into DataTable, this provides a way to style separate from the theme if desired.

src/js/components/Button/Button.js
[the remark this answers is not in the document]
taysea: Done.

src/js/components/List/List.js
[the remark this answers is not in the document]
taysea: Done.

src/js/components/DataTable/DataTable.js
```diff
@@ -283,50 +274,87 @@ const DataTable = ({
...
+            primaryProperty={primaryProperty}
+            onToggle={onToggleGroup}
+            size={size}
+          />
+        ) : (
+          <Body
+            background={normalizeProp(background, 'body')}
+            border={normalizeProp(border, 'body')}
+            columns={columns}
+            data={!paginate ? adjustedData : items}
```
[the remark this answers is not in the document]
taysea: Yes, we can.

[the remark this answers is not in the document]
taysea: Whoops, correcting my above comment. We need the paginate condition when passing `show` through to InfiniteScroll. If paginating, then pagination controls the data that is viewed. However, if not paginating it is fine to pass it through to InfiniteScroll as long as it is a number value (indicating the index of a data item to show).

src/js/components/DataTable/doc.js
taysea: Done.

src/js/themes/base.d.ts
taysea: Done.

src/js/components/DataTable/doc.js
ShimiSun: Did you push the changes?

src/js/themes/base.js
ShimiSun: OK, so I'll keep this thread open so we can circle back to it before merging.

src/js/components/Pagination/StyledPageControl.js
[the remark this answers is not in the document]
halocline: I am thinking we will still need some logic to accommodate.  We will pull in his branch and investigate.

src/js/components/Pagination/StyledPageControl.js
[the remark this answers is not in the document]
halocline: If we like the concept, will add a story to demonstrate the functionality.

src/js/components/Pagination/stories/Grid.js
halocline: Done.

src/js/components/DataTable/doc.js
taysea: Not yet, pushing now. sorry for confusion.

src/js/components/Pagination/doc.js
taysea: I have changed to use `numberMiddlePages`, etc.

src/js/components/Pagination/Pagination.js
taysea: So for this, we are setting  numberMiddlePages to 1 if they set it to anything less than that. The warning is intended to just alert the user that we are changing something for them. If this warning feels unnecessary, we can just remove it.

halocline:
TODO:
- [ ] Update themeDocs for pagination

src/js/components/Pagination/stories/Custom.js
taysea: Done.

src/js/themes/base.d.ts
taysea: Done.

src/js/components/DataTable/doc.js
taysea: Done.

src/js/components/Pagination/StyledPageControl.js
taysea: These values are now coming from theme and based on a calculation that takes line-height and padding into account.

src/js/components/Pagination/StyledPageControl.js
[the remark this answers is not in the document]
taysea: Story has been added.

ShimiSun:
@YuriyKovalcuk Since your team is already using the pagination concept, we'd appreciate your feedback on the [functionality]( https://deploy-preview-4646--sad-tereshkova-173d07.netlify.app/?path=/story/*) of the Pagination component and your inputs on whether the component will be sufficient for the use cases of your team. Bests.

package.json
```diff
@@ -82,33 +82,33 @@
...
+    "eslint-plugin-jsx-a11y": "^6.2.1",
+    "eslint-plugin-prettier": "^3.0.0",
+    "eslint-plugin-react": "^7.9.1",
     "eslint-plugin-react-hooks": "^2.2.0",
     "fs-extra": "^8.1.0",
     "grommet-theme-aruba": "^0.1.1",
     "grommet-theme-dxc": "^0.1.1",
     "grommet-theme-hp": "^0.1.1",
-    "grommet-theme-hpe": "^2.0.2",
+    "grommet-theme-hpe": "^2.0.0",
```
ShimiSun, line 111: The `package.json` changes look like a result of a bad merge conflict, please be mindful it downgrades the mentioned dependencies.

src/js/components/Button/Button.js
```diff
@@ -152,23 +160,30 @@ const Button = forwardRef(
...
         }
-        result.hover.push('hover', `hover.${kind}`);
+        result.hover.push('hover');
+        if (!kindObj) result.hover.push(`hover.${kind}`);
         if (active) {
-          result.hover.push(`hover.active`, `hover.active.${kind}`);
+          result.hover.push(`hover.active`);
+          if (!kindObj) {
+            result.hover.push(`hover.active.${kind}`);
+          }
```
ShimiSun, line 182: Are these changes are part of this PR or also a result of a merge conflict? @MikeKingdom might have an insight as well.

src/js/themes/base.d.ts
```diff
@@ -952,6 +956,77 @@ export interface ThemeType {
     colors?: GraphColorsType;
     extend?: ExtendType;
   };
+  pagination?: {
+    button?: {
```
ShimiSun, line 960: If the button type will be changed in the future, we might not be mindful to change the type instance on the pagination component as well, in order to avoid that, can we reuse the button type in a generic way so there will be one reusable type instance?

src/js/components/Tabs/__tests__/__snapshots__/Tabs-test.js.snap
```diff
@@ -2615,6 +2615,7 @@ exports[`Tabs should apply custom theme disabled style when theme.button.default
         aria-expanded="true"
         aria-selected="true"
         class="c3"
+        kind="default"
```
ShimiSun, line 2618: How are these changes related to this PR?

ShimiSun:
This behavior of pagination doesn't look right to me, if the active element isn't really changing its position, why are we going back and forth with the active background on every page change.

https://user-images.githubusercontent.com/6320236/107295386-7e008280-6a2c-11eb-9f42-532df53bf86d.mov

src/js/components/Pagination/stories/Custom.js
```diff
@@ -0,0 +1,101 @@
...
+        },
+      },
+    },
+  },
+  pagination: {
+    button: 'bright',
+  },
+});
+
+export const Custom = () => (
```
ShimiSun, line 70: Can we have more spacing between the examples? the render section is looking too dense.

src/js/components/Pagination/stories/Size.js
```diff
@@ -0,0 +1,27 @@
...
+import { Box, Grommet, Pagination, Text } from 'grommet';
+import { grommet } from 'grommet/themes';
+
+export const Size = () => (
+  <Grommet theme={grommet}>
+    <Box align="start" pad="small" gap="medium">
+      <Box>
+        <Text>Default</Text>
+        <Pagination numberItems={237} />
+      </Box>
```
ShimiSun, line 12: I think we can remove this example, and be more explicit about 'Medium' as the default.

src/js/components/Pagination/stories/Size.js
```diff
@@ -0,0 +1,27 @@
...
+      <Box>
+        <Text>Small</Text>
+        <Pagination numberItems={237} size="small" />
+      </Box>
+      <Box>
+        <Text>Medium</Text>
```
ShimiSun, line 18:
```suggestion
        <Text>Medium (Default)</Text>
```

src/js/components/Pagination/stories/Pagination.stories.js
```diff
@@ -0,0 +1,10 @@
+export { Custom } from './Custom';
+export { Grid } from './Grid';
+export { NumberEdgePages } from './NumberEdgePages';
+export { NumberMiddlePages } from './NumberMiddlePages';
+export { Simple } from './Simple';
+export { Size } from './Size';
+
+export default {
+  title: 'Controls/Pagination',
+};
```
ShimiSun, line 10: With the new storybook upgrade, this file should be removed.

src/js/components/Pagination/stories/Size.js
halocline: Removed and incorporated recommendation.

src/js/components/Pagination/stories/Size.js
halocline: Done.

src/js/components/Pagination/stories/Custom.js
halocline: Done. Added spacing.

src/js/components/Pagination/stories/Pagination.stories.js
halocline: Removed.

src/js/components/Button/Button.js
halocline: This does touch work done by @MikeKingdom, however these changes are related to this PR. Allows for callers to theme their pagination controls by specifying a kind.

package.json
halocline: `package.json` is inline with master and diffs are no longer present.

src/js/components/Tabs/__tests__/__snapshots__/Tabs-test.js.snap
halocline: Unrelated. This has been resolved.

src/js/themes/base.d.ts
halocline: Good question. I believe our thinking here was that we are only allowing customization of background colors and not supporting other customizations such as border. Would we be mis-setting expectations if we used `ButtonKindType`?

halocline:
> This behavior of pagination doesn't look right to me, if the active element isn't really changing its position, why are we going back and forth with the active background on every page change.

I will take a look at the implementation here to investigate our options here.

src/js/themes/base.d.ts
ShimiSun: Understood, thanks for clarifying. I think that if that's the case we should be good.

halocline:
> > This behavior of pagination doesn't look right to me, if the active element isn't really changing its position, why are we going back and forth with the active background on every page change.
> 
> I will take a look at the implementation here to investigate our options here.
> 
> [Comment from Taylor]: Investigating here, and I'm testing locally and noticing it only happens when we're moving forward, but doesn't happen when moving backwards. Trying to understand what may be different between selecting "next" vs "previous"

Fix is in place to eliminate the jump: [Preview Here](https://deploy-preview-4646--sad-tereshkova-173d07.netlify.app/?path=/story/controls-pagination-custom--custom)

src/js/components/List/List.js
```diff
@@ -48,6 +51,14 @@ const StyledItem = styled(Box)`
     props.theme.list && props.theme.list.item && props.theme.list.item.extend}
 `;
 
+// when paginated, this wraps the data table and pagination component
+const StyledContainer = styled(Box)`
+  ${props =>
+    props.theme.list &&
+    props.theme.list.container &&
```
taysea, line 58: Need to add `list.container` to docs

src/js/components/DataTable/doc.js
```diff
@@ -452,6 +481,12 @@ export const themeDoc = {
     type: 'React.Element',
     defaultValue: undefined,
   },
+  'dataTable.pagination': {
+    description: `If paginating, any valid Box props to style the pagination 
+    component.`,
+    type: 'object',
+    defaultValue: `margin: { top: 'xsmall' }`,
```
taysea, line 488: I don't see this default value in base.js, I think we achieve the desired styling with dataTable.container. Can this be removed?

src/js/components/List/doc.js
```diff
@@ -171,4 +189,10 @@ export const themeDoc = {
     type: 'string | (props) => {}',
     defaultValue: undefined,
   },
+  'list.pagination': {
+    description: `If paginating, any valid Box props to style the pagination 
+    component.`,
+    type: 'object',
+    defaultValue: `margin: { top: 'xsmall' }`,
```
taysea, line 196: I don't see this default value in base.js, I think we achieve the desired styling with list.container. Can this be removed?

taysea: Fixed.

src/js/components/DataTable/doc.js
taysea: Fixed.

src/js/components/List/List.js
taysea: Done.

taysea (approved):
Behavior and file changes lgtm.

src/js/components/Pagination/stories/Custom.js
```diff
@@ -0,0 +1,120 @@
...
```
ShimiSun, line 70: Can we please move this file under stories/typescript/Custom.tsx

This will make sure all TS definitions are working as expected?

taysea: Done.

src/js/components/Pagination/stories/typescript/Custom.tsx
```diff
@@ -1,11 +1,10 @@
...
 
-import { Box, Grommet, Pagination, Text } from 'grommet';
+import { Box, Grommet, Pagination, Text, ThemeContext } from 'grommet';
 import { grommet } from 'grommet/themes';
 import { deepMerge } from 'grommet/utils';
 import { FormPreviousLink } from 'grommet-icons/icons/FormPreviousLink';
 import { FormNextLink } from 'grommet-icons/icons/FormNextLink';
-import { ThemeContext } from 'styled-components';
 
 const customTheme = deepMerge(grommet, {
```
ShimiSun, line 9: You should add 'ThemeType',  it will surface any potential issues with theme typing.
See how it is done on other TS examples.

taysea: Got it, I've just tested locally and fixed one item that was flagged. However, it only flags when we're not using `deepMerge`. We'd deepMerged with grommet theme to inherit the grommet font as opposed to the browser default. Would you like me to push up the code with `ThemeType` and not use `deepMerge`? Or is it fine that I tested locally and we can leave the published story as is?

ShimiSun: Let's avoid using deepMerge and grommet theme, you can explicitly add the font to the theme.

taysea: Update, I will just add "font-family: "Arial"" to the TS theme to make it look a little nicer and allow us to leave ThemeType.

ShimiSun: Please add a comment to the ThemeType usage:
```
```

ericsoderberghp (approved):

ShimiSun:
Linking to #3489

ShimiSun (approved):

PR #4650, opened 2020-10-26
Enhance Calendar logic when using 'range'

### LINKED ISSUE #4507: Not possible to select a date range from a previous month with DateInput
### Expected Behavior
When a user goes back a few months in the DateInput, they should be able to select a start date and end date in that month, regardless of the existing range value. This is common behavior (e.g. when the page loads, show the current period but let the user select a date range in the past.)

### Actual Behavior
When you go to a previous month and click on a new start date it updates correctly. However, then selecting a new end date in that month only moves the start date forward. I think we need to keep track of the previous click and if a subsequent click is on a date later than the previous click's date, it should set the end date, not the start date.

[https://storybook.grommet.io/?path=/story/dateinput--range](https://storybook.grommet.io/?path=/story/dateinput--range)

### Steps to Reproduce

1. Click the button to open the DateInput drop
2. Click the "<" previous month button, say 2 times for good measure
3. Click the first day of the that month
4. Click the last day of that month

### Your Environment

Seeing this locally and on the hosted storybook.

- Grommet version: 2.15
- Browser Name and version: Tested on Firefox 80 and Chrome 85.0.4183.83
- Operating System and version (desktop or mobile): macOS 10.15.5

taysea:
#### What does this PR do?
This PR enhances the Calendar logic to allow for more predictable date handling when using range. Previously, whichever date the user last selected (end or start date of range) would be the one affected on the next selection. This posed problems when users left the current month and selected from a previous/future month because there was no way to close the range. These enhancements align the Calendar logic more closely with what other online Calendars are demonstrating which is that once a start date is selected, then the end date is the active range bound. once end date is selected, then start date becomes active again. By default, Calendar will now behave like this. 

Additionally, when working with range, Calendar will return an array of dates where if one end of the range is not selected, the value will be undefined. This differs from previous behavior where range would return a string if only one range bound was selected. To maintain backwards compatibility, a user must use `range="array"` to opt into this new dates array behavior.

This also introduces a new prop `activeDate` which allows the caller to explicitly state which range bound they want to be affected on date selection. The accepted values are `start` and `end`. This prop can be leveraged internally in future for DateInput to toggle which date is active depending on where a user has clicked in a MaskedInput date range, for example. I have added a storybook example that demonstrates a behavior that mirrors that of booking sites.

Added tests for new logic edge cases and added a storybook example.

Engage with `ActiveDate` storybook example or other calendar examples.

Closes #4507 

Yes, updated here.

Yes.

Yes, backwards compatible.


src/js/components/Calendar/doc.js
```diff
@@ -95,12 +101,13 @@ to disable the previous and next buttons.
       For single select, make this the subsequent \`date\` property value.
       For multiple select or ranges, toggle values in \`dates\`.
       Not specifying this property makes the component read only.`),
-    range: PropTypes.bool
+    range: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf(['bounds'])])
```
ericsoderberghp, line 104: I think `bounds` might be a bit confusing here, as Calendar already has a `bounds` property. Some other naming options:
- "uncompressed" - to indicate that `true` implies "compressed" into a single date.
- "always separate", "always distinct", or "always both" - too long but maybe conveys the concept a bit better
- "always" - shortening the previous
- "verbose" - a little odd, but hints at providing further detail
- "long" - similar to previous

src/js/components/Calendar/doc.js
```diff
@@ -20,6 +20,12 @@ export const doc = Calendar => {
 
   DocumentedCalendar.propTypes = {
     ...genericProps,
+    activeRangeBound: PropTypes.oneOf(['start', 'end'])
```
ericsoderberghp, line 23: What about `activeDate`, which would always be a single date value string? This could be leveraged to indicate where keyboard navigation should start. Just thinking out loud.

src/js/components/Calendar/Calendar.js
```diff
@@ -154,6 +155,19 @@ const Calendar = forwardRef(
   ) => {
     const theme = useContext(ThemeContext) || defaultProps.theme;
 
+    // set activeRangeBound when caller changes it, allows us to change
+    // it internally too
+    const activeRangeBounds = {
```
ericsoderberghp, line 160: This could be outside the component function, since it's a constant, perhaps line 43.

src/js/components/Calendar/Calendar.js
```diff
@@ -154,6 +155,19 @@ const Calendar = forwardRef(
...
+      start: 'start',
+      end: 'end',
+    };
+    const [activeRangeBound, setActiveRangeBound] = useState(
+      dateProp && range ? activeRangeBounds.end : activeRangeBounds.start,
+    );
+    useEffect(() => {
+      if (activeRangeBoundProp) setActiveRangeBound(activeRangeBoundProp);
+    }, [activeRangeBoundProp, setActiveRangeBound]);
```
ericsoderberghp, line 169: I don't think you need `setActiveRangeBound` in this array.

src/js/components/Calendar/Calendar.js
```diff
@@ -314,61 +325,99 @@ const Calendar = forwardRef(
...
+          if (!nextDates[0][0] && !nextDates[0][1]) nextDates = undefined;
         }
 
         setDates(nextDates);
         if (!dates) setDate(nextDate);
         setActive(new Date(selectedDate));
-        setLastSelectedDate(selectedDate);
-        if (onSelect) onSelect(nextDates || nextDate);
+        let adjustedDates;
+        if (
```
ericsoderberghp, line 399: Can we put all of this inside the `if (onSelect)`, since `adjustedDates` isn't used at all of `!onSelect`.

src/js/components/DateInput/__tests__/DateInput-test.js
```diff
@@ -274,10 +274,13 @@ describe('DateInput', () => {
     );
     expect(container.firstChild).toMatchSnapshot();
 
+    // new calendar logic adjust start date by default
     fireEvent.click(getByText('10'));
     expect(onChange).toHaveBeenCalled();
     expect(onChange).toHaveReturnedWith([
-      '2020-07-02T00:00:00-08:00',
+      // Calendar is returning UTC, but should return local
+      // this should be resolved when https://github.com/grommet/grommet/issues/4352 ågets fixed
```
ericsoderberghp, line 282: Can we fix this as part of this change?

src/js/components/Calendar/doc.js
taysea: funny you say that, I had `activeDate` initially and then changed it. But maybe since that was your inclination as well, I will go back to that.

src/js/components/Calendar/Calendar.js
taysea: good call.

src/js/components/Calendar/doc.js
taysea: That makes sense with potential confusion with the `bounds` prop. uncompressed makes most sense to me out of these options, but it's not sitting perfectly well right off the bat. Let me think a little more on some other suggestions.

src/js/components/DateInput/__tests__/DateInput-test.js
taysea: It could be if I'm able to sort out that timezone issue soon. I didn't want to hold up this PR waiting for that to merge.

src/js/components/Calendar/doc.js
taysea: Some thoughts:
- `array` -- just being explicit about the return type?
- `expanded` -- similar to uncompressed mindset
- `unfiltered`
?

ericsoderberghp: I'm leaning towards `array`

taysea: That was my inclination as well.

taysea: I've updated the value to be called "array".

src/js/components/Calendar/doc.js
```diff
@@ -95,12 +101,13 @@ to disable the previous and next buttons.
...
       For multiple select or ranges, toggle values in \`dates\`.
       Not specifying this property makes the component read only.`),
-    range: PropTypes.bool
+    range: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf(['array'])])
       .description(
         `Whether to automatically manage multiple date selection as a range.
         When the user clicks the first date, onSelect will be called with that
         date. When the user selects another date, onSelect will be called with
-        an array of two dates.`,
+        an array of two dates. If range = bounds, then an array of dates will
```
ShimiSun, line 109: What does `bounds` mean in this context?

taysea: Whoops, need to update the description. We've updated it to be `range="array"`.

ShimiSun: I thought so :)

taysea: Should be updated now, let me know if the description makes sense.

ericsoderberghp (approved):

src/js/components/Calendar/doc.js
```diff
@@ -95,12 +101,13 @@ to disable the previous and next buttons.
...
       For multiple select or ranges, toggle values in \`dates\`.
       Not specifying this property makes the component read only.`),
-    range: PropTypes.bool
+    range: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf(['array'])])
       .description(
         `Whether to automatically manage multiple date selection as a range.
         When the user clicks the first date, onSelect will be called with that
         date. When the user selects another date, onSelect will be called with
-        an array of two dates.`,
+        an array of two dates. If range = array, then an array of dates will
```
ShimiSun, line 109:
```suggestion
        an array of two dates. If range = 'array', then an array of dates will 
```
To make sure the users understand they should use the 'array' string vs the array data structure.

src/js/components/Calendar/doc.js
```diff
@@ -20,6 +20,12 @@ export const doc = Calendar => {
 
   DocumentedCalendar.propTypes = {
     ...genericProps,
+    activeDate: PropTypes.oneOf(['start', 'end'])
+      .description(
+        `Whether the next date selection will affect the start or end bound of 
+        the range.`,
```
ShimiSun, line 26: Can we add a clarification that this prop is only relevant for `range` users? I know we imply it at the end of the line, but I'm thinking we should be more clear about that right at the beginning. Otherwise the context of 'next date selection' doesn't make sense if you are not using `range`.

ShimiSun (approved):
LGTM. 
I dropped two tiny comments on the docs, because I'm a little concerned that the API isn't intuitive enough and we should make it clearer.

src/js/components/Calendar/doc.js
taysea: Good idea.

ShimiSun:

Issue #4651, opened 2020-10-26
Storybook Component Story Format


jcfilben:
### Expected Behavior
Organize the storybook stories by type of component. For example Box is a type of Layout component. All the Layout components should be grouped together like they are in the documentation https://v2.grommet.io/components.

### Actual Behavior
The stories are organized alphabetically by component.

### URL, screenshot, or Codepen exhibiting the issue
![image](https://user-images.githubusercontent.com/54560994/97214999-81c40e00-1789-11eb-8c42-19c15134b80b.png)

### Steps to Reproduce
1. Run `yarn storybook`
2. Open http://localhost:9001/

### Additional Info
Component Types:
- Layout
- Type
- Color
- Controls
- Input
- Visualizations
- Media
- Utilities

See https://github.com/grommet/grommet/pull/4316 for more info

minicrashxv:
Hello, Can I take this issue?

jcfilben:
@minicrashxv yes, go for it!

minicrashxv:
thanks ;)

ShimiSun:
@minicrashxv thank you for showing interest in working on this issue buddy! 💟 
There are more than 50 components, so small PRs per component will be appreciated.

minicrashxv:
@ShimiSun Ok ! Great. I should just finish Color part and it will be finish 😉

jcfilben:
The stories left are 'All', 'Theme', and 'Typography'. Currently these are listed under Other. Is there a different place these stories should go? If not I think this issue is done and can be closed

ShimiSun:
'Typography' should go under `Type`.
I think we should be good with the others. 
I don't think we should close this issue until we'll resolve the URL issues from the site.

PR #4654, opened 2020-10-26
Accordion Simple stories changed to new structure

jcfilben:
### What does this PR do?
Updates the stories in Accordion/stroies/simple.js to use the new structure.

yarn storybook

#4651 

No

No

Backwards compatible

### DIFF
--- a/src/js/components/Accordion/stories/Accordion.stories.js
+++ b/src/js/components/Accordion/stories/Accordion.stories.js
@@ -1,0 +1,8 @@
+export { Custom } from './Custom';
+export { Header } from './CustomHeader';
+export { Rich } from './typescript/Rich.tsx';
+export { Simple, DarkNoAnimation, Multiple } from './Simple';
+
+export default {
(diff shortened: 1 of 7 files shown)


ShimiSun:
I think it looks good, did you get a chance to see if `addParameters` is a deprecated function before we are adding it again?

Also, I think PR should be per component, and at the moment, Accordion has its stories split into two different places in storybook:
image

jcfilben:
I checked to see if `addParameters` was deprecated and it looks like it is not https://storybooks.netlify.app/configurations/options-parameter/ (We are using Storybook version 5.3.20) for version 6 I think you use `addons.setConfig` to set `showRoots`. I think we are fine using `addParameters`, I couldn't find anything about it being deprecated and no warnings in the console.

Make sure you are testing storybook with the updated branch. Should look like this:
Screen Shot 2020-10-26 at 7 24 29 PM

ShimiSun (approved):

PR #4655, opened 2020-10-26
Update Menu Button logic to account for default button

taysea:
#### What does this PR do?
This PR updates the internal logic of Menu so that if a theme defines `theme.button.default`, then the Menu button will use the default button styling for its control as opposed to a `plain` button. The existing issue was that if Button has children, it acts as plain, so a Menu button was built by creating a Box (with direction row) and adding in the label/icon to appear like a Button. However, any button hover styles etc. no longer apply.

It makes more sense that the button control would not rely on children/plain (as this is a practice we want to avoid when possible) and instead relied on the `label` and `icon` props. This approach enhances the logic so that if a theme has `theme.button.default`, the menu control will inherit all the stylings of default button. This seemed like a more swift solution than try to extend all of the styling options into menu's theme object, particularly because Menu button was currently being implemented as a Box.

This new behavior applies only when `theme.button.default` is defined in the theme, but for backwards compatibility all of the previous logic remains as is.

Tested `grommet-theme-hpe` locally in Menu stories. Added a jest test.

Apply a customTheme with `theme.button.default` styles defined (or use grommet-theme-hpe). Then, create a menu.

Related to https://github.com/grommet/hpe-design-system/issues/1199

No.

No.

Yes, backwards compatible.

### DIFF
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -225,6 +225,7 @@ const Menu = forwardRef((props, ref) => {
         plain={plain}
         onClick={onDropClose}
         onFocus={() => setActiveItemIndex(controlButtonIndex)}
+        reverse={theme.button.default && !children ? true : undefined}
         // On first tab into menu, the control button should not
         // be able to receive tab focus because the focus should
         // go to the first menu item instead.
@@ -261,7 +262,7 @@ const Menu = forwardRef((props, ref) => {
         open={isOpen}
         onOpen={onDropOpen}
         onClose={onDropClose}
-        reverse={theme.button.default ? true : undefined}
+        reverse={theme.button.default && !children ? true : undefined}
         dropContent={
           <Keyboard
             onTab={event =>
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -230,13 +230,13 @@ const Menu = forwardRef((props, ref) => {
         active={activeItemIndex === controlButtonIndex}
         focusIndicator={false}
         hoverIndicator="background"
-        {...buttonProps}
         onClick={onDropClose}
         onFocus={() => setActiveItemIndex(controlButtonIndex)}
         // On first tab into menu, the control button should not
         // be able to receive tab focus because the focus should
         // go to the first menu item instead.
         tabIndex={activeItemIndex === constants.none ? '-1' : undefined}
+        {...buttonProps}
       >
         {typeof content === 'function'
           ? () => content({ ...props, drop: true })
@@ -258,12 +258,12 @@ const Menu = forwardRef((props, ref) => {
       <DropButton
         ref={ref}
         {...rest}
+        {...buttonProps}
         a11yTitle={a11yTitle || messages.openMenu || 'Open Menu'}
         disabled={disabled}
         dropAlign={align}
         dropTarget={dropTarget}
         dropProps={dropProps}
-        {...buttonProps}
         open={isOpen}
         onOpen={onDropOpen}
         onClose={onDropClose}
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -184,20 +184,30 @@ const Menu = forwardRef((props, ref) => {
     }
   };
 
-  const content = children || (
-    <Box
-      direction="row"
-      justify={justifyContent}
-      align="center"
-      pad={theme.button.default ? undefined : 'small'}
-      gap={label && icon !== false ? 'small' : undefined}
-    >
-      <Text size={size}>{label}</Text>
-      {icon !== false
-        ? (icon !== true && icon) || <MenuIcon color={iconColor} size={size} />
-        : null}
-    </Box>
-  );
+  const menuLabel = <Text size={size}>{label}</Text>;
+  const menuIcon =
+    icon !== false
+      ? (icon !== true && icon) || <MenuIcon color={iconColor} size={size} />
+      : null;
+
+  let content;
+  if (children) content = children;
+  else if (!theme.button.default) {
+    content = (
+      <Box
+        direction="row"
+        justify={justifyContent}
+        align="center"
+        pad="small"
+        gap={label && icon !== false ? 'small' : undefined}
+      >
+        {menuLabel}
+        {menuIcon}
+      </Box>
+    );
+  } else {
+    content = undefined;
+  }
 
   const controlMirror = (
     <Box flex={false}>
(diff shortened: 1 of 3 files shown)


src/js/components/Menu/__tests__/Menu-test.js
```diff
@@ -522,4 +535,16 @@ describe('Menu', () => {
     );
     expect(component.toJSON()).toMatchSnapshot();
   });
+
+  test('custom theme with default button', () => {
```
halocline, line 539: Would it be helpful to build a snapshot test to cover the scenario where `button.default` is defined and the Menu has children? This would help ensure that the logic of rendering children if present continues to be applied.

src/js/components/Menu/Menu.js
```diff
@@ -184,20 +184,30 @@ const Menu = forwardRef((props, ref) => {
...
+      ? (icon !== true && icon) || <MenuIcon color={iconColor} size={size} />
+      : null;
+
+  let content;
+  if (children) content = children;
+  else if (!theme.button.default) {
+    content = (
+      <Box
+        direction="row"
+        justify={justifyContent}
```
halocline, line 199: Per our convo, investigating the `justifyContent` + default button scenarion.

halocline (commented):
Looking good. Two comments related to our discussion.

src/js/components/Menu/__tests__/Menu-test.js
taysea: Yes, I think you're right. We should have that as a baseline.

taysea: Added this.

src/js/components/Menu/Menu.js
taysea: An update here, currently button does not support ability to control justification of content within the width of the button. However, this could be considered a future enhancement. That being said, the benefits of `justifyContent` on Menu could still be achieved by providing Menu with children, such as:
```
<Menu>
        {() => (
          <Box align="center" gap="small" direction="row">
            <Text>Menu</Text>
            <FormDown />
          </Box>
        )}
</Menu>
```
This is not as seamless as a prop name, but the functionality is still possible. I feel like this is sufficient for now and a discussion of a button enhancement could be a future discussion. What do you think? @halocline

src/js/components/Menu/__tests__/Menu-test.js
```diff
@@ -522,4 +535,27 @@ describe('Menu', () => {
...
```
ShimiSun, line 539: Thanks for enhancing the testing! Can you try to separate the testing into a separate PR? After the testing PR will be merged it will be easier to understand the impact of changes this PR is suggesting.

taysea: Yes, will do now.

src/js/components/Menu/__tests__/__snapshots__/Menu-test.js.snap
```diff
@@ -1401,11 +1401,20 @@ exports[`Menu custom theme with default button 1`] = `
   background: transparent;
   overflow: visible;
   text-transform: none;
-  outline: none;
   border: none;
-  padding: 0;
-  text-align: inherit;
-  color: inherit;
+  border-radius: 18px;
```
taysea, line 1405: These changes are expected because `default button` is now being used instead of `plain`

src/js/components/Menu/Menu.js
```diff
@@ -184,20 +184,30 @@ const Menu = forwardRef((props, ref) => {
...
+        align="center"
+        pad="small"
+        gap={label && icon !== false ? 'small' : undefined}
+      >
+        {menuLabel}
+        {menuIcon}
+      </Box>
+    );
+  } else {
+    content = undefined;
```
taysea, line 209: add comment to capture "For button default use case, keep content as undefined and rely on Button label and icon props instead"

src/js/components/Menu/Menu.js
```diff
@@ -210,9 +220,12 @@ const Menu = forwardRef((props, ref) => {
         active={activeItemIndex === controlButtonIndex}
         focusIndicator={false}
         hoverIndicator="background"
+        label={theme.button.default && !children ? menuLabel : undefined}
```
taysea, line 223: extract logic sure to separate condition --> and name single variable

taysea: to make more readable

src/js/components/Menu/Menu.js
```diff
@@ -210,9 +220,12 @@ const Menu = forwardRef((props, ref) => {
...
+        icon={theme.button.default && !children ? menuIcon : undefined}
         plain={plain}
         onClick={onDropClose}
         onFocus={() => setActiveItemIndex(controlButtonIndex)}
+        reverse={theme.button.default && !children ? true : undefined}
```
taysea, line 228:
```suggestion
        reverse={theme.button.default && !children}
```
try this

taysea: Pull out these 3 variables and any other shared props and spread across controlMirror and DropButton

ShimiSun (approved):
Thanks for the clean-ups, much more readable. LGTM

halocline (approved):
Good to go.

src/js/components/Menu/Menu.js
```diff
@@ -184,20 +184,40 @@ const Menu = forwardRef((props, ref) => {
...
-      pad={theme.button.default ? undefined : 'small'}
-      gap={label && icon !== false ? 'small' : undefined}
-    >
-      <Text size={size}>{label}</Text>
-      {icon !== false
-        ? (icon !== true && icon) || <MenuIcon color={iconColor} size={size} />
-        : null}
-    </Box>
-  );
+  const menuLabel = <Text size={size}>{label}</Text>;
```
ericsoderberghp, line 187: Can we be using `<Button size={size} label={label} />` instead of using `<Text />` for the label?

src/js/components/Menu/Menu.js
```diff
@@ -184,20 +184,40 @@ const Menu = forwardRef((props, ref) => {
...
+  const menuIcon =
+    icon !== false
+      ? (icon !== true && icon) || <MenuIcon color={iconColor} size={size} />
+      : null;
+
+  const isDefaultButton = theme.button.default && !children;
```
ericsoderberghp, line 193: This is a minor comment...

Can we use `let buttonProps;` and then assign it in the same way we assign `content`? That would have a single source for the conditional logic.

src/js/components/Menu/Menu.js
taysea: Good thought, let me have a look.

ericsoderberghp (approved):

PR #4656, opened 2020-10-27
Update clock theme props doc

leossantiago:
#### What does this PR do?

- Adds size-related props of Digital Clock, that aren't listed in docs;

- **PROPOSAL** add possibility of using `extend` prop throgh theme for Digital Clock; It works fine for props such `color` and `background` but doesn't look so good when using props like `margin` as those styles will be appied to each digit (and separators) individually. Still, I don't see any other way to customize Digital Clock. I think using ThemeContext and styling a Box component could work aswell, but this approach looks too cumbersome. TL;DR: this works, but it's not perfect.

- Update usage of sizes (on Digital Clock) that aren't listed on theme. Before this PR, code breaks when a size thats not set (or one/both props (`size` and `height`) aren't set aswell. Now they all default to `medium` size when trying to use a prop that wasn't set;

- Update Clock typing to cover all sizes;

- Update Analog Clock sizes provided by default theme to match sizes from other components (add `xsmall` and change `huge` to `xxlarge`);

- Update `Clock/Digital` story, to showcase different/default/custom clock sizes;

- Update tests to cover all sizes + add test to custom size Digital clock; 

- Update snapshots (most changes are related to the proposed height values in Digital Clock)

- **PROPOSAL** update line-height factor on Digital Clock, to 'linear' values according to the multiplier of fontScale. In that proposal, each time fontScale is added/removed of `size`, it represents 0.0625 in `height`. 
Also: searched in `base.js` theme file history, but looks like those values were already used in first version, so I assume they're empirically determined. Correct-me if I'm wrong.

Previously: 
```
text: {
          xsmall: { size: `${baseFontSize - 2 * fontScale}px`, height: 1.5 },
          small: { size: `${baseFontSize - fontScale}px`, height: 1.43 },
          medium: { size: `${baseFontSize}px`, height: 1.375 },
          large: { size: `${baseFontSize + fontScale}px`, height: 1.167 },
          xlarge: { size: `${baseFontSize + 2 * fontScale}px`, height: 1.1875 },
          xxlarge: { size: `${baseFontSize + 4 * fontScale}px`, height: 1.125 },
        },
```
Proposed:
```
text: {
          xsmall: { size: `${baseFontSize - 2 * fontScale}px`, height: 1.5 },
          small: { size: `${baseFontSize - fontScale}px`, height: 1.4375 },
          medium: { size: `${baseFontSize}px`, height: 1.375 },
          large: { size: `${baseFontSize + fontScale}px`, height: 1.3125 },
          xlarge: { size: `${baseFontSize + 2 * fontScale}px`, height: 1.25 },
          xxlarge: { size: `${baseFontSize + 4 * fontScale}px`, height: 1.125 },
        },
```

* Note for anyone curious who might end here when searching for Clock: I thought about using `${genericStyles}` in Digital Clock, to match Analog, but AnalogClock is created using a Styled Component `svg`, which receives those genericStyles. Digital clock is created using a Box component as outer element, which receives `{...rest}` as prop. That being said, I don't think there's need for that.

`src/js/components/Clock/StyledClock.js`
and
`src/js/themes/base.js`

Storybook and jest tests.

Test default sizes on Clock component, and create custom sizes using themes (example in storybook: `Clock/Digital`)

Already did.

Probably not.

Considering the change from `huge` to `xxlarge` (Analog Clock) and the proposed `height` values (Digital Clock), it might be a breaking change, as prop values will be changed, but it looks like everything else is backwards compatible (in case we drop the breaking changes).


leossantiago:
Sorry for the huge wall of text. 

About sending PRs involving fixes that we're not related to issues, or proposals, should I always create an issue before sending a PR?

ShimiSun:
>About sending PRs involving fixes that we're not related to issues, or proposals, should I always create an issue before sending a PR?
I think that if the PR has enough details on the problem statement you should be good. 💟

ericsoderberghp:
There are some good ideas in this. However, I think it would be better to split it up into separate pull requests, as I think some are orthogonal to each other. For example:
- enhancing the documentation for the theme in doc.js
- adding sizes, we need to preserve `huge` for backwards compatibility
- revisiting the digital text font sizing
- using `extend` more. For digital, how about `theme.clock.digital.digit.extend`?

leossantiago:
Okay! I'll be splitting them soon! Should I convert this PR to a draft?

ShimiSun:
> Okay! I'll be splitting them soon! Should I convert this PR to a draft?

Yes, thank you!! ❤️

ShimiSun:
Hi buddy, I'll be closing this PR since it was broken into two separate PRs that got merged. Feel free to reopen if needed, cheers!

PR #4660, opened 2020-10-27
Changed background colors to support dark/light mode

ShimiSun:
#### What does this PR do?
Changed background-* colors to support dark and light more accurately.

base.js
codesandbox examples.
Review.
no
yes
backwards compatible

### DIFF
--- a/src/js/themes/base.js
+++ b/src/js/themes/base.js
@@ -51,16 +51,16 @@ const focusColor = accentColors[0];
 const colors = {
   active: rgba(221, 221, 221, 0.5),
   'background-back': {
-    dark: '#33333308',
-    light: '#EDEDED',
+    light: '#33333308',
+    dark: '#EDEDED',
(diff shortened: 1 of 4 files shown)


src/js/themes/base.js
```diff
@@ -51,16 +51,16 @@ const focusColor = accentColors[0];
 const colors = {
   active: rgba(221, 221, 221, 0.5),
   'background-back': {
-    dark: '#33333308',
-    light: '#EDEDED',
+    light: '#33333308',
```
ericsoderberghp, line 54: I don't think background-back or background-front need to change, just background-contrast.

ericsoderberghp (approved):

ericsoderberghp:
The Card story change that Chromatic flagged looks a bit odd, like there's padding outside of the CardFooter that shouldn't be there.

PR #4675, opened 2020-10-30
Added tests for Button kind events and icon colors

ShimiSun:
#### What does this PR do?
Added tests for Button kind events and icon colors

test file
Jest
Jest
image

image

Updated:
image

#4539 
no
no
b/c

### DIFF
--- a/src/js/components/Button/__tests__/Button-kind-test.js
+++ b/src/js/components/Button/__tests__/Button-kind-test.js
@@ -6,6 +6,8 @@ import 'regenerator-runtime/runtime';
 
 import { axe } from 'jest-axe';
 import { cleanup, fireEvent, render } from '@testing-library/react';
+import { Add } from 'grommet-icons';
+
 import { Grommet, Button } from '../..';
 import { buttonKindTheme } from './theme/buttonKindTheme';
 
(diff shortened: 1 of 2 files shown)


src/js/components/Button/__tests__/Button-kind-test.js
```diff
@@ -37,6 +39,46 @@ describe('Button kind', () => {
...
+        theme={{
+          button: {
+            default: {
+              color: '#FFF',
+              background: '#000',
+            },
+          },
+        }}
+      >
+        <Button label="label" icon={<Add />} hoverIndicator />
```
taysea, line 54: The docs note that `hoverIndicator` should only be used with `plain` buttons, but this isn't a plain button. Do we still want to have a test case?

taysea: also if this test just has a single expect, i'm not sure we are capturing the `hoverIndicator` in the snapshot

ShimiSun: Good catch, the `hoverIndicator` slipped in and is not needed for this test, let me clean it up.

ShimiSun:
Cleaned up the tests and improved coverage:

taysea (approved):
LGTM

PR #4682, opened 2020-10-31
Added tests to cover events of places and continents on WorldMap

ShimiSun:
#### What does this PR do?
* Added tests to cover events of places and continents on WorldMap
* Enable continents to accept rest props
WorldMap.js
Jest
Jest
Nop
#4539 
image

no
no
b/c

### DIFF
--- a/src/js/components/WorldMap/WorldMap.js
+++ b/src/js/components/WorldMap/WorldMap.js
@@ -551,7 +551,7 @@ const WorldMap = forwardRef(
     );
 
     const continentElements = world.continents.map(({ area, dots, name }) => {
-      const { color: continentColor, onClick, onHover } =
+      const { color: continentColor, onClick, onHover, ...restContinents } =
         continents[name] || {};
       const active = activeContinent && activeContinent === name;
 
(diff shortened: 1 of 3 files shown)


src/js/components/WorldMap/WorldMap.js
```diff
@@ -551,7 +551,7 @@ const WorldMap = forwardRef(
     );
 
     const continentElements = world.continents.map(({ area, dots, name }) => {
-      const { color: continentColor, onClick, onHover } =
+      const { color: continentColor, onClick, onHover, ...restContinents } =
```
ericsoderberghp, line 554: This will assign `name` to the `<g />` element, which I'm not sure we want. I'm thinking we should pull `name` out of `restContinents`. I'm also thinking we should document that all other properties of the content will be applied to the content graphical element.

ericsoderberghp: Actually, I'm thinking we should make the continents more accessible by using an aria-label for the continent name. We could then use that label in the tests.

ShimiSun: I made the behavior consistent with how we are accepting rest props on `places` and applied it on `continents` in a symmetric way. 
Even if we'll be using `aria-label` for 'continent', it might solve the issue for the specific test, but it would still be missing other user props similarly to how the `places` prop is allowing it. 
One scenario that I have in mind, a user would like to draw a Diagram stacked on top of the WorldMap, and the diagram connections should be the continents, there wouldn't be a way to pass the ids to the continent if `continents` isn't accepting `rest`.

ericsoderberghp (approved):

PR #4690, opened 2020-10-31
Added tests for Video

ShimiSun:
#### What does this PR do?
Added tests for Video

video.js
Jest
Jest + review
#4539 
image

Nope
Nope
b/c

### DIFF
--- a/src/js/components/Video/Video.js
+++ b/src/js/components/Video/Video.js
@@ -182,11 +182,7 @@ const Video = forwardRef(
           const scrubberRect = scrubberRef.current.getBoundingClientRect();
           const percent =
             (event.clientX - scrubberRect.left) / scrubberRect.width;
-          if (!duration) {
-            console.warn('The video provided has no valid duration property.');
-          } else {
-            videoRef.current.currentTime = duration * percent;
-          }
+          if (duration) videoRef.current.currentTime = duration * percent;
         }
       },
       [duration, videoRef],
--- a/src/js/components/Video/Video.js
+++ b/src/js/components/Video/Video.js
@@ -182,7 +182,11 @@ const Video = forwardRef(
           const scrubberRect = scrubberRef.current.getBoundingClientRect();
           const percent =
             (event.clientX - scrubberRect.left) / scrubberRect.width;
-          videoRef.current.currentTime = duration * percent;
+          if (!duration)
+            console.warn('The video provided has no valid duration property.');
+          else {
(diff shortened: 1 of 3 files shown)


src/js/components/Video/Video.js
```diff
@@ -182,7 +182,11 @@ const Video = forwardRef(
           const scrubberRect = scrubberRef.current.getBoundingClientRect();
           const percent =
             (event.clientX - scrubberRect.left) / scrubberRect.width;
-          videoRef.current.currentTime = duration * percent;
+          if (!duration)
```
ericsoderberghp, line 185: The use of curly braces is a bit inconsistent in this if/else.

src/js/components/Video/Video.js
```diff
@@ -182,7 +182,11 @@ const Video = forwardRef(
...
+            console.warn('The video provided has no valid duration property.');
```
ericsoderberghp, line 186: Is there a chance that the video hasn't been loaded from the server yet? If so, I don't know that we'd want to add this message into the browser console.

ShimiSun: Not sure if a server error will land in this space, but in some cases, you might have a malfunctioning video that has an issue with its duration.

src/js/components/Video/Video.js
ShimiSun: done.

src/js/components/Video/Video.js
ericsoderberghp: I wasn't thinking there was a server error, just that the latency or bandwidth was slower. I wouldn't want to show a console warning in that case. I think that's more likely than a malfunctioning video.

ShimiSun: So, should I change the message to:
"The video provided has no valid duration property, check your server connection."
Or you are suggesting not to have a warning at all?

ericsoderberghp: no warning

ShimiSun: In this case, the next line would break the code and send an error, are we good with it?

ericsoderberghp: No, I was thinking we'd do:
```
if (duration) videoRef.current.currentTime = duration * percent;
```

ShimiSun: Thanks. ready.

ericsoderberghp (approved):

Issue #4668, opened 2020-10-29
Select onChange target is null when using a jsx element as initial value


jimenezff:
### Expected Behavior
Select a new value and `onChange` `event.target`  shouldn't be undefined
### Actual Behavior
`event.target` is undefined
### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/s/wonderful-spence-r3ew0
### Steps to Reproduce
Sandbox: change the value and see value logged for target.
Or running this test:
failing test
```
  test("onChange target isn't undefined", () => {
    const onChange = jest.fn();
    const { getByText } = render(
      <Select
        id="test-select"
        value={<span>one</span>}
        options={[<span>one</span>, <span>two</span>]}
        onChange={onChange}
      />,
    );
    fireEvent.click(getByText('one'));
    fireEvent.click(
      document
        .getElementById('test-select__drop')
        .querySelectorAll('button')[1],
    );
    expect(onChange).toBeCalledWith(
      expect.objectContaining({ value: <span>two</span> }),
    );
    expect(onChange).toBeCalledWith(
      expect.not.objectContaining({ target: undefined }),
    );
  });
```
### Your Environment
- Grommet version: master branch
- Browser Name and version: Chrome 86
- Operating System and version (desktop or mobile): Windows 10
___

I noticed this while working on #3173. The solution for that issue should also fix this one but wanted to check if anyone has any thoughts about it.

jimenezff:
Actually, I just noticed this in the docs
```
Passing an element is deprecated. Instead, use the 'valueLabel' property.
```

PR #4694, opened 2020-11-01
Changes to onChange target and type definitions

### LINKED ISSUE #3173: Select event.value !== event.target.value
When I change the option in a Select, `event.value` is the one I selected, and `event.target.value` is the previously selected option. This is obviously not the correct behavior.

This is a pretty simple issue so unless this is unknown behavior / hard to repro, I'll leave it at that.

### Your Environment

- Grommet version: 2.7.1
- Browser Name and version: Chrome 74.0.3729.169

jimenezff:
#### What does this PR do?
- Keep onChange original target and updated value. Fixes #3173 and #4668
- Updated type definition to reflect this change, among other changes.
- Updated ts story to correctly use the new type definition.
- Added new tests to cover some of the changes to Select.js and new type definitions.

Changes to `Select.js`
 
The added tests to `Select-test.js`

Could be by running tests `event.target.value === event.target` and `onChange target isn't undefined when using valueLabel` with master branch (they will fail)

 #3173  #4668  #3165

no
no
might be an issue if some people was using onChange's `event.target.value` as a way to obtain the previous value (see #3173 for more details). However, that is not the expected behavior and it wasn't documented as such.


jcfilben (commented):
When testing with some of the storybook stories (Controlled and CustomValueAndIcon) I got the following error "Argument of type 'SelectOption' is not assignable to parameter of type 'SetStateAction<string>'. Type 'number' is not assignable to type 'SetStateAction<string>'."

jimenezff:
Thanks! I'll take a look later today

jimenezff:
@jcfilben what do you think about this implementation with generics?

- Also, `selected` prop (also in `onChange`'s event) is deprecated according to the docs. So, what do you think about removing them or adding a JSDoc `@deprecated` tag?

- This story `src\js\components\Select\stories\Seasons.js` makes use of selected. I adapted it to use `value` instead, and added it to the typescript stories.

src/js/components/Select/stories/typescript/Controlled.tsx
```diff
@@ -0,0 +1,31 @@
...
+          value={value}
+          options={options}
+          onChange={({ option }) => setValue(option)}
+        />
+      </Box>
+    </Grommet>
+  );
+};
+
+storiesOf('Select', module).add('TS-Controlled', () => <SimpleSelect />);
```
jcfilben, line 31: Story should use new storybook story structure (See #4651)

src/js/components/Select/stories/typescript/CustomValueAndIcon.tsx
```diff
@@ -0,0 +1,46 @@
...
+              <CaretDown color="black" />
+            </Box>
+          }
+        />
+      </Box>
+    </Grommet>
+  );
+};
+
+storiesOf('Select', module).add('TS-Custom', () => <CustomSelect />);
```
jcfilben, line 46: This story should also use new storybook story structure

src/js/components/Select/stories/typescript/Seasons.tsx
```diff
@@ -0,0 +1,110 @@
...
+          }
+        >
+          {renderOption}
+        </Select>
+      </Box>
+    </Grommet>
+  );
+};
+
+storiesOf('Select', module).add('TS-Seasons', () => <SeasonsSelect />);
```
jcfilben, line 110: This story should also use new storybook story structure

jcfilben (commented):
Nice job switching the Seasons story to use `value` instead of `selected`. I think a comment could be added to the Select/index.d.ts story about `selected` being deprecated but I don't think we should completely remove `selected` because of backwards compatibility issues.

ShimiSun (commented):
I think we should avoid adding more stories to TS issues unless they provide additional functional value that other stories don't, it will be helpful if you can attach a snippet of the code (story) you were testing with instead of adding it to the PR.

jimenezff:
Sure I can remove the new ones. Two of them are simply rewrites of existing js ones. However, maybe we could keep the new `Seasons.tsx` story which is making use of `value` instead of the deprecated `selected`. If so, should I delete the old `Seasons.js`?

ShimiSun:
> Sure I can remove the new ones. Two of them are simply rewrites of existing js ones. However, maybe we could keep the new Seasons.tsx story which is making use of value instead of the deprecated selected. If so, should I delete the old Seasons.js?

Good comment. Let's remove the deprecated usage and keep the more recommended usage story.

src/js/components/Select/index.d.ts
```diff
@@ -26,35 +44,43 @@ export interface SelectProps {
...
   open?: boolean;
   placeholder?: PlaceHolderType;
   plain?: boolean;
   replace?: boolean;
   searchPlaceholder?: string;
-  selected?: number | number[];
+  selected?: number[] | number;
   size?: 'small' | 'medium' | 'large' | 'xlarge' | string;
-  value?: string | JSX.Element | object | (string | number | object)[];
+  value: ValueType;
```
jcfilben, line 73:
```suggestion
  value?: ValueType;
```
Value should have `?` to support uncontrolled forms (see Uncontrolled Select story)

ShimiSun:
@jcfilben any other feedback on this PR?

jcfilben (approved):
No other feedback, this PR looks good to me

src/js/components/Select/Select.js
```diff
@@ -143,7 +141,7 @@ const Select = forwardRef(
         if (onChange) {
           event.persist();
           const adjustedEvent = event;
-          adjustedEvent.target = inputRef.current;
+          adjustedEvent.target.value = nextValue;
```
ShimiSun, line 144: I suspect this will be a breaking change.

jcfilben: Yeah it would be a breaking change, here is what was said about it in the PR description:

> might be an issue if some people was using onChange's event.target.value as a way to obtain the previous value (see #3173 for more details). However, that is not the expected behavior and it wasn't documented as such.

src/js/components/Select/stories/typescript/Seasons.tsx
```diff
@@ -17,17 +17,16 @@ const allSeasons = [
   'S10',
 ];
 
-export const Seasons = () => {
-  const [selected, setSelected] = useState([]);
+export const Seasons: React.FC = () => {
```
ShimiSun, line 20:
```suggestion
// Type annotations can only be used in TypeScript files.
// Remove ': React.FC'  if you are not using TypeScript.
export const Seasons: React.FC = () => {
```

src/js/components/Select/stories/typescript/Seasons.tsx
```diff
@@ -17,17 +17,16 @@ const allSeasons = [
...
-    setSelected(
-      selected.filter(selectedSeason => selectedSeason !== seasonIndex),
+  const onRemoveSeason = (season: string) => {
+    setValue(value =>
+      value.filter(selectedSeason => selectedSeason !== season),
     );
   };
 
-  const renderSeason = season => (
+  const renderSeason = (season: string) => (
```
ShimiSun, line 29:
```suggestion
  // Type annotations can only be used in TypeScript files.
  // Remove ': string'  if you are not using TypeScript.
  const renderSeason = (season: string) => (
```

src/js/components/Select/stories/typescript/CreateOption.tsx
```diff
@@ -27,7 +27,7 @@ const theme: ThemeType = {
   },
 };
 
-const defaultOptions = [];
+const defaultOptions: string[] = [];
```
ShimiSun, line 30:
```suggestion
// Type annotations can only be used in TypeScript files.
// Remove ': string[]'  if you are not using TypeScript.
const defaultOptions: string[] = [];
```

ShimiSun (approved):
@ericsoderberghp I approve this PR, due to backward compatibility sensitivity I wanted another pair of eyes on this PR before merging.

src/js/components/Select/Select.js
jimenezff: @ShimiSun I think normally `event.target.value` is used to retrieve to the current value of the input element that changed, but for Grommet's Select component it's actually the previous value.

jimenezff:
I added a deprecated tag in the latest commit https://github.com/grommet/grommet/pull/4694/commits/3129ae1f5b7a0f0ef11f8b2bee7d6c47f2cf5b42

src/js/components/Select/Select.js
ericsoderberghp: I think this approach is too narrowly focused, as the `event.target` will not be referring to the `<select />` DOM element but instead the `<button />` inside the drop. I think we should be using the technique we use in MaskedInput to first update the `<select />` value and then dispatching an event from it. See: https://github.com/grommet/grommet/blob/186b94413318dd7c4e30e9f8f47c124693acfd63/src/js/components/MaskedInput/MaskedInput.js#L238

ShimiSun: @jimenezff Any updates on this PR? It will be nice to get it on our next release.

ShimiSun:
Hi @jimenezff Thanks for working on this fix! I'd like to get it in our next release, any chance we can get it to the finish line and incorporate the feedback?

src/js/components/Select/Select.js
jimenezff: @ShimiSun sorry for the delay. It looks like this may have already been solved with #5155. I'll check and revert my changes to `Select.js`

jimenezff: @ShimiSun I tried it but there are still issues with #3173. How about I remove the code related to that issue and we can just make this PR about type definition and the new TS story. If you agree, I can make that change today.

jimenezff:
Hi @ShimiSun, regarding https://github.com/grommet/grommet/pull/4694#discussion_r617071599, would you prefer I create a new PR for Select's TS type definitions and story (I have the branch ready to go https://github.com/jimenezff/grommet/commit/cfd2702913e5fd2e880cca074a810b095e729cb1) or should I update this PR title and description?

jcfilben:
Hi @jimenezff due to the age and merge conflicts on this PR lets close this one and create a new PR to update the type definitions

PR #4705, opened 2020-11-04
Drop - Added DOM handling for elements without viewport position

ShimiSun:
#### What does this PR do?
Currently, Drop target doesn't open correctly on any HTML element that doesn't have the offsetParent property, which includes SVG (translated in grommet to Chart, Meter, and Icon components), the target will be opened on the component's target parent position (as long as it allows offsetParent, recursively) which may be dramatically off from the desired position, and in some cases even not placed on the visible browser screen.

This PR is suggesting a way to calculate the properties of x, y, height, width, and more in order for those components to be placed correctly on the browser/DOM.

In case we are not comfortable with the way the positioning is being handled, here are more ways to consider for the problem:
1. Calculate the left/top/right/bottom values according to the containerRect.
1. Use d3 function to calculate the positing of SVG within the viewport (I'm not a fan, but thought I'll mention it)
https://stackoverflow.com/questions/56296571/how-do-i-get-an-internal-svg-elements-position-relative-to-the-viewport-of-an-o
1. Wrap the target in an HTML element that has an offsetParent, and calculate positioning according to it -- might be a problem with changing the target.
1. Since we have been recursively looking for a parent with the offsetParent property, it made me think that maybe the current behavior is desired. Hence, we can continue not supporting this type of HTML element, and document it on the Drop children that we would be looking for the child's parent in case it wouldn't have a position.

Disclaimer: We'll need to research and handle other components that don't have offsetParent in addition to SVG, e.g. `li` element and might be more.

Relevant links:
1. https://stackoverflow.com/questions/30158656/positioning-svg-elements-via-getboundingclientrect-in-variable-width-div
1. https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect VS. https://developer.mozilla.org/en-US/docs/Web/API/SVGGraphicsElement/getBBox
1. https://stackoverflow.com/questions/14832176/convert-svg-rect-x-y-to-screen-left-and-top-jquery-svg
1. https://svn.apache.org/repos/asf/commons/sandbox/gsoc/2010/scxml-js/trunk/src/javascript/scxml/cgf/util/svg.js

DropContainer.js
storybook
storybook
Relevant for #4599


src/js/components/Drop/DropContainer.js
```diff
@@ -81,8 +82,21 @@ const DropContainer = forwardRef(
             container.style.maxHeight = '';
           }
           // get bounds
-          const targetRect = findVisibleParent(target).getBoundingClientRect();
```
ericsoderberghp, line 84: I'm wondering why we even need `findVisibleParent()` here. I would think that we could just call `target.getBoundingClientRect()`. Is there a reason that won't work?

ShimiSun: `findVisibleParent()` is a recursive function that will look for the parent's element with a valid offsetParent property incase the element itself doesn't have offsetParent. 
SVG is one example of an element not having an offsetParent, but there can be more reasons for that that we haven't researched yet, so it is too soon to remove the function completely.
I agree that if we'll find a valid solution that will be generic for all elements without offsetParent, this recursive lookup will no longer be required.

ericsoderberghp: I just changed it to: `export const findVisibleParent = element => element;` and ran through the stories in storybook and ran the unit tests and didn't notice anything amiss. I'm not sure why we need this function.

ShimiSun: Do we have tests that check elements such as `li` or `svg`? You should see the difference using those as containers, and probably more elements that we need to research.

ericsoderberghp: If the DOM element hasn't loaded, I don't think we should show the Drop.

If other elements are different, we should add stories.

ShimiSun: Your suggestion made me realize how much of the complexity could be removed by not accounting for offsetParent. So I cleaned-up the PR -- Thank you. 
`findVisibleParent` is still used within the Layer component so I'll keep it for now, and if relevant, we could remove the usage of it there as well after verifying that it has the correct test coverage.

ericsoderberghp (approved):

PR #4711, opened 2020-11-05
Checkboxgroup stories changed to new structure

leossantiago:
#### What does this PR do?

Organized storybook stories for CheckBoxGroup.

- **Note:** `stories/typescript/Typescript.tsx` was an improved version of `stories/ObjectOptions.js`, with two groups instead of one and with states. Since both of them cover the same example (using objects as options) and the .tsx one doesn't seem to use any typescript feature, I don't see any need to have both. I moved the code to the .js file and added some logs, and deleted the typescript folder. Let me know if I should leave it! :smile:

`components/CheckBoxGroup/stories/CheckBoxGroup.stories.js`

Storybook

Storybook using PR's branch.

#4651

image

No.

Backwards compatible.


ShimiSun (approved):

ShimiSun:
Thank you buddy!! ❤️

PR #4723, opened 2020-11-06
Updated dependencies

### LINKED ISSUE #4574: styled-components upgrade to 5.2 breaks jest
### Expected Behavior
The upgrade to styled-components 5.2.0 should pass all tests.
Currently we are locked on version 5.1.1, trying to upgrade to 5.2 while it causes failures. 
In the 5.2 [release notes](https://github.com/styled-components/styled-components/releases/tag/v5.2.0), it says, "We no longer emit dynamic classNames for empty rulesets, so some className churn may occur in snapshots". Not sure if this is related to the issue, I kept looking...

### Actual Behavior
the upgrade breaks the tests.

Related PR #4531 

### Steps to Reproduce

1. Go to package.json
2. Change the current locked version of styled-components to a caret version that will deploy the latest styled-components
3. Install the new version
4. Run tests `yarn test` and see the failures.

### Your Environment
Any.

- Grommet version: 2.15.0
- Operating System and version (desktop or mobile): Any

ShimiSun:
#### What does this PR do?
* Removed lock from styled-components version
* Upgrade icons version so we can use new icons on the stories

package.json
jest+chromatic
There are many snapshot changes due to styled-components upgrade, but chromatic shows only 3 changes that have no visual impact, so I think it should be good.
Closes #4574 
No
No
backwards compatible

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,7 +129,7 @@
     "regenerator-runtime": "^0.13.1",
     "require-reload": "^0.2.2",
     "simple-git": "^1.95.0",
-    "styled-components": "^5.1.1",
+    "styled-components": "^5.2.1",
     "tarball-extract": "^0.0.6",
     "typescript": "^3.6.4",
     "webpack": "^4.39.3",
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "styled-components": ">= 5.1"
   },
   "dependencies": {
-    "grommet-icons": "^4.2.0",
+    "grommet-icons": "^4.5.0",
     "hoist-non-react-statics": "^3.2.0",
     "markdown-to-jsx": "^6.11.4",
     "polished": "^3.4.1",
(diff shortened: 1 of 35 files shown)


package.json
```diff
@@ -129,7 +129,7 @@
     "regenerator-runtime": "^0.13.1",
     "require-reload": "^0.2.2",
     "simple-git": "^1.95.0",
-    "styled-components": "5.1.1",
+    "styled-components": "^5.1.1",
```
MikeKingdom, line 132: Thanks @ShimiSun . When I tried this the differences were mostly just classnames, but there were some tests that failed outright when they couldn't find elements they were looking for. Did you run into this or find a solution?

ShimiSun: Worked like a charm. which version exactly did you try using?
I just pushed a commit that uses 5.2.1. the latest directly to make sure the errors don't appear and don't break anything.

MikeKingdom: Awesome. I ended up with 5.2.0 at the time

ericsoderberghp (approved):

PR #4724, opened 2020-11-06
Added Tip to Button

ShimiSun:
#### What does this PR do?
A POC for adding Tip to Button.

Few thoughts that came from this quick PR:
* 90% of my use cases were to use the Tip on the right side of the target, so it's not too late to change the tip's default behavior from opening on the bottom, just a thought 😉 
* Let's say the user would like to customize the Tip behavior on the button, for example changing the alignment (like described on the first bullet). So, should we allow `tipProps={{dropProps: {align: {top: 'bottom'}} }}` as well as the `tip` prop? I wish we could combine those two together, but it might not be consistent with how we handle similar cases in other components.
* I'm thinking maybe the Button's tip should have its own themeing, rather than using the default tip themeing, but I'm OK if it's too soon in the game, and we can wait for a need to arise.


ericsoderberghp:
This is a great place to start, the 80% case. I like `tip: string` as the main focus.

What about `<Button tip={{ dropProps: { align: { top: 'bottom' } }, content: "shimi" }} />` as a solution for customizing the Tip props? Although, I think just falling back on `<Tip dropProps={{ align: { top: 'bottom' } }} content="shimi"><Button /></Tip>` is a reasonable place to start.

I'm wondering if there's a way to have a more robust automatic alignment capability for Tip. For instance, align to the right if the target is near the left side of the screen, align to the bottom of the target is near the top of the screen.

I agree that it might be nice to theme the Button Tip separately at some point. But, I think we can add that later, once a concrete need emerges.

ericsoderberghp (approved):

ShimiSun:
#4599

PR #4735, opened 2020-11-10
Changed LayerContainer to use trapFocus

### LINKED ISSUE #4721: When Layer component is set up as a modal, it still allows keyboard focus to go behind the modal
When Layer component is set up as a modal, it still allows keyboard focus to go behind the modal.

### Expected Behavior

I would expect that when a modal is open, keyboard navigation should not leave the modal (except Esc which should close the modal). This is supported by the description of the `modal` attribute in the doc: "Whether there should be an overlay preventing interaction underneath the layer."
I'd expect that continuing to Tab through the modal should loop through the focusable elements inside it.

### Actual Behavior

When a modal is open, tabbing allows you to leave the modal and select elements behind it.

https://v2.grommet.io/layer

### Steps to Reproduce

1. Create a Layer component with modal set to true (which is the default)
2. Open the modal (with keyboard or mouse, doesn't matter)
3. First Tab gets you in to the modal, but when you keep tabbing and run out of focusable elements inside the modal it continues to focus on elements outside of the modal instead of looping back through the modal.

### Your Environment

I was testing the modal right in the example here https://v2.grommet.io/layer
But another user has confirmed they have the same issue in their app built with Grommet.

ShimiSun:
#### What does this PR do?
Changed LayerContainer to use trapFocus by default, so the background of the modal won't trap the focus.

LayerContainer
Storybook https://storybook.grommet.io/?path=/story/layout-layer--corner-layer

Storybook https://storybook.grommet.io/?path=/story/layout-layer--corner-layer
The root-cause was https://github.com/grommet/grommet/pull/4378
Closes #4721 
No
Yes
b/c

### DIFF
--- a/src/js/components/Layer/LayerContainer.js
+++ b/src/js/components/Layer/LayerContainer.js
@@ -156,7 +156,11 @@ const LayerContainer = forwardRef(
     }
     if (modal) {
       content = (
-        <FocusedContainer hidden={position === 'hidden'} restrictScroll>
+        <FocusedContainer
+          hidden={position === 'hidden'}
+          restrictScroll
+          trapFocus
+        >
           {content}
         </FocusedContainer>
       );


ericsoderberghp:
Is this the right behavior when Layer is used for something like a notification message? For example: https://storybook.grommet.io/?path=/story/layout-layer--notification-layer

ShimiSun:
> Is this the right behavior when Layer is used for something like a notification message? For example: https://storybook.grommet.io/?path=/story/layout-layer--notification-layer

The principle and the guiding thought that I was following, is that keyboard navigation should be feasible for clickable elements, and when a `modal` Layer is opened, it doesn't allow the interactive elements on the backdrop to be clickable, so it would make sense to not allow keyboard navigation to those interactive elements and to `trapFocus` on the `modal` Layer.
As for non-modal Layers, like the example you shared, when the Layer opens, interactive elements outside of the Layer are still clickable by the mouse-user so there is no reason that the keyboard-user wouldn't be able to access those interactive elements as well, hence I don't think the focus should be trapped on the Layer for this case, and the current example works as expected.

ericsoderberghp:
> > Is this the right behavior when Layer is used for something like a notification message? For example: https://storybook.grommet.io/?path=/story/layout-layer--notification-layer
> 
> The principle and the guiding thought that I was following, is that keyboard navigation should be feasible for clickable elements, and when a `modal` Layer is opened, it doesn't allow the interactive elements on the backdrop to be clickable, so it would make sense to not allow keyboard navigation to those interactive elements and to `trapFocus` on the `modal` Layer.
> As for non-modal Layers, like the example you shared, when the Layer opens, interactive elements outside of the Layer are still clickable by the mouse-user so there is no reason that the keyboard-user wouldn't be able to access those interactive elements as well, hence I don't think the focus should be trapped on the Layer for this case, and the current example works as expected.

oops, I missed the `if (modal)` above. Makes sense.

PR #4725, opened 2020-11-07
Add new theme sizes to Analog Clock

leossantiago:
#### What does this PR do?

The first part of splitting PR #4656 

- Add sizes `xsmall` and `xxlarge` to Analog Clock, to match most components of Grommet that have `size` prop. As suggested by @ericsoderberghp , `huge` size was kept for backwards compatibility purposes but it's equal to `xxlarge`. 
- Update docs.
- Update tests related to `size` to cover all sizes of Analog and Digital Clocks.
- Update typing.
- Update `Clock/Analog` story to showcase all sizes.
- Update snapshots.
- Add some notes in docs/story about `huge` being equal to `xxlarge` but being kept for backwards compatibility.

#### Where should the reviewer start?

`src/js/themes/base.js`

#### What testing has been done on this PR?

Storybook and jest tests.

#### How should this be manually tested?

Storybook

#### Do the grommet docs need to be updated?

Already did.

#### Is this change backwards compatible or is it a breaking change?

Backwards compatible.

PR #4745, opened 2020-11-13
Fix Digital Clock breaking with undefined/incomplete size prop

leossantiago:
#### What does this PR do?

The second part of splitting PR #4656

- Fix component breaking when providing undefined or incomplete object size to digital clock. Now, if the size passed as prop doesn't exist, `medium` size will be used. Also, considering that for Digital Clock size is defined as `{ size: "VALUEpx", height: value}`, if any of these props is missing, `medium` size/height will be used instead.

- Add unit test for custom size Digital Clock.

- Add Digital Clock styling properties to docs.

- Update Digital story to showcase all sizes.

`src/js/components/Clock/StyledClock.js`

Storybook and unit testing.

```js
const override = {
  clock: {
    digital: {
      text: {
        customSize: {
          size: '30px',
          height: 1.234,
        },
        customMissingSize: {
          height: 1.234,
        },
        customMissingHeight: {
          size: '30px',
        }
      },
    },
  },
};

const theme = deepMerge(grommet, override);

export const Digital = () => (
  <Grommet theme={theme}>
        <Clock type="digital" />
        <Clock type="digital" size="customSize" />
        <Clock type="digital" size="customMissingSize" />
        <Clock type="digital" size="customMissingHeight" />
  </Grommet>
);
```

Already did.

Backwards compatible.

### DIFF
--- a/src/js/components/Clock/doc.js
+++ b/src/js/components/Clock/doc.js
@@ -166,4 +166,64 @@ export const themeDoc = {
     type: 'string',
     defaultValue: '288px',
   },
+  'clock.digital.text.xsmall.size': {
+    description: 'Defines the font size of the Digital Clock',
+    type: 'string',
+    defaultValue: '10px',
+  },
+  'clock.digital.text.xsmall.height': {
+    description: 'Defines the line height of the Digital Clock',
+    type: 'string',
+    defaultValue: '1.5',
+  },
+  'clock.digital.text.small.size': {
+    description: 'Defines the font size of the Digital Clock',
+    type: 'string',
+    defaultValue: '14px',
+  },
+  'clock.digital.text.small.height': {
+    description: 'Defines the line height of the Digital Clock',
+    type: 'string',
+    defaultValue: '1.43',
+  },
+  'clock.digital.text.medium.size': {
+    description: 'Defines the font size of the Digital Clock',
+    type: 'string',
+    defaultValue: '18px',
+  },
+  'clock.digital.text.medium.height': {
+    description: 'Defines the line height of the Digital Clock',
+    type: 'string',
+    defaultValue: '1.375',
+  },
+  'clock.digital.text.large.size': {
+    description: 'Defines the font size of the Digital Clock',
+    type: 'string',
+    defaultValue: '22px',
+  },
+  'clock.digital.text.large.height': {
+    description: 'Defines the line height of the Digital Clock',
+    type: 'string',
+    defaultValue: '1.167',
+  },
+  'clock.digital.text.xlarge.size': {
+    description: 'Defines the font size of the Digital Clock',
+    type: 'string',
+    defaultValue: '26px',
+  },
+  'clock.digital.text.xlarge.height': {
+    description: 'Defines the line height of the Digital Clock',
+    type: 'string',
+    defaultValue: '1.1875',
+  },
+  'clock.digital.text.xxlarge.size': {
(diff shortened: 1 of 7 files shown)


src/js/components/Clock/doc.js
```diff
@@ -166,4 +166,64 @@ export const themeDoc = {
...
   },
+  'clock.digital.text.xsmall.size': {
+    description: 'Defines the font size of the Digital Clock',
+    type: 'string',
+    defaultValue: '10px',
+  },
+  'clock.digital.text.xsmall.height': {
+    description: 'Defines the line height of the Digital Clock',
+    type: 'string',
+    defaultValue: '1.5',
```
ShimiSun, line 177: I think a more accurate definition for height is the type of 'number' rather than 'string'.
An example of that will be the test you created for the use case of customSize with using height as a number and not as a string.
(This comment is relevant for all height values on different sizes)

src/js/components/Clock/doc.js
```diff
@@ -166,4 +166,64 @@ export const themeDoc = {
...
+    description: 'Defines the line height of the Digital Clock',
+    type: 'string',
+    defaultValue: '1.1875',
+  },
+  'clock.digital.text.xxlarge.size': {
+    description: 'Defines the font size of the Digital Clock',
+    type: 'string',
+    defaultValue: '34px',
+  },
+  'clock.digital.text.xxlarge.height': {
```
ShimiSun, line 224: Thank you for adding all those sizes - highly useful for users of digital Clock! ❤️

ShimiSun (commented):
Looks great-- Thank you! 
I dropped a small comment regarding the docs, but overall looks good.

Can you also fix the allowance and usage of custom sizes and sizes in general, on the base.d.ts file to support the typescript users?

src/js/components/Clock/doc.js
leossantiago: Sure! Didn't notice that. Will fix them all :smile:

leossantiago:
Sizes were already defined in Clock interface prior to this PR.

https://github.com/grommet/grommet/blob/25222421d622b6b7f0923bb9c3e79b9bf12fb1e2/src/js/components/Clock/index.d.ts#L18-L26

(`huge` is related to Analog Clock, and was added in #4725).

Should we add something else to its interface?

ShimiSun:
> Sizes were already defined in Clock interface prior to this PR.
> 
> https://github.com/grommet/grommet/blob/25222421d622b6b7f0923bb9c3e79b9bf12fb1e2/src/js/components/Clock/index.d.ts#L18-L26
> 
> (`huge` is related to Analog Clock, and was added in #4725).
> 
> Should we add something else to its interface?

I'm OK if we'll mention on the documentation that `huge` only applies to Analog.

leossantiago:
> I'm OK if we'll mention on the documentation that `huge` only applies to Analog.

https://github.com/grommet/grommet/blob/25222421d622b6b7f0923bb9c3e79b9bf12fb1e2/src/js/components/Clock/doc.js#L48-L52

Already did that on #4725 , which is already merged. Is there any way besides `description` we should use to enforce that `huge` works only for Analog Clock?

ShimiSun (approved):
Thank you! I think this PR is good to go...

leossantiago:
Okay! Let me know if anything can be improved before merge! :heart:

ericsoderberghp (approved):

PR #4755, opened 2020-11-16
Add event listener for scroll on layer

taysea:
#### What does this PR do?
Adds an event listener for scroll so that when Layer has a target it will scroll with the target. `true` is needed as a third parameter of `addEventListener` otherwise scroll event was not being registered.

#### Where should the reviewer start?
src/js/components/Layer/LayerContainer.js

#### What testing has been done on this PR?
Locally in storybook.

#### How should this be manually tested?
Apply this code locally to the Layer target story:
```
<Paragraph>
Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus blandit ut lectus at laoreet. Curabitur eu erat nec metus aliquet eleifend. Aliquam lectus eros, tincidunt sit amet condimentum at, sodales eleifend ipsum. Donec nibh erat, euismod sit amet nulla et, elementum finibus metus. Proin non justo et nunc viverra facilisis pulvinar vitae tortor. Curabitur fermentum a ipsum vel aliquet. Aenean aliquet sodales elit, ac mollis sem rutrum et.
</Paragraph>
<Box height="large" pad="large" background="background-contrast">
   <Button label="Open Layer" onClick={() => setOpen(true)} />
</Box>
{open && (
  <Layer position="right" full="vertical">
      <Box>
        Hello there
      </Box>
   </Layer>
)}
<Paragraph>
Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus blandit ut lectus at laoreet. Curabitur eu erat nec metus aliquet eleifend. Aliquam lectus eros, tincidunt sit amet condimentum at, sodales eleifend ipsum. Donec nibh erat, euismod sit amet nulla et, elementum finibus metus. Proin non justo et nunc viverra facilisis pulvinar vitae tortor. Curabitur fermentum a ipsum vel aliquet. Aenean aliquet sodales elit, ac mollis sem rutrum et.
</Paragraph>
<Paragraph>
Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus blandit ut lectus at laoreet. Curabitur eu erat nec metus aliquet eleifend. Aliquam lectus eros, tincidunt sit amet condimentum at, sodales eleifend ipsum. Donec nibh erat, euismod sit amet nulla et, elementum finibus metus. Proin non justo et nunc viverra facilisis pulvinar vitae tortor. Curabitur fermentum a ipsum vel aliquet. Aenean aliquet sodales elit, ac mollis sem rutrum et.
</Paragraph>
[...]

PR #4761, opened 2020-11-18
Dont restrict scroll when layerTarget present

taysea:
#### What does this PR do?
Does not restrict page scroll when a layerTarget is present. A previously merged PR monitored for scroll event and that was working for the tested implementation. However, if a `Grommet` component is provided with `height="auto"` (to properly calculate the height of the entire page), suddenly the page still would not scroll. In order to overcome this, we need to adjust the `restrictScroll` prop, that is used internally for Layer in modal cases, to only be true if a Layer does not have a target.

Tested locally with a long page with Layer with a target on a Box contained in the page. Grommet component of the example was given `style={{height: 'auto'}}`. 

Same as above.

Modal layer + target was restricting scroll on pages with height="auto".

https://github.com/grommet/hpe-design-system/issues/1115

No.

No.

Backwards compatible.

### DIFF
--- a/src/js/components/Layer/LayerContainer.js
+++ b/src/js/components/Layer/LayerContainer.js
@@ -163,7 +163,7 @@ const LayerContainer = forwardRef(
       content = (
         <FocusedContainer
           hidden={position === 'hidden'}
-          restrictScroll
+          restrictScroll={!layerTarget ? true : undefined}
           trapFocus
         >
           {content}
--- a/src/js/components/Layer/LayerContainer.js
+++ b/src/js/components/Layer/LayerContainer.js
@@ -163,6 +163,9 @@ const LayerContainer = forwardRef(
       content = (
         <FocusedContainer
           hidden={position === 'hidden'}
+          // if layer has a target, do not restrict scroll.
+          // restricting scroll  could inhibit the user's
+          // ability to scroll the page while the layer is open.
           restrictScroll={!layerTarget ? true : undefined}
           trapFocus
         >


src/js/components/Layer/LayerContainer.js
```diff
@@ -163,7 +163,7 @@ const LayerContainer = forwardRef(
       content = (
         <FocusedContainer
           hidden={position === 'hidden'}
-          restrictScroll
+          restrictScroll={!layerTarget ? true : undefined}
```
ericsoderberghp, line 166: Can you add some comments here, just so we don't lose the "why"?

taysea: I've added a comment for future reference.

ericsoderberghp (approved):

ShimiSun:
Can we add any tests to these varied bug fixes of scrolling when using Layer? Could also be relevant to #4755/

taysea:
> Can we add any tests to these varied bug fixes of scrolling when using Layer? Could also be relevant to #4755/

I am not sure that we can since capturing scroll behavior isn't reliable in jest. it would be more suited to test in an e2e environment where the tests are running in a browser environment.

ShimiSun:
@taysea undestood. Can we create a test on the DS site that ensures the correct behavior?

taysea:
> @taysea understood. Can we create a test on the DS site that ensures the correct behavior?

Yes I will make sure to add one in once I implement the work on the site that leverages this work.

PR #4764, opened 2020-11-18
Enhance CheckBoxGroup theming

taysea:
#### What does this PR do?
Adds `checkBoxGroup` to theme and supports `checkBoxGroup.container` which accepts all Box props (aligned with `radioButtonGroup.container`. This allows users to control gap, margin, etc. of checkboxgroup from theme. Previously `gap` was only available to be controlled via prop.

src/js/components/CheckBoxGroup/CheckBoxGroup.js

Added jest test and tested locally with storybook.

Create custom theme in storybook.

Users could only control the gap between checkboxes in checkboxgroup via prop. This is not ideal from applications that want to consistently apply gap across many instances (or for design system usage).

Related to https://github.com/grommet/hpe-design-system/issues/1246

In this example, a margin has been applied to the checkboxgroup itself to create the desired dimensions within a formfield and gap has been removed:

Screen Shot 2020-11-18 at 8 43 47 AM

Yes, updated here.

Yes. CheckBoxGroup can now be themed using `checkBoxGroup.container` where all Box props are accepted.

Yes, backwards compatible.

### DIFF
--- a/src/js/components/CheckBoxGroup/CheckBoxGroup.js
+++ b/src/js/components/CheckBoxGroup/CheckBoxGroup.js
@@ -1,15 +1,16 @@
 import React, { forwardRef, useContext, useMemo } from 'react';
+import { ThemeContext } from 'styled-components';
 
-import { Box } from '../Box';
 import { CheckBox } from '../CheckBox';
 import { FormContext } from '../Form/FormContext';
+import { StyledCheckBoxGroup } from './StyledCheckBoxGroup';
 
 export const CheckBoxGroup = forwardRef(
   (
     {
       value: valueProp,
       disabled: disabledProp,
-      gap = 'small', // consistent with RadioButtonGroup default
+      gap,
       labelKey,
       valueKey,
       onChange,
@@ -20,6 +21,7 @@ export const CheckBoxGroup = forwardRef(
     ref,
   ) => {
     const formContext = useContext(FormContext);
+    const theme = useContext(ThemeContext) || defaultProps.theme;
 
     // In case option is a string, normalize it to be an object
     const options = useMemo(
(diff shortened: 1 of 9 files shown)


src/js/components/CheckBoxGroup/doc.js
```diff
@@ -53,3 +53,11 @@ export const doc = CheckBoxGroup => {
 
   return DocumentedCheckBoxGroup;
 };
+
+export const themeDoc = {
+  'checkBoxGroup.container': {
+    description: 'Any valid Box props for the RadioButtonGroup container.',
```
halocline, line 59:
```suggestion
    description: 'Any valid Box props for the CheckBoxGroup container.',
```

src/js/themes/base.js
```diff
@@ -614,6 +614,9 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         // extend: undefined,
       },
     },
+    checkBoxGroup: {
+      // container: {}, // any box props
```
halocline, line 618: For consistency, are we using the `// any box props` comment anywhere else? Haven't noticed it as a convention, but do think it is helpful. Should `extend` be documented as well?

```suggestion
      // container: {
        // any box props
        // extend: undefined,
      //},
```

taysea: We used it in the radioButtonGroup.container so I had carried it over from there.

taysea: I think it would be beneficial to document extend.

src/js/components/CheckBoxGroup/README.md
```diff
@@ -87,4 +87,15 @@ string
...
 ```
 div
-```
\ No newline at end of file
+```
+## Theme
+  
+**checkBoxGroup.container**
+
+Any valid Box props for the RadioButtonGroup container. Expects `object`.
```
halocline, line 95: Probably best to re-run the generate-docs command to capture this and the corresponding snapshot.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -60,7 +62,17 @@ export const CheckBoxGroup = forwardRef(
     };
 
     return (
-      <Box ref={ref} gap={gap} {...rest}>
+      <StyledCheckBoxGroup
+        ref={ref}
+        {...theme.checkBoxGroup.container}
+        gap={
+          gap ||
+          (theme.checkBoxGroup.container && theme.checkBoxGroup.container.gap
```
halocline, line 70: What happens if a caller has set `theme.checkBoxGroup` to `undefined`?

taysea: I set `checkBoxGroup` as defined in base.js and I believe `Grommet` component merges any custom theme with `base.js` so it will always be defined.

ericsoderberghp (approved):

PR #4766, opened 2020-11-19
Allow onClickOutside for Layer when modal={false}

taysea:
#### What does this PR do?
Allows `onClickOutside` prop to work on Layer when `modal={false}`. Previously, this only functioned when modal={true}. Referenced the behavior that `DropContainer` was applying as it already provides the desired behavior.

Updated jest test and tested on storybook.

On Layer Target story example, switch modal to "false" using the toggle. Then, open the Layer and click outside of it. Layer should close.

Design use cases for a non-modal Layer where `onClickOutside` still functions created a need for this functionality. For example, a Sidebar layer that opens over screen content that does not prohibit the user from interacting with the rest of the page should still be able to be closed by clicking outside.

Related to https://github.com/grommet/grommet/issues/3771

No.

`onClickOutside` is now supported for Layer with modal={false}.

Backwards compatible.


ericsoderberghp (approved):

src/js/components/Layer/LayerContainer.js
```diff
@@ -11,6 +12,7 @@ import { defaultProps } from '../../default-props';
 import { FocusedContainer } from '../FocusedContainer';
 import { Keyboard } from '../Keyboard';
 import { backgroundIsDark, findVisibleParent } from '../../utils';
+import { PortalContext } from './PortalContext';
```
ShimiSun, line 15: Would it make sense to create a PortalContext util that will be shared between Layer and DropContainer?

src/js/components/Layer/LayerContainer.js
```diff
@@ -46,6 +49,12 @@ const LayerContainer = forwardRef(
     const anchorRef = useRef();
     const containerRef = useRef();
     const layerRef = useRef();
+    const portalContext = useContext(PortalContext) || defaultPortalContext;
```
taysea, line 52: look to simplify, do we need `defaultPortalContext`

src/js/components/Layer/LayerContainer.js
```diff
@@ -77,6 +86,28 @@ const LayerContainer = forwardRef(
     }, [position, ref]);
 
     useEffect(() => {
+      const onClickDocument = event => {
+        // determine which portal id the target is in, if any
+        let clickedPortalId = null;
+        let node = event.target;
+        while (clickedPortalId === null && node !== document) {
```
taysea, line 93: should check for "" on getAttribute: https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute

src/js/components/Layer/LayerContainer.js
```diff
@@ -77,6 +86,28 @@ const LayerContainer = forwardRef(
...
+          // check if user click occurred within the layer
+          const attr = node.getAttribute('data-l-portal-id');
+          if (attr !== null) clickedPortalId = parseInt(attr, 10);
+          node = node.parentNode;
```
taysea, line 97: add comment that we are looping up to the parent to find if it has the `data-l` attr

src/js/components/Layer/LayerContainer.js
```diff
@@ -77,6 +86,28 @@ const LayerContainer = forwardRef(
...
+        }
+        if (
+          clickedPortalId === null ||
```
taysea, line 100: check for "" here because it could return null || ""

src/js/components/Layer/LayerContainer.js
```diff
@@ -77,6 +86,28 @@ const LayerContainer = forwardRef(
...
+          portalContext.indexOf(clickedPortalId) !== -1
+        ) {
+          onClickOutside(event);
```
taysea, line 103: this is the user onClickOutside

src/js/components/Layer/LayerContainer.js
```diff
@@ -114,6 +152,9 @@ const LayerContainer = forwardRef(
         plain={plain}
         responsive={responsive}
         dir={theme.dir}
+        // portalId is used to determine if click occurred inside
+        // or outside of the layer
+        data-l-portal-id={portalId}
```
taysea, line 157: Add tests to check if onClickOutside is being called only once.

src/js/components/Layer/LayerContainer.js
```diff
@@ -77,6 +86,28 @@ const LayerContainer = forwardRef(
...
+          node = node.parentNode;
+        }
+        if (
+          clickedPortalId === null ||
+          portalContext.indexOf(clickedPortalId) !== -1
+        ) {
+          onClickOutside(event);
+        }
+      };
+
```
taysea, line 106: add a comment of why we are checking for `onClickOutside`

src/js/components/Layer/LayerContainer.js
```diff
@@ -46,6 +49,12 @@ const LayerContainer = forwardRef(
     const anchorRef = useRef();
     const containerRef = useRef();
     const layerRef = useRef();
+    const portalContext = useContext(PortalContext) || defaultPortalContext;
+    const portalId = useMemo(() => portalContext.length, [portalContext]);
+    const nextPortalContext = useMemo(() => [...portalContext, portalId], [
```
taysea, line 54: Can we use the same PortalContext and reference both for Drop and Layer.

src/js/components/Layer/LayerContainer.js
```diff
@@ -146,7 +187,21 @@ const LayerContainer = forwardRef(
     }
 
     if (onEsc) {
-      content = <Keyboard onEsc={onEsc}>{content}</Keyboard>;
+      content = (
+        <Keyboard
+          onEsc={
+            onEsc
+              ? event => {
+                  event.stopPropagation();
```
taysea, line 195: Solves the onEsc for modal={false} (Add comment)

src/js/components/Layer/LayerContainer.js
```diff
@@ -159,20 +214,24 @@ const LayerContainer = forwardRef(
...
+        // restricting scroll could inhibit the user's
+        // ability to scroll the page while the layer is open.
+        restrictScroll: !layerTarget ? true : undefined,
+        trapFocus: true,
+      };
+    else containerProps = undefined;
+
+    content = (
+      <PortalContext.Provider value={nextPortalContext}>
+        <FocusedContainer {...containerProps}>{content}</FocusedContainer>
```
taysea, line 232: Change the order so that it behaves as is

src/js/components/Layer/LayerContainer.js
taysea: Done.

src/js/components/Layer/LayerContainer.js
taysea: Done.

taysea:
Converting to draft because I need to think through the use case of when a Drop is inside a Layer and how the `onClickOutside` is being determined. Clicking within a Drop that is open inside a Layer should still be considered "inside" the Layer.

src/js/components/Layer/LayerContainer.js
taysea: Done.

src/js/components/Layer/LayerContainer.js
taysea: Added comment.

src/js/components/Layer/LayerContainer.js
taysea: Not applicable. the "" is only potentially returned on getAttribute.

src/js/components/Layer/LayerContainer.js
taysea: Done.

src/js/components/Layer/LayerContainer.js
taysea: They can share. Docs here: https://reactjs.org/docs/context.html#contextprovider

"Providers can be nested to override values deeper within the tree.

All consumers that are descendants of a Provider will re-render whenever the Provider’s value prop changes."

src/js/components/Layer/LayerContainer.js
taysea: Done.

src/js/components/Layer/LayerContainer.js
```diff
@@ -77,6 +89,35 @@ const LayerContainer = forwardRef(
...
+            portalContext.indexOf(clickedPortalId) !== -1) &&
+          node !== null
+        ) {
+          // if the click occurred outside of the Layer portal, call
+          // the user's onClickOutside function
+          onClickOutside(event);
+        }
+      };
+
+      // if user provides an onClickOutside function, list for mousedown event
```
taysea, line 116:
```suggestion
      // if user provides an onClickOutside function, listen for mousedown event
```

ShimiSun (approved):
Thank you 💟

PR #4777, opened 2020-11-24
Added Number component

ericsoderberghp:
#### What does this PR do?

THIS IS A DRAFT FOR REVIEW - DO NOT MERGE YET

Adds a new component for a numeric value with optional units.

Some things to consider still:
- do we even need this?
- would it be better to add more size options for Text?
- are their accessibility aspects we should include?
- should a label be included as well, probably below the number + units?

stories

stories
unit test

stories

yes

yes

backwards compatible

### DIFF
--- a/src/js/components/Number/Number.js
+++ b/src/js/components/Number/Number.js
@@ -1,0 +1,66 @@
+import React, { useContext } from 'react';
+import { ThemeContext } from 'styled-components';
+import { Box } from '../Box';
+import { Text } from '../Text';
+
+const Number = ({
+  color,
+  gap = 'xsmall', // theme someday
+  size = 'medium',
+  units,
+  value,
+  weight,
+  ...rest
+}) => {
+  const theme = useContext(ThemeContext) || defaultProps.theme;
+  const { size: themeSize, units: themeUnits, ...valueProps } = {
+    ...theme.number,
+  };
+  valueProps.size = themeSize[size] || size;
+  if (color) valueProps.color = color;
+  if (weight) valueProps.weight = weight;
+  let content = (
+    <Text {...valueProps} {...rest}>
+      {value}
+    </Text>
+  );
+
+  if (units) {
+    const { size: themeUnitsSize = {}, ...themeUnitsRest } = theme.number.units;
+    let unitsProps;
+    let unitsLabel;
+    if (typeof units === 'object') {
+      const { label, size: unitsSize, ...unitsRest } = units;
+      unitsLabel = label;
+      unitsProps = { ...themeUnitsRest, ...unitsRest, size: unitsSize };
+      if (!unitsProps.size)
+        unitsProps.size = themeUnitsSize[size] || valueProps.size;
+      if (!unitsProps.color) unitsProps.color = valueProps.color;
+      if (!unitsProps.weight) unitsProps.weight = valueProps.weight;
+    } else {
+      unitsLabel = units;
+      unitsProps = themeUnitsRest;
+      unitsProps.size = themeUnitsSize[size];
+      if (!unitsProps.size) unitsProps.size = valueProps.size;
+      if (!unitsProps.color) unitsProps.color = valueProps.color;
+      if (!unitsProps.weight) unitsProps.weight = valueProps.weight;
+    }
+    content = (
+      <Box direction="row" gap={gap} align="baseline">
+        {content}
+        <Text {...unitsProps}>{unitsLabel}</Text>
+      </Box>
+    );
+  }
+
+  return content;
+};
+
+let NumberDoc;
+if (process.env.NODE_ENV !== 'production') {
+  // eslint-disable-next-line global-require
+  NumberDoc = require('./doc').doc(Number);
+}
+const NumberWrapper = NumberDoc || Number;
+
+export { NumberWrapper as Number };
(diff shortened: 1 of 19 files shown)


src/js/themes/base.d.ts
```diff
@@ -943,6 +943,58 @@ export interface ThemeType {
     colors?: GraphColorsType;
     extend?: ExtendType;
   };
+  number?: {
```
ShimiSun, line 946: I'm wondering if we can shorten the all number type of definition by leveraging TextProps.

src/js/themes/base.js
```diff
@@ -940,6 +940,19 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       // colors: [] || colors: ['graph-0', 'graph-1', 'graph-2', 'graph-3'],
       // extend: undefined,
     },
+    number: {
+      size: {
+        small: { ...fontSizing(2) },
+        medium: { ...fontSizing(4) },
+        large: { ...fontSizing(8) },
+        xlarge: { ...fontSizing(16) },
```
ShimiSun, line 948: Why not defaulting to regular text sizing?

src/js/themes/base.js
```diff
@@ -940,6 +940,19 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
+      },
+      weight: 'bold',
```
ShimiSun, line 950: A little opinionated to me, do we have designs that recommend a bold weight?

src/js/components/Number/Number.js
```diff
@@ -0,0 +1,66 @@
...
+  weight,
+  ...rest
+}) => {
+  const theme = useContext(ThemeContext) || defaultProps.theme;
+  const { size: themeSize, units: themeUnits, ...valueProps } = {
+    ...theme.number,
+  };
+  valueProps.size = themeSize[size] || size;
+  if (color) valueProps.color = color;
+  if (weight) valueProps.weight = weight;
```
ShimiSun, line 21: Are you ok with potentially overriding an existing theme value?

src/js/components/Number/Number.js
```diff
@@ -0,0 +1,66 @@
...
+    } else {
+      unitsLabel = units;
+      unitsProps = themeUnitsRest;
+      unitsProps.size = themeUnitsSize[size];
+      if (!unitsProps.size) unitsProps.size = valueProps.size;
+      if (!unitsProps.color) unitsProps.color = valueProps.color;
+      if (!unitsProps.weight) unitsProps.weight = valueProps.weight;
+    }
+    content = (
+      <Box direction="row" gap={gap} align="baseline">
```
ShimiSun, line 49: wondering if `direction="row"` should be more flexible so it will work better on small windows using `column` or `row-responsive` instead of `row`.

src/js/components/Number/doc.js
```diff
@@ -0,0 +1,112 @@
...
+
+  DocumentedNumber.propTypes = {
+    color: colorPropType.description(
+      'A color identifier to use for the text color.',
+    ),
+    size: PropTypes.oneOfType([
+      PropTypes.oneOf(['small', 'medium', 'large', 'xlarge']),
+      PropTypes.string,
+    ])
+      .description('The size of the text.')
```
ShimiSun, line 23: I think we want to be a little more accurate and say it is for the 'value text' instead of just 'text'. Or in cases, it affects both value-text and unit-text, we should say something like 'the Number text'

src/js/components/Number/doc.js
```diff
@@ -0,0 +1,112 @@
...
+          'A color identifier to use for the units.',
+        ),
+        label: PropTypes.string.description('The label for the units.'),
+        size: PropTypes.oneOfType([
+          PropTypes.oneOf(['small', 'medium', 'large', 'xlarge']),
+          PropTypes.string,
+        ]).description('The size of the units.'),
+        weight: PropTypes.oneOfType([
+          PropTypes.oneOf(['normal', 'bold']),
+          PropTypes.number,
```
ShimiSun, line 38: Could also be string

src/js/components/Number/doc.js
```diff
@@ -0,0 +1,112 @@
...
+        ]).description('The font weight of the units'),
+      }),
+    ]).description('Whether and how any units should be shown.'),
+    weight: PropTypes.oneOfType([
+      PropTypes.oneOf(['normal', 'bold']),
+      PropTypes.number,
```
ShimiSun, line 44: Could also be string

ShimiSun (commented):
> do we even need this?

On one hand, I think "why not", but I feel like I need to see more design requirements to make sure we are providing the grommet-ish solution.


>would it be better to add more size options for Text?

Instead of the Number component?


> are their accessibility aspects we should include?

I think that as long as we allow passing `a11yTitle` to Text we should be fine.


> should a label be included as well, probably below the number + units?

It will be nice to attach some design requirements.

ShimiSun:
One more comment regarding accessibility that we can use for reference,
https://www.airbnb.com/rooms/950167?adults=2&children=1&check_in=2020-12-12&check_out=2021-01-09&display_extensions%5B0%5D=MONTHLY_STAYS&display_extensions%5B1%5D=DATED_LTS&source_impression_id=p3_1606254976_WBw%2FfS%2B6h%2Fx7O4sA&guests=1

(Try to ignore the views and location :) )

The Number value/units are not accessible by keyboard navigation, so I think we approached it correctly by not setting it a focus
image

nomadus:
@ericsoderberghp Actually, I was hoping that this will be a number input component. I wonder why Grommet still misses the number input component?

ShimiSun:
Closing this PR, feel free to reopen if still relevant.

Issue #2961, opened 2019-03-21
Select multiple with search breaks


alansouzati:
### Expected Behavior
Select multiple with search should allow me to filter the list of options and keep selecting and unselecting things

### Actual Behavior
When using Select multiple and I perform a search, select breaks.

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/s/2pop0013q0

### Steps to Reproduce

1. Open sandbox link
2. Open Select and choose option one
3. Open select again and this search search "two"
4. Select "two"
5. open select one last time.
6. Notice that only two is selected where one and two should be selected.

### Your Environment

- Grommet version: v2 latest
- Browser Name and version: any
- Operating System and version (desktop or mobile): any

oorestisime:
There is a bigger issue here i think. The selected indexes are wrong when there is a search involved. It reflects the index relative to the search and not the absolute index.

oorestisime:
And `options.indexOf(v)` for each value can return -1 if the value  is not in the search container.

ShimiSun:
Yea, I noticed the same type of behavior and hence labeled the issue as `arch` (i.e architectural problem). 
whether you had previous values selected before the search, the values won't be available on the `options` array due to filtered options resulted from the search, hence you'll be losing their indexes. An idea I had was that we need a helper method that will be called after the search and before `selectOption` that will send the original values of indexes and options.

oorestisime:
Another solution would be to store initial options in the state something like:

```
constructor(props) {
    super(props);
    this.state = {
      initialOptions: props.options,
    };
  }
```

and then calculate the selected with regards to that

romanalesenkov1:
I can reproduce this issue now here: 
https://codesandbox.io/s/github/grommet/grommet-sandbox?initialpath=/select&module=/src/Select.js&file=/src/Select.js

ShimiSun:
@romanalesenkov1 , thanks for reporting, reopening the issue...

ShimiSun:
The fix is available @grommet version 2.16.0. Thank you @romanalesenkov1 for the fix.

PR #4778, opened 2020-11-24
fix selectMultiple + search

romanalesenkov1:
#### What does this PR do?
https://github.com/grommet/grommet/issues/2961
Previously it was impossible to select multiple values using search. Fix this case.

The problem was that options change when they are filtered using search. So indexes of selected values are invalid. Now we always use the initial list of options, so the indexes will always be valid.

tested selecting multiple values using search and without using search
```
       <Select
          size="medium"
          placeholder="Select"
          multiple
          value={value}
          options={options}
          closeOnChange={false}
          onChange={({ value: nextValue }) => setValue(nextValue)}
          onSearch={text => {
            setOptions(defaultOptions.filter(o => exp.test(o)));
          }}
        />
```

e.g. storybook

- simple selection
- searching and then select stores correct values
- multiple select with/without searching
- selecting already selected in multiple select removes the selection
https://github.com/grommet/grommet/issues/2961

No
No

### DIFF
--- a/src/js/components/Select/Select.js
+++ b/src/js/components/Select/Select.js
@@ -101,10 +101,12 @@ const Select = forwardRef(
         );
       return valueKey && valueKey.reduce ? value : applyKey(value, valueKey);
     }, [value, valueKey]);
+
+    const [initialOptions] = useState(options);
     // the option indexes present in the value
     const optionIndexesInValue = useMemo(() => {
       const result = [];
-      options.forEach((option, index) => {
+      initialOptions.forEach((option, index) => {
         if (selected !== undefined) {
           if (Array.isArray(selected)) {
             if (selected.indexOf(index) !== -1) result.push(index);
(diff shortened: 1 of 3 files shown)


ShimiSun (approved):
Nicely done, works as expected! 
I added a story example so it will more straightforward to have sanity checks on the serach+multiple scenario in the future, the story demonstrates the behavior as expected. Thank you for your contribution! 💟

ericsoderberghp:
I think something isn't quite right. Consider this sequence:
1. click to open -> drop opens
2. type "2" -> options are pruned
3. click "option 2" -> value shows "option 2", drop is closed
4. click to open -> drop opens, showing "option 2" selected
5. type "3" -> options are pruned, BUT: value shows "option 13"

If I have no value, or if I have multiple options selected already, the search + drop + value all behave well. But if I have one option selected and then perform a search, the visible value is being changed incorrectly.

ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution!

ShimiSun:
Right on time for the release! Thank you @romanalesenkov1!

romanalesenkov1:
Thank you guys!

PR #4782, opened 2020-11-25
Changed Chart to support pattern area fill

ericsoderberghp:
#### What does this PR do?

Changed Chart to support pattern area fill

Chart/doc.js

Added storybook stories
Added unit tests

Storybook

automatic

yes

backwards compatible


src/js/components/Chart/Chart.js
```diff
@@ -218,6 +217,9 @@ const Chart = React.forwardRef(
       ];
     };
 
+    const useGradient = color && Array.isArray(color);
+    const patternId = pattern && `${pattern}-${id}-pattern`;
```
ShimiSun, line 221: Two questions:
1. Should we document that `id` is required when using `pattern`?
2. Shouldn't `patternId` be relevant only when `type` is 'area' ? why is it calculated for all the other use cases?

src/js/components/Chart/Chart.js
```diff
@@ -280,6 +282,7 @@ const Chart = React.forwardRef(
               <title>{label}</title>
               <path
                 d={d}
+                fill={patternId ? `url(#${patternId})` : undefined}
```
ShimiSun, line 285: Interesting, according to docs I was under the impression that `pattern` is only relevant for `area` and not for bars.

src/js/components/Chart/Chart.js
```diff
@@ -511,30 +523,39 @@ const Chart = React.forwardRef(
...
+          x2={0}
+          y2={1}
+        >
+          {color
+            .sort((c1, c2) => c2.value - c1.value)
+            .map(({ value, color: gradientColor }) => (
+              <stop
+                key={value}
+                offset={
+                  // TODO:
```
ShimiSun, line 547: Is it still relevant?

src/js/components/Chart/Chart.js
ericsoderberghp: yes to both, fixed

src/js/components/Chart/Chart.js
ericsoderberghp: yep, fixed

src/js/components/Chart/Chart.js
ericsoderberghp: Hmm... not part of this change.

src/js/components/Chart/Chart.js
```diff
@@ -501,7 +510,11 @@ const Chart = React.forwardRef(
     const drawing = (
       <g
         stroke={stroke}
-        strokeWidth={type !== 'point' ? strokeWidth : undefined}
+        strokeWidth={
+          type !== 'point' && (type !== 'area' || !pattern)
```
ShimiSun, line 514: This change will behave strangely if the user tries out the type of 'bar' with a pattern. I agree this isn't something we should support since it isn't within the API, but since it still may impact the UI behavior, I would consider adding a warning for the usage of `pattern` without the type of 'area'.
Which leads me to another thought I had and wanted to share since `pattern` is only part of 'area', should we consider adding it to the area type as an object? It seems a little loud to have a whole prop for a singular use case of `type='area'`. so something like  `type={area: {pattern: '...'}}` or any other idea that will bind `pattern` directly to `area` and not let `pattern` float as an independent prop. I'm well aware that this idea may not be relevant in case we consider to allow pattern on other types than 'area', but thought I should share.

ericsoderberghp: I think this specific line will still behave correctly if the caller uses "bar" and "pattern", since `type != 'area'`.

I was thinking that we might want to support patterns for bars at some point down the road, so I was thinking to allow API growth in that direction. Also, we already have the `point` prop, which is currently only used when `type="point"`.

ShimiSun (approved):

PR #4785, opened 2020-11-27
chore: Updates index.d.ts with Tip export

apawson:
#### What does this PR do?
Adds missing `Tip` TypeScript export

Tested/confirmed on local dev ENV.

Attempt to `import { Tip } from 'grommet';`

image

No

No

Backwards compatible


ShimiSun (approved):
I agree with this fix. And I think the reason we have not seen it reproducible locally is due to the way we import the Tip component on our tsx stories. We should change the import of Tip to be imported from `grommet` rather than from the component folder, feel free to file a fix for it as well.

Issue #1190, opened 2017-02-15
List onSelect argument type


dioklecijan:
IMHO, for the List component, the argument for the onSelect callback handler should always be an array when multiple rows may be selected. Besides semantic improvement, the code looks more elegant:

`<List selectable="multiple" onSelect={this.onSelect}>`

Instead:

```
  onSelect = (indexes) => {
    if (Array.isArray(indexes))
      indexes.forEach(idx => console.log(this.props.items[idx]));
    else
      console.log(this.props.items[indexes]);
  }
```

we could write:

```
  onSelect = (indexes) => indexes.forEach(idx => console.log(this.props.items[idx]));
```

### Your Environment
* Grommet version: 1.2

tracybarmore:
Hey @dioklecijan, we agree. We just need to work through how to manage and communicate this change since it would mean a breaking change.

karatechops:
Tagging 2.x List (#1472)

PR #4793, opened 2020-12-02
Fix/datatable units

### LINKED ISSUE #4610: Add `unit` to DataTable column structure
### Expected Behavior
Allow user to specify a unit value that would appear next to the column label. User should be able to style this value via the theme and provide a custom render if desired. It is a common case that users may have a unit accompany a column label that should receive separate style treatment than the label itself.

### Actual Behavior
Not currently existing.

Screen Shot 2020-10-14 at 10 12 14 AM

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

MikeKingdom:
#### What does this PR do?
Adds a `units` string to the column definition for DataTables. This string will be displayed in the header with a default text size of `xsmall` and the same text properties used for the column header. It also will use any properties specified under `datatable.header.units` in the theme for units Text component.

DataTable/Header.js

A 'Units' story has been added to the DataTables storybook.
A Jest test case has been added to the DataTables Jest test.

Look at the Units story in the DataTables storybook.

See #4610 and #1190

Fixes #4610 

Done

Sure.

Backwards compatible

### DIFF
--- a/src/js/components/DataTable/Header.js
+++ b/src/js/components/DataTable/Header.js
@@ -173,7 +173,7 @@ const Header = forwardRef(
             }) => {
               let content;
               const unitsContent = units ? (
-                <Text {...textProps} {...theme.dataTable.header.units.label}>
+                <Text {...textProps} {...theme.dataTable.header.units}>
                   {units}
                 </Text>
               ) : (
@@ -233,7 +233,7 @@ const Header = forwardRef(
 
               if (unitsContent) {
                 content = (
-                  <Box {...theme.dataTable.header.units.container}>
+                  <Box align="baseline" direction="row">
                     {content}
                     {unitsContent}
                   </Box>
--- a/src/js/components/DataTable/Header.js
+++ b/src/js/components/DataTable/Header.js
@@ -112,7 +112,6 @@ const Header = ({
   ...rest
 }) => {
   const theme = useContext(ThemeContext) || defaultProps.theme;
-  console.log(theme);
   const [cellProps, layoutProps, textProps] = separateThemeProps(theme);
   const unitsProps = theme.dataTable.header.units;
 
--- a/src/js/components/DataTable/Header.js
+++ b/src/js/components/DataTable/Header.js
@@ -30,6 +30,7 @@ const separateThemeProps = theme => {
     color,
     font,
     gap, // gap is used for space between header cell elements only
+    units: { label: unitsTextProps },
     ...rest
   } = theme.dataTable.header;
 
@@ -38,7 +39,7 @@ const separateThemeProps = theme => {
   const iconProps = { color };
   const layoutProps = { ...rest };
 
-  return [cellProps, layoutProps, textProps, iconProps];
+  return [cellProps, layoutProps, textProps, iconProps, unitsTextProps];
 };
 
 // build up CSS from basic to specific based on the supplied sub-object paths.
@@ -112,8 +113,9 @@ const Header = ({
   ...rest
 }) => {
   const theme = useContext(ThemeContext) || defaultProps.theme;
-  const [cellProps, layoutProps, textProps] = separateThemeProps(theme);
-  const unitsProps = theme.dataTable.header.units;
+  const [cellProps, layoutProps, textProps, , unitsProps] = separateThemeProps(
+    theme,
+  );
 
   let background;
   if (backgroundProp) background = backgroundProp;
@@ -167,7 +169,7 @@ const Header = ({
           }) => {
             let content;
             const unitsContent = units ? (
-              <Text {...textProps} size="xsmall" {...unitsProps}>
+              <Text {...textProps} {...unitsProps}>
                 {units}
               </Text>
             ) : (
@@ -193,7 +195,11 @@ const Header = ({
 
             if (unitsContent) {
               content = (
-                <Box direction="row" align="baseline" gap="xsmall">
+                <Box
+                  direction="row"
+                  align={theme.dataTable.header.units.align}
+                  gap={theme.dataTable.header.units.gap}
+                >
                   {content}
                   {unitsContent}
                 </Box>
--- a/src/js/components/DataTable/Header.js
+++ b/src/js/components/DataTable/Header.js
@@ -30,7 +30,7 @@ const separateThemeProps = theme => {
     color,
     font,
     gap, // gap is used for space between header cell elements only
-    units: { label: unitsTextProps },
+    units,
     ...rest
   } = theme.dataTable.header;
 
@@ -39,7 +39,7 @@ const separateThemeProps = theme => {
   const iconProps = { color };
   const layoutProps = { ...rest };
 
-  return [cellProps, layoutProps, textProps, iconProps, unitsTextProps];
+  return [cellProps, layoutProps, textProps, iconProps];
 };
 
 // build up CSS from basic to specific based on the supplied sub-object paths.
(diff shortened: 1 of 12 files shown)


src/js/components/DataTable/doc.js
```diff
@@ -122,6 +122,7 @@ export const doc = DataTable => {
           ]),
           PropTypes.string,
         ]),
+        units: PropTypes.string,
```
ShimiSun, line 125: This is actually an object that accepts all props of Text component

src/js/components/DataTable/doc.js
```diff
@@ -393,6 +394,12 @@ export const themeDoc = {
     type: 'string | object',
     defaultValue: undefined,
   },
+  'dataTable.header.units': {
+    description: `The styles for the units display in the header. This
```
ShimiSun, line 398:
```suggestion
    description: `Any Text component properties for styling the header's units text. This
```
Or something along the lines that implies the user to use Text props for `units`

src/js/components/DataTable/doc.js
```diff
@@ -393,6 +394,12 @@ export const themeDoc = {
...
+    can be any valid Text properties`,
+    type: 'object',
+    defaultValue: '{ size: "xsmall" }',
```
ShimiSun, line 401: Since you are defining the default properties of the theme, I would have expected to see this default value on base.js and not hardcode on the component file.

src/js/components/DataTable/doc.js
MikeKingdom: This one is actually the string units rather than props. For example, "(TiB)" or "%" etc.

src/js/components/DataTable/Header.js
```diff
@@ -183,6 +191,15 @@ const Header = ({
               }
             } else content = header;
 
+            if (unitsContent) {
+              content = (
+                <Box direction="row" align="baseline" gap="xsmall">
```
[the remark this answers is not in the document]
MikeKingdom: Good point. I'll add the align/gap to the header.units part of the theme. Is there a good place to put defaults other than adding them to the base grommet theme or is that the place?

The reason it's baseline is otherwise it's 'center' which doesn't work well for something with 2 different font sizes like this.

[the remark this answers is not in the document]
ShimiSun: >Is there a good place to put defaults other than adding them to the base grommet theme or is that the place?

Since the styling of the unit is retrieved from the theme I think it will be the best place.

src/js/components/DataTable/doc.js
MikeKingdom: Perfect. Done

ShimiSun (commented):
Needs to address changes of index.d.ts and base.d.ts to support typescript users

src/js/themes/base.d.ts
```diff
@@ -625,6 +625,11 @@ export interface ThemeType {
         background?: BackgroundType;
       };
       pad?: PadType;
+      units?: {
+        gap?: GapType;
+        align?: string;
```
ShimiSun, line 630: Should this be `AlignType`?

src/js/components/DataTable/Header.js
```diff
@@ -183,6 +193,19 @@ const Header = ({
               }
             } else content = header;
 
+            if (unitsContent) {
+              content = (
+                <Box
+                  direction="row"
+                  align={theme.dataTable.header.units.align}
+                  gap={theme.dataTable.header.units.gap}
```
ShimiSun, line 201: I think the three of those should be under `theme.dataTable.header.units.container` and let the container accept any BoxProps. hence will clean up this code to:
```
  <Box {...theme.dataTable.header.units.container}> ... </Box>
```

MikeKingdom: I like it...

src/js/themes/base.d.ts
MikeKingdom: Somewhere else was just doing string but that would be much better

src/js/components/DataTable/Header.js
MikeKingdom: Done

src/js/themes/base.d.ts
MikeKingdom: Done

ShimiSun: Feel free to fix the other place as well.

MikeKingdom: Can we merge this without updating the other usages, especially since I didn't have to specify it here with the container change? I'm actually supposed to be on FTO since Wednesday.

ShimiSun: No rush. we rather have all the relevant fixes in one PR, easier to track.

src/js/components/DataTable/Header.js
```diff
@@ -183,6 +191,15 @@ const Header = ({
               }
             } else content = header;
 
+            if (unitsContent) {
+              content = (
+                <Box {...theme.dataTable.header.units.container}>
```
ericsoderberghp, line 196: Something feels a little odd to have `units.container` driving the Box containing the main header content.

I'm also wondering how this interacts with `columns.align`, since that also has some layout aspects to it.

With the way this PR is structured, the sort icon will be placed between the heading label and the units. I'm wondering if that's the typical design approach. I could see designs wanting the units to be next to the label with the sort icon following the units.

Here are some rough possible designs:
| label units            |
|            label units |
| label            units |
| label units sort    |
|    label units sort |
| label sort    units |

At this point, I'm wondering if the last design should require a custom column.header render function and that the theme should be more focused around just the Text for the units, so the theme could have just a color, size, margin.

ericsoderberghp: I think part of my confusion is around the naming of "units.container" compared to what it is affecting.

Would it work to just do: `<Box align="baseline">{content}<Text {...theme.dataTable.units} /></Box>` when there are units and put a left margin on the Text?

MikeKingdom: That's certainly possible if that works for you (+ direction="row"). The nice part about having all the box props in the theme under something like container is it gives you full control rather than having the hardcoded parts (kinda Shimi's earlier comment.) Although most of the other layout in Header.js is hardcoded.

ericsoderberghp: If we decide to keep it, I think we would need to put it somewhere other than under `units`. For example, if I set `units.container.background = 'brand'`, it will be the background of both the label and the units. Something on the order of `theme.dataTable.labelUnitsContainer` would be more accurate.

However, I tend towards not putting it in the theme at the moment. As you mention, there's a bunch of other styling stuff hardcoded in the Header already. I'd rather bring it all into the theme at the same time, driven by some specific design references.

MikeKingdom: Simplified...Does this look a little better?

ericsoderberghp: yep

src/js/themes/base.js
```diff
@@ -723,6 +723,10 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         //   background: undefined,
         // },
         // pad: undefined,
+        units: {
+          size: 'xsmall',
```
ericsoderberghp, line 727: I think we should default to `color = "text-xweak"` to align more with `theme.formField.info` and not change the size.

ericsoderberghp (commented):
Structure looks great, just one small style suggestion. Sorry this has dragged on long.

src/js/themes/base.js
MikeKingdom: This is the difference with color="text-xweak" and the default header size:
image

ericsoderberghp (approved):

Issue #4743, opened 2020-11-12
Form's onValidate() not returning updated elements when using dynamic add remove functionality


shreyashah7:
We have created a Dynamic Add Remove Component inside a form. Whenever i delete any input field it gets deleted from the UI but in form's onValidate Function, I still see removed input's validation.

### Expected Behavior

onValidate() Function returning {errors, infos} should only return existing elements on Form

### Actual Behavior

But it is still persisting removed elements from form.

### URL, screen shot, or Codepen exhibiting the issue

<img width="402" alt="Screen Shot 2020-11-12 at 10 39 36 AM" src="https://user-images.githubusercontent.com/6551265/98985764-e2b54b00-24d8-11eb-818c-7dde6e1a9b96.png">

<img width="837" alt="Screen Shot 2020-11-12 at 11 14 29 AM" src="https://user-images.githubusercontent.com/6551265/98985769-e3e67800-24d8-11eb-87c3-89a9fcc381f8.png">

![ezgif com-gif-maker](https://user-images.githubusercontent.com/6551265/98986590-f7dea980-24d9-11eb-969e-1303f4c75637.gif)

### Steps to Reproduce

1. create a dynamic add remove component, add elements and then remove elements using splice.
2. Check the form's onValidate() results.

CC: @ShimiSun

ypinneka:
https://codesandbox.io/s/frosty-snowflake-yfndq?file=/src/App.js:1140-1166

Replicating the issue in CodeSandBox with plain grommet.

halocline:
Thank you for the [CodeSandbox](https://codesandbox.io/s/frosty-snowflake-yfndq?file=/src/App.js:1140-1166) @ypinneka. I also [created an implementation](https://codesandbox.io/s/epic-poitras-ebds1?file=/src/App.js) where the inputs are in an array and remove action splices the item from the array to mimic @shreyashah7's original post.

I have been able to replicate the issue with both implementations.

**A couple of initial findings:**
- The validation results returned when triggered by onSubmit are returned as desired (as a `{errors: {}, infos: {}}`)
- The validation results returned when triggered by onBlur do include the removed input
  - The current implementation retains previous errors until each are individually resolved. This is desired if multiple errors were surfaced on an onSubmit event, where each error would need to be resolved individually.
  - We could add in a step to check to only send errors for inputs in the form's context, prior to returning the validation results, however the form's `valueState` (determined by the form's context) still retains the removed input.
[...]

PR #4794, opened 2020-12-02
DRAFT - Do Not Merge - Fix 4743 - Form onValidate and onSubmit return only relevant form input values

halocline:
#### What does this PR do?

Work in progress solution to receive feedback on the solution's approach.

Addresses concern that "onValidate() function should only return {errors, infos} for existing elements on Form." Additionally, ensures only existing inputs on Form are returned in Form's value.

Form.js

Jest + Storybook

Storybook
```
import React from 'react';
import { Box, Button, Grommet, Form, FormField, TextInput } from 'grommet';

const servers = [
  { name: 'Server-1', ipaddr: '10.10.10.10' },
  { name: 'Server-2', ipaddr: '20.20.20.20' },
];

export const Test = () => {
  const [entries, setEntries] = React.useState(servers);
  const [formValues, setFormValues] = React.useState();

  return (
    <Grommet full>
        <Form
          value={formValues}
          validate="blur"
          onChange={nextValue => setFormValues(nextValue)}
          onSubmit={({ value }) => console.log(value)}
          onValidate={obj => console.log('obj = ', obj)}
        >
          {entries &&
            entries.map(entry => {
              return (
                <FormField
                  key={entry.name}
                  htmlFor={entry.name}
                  name={entry.name}
                  label={entry.name}
                  required
                >
                  <TextInput
                    id={entry.name}
                    name={entry.name}
                    value={entry.ipaddr}
                  />
                </FormField>
              );
            })}
          <Button
            type="button"
            label="Add"
            secondary
            onClick={() => {
              setEntries([...entries, { name: 'New server', ipaddr: '' }]);
            }}
          />
          <Button type="submit" label="Submit" primary />
        </Form>
      </Box>
    </Grommet>
  );
};
```

https://github.com/grommet/grommet/issues/4743

Initial findings: https://github.com/grommet/grommet/issues/4743#issuecomment-733289505
Suggested approach: https://github.com/grommet/grommet/issues/4743#issuecomment-733295563

This is a work in progress to receive feedback on the solution's approach: 
- [ ] The updated `formContext.useFormInput` has only been applied to `TextInput` and potentially needs to be applied to other inputs as well.
- [ ] Consideration for how this approach would affect implementations of a progressive or multi-step form or wizard where inputs are shown/hidden. Current implementation may be breaking change in these scenarios. Have not yet tested. May require an "opt-in" mechanism by the caller to make backwards compatible.

https://github.com/grommet/grommet/issues/4743

No.

Yes.

See "background information" comments regarding progressive / multi-step forms.


src/js/components/Form/Form.js
```diff
@@ -148,6 +149,8 @@ const Form = forwardRef(
         useValue = formValue;
       else useValue = inputValue;
 
+      valueRef.current = valueState;
```
ericsoderberghp, line 152: Something feels odd that each input is resetting `valueRef` at the Form level.

src/js/components/Form/Form.js
```diff
@@ -168,6 +171,14 @@ const Form = forwardRef(
           }
           if (initialValue !== undefined) setInputValue(nextComponentValue);
         },
+        () => {
+          // Clean up valueState if input component is removed from
+          // the form (e.g. dynamic forms where inputs may be added
+          // or removed).
+          const nextValueState = valueRef.current;
+          delete nextValueState[name];
```
ericsoderberghp, line 179: To keep it clean, should we be deleting any errors, infos, and touched here as well?

src/js/components/Form/FormContext.js
```diff
@@ -14,6 +14,11 @@ const useFormInput = (_, valueProp, initialValue) => {
     nextValue => {
       if (initialValue !== undefined) setValue(nextValue);
     },
+    () => {
+      const nextValue = { ...value };
```
ericsoderberghp, line 18: I don't think `valueProp`, and thus `value`, is necessarily an object.

I think this function can be a no-op in this situation`() => {}`.

ericsoderberghp (commented):
Good draft for discussion.

I'm wondering if unmounting should always dictate removal from the Form. I think this is your point about something like a wizard context, where some fields are shown and some are not at any point in time. I do think we should solve for both cases at the same time.

A few thoughts:
- I am thinking the wizard scenario will be more likely than the add/remove fields scenario.
- We should consider the case where the `value` supplied to Form contains names that don't have any associated input but might be involved in validation.
- We should consider the case where the `value` supplied to Form doesn't contain a name that is used in an input component.
- We could argue that removing something, as this pull request addresses, should require the caller to use a controlled Form and take care of removing the property from the supplied `value` themselves.

PR #4795, opened 2020-12-02
Changed Text to add more sizes

ericsoderberghp:
#### What does this PR do?

Changed Text to add more sizes

Text -> All story

enhanced All story, enhanced unit tests

storybook

automatic

yes

backwards compatible


src/js/components/Text/doc.js
```diff
@@ -74,6 +74,11 @@ export const doc = Text => {
         'large',
         'xlarge',
         'xxlarge',
+        '2xl',
```
ShimiSun, line 77: Line 76 seems like a duplicate of 77.

ericsoderberghp: 76 is backwards compatible, 77 is friendlier for the new naming, both seem reasonable to me

ShimiSun: seeing them both as consecutive sizes gives the impression that '2xl', is larger than 'xxlarge'.

ericsoderberghp: I believe it's somewhat standard in clothing sizes that "2XL" and "XXLARGE" mean the same thing. We can add a note in the documentation that they are identical.

ShimiSun (approved):

Issue #4288, opened 2020-07-09
Select component should capture keyboard input


bdesouza:
I believe the [select component](https://storybook.grommet.io/?path=/story/select--simple) should mimic the native HTML select component. ~A core aspect of this is allowing the user to select an option with a keypress and using arrow keys to move up and down the list of options. Selection is done by pressing `space` or `enter`.~ The component doesn't let you scroll to the option by pressing the letter key associated with the option.

Example: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select

ShimiSun:
The Select does mimic the native HTML Select. https://storybook.grommet.io/?path=/story/select--simple
You are opening the drop with 'enter'  and navigate with the arrows through the options, and choose an option with 'enter' as well. I agree we can extend the support of both 'enter' and 'space',  but other than that I'm not sure what else is missing?

bdesouza:
@ShimiSun - You're right! It actually does mimic the native HTML element for the arrow keys and the select. 

I think where it breaks from the native HTML Select is when I do a keypress. If I have a list of States and hit the letter C, it should take me to the first option with the letter C.

bdesouza:
Any idea how I can pass in the `onKeyDown` to make this happen? We've been struggling with this for a while.

ShimiSun:
@bdesouza have you tried to warp the Select component with the Keyboard component?

bdesouza:
@ShimiSun - I tried, but the issues is that it doesn't seem to work if the Select field is used inside a form field component.

ShimiSun:
@bdesouza Interesting, would you like to work on a PR as a POC?

brandly:
Came here looking for this! The native `<select>` also scrolls the selected item into view to keep it visible while using arrow keys, but the grommet component does not. These two things make longs lists difficult to navigate.

bdesouza:
@ShimiSun - Sorry for the tardy reply.

I was playing around with the component and managed to get something working by updating `SelectContainer.js`.

```
//Added
const onKeyDownOption = useCallback(
  event => {
      event.preventDefault();
      let nextActiveIndex = options.findIndex((e, index) => {
        let label = (typeof e === 'object') ? e.label : e
        return label.charAt(0).toLowerCase() === event.key.toLowerCase() && !isDisabled(index)
        });
      if (nextActiveIndex >= 0) {
        setActiveIndex(nextActiveIndex);
        setKeyboardNavigation(true);
      }
[...]

PR #4797, opened 2020-12-03
Select an option in Select component using keypress

bdesouza:
#### What does this PR do?
Add functionality to the Select component to support keypress based navigation.

Run tests for the Select component

Load the storybook and navigate to the Select component. Open the Select component and then press a key to select an option based on the first letter of the option.

[Issue #4288](https://github.com/grommet/grommet/issues/4288)

screenshot

No

Don't think so

It's not a breaking change


ShimiSun (approved):

ShimiSun (changes requested):
Taking another look at the stories, I noticed that all the examples with a search box had stopped receiving the user inputs.
Here is a couple for reference:
https://deploy-preview-4797--sad-tereshkova-173d07.netlify.app/?path=/story/input-select--search

https://deploy-preview-4797--sad-tereshkova-173d07.netlify.app/?path=/story/input-select--create-option

Try to `onKeyDown` on the search and notice the missing input.

bdesouza:
Good catch @ShimiSun. Not sure how I missed it.

I was thinking about it and I think the dropdown with a search shouldn't have keydown navigation. The point of the keydown navigation is to help bring the user to the correct option. The search input accomplishes that. Perhaps what I could do is disable the above for select components with a search input?

src/js/components/Select/SelectContainer.js
```diff
@@ -277,6 +277,58 @@ const SelectContainer = forwardRef(
       [activeIndex, isDisabled],
     );
 
+    const findNextActiveIndex = (
+      event,
+      selectOptions,
+      currentActiveIndex,
```
ericsoderberghp, line 283: I think we can simplify this to not pass currentActiveIndex and afterActiveIndex. Looking at https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select, adding another option for Crocodile after Hamster, typing H selects H Hamster, then typing C goes back to Cat instead of down to Crocodile.

bdesouza: Thanks for the feedback, Eric. Will simplify. I was attempting to make it easier to iterate across options in the case of multiple options with the same first letter (Car, Cat, Caterpillar, ...). That said, makes sense to follow the HTML Select spec as close as possible.

ericsoderberghp (approved):

ShimiSun (approved):

PR #4799, opened 2020-12-03
Adjust layer zindex to match drop

### LINKED ISSUE #4756: Modal Layer leaves underlying Opened DropButton active
Related to #4721.  In this case a DropButton remains active after a Modal Layer is open.  In my case, I have a modal layer that appears after a timeout.  If the DropButton happens to be open at that time, then the opened DropButton remains active.  As soon as the DropButton is closed, then the modal Layer works as advertised.

### Expected Behavior

A Modal Layer ought to prevent interaction with all other "underlayer" components.

### Actual Behavior

If a DropButton component is open at the time the Modal Layer is displayed, then the DropButton can continue to be interacted with.

https://codesandbox.io/s/modal-layer-does-not-block-tab-to-underlayer-4srul?file=/src/App.js

### Steps to Reproduce

Open a DropButton.
Open a modal Layer.

Notice the DropButton remains active while the modal Layer is open.

### Your Environment

- Grommet version: 2.14, 2.15.2, stable (with the fix from #4721.
- Browser Name and version:

taysea:
#### What does this PR do?
Sets Layer z-index and Drop z-index to the same value. As Layer/Drop are appended to the DOM, the most recent one will take precedence over any other open portals. So, a Layer opened after a Drop will be on top, and vice versa.

Previously, Layer had a lower z-index than Drop which caused conflicts if a Drop was open when a Layer opened (the Drop would appear on top of the layer instead of behind).

src/js/themes/base.js

Tested scenarios with Drops inside Drops, Drops inside Layers, etc. to ensure the expected behavior remained.

Implement this as a story:
```
export const TestLayer = () => {
  const [openDrop, setOpenDrop] = React.useState(true);
  const [openLayer, setOpenLayer] = React.useState(false);
  const [countdown, setCountdown] = React.useState(5); // seconds
  const intervalId = React.useRef();

  // open the Layer after countdown reaches zero
  React.useEffect(() => {
    intervalId.current = setInterval(() => {
      setCountdown(prev => prev - 1);
    }, 1000);
  }, []);

  React.useEffect(() => {
    if (countdown === 0) {
      setOpenLayer(true);
      clearInterval(intervalId.current);
    }
  }, [countdown]);

  return (
    <Grommet theme={grommet}>
      <Heading>Modal Layer does not block TAB to underlayer</Heading>
      <Paragraph>
        Layer will open in <b>{countdown}</b> seconds.
      </Paragraph>
      <Paragraph>
        <Anchor href="https://v2.grommet.io/" label="https://v2.grommet.io/" />
      </Paragraph>
      <DropButton
        open={openDrop}
        onOpen={() => setOpenDrop(true)}
        onClose={() => setOpenDrop(false)}
        label="Click me!"
        dropAlign={{ top: 'bottom', right: 'left' }}
        trapFocus={false}
        dropContent={
          <Button
            label={`I'm the Drop - Wait for Layer to be displayed in ${countdown} seconds`}
            pad="large"
          />
        }
      />
      {openLayer && (
        <Layer onClickOutside={() => setOpenLayer(false)} modal={false}>
          <Box align="end">
            <Button icon={<Close />} onClick={() => setOpenLayer(false)} />
          </Box>
          <Heading margin="none" level="2">
            Instructions
          </Heading>
          <Paragraph margin="xsmall">This is the Modal Layer</Paragraph>
          <Paragraph margin="xsmall">
            Use the TAB key to access the link in the underlayer. This bug was
            introduced in Grommet v2.15.0 and fixed in v2.15.3 (yet to be
            released).
          </Paragraph>
          <DropButton
            label="Click me!"
            dropAlign={{ top: 'bottom' }}
            trapFocus={false}
            dropContent={
              <Box>
                <Select options={['one', 'twi', 'three']} />
                <Paragraph>Hey there!!!</Paragraph>
                <DropButton
                  label="Click me!"
                  dropAlign={{ top: 'bottom' }}
                  trapFocus={false}
                  dropContent={
                    <Box>
                      <Select options={['one', 'twi', 'three']} />
                      <Paragraph>Hey there!!!</Paragraph>
                    </Box>
                  }
                />
              </Box>
            }
          />
          <Paragraph margin="xsmall">
            Also note that if a DropButton is open, then the most recent modal
            does not prevent it from being accessed via the mouse. This
            behaviour exists on v2.14.0.
          </Paragraph>
          <Paragraph margin="xsmall">
            To observe the behaviour in code sandbox, rebuild with grommet
            v2.15.0, then open this little app in a new window. The problem is
            not evident with v2.14.0.
          </Paragraph>
          <Paragraph margin="xsmall">
            <Anchor
              href="https://icons.grommet.io/"
              label="https://icons.grommet.io/"
            />
          </Paragraph>
        </Layer>
      )}
    </Grommet>
  );
};
```
Closes #4756 

Updated here.

No.

Yes, backwards compatible.

### DIFF
--- a/src/js/components/Layer/README.md
+++ b/src/js/components/Layer/README.md
@@ -224,7 +224,7 @@ The stack order of Layer Container. Expects `number`.
 Defaults to
 
 ```
-15
+20
 ```
 
 **layer.extend**
@@ -265,7 +265,7 @@ The stack order of Layer. Expects `number`.
 Defaults to
 
 ```
-10
+20
 ```
 
 **global.breakpoints**
--- a/src/js/components/Layer/__tests__/__snapshots__/Layer-test.js.snap
+++ b/src/js/components/Layer/__tests__/__snapshots__/Layer-test.js.snap
@@ -11,7 +11,7 @@ exports[`Layer animation fadeIn 1`] = `
   -webkit-font-smoothing: antialiased;
   background: transparent;
   position: relative;
-  z-index: 10;
+  z-index: 20;
   pointer-events: none;
   outline: none;
   position: fixed;
@@ -45,7 +45,7 @@ exports[`Layer animation fadeIn 1`] = `
   color: #444444;
   outline: none;
   pointer-events: all;
-  z-index: 15;
+  z-index: 20;
   position: absolute;
   max-height: calc(100% - 0px - 0px);
   max-width: calc(100% - 0px - 0px);
@@ -125,7 +125,7 @@ exports[`Layer animation fadeIn 1`] = `
 `;
 
 exports[`Layer animation fadeIn 2`] = `
-".gNIvtm {
+".gmvEoJ {
   font-size: 18px;
   line-height: 24px;
   box-sizing: border-box;
@@ -135,7 +135,7 @@ exports[`Layer animation fadeIn 2`] = `
   -webkit-font-smoothing: antialiased;
   background: transparent;
   position: relative;
-  z-index: 10;
+  z-index: 20;
   pointer-events: none;
   outline: none;
   position: fixed;
@@ -145,7 +145,7 @@ exports[`Layer animation fadeIn 2`] = `
   bottom: 0px;
 }
 @media only screen and (max-width: 768px) {
-  .gNIvtm {
+  .gmvEoJ {
     position: absolute;
     width: 100%;
     height: 100%;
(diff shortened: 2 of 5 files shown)


ShimiSun (approved):
why not using 15 for both so no change will have to be applied on drop?

taysea:
> why not using 15 for both so no change will have to be applied on drop?

That would also work, it just needs to be the same zIndex. I've adjusted Layer to use `20` currently so that it matches Drop. Currently, the changes are only occurring in Layer.

ericsoderberghp (approved):

PR #4801, opened 2020-12-04
Changed DataTable to allow setting the pinned background from the background context

ericsoderberghp:
#### What does this PR do?

Changed DataTable to allow setting the pinned background from the background context

Box.js
Cell.js

Added unit tests, tweaked Fill and Pin storybook

storybook

https://github.com/grommet/grommet-theme-hpe/issues/145

no

yes

Not 100%. If a theme sets a pinned background, there is no change. Using the grommet theme, currently pinned DataTable elements have no background, which doesn't look very good. This case will look different as we will use the background color + opacity.


halocline:
I am finding myself wanting to articulate the hierarchy / cascading of the background colors. Let me see if I have this correct.

Working from acute to general:

- `<DataTable background={{ header: ..., body: ..., footer: ... }} />`
- `<DataTable background="..." />`
- `theme.dataTable.pinned[context].background.color`
- `theme.table.[header, body, footer].background.color`
- background from Box background prop via ThemeContext.Provider

If I have that correct, I think I like the direction. Background context gets established by a Box's background prop. I am curious where you have reservations.

Additionally, and again if I have this hierarchy correct, I think the current implementation is not taking into consideration any theme defined in `theme.table.[header, body, footer]`.

ericsoderberghp:
> I am finding myself wanting to articulate the hierarchy / cascading of the background colors. Let me see if I have this correct.
> 
> Working from acute to general:
> 
> * `<DataTable background={{ header: ..., body: ..., footer: ... }} />`
> * `<DataTable background="..." />`
> * `theme.dataTable.pinned[context].background.color`
> * `theme.table.[header, body, footer].background.color`
> * background from Box background prop via ThemeContext.Provider
> 
> If I have that correct, I think I like the direction. Background context gets established by a Box's background prop. I am curious where you have reservations.
> 
> Additionally, and again if I have this hierarchy correct, I think the current implementation is not taking into consideration any background defined in `theme.table.[header, body, footer]`.

I think you're on the right track. But, from Table's point of view, `theme.dataTable.*` becomes `<Table background />`. So, if the DataTable theme doesn't specify background colors, then the `theme.table.*` backgrounds will be used.

ShimiSun (approved):

Issue #4784, opened 2020-11-26
DataTable onSelect TypeScript types


oskari:
I think the types for onSelect -function are inferred wrong

### Expected Behavior

Using DataTable onSelect as seen on story [DataTable / Select](https://storybook.grommet.io/?path=/story/visualizations-datatable--select) should work.

### Actual Behavior

![image](https://user-images.githubusercontent.com/506596/100356564-12e20b00-2ffc-11eb-9bd9-8445f9fd7f4f.png)

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-typescript-datatable-onselect-typing-hj11t?file=/src/App.tsx

### Your Environment

name | version
-|-
grommet |  2.16.0
react | 17.0.0
react-dom | 17.0.0
react-scripts | 3.3.0
styled-components | 5.2.1

- Grommet version: 2.16.0
- Browser Name and version: Firefox Developer Edition 83.0b9
- Operating System and version (desktop or mobile): macOS Catalina 10.15.7

apawson:
Similarly, the new `Tip` component appears to be missing from `node_modules/grommet/index.d.ts`

![image](https://user-images.githubusercontent.com/31909523/100475060-0068df80-3097-11eb-9866-cddcb1024331.png)

ShimiSun:
If you guys already have an idea of how to fix the interface file or the export, please feel free to make a contribution.

apawson:
> If you guys already have an idea of how to fix the interface file or the export, please feel free to make a contribution.

https://github.com/grommet/grommet/pull/4785

jcfilben:
[Here](https://github.com/grommet/grommet/pull/4483) is the initial PR where they type for `onSelect` in DataTable was defined. I tried to change the type from `onSelect?: (string | number)[];` to `onSelect?: (select: React.Dispatch<React.SetStateAction<(string | number)[]>>) => void;` which is the suggested type for the setter in useState but I was still getting an error when trying to use typescript in the DataTable / Select story. I have also found some information suggesting the use of Dispatch with useState typing but I haven't had success with that either. I am not sure what the correct type for `onSelect` is just wanted to give an update of the information I have found out about so far for this issue.

PR #4804, opened 2020-12-04
Fix DataTable onSelect type

oskari:
#### What does this PR do?

Fixes DataTable onSelect types by excluding the onSelect inferred from table

src/js/components/DataTable/index.d.ts

Storybook stories

See added storybook story or by linking the package and using DataTables onSelect -function

#4784 

No

### DIFF
--- a/src/js/components/DataTable/index.d.ts
+++ b/src/js/components/DataTable/index.d.ts
@@ -100,7 +100,7 @@ export interface DataTableProps<TRowType = any> {
 }
 
 declare class DataTable<TRowType = any> extends React.Component<
-  DataTableProps<TRowType> & JSX.IntrinsicElements['table']
+  DataTableProps<TRowType> & Omit<JSX.IntrinsicElements['table'], 'onSelect'>
 > {}
 
 export { DataTable };
--- a/src/js/components/DataTable/stories/DataTable.stories.js
+++ b/src/js/components/DataTable/stories/DataTable.stories.js
@@ -7,6 +7,7 @@ export { Custom } from './Custom';
 export { Fill } from './Fill';
 export { GroupedDataTable } from './Grouped';
 export { InfiniteScrollDataTable } from './InfiniteScrollDataTable';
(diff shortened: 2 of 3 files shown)


jcfilben:
I am curious as to why you think `onSelect` should be omitted rather than assigned a type. I know that currently the type definition for `onSelect` is incorrect but wouldn't the ideal fix be to change the type definition rather than omitting it? Is there a reason you think that we should omit it?

oskari:
@jcfilben I'm sorry, my wording is probably a bit misleading. We should omit the `onSelect` from `JSX.IntrinsicElements['table']` as that is not our `onSelect` types. By omitting that, the types are exactly what we define, and not something that is an union of our types and the inferred Table element onSelect types. 

The types defined in [DataTable/index.d.ts#L98](https://github.com/grommet/grommet/blob/5e8d2eb058382e5e5c8e583f7a38352c4471654a/src/js/components/DataTable/index.d.ts#L98) are okay for now, as the DataTable will (according to documentation) call the passed `onSelect` -function with _"array of primary key values, suitable to be passed to the 'select' property"_. What you're exploring in your comment on the relevant issue is not something I think we should be doing, as that would force us to use `useState` hook in the `onSelect` which is only one possible use-case out of several others.

jcfilben:
Ah okay I see, thank you for the explanation, very helpful!

ShimiSun:
@jcfilben we'll need to check whether React is providing a default `onSelect` for table before safely removing it. Also, try to apply reverse engineering on when and why the Omit was added and whether it was a community request issue/PR that may explain the existence of it.

jcfilben:
@ShimiSun There is a default type for on select in Table. I tested by adding onSelect to the Table/Default story. I also tested onSelect in DataTable with different value/uses and wasn't able to find any that didn't work.

ShimiSun (approved):

ShimiSun:
Thank you for your patience and contribution! 💟

Issue #4781, opened 2020-11-24
Is there a way to control the activeSuggestionIndex property of TextInput?


ddohler:
We've had a request to auto-select the first suggestion in a TextInput's suggestions list when the user hits Enter, so that they are not required to click or use the down-arrow first. I've found that the active suggestion is managed by `activeSuggestionIndex` within the TextInput, but that state seems to be internal to the component, without any way to control it from the outside. Is this possible?

### Expected Behavior

(Not sure this is helpful in this case since this is more of a question, but here's an example just in case):
Type an entry into a TextInput that has suggestions, so that one or more suggestions are shown.
Hit Enter
`onSuggestionSelect` is fired with the value of the first item in the suggestions list.

### Actual Behavior

The suggestions Drop disappears, and `onSuggestionSelect` is not fired. I haven't found a way to customize this behavior so far.

### URL, screen shot, or Codepen exhibiting the issue

I believe this behavior occurs with TextInputs that use suggestions, but I am specifically referring to the geography name search at the top of this page: https://wildfirerisk.org/explore/0/30/30063/3000050200/

<img width="469" alt="Screen Shot 2020-11-24 at 4 58 55 PM" src="https://user-images.githubusercontent.com/447977/100156333-b1446400-2e76-11eb-8286-c2083eadd074.png">

### Steps to Reproduce

1. Type an entry into a TextInput that has suggestions, so that one or more suggestions are shown.
2. Hit Enter

### Your Environment

- Grommet version: 2.10.0
- Browser Name and version: All desktop browsers.
- Operating System and version (desktop or mobile): Desktop

Thanks!

ShimiSun:
Just a quick thought here, have you tried wrapping the component with Keyborad component and control the Enter behavior yourself?

ddohler:
Thanks for the reply! Yep, that's a potential fallback we've discussed, but the trouble with that approach is that there will be no visual indication that the first suggestion will be selected if the user presses Enter. Ideally we'd like to also highlight the first suggestion. 

Looking at other autoselect libraries, we've seen a `highlightFirstSuggestion` prop used to accomplish this goal -- would you all be open to a PR adding something like that to `TextInput`? My quick read of the `TextInput` makes me think that it should primarily require adjusting all the occasions where `setActiveSuggestionIndex` is called with `-1` to instead call with `0` if the new prop is true. Does that sound like a fair characterization of the task to you, or is there something major I'm missing?

[...]

PR #4806, opened 2020-12-04
Add highlightFirstSuggestion prop to TextInput

ddohler:
#### What does this PR do?

Adds a `highlightFirstSuggestion` prop to TextInput, which, when set, causes the suggestions list to default to highlighting the first suggestion. Default behavior is that the first suggestion is not highlighted.

Either the two added tests, or the storybook, to get a sense of the desired behavior, unless you've touched `TextInput` recently and it's fresh, in which case you could jump straight there.

Tested by adding the new prop to the `Suggestions` storybook. Wrote two unit tests describing the desired behavior.

I did most of my development testing by editing and debugging the `Suggestions` storybook.

I attempted to follow the contribution instructions and I believe that I edited all the necessary ancillary files (storybook, tests, `index.d.ts`) but if I've missed something, please let me know. 

There were two internal state variables, `activeSuggestionIndex` and `selectedSuggestionIndex`, that appeared to do the same thing within `TextInput`. `activeSuggestionIndex` was the more widespread of the two, so I removed `selectedSuggestionIndex` and converted its uses to use `activeSuggestionIndex` instead. All tests continue to pass after this change, and I haven't detected any changes to the suggestions functionality otherwise, but I'm happy to revert this part if `selectedSuggestionIndex` is important to maintain separately for some reason that I've missed.

#4781 

highlightFirstSuggestion

Yes. My understanding is that this happens automatically, but I'm happy to make further changes as necessary to support this.

I'm not sure what the project standards are here, but as a library consumer, I would find it helpful for this to be included. Please let me know if there are other files I should edit in order to make this happen.

This change is backwards compatible. If the new property is not set, the existing behavior is the default.


ericsoderberghp:
I apologize for the delay responding to this issue. From looking at the site referenced in the issue, it seems like the intent is to provide a search control where picking one of the suggestions is required in order to have any effect. Just typing "12" in the search on that site and then pressing Enter doesn't do anything in that situation. Is that expected?

In some since, for this use, "suggestions" isn't really the right term, "options" would be better. It's almost more like a Select control with options and a search control to filter the options. But, even in the Select method, showing the selected option in a separate element isn't desired or required either, as with the TextInput method.

Given all of this, I think TextInput is the right basis. And from the example site shown, I don't think activating the first suggestion is as ideal as maybe something like "defaultActiveSuggestionIndex". Something like defaultActiveSuggestionIndex would allow the caller to pre-set to the last selected suggestion. This would facilitate a flow where the user is not sure which one to select, or selects the wrong one, then the user can proceed quickly to the next suggestion. "defaultActiveSuggestionIndex" is a bit long in the tooth though, maybe "defaultSuggestion" that takes either a number or a string?

ddohler:
@ericsoderberghp Thanks for responding! Let me try to clear up your questions and confirm that I understand everything.
> Just typing "12" in the search on that site and then pressing Enter doesn't do anything in that situation. Is that expected?

There are no options that contain numbers (all the options are the names of cities, counties, and states), so the specific example you gave of typing "12" is never expected to do anything because there will be no matching suggestions.

However, if you were to type, say, `New`, and then press Enter, the desired behavior is for the first suggestion to be selected and for the app to move to the next page immediately.

Screen Shot 2020-12-16 at 8 49 54 AM


We help maintain this site, and we haven't figured out a way to do this using the existing Grommet functionality, so that's why we've opened the current PR.

> Something like defaultActiveSuggestionIndex would allow the caller to pre-set to the last selected suggestion.

I'm not 100% sure I understand what you mean -- is something like the following what you had in mind?
```
<TextInput
  value={value}
  onChange={onChange}
  onSelect={onSelect}
  suggestions={suggestions}
  defaultActiveSuggestionIndex={prevSelectionIndex} // Could replicate highlightFirstSuggestion by passing 0 here
/>
```

> maybe "defaultSuggestion" that takes either a number or a string?

The number case seems reasonably clear to me, assuming that the example above matches what you had in mind. I'm not sure what kinds of values would be expected in the case where the user passed a string--could you clarify what you were thinking here?

Overall, I think something like what you're proposing would be a relatively small change to what's here, so I'm happy to implement it that way if you think that is what you want. But I also think, if I'm understanding what you're proposing, that it would be possible to achieve similar results using `highlightFirstSuggestion` by manipulating the order of the `suggestions` prop to place the chosen item in the first slot. For example:
```
// Assuming `previousSuggestion` is tracked somewhere
<TextInput
  value={value}
  onChange={onChange}
  onSelect={onSelect}
  suggestions={[previousSuggestion, ...otherSuggestions]}
  highlightFirstSuggestion
/>
```
There would be some differences in the UX, of course, but otherwise the result should be roughly the same. Again though, I'm happy to go either way. 🙂

ericsoderberghp:
> There are no options that contain numbers (all the options are the names of cities, counties, and states), so the specific example you gave of typing "12" is never expected to do anything because there will be no matching suggestions.

Thanks, just confirming that.

> The number case seems reasonably clear to me, assuming that the example above matches what you had in mind. I'm not sure what kinds of values would be expected in the case where the user passed a string--could you clarify what you were thinking here?

I was thinking the caller could pass the actual suggestion string, such as "East Newark, NJ" . I'd be fine just starting with a number though.

> But I also think, if I'm understanding what you're proposing, that it would be possible to achieve similar results using highlightFirstSuggestion by manipulating the order of the suggestions prop to place the chosen item in the first slot.

I was thinking that if the user initially arrowed down and pressed Enter on the fourth suggestion but then realized afterwards that he wanted to next one, he would be right where he left off and could get to the fifth one quickly. I don't think re-ordering the suggestions is a good idea, as the user will lose the familiarity from the last time they were opened.

From a naming standpoint, I'm leaning towards `defaultSuggestion`.

ericsoderberghp (approved):
Nicely done.

ddohler:
@ericsoderberghp Okay, sounds good! I've gone ahead and updated this branch to switch `highlightFirstSuggestion` to `defaultSuggestion`, which takes an index into the suggestion array as  and then highlights the corresponding suggestion. One thing I wanted to highlight is that this doesn't attempt to prevent callers from passing in indices that are outside the bounds of the `suggestions` array. Doing that doesn't cause any exceptions to be thrown, but the selected suggestion will be `undefined`. 

The only other option I could think of was to attempt to recover from a bad `defaultSuggestion` by snapping to the upper or lower bounds of the suggestion array, but that would increase the complexity of this change, and would be less likely to alert the caller that they have some kind of logic error in the calling code. But please let me know if you think differently!

Otherwise, I just implemented the numeric version of this; since we don't have this need I didn't want to stray too far from the known use-case.

I see you've already approved -- thanks for reviewing! Please let me know if there is anything else I can or should do to move this forward.

ericsoderberghp:
Thanks for your contribution!

ShimiSun:
@ddohler Hi buddy, thanks for your contribution 🎉 . 
Can you please add the new prop `defaultSuggestion` to the doc.js file? this way users of the component could leverage from your work and have the new prop documented.

PR #4810, opened 2020-12-05
NEW Tree component

ShimiSun:
#### What does this PR do?
Provide 'column' display of Tree component and some foundation ideas for 'nested' Tree
Looking at the examples and Tree.js
Storybook+jest as a starter
storybook and review
CCP designs
https://www.figma.com/proto/qxdE6A2qswCVbW3R6Ti7KG/CCS-v0.09-(New-Navigation)?node-id=5945%3A39378&viewport=446%2C101%2C0.015625&scaling=scale-down
#3489 
https://github.com/grommet/hpe-design-system/issues/1383
https://github.com/grommet/hpe-design-system/issues/1322
Done, might need further adjustments as we progress
Yes. once we'll get there

b/c


src/js/components/Tree/Tree.js
```diff
@@ -0,0 +1,161 @@
...
+import { Button } from '../Button';
+import { List } from '../List';
+import { Text } from '../Text';
+
+const Tree = ({ children, mode, data }) => {
+  const theme = useContext(ThemeContext);
+
+  const [depth, setDepth] = useState(0);
+  // an array of the path names
+  const [breadcrumbs, setBreadcrumbs] = useState([]);
```
ericsoderberghp, line 17: A minor idea: rename "breadcrumbs" to "ancestors"? Technically, breadcrumbs mean "how I got here". In the case of a deep link, the user may not have navigated sequentially through the tree.

Actually, based on comments further down in this file, perhaps "path" would be a more precise name?

src/js/components/Tree/Tree.js
```diff
@@ -0,0 +1,161 @@
...
+  const [selectedNode, setSelectedNode] = useState();
+  // treeLevel should only be updated on useEffect to reflect correct depth
+  const [treeLevel, setTreeLevel] = useState(data);
```
ericsoderberghp, line 20: Initializing the "level" of the tree directly from the "data" looks odd to me here, conceptually.

src/js/components/Tree/Tree.js
```diff
@@ -0,0 +1,161 @@
...
+
+import { Box } from '../Box';
+import { Button } from '../Button';
+import { List } from '../List';
+import { Text } from '../Text';
+
+const Tree = ({ children, mode, data }) => {
+  const theme = useContext(ThemeContext);
+
+  const [depth, setDepth] = useState(0);
```
ericsoderberghp, line 15: Does depth signify the deepest depth of the data? How does "depth" relate to "level"? The two sound similar. Perhaps a comment?

src/js/components/Tree/Tree.js
```diff
@@ -0,0 +1,161 @@
...
+import { FormDown } from 'grommet-icons/icons/FormDown';
+import { FormNext } from 'grommet-icons/icons/FormNext';
+import { FormPrevious } from 'grommet-icons/icons/FormPrevious';
+
+import { Box } from '../Box';
+import { Button } from '../Button';
+import { List } from '../List';
+import { Text } from '../Text';
+
+const Tree = ({ children, mode, data }) => {
```
ericsoderberghp, line 12: Can the caller start from a node already selected?

src/js/components/Tree/Tree.js
```diff
@@ -0,0 +1,161 @@
...
+      // follow the breadcrumbs path
+      parent = level.find(element => element.name === breadcrumbs[i]);
+      level = parent.children;
+    }
+    setParentLevel(parent);
+    setTreeLevel(level);
+  }, [breadcrumbs, data, depth]);
+
+  const ColumnNavigationList = () => (
+    <Box width="medium">
```
ericsoderberghp, line 37: How will this behave in a mobile context?

src/js/components/Tree/Tree.js
```diff
@@ -0,0 +1,161 @@
...
+      <List
+        data={treeLevel}
+        onClickItem={event => setSelectedNode(event.item)}
+        {...theme.tree.column.listProps}
+      >
+        {datum => {
+          return (
+            <Box direction="row" justify="between" align="center" gap="small">
+              <Text size="small">{datum.name}</Text>
```
ericsoderberghp, line 46: Why "small" here? I would have expected "medium".

src/js/components/Tree/Tree.js
```diff
@@ -0,0 +1,161 @@
...
+              <Box direction="row" gap="hair" align="center">
+                <Text size="xsmall">
```
ericsoderberghp, line 48: This part, where we see the count and the label "sub-resources" feels too opinionated to me. I'm inclined to not include this in our first go at Tree.

src/js/components/Tree/Tree.js
```diff
@@ -0,0 +1,161 @@
...
+              // remove the last breadcrumb item when Previous is selected
+              breadcrumbs.pop();
+              setBreadcrumbs(breadcrumbs);
+              setDepth(depth - 1);
+              setSelectedNode(parentLevel);
+            }}
+          />
+        )}
+        {/* Layout Containers for List and user content */}
+        <Box direction="row" border width="large">
```
ericsoderberghp, line 114: I don't think "large" will work on mobile.

src/js/components/Tree/Tree.js
```diff
@@ -0,0 +1,161 @@
...
+                label={
+                  <Text size="small">
+                    {`${selectedNode.children.length} sub-resources`}
+                  </Text>
+                }
+                icon={<FormNext />}
+                reverse
+                onClick={() => {
+                  // add a new selected entry to the breadcrumbs path
+                  setBreadcrumbs([...breadcrumbs, selectedNode.name]);
```
ericsoderberghp, line 131: I think we should allow something like `nameKey` or `labelKey` as a prop such that this would be `selectedNode[labelKey]`.

src/js/components/Tree/doc.js
```diff
@@ -0,0 +1,52 @@
...
+    .description(`Hierarchical list structure.`)
+    .usage(
+      `import { Tree } from 'grommet';
+<Tree />`,
+    );
+
+  DocumentedTree.propTypes = {
+    data: PropTypes.arrayOf(PropTypes.object).description(
+      `The data structure provided to the tree, for nested behavior 
+      an object should include a children key.`,
```
ericsoderberghp, line 17: I think we also expect `name` currently. Although, see my other comment about `labelKey`.
Similarly, I also think we should support `childrenKey`. That would allow the caller to pass any object structure, using their existing data.

src/js/components/Tree/stories/Tree.stories.js
```diff
@@ -0,0 +1,7 @@
+export { Busy } from './Busy';
+export { Flat } from './Flat';
+export { Nested } from './Nested';
+
+export default {
+  title: 'Controls/Tree',
```
ericsoderberghp, line 6: Are we sure it should be "Controls"? It seems like it goes where Table and List are, more so than "Controls".

ericsoderberghp (commented):
Nice progress! I think most of my comments are around the variable names.

src/js/components/Tree/Tree.js
ShimiSun: I agree, it more about the data of the level rather than the level itself.
would `treeLevelData` make more sense?

src/js/components/Tree/Tree.js
ShimiSun: Changed to `path`.

src/js/components/Tree/Tree.js
ShimiSun: Yep, not obvious ondeed. I added more descriptions and comments about the state variables and how they are related to each other.

src/js/components/Tree/Tree.js
ShimiSun: Not at the moment. Nice enhancement.

src/js/components/Tree/Tree.js
ShimiSun: Bad. I know the code needs cleanups from the current opinionated styling, I'm using it temporarily to get the functionality nailed down first, and I'll do the styling afterward.

src/js/components/Tree/stories/Tree.stories.js
ShimiSun: Fixed.

Issue #4796, opened 2020-12-03
Uncontrolled Form - Checkbox is not visually reset when form is reset


ArrayKnight:
### Expected Behavior

If a form includes a checkbox, and a user checks that checkbox, then the form is reset, the checkbox should visually represent that the checkbox is no longer checked.

### Actual Behavior

The checkbox appears checked visually (unexpected) but validates as if not checked (expected).

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-forked-e5bpe

![image](https://user-images.githubusercontent.com/5431637/100958206-c3ef1680-34d0-11eb-92bf-0e2b30d43288.png)

### Steps to Reproduce

1. Create a Form with an uncontrolled Checkbox inside
2. Check the checkbox manually
3. Reset the form

### Your Environment

- Grommet version: 2.16.0
- Operating System and version (desktop or mobile): Windows 10

ShimiSun:
also true for RadioButton and Select https://storybook.grommet.io/?path=/story/input-form--uncontrolled-typed

ShimiSun:
@ArrayKnight thanks for reporting, the issue is fixed, we'll release it officially shortly.

idanhaviv:
@ShimiSun hey, any updates on this?

PR #4817, opened 2020-12-08
Reset inputs when form resets

taysea:
#### What does this PR do?
Resets uncontrolled input values when form has reset. Previously, if a Form reset, CheckBox/RadioButtonGroup/Select would not reset their value.

Tested with "Uncontrolled" storybook example and added jest test.

Uncontrolled Form example storybook.

Closes https://github.com/grommet/grommet/issues/4796

No.

Yes. "Uncontrolled input values properly reset when Form is reset."

Backwards compatible.


ShimiSun (approved):
Nice work, thanks!

Issue #4818, opened 2020-12-09
Broken Select with async options


benschiefer:
Commit https://github.com/grommet/grommet/commit/98368af6395a3f6e36fab23c111b5d517cffc9d3 has broken select with async options

When a select is rendered with options populated asynchronously (ie: after initial render), the selected value is not rendered in the input (it does show as selected in the option layer when it is opened)

### Expected Behavior

Select should behave as it did before this commit, where options that are updated AFTER the initial render will still render the selected value properly.

### Actual Behavior

Selected Value does not render if options are updated after initial render.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-forked-nqvql?file=/index.js

### Steps to Reproduce

1. Render a <Select /> element with options that are NOT populated at initial render
2. Notice that the selected value is not displayed properly

### Your Environment

- Grommet version: 2.16.0
- Browser Name and version: All
- Operating System and version (desktop or mobile): All

Issue #4802, opened 2020-12-04
Dynamically loaded options and value are not shown in Select component


SagarTrimukhe:
### Expected Behavior
If the user dynamically passes the options and a value to a Select component, then the value passed should be shown as selected in the component.

### Actual Behavior
The component is not showing the passed value as selected.
It looks like a regression. This behavior is achievable in grommet version: 2.15.2

### URL, screen shot, or Codepen exhibiting the issue
Please check this code box example. (using Grommet version: 2.15.2) **Working as expected**
https://codesandbox.io/s/snowy-haze-bpe4j?file=/src/App.js

This is the same example with Grommet version 2.16.0  **Not working**
https://codesandbox.io/s/adoring-visvesvaraya-5piwx?file=/src/App.js

### Your Environment

- Grommet version: 2.16.0
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): Desktop

Issue #4792, opened 2020-12-02
Unable to select a dynamically loaded option from Select Field


sindhoora-r:
Unable to select a dynamically loaded option from Select Field. 
In a Select component, where we lazy load more options, we are unable to select an option from the lazy loaded options.

### Expected Behavior
We should be able to select any option from the Select Component

### Actual Behavior
We are unable to select an option from the latest list of options(load more)
https://storybook.grommet.io/?path=/story/input-select--lazy-loading-options
Unable to select the options after the Option 199 in this story book.

### URL, screen shot, or Codepen exhibiting the issue
https://storybook.grommet.io/?path=/story/input-select--lazy-loading-options

### Steps to Reproduce

Open https://storybook.grommet.io/?path=/story/input-select--lazy-loading-options
Scroll to Option 200
Select Option 200, it is not selected, if we close and open it, the option is selected but the selected value is not displayed in the Select Field.

### Your Environment

- Grommet version: 2.16.0
- Browser Name and version: chrome
- Operating System and version (desktop or mobile): Mac, Windows

ShimiSun:
Thanks for reporting, we'll prioritize the issue.

sabino-gadaleta:
> Thanks for reporting, we'll prioritize the issue. Can you confirm the behavior was working as expected on v2.15.2?

We did confirm inside our UI that the problem did not show up if we added a resolution that forced use of the v2.15.2 grommet version.

halocline:
FYI, I did some initial investigation to track down where this regression was introduced.

Using git bisect (starting on the master branch) as follows:
```
git bisect start
git bisect bad
git bisect good v2.15.2
```

Identified https://github.com/grommet/grommet/pull/4778 as the commit which introduced the regression. 

I haven't researched beyond here, but the description in the PR "The problem was that options change when they are filtered using search. So indexes of selected values are invalid. Now we always use the initial list of options, so the indexes will always be valid." sounds as if the approach taken may not work in scenarios where the options are loaded using InfiniteScroll's onMore.

ShimiSun:
@romanalesenkov1 FYI ⬆️

sabino-gadaleta:
Just for more information. We had seen the issue too in our UI if we just render the select component before the options are fully loaded. So it's not only an issue with onMore maybe but a more general loss of binding if the options change after render.

ShimiSun:
[...]

PR #4821, opened 2020-12-10
Fix 4792 - Refine multi-select with search to work with lazy-load options

halocline:
#### What does this PR do?

- Closes https://github.com/grommet/grommet/issues/4792
- Closes https://github.com/grommet/grommet/issues/4818
- I believe this also closes https://github.com/grommet/grommet/issues/4802
- Refine multi-select with search to work with lazy-load options
  - Modifies previous approach using `initialOptions` to maintain flexibility when additional options are added in lazy-load / onMore scenarios
- Adds tests for:
  1.  search+multiple select
  1. lazy load more options and select

Select.js

Additional Jest tests
Storybook:
  - ?path=/story/input-select--search
  - ?path=/story/input-select--lazy-loading-options

Storybook:
  - ?path=/story/input-select--search
  - ?path=/story/input-select--lazy-loading-options

- https://github.com/grommet/grommet/issues/4792
- https://github.com/grommet/grommet/issues/4818
- https://github.com/grommet/grommet/issues/4802

No.

Yes.

Backwards compatible.


ShimiSun (approved):

src/js/components/Select/Select.js
```diff
@@ -101,12 +101,20 @@ const Select = forwardRef(
         );
       return valueKey && valueKey.reduce ? value : applyKey(value, valueKey);
     }, [value, valueKey]);
+    // maintain master list of all options to reference when
+    // option values are selected from a filtered set of options
+    // (e.g. such as when searching)
+    const [optionsTotal, setOptionsTotal] = useState(options);
+    useEffect(() => {
+      if (options.length > optionsTotal.length) {
```
ericsoderberghp, line 109: I think this assumes that the caller is only reducing via search. For example, what if is a Select whose options depend on another form input element? For instance, if the Select contains "countries" whose options change based on another form input that selects from "continents", this would only reset optionsTotal when going from North America to Europe, but not vice-versa.

halocline: Good point and example. Revisiting a concept I had explored prior.

halocline: @ericsoderberghp I modified the approach to set `optionsTotal` to `initialOptions` only when searching. When not searching, use `options` like so:

```
const [search, setSearch] = useState();
const [initialOptions] = useState(options);
// if searching, use initialOptions
const optionsTotal = search ? initialOptions : options;
```

src/js/components/Select/Select.js
```diff
@@ -101,12 +101,18 @@ const Select = forwardRef(
         );
       return valueKey && valueKey.reduce ? value : applyKey(value, valueKey);
     }, [value, valueKey]);
-
+    // search input value
+    const [search, setSearch] = useState();
+    // track initialOptions for reference when option values are selected
+    // from filtered options set (e.g. such as when searching)
     const [initialOptions] = useState(options);
```
ericsoderberghp, line 108: I think initialOptions will never change. If the caller calls with a different `options`, `initialOptions` remains as it was initially.

ericsoderberghp (approved):

PR #4822, opened 2020-12-10
Enabled customization of DataTable selected column

ShimiSun:
#### What does this PR do?
Enables customization of DataTable selected column
DataTable.js
storybook + jest
storybook + jest (storybook: dataTable -> OnSelect)
The issue was driven from CCP, currently, there is no way to control the styling of the auto-generated selected column.
Before (too much space between the Name and selected column
image

The desired behavior:
image

Done.
Yes
b/c

### DIFF
--- a/src/js/components/DataTable/Body.js
+++ b/src/js/components/DataTable/Body.js
@@ -1,5 +1,6 @@
 import React, { forwardRef } from 'react';
 
+import { Box } from '../Box';
 import { CheckBox } from '../CheckBox';
 import { InfiniteScroll } from '../InfiniteScroll';
 import { TableRow } from '../TableRow';
@@ -17,6 +18,7 @@ const Body = forwardRef(
       border,
       columns,
       data,
+      gdtSelected,
       onMore,
       replace,
       onClickRow,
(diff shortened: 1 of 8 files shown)


src/js/components/DataTable/Body.js
```diff
@@ -17,6 +18,7 @@ const Body = forwardRef(
       border,
       columns,
       data,
+      gdtSelected,
```
ericsoderberghp, line 21: "gdt"?

src/js/components/DataTable/Body.js
```diff
@@ -112,18 +114,22 @@ const Body = forwardRef(
                 >
                   {(selected || onSelect) && (
                     <TableCell background={background}>
```
ericsoderberghp, line 116: Can we do something to restrict this cell's width in all cases?

src/js/components/DataTable/DataTable.js
```diff
@@ -70,6 +70,17 @@ const DataTable = ({
   step = 50,
   ...rest
 }) => {
+  // Pulling out the data object for the selected column
+  const gdtSelected = columnsProp.find(
+    column => column.property === 'gdt-selected',
```
ericsoderberghp, line 75: This look like "gdt-selected" becomes a special property name?

src/js/components/DataTable/README.md
```diff
@@ -447,7 +447,12 @@ When supplied, causes checkboxes to be added to each row such that
       will be called with an array of primary key values, suitable to be
       passed to the 'select' property. If you are storing select state via
       a 'useState' hook, you can do something like:
-      '<DataTable select={select} onSelect={setSelect} />'.
+      '<DataTable select={select} onSelect={setSelect} />'. The rendering of the
+      selected column is generated automatically, if you like to customize its 
+      column size or align properties, add the reserved property name of 
+      'gdt-selected' with the attributes of 
+      {property: 'gdt-selected', size: '0px', align: 'end' }
```
ericsoderberghp, line 454: Hmm... This feels a bit hack-y to me.

1. I'm thinking we should change the default behavior to restrict the size of the cell containing the selection checkbox to never grow or shrink.
2. If we end up customizing via `columns`, I think we should omit `property` and use something like `selection: true`.

src/js/components/DataTable/Body.js
ShimiSun: "grommet data table". I chose a name that wouldn't have a potential collision with names that the user may select for new columns.

src/js/components/DataTable/DataTable.js
ShimiSun: Yea, that's a reserved name for the selected column. Currently, the 'selected' column doesn't have an entry on the `columns` prop, this PR is enabling styling for this column under this reserved name.

src/js/components/DataTable/README.md
ShimiSun: OK. I'll close this PR in this case.

PR #4825, opened 2020-12-11
Changed Select to add defaultValue

ericsoderberghp:
#### What does this PR do?

Changed Select to add defaultValue

doc.js

Added unit tests
White box test with stories

manually

automatic

yes

backwards compatible


src/js/components/Select/Select.js
```diff
@@ -92,7 +93,11 @@ const Select = forwardRef(
     const formContext = useContext(FormContext);
     // value is used for what we receive in valueProp and the basis for
     // what we send with onChange
-    const [value, setValue] = formContext.useFormInput(name, valueProp, '');
+    const [value, setValue] = formContext.useFormInput(
+      name,
+      valueProp,
+      defaultValue || '',
```
ShimiSun, line 99: I'm wondering if we should do value verification and make sure that defaultValue exists within the Select options and only apply the assignment if there is a match, and put a proper warning in case it doesn't. Thoughts?

ShimiSun (commented):
I dropped one small question.

src/js/components/Select/Select.js
ericsoderberghp: Well, we don't do that correlation with `value` and `options` currently. You can pass a `value` that isn't checked against the `options`.

ShimiSun (approved):

PR #4835, opened 2020-12-17
Update storybook from version 5.3.20 to 6.1.11

jcfilben:
#### What does this PR do?
Updates the storybook version to the latest, 6.1.11

This update includes changes to the CSF format for the stories. Previously the annotations looked like this:
```javascript
export const Basic = () => <Button />
Basic.story = {
  name: 'foo',
  parameters: { ... },
  decorators: [ ... ],
};
```
The above format is deprecated and the annotation should be in this format instead:
```javascript
export const Basic = () => <Button />
Basic.storyName = 'foo';
Basic.parameters = { ... };
Basic.decorators = [ ... ];
};
```
This PR fixes the stories to use the new format. This is where the majority of the changes come from. Also stories should no longer use `storiesOf`.

In the `storybook/preview.js` file I removed `options: { showRoots: true, }` because `showRoots` is deprecated and the behavior is now done by default and controlled within the `.storybook/manager.js` file instead of the `preview.js` file.

Also in `storybook/preview.js` I added `layout: 'fullscreen'` so the component takes up the full canvas instead of having a margin around it. The previous behavior in 5.3.20 was for the component to take up the full canvas.

More information about migrating to version 6 of storybook can be found in the [version 6 migration guide](https://storybook.js.org/blog/storybook-6-migration-guide/) as well as this more detailed [markdown file](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md).

Start with the `package.json` and `storybook/preview.js` files.
Then look the stories to see the changes from the old CSF annotation format to the new format.

storybook

storybook

No

No

Backwards compatible


ericsoderberghp (approved):

jcfilben:
I think we should roll back this PR since the code on the stories is not showing up. 

I did find a solution based on information found [here](https://github.com/storybookjs/storybook/issues/3566). But it requires moving `export default {title: ''};` into each story with a unique name. Currently we have stories files with multiple stories within so this requires some restructuring. I think one I have this part done the storybook version can be update to 6 at the same time as the story restructuring.

Issue #4840, opened 2020-12-20
Unable to type form values


mattoni:
### Expected Behavior

I should be able to pass a generic type to the form, i.e. `<Form<FormValueType>>`, in order to get proper typing for `onSubmit` etc

### Actual Behavior

It's not possible to type the form, and passing in external onSubmit handlers have type errors if you attempt to type the params manually.

### URL, screen shot, or Codepen exhibiting the issue

<img width="492" alt="image" src="https://user-images.githubusercontent.com/5110855/102720697-aed4fe80-42aa-11eb-9d19-592b099e9591.png">

<img width="989" alt="image" src="https://user-images.githubusercontent.com/5110855/102720703-bac0c080-42aa-11eb-8950-2139909dfb12.png">

### Steps to Reproduce

1. Attempt to pass a typed onSubmit handler to the <Form> component

### Your Environment

- Grommet version: 2.16.2

kenjdavidson:
You need to create a typed form. Don't mind my comments, but it took a while.

```
// Super intuitive and not documented anywhere but a commit message
// https://github.com/grommet/grommet/commit/ed480a02d0fb053e9466bc137541e436e1e77c15
const TypedCourseForm = Form as TypedForm<Course>;
```

But it works great.

mattoni:
That does indeed work, though it seems a bit redundant to do instead of what most other libraries do, i.e. `Form<FormProps>`. I've opened a PR to allow that method, in case there is more interest.  #4841

ShimiSun:
@mattoni Thank you for your contribution, the fix will be available on our next release (later than v2.16.3)

PR #4841, opened 2020-12-20
Add ability to type form component

mattoni:
Signed-off-by: Alexander Mattoni <5110855+mattoni@users.noreply.github.com>

#### What does this PR do?

This PR adds the ability to type Form components with the values expected.

```
// event.values is typed to InputTypes now.
<Form<InputTypes> onSubmit={event => console.log(event.values)} />
```

`src/js/components/Form/index.d.ts`

Confirmed the prop types work on <Form> components, and that it works as it used to if not typed.

Create a <Form> component and type it with an interface of expected input values. Then, ensure onSubmit event.values is typed properly.

#4840 

Probably

Backward compatible


ericsoderberghp:
Thanks for working on this. If I understand it correctly, if seems like a good issue to resolve. I'm trying to ensure I've wrapped my head around it all. Does the example below seem like the right way to think of it with a bit more context?
```
interface FormValue {
  name?: string;
  email?: string;
}

const MyComponent = () => {
  const [value, setValue] = React.useState<FormValue>({ name: '', email: '' });
  return (
    <Form<FormValue> value={value} onSubmit={({ value: nextValue }) => setValue(nextValue)}>
      ...
    </Form>
  );
}
```
where the `value` prop of Form gets fully typed, instead of `{}` as it is today, and the `onSubmit` `event.value` gets fully typed as well. `value` and `setValue` are already fully typed via `useState<FormValue>`.

mattoni:
@ericsoderberghp precisely! This allows us to make the form specific, and ensures type safety to all relevant properties.

ericsoderberghp:
> @ericsoderberghp precisely! This allows us to make the form specific, and ensures type safety to all relevant properties.

Great. Thanks for clarifying. In that case, can we also amend the type of `value?` to be `T` instead of `{}`?

ericsoderberghp:
Also, can we change the UncontrolledType story to do this, changing it from uncontrolled to controlled, to ensure all of the typings are handled correctly? I'm also thinking we wont need `FormExtendedEvent` inside that story anymore, as these changes should take care of it.

mattoni:
Yes. In that case though, we should probably make the type of `T` default to `{}`  instead of `unknown` for backward compatibility.

mattoni:
Sounds good, I'll try and get to that this week.

ShimiSun:
@mattoni hi buddy, thanks for working on this PR, any updates on incorporating the feedback?

mattoni:
Hi there, sorry I haven't had a chance yet but can look into it this week :)

mattoni:
Sorry this took awhile to get to, had an unusually turbulent month! I think this is in-line with all the changes requested, but please let me know if there's anything else needed.

ericsoderberghp (approved):

ericsoderberghp:
merged manually. Thanks for your contribution!

PR #4850, opened 2020-12-30
Fix ResponsiveGrid story occured on size argument with small

mxxnhc:
#### What does this PR do?
Fix storybook code in an example of ResponsiveGrid.

```javascript
let a = "arbitary string"
let b = "small" // This is problem
// We don't expect below is true, but it is
typeof a[b] === "function"
```

I'm not professional on javascript.
What I've figured out on this problem is that javascript string object has a method called `small`. So if we try `string["small"]`, it returns a `small` function that is method of string object in javascript.

This cause an error to someone who follows storybook example code of `ResponsiveGrid`.
This error only happens when the responsive size is small.

No. It is not a problem of grommet. It's just example.

No.

No effect on main code stream I guess.

### DIFF
--- a/src/js/components/Grid/stories/Responsive.js
+++ b/src/js/components/Grid/stories/Responsive.js
@@ -114,14 +114,14 @@ const Responsive = ({
       // Take into consideration if not array is sent but a simple string
       let columnsVal = columns;
       if (columns) {
-        if (columns[size]) {
+        if (columns[size] && typeof columns !== "string") {
           columnsVal = columns[size];
         }
       }
 
       let rowsVal = rows;
       if (rows) {
-        if (rows[size]) {
+        if (rows[size] && typeof columns !== "string") {
           rowsVal = rows[size];
         }
       }


jcfilben:
For this particular example I don't think `&& typeof columns !== "string"` needs to be added as a check because in the Responsive Grid story columns is defined as 
``` javascript
const columns = {
  small: ['small'],
  medium: ['auto', 'auto'],
  large: ['auto', 'auto', 'auto'],
  xlarge: ['auto', 'auto', 'auto'],
};
```
It is an object not a string. In another example if you wanted to set columns to a string such as `columns = 'small'` then you would just change `columnsVal = columns[size]` to `columnsVal = [columns]`. But I don't think any changes need to be made to this example.

mxxnhc:
Thanks you for the reply! :smiley: 
I misunderstood. I thought this code block means it considers a simple string to the column argument.
```javascript
      // Take into consideration if not array is sent but a simple string
      let columnsVal = columns;
      if (columns) {
        if (columns[size]) {
          columnsVal = columns[size];
        }
      }
```
Rather, it means that the column object that can have a string or an array for each size key value.

Then I have a question, do we must put an object to the argument even though we want same shape for each grid size? for example,
```javascript
const rows = {
  small: ['small'],
  medium: ['small'],
  large: ['small'],
  xlarge: ['small'],
};
```
If there is an better way, plz let me know. It will be helpful to understand and use this toolkit.

I close this PR cuz no longer needs

PR #4861, opened 2021-01-05
Refined Select stories

ShimiSun:
#### What does this PR do?
Refined Select stories

1. Removed Dark story and replace it with a unit test.
2. Renamed other existing stories to be more aligned with prop names
3. Cleaned up imports
any select story changed
storybook + jest
storybook
Post discussion on Select stories
None
N/A
Nope
No
b/c


ericsoderberghp:
A few thoughts from just looking at Storybook:
1. How about combining "Themed" and "Custom search"?
2. How about splitting disabled out of "Simple"?
3. How about aligning story names more with property names, such as "Children render" -> "Children"?
4. Can we remove "Object options with search and multiple" since it isn't focused on one property?
5. Can we simplify "Clear"? It seems a bit too complex.
6. Can we remove the custom theming from "Create option", since theming is a separate aspect?
7. Can we improve the styling of "Value label"? I don't think we'd ever recommend having it look like a Button.

ericsoderberghp (approved):

PR #4864, opened 2021-01-06
Allow onEsc to be applied to Layer when modal={false}

### LINKED ISSUE #3771: Grommet Layer onEsc and onClickOutside not working when modal is set false

taysea:
#### What does this PR do?
Allows onEsc to be applied to Layer when modal={false} by tracking the keyboard event on the document in cases where modal={false}. An adjustment was made to Keyboard.js and DropContainer is order to allow the DropContainer to capture the keyboard event before the Layer. Documentation on capture=true can be found here: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

This is needed so that if a Drop opens inside a layer, the user can hit `esc` without the Layer closing as well.

Tested in Layer Target story by adding a Select to the Layer content.Esc could be used to close the Select and then used again to close the Layer.

A jest test has been added as well to test the same scenario.

Same as noted above.

onEsc was noting working when modal={false}

Closes #3771 

No.

Yes.

Backwards compatible.


src/js/components/Drop/DropContainer.js
```diff
@@ -338,6 +338,9 @@ const DropContainer = forwardRef(
           trapFocus={trapFocus}
         >
           <Keyboard
+            // should capture keyboard event before other elements,
+            // such as Layer
+            capture
```
taysea, line 343: add link to documentation on addEventListener

src/js/components/Layer/LayerContainer.js
```diff
@@ -198,7 +198,21 @@ const LayerContainer = forwardRef(
     }
 
     if (onEsc) {
-      content = <Keyboard onEsc={onEsc}>{content}</Keyboard>;
+      content = (
+        <Keyboard
+          onEsc={
+            onEsc
+              ? event => {
+                  event.stopPropagation();
```
taysea, line 206:
```suggestion
                  // prevent event from propagating upward to any parent of Layer
                  event.stopPropagation();
```

ShimiSun (approved):
Thank you! 💟

Issue #4855, opened 2021-01-04
Custom Select - Select with a custom Footer


padmacnu:
@ShimiSun We have requirement where we want to customize the Select component to render something at bottom. I have mocked the Grommet select to show the requirement. It's not an option, should be outside `Select` element. Similar to `Search bar`

![image](https://user-images.githubusercontent.com/6428437/103583245-3ff3ca00-4e94-11eb-8fc2-dcaeca7e6379.png)

ericsoderberghp:
Can you provide a real world example of what you would like to put there? That will help us ensure that we don't miss something. Of note, are you thinking the custom JSX might affect the width of the drop?

padmacnu:
@ericsoderberghp Here is screenshot of live example:
![image](https://user-images.githubusercontent.com/6428437/103685438-14c7b400-4f42-11eb-97b3-667119461bf4.png)

ShimiSun:
Hi @padmacnu, thanks for filing this issue, the requested functionality could currently be achieved by Select `children`, please take a look at a POC I've prepared and let me know what you think https://deploy-preview-4869--sad-tereshkova-173d07.netlify.app/?path=/story/input-select--footer-option

ericsoderberghp:
For the example provided, I think `searchPlaceholder` would be a more appropriate way to approach providing the "Type ..." guidance on using the search control within Select. Did you consider that?

padmacnu:
Thanks for the POC!

padmacnu:
> For the example provided, I think `searchPlaceholder` would be a more appropriate way to approach providing the "Type ..." guidance on using the search control within Select. Did you consider that?

I will check with Scott, thanks @ericsoderberghp

PR #4869, opened 2021-01-07
POC - Added a Select with footer option story example

ShimiSun:
#### What does this PR do?
Added a Select with footer option story example
FooterOption.js
storybook+jest

This PR is a POC for https://github.com/grommet/grommet/issues/4855
The POC shows that Footer rendering could be achieved without additional enhancements to Select.

No
No
b/c

### DIFF
--- a/src/js/components/Select/stories/FooterOption.js
+++ b/src/js/components/Select/stories/FooterOption.js
@@ -1,0 +1,55 @@
+import React from 'react';
+
+import { Box, Text, Grommet, Select } from 'grommet';
+import { grommet } from 'grommet/themes';
+
+const Option = React.memo(({ label, isFooter }) => (
+  <>
+    {isFooter ? (
+      <Box border="top">
+        <Text margin="small">{label}</Text>
+      </Box>
+    ) : (
+      <Box margin="small">{label}</Box>
+    )}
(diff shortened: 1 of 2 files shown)


src/js/components/Select/stories/FooterOption.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+
+import { Box, Text, Grommet, Select } from 'grommet';
+import { grommet } from 'grommet/themes';
+
+const Option = React.memo(({ label, isFooter }) => (
+  <>
+    {isFooter ? (
+      <Box border="top">
```
ericsoderberghp, line 9: Can we do anything to cause the text to wrap as opposed to increasing be drop width? The single line message looks a bit long to me. Perhaps use `<Paragraph />`? Just wondering.

ericsoderberghp (commented):
I'm on the fence about whether this is an approach we'd want to recommend. It feels a bit of a hack to include within `options` and with `disabled`.

For the use case desired, it seems like putting a placeholder message in the search field, via `searchPlaceholder` would be a better approach.

ShimiSun:
POC solved issue #4855, closing this PR.

Issue #4865, opened 2021-01-07
Tooltip showing on disabled button with custom label


Jam42:
When using Tip with disabled Button with custom label tooltip still appears

### Expected Behavior

Tooltip shouldn't appear

### Actual Behavior

Tooltip is appearing – and not hiding when mouse leave it

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-forked-wlr22?file=/index.js

### Steps to Reproduce

1. Add Button with custom label and disable it
2. Wrap it with Tip
3. Hover over it

### Your Environment

- Grommet version: 2.16.2
- Browser Name and version: Brave v1.16.68 Chromium: 86.0.4240.111
- Operating System and version (desktop or mobile): macOS Big Sur

ShimiSun:
I think the issue is with onMouseLeave of Button, and not with the Tip itself.

sabarnix:
We have a use case to show tooltip on disabled button. We want to show the reason of disabling on tooltip. This could be a common use case. Can we allow tooltip on disabled buttons?

jcfilben:
Hey @sabarnix currently we don't allow a tooltip to show on a disabled button because that would be an accessibility violation. People who use assistive technology wouldn't be able to interact with the tooltip.

Issue #4846, opened 2020-12-26
Button.onMouseOut is applied to the children of the button (icon/text) not to the button itself 


Samuel-Lewis:
When using the Button's event listener `onMouseOut`, it seems to be applied to all of the children of the button, causing errant calls.

Eg: For a text with an icon and label, mousing "out" of just the label (while staying within the button bounds) will fire `onMouseOut`. Similarly, moving over SVG elements of an icon will also trigger the change.

### Expected Behavior

Expect `onMouseOut` to only fire once the mouse leaves the bounds of the button.

### Actual Behavior

`onMouseOut` is every time the mouse leaves the element and any of the child elements.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-forked-nbjwt?file=/index.js

https://user-images.githubusercontent.com/12661033/103151138-5a121600-47cb-11eb-946a-ed7f2bc67dd2.mov

### Your Environment

- Grommet version: 2.16.2
- Browser Name and version: Google Chrome 87.0.4280.88
- Operating System and version (desktop or mobile): Desktop, macOS Catalina 10.15.7

Samuel-Lewis:
It's also possible that I have just misunderstood the difference between `onMouseOut` and `onMouseLeave`. But I would assume applying the `onMouseOut` to the Button would only apply it to the button layer, and not the children.

On the flip side, `onMouseLeave` never seems to fire 😢 . I'll make a seperate issue for that if needed.

jcfilben:
Hi @Samuel-Lewis, `onMouseOut` should fire when the mouse leaves the element or moves out of one of the children of the element, so `onMouseOut` for button is working as expected. `onMouseLeave` should only fire when the mouse leaves the element not when it leaves a child of the element. However, Button doesn't currently have an `onMouseLeave` event which is why it is not firing

jcfilben:
Here is a codesandbox using grommet v2.16.3 showing onMouseLeave and onMouseOut working as expected for button https://codesandbox.io/s/grommet-v2-template-forked-7nl1y?file=/index.js. I think this issue can be closed.

PR #4872, opened 2021-01-09
Fix: add onMouseLeave prop to button to fix bug with tip

netochaves:
#### What does this PR do?
Fix bug with tip on button using a element as label.
`src/js/components/Button/Button.js`
Unit and manually tests.
With storybook
No
#4865
No
backwards compatible

### DIFF
--- a/src/js/components/Button/Button.js
+++ b/src/js/components/Button/Button.js
@@ -109,6 +109,7 @@ const Button = forwardRef(
       onBlur,
       onClick,
       onFocus,
+      onMouseLeave,
       onMouseOut,
       onMouseOver,
       plain,
@@ -202,8 +203,9 @@ const Button = forwardRef(
 
     const onMouseOutButton = event => {
       setHover(false);
-      if (onMouseOut) {
-        onMouseOut(event);
+      const outEvent = onMouseLeave || onMouseOut;
+      if (outEvent) {
+        outEvent(event);
       }
     };
 
--- a/src/js/components/Button/__tests__/Button-test.js
+++ b/src/js/components/Button/__tests__/Button-test.js
@@ -159,6 +159,21 @@ describe('Button', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
+  test('tip with label using mouseLeave', () => {
+    const { container, queryByText } = render(
+      <Grommet>
+        <Button label={<Text> Label </Text>} onClick={() => {}} tip="tooltip" />
+      </Grommet>,
+    );
+
+    fireEvent.mouseOver(queryByText('Label'));
+    expect(queryByText('tooltip')).toBeTruthy();
+    fireEvent.mouseLeave(queryByText('Label'));
+    expect(queryByText('tooltip')).toBeFalsy();
+
+    expect(container.firstChild).toMatchSnapshot();
+  });
+
   test('disabled', () => {
     const component = renderer.create(
       <Grommet>
--- a/src/js/components/Button/__tests__/__snapshots__/Button-test.js.snap
+++ b/src/js/components/Button/__tests__/__snapshots__/Button-test.js.snap
@@ -3149,6 +3149,107 @@ exports[`Button tip 1`] = `
 </div>
 `;
 
+exports[`Button tip with label using mouseLeave 1`] = `
+.c1 {
+  display: inline-block;
+  box-sizing: border-box;
+  cursor: pointer;
+  font: inherit;
+  -webkit-text-decoration: none;
+  text-decoration: none;
+  margin: 0;
+  background: transparent;
+  overflow: visible;
+  text-transform: none;
+  border: 2px solid #7D4CDB;
+  border-radius: 18px;
+  color: #444444;
+  padding: 4px 22px;
+  font-size: 18px;
+  line-height: 24px;
+  -webkit-transition-property: color,background-color,border-color,box-shadow;
+  transition-property: color,background-color,border-color,box-shadow;
+  -webkit-transition-duration: 0.1s;
+  transition-duration: 0.1s;
+  -webkit-transition-timing-function: ease-in-out;
+  transition-timing-function: ease-in-out;
+}
+
+.c1:hover {
+  box-shadow: 0px 0px 0px 2px #7D4CDB;
+}
+
+.c1:focus {
+  outline: none;
+  box-shadow: 0 0 2px 2px #6FFFB0;
+}
+
+.c1:focus > circle,
+.c1:focus > ellipse,
+.c1:focus > line,
+.c1:focus > path,
+.c1:focus > polygon,
+.c1:focus > polyline,
+.c1:focus > rect {
+  outline: none;
+  box-shadow: 0 0 2px 2px #6FFFB0;
+}
+
+.c1:focus::-moz-focus-inner {
+  border: 0;
+}
+
+.c2 {
+  font-size: 18px;
+  line-height: 24px;
+}
+
+.c0 {
+  font-size: 18px;
+  line-height: 24px;
+  box-sizing: border-box;
+  -webkit-text-size-adjust: 100%;
+  -ms-text-size-adjust: 100%;
+  -moz-osx-font-smoothing: grayscale;
+  -webkit-font-smoothing: antialiased;
+}
+
+@media only screen and (max-width:768px) {
+
+}
+
+@media only screen and (max-width:768px) {
+
+}
+
+@media only screen and (max-width:768px) {
+
+}
+
+@media only screen and (max-width:768px) {
+
+}
+
+@media screen and (-ms-high-contrast:active),(-ms-high-contrast:none) {
+
+}
+
+<div
+  class="c0"
+>
+  <button
+    class="c1"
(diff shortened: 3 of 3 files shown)


src/js/components/Button/Button.js
```diff
@@ -202,8 +203,9 @@ const Button = forwardRef(
 
     const onMouseOutButton = event => {
       setHover(false);
-      if (onMouseOut) {
-        onMouseOut(event);
+      const outEvent = onMouseLeave || onMouseOut;
```
ShimiSun, line 206: What are your thoughts on handling `onMouseLeave` separately from `onMouseOut`?

ShimiSun (commented):
Thanks for the PR buddy, good to see you back 💟 
I dropped a small question, let me know what you think.

src/js/components/Button/Button.js
netochaves: So @ShimiSun, `onMouseLeave` is used by `Tip` but button expects to receive an `onMouseOut` and I decided to stay with both because I was worried about backwards compatible, I saw that both are used in other files. If I remove `onMouseOut` some components that send it to `Button` could break.  

What do you think?

ShimiSun: I think it makes sense.  Thank you for the detailed answer.

jHwls:
Thanks for work on this fix! Dealing with this exact issue right now.

Noticing that the issue with binding `onMouseLeave` to the `onMouseOut` event is that that event is firing many times while the cursor is still over the button. You can see this on the storybook deploy preview as the tooltip now flickers when moving the cursor over the button.

If you bind to `onMouseLeave` the Tip functions as expected. Unsure how other the components sending onMouseOut to Button @netochaves mentions would be affected by this however.

netochaves:
Thanks for the comment @jHwls 
I was able to reproduce the issue with hover event firing many times but I notice that this only happens with SVG elements like the  [Caret story](https://deploy-preview-4872--sad-tereshkova-173d07.netlify.app/?path=/story/controls-tip--caret) other stories like [Simple](https://deploy-preview-4872--sad-tereshkova-173d07.netlify.app/?path=/story/controls-tip--simple) don't have this behavior.

I did some research and this is an old problem on react, you can see more here: https://github.com/facebook/react/issues/4492

A workaround is to add `pointerEvents = 'none'` on the SVG element, e.g
```js
<Button icon={<Trash pointerEvents="none" />} plain={false} />
```

I think that we don't have much to do about this problem on the grommet repo.

ShimiSun (approved):

jHwls:
> I did some research and this is an old problem on react, you can see more here: [facebook/react#4492](https://github.com/facebook/react/issues/4492)

Wow that's a great find @netochaves ! Never run into that before. Is it worth adding this prop to the storybook examples that are affected at least, or to the docs somewhere, as it seems like others might run into the same issue?

src/js/components/Button/Button.js
ericsoderberghp: Given that the events are actually different, based on https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseleave_event, it doesn't seem correct to be merging the two here. I think having both wired up would allow the caller to get either behavior they want correctly.

netochaves: I agree @ericsoderberghp that this is not the best solution, but I couldn't figure out how to fix this without merging them. If I left only `onMouseOut` the problem will continue, btw the MDN docs that you sent it's a great resource to understand why the problem happens in the first place. 

If I left just `onMouseLeave` this could break some apps that are sending a custom `onMouseOut` function or some components on grommet that use this prop, like `AccordionPanel` .

I'm open for better ideas on how to handle this, thanks

ericsoderberghp: From Button's point of view, it should pass along either or both of those listeners and caller's should be responsible to listen as needed. Is that not viable?

taysea: I noticed in the Codesandbox demo on the original issue (https://codesandbox.io/s/grommet-v2-template-forked-wlr22?file=/index.js) that the tooltip on the button with custom label only appears if I hover directly over the label but not if I just hover somewhere else on the disabled button. This makes me think the issue is rooted in how Button states are passed along to custom elements, like a custom label.

Is it possible instead to make sure that if button is `disabled` that no mouse events are registered on any of the button's children/custom elements? It seems that as long as the mouseover is not registered on the custom label, then we wouldn't be seeing this issue. If we could do that, we wouldn't need to merge the `onMouseOut` and `onMouseLeave` events.

netochaves: @ericsoderberghp Do you mean pass both props `onMouseOver` and `onMouseOut` to button? I tried that and didn't work out, has the same bug behaviour of the demo.

netochaves: @taysea To do this I think we will have to use something like `cloneElement` to pass information to a custom element in label or something else, I don't know if this is viable and we also have to know if it is a react element or no wich we don't need RN. See code bellow on how is implemented today.
https://github.com/grommet/grommet/blob/b68eac70c8e2c6a6b0013ac9f4c606dca6c6164a/src/js/components/Button/Button.js#L227-L244

ericsoderberghp: I was thinking Button would support either `onMouseOver` or `onMouseOut` or both and that Tip would use whatever was appropriate.

taysea: Thanks for the reply @netochaves -- maybe exploring Eric's proposal would be more viable given the current implementation.

netochaves: @ericsoderberghp I said `onMouseOver` but I meant `onMouseLeave` if I pass that event and the `onMouseOut` the tip behaviour continues like the issue demo.

jcfilben (commented):
Currently with this solution the tooltip is still showing up on the disabled button with custom label and onMouseLeave is not the expected behavior. What if instead we did something like this:
In Tip.js change 
```javascript
const clonedChild = cloneElement(child, {
    onMouseOver: () => setOver(true),
    onMouseLeave: () => setOver(false),
```
to
```javascript
const clonedChild = cloneElement(child, {
    onMouseEnter: () => setOver(true),
    onMouseLeave: () => setOver(false),
```

and then in Button.js have both onMouseOut and onMouseLeave
```javascript
    const onMouseOutButton = event => {
      setHover(false);
      const outEvent = onMouseOut;
      if (outEvent) {
        outEvent(event);
      }
    };

    const onMouseLeaveButton = event => {
      setHover(false);
      const outEvent = onMouseLeave;
      if (outEvent) {
        outEvent(event);
      }
    };
```
...
```javascript
onMouseOut={onMouseOutButton}
onMouseLeave={onMouseLeaveButton}
```
I believe this would solve both #4846 and #4865 without compromising standard behavior for `onMouseLeave`. Here are the storybook examples I used to test these:
```javascript
import React from 'react';

import { grommet, Grommet, Button, Box, Text } from "grommet";
import { Copy } from "grommet-icons";

export const TEST = () => {
    const [outEvents, setOutEvents] = React.useState(0);

    function onOut() {
        setOutEvents(outEvents + 1);
    }

    return(
        <Grommet theme={grommet}>
        <Box pad="large" align="center">
            <Button
            icon={<Copy />}
            size="large"
            label="Example button"
            onMouseLeave={onOut}
            />
            <Text>Out events: {outEvents}</Text>
        </Box>
        </Grommet>
    );
};

export default {
  title: 'Controls/Button/TEST',
};
```

```javascript
import React from 'react';

import { grommet, Text, Box, Button, Grommet, Tip } from "grommet";

export const TEST2 = () => (
    <Grommet theme={grommet}>
    <Box pad="small" gap="xlarge">
      <Tip content="Tooltip will hide as expected">
        <Button label="Button" />
      </Tip>
      <Tip content="Tooltip will not render">
        <Button label="Disabled button" disabled />
      </Tip>
      <Tip content="Tooltip will stay here">
        <Button
          label={<Text>Disabled button with custom label</Text>}
          disabled
        />
      </Tip>
    </Box>
  </Grommet>
);

export default {
  title: 'Controls/Button/TEST 2',
};
```

netochaves:
Hi @jcfilben thanks for you proposal, I tried something like that but I din't use `onMouseEnter` and now I see why din't worked out :smile: 

But, we still will have a problem in another implementations with tip, in specific implementations that use SVG elements, because with `onMouseEnter`  when your mouse is over the icon the tooltip will not appears, it will only appears when is over the container of the icon, see gifs bellow:

With onMouseOver:

https://user-images.githubusercontent.com/19830660/107517312-5b857b00-6b8c-11eb-8b25-43643a4180f5.mp4

With onMouseEnter:

https://user-images.githubusercontent.com/19830660/107517343-68a26a00-6b8c-11eb-9770-c4681b77d454.mp4

jcfilben:
Hi @netochaves I created a pull request switching Tip to use onMouseEnter and I believe it is working with the button with custom label case and the tip with SVG element case. You can take a look at it here https://github.com/grommet/grommet/pull/4970

ShimiSun:
Thank you @netochaves and @jcfilben 💟

PR #4874, opened 2021-01-10
Menu control position according Drop mirroring

### LINKED ISSUE #3540: List actions sometimes aligns with wrong list item
### Expected Behavior

There are two approaches I see, but in both I think the "..." should remain aligned with the list item clicked. The two ideas I have are:
1. Leave the action items below the "...", and maybe the page scrolls down to make space when this drop is opened.
2. Move the action items so they appear above the "...". This would essentially be making the control button ("...") on the bottom of the drop.

### Actual Behavior

When the actions icon of a list item at the bottom portion of a list is clicked, the drop shifts up so the entire drop fits in the window. The issue is that this makes the "..." align with a previous list item. This could cause confusion to the user about which list item they had clicked.

In this screenshot, I had clicked on entry-9, but you can see how it looks as though I would've clicked entry-8.

Screen Shot 2019-11-08 at 12 54 08 PM

### Steps to Reproduce

1. https://storybook.grommet.io/?path=/story/list--action
2. Click the "..." on entry-9

### Your Environment

- Grommet version: 2.7.11
- Browser Name and version: Firefox 70.0
- Operating System and version (desktop or mobile): macOS Mojave 10.14.6

leossantiago:
Right now, Menu doesn't have a proper way to update Menu Control button when `Drop` is mirrored:

Issue3540
(this can be reproduced on https://storybook.grommet.io/?path=/story/visualizations-list--action)

Even though it's on List stories, this problem is caused by `Menu`.
When using a `Menu` , it renders this sequence of components: Menu -> DropButton -> Button ->  Drop -> DropContainer. 

While Menu control button (three dots) position changes according to Menu `align` prop (which defaults to `top: 'top'` vertically, but can be set as `bottom: 'bottom'`), `DropContainer` calculates remaining space on view and forces a mirrored position if there's not enough space to allocate the container. And since those actions are not combined, it results on the problem showed above: a mirrored Drop with Menu-control button on top.

#### What does this PR do?

Fixes #3540 by keeping track of `DropContainer` align on `Menu`, forcing a change on control button position when Drop is mirrored.

Issue3540_2

setState function is passed down to DropContainer, and it changes according to its align prop whenever there's an update to DropContainer style.

This PR also fixes a typing problem on `List` `pad` prop, where `'none'` wasn't being recognized as a valid type (`pad={{ left: 'small', right: 'none' }}` on example below).

src/js/components/Drop/DropContainer.js

Manual testing using code below.

```js
  <Grommet theme={grommet}>
    <Box pad="large">
      <List
        data={data.slice(0, 10)}
        pad={{ left: 'small', right: 'none' }}
        action={(item, index) => {
          return (
            <Menu
              dropAlign={{ top: 'top', left: 'left' }}
              key={index}
              icon={<More />}
              hoverIndicator
              items={[{ label: 'one' }]}
            />
          );
        }}
      />
    </Box>
  </Grommet>
```

Issue #3540 

GIFs above.

No.

Breaking change if you consider position change when mirrored. Otherwise, backwards compatible.


ShimiSun:
@leossantiago we missed you buddy!! I'll try to review the PR today!

src/js/components/List/stories/Action.js
```diff
@@ -17,7 +17,8 @@ export const Action = () => (
     <Box pad="large">
       <List
         data={data.slice(0, 10)}
-        pad={{ left: 'small', right: 'none' }}
+        pad={{ left: 'large', right: 'small' }}
+        // pad={{ left: 'small' }}
```
ShimiSun, line 21: Why is this comment needed?

ShimiSun:
> Breaking change if you consider position change when mirrored. Otherwise, backward compatible.

Good point, I think that it would make sense for Menu to open in the way you have suggested. I'll take a few hours to think of counter examples.

src/js/components/Menu/Menu.js
```diff
@@ -275,7 +281,9 @@ const Menu = forwardRef((props, ref) => {
             onEnter={onSelectMenuItem}
           >
             <ContainerBox background={dropBackground || theme.menu.background}>
-              {align.top === 'top' ? controlMirror : undefined}
+              {menuMirrorReference === 'top' && align.top === 'top'
```
taysea, line 284: Thinking functionality looks great. Just wondering if this variable could have a more clear name.

maybe, `alignControlMirror` and `setAlignControlMirror` ?

src/js/components/List/stories/Action.js
leossantiago: Thanks for pointing that! I was using this story as a playground and didn't realize this test was added to the commit. I'll remove it on the next commit. :smile:

leossantiago:
@ShimiSun feels good to be back. Missed this community as well :smile: 
I'll be addressing reviews soon, and we can discuss counter examples when you come up with any.

src/js/components/Drop/DropContainer.js
```diff
@@ -62,6 +63,13 @@ const DropContainer = forwardRef(
     ]);
     const dropRef = useRef();
     useEffect(() => {
+      const updateMenuMirrorReference = () => {
+        const styleCurrent = (ref || dropRef).current.style;
+        const sideToAlign = styleCurrent.top !== '' ? 'top' : 'bottom';
```
taysea, line 68: soft opinion for variable name.

```suggestion
        const alignControl = styleCurrent.top !== '' ? 'top' : 'bottom';
```

src/js/components/Menu/Menu.js
leossantiago: I'm open to any suggestions. I must confess I spent a couple minutes renaming and trying new names before committing so it would be more clear haha. I'll be pushing a new commit to remove a comment that wasn't supposed to be there and guess  we'll go with `alignControlMirror` and `setAlignControlMirror`. :smile:

leossantiago:
Addressed comments :tada:

ShimiSun (approved):

src/js/components/Drop/DropContainer.js
```diff
@@ -37,6 +37,7 @@ const DropContainer = forwardRef(
   (
     {
       align = defaultAlign,
+      setAlignControlMirror,
```
ericsoderberghp, line 40: Can we call this `onAlign`? I think that will be more consistent and generic, caller is notified when the actual alignment changes.

ericsoderberghp (commented):
Overall, this looks good to me. Just one small suggestion for better consistency.

leossantiago:
@ericsoderberghp updated props name on `DropContainer` and `DropButton` but kept it as it is on `Menu` so the state stays self-explanatory since `Menu` is the one handling the change of position. What do you think?

Also, on `DropContainer` I kept the whole updating function as `updateAlignControlMirror`. I'd like your opinion on that as well!

src/js/components/Drop/DropContainer.js
```diff
@@ -62,6 +63,13 @@ const DropContainer = forwardRef(
     ]);
     const dropRef = useRef();
     useEffect(() => {
+      const updateAlignControlMirror = () => {
```
ericsoderberghp, line 66: Agreed with leaving Menu as is. But, can we call this function something like "notifyAlign", since DropContainer doesn't know anything about a "control mirror".

src/js/components/Drop/DropContainer.js
```diff
@@ -225,6 +233,7 @@ const DropContainer = forwardRef(
             container.style.maxHeight = `${maxHeight}px`;
           }
         }
+        updateAlignControlMirror();
```
ericsoderberghp, line 236: Can we move the `if (onAlign)` check here, since it won't do anything unless the caller passes it in?

ericsoderberghp (commented):
One more small naming suggestion.

ShimiSun:
@leossantiago hi buddy, I would love to get this PR to the upcoming release, any chance you can address the comments today? Thank you for working on it 💟

src/js/components/Drop/DropContainer.js
leossantiago: Agreed! Addressed!

src/js/components/Drop/DropContainer.js
leossantiago: Agreed! Addressed!

leossantiago:
Hey @ShimiSun! Sorry for not addressing that earlier. It was already late night around here when you tagged me here 😅 
Addressed Eric's suggestions. Hope it's still on time. Ping me if something else should be changed! 🎉

ericsoderberghp (approved):

ShimiSun:
Thank you @leossantiago! I'm glad this fix will be available on the release. 💟

Issue #4867, opened 2021-01-07
TextInput renders styled placeholders even when there is a value


MikeKingdom:
A TextInput that has a styled placeholder (`placeholder={<someJSX>}` rather than `placeholder="somestring"`) renders the placeholder even when the textinput has a value, at least until another render happens. 

### Expected Behavior

If a TextInput has a value (either from a value prop or by typing text in the input), the placeholder should not be shown.

### Actual Behavior

One first render the placeholder will be drawn whether or not the input has a value.
On next render, if the input has a value the placeholder will not be rendered. Note that typing in the text field does not cause a re-render so the typed text will overlap the placeholder.

### URL, screen shot, or Codepen exhibiting the issue

![image](https://user-images.githubusercontent.com/12054362/103911440-487e1780-50c3-11eb-97cd-bd6465f832be.png)

### Steps to Reproduce
```jsx
<Grommet>
    <h1>some content</h1>
    <TextInput
      data-testid="styled-placeholder"
      id="styled-placeholder"
      name="styled-placeholder"
      placeholder={<Text>placeholder text</Text>}
    />
</Grommet>
```

This happens because on first render, the inputRef is not defined yet and so this check in TextInput's render returns true even though a value was specified:
```javascript
const showStyledPlaceholder =
      placeholder &&
      typeof placeholder !== 'string' &&
      !(inputRef.current && inputRef.current.value);
```

There is also nothing that causes a re-render on typing text. Some onChange/setValue type of code is necessary to cause the input to re-render on input.

Note that there is a test case in TextInput-test that tries to make sure the placeholder doesn't get rendered but it only works if something causes the input to get rendered twice.
 
### Your Environment

- Grommet version: 2.16.2
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): Windows10

ShimiSun:
Might be related to https://github.com/grommet/grommet/pull/3659

Issue #2556, opened 2018-12-07
Leverage the new hooks API to replace the context API


oorestisime:
Hooks will be released sometime at Q1 2019 and besides the excitement around them Grommet could simplify some of its internal core by leveraging them.

Instead of having multiple context providers, which seem to also slow down https://github.com/styled-components/styled-components/issues/2215 , the hooks API (useContext, https://reactjs.org/docs/hooks-overview.html#-other-hooks) could come in handy to simplify the codebase (avoiding to always needing to wrap code in a consumer to get theme or responsive) and improve user experience.

While i dont propose to convert all at once i think a nice step forward would be to experiment on a side-branch by just converting the responsive context to be used with hooks. this would give the opportunity to test them and if desired proceed to incremental changes in the other components.

oorestisime:
So this will be kind of a learning experience for me but i am willing to give this a go in another branch to see how it behaves.  i ll just modify the responsive context and the components that are using it in order to have something functional and a base for further testing!

dheerajvs:
@ericsoderberghp Now that hooks are released, is there a plan to leverage them?

ericsoderberghp:
We've been using this pattern in apps built with grommet recently:
```
...
import { ResponsiveContext } from 'grommet';
...
const MyComponent = props => {
    const responsive = React.useContext(ResponsiveContext);
    ...
}
```
So, you can use the `useContext()` hook with grommet today.

I think the intent of this issue is to use hooks internally in the grommet implementation more. @ShimiSun and I discussed migrating grommet code, both the components and the stories, to use hooks over time. Any help in this effort would be most welcome.

alexlabrioche:
@ericsoderberghp  hello Eric, can we see an exemple of how to use the responsive context ? without Hooks I must wrap my component around the ResponsiveContext.consumer every time i need (but I'm pretty sure it's not the good way) 
And maybe an example with the new useContext Hooks ?

thanks so much

ericsoderberghp:
I would have thought that the above code snippet would be a useful example. Can you clarify a bit more on what else you'd like to see?

MikeWalrath:
Component left for refactoring:
- Grommet.js

@MikeKingdom

PR #4876, opened 2021-01-12
Refactor Grommet component as functional with hooks

MikeKingdom:
The Grommet component is the only component remaining that hasn't yet been refactored to be functional and useHooks instead of a higher order component.

#### What does this PR do?
Refactor Grommet to be functional and use hooks. (the last remaining piece of #2556)
 
Improves a few test cases which fail due to the component not getting re-rendered before the test renderer takes a snapshot.

Also fixed to TextInput to render the first time correctly when a styled placeholder was specified along with an initial value. Previously it would render both the placeholder and value on top of each other until something causes the component to re-render. Note that it still won't remove the placeholder as the user types characters until something causes the onChange to trigger (like a focus change, etc.) This is captured in a different issue #4867

Grommet.js

Individual testing with Storybook and custom examples to test theme and themeMode changes and window size changes.
 
Storybook cases

#2556 

No

Don't think so.

Should be backward compatible.


src/js/components/Grommet/Grommet.js
```diff
@@ -21,144 +21,88 @@ const FullGlobalStyle = createGlobalStyle`
...
+    if (/Mobile|iPhone|Android/.test(userAgent)) {
+      return getDeviceBreakpoint('phone', theme);
+    }
+    return getDeviceBreakpoint('computer', theme);
   }
+  return undefined;
+};
 
-  state = {};
+const Grommet = props => {
```
MikeKingdom, line 45: One thought that just occurred to me...does this need a forwardRef?

ShimiSun: Yes, good catch, we try to have forwardRef on all grommet component.

src/js/components/Form/__tests__/Form-test-uncontrolled.js
```diff
@@ -127,25 +127,31 @@ describe('Form uncontrolled', () => {
...
   test('errors', () => {
-    const component = renderer.create(
-      <Grommet>
-        <Form errors={{ test: 'missing' }}>
-          <FormField name="test" />
-        </Form>
-      </Grommet>,
-    );
+    let component;
+    renderer.act(() => {
```
ericsoderberghp, line 131: Can you add a brief comment on why we need `act()`? I suspect we might forget.

ericsoderberghp (approved):

src/js/components/Form/__tests__/Form-test-uncontrolled.js
MikeKingdom: Done

src/js/components/Grommet/Grommet.js
MikeKingdom: done

src/js/components/Form/__tests__/Form-test-uncontrolled.js
ShimiSun: My comment on `act` did not come through the first time, so I'll try again - sorry about that.

You might be able to clean up the usage of `act` and `component = renderer.create(...)` with a more updated API of jest and '@testing-library/react', which is simply using the concept of `render` as follows

```
  const { container } = render(
      <Grommet>
         ...
      </Grommet>,
    );
```
    
`act` and `renderer` are limited in their functionality, and will probably be deprecated at one point, it is better not to introduce any new ones.

MikeKingdom: I did actually try that first and it didn't help but I can double check and see if that's doable.

MikeKingdom: Done

src/js/components/Tabs/__tests__/__snapshots__/Tabs-test.js.snap
```diff
@@ -469,54 +469,44 @@ exports[`Tabs Tab 1`] = `
...
-        aria-selected={false}
-        className="c3"
-        onBlur={[Function]}
-        onClick={[Function]}
-        onFocus={[Function]}
-        onMouseOut={[Function]}
-        onMouseOver={[Function]}
+        aria-expanded="false"
+        aria-selected="false"
+        class="c3"
```
ShimiSun, line 501: Seems like this button has missing function props (onBlur, onFocus, onClick, onMouseOut...), is this change could cause a potential regression, or does it have a reasonable explanation?

src/js/components/Grommet/Grommet.js
```diff
@@ -1,15 +1,15 @@
...
+  ContainerTargetContext,
   ResponsiveContext,
   ThemeContext,
-  ContainerTargetContext,
 } from '../../contexts';
 
 import {
-  backgroundIsDark,
   deepMerge,
+  backgroundIsDark,
```
ShimiSun, line 12: why did this prop move a line?

src/js/components/Grommet/Grommet.js
```diff
@@ -21,144 +21,90 @@ const FullGlobalStyle = createGlobalStyle`
...
-      // It is necessary that we set it twice. We have to handle two cases:
-      // 1. Caller passes in a color object or a color name that resolves an
-      //    object. In this case, we want to set dark as line 38 shows. The
-      //    second set, in line 46, is a no-op.
-      // 2. Caller passes a specific color value or a color name that resolves
-      //    to a specific color value. In this case, we want dark to be set
-      //    based on that color, which line 46 will do.
-      // The double set of nextTheme.dark allows us to handle both cases here
-      // without having to duplicate color object + name + dark mode detection
-      // code here that is already in normalizeColor and backgroundIsDark.
```
ShimiSun, line 59: Is this comment not necessary anymore or can we include it on the component's refactoring?

src/js/components/Tabs/__tests__/__snapshots__/Tabs-test.js.snap
MikeKingdom: Not a regression. The old test mechanism does snapshots by JSON-izing the component and therefore it lists props with camelCase and a toString of the handler functions. The newer mechanism is basically a more stringified version of the element so you see class vs className etc and the handlers don't show up as properties on the element. It's still the same DOM

src/js/components/Grommet/Grommet.js
MikeKingdom: The comment was a bit hard to figure out since the line numbers were off from some earlier change (a hazard of using line numbers in comments). I think I could probably re-write the description to convey the info with the new hook way of tweaking the theme (it's the same logic and effect.)

src/js/components/Grommet/Grommet.js
MikeKingdom: the import just moved since I pretty much re-worked the entire file

ShimiSun: I think we should lean towards alphabetized order.

PR #4878, opened 2021-01-13
Don't let kind attribute leak to DOM

taysea:
#### What does this PR do?
`kind` prop was leaking to the DOM. This change makes sure that it does not. In React 16, unknown attributes passed onto elements are no longer sifted out. See documentation here (https://reactjs.org/blog/2017/09/08/dom-attributes-in-react-16.html). Therefore, this change makes sure that no kind value is passed onto the DOM element.

Snapshot updates, only the `kind` attribute is being removed. no styling changes.

kind prop was leaking to the DOM, this removes it.

before:
Screen Shot 2021-01-13 at 10 05 42 AM

after:
Screen Shot 2021-01-13 at 10 06 13 AM

No.

No.

Backwards compatible.

### DIFF
--- a/src/js/components/Button/StyledButtonKind.js
+++ b/src/js/components/Button/StyledButtonKind.js
@@ -152,7 +152,10 @@ const plainStyle = () => css`
   }
 `;
 
-const StyledButtonKind = styled.button`
+const StyledButtonKind = styled.button.attrs(() => ({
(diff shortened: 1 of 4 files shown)


halocline (approved):
Good catch and clean up.

ShimiSun (approved):
THANKS! 💟

PR #4879, opened 2021-01-14
Remove styled placeholder when input has value

### LINKED ISSUE #4867: TextInput renders styled placeholders even when there is a value

taysea:
#### What does this PR do?
Manages presence of StyledPlaceholder via state (showStyledPlaceholder). When `onChange` is called within the TextInput, it checks if the input will have a value or not. If yes, it removes the styled placeholder. If no, it keeps it.

This specifically addresses the case when a TextInput is not in a Form. When the TextInput is not in a Form, no rerender was happening when a value changed and therefore the styledplaceholder was not disappearing when the textinput had a value. A detailed explanation of this issue is [well documented here](https://github.com/grommet/grommet/issues/4867).

When a TextInput is inside a form, all worked correctly and continues to work correctly.

https://github.com/grommet/grommet/issues/4867

Test in storybook with https://storybook.grommet.io/?path=/story/input-textinput--styled-placeholder by removing the `Form` tags. As you type, the placeholder should disappear and if there value is cleared, the placeholder should reappear.

Same as above.

Closes #4867 

No.

Yes. Removes custom JSX placeholder from TextInput when input has a value.

Backwards compatible.


src/js/components/TextInput/TextInput.js
```diff
@@ -448,6 +450,11 @@ const TextInput = forwardRef(
               readOnly
                 ? undefined
                 : event => {
+                    // when TextInput is not contained in a Form, no re-render
+                    // will come from this onChange and remove the placeholder,
+                    // so we need to update state to ensure the styled
+                    // placholder only appears when there is no value
+                    setShowStyledPlaceholder(!event.target.value);
```
MikeKingdom, line 457: When I investigated this, it seems the onChange doesn't get called while the user is typing until they tab out or change focus. As a result the placeholder remains and the typed text overlaps it. Are you seeing something similar?

taysea: onChange was getting called as soon as they were typing. However, it wasn't causing a rerender for whatever reason (which made the placeholder remain). I think it has to do with the fact that `value` is looking to formContext and so if the TextInput isn't in a formContext it's not registering the value change as a state change. 

by converting showStyledPlaceholder to something that is controlled by TextInput state, it allows us to toggle when it's shown by checking if the event.target has a value or not. Let me know if this makes sense.

taysea: I was running this locally in Chrome. Let me confirm that it works across browsers.

taysea: It is working as expected on Chrome, Firefox, and Safari.

MikeKingdom: Looks great to me. I gave it a try and it definitely removes the placeholder on typing.

ericsoderberghp (approved):

Issue #4882, opened 2021-01-15
Input inside of Form without name does not receive value


taysea:
### Expected Behavior

Selecting a value of an input within a Form should show that value. For example, selecting an option from select should show that option. 
 

### Actual Behavior

Right now, if no `name` is on the Select, for example, then no value is displayed when selected. Potentially related to https://github.com/grommet/grommet/pull/4817

Demonstrated here: https://designer.grommet.io/?id=Select-contexts-eric-soderberg-hpe-com
### Steps to Reproduce

1. Select an option from 'unnamed' Select
 2. No value shows up

### Your Environment

- Grommet version: Stable
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): MacOS

PR #4884, opened 2021-01-18
Fix components update in Form without name prop

leossantiago:
#### What does this PR do?

Fixes components updating behavior when used inside an uncontrolled Form (related to PR #4817 )

Components that make use of `FormContext` (CheckBox, CheckBoxGroup, FormField, MaskedInput, RadioButtonGroup, RangeInput,Select, TextArea and TextInput) use the `useFormInput` function to handle its values (and updates). This function has two versions:
1. The original context function (when the component isn't inside a Form):
https://github.com/grommet/grommet/blob/5b5113e9dab0239eb19bfa117960580affb3a846/src/js/components/Form/FormContext.js#L8
2. The overwritten one (when component IS inside a Form):
https://github.com/grommet/grommet/blob/5b5113e9dab0239eb19bfa117960580affb3a846/src/js/components/Form/Form.js#L121

From PR #4817, a condition for resetting uncontrolled forms was introduced:
`Resets uncontrolled input values when form has reset. Previously, if a Form reset, CheckBox/RadioButtonGroup/Select would not reset their value.`

https://github.com/grommet/grommet/blob/5b5113e9dab0239eb19bfa117960580affb3a846/src/js/components/Form/Form.js#L149-L152

Even though it's a fix for uncontrolled forms, this seems to work mostly for controlled forms, as there would be situations without an undefined `value`. For uncontrolled Forms, the last `else` would never be reached, and the component value would always be updated to `""` (empty string). This PR fixes that by introducing a state to track whether a reset was called or not, allowing the listed components to update values properly.

`src/js/components/Form/Form.js`

Manual testing, but also using uncontrolled form tests present on `Forms/__tests__/Form-test-uncontrolled.js` for Select update, and tests introduced in #4817

https://github.com/grommet/grommet/blob/5b5113e9dab0239eb19bfa117960580affb3a846/src/js/components/Form/__tests__/Form-test-uncontrolled.js#L890
https://github.com/grommet/grommet/blob/5b5113e9dab0239eb19bfa117960580affb3a846/src/js/components/Form/__tests__/Form-test-uncontrolled.js#L945

**Note:** added a test for Select without name inside a Form.

```js
      <Form>
        <FormField>
          <Select options={['morning', 'evening']} />
        </FormField>
        <Button type="reset" primary label="Reset" />
      </Form>
    </Grommet>
```

#4882

No.

Backwards compatible, as it fixes an introduced bug.


src/js/components/Form/Form.js
```diff
@@ -139,17 +145,26 @@ const Form = forwardRef(
...
+
       let useValue;
       if (componentValue !== undefined)
         // input component drives, pattern #2
         useValue = componentValue;
       else if (valueProp && name && formValue !== undefined)
         // form drives, pattern #1
         useValue = formValue;
-      else if (formValue === undefined)
+      else if (formValue === undefined && shouldReset) {
```
taysea, line 162: This all works great for the reseting, but I think a bigger question is if an input doesn't have a `name` prop, should the form be reseting its value at all? I ask because if an input doesn't have `name` prop, when the form is submitted, then that input does not appear in the onSubmit since `name` is required to link the input with the form values.

If we decide that for a Form to control the reset of any input, that input should have a `name` prop (similar to what is needed for that input to appear in the submit), then I think we can remove the reset state variables you've created and just adjust line 162 to be:
```
else if (formValue === undefined && name)
```
I'd like to hear @ericsoderberghp or @ShimiSun thoughts though.

ericsoderberghp: I'm inclined to agree, that if an input doesn't have a name, it shouldn't be touched, since we don't touch un-named inputs anywhere else.

taysea: Great, so I'd recommend this PR removes the logic around `shouldReset`, etc. and just applies:
```
else if (formValue === undefined && name)
```
@leossantiago

taysea: Closes #4882

leossantiago:
Updated! The test was kept as it checks if we were able to change the value of a component without `name` even if inside a Form.

Peek 2021-01-25 10-00

taysea (approved):
LGTM. Nice work.

ericsoderberghp (approved):

PR #4888, opened 2021-01-20
Enhance Layer border radius and allow background prop to be applied

taysea:
#### What does this PR do?
Allows user to style Layer border radius via grommet size such as "small" or px value such as "4px". If user opts into `intelligentRounding`, then the Layer will automatically apply a border-radius of 0 to any edge of the layer that's touching the edge of the screen.

For the positions "start" and "end" that are dependent on if a user is in LTR or RTL, there is a new CSS style that can be used to apply border-radius properly depending on which direction the app is in. However, it currently only functions in Firefox. So, we need to decide what the fallback behavior should be for all other browsers.

Additionally, added background prop to Layer so that user could style background and maintain border-radius theming.

Added Border Radius storybook example to Layer and added jest tests for all positions/full combinations.

Look at border radius storybook example.

Layer would always round all corners. This is not always desirable for Layers with position "right" where one edge is flush against the edge of the window.

Related to https://github.com/grommet/hpe-design-system/issues/1341

Yes, updated here.

Yes. Enhance Layer to accept background prop and grommet sizes for theme.layer.border.radius.
Backwards compatible.


src/js/components/Layer/StyledLayer.js
```diff
@@ -651,8 +742,15 @@ const desktopContainerStyle = css`
       props.theme,
       'right',
     )}px)`};
-  border-radius: ${props =>
-    props.plain ? 0 : props.theme.layer.border.radius};
+  ${props =>
+    props.plain || props.full === true
```
ericsoderberghp, line 746: Can we have `full` with `margin`? If so, then we should use `roundStyle()` in that case too.

src/js/components/Layer/doc.js
```diff
@@ -146,6 +148,18 @@ export const themeDoc = {
...
+    description: 'The rounding of the Layer corners.',
+    type: 'string',
+    defaultValue: 'white',
+  },
+  'layer.border.intelligentRounding': {
+    description: `Whether the border-radius of the Layer should adapt based on 
+    the Layer's position. Wherever the Layer is touching the edge of the 
+    screen, a border-radius of 0 will be applied.`,
+    type: 'boolean',
+    defaultValue: undefined,
```
ericsoderberghp, line 161: Given that `layer.border.radius` is new, should the default value for this key should be `true`, given that's the more likely desire?

taysea: layer.border.radius is not new, it just wasn't documented prior

src/js/components/Layer/StyledLayer.js
taysea: good call, let me look

src/js/components/Layer/doc.js
ericsoderberghp: k

ericsoderberghp (approved):

PR #4892, opened 2021-01-22
Disable animation on layer tests

taysea:
#### What does this PR do?
Disables animation for Layer tests that involved onClickOutside/onEsc. There have been occasional test failures, and I am wondering if the layer animation in the jest environment was causing timing issues. Disabling animation to see if that removes test failures consistently. If we see Layer test failures again in the future, then more investigation will be needed.

src/js/components/Layer/__tests__/Layer-test.js

Updated snapshots.

Occasional Layer test failures. They would pass sometimes, and fail on others. Animation was potentially causing that unreliability.

No.

No.

Backwards compatible.


ShimiSun (approved):
Thank you! 💟

Issue #4887, opened 2021-01-20
Datatable thead not aligned with tbody when overflow is set to true


smartnexus:
### Expected Behavior
When I set the size of Datatable to a fixed height, it automatically fits table data to the fixed height adding an overflow bar(as html by default), but when that scroll bar is being displayed table tbody width reduces and thead does not reduces with the same anchor.

### Actual Behavior
Datatable does not resize thead to keep the columns aligned.

### URL, screen shot, or Codepen exhibiting the issue
Columns not being aligned.
![image](https://user-images.githubusercontent.com/6134261/105168763-dd98fb80-5b1a-11eb-9f12-e4756518ee5f.png)
------------------------------------------------------------------------------------------------------------
Fixed adding padding to last th element of thead.
![image](https://user-images.githubusercontent.com/6134261/105168627-aa566c80-5b1a-11eb-84ad-0c1f811f8b1c.png)

### Steps to Reproduce

1. Create datatable element with data.
2. Add "size" property to DataTable component.

### Your Environment

- Grommet version: 2.16.2
- Browser Name and version: Firefox 84.0.2
- Operating System and version (desktop or mobile): Windows 10 Desktop

ShimiSun:
Thank you for reporting!

smartnexus:
What is the expected time to solve this?

ericsoderberghp:
We are always ready to review pull requests. :)

ShimiSun:
Thanks for reporting, the fix will be available in the coming release.

PR #4896, opened 2021-01-25
Adjust DataTable Header/Footer on Body overflow

leossantiago:
#### What does this PR do?

This fix forces `Header` and `Footer` of `DataTable` to adjust themselves according to `Body` content overflow (and showing a scrollbar).

By setting a Ref to Body's content, we can track its width and its parent's width. This might be important when dealing with custom scrollbars. The difference in size, saved as a state, is passed down to Header and Footer, where it's used inside their widths calculations.

Mainly storybook using its examples.

Use any DataTable example (i.e. https://storybook.grommet.io/?path=/story/visualizations-datatable--clickable-data-table) and pass `size="small"` (or any size which leads to an overflow, considering the length of the provided data) as prop to DataTable.

Issue #4887

Before:
image

After:
image

Backwards compatible functionality-wise. Breaking change considering Header and Footer width shrink.


src/js/components/DataTable/DataTable.js
```diff
@@ -127,10 +127,22 @@ const DataTable = ({
 
   // placeholder placement stuff
   const headerRef = useRef();
+  const bodyRef = useRef();
   const footerRef = useRef();
   const [headerHeight, setHeaderHeight] = useState();
   const [footerHeight, setFooterHeight] = useState();
 
+  const [overflowComp, setOverflowComp] = useState(0);
```
ShimiSun, line 135: would it make sense to call this state `bodyOverflow` vs `overflowComp`?

ShimiSun (commented):
Overall looks good to me and that it is doing the trick. Dropped a small comment regarding readability, let me know what you think.
💟

src/js/components/DataTable/DataTable.js
leossantiago: I was going for `overflowCompensation` but I do agree it could use some work on readability. I like `bodyOverflow` but I'm not sure if it's related to what it does. Even though `body` is the one overflowing, I don't think it would explain its use.

WDYT about `scrollBarWidth` or maybe `scrollBarCompensation` ?

Also, I'll try pushing at least one snapshot later on, so we can keep track of this behavior.

ericsoderberghp: We've used the term `offset` for situations like this elsewhere. Perhaps `scrollOffset`?

leossantiago:
While adding tests to this, I noticed GroupedBody needed a fix as well, but since its overflow can be triggered after rendering and wouldn't notify us about changes to the current ref value, the useEffect approach wouldn't solve it. So I migrated to a useCallback ref. That way we can use the same code to both Body and GroupedBody. 

I'd appreciate your opinions on that, @ShimiSun @ericsoderberghp ! 

Body:
image

GroupBody:
Peek 2021-01-25 23-32

leossantiago:
**NOTE:** removed tests. It seems jest/RTL make use of jsdom, which can't handle what this code does. It passes `width: calc(100% - 0px)` even on examples that should overflow, making those tests useless IMHO. Let me know what you think about that as well, and if there's another approach we could try to test this. Otherwise, we'll probably stick to manual testing done with storybook.

src/js/components/DataTable/DataTable.js
```diff
@@ -131,6 +131,16 @@ const DataTable = ({
   const [headerHeight, setHeaderHeight] = useState();
   const [footerHeight, setFooterHeight] = useState();
 
+  // offset compensation when body overflows
+  const [scrollOffset, setScrollOffset] = useState(0);
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const bodyRef = useCallback(ref => {
```
ericsoderberghp, line 138: One suggestion, can we do something like this instead?
```
  const bodyRef = useRef();
  useLayoutEffect(() => {
    const nextScrollOffset =
      bodyRef.current.parentElement?.clientWidth - bodyRef.current.clientWidth;
    if (nextScrollOffset !== scrollOffset) setScrollOffset(nextScrollOffset);
  });

leossantiago: Sure thing! I'll push a commit addressing that! Also, thanks for pointing that out. Never had a use case for useLayoutEffect so I've been reading about it a little bit 😄

ericsoderberghp (approved):

ericsoderberghp:
Thanks!

ShimiSun:
Thank you @leossantiago 💟

PR #4897, opened 2021-01-25
Storybook v6 upgrade

jcfilben:
#### What does this PR do?
Updates the storybook version to the latest, 6.1.17

This update includes changes to the CSF format for the stories. Previously the annotations looked like this:
```javascript
export const Basic = () => <Button />
Basic.story = {
  name: 'foo',
  parameters: { ... },
  decorators: [ ... ],
};
```
The above format is deprecated and the annotation should be in this format instead:
```javascript
export const Basic = () => <Button />
Basic.storyName = 'foo';
Basic.parameters = { ... };
Basic.decorators = [ ... ];
};
```
This PR fixes the stories to use the new format. This is where the majority of the changes come from. Also stories should no longer use `storiesOf`.

Files with multiple stories have been separated so that each file has 1 story and there are no duplicated exports.

In the `storybook/preview.js` file I removed `options: { showRoots: true, }` because `showRoots` is deprecated and the behavior is now done by default and controlled within the `.storybook/manager.js` file instead of the `preview.js` file.

Also in `storybook/preview.js` I added `layout: 'fullscreen'` so the component takes up the full canvas instead of having a margin around it. The previous behavior in 5.3.20 was for the component to take up the full canvas.

More information about migrating to version 6 of storybook can be found in the [version 6 migration guide](https://storybook.js.org/blog/storybook-6-migration-guide/) as well as this more detailed [markdown file](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md).

Start with the `package.json` and `storybook/preview.js` files.
Then look at the stories to see the changes from the old CSF annotation format to the new format.

storybook

storybook

No

No

Backwards compatible


package.json
```diff
@@ -69,15 +69,15 @@
...
+    "@storybook/addon-storysource": "^6.1.11",
+    "@storybook/addons": "^6.1.11",
+    "@storybook/react": "^6.1.11",
+    "@storybook/source-loader": "^6.1.11",
     "@testing-library/dom": "^6.1.0",
     "@testing-library/jest-dom": "^4.2.4",
     "@testing-library/react": "^9.1.3",
     "@types/node": "^14.14.22",
-    "@types/react": "^16.9.11",
+    "@types/react": "^17.0.0",
```
ShimiSun, line 80: I wonder whether this is a good approach since we haven't upgrade React to 17.0.0 yet, why is this change needed as part of this PR?

jcfilben: Agreed, I switched it back to `"@types/react": "^16.9.11"`

package.json
```diff
@@ -69,10 +69,10 @@
...
     "@babel/preset-env": "^7.12.11",
     "@babel/preset-react": "^7.12.10",
-    "@storybook/addon-storysource": "^5.3.19",
-    "@storybook/addons": "^5.3.19",
-    "@storybook/react": "^5.3.19",
-    "@storybook/source-loader": "^5.3.19",
+    "@storybook/addon-storysource": "^6.1.11",
+    "@storybook/addons": "^6.1.11",
+    "@storybook/react": "^6.1.11",
+    "@storybook/source-loader": "^6.1.11",
```
ShimiSun, line 75: Why not the latest version? I believe its 6.1.15

jcfilben: Fixed version to be 6.1.15 in latest push. Note: when running storybook it says version 6.1.14 but in the console it correctly shows 6.1.15 as the version being used and the package.json and yarn.lock show 6.1.15

ShimiSun (commented):
I think this PR is in a good place.
Since we are prioritizing the release, I'll circle back to it once the release is out. Thanks 💟

PR #4898, opened 2021-01-26
Allow clicks to complete before running onBlur validation

### LINKED ISSUE #4863: Form intermittently doesn't acknowledge reset button when validation on blur fails
### Expected Behavior

I expect that after entering an invalid value in a form input and clicking reset, that the reset will occur. The input field and any errors should be cleared.

### Actual Behavior

I'm occasionally seeing an odd "timing" issue when working on a Form that has a validation on blur for an input field. Sometimes when I enter an invalid value in an input with a "blur" validation setting, then I click on a Reset button (but the problem may occur with any non-submit button), the form doesn't register that I clicked the button. It will display the validation error but it won't clear the field. The button will be highlighted as if clicked, but if I put a log statement on its onClick handler or on the forms onReset handler, I see nothing in the logs.

This is reproducible occasionally for the Name field here: https://storybook.grommet.io/?path=/story/input-form--validate-on-blur. It seems to happen more often when the click happens immediately after the text is entered. Once the problem occurs, it is more likely to occur again.

Reproducible here, and also reproduced by @jcfilben :
https://storybook.grommet.io/?path=/story/input-form--validate-on-blur
reset_not_working

### Steps to Reproduce

1. Create a form with a reset button and a validation on blur for an input
2. Enter an invalid input
3. Immediately click reset
4. Intermittently, the reset button will appear to be clicked/highlighted, but the value will remain in the input field with the corresponding validation error.

### Your Environment

Grommet: 2.15.2
Firefox 84.0.1 (Ubuntu 20.04.1 and Windows 10, desktop)
Chrome 87.0.4280.88 (Ubuntu 20.04.1, desktop)

taysea:
#### What does this PR do?
This PR allows click events, such as reset, to complete prior to onBlur validation. A state variable called `pendingValidation` and a useEffect hook track what input (if any) is awaiting validation from a blur event. A timeout is set to ensure that click events complete prior to running the validation. Without this, click events like reset, etc. were sometimes not being called because the click event was not being registered. See https://github.com/grommet/grommet/issues/4863 for detailed scenario.

The timeout value was tested across Chrome, Safari, and Firefox to empirically determine the minimum value.

Tested in Validate onBlur form story.

1. Type a letter into the first input
2. Click "Reset" button

Expected outcome: Form resets, no validation error is displayed.

Same as above.

Closes #4863 

No.

Yes. Improve Form onBlur validation.

Backwards compatible.

### DIFF
--- a/src/js/components/Form/Form.js
+++ b/src/js/components/Form/Form.js
@@ -64,10 +64,12 @@ const Form = forwardRef(
       defaultValidationResults,
     );
 
-    // when onBlur input validation is triggered by a click event,
-    // we need to complete the click event before running the validation
+    // when onBlur input validation is triggered, we need to complete any
+    // potential click events before running the onBlur validation.
     // otherwise, click events like reset, etc. may not be registered.
     // for a detailed scenario/discussion, see: https://github.com/grommet/grommet/issues/4863
+    // the value of pendingValidation is the name of the FormField
+    // awaiting validation.
     const [pendingValidation, setPendingValidation] = useState(undefined);
 
     useEffect(() => {
@@ -76,6 +78,7 @@ const Form = forwardRef(
     }, [errorsProp, infosProp]);
     const validations = useRef({});
 
+    // Currently, onBlur validation will trigger after a timeout of 120ms. #4863
     useEffect(() => {
       const timer = setTimeout(() => {
         if (pendingValidation) {
--- a/src/js/components/Form/Form.js
+++ b/src/js/components/Form/Form.js
@@ -64,14 +64,55 @@ const Form = forwardRef(
       defaultValidationResults,
     );
 
-    useEffect(
-      () => setValidationResults({ errors: errorsProp, infos: infosProp }),
-      [errorsProp, infosProp],
-    );
+    // when onBlur input validation is triggered by a click event,
+    // we need to complete the click event before running the validation
+    // otherwise, click events like reset, etc. may not be registered.
+    // for a detailed scenario/discussion, see: https://github.com/grommet/grommet/issues/4863
+    const [pendingValidation, setPendingValidation] = useState(undefined);
+
+    useEffect(() => {
+      setPendingValidation(undefined);
+      setValidationResults({ errors: errorsProp, infos: infosProp });
+    }, [errorsProp, infosProp]);
     const validations = useRef({});
+
+    useEffect(() => {
+      const timer = setTimeout(() => {
+        if (pendingValidation) {
+          // run validations on touched keys
+          const [nextErrors, nextInfos] = validate(
+            Object.entries(validations.current).filter(
+              ([n]) => touched[n] || n === pendingValidation,
+            ),
+            value,
+          );
+          setPendingValidation(undefined);
+          // give user access to errors that have occurred on validation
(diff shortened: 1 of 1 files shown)


src/js/components/Form/Form.js
```diff
@@ -64,14 +64,55 @@ const Form = forwardRef(
...
     );
 
-    useEffect(
-      () => setValidationResults({ errors: errorsProp, infos: infosProp }),
-      [errorsProp, infosProp],
-    );
+    // when onBlur input validation is triggered by a click event,
+    // we need to complete the click event before running the validation
+    // otherwise, click events like reset, etc. may not be registered.
+    // for a detailed scenario/discussion, see: https://github.com/grommet/grommet/issues/4863
```
taysea, line 70: anything causing input to blur, 

add a note: value is the name of the form field waiting to be validated.

src/js/components/Form/Form.js
```diff
@@ -64,14 +64,55 @@ const Form = forwardRef(
       defaultValidationResults,
     );
 
-    useEffect(
-      () => setValidationResults({ errors: errorsProp, infos: infosProp }),
-      [errorsProp, infosProp],
-    );
+    // when onBlur input validation is triggered by a click event,
+    // we need to complete the click event before running the validation
```
taysea, line 68:
```suggestion
    // we need to complete the click event before running the onBlur validation
```

src/js/components/Form/Form.js
```diff
@@ -64,14 +64,55 @@ const Form = forwardRef(
...
+            return nextValidationResults;
+          });
+        }
+        // a timeout is needed to ensure that a click event (like one on a reset
+        // button) completes prior to running the validation. without a timeout,
+        // the blur will always complete and trigger a validation prematurely
+        // The following values have been empirically tested, but 120 was
+        // selected because it is the largest value
+        // Chrome: 100, Safari: 120, Firefox: 80
+      }, 120);
```
taysea, line 108: comment from shimi: we might want to make this value themeable, so if callers still experience the issue it would be configurable to manipulate the timeout value to make the onBlur trigger faster or slower.

src/js/components/Form/Form.js
```diff
@@ -64,14 +64,55 @@ const Form = forwardRef(
...
+    // for a detailed scenario/discussion, see: https://github.com/grommet/grommet/issues/4863
+    const [pendingValidation, setPendingValidation] = useState(undefined);
+
+    useEffect(() => {
+      setPendingValidation(undefined);
+      setValidationResults({ errors: errorsProp, infos: infosProp });
+    }, [errorsProp, infosProp]);
     const validations = useRef({});
 
+    useEffect(() => {
```
taysea, line 79:
```suggestion
    // Currently, onBlur validation will trigger after a timeout of 120ms. #4863
    useEffect(() => {
```

taysea: Done.

src/js/components/Form/Form.js
taysea: Done.

src/js/components/Form/Form.js
taysea: Done.

PR #4899, opened 2021-01-26
Fix border radius when full

taysea:
#### What does this PR do?
When Layer receives full of any kind (true, vertical, horizontal, etc), the corners are touching the edge of the screen and should receive a radius of 0.

Used border radius Layer story.

Necessary to apply desired behavior as demonstrated here and other designs such as global header menu layer:
Screen Shot 2021-01-26 at 11 06 26 AM

No.

No.

Backwards compatible, this functionality has not yet been included in a release.

### DIFF
--- a/src/js/components/Layer/StyledLayer.js
+++ b/src/js/components/Layer/StyledLayer.js
@@ -743,7 +743,7 @@ const desktopContainerStyle = css`
       'right',
     )}px)`};
   ${props =>
-    props.plain || (props.full === true && props.margin === 'none')
+    props.plain || (props.full && props.margin === 'none')
       ? `border-radius: 0;`
       : roundStyle(
           props.theme.layer.border.radius,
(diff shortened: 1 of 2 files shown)


src/js/components/Layer/StyledLayer.js
```diff
@@ -743,7 +743,7 @@ const desktopContainerStyle = css`
       'right',
     )}px)`};
   ${props =>
-    props.plain || (props.full === true && props.margin === 'none')
+    props.plain || (props.full && props.margin === 'none')
```
halocline, line 746: **Note:** Chatted through this with @taysea, but leaving these thoughts for others to consider.

How does this work if I want my Layer to positioned incoming from the position="right", full="vertical", but I want my corners to be `24px 0 0 24px`? Am I able to do that with this approach? Seems like all corners get set to `0` but I may be missing something.

**Desire**
Screen Shot 2021-01-26 at 2 55 03 PM

--> **Outcome of our conversation:** Decided that this implementation as is advances the current asks and we can wait for the community to express the need for "smart corners" when in full modes.

halocline (approved):
Good with the functionality. Had question regarding whether "full" mode should support "smart corners" when full="vertical" or full="horizontal" and whether this is something that should be implemented now vs. waiting for request from the Grommet community.

ericsoderberghp (approved):

Issue #4875, opened 2021-01-12
TextInput with suggestions with keyboard skips every other suggestion


taysea:
### Expected Behavior

Using down arrow with textinput + suggestions should go through each suggestion.

### Actual Behavior

Down arrow goes to every other suggestion (skipping every other one).

### URL, screen shot, or Codepen exhibiting the issue
https://storybook.grommet.io/?path=/story/input-textinput--suggestions

### Steps to Reproduce

1. Click into text input
2. Use down arrow to navigate through suggestions
3. Notice that you only navigate to every other one

### Your Environment

- Grommet version: stable
- Browser Name and version: Chrome Version 87.0.4280.141
- Operating System and version (desktop or mobile): MacOS Mojave 10.14.6

netochaves:
This problem is because `onUp` and `onDown` is called two times
First here:
https://github.com/grommet/grommet/blob/8a0fd52b8ae330a245525033b2f72c2ab1b09bbb/src/js/components/TextInput/TextInput.js#L265-L268

And here:
https://github.com/grommet/grommet/blob/8a0fd52b8ae330a245525033b2f72c2ab1b09bbb/src/js/components/TextInput/TextInput.js#L395-L415

I can open a PR to remove one of this but which one is safely to remove?

halocline:
@ShimiSun I had just started looking into this but saw that you had assigned yourself.

Based on running `git bisect` I believe this was introduced in [this commit](https://github.com/grommet/grommet/pull/4864).

![Screen Shot 2021-01-26 at 4 17 41 PM](https://user-images.githubusercontent.com/1756948/105919228-75528a00-5ff2-11eb-9f95-b938a3806680.png)

I will shift focus to another issue.

ShimiSun:
@netochaves Thank you!! Please review my associated PR and let me know what you think. I took the opportunity to do some cleanups along with the fix.

@halocline just saw your comment, feel free to read the analysis of the root cause on my associated PR.

taysea:
Closing this issue since it was solved in the recent release: https://github.com/grommet/grommet/pull/4901

PR #4901, opened 2021-01-27
Changed keyboard handling of TextInput suggestions

ShimiSun:
#### What does this PR do?
Handling the double trigger of keypress up/down of TextInput suggestion.
The issue only reproduces on stable and not on an official release, the issue was introduced ~2 weeks ago with the new `capture` prop addition for the Keyboard component.
https://github.com/grommet/grommet/pull/4864

TextInput is currently using two instances of the Keyboard component and each one fired the onDown/onUp event, since `capture` was added, none of the events got waved, and each single keyboard arrow action was triggered twice.

Along with fixing the issue, this PR is attempting to clean-up some of the `onEnter` code duplications, and as a result of making `onEnter` as generic as possible for both keyboard instances the test file had to be changed since `onSelect` would no longer be called in case there is no active index -- let me know what you think about this change.

I was also contemplating removing completely one of the Keyboard instances, but each one serves a separate scenario according to the code's comments that I found to be merit during testing -- let me know if you have any thoughts on consolidating the component to a single instance.

TextInput.js
jest, storybook
story/input-textinput--suggestions
story/input-textinput--default-suggestion

Try to hover, move scroll, click up/down, enter, and all the other keyboard Shabang.

https://github.com/grommet/grommet/pull/4864

#4875 

No
No - the issue was never introduced on an official release.
b/c - unless we think that the fact `onSelect` isn't being called with the empty suggestion object, is not b/c.

### DIFF
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -258,6 +258,21 @@ const TextInput = forwardRef(
     const extraProps = {
       onSelect: handleTextSelect,
     };
+
+    const onEnterEventHandler = (event, handleForm) => {
+      closeDrop();
+      if (activeSuggestionIndex >= 0 && handleSuggestionSelect) {
+        // prevent submitting forms when choosing a suggestion
+        if (handleForm) {
+          event.preventDefault();
+          event.persist();
+        }
+        const adjustedEvent = event;
+        adjustedEvent.suggestion = suggestions[activeSuggestionIndex];
+        handleSuggestionSelect(adjustedEvent);
+      }
+    };
+
     if (showDrop) {
       drop = (
         // keyboard access needed here in case user clicks
@@ -268,12 +283,7 @@ const TextInput = forwardRef(
           onEnter={event => {
             // we stole the focus, give it back
             inputRef.current.focus();
-            closeDrop();
-            if (handleSuggestionSelect) {
-              const adjustedEvent = event;
-              adjustedEvent.suggestion = suggestions[activeSuggestionIndex];
-              handleSuggestionSelect(adjustedEvent);
-            }
+            onEnterEventHandler(event);
             setValue(suggestions[activeSuggestionIndex]);
           }}
         >
@@ -366,19 +376,9 @@ const TextInput = forwardRef(
             {icon}
           </StyledIcon>
         )}
+
         <Keyboard
-          onEnter={event => {
-            closeDrop();
-            if (activeSuggestionIndex >= 0 && handleSuggestionSelect) {
-              // prevent submitting forms when choosing a suggestion
-              event.preventDefault();
-              event.persist();
-              const adjustedEvent = event;
-              adjustedEvent.suggestion = suggestions[activeSuggestionIndex];
-              adjustedEvent.target = inputRef.current;
-              handleSuggestionSelect(adjustedEvent);
-            }
-          }}
+          onEnter={event => onEnterEventHandler(event, true)}
           onEsc={
             showDrop
               ? event => {
@@ -393,7 +393,7 @@ const TextInput = forwardRef(
           }
           onTab={showDrop ? closeDrop : undefined}
           onUp={
-            showDrop &&
+            !showDrop &&
             suggestions &&
             suggestions.length > 0 &&
             activeSuggestionIndex
@@ -403,7 +403,7 @@ const TextInput = forwardRef(
               : undefined
           }
           onDown={
-            suggestions && suggestions.length > 0
+            !showDrop && suggestions && suggestions.length > 0
               ? event => {
                   if (!showDrop) {
                     openDrop();
@@ -451,9 +451,9 @@ const TextInput = forwardRef(
                 ? undefined
                 : event => {
                     // when TextInput is not contained in a Form, no re-render
-                    // will come from this onChange and remove the placeholder,
+                    // will come from this onChange and remove the placeholder
                     // so we need to update state to ensure the styled
-                    // placholder only appears when there is no value
+                    // placeholder only appears when there is no value
                     setShowStyledPlaceholder(!event.target.value);
                     setValue(event.target.value);
                     if (onChange) onChange(event);
@@ -461,6 +461,7 @@ const TextInput = forwardRef(
             }
           />
         </Keyboard>
+
         {drop}
       </StyledTextInputContainer>
     );
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -177,7 +177,7 @@ const TextInput = forwardRef(
 
     // Only update active suggestion index when the mouse actually moves,
     // not when suggestions are moving under the mouse.
-    const [mouseArmed, setMouseArmed] = useState();
+    const [mouseMovedSinceLastKey, setMouseMovedSinceLastKey] = useState();
 
     // reset activeSuggestionIndex when the drop is closed
     useEffect(() => {
@@ -262,7 +262,7 @@ const TextInput = forwardRef(
           suggestions.length - 1,
         );
         setActiveSuggestionIndex(nextActiveIndex);
-        setMouseArmed(false);
+        setMouseMovedSinceLastKey(false);
       },
       [activeSuggestionIndex, suggestions],
     );
@@ -272,7 +272,7 @@ const TextInput = forwardRef(
         event.preventDefault();
         const nextActiveIndex = Math.max(activeSuggestionIndex - 1, 0);
         setActiveSuggestionIndex(nextActiveIndex);
-        setMouseArmed(false);
+        setMouseMovedSinceLastKey(false);
       },
       [activeSuggestionIndex],
     );
@@ -305,7 +305,7 @@ const TextInput = forwardRef(
             ref={suggestionsRef}
             overflow="auto"
             dropHeight={dropHeight}
-            onMouseMove={() => setMouseArmed(true)}
+            onMouseMove={() => setMouseMovedSinceLastKey(true)}
           >
             <StyledSuggestions>
               <InfiniteScroll items={suggestions} step={theme.select.step}>
@@ -345,7 +345,8 @@ const TextInput = forwardRef(
                           setValueFromSuggestion(event, suggestion)
                         }
                         onMouseMove={
-                          mouseArmed && activeSuggestionIndex !== index
+                          mouseMovedSinceLastKey &&
+                          activeSuggestionIndex !== index
                             ? () => setActiveSuggestionIndex(index)
                             : undefined
                         }
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -233,7 +233,7 @@ const TextInput = forwardRef(
       if (onSuggestionsClose) onSuggestionsClose();
     }, [messages.onSuggestionsClose, onSuggestionsClose]);
 
-    const setFromSuggestion = (event, suggestion) => {
+    const setValueFromSuggestion = (event, suggestion) => {
       // if we stole the focus in the drop, perhaps by interacting with
       // a suggestion button or the scrollbar, give it back
       inputRef.current.focus();
@@ -242,7 +242,6 @@ const TextInput = forwardRef(
         if (event.persist) event.persist();
         const adjustedEvent = event;
         adjustedEvent.suggestion = suggestion;
-        // adjustedEvent.target = inputRef.current;
         handleSuggestionSelect(adjustedEvent);
       }
       inputRef.current.value = suggestion; // needed for uncontrolled cases
@@ -328,7 +327,9 @@ const TextInput = forwardRef(
                         kind={!child ? 'option' : undefined}
                         hoverIndicator={!child ? undefined : 'background'}
                         label={!child ? renderedLabel : undefined}
-                        onClick={event => setFromSuggestion(event, suggestion)}
+                        onClick={event =>
+                          setValueFromSuggestion(event, suggestion)
+                        }
                         onMouseOver={() => setActiveSuggestionIndex(index)}
                         onFocus={() => setActiveSuggestionIndex(index)}
                       >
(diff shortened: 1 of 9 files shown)


src/js/components/TextInput/TextInput.js
```diff
@@ -393,7 +393,7 @@ const TextInput = forwardRef(
           }
           onTab={showDrop ? closeDrop : undefined}
           onUp={
-            showDrop &&
+            !showDrop &&
```
ericsoderberghp, line 396: I think this is the key, that we don't process "Up" with both Keyboard listeners. Am I right?

src/js/components/TextInput/TextInput.js
```diff
@@ -258,6 +258,21 @@ const TextInput = forwardRef(
...
       onSelect: handleTextSelect,
     };
+
+    const onEnterEventHandler = (event, handleForm) => {
+      closeDrop();
+      if (activeSuggestionIndex >= 0 && handleSuggestionSelect) {
+        // prevent submitting forms when choosing a suggestion
+        if (handleForm) {
+          event.preventDefault();
+          event.persist();
```
ericsoderberghp, line 268: I think we should be persisting all the time. And, it would be nice to not need `handleForm` and put the `event.preventDefault()` in the non-drop Keyboard only.

src/js/components/TextInput/TextInput.js
ShimiSun: Exactly.

src/js/components/TextInput/TextInput.js
ShimiSun: Good comment. I thought so too and I tried to avoid this condition and the extra argument of `handleForm` on the function and make it generic for the two use cases, unfortunately, it didn't work well with the `showDrop` Keyboard, there was an issue with the selection of a keyboard item after scrolling. 
Let me take a second look since I tried it in the early stages of the PR and maybe the code was not fully mature for it at that time.

ShimiSun: So I was able to confirm that the functionality wouldn't work well with `event.preventDefault()` for the `showDrop` scenario.
Let's say we remove the `handleForm` condition so `event.preventDefault() && event.persist()` are always there, once you open the drop, use the scroll down -> hover on an item -> arrow from it up/down and click enter, the item should be selected but it's not which is an issue caused by the event actions.

ericsoderberghp: I was suggesting that `event.preventDefault()` be moved out of this function and only put in the Enter handler for the non-drop Keyboard.

ericsoderberghp:
OK, I made some bigger scale changes. I've reduced it to one `<Keyboard />`.

A few key areas to take take note of:
- `bestSuggestion` ensures we always reset the active suggestion index consistently. Previously, opening and closing the drop didn't always preserve it. You can see the difference in the "Suggestions" story. Select one suggestion and then re-open the drop via the down key.
- `setFromSuggestion` is used for both keyboard and mouse interaction, to ensure consistency.
- `keyboardProps` is built separately to keep the logic more consistent across event handlers.
- `event.relatedTarget` is used in the blur handler to not clear the focus state if the focus is moving to the drop. This is the key change that removes a lot of the more complicated event handling.
- we no longer listen for Esc in our Keyboard and instead rely on `onEsc` via Drop.

testing:
- all unit tests are passing
- I've added suggestions to the Form Controlled, Form Uncontrolled, and Layer Form stories. I recommend testing with these stories, mixing up keyboard and mouse interactions.

src/js/components/TextInput/TextInput.js
```diff
@@ -232,6 +233,22 @@ const TextInput = forwardRef(
       if (onSuggestionsClose) onSuggestionsClose();
     }, [messages.onSuggestionsClose, onSuggestionsClose]);
 
+    const setFromSuggestion = (event, suggestion) => {
```
taysea, line 236: This function name felt a little unclear to me. Maybe `setValueFromSuggestion` ?

src/js/components/TextInput/TextInput.js
```diff
@@ -232,6 +233,22 @@ const TextInput = forwardRef(
...
+      // if we stole the focus in the drop, perhaps by interacting with
+      // a suggestion button or the scrollbar, give it back
+      inputRef.current.focus();
+      closeDrop();
+      if (handleSuggestionSelect) {
+        if (event.persist) event.persist();
+        const adjustedEvent = event;
+        adjustedEvent.suggestion = suggestion;
+        // adjustedEvent.target = inputRef.current;
```
taysea, line 245: remove comment if not needed

src/js/components/TextInput/TextInput.js
```diff
@@ -435,32 +394,47 @@ const TextInput = forwardRef(
...
               }
-              if (onFocus) onFocus(event);
             }}
             onBlur={event => {
-              setFocus(false);
-              if (onBlur) onBlur(event);
+              // Only treat it as a blur if the element receiving focus
+              // isn't in our drop. The relatedTarget will be our drop
+              // when the user clicks on a suggestion or interacts with the
+              // scrollbar in the drop.
```
taysea, line 413: helpful comment, thanks. I confirmed the behavior works properly with a long list of suggestions that introduces a scrollbar.

src/js/components/TextInput/TextInput.js
```diff
@@ -258,104 +275,94 @@ const TextInput = forwardRef(
...
+                  );
+                }}
+              </InfiniteScroll>
+            </StyledSuggestions>
+          </ContainerBox>
+        </Drop>
       );
     }
 
+    // If we have focus, then we set the target to the document, otherwise
```
taysea, line 347: Why do we set the target to the document when we're in focus?

ericsoderberghp: `focus` being true doesn't mean the input has focus, it means that either the input or the drop has focus. We need to target the document to ensure we cover both contexts when the key is pressed.

src/js/components/TextInput/TextInput.js
ericsoderberghp: done

src/js/components/TextInput/TextInput.js
ericsoderberghp: good idea, changed

src/js/components/TextInput/TextInput.js
ericsoderberghp: I've put more into the comment to attempt to clarify.

ShimiSun:
Default Suggestions story is showing inconsistency behavior
image
After selecting option with a scroll, and arrow down with keyboard, the wrong suggestion is marked as active, and from time to time it will show two active suggestions.

ShimiSun:
Great, much better.
I still get inconsistent behavior when selecting an option and opening the drop, for example, select option #20, and open the drop, you'll see it is not highlighting #20 and starts from a random option.
Another thing that I noticed, is that when reaching the end of the drop via keyboard it doesn't skip to the next item correctly and steps back a few options.

taysea (commented):
I notice when I first click into the input on Default Suggestions, the second option is highlighted as opposed to the first.

ericsoderberghp:
> I notice when I first click into the input on Default Suggestions, the second option is highlighted as opposed to the first.

That's because the "Default Suggestion" story passes `defaultSuggestion={1}`, which aligns with the documentation for that property:
      Default suggestion to highlight, as an index into the suggestions array.
      If set, the suggestion at the specified index in the suggestions array
      will be highlighted by default when the suggestions drop opens.

ericsoderberghp:
> I still get inconsistent behavior when selecting an option and opening the drop, for example, select option #20, and open the drop, you'll see it is not highlighting #20 and starts from a random option.

Is your mouse over the suggestions drop after you open the drop, because that will cause the option being hovered over to be active. If I open the drop with the down arrow, option #20 is active, although not visible as such given the scroll position.

ShimiSun:
> > I notice when I first click into the input on Default Suggestions, the second option is highlighted as opposed to the first.
> 
> That's because the "Default Suggestion" story passes `defaultSuggestion={1}`, which aligns with the documentation for that property:
> Default suggestion to highlight, as an index into the suggestions array.
> If set, the suggestion at the specified index in the suggestions array
> will be highlighted by default when the suggestions drop opens.

I think that would have been true if it was acting consistently, here is an example where it is working until reaching 'Suggestion 15'

https://user-images.githubusercontent.com/6320236/106531127-dd99e380-64aa-11eb-82b4-433bbca44c9f.mov

src/js/components/TextInput/TextInput.js
```diff
@@ -152,18 +153,39 @@ const TextInput = forwardRef(
...
+    // activeSuggestionIndex unifies mouse and keyboard interaction of
+    // the suggestions
     const [activeSuggestionIndex, setActiveSuggestionIndex] = useState(
-      typeof defaultSuggestion === 'number' ? defaultSuggestion : -1,
+      bestSuggestion(),
     );
 
+    // Only update active suggestion index when the mouse actually moves,
+    // not when suggestions are moving under the mouse.
+    const [mouseArmed, setMouseArmed] = useState();
```
taysea, line 180: this variable name wasn't immediately clear to me -- maybe a different name. from my understanding, this is a way to discern that if we should be relying on the mouse placement or not to decide on the suggestion, some ideas:
- `mouseReady`
- `mouseControl`
- `mouseDetermineSuggestion`
- `mouseReadyToSelect`

no necessarily sold on some of these, but figured I'd leave as a brainstorm.

ericsoderberghp:
> I think that would have been true if it was acting consistently, here is an example where it is working until reaching 'Suggestion 15'

What is the issue demonstrated by the video? I can't tell which are keyboard and mouse interactions. It looks like you use the keyboard to select suggestions up until suggestion 15, at which point I can't tell what happens next. My guess is the drop is opened and suggestion 15 is active, only you don't see it because it is scrolled out of view.

src/js/components/TextInput/TextInput.js
ericsoderberghp: Renamed to `mouseMovedSinceLastKey`

ShimiSun:
Typing a Suggestion is not receiving the user input and randomly falls back to other Suggestion option, in this example, it fell randomly to 21 although the user was typing 25.

https://user-images.githubusercontent.com/6320236/106531807-2bfbb200-64ac-11eb-99f2-667e91f92501.mov

src/js/components/TextInput/TextInput.js
```diff
@@ -152,18 +153,39 @@ const TextInput = forwardRef(
     }, [onSuggestionsOpen, suggestions]);
     /* eslint-enable react-hooks/exhaustive-deps */
 
+    // choose the best suggestion, either the explicit default or the one
+    // that matches the current value
+    const bestSuggestion = useCallback(() => {
```
taysea, line 158: After I select a value, if I place focus in the input and press "down arrow", the drop opens but my current selection is not in view but is still the value I selected. If I start using arrow keys, I can't see where I am navigating.

I am doing the following in the video:
1. Mouse click into the input and selecting an option with my mouse.
2. Down arrow to open the suggestions again.
3. Down arrow to continue through selections (not in view since my original selection was outside of view).
4. Enter to select an option.

My expectation would be that upon using the down arrow after selecting an item, either:
a) My previously selected suggestion is in view when suggestions open
or
b) If I can't see my previous selection, once I press down arrow I should be navigating from the top of the suggestions list so I can see the active value.


https://user-images.githubusercontent.com/12522275/106532025-40d44780-64a4-11eb-84a2-a89ced64557c.mov

ShimiSun:
> > I think that would have been true if it was acting consistently, here is an example where it is working until reaching 'Suggestion 15'
> 
> What is the issue demonstrated by the video? I can't tell which are keyboard and mouse interactions. It looks like you use the keyboard to select suggestions up until suggestion 15, at which point I can't tell what happens next. My guess is the drop is opened and suggestion 15 is active, only you don't see it because it is scrolled out of view.

I think that you might be right regarding not seeing the option since it's out of view, though my problem was that I lost the keyboard indicator, I was going up/down and couldn't see the Suggestion that is being pointed out by the keyboard, it behaves random and I wasn't able to catch it on a video yet. Try to see if you can reproduce it locally.

ericsoderberghp:
> Typing a Suggestion is not receiving the user input and randomly falls back to other Suggestion option, in this example, it fell randomly to 21 although the user was typing 25.

When you say "typing" do you mean that the user was using the up/down arrow and Enter keys to select suggestion 25? And then, when opening the drop again, it shows "Suggestion 21" active instead of "Suggestion 25"? I didn't see that in the video referenced.

ericsoderberghp:
Thanks for the testing. I'll look into it later tonight.

ShimiSun:
> > Typing a Suggestion is not receiving the user input and randomly falls back to other Suggestion option, in this example, it fell randomly to 21 although the user was typing 25.
> 
> When you say "typing" do you mean that the user was using the up/down arrow and Enter keys to select suggestion 25? And then, when opening the drop again, it shows "Suggestion 21" active instead of "Suggestion 25"? I didn't see that in the video referenced.

When I say 'typing' I mean using the keyboard for typing a number, the video demonstrates the following flow:
1. Select "Suggestion 15"
2. Use the keyboard to change the number 15 and manually type "25" instead.
3. Mind the suggestion that was actually selected, in the video it is "Suggestion 21", but could randomly be another suggestion.

ericsoderberghp:
> > > Typing a Suggestion is not receiving the user input and randomly falls back to other Suggestion option, in this example, it fell randomly to 21 although the user was typing 25.
> > 
> > 
> > When you say "typing" do you mean that the user was using the up/down arrow and Enter keys to select suggestion 25? And then, when opening the drop again, it shows "Suggestion 21" active instead of "Suggestion 25"? I didn't see that in the video referenced.
> 
> When I say 'typing' I mean using the keyboard for typing a number, the video demonstrates the following flow:
> 
> 1. Select "Suggestion 15"
> 2. Use the keyboard to change the number 15 and manually type "25" instead.
> 3. Mind the suggestion that was actually selected, in the video it is "Suggestion 21", but could randomly be another suggestion.

Hmm... I can't reproduce this. When I type replace "15" with "25" and then open the drop, it shows 25, on both Safari and Chrome.

ShimiSun:
> > > > Typing a Suggestion is not receiving the user input and randomly falls back to other Suggestion option, in this example, it fell randomly to 21 although the user was typing 25.
> > > 
> > > 
> > > When you say "typing" do you mean that the user was using the up/down arrow and Enter keys to select suggestion 25? And then, when opening the drop again, it shows "Suggestion 21" active instead of "Suggestion 25"? I didn't see that in the video referenced.
> > 
> > 
> > When I say 'typing' I mean using the keyboard for typing a number, the video demonstrates the following flow:
> > 
> > 1. Select "Suggestion 15"
> > 2. Use the keyboard to change the number 15 and manually type "25" instead.
> > 3. Mind the suggestion that was actually selected, in the video it is "Suggestion 21", but could randomly be another suggestion.
> 
> Hmm... I can't reproduce this. When I type replace "15" with "25" and then open the drop, it shows 25, on both Safari and Chrome.

Interesting, maybe your previous commit handled it without knowing :) I'll try to reproduce it again.

ShimiSun:
LGTM! I can't officially approve the PR since I'm the author.

Issue #4881, opened 2021-01-15
DataTable has lag showing hover style with large number of rendered rows


kariwhitcomb:
### Expected Behavior
DataTable's hover styling should be applied consistently fast, even with a large amount of data being shown

### Actual Behavior
Using DataTable with onClickRow and large number of rendered rows (ex 200) experiences noticeable lag in showing the hover style. 

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/s/dark-hooks-ctdl3?file=/src/App.js

![Screen Shot 2021-01-15 at 9 47 18 AM](https://user-images.githubusercontent.com/12547826/104756478-18fd8980-5719-11eb-887e-5a6e2fc86f5a.png)

### Steps to Reproduce

1. Use DataTable component with step={200}, data (with at least 200 rows of info), and onClickRow handler set
2. Use mouse to hover over rows in the table.  There is a noticeable lag in the hover styling being applied after mousing onto a row.
3. Check browser console (I use chrome) and you may see the Violation messages I screenshotted above indicating the mouseover and mouseout handlers are taking too long

I did poke through some of the performance/react profiler behavior... it appears that every mouse event is causing the whole table to rerender.  So the larger the table, the longer this takes.  It seems like it should only have to rerender a max of 2 rows (the one that used to have hover, and the one that does now) so not sure why it is rerendering everything.

### Your Environment

- Grommet version: 2.16.2
- Browser Name and version: Chrome latest
- Operating System and version (desktop or mobile): MacOS

halocline:
Tagging potentially related https://github.com/grommet/grommet/issues/4559 & https://github.com/grommet/grommet/issues/3935.

PR #4904, opened 2021-01-27
Changed DataTable to use onMouseEnter for hover detection

ericsoderberghp:
#### What does this PR do?

Changed DataTable to use onMouseEnter for hover detection. This reduces the amount of re-rendering needed. There might be even more aggressive ways to tackle this down the road. But, this seems a good quick win.

whitebox 400 rows and onClickRow

same

#4881

no

yes

backwards compatible

### DIFF
--- a/src/js/components/DataTable/Body.js
+++ b/src/js/components/DataTable/Body.js
@@ -103,8 +103,8 @@ const Body = forwardRef(
                         }
                       : undefined
                   }
-                  onMouseOver={onClickRow ? () => setActive(index) : undefined}
-                  onMouseOut={
+                  onMouseEnter={onClickRow ? () => setActive(index) : undefined}
+                  onMouseLeave={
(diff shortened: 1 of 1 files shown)


src/js/components/DataTable/Body.js
```diff
@@ -103,8 +103,8 @@ const Body = forwardRef(
                         }
                       : undefined
                   }
-                  onMouseOver={onClickRow ? () => setActive(index) : undefined}
-                  onMouseOut={
+                  onMouseEnter={onClickRow ? () => setActive(index) : undefined}
```
MikeKingdom, line 106: So that's simply fewer events and fewer calls to setActive? I'm still a little surprised the redudant setActive() calls cause a re-render. Seems like a decent improvement.

ericsoderberghp: yep

ShimiSun (approved):

PR #4905, opened 2021-01-27
Allow required formfield asterisk to be enabled in the theme

taysea:
#### What does this PR do?

Enhances the theme to allow users to enable an asterisk * to appear next to required form field labels. Alternately, they can provide a string or a custom JSX element (if they want their own symbol for example). Currently, grommet users have to create a custom label render component any time they want to have this functionality. Having a place in the theme to control this would remove the need for users to use a custom component on every instance of a required formfield.

If the user flags this theme item as "true", then grommet will automatically apply an a11yTitle of "required". This a11yTitle is necessary so that screen readers read out the symbol as "required" as opposed to "star". If a user provides a custom element, such as a custom icon, then it will render this.

Tested/updated Required Label Form storybook and added jest tests.

Look at Required Label Form storybook example.

Related to https://github.com/grommet/hpe-design-system/issues/1378

Yes, updated here.

Yes. Use `theme.formField.label.requiredAsterisk` to have a * or custom element appear next to the label on required FormFields.

Backwards compatible.

### DIFF
--- a/src/js/components/Form/stories/RequiredLabel.js
+++ b/src/js/components/Form/stories/RequiredLabel.js
@@ -1,14 +1,6 @@
 import React from 'react';
 
-import {
-  Box,
-  Button,
-  Form,
-  FormField,
-  Grommet,
-  Text,
-  TextInput,
-} from 'grommet';
+import { Box, Button, Form, FormField, Grommet, Text } from 'grommet';
 import { grommet } from 'grommet/themes';
 import { deepMerge } from 'grommet/utils';
 
--- a/src/js/components/Form/stories/RequiredLabel.js
+++ b/src/js/components/Form/stories/RequiredLabel.js
@@ -1,35 +1,48 @@
 import React from 'react';
 
-import { Box, Button, Form, FormField, Grommet, Text } from 'grommet';
+import {
+  Box,
+  Button,
+  Form,
+  FormField,
+  Grommet,
+  Text,
+  TextInput,
+} from 'grommet';
 import { grommet } from 'grommet/themes';
+import { deepMerge } from 'grommet/utils';
 
-const FormFieldLabel = props => {
-  const { required, label, ...rest } = props;
-  return (
-    <FormField
-      label={
-        required ? (
-          <Box direction="row">
-            <Text>{label}</Text>
-            <Text color="status-critical">*</Text>
-          </Box>
-        ) : (
-          label
-        )
-      }
-      required={required}
-      {...rest}
-    />
-  );
-};
+const myTheme = deepMerge(grommet, {
+  formField: {
+    label: {
+      requiredAsterisk: true,
+    },
+  },
+});
 
 export const RequiredLabel = () => (
-  <Grommet theme={grommet}>
+  <Grommet theme={myTheme}>
     <Box align="center" pad="large">
       <Form>
-        <FormFieldLabel name="firstName" label="FirstName" required />
-        <FormFieldLabel name="LastName" label="LastName" required />
-        <FormFieldLabel name="email" label="Email" />
+        <FormField
+          name="firstName"
+          label="First Name"
+          htmlFor="firstName"
+          required
+        >
+          <TextInput id="firstName" name="firstName" />
+        </FormField>
+        <FormField
+          name="lastName"
+          label="Last Name"
+          htmlFor="lastName"
+          required
+        >
+          <TextInput id="lastName" name="lastName" />
+        </FormField>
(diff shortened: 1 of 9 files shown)


src/js/components/FormField/FormField.js
```diff
@@ -356,6 +356,12 @@ const FormField = forwardRef(
           }
         : {};
 
+    let { requiredAsterisk } = theme.formField.label;
+    if (requiredAsterisk === true)
+      // a11yTitle necessary so screenreader announces as "required"
+      // as opposed to "star"
```
halocline, line 362: Is there documentation to reference this?

src/js/components/Form/stories/RequiredLabel.js
```diff
@@ -2,34 +2,23 @@ import React from 'react';
...
 
 export const RequiredLabel = () => (
-  <Grommet theme={grommet}>
+  <Grommet theme={customTheme}>
     <Box align="center" pad="large">
       <Form>
-        <FormFieldLabel name="firstName" label="FirstName" required />
-        <FormFieldLabel name="LastName" label="LastName" required />
-        <FormFieldLabel name="email" label="Email" />
+        <FormField name="firstName" label="First Name" required />
```
halocline, line 19: Should these be presented as: 
```
<FormField htmlFor="firstName" ... />
  <TextInput id="firstName" ... />
</FormField>
```
?

I thought we had been trying to promote this pattern as opposed to just `<FormField />`. If just FormField,
```suggestion
        <FormField htmlFor="firstName" id="firstName" name="firstName" label="First Name" required />
```

Either way, I think we want id and htmlFor so the field has an associated label and passes WAVE test.

taysea: Yes, I can enhance to our recommended format.

src/js/themes/base.d.ts
```diff
@@ -729,7 +729,10 @@ export interface ThemeType {
       container?: BoxProps;
       icon?: any;
     };
-    label?: TextProps;
+    label?: {
+      margin?: MarginType;
+      requiredAsterisk?: boolean | JSX.Element | string;
```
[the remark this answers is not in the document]
taysea: I like that. I'll update.

src/js/components/Form/stories/RequiredLabel.js
taysea: Done.

src/js/themes/base.d.ts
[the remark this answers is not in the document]
taysea: Done.

src/js/components/FormField/FormField.js
taysea: Done.

halocline (approved):
Looks good to me.

src/js/components/FormField/__tests__/FormField-test.js
```diff
@@ -362,4 +362,45 @@ describe('FormField', () => {
     const tree = component.toJSON();
     expect(tree).toMatchSnapshot();
   });
+
+  test('should render asterisk when requiredIndicator === true', () => {
```
ShimiSun, line 366: We are trying to stay away from using `renderer`, please take a look at the test pattern of 
`const { container } = render(...)` instead, and see if it can work as well. 
It will keep us more up to date with using `@testing-library/react`.

src/js/themes/base.d.ts
```diff
@@ -729,7 +729,10 @@ export interface ThemeType {
       container?: BoxProps;
       icon?: any;
     };
-    label?: TextProps;
+    label?: {
```
ShimiSun, line 732: According to the changed file, `label` used to accept all types of Text props, any reason we are singling out only two options for props now? was the implementation wrong earlier?

taysea: You're right, all TextProps is fine, but we'd also like to allow `requiredIndicator` in addition. Looking into how to do that in the declaration file.

src/js/components/FormField/__tests__/FormField-test.js
taysea: Yes, I'll update the tests accordingly.

taysea: Done.

src/js/themes/base.d.ts
taysea: Done. Created `FormFieldLabelType` interface that extends TextProps and adds requiredIndicator.

ericsoderberghp (approved):

PR #4907, opened 2021-01-28
Changed Meter to accept a single value property

ericsoderberghp:
#### What does this PR do?

Changed Meter to accept a single value property and a color property.
This makes it easier to use for the 80% case where there is only one value.

index.d.ts

Revised the "Bar" storybook to use it.
Added a unit test.

Bar story

automatic

yes

backwards compatible

### DIFF
--- a/src/js/components/Meter/index.d.ts
+++ b/src/js/components/Meter/index.d.ts
@@ -11,6 +11,7 @@ export interface MeterProps {
   a11yTitle?: A11yTitleType;
   alignSelf?: AlignSelfType;
   background?: string | { color?: string; opacity?: OpacityType };
+  color?: string;
   gridArea?: GridAreaType;
   margin?: MarginType;
   max?: number;
@@ -18,6 +19,7 @@ export interface MeterProps {
   size?: 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | 'full' | string;
   thickness?: 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | string;
   type?: 'bar' | 'circle';
+  value?: number;
   values?: {
     color?: string;
     highlight?: boolean;
--- a/src/js/components/Meter/Meter.js
+++ b/src/js/components/Meter/Meter.js
@@ -18,15 +18,25 @@ const Meter = forwardRef(
   (
     {
       background = { color: 'light-2', opacity: 'medium' },
+      color,
       size = 'medium',
       thickness = 'medium',
       type = 'bar',
-      values,
+      value,
+      values: valuesProp,
       ...rest
     },
     ref,
   ) => {
+    // normalize values to an array of objects
+    const values = useMemo(() => {
+      if (valuesProp) return valuesProp;
+      if (value) return [{ color, value }];
+      return [];
+    }, [color, value, valuesProp]);
+
     const memoizedMax = useMemo(() => deriveMax(values), [values]);
+
     let content;
     if (type === 'bar') {
       content = (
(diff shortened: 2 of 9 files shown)


ShimiSun (commented):
Why not leverage the existing values prop to accept a single value? 
I’m concerned similar name props will cause confusion of when to use which prop.
same for leveraging `background` vs creating a new `color` prop.

ericsoderberghp:
> same for leveraging `background` vs creating a new `color` prop.

`background` is the "track", `color` is the value piece only. So, a value of 0 will show no color and a value of 100 will show no background.

ericsoderberghp:
> Why not leverage the existing values prop to accept a single value?

Something just feels weird about `values={20}`. Trying to figure out if there's another way to address it. We're a bit stuck in that the current API is more oriented towards the 20% use cases than the 80% ones. :(

ShimiSun:
One thought, deprecate `values` and use `value` instead. So any calls to `values` will fall back to `value` for backward compatibility.

src/js/components/Meter/index.d.ts
```diff
@@ -11,13 +11,15 @@ export interface MeterProps {
...
   background?: string | { color?: string; opacity?: OpacityType };
+  color?: string;
   gridArea?: GridAreaType;
   margin?: MarginType;
   max?: number;
   round?: boolean;
   size?: 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | 'full' | string;
   thickness?: 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | string;
   type?: 'bar' | 'circle';
+  value?: number;
```
taysea, line 22: If I have a single value but want to have so `onHover` or `onClick` behavior, do I still have to use `values`? Or does this fall outside of the 80% use case where people just want to display a single value and no interaction is desired

ericsoderberghp: For that scenario, the default react/browser behavior of event listeners on the DOM element will work without grommet needing to do anything. Although, for a single value Meter, onClick might be better represented in the DOM via `<Button onClick={}><Meter /></Button>`.

ericsoderberghp:
> One thought, deprecate `values` and use `value` instead. So any calls to `values` will fall back to `value` for backward compatibility.

After thinking about this some more, I'm inclined to keep them separate, as I think this aligns more closely with how we think. It also aligns with how the `date` and `dates` properties work in Calendar.

ShimiSun:
> > One thought, deprecate `values` and use `value` instead. So any calls to `values` will fall back to `value` for backward compatibility.
> 
> After thinking about this some more, I'm inclined to keep them separate, as I think this aligns more closely with how we think. It also aligns with how the `date` and `dates` properties work in Calendar.

Interesting comparison, I had an in mind similar comparison of DateInput `value` which could be a single date string or array of string date, similar to how I thought Meter `value` should work. 
Do you think Calendar `date` and `dates` is a better comparison than DateInput `value`?

ericsoderberghp:
> > > One thought, deprecate `values` and use `value` instead. So any calls to `values` will fall back to `value` for backward compatibility.
> > 
> > 
> > After thinking about this some more, I'm inclined to keep them separate, as I think this aligns more closely with how we think. It also aligns with how the `date` and `dates` properties work in Calendar.
> 
> Interesting comparison, I had an in mind similar comparison of DateInput `value` which could be a single date string or array of string date, similar to how I thought Meter `value` should work.
> Do you think Calendar `date` and `dates` is a better comparison than DateInput `value`?

Well, DateInput needs to be tightly aligned with how form input controls work, with `value` and `onChange`, as with TextInput, Select, CheckBoxGroup, etc.

Meter feels more akin to the other Visualization components, to me.

PR #4909, opened 2021-01-29
Adds onChange to TextInput types

abnersajr:
#### What does this PR do?

It adds `onChange` type definition for the `TextInput` comment.

`TextInput/index.d.ts`

Manual test to check if the type is allowing `onChange` properly.

Use the `TextInput` passing a onChange without any error.

Recently using the component I had to extend it.

```
textInput: TextInputProps & {
    onChange: (e: React.ChangeEvent<HTMLInputElement>) => void
  }
```

No

Not sure.

No.


src/js/components/TextInput/index.d.ts
```diff
@@ -23,6 +23,7 @@ export interface TextInputProps {
     suggestionIsOpen?: string;
   };
   name?: string;
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
```
ShimiSun, line 26: hi buddy, isn't it the default type of `onChange` by React? we try not to override the react types when unnecessary.

src/js/components/TextInput/index.d.ts
```diff
@@ -42,7 +43,9 @@ export interface TextInputProps {
   value?: string | number;
 }
 
-declare const TextInput: React.ComponentClass<TextInputProps &
-  Omit<JSX.IntrinsicElements['input'], 'onSelect' | 'size' | 'placeholder'>>;
+declare const TextInput: React.ComponentClass<
```
ShimiSun, line 46: I think this should be changed to FC (functional component) vs component class.

abnersajr: Absolutely! I can change it.

src/js/components/TextInput/index.d.ts
abnersajr: So, It's exist in React for sure. However look here:
https://codesandbox.io/s/dreamy-field-0nhrg?file=/src/App.tsx

If I try to use the `TextInputProps` as a type for one object Key, this doesn't work.

abnersajr: I figured-out that might be a issue where the `Omit` with the HTML Event stuff is being declared in the `declare const TextInput`

So when you import the `TextInputProps` interface that is being declared above. This doesn't have this `Omit....`

src/js/components/TextInput/index.d.ts
ShimiSun: This comment was already addressed on a PR that was merged yesterday, so feel free to resolve the conflicts or roll it back.

abnersajr: Done

ShimiSun (commented):
How is this change fixes the problem? Looks like we haven't changed much except of the ordering of things?

abnersajr:
> How is this change fixes the problem? Looks like we haven't changed much except of the ordering of things?

If for some reason you need to extend or use the type from text input, you will need to import it.
Ex:
```
import { TextInputProps } from 'grommet'

type FormField = {
 inputProps: TextInputProps,
 otherProps...
}
```

This TextInputProps export, doesn't have the HTML Stuff typed in the `declare TextInput`
It's a small change, that allows you complete use and re-use the type.

ericsoderberghp:
Want to think about if we want to go this direction and make all of our type references work this way as well.

ShimiSun:
> > How is this change fixes the problem? Looks like we haven't changed much except of the ordering of things?
> 
> If for some reason you need to extend or use the type from text input, you will need to import it.
> Ex:
> 
> ```
> import { TextInputProps } from 'grommet'
> 
> type FormField = {
>  inputProps: TextInputProps,
>  otherProps...
> }
> ```
> 
> This TextInputProps export, doesn't have the HTML Stuff typed in the `declare TextInput`
> It's a small change, that allows you complete use and re-use the type.

Great inputs @abnersajr, thanks for sharing. I'm also inclined to proceed with this direction as it provides a better typing experience for the users, that being said here are a few thoughts:
1. If we are to change the type of the component now (either for TextInput or other components) would that cause backward compatibility issues for users that are already importing and modifying the component's props?
2. In order to maintain consistent behavior across the framework, should we make sure all of the components are behaving in a similar way, and TextInput isn't the only one that is importing the IntrinsicElements. I'm OK proceeding with this route and migrating types gradually, just wanted to hear your inputs on this matter and whether we should create an issue to track this effort.

abnersajr:
> Want to think about if we want to go this direction and make all of our type references work this way as well.

@ericsoderberghp, one of the good things about Grommet is the composability.
The way the types for this component exports/declare the type blocks the import and re-usage of the component type properly.

I think the biggest concern here is to maintain the backward compatibility for who already use the import of the type for some reason like the example.

I would like to hear your opinion about this direction, and looking for suggestions 😁

ericsoderberghp:
> The way the types for this component exports/declare the type blocks the import and re-usage of the component type properly.

I think this is the right direction to head. I would suggest rolling it out slowly, rather than a single big pull request.

abnersajr:
> > The way the types for this component exports/declare the type blocks the import and re-usage of the component type properly.
> 
> I think this is the right direction to head. I would suggest rolling it out slowly, rather than a single big pull request.

I think we could let this PR fly away.
In the next chapters, we could create an issue to track the transition progress. Listing components and checking which needs and which one doesn't need any change on this.

ericsoderberghp:
> I think we could let this PR fly away.

As in close or merge?

ericsoderberghp (approved):

PR #4910, opened 2021-01-29
Provide Checkbox with FormField context

halocline:
#### What does this PR do?

This is a draft for which to solicit reaction. Alternate approach provided here https://github.com/grommet/grommet/pull/4911

Fixes https://github.com/grommet/grommet-theme-hpe/issues/151 which restores focus to Checkboxes when not contained in a FormField.

More specifically, it allows a Checkbox to alter its styling (e.g. focus) based on whether or not it is contained within a FormField.

**Background**
The HPE Design System provides direction to place focus around the FormField (as opposed to the checkbox itself). However, that did not account for Checkboxes used in other contexts such as a standalone instance or when contained in a DataTable.

The desired design is:
- When a Checkbox is contained in a FormField, the [FormField receives the focus](https://www.figma.com/file/7Mm1xDBTOtPHqggEVpaD2N/HPE-Checkbox-Component?node-id=1388%3A77).
- When a Checkbox is not in a FormField, the [Checkbox receives focus](https://www.figma.com/file/7Mm1xDBTOtPHqggEVpaD2N/HPE-Checkbox-Component?node-id=2120%3A252).

**Approach**
This approach creates a FormFieldContext which allows components to know whether they are contained in a FormField or not.

Pros - Solves issue, allowing for contextual Checkbox styling.
Cons - Feels a bit off and is particular to inputs such as Checkbox, Checkbox toggle, RangeInput, and RangeSelector. Concept is limited to FormField - is there benefit to a more generic, adaptable solution?

#### Where should the reviewer start?

Checkbox.js and FormField.js

#### What testing has been done on this PR?

Storybook using the story "input-checkbox--temp" which will be removed from the PR,  but is useful for testing.

#### How should this be manually tested?

Storybook - Checkbox/stories/Temp.js

#### Any background context you want to provide?

From above: "The HPE Design System provides direction to place focus around the FormField (as opposed to the checkbox itself). However, that did not account for Checkboxes used in other contexts such as a standalone instance or when contained in a DataTable.

The desired design is:
- When a Checkbox is contained in a FormField, the FormField receives the focus.
- When a Checkbox is not in a FormField, the Checkbox receives focus."

#### What are the relevant issues?

https://github.com/grommet/grommet-theme-hpe/issues/151 

#### Screenshots (if appropriate)

**Different focus treatment in different contexts**
Screen Shot 2021-01-29 at 11 36 49 AM
Screen Shot 2021-01-29 at 11 37 03 AM
[...]

PR #4911, opened 2021-01-29
provide component context to checkbox

halocline:
#### What does this PR do?

This is a draft for which to solicit reaction. It is an alternate approach to that proposed in https://github.com/grommet/grommet/pull/4910.

Fixes https://github.com/grommet/grommet-theme-hpe/issues/151 which restores focus to Checkboxes when not contained in a FormField.

More specifically, it allows a Checkbox to alter its styling (e.g. focus) based on whether or not it is contained within a FormField.

**Background**
The HPE Design System provides direction to place focus around the FormField (as opposed to the checkbox itself). However, that did not account for Checkboxes used in other contexts such as a standalone instance or when contained in a DataTable.

The desired design is:
- When a Checkbox is contained in a FormField, the [FormField receives the focus](https://www.figma.com/file/7Mm1xDBTOtPHqggEVpaD2N/HPE-Checkbox-Component?node-id=1388%3A77).
- When a Checkbox is not in a FormField, the [Checkbox receives focus](https://www.figma.com/file/7Mm1xDBTOtPHqggEVpaD2N/HPE-Checkbox-Component?node-id=2120%3A252).

**Approach**
This approach creates a ComponentContext/ContainerContext which allows components to know whether they are contained within a specific component.

Pros - Solves issue, allowing for contextual Checkbox styling. More flexible/extendable than what is proposed in https://github.com/grommet/grommet/pull/4910 in that the context could be used in additional situations.
Cons - Is this going too broad? I don't have clear use cases for where else this might be used (I imagine it could be applicable to other layout scenarios where properties of a component change based on where it appears).

Checkbox.js

Storybook using the story "input-checkbox--temp" which will be removed from the PR,  but is useful for testing.

Storybook - Checkbox/stories/Temp.js

From above: "The HPE Design System provides direction to place focus around the FormField (as opposed to the checkbox itself). However, that did not account for Checkboxes used in other contexts such as a standalone instance or when contained in a DataTable.

The desired design is:
- When a Checkbox is contained in a FormField, the FormField receives the focus.
- When a Checkbox is not in a FormField, the Checkbox receives focus."

https://github.com/grommet/grommet-theme-hpe/issues/151 

**Different focus treatment in different contexts**
Screen Shot 2021-01-29 at 11 36 49 AM
Screen Shot 2021-01-29 at 11 37 03 AM
Screen Shot 2021-01-29 at 11 37 17 AM

Backwards compatible.

### DIFF
--- a/src/js/components/CheckBox/stories/Temp.js
+++ b/src/js/components/CheckBox/stories/Temp.js
@@ -19,7 +19,9 @@ const myTheme = deepMerge(hpe, {
   checkBox: {
     check: {
       extend: props =>
-        props.inFormField ? `box-shadow: none` : `box-shadow: undefined`,
+        props.componentContext === 'FormField'
+          ? `box-shadow: none`
+          : `box-shadow: undefined`,
     },
   },
 });
--- a/src/js/components/CheckBox/stories/Temp.js
+++ b/src/js/components/CheckBox/stories/Temp.js
@@ -18,15 +18,8 @@ delete controlledColumns[4].aggregate;
 const myTheme = deepMerge(hpe, {
   checkBox: {
     check: {
-      extend: props => {
-        return `
-          ${
-            props.componentContext === 'FormField'
-              ? `box-shadow: none`
-              : `box-shadow: undefined`
-          };
-        `;
-      },
+      extend: props =>
+        props.inFormField ? `box-shadow: none` : `box-shadow: undefined`,
     },
   },
 });
--- a/src/js/components/CheckBox/stories/Temp.js
+++ b/src/js/components/CheckBox/stories/Temp.js
@@ -1,0 +1,96 @@
+import React, { useState } from 'react';
+
+import { Box, Grommet, CheckBox, FormField, DataTable } from 'grommet';
+// import { grommet } from 'grommet/themes';
+import { deepMerge } from 'grommet/utils';
+import { hpe } from 'grommet-theme-hpe';
+
+// Source code for the data can be found here
+// https://github.com/grommet/grommet/blob/master/src/js/components/DataTable/stories/data.js
+import { columns, DATA } from '../../DataTable/stories/data';
+
+const controlledColumns = columns.map(col => ({ ...col }));
+delete controlledColumns[0].footer;
+delete controlledColumns[3].footer;
+delete controlledColumns[4].footer;
+delete controlledColumns[4].aggregate;
+
+const myTheme = deepMerge(hpe, {
+  checkBox: {
+    check: {
+      extend: props => {
+        return `
+          ${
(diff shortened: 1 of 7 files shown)


src/js/components/CheckBox/stories/Temp.js
```diff
@@ -0,0 +1,89 @@
...
+delete controlledColumns[0].footer;
+delete controlledColumns[3].footer;
+delete controlledColumns[4].footer;
+delete controlledColumns[4].aggregate;
+
+const myTheme = deepMerge(hpe, {
+  checkBox: {
+    check: {
+      extend: props =>
+        props.inFormField ? `box-shadow: none` : `box-shadow: undefined`,
```
taysea, line 22: Maybe I got lost, but should this be check for `componentContext` value of "FormField" in this approach?

taysea (commented):
This approach makes sense, but I feel unsure about providing something so generic when we're not fully clear on other use cases for non-form contexts. Interested to hear Eric's thoughts on this vs the other approach.

src/js/components/CheckBox/stories/Temp.js
halocline: You are correct. Fixed.

taysea: Thanks. Another question on this approach. Is ComponentContext something that would just live internally, or is it a Context, like responsive/theme contexts that we'd exposed to users? For example, I want my buttons in my navigation bar to be styled like X but want them to be styled in my main "blog" area as Y.

Is that something we want to introduce or is does that create too many ways to theme? (posing this as an open-ended question)

halocline: I think it has the potential to serve externally and be consumed similarly to ResponsiveContext. With ResponsiveContext, I choose to affect prop values, layout, or styling based on screen width. Similarly, I could see ComponentContext allowing me to affect prop values, layout, and styling based on its container. I think your navigation vs. "blog" is a good example. Another example might be if I have a Notifications component that I'd like to re-use throughout my app. That notification behavior may vary based on where the notification is surfaced.

In a way, TableCell is applying a similar concept; when I am in a TableHeader context, do X; when I am in a TableBody context, do Y ... . Different implementation, but similar in concept that my component's properties may change based on their container.

That said, the direction should be informed and driven by concrete use cases rather than speculative.

halocline: Another use case would be situations where I may want more granular control to alter my component based on its container's dimensions. For example, ResponsiveContext lets me do that based on window dimension, however sometimes I want my component to behave differently when contained in a small Box vs. a medium Box.

halocline:
Closing in favor of this approach: https://github.com/grommet/grommet/pull/4930

Issue #4870, opened 2021-01-07
Ability to provide search value to Select


jasnross:
### Expected Behavior

I would like to have a way to control the search value for the Select component so that I can ensure that the search term is controlled (controlled as in a "controlled input") and always in sync with my external state that is tracking the search term.

For example, it would be nice if we could do:

```javascript
const [searchValue, setSearchValue];

<Select
  onSearch={value => setSearchValue(value)}
  searchValue={searchValue}
/>
```

One use-case for this is being able to clear the search term according to logic outside the Select.

### Actual Behavior

Currently it seems that the Select component's search value is controlled entirely within the Select component, and I don't see any way to pass the current value of the search field to the component instead.

See this Codepen for an example: https://codesandbox.io/s/dry-resonance-7ceyo?file=/src/App.js

### Your Environment

- Grommet version: 2.16.2
- Browser Name and version: Firefox 84.0.1
- Operating System and version (desktop or mobile): MacOS 10.14.6

joaoGabriel55:
Can I work on this? I think found a good solution

joaoGabriel55:
Almost done guys...

PR #4924, opened 2021-02-03
Added search value to Select

joaoGabriel55:
#### What does this PR do?

Added a search value prop for the Select component so ensure that the search term is controlled (controlled as in a "controlled input") and always in sync with some external state that is tracking the search term.

First of all, access the storybook and see the new feature.

The test wrote is called `search with search value prop`. I think it need some improvement. I am open to suggestions, cause I have few familiarity with grommet's test cases

Test the Select component inform some value to prop `searchValue`. Something like this:

image

So, You will see that search input will filled with prop informed.

I think not

I think so.

Issue #4870

I think not

Waiting your feedback.

### DIFF
--- a/src/js/components/Select/SelectContainer.js
+++ b/src/js/components/Select/SelectContainer.js
@@ -342,6 +342,13 @@ const SelectContainer = forwardRef(
                 type="search"
                 value={search || ''}
                 placeholder={searchPlaceholder}
+                onFocus={() => {
+                  if (search) {
+                    setSearch(search);
+                    setActiveIndex(-1);
+                    onSearch(search);
+                  }
+                }}
                 onChange={event => {
                   const nextSearch = event.target.value;
                   setSearch(nextSearch);
--- a/src/js/components/Select/__tests__/Select-test.js
+++ b/src/js/components/Select/__tests__/Select-test.js
@@ -201,6 +201,27 @@ describe('Select', () => {
     expect(window.scrollTo).toBeCalled();
   });
 
+  test('search with search value prop', () => {
+    jest.useFakeTimers();
+    const onSearch = jest.fn();
+    const { container } = render(
+      <Select
+        id="test-select"
+        options={['one', 'two']}
+        onSearch={onSearch}
+        searchValue="two"
+      />,
+    );
+    expect(container.firstChild).toMatchSnapshot();
+
+    // advance timers so select can open
+    act(() => {
+      jest.advanceTimersByTime(200);
+    });
+
+    expect(document.activeElement).toMatchSnapshot();
+  });
+
   test('size', () => {
     const component = renderer.create(
       <Select
--- a/src/js/components/Select/index.d.ts
+++ b/src/js/components/Select/index.d.ts
@@ -40,6 +40,7 @@ export interface SelectProps {
   onMore?: (...args: any[]) => any;
   onOpen?: (...args: any[]) => any;
   onSearch?: (search: string) => void;
+  searchValue?: string | JSX.Element | number | object;
   options: (string | boolean | number | JSX.Element | object)[];
   open?: boolean;
   placeholder?: PlaceHolderType;
--- a/src/js/components/Select/Select.js
+++ b/src/js/components/Select/Select.js
@@ -73,6 +73,7 @@ const Select = forwardRef(
       onMore,
       onOpen,
       onSearch,
+      searchValue,
       open: propOpen,
       options: optionsProp,
       placeholder,
@@ -115,7 +116,8 @@ const Select = forwardRef(
     // to be referenced when filtered by search.
     useEffect(() => {
       if (!search) setAllOptions(optionsProp);
-    }, [optionsProp, search]);
+      if (searchValue) setSearch(searchValue);
+    }, [optionsProp, search, searchValue]);
 
     // the option indexes present in the value
     const optionIndexesInValue = useMemo(() => {
--- a/src/js/components/Select/stories/Search.js
+++ b/src/js/components/Select/stories/Search.js
@@ -11,6 +11,7 @@ for (let i = 1; i <= 200; i += 1) {
 export const Search = () => {
   const [options, setOptions] = useState(defaultOptions);
   const [value, setValue] = useState('');
+  const [searchValue, setSearchValue] = useState('');
   const [valueMultiple, setValueMultiple] = useState([]);
 
   return (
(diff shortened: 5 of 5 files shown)


src/js/components/Select/__tests__/Select-test.js
```diff
@@ -201,6 +201,27 @@ describe('Select', () => {
...
+    expect(container.firstChild).toMatchSnapshot();
+
+    // advance timers so select can open
+    act(() => {
+      jest.advanceTimersByTime(200);
+    });
+
+    expect(document.activeElement).toMatchSnapshot();
+  });
+
```
negreirosleo, line 224: Hi Gabriel, that test isnt very effective, I recommend this approach:

```js
  test('search with search value prop', () => {
    const onSearch = jest.fn();
    const { getByPlaceholderText,getByTestId } = render(
      <Select
        id="test-select"
        options={['one', 'two']}
        onSearch={onSearch}
        searchValue="two"
        searchPlaceholder="search-placeholder"
        data-testid="test"
      />,
    );
    fireEvent.click(getByTestId('test'));
    const searchInput = getByPlaceholderText('search-placeholder');
    expect(searchInput.value).toBe("two");
  });
  ```

Comment (joaoGabriel55) [reply src/js/components/Select/__tests__/Select-test.js:224]: Nice! Thanks a lot @negreirosleo.

File: src/js/components/Select/SelectContainer.js [hunk 2]
```diff
@@ -342,6 +342,13 @@ const SelectContainer = forwardRef(
                 type="search"
                 value={search || ''}
                 placeholder={searchPlaceholder}
+                onFocus={() => {
+                  if (search) {
+                    setSearch(search);
+                    setActiveIndex(-1);
+                    onSearch(search);
+                  }
+                }}
```
negreirosleo, line 351: This code is also used on the onChange, since they are the same behaviour maybe we could have a function that holds it, something like:

```js
const executeSearch = useCallback( search =>
  setSearch(search);
  setActiveIndex(-1);
  onSearch(search)
  ,[ hook dependencies here ]
  )
  ...
onFocus={() => search && executeSearch(search)}
onChange={event => executeSearch(event.target.value)}
```

negreirosleo:
@joaoGabriel55 great work! Just one more thing, how about creating a specific story for that new feature? like: `SearchWithControlledInput.js`, I believe it will make better for the users to check on the search capabilities!

src/js/components/Select/__tests__/Select-test.js
```diff
@@ -201,6 +201,25 @@ describe('Select', () => {
     expect(window.scrollTo).toBeCalled();
   });
 
+  test('search with search value prop', () => {
+    jest.useFakeTimers();
```
negreirosleo, line 205: Hey, no need for this fakeTimer here!

negreirosleo (approved):
LGTM!

src/js/components/Select/index.d.ts
```diff
@@ -40,6 +40,7 @@ export interface SelectProps {
   onMore?: (...args: any[]) => any;
   onOpen?: (...args: any[]) => any;
   onSearch?: (search: string) => void;
+  searchValue?: string | JSX.Element | number | object;
```
ericsoderberghp, line 43:
```suggestion
  searchValue?: string;
```

ericsoderberghp: I think the `searchValue` can only be a string.

src/js/components/Select/SelectContainer.js
```diff
@@ -342,12 +351,8 @@ const SelectContainer = forwardRef(
                 type="search"
                 value={search || ''}
                 placeholder={searchPlaceholder}
-                onChange={event => {
-                  const nextSearch = event.target.value;
-                  setSearch(nextSearch);
-                  setActiveIndex(-1);
-                  onSearch(nextSearch);
-                }}
+                onFocus={() => search && executeSearch(search)}
```
ericsoderberghp, line 354: Why are we updating the search value `onFocus`?

src/js/components/Select/SelectContainer.js
```diff
@@ -320,6 +320,15 @@ const SelectContainer = forwardRef(
         }
       : {};
 
+    const executeSearch = useCallback(
```
ericsoderberghp, line 323: We don't use the term "execute" in a context like this. Instead, we tend to rename the property passed in to "onSearchProp" and then call this internal callback function "onSearch".

src/js/components/Select/SelectContainer.js
joaoGabriel55: Because I guess that is good to allow the dev set some default  `searchValue` and when user click on Select, already filter options based on this default  `searchValue`  by  `onFocus` event.

src/js/components/Select/__tests__/Select-test.js
```diff
@@ -201,6 +201,24 @@ describe('Select', () => {
...
+        searchValue="two"
+        searchPlaceholder="search-placeholder"
+        data-testid="test"
+      />,
+    );
+    fireEvent.click(getByTestId('test'));
+
+    const searchInput = getByPlaceholderText('search-placeholder');
+    expect(searchInput.value).toBe('two');
+  });
```
jcfilben, line 220: I think a better test would be to actually change the value of `searchValue` outside of the render function and then expect is toBe the new value and check that the items in the dropdown match what we are expecting to see from the `searchValue`

jcfilben (commented):
Doesn't look like the functionality is working as expected. When I tested with your storybook example the search was saying 'no matches found' when there should have been matches.
Apr-29-2021 11-48-53

jcfilben:
Still looks like the functionality isn't quite there yet. I am seeing some strange behavior in the story you added when you search no options are shown.
May-26-2021 11-40-29

joaoGabriel55:
> Still looks like the functionality isn't quite there yet. I am seeing some strange behavior in the story you added when you search no options are shown.
> May-26-2021 11-40-29

Oh! Sorry, I will take a look.

jcfilben:
Thanks for updating the story! Can you also fix the merge conflicts? (Just merge master into your branch and fix conflicts locally and then push up to GitHub)

src/js/components/Select/README.md
```diff
@@ -468,6 +468,15 @@ number
 ]
 ```
 
+**searchValue**
+
+Search Value text to use in the search box to
```
taysea, line 473: Docs suggestion: Text to use in the search input to filter Select options.

taysea (commented):
Looking good here! Left a docs suggestion and merge conflicts need to be resolved.

halocline (commented):
Looks good overall. Thank you, @joaoGabriel55.

Looks like a couple minor items to address and then this should be good for a final review.

jcfilben:
Hey @joaoGabriel55, we would love to get this PR merged. I think the checks are failing due to some references to outdated stories (specifically the Themed story for Select which no longer exists). I think if we clean up the stories and update the documentation to move to grommet-site this PR will be ready. Let me know if you have questions

jkalberer:
Could I commandeer this branch? 
The current experience where the `search` text is automatically cleared after selecting items is not ideal.

jcfilben:
Hi @jkalberer thanks for your interest in this! I think you could take the changes here and put them into a new branch and open a new PR from that branch.

jcfilben:
Hi @jkalberer just wanted to check in and see if you were still interested in opening a new PR for this

jcfilben:

PR #4930, opened 2021-02-05
Allow CheckBox in a FormField to control where focus is applied

halocline:
#### What does this PR do?

This is a draft for which to solicit reaction. It is an alternate approach to that proposed in https://github.com/grommet/grommet/pull/4910 && https://github.com/grommet/grommet/pull/4911.

Fixes https://github.com/grommet/grommet-theme-hpe/issues/151 which restores focus to Checkboxes when not contained in a FormField.

More specifically, it allows a Checkbox to alter its styling (e.g. focus) based on whether or not it is contained within a FormField.

**Background**
The HPE Design System provides direction to place focus around the FormField (as opposed to the checkbox itself). However, that did not account for Checkboxes used in other contexts such as a standalone instance or when contained in a DataTable.

The desired design is:
- When a Checkbox is contained in a FormField, the [FormField receives the focus](https://www.figma.com/file/7Mm1xDBTOtPHqggEVpaD2N/HPE-Checkbox-Component?node-id=1388%3A77).
- When a Checkbox is not in a FormField, the [Checkbox receives focus](https://www.figma.com/file/7Mm1xDBTOtPHqggEVpaD2N/HPE-Checkbox-Component?node-id=2120%3A252).

**Approach**
This approach passes `focusIndicator` from FormField to CheckBox, similar to other inputs such as TextInput and TextArea and [suggested here](https://github.com/grommet/grommet-theme-hpe/issues/151#issuecomment-772219903).

Pros - Solves issue, allowing for contextual Checkbox styling. Leverages existing pattern for other Grommet inputs contained in a FormField.
Cons - A little awkward to achieve backwards compatibility.

Checkbox.js

Storybook using the story "input-checkbox--temp" which will be removed from the PR,  but is useful for testing.

Storybook - Checkbox/stories/Temp.js

From above: "The HPE Design System provides direction to place focus around the FormField (as opposed to the checkbox itself). However, that did not account for Checkboxes used in other contexts such as a standalone instance or when contained in a DataTable.

The desired design is:
- When a Checkbox is contained in a FormField, the FormField receives the focus.
- When a Checkbox is not in a FormField, the Checkbox receives focus."

https://github.com/grommet/grommet-theme-hpe/issues/151 

**Different focus treatment in different contexts**
Screen Shot 2021-01-29 at 11 36 49 AM
Screen Shot 2021-01-29 at 11 37 03 AM
Screen Shot 2021-01-29 at 11 37 17 AM

Yes.

Backwards compatible.


src/js/components/CheckBox/stories/Temp.js
```diff
@@ -0,0 +1,179 @@
...
+delete controlledColumns[0].footer;
+delete controlledColumns[3].footer;
+delete controlledColumns[4].footer;
+delete controlledColumns[4].aggregate;
+
+const myTheme = deepMerge(hpe, {
+  checkBox: {
+    check: {
+      extend: props =>
+        props.inFormField ? `box-shadow: none` : `box-shadow: undefined`,
```
taysea, line 29: Why do we still need this is doing the `focusIndicator` flag?

src/js/components/CheckBox/StyledCheckBox.js
```diff
@@ -89,7 +92,10 @@ const StyledCheckBoxToggle = styled.span`
       ? normalizeColor(props.theme.checkBox.toggle.background, props.theme)
       : 'transparent'};
 
-  ${props => props.focus && focusStyle()};
+  ${props =>
+    props.focusIndicator === false
+      ? props.focus && props.theme.checkBox.focusIndicator && focusStyle()
```
taysea, line 97: It looks like theme value is overriding props here. is that true? I would anticipate if focusIndicator is false, then focusStyle should not be applied.

src/js/components/CheckBox/stories/Temp.js
halocline: Good call. The conditional can be removed as it is irrelevant and always false (it was a leftover from alternate approach). I will clean up.

The `box-shadow: undefined` is needed to undo the `box-shadow: none` which is in the HPE theme currently and would be removed.

src/js/components/CheckBox/StyledCheckBox.js
halocline: I agree that this logic is confusing. It gets convoluted trying to maintain backwards compatibility.

Ideally this line would read
```
${props => props.focus && props.focusIndicator !== false && focusStyle()};
```
but then we lose backwards compatibility. In the grommet theme, currently both the checkbox and the formfield bottom border get focus styling:
Screen Shot 2021-02-05 at 9 03 48 AM
but if we implement as shown above, we lose the focus around the checkbox and only the formfield bottom border receives the treatment:
Screen Shot 2021-02-05 at 9 04 09 AM

So, if we want to avoid this, then we need some sort of opt in mechanism. So, this is trying to say:
"IF focusIndicator is false, THEN check whether I have opted into this behavior." I will play around with alternatives, but open to ideas if you have any.

ericsoderberghp: I would be OK with the grommet theme behavior having the focus indicator only on the bottom line, as I think that fits with the general behavior of focus in a form context. You could argue that the current behavior is a bug, being inconsistent.

halocline: Yes, I had thought that as well, but wanted to bring visibility and discussion.

If we adapt that view, it eliminates the confusing aspects here. Proceed with the "dual focus is a bug" viewpoint? If so, I will clean up.

src/js/components/CheckBox/stories/Temp.js
halocline: Done.

src/js/components/CheckBox/StyledCheckBox.js
halocline: @ShimiSun , would appreciate your input on whether we should look at the focus treatment of a CheckBox in a FormFiled as a bug since it is inconsistent with the treatment for other inputs? Or, whether we should maintain the current treatment?

src/js/themes/base.d.ts
```diff
@@ -512,6 +512,7 @@ export interface ThemeType {
     };
     color?: ColorType;
     extend?: ExtendType;
+    focusIndicator: boolean;
```
ShimiSun, line 515: I don't think we should force the user to use this prop.
```suggestion
    focusIndicator?: boolean;
```

src/js/components/CheckBox/stories/Temp.js
```diff
@@ -0,0 +1,181 @@
...
+  Box,
+  Grommet,
+  CheckBox,
+  FormField,
+  DataTable,
+  ThemeContext,
+} from 'grommet';
+import { grommet } from 'grommet/themes';
+import { deepMerge } from 'grommet/utils';
+import { hpe } from 'grommet-theme-hpe';
```
ShimiSun, line 13: Let's remove this import after the PR is approved, it's good for POC but not needed on a real deployment, we can use test-cafe for this type of testing.

src/js/components/CheckBox/stories/Temp.js
```diff
@@ -0,0 +1,181 @@
...
+
+// Source code for the data can be found here
+// https://github.com/grommet/grommet/blob/master/src/js/components/DataTable/stories/data.js
```
ShimiSun, line 16: Thanks for adding an import.

src/js/components/CheckBox/StyledCheckBox.js
ShimiSun: I think that the current behavior of the theme is a bug and I'm OK fixing it without worrying about the backward compatibility aspect -- in case it affects only the HPE theme users. 
Few questions:
1. Is this issue also reproducible for a single RadioButton outside of Form content?
2. Would there be any b/c issues for grommet itself not in the context of HPE?
3. Would the `focusIndicator` behavior be consistent with other input elements when used with `plain`?

halocline: > 1. Is this issue also reproducible for a single RadioButton outside of Form content?
RadioButton has its own focus issues which should be looked at. For example noticed how focus is inconsistently applied in [this example](https://storybook.grommet.io/?path=/story/input-radiobuttongroup-simple--simple).

> 2. Would there be any b/c issues for grommet itself not in the context of HPE?
The backwards compatibility vs. bug question is being raised because it does affect Grommet users not in the context of HPE. That is the decision which needs to be made in order to inform what direction this PR takes.

> 3. Would the `focusIndicator` behavior be consistent with other input elements when used with `plain`?
That is the intent.

src/js/components/CheckBox/stories/Temp.js
halocline: I think we should remove this story altogether as I don't see it adding value outside the context of this PR demonstrating the differences. I had just included it to help reviewers visualize the different behaviors.

> #### What testing has been done on this PR?
> Storybook using the story "input-checkbox--temp" which will be removed from the PR, but is useful for testing.
> 
> #### How should this be manually tested?
> Storybook - Checkbox/stories/Temp.js
>

src/js/components/CheckBox/StyledCheckBox.js
ShimiSun: I might be an idealist but I would love to leverage the final approach of the fix to fix RadioButton as well so this type of fixes won't be spread across different releases. Regarding the link you have attached of RadioButton it looks like the focus is displayed, so I'm not sure what you are pointing at that I might be missing.

I would like us to have a quick chat about the b/c issue and how it affects grommet users (tomorrow morning?), since the current reported issue is only to HPE users I wonder how far we should go with risking grommet b/c to fix it, hence I would like a discussion there. 

Also, it will be nice to see how the new approach is working with `plain`.

taysea: I'm looking at the RadioButton example linked. I have not looked into the component implementation, but what I see behaviorally is:
1. I tab to the component, but there's no focus indication that I'm on it.
2. I press down arrow and a focus indicator appears on the selected RadioButton.
3. I press down arrow again and the next RadioButton gets selected but no focus indicator is shown.
4. I press down arrow again and the focus indicator appears again.

Overall, getting this back and forth behavior where focus indicator is shown, then removed, then shown again.

I would hope that Matt's work on Checkbox could be leveraged in RadioButton, but it also feels like there are some other funky focus behavior issues at play in RadioButtonGroup. Might take a little bit of separate investigation to understand.

ShimiSun:
I'll be closing this Draft for now, feel free to resolve conflicts and reopen this when it'srelevant.

Issue #4927, opened 2021-02-04
RangeSelector selected bar can't reach ends if dragged very fast and range is large


iicfcii:
### Expected Behavior
When the middle bar is selected and dragged very fast towards one end, the bar should reach the end. 

### Actual Behavior
If the range is large(about 100) and the bar is dragged quickly to one end, the bar cannot reach the end. If dragged slowly, it can reach the end. 

### URL, screen shot, or Codepen exhibiting the issue
This is easily reproducible using Thin RangeSelector in the Storybook. 
https://storybook.grommet.io/?path=/story/input-rangeselector--thin-story

### Your Environment
- Grommet version: 2.16.2
- Browser Name and version: Google Chrome Version 88.0.4324.104 (Official Build) (64-bit)
- Operating System and version (desktop or mobile): Windows 10, desktop. 

### Possible fix
Under `src\js\components\RangeSelector\RangeSelector.js`, I changed line 84 to 86 to below code and it seems to fix the issue but I am not sure if this is the best way to fix it. The added code makes sure the values are updated if max or min is reached. 
```
if (value === max) {
  nextValues = [max-(values[1]-values[0]),max];
} else if (value === min) {
  nextValues = [min,min+(values[1]-values[0])];
} else {
  const delta = value - moveValue;
  if (values[0] + delta >= min && values[1] + delta <= max) {
    nextValues = [values[0] + delta, values[1] + delta];
  }
}
```

ShimiSun:
hi buddy, do you want to file a PR with your potential fix? It will help us speed up the fix process.

PR #4931, opened 2021-02-05
Made sure RangeSelector values are updated when bar is dragged quickly to ends

iicfcii:
#### What does this PR do?
This PR made sure values of RangeSelector are updated when its bar is dragged quickly to max or min end. If the drag is too fast, values may not be updated in time. The previous method using `delta` won't apply to the correct values. 
The only file changed.
Manual testing using storybook. 
Test and compare the behavior of the Thin RangeSelector example in storybook. 
NA
#4927
NA
No.
Probably not. 


ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution!

PR #4932, opened 2021-02-05
Improve Layer + Target scroll behavior

taysea:
#### What does this PR do?
Updates the way layer is styled to move with a target to remove scroll lag. Previously, Layer bounds were being handled by a state variable, so every time a user scrolled or resized the browser, a state variable was being updated and then affecting the CSS. This was a costly computation which caused the movement of the Layer to lag behind the movement of the target.

These updates align the behavior of Layer positioning closer to how Drop behaves (where no scroll lag is exhibited). Mainly, instead of relying on a state variable, the layer container style is now adjusted directly in the scroll/resize listener functions.

Some other small changes to take note of:
1. When Layer (modal or non-modal) has a target, it should be positioned `absolute` to allow it to align properly and move with its target.
2. A CSS property of `will-change: transform` is added to LayerOverlay in order to improve scroll performance. Prior, the overlay was being repainted on every render which also could have been contributing to some lag. This CSS property enhances the way the browser renders this element so it is not repainted on scroll.

Snapshot changes are merely additions of the `will-change` property and addition of the Layer wrapper to Layers with `modal={false} + target`.

Tested scroll behavior with: https://storybook.grommet.io/?path=/story/layout-layer--target-layer
and checked other Layer stories to ensure no changed behavior.

Storybook deploy of "Layer Target" story. Open layer, (modal or non-modal), and scroll. No lag should be seen.

Related to https://github.com/grommet/hpe-design-system/issues/1115

No.

Enhanced Layer + `target` scroll behavior to remove scroll lag

Backwards compatible.


halocline (approved):
This looks really good. Nice work.

Behavior across browsers looks great. Confirmed snapshot changes are all valid.

ericsoderberghp (approved):

ShimiSun:
@halocline can you please share what are the snapshot changes you were seeing on the site?

ShimiSun:
The build failures are not related to this PR, we are working on fixing it internally.

halocline:
> @halocline can you please share what are the snapshot changes you were seeing on the site?

I reviewed the Jest snapshot changes, just to confirm the changes were related to the new CSS property `will-change` which affected all snapshots (appropriately adding `will-change` and new class ids), and the other changes were related to snapshots where `layerTarget` is being used, but `modal === false`.

These were all valid. I did not see any other changes.

ShimiSun (approved):

PR #4933, opened 2021-02-05
Changed CheckBox to add defaultChecked

ericsoderberghp:
#### What does this PR do?

Changed CheckBox to add defaultChecked

index.d.ts

manual story + added unit test

aligns with how defaultValue can be used for TextInput and Select.

automatic

yes

backwards compatible

### DIFF
--- a/src/js/components/CheckBox/doc.js
+++ b/src/js/components/CheckBox/doc.js
@@ -20,6 +20,9 @@ export const doc = CheckBox => {
     checked: PropTypes.bool
       .description('Same as React <input checked={} />')
       .defaultValue(false),
+    defaultChecked: PropTypes.bool
+      .description('Same as React <input defaultChecked={} />')
+      .defaultValue(false),
     disabled: PropTypes.bool
       .description(
         `Same as React <input disabled={} />. Also adds a hidden input element
--- a/src/js/components/CheckBox/doc.js
+++ b/src/js/components/CheckBox/doc.js
@@ -20,9 +20,6 @@ export const doc = CheckBox => {
     checked: PropTypes.bool
       .description('Same as React <input checked={} />')
       .defaultValue(false),
-    defaultChecked: PropTypes.bool
-      .description('Same as React <input defaultChecked={} />')
-      .defaultValue(false),
     disabled: PropTypes.bool
       .description(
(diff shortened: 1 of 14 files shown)


halocline (approved):
Looks good. No concerns.

Worked as expected. Snapshot ✔️ . Type ✔️ . Docs ✔️ .

src/js/components/CheckBox/doc.js
```diff
@@ -20,6 +20,9 @@ export const doc = CheckBox => {
     checked: PropTypes.bool
       .description('Same as React <input checked={} />')
       .defaultValue(false),
+    defaultChecked: PropTypes.bool
+      .description('Same as React <input defaultChecked={} />')
+      .defaultValue(false),
```
ShimiSun, line 25: I thought our policy was not to document any React's given prop unless we modify them internally

src/js/components/CheckBox/index.d.ts
```diff
@@ -4,6 +4,7 @@ import { A11yTitleType } from '../../utils';
 export interface CheckBoxProps {
   a11yTitle?: A11yTitleType;
   checked?: boolean;
+  defaultChecked?: boolean;
```
ShimiSun, line 7: If this is out of the box React input prop, it should be added automatically by specifying `JSX.IntrinsicElements['input']`. we only need the explicit addition to the type file when we are applying existing prop modification, and if that is the case, it should also be added to the TS omit functionality.

src/js/components/CheckBox/doc.js
halocline: Good to know that policy/perspective. Would like to know the outcome of this discussion.

ShimiSun: Following the thread should do the trick.

ericsoderberghp: Well, I added it in because we are actually doing something with it, a bit like `defaultValue` for Select. However, I agree from an API point of view, the caller isn't being given any new information or behavior that they need to care about. I will remove it.

src/js/components/CheckBox/index.d.ts
ericsoderberghp: removed

ShimiSun (approved):

Issue #4934, opened 2021-02-05
DataTable InfiniteScroll onMore isn't being called when data resets


TylerKienVu:
When resetting the `data` on the `DataTable` after the `DataTable` has already called `onMore` from scrolling down at least once, the `DataTable` will no longer call `onMore`.

### Expected Behavior

The DataTable should still have the same behavior of calling `onMore` when it scrolls to the bottom.

### Actual Behavior

The DataTable stops calling `onMore` when it scrolls to the bottom after a data reset.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/datatable-infinitescroll-reset-issue-6wbxc?file=/src/Example.jsx

### Steps to Reproduce

1. Scroll down in the data table until you see 2 console.logs (this means 20 items have loaded, 10 on initial render and then 10 due to the `onMore` being called)
2. Press the **Reset data** button at the top of the page (this will reset the data to have the first 10 data items)
3. Try to scroll down the `DataTable` again
4. `onMore` no longer gets triggered and the `DataTable` is stuck

Here is a gif showing the issue using the codebox above.

![2021-02-05 15 14 39](https://user-images.githubusercontent.com/15975780/107098801-f5cc7400-67c4-11eb-9ea4-2af7198d668a.gif)

### Your Environment

- Grommet version: v2.16.3
- Browser Name and version: Google Chrome Version 87.0.4280.141 (Official Build) (x86_64)
- Operating System and version (desktop or mobile):macOS Catalina Version 10.15.4

leossantiago:
Addressed :) #4936 

My only concern is that this is a really specific use case. I wonder if we should support dynamic changes to the data of a DataTable with `onMore`. I'd love to hear some inputs on that.

TylerKienVu:
> Addressed :) #4936
> 
> My only concern is that this is a really specific use case. I wonder if we should support dynamic changes to the data of a DataTable with `onMore`. I'd love to hear some inputs on that.

Thank you for working on this! 

The main reason that we are using this in the project that I'm working on is to "refresh" the data once a mutation happens that would result in needing to requery the data for the table. Our data query works in increments so that would mean grabbing the first page again and starting from the beginning. I'm not sure how many other people encounter this use case but do you think there would be a different solution that would be a better fit for this kind of scenario?

PR #4936, opened 2021-02-08
Handle DataTable data reset with InfiniteScroll

leossantiago:
#### What does this PR do?

Provides a fix for #4934, allowing `onMore` to be triggered when manually resetting `DataTable` data.
Even though it's related to `DataTable`, the fix was made on `InfiniteScroll`, as it's the one handling `onMore` calls.

This issue has a use case that updates data dynamically, but when a manual reset is made, some internal states are kept. This fix verifies a condition that indicates a manual data reset (data length = 0, last page = 0, and pending length != 0). If the condition is valid, it resets pageHeights, pendingLength and renderPageBounds, which will make the DataTable behave just as how it works when it first renders.

Note that the given example starts with `currentData = [ ]`, and this should be consistently followed in order for this to work, meaning that the reset should set `currentData = [ ]`.

`src/js/components/InfiniteScroll/InfiniteScroll.js`

Manual testing using storybook, as this issue involves updates to `BoundingCliendRect`, and I do believe those can't be easily tested using RTL/JSDom.

An updated version of the example provided by the issue reporter. Based on storybook/DataTable/Infinite Scroll

https://gist.github.com/leossantiago/cfcbd3de1e88d28c2d96df98b199c4b9

#4934 

Peek 2021-02-08 02-41

Backwards compatible.

### DIFF
--- a/src/js/components/InfiniteScroll/InfiniteScroll.js
+++ b/src/js/components/InfiniteScroll/InfiniteScroll.js
@@ -13,6 +13,9 @@ import {
 } from '../../utils';
 import { Box } from '../Box';
 
+const calculateLastPageBound = (show, step) =>
+  show ? Math.floor((show + step) / step) - 1 : 0;
+
 const InfiniteScroll = ({
   children,
   items = [],
@@ -34,7 +37,7 @@ const InfiniteScroll = ({
   // the pages we are rendering
   const [renderPageBounds, setRenderPageBounds] = useState([
     0,
-    show ? Math.floor((show + step) / step) - 1 : 0,
+    calculateLastPageBound(show, step),
   ]);
 
   // the heights of the pages, approximated after we render the first page
@@ -98,7 +101,7 @@ const InfiniteScroll = ({
 
       if (show) {
         // ensure we try to render any show page
-        const showPage = Math.floor((show + step) / step) - 1;
+        const showPage = calculateLastPageBound(show, step);
         nextBeginPage = Math.min(showPage, nextBeginPage);
         nextEndPage = Math.max(showPage, nextEndPage);
       }
@@ -151,7 +154,7 @@ const InfiniteScroll = ({
     if (lastPage === 0 && pendingLength !== 0) {
       setPageHeights([]);
       setPendingLength(0);
-      setRenderPageBounds([0, show ? Math.floor((show + step) / step) - 1 : 0]);
+      setRenderPageBounds([0, calculateLastPageBound(show, step)]);
     }
   }, [lastPage, pendingLength, show, step]);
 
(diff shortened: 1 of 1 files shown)


src/js/components/InfiniteScroll/InfiniteScroll.js
```diff
@@ -147,6 +147,14 @@ const InfiniteScroll = ({
     }
   }, [items.length, lastPage, onMore, pendingLength, renderPageBounds, step]);
 
+  useEffect(() => {
+    if (items.length === 0 && lastPage === 0 && pendingLength !== 0) {
```
ericsoderberghp, line 151: Can we just key off of `items.length` here?

src/js/components/InfiniteScroll/InfiniteScroll.js
```diff
@@ -147,6 +147,14 @@ const InfiniteScroll = ({
...
+      setPageHeights([]);
+      setPendingLength(0);
+      setRenderPageBounds([0, 0]);
```
ericsoderberghp, line 154: I think we should reset `renderPageBounds` consistent with how we initialize it on line 37.

ericsoderberghp (commented):
I think this is a reasonable step to take.

However, at some point we might want to look a little deeper and detect when the items have shrunk such that we don't have enough items for the `pageHeights` we've got. That can be a later change though.

leossantiago:
@ericsoderberghp could you check this once again?

Also, should we keep this open while we investigate `detect when the items have shrunk such that we don't have enough items for the pageHeights we've got` ?

I can keep working on that. Just let me know if we'll be keeping it here, or separating in another issue 😄

src/js/components/InfiniteScroll/InfiniteScroll.js
```diff
@@ -147,6 +147,14 @@ const InfiniteScroll = ({
     }
   }, [items.length, lastPage, onMore, pendingLength, renderPageBounds, step]);
 
+  useEffect(() => {
+    if (lastPage === 0 && pendingLength !== 0) {
+      setPageHeights([]);
+      setPendingLength(0);
+      setRenderPageBounds([0, show ? Math.floor((show + step) / step) - 1 : 0]);
```
ericsoderberghp, line 154: I think this is fine functionally. One small improvement would be to create something like:
`const calculateLastPage = (show, step) => (show ? Math.floor((show + step) / step) - 1 : 0);`
outside of the component in this file and then call it in the three places we have that expression.

leossantiago: Done!
Went with `calculateLastPageBound` as we already have a `lastPage` variable that isn't directly related to this function.

ericsoderberghp (approved):

PR #4941, opened 2021-02-09
AllComponents Story Loaded First

jcfilben:
#### What does this PR do?
In the new storybook format when storybook is opened, the AllComponents story is displayed. It is also listed first before every thing else. I moved the 'All' story outside of the 'Others' category so that the entire 'Others' category was not moved to the top.

storybook

Screen Shot 2021-02-08 at 5 15 52 PM

No

No

Backwards compatible

### DIFF
--- a/storybook/preview.js
+++ b/storybook/preview.js
@@ -3,6 +3,7 @@ export const parameters = {
   options: {
     storySort: {
       method: 'alphabetical',
+      order: ['All'],
     },
   },
 };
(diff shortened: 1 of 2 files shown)


storybook/preview.js
```diff
@@ -3,6 +3,7 @@ export const parameters = {
   options: {
     storySort: {
       method: 'alphabetical',
+      order: ['All'],
```
ShimiSun, line 6: I'm so glad they added this property!!

ShimiSun (approved):

ShimiSun:
Looking good, I don't think we need the "Others" package anymore, we can leave the one last item of "Theme" to utils and remove this package. we can do it on a separate PR though.

PR #4949, opened 2021-02-10
Select Stories Improvements

### LINKED ISSUE #4868: Select stories improvements
A few thoughts from just looking at Storybook:
1. How about combining "Themed" and "Custom search"?
2. How about splitting disabled out of "Simple"?
3. How about aligning story names more with property names, such as "Children render" -> "Children"?
4. Can we remove "Object options with search and multiple" since it isn't focused on one property?
5. Can we simplify "Clear"? It seems a bit too complex.
6. Can we remove the custom theming from "Create option", since theming is a separate aspect?
7. Can we improve the styling of "Value label"? I don't think we'd ever recommend having it look like a Button.

_Originally posted by @ericsoderberghp in https://github.com/grommet/grommet/issues/4861#issuecomment-754947498_

jcfilben:
#### What does this PR do?
Applied the following changes to the Select stories
1. Combined "Themed" into "Custom search" and removed "Themed"
2. Moved "Disabled" out of "Simple"
3. Renamed "Children render" to "Children"
4. Removed "Object options with search and multiple"
5. Simplified "Clear" to only have one Select
6. Removed custom theming from "Create option"
7. Changed "Value label" styling so it doesn't look like a button

storybook

storybook

Closes #4868 

No

No

Backwards Compatible


ShimiSun (approved):

ShimiSun:
We can go ahead and merge this PR. 
@ericsoderberghp, I've put you as the reviewer since you requested the above changes and I thought you may want to confirm your thoughts.

ericsoderberghp:
Thanks for the cleanup! Lot's of great improvements already.

A few more suggestions:

Screen Shot 2021-02-11 at 2 17 41 PM
doesn't look so great as an example

Screen Shot 2021-02-11 at 2 18 45 PM
This also doesn't seem like a good recommendation

I don't know that we still need the "Dark" story. I think that is covered well enough by the "All Components" story where you can toggle dark mode.

"Search" story multiple case closes the drop when selecting an option. "Multiple" story leaves the drop open when selecting an option. It seems like both should be leaving the drop open, like "Multiple" does it.

Screen Shot 2021-02-11 at 2 22 24 PM
wondering if we can make this look like something we might recommend. It's a bit blah as is.

jcfilben:
Update: 

- Changed Children story to have better contrast. Let me know if there are other changes I should make here

Screen Shot 2021-02-24 at 4 00 58 PM

- Changed Custom Search story to say 'multiple' when multiple options are selected instead of listing them all

Screen Shot 2021-02-24 at 4 02 15 PM

- Removed Dark story

- Changed Search story multiple case to leave the drop open after selecting an option.

- Changed Value Label story. Let me know if I should change styling to something else

Screen Shot 2021-02-24 at 4 04 35 PM

ericsoderberghp (approved):

PR #4954, opened 2021-02-11
Fix Layer responsive positioning

taysea:
#### What does this PR do?
This PR fixes the way that Layer positions once it hits its responsive breakpoint. In all variations of a responsive layer (modal, non-modal, target, no target), once the layer hits its breakpoint it should take up the full viewport.

The main adjustment here is that the LayerContainer always needs to be positioned inside of StyledLayer, where Styled Layer has a position of `fixed` to ensure it is positioned with respect to the viewport.

This PR fixes existing Layer positioning issues for the following scenarios:
1. Non-Modal Layer + No Target: once breakpoint is hit, Layer is placed below main content
2. Modal + Target: Layer positions with the target but is styled to match viewport dimensions (leading it to cut off on the sides and not allow proper scrolling)
3. Non-Modal + Target: Layer positions with the target but is styled to match viewport dimensions (leading it to cut off on the sides and not allow proper scrolling)
4. Grommet tag uses `minHeight: 100%` instead of `full`, Layer at the responsive breakpoint would not properly cover the viewport, instead is would be positioned relative to the body which may have scrolled -- causing the Layer to be out of view and the page could not be interacted with. See https://github.com/grommet/hpe-design-system/issues/1326

Tested locally in storybook Layer "Target" example using all combinations of target + modal. Checked other Layer stories to confirm no other functionality was affected.

Using the Layer Target example locally and with the above combinations.
OR
Explore the functionality on the current released version 2.16.3 (https://codesandbox.io/s/layer-positioning-4fjb4?file=/index.js), keep in mind some functionality regarding Layer styling is only on stable atm.

Layer was not positioning properly once a responsive breakpoint was hit.

Closes https://github.com/grommet/hpe-design-system/issues/1326

No.

Fixed Layer responsive positioning.

Backwards compatible.


halocline:
I guess I have a few behavioral questions:
- When target + hitResponsiveBreakpoint + position, what is the expected positioning of the layer content? In this implementation, the content is always placed in the upper, left-hand corner:
Screen Shot 2021-02-16 at 11 10 21 AM
Compared to the current implementation the content is horizontally centered.
- When target + hitResponsiveBreakpoint + layer content height > window height, what is the expected overflow behavior? In this implementation, I am unable to access the entirety of the layer's content:
Screen Shot 2021-02-16 at 11 25 12 AM
- Not directly related to this PR, but may want to re-think how this story is written --> when hitReponsiveBreakpoint, the "Toggle gutter size" no longer affects the gutter of the content which leads me to think something is broken. Seems more related to how the story is implemented / what it is trying to demonstrate, but might be worth a little refinement.

taysea:
> I guess I have a few behavioral questions:
> 
> * When target + hitResponsiveBreakpoint + position, what is the expected positioning of the layer content? In this implementation, the content is always placed in the upper, left-hand corner:
>   Screen Shot 2021-02-16 at 11 10 21 AM
>   Compared to the current implementation the content is horizontally centered.

Good comment, can you show me where you see it horizontal centered?

> * When target + hitResponsiveBreakpoint + layer content height > window height, what is the expected overflow behavior? In this implementation, I am unable to access the entirety of the layer's content:
>   Screen Shot 2021-02-16 at 11 25 12 AM

Good question -- currently Layer doesn't handle overflow for you. The user would overcome this by adding an `overflow="auto"` to their internal Box. I had played around with adding an `overflow` to StyledLayer, but ran into issues where [this story](https://storybook.grommet.io/?path=/story/layout-layer-fixed-header-scroll-body--scroll-body-layer) would no longer scroll once the responsive breakpoint was hit. Thoughts? 

> * Not directly related to this PR, but may want to re-think how this story is written --> when hitReponsiveBreakpoint, the "Toggle gutter size" no longer affects the gutter of the content which leads me to think something is broken. Seems more related to how the story is implemented / what it is trying to demonstrate, but might be worth a little refinement.

Nice, would be good to clean that up. I'll think around how to update to remove any confusion there.

halocline:
> Good question -- currently Layer doesn't handle overflow for you. The user would overcome this by adding an `overflow="auto"` to their internal Box. I had played around with adding an `overflow` to StyledLayer, but ran into issues where [this story](https://storybook.grommet.io/?path=/story/layout-layer-fixed-header-scroll-body--scroll-body-layer) would no longer scroll once the responsive breakpoint was hit. Thoughts?
> 

Works for me. Good example.

halocline:
> > I guess I have a few behavioral questions:
> > 
> > * When target + hitResponsiveBreakpoint + position, what is the expected positioning of the layer content? In this implementation, the content is always placed in the upper, left-hand corner:
> >   Screen Shot 2021-02-16 at 11 10 21 AM
> >   Compared to the current implementation the content is horizontally centered.
> 
> Good comment, can you show me where you see it horizontal centered?
> 

I think it was the Grid being used in the Target story where I was seeing this behavior. 

I guess when hitting the responsiveBreakpoint, the jumping up to the corner really stood out to me (Corner, Form, and Target stories especially) which left me wondering what implementers would expect. I suppose they'd have the ability to adjust the contents at that breakpoint as well.

taysea:
> > > I guess I have a few behavioral questions:
> > > 
> > > * When target + hitResponsiveBreakpoint + position, what is the expected positioning of the layer content? In this implementation, the content is always placed in the upper, left-hand corner:
> > >   Screen Shot 2021-02-16 at 11 10 21 AM
> > >   Compared to the current implementation the content is horizontally centered.
> > 
> > 
> > Good comment, can you show me where you see it horizontal centered?
> 
> I think it was the Grid being used in the Target story where I was seeing this behavior.
> 
> I guess when hitting the responsiveBreakpoint, the jumping up to the corner really stood out to me (Corner, Form, and Target stories especially) which left me wondering what implementers would expect. I suppose they'd have the ability to adjust the contents at that breakpoint as well.

Got it, I think what you were seeing was the Layer aligning to the target box (which happens to make the layer content look like it's centered). You're correct though that a user could leverage "ResposniveContext" to align their content to the center once their responsive breakpoint is hit.

ericsoderberghp:
Target behavior doesn't seem right.

From this screen:

Screen Shot 2021-02-19 at 7 55 33 AM

I would expect the Layer to still stay within the target bounds. But, it takes over the whole page instead.

Screen Shot 2021-02-19 at 7 56 19 AM

taysea:
> Target behavior doesn't seem right.
> 
> From this screen:
> 
> Screen Shot 2021-02-19 at 7 55 33 AM
> 
> I would expect the Layer to still stay within the target bounds. But, it takes over the whole page instead.
> 
> Screen Shot 2021-02-19 at 7 56 19 AM

Good comment. Matt and I had discussed if what a Layer with Target should do when responding on mobile, and looking at the grommet docs (https://v2.grommet.io/layer#responsive) it says "Whether the layer should take full width and height on mobile" which made us inclined to have it fill the screen on mobile. But also understand the perspective of "When there's a target, do everything with respect to the target bounds". Happy to adjust the behavior if that's the direction we want to go and will adjust the "responsive" doc definition to reflect that as well.

ericsoderberghp:
> Good comment. Matt and I had discussed what a Layer with Target should do when responding on mobile, and looking at the grommet docs (https://v2.grommet.io/layer#responsive) it says "Whether the layer should take full width and height on mobile" which made us inclined to have it fill the screen on mobile. But also understand the perspective of "When there's a target, do everything with respect to the target bounds". Happy to adjust the behavior if that's the direction we want to go and will adjust the "responsive" doc definition to reflect that as well.

Hmm.... on this page: https://design-system.hpe.design/components/layer, I was expecting the Layer to be within the target. I think when I am trying to show an app inside a site, I want the target to be honored. Perhaps another way to put it, if I am supplying a target, I want the Layer inside that target, otherwise, why am I providing a target. I'm more inclined to change the documentation to something like: "Whether the layer should take full width and height of on mobile. If a target is provided, the Layer will take the full width and height within the target."

taysea:
> > Good comment. Matt and I had discussed what a Layer with Target should do when responding on mobile, and looking at the grommet docs (https://v2.grommet.io/layer#responsive) it says "Whether the layer should take full width and height on mobile" which made us inclined to have it fill the screen on mobile. But also understand the perspective of "When there's a target, do everything with respect to the target bounds". Happy to adjust the behavior if that's the direction we want to go and will adjust the "responsive" doc definition to reflect that as well.
> 
> Hmm.... on this page: https://design-system.hpe.design/components/layer, I was expecting the Layer to be within the target. I think when I am trying to show an app inside a site, I want the target to be honored. Perhaps another way to put it, if I am supplying a target, I want the Layer inside that target, otherwise, why am I providing a target. I'm more inclined to change the documentation to something like: "Whether the layer should take full width and height of on mobile. If a target is provided, the Layer will take the full width and height within the target."

Makes sense, I am comfortable with that approach. I'll make the adjustments.

ericsoderberghp (approved):

src/js/components/Layer/README.md
```diff
@@ -214,7 +214,9 @@ top-right
 
 **responsive**
 
-Whether the layer should take full width and height on mobile Defaults to `true`.
+Whether the layer should take full width and height on mobile. If a 
+        target is provided, the Layer will take the full width and height of 
+        the target. Defaults to `true`.
```
halocline, line 219: 👍

halocline (approved):
LGTM

Issue #4843, opened 2020-12-21
ResponsiveContext doesn't match CSS media queries on mobile devices


ArrayKnight:
When loading a page on a mobile device or device emulation, and that content lays out wider than the initial (default) breakpoint, the JS breakpoint and CSS breakpoint stop matching each other

**But what causes this?**

Because Grommet uses medium as it's default breakpoint (_I have read this in comments but haven't found it documented_), on initial load its possible to have a component layout horizontally when it would be expected to layout vertically on a mobile device (say a dynamic set number of columns or a list of things, etc)

When this horizontal alignment ends up being wider than the medium breakpoint, Grommet doesn't correctly identify the actual breakpoint since the width of window is being forced open by the content

### Expected Behavior

CSS and JS breakpoints should always match

### Actual Behavior

When on a mobile device, when content is wider than the body and the next breakpoint, `window.innerWidth` is greater than the `document.body.clientWidth`, the JS breakpoint no longer matches the CSS breakpoint

### URL, screen shot, or Codepen exhibiting the issue

![image](https://user-images.githubusercontent.com/5431637/102832113-58e38200-43a2-11eb-8806-8a6b185a14f3.png)

https://codesandbox.io/s/grommet-v2-template-forked-btyzj

### Steps to Reproduce

1. Go to: https://btyzj.csb.app/ (the output of the above sandbox)
2. Open developers dev tools
3. Click device toolbar toggle to enable device mode
4. Notice that because content has forced a difference between window and html/body, the JS breakpoint no longer matches the CSS breakpoint

### Possible solution

Use `document.body.clientWidth` in place of `window.innerWidth` for `ResponsiveContext` `getBreakpoint` calculation

Switch to using a `window.matchMedia` approach since this doesn't care about `window.innerWidth`:
![image](https://user-images.githubusercontent.com/5431637/102834182-29d00f00-43a8-11eb-8b6d-a7037d209dd5.png)

### Hacky work-around

Apply `overflow-x: hidden` to `html, body` which forces `window.innerWidth` to match `document.body.clientWidth`

### Your Environment

- Grommet version: ^2.16.2
- Browser Name and version: Any
- Operating System and version (desktop or mobile): Mobile (Android, iOS, any device in DevTools as well)

### References

https://github.com/grommet/grommet/blob/adb07f11eea988777fbe0df193f5c2dd1bc8dc5f/src/js/components/Grommet/Grommet.js#L97
[...]

PR #4959, opened 2021-02-14
Fix ResponsiveContext device breakpoint

jeferson-sb:
#### What does this PR do?
Fix ResponsiveContent breakpoint to match css media queries when content is wider than the viewport screen.
Unlike `window.innerWidth`, `document.body.clientWidth` returns the viewport size excluding any scrollbars and borders

`src/js/utils/responsive.js`

`yarn test` and `yarn storybook`

`yarn storybook`

#4843 

no

backwards compatible


src/js/utils/responsive.js
```diff
@@ -1,29 +1,24 @@
...
+export const getBreakpoint = (viewportWidth, theme) => {
   const sortedBreakpoints = Object.keys(theme.global.breakpoints).sort(
     (a, b) => {
       const first = theme.global.breakpoints[a];
       const second = theme.global.breakpoints[b];
-      if (!first) return 1;
-      if (!second) return -1;
-      if (!first.value) return 1;
-      if (!second.value) return -1;
+      if (!first || !first.value) return 1;
```
ericsoderberghp, line 6: I think this might sort differently. In the old code, `!second && !first.value` will return -1 but will return 1 here.

ericsoderberghp (commented):
just a small comment, good otherwise

src/js/utils/responsive.js
jeferson-sb: gotcha

ericsoderberghp (approved):

PR #4961, opened 2021-02-15
Fix Drop content text color behavior when plain

### LINKED ISSUE #4950: Tip component ignoring global theme text color
Text in the Tip component does not use the global theme text color, and simply uses black.

### Expected Behavior
Text in the Tip component should be the color specified in the theme.

### Actual Behavior
See above. It does not matter if you wrap the content in a Text tag.

https://codesandbox.io/s/grommet-v2-template-forked-402ly

### Your Environment
- Grommet version: 2.16.3
- Browser Name and version: Firefox 85, Chrome 87
- Operating System and version (desktop or mobile): Desktop Ubuntu

leossantiago:
#### What does this PR do?

Fixes #4950 , where components such <Tip /> wouldn't follow `theme.global.colors.text`.

Even though being related as a <Tip /> issue, after some debugging, this happens because of <Drop /> component.

Tip forces a `plain` state on Drop, which removes some styling related to background and elevation, but it also includes text-color styling. 

This PR fixes the issue by providing text color styling when `plain` is set to true on <Drop /> component.

Note that manual content styling such as `<Text color="red">Text in red</Text>` worked just fine prior to this.

`./src/js/components/Drop/StyledDrop.js`

Manual testing, and update snapshots that already cover Drop `plain` usage.

```js
const theme = {
  global: {
    colors: {
      // text: 'blue',
      text: {
        light: 'blue',
        dark: 'blue',
      },
    },
  },
};

const override = deepMerge(grommet, theme);

export const Simple = () => (
  <Grommet theme={override}>
    <Text>Grommet &gt; Text</Text>
    <Box pad="small" gap="xlarge">
      <Text>Global text color theme is set to blue</Text>

      <Tip content="Tip with regular text">
        <Button label="Hover" />
      </Tip>

      <Button label="Hover" tip="Tip as Button prop" />

      <Tip content={<Text>Tip using Text component as content</Text>}>
        <Button label="Hover" />
      </Tip>

      <Button
        label="Hover"
        tip={{
          content: (
            <Text>Tip as button prop using Text component as content</Text>
          ),
        }}
      />
    </Box>
  </Grommet>
);
```

#4950 

Backwards compatible considering general behavior. Breaking change if you consider global styling usage.

### DIFF
--- a/src/js/components/Drop/StyledDrop.js
+++ b/src/js/components/Drop/StyledDrop.js
@@ -1,7 +1,10 @@
 import styled, { keyframes } from 'styled-components';
 
 import { baseStyle } from '../../utils';
-import { backgroundStyle } from '../../utils/background';
+import {
+  backgroundStyle,
+  backgroundAndTextColors,
+} from '../../utils/background';
 import { defaultProps } from '../../default-props';
 
 function getTransformOriginStyle(align) {
@@ -36,8 +39,18 @@ const StyledDrop = styled.div`
   outline: none;
 
   ${props =>
-    !props.plain &&
-    backgroundStyle(props.theme.global.drop.background, props.theme)}
+    !props.plain
+      ? backgroundStyle(props.theme.global.drop.background, props.theme)
+      : () => {
+          const [, textColor] = backgroundAndTextColors(
+            props.theme.global.drop.background,
+            undefined,
+            props.theme,
+          );
+          return `${textColor ? `color: ${textColor};` : ''}`;
+        }}
+
+
 
   opacity: 0;
   transform-origin: ${props => getTransformOriginStyle(props.alignProp)};
--- a/src/js/components/Drop/__tests__/__snapshots__/Drop-test.js.snap
+++ b/src/js/components/Drop/__tests__/__snapshots__/Drop-test.js.snap
@@ -1329,6 +1329,7 @@ exports[`Drop plain renders 1`] = `
   position: fixed;
   z-index: 20;
   outline: none;
+  color: #444444;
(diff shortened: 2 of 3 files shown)


ericsoderberghp:
Hmm... something doesn't feel quite right here in setting the text color even though `plain` has been passed in to Drop. `plain` typically indicates that the caller will take care of any styling. I think we should instead revisit how Tip is behaving and ensure it is setting up the styling. I'm inclined to say that if Tip is called with `plain`, that works as it does today. If Tip is not called with `plain`, then the Box it wraps the content in should include background and text color styling. Although, since that's all that StyledDrop has removed due to `plain`, I'm wondering why we don't just call Drop with `plain={plain}`?

ShimiSun:
There are two concepts of `plain` within Tip:
1. The `plain` of the drop that is currently always set to true since the Tip has its own content Box with its own elevation and background, if the drop wouldn't be `plain` we will see a double background and weird styling. The caller could always override this value via `dropProps`, but in 80% of the cases, it won't look good.
2. The `plain` of the Tip content, which basically means whether you'd like to use Tip's content styling or not, the user can decide this `plain` value via prop (but it won't affect the plain set on the Drop). Unfortunately, this prop doesn't have control over the text color, since it was already inherited from the Drop.

I'm wondering if the approach should be to allow text color as part of the Box's content. Feel free to share your thoughts.

ShimiSun:
Thank you for this PR 💟  we took the approach of removing `plain` from the Tip's Drop and that did the trick. Check out related PRs #5046, #5047

PR #4963, opened 2021-02-15
Ensure TextInput respects defaultSuggestion prop when onChange is triggered

ddohler:
#### What does this PR do?

Ensures that TextInput respects the `defaultSuggestion` prop when the input's `onChange` is triggered. The `onChange` method was setting `activeSelectionIndex` to -1 when it should have been using `resetSelectionIndex`. This caused the `defaultSuggestion` prop not to be respected if the value of the text field changed.

The easiest thing to do is probably to revert the one-line change to `TextInput.js` and then run tests, and confirm you get a failure.

This was discovered when attempting to integrate the `defaultSuggestion` prop into the app discussed in #4781 . I diagnosed the issue, confirmed that there was a gap in the test coverage, and wrote a test to cover the gap. Then I made the change to resolve the issue.

If you open the existing `Default suggestion` storybook, you can verify that with this change, typing in the text input does not affect the highlighted default suggestion. However, if you revert the change in `TextInput.js`, you will see that entering input into the text input will cause the highlight over the default suggestion to be cleared.

The user-requested goal here is to be able to auto-highlight the first suggestion in the list at all times, so that the user can type a partial match and then hit Enter to select the first suggestion without having to use the mouse or arrow keys to explicitly select it.

I have not opened an issue for this because it was a simple fix and it was quicker to just open the PR directly, but I am happy to open one if that would be helpful.

N/A

No.

Yes, this is a bugfix.

Backwards compatible.


ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution!

PR #4968, opened 2021-02-17
Fix1396 add elevation prop

britt6612:
I have 2 suggestions for the current bug in the design-system
1. The first suggestion is taking elevation out of the theme and making it a prop for grommet, that can be used in the design system. This would prevent the bug that is happening now with the elevation. See below. If this is a prop then we can prevent this from happening.  
Screen Shot 2021-02-17 at 9 54 29 AM

2. The other  approach would be to change the design. We can use `flex` within the`box` in the `layer` to take up the space. The `border-radius` is different from the outside box.
Screen Shot 2021-02-17 at 9 55 34 AM

This would have to be changed in figma to make the `border-radius` match. 
Screen Shot 2021-02-17 at 10 02 02 AM

#### What does this PR do?
This PR takes the first suggestion by adding an elevation prop
layer.js
StoryBook 

```
       <Layer
          position="bottom"
          modal={false}
          margin={{ vertical: 'medium', horizontal: 'small' }}
          onEsc={onClose}
          responsive={false}
          plain
          elevation="large"
        >
          <Box
            align="center"
            direction="row"
            gap="small"
            justify="between"
            round="medium"
            elevation="medium"
            pad={{ vertical: 'xsmall', horizontal: 'small' }}
            background="status-ok"
          >
            <Box align="center" direction="row" gap="xsmall">
              <StatusGood />
              <Text>
                A new virtual machine has been successfully added (this Layer
                will close after 3 seconds)
              </Text>
            </Box>
            <Button icon={<FormClose />} onClick={onClose} plain />
          </Box>
        </Layer>

```
issue
https://github.com/grommet/hpe-design-system/issues/1396
already  updated
no
yes

### DIFF
--- a/src/js/components/Layer/LayerContainer.js
+++ b/src/js/components/Layer/LayerContainer.js
@@ -30,6 +30,7 @@ const defaultPortalContext = [];
 const LayerContainer = forwardRef(
   (
     {
+      elevation,
       background,
       children,
       full = false,
@@ -169,7 +170,8 @@ const LayerContainer = forwardRef(
       <StyledContainer
         ref={ref || containerRef}
         background={background}
-        elevation={theme.layer.container.elevation}
+        // elevation munged to avoid styled-components putting it in the DOM
+        elevationProp={elevation}
         id={id}
         full={full}
         margin={margin}
--- a/src/js/components/Layer/README.md
+++ b/src/js/components/Layer/README.md
@@ -73,6 +73,21 @@ string
 }
 ```
 
+**elevation**
+
+Elevated height above the underlying context, indicated
+        via a drop shadow. Defaults to `none`.
+
+```
+none
+xsmall
+small
+medium
+large
+xlarge
+string
+```
+
 **full**
 
 Whether the width and/or height should fill the current viewport
--- a/src/js/components/Layer/StyledLayer.js
+++ b/src/js/components/Layer/StyledLayer.js
@@ -766,7 +766,7 @@ const responsiveContainerStyle = css`
 const elevationStyle = css`
   box-shadow: ${props =>
     props.theme.global.elevation[props.theme.dark ? 'dark' : 'light'][
-      props.theme.layer.container.elevation
+      props.elevationProp
     ]};
 `;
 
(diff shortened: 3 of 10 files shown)


ShimiSun:
It will be helpful to know which bug you are referring
> I have 2 suggestions for the current bug in the design-system

The issue that is linked from this PR isn't the DS system issue, and I'm not sure it is relevant to this PR since it's grommet v1.

britt6612:
> It will be helpful to know which bug you are referring
> 
> > I have 2 suggestions for the current bug in the design-system
> 
> The issue that is linked from this PR isn't the DS system issue, and I'm not sure it is relevant to this PR since it's grommet v1.

Sorry about that I added the link to the issue in design-system

ShimiSun:
Per eric's recommendation, 

>I think we need to start by ensuring the notification example fits with the theme, just adding a flex to the Box and removing plain from the Layer might be enough

britt6612:
I did try using `flex` and taking out `plain`  in the example. This will fill the majority of the `elevation` in the `box`  however we have the corners that do not have the same `border-radius` which causes it to not fill completely. 

```
<Layer
          position="top"
          modal={false}
          margin={{ vertical: 'medium', horizontal: 'small' }}
          responsive={false}
        >
          <Box
            flex
            align="center"
            direction="row"
            gap="xsmall"
            justify="between"
            round="medium"
            elevation="medium"
            pad={{ vertical: 'xxsmall', horizontal: 'small' }}
            background="status-ok"
          >
            <Box align="center" direction="row" gap="small">
              <StatusGood color="text-strong" />
              <Text color={{ light: 'text', dark: 'text-strong' }}>
                This is a notifcation with position="top"
              </Text>
            </Box>
            <Button icon={<FormClose />} onClick={onClose} plain />
          </Box>
        </Layer>
```

Screen Shot 2021-02-17 at 11 31 55 AM

ericsoderberghp:
> however we have the corners that do not have the same `border-radius` which causes it to not fill completely.

I'd remove the `round` from the Box to align with what Layer already has from the theme. I think we should start by being more consistent with the theme

britt6612:
Chatting with designer about changing the `border-radius`.

So we wouldn't have to change grommet

britt6612:
> > however we have the corners that do not have the same `border-radius` which causes it to not fill completely.
> 
> I'd remove the `round` from the Box to align with what Layer already has from the theme. I think we should start by being more consistent with the theme

I agree we should be consistent. The `layer.border.radius` only affects the layer. So if I take out the `round` from the `box` it would look like this. 

Screen Shot 2021-02-17 at 12 33 52 PM

we already have this issue in the modal example. We are not putting a `round` in the box and if you look closely you can see the bottom two corner edges.

https://design-system.hpe.design/components/layer
You can see it more on the dark version

britt6612:
going to close issue have a PR fix on design system
https://github.com/grommet/hpe-design-system/pull/1430

PR #4973, opened 2021-02-18
Button padding should use size "medium" by default

taysea:
#### What does this PR do?
This PR ensures that StyledButtonKind uses size "medium" by default. Currently, it wasn't falling back to any size and this was causing discrepancies between buttons implemented as `<Button />` vs `<Button size="medium" />` but they should be identical.

Tested local in storybook, created custom theme with size definitions. Default (no size prop specified) and medium should be the same always.

Before (notice padding difference between default and medium):
Screen Shot 2021-02-17 at 3 52 14 PM

After:
Screen Shot 2021-02-17 at 3 52 23 PM

Noticed the HPE button dimensions had shifted after we removed padding values in this PR but when I looked into it further I also noticed that buttons implemented as `<Button label="Label" />` were not receiving styling for "medium" size like they should.

Noticed issue after https://github.com/grommet/grommet-theme-hpe/pull/163 merged in HPE theme and a subsequent PR on design system site noted snapshot changes.

See above.

No.

Backwards compatible.

### DIFF
--- a/src/js/components/Button/StyledButtonKind.js
+++ b/src/js/components/Button/StyledButtonKind.js
@@ -32,7 +32,7 @@ const fontStyle = props => {
   `;
 };
 
-const padFromTheme = (size, theme) => {
+const padFromTheme = (size = 'medium', theme) => {
   if (
     size &&
     theme.button.size &&
--- a/src/js/components/Button/StyledButtonKind.js
+++ b/src/js/components/Button/StyledButtonKind.js
@@ -32,7 +32,8 @@ const fontStyle = props => {
   `;
 };
 
-const padFromTheme = (size, theme) => {
+const padFromTheme = (sizeProp, theme) => {
+  const size = sizeProp || 'medium';
   if (
     size &&
     theme.button.size &&
--- a/src/js/components/Button/StyledButtonKind.js
+++ b/src/js/components/Button/StyledButtonKind.js
@@ -32,8 +32,7 @@ const fontStyle = props => {
   `;
 };
 
-const padFromTheme = (sizeProp, theme) => {
-  const size = sizeProp || 'medium';
+const padFromTheme = (size, theme) => {
   if (
     size &&
     theme.button.size &&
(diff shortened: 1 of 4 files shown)


src/js/components/Button/__tests__/Button-kind-test.js
```diff
@@ -381,4 +381,53 @@ describe('Button kind', () => {
     fireEvent.mouseOver(getByText('Button'));
     expect(container.firstChild).toMatchSnapshot();
   });
+
+  test(`should apply size styling correctly`, () => {
```
halocline, line 385: What is "correctly"?

```suggestion
  test(`should apply styling according to theme size definition`, () => {
```

halocline (commented):
One wording suggestion for test name, the rest looks good.

halocline (approved):

halocline:
@taysea looks like the test name update needs the associated snapshot updated.

src/js/components/Button/StyledButtonKind.js
```diff
@@ -32,7 +32,8 @@ const fontStyle = props => {
   `;
 };
 
-const padFromTheme = (size, theme) => {
+const padFromTheme = (sizeProp, theme) => {
```
ericsoderberghp, line 35: FYI, I think you could have just done: `(size = 'medium', theme)`

MikeKingdom (approved):
Looks good and I agree with Eric, you can probably use the arg default mechanism.

Sorry I missed that. I thought I remembered it doing that up higher in Button itself.

src/js/components/Button/StyledButtonKind.js
taysea: Good reminder, updating now.

taysea:
@ericsoderberghp @MikeKingdom @halocline I've adjusted the PR to define "medium" as the fallback size directly in Button.js. I noticed that in `StyledButton` for themes without any `theme.button.default` the same discrepancies were also showing (see screenshots). I think setting size higher up is a smoother solution:

For this I defined some padding/border radius for the various sizes, but notice how it's not carried through to the Default Button (`<Button />`)

Screen Shot 2021-02-18 at 8 38 18 AM

MikeKingdom (commented):
Coolness. Makes sense to me

taysea:
> Coolness. Makes sense to me

Hmm need to look at some chromatic changes actually.

ericsoderberghp (approved):

Issue #4952, opened 2021-02-11
Child component for DataTable component


anoopadvaitha:
I was looking at having a sub component for the Datatable rows. At HPE we use tables a lot and this is one of the requirement for us. It is something available in majority of Table components available outside Grommet.

### Expected Behavior
Option to have a childComponent to a Datatable row.

### Actual Behavior
No way to provide a chidComponent.

### URL, screen shot, or Codepen exhibiting the issue
![image](https://user-images.githubusercontent.com/5357867/107683204-4e20db80-6cc7-11eb-8902-124eb5fb9daa.png)

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.16.3
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): Windows 10

anoopadvaitha:
I have managed to implement something in these lines. Is it open for a PR ? 
Will share the storybook link soon

anoopadvaitha:
Please find the link for the storybook
Let me know if this is acceptable for you.
https://www.chromatic.com/component?appId=602aa3d849721200238dcf29&name=Visualizations%2FDataTable%2FchildComponent&buildNumber=2

ericsoderberghp:
I would use this story as the basic design direction: https://storybook.grommet.io/?path=/story/visualizations-datatable-grouped--grouped-data-table
I think we can restrict that you cannot group and have row detail rendering in the same DataTable.
I am open to a pull request for this functionality.

anoopadvaitha:
@ericsoderberghp  Thanks for the input will try to address your comments and open a PR.

anoopadvaitha:
@ericsoderberghp Regarding the restriction of not allowing both group and childComponent. I'm thinking of going with an approach of ignoring the childComponent if groupBy prop is provided rather than throwing the error. Please let me know your thoughts will change accordingly

PR #4975, opened 2021-02-18
Added 'rowDetails' to enable expansion of DataTable Rows 

anoopadvaitha:
#### What does this PR do?
Changed Datatable to include ChildComponent prop which will take a ReactNode.

src/js/components/Datatable/Datatable.js

Tested local in storybook. 

We were using Datatable for majority of the pages and having a childComponent for it was a critical requirement for us.

#4952 

https://www.chromatic.com/component?appId=602aa3d849721200238dcf29&name=Visualizations%2FDataTable%2FchildComponent&buildNumber=4

Yes.

Not sure.

It should be backward compatible.


src/js/components/DataTable/DataTable.js
```diff
@@ -67,6 +67,7 @@ const DataTable = ({
   size,
   sort: sortProp,
   sortable,
+  childComponent,
```
ericsoderberghp, line 70: I think we should just use `children`, rather than adding a separate property for this. This will be more consistent with how InfiniteScroll, Select, Distribution, etc. work.

Also, need to update `doc.js` to explain how `children` will be called.

src/js/components/DataTable/Body.js
```diff
@@ -85,67 +89,101 @@ const Body = forwardRef(
                 ? datumValue(datum, primaryProperty)
                 : undefined;
               const isSelected = selected && selected.includes(primaryValue);
+              const childExist = childExpand.includes(index);
+              const ischildExpanded = childExpand && childExist;
```
ericsoderberghp, line 93: Seems odd to test for `childExpand` here when we've already dereferenced it on the line above.

src/js/components/DataTable/Body.js
```diff
@@ -85,67 +89,101 @@ const Body = forwardRef(
...
+                      onClickRow ? () => setActive(index) : undefined
+                    }
+                    onMouseLeave={
+                      onClickRow ? () => setActive(undefined) : undefined
+                    }
+                    onFocus={onClickRow ? () => setActive(index) : undefined}
+                    onBlur={onClickRow ? () => setActive(undefined) : undefined}
+                  >
+                    {childComponent && (
+                      <ExpanderCell
```
ericsoderberghp, line 123: I'm wondering if the CheckBox cell for selecting the row should come first, since whether the row is expanded or not seems related to "within" whereas the selection state seems more about the whole thing.

src/js/components/DataTable/Body.js
```diff
@@ -85,67 +89,101 @@ const Body = forwardRef(
...
+                        rowProp={rowProps && rowProps[primaryValue]}
+                        scope={
+                          column.primary || column.property === primaryProperty
+                            ? 'row'
+                            : undefined
+                        }
+                      />
+                    ))}
+                  </StyledDataTableRow>
+                  <StyledDataTableRow key={`${index.toString()}_expand`}>
```
ericsoderberghp, line 176: We shouldn't be adding an extra table row unless we actually have a child.

ericsoderberghp (commented):
Great start.

src/js/components/DataTable/DataTable.js
anoopadvaitha: Since the expanded data should be shown for all the rows if required, I feel it'll be better to use it as a property rather than having it as a child as it may confuse users. If it's pagination or InfiniteScroll then I agree that it can be a children instead of property. Let me know your thoughts.

src/js/components/DataTable/Body.js
anoopadvaitha: Addressed this.

src/js/components/DataTable/Body.js
anoopadvaitha: Addressed this.

src/js/components/DataTable/Body.js
anoopadvaitha: Addressed this.

src/js/components/DataTable/DataTable.js
ericsoderberghp: Hmm... perhaps you have a point, in that for Distribution, Select, and InfiniteScroll it applies to all items. Whereas, here it applies to only the expanded item. In which case, I think we should relabel the property to something like `rowDetails` rather than `childComponent`.

src/js/components/DataTable/DataTable.js
```diff
@@ -122,6 +123,8 @@ const DataTable = ({
     select,
   ]);
 
+  const [childExpand, setChildExpand] = useState([]);
```
ericsoderberghp, line 126: Upon reflection, I think we call this `rowExpand`, because it's the "row" we are expanding, not the "child" we are expanding.

ericsoderberghp (commented):
steady progress

src/js/components/DataTable/DataTable.js
anoopadvaitha: Addressed the change

src/js/components/DataTable/DataTable.js
anoopadvaitha: Addressed the change

src/js/components/DataTable/Body.js
```diff
@@ -26,8 +27,11 @@ const Body = forwardRef(
       primaryProperty,
       rowProps,
       selected,
+      rowDetails,
       size,
       step,
+      rowExpand,
+      setrowExpand,
```
ericsoderberghp, line 34: setrowExpand -> setRowExpand

src/js/components/DataTable/Body.js
```diff
@@ -85,67 +89,96 @@ const Body = forwardRef(
...
-                        a11yTitle={`${
-                          isSelected ? 'unselect' : 'select'
-                        } ${primaryValue}`}
-                        checked={isSelected}
-                        disabled={!onSelect}
-                        onChange={() => {
-                          if (isSelected)
-                            onSelect(selected.filter(s => s !== primaryValue));
-                          else onSelect([...selected, primaryValue]);
+                <Fragment key={`${index.toString()}_expanded`}>
```
ericsoderberghp, line 94: I think we can move the key from line 96 to here and omit the key at line 96 below.

src/js/components/DataTable/Body.js
```diff
@@ -85,67 +89,96 @@ const Body = forwardRef(
                 ? datumValue(datum, primaryProperty)
                 : undefined;
               const isSelected = selected && selected.includes(primaryValue);
+              const isrowExpanded = rowExpand && rowExpand.includes(index);
```
ericsoderberghp, line 92: isrowExpanded -> isRowExpanded

src/js/components/DataTable/stories/childComponent.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
```
ericsoderberghp, line 1: I think the file name should be "rowDetails.js" instead of "childComponent.js"

ericsoderberghp (commented):
almost there, just a few naming comments

Also, it needs a fresh merge with master.

ericsoderberghp:
The functionality is looking good. Forgot one thing, we need a unit test. I think we can just leverage the current "groupBy" unit test as a starting point.

anoopadvaitha:
> 
> 
> The functionality is looking good. Forgot one thing, we need a unit test. I think we can just leverage the current "groupBy" unit test as a starting point.

Sure, will get test cases added in DataTable-test.js and update the PR.
Thanks !

anoopadvaitha:
@ericsoderberghp  Added two unit test cases. Please validate and let me know if any changes required

src/js/components/DataTable/__tests__/DataTable-test.js
```diff
@@ -408,6 +408,64 @@ describe('DataTable', () => {
...
+            { a: 'two', b: 2.1 },
+            { a: 'two', b: 2.2 },
+          ]}
+          rowDetails={row => <Box>{row.a}</Box>}
+          primaryKey="b"
+        />
+      </Grommet>,
+    );
+    expect(container.firstChild).toMatchSnapshot();
+    const headerCell = getByText('A');
```
ericsoderberghp, line 431: This will click in the table header cell. I think we need to click the expand/collapse control in the row.

anoopadvaitha: Addressed the change.

anoopadvaitha:
Please find the updated storybook
https://602aa3d849721200238dcf29-zujcwytlfi.chromatic.com/?path=/story/visualizations-datatable-rowdetails--controlled-data-table

ericsoderberghp (approved):
Looks OK to me. I'd like @ShimiSun to take a look as well now.

ShimiSun (approved):

ShimiSun:
Looking good!! Thank you for the contribution @anoopadvaitha! Make sure to join our Slack channels, we'd love to see more of your work https://slack-invite.grommet.io/ 💟

PR #4977, opened 2021-02-18
Changed Announce Polite story

ShimiSun:
#### What does this PR do?
Changed Announce Polite story to describe a better flow of n announcement scenario.


src/js/contexts/AnnounceContext/stories/Polite.js
```diff
@@ -1,47 +1,67 @@
...
+      <>
+        <Paragraph textAlign="center">
+          This component will demonstrate a screen reader accessibility feature
+          via turning on the
+          <Anchor
+            label=" turning on the Screen reader mode."
+            href="https://www.codecademy.com/articles/how-to-setup-screen-reader#:~:text=(OS%20X)%20VoiceOver,Command%2DF5%20turns%20it%20off."
+          />{' '}
+          An The announcement will be broadcast once clicking on the announce
+          Button, turn on your clients Screen Reader mode to listen:
```
ericsoderberghp, line 57: I think we can simplify this text block:

"Announce can only be "observed" via a screen reader. Here's how to <Anchor href="">how turn it on</Anchor>, hint: Command-F5 on OSX. Clicking the Button below will trigger an announcement."

src/js/contexts/AnnounceContext/stories/Polite.js
```diff
@@ -1,47 +1,67 @@
...
+        </Paragraph>
+      </>
+      <Announcer mode="polite" role="log" />
```
ericsoderberghp, line 60: I know we need to be a separate component to simplify access to the AnnounceContext. But, I think we should call it something else, and maybe put the Paragraph inside it, to make it clear you can embed it inside your existing content and don't need to create a separate "Announce" component to get announcements.

src/js/contexts/AnnounceContext/stories/Polite.js
```diff
@@ -1,47 +1,67 @@
...
+        }}
+      />
 
-Announcer.defaultProps = {
-  message: 'Here is a simple announcement. This will soon disappear',
-  mode: 'polite',
-  role: 'log',
+      {show && (
+        <Text role={role} aria-live={mode}>
+          {message}
```
ericsoderberghp, line 37: I don't think we should show the same message we are announcing. Screen readers will already do that via `aria-live`. I think we can remove this Text.

src/js/contexts/AnnounceContext/stories/Polite.js
```diff
@@ -1,47 +1,67 @@
...
-  role: PropTypes.string,
-};
+    <Box align="center" gap="medium">
+      <Button
+        label="Announce"
+        icon={<Announce />}
+        a11yTitle="Announce button"
+        reverse
+        onClick={() => {
+          setShow(true);
```
ericsoderberghp, line 30: I think we can simplify this to just call `announce()` on click. That keeps the code simpler and more focused on the key feature.

src/js/contexts/AnnounceContext/stories/Polite.js
```diff
@@ -1,47 +1,42 @@
...
+const PageContent = ({ mode }) => {
+  const announce = useContext(AnnounceContext);
   return (
-    <Text align="center" role={role} aria-live={mode}>
-      {message}
-    </Text>
+    <Box align="center" gap="medium">
+      <Paragraph textAlign="center">
+        Announce can only be &quot;observed&quot; via a screen reader.
+        Here&apos;s how to how turn it on, hint: Command-F5 on OSX. Clicking the
```
ericsoderberghp, line 16: Can we use an `<Anchor />` for the "how to turn it on"?

src/js/contexts/AnnounceContext/stories/Polite.js
```diff
@@ -1,47 +1,55 @@
...
-  role: 'log',
+const PageContent = ({ mode }) => {
+  const announce = useContext(AnnounceContext);
+  return (
+    <Box align="center" gap="medium">
+      <Paragraph textAlign="center">
+        Announce can only be &quot;observed&quot; via a screen reader.
+        Here&apos;s{' '}
+        <Anchor
+          label=" how turn it on"
```
ericsoderberghp, line 26:
```suggestion
          label=" how to turn it on"
```

ericsoderberghp (approved):

Issue #4739, opened 2020-11-11
aria-hidden toggled outside of application


joelachance:
I'm using the Select component inside of the Layer component. When Select is clicked, this toggles all aria-hidden labels outside of the application.

This application is a third-party app on a Shopify page, and the theme is toggling overlays based on the value of aria-hidden, causing some issues. 

### Expected Behavior

Only toggle aria-hidden labels associated with the focused components.

### Actual Behavior

All aria-hidden labels on the page are being toggled.

### URL, screen shot, or Codepen exhibiting the issue

NOTE: The bug doesn't work in CodePen. This code needs to be rendered locally to see the effect.

https://codepen.io/joelachance/pen/rNLQYZK?editors=1111

### Steps to Reproduce

Click select, and watch aria-hidden labels in parent elements toggle.

### Your Environment

- Grommet version: 2.15.1
- Browser Name and version: Firefox 81.0.2
- Operating System and version (desktop or mobile): Catalina 10.15.7

jeferson-sb:
@ShimiSun I believe this somewhat related to this function 
https://github.com/grommet/grommet/blob/6e68e5039bf467eaababd42110e382da3cbe97a5/src/js/utils/DOM.js#L118

ShimiSun:
@joelachance thanks for reporting. I'm guessing it is because you are using a modal Layer, have you tried `modal={false}` and see if it works for you?
Also, Select had gone through lots of changes on the last few releases, can you please verify that you still see the problem on grommet v2.16.2?

joelachance:
I also thought it was also due to using a Layer, but moving it outside of the Layer didn't seem to change this behavior.
I'm currently using 2.16.2 and still seeing this issue.

@ShimiSun thanks for taking a look!

taysea:
@joelachance Just to confirm I understand your scenario, the theme you're referencing is a Shopify theme that is relying on `aria-hidden` to decide the styling of certain things?

joelachance:
Corret @taysea

taysea:
Just summarizing my understanding of what is happening. The situation is centered around how Grommet toggles `aria-hidden` for elements in the DOM. Currently, `aria-hidden` values in the DOM are controlled by two functions: `makeNodeFocusable` (sets aria-hidden to false) and `makeNodeUnfocusable` (sets aria-hidden to true). These functions are called by `FocusedContainer` which is only used by Drop (and subsequently, any components built using Drop, like Select).

[...]

PR #4979, opened 2021-02-18
Only toggle aria-hidden for DOM nodes coming from grommet

taysea:
#### What does this PR do?
This PR updates how Grommet toggles `aria-hidden` values on DOM elements. Now, Grommet will only touch elements that have been added to the DOM by grommet. It will not touch other children of `<body>` that may not be related to Grommet.

This is done by creating a context called `RootsContext` that holds an array of refs of any Grommet related elements appended as a child of `body`. By default, it holds the ref of the `<Grommet>` tag, but any time a FocusedContainer is added (Drop or Modal Layer), the ref of that div is added to the array. Instead of iterating on all body children, we now iterate over this Grommet specific array.

Adjustments were also made to the useEffects in `FocusedContainer` to separate unrelated functionality.

Additionally, the `removeTrap` function was enhanced to fix a bug where all elements with`aria-hidden="true"` were toggling back to `aria-hidden="false"`.

src/js/components/Grommet/Grommet.js

Tested locally in storybook and jest.

I've confirmed that in the case that you have nested Grommets, the inner Grommet will not toggle `aria-hidden` values for an outer Grommet.

Interact with any Drop story and open dev tools. `aria-hidden` should not be added to body children not related to Grommet.

Closes https://github.com/grommet/grommet/issues/4739
 
No.

No.

Backwards compatible.


ericsoderberghp:
> One edge case to account for is if the user does not wrap their application in a `<Grommet>` (as was the case in some jest tests). In this case, we add a `data-g-id="grommet"` to the FocusedContainer which `getBodyChildElements` will look for. Since the GrommetContext is not available, we need some way to mark the element directly in the DOM.

I don't think we should be handling this case. I think a `<Grommet />` should be required for us to manage `aria-hidden` elsewhere. We will need this behavior to handle multiple versions of grommet in the same browser DOM.

taysea:
> > One edge case to account for is if the user does not wrap their application in a `<Grommet>` (as was the case in some jest tests). In this case, we add a `data-g-id="grommet"` to the FocusedContainer which `getBodyChildElements` will look for. Since the GrommetContext is not available, we need some way to mark the element directly in the DOM.
> 
> I don't think we should be handling this case. I think a `<Grommet />` should be required for us to manage `aria-hidden` elsewhere. We will need this behavior to handle multiple versions of grommet in the same browser DOM.

Okay that's fair. I will remove this functionality then.

src/js/components/Form/__tests__/__snapshots__/Form-test-uncontrolled.js.snap
```diff
@@ -1941,6 +1941,7 @@ exports[`Form uncontrolled reset clears select, checkbox, radiobuttongroup 1`] =
 }
 
 <div
+  aria-hidden="false"
```
ericsoderberghp, line 1944: Why are we adding this? It seems like it isn't needed.

src/js/components/Grommet/Grommet.js
```diff
@@ -93,15 +95,23 @@ const Grommet = forwardRef((props, ref) => {
     deviceResponsive(userAgent, theme) ||
     theme.global.deviceBreakpoints.tablet;
 
+  const grommetRef = useForwardedRef(ref);
+  // When toggling aria-hidden values, we only want to affect elements
+  // in the DOM that come from Grommet, so we track those elemnts in this
+  // context value. See FocusedContainer.js
+  const bodyChildrenFromGrommet = [grommetRef.current];
```
ericsoderberghp, line 102: What if I have nested `<Grommet />` elements?

src/js/components/Grommet/GrommetContext.js
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+// Track which components mounted onto the DOM are coming from
+// Grommet (see FocusedContainer.js)
+export const GrommetContext = React.createContext(undefined);
```
ericsoderberghp, line 5: Wondering if "GrommetContext" is the right term here, perhaps "RootsContext"?

taysea: Good comment, I was trying to think what would be appropriate an appropriate name. I could see `Roots` working but it didn't click right off the bat. Let me brainstorm a little more while I address the other comments.

src/js/components/Form/__tests__/__snapshots__/Form-test-uncontrolled.js.snap
taysea: Good call, let me look into it.

src/js/components/Grommet/Grommet.js
taysea: Hmm that's a good point. My understanding is that Context is contained, so I'm not sure if you can have an outer/inner context inherit the value of the other -- do you know?

Otherwise, we may need to rely more explicitly on a DOM attribute like "data-g-id" or something that we could scan the DOM for. What do you think?

ericsoderberghp: I think we have to avoid scanning the DOM. I was thinking that we just keep the additive nature, so a nested Grommet just adds another root to the array started by the outermost one.

src/js/components/Form/__tests__/__snapshots__/Form-test-uncontrolled.js.snap
taysea: Resolved.

src/js/components/Grommet/GrommetContext.js
taysea: I've gone with RootsContext. I think once I equated it to the concept of "root directory", it makes more sense. The other idea I had was `OriginContext`, but that didn't have any programming equivalent so it doesn't feel like it's a better option.

src/js/components/FocusedContainer.js
```diff
@@ -18,44 +13,51 @@ export const FocusedContainer = ({
...
-      const child = ref.current;
-      getBodyChildElements()
-        .filter(isNotAncestorOf(child))
-        .forEach(makeNodeUnfocusable);
+  useEffect(() => {
+    if (
+      bodyOverflowStyle !== 'hidden' &&
+      !hidden &&
+      restrictScroll &&
+      trapFocus
```
ericsoderberghp, line 29: do we need `trapFocus` here?

src/js/components/FocusedContainer.js
```diff
@@ -18,44 +13,51 @@ export const FocusedContainer = ({
...
+    ) {
+      setBodyOverflowStyle(document.body.style.overflow);
+      document.body.style.overflow = 'hidden';
+    }
 
-      if (restrictScroll && bodyOverflowStyle !== 'hidden') {
-        setBodyOverflowStyle(document.body.style.overflow);
-        document.body.style.overflow = 'hidden';
+    return () => {
+      if (restrictScroll) {
```
ericsoderberghp, line 36: It feels like this test should align more with the test at lines 26-28.

src/js/components/FocusedContainer.js
```diff
@@ -18,44 +13,51 @@ export const FocusedContainer = ({
...
+        document.body.style.overflow = bodyOverflowStyle;
       }
     };
+  }, [bodyOverflowStyle, hidden, trapFocus, restrictScroll]);
 
+  useEffect(() => {
     const timer = setTimeout(() => {
-      if (!hidden && trapFocus) {
-        handleTrapFocus();
+      if (!hidden && trapFocus && root && root[0] !== null) {
```
ericsoderberghp, line 44: I would think `roots` would never have null entries.

src/js/components/FocusedContainer.js
```diff
@@ -18,44 +13,51 @@ export const FocusedContainer = ({
...
+        root.forEach(makeNodeUnfocusable);
       }
     }, 0);
 
     return () => {
-      removeTrap();
+      // remove trap
+      if (root && root[0] !== null) makeNodeFocusable(root[root.length - 1]);
```
ericsoderberghp, line 51: I think a comment about why we only allow focus on the last root here would be helpful.

src/js/components/FocusedContainer.js
taysea: In the browser, I wasn't having any issues, but when I ran the jest tests I was getting failures. Not sure if it was specific to the jest environment but adding a check resolved it

src/js/components/FocusedContainer.js
```diff
@@ -18,44 +13,51 @@ export const FocusedContainer = ({
   const [bodyOverflowStyle, setBodyOverflowStyle] = useState('');
   const ref = useRef(null);
 
+  const root = useContext(RootsContext);
```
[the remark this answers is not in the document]
taysea: Good point, I'll update.

src/js/components/FocusedContainer.js
taysea: I believe so, because if we're not trapping focus in the drop we would want the body to still be able to scroll in case an element on the page is receiving focus

src/js/components/FocusedContainer.js
taysea: That makes sense. I've updated to condition.

src/js/components/FocusedContainer.js
taysea: However, after thinking through it some more, it seems that "restrictScroll" should always be `true` when we want this to happen. The reason I had included it though was because it is part of the existing logic, but doing some testing I think we may be able to take it out.

src/js/components/FocusedContainer.js
taysea: Done.

src/js/components/FocusedContainer.js
taysea: Done.

src/js/components/FocusedContainer.js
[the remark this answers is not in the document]
taysea: Done.

src/js/components/FocusedContainer.js
```diff
@@ -18,44 +13,51 @@ export const FocusedContainer = ({
...
+  const [updatedRoot, setUpdatedRoot] = useState(root);
+
   useEffect(() => {
```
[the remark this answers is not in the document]
taysea: When I place it as a memo, it doesn't rerender to include the `ref.current`. Thoughts?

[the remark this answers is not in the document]
ericsoderberghp: I don't know enough to know why. I'm fine with leaving it as `useEffect`.

ericsoderberghp (approved):

halocline (commented):
@taysea This is looking good. I like this approach. I made a first pass this afternoon, but would like to test manually tomorrow AM.

src/js/components/FocusedContainer.js
```diff
@@ -18,44 +13,56 @@ export const FocusedContainer = ({
...
-      if (!hidden && trapFocus) {
-        handleTrapFocus();
+      if (!hidden && trapFocus && roots && roots[0] !== null) {
+        roots.forEach(makeNodeUnfocusable);
       }
     }, 0);
 
     return () => {
-      removeTrap();
+      // remove trap and restore ability to focus on the last root only
```
halocline, line 54: Question/Wondering: This is a style related comment, but since useEffects aren't named you don't get the benefit of having a clue of what's happening inside like you do with a well named function. 

The comment here is helpful, but I have been wondering about moving them just ahead of the useEffect so the reader knows why the effect is being run sooner.

In this particular instance, its positioning isn't really an issue, but when effects have more going on it can be quite helpful.

Just a wondering... / thought to share.

taysea: that's a good point -- in this case this comment is a little more specific to the clean-up function but I agree that a comment at the start of this effect could be helpful to say "toggle aria-hidden value for roots", and a similar pattern adopted in future

halocline (approved):
Looks good!

Issue #4653, opened 2020-10-26
Collapsible Render Issues


MJordanMilliman:
On occasion, the collapsible component will double render an item or group of items, leading to a jolty behavior, where the content flashes in and then out and then back in a again when opening the collapsible. When closing the collapsible, the opposite scenario is true (content flashes out, then back in and finally out again). 

### Expected Behavior

The content should render once. 

### Actual Behavior

The content seems to double render

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/s/grommetgrommet-sandbox-forked-bhns3?file=/src/Calendar.js

https://bhns3.sse.codesandbox.io/calendar

### Steps to Reproduce

1. Keep clicking the 'Click Me' button after the underlying components finish rendering 
2. Component will occasionally flash. I've consistently been able to reproduce the issue when closing the collapsible for the first time.

### Your Environment

Grommet: 2.15.2;
Chrome:  85.0.4183.102;
Windows 10 Enterprise: 1903

PR #4984, opened 2021-02-20
Collapsibe render issue fix

bojan88:
#### What does this PR do?
fixes https://github.com/grommet/grommet/issues/4653
The problem is nicely described in this issue. It's also noticeable [here](https://storybook.grommet.io/?path=/story/controls-accordion-rich--rich) when you open the first accordion, and on [rich footer card](https://storybook.grommet.io/?path=/story/layout-card-rich-footer--rich-footer). 

It should be straightforward. All the changes are in `src/js/components/Collapsible/Collapsible.js`

Tested in Chrome 88  and Firefox 86.
Did a quick round of testing on BrowserStack in Safari and Edge

We should test Collapsible, Accordion, Rich footer calendar, and other components that use Collapsible component

The solution is a bit hacky.
We have to know animated element size in order to animate from 0px -> [X]px
I wanted to use `scrollHeight` but that wasn't working, mostly with components that have height set - like Panel 1 [here](http://localhost:9001/?path=/story/controls-accordion-custom--custom).

So I ended up setting the content position to absolute (also parent to relative) and hiding it, which renders an invisible DOM element. This way we can read actual element size without displaying the element.

I did not find any issues with this solution, which doesn't mean there are no issues. Please try to find issues if you can.

###### Before: 

https://user-images.githubusercontent.com/1783133/108610214-032b5600-73a2-11eb-9c7b-bbd42ed9801e.mp4

---

###### After:

https://user-images.githubusercontent.com/1783133/108610245-3968d580-73a2-11eb-9d50-ae393eac4c46.mp4

No. Component API is not changed.

I guess, as a bug fix.

No breaking changes.


jcfilben (commented):
Looks good. I am no longer seeing the double render on collapsible. I tested with Collapsible, Card/Rich Footer, Nav/Collapsible, and Accordion stories and didn't find any problems. 

I was initially worried that  rendering an invisible element would cause issues, but since the element use aria-hidden there weren't any accessibility issues that I found. Tabbing/keyboard navigation is still working as expected.

I think it would be helpful to get another pair of eyes on this to review just to make sure I didn't miss anything.

MikeKingdom (approved):
This looks good to me and works well. I've also seen this issue and spent some time testing these changes.

The solution is a little more complicated than I'd like, but I didn't see a simpler way to solve it.

ericsoderberghp (approved):

Issue #4964, opened 2021-02-16
FormField error size theme property not documented


raywong-fb:
### Expected Behavior

The documentation for [formField](https://github.com/grommet/grommet/blob/master/src/js/components/FormField/doc.js) should include an entry for `formField.error.size`.

### Actual Behavior

There is no documentation for the theme field `formField.error.size` even though it appears to work.

- Grommet version: 2
- Browser Name and version: all
- Operating System and version (desktop or mobile): both

g4rry420:
Can I work on this issue ?

PR #4985, opened 2021-02-21
Form field error size theme doc

g4rry420:
#### What does this PR do?
Added the docs for the FormField.error.size

components => FormField => doc.js

I had run the yarn test-update command to update the snapshots

N/A

N/A

N/A

N/A

No

N/A

N/A


g4rry420:
#4964

g4rry420:
I got a question. How can we specify the other value of size property to users such as xsmall, small, large and xlarge ?

ShimiSun:
> I got a question. How can we specify the other value of size property to users such as xsmall, small, large and xlarge ?

Good question, you can add the theme docs for `formField.error.size.small`, `formField.error.size.medium`(...) which will specify the actual pixel value for each.

g4rry420:
How about if we do the same as in the global.borderSize in the formField ?

ShimiSun:
> How about if we do the same as in the global.borderSize in the formField?

Sure. Thanks. It should already exist on the doc, might be worth moving it to docs utils and reuse it.

g4rry420:
I am trying to use global.borderSize but somehow it's not working. Can you please tell how to use them ?

ShimiSun:
```
  'global.borderSize': {
    description: 'The size of the border.',
    type: 'string',
    defaultValue: `{
      xsmall: '1px',
      small: '2px',
      medium: '4px',
      large: '12px',
      xlarge: '24px,
    }`,
  },
```

Taken from Tabs/doc.js

g4rry420:
https://codesandbox.io/s/strange-newton-nlizl?file=/src/App.js

Some borderSize doesn't seem to work.

ShimiSun:
I'm not sure what you are asking, on your sandbox you are using `formField.error.size` which is a prop that isn't mentioned on the docs and not something that grommet supports, so what were you expecting to see?
Also, `borderSize` is just the definition of sizes and it isn't stating default values. This definition is used in the background by grommet to look up what are the values of small/medium/large should be mapped to, e.g. `Box border={{size="large"...}}` would map to border size large according to the large value on `borderSize`.

g4rry420:
ohh, okay, Thank You for the clarification. Btw, `formField.error.size ` in the theme does seems to work if I add different values such as xsmall, small, medium , large and xlarge.

raywong-fb:
> ohh, okay, Thank You for the clarification. Btw, `formField.error.size ` in the theme does seems to work if I add different values such as xsmall, small, medium , large and xlarge.

Yes, `formField.error.size` does work as stated in my [original issue](https://github.com/grommet/grommet/issues/4964). The request was to simply add it to the documentation since it is a working spec.

src/js/components/FormField/doc.js
```diff
@@ -175,6 +175,18 @@ export const themeDoc = {
...
+     The default size is medium.`,
+    type: 'string',
+    defaultValue: `
+      xsmall: '12px'
+      small: '14px'
+      medium: '18px'
+      large: '22px'
+      xlarge: '26px
+    `,
+  },
```
ShimiSun, line 189:
```suggestion
  'formField.error.size': {
    description: `The size of the error messsage to be displayed.
     The default size is medium.`,
    type: 'string',
    defaultValue: medium',
  },
  'formField.error.size.small': {
    description: `The size of the error messsage to be displayed.
     The default size is medium.`,
    type: 'string',
    defaultValue: '12px',
  },
    'formField.error.size.medium': {
    description: `The size of the error messsage to be displayed.
     The default size is medium.`,
    type: 'string',
    defaultValue: '18px',
  },
  // same for the rest of the sizes.
```
It is a little confusing when we say the type of the prop is a  string, but showing an example of an object, the above is a cleaner way to suggest the changes.

ShimiSun (commented):
Let me know if the approach I suggested makes sense?

src/js/components/FormField/doc.js
g4rry420: yes, this make more sense, shall I do that for the rest of it too ?

g4rry420:
I don't understand why some tests are failing.

src/js/components/FormField/doc.js
```diff
@@ -175,6 +175,42 @@ export const themeDoc = {
...
   },
+  'formField.error.size': {
+    description: `The size of the error messsage to be displayed.
+     The default size is medium.`,
+    type: 'string',
+    defaultValue: 'medium',
+  },
+  'formField.error.size.xsmall': {
+    description: `The size of the error messsage to be displayed.
+     The default size is medium.`,
```
ShimiSun, line 186:
```suggestion
    description: `The size of a 'xsmall' error message.`,
```
The same fix for description should be applied to the rest of the sizes.
Also, there is a type for  'messsage' that should be 'message' instead across the other sizes..

ShimiSun (commented):
Almost there! dropped a  comment.

src/js/components/FormField/doc.js
g4rry420: All done, Thank You for all your help and suggestions.

ShimiSun (approved):

Issue #4945, opened 2021-02-09
Enhance RangeInput for keyboard entry


ericsoderberghp:
It could be interesting to explore allowing numeric keyboard input when the RangeInput has focus. Some initial thoughts:
- when the user types the first numeric character, the down arrow, or the space bar while the RangeInput has focus, a Drop is automatically added with a MaskedInput that ties in to min/max.
- the user can then type a specific value and the visual knob tracks along with it
- Esc closes the drop
- Tab closes the drop and advances to the next component.

joaoGabriel55:
It would be a nice enhancement.

joaoGabriel55:
I am working on this

joaoGabriel55:
I am working on this

ericsoderberghp:
I think this needs some design work first, around where the input is placed and how accessibility works. Then, implementation can follow.

PR #4986, opened 2021-02-22
Added Input number to RangeInput component

joaoGabriel55:
#### What does this PR do?

Added a Input number component based on max and min of  RangeInput component. This Input number is showed when user type any number or some arrow key and closed when user type "Tab" or "Esc".

Open storybook and go to SimpleRangeUseInputValue

Need a test

#### Screenshots

image

I think yes

Issue #4945

### DIFF
--- a/src/js/components/RangeInput/README.md
+++ b/src/js/components/RangeInput/README.md
@@ -84,6 +84,7 @@ string
 **inputValue**
 
 The input number type to inform range input value.
+      It default value is false.
 
 ```
 boolean
--- a/src/js/components/RangeInput/README.md
+++ b/src/js/components/RangeInput/README.md
@@ -81,6 +81,15 @@ number
 string
 ```
 
+**inputDirection**
+
+The input direction informed that will define
+      the input number direction relative to range input.
+
+```
+string
+```
+
 **inputValue**
 
 The input number type to inform range input value.
--- a/src/js/components/RangeInput/README.md
+++ b/src/js/components/RangeInput/README.md
@@ -80,6 +80,14 @@ The current value.
 number
 string
 ```
+
+**inputValue**
+
+The input number type to inform range input value.
+
+```
+boolean
+```
   
 ## Intrinsic element
 
--- a/src/js/components/RangeInput/README.md
+++ b/src/js/components/RangeInput/README.md
@@ -84,7 +84,7 @@ string
 **inputValue**
 
 The input number type to inform range input value.
-      It default value is false.
+      The default value is false.
 
 ```
 boolean
--- a/src/js/components/RangeInput/RangeInput.js
+++ b/src/js/components/RangeInput/RangeInput.js
@@ -4,6 +4,20 @@ import { TextInput } from '../TextInput';
 import { FormContext } from '../Form/FormContext';
 import { StyledRangeInput } from './StyledRangeInput';
 
+const DIRECTIONS = {
+  RIGHT: 'right',
+  LEFT: 'left',
+  ABOVE: 'above',
+  BELOW: 'below',
+};
+
+const INPUT_DIRECTIONS = {
+  [DIRECTIONS.RIGHT]: 'row',
+  [DIRECTIONS.LEFT]: 'row-reverse',
+  [DIRECTIONS.ABOVE]: 'column-reverse',
+  [DIRECTIONS.BELOW]: 'column',
+};
+
 const allowedKeysToOpenInputValue = [
   'ArrowRight',
   'ArrowLeft',
@@ -24,6 +38,7 @@ const RangeInput = forwardRef(
       min = 0,
       max = 100,
       step = 1,
+      inputDirection,
       inputValue = false,
       value: valueProp,
       ...rest
@@ -31,6 +46,14 @@ const RangeInput = forwardRef(
     ref,
   ) => {
     const formContext = useContext(FormContext);
+
+    const inputDirectionSelected =
+      INPUT_DIRECTIONS[inputDirection] || INPUT_DIRECTIONS.right;
+
+    const isFillInput =
+      inputDirection === DIRECTIONS.ABOVE || inputDirection === DIRECTIONS.BELOW
+        ? 'horizontal'
+        : false;
 
     const [value, setValue] = formContext.useFormInput(name, valueProp);
 
@@ -50,7 +73,7 @@ const RangeInput = forwardRef(
     };
 
     return (
-      <>
+      <Box fill direction={inputDirectionSelected} align="center">
         <StyledRangeInput
           aria-label={a11yTitle}
           ref={ref}
@@ -91,10 +114,11 @@ const RangeInput = forwardRef(
           type="range"
         />
         {inputValue && showInputValue && (
-          <Box align="center" width="small">
+          <Box align="start" fill={isFillInput}>
             <TextInput
               id="input-range-value"
               type="number"
+              style={{ border: 'none' }}
               min={min}
               max={max}
               step={step}
@@ -113,7 +137,7 @@ const RangeInput = forwardRef(
             />
           </Box>
         )}
-      </>
+      </Box>
     );
   },
 );
--- a/src/js/components/RangeInput/RangeInput.js
+++ b/src/js/components/RangeInput/RangeInput.js
@@ -4,13 +4,13 @@ import { TextInput } from '../TextInput';
 import { FormContext } from '../Form/FormContext';
 import { StyledRangeInput } from './StyledRangeInput';
 
-const allowedKeysToOpenInputMask = [
+const allowedKeysToOpenInputValue = [
   'ArrowRight',
   'ArrowLeft',
   'ArrowUp',
   'ArrowDown',
 ];
-const allowedKeysToCloseInputMask = ['Tab', 'Escape'];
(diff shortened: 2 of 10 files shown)


src/js/components/RangeInput/RangeInput.js
```diff
@@ -1,40 +1,117 @@
...
+            if (allowedKeysToOpenInputMask.includes(event.key)) {
+              setShowInputValue(true);
+            } else if (allowedKeysToCloseInputMask.includes(event.key)) {
+              setShowInputValue(false);
+            }
+          }}
+          type="range"
+        />
+        {inputValue && showInputValue && (
+          <Box align="center" width="100px">
```
ShimiSun, line 92: We try not to hardcode `width` values, so the component will be responsive on small/medium/large devices.

src/js/components/RangeInput/__tests__/RangeInput-test.js
```diff
@@ -105,4 +105,13 @@ describe('RangeInput', () => {
...
   });
+
+  test('inputValue', () => {
+    const { container } = render(
+      <Grommet>
+        <RangeInput min={0} max={10} step={1} value={5} inputValue />
+      </Grommet>,
+    );
+    expect(container.firstChild).toMatchSnapshot();
+  });
```
ShimiSun, line 116: What is this test actually testing? that the component was created and rendered? Can we also check that it accepts the user's input?

src/js/components/RangeInput/doc.js
```diff
@@ -38,6 +38,9 @@ export const doc = RangeInput => {
       PropTypes.number,
       PropTypes.string,
     ]).description('The current value.'),
+    inputValue: PropTypes.oneOfType([PropTypes.bool]).description(
+      'The input number type to inform range input value.',
+    ),
```
ShimiSun, line 43: Should mention a default value.

ShimiSun (commented):
Thanks for the PR buddy! 
I'll probably have more feedback once I'll check the deployment, Netlify is currently not working for me.

src/js/components/RangeInput/__tests__/RangeInput-test.js
joaoGabriel55: Yes, I need to improve this test.

joaoGabriel55:
Thanks a lot for your feedback @ShimiSun!

src/js/components/RangeInput/RangeInput.js
```diff
@@ -1,40 +1,117 @@
...
+          }}
+          onBlur={event => {
+            setFocus(false);
+            if (onBlur) onBlur(event);
+          }}
+          onChange={event => {
+            setValue(event.target.value);
+            if (onChange) onChange(event);
+          }}
+          onKeyDown={event => {
```
abnersajr, line 75: This listener fn becomes a little bit bigger, I would suggest extract it to a `const`.
In the other listeners(onBlur, onFocus, onChange) you are able to pass an extra func.
What about add this:

```ts
if (onKeyDown) onKeyDown(event);
```

src/js/components/RangeInput/__tests__/RangeInput-test.js
```diff
@@ -105,4 +105,36 @@ describe('RangeInput', () => {
     expect(container.firstChild).toMatchSnapshot();
     expect(onChange).toBeCalledTimes(1);
   });
+
+  test('inputValue', () => {
```
abnersajr, line 109: Could we add a test with one of the keys to open?
https://github.com/grommet/grommet/pull/4986/files#diff-b99e2906891e71a7ce78554badfd2fd18f035cd8bc4caff6c595d0cd64cfd4aaR7

src/js/components/RangeInput/RangeInput.js
joaoGabriel55: Hello @abnersajr! Thanks for your feedback, I do not used the `if (onKeyDown) onKeyDown(event);` cause I wanted to keep internal this keydown behavior on RangeInput component.

src/js/components/RangeInput/__tests__/RangeInput-test.js
joaoGabriel55: Nice!!! Thanks a lot

src/js/components/RangeInput/RangeInput.js
abnersajr: `onKeyDown` it's already a value event handler. Some projects might be using it, and remove this can break something.

src/js/components/RangeInput/README.md
```diff
@@ -80,6 +80,15 @@ The current value.
 number
 string
 ```
+
+**inputValue**
+
+The input number type to inform range input value.
+      It default value is false.
```
britt6612, line 87:
```suggestion
      The default value is false.
```

src/js/components/RangeInput/RangeInput.js
```diff
@@ -1,40 +1,119 @@
...
   (
-    { a11yTitle, name, onChange, onFocus, onBlur, value: valueProp, ...rest },
+    {
+      a11yTitle,
+      name,
+      onChange,
+      onFocus,
+      onBlur,
+      onKeyDown,
+      min = 0,
```
britt6612, line 24: I dont think these should be hardcoded as 0,100,1 
If someone wants other values I don't think this will work properly

britt6612:
The new Prop `inputValue` would need to be added to the `index.d.ts` for typescript. This can just be 
`iniputValue?: boolean;`

src/js/components/RangeInput/RangeInput.js
```diff
@@ -1,40 +1,119 @@
...
+      max = 100,
+      step = 1,
+      inputValue = false,
```
britt6612, line 27: Do we need to hardcode these values ?
`min` `max` `step`

src/js/components/RangeInput/index.d.ts
```diff
@@ -9,6 +9,7 @@ export interface RangeInputProps {
   name?: string;
   step?: number;
   value?: number | string;
+  inputValue?: boolean;
```
britt6612, line 12: Can we move this above `min` just to keep things nice and clean and alphabetical

src/js/components/RangeInput/RangeInput.js
joaoGabriel55: Yes, because if these values are not entered, the component will assume the default values

ericsoderberghp:
I'm finding the keyboard interaction to be a bit cumbersome with this. I'm wondering if we should instead explore something that doesn't use a Drop but instead puts a numeric MaskedInput next to the slider. Here are some design concepts: https://designer.grommet.io/?id=RangeInput-eric-soderberg-hpe-com

src/js/components/RangeInput/stories/SimpleRangeUseInputValue.js
```diff
@@ -0,0 +1,29 @@
...
+          min={0}
+        />
+      </Box>
+    </Grommet>
+  );
+};
+
+export default {
+  title: 'Input/RangeInput/SimpleRangeUseInputValue',
+};
```
jcfilben, line 29: Change story name to something shorter and use CSF annotation format
```suggestion
InputValue.storyName = 'Input Value';

export default {
  title: 'Input/RangeInput/Input Value',
};
```

jcfilben (commented):
If the user tabs to the RangeInput and uses the arrowkeys to change the value, when the input number is displayed it shifts the RangeInput down. I think this is happening because the number has some extra padding around it. I wonder if we could do something where if the RangeInput has the `inputValue` prop the extra padding is added initially so that there isn't a shift when the input number is displayed.

Jun-16-2021 10-31-01

vavalos5:
This is a great PR request! I did some research on best practices on RangeInputs, here's what i've gathered:

According to NNG, the simplest way to get a precise value when using a RangeInput is to use a MaskedInput as @ericsoderberghp suggested so that the user can type the exact value for the parameter desired. I think that the explorations Eric added are all possible solutions and are aligned with NNG's guidance on best practices. 

However, in the example that was included, the RangeInput does not have numerical values on either side. If we were to include the MaskedInput, this solution does offer precision if they know which number to insert, but it doesn't allow the user to efficiently explore the ranges that is offered on the RangeInput if the user doesn't know the desired value beforehand as the MaskedInput doesn't reveal the range of possible values. This can make the experience tedious if the user is randomly typing values to figure out the range. According to NNG, a common solution for this is to use a separate linked control such as up and down arrow Buttons for fine adjustment and rough approximation within the same parameter. This would free the user from considering the exact value and allow the user to freely use the up and down arrows to view the range move on the RangeInput.

While I don't recommend the solution with added controls, I would suggest having the numerical range increments included on both sides of the RangeInput for precision with a MaskedInput.

ericsoderberghp:
A few behavioral suggestions first, before digging into the code...

I don't think the input element should be tied to the keyboard keys. I used it via the mouse and had no idea how to get the ability to type a number to show up. I think the caller should be able to specify whether they want the input element or not and when they want it we show it all the time. This will also help avoid the track growing and shrinking while using it.

I'm wondering if the API surface should just be `textInput`, if true it adds it. If an object, that object is passed as props to the underlying `<TextInput />`.

jcfilben:

PR #4991, opened 2021-02-22
React 17 Upgrade

### LINKED ISSUE #4938: Upgrade to major React release
Upgrade to react 17.0.1 including all react plugins and sub dependencies.

jcfilben:
#### What does this PR do?
Upgrade React from 16.12.0 to 17.0.1 (allows for either v16 or v17 to be used)

package.json

yarn test
yarn storybook

See above

More info about react 17 can be found [here](https://reactjs.org/blog/2020/10/20/react-v17.html). New features were not added in this release but there were testing changes

Closes #4938 

No

Yes


package.json
```diff
@@ -44,8 +44,8 @@
     "chromatic": "chromatic -t 9q99in2ygnh"
   },
   "peerDependencies": {
-    "react": ">= 16.6.1",
-    "react-dom": ">= 16.6.1",
+    "react": ">= 17.0.1",
+    "react-dom": ">= 17.0.1",
```
ericsoderberghp, line 48: My one concern with this is that we will be forcing teams using grommet to update react as well. Would it be possible to make the changes on our side that are needed for react 17 in a way that is compatible with both 16 and 17, and then set up our dependencies to allow either?

package.json
```diff
@@ -73,12 +73,12 @@
     "@storybook/addons": "^6.1.17",
     "@storybook/react": "^6.1.17",
     "@storybook/source-loader": "^6.1.17",
-    "@testing-library/dom": "^6.1.0",
-    "@testing-library/jest-dom": "^4.2.4",
-    "@testing-library/react": "^9.1.3",
+    "@testing-library/dom": "^7.29.4",
```
ericsoderberghp, line 76: Would it be possible to split out the updates for `@testing-library` from the updates for `react` into separate pull requests?

src/js/components/Calendar/__tests__/Calendar-test.js
```diff
@@ -185,6 +195,7 @@ describe('Calendar', () => {
     );
     const tree = component.toJSON();
     expect(tree).toMatchSnapshot();
+    component.unmount();
```
ShimiSun, line 198: Is there a configuration on the suit level that can unmount the component after each test run instead of explicitly unmounting it at the end of each test?

package.json
jcfilben: Yes, I created a separate PR for the testing library https://github.com/grommet/grommet/pull/4999

src/js/components/Calendar/__tests__/Calendar-test.js
jcfilben: Within the testing suites there is a combination of tests that use `render` and tests that use `renderer.create` so unmounting the component on all the tests would cause an error in tests that use `render`

package.json
jcfilben: React 17 doesn't have new features from v16 so this should be possible. Would I just need to change the react peerDependencies back to 16.6.1 and leave the react devDependencies at 17.0.1? Or is there more to this

ericsoderberghp: I think there's a richer syntax needed, to allow both 16 and 17 major versions

ericsoderberghp (approved):

package.json
```diff
@@ -44,8 +44,8 @@
     "chromatic": "chromatic -t 9q99in2ygnh"
   },
   "peerDependencies": {
-    "react": ">= 17.0.1",
-    "react-dom": ">= 17.0.1",
+    "react": ">= 16.6.1",
+    "react-dom": ">= 16.6.1",
```
ShimiSun: I think this will throw warnings for users of React 17.
I personally have a side project that uses React 17 and I'm getting this peer dependency warning from grommet for not using the correct version of React.

package.json
```diff
@@ -77,8 +77,8 @@
     "@testing-library/jest-dom": "^5.11.9",
     "@testing-library/react": "^11.0.3",
     "@types/node": "^14.14.22",
-    "@types/react": "^17.0.1",
-    "@types/react-dom": "^17.0.1",
+    "@types/react": "^16.9.11 || ^17.0.1",
+    "@types/react-dom": "^16.9.3 || ^17.0.1",
```
ShimiSun, line 81: I don't think that we need to support the older React version for devDependencies, just for peer dependencies.

jcfilben: switched dev dependencies, "@types/react" and "@types/react-dom" to be 17.0.1 in latest push

package.json
jcfilben: I updated this to be
```javascript
"react": "^16.6.1 || ^17.0.0",
"react-dom": "^16.6.1 || ^17.0.0",
```

package.json
```diff
@@ -118,11 +118,11 @@
...
-    "react": "^16.12.0",
-    "react-dev-utils": "^9.0.3",
-    "react-dom": "^16.12.0",
-    "react-is": "16.12.0",
-    "react-test-renderer": "^16.12.0",
+    "react": "^16.12.0 || ^17.0.1",
+    "react-dev-utils": "^11.0.2",
+    "react-dom": "^16.12.0 || ^17.0.1",
+    "react-is": "^16.12.0 || ^17.0.1",
+    "react-test-renderer": "^16.12.0 || ^17.0.1",
```
ShimiSun, line 125: Let's remove the support for React 16, we should support it only for peer dependency.

jcfilben: Done in latest commit

ShimiSun (approved):

PR #4994, opened 2021-02-24
Drop should always scroll with target

taysea:
#### What does this PR do?
DropContainer should listen for scroll on the window, in addition to any scrollParents of itself to ensure that it always moves with its target.

Previously, if a Drop opened inside of a scroll region and the overall window was scrollable, the Drop would not scroll with its target when the overall window was scrolled.

src/js/components/Drop/DropContainer.js

Adjusted List Action story to be the following and ensured that the "Action" menu that opens on each list item would scroll with its target whether the inner scrollable region was scrolled or the overall window:
```
import React from 'react';

import { Grommet, Box, List, Menu } from 'grommet';
import { More } from 'grommet-icons';

const data = [];

for (let i = 0; i < 40; i += 1) {
  data.push({
    entry: `entry-${i + 1}`,
  });
}

export const Action = () => (
  <Grommet theme={grommet}>
    <Box pad="large">
      <Box height="small" overflow="auto">
        <List
          data={data.slice(0, 10)}
          pad={{ left: 'small', right: 'none' }}
          action={(item, index) => (
            <Menu
              key={index}
              icon={<More />}
              hoverIndicator
              items={[{ label: 'one' }]}
            />
          )}
        />
      </Box>
      [INSERT LONG TEXT TO FORCE BODY TO SCROLL]
    </Box>
  </Grommet>
);
```
Same as above.

Closes https://github.com/grommet/hpe-design-system/issues/1420

No.

Fixes scroll bug where Drop would not always scroll with its target.

Backwards compatible.


src/js/components/Drop/DropContainer.js
```diff
@@ -243,13 +243,15 @@ const DropContainer = forwardRef(
         scrollParents.forEach(scrollParent =>
           scrollParent.addEventListener('scroll', place),
         );
+        window.addEventListener('scroll', place, true);
```
ericsoderberghp, line 246: How is this different from ensuring that `document` is in the `scrollParents`?

taysea: Good comment. I've adjusted findScrollParents to add `document` as a scrollParent and removed it from here. Previously, it was only being added to the list if there were no other scrollable regions on the page, but it seems that we always want to have it included.

ericsoderberghp (approved):

halocline (approved):

PR #4996, opened 2021-02-24
Adding a new prop to validate form

ypinneka:
#### What does this PR do?
Lets the user know if the form is valid.

Form.js

Verified stories with the new behavior

Check for isFormValid returned value from onValidate callback.

No

This is an additional feature which was used in CURO.

Not necessary

Yes

Yes

No breaking changes


src/js/components/Form/Form.js
```diff
@@ -289,6 +299,12 @@ const Form = forwardRef(
           return result;
         };
 
+        if (required) {
+          setRequiredFields(prevValue =>
+            !prevValue.includes(name) ? [...prevValue, name] : [...prevValue],
```
ericsoderberghp, line 304: I think we can just return `prevValue` directly in the false case, don't need to create a new array.

src/js/components/Form/index.d.ts
```diff
@@ -16,6 +16,7 @@ export interface FormProps<T> {
   onValidate?: (validationResults: {
     errors: Record<string, any>;
     infos: Record<string, any>;
+    isFormValid: boolean;
```
ericsoderberghp, line 19: I think we should name this just "valid". We are already in a Form context, so we don't need that term. And, we tend to just use the adjective alone without the "is" prefix elsewhere, such as
```
<Button>
  {({ disabled, hover, focus }) => {}}
</Button>
```

src/js/components/Form/stories/ValidateOnChange.js
```diff
@@ -38,8 +41,8 @@ export const ValidateOnChange = () => (
             ]}
           />
 
-          <FormField label="Email" name="email" required>
-            <TextInput name="email" type="email" />
+          <FormField label="Address" name="address" required>
```
ericsoderberghp, line 44: Any reason we changed from email to address?

ericsoderberghp (commented):
Good progress. A few simple comments. Still need a unit test.

src/js/components/Form/Form.js
ypinneka: missed it.

src/js/components/Form/index.d.ts
ypinneka: Renaming isFormValid to valid

src/js/components/Form/stories/ValidateOnChange.js
ypinneka: If we have email, we were not checking for the email format case from the form validation. It throws an error saying improper format only when we click on submit (as we pass type to TextInput). Ideally, it would be better if we use MaskedInput for email.
So, I preferred a simple use case depicting only what we wanted to demo in the story. 'Address' was something which was on top of my mind at that moment.
Will try to update the story with First Name & Last Name as its more appropriate.

ericsoderberghp:
All is looking good, just need a unit test now

ericsoderberghp (approved):

ShimiSun (commented):
I think this test should pass, and it currently breaks the build
Screen Shot 2021-02-25 at 9 30 52 AM

ShimiSun:
Can you please add more details why you think this PR has breaking changes, quoting from the description above:
>Is this change backwards compatible or is it a breaking change?
Yes

src/js/components/Form/stories/ValidateOnChange.js
```diff
@@ -12,36 +11,38 @@ export const ValidateOnChange = () => (
           validate="change"
           onReset={event => console.log(event)}
           onSubmit={({ value }) => console.log('Submit', value)}
+          onValidate={validationResults => {
```
ShimiSun, line 14: I think this example has improved with these changes, but it does not demonstrate the whole `valid` story. For users to be able to leverage this new functionality, a real use case example of how this prop could be used will be helpful. For example, you can change the story so `valid` will have an effect on the chain of events, from what I understand a common use case for your team is to disable the Submit button when valid is false. It will be nice to see how this end-to-end flow is working on the story.

ypinneka: > Can you please add more details why you think this PR has breaking changes, quoting from the description above:
> 
> > Is this change backwards compatible or is it a breaking change?
> > Yes

Meant to say this change should be compatible, not a breaking change.

ypinneka:
> Can you please add more details why you think this PR has breaking changes, quoting from the description above:
> 
> > Is this change backwards compatible or is it a breaking change?
> > Yes

Meant to say this change should be compatible, not a breaking change.

src/js/components/Form/stories/ValidateOnChange.js
ypinneka: Will add. Makes sense.

src/js/components/Form/__tests__/Form-test-uncontrolled.js
```diff
@@ -609,7 +679,6 @@ describe('Form uncontrolled', () => {
       target: { value: 'J' },
     });
     act(() => jest.advanceTimersByTime(1000)); // allow validations to run
-    expect(queryAllByText('must be >1 character')).toHaveLength(1);
```
ShimiSun: Normally, changing the tests due to build breaks is a recipe for potential breaking changes.
Can you please explain why is this change needed?

ypinneka: I moved it to a new test suite name 'validate on change'.
Wanted to keep this test suite only for testing valid prop.

ShimiSun (approved):

src/js/components/Form/__tests__/Form-test-uncontrolled.js
ypinneka: I get it.
I have moved the line to a different test case 'validate on change'.

Will keep you updated on why this line was failing in this specific case.

ypinneka: // act(() => jest.advanceTimersByTime(1000)); // allow validations to run
    setTimeout(() => {
      expect(queryByText('must be >1 character')).toBe(null);
      expect(valid).toBeFalsy();
    }, 600);

Using setTimeout works, looks like advanceTimersByTime is not the right way here.

PR #4999, opened 2021-02-24
Testing upgrade

jcfilben:
#### What does this PR do?
Upgrades Testing Library to the latest version

`component.unmount()` is necessary to prevent tests from affecting each other when `renderer.create` is used because `afterEach(cleanup)` only covers tests that use `render`.

Note
I didn't add `component.unmount()` to every test that uses renderer.create, only the ones causing test failures. Wondering if I should add it to every test that uses `renderer.create`? Another option is to update these tests to use `render` but this would cause a lot of snapshot changes.

package.json

yarn test

yarn test

No

No

Backwards compatible


ericsoderberghp (approved):

src/js/__tests__/default-props-test.js
```diff
@@ -16,13 +16,15 @@ test('default theme is used', () => {
   const component = renderer.create(<Box background="brand" />);
   const tree = component.toJSON();
   expect(tree).toMatchSnapshot();
+  component.unmount();
```
ShimiSun, line 19: Can we avoid the repetitive `component.unmount();` and leverage a certain clean-up method on the test suite level after each test ends?

jcfilben: Within the testing suites there is a combination of tests that use `render` and tests that use `renderer.create` so unmounting the component on all the tests would cause an error in tests that use `render`

ShimiSun: Saw this comment https://github.com/grommet/grommet/pull/4991#discussion_r582149917

ShimiSun (approved):

ShimiSun:
Linked to  https://github.com/grommet/grommet/pull/4991

PR #5003, opened 2021-02-25
Add extend to drop on Tip

britt6612:
#### What does this PR do?
This PR adds the extend option for `tooltip.drop` in the theme.
tip.js
storybook
Currently, in hpe-design-system we have a alignment issue with the tooltip. We had to add `margin-top` and `margin-bottom` to the theme for `select` that can be found here.
https://github.com/grommet/grommet-theme-hpe/blob/master/src/js/index.js#L226

One idea was to add this for tooltip so that we can set the  
```
margin-top: 0;
margin-bottom: 0;

```

that way the alignment will not be off.

https://github.com/grommet/hpe-design-system/issues/1449
This is the problem

Screen Shot 2021-02-25 at 9 47 43 AM

If the margin is set to 0 then it will be centered aligned

Screen Shot 2021-02-25 at 9 50 14 AM

auto
no
backward compatible

### DIFF
--- a/src/js/components/Tip/README.md
+++ b/src/js/components/Tip/README.md
@@ -62,3 +62,13 @@ Defaults to
 ```
 {align: { top: 'bottom' }}
 ```
+
+**tip.drop.extend**
+
+Any additional style for Drop within tooltip. Expects `string | (props) => {}`.
+
+Defaults to
+
+```
+undefined
+```
--- a/src/js/themes/base.js
+++ b/src/js/themes/base.js
@@ -1255,6 +1255,7 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       drop: {
         // any props for the drop
         align: { top: 'bottom' }, // most common use case is Header with Buttons
+        // extend: undefined,
       },
     },
     video: {
--- a/src/js/components/Tip/Tip.js
+++ b/src/js/components/Tip/Tip.js
@@ -5,11 +5,15 @@ import React, {
   useContext,
   useState,
 } from 'react';
-import { ThemeContext } from 'styled-components';
+import styled, { ThemeContext } from 'styled-components';
 
 import { Box } from '../Box';
 import { Drop } from '../Drop';
 import { useForwardedRef } from '../../utils/refs';
+
+const StyledDrop = styled(Drop)`
+  ${props => props.theme.tip.drop && props.theme.tip.drop.extend}}
+`;
 
 const Tip = forwardRef(({ children, content, dropProps, plain }, tipRef) => {
   const theme = useContext(ThemeContext);
@@ -59,7 +63,7 @@ const Tip = forwardRef(({ children, content, dropProps, plain }, tipRef) => {
   return [
     clonedChild,
     over && (
-      <Drop
+      <StyledDrop
         target={componentRef.current}
         trapFocus={false}
         key="tip-drop"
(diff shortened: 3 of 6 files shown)


src/js/components/Tip/README.md
```diff
@@ -62,3 +62,13 @@ Defaults to
 ```
 {align: { top: 'bottom' }}
 ```
+
+**tip.drop.extend**
+
+Any additional style for Drop within tooltip. Expects `string | (props) => {}`.
```
taysea, line 68:
```suggestion
Any additional style for Drop within Tip. Expects `string | (props) => {}`.
```

src/js/themes/base.js
```diff
@@ -1255,6 +1255,7 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       drop: {
         // any props for the drop
         align: { top: 'bottom' }, // most common use case is Header with Buttons
+        // extend: undefined,
```
taysea, line 1258: Add to base.d.ts also

ShimiSun (commented):
The root cause for this PR is a little odd to me, looking at the DS original issue, it seems like enabling this feature on grommet is for covering up an issue that is caused by the hpe themed drop. Is there a way to fix the issue of the drop on HPE theme side or on the drop side instead?

halocline:
> The root cause for this PR is a little odd to me, looking at the DS original issue, it seems like enabling this feature on grommet is for covering up an issue that is caused by the hpe themed drop. Is there a way to fix the issue of the drop on HPE theme side or on the drop side instead?

I agree that this may not address the root cause. I think we should take a pause on this direction and consider some alternatives.

To me, the root cause is more aligned to the fact that Drop is consumed by multiple components, but has a single styling. However, with various callers in different contexts, the callers may have different styling requirements (e.g. in this case the amount of margin being different in a Menu or Select context from that of the amount desired in a Tip context).

I was wondering about using `<StyledTableCell />` as a possible pattern. It is able to style a table cell depending on the cell's context (e.g. is it in the header? body? footer? Is it pinned? etc.).

There are probably other solutions for us to consider, but to me the root is that I may want my Drop to be styled differently, dependent on context.

britt6612:
> > The root cause for this PR is a little odd to me, looking at the DS original issue, it seems like enabling this feature on grommet is for covering up an issue that is caused by the hpe themed drop. Is there a way to fix the issue of the drop on HPE theme side or on the drop side instead?
> 
> I agree that this may not address the root cause. I think we should take a pause on this direction and consider some alternatives.
> 
> To me, the root cause is more aligned to the fact that Drop is consumed by multiple components, but has a single styling. However, with various callers in different contexts, the callers may have different styling requirements (e.g. in this case the amount of margin being different in a Menu or Select context from that of the amount desired in a Tip context).
> 
> I was wondering about using `<StyledTableCell />` as a possible pattern. It is able to style a table cell depending on the cell's context (e.g. is it in the header? body? footer? Is it pinned? etc.).
> 
> There are probably other solutions for us to consider, but to me the root is that I may want my Drop to be styled differently, dependent on context.

I can try this approach as well.. I can dig a little more to see everything that uses drop under the hood and create different files that style depending on what is being used by the drop.

britt6612:
> > I can try this approach as well.. I can dig a little more to see everything that uses drop under the hood and create different files that style depending on what is being used by the drop.
> 
> If you have some time to review the approach, I think that would be worthwhile, but it might be helpful to have others weigh in on the validity of the approach before investing too much time.

I think it would be worthwhile to run this approach by @ShimiSun or @ericsoderberghp

halocline:
> I think it would be worthwhile to run this approach by @ShimiSun or @ericsoderberghp

Agree. @taysea and I were talking through this approach as well https://github.com/grommet/grommet/pull/5005 which might be a better alternative.

ShimiSun:
Consolidating this work effort with to #5005. I'll be closing this PR for now as we'd be taking a new approach on the Drop alignment.

PR #5005, opened 2021-02-25
Allow Menu drop to be themed, any Drop props accepted

taysea:
#### What does this PR do?
Adds `theme.menu.drop` to theme to allow users to style their Menu drop from the theme rather than having to repeat the styling every time via props. This is beneficial for apps that always want a Menu to have drop alignment of, for example, `{ top: 'bottom', left: 'left' }`.

Tested in storybook example by creating custom theme and added jest test.

Create custom theme.

HPE Design System calls for Menu that is always aligned `{ top: 'bottom', left: 'left' }`. Instead of having teams implement on every Menu instance, it should come automatically from theme.

Screen Shot 2021-02-25 at 2 51 03 PM

Related to https://github.com/grommet/hpe-design-system/issues/1416

Yes, updated here.

Added `theme.menu.drop` to allow theming of Menu's drop alignment.

Backwards compatible.

### DIFF
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -72,9 +72,9 @@ const Menu = forwardRef((props, ref) => {
   const iconColor = normalizeColor(theme.menu.icons.color || 'control', theme);
   // need to destructure the align otherwise it will get passed through
   // to DropButton and override prop values
-  const { align: menuAlign, ...themeDropProps } = theme.menu.drop;
+  const { align: themeDropAlign, ...themeDropProps } = theme.menu.drop;
 
-  const align = (dropProps && dropProps.align) || dropAlign || menuAlign;
+  const align = (dropProps && dropProps.align) || dropAlign || themeDropAlign;
   const controlButtonIndex = useMemo(() => {
     if (align.top === 'top') return -1;
     if (align.bottom === 'bottom') return items.length;
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -70,7 +70,7 @@ const Menu = forwardRef((props, ref) => {
   } = props;
   const theme = useContext(ThemeContext) || defaultProps.theme;
   const iconColor = normalizeColor(theme.menu.icons.color || 'control', theme);
-  const align = dropProps.align || dropAlign;
+  const align = dropProps.align || dropAlign || theme.menu.drop.align;
   const controlButtonIndex = useMemo(() => {
     if (align.top === 'top') return -1;
     if (align.bottom === 'bottom') return items.length;
@@ -357,10 +357,6 @@ const Menu = forwardRef((props, ref) => {
 });
 
 Menu.defaultProps = {
-  dropAlign: {
-    top: 'top',
-    left: 'left',
-  },
   dropProps: {},
   items: [],
   messages: {
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -70,7 +70,11 @@ const Menu = forwardRef((props, ref) => {
   } = props;
   const theme = useContext(ThemeContext) || defaultProps.theme;
   const iconColor = normalizeColor(theme.menu.icons.color || 'control', theme);
-  const align = dropProps.align || dropAlign || theme.menu.drop.align;
+  // need to destructure the align otherwise it will get passed through
+  // to DropButton and override prop values
+  const { align: menuAlign, ...themeDropProps } = theme.menu.drop;
+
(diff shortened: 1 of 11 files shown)


halocline:
So, I like the idea of having smart defaults in the theme so that implementors don't need to remember to set every time.

I also think this might be a better approach than what I suggested in this discussion: https://github.com/grommet/grommet/pull/5003#pullrequestreview-598937494


> HPE Design System calls for Menu that is always aligned `{ top: 'bottom', left: 'left' }`. Instead of having teams implement on every Menu instance, it should come automatically from theme.
> 
> Screen Shot 2021-02-25 at 2 51 03 PM
> 

Regarding this, I think "always" might be a bit strong stance to have as I am seeing a lot of menu instances where the menu is aligned to the right of a table which will bump up against other elements and/or be near the edge of the screen, like so:
Screen Shot 2021-02-25 at 4 36 46 PM

src/js/themes/base.d.ts
```diff
@@ -970,6 +970,9 @@ export interface ThemeType {
   };
   menu?: {
     background?: BackgroundType;
+    drop?: {
+      align?: AlignContentType;
+    };
```
halocline, line 975: Should we be thinking about this a bit more broadly such as a DropType and support other drop props? This would be relevant to this discussion: https://github.com/grommet/grommet/pull/5003#pullrequestreview-598937494

halocline (commented):
No concerns with the implementation. Just some questions for discussion around whether we want to adopt this pattern and its applicability elsewhere.

src/js/themes/base.d.ts
taysea: I think that would be fair to consider accepting all dropProps (considering they can already be adjusted on Menu via dropProps prop). I had just done align to start because I didn't want to over-expand the theme if unnecessary, but I think given the issue linked, that may be a fair extension of the theme.

src/js/components/Menu/doc.js
```diff
@@ -139,6 +139,14 @@ export const themeDoc = {
     type: 'string',
     defaultValue: undefined,
   },
+  'menu.drop.align': {
+    description: 'The alignment of the drop with respect to the menu button.',
+    type: 'string',
```
ShimiSun, line 144:
```suggestion
    type: 'string | object',
```
Like the default value suggests

taysea: Good catch, thank you.

src/js/themes/base.js
```diff
@@ -982,6 +982,12 @@ export const generate = (baseSpacing = 24, scale = 6) => {
     menu: {
       // background: undefined,
       // extend: undefined,
+      drop: {
+        align: {
+          top: 'top',
+          left: 'left',
+        },
+      },
```
halocline, line 990: I was starting to ask "How will this affect existing implementations? Do we want this undefined and let users opt-in?", but now see that these were previously defaultProps on Menu.js and we are simply moving where the defaults are applied, correct?

src/js/components/Menu/doc.js
taysea: Done.

src/js/themes/base.js
taysea: Yes exactly. Previously, this same default was defined as Menu.defaultProps `dropAlign`. With the adjustment of the logic to fall back to the theme, I've just moved the default into the theme. No changes will be experienced on existing applications

src/js/components/Menu/Menu.js
```diff
@@ -70,7 +70,11 @@ const Menu = forwardRef((props, ref) => {
   } = props;
   const theme = useContext(ThemeContext) || defaultProps.theme;
   const iconColor = normalizeColor(theme.menu.icons.color || 'control', theme);
-  const align = dropProps.align || dropAlign;
+  // need to destructure the align otherwise it will get passed through
+  // to DropButton and override prop values
+  const { align: menuAlign, ...themeDropProps } = theme.menu.drop;
```
ericsoderberghp, line 75: Can we rename "menuAlign" to "themeDropAlign"? It feels a bit more descriptively accurate.

ericsoderberghp (commented):
one small naming suggestion

src/js/components/Menu/Menu.js
taysea: Yep! Will adjust now.

taysea: Done.

ericsoderberghp (approved):

halocline (approved):

ShimiSun (approved):

Issue #4978, opened 2021-02-18
[TS]: Fix Component Props declarations


abnersajr:
Pay attention to rules for migration at the end of the issue.

### Expected Behavior

Be able to import and extend the `<component>Props` correctly, bringing the ` JSX.IntrinsicElements` props together.

### Actual Behavior

Components have theirs on `<component>Props` combined with ` JSX.IntrinsicElements` being exported using the `declare` block. The import of `<component>Props` doesn't have all the props defined in the `declare` block.

References:
https://github.com/grommet/grommet/pull/4909
https://github.com/grommet/grommet/issues/4976

### Proposed Solution

As discussed in the PR: https://github.com/grommet/grommet/pull/4909,
We need to check which components follow the pattern of combining the props on `declare` blocks.
A second item is that many components are defined as `React.ComponentClass` but can be modified to `React.FC`.

That being said, we should start refining these declarations following the examples below, where we create and export an extended version of the props instead of removing the actual export:

```ts
export interface LayerProps {...}
export type LayerPropsExtended = LayerProps & JSX.IntrinsicElements['div'] 

// React.FC or React.componentClass must be evaluated to see if the change is possible.
// Preference to use React.FC
declare const Layer: React.FC<LayerPropsExtended>;

export { Layer };
```
```ts
// another example
export interface TextInputProps { ... };

export interface TextInputExtendedProps extends Omit<
JSX.IntrinsicElements['input'],
'onSelect' | 'size' | 'placeholder'
>, TextInputProps {};

declare const TextInput: React.FC<TextInputExtendedProps>;

export { TextInput };
```

### Roadmap

Thanks to @Isaius for check and get this list for us.

 These do not extend `JSX.IntrinscElements` in their props interface but in the `declare` block.
 
- [ ] [Accordion](https://github.com/grommet/grommet/blob/master/src/js/components/Accordion/index.d.ts)
- [ ] [AccordionPanel](https://github.com/grommet/grommet/blob/master/src/js/components/AccordionPanel/index.d.ts)
- [ ] [Anchor](https://github.com/grommet/grommet/blob/master/src/js/components/Anchor/index.d.ts)
- [ ] [Carroussel](https://github.com/grommet/grommet/blob/master/src/js/components/Carousel/index.d.ts) [ComponentClass]
- [ ] [CheckBox](https://github.com/grommet/grommet/blob/master/src/js/components/CheckBox/index.d.ts)
[...]

Issue #4998, opened 2021-02-24
[TS]: Extended types export


abnersajr:
Pay attention to rules for migration at the end of the issue.
This is a tracking ticket for #4978 
### Expected Behavior

Be able to import and extend the `<component>Props` correctly, bringing the ` JSX.IntrinsicElements` props together.

### Actual Behavior

Components have theirs on `<component>Props` combined with ` JSX.IntrinsicElements` being exported using the `declare` block. The import of `<component>Props` doesn't have all the props defined in the `declare` block.

References:
https://github.com/grommet/grommet/pull/4909
https://github.com/grommet/grommet/issues/4976

### Proposed Solution

As discussed in the PR: https://github.com/grommet/grommet/pull/4909,
We need to check which components follow the pattern of combining the props on `declare` blocks.
A second item is that many components are defined as `React.ComponentClass` but can be modified to `React.FC`.

That being said, we should start refining these declarations following the examples below, where we create and export an extended version of the props instead of removing the actual export:

```ts
// another example
export interface TextInputProps { ... };

export interface TextInputExtendedProps extends Omit<
JSX.IntrinsicElements['input'],
'onSelect' | 'size' | 'placeholder'
>, TextInputProps {};

declare const TextInput: React.FC<TextInputExtendedProps>;

export { TextInput };
```

### Roadmap

Thanks to @Isaius for check and get this list for us.

 These do not extend `JSX.IntrinscElements` in their props interface but in the `declare` block.
 
- [x] [Accordion](https://github.com/grommet/grommet/blob/master/src/js/components/Accordion/index.d.ts) (#5010)
- [x] [AccordionPanel](https://github.com/grommet/grommet/blob/master/src/js/components/AccordionPanel/index.d.ts) (#5010)
- [x] [Anchor](https://github.com/grommet/grommet/blob/master/src/js/components/Anchor/index.d.ts)
- [x] [Avatar](https://github.com/grommet/grommet/blob/master/src/js/components/Avatar/index.d.ts)
- [x] [Carroussel](https://github.com/grommet/grommet/blob/master/src/js/components/Carousel/index.d.ts) (#5010)[ComponentClass]
- [x] [CheckBox](https://github.com/grommet/grommet/blob/master/src/js/components/CheckBox/index.d.ts) (#5010)
- [x] [CheckBoxGroup](https://github.com/grommet/grommet/blob/master/src/js/components/CheckBoxGroup/index.d.ts)
[...]

PR #5010, opened 2021-02-26
[TS] Fix Accordion, AccordionPanel, Carousel and CheckBox extended types definition

Isaius:
#### What does this PR do?
This PR adds and exports <component>ExtendedProps declarations for the Accordion, AccordionPanel, Carousel and CheckBox components, forming part of issue #4998.

The changes are in the type definition files, and are simple and do not have any dependencies from external files. As they are individual, the review can start for any of them.

Since it's type definition, no tests were added. All the existing tests passed locally.

Extending any of the changed components have to contain the respective extended `JSX.IntrinsicElements` 

This changes were discussed in the #4978 

Progress track: #4998 
Discussion: #4978 

No.

No. Just all the components tracked in #4998 that now are fixed when extending. 

backwards compatible.


abnersajr (approved):

src/js/components/Accordion/index.d.ts
```diff
@@ -2,16 +2,17 @@ import * as React from 'react';
...
   animate?: AnimateType;
   onActive?: (activeIndexes: number[]) => void;
   multiple?: boolean;
   messages?: { tabContents?: string };
 }
 
-declare const Accordion: React.FC<BoxProps &
-  AccordionProps &
-  JSX.IntrinsicElements['div']>;
+export type AccordionExtendedProps = AccordionProps &
```
ShimiSun, line 13: I think we should export an interface rather than a type.

ShimiSun (changes requested):

src/js/components/Accordion/index.d.ts
Isaius: That's a problem we discussed in the closed issue #4978 . `extends` only allow extend one item, components that extends more than that need a special form.
I suggested use a intermediate variable, something like that:

```ts
interface a extends b { }
export interface c extends a { }
```

But we agree that isn't good, so @abnersajr came with this approach using `type` and the `&`.

abnersajr: There is a way to do it.
Here is an example.
https://github.com/grommet/grommet/pull/5040/files#diff-5f4f62bcd0e71612e1588e56d33792b4a74c8e011fb6b5fe9b06bb8436ddc307R23

abnersajr (changes requested):
Change to export interfaces.

src/js/components/Accordion/index.d.ts
Isaius: sure, i'll change that

src/js/components/Accordion/index.d.ts
```diff
@@ -2,16 +2,18 @@ import * as React from 'react';
 import { AnimateType } from '../../utils';
 import { BoxProps } from '../Box';
 
-export interface AccordionProps {
+export interface AccordionProps extends BoxProps {
```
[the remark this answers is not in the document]
Isaius: Done. But the branch is outdated and IDK how to do with it.
I've tried to update and... things went wrong. So, I changed back to the previous commit.

ShimiSun (approved):

Issue #5012, opened 2021-02-26
Form's valid prop inside onValidate() callback not returning accurate value for checkbox


shreyashah7:
I have a form with some required input fields and required checkbox. onValidate() callback is not returning correct value for checkbox and it always returns false. If i remove checkbox, it works fine.

### Expected Behavior

Needs to check checkbox and return correct value

### Actual Behavior

Not returning correct value

CC: @ypinneka

ypinneka:
https://github.com/grommet/grommet/pull/5014

ShimiSun:
Thanks for the contribution @ypinneka , can we close this issue?  Or are there more functionalities that are expected as part of this issue?

ShimiSun:
I think this issue could be closed. Feel free to reopen if needed.

PR #5014, opened 2021-02-26
Valid prop inside onValidate callback of Form was not considering the required case of Checkbox.

ypinneka:
#### What does this PR do?
Addition to the PR (https://github.com/grommet/grommet/pull/4996), where checkbox's case check was missing

Form.js

Tested on new stories covering different use cases & creating new test cases.

Submit button should be enabled in the story, 'validate on change' after passing validity checks.

https://github.com/grommet/grommet/pull/4996

https://github.com/grommet/grommet/issues/5012

No

No

Not a breaking change


src/js/components/Form/Form.js
```diff
@@ -118,6 +118,12 @@ const Form = forwardRef(
 
             let valid = false;
 
+            setRequiredFields(
```
ericsoderberghp, line 121: This seems like the wrong place to put this. Why isn't this covered by this block later on?:
```
        if (required) {
          setRequiredFields(prevValue =>
            !prevValue.includes(name) ? [...prevValue, name] : prevValue,
          );
        }
```

ypinneka: I can do something like this,

        if (validateArg || required) {
          validations.current[name] = validateField;
          setRequiredFields(prevValue =>
            required 
            && !prevValue.includes(name) 
            && Object.keys(validations.current).includes(name) 
            ? [...prevValue, name] : prevValue,
          );

But in useFormField, on every mount, the requiredFields will be set to empty unless validation triggers.

ypinneka: Wanted to make the filter below where you were deleting unmounted components (error/info object keys).
You are right, made the change.

src/js/components/Form/Form.js
```diff
@@ -300,9 +301,7 @@ const Form = forwardRef(
         };
 
         if (required) {
-          setRequiredFields(prevValue =>
-            !prevValue.includes(name) ? [...prevValue, name] : prevValue,
-          );
+          setRequiredFields(() => Object.keys(validations.current));
```
ericsoderberghp, line 304: This doesn't seem quite right, there might be fields that are marked required that do not have a validation.

Instead, I think we want to both add and remove here:
```
if (required) setRequiredFields(prevValue => !prevValue.includes(name) ? [...prevValue, name] : prevValue);
else setRequiredFields(prevValue => prevValue.filter(v => v !== name));
```

ypinneka: This does not work in the case when a field is unmounted.

Eg.,
  const defaultValue = {
    ampm: 'Field One',
    name1: '',
    name2: '',
  };
  const [value, setValue] = useState(defaultValue);
  const [selectedOption, setSelectedOption] = useState('Field One');

  const optionOne = (
    <FormField
      label="Name 1"
      required
      name="name1"
    >
      <TextInput name="name1" />
    </FormField>
  );

  const optionTwo = (
    <FormField
      label="Name 2"
      validate={[
        (val) => {
          if (val && val.length === 1)
            return { message: 'Text must be >1 character', status: 'error' };
          return undefined;
        },
        (val) => {
          if (val) return { message: 'Informational Status', status: 'info' };
          return undefined;
        },
      ]}
      tooltip="This information is publicly available !"
      required
      name="name2"
    >
      <TextInput name="name2" />
    </FormField>
  );

  return (
    <Grommet>
      <Form
        validate="change"
        onChange={(obj) => {
          setValue(obj);
          setSelectedOption(obj.ampm);
        }}
        value={value}
        onValidate={(validationResults) => {
          console.log('validationResults = ', validationResults);
        }}
      >
          <FormField
            label="Select a Form"
            name="ampm"
            required
          >
            <RadioButtonGroup name="ampm" options={['Field One', 'Field Two']} />
          </FormField>
          {selectedOption === 'Field One' ? optionOne : optionTwo}
      </Form>
    </Grommet>
  );

ericsoderberghp: I didn't quite follow the code snippets above. I think I'm getting it though, we need to `setRequiredFields()` in the case where fields are unmounted. I'm wondering if we should do this inside the effect at line 84, but outside of the call to `setValidationResults()` around line 96.

ypinneka: I think `validations.current` would be better if placed in useEffect rather than in useFormField.
Reverting back to initial commit where required fields are filtered based on current validations.

ypinneka: The filter can even be in validateField function.

        const validateField = (value2, data) => {
          let result;
          if (
            required &&
            // false is for CheckBox
            (value2 === undefined || value2 === '' || value2 === false)
          ) {
            result = messages.required;
          } else if (validateArg) {
            if (Array.isArray(validateArg)) {
              validateArg.some(aValidate => {
                result = validateSingle(aValidate, value2, data);
                return !!result;
              });
            } else {
              result = validateSingle(validateArg, value2, data);
            }
          }
          setRequiredFields(
            requiredFields.filter(n =>
              [Object.keys(validations.current)].includes(n),
            ),
          );
          return result;
        };

ypinneka: Didnt even notice that am setting it inside `setValidationResults`.
Made the change, thanks !

src/js/components/Form/Form.js
```diff
@@ -93,6 +93,12 @@ const Form = forwardRef(
           );
           setPendingValidation(undefined);
 
+          setRequiredFields(
```
ericsoderberghp, line 96: Just to be a bit more tight, can we parallel how we use setValidationResults()? For example:
`setRequiredFields(prevRequiredFields => prevRequiredFields.filter(n => Object.keys(validations.current).includes(n)));`

ericsoderberghp (commented):
One small comment. Looking great overall.

src/js/components/Form/Form.js
ypinneka: ok

ericsoderberghp (approved):

ShimiSun (approved):

PR #5018, opened 2021-02-28
[TS] Fix RadioButton, RadioButtonGroup, RangeInput and RangeSelector type interfaces

alexmirrington:
#### What does this PR do?

This PR adds and exports `<component>ExtendedProps` declarations for the RadioButton, RadioButtonGroup, RangeInput and RangeSelector components, forming part of issue #4998.

The only files changed are the TS declaration files for each of the four components. Any of these files are good places to start.

Importing the extended prop interfaces from a separate project should allow `JSX.IntrinsicElements` keys to be passed through to the imported interface, e.g. importing `RadioButtonExtendedProps` should make `input` keys available, as well as general HTML keys like `children`. Also double check any keys included in `Omit<..., ...>` types for consistency with original intersected types.

Progress tracking: #4998
Design/discussion: #4978

No

Backwards compatible.

### DIFF
--- a/src/js/components/RadioButton/index.d.ts
+++ b/src/js/components/RadioButton/index.d.ts
@@ -14,6 +14,6 @@ export interface RadioButtonExtendedProps
   extends RadioButtonProps,
     Omit<JSX.IntrinsicElements['input'], 'name'> {}
 
-declare const RadioButton: React.ComponentClass<RadioButtonExtendedProps>;
+declare const RadioButton: React.FC<RadioButtonExtendedProps>;
 
 export { RadioButton };
--- a/src/js/components/RadioButton/index.d.ts
+++ b/src/js/components/RadioButton/index.d.ts
@@ -10,7 +10,10 @@ export interface RadioButtonProps {
   name: string;
 }
 
-declare const RadioButton: React.ComponentClass<RadioButtonProps &
-  JSX.IntrinsicElements['input']>;
+export interface RadioButtonExtendedProps
+  extends RadioButtonProps,
+    Omit<JSX.IntrinsicElements['input'], 'name'> {}
+
+declare const RadioButton: React.ComponentClass<RadioButtonExtendedProps>;
 
 export { RadioButton };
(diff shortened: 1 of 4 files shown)


abnersajr (approved):

ShimiSun (commented):
Thanks for the PR, overall it looks well.
I see that you've added an Omit for certain props --a functionality that wasn't there on the original code. How did you come up with those props and why were they omitted?

alexmirrington:
> I see that you've added an Omit for certain props --a functionality that wasn't there on the original code. How did you come up with those props and why were they omitted?

@ShimiSun The main reason for inclusion of the `Omit` generics is that we are exporting interfaces instead of type intersections. Let's take the `RadioButtonExtendedProps` as an example:

We have the existing `RadioButtonProps` interface which defines a `name: string` field, as well as `JSX.IntrinsicElements['input']`, which contains field for relevant HTML `input` tag attributes. Since `JSX.IntrinsicElements['input']` also defines a `name?: string` field, we can create type intersections but not interfaces extending both `RadioButtonProps` and `JSX.IntrinsicElements['input']`:

Screen Shot 2021-03-09 at 3 08 12 pm

In this example, taking the type intersection of `RadioButtonProps & JSX.IntrinsicElements['input']` yields a new type with a `name: string` field, so omitting the `name?: string` field from `JSX.IntrinsicElements['input']` allows us to create an interface without any issues. I noticed you also used the same approach [here](https://github.com/grommet/grommet/pull/5021/files#diff-37524e012a5eb198c557ffc83a19486994866790935ef45e467d68096560ca90R26).

When doing some further reading however, I noticed there is something inherently wrong with the way the type intersections have been used in Grommet in the first place; consider the following example:

Screen Shot 2021-03-09 at 3 33 09 pm

We see that since `y` is a `string` in `Foo` type but a `number` in `Bar`, the type intersection `FooBar` actually contains the field `y` with type `never`, i.e. the field `y` should **never occur at all**. We got away with this for the `RadioButton` example above since `string` is more strict than `string?`, but I imagine there are other components where we will run into fields with `never` types.

I'm happy to discuss the best way forward, but in my opinion, it makes more sense to omit conflicting fields from the `JSX.IntrinsicElements` types and make sure the fields from Grommet's component prop interfaces (e.g. `RadioButtonProps`) are used, since those are the types expected by the React components themselves. Then the react components can internally handle any potential type conflicts between the received prop field and what is passed through to the underlying HTML tag.

ShimiSun (approved):
Good Stuff!! Thank you for all the helpful details and for your contribution!

Issue #5001, opened 2021-02-24
Feature request: label and toggle alignment


jlarso11:
### Expected Behavior

We would like a way to add space between the label and the toggle when they are in reverse.  This way, when added many toggles in a row, we do not see a ragged line of toggles that are dependent on the length of the label.  

Designed: 
![image](https://user-images.githubusercontent.com/14100034/109072850-d2556480-76b2-11eb-8757-172b8724aa49.png)

Actual: 
![image](https://user-images.githubusercontent.com/14100034/109072872-da150900-76b2-11eb-83cd-966e83b54fd6.png)

### Actual Behavior

The toggle is set just to the right of the label with some padding.  No matter the width of the containing box

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-forked-j78lj?file=/index.js

### Your Environment

- Grommet version: 2.16.2
- Browser Name and version: Chrome 88

PR #5019, opened 2021-02-28
Feature label toggle alignment

sosolidkk:
#### What does this PR do?
This PR aims to implement the improvement mentioned in issue #5001 . The idea here is to pass a props called `containerWidth` so that each `Checkbox` can have a variable size inside a `Container`.

`./src/js/components/Checkbox/CheckBox.js`

Manual testing, storybook testing (including a brand new example) and update snapshots that already cover Checkbox plain usage.

```js
import React from 'react';

import { grommet, Box, Grommet, FormField, CheckBox } from 'grommet';

export const CustomWidth = () => (
  <Grommet theme={grommet}>
    <Box pad="small" width="20%">
      <FormField name="required-field" label="Label" required>
        <CheckBox
          name="checkbox"
          id="required-field"
          label="Validation"
          toggle
          reverse
          containerWidth="50%"
        />

        <CheckBox
          name="checkbox"
          id="required-field"
          label="Validation"
          toggle
          reverse
          containerWidth="40%"
        />
      </FormField>
    </Box>
  </Grommet>
);
```
#5001 

image

Yes

I believe it's backwards compatible, since it's only a new prop.

### DIFF
--- a/src/js/components/CheckBox/StyledCheckBox.js
+++ b/src/js/components/CheckBox/StyledCheckBox.js
@@ -37,6 +37,8 @@ const StyledCheckBoxContainer = styled.label`
   align-items: center;
   user-select: none;
   width: fit-content;
+  ${props =>
+    props.width && `width: ${props.width}; justify-content: space-between;`}
   ${props => props.disabled && disabledStyle}
   ${props => !props.disabled && 'cursor: pointer;'}
   ${props => props.theme.checkBox.hover.border.color && hoverStyle}
--- a/src/js/components/CheckBox/doc.js
+++ b/src/js/components/CheckBox/doc.js
@@ -54,6 +54,12 @@ export const doc = CheckBox => {
 NOTE: This can only be used with non-toggle components`,
       )
       .defaultValue(false),
+    containerWidth: PropTypes.string
+      .description(
+        `Define a value for the container width and 
+        set value for justify content to space between.`,
+      )
+      .defaultValue(false),
   };
 
   return DocumentedCheckBox;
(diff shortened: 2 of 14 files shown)


src/js/components/CheckBox/doc.js
```diff
@@ -54,6 +54,12 @@ export const doc = CheckBox => {
 NOTE: This can only be used with non-toggle components`,
       )
       .defaultValue(false),
+    containerWidth: PropTypes.string
+      .description(
+        `Define a value for the container width and 
+        set value for justify content to space between.`,
+      )
+      .defaultValue(false),
```
ShimiSun, line 62: How is the default value set to boolean when the type of the prop is requesting a string?

src/js/components/CheckBox/StyledCheckBox.js
```diff
@@ -37,6 +37,8 @@ const StyledCheckBoxContainer = styled.label`
   align-items: center;
   user-select: none;
   width: fit-content;
+  ${props =>
+    props.width && `width: ${props.width}; justify-content: space-between;`}
```
ShimiSun, line 41: A couple of questions:
1. Are we always assuming that if `width` is provided we'll use justify between layout?
2. The current implementation isn't supporting t-shirt sizing of width such as 'small'/'medium'/... which isn't consistent with grommet `width` functionalities.

ericsoderberghp:
`containerWidth` doesn't feel very consistent with the rest of grommet. I think the primary context of the related issue is that FormFields containing CheckBox components should align their toggle buttons to the right. As such, I think we should add a `fill` property to CheckBox. This `fill` would be passed to the `<label />` Box. This would align more with how `fill` is used in other places in grommet.

britt6612:
Hi, @sosolidkk any updates with this PR?

ShimiSun:
Thank you for submitting this PR! 
We'll be closing this PR due to its slow traffic and conflicted files with the master branch, feel free to reopen the PR after reviewing our comments and addressing the branch conflicts.

PR #5020, opened 2021-03-01
Add linearGradient to Meter

leossantiago:
#### What does this PR do?

This PR proposes a way of using linearGradients as `color` in `<Meter /> component, instead of just regular solid colors.
This was achieved by adding `<linearGradient />` definitions to the svg generated by Meter. This works for both Bar and Circle types.

Gradient properties and stops are set using grommet's Theme, and any valid linearGradient/stop property can be easily added.

This PR also reverses `paths` and `pathCaps` on Circle Meter, in order to prevent weird caps when user is not using `gradientUnits: 'userSpaceOnUse'` as gradient prop. This also prevents caps border leaking on top of previous paths.

Before:
image

After:
image

In order to use gradients, user should pass a string to `color` following this format: `#grad-[gradientName]`, ex: `grad-linear`, `grad-grommet`, `grad-pretty`.

And Theme should have:
```
//corresponds to #grad-linear
  meter: {
    gradients: {
      linear: {
        props: {
          id: 'linear',
          gradientTransform: 'rotate(90)',
          gradientUnits: 'userSpaceOnUse',
        },
        stops: [
          { offset: '5%', stopColor: 'red' },
          { offset: '95%', stopColor: 'gold' },
        ],
      },
    },
  },

```

`src/js/components/Meter/Circle.js`
`src/js/components/Meter/Bar.js`

Manual and storybook testing.

A story was created to showcase different approaches to gradients. 
`Visualizations/Meter/Gradient`

image

Yes. Already did.

Backwards compatible, as Gradients do not impact on current behavior of Meter. Some snapshots could break due to switching pathCaps and paths order on Circle Meter.


ericsoderberghp:
I am very open to supporting gradients with Meter. I think we shouldn't tie it as much to the theme and make it easier to use via props. Did you look at how we support gradients with Chart to see if we can do something similar?
https://storybook.grommet.io/?path=/story/visualizations-chart-gradient--gradient-charts
https://v2.grommet.io/chart#color

leossantiago:
Hey everyone :)
I see that you added a new flag to it.
Since the other PR I was working on seems to have been addressed by another community member, I'll be moving on to this one. Should have the comments addressed soon.

leossantiago:
~I'm finally back. Yay~

@ericsoderberghp I actually LOVE how gradients are supported with `Chart`. I was trying to decouple some complexity from props but ended cluttering the theme. I agree that we can achieve an easier approach by passing down the gradient as an array of objects. I'm currently working on an update to this PR and noticed a couple of things by checking `Chart` component approach:

While Chart applies a gradient to a single `rect` element, Meter creates one or more sections, so my understanding is that each entry on `values` should be able to receive an array of objects as color. I'm still wondering how to include `gradientUnits` and `gradientTransform` properties, as those are props from the gradient itself, and not from the gradient section. Maybe an object containing the gradient props and the array of objects with colors and percentages 🤔 

Right now, I'm leaning towards something like this:

```js
values=
{
  [
    { value: 21,
      color: [
        { percentage: 50, color: "red" }, 
        { percentage: 25, color: "#ababab" }
      ]
    }
  ],
  [
    { value: 40,
      color: [
        { percentage: 33, color: "blue" }, 
        { percentage: 67, color: "#00ddaa" }
      ]
    }
  ]
}
```

Or maybe this, if we consider gradient's props:

```js
values=
{
  [
    { value: 21,
      color: {
        sections: [
          { percentage: 50, color: "red" }, 
          { percentage: 25, color: "#ababab" }
        ],
        transform: "rotate(45)",
        units: "userSpaceOnUse",
      }
    }
  ],
    { value: 40,
      color: {
        sections: [
          { percentage: 50, color: "blue" }, 
          { percentage: 25, color: "#00bbaa" }
        ],
      }
    }
  ],
}
```

Any thoughts on that?

ericsoderberghp:
I think I was leaning towards two flavors:
`<Meter color={[{ color, value }, ...]} value={} />` which works identically to Chart
and
`<Meter values={[{ color: [{ color, value }], value }, ...]} />` where the `value` in the gradient color array are with respect to the overall min/max of the Meter. For example:
```
<Meter
  values={[
    { color: [{ color: 'purple', value: 0}, { color: 'blue', value: 30}], value: 30 }] },
    { color: [{ color: 'red', value: 30}, { color: 'gold', value: 60}], value: 60 }] },
  ]}
/>
```

jcfilben:
Hey @leossantiago just wanted to check in and see if you are still working on this PR or if you had any questions related to it

leossantiago:
Hey @jcfilben. It's been a while 😅 sorry about that.
I actually am still working on it. I did some reworks to it over a month ago but started struggling with gradients behavior on Bar Meters. Turns out SVGs don't like it very much when there's no change on the Y-Axis lol. Either way, I'm working on a proposal to change Bar meters to be based on rect elements. Looking forward to finishing it over the weekend so we can finally discuss it next week as this proposal might represent a breaking change. Hopefully, we'll have a working solution and will be able to FINALLY close this 🎉

jcfilben:
Great sounds good! No rush, just wanted to check in

leossantiago:
Following an offline discussion with @taysea, I will be closing this. The discussion was moved to #5773 as it completely deviates from the initial approach. We can always have this PR as a reference.

PR #5022, opened 2021-03-02
Prevent app from crashing when Distribution value in values is not defined

jcfilben:
#### What does this PR do?
If value in values for Distribution ins't defined and there is more than 1 item in values the app will crash.
This PR changes the behavior so that Distribution won't cause the app to crash when value isn't defined.

More discussion here: https://grommet.slack.com/archives/C09QQEG69/p1614624627045600

Distribution.js

tested with this storybook test:
```javascript
import React from 'react';

import { Grommet, Box, Distribution, Text } from 'grommet';

export const TEST = () => (
  <Grommet theme={grommet} full>
    <Distribution
      basis="xxlarge"
      values={[
            { title: 'Titulek', content: 'Obsah prvního článku', color: 'light-2' },
            { title: 'Titulek', content: 'Obsah druhého článku', color: 'graph-0' },
            { title: 'Titulek', content: 'Obsah třetího článku', color: 'graph-0' },
            { title: 'Titulek', content: 'Obsah čtvrtého článku', color: 'light-2' },
        ]}
    >
      {value => (
          <Box pad="small" background={value.color} size="small" margin="small">
            <Text size="medium">{value.title}</Text>
            <Text size="large">{value.content}</Text>
        </Box>
      )}
    </Distribution>
  </Grommet>
);

export default {
  title: 'Visualizations/Distribution/TEST',
};
```
storybook

No

No

Backwards compatible

### DIFF
--- a/src/js/components/Distribution/Distribution.js
+++ b/src/js/components/Distribution/Distribution.js
@@ -63,7 +63,9 @@ const Distribution = ({
     }
 
     let childBasis;
-    if (subTotal === total) {
+    if (subTotal === 0) {
+      childBasis = ['0px', '0px'];
+    } else if (subTotal === total) {
       childBasis = ['full', '0px'];
     } else if (subTotal > total * 0.7) {
       childBasis = ['3/4', '1/4'];
--- a/src/js/components/Distribution/Distribution.js
+++ b/src/js/components/Distribution/Distribution.js
@@ -33,23 +33,19 @@ const Distribution = ({
     );
   }
   if (values.length > 1) {
-    for (let i = 0; i < values.length; i += 1) {
-      if (!('value' in values[i]))
-        /* eslint-disable-next-line max-len */
-        throw new Error(
-          `Distribution should have set 'value' for each item in 'values'.`,
-        );
-    }
-
     const reducer = (accumulator, { value }) => accumulator + (value || 0);
-    const total = values.reduce(reducer, 0);
+    const total = values
+      .filter(v => Object.prototype.hasOwnProperty.call(v, 'value'))
+      .reduce(reducer, 0);
 
     // figure out how many of the values area needed to represent half of the
     // total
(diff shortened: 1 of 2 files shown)


src/js/components/Distribution/Distribution.js
```diff
@@ -33,6 +33,14 @@ const Distribution = ({
     );
   }
   if (values.length > 1) {
+    for (let i = 0; i < values.length; i += 1) {
```
ericsoderberghp, line 36: Hmm... We don't usually perform this level of property validation. Instead, we typically try to at least not crash the browser by handling the scenario where there is no value property set. For example:
`const total = values.filter(v => v.hasOwnProperty('value')).reduce(reducer, 0);`

jcfilben: Ok I changed the code so instead of throwing and error, it handles when there is no value in values. I am wondering what the default behavior should be for this. Right now if value isn't set in values the first item in values will still be displayed. I am wondering if instead none of the items should be displayed if value isn't set. Also if this happens should there be a warning in the console? Based on the conversation [here](https://grommet.slack.com/archives/C09QQEG69/p1614678325052000?thread_ts=1614624627.045600&cid=C09QQEG69) it may be confusing to users if there isn't something to signify that value needs to be added to values.

ericsoderberghp: > I am wondering if instead none of the items should be displayed if value isn't set.

Great idea.

I'm still reluctant to add a console message when no values are provided. We don't do that for DataTable when no `data` is provided or Meter when no `values` are provided.

jcfilben: Ok makes sense, I didn't add a console message and I changed distribution so that if value is not set then the items won't be displayed

ericsoderberghp (approved):

Issue #5023, opened 2021-03-02
FileInput onChange event not called when files are removed


vincentl:
### Expected Behavior

The new FileInput component (introduced from issue #2535) `onChange` event should be called when files are removed so client code can responded to removal.

### Actual Behavior

`onChange` is only called when files are added.

abnersajr:
Good call @vincentl. I think to be more flexible we can add a callback `onRemove`  to be triggered when you remove a file using the button.

ericsoderberghp:
Agreed. This shouldn't be too hard to create a pull request for. Any takers?

abnersajr:
@ericsoderberghp I'm doing. Will add `onRemove`  and `onRemoveAll`

ericsoderberghp:
I think leveraging `onChange` would be sufficient. Just have it reflect the current set of files. The callee already has to track when multiple files are added sequentially. Removing would mirror in reverse.

abnersajr:
IMO,  Give to the user the possibility to declare a specific callback for `remove` and `removeAll`  would be nice, in case he wants he can simply pass the same function used in the `onChange`  to these callbacks.

One possibility is to add the `onChange`  as default. In case the user doesn't pass callbacks to `onRemove`  and `onRemoveAll` by default we can use the `onChange` if it exists.
Adding this to the documentation to let the user know about it.
@ericsoderberghp what you think?

abnersajr:
Opening the PR and thinking in the backward compatibility, only leveraging the `onChange` function, might lead users to unexpected behaviors.
The addition of a new callback does not impact old users that don't want this behavior.

ericsoderberghp:
Part of my point was we don't have `onAdd`, when adding a subsequent file to a multiple file input. Also, Select that allows multiple options being selected only indicates changes to the selected options via `onChange`. I think it is more consistent to stay with just `onChange`.

abnersajr:
Cool. PR is opened.

jcfilben:
Closing since #5028 was merged.

PR #5028, opened 2021-03-03
[FileInput] Add onChange callback to remove actions

abnersajr:
#### What does this PR do?

Adds `onChange` callback to remove actions

The FileInput component

Manually checking if the `onChange` is being properly executed

Importing the component, pass an `onChange` prop, adds an item, and then remove it. The `onChange` callback should be dispatched on both add and remove actions.

#5023

No.

That would be good.

It depends. Not necessarily breaks, but might lead the user to unexpected behavior.


src/js/components/FileInput/FileInput.js
```diff
@@ -149,6 +149,7 @@ const FileInput = forwardRef(
                 hoverIndicator
                 onClick={event => {
                   event.stopPropagation();
+                  if (onChange) onChange(event);
```
ericsoderberghp, line 152: I think this will send the click event from the button, which isn't going to return the same kind of event that the input will return, i.e. there is no `files` in the event.

ShimiSun:
Hi @abnersajr! Any updates regarding this PR comment?

src/js/components/FileInput/FileInput.js
abnersajr: I added as a second argument the `files` from the hook, however, I'm not sure if it's the best option.
If you have any suggestion for this, let me know.

abnersajr:
> Hi @abnersajr! Any updates regarding this PR comment?

Just pushed a change. I'm looking if need new changes.

src/js/components/FileInput/FileInput.js
ericsoderberghp: I think this is the right strategy, to add an argument, as it parallels how grommet Select onChange works. To make it align even more, I think we should do: `onChange(event, { files })`. That will make it easier to extend in the future.

I also think we should change the onChange delivered from StyledFileInput to include the same second argument, that way the caller can always refer to the files in the same way.

src/js/components/FileInput/FileInput.js
```diff
@@ -171,6 +171,7 @@ const FileInput = forwardRef(
                 hoverIndicator
                 onClick={event => {
                   event.stopPropagation();
+                  if (onChange) onChange(event, files);
```
ericsoderberghp, line 174:
```suggestion
                  if (onChange) onChange(event, { files: [] });
```

src/js/components/FileInput/FileInput.js
```diff
@@ -210,6 +211,7 @@ const FileInput = forwardRef(
                     const nextFiles = [...files];
                     nextFiles.splice(index, 1);
                     setFiles(nextFiles);
+                    if (onChange) onChange(event, files);
```
ericsoderberghp, line 214:
```suggestion
                    if (onChange) onChange(event, { files: nextFiles });
```

ericsoderberghp (commented):
Still looking for the signature to change.

ericsoderberghp (approved):

PR #5042, opened 2021-03-09
Changed react-dom and elliptic dependencies to fix a security issue

ShimiSun:
#### What does this PR do?
Changed react-dom and elliptic dependencies to fix a security issue
https://github.com/grommet/grommet/security/dependabot/yarn.lock/elliptic/open

package.json

jest
review
https://github.com/grommet/grommet/security/dependabot/yarn.lock/elliptic/open
https://github.com/grommet/grommet/security/dependabot/yarn.lock/elliptic/open
no
no
b/c


package.json
```diff
@@ -95,6 +95,7 @@
     "core-js": "^3.8.3",
     "cross-env": "^5.1.6",
     "del": "^5.1.0",
+    "elliptic": "^6.5.4",
```
ericsoderberghp, line 98: Why do we need to add this dependency? Can't we just re-create yarn.lock?

ShimiSun: >Why do we need to add this dependency
It's a  dev-dependency, so not impacting the project in any way.

>Can't we just re-create yarn.lock?
We can, but since this package is resolved from other packages and not directly from grommet's dependencies, it's bound to be erased each time we are rebuilding yarn.lock unless we specify it as a dependency.

ericsoderberghp (approved):

PR #5046, opened 2021-03-09
Added Drop props and theme support for 'margin', 'round' & 'background'

ShimiSun:
#### What does this PR do?
Added Drop props and theme support for 'margin', 'round', 'background' & 'elevation' 
Priority is given to props over theme props.

`elevation` is translated to `shadowSize` on the theme.
`border.radius` was already supported on the theme.
`background` was only supported by the theme.
`margin` was added from scratch for both props & theme.

doc.js
jest + storybook. Added and refined stories.

jest + storybook.
This PR is the groundwork for #4950 

Yes, done.
YES
b/c

### DIFF
--- a/src/js/components/Box/StyledBox.js
+++ b/src/js/components/Box/StyledBox.js
@@ -145,7 +145,7 @@ const wrapStyle = css`
   flex-wrap: ${props => WRAP_MAP[props.wrapProp]};
 `;
 
-const ROUND_MAP = {
+export const ROUND_MAP = {
   full: '100%',
 };
 
--- a/src/js/components/Drop/DropContainer.js
+++ b/src/js/components/Drop/DropContainer.js
@@ -37,6 +37,7 @@ const DropContainer = forwardRef(
   (
     {
       align = defaultAlign,
+      background,
       onAlign,
       children,
       dropTarget,
@@ -314,6 +315,7 @@ const DropContainer = forwardRef(
       <StyledDrop
         ref={ref || dropRef}
         as={Box}
+        background={background}
         plain={plain}
         elevation={
           !plain
@@ -330,8 +332,11 @@ const DropContainer = forwardRef(
       </StyledDrop>
     );
 
-    if (theme.global.drop.background) {
-      const dark = backgroundIsDark(theme.global.drop.background, theme);
+    if (background || theme.global.drop.background) {
+      const dark = backgroundIsDark(
+        background || theme.global.drop.background,
+        theme,
+      );
       if (dark !== undefined && dark !== theme.dark) {
         content = (
           <ThemeContext.Provider value={{ ...theme, dark }}>
--- a/src/js/components/Drop/StyledDrop.js
+++ b/src/js/components/Drop/StyledDrop.js
@@ -49,11 +49,11 @@ const StyledDrop = styled.div`
     )}
 
   ${props =>
-    props.margin &&
+    (props.margin || props.theme.global.drop.margin) &&
     props.theme.global &&
     edgeStyle(
       'margin',
-      props.margin,
+      props.margin || props.theme.global.drop.margin,
       props.responsive,
       props.theme.global.edgeSize.responsiveBreakpoint,
       props.theme,
--- a/src/js/components/Drop/StyledDrop.js
+++ b/src/js/components/Drop/StyledDrop.js
@@ -1,6 +1,7 @@
 import styled, { keyframes } from 'styled-components';
 
-import { baseStyle } from '../../utils';
+import { ROUND_MAP } from '../Box/StyledBox';
+import { baseStyle, edgeStyle } from '../../utils';
 import { backgroundStyle } from '../../utils/background';
 import { defaultProps } from '../../default-props';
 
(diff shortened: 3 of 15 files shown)


src/js/components/Box/StyledBox.js
```diff
@@ -145,7 +145,7 @@ const wrapStyle = css`
   flex-wrap: ${props => WRAP_MAP[props.wrapProp]};
 `;
 
-const ROUND_MAP = {
+export const ROUND_MAP = {
```
ericsoderberghp, line 148: Should we put this in `utils` somewhere?

src/js/components/Drop/DropContainer.js
```diff
@@ -314,6 +315,7 @@ const DropContainer = forwardRef(
       <StyledDrop
         ref={ref || dropRef}
         as={Box}
+        background={background}
```
ericsoderberghp, line 318: On first glance, this seems like it should work similar to `elevation` below. Or, perhaps StyledDrop should take care of checking for `plain` for `elevation`?

src/js/components/Drop/StyledDrop.js
```diff
@@ -1,6 +1,7 @@
 import styled, { keyframes } from 'styled-components';
 
-import { baseStyle } from '../../utils';
+import { ROUND_MAP } from '../Box/StyledBox';
```
ericsoderberghp, line 3: Perhaps `roundStyle` should be added under utils?

src/js/components/Drop/__tests__/Drop-test.js
```diff
@@ -15,6 +15,10 @@ const customTheme = {
   global: {
     drop: {
       shadowSize: 'large',
```
ericsoderberghp, line 17: can we deprecate `shadowSize` and use `elevation` more?

src/js/components/Drop/__tests__/Drop-test.js
```diff
@@ -171,7 +175,7 @@ describe('Drop', () => {
     expectPortal('drop-node').toMatchSnapshot();
   });
 
-  test('props elevation renders', () => {
+  test('elevation prop renders', () => {
```
ericsoderberghp, line 178: since we're renaming this test, how about just calling it "elevation"? The "renders" suffix is redundant. The "prop" term is also redundant.

src/js/components/Drop/doc.js
```diff
@@ -73,11 +81,13 @@ export const doc = Drop => {
       `Elevated height of the target, indicated via a drop shadow. 
       Only applicable if the Drop isn't plain.`,
     ),
+    margin: marginProp,
```
ericsoderberghp, line 84: Can we clean this up to be ordered alphabetically?

src/js/components/Drop/__tests__/Drop-test.js
ShimiSun: You don't think that in some cases we'd want the 'prop' to indicate whether it's a component prop or a theme prop?

ericsoderberghp: I was thinking we always use the term "property" or "prop" to indicate a React property on a component. I would use the term "theme" for something from the theme. I've been trying to keep test names simpler to make it easier to scan and match with -t. DateInput-test.js shows the naming approach I've been working towards. Worth a discussion.

src/js/components/Drop/__tests__/Drop-test.js
ShimiSun: Done.

src/js/components/Drop/DropContainer.js
ShimiSun: `background` is a little tricky since it is being used twice, once in DropContainer.js (line 338), just to determine whether the background is dark or not (regardless of `plain` property), and the second time is on StyledDrop where the actual style is being applied (line 44).
I tried not to reinvent the wheel and maintain backward compatibility of the theme 'background' style that already existed, by keeping the existing behavior and only enhancing it to support the `background` prop. So the `plain` logic on the prop is kept intact.

src/js/components/Box/StyledBox.js
ShimiSun: Done.

src/js/components/Drop/stories/Themed.js
```diff
@@ -7,14 +7,16 @@ import { deepMerge } from 'grommet/utils';
 const customTheme = deepMerge(grommet, {
   global: {
     drop: {
-      background: { dark: 'neutral-2', light: 'neutral-2' },
-      border: { radius: '10px' },
+      background: { dark: 'neutral-2', light: 'background-contrast' },
+      border: { radius: '10px' }, // impacting 'round' behavior
       zIndex: '13',
+      shadowSize: 'large', // impacting the elevation
```
ericsoderberghp, line 13:
```suggestion
      elevation: 'large',
```

ShimiSun: Good catch

ericsoderberghp (approved):

PR #5047, opened 2021-03-10
Changed Tip to use explicit props setting of Drop instead of using plain

### LINKED ISSUE #4950: Tip component ignoring global theme text color

ShimiSun:
#### What does this PR do?
Changed Tip to use explicit props setting of Drop instead of using `plain`

Tip.js
Added Themed story to storybook, and reviewed snapshot changes.

storybook, manual review of snapshots.

Related to #5046 
Fixes #4950 
Not really, I just added a refinement.
Yes.
b/c


src/js/components/Tip/Tip.js
```diff
@@ -63,7 +63,9 @@ const Tip = forwardRef(({ children, content, dropProps, plain }, tipRef) => {
         target={componentRef.current}
         trapFocus={false}
         key="tip-drop"
-        plain
+        background="none"
+        elevation="none"
+        margin="none"
```
ShimiSun, line 68: I can also add `round="0px"` but since it didn't have an impact on the UI, I preferred to keep the CSS clean. Let me know if you think it should be added.

src/js/components/Tip/stories/typescript/Children.tsx
```diff
@@ -2,62 +2,60 @@ import React from 'react';
...
-              pad="small"
-              margin="small"
-            >
-              <Text color="brand">Tooltip</Text>
-            </Box>
-          }
-        >
-          <Box background="brand" pad="small" flex={false}>
-            Box Child
+export const Children = () => (
```
ShimiSun, line 5: The only thing that was changed on this component is the removal of the explicit return statement.

src/js/components/Tip/Tip.js
```diff
@@ -63,7 +63,9 @@ const Tip = forwardRef(({ children, content, dropProps, plain }, tipRef) => {
         target={componentRef.current}
         trapFocus={false}
         key="tip-drop"
-        plain
+        background="none"
```
ericsoderberghp, line 66: Do we even need these? I was expecting `theme.tip.drop` to take care of resetting anything from the themed drop styling.

ShimiSun: That would have been true if we had default `theme.tip.drop` props that are overriding the `global.drop` styles, but we do not. I'm reluctant to add those due to backward compatibility concerns.
So if the Drop isn't `plain`, and the theme are not overriding the base values, Tip will pick up all of its default theme styling of `global.drop` as follows:
```
      drop: {
        background: {
          dark: 'black',
          light: 'white',
        },
        border: {
          radius: '0px',
        },
        shadowSize: 'small', // shadowSize is deprecated, use 'elevation'
        zIndex: '20',
      },
```
Hence, we need the explicit props suggested on this PR to override those default values.
I'm open to suggestions on cleaner ideas.

ericsoderberghp: Would it be essentially equivalent to `theme.tip.drop = { background: 'none', elevation: 'none', margin: 'none' }`?

ShimiSun: Yes, I added backward compatibility tests for the theme to verify it wouldn't break anything. The change had been made.

ericsoderberghp (approved):

Issue #2747, opened 2019-02-12
Would be nice if meter could be turned vertical.


RobertGary1:
It would be helpful if we could support turning the meter vertical as we do with chart.

g4rry420:
Can I work on this issue ?

ShimiSun:
This feature will be available on our next release, thank you @g4rry420 for your contribution!

PR #5052, opened 2021-03-10
Vertical meter

g4rry420:
#### What does this PR do?
Modified Meter so that it can also have vertical direction.
#2747

components => Meter.js, Bar.js, 

None

Added some temporary stories.

N/A

Yes, Already did.

backwards compataible


src/js/components/Meter/Bar.js
```diff
@@ -8,55 +8,169 @@ import { StyledMeter } from './StyledMeter';
...
-        color || defaultColor(index, theme, values ? values.length : 0);
-      let hoverProps;
-      if (onHover) {
-        hoverProps = {
-          onMouseOver: () => onHover(true),
-          onMouseLeave: () => onHover(false),
-        };
+
+  let paths;
+  if (direction === 'horizontal') {
```
ericsoderberghp, line 47: There's a fair amount of logic duplication with this coarse grained if/else approach. I'm wondering if we can merge these more, perhaps using a term like `length` for the calculations, and then just using the if/else on `direction` to construct the path.

src/js/components/Meter/Bar.js
```diff
@@ -8,55 +8,169 @@ import { StyledMeter } from './StyledMeter';
...
 
 const Bar = forwardRef((props, ref) => {
-  const { background, max, round, size, thickness, values, ...rest } = props;
+  const {
+    background,
+    max,
+    round,
+    size,
+    thickness,
+    direction,
```
ericsoderberghp, line 17: We try to keep the props alphabetized.

src/js/components/Meter/stories/newStory.js
```diff
@@ -0,0 +1,63 @@
...
+            </Box>
+            <Text>{label || 'total'}</Text>
+          </Box>
+        </Stack>
+      </Box>
+    </Grommet>
+  );
+};
+
+MultipleValues.storyName = 'New Story';
```
ericsoderberghp, line 59: Should this be called something like "Vertical Bar Multiple" rather than "New Story"?

src/js/components/Meter/__tests__/__snapshots__/Meter-test.js.snap
```diff
@@ -287,58 +287,58 @@ exports[`Meter many values 1`] = `
       strokeWidth={24}
     />
     <path
-      d="M 336,12 L 384,12"
```
ericsoderberghp, line 290: I wouldn't expect any snapshot changes to existing tests for this.

src/js/components/Meter/doc.js
```diff
@@ -57,6 +57,10 @@ export const doc = Meter => {
     type: PropTypes.oneOf(['bar', 'circle'])
       .description('The visual type of meter.')
       .defaultValue('bar'),
+    direction: PropTypes.oneOf(['horizontal', 'vertical'])
+      .description(`The direction of the Meter you want to display.
+       The default direction is horizontal.But, direction works only
```
ericsoderberghp, line 62:
```suggestion
       The default direction is horizontal. Currently, direction works only
```

ericsoderberghp (commented):
Massive apologies for not getting to review this sooner. I really like the functionality.

I think we also will need some new jest tests for `direction`.

src/js/components/Meter/Bar.js
g4rry420: I am sorry, I didn't understand it. Can you please elaborate it bit further ?

ericsoderberghp: For example, in this set of changes, lines 63-89 are identical to lines 136-162. I think we should restructure the code to avoid that duplication. I am also wondering if we can do the calculations consistently and then only when constructing the actual path coordinates switch on horizontal vs. vertical. For example:

```
const d = (direction === 'horizontal')
  ? `M ${start},${mid} L ${start + delta},${mid}`
  : `M ${mid},${end - start} L ${mid},${(end - start) + delta}`
;
```
Note that this above code draws the vertical bar elements from the bottom, rather than handling that aspect in how the value of `start` is managed.

g4rry420: I have refactored the codebase for the component and also added the prop in typescript `index.d.ts` file. Let me know if any other changes required.

src/js/components/Meter/__tests__/__snapshots__/Meter-test.js.snap
g4rry420: I am not sure why this changed.

ericsoderberghp (commented):
Good progress on reducing code duplication. I have one more key comment here.

src/js/components/Meter/Bar.js
```diff
@@ -18,17 +27,51 @@ const Bar = forwardRef((props, ref) => {
...
-  const someHighlight = (values || []).some(v => v.highlight);
 
+  const someHighlight = (values || []).some(v => v.highlight);
   let start = capOffset;
-  const paths = (values || [])
-    .filter(v => v.value > 0)
-    .map((valueArg, index) => {
-      const { color, highlight, label, onHover, value, ...pathRest } = valueArg;
 
+  // suppose if the user only passes the
```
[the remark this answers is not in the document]
g4rry420: Sorry, if I am misunderstood, how do you wanna change setting of `width` of StyledMeter as it will create the nested ternary expression?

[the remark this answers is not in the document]
ericsoderberghp: I am proposing we do all of the calculations based on `length` and `thickness`, where length is the old horizontal `width` and thickness is the old horizontal `height`. So, we only care about `width` at the very end, when building `viewBox` and the `width` and `height` of the StyledMeter.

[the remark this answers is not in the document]
g4rry420: Got it, I just pushed the code, let me know about it.

src/js/components/Meter/Bar.js
```diff
@@ -49,30 +67,49 @@ const Bar = forwardRef((props, ref) => {
...
           )}
-          strokeWidth={height}
+          strokeWidth={direction === 'horizontal' ? thickness : length}
           strokeLinecap={round ? 'round' : 'butt'}
           {...hoverProps}
           {...pathRest}
         />
       );
-    })
-    .reverse(); // reverse so the caps looks right
```
ericsoderberghp, line 59: I think we still need this `reverse()` to ensure rounded caps look right. I think this is the source of the unexpected snapshot changes.

src/js/components/Meter/index.d.ts
```diff
@@ -12,6 +12,7 @@ export interface MeterProps {
   alignSelf?: AlignSelfType;
   background?: string | { color?: string; opacity?: OpacityType };
   color?: string;
+  direction?: string;
```
ericsoderberghp, line 15:
```suggestion
  direction?: 'horizontal' | 'vertical';
```

ericsoderberghp (commented):
Very close. A couple of small items and I think it will be good to go.

g4rry420:
@ericsoderberghp Thank you for all your help for making this pr successful.

Issue #3551, opened 2019-11-15
Enhance Box `direction`


ShimiSun:
This request is coming from slack.
>Q: Is there a way to make a Box "row-responsive", but reverse? I have 2 Boxes that I want to be horizontal if space allows, but if not, I want the right Box to be on top.

Currently, we don't have a straight forward way to provide a solution (besides manipulating the layout with ResponsiveContext). 
A nice enhancement that will allow this functionality in a more straight forward way will be to extend the `direction` prop:
`direction={{ direction: 'row', responsive: true, reverse: true }}`

https://grommet.slack.com/archives/C04LMHN59/p1573747488463900

oorestisime:
That looks to be abreaking change. How about adding a just a reverse prop? that defaults to false?

ShimiSun:
Why do you think it's a breaking change? 'direction' prop will still accept a string along with the suggested object

oorestisime:
yeah if we add both types then it wouldn't be a breaking change.
but then we would have to support two code path for the same thing (direction with responsive) and keep the string limited to direction-responsive since we wouldn't be adding row-responsive-reverse i suppose. 
no strong opinion on this :D  just felt it might be smoother to implement

PR #5060, opened 2021-03-14
Added support for the box row-reverse with column reverse.

AkshayChoulwar:
Added support for the box row-reverse with column reverse.

Currently, we don't have a straight forward way to provide a solution.
A nice enhancement that will allow this functionality in a more straight forward way will be to extend the direction prop:
direction={{ direction: 'row', responsive: true, reverse: true }}

#### What does this PR do?
We can use column reverse functionality for the Box.

- src/js/components/Box/StyledBox.js
- src/js/components/Box/stories/RowResponsiveBox.js
- src/js/utils/index.d.ts

Added RowResponsiveBox story and tested with npm run storybook.

Use Box with relevant props.

https://github.com/grommet/grommet/issues/3551

NA

Before Reducing Viewport Width: 
image

After Reducing Viewport Width:
image

Yes

Yes

This PR is backward compatible.

### DIFF
--- a/src/js/components/Box/StyledBox.js
+++ b/src/js/components/Box/StyledBox.js
@@ -68,20 +68,14 @@ const basisStyle = css`
 // we assume we are in the context of a Box going the other direction
 // TODO: revisit this
 const directionStyle = (directionWrapper, theme) => {
-  let direction;
-  let isResponsive;
-  let isReverse;
+  let { direction } = directionWrapper;
+  const { responsive, reverse } = directionWrapper;
 
   if (
     typeof directionWrapper === 'string' ||
     directionWrapper instanceof String
   ) {
     direction = directionWrapper;
-  } else {
-    // DirectionWrapper is object now.
-    direction = directionWrapper.direction;
-    isResponsive = directionWrapper.responsive;
-    isReverse = directionWrapper.reverse;
   }
 
   const styles = [
@@ -92,43 +86,46 @@ const directionStyle = (directionWrapper, theme) => {
     `,
   ];
 
-  const breakpoint = getBreakpointStyle(theme, theme.box.responsiveBreakpoint);
-
-  if (
-    direction === 'row-responsive' &&
-    theme.box.responsiveBreakpoint &&
-    breakpoint
-  ) {
-    styles.push(
-      breakpointStyle(
-        breakpoint,
-        `
-        flex-direction: column;
-        flex-basis: auto;
-        justify-content: flex-start;
-        align-items: stretch;
-      `,
-      ),
+  if (direction === 'row-responsive' && theme.box.responsiveBreakpoint) {
+    const breakpoint = getBreakpointStyle(
+      theme,
+      theme.box.responsiveBreakpoint,
     );
+
+    if (breakpoint) {
+      styles.push(
+        breakpointStyle(
+          breakpoint,
+          `
+          flex-direction: column;
+          flex-basis: auto;
+          justify-content: flex-start;
+          align-items: stretch;
+        `,
+        ),
+      );
+    }
   }
 
-  if (
-    direction === 'row' &&
-    isResponsive === true &&
-    isReverse === true &&
-    breakpoint
-  ) {
-    styles.push(
-      breakpointStyle(
-        breakpoint,
-        `
-         flex-direction: column-reverse;
-         flex-basis: auto;
-         justify-content: flex-start;
-         align-items: stretch;
-      `,
-      ),
+  if (direction === 'row' && responsive && reverse) {
+    const breakpoint = getBreakpointStyle(
+      theme,
+      theme.box.responsiveBreakpoint,
     );
+
+    if (breakpoint) {
+      styles.push(
+        breakpointStyle(
+          breakpoint,
+          `
+           flex-direction: column-reverse;
+           flex-basis: auto;
+           justify-content: flex-start;
+           align-items: stretch;
+        `,
+        ),
+      );
+    }
   }
   return styles;
 };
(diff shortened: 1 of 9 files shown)


src/js/components/Box/StyledBox.js
```diff
@@ -457,7 +493,10 @@ const gapStyle = (directionProp, gap, responsive, border, theme) => {
     if (responsiveMetric) {
       if (directionProp === 'row' || directionProp === 'row-reverse') {
         styles.push(breakpointStyle(breakpoint, `width: ${responsiveMetric};`));
-      } else if (directionProp === 'row-responsive') {
+      } else if (
+        directionProp === 'row-responsive' ||
+        (directionProp.direction === 'row' && directionProp.responsive === true)
```
ShimiSun, line 498:
```suggestion
        (directionProp.direction === 'row' && directionProp.responsive)
```

src/js/components/Box/StyledBox.js
```diff
@@ -539,7 +578,11 @@ const gapStyle = (directionProp, gap, responsive, border, theme) => {
               }`,
             ),
           );
-        } else if (directionProp === 'row-responsive') {
+        } else if (
+          directionProp === 'row-responsive' ||
+          (directionProp.direction === 'row' &&
+            directionProp.responsive === true)
```
ShimiSun, line 584:
```suggestion
            directionProp.responsive)
```

src/js/components/Box/StyledBox.js
```diff
@@ -67,32 +67,68 @@ const basisStyle = css`
...
-      theme,
-      theme.box.responsiveBreakpoint,
-    );
-    if (breakpoint) {
-      styles.push(
-        breakpointStyle(
-          breakpoint,
-          `
+
+  const breakpoint = getBreakpointStyle(theme, theme.box.responsiveBreakpoint);
```
ShimiSun, line 95: why is this line being extracted outside of the 'if' (where it was placed before)?

src/js/components/Box/StyledBox.js
```diff
@@ -67,32 +67,68 @@ const basisStyle = css`
...
-      );
-    }
+      ),
+    );
+  }
+
+  if (
+    direction === 'row' &&
+    isResponsive === true &&
+    isReverse === true &&
```
ShimiSun, line 118:
```suggestion
    isResponsive &&
    isReverse &&
```

src/js/components/Box/StyledBox.js
```diff
@@ -67,32 +67,68 @@ const basisStyle = css`
...
+  if (
+    typeof directionWrapper === 'string' ||
+    directionWrapper instanceof String
+  ) {
+    direction = directionWrapper;
+  } else {
+    // DirectionWrapper is object now.
+    direction = directionWrapper.direction;
+    isResponsive = directionWrapper.responsive;
+    isReverse = directionWrapper.reverse;
```
ShimiSun, line 84: Could those be deconstructed more nicely with ES6? Something like...
```
const { direction, responsive: isResponsive .... } = directionWrapper;
```

src/js/components/Box/StyledBox.js
```diff
@@ -67,32 +67,68 @@ const basisStyle = css`
 // https://stackoverflow.com/questions/36247140/why-doesnt-flex-item-shrink-past-content-size
 // we assume we are in the context of a Box going the other direction
 // TODO: revisit this
-const directionStyle = (direction, theme) => {
+const directionStyle = (directionWrapper, theme) => {
+  let direction;
+  let isResponsive;
+  let isReverse;
```
ShimiSun, line 73: What would be the expected behavior in case the user set the `direction` but hadn't set these two values?

ShimiSun (commented):
Thank you for working on this enhancement! 
I dropped small comments, let me know what you think.

A one-piece that is missing from this PR is the doc.js changes. After that run `yarn test-update` to create new snapshots for the README files and docs, and push those files up the branch.

src/js/components/Box/StyledBox.js
AkshayChoulwar: If user passes only direction like below:
```

<Box direction={
  direction: "row" // or column
}/>
```

then it will be non responsive like `<Box direction="row">` or `<Box direction="column">`

AkshayChoulwar: Added story for above scenario.

src/js/components/Box/StyledBox.js
AkshayChoulwar: Done

src/js/components/Box/StyledBox.js
AkshayChoulwar: It was placed inside the if statement. Now we have two if's so don't wanted to duplicate it. Now added back to the if statement.

src/js/components/Box/doc.js
```diff
@@ -127,12 +127,19 @@ export const doc = Box => {
...
+        'row-reverse',
+        'column-reverse',
+      ]),
+      PropTypes.shape({
+        direction: PropTypes.string,
+        responsive: PropTypes.bool,
+        reverse: PropTypes.bool,
+      }),
     ])
       .description('The orientation to layout the child components in.')
```
ericsoderberghp, line 144: I think we need more description here to explain what the values mean.

src/js/components/Box/doc.js
```diff
@@ -127,12 +127,19 @@ export const doc = Box => {
...
+        'row',
+        'column',
+        'row-responsive',
+        'row-reverse',
+        'column-reverse',
+      ]),
+      PropTypes.shape({
+        direction: PropTypes.string,
+        responsive: PropTypes.bool,
+        reverse: PropTypes.bool,
```
ericsoderberghp, line 141: If I'm understanding the intent, that `reverse` only applies when `responsive` is true, I'm wondering if we should structure this object a bit differently to be clearer. For example:
`direction={{ direction: 'row', responsive: { direction: 'column-reverse } }}`

jcfilben:

PR #5061, opened 2021-03-14
fix(StyledAnchor): prevent custom props from bleeding into DOM

mihai-ro:
#### What does this PR do?
Uses styled-components' util called `withConfig` to prevent custom props such as `as`, `colorProp`, `hasIcon`, and `hasLabel` from bleeding into the DOM tree and throwing a bunch of React warnings in the console.

Start from here -  https://codesandbox.io/s/charming-lumiere-7ruv2?file=/src/App.js to see the example.

Manual

Replicate the example from the sandbox.

This PR is a follow-up to the short conversation we had with @halocline on Slack.

None, as far as i can tell. If needed for documentation purposes, I can open one so it can be backtracked at a later point in time.

No

No

Not a breaking change

### DIFF
--- a/src/js/components/Anchor/StyledAnchor.js
+++ b/src/js/components/Anchor/StyledAnchor.js
@@ -28,8 +28,9 @@ const StyledAnchor = styled.a.withConfig({
   // prevent custom props from bleeding into DOM
   // https://styled-components.com/docs/api#shouldforwardprop
   shouldForwardProp: (prop, defaultValidatorFn) =>
-    !['as', 'colorProp', 'hasIcon', 'hasLabel'].includes(prop) &&
-    defaultValidatorFn(prop),
+    !['as', 'colorProp', 'focus', 'hasIcon', 'hasLabel', 'reverse'].includes(
+      prop,
+    ) && defaultValidatorFn(prop),
 })`
   box-sizing: border-box;
   ${props => sizeStyle(props)}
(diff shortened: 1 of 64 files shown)


src/js/components/Anchor/StyledAnchor.js
```diff
@@ -24,7 +24,13 @@ const sizeStyle = props => {
...
 };
 
-const StyledAnchor = styled.a`
+const StyledAnchor = styled.a.withConfig({
+  // prevent custom props from bleeding into DOM
+  // https://styled-components.com/docs/api#shouldforwardprop
+  shouldForwardProp: (prop, defaultValidatorFn) =>
+    !['as', 'colorProp', 'hasIcon', 'hasLabel'].includes(prop) &&
+    defaultValidatorFn(prop),
+})`
```
ShimiSun, line 33: I wonder how come we are not seeing those props leaking on the tests/snapshots.

halocline: I think the issue has to do when the components are being composed together and they share prop names. In this case, @mihai-ro was [trying to use Anchor + React-router Link](https://grommet.slack.com/archives/C09QQEG69/p1615558841146600).

We have also run into this with [StyledText](https://github.com/grommet/grommet/issues/4003#issuecomment-656820442) and [StyledButtonKind](https://github.com/grommet/grommet/blob/c65701c11c791169e5be90a330e41d5ad9f61f3e/src/js/components/Button/StyledButtonKind.js#L214).

Documentation: https://styled-components.com/docs/api#shouldforwardprop

PR #5062, opened 2021-03-15
Type object added on row property for Grid

andrewcovenant:
#### What does this PR do?
Adds an additional type on the row property.
By comparing the Grid Documentation (scroll down to row and check the recommended types) vs the Grid/stories/Responsive.js and checking the "const rows" what is passed as a value.
The default ones
N/A
We have built a custom Responsive Grid component on top of the Grommet one but using TypeScript when we are passing an object on the row property, the compiler will give us "Object not assignable to type...etc". Because the one in Grommet is built in Javascript, it won't check for the type and the tests are not failing.
Need an extra type on the row property to avoid TypeScript error when building a reusable component and passing an object on the Grid row property.
[Screenshot](https://ibb.co/McYGH5y)
Yes
Yes
Backward compatible

### DIFF
--- a/src/js/components/Grid/index.d.ts
+++ b/src/js/components/Grid/index.d.ts
@@ -75,6 +75,26 @@ export interface GridProps {
   pad?: PadType;
   responsive?: boolean;
   rows?:
+    | {
+        count?: 'fit' | 'fill' | number;
+        size?:
+          | 'xsmall'
+          | 'small'
+          | 'medium'
+          | 'large'
+          | 'xlarge'
+          | 'full'
+          | '1/2'
+          | '1/3'
+          | '2/3'
+          | '1/4'
+          | '2/4'
+          | '3/4'
+          | 'flex'
+          | 'auto'
+          | string
+          | string[];
+      }
     | (
         | 'xsmall'
         | 'small'
--- a/src/js/components/TextInput/index.d.ts
+++ b/src/js/components/TextInput/index.d.ts
@@ -44,6 +44,7 @@ export interface TextInputProps
   suggestions?: ({ label?: React.ReactNode; value?: any } | string)[];
   textAlign?: TextAlignType;
   value?: string | number;
+  ref?: React.Ref<HTMLInputElement>;
 }
 
 declare const TextInput: React.FC<TextInputProps>;


src/js/components/Grid/index.d.ts
```diff
@@ -75,6 +75,26 @@ export interface GridProps {
   pad?: PadType;
   responsive?: boolean;
   rows?:
+    | {
+        count?: 'fit' | 'fill' | number;
```
ShimiSun, line 79:
```suggestion
        count: 'fit' | 'fill' | number;
```
I think we have to have the `count` prop defined for the object type, let me know what you think.

andrewcovenant: Hmmm, I get what you mean, but I think this will add extra complexity when using or re-using a component.
Either way, we will have something like this (see code below) when creating the row.
Adding extra keys will make the code maintenance/debugging harder and we will add a mandatory track to the rows based on the size.
` const rows = {
    small: ["xsmall", "xsmall", "xsmall", "xsmall"],
    medium: ["xsmall", "xsmall","xsmall"],
    large: ["xsmall", "xsmall","xsmall"]
  };`

ShimiSun: @abnersajr & @alexmirrington  I'd love to hear your thoughts here.

alexmirrington: To start, the `const rows = { small: ['xsmall', 'xsmall', 'xsmall'], medium: ['xsmall', 'xsmall'], large: ['xsmall'], xlarge: ['xsmall']};` in the responsive example storybook is not actually passed directly to the `Grid` component. The `ResponsiveContext` context is used to provide a key for the current responsive size _e.g._ `"small"`, which is used to lookup the corresponding array in the `rows` object _e.g._  `['xsmall', 'xsmall', 'xsmall']`. It is this array that is passed to the `Grid` component. TL;DR: no objects are ever passed to the `rows` prop of the `Grid` component.

Delving a bit deeper and looking into what is actually supported by the Grid component, the following function in `StyledGrid.js` handles `rows` props. I'll paste the function here for reference:

```Python
const rowsStyle = props => {
  if (Array.isArray(props.rowsProp)) {
    return css`
      grid-template-rows: ${props.rowsProp
        .map(s => {
          if (Array.isArray(s)) {
            return `minmax(${sizeFor(s[0], props, true)}, ${sizeFor(
              s[1],
              props,
              true,
            )})`;
          }
          return sizeFor(s, props, true);
        })
        .join(' ')};
    `;
  }
  return css`
    grid-auto-rows: ${props.theme.global.size[props.rowsProp]};
  `;
};
```

Looking at this function, the `rows` prop is either:

- A key for `props.theme.global.size`, _i.e._ `"xxsmall" | "xsmall" | "small" | "medium" | "large" | "xlarge" | "xxlarge" | string`(looking at `src/js/themes/base.d.ts`)
- An array containing any mixture of:
  1. the above keys
  2. any css size (incl. "auto")
  3. `"full" | "flex" | "1/2" | "1/3" | "2/3" | "1/4" | "2/4" | "3/4"
  4. a two-element array containing the minimum and maximum size for the row, where the minum and maximum sizes may be one of 1, 2 or 3.

This raises two main issues:

1. The `rows` prop should **NOT** accept objects with `count` and `size` keys, these are only for `columns` props.
2. The documentation does not explicitly contain `xxsmall` or `xxlarge` as valid theme keys, although they are explicitly defined as keys of `props.theme.global.size`. This needs to be updated.
3. "any CSS size" as mentioned in the docs is actually only valid inside an array. If something like `"4em"` is given to the `rows` prop as a single string, that will be used as a lookup key in `props.theme.global.size`. Technically any string is valid here, but I think the docs could be clearer.

Please let me know if I missed anything regarding the `count` and `size` keys.

andrewcovenant: I agree with you on the count, not sure on the size. But either way, the row prop on the Grid needs to understand an "object", else when re-using the component the TypeScript will scream that rows do not expect an object

andrewcovenant: @ShimiSun What do you think?

andrewcovenant: @ShimiSun any updates on this?

ShimiSun: @andrewthamcc , sorry for the late response here. 
I think I agree we should allow the count to be optional as you suggested. Feel free to reopen the PR for another review. 
again, sorry for the long overdue.

PR #5077, opened 2021-03-17
Update DataTable Checkbox styling, allow plain column render

### LINKED ISSUE #5000: DataTable: Size of column with select boxes in it
### What should happen

Checkbox column width should either be customizable by the consumer or the width of the checkbox

### What does happen

The width is set automatically and Can lead to extra white space as shown below

https://codesandbox.io/s/grommet-v2-template-forked-vp81p?file=/index.js

### Your Environment

- Grommet version: 2.16.1
- Browser Name and version: Chrome 88

taysea:
#### What does this PR do?
This PR improves the way DataTable renders the content of its cells (specifically CheckBoxes from using `onSelect` or an enhancement where column data can be rendered `plain`, meaning cell padding won't apply).

In order to maintain backwards compatibility, we introduce `plain="noPad"` which means that cell contents will fill the entirety of the cell (no pad applied). This is leverage internally on the CheckBoxes that appear when a DataTable has onSelect, but the caller can also specify `plain` on a column to flag that the data in that column should render with no cell padding.

In addition, a prop of `pad` is added onto CheckBox that applies pad to the container surrounding the CheckBox and its label. In the context of a DataTable, the DataTable will pass the intended cell padding onto the CheckBox directly. This allows for hover backgrounds of CheckBoxes (if present in a theme) to fill an entire cell.

Lastly, `size="auto"` is applied to the CheckBoxes in DataTable which means that the width will be drive by the size of the content (in this case the CheckBox). The auto-width is driven by internally setting the column width to 1px which will allow the checkbox dimensions to drive the width. References for the CSS pattern:
https://stackoverflow.com/questions/4757844/css-table-column-autowidth?noredirect=1&lq=1
https://www.w3schools.com/cssref/pr_tab_table-layout.asp

src/js/components/DataTable/Body.js

Tested in storybook:

Screen Shot 2021-03-24 at 3 33 40 PM

3. `plain: 'noPad'` in column definition:
4. https://user-images.githubusercontent.com/12522275/111552937-9a38c300-8740-11eb-9727-2cb5742271e2.mov

```
...
{
    property: 'location',
    header: 'Location',
    render: datum => datum.location ? (
        <Box pad={{horizontal: 'small', vertical: 'xsmall'}} onClick={() => {}} hoverIndicator fill>
          <Text truncate weight="bold">
            {datum.location}
          </Text>
        </Box>
    ) : undefined,
    plain: 'noPad',
  },
...
```

Added jest test.

See above.

Closes #5000
Closes https://github.com/grommet/grommet-theme-hpe/issues/170
Yes. Updated here.

Enhanced DataTable columns to allow for `plain` render of column data.
Enhanced CheckBox theming to allow `checkBox.pad` and added `pad` prop.
Improved DataTable rendering of CheckBox with `onSelect` to accommodate pad on CheckBox.

@ericsoderberghp this will shrink the width of the checkbox column (as opposed to before it could grow to receive equal width with other columns).


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "grommet",
-  "version": "2.17.0",
+  "version": "2.16.3",
```
ShimiSun: Please clean up your branch from merge conflicts, I'll hold off the review until you'll take a look at it.

ShimiSun (changes requested):

package.json
taysea: Thanks. will do.

taysea: All merge conflicts have been resolved @ShimiSun

src/js/components/CheckBox/doc.js
```diff
@@ -42,6 +43,14 @@ export const doc = CheckBox => {
...
       via event.target.checked. Same as React <input onChange={} />.`,
     ),
+    pad: padPropType
+      .description(
+        `The amount of padding on the container 
+    surrounding the checkbox and its label. An object can be specified to 
+    distinguish horizontal padding, vertical padding, and padding on a 
+    particular side of the container.`,
+      )
+      .defaultValue('none'),
```
ShimiSun, line 53: I don't think we are setting the default to 'none' anywhere.

src/js/components/DataTable/Header.js
```diff
@@ -139,7 +139,11 @@ const Header = forwardRef(
           )}
 
           {(selected || onSelect) && (
-            <TableCell background={background || cellProps.background}>
+            <TableCell
+              background={background || cellProps.background}
+              plain="fill"
```
ShimiSun, line 144: I find `plain="fill"`to be counter intuative.

src/js/components/CheckBox/doc.js
taysea: You're right, I can update to undefined.

src/js/components/DataTable/Header.js
taysea: Valid feedback. The reason we cannot just say `plain={true}` is because `plain` as a boolean is already a documented property on TableCell. Unfortunately, its behavior does not align with general grommet stylings, where assigning `plain` would remove padding (in the TableCell case, padding is still applied when caller applies `plain`). For backwards compatibility, we needed to introduce either a new prop or a new value on `plain` that would indicate for the TableCell case that we want it to be trulyyy plain (no pad, the content should take up the entirety of the cell). 

Eric and I were thinking `fill` because in other areas of Grommet, `fill` means to take up all available space. Any other words that come to mind that mean a similar thing?

src/js/components/CheckBox/doc.js
taysea: Done.

src/js/components/DataTable/Header.js
taysea: Have updated to `noPad` to be more intuitive.

src/js/components/TableCell/TableCell.js
```diff
@@ -33,66 +34,106 @@ const TableCell = forwardRef(
...
-          delete mergedProps.verticalAlign;
+        }
+      };
+
+      window.addEventListener('resize', updateHeight);
+      updateHeight();
+      return () => {
+        window.removeEventListener('resize', updateHeight);
+      };
+    });
```
ericsoderberghp, line 71: Should this effect be rebuilt it the value for `plain` changes?

src/js/components/DataTable/doc.js
```diff
@@ -102,6 +102,7 @@ export const doc = DataTable => {
           }),
         ]),
         pin: PropTypes.bool,
+        plain: PropTypes.oneOf(['noPad']),
```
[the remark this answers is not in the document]
taysea: Yes, I think that should work.

src/js/components/DataTable/index.d.ts
```diff
@@ -42,6 +42,7 @@ export interface ColumnConfig<TRowType> {
   footer?: React.ReactNode | { aggregate?: boolean };
   header?: string | React.ReactNode | { aggregate?: boolean };
   pin?: boolean;
+  plain?: boolean | 'noPad';
```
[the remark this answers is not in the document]
taysea: Will update to just be `boolean`.

src/js/components/TableCell/TableCell.js
taysea: Thanks yes I was missing the dependency array.

src/js/components/Table/StyledTable.js
```diff
@@ -19,9 +19,17 @@ const SIZE_MAP = {
 
 const sizeStyle = css`
   width: ${props =>
-    SIZE_MAP[props.size] || props.theme.global.size[props.size] || props.size};
+    props.size === 'auto'
+      ? '1px' // table cell will respect content width
```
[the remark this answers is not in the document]
taysea: Updating the comment for clarity but also adding resource links to PR description in case needed for reference

ericsoderberghp (approved):

src/js/components/TableCell/doc.js
```diff
@@ -10,8 +10,12 @@ export const doc = TableCell => {
     .intrinsicElement('td');
 
   DocumentedTableCell.propTypes = {
-    plain: PropTypes.bool
-      .description('Whether default styling context should be removed.')
+    plain: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf(['noPad'])])
```
ShimiSun, line 13: We should not expose `noPad`

src/js/components/TableCell/index.d.ts
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { BoxTypes } from '../Box';
 
 export interface TableCellProps {
-  plain?: boolean;
+  plain?: boolean | 'noPad';
```
ShimiSun, line 5:
```suggestion
  plain?: boolean;
```
We should not expose `noPad`

src/js/components/TableCell/doc.js
taysea: We need to leave here for typing to pass, but since it's internal use only we can remove it from site documentation.

ShimiSun (approved):

ShimiSun:
Thank you @taysea 💟

PR #5081, opened 2021-03-19
[TS] Add Avatar, CheckBoxGroup, Chart and Clock extended props interfaces

alexmirrington:
#### What does this PR do?

This PR adds and exports `<component>ExtendedProps` declarations for the Avatar, CheckBoxGroup, Chart and Clock components, forming part of issue #4998.

The only files changed are the TS declaration files for each of the four components. The `Clock` component required some discussion, so should be reviewed last.

Importing the extended prop interfaces from a separate project should allow `JSX.IntrinsicElements` keys to be passed through to the imported interface, e.g. importing `AvatarExtendedProps` should make `div` keys available, as well as general HTML keys like `children`. Also double check any keys included in `Omit<..., ...>` types for consistency with original intersected types.

Progress tracking: #4998
Design/discussion: #4978

No

Backwards compatible.

### DIFF
--- a/src/js/components/Clock/index.d.ts
+++ b/src/js/components/Clock/index.d.ts
@@ -28,7 +28,17 @@ export interface ClockProps {
   type?: 'analog' | 'digital';
 }
 
-declare const Clock: React.ComponentClass<ClockProps &
-  (JSX.IntrinsicElements['div'] | JSX.IntrinsicElements['svg'])>;
+/**
+ * Ideally this would be an interface, however since the Clock component can be
+ * either analog (svg) or digital (div), we cannot know at compile time whether
+ * ClockExtendedProps should contain svg or div props.
+ */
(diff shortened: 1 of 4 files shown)


src/js/components/Clock/index.d.ts
```diff
@@ -28,7 +28,17 @@ export interface ClockProps {
   type?: 'analog' | 'digital';
 }
 
-declare const Clock: React.ComponentClass<ClockProps &
-  (JSX.IntrinsicElements['div'] | JSX.IntrinsicElements['svg'])>;
+/**
+ * Ideally this would be an interface, however since the Clock component can be
+ * either analog (svg) or digital (div), we cannot know at compile time whether
+ * ClockExtendedProps should contain svg or div props.
+ */
```
ShimiSun, line 35: Great comment!

ShimiSun (approved):

PR #5084, opened 2021-03-19
[TS] Add Spinner and MaskedInput extended props interfaces

alexmirrington:
#### What does this PR do?

This PR adds and exports `<component>ExtendedProps` declarations for the Spinner and MaskedInput components, forming part of issue #4998

The only files changed are the TS declaration files for each of the two mentioned components. The spinner component is more straightforward, since it is more similar to the majority of the components changed as part of issue #4998.

The `MaskedInput` component previously defined a `MaskedInputType`, which has been kept with the addition of the `MaskedInputExtendedProps` interface for backwards compatibility. The reviewer should double check that the changes are backwards compatible.

Importing the extended prop interfaces from a separate project should allow `JSX.IntrinsicElements` keys to be passed through to the imported interface, e.g. importing `LayerExtendedProps` should make `div` keys available, as well as general HTML keys like `children`. Also double check any keys included in `Omit<..., ...>` types for consistency with original intersected types.

Progress tracking: #4998
Design/discussion: #4978

No

Backwards compatible.

### DIFF
--- a/src/js/components/MaskedInput/index.d.ts
+++ b/src/js/components/MaskedInput/index.d.ts
@@ -21,9 +21,13 @@ export interface MaskedInputProps {
   value?: string | number;
 }
 
-declare const MaskedInput: React.ComponentClass<MaskedInputProps &
-  Omit<JSX.IntrinsicElements['input'], keyof MaskedInputProps>>;
-export type MaskedInputType = MaskedInputProps &
-  Omit<JSX.IntrinsicElements['input'], keyof MaskedInputProps>;
+export interface MaskedInputExtendedProps
+  extends MaskedInputProps,
+    Omit<JSX.IntrinsicElements['input'], keyof MaskedInputProps> {}
+
+// Keep type alias for backwards compatibility.
+export type MaskedInputType = MaskedInputExtendedProps;
+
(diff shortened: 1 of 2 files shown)


src/js/components/MaskedInput/index.d.ts
```diff
@@ -21,9 +21,13 @@ export interface MaskedInputProps {
   value?: string | number;
 }
 
-declare const MaskedInput: React.ComponentClass<MaskedInputProps &
-  Omit<JSX.IntrinsicElements['input'], keyof MaskedInputProps>>;
-export type MaskedInputType = MaskedInputProps &
-  Omit<JSX.IntrinsicElements['input'], keyof MaskedInputProps>;
+export interface MaskedInputExtendedProps
+  extends MaskedInputProps,
+    Omit<JSX.IntrinsicElements['input'], keyof MaskedInputProps> {}
```
ShimiSun, line 26: Why is `keyof` being used in this refactoring?

alexmirrington: Please refer to #5085, the reasining is the same.

ShimiSun (approved):

PR #5085, opened 2021-03-19
[TS] Add Box, Button, Calendar and Drop extended props interfaces

alexmirrington:
#### What does this PR do?

This PR adds and exports `<component>ExtendedProps` declarations for the Box, Button, Calendar and Drop components, forming part of issue #4998

The only files changed are the TS declaration files for each of the four mentioned components. All components previously defined a `<component>Type`, which has been kept in addition to the `<component>ExtendedProps` interfaces for backwards compatibility. The reviewer should double check that the changes are backwards compatible. I am happy to discuss potential deprecation warnings etc. and log a separate issue for deprecation notices if required.

Importing the extended prop interfaces from a separate project should allow `JSX.IntrinsicElements` keys to be passed through to the imported interface, e.g. importing `CalendarExtendedProps` should make `div` keys available, as well as general HTML keys like `children`. Also double check any keys included in `Omit<..., ...>` types for consistency with original intersected types.

Progress tracking: #4998
Design/discussion: #4978

No

Backwards compatible.

### DIFF
--- a/src/js/components/Box/index.d.ts
+++ b/src/js/components/Box/index.d.ts
@@ -190,7 +190,13 @@ export interface BoxProps {
   wrap?: boolean | 'reverse';
 }
 
-declare const Box: React.FC<BoxProps & JSX.IntrinsicElements['div']>;
+export interface BoxExtendedProps
+  extends BoxProps,
+    Omit<JSX.IntrinsicElements['div'], keyof BoxProps> {}
+
+// Keep type alias for backwards compatibility.
 export type BoxTypes = BoxProps & JSX.IntrinsicElements['div'];
+
+declare const Box: React.FC<BoxExtendedProps>;
 
 export { Box };
--- a/src/js/components/Drop/index.d.ts
+++ b/src/js/components/Drop/index.d.ts
@@ -45,6 +45,6 @@ export interface DropExtendedProps extends DropProps, divProps {}
 // Keep type alias for backwards compatibility.
 export type DropType = DropProps & JSX.IntrinsicElements['div'];
 
-declare const Drop: React.ComponentClass<DropExtendedProps>;
+declare const Drop: React.FC<DropExtendedProps>;
 
 export { Drop };
--- a/src/js/components/Drop/index.d.ts
+++ b/src/js/components/Drop/index.d.ts
@@ -1,5 +1,11 @@
 import * as React from 'react';
-import { BackgroundType, ElevationType, KeyboardType, MarginType, RoundType } from '../../utils';
+import {
+  BackgroundType,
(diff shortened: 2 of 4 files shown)


src/js/components/Box/index.d.ts
```diff
@@ -190,7 +190,13 @@ export interface BoxProps {
   wrap?: boolean | 'reverse';
 }
 
-declare const Box: React.FC<BoxProps & JSX.IntrinsicElements['div']>;
+export interface BoxExtendedProps
+  extends BoxProps,
+    Omit<JSX.IntrinsicElements['div'], keyof BoxProps> {}
```
ShimiSun, line 195: why is `keyof` needed in this case?

src/js/components/Drop/index.d.ts
```diff
@@ -32,8 +38,13 @@ export interface DropProps {
...
-declare const Drop: React.ComponentClass<DropProps &
-  JSX.IntrinsicElements['div']>;
+type divProps = JSX.IntrinsicElements['div'];
+
+export interface DropExtendedProps extends DropProps, divProps {}
+
+// Keep type alias for backwards compatibility.
 export type DropType = DropProps & JSX.IntrinsicElements['div'];
 
+declare const Drop: React.ComponentClass<DropExtendedProps>;
```
ShimiSun, line 48: Should we change ComponentClass to FC?

src/js/components/Box/index.d.ts
alexmirrington: I noticed `keyof` was already being used in other components, and it is a really nice solution to overiiding keys defined in `JSX.IntrinsicElements`, as it will mean `BoxExtendedProps` doesn't need to be changed if extra keys are ever added to `BoxProps`. This is because `keyof BoxProps` is essentially `key_0 | key_1 | ... | key_n` where `key_i` is a key in the `BoxProps` interface. Adding a new key to `BoxProps` automatically extends this list of keys to exclude.

alexmirrington: If you would prefer explicit declarations of which keys are being omitted, let me know and I will change `keyof BoxProps` to the union of all offending keys :)

src/js/components/Drop/index.d.ts
alexmirrington: Good catch, thanks!

src/js/components/Box/index.d.ts
ShimiSun: Nice. I think it would make sense to use it for use cases that have an overlap of ~3 props.

ShimiSun (approved):

Issue #5087, opened 2021-03-19
[TS] FormField should accept type of component props


exromany:
I'm trying to use `FormField` with custom component and pass it's own prop
```tsx
<FormField component={MyComponent} myExtraProp />
```
But TS swears at unknown prop `myExtraProp`

### Expected Behavior

I suppose `FormField` should be generic type
I suppose it is necessary to pass type of component props in generic 
```tsx
<FormField<{myExtraProp: boolean}> component={MyComponent} myExtraProp />
```

### Actual Behavior

There is no way to pass component's props type to `FormField`

jcfilben:
I think this issue can be closed since #5088 was merged

PR #5088, opened 2021-03-19
formfield as generic

exromany:
#5087


ShimiSun (approved):

ShimiSun:
@exromany Thank you for this PR!
@alexmirrington I'd appreciate another pair of eyes on this PR.

alexmirrington (approved):
I haven't ever used forms with Grommet, however from 20 mins of playing around with them, the idea of passing a component as a prop and then passing through its props as `{ ...rest }` when react inherently supports children seems like a bit of an anti-pattern to me. I understand that users rely on this feature (otherwise we wouldn't be seeing this PR), but it seems to be a bit of a code smell to me. Please correct me if there is an important justification behind why `FormFields` are implemented this way @ShimiSun.

That being said, in the meantime I am all for making `FormField` generic, since `Form` is as well. May as well add TypeScript support for those who are using this feature. PR looks good, it's the same as the `Form` types which is good, though missing some of the things we might like to see from `React.FC` like `defaultProps`. I think that is an issue for another day though, since it would require changes to `Form` types as well. Perhaps it's worth getting Eric to take a quick look since he reviewed the `Form` typing ~1 month ago? (#4841)

ShimiSun:
> I haven't ever used forms with Grommet, however from 20 mins of playing around with them, the idea of passing a component as a prop and then passing through its props as { ...rest } when react inherently supports children seems like a bit of an anti-pattern to me. I understand that users rely on this feature (otherwise we wouldn't be seeing this PR), but it seems to be a bit of a code smell to me. Please correct me if there is an important justification behind why FormFields are implemented this way @ShimiSun.

Sorry for the overdue, I agree in some ways with what you have shared, we are not encouraging to use the `component` prop of FormField due to its complexity with types & functionality and we do encourage to use the child component instead. Unfortunately, we should still support this functionality for backward compatibilty, so this PR is valid.

PR #5089, opened 2021-03-19
Only set focus on FormField if it containsFocus in the DOM

britt6612:
#### What does this PR do?
This PR adjusts FormField setFocus to check if the FormField contains DOM focus. If true, it sets focus to true and applies focus styling. If false, it does not apply focus styling since another DOM element has focus.

This resolves a bug where the TextInput in Select + Search was displaying focus on both the FormField and the TextInput. See "Background context" below for more details.

with fix
Screen Shot 2021-03-19 at 1 00 43 PM

without fix
Screen Shot 2021-03-19 at 1 01 07 PM

Formfield.js
storyboook
When `select` is in `formfield` there should only be one focus
```
        <FormField>
          <Select
            size="medium"
            placeholder="Select single option"
            value={value}
            options={options}
            onSearch={text => {
              // The line below escapes regular expression special characters:
              // [ \ ^ $ . | ? * + ( )

              // Create the regular expression with modified value which
              // handles escaping special characters. Without escaping special
              // characters, errors will appear in the console
              setOptions(defaultOptions.filter(o => exp.test(o)));
            }}
          />
        </FormField>

```

Select with Search applies focus onto the Search Input when the Select Drop opens. However, FormField is controlling its own focus internally and was not aware that the DOM focus had been placed on the search input in the select drop. This resulted in a focus indicator being placed on both the search input and the formfield (even though the Search Input was the active DOM element). By adding this check, we ensure that FormField only sets its focus to true when it also has focus in the DOM.

https://github.com/grommet/grommet-theme-hpe/issues/171
no
backwards compatible

### DIFF
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -8,7 +8,12 @@ import React, {
 import styled, { ThemeContext } from 'styled-components';
 import { defaultProps } from '../../default-props';
 
-import { focusStyle, parseMetricToNum } from '../../utils';
+import {
+  containsFocus,
+  useForwardedRef,
+  focusStyle,
+  parseMetricToNum,
+} from '../../utils';
 import { Box } from '../Box';
 import { CheckBox } from '../CheckBox';
 import { CheckBoxGroup } from '../CheckBoxGroup';
@@ -171,6 +176,7 @@ const FormField = forwardRef(
       validate,
     });
     const [focus, setFocus] = useState();
+    const formFieldRef = useForwardedRef(ref);
 
     const { formField: formFieldTheme } = theme;
     const { border: themeBorder } = formFieldTheme;
@@ -380,14 +386,14 @@ const FormField = forwardRef(
 
     return (
       <FormFieldBox
-        ref={ref}
+        ref={formFieldRef}
         className={className}
         background={outerBackground}
         margin={abut ? abutMargin : margin || { ...formFieldTheme.margin }}
         {...outerProps}
         style={outerStyle}
         onFocus={event => {
-          setFocus(true);
+          setFocus(containsFocus(formFieldRef.current));
           if (onFocus) onFocus(event);
         }}
         onBlur={event => {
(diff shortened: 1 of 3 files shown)


taysea (approved):
This LGTM.

ShimiSun:
Can we please not update `yarn.lock` as part of this PR?

src/js/components/FormField/FormField.js
```diff
@@ -8,7 +8,12 @@ import React, {
 import styled, { ThemeContext } from 'styled-components';
 import { defaultProps } from '../../default-props';
 
-import { focusStyle, parseMetricToNum } from '../../utils';
+import {
+  containsFocus,
+  useForwardedRef,
+  focusStyle,
+  parseMetricToNum,
+} from '../../utils';
```
ShimiSun, line 16: Please use explicit imports and alphabetized order. e.g.
```suggestion
import {
  containsFocus,
} from '../../utils/DOM';
....
```

ShimiSun:
Also @taysea, can you please add details on why does this fix solves the problem. It will help with the context of the solution.

taysea:
> Also @taysea, can you please add details on why does this fix solves the problem. It will help with the context of the solution.

Yes, I'll add to the PR description.

src/js/components/FormField/FormField.js
```diff
@@ -8,7 +8,10 @@ import React, {
 import styled, { ThemeContext } from 'styled-components';
 import { defaultProps } from '../../default-props';
 
-import { focusStyle, parseMetricToNum } from '../../utils';
+import { containsFocus } from '../../utils/DOM';
+import { useForwardedRef } from '../../utils/refs';
```
taysea, line 12: Can we move this to the bottom? so the imports are alphabetized

ShimiSun (approved):

ShimiSun:
Good stuff ladies!! ❤️ 

@leossantiago I thought you might e interested in seeing this fix 💟

PR #5094, opened 2021-03-23
Adjust rightOffset when caller applies pad on FileInput

taysea:
#### What does this PR do?
When caller applies pad to FileInput, the dimensions of the pad need to be taken into account in the rightOffset of the `input`. If not, the `input` will cover the RemoveButton which allows the user to delete a file they have uploaded.

Previously, the pad was not taken into account which meant that the RemoveButton would at times be inaccessible by the mouse.

Also resolved incorrect propType in docs for `renderFile`.

src/js/components/FileInput/FileInput.js

Test in Custom Story.

In CustomStory. Upload a file then click to remove.

Previously, the input would cover the remove button when pad was applied.
Screen Shot 2021-03-22 at 7 50 25 PM

Updated here.

New component.

Backwards compatible.

### DIFF
--- a/src/js/components/FileInput/FileInput.js
+++ b/src/js/components/FileInput/FileInput.js
@@ -111,6 +111,25 @@ const FileInput = forwardRef(
         : result;
     };
 
+    let rightPad;
+    if (mergeTheme('pad')) {
+      const { horizontal, right } = mergeTheme('pad');
+      if (right) {
+        rightPad = theme.global.edgeSize[right] || right;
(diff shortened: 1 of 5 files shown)


ShimiSun (approved):

ericsoderberghp:
darn, too slow to get in on the action here

taysea:
> darn, too slow to get in on the action here

Feel free to get in on the action though if anything seems fishy : )

Issue #4461, opened 2020-08-27
Preact support effort - we need your inputs


ShimiSun:
Due to the [PR filed to improve Preact support](https://github.com/grommet/grommet/pull/4346)
We would like to understand what is the current demand for supporting Preact around the community.

Currently, we rather spend our efforts and resources on higher-profile requirements and enhancements, and unfortunately, Preact support isn't part of our backlog -- if you would like grommet to invest more resources in making grommet a more Preact friendly framework, with the right Preact testing coverage, please comment on this issue and add whether you'll be able to help and contribute to the open-source to complete this effort.

ShimiSun:
Posted on Slack to receive more feedback.
![image](https://user-images.githubusercontent.com/6320236/98709259-63404400-233f-11eb-818e-92869d47b288.png)

sebringrose:
It's been a while since this issue was raised but I would like preact support. However, I appreciate that it won't be needed by the vast majority of users. I just started a preact project and would have liked to use grommet components but no big deal :)

jcfilben:
Are there certain Grommet components that are not working well with Preact? If so it would be great to get a list of these

jcfilben:
I am going to close this due to lack of recent activity but if there is still interest in this feel free to add a comment on the issue and we can re-evaluate

PR #5098, opened 2021-03-24
Fix for native event in Select component (Preact support)

bojan88:
Signed-off-by: Bojan Durdevic <bojan88@gmail.com>

#### What does this PR do?
Fixes https://github.com/grommet/grommet/issues/4340 - error throw by Select component in Preact `Cannot assign to read only property 'target' of object '#<MouseEvent>'`.

I copied a few examples from the storybook into a Preact component and tested both mouse and keyboard events.

The best way is to start a new Preact project, installe Grommet, replace with locally built Grommet (I'm using symbolic link), and try different examples from Grommet storybook.

I'm working on a project that's using Grommet, and most of the components work nicely, but it's very annoying that we need to always throw in a replacement for Select component.

This change should not affect React projects in any way.
I didn't want to modify the original native Mouse/Keyboard event, so I created a copy.

Bug report - https://github.com/grommet/grommet/issues/4340
General Preact support - https://github.com/grommet/grommet/issues/4461
Previous attempt to fix this with conversation about the solution - https://github.com/grommet/grommet/pull/4346

No.

Maybe as a fix for Preact.

No breaking changes


src/js/components/Select/Select.js
```diff
@@ -158,11 +158,23 @@ const Select = forwardRef(
...
         if (onChange) {
           event.persist();
-          const adjustedEvent = event;
-          adjustedEvent.target = inputRef.current;
-          adjustedEvent.value = nextValue;
-          adjustedEvent.option = option;
-          adjustedEvent.selected = nextSelected;
+          let adjustedEvent;
+          // support for native event used by Preact
+          if (event instanceof Event) {
```
bojan88, line 163: Another option is to have this condition `event.constructor.name !== 'SyntheticBaseEvent'`

ShimiSun:
I'm not sure we'd like to support this functionality due to the existing issue of #4461 , let us know what you think.

bojan88:
> I'm not sure we'd like to support this functionality due to the existing issue of #4461 , let us know what you think.

I guess it's up to you and other Grommet maintainers.
If you wanted my opinion... I think you should support it if it's not taking too much of your time.

ericsoderberghp:
Agreed that we shouldn't be changing the `target` of the event the way it currently is done. I'm wondering though if the recent changes to add `triggerChangeEvent` might help us think through how this should work. For instance, `triggerChangeEvent` will cause an onChange event with the correct target, controlled by the DOM. This leaves the need to still send an `onChange` with the `value`, `option`, and `selected` keys. Can we either remove the call to `onChange` from `onSelectChange` and instead listen for `onChange` from SelectTextInput and tie those keys in there? Since we wouldn't need to adjust `target`.

bojan88:
@ericsoderberghp `onChange` is not triggered on `SelectTextInput`.
We can listen to `onInput` on `SelectTextInput`, and call our `onChange` from there. But note that that will not work for multi value select. With the latest changes, `input` event is dispatch only if `nextValue` is a string, and in case of multiple values (array) we would still need to have `onChange` called from `onSelectChange`.

ericsoderberghp:
> @ericsoderberghp `onChange` is not triggered on `SelectTextInput`.
> We can listen to `onInput` on `SelectTextInput`, and call our `onChange` from there. But note that will not work for multi value select. With the latest changes, `input` event is dispatched only if `nextValue` is a string, and in case of multiple values (array) we would still need to have `onChange` called from `onSelectChange`.

That sounds like `<Select multiple />` will have issues with behaving correctly in an uncontrolled Form context, where the `onChange` needs to be tied to a native change event on the `<select />` DOM element.

bojan88:
@ericsoderberghp We might need to confirm that, but based on this line https://github.com/grommet/grommet/blob/518373864c2608df9faf4a6de8900ef5a55d8f20/src/js/components/Select/Select.js#L175 `triggerChangeEvent` is only invoked for non multi select.

src/js/components/Select/Select.js
```diff
@@ -182,11 +182,23 @@ const Select = forwardRef(
...
+            adjustedEvent = new event.constructor(event.type, event);
+            Object.defineProperties(adjustedEvent, {
+              target: { value: inputRef.current },
```
ericsoderberghp, line 190: This looks like `target.value` will be a reference to the input element. Can this be `target: inputRef.current,` instead?

src/js/components/Select/Select.js
ericsoderberghp: Would it be reasonable to use the `new event.constructor` approach in all cases?

src/js/components/Select/Select.js
bojan88: You confused me for a moment...
`Object.defineProperties` expects this type of value where we have `{ propertyName: { value: 'someValue', writable: true/false, configurable: true/false, ... } }`. Here's [an MDN example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperties#examples).
And a quick test:
screenshot

src/js/components/Select/Select.js
```diff
@@ -186,11 +186,26 @@ const Select = forwardRef(
...
-          adjustedEvent.target = inputRef.current;
-          adjustedEvent.value = nextValue;
-          adjustedEvent.option = option;
-          adjustedEvent.selected = nextSelected;
+          const adjustedEvent =
+            event instanceof Event
+              ? // native event
+                new event.constructor(event.type, event)
+              : // react SyntheticBaseEvent
+                new event.constructor(
```
ericsoderberghp, line 194: Well, I didn't realize it would be this involved. This feels like it's too intimate with the internals of React. I think the way you had it before is preferable. Thanks for trying this out though. Sorry for the flip-flop.

bojan88: No problem. I had the same idea, but gave up when I saw that it looks complicated.
I reverted it.

src/js/components/Select/Select.js
ericsoderberghp: Ah, I learned something new today, thanks.

ericsoderberghp (approved):

Issue #5095, opened 2021-03-23
Select: Warning Maximum update depth exceeded.


versoworks:
Updated to the latest v2.17 and am seeing the following warning when opening the `Drop` on a `Select` component;

`Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
    at InfiniteScroll (https://p86yu.csb.app/node_modules/grommet/es6/components/InfiniteScroll/InfiniteScroll.js:21:23)`

Here's a [Codesandbox ](https://codesandbox.io/s/elated-proskuriakova-p86yu?file=/src/App.js) reproducing the issue.

ShimiSun:
Thanks for reporting.

Notes: 
1. The warning appears when trying to open Select.
2. The `fill` prop on the Grommet isn't valid and throwing warning as well - `fill` could be removed.

halocline:
Investigation Update:
It looks like this might have been introduced in:
![Screen Shot 2021-03-23 at 11 13 10 AM](https://user-images.githubusercontent.com/1756948/112188791-ec5a6800-8bc8-11eb-987e-322be449ade8.png)

ShimiSun:
Thanks, @halocline. I'll tag @leossantiago for awareness and potential insights.

halocline:
These two useEffects seem to be entering into an infinite loop. Each have dependencies of `pendingLength` which gets set in each, which then satisfies the dependency condition as well as the checks within the useEffects. Seems like we need to add an additional condition around `pendingLength`.
![Screen Shot 2021-03-23 at 11 46 26 AM](https://user-images.githubusercontent.com/1756948/112195437-93da9900-8bcf-11eb-9284-43f1e3a99c21.png)

leossantiago:
Thanks for the inputs @ShimiSun and @halocline . I'll definitely take a look at this later today. Hope we can have a fix out ASAP 😄

halocline:
@leossantiago It looks like you originally had a check for `items.length === 0` which would prevent the infinite loop.

I believe with that approach, it assumes the caller would always be resetting their data to an empty array. Something just feels a bit off about assuming that is what the caller would want to do.

I was exploring some other approaches to see if we could avoid that assumption, but no luck so far.

We may need to put `items.length === 0` in as a patch to resolve the regression, and then explore working around the assumption later. @ShimiSun thoughts on this approach?

[...]

PR #5100, opened 2021-03-25
Fix InfiniteScroll loop when Data length changes

leossantiago:
#### What does this PR do?

Introduces a patch to #5095,  introduced on PR #4936 .

This fix uses two states to help tracking changes to data length. On this approach, we avoid going back to using `items.length === 0` as a check and thus we wouldn't assume that the caller wants to reset the data to an empty array before repopulating it.

Please note that this covers a very specific use case ( #4934 ) where the reporter wanted to reset data to an empty array before setting it. The condition `lastPage === 0` will make this work only for situations where the modified length is smaller than the initial step. 

Would love to hear your thoughts on that @ShimiSun @halocline, and I'd suggest we open this matter to discussion so we can have a clear view of use cases we want to cover before any attempt of broadening the whole reset-onMore-calls thing.

`src/js/components/InfiniteScroll/InfiniteScroll.js`

Manual testing as JSDom and Jest do not support calls such as getBoundingClientRect, making this PR unable to be tested using automated tests.

Code provided on PR #4936 and issue #5095.

#5095

No.

Backwards compatible.

### DIFF
--- a/src/js/components/InfiniteScroll/InfiniteScroll.js
+++ b/src/js/components/InfiniteScroll/InfiniteScroll.js
@@ -151,12 +151,12 @@ const InfiniteScroll = ({
   }, [items.length, lastPage, onMore, pendingLength, renderPageBounds, step]);
 
   useEffect(() => {
-    if (lastPage === 0 && pendingLength !== 0) {
+    if (items.length === 0 && lastPage === 0 && pendingLength !== 0) {
       setPageHeights([]);
       setPendingLength(0);
       setRenderPageBounds([0, calculateLastPageBound(show, step)]);
     }
-  }, [lastPage, pendingLength, show, step]);
+  }, [lastPage, pendingLength, show, step, items.length]);
 
   // scroll to any 'show'
   useLayoutEffect(() => {


halocline:
Thank you @leossantiago. Much appreciated. I will check this out within the next hour. 

I agree that discussion should happen around the use cases we want to support.

halocline:
> Would love to hear your thoughts on that @ShimiSun @halocline, and I'd suggest we open this matter to discussion so we can have a clear view of use cases we want to cover before any attempt of broadening the whole reset-onMore-calls thing.
> 
💯  We definitely need to discuss the scenarios we want to support.

> On this approach, we avoid going back to using `items.length === 0` as a check and thus we wouldn't assume that the caller wants to reset the data to an empty array before repopulating it.
> 
> Please note that this covers a very specific use case ( #4934 ) where the reporter wanted to reset data to an empty array before setting it. The condition `lastPage === 0` will make this work only for situations where the modified length is smaller than the initial step.
> 

@leossantiago This approach works around the assumption of wanting to reset to an empty array, but to your point about:
> this [will only] work for situations where the modified length is smaller than the initial step.
> 

In your opinion, with the constraint of `items.length < step`, does the incremental value of not assuming an empty array justify the cost of the additional complexity? Would we be better off keeping things simple for now using the `items.length === 0` approach and then having a proper use case and design discussion as follow up?

leossantiago:
I believe using `items.length === 0` would be the best option for now. The whole point of having the `useEffect` introduced on #4936 was to deal with `onMore` not working as intended when resetting the data passed to `DataTable`. Even though we are trying to find a solution that works for any changes on data, we are still bound to having a `lastPage` equal to zero in order for this to work.

> and then having a proper use case and design discussion

Definitely. This was introduced based on a very specific case. We need to figure out what we want to have as general behavior.. 
I wonder if removing the `lastPage` condition would do the trick, but I'm worried this could lead us to another loop.

halocline:
Appreciate you sharing your thoughts.

> I believe using `items.length === 0` would be the best option for now. 

I agree and would support moving forward with this direction to get the fix in. We can then discuss how/whether we want to enhance the onMore/reset concept.

Would you be willing to modify this PR to head in this direction? Or, I am have ability this afternoon and would be happy to open a separate PR.

leossantiago:
> Would you be willing to modify this PR to head in this direction? 

Sure thing. Will do that in a couple of minutes.

leossantiago:
@halocline done! Just force pushed an update to this PR.

src/js/components/InfiniteScroll/InfiniteScroll.js
```diff
@@ -151,12 +151,12 @@ const InfiniteScroll = ({
   }, [items.length, lastPage, onMore, pendingLength, renderPageBounds, step]);
 
   useEffect(() => {
-    if (lastPage === 0 && pendingLength !== 0) {
+    if (items.length && lastPage === 0 && pendingLength !== 0) {
```
halocline, line 154: Think we lost the `=== 0`

```suggestion
    if (items.length === 0 && lastPage === 0 && pendingLength !== 0) {
```

leossantiago: Oops. My bad. Will fix it in a sec!

halocline (approved):
Thank you. Again, appreciate you working through this with us.

Tested manually with the following stories:
[Select + onMore](https://gist.github.com/halocline/d8d7d11aca684eaa624f840d6c703106)

[InfiniteScroll + onMore + reset](https://gist.github.com/halocline/f805d07559981dda24a19f870d6666bf)

leossantiago:
Np, @halocline! Feels great to be part of this community! Glad to be helpful 😄

ShimiSun (approved):
Thank you @leossantiago ! Enviando amor do Colorado!

PR #5101, opened 2021-03-25
Enhance DataTable to support groupBy and onSelect together

### LINKED ISSUE #5067: Enhance DataTable to allow `groupBy` and `select` at the same time
### Expected Behavior

User should be able to group and apply selections to DataTable data. Some work will need to by considered around where the checkboxes will appear when data is under a group/how indeterminate behavior will function when selecting a subset of a group.

Motivated by designs coming from Financial Services. 

### Actual Behavior

User must choose one or the other. Applying both incorrectly renders the data in the wrong columns/checkboxes don't appear.

Screen Shot 2021-03-15 at 6 06 26 PM

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

taysea:
#### What does this PR do?
This PR enhances DataTable to support `groupBy` and `onSelect`/`select` together. The intended behaviors are:
1. If all data in a table is selected, the checkbox in the group header should be checked.
2. If a portion of a group is selected, the checkbox in the group header should be indeterminate and the checkbox in the table header should be indeterminate.

src/js/components/DataTable/GroupedBody.js

Added storybook example "Grouped and onSelect"

Added jest tests.

Explore "Grouped and onSelect" storybook example in deploy.

Previously, if users tried to do `groupBy` and `onSelect` simultaneously, the table would break -- meaning, checkboxes would render in the table header but not in the body. Data would then be improperly rendered in the wrong columns.

Closes #5067 

Updated here.

Yes.  Enhanced DataTable to support use of `groupBy` and `onSelect`/`select` together.

Backwards compatible.


src/js/components/DataTable/Header.js
```diff
@@ -142,6 +142,11 @@ const Header = forwardRef(
             <TableCell background={background || cellProps.background}>
               {onSelect && (
                 <CheckBox
+                  a11yTitle={
```
halocline, line 145: 👍

src/js/components/DataTable/__tests__/DataTable-test.js
```diff
@@ -1116,4 +1116,122 @@ describe('DataTable', () => {
...
+      </Grommet>,
+    );
+    expect(container.firstChild).toMatchSnapshot();
+
+    let headerCheckBox;
+    headerCheckBox = getByLabelText('select all');
+    fireEvent.click(headerCheckBox);
+    expect(container.firstChild).toMatchSnapshot();
+
+    // aria-label should have changed since all entries
```
halocline, line 1147: 👍

src/js/components/DataTable/__tests__/DataTable-test.js
```diff
@@ -1116,4 +1116,122 @@ describe('DataTable', () => {
 
     expect(container.firstChild).toMatchSnapshot();
   });
+
```
halocline, line 1119: Would it be worthwhile to add a test checking that `value` is being stored appropriately in these scenarios? I.e. no data duplications, only contains unique items, etc.?

I don't see anything that would cause it in this work, but wondering from a defensive standpoint?

halocline (commented):
Looks solid. Good catches on the `a11yTitles`.  One question about adding an assertion to make sure the value is tracked/stored as expected.

src/js/components/DataTable/__tests__/DataTable-test.js
taysea: It's a good point. the `selected` value will hold the primaryKeys. However, it's up to the grommet user to denote which column holds the data that will serve as the primaryKey. Docs: https://v2.grommet.io/datatable#columns

Because we can't anticipate the data structure provided by the caller, it would be on them to provide the information around which column has primary data. However, I've added a check to my test to make sure onSelect has been called with the desired values.

halocline: Nice. The onSelect assertion is the right approach and is the check I was hoping for. Thanks.

ericsoderberghp (approved):

Issue #5105, opened 2021-03-26
DataTable: when a header has line break (or spaces in it) and the viewport is small, the sort & search icons overflow into the next column and are sometimes unusable


sbytnar:
The Grommet storybook for DataTable demonstrates the problem.
Resize it so the view is about 800 pixels wide or smaller.
https://storybook.grommet.io/?path=/story/visualizations-datatable-sort--sort
My actual use case is to have a very large number of columns in the table, but it happens in the smaller tables too.

### Expected Behavior

When search & sort are enabled on columns, those button elements must stay within their original column and must not overflow into other columns.

### Actual Behavior

The search & sort & column divider lines overflow into other header columns making them inaccessible in some cases.
I haven't found a combination of adding a Box/Text wrapper on the header content that makes it render nicely. Text workBreak options don't seem to help and not breaking on words isn't acceptable....
One (unacceptable) way to make it work as desired is to make sure there are no breaks or spaces in the header row content.

### URL, screen shot, or Codepen exhibiting the issue

![image](https://user-images.githubusercontent.com/198164/112686433-ede39480-8e43-11eb-8552-332894540e00.png)

### Steps to Reproduce

Load https://storybook.grommet.io/?path=/story/visualizations-datatable-sort--sort
Resize it to smaller sizes and watch the column sort/search icons and the column divider indicator.

### Your Environment

Safari, Chrome.

PR #5114, opened 2021-03-29
Fix DataTable Searchers wrap on small screens

leossantiago:
#### What does this PR do?

Provides a fix for #5105 by forcing column headers to wrap content and thus avoiding them overflowing on top of each other on small screens.

Searcher and Resizer are handled in the same part of the code. They are added to a content `Box` alongside the column header content. Since it's a Box, we are able to force a `wrap` in case its content doesn't fit.

`src/js/components/DataTable/Header.js`

Manual testing through storybook, as this requires a small screen size to be observed properly.

Storybook: Visualizations / DataTable / Sort

#5105 

Peek 2021-03-29 02-38

No.

Backwards compatible.

### DIFF
--- a/src/js/components/DataTable/Header.js
+++ b/src/js/components/DataTable/Header.js
@@ -268,6 +268,7 @@ const Header = forwardRef(
                     justify={!align || align === 'start' ? 'between' : align}
                     gap={theme.dataTable.header.gap}
                     fill="vertical"
+                    wrap
                     style={onResize ? { position: 'relative' } : undefined}
                   >
                     {content}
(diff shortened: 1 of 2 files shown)


ShimiSun (approved):

ericsoderberghp:
Screen Shot 2021-04-16 at 8 42 50 AM

- The placement of the vertical border between cells doesn't look right here.
- There will always be a point where the horizontal space isn't enough to fit everything without wrapping. Before wrapping, I think it would be preferable from a design standpoint to first add horizontal scrolling and then fall back to truncating the header text.
- FYI, we are intending to deprecate this search behavior embedded in the table cell header in the next major version (nothing imminent). Instead, we recommend either a single search across all columns or a separate per-column filter/search panel or layer.

ShimiSun:
Hi @leossantiago ! Any chance we circle back to this PR and see if we can incorporate the feedback?

leossantiago:
I'm converting this to a draft to give anyone the chance of working on it as I'm currently working on another open pull request.

jcfilben:

PR #5115, opened 2021-03-29
Fix Carousel showed slide when mounted

### LINKED ISSUE #5093: Carousel shows different image on mount
When you mount a Carousel component containing images (probably also with other children) it shows the last image for a brief moment before displaying the first image (initialChild has to be set to 0).

### Expected Behavior
Show the first image instantly without a fading animation.

### Actual Behavior
Shows the last image for a brief moment before showing the first image.

The best way to showcase this is using a layer because this forces the initial render after a button click.
For demonstration purposes: https://codesandbox.io/s/grommet-carousel-fix-kokh7?file=/src/BrokenModal.tsx
Note that I copied the source code of the Carousel component and had to adapt it to TypeScript. However the only line I changed in the "fix" was the rendering of the wrappedChildren, where I reversed the array (CarouselFixed.tsx line 187). This fix does only work when initialChild is set to 0. If it is set to another value, e.g 1 or 2, the first image is shown for a brief moment before displaying the specified image (which applies to both the broken and the fixed version). 

### Steps to Reproduce

1. Create Carousel with images (best in modal)
2. Watch render of Carousel

### Your Environment

- React version: 17.0.0
- Grommet version: v2.17.0
- Browser Name and version: Google Chrome 89.0.4389.90
- Operating System and version (desktop or mobile): MacOS Big Sur 11.2.3

leossantiago:
#### What does this PR do?

Provides a fix for #5093.

Following the insights of the issue reporter, reversing the Carousel children fixes the behavior for some cases (initialChild = 0). However, passing any other value as initialChild results in the same problem. 

The problem itself is related to children animations. That's how Carousel navigation is handled, by changing their visibility. This fix solves the problem by reordering the children based on which one is the activeIndex, and also changing `z-Index` and `opacity` for the other children.

`src/js/components/Carousel/Carousel.js`

Manual testing and the existing tests (that were updated)

```js
const CarouselModal = ({ close }) => (
  <Layer onClickOutside={close}>
    <Carousel controls="arrows" initialChild={0}>
      <Image fit="cover" src="//v2.grommet.io/assets/Wilderpeople_Ricky.jpg" />
      <Image fit="cover" src="//v2.grommet.io/assets/IMG_4245.jpg" />
      <Image fit="cover" src="//v2.grommet.io/assets/IMG_4210.jpg" />
    </Carousel>
  </Layer>
);

export const Simple = () => {
  const [showModal, setShowModal] = useState(false);

  return (
    <Grommet theme={grommet}>
      <Button label="Show Modal Carousel" onClick={() => setShowModal(true)} />

      {showModal && <CarouselModal close={() => setShowModal(false)} />}
    </Grommet>
  );
};
```

Closes #5093

No.

Backwards compatible, but it might break some snapshots as children's order might differ a little bit.

### DIFF
--- a/src/js/components/Carousel/Carousel.js
+++ b/src/js/components/Carousel/Carousel.js
@@ -106,7 +106,8 @@ const Carousel = ({
   );
 
   const selectors = [];
-  const wrappedChildren = Children.map(children, (child, index) => {
+  const wrappedChildren = [];
+  Children.map(children, (child, index) => {
     selectors.push(
       <Button
         // eslint-disable-next-line react/no-array-index-key
@@ -136,13 +137,33 @@ const Carousel = ({
       animation = { type: 'fadeOut', duration: 0 };
     }
 
-    return (
-      <Box fill={fill} overflow="hidden">
-        <Box fill={fill} animation={animation}>
-          {child}
-        </Box>
-      </Box>
-    );
+    if (index === activeIndex) {
+      wrappedChildren.push(
+        // eslint-disable-next-line react/no-array-index-key
+        <Box key={index} fill={fill} overflow="hidden" style={{ zIndex: 2 }}>
+          <Box fill={fill} animation={animation}>
+            {child}
+          </Box>
+        </Box>,
+      );
+    } else {
+      wrappedChildren.unshift(
+        <Box
+          // eslint-disable-next-line react/no-array-index-key
+          key={index}
+          fill={fill}
+          overflow="hidden"
+          style={{
+            zIndex: index === priorActiveIndex ? 1 : 0,
+            opacity: index === priorActiveIndex ? 1 : 0,
+          }}
+        >
+          <Box fill={fill} animation={animation}>
+            {child}
+          </Box>
+        </Box>,
+      );
+    }
   });
 
   const NextIcon = theme.carousel.icons.next;
--- a/src/js/components/Carousel/__tests__/__snapshots__/Carousel-test.js.snap
+++ b/src/js/components/Carousel/__tests__/__snapshots__/Carousel-test.js.snap
@@ -106,6 +106,9 @@ exports[`Carousel basic 1`] = `
   -webkit-flex-direction: column;
   -ms-flex-direction: column;
   flex-direction: column;
+  opacity: 1;
+  -webkit-animation: hftlBD 1s 0s forwards;
+  animation: hftlBD 1s 0s forwards;
 }
 
 .c6 {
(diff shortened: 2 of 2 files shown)


src/js/components/Carousel/Carousel.js
```diff
@@ -136,13 +137,33 @@ const Carousel = ({
...
-      <Box fill={fill} overflow="hidden">
-        <Box fill={fill} animation={animation}>
-          {child}
-        </Box>
-      </Box>
-    );
+    if (index === activeIndex) {
+      wrappedChildren.push(
+        // eslint-disable-next-line react/no-array-index-key
+        <Box key={index} fill={fill} overflow="hidden" style={{ zIndex: 2 }}>
```
taysea, line 143: zIndex of `2` feels a bit off. Do we need to introduce zIndex at all?

src/js/components/Carousel/Carousel.js
```diff
@@ -136,13 +137,33 @@ const Carousel = ({
...
+      );
+    } else {
+      wrappedChildren.unshift(
+        <Box
+          // eslint-disable-next-line react/no-array-index-key
+          key={index}
+          fill={fill}
+          overflow="hidden"
+          style={{
+            zIndex: index === priorActiveIndex ? 1 : 0,
```
taysea, line 157: Same as above, do we need `zIndex`/`opacity`? Is there any way to reorder the children like you're doing and leverage the existing way that the other children's visibility was controlled?

taysea (commented):
Thanks for looking into this! Dropped a few comments seeing if there's anyway to eliminate the inline styling/leverage existing visibility behavior

DolceTriade:
Since this appears abandoned, I addressed the comments in the PR here https://github.com/grommet/grommet/pull/5315

It seems the style overrides are not necessary based on the examples in the sandbox.

leossantiago:
@DolceTriade Even though it's been a while, I'm still trying to figure this out. Guess I should have left a comment here.

I see that you closed the PR you opened earlier. Could you elaborate on why that approach doesn't work?

Any insights are welcome. Feel free to suggest changes or turn this into a conversation so we can try to solve this issue :)

DolceTriade:
@leossantiago 

Ahh, I mean when I just removed the zIndex changes and the opacity changes. They seemed to work in the story book but when I tried it in my real project, the issue persisted. 

I was trying to fix this another way with a "hidden" flag, so I'll see how this goes. Ultimately, this is a pretty ugly bug and I just wanna see it fixed :smile:

leossantiago:
@DolceTriade agreed! I want to see this fixed as well 😛 

Well, I'll keep working on it, but please, feel free to try alternative fixes and open another pull request if needed!
Also, I see that was your first PR here, so welcome aboard. You'll soon learn how great this community is, and I'm glad we are all looking forward to helping grommet grow :)

DolceTriade:
Ok, this time I have a real fix :smile: 

#5326 works on both the storyboard *AND* my project. I believe it is a simpler fix.

leossantiago:
Nice! I'll be closing this as #5326  was already merged! 🎉

ShimiSun:
@leossantiago Thank you for starting the work and the conversation that lead to fixing this issue! ❤️

PR #5116, opened 2021-03-29
Controlled Carousel use case

### LINKED ISSUE #5079: Allow the option to make the Carousel component controlled
**Background**:
The current implementation of the Carousel is that it is an uncontrolled component, where a list of slides should be provided as children, and the internal logic of the Carousel hooks up all the actions. There is also a prop that acts as an eventListener called onChild, but it is more about informing outside components when a slide changes.

**Issue**
The Carousel component is uncontrolled; i.e. there is no programatic way of triggering navigations or other actions. E.g. I have this use case where I set of external buttons sit above the Carousel, and each button is associated with a Carousel slide. When tapping on a button, the Carousel should flip to that slide. With the current implementation, there is no right way to do it, so my temporary solution is hacky, and relies on having the Carousel buttons in place and that their order in the DOM never changes.

Please see this sandbox for my use case and my temporary solution - https://codesandbox.io/s/throbbing-surf-jo01c?file=/src/App.js

### Expected Behavior
It would be great if by default the Carousel is kept uncontrolled, but when providing an activeSlide and onSlideChange function, the developers could trigger these actions from outside the parent component.

### Actual Behavior
The component is uncontrolled and there is no programatic way of extending its behaviour.

### Your Environment

- Grommet version: 2.16.3
- Operating System and version (desktop or mobile): OSX

leossantiago:
#### What does this PR do?

Addresses #5079 by adding the option of making Carousel a controlled component.

By adding the prop `activeIndex` we can control which Carousel slide will be active, and combined with a rework on the `onChild` function logic, this can be used as a listener to state changes (through Carousel controls). Extra checks were also added to avoid states updating to the same value.

`src/js/components/Carousel/Carousel.js`

Manual testing (storybook) and Jest tests

Storybook: Media / Carousel / Controlled

Closes #5079

Yes. Already updated.

Backwards compatible. Even though a hook was replaced by function calls, this shouldn't affect the component's behavior.


src/js/components/Carousel/doc.js
```diff
@@ -20,16 +20,21 @@ export const doc = Carousel => {
 
   DocumentedCarousel.propTypes = {
     ...genericProps,
+    activeIndex: PropTypes.number.description(
```
ericsoderberghp, line 23: Can we rename this to `activeChild`, to align with `initialChild` and `onChild`? In retrospect, it would have been better to use "Index" as the suffix instead of "Child". But, here we are.

ericsoderberghp (commented):
Just a naming comment, otherwise looking good.

src/js/components/Carousel/Carousel.js
```diff
@@ -32,12 +34,30 @@ const Carousel = ({
   const timerRef = useRef();
 
   const [indexes, setIndexes] = useState({
-    activeIndex: initialChild,
+    activeIndex: activeChild || initialChild,
```
ericsoderberghp, line 37:
```suggestion
    activeIndex: activeChild !== undefined ? activeChild : initialChild,
```

ericsoderberghp: If `activeChild` is `0`, we would incorrectly fall back to `initialChild`.

ericsoderberghp (approved):

Issue #2529, opened 2018-12-04
Property consolidation


tbranyen:
There are many unique property names to Grommet that could be aligned better with the web specification. For instance the property `a11yTitle` is odd since it has numbers and is an abbreviation of accessibility. I would expect this property to be named `ariaLabel`.

Other examples of uniquely named props:

- `gap` vs `margin`
- `pad` vs `padding`
- `tag` vs `as`

A suggestion would be to consolidate the properties over the course of two full major version changes. In v3 introduce deprecation warnings for using the old props with suggestions for the new ones, possibly produce a codemod to do this for your users, and in v4 completely remove the old props.

oorestisime:
the `tag` vs `as` part is already at the consolidation stage. in case you found out a component that supports `tag` but not `as` shout and i could handle it as well!

tbranyen:
@oorestisime awesome, where can I find out more about the consolidation effort? I don't see any indication in the code that `tag` is being deprecated.

If `tag` and `as` are getting aligned that's great, but what about the other props I listed?

oorestisime:
The issue is: https://github.com/grommet/grommet/issues/2487
There is no warning when a user uses `tag` right now. But there is a Note in the documentation https://github.com/grommet/grommet/blob/master/src/js/components/Box/doc.js#L240-L242
There should be the same Note in the other components, unless i missed some.

Can't speak for the other props, i just know about the `tag` and `as` since i worked recently  on them.

ericsoderberghp:
Box currently supports both `margin` and `gap`. How were you envisioning moving `gap` to `margin`?

ericsoderberghp:
`pad` was a legacy from grommet v1, partly due to trying to keep things shorter and simpler. It hasn't been an issue before that it isn't called `padding`. We have had questions about `a11tTitle`.

tbranyen:
@ericsoderberghp I'll dig more into `gap`, on the surface it looks to be very similar to `margin`.

alansouzati:
I agree with deprecating `a11yTitle` and rely on `aria-label` instead.

episodeyang:
The type for this is currently set to `String` by mistake. Shall we change the type to accommodate custom components the same way that the styled components do it?

#2869

powerglove-eth:
I like gap and pad.

PR #5119, opened 2021-03-30
Pass aria-label prop through as button attribute.

kid-icarus:
#### What does this PR do?

Previously we only used the a11yTitle prop for button aria labels. This is a little unintuitive to someone who is already familiar with a11y. Rather than deprecate the a11yTitle prop we allow for it in addition to aria-label.

At the Button component.

Added a unit test to ensure the Button has no aXe a11y violations, manually tested to see the attribute rendered.

Create a Button, passing in aria-label. Then, check to see that the rendered button does in fact contain the given `aria-label` attribute.

I created a `Button` and gave it an `aria-label`. TypeScript allowed this, which tripped me up. An alternative here is to just ensure that the [ButtonExtendedProps](https://github.com/kid-icarus/grommet/blob/cea630132b637c223dd155292afdcd5399afbdce/src/js/components/Button/index.d.ts#L45) disallows `aria-label`, which I tried, but was unable to get it to work properly by adding `aria-label` to the list of Omitted keys.

Someone mentioned this [here](https://github.com/grommet/grommet/issues/2529)

It is technically modifying the API, a minor bump may be warranted.

It is backwards compatible

### DIFF
--- a/src/js/components/Button/Button.js
+++ b/src/js/components/Button/Button.js
@@ -98,6 +98,7 @@ const Button = forwardRef(
       a11yTitle,
       active,
       align = 'center',
+      'aria-label': ariaLabel,
       color, // munged to avoid styled-components putting it in the DOM
       children,
       disabled,
@@ -300,7 +301,7 @@ const Button = forwardRef(
           {...rest}
           as={domTag}
           ref={ref}
-          aria-label={a11yTitle}
+          aria-label={ariaLabel || a11yTitle}
           colorValue={color}
           active={active}
           selected={selected}
--- a/src/js/components/Button/__tests__/Button-test.js
+++ b/src/js/components/Button/__tests__/Button-test.js
@@ -26,6 +26,22 @@ describe('Button', () => {
     expect(results).toHaveNoViolations();
   });
 
+  test('passes through the aria-label prop', async () => {
+    const TEST_LABEL = 'Test Label';
+    const { container, getByText } = render(
+      <Grommet>
+        <Button aria-label={TEST_LABEL} label="Test" onClick={() => {}} />
+      </Grommet>,
+    );
+
+    const button = container.querySelector('button');
+    expect(button.getAttribute('aria-label')).toEqual(TEST_LABEL);
+
+    fireEvent.click(getByText('Test'));
+    const results = await axe(container);
+    expect(results).toHaveNoViolations();
+  });
+
   test('basic', () => {
     const component = renderer.create(
       <Grommet>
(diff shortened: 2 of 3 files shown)


ShimiSun:
quick question, have you tried to pass `aria-label` via `rest` and it didn't work?

kid-icarus:
Oh good point @ShimiSun, I did actually try that but we spread `...rest` _before_  setting `aria-label` to `a11yTitle,` so it is always overridden to undefined.

Perhaps the solution here just to put the spread at the end, happy to change this PR and test that solution.

kid-icarus:
I actually just re-ordered the ...rest spread, and added a unit test to assert that the attribute is set correctly.

FYI, a strange issue here that I noticed: Even rendering a button with no aria-label, no a11yTitle, and no label still passes the aXe violation test, which seems incorrect to me.

kid-icarus:
Additionally, I'd be happy to check the rest of the components here to ensure that aria-labels are appropriate spread onto the correct element for all components, we seem to have inconsistent ...rest spread positioning between components.

See [Anchor](https://github.com/kid-icarus/grommet/blob/3b93319c04e60f8caa6b9f3881131c855f09dc9f/src/js/components/Anchor/Anchor.js#L59) vs [Box](https://github.com/kid-icarus/grommet/blob/1c1b69df83cca1834a287a6287982c6ef1114809/src/js/components/Box/Box.js#L151)

src/js/components/Button/Button.js
```diff
@@ -297,7 +297,6 @@ const Button = forwardRef(
     } else {
       styledButtonResult = (
         <StyledButton
-          {...rest}
           as={domTag}
           ref={ref}
           aria-label={a11yTitle}
```
ericsoderberghp, line 302:
```suggestion
          aria-label={aria-label || a11yTitle}
```

ericsoderberghp (commented):
I think the approach suggested in this review is less destabilizing. If a caller passes in any property via rest that conflicts with something grommet is setting internally, it will break how Button works, such as `hasIcon`.

kid-icarus:
This makes sense to me @ericsoderberghp and I think it may be beneficial to standardize (perhaps a lint rule exists for this purpose) spreading first so as to not override internally set properties.

One could argue that a TypeSript would prevent people from passing props that aren't allowed, but alas this is JavaScript world and propTypes only go so far.

ericsoderberghp (approved):

src/js/components/Button/__tests__/Button-test.js
```diff
@@ -26,6 +26,23 @@ describe('Button', () => {
...
+    );
+
+    const button = container.querySelector('button');
+    expect(button).toBeTruthy();
+    expect(button.getAttribute('aria-label')).toEqual('Test Label');
+
+    fireEvent.click(getByText('Test'));
+    const results = await axe(container);
+    expect(results).toHaveNoViolations();
+  });
```
ShimiSun, line 44: Where are the snapshots created for this test that are indicating that `aria-label` is applied as expected?

kid-icarus: There are no snapshots, I'm just manually asserting that the attribute is correctly set. If you find a snapshot valuable, I'm happy to add it.

kid-icarus: @ShimiSun I've added a snapshot that looks at both `a11yTitle` and `aria-label` props.

ShimiSun (approved):

ShimiSun:

Issue #5113, opened 2021-03-29
Bug: cannot properly theme buttons


jeznag:
### Expected Behavior

I should be able to change the background and border color of a secondary button.

### Actual Behavior

When I try and style secondary buttons I get really unexpected behaviour. The background color property seems to be ignored. Instead it uses the color prop

![image](https://user-images.githubusercontent.com/1034890/112780405-7f494580-9094-11eb-837f-713c0464229d.png)

![image](https://user-images.githubusercontent.com/1034890/112780428-896b4400-9094-11eb-9545-7ccf623b2556.png)

The only way I can get it to work is if I use extend which seems hacky.

```
active: {
      extend: css`
        background-color: green;
      `
    },
```

Even if I use extend on active, I then screw up the appearance of my primary buttons. I want to use the default styling for primary.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/themed-button-v2130-4bei0?file=/theme.js:3718-3794

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.16.2
- Browser Name and version:
- Operating System and version (desktop or mobile):

britt6612:
@jeznag I believe this is because you are also using `active` in the secondary buttons. If you look at this story 
https://storybook.grommet.io/?path=/story/controls-button-custom--custom 
you can see that the background color is different depending on `active` `disabled` etc... 

Now if I understand you correctly you want to set the border and background of the active secondary button. 
to do so I would add-in:

```
    secondary: {
      background: {
        color: "green"
      }
    },
    active: {
      background: { color: "pink" },
      secondary: {
        background: "green"
      }
    },

```

You would have your background color to be `green` when the `secondary` button is `active` as well as when it is not `active` then you can also specify a color for when a button is only `active`

Also as far as the `color prop` in the code it looks for the `color prop` first then applies that if provided if that is not there it looks at the theme and goes applies the theme color second.

Let me know if that helps. 👍

jeznag:
Hi Britt,
I don't believe I am using active secondary. My codesandbox did not include that.

The fork still has the same two problems. 
1. If you add a color to the secondary button, the background color changes. Please refer to https://codesandbox.io/s/themed-button-v2130-with-active-secondary-b92bg

2. The primary button styles are gone - I want to use the same primary styling as I get if I don't customise the theme

britt6612:
[...]

PR #5120, opened 2021-03-30
update docs for color prop in button

britt6612:
#### What does this PR do?
This PR updates the `color` prop for `button`
docs
There was some changes when `button.default` got added in 
the  `color` prop  no longer has the same definition 
When someone has `button.default` in their theme then the 
color prop will apply background color
issue #5113 
already done
no
backwards compatible


src/js/components/Button/doc.js
```diff
@@ -45,7 +45,10 @@ export const doc = Button => {
       `The DOM tag or react component to use for the element.`,
     ),
     color: colorPropType.description(
-      'Fill color for primary, label color for plain, border color otherwise.',
+      `Fill color for primary, label color for plain, border color otherwise.
+       If button.default is defined in the theme, then the color prop will
```
taysea, line 49:
```suggestion
       If button.default is defined in the theme, the color prop will
```

src/js/components/Button/doc.js
```diff
@@ -45,7 +45,10 @@ export const doc = Button => {
       `The DOM tag or react component to use for the element.`,
     ),
     color: colorPropType.description(
-      'Fill color for primary, label color for plain, border color otherwise.',
+      `Fill color for primary, label color for plain, border color otherwise.
+       If button.default is defined in the theme, the color prop will
+       affect the background color of the button, regardless of which
```
ShimiSun, line 50:
```suggestion
       affect the background color of the primary button, regardless of which
```
Should we be more accurate and point out that this color will fill the background only when the button is primary? also, how would it affect non-primary button?

britt6612: This will fill the background color if primary, secondary for default it will only change the text color I can add that

taysea (approved):
Thanks for the updates and specificity, LGTM.

ShimiSun (approved):

PR #5126, opened 2021-03-31
Fix formfield ts

### LINKED ISSUE #5118: Grommet global theme for formField complains that 'side' is incompatible (TypeScript)
### Expected Behavior

The theme prop added to the `Grommet` component should recognize the key `side` in `BorderType` when configuration global theme for formField

### Actual Behavior

I am trying to set a global theme for my formField where the border is displayed all around the TextInput. This is my theme configuration, which I took from the story book found here: https://storybook.grommet.io/?path=/story/input-form-field-custom-states--field-custom-states

```TypeScript
const theme = {
  formField:{
    border: {
      error: {
        color: 'border'
      },
      color: 'border',
      side: 'all'
    },
    round: '4px'
  }
}
```

The theme prop that is passed to the `Grommet` component complains that `side` is incompatible. I get this message
image

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.17.1
- Browser Name and version:

britt6612:
#### What does this PR do?
This PR adds `formfield` theme to the base.d.ts for typescript
theme/base.d.ts
manually 

```
const theme = {
  formField:{
    border: {
      error: {
        color: 'border'
      },
      color: 'border',
      side: 'all'
    },
    round: '4px'
  }
}

```
storybook with a ts story
previously there was `formfield.border` which was set to `bordertype` however this does not include colors. This is why if someone was to use  `formfield.border.color` it would throw an error since `color` string is not  included in the `bordertype`
Also found that `formfield.extend` was missing so I added it in.
closes #5118 
no
no
backwards compatible


src/js/themes/base.d.ts
```diff
@@ -726,7 +726,14 @@ export interface ThemeType {
...
   formField?: {
-    border?: BorderType;
+    border?: {
+      color?: ColorType;
+      position?: BorderType;
+      side?: BorderType;
+      error?: {
+        color?: ColorType;
+      };
+    };
```
ShimiSun, line 736: Wouldn't it be a better practice to extend the `BroderType` types instead of hard coding the prop values?

britt6612: I was inclined to do this but saw other areas that used `border.color` so I can add colors to extend `BorderType` 
Would you want me to change all of the other `border` areas to  be `BorderType` ?

taysea: I think what can be done instead is such:
```
border?: BorderType;
```
BorderType already includes the necessary typings for color, position, etc

britt6612: Your right i was making it to complicated 😓 I just added the `error.color` to the  `border?:`

src/js/themes/base.d.ts
```diff
@@ -726,7 +726,11 @@ export interface ThemeType {
     round?: RoundType;
   };
   formField?: {
-    border?: BorderType;
+    border?: BorderType & {
+      error?: {
+        color?: ColorType;
```
taysea, line 731: Is there anything else that would be on border.error? Or it's just color that we can change

britt6612: https://v2.grommet.io/formfield#formField.border.error.color 
Thats it according to the docs

britt6612: also checked code:
https://github.com/grommet/grommet/blob/master/src/js/components/FormField/FormField.js#L264

taysea (approved):
LGTM!

src/js/themes/base.d.ts
taysea: Great thanks!

ShimiSun (approved):

PR #5131, opened 2021-04-01
Don't let Layer elevation leak to DOM, don't apply elevation when plain

### LINKED ISSUE #5129: Layer with plain set now produces elevation shadow
A Layer with the `plain` property set now puts a box shadow on its container div and the `elevation` property on the container is leaking through to the `<div>` in the DOM.

### Expected Behavior

The plain property should result in no extra border or elevation on the layer.

### Actual Behavior

`<Layer plain>` results in:

image

image

### Steps to Reproduce

 ```
<Layer
        position="bottom"
        modal={false}
        margin={{ vertical: 'medium', horizontal: 'small' }}
        responsive={false}
        plain
  >
       <Text> example layer </Text>
  </Layer>
```
### Your Environment

- Grommet version: 2.17.1

### PR DESCRIPTION
**Pull request by taysea**
#### What does this PR do?
Checks if Layer is plain before applying elevation. Also, used styled-components `withConfig` to ensure that the elevation prop doesn't leak to the DOM.

Tested with this code snippet, no elevation was applied. Updated jest test.

```
<Layer
       position="bottom"
       modal={false}
       margin={{ vertical: 'medium', horizontal: 'small' }}
       responsive={false}
       plain
 >
      <Text> example layer </Text>
 </Layer>
```

With code snippet above, make a theme that has a value for `theme.layer.container.elevation`.

Closes #5129 

Updated here to clarify that Layer `plain` will not apply elevation.

Fixed Layer elevation styling, should not apply when plain.

Backwards compatible.

### DIFF
--- a/src/js/components/Layer/__tests__/__snapshots__/Layer-test.js.snap
+++ b/src/js/components/Layer/__tests__/__snapshots__/Layer-test.js.snap
@@ -1005,7 +1005,6 @@ exports[`Layer custom theme 1`] = `
   <div
     class="c2"
     data-g-portal-id="0"
-    elevation="large"
     id="custom-theme-test"
   >
     <a
--- a/src/js/components/Layer/README.md
+++ b/src/js/components/Layer/README.md
@@ -187,7 +187,8 @@ function
 
 **plain**
 
-Whether this is a plain Layer with no background color or border.
+Whether this is a plain Layer with no background color, border, or 
+        elevation.
 
 ```
 boolean
--- a/src/js/components/Layer/StyledLayer.js
+++ b/src/js/components/Layer/StyledLayer.js
@@ -774,7 +774,12 @@ const elevationStyle = css`
     ]};
 `;
 
-const StyledContainer = styled.div`
+const StyledContainer = styled.div.withConfig({
+  // don't let elevation leak to DOM
+  // https://styled-components.com/docs/api#shouldforwardprop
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

**Review by MikeKingdom — COMMENTED** [review]
Awesome. Thanks Taylor!

File: src/js/components/Layer/__tests__/__snapshots__/Layer-test.js.snap [hunk 1]
```diff
@@ -1005,7 +1005,6 @@ exports[`Layer custom theme 1`] = `
   <div
     class="c2"
     data-g-portal-id="0"
-    elevation="large"
```
ShimiSun, line 1008: Is this line removal from snapshot due to using Layer `plain` on the test?

ShimiSun (commented):
Does this PR fixes a regression or was it always working this way?

src/js/components/Layer/__tests__/__snapshots__/Layer-test.js.snap
taysea: This line removal is due to an existing behavior where elevation was leaking to the DOM but shouldn't.

We did not have an existing test with elevation + plain combination. I can add one to this PR for future.

taysea: Done.

ShimiSun (approved):
Perfect! Thanks!

Issue #5123, opened 2021-03-31
DataTable TypeError when Data Array is empty with `paginate`


TPK2002:
### Expected Behavior
When data array is empty, the table should display no rows but the header.

### Actual Behavior
This perfectly works in most cases, but one. When I set the paginate prop in combination with the onClickRow Prop it crashes on empty data array.

### URL, screen shot, or Codepen exhibiting the issue

Error reproduced in Codesandbox
https://codesandbox.io/s/datatable-empty-data-bug-94i1z
![Bildschirmfoto 2021-03-31 um 11 15 26](https://user-images.githubusercontent.com/19915810/113120943-6797cb80-9212-11eb-8a5c-a86564cce363.png)

### Steps to Reproduce

1. SetUp a React App
2. Install Grommet
3. Create a DataTable
4. Set data to '[]'
5. Set the paginate and onClickRow Props

### Your Environment

- Grommet version: 2.17.0
- Browser Name and version: Safari/Firefox/Chrome (all newest versions)
- Operating System and version (desktop or mobile): macOS 11

ShimiSun:
For troubleshooting, I think the root cause of this issue is the `paginate` prop, as this functionality seems to work in the past.

PR #5135, opened 2021-04-02
Pagination utility to return empty array of current items if data is passed in as array

taysea:
#### What does this PR do?
This PR updates the pagination utility function to review an empty array `[]` when the data passed in is an array but there are no current items. It also updates the function to `useMemo` instead of `useCallback` to avoid an infinite loop that could occur in the `useEffect` `useState` combination.

This fixes an issue with DataTable using `paginate` + `onClickRow` because DataTable relies on data being defaulted to `[]` however usePagination was returning back `undefined` even when the data passed in was an empty array. This ensures that usePagination returns a value that is the same type as the data it received.

src/js/utils/pagination.js

Tested in storybook by adjusting Paginated DataTable with:
```
      <DataTable
        columns={[
          {
            property: "a",
            header: "a"
          },
          {
            property: "b",
            header: "b"
          }
        ]}
        data={[]}
        paginate
        onClickRow={() => {}}
      />
```
Also added jest test for the combination of `paginate` and empty data array to ensure that the pagination controls don't render.

Same as above.

DataTable render relies on fact that data will default to `[]` but `usePagination` was return data as `undefined` which caused failed render.

Closes https://github.com/grommet/grommet/issues/5123

No.

Fixed error that occurred when `paginate` was used with `onClickRow` on DataTable.

Backwards compatible.

### DIFF
--- a/src/js/components/DataTable/DataTable.js
+++ b/src/js/components/DataTable/DataTable.js
@@ -406,7 +406,7 @@ const DataTable = ({
           )}
         </StyledDataTable>
       </OverflowContainer>
-      {paginate && items.length ? (
+      {paginate && items && items.length ? (
         <Pagination alignSelf="end" {...paginationProps} />
       ) : (
         undefined
--- a/src/js/components/DataTable/DataTable.js
+++ b/src/js/components/DataTable/DataTable.js
@@ -408,9 +408,7 @@ const DataTable = ({
       </OverflowContainer>
       {paginate && items && items.length ? (
         <Pagination alignSelf="end" {...paginationProps} />
-      ) : (
-        undefined
-      )}
+      ) : null}
     </Container>
   );
 };
--- a/src/js/components/DataTable/DataTable.js
+++ b/src/js/components/DataTable/DataTable.js
@@ -406,7 +406,7 @@ const DataTable = ({
           )}
         </StyledDataTable>
       </OverflowContainer>
-      {paginate && items && items.length ? (
+      {paginate && data.length > step && items && items.length ? (
         <Pagination alignSelf="end" {...paginationProps} />
       ) : null}
     </Container>
--- a/src/js/components/DataTable/DataTable.js
+++ b/src/js/components/DataTable/DataTable.js
@@ -406,7 +406,11 @@ const DataTable = ({
           )}
         </StyledDataTable>
       </OverflowContainer>
-      {paginate && items && <Pagination alignSelf="end" {...paginationProps} />}
+      {paginate && items.length ? (
+        <Pagination alignSelf="end" {...paginationProps} />
+      ) : (
+        undefined
+      )}
     </Container>
   );
 };
--- a/src/js/utils/pagination.js
+++ b/src/js/utils/pagination.js
@@ -20,10 +20,8 @@ export const usePagination = ({ data, page, step, ...rest }) => {
   const itemsEndIndex = itemsBeginIndex + step;
 
   const currentItems = useMemo(() => {
-    if (Array.isArray(data)) {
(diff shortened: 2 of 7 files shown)


src/js/components/DataTable/DataTable.js
```diff
@@ -406,7 +406,11 @@ const DataTable = ({
           )}
         </StyledDataTable>
       </OverflowContainer>
-      {paginate && items && <Pagination alignSelf="end" {...paginationProps} />}
+      {paginate && items && items.length ? (
+        <Pagination alignSelf="end" {...paginationProps} />
+      ) : (
+        undefined
+      )}
```
halocline, line 413: This would also be valid and result the same:

```suggestion
      {paginate && items && !!items.length && (
        <Pagination alignSelf="end" {...paginationProps} /> 
      )}
```

Or by convention, I think Grommet has tended to use `null` instead of `undefined` in these circumstances.

```suggestion
      {paginate && items && items.length ? (
        <Pagination alignSelf="end" {...paginationProps} />
      ) : null}
```

halocline (approved):

src/js/components/DataTable/DataTable.js
taysea: Thanks! I'll update to `null` to align with common Grommet convention.

src/js/components/DataTable/DataTable.js
```diff
@@ -406,7 +406,9 @@ const DataTable = ({
           )}
         </StyledDataTable>
       </OverflowContainer>
-      {paginate && items && <Pagination alignSelf="end" {...paginationProps} />}
+      {paginate && items && items.length ? (
```
ericsoderberghp, line 409: I think this might cause a jump in vertical layout if the table has a fixed height but is lazily loaded. I wonder if instead Pagination should treat no items the same a single item, with just one page and disabled controls.

src/js/utils/pagination.js
```diff
@@ -18,24 +18,13 @@ export const usePagination = ({ data, page, step, ...rest }) => {
...
-    [itemsBeginIndex, itemsEndIndex],
-  );
-  const [currentItems, setCurrentItems] = useState(getCurrentItems(data));
 
-  useEffect(() => {
-    if (data) {
-      setCurrentItems(getCurrentItems(data));
+  const currentItems = useMemo(() => {
+    if (Array.isArray(data)) {
+      return data.length ? data.slice(itemsBeginIndex, itemsEndIndex) : [];
```
ericsoderberghp, line 24: if data is an array of length 0, wont' `data.slice(...)` return `[]` already?

src/js/utils/pagination.js
```diff
@@ -18,24 +18,13 @@ export const usePagination = ({ data, page, step, ...rest }) => {
...
     }
-  }, [data, getCurrentItems, setCurrentItems]);
+    return data;
```
ericsoderberghp, line 26: Should this instead return `[]` when data isn't an array?

src/js/utils/pagination.js
taysea: You're right. I think we can remove the check for length now

src/js/utils/pagination.js
taysea: Yes I think that would be a safe return. No errors were being caused because the only case this return was being called was when List provided `data` as `undefined`, but I think for consistency it would make sense for pagination to return an empty array for currentItems.

src/js/components/DataTable/DataTable.js
taysea: Summarizing offline discussion: Pagination controls should only show if the number of items is great than the step (meaning that there would be more than one page).

ericsoderberghp (approved):

PR #5137, opened 2021-04-02
Ensure aria-label can be used anywhere a11yTitle can.

kid-icarus:
This is just applying the same logic that was approved in https://github.com/grommet/grommet/pull/5119 to all components using a11yTitle. I'm a little curious about the history of this a11yTitle prop, and in my opinion it should be deprecated. There is already an DOM attribute for this specific purpose and I'm not understanding the value of renaming it, but I'm sure there was a good reason to do so.

#### What does this PR do?

Ensures aria-label can be used anywhere a11yTitle can.

Unit tests / snapshots should be the only thing necessary here.

Ensure that all components modified render an aria-label via either a11yTitle or aria-label props

This is just applying the same logic that was approved in https://github.com/grommet/grommet/pull/5119 to all components using a11yTitle. 

I don't believe so

It may make sense to announce it, it is a new feature for the components where we're not spreading props after setting aria-label

It is backwards compatible


src/js/components/CheckBox/CheckBox.js
```diff
@@ -192,7 +193,7 @@ const CheckBox = forwardRef(
 
     return (
       <StyledCheckBoxContainer
-        aria-label={a11yTitle}
+        aria-label={ariaLabel || a11yTitle}
```
ShimiSun, line 196: I think `aria-label` will work transparently via the `...rest` spread. So I don't think we need an explicit attention for that.

src/js/components/DropButton/DropButton.js
```diff
@@ -10,6 +10,7 @@ const DropButton = forwardRef(
   (
     {
       a11yTitle = 'Open Drop',
+      'aria-label': ariaLabel,
```
ShimiSun, line 13: I think the `aria-label` will work transparently via the `...rest` spread. So I don't think we need explicit attention for that.

src/js/components/Image/Image.js
```diff
@@ -2,7 +2,19 @@ import React, { forwardRef, useState } from 'react';
 import { StyledImage } from './StyledImage';
 
 const Image = forwardRef(
-  ({ a11yTitle, fallback, onError, opacity, fill, src, ...rest }, ref) => {
+  (
+    {
+      a11yTitle,
+      'aria-label': ariaLabel,
```
ShimiSun, line 8: I think the `aria-label` will work transparently via the `...rest` spread. So I don't think we need explicit attention for that.

src/js/components/Menu/Menu.js
```diff
@@ -236,7 +237,7 @@ const Menu = forwardRef((props, ref) => {
           // make it accessible at the end of all menu items
           buttonRefs[items.length] = r;
         }}
-        a11yTitle={a11yTitle || messages.closeMenu || 'Close Menu'}
+        a11yTitle={ariaLabel || a11yTitle || messages.closeMenu || 'Close Menu'}
```
ShimiSun, line 240: I think that with your recent Button fix, the `aria-label` would work automatically, so I don't think we need to explicitly handle it.

src/js/components/Pagination/Pagination.js
```diff
@@ -179,7 +180,10 @@ const Pagination = forwardRef(
 
     return (
       <StyledPaginationContainer {...theme.pagination.container} {...rest}>
-        <Nav a11yTitle={a11yTitle || 'Pagination Navigation'} ref={ref}>
+        <Nav
+          a11yTitle={ariaLabel || a11yTitle || 'Pagination Navigation'}
```
ShimiSun, line 184: Box is handling `aria-label` properly, and Nav is using Box. Can you please double-check if the explicit handling here is really needed?

src/js/components/Select/Select.js
```diff
@@ -88,6 +89,7 @@ const Select = forwardRef(
     },
     ref,
   ) => {
+    const label = ariaLabel || a11yTitle;
```
ShimiSun, line 92: I think the `aria-label` will work automatically via the `...rest` spread of TextInput. So I don't think we need explicit mention of `aria-label` in this case.

ShimiSun (commented):
Thank you for this PR and for helping us improve the grommet's accessibility!
Dropped a few comments, let us know what you think.

ShimiSun:
@kid-icarus hi buddy, thanks for the PR! I was hoping we can get it in our coming release. 
Can you please incorporate the feedback or let us know what you think? Thanks!

kid-icarus:
@ShimiSun Sure thing! I've been a little preoccupied with work, will try to tackle this tonight :)

src/js/components/CheckBox/CheckBox.js
jcfilben: StyledCheckBoxContainer doesn't have `{...rest}` on it

src/js/components/Pagination/Pagination.js
jcfilben: Without explicit handling the aria-label would be placed on the div around the nav instead of on the nav like it would be if you were using a11yTitle

ericsoderberghp:
Due to inactivity, I'm thinking we should split this up into separate pull request for individual components.

taysea:
Closing since these changes were handled in #5444

PR #5138, opened 2021-04-03
Add default title for truncated Text.

### LINKED ISSUE #5132: Enhance Text with truncate to use Tip when truncating text

kid-icarus:
This is a minor UX improvement. It ensures that truncated text is
still visible if the user hovers over it. The title can be overridden by
the consumer via the props spread at the end of the component. This also
helps with accessibility in the case that text is truncated without the
consumer's knowledge.

#### What does this PR do?

Adds default title text.

The Text component

Updated snapshot, manually verified the title comes through on truncated text and is read by screen readers, and displayed as a native tooltip.
 
See above

At work we were manually doing this in a Text wrapper and thought we'd contribute it back.

Closes #5132 

Potentially, we might want to add a note there.

I don't see why not

It is backwards compatible. Due to the fact that we spread props at the end of this component, if a consumer passes a title, it will override the default.

Additionally, if text is trunctated, but the consumer passes in an `aria-label` or `a11yTitle`, screen readers give precedence to `aria-label` over `title`


src/js/components/Text/Text.js
```diff
@@ -2,15 +2,27 @@ import React, { forwardRef } from 'react';
...
-  />
-));
+const Text = forwardRef(
+  ({ color, tag, as, a11yTitle, truncate, children, ...rest }, ref) => {
+    const conditionalProps = {
+      truncate,
+      children,
+    };
+    if (truncate) {
+      conditionalProps.title = children;
```
ShimiSun, line 12: I think we would like to use the Tip component instead of the plain HTML title, this way we'll maintain the grommet's look & feel consistency.

ShimiSun (commented):
Thank you for starting the work and discussion! 💟 
I dropped a comment, let me know what you think.

ShimiSun:
Also, please take a look at this recent issue https://github.com/grommet/grommet/issues/5132

kid-icarus:
Hi @ShimiSun, thanks for the feedback!

I took a look at the issue in question and reworked my PR to use a Tip, along with an optional tip prop for the Text component. My work here isn't yet complete, as there are a couple of outstanding issues:

1. Accessibility - In terms of accessibility, I went through a story in storybook with the truncated text and a tip. The screenreader ends up reading both the text and the tip, making for a confusing user experience.

2. Developer experience - It seems strange to have to specify the tip content here, and in my opinion, we should default the tip's content to its children, since we know the children _should_ just be text (i.e. there are no child nodes inside Text other than a nHTML TextNode)

I still need to write unit tests, and I created a storybook story (largely copy/paste from Button) around the tooltip, which I need to clean up a bit, but I just wanted to move the discussion forward with some code.

Let me know what you think about my points and I'll happily proceed forward :)

britt6612:
@kid-icarus Thank you for your explanation!

I agree with the developer experience `tip` should `if text is truncated, show a tooltip on hover that displays the full text.` 

Accessibility: You make a good point by being able to default the a11yTitle/aria-label to the tip contents. Did you by change test if a user passed in something like 
```
        content={
          <Box align="center">
            <Text>Hello</Text>
          </Box>
        }

``` 
into `Tip`?

Also PR looks great can you add the new  `Tip` prop to `Text` docs ? 

```
    tip: PropTypes.oneOfType([
      PropTypes.shape({
        content: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
        dropProps: PropTypes.shape({}),
        plain: PropTypes.bool,
      }),
      PropTypes.string,
    ]).description(`tooltip when text is truncated.`),

```

ShimiSun:
@britt6612 following the offline discussion, can you please add visuals that compare the native `title behavior with the actual `Tip` behavior.

britt6612:
The first images are with using `Tip`
Screen Shot 2021-04-06 at 10 41 15 AM
Screen Shot 2021-04-06 at 10 41 10 AM

Here is without `Tip` 
This does take a little while to show up in my opinion 

https://user-images.githubusercontent.com/42451602/113749158-8db3e480-96c6-11eb-8910-114feeb1fed1.mov

ShimiSun:
@britt6612 thanks for the quick turnaround, yea, I think it will look more consistent to use the grommet's Tip and that will be the better direction to follow.

britt6612:
> @britt6612 thanks for the quick turnaround, yea, I think it will look more consistent to use the grommet's Tip and that will be the better direction to follow.

Yes sorry about that totally spaced it out yesterday! But I agree I think `Tip` is more consistent 👍

src/js/components/Text/index.d.ts
```diff
@@ -35,6 +37,7 @@ export interface TextProps {
   truncate?: boolean;
   weight?: 'normal' | 'bold' | number;
   wordBreak?: 'normal' | 'break-all' | 'keep-all' | 'break-word';
+  tip?: TipProps;
```
ShimiSun, line 40: Great! Thank you! Can you please also add this new prop to Text/doc.js? 
It will require updating our README and snapshots, so please use `yarn test-update` and push the new files up the branch.

src/js/components/Text/stories/TipOnText.js
```diff
@@ -0,0 +1,53 @@
...
+        a b c d e f g h i j k l m n o p q r s t u v w x y z
+      </Text>
+    </Box>
+    <Box align="center" pad="medium" gap="large">
+      <Text
+        label="Tip Drop props"
+        onClick={() => {}}
+        tip={{ dropProps: { align: { left: 'right' } }, content: 'tooltip' }}
+      >
+        Tip Drop props
```
ShimiSun, line 18:
```suggestion
        Tip with dropProps
```

src/js/components/Text/stories/TipOnText.js
```diff
@@ -0,0 +1,53 @@
...
+              margin="xsmall"
+              overflow="hidden"
+              align="center"
+            >
+              tooltip
+            </Box>
+          ),
+        }}
+      >
+        Tip Content
```
ShimiSun, line 39:
```suggestion
        Tip with content prop
```

src/js/components/Text/stories/TipOnText.js
```diff
@@ -0,0 +1,53 @@
...
+      </Text>
+    </Box>
+  </Grommet>
+);
+
+TipOnText.storyName = 'Tip on text';
```
ShimiSun, line 45:
```suggestion
TipOnText.storyName = 'Tip';
```

src/js/components/Text/stories/TipOnText.js
```diff
@@ -0,0 +1,53 @@
...
+
+TipOnText.parameters = {
+  chromatic: { disable: true },
```
ShimiSun, line 48: Good call!

src/js/components/Text/stories/TipOnText.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+
+import { grommet, Box, Text, Grommet } from 'grommet';
+
+export const TipOnText = () => (
+  <Grommet theme={grommet}>
+    <Box align="center" width="small" pad="small">
+      <Text truncate tip="a b c d e f g h i j k l m n o p q r s t u v w x y z">
```
ShimiSun, line 8: This Text isn't aligned to the center and it looks inconsistent with the below tip examples.
Feel free to align all the examples to start (i.e. remove the align prop since its default to 'start' anyway)

src/js/components/Text/stories/TipOnText.js
```diff
@@ -0,0 +1,53 @@
+import React from 'react';
+
+import { grommet, Box, Text, Grommet } from 'grommet';
+
+export const TipOnText = () => (
+  <Grommet theme={grommet}>
+    <Box align="center" width="small" pad="small">
```
ShimiSun, line 7: You might want to increase the vertical pad so the tooltip won't overlap with the below example.

ShimiSun:
Hi buddy @kid-icarus, thanks for working on this PR, I think it is really close, would you like to continue working on it so we can get it in our coming release?

kid-icarus:
@ShimiSun Sure thing! I've been a little preoccupied with work, will try to tackle this tonight :)

ericsoderberghp (approved):

ShimiSun (approved):

ShimiSun:
Thank you for your contribution @kid-icarus! And thank you @taysea for getting this PR to the finish line 💟

Issue #5076, opened 2021-03-17
Range Selector not draggable on mobile


leifg:
### Expected Behavior

When I have a range selector that I use on a mobile device I expect to drag the endpoints like I would with a mouse curser

### Actual Behavior

On a mobile device (I tested with an iPhone) I am not able to drag the endpoints. Instead I tap where I want the individual endpoints to be.

### URL, screenshot, or Codepen exhibiting the issue

- [Codepen example](https://codesandbox.io/s/grommet-v2-template-forked-ybb73?file=/index.js)
- [Slack Discussion](https://grommet.slack.com/archives/C09QQEG69/p1615942603000700)
- [Video outlining the issue](https://user-images.githubusercontent.com/110321/111541494-ce56b880-872d-11eb-9184-96cf1d7d878b.MOV)

### Steps to Reproduce

- Go to codeine example with your phone
- try to drag the endpoints to adjust the range
- nothing happens

### Your Environment

- Grommet version: 2.16.2
- Safari on IOs
- iOS 14.4.1

ShimiSun:
Thank you for reporting, the issue is fixed and will be available on the following release of ^2.17.2.

PR #5139, opened 2021-04-04
Fix RangeSelector to respond to touch gestures

jeferson-sb:
#### What does this PR do?

Fix RangeSelector change events to able to respond to mobile touch gestures for when the user drag on of the edge points

`src/js/components/RangeSelector/RangeSelector.js`

`yarn storybook`, `yarn test` and test on a real device

`yarn storybook`, go to a range selector story and then enabling device emulation on browser

See #5076 

range-selector

No

backwards compatible

### DIFF
--- a/src/js/components/RangeSelector/RangeSelector.js
+++ b/src/js/components/RangeSelector/RangeSelector.js
@@ -68,8 +68,8 @@ const RangeSelector = forwardRef(
       [direction, max, min, step],
     );
 
-    useEffect(() => {
-      const mouseMove = event => {
+    const onMouseMove = useCallback(
+      event => {
         const value = valueForMouseCoord(event);
         let nextValues;
         if (changing === 'lower' && value <= values[1] && value !== moveValue) {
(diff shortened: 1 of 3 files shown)


src/js/components/RangeSelector/__tests__/RangeSelector-test.js
```diff
@@ -209,4 +209,40 @@ describe('RangeSelector', () => {
...
+  test('handle touch gestures', () => {
+    const onChange = jest.fn();
+    const { container, getByLabelText } = render(
+      <Grommet>
+        <RangeSelector values={[10, 20]} onChange={onChange} />
+      </Grommet>,
+    );
+    const rangeContainer = container.firstChild.firstChild;
+
+    expect(container.firstChild).toMatchSnapshot();
```
jcfilben, line 222: Should the `expect(container.firstChild).toMatchSnapshot();` go after the fireEvent stuff at the end of the test? Otherwise not sure what the snapshot is really capturing since it is being taken before any of the touch events happen.

jcfilben (commented):
Looks good! I tested with storybook and it is working with touch gestures, code looks good too. Also thanks for adding a unit test as well! I just left a comment about the test below

src/js/components/RangeSelector/__tests__/RangeSelector-test.js
jeferson-sb: hmm yeah, that makes more sense

jcfilben (approved):
Looks good, on mobile you still can't drag the entire RangeSelector to reposition it like you can on desktop but being able to drag the endpoints on mobile is a big improvement. I think being able to drag the entire RangeSelector on mobile is a separate issue though, so approving this PR.

src/js/components/RangeSelector/RangeSelector.js
```diff
@@ -153,6 +173,7 @@ const RangeSelector = forwardRef(
         {...rest}
         tabIndex="-1"
         onClick={onChange ? onClick : undefined}
+        onTouchMove={onTouchMove}
```
ericsoderberghp, line 176:
```suggestion
        onTouchMove={onChange ? onTouchMove : undefined}
```

ericsoderberghp (commented):
Looking good. Just one suggestion, which would eliminate the snapshot changes.

ericsoderberghp:
I think you need to now re-run the tests to update the snapshots.

PR #5140, opened 2021-04-04
Make DataTable's groupBy togglable

watiko:
#### What does this PR do?

Make DataTable's groupBy togglable.

`DataTable.js`

Check if the error below was not thrown when toggle `groupBy` props between `undefined` and `"a"`.

>Uncaught [TypeError: Cannot read property 'expanded' of undefined]

### DIFF
--- a/src/js/components/DataTable/Body.js
+++ b/src/js/components/DataTable/Body.js
@@ -204,7 +204,7 @@ const Body = forwardRef(
               const isRowExpanded = rowExpand && rowExpand.includes(index);
               return (
                 <Row
-                  key={primaryValue || index}
+                  key={index}
                   rowRef={rowRef}
                   primaryValue={primaryValue}
                   isSelected={isSelected}
--- a/src/js/components/DataTable/DataTable.js
+++ b/src/js/components/DataTable/DataTable.js
@@ -53,12 +53,24 @@ const normalizeProp = (prop, context) => {
   return undefined;
 };
 
+// https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is
+const objectIs = (x, y) => {
+  if (x === y) {
+    return x !== 0 || 1 / x === 1 / y;
+  } else {
+    return x !== x && y !== y;
+  }
+};
+
+const equals = (x, y) =>
+  typeof Object.is === 'function' ? Object.is(x, y) : objectIs(x, y);
+
 const useStateWithDeps = (initialStateFn, deps) => {
   const [state, setState] = useState(initialStateFn);
   const [prevDeps, setPrevDeps] = useState(deps);
 
   for (let i = 0; i < deps.length; i++) {
-    if (prevDeps[i] !== deps[i]) {
+    if (!equals(prevDeps[i], deps[i])) {
       setPrevDeps(deps);
       const nextState = initialStateFn();
       setState(nextState);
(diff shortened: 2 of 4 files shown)


src/js/components/DataTable/Body.js
```diff
@@ -204,7 +204,7 @@ const Body = forwardRef(
               const isRowExpanded = rowExpand && rowExpand.includes(index);
               return (
                 <Row
-                  key={primaryValue || index}
```
watiko, line 207: `primaryValue` is not always unique depending on what is passed.

ShimiSun:
Hi buddy, thanks for this PR!
We are missing some context about the original issue and what this PR is trying to solve, can you please elaborate?

watiko:
@ShimiSun Here is the context.

I has an app with a togglable table component using (old) Grommet's DataTable and the table component can be toggle between plain state (`groupBy={undefined}`) and grouped by state(`groupBy="some column name"`).
When I was trying to update Grommet's version, I had a trouble about the application's crash by toggling table state.

>Uncaught [TypeError: Cannot read property 'expanded' of undefined]

After some investigations, I found that the root cause of the problem. `groupState` can be inconsistent with `groups` and `groupBy`.

ShimiSun:
@watiko Can you please share a reproducible code of the issue you are describing? It will help us understand your fix better. Here is a [template](https://codesandbox.io/s/m7mml8l0zj)

watiko:
@ShimiSun Here is [the reproducible code](https://codesandbox.io/s/grommet-v2-template-forked-njh96?file=/index.js)

```javascript
function CrashOnToggling() {
  const [groupBy, setGroupBy] = React.useState();
  const toggle = () => setGroupBy(groupBy === undefined ? "a" : undefined);

  return (
    <Grommet>
      <button type="button" onClick={toggle}>
        toggle
      </button>
      <DataTable
        columns={[
          { property: "a", header: "A" },
          { property: "b", header: "B" }
        ]}
        data={[
          { a: "one", b: 1.1 },
          { a: "one", b: 1.2 },
          { a: "two", b: 2.1 },
          { a: "two", b: 2.2 }
        ]}
        groupBy={groupBy}
      />
    </Grommet>
  );
}
```

Naguumo:
I have been having this same issue with my use case and would love to see more attention from the team on this issue.

src/js/components/DataTable/DataTable.js
```diff
@@ -53,6 +53,34 @@ const normalizeProp = (prop, context) => {
...
+  if (x === y) {
+    return x !== 0 || 1 / x === 1 / y;
+  }
+  return x !== x && y !== y; // eslint-disable-line no-self-compare
+};
+
+const equals = (x, y) =>
+  typeof Object.is === 'function' ? Object.is(x, y) : objectIs(x, y);
+
+const useStateWithDeps = (initialStateFn, deps) => {
```
ericsoderberghp, line 67: This approach feels a bit too generic-ish. I'm wondering instead if we should create a more specialized hook called something like `useGroupState` which pulls `buildGroupState` inside it and then has an effect that updates the state when `groups` or `groupBy` change.

watiko: Thank you for your review. I made `useGroupState` hook. Would you review this?

ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution!

Issue #3642, opened 2020-01-03
Hide focus outline


steventilator:
Sometimes the focus outline stays visible after clicking on a link/button/anchor (see screenshot). This happens when navigating by mouse and touch, not keyboard. It happens on Chrome but not on Safari and Firefox.

Is there a way to hide the outline?

<img width="258" alt="focus-outline" src="https://user-images.githubusercontent.com/39826675/71725074-f0b72b80-2e32-11ea-9bcf-ff24459c5220.png">

- Grommet version: 2.9.0
- Browser Name and version: Chrome 79.0.3945.88 
- Operating System and version (desktop or mobile): Various

steventilator:
Okay, I realized that I can set `focusIndicator={false}` to hide it. Is there a way to set this on every Button or Box? Or do I have to add it manually to every instance?

CR1AT0RS:
I agree focus is also not consistent across Mobile UI so I think it should be optional.

iMerica:
TLDR from Slack Discussion:

`focusIndicator` needs to be false by default and it needs to update its definition as "focus indicator that displays after clicking something". And add a separate capability for displaying a focus indicator when tabbing through the page.

srallen:
For accessibility, it's actually best to typically have the styles for the click, hover, and focus events to be the same both for maintainability and predictability for users who use a variety of assistive tools. Disabling a focus indicator and or separating the styles between keyboard and mouse events inadvertently create less accessible components. Any configurability supported here should be done with great caution. 

I would agree that there's a bug with the focus outline persisting after the click event, but I don't think the solution is to necessarily allow devs to configure this to potentially make less accessible components.

srallen:
Caution should be taken when separating the focus styles between click and keyboard events as this may have unintended side effects for screen readers:

> Some browser/screen reader combinations fire mouse events, which could cause outlines to disappear while using this method.

https://a11yproject.com/posts/never-remove-css-outlines/

taysea:
Agreed, I don't think we should separate the focusIndicator into multiple props for mouse vs keyboard. I think it would be best to look into why the focus indicator persists after clicking.

iMerica:
> Disabling a focus indicator and or separating the styles between keyboard and mouse events inadvertently create less accessible components

Grommet is creating less accessible components because it leaves us with only one easy way to remove this on/after click outline: `focusIndicator={false}`.  This is having the unintended consequence of removing the outline for every type of event.

[...]

Issue #2785, opened 2019-02-18
focus on text area


ShimiSun:
c/p from @alansouzati 
```
The focus outline is showing if I click in the textinput
IMHO the focus outline should only show if you are using the tab

if I use grommet 2.3.1 this problem goes away
so for sure was something that changed from 2.3.1 to 2.4.0
```
https://codesandbox.io/s/github/grommet/grommet-sandbox?initialpath=textinput&module=%2Fsrc%2FTextInput.js

ShimiSun:
contradicts #2687

atanasster:
```
The focus outline is showing if I click in the textinput
IMHO the focus outline should only show if you are using the tab
```
I would disagree - the focus outline is an indicator of where the text input will go, regardless of how the focus was achieved (by mouse of by keyboard). 

Any real world examples where its implemented the way described in this issue (no outline if TextInput focused by mouse click)?

oorestisime:
yeah i also think it makes sense to have the same behavior on mouse click

ericsoderberghp:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input sets focus on click of `<input />`,

ShimiSun:
I'll be closing the issue if there is no objection.

MikeWalrath:
Closing. Please feel free to re-open if further discussion is needed.

alansouzati:
Take this for example: https://codesandbox.io/s/rrjz94lj7m

If you open the filter on the left and start selecting things, you will see the outline. this is a regression.

The outline is very distracting and I feel we should only show this if we are using the keyboard.

PR #5148, opened 2021-04-06
fix: only show focus on keyboard use

alansouzati:
Inspired by [this](https://nelo.is/writing/styling-better-focus-states/) I'm adding support for `:focus-visible` if available so that we only show the focus ring if you are using the keyboard

#### What does this PR do?

Show focus only when using keyboard

StyledButton

manual

Make sure you see the focus outline when using the keyboard and you dont see it when clicking on buttons

nope
yes
backwards compatible

### DIFF
--- a/src/js/components/Button/StyledButton.js
+++ b/src/js/components/Button/StyledButton.js
@@ -5,6 +5,7 @@ import {
   backgroundStyle,
   disabledStyle,
   focusStyle,
+  unfocusStyle,
   genericStyles,
   getHoverIndicatorStyle,
   normalizeColor,
@@ -206,9 +207,13 @@ const StyledButton = styled.button`
     props.theme.button &&
     props.theme.button.disabled &&
     disabledButtonStyle(props)}
-  
+
   &:focus {
     ${props => (!props.plain || props.focusIndicator) && focusStyle()}
+  }
+
+  &:focus:not(:focus-visible) {
+    ${unfocusStyle()}
   }
 
   ${props =>
--- a/src/js/components/Accordion/__tests__/__snapshots__/Accordion-test.js.snap
+++ b/src/js/components/Accordion/__tests__/__snapshots__/Accordion-test.js.snap
@@ -173,6 +173,26 @@ exports[`Accordion AccordionPanel 1`] = `
   border: 0;
 }
 
+.c3:focus:not(:focus-visible) {
+  outline: none;
+  box-shadow: none;
+}
+
+.c3:focus:not(:focus-visible) > circle,
+.c3:focus:not(:focus-visible) > ellipse,
+.c3:focus:not(:focus-visible) > line,
+.c3:focus:not(:focus-visible) > path,
+.c3:focus:not(:focus-visible) > polygon,
+.c3:focus:not(:focus-visible) > polyline,
+.c3:focus:not(:focus-visible) > rect {
+  outline: none;
+  box-shadow: none;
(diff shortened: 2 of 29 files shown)


src/js/components/Button/StyledButton.js
```diff
@@ -206,11 +207,15 @@ const StyledButton = styled.button`
...
     props.theme.button.disabled &&
     disabledButtonStyle(props)}
-  
+
   &:focus {
     ${props => (!props.plain || props.focusIndicator) && focusStyle()}
   }
 
+  &:focus:not(:focus-visible) {
+    ${unfocusStyle()}
```
taysea, line 216: When I click on a (such as the one with label "Default")  Button in [this story](https://deploy-preview-5148--sad-tereshkova-173d07.netlify.app/?path=/story/controls-button-basic--basic), I don't get the hover behavior unless I click outside the button (somewhere on the page). I would expect that I would still get hover behavior again after clicking.

Also, would we want to apply similar styling to StyledButtonKind so the same focus behavior applies to buttons with themes relying on `button.default`?

alansouzati: cool, let me look where we apply the focus styling elsewhere. In theory, everywhere we use `:focus` we could augment it with the logic proposed in this PR.

I'm not sure I follow your first comment though. I would not expect this PR to affect the hover behavior at all. I'm only changing the focus behavior when using the mouse. Do you mind sharing more details?

alansouzati: FYI I've updated this PR with the other components that needed the `:focus-visible` extension. I've skip input components like TextInput and FileInput.

taysea: Summarizing from offline discussion: It seems that grommet does not apply `hover` styling if an element is in focus (existing behavior). I am wondering if a separate enhancement would be to always apply hover styling when moused over (whether in focus or not). This would resolve the behavior I noted in my initial comment.

src/js/components/Button/StyledButton.js
```diff
@@ -5,6 +5,7 @@ import {
   backgroundStyle,
   disabledStyle,
   focusStyle,
+  unfocusStyle,
```
taysea, line 8: I think this functionality looks great and works as expected. Wondering if we should update the comment above `unfocusStyle` to explain why this is a valid use case of the function: https://github.com/grommet/grommet/blob/master/src/js/utils/styles.js#L331

May be helpful add the link to the blog article you referenced in the comment, it was a helpful read: https://nelo.is/writing/styling-better-focus-states/

alansouzati: good idea! I will update the comment accordingly

alansouzati: i forgot to update the PR. ive just now changed the unfocusStyle to include the new description. sorry about the delay

ShimiSun (approved):
I think this PR is a great way to clear out some of the focus concerns we ran into the last few months. Thank you for your contribution!

I've verified that all components that apply focus will now have the alternative of unfocusing when the user is navigating through the mouse and not via keyboard. 
Also, this PR behavior is recommended as the more accessible approach for focus handling. Thanks again!

ShimiSun:
This PR closes https://github.com/grommet/grommet/issues/3642

ShimiSun:
Closes https://github.com/grommet/grommet/issues/2785

PR #5153, opened 2021-04-08
add new fill prop to checkbox

### LINKED ISSUE #5001: Feature request: label and toggle alignment

britt6612:
#### What does this PR do?
Orginal work was done [here](https://github.com/grommet/grommet/pull/5019) 

This PR adds a new `fill` prop to `checkbox` 
When a user includes the `fill` prop the box containing the label and checkbox/toggle will fill the width and set `justify-content:  space-between`
styledcheckbox.js
storybook and added a test for using the fill prop
```

        <FormField
          label="Toggle fill"
          name="toggle"
          htmlFor="check-box-toggle"
          {...props}
        >
          <Box pad={{ horizontal: 'small', vertical: 'xsmall' }}>
            <CheckBox
              id="check-box-toggle"
              name="toggle"
              label="CheckBox"
              toggle
              fill
              reverse
            />
          </Box>
        </FormField>
```
feature request: see below
Screen Shot 2021-04-08 at 9 42 29 AM

closes  #5001 
already done
yes
backwards compatible

### DIFF
--- a/src/js/components/CheckBox/CheckBox.js
+++ b/src/js/components/CheckBox/CheckBox.js
@@ -33,6 +33,7 @@ const CheckBox = forwardRef(
       checked: checkedProp,
       defaultChecked = false,
       disabled,
+      fill,
       focus: focusProp,
       id,
       label,
@@ -194,6 +195,7 @@ const CheckBox = forwardRef(
     return (
       <StyledCheckBoxContainer
         aria-label={a11yTitle}
+        fillProp={fill}
         reverse={reverse}
         {...removeUndefined({ htmlFor: id, disabled })}
         checked={checked}
--- a/src/js/components/CheckBox/CheckBox.js
+++ b/src/js/components/CheckBox/CheckBox.js
@@ -195,7 +195,7 @@ const CheckBox = forwardRef(
     return (
       <StyledCheckBoxContainer
         aria-label={a11yTitle}
-        fillProp={fill}
+        fill={fill}
         reverse={reverse}
         {...removeUndefined({ htmlFor: id, disabled })}
         checked={checked}
--- a/src/js/components/CheckBox/StyledCheckBox.js
+++ b/src/js/components/CheckBox/StyledCheckBox.js
@@ -3,6 +3,18 @@ import styled, { css } from 'styled-components';
 import { edgeStyle, focusStyle, normalizeColor } from '../../utils';
 import { defaultProps } from '../../default-props';
 
+const fillStyle = fillContainer => {
+  if (fillContainer) {
+    return `
+      width: 100%;
+      height: 100%;
+      max-width: none;
+      flex: 1 0 auto;
+    `;
+  }
+  return undefined;
+};
+
 const disabledStyle = `
   opacity: 0.5;
   cursor: default;
@@ -36,7 +48,11 @@ const StyledCheckBoxContainer = styled.label`
   flex-direction: row;
   align-items: center;
   user-select: none;
-  width: fit-content;
+  ${props =>
+    props.fillProp &&
+    fillStyle(props.fillProp) &&
+    `justify-content: space-between;`} ${props =>
+  !props.fillProp && 'width: fit-content;'}
   ${props =>
     (props.pad || props.theme.checkBox.pad) &&
     edgeStyle(
--- a/src/js/components/CheckBox/StyledCheckBox.js
+++ b/src/js/components/CheckBox/StyledCheckBox.js
@@ -3,16 +3,13 @@ import styled, { css } from 'styled-components';
 import { edgeStyle, focusStyle, normalizeColor } from '../../utils';
 import { defaultProps } from '../../default-props';
 
-const fillStyle = fillContainer => {
-  if (fillContainer) {
+const fillStyle = () => {
     return `
       width: 100%;
       height: 100%;
       max-width: none;
       flex: 1 0 auto;
     `;
-  }
-  return undefined;
 };
 
 const disabledStyle = `
@@ -50,7 +47,7 @@ const StyledCheckBoxContainer = styled.label`
   user-select: none;
   ${props =>
     props.fill
-      ? fillStyle(props.fill) && `justify-content: space-between;`
+      ? fillStyle() && `justify-content: space-between;`
       : 'width: fit-content;'}
   ${props =>
     (props.pad || props.theme.checkBox.pad) &&
--- a/src/js/components/CheckBox/StyledCheckBox.js
+++ b/src/js/components/CheckBox/StyledCheckBox.js
@@ -3,14 +3,12 @@ import styled, { css } from 'styled-components';
 import { edgeStyle, focusStyle, normalizeColor } from '../../utils';
 import { defaultProps } from '../../default-props';
 
-const fillStyle = () => {
-    return `
+const fillStyle = () => `
       width: 100%;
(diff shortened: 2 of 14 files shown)


src/js/components/CheckBox/CheckBox.js
```diff
@@ -194,6 +195,7 @@ const CheckBox = forwardRef(
     return (
       <StyledCheckBoxContainer
         aria-label={a11yTitle}
+        fillProp={fill}
```
taysea, line 198:
```suggestion
        fill={fill}
```
Can we just pass through as `fill`? If so, would also need to update the StyledCheckBox stuff

src/js/components/CheckBox/StyledCheckBox.js
```diff
@@ -36,7 +48,11 @@ const StyledCheckBoxContainer = styled.label`
   flex-direction: row;
   align-items: center;
   user-select: none;
-  width: fit-content;
+  ${props =>
+    props.fillProp &&
```
taysea, line 52: Can this take structure of `props.fill ? fillStyle : 'width: fit-content;'`?

src/js/components/CheckBox/CheckBox.js
britt6612: oh yes you just pass in `fill`

src/js/components/CheckBox/StyledCheckBox.js
```diff
@@ -36,7 +48,11 @@ const StyledCheckBoxContainer = styled.label`
...
+    fillStyle(props.fillProp) &&
+    `justify-content: space-between;`} ${props =>
```
taysea, line 54: Do we always want it to justify with space-between? what about if I wanted `fill` so my hover background would show, but the CheckBox/label was still justified to the start

taysea: Should caller have control of that

src/js/components/CheckBox/doc.js
```diff
@@ -27,6 +27,13 @@ export const doc = CheckBox => {
       with the same name so form submissions work.`,
       )
       .defaultValue(false),
+    fill: PropTypes.bool
+      .description(
+        `Whether the checkbox expands to fill all of the available width and/or 
+          height, in order to have set the label and checkbox/toggle justify
+          content to space between.`,
+      )
+      .defaultValue(false),
```
taysea, line 36:
```suggestion
      .defaultValue(undefined),
```

taysea: Is default value undefined?

src/js/components/CheckBox/stories/InsideFormField.js
```diff
@@ -32,6 +32,23 @@ export const InsideFormField = props => (
...
+        <FormField
+          label="Toggle fill"
+          name="toggle"
+          htmlFor="check-box-toggle"
+          {...props}
+        >
+          <Box pad={{ horizontal: 'small', vertical: 'xsmall' }}>
+            <CheckBox
+              id="check-box-toggle"
+              name="toggle"
```
taysea, line 44: Seems like you need to update the id/name stuff here because clicking on this toggle is just toggling the first one

src/js/components/CheckBox/StyledCheckBox.js
britt6612: I was following the pattern of the other PR. In this case yes this would be always set to  ` justify with space-between` however I agree with your point of if you  just wanted the hover background and not the `space-between` Wondering if we can add another prop for adding the `space-between`

britt6612: @ericsoderberghp would you agree to adding another prop that accounts for the `justify`?

ericsoderberghp: I would be OK with not adding a `justify` property and just adding `fill` for now. We can add other layout properties later when we have a concrete need, as opposed to a hypothetical one. I think defaulting to `justify-between` behavior is OK as I think that's the 80% case.

src/js/components/CheckBox/StyledCheckBox.js
```diff
@@ -36,7 +48,10 @@ const StyledCheckBoxContainer = styled.label`
   flex-direction: row;
   align-items: center;
   user-select: none;
-  width: fit-content;
+  ${props =>
+    props.fill
+      ? fillStyle(props.fill) && `justify-content: space-between;`
```
taysea, line 53:
```suggestion
      ? fillStyle() && `justify-content: space-between;`
```
Since `fillStyle` will only be applied when `props.fill` is true, I think we can simplify the function to not have any  parameters.

src/js/components/CheckBox/StyledCheckBox.js
```diff
@@ -3,6 +3,18 @@ import styled, { css } from 'styled-components';
 import { edgeStyle, focusStyle, normalizeColor } from '../../utils';
 import { defaultProps } from '../../default-props';
 
+const fillStyle = fillContainer => {
```
taysea, line 6:
```suggestion
const fillStyle = () => {
```
continuing from other comment around not needing any parameters here.

taysea: Then below, just a single return with
```
return `
      width: 100%;
      height: 100%;
      max-width: none;
      flex: 1 0 auto;
`
```

src/js/components/CheckBox/doc.js
```diff
@@ -27,6 +27,13 @@ export const doc = CheckBox => {
       with the same name so form submissions work.`,
       )
       .defaultValue(false),
+    fill: PropTypes.bool
+      .description(
+        `Whether the checkbox expands to fill all of the available width and/or
```
taysea, line 32: Wording suggestion:
```
Whether the checkbox and label expand to fill all of the available width and/or height of their container. When fill === true, any additional space around the checkbox and label will be applied as justify="between".
```

taysea (commented):
Need to add `fill` to `index.d.ts` but other than that looks good.

src/js/components/CheckBox/doc.js
```diff
@@ -27,6 +27,14 @@ export const doc = CheckBox => {
       with the same name so form submissions work.`,
       )
       .defaultValue(false),
+    fill: PropTypes.bool
+      .description(
+        `Whether the checkbox and label expand to fill all of the available
+         width and/or height of their container. When fill === true,
```
ericsoderberghp, line 33: I don't think we need this last sentence as I think it's describing the internals more than we usually do.

ericsoderberghp (commented):
As Taylor mentioned, still need typescript typings.

britt6612:
> As Taylor mentioned, still need typescript typings.

yes sorry about that thought I pushed up but only ran `git commit` should be up now!

ericsoderberghp (approved):

taysea (approved):
LGTM.

Issue #5122, opened 2021-03-30
Validate/onChange Function on Formfield not getting trigger when used with Select


shreyashah7:
I have a select field which is wrapped with formfield. I have provided my function in validate prop of that formfield which validations happening onChange. My function is not getting triggered at all even when a select any of the option from select field.

### Expected Behavior

It should trigger validate function on formfield when selecting options from select field

### Actual Behavior

Not getting triggered

### URL, screen shot, or Codepen exhibiting the issue

```
<FormField
   dataTestId="formfield-select"
   label="test"
   name="myselect"
   required
   validate={[(value) => {
       console.log("value:", value);
   ]}
>
      <Select
         dataTestId="selectfield"
         name="myselect"
         options={["small", "medium", "large"]}
       />
</FormField>
```

### Steps to Reproduce

On the above code, you will not see console.log in developer tools when selecting any option from the Select Input field.

CC: @ypinneka @shen6653

taysea:
Documenting what I have found so far. It seems that the `onChange` function being called in Select is being registered as an `click`/`keyDown` event type as opposed to `change` like other components such as TextInput or CheckBox. Because of this, the Form is not receiving any notice that the `onChange` happened and therefore does not run the validation.

Looking into how a `change` event may be able to be registered or the `click`/`keyDown` event may be able to trigger the validation.

bbansalWolfPack:
I am also facing the same issue, do we have any ETA on the fix?

PR #5155, opened 2021-04-08
Trigger `change` event when Select value changes

taysea:
#### What does this PR do?
This PR ensures that when Select value changes, a `change` event is dispatched in order for upstream elements, such as FormField/Form to register this change event and trigger any validations (when necessary).

The way Grommet Select results in only `click` or `keyDown` event type to occur when a Select option is chosen. However, FormField validation relies on a `change` event bubbling up from the contained component to run validations. This PR mirrors what is implemented in MaskedInput and was inspired by this [Stack Overflow discussion](https://stackoverflow.com/questions/23892547/what-is-the-best-way-to-trigger-onchange-event-in-react-js/46012210#46012210). If a select value is changing and it has a `name` prop, we ensure that a change event is triggered on the underlying input so it will bubble up to the FormField and trigger validation to run.

Added jest test and adjusted Validate onChange storybook example. Choose "small" from the Select to see the validation run.

With Validate onChange storybook example.

Grommet Select option change results in a `click` or `keyDown` event type but in order for validation to run `onChange`, an event type of `change` needs to bubble up.

Closes https://github.com/grommet/grommet/issues/5122

No.

Fixed Select onChange validation in Form.

Backwards compatible.

### DIFF
--- a/src/js/components/Form/__tests__/Form-test-uncontrolled.js
+++ b/src/js/components/Form/__tests__/Form-test-uncontrolled.js
@@ -468,6 +468,9 @@ describe('Form uncontrolled', () => {
 
   test('validate on change', async () => {
     jest.useFakeTimers();
+    const onChange = jest.fn();
+    window.scrollTo = jest.fn();
+
     const { getByPlaceholderText, queryAllByText } = render(
       <Grommet>
         <Form validate="change">
@@ -501,6 +504,29 @@ describe('Form uncontrolled', () => {
               placeholder="email"
             />
           </FormField>
+          <FormField
+            label="Size"
+            name="test-select"
+            htmlFor="test-select__input"
+            required
+            validate={val => {
+              if (val === 'small')
+                return {
+                  message: 'good',
+                  status: 'info',
+                };
+              return undefined;
+            }}
+          >
+            <Select
+              a11yTitle="select form"
+              id="test-select"
+              name="test-select"
+              placeholder="test input"
+              options={['small', 'medium', 'large']}
+              onChange={onChange}
+            />
+          </FormField>
           <Button label="submit" type="submit" />
         </Form>
       </Grommet>,
@@ -524,8 +550,15 @@ describe('Form uncontrolled', () => {
       target: { value: 'a' },
     });
     act(() => jest.advanceTimersByTime(1000)); // allow validations to run
+    // change value of select
+    fireEvent.click(getByPlaceholderText('test input'));
+    fireEvent.click(document.activeElement.querySelector('button'));
+    window.scrollTo.mockRestore();
+    act(() => jest.advanceTimersByTime(1000)); // allow validations to run
+
     expect(queryAllByText('required')).toHaveLength(1);
     expect(queryAllByText('must be >1 character')).toHaveLength(1);
+    expect(queryAllByText('good')).toHaveLength(1);
   });
 
   test('validate on blur', async () => {
--- a/src/js/components/Form/__tests__/Form-test-uncontrolled.js
+++ b/src/js/components/Form/__tests__/Form-test-uncontrolled.js
@@ -507,7 +507,7 @@ describe('Form uncontrolled', () => {
           <FormField
             label="Size"
             name="test-select"
-            htmlFor="test-select__input"
+            htmlFor="test-select"
             required
             validate={val => {
               if (val === 'small')
--- a/src/js/components/Select/Select.js
+++ b/src/js/components/Select/Select.js
@@ -169,9 +169,9 @@ const Select = forwardRef(
     const onSelectChange = useCallback(
       (event, { option, value: nextValue, selected: nextSelected }) => {
         if (closeOnChange) onRequestClose();
-        // value must not be of type object to set value directly on the
+        // nextValue must not be of type object to set value directly on the
         // input. if it is an object, then the user has not provided necessary
-        // props to reduce complex option object
+        // props to reduce object option
         if (typeof nextValue !== 'object' && nextValue !== event.target.value) {
           // select registers changing option as a click event or keydown.
           // when in a form, we need to programatically trigger a change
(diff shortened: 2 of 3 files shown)


src/js/components/Select/Select.js
```diff
@@ -152,9 +152,33 @@ const Select = forwardRef(
...
+        inputRef.current.dispatchEvent(event);
+      },
+      [inputRef],
+    );
+
     const onSelectChange = useCallback(
       (event, { option, value: nextValue, selected: nextSelected }) => {
         if (closeOnChange) onRequestClose();
+        if (name && nextValue !== event.target.value) {
+          // select registers changing option as a click event or keydown
```
taysea, line 176:
```suggestion
          // select registers changing option as a click event or keydown.
```

ericsoderberghp:
I seem to recall that we had this `Object.getOwnPropertyDescriptor` stuff in an earlier version of Select. I'm guessing it was removed at some point.

src/js/components/Form/__tests__/Form-test-uncontrolled.js
```diff
@@ -501,6 +504,29 @@ describe('Form uncontrolled', () => {
               placeholder="email"
             />
           </FormField>
+          <FormField
+            label="Size"
+            name="test-select"
+            htmlFor="test-select__input"
```
ericsoderberghp, line 510: Why are we having to use the "__input" suffix? That seems like knowing too much about the internals. I would have expected this to be `htmlFor="text-select"` to match with the `id` of the contained Select.

src/js/components/Select/Select.js
```diff
@@ -152,9 +152,33 @@ const Select = forwardRef(
...
+        const event = new Event('input', { bubbles: true });
+        inputRef.current.dispatchEvent(event);
+      },
+      [inputRef],
+    );
+
     const onSelectChange = useCallback(
       (event, { option, value: nextValue, selected: nextSelected }) => {
         if (closeOnChange) onRequestClose();
+        if (name && nextValue !== event.target.value) {
```
ericsoderberghp, line 175: Why do we need the check for `name` here? I would expect any caller listening would want their `onChange` to fire.

src/js/components/Select/Select.js
```diff
@@ -152,9 +152,33 @@ const Select = forwardRef(
...
+        // https://stackoverflow.com/a/46012210
+        const nativeInputValueSetter = Object.getOwnPropertyDescriptor(
+          window.HTMLInputElement.prototype,
+          'value',
+        ).set;
+        nativeInputValueSetter.call(inputRef.current, nextValue);
+        const event = new Event('input', { bubbles: true });
+        inputRef.current.dispatchEvent(event);
+      },
+      [inputRef],
```
ericsoderberghp, line 169: Is `inputRef` really needed, since the `current` property is what can change.

src/js/components/Form/__tests__/Form-test-uncontrolled.js
taysea: Would probably need to look into this as a separate fix since this is existing behavior, but in order for the label to associate with the input (using WAVE browser plugin to check this), we currently have to add `__input`. Agreed grommet should probably handle this internally.

src/js/components/Select/Select.js
taysea: You're right, I can remove `inputRef` and it doesn't flag any missing dependencies.

src/js/components/Form/__tests__/Form-test-uncontrolled.js
taysea: I think this is due to the fact that we internally manipulate the `id` passed to select input to be `${id}__input`, so we would probably want to to the same adjustment to `htmlFor` if the component is a Select.

src/js/components/Select/Select.js
taysea: Updated.

src/js/components/Form/__tests__/Form-test-uncontrolled.js
ericsoderberghp: Something seems broken if the `id` isn't associated with the underlying `<input />` element, since that's what `htmlFor` is typically used for. For instance, CheckBox puts the `id` property on the underlying `<input type="checkbox" />` element.

It looks like Select puts the `id` on the DropButton. I can see where this might seem a good choice, but I think the input element is a better place. However, this would be a breaking change. Sigh. I wonder if we should support: `id={{ input: ..., button: ... }}`? It looks crazy, but is at least semantic. Hmm ...

taysea: Yes, it would be a breaking change to move the placement of the input. We don't need to have the `__input` for the storybook example to work, I just wanted it to pass the WAVE check.

How would you feel about handling this work in a separate PR and having a little more discussion around implementation? I can remove the `__input` for now from the story for clarity if that'd be better.

ericsoderberghp: I'm fine taking care of the `id` element mapping in a separate PR. I would remove the `__input` from this PR.

taysea: Done.

ericsoderberghp (approved):

ShimiSun (approved):

daniel-parlapan:
Is there a plan to version this an have a new release?

ShimiSun:
@daniel-parlapan thanks for reaching out with a release request. We are planning a release in the coming couple of weeks after completing high-priority issues and perform testing.

src/js/components/Select/Select.js
```diff
@@ -152,9 +152,33 @@ const Select = forwardRef(
...
+      inputRef.current.dispatchEvent(event);
+    }, []);
+
     const onSelectChange = useCallback(
       (event, { option, value: nextValue, selected: nextSelected }) => {
         if (closeOnChange) onRequestClose();
+        // nextValue must not be of type object to set value directly on the
+        // input. if it is an object, then the user has not provided necessary
+        // props to reduce object option
+        if (typeof nextValue !== 'object' && nextValue !== event.target.value) {
```
daniel-parlapan, line 175: Seems that this does not cover the select capability of having options as objects. 
We should take into account if (typeof nextValue === 'object')  and compare define actualValue = nextValue[valueKey] and then compare it to previous

Issue #5152, opened 2021-04-08
Button tip string should be used for a11yTitle default value if none provided


samuelgoff:
`Buttons` and icon `Buttons` quite often need a `tip` value and `a11yTitle` value that is (or could suitably be) identical. The result is you end up repeating yourself, or declaring a `const` just so you only have to repeat the reference rather than the literal (or derivation), but it's still redundant.

### Expected Behavior

If `Button` `tip` attribute is populated with a string but `a11yTitle` is not, then `a11yTitle` should get default value from the Button `tip` value. Not only does this solve a DRY opportunity, but this also ensures that if the `a11yTitle` value isn't populated, it will still allow the component to add a useful/meaningful `aria-label` value, improving accessibility with no additional effort on the part of the developer.

### Actual Behavior

Create a `Button` and populate `tip` with a string but omit supplying an `a11yTitle` value, then elicit `Tip` visibility so you can examine the DOM. Note that the `aria-label` is not populated in the resulting `button` DOM.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-forked-w597x?file=/index.js

### Steps to Reproduce

1. Create a `Button` 
2. populate `tip` with a string but omit supplying an `a11yTitle` value
3. elicit `Tip` visibility so you can examine the DOM
4. Note that the `aria-label` is not populated in the resulting `button` DOM.

### Your Environment

- Grommet version: grommet@2.17.1
- Browser Name and version: Chrome Version 89.0.4389.114 (Official Build) (x86_64) & Safari Version 14.0.3 (16610.4.3.1.7)
- Operating System and version (desktop or mobile): macOS Big Sur Version 11.2.3

samuelgoff:
Thanks Shimi, I already have it done & PR ready, but the battery on my laptop died before I could submit. Trekking home from getting the COVID vaccine, then I’ll be able to plug in & submit the PR.

Thanks,

Sam

samuelgoff:
> I like this enhancement idea!
> The code contribution for this enhancement should be straightforward, would you like to take a stab at it? @samuelgoff

Hi @ShimiSun, hopefully this works for you.

PR #5156, opened 2021-04-09
Button tip text default for a11y

samuelgoff:
Signed-off-by: Samuel Goff <samuel@defpix.com>

#### What does this PR do?
Addresses #5152 Button tip string should be used for a11yTitle default value if none provided

Button.js

`yarn test-update`, `yarn test`, & `yarn storybook`. Results shown below.

1. Create a Button
2. populate tip with a string but omit supplying an a11yTitle value
3. elicit Tip visibility so you can examine the DOM
4. Note that the aria-label is now populated in the resulting button DOM.

`yarn test-update`:
Screen Shot 2021-04-08 at 7 15 54 PM

`yarn test`:
Screen Shot 2021-04-08 at 7 17 32 PM

Already done :) This has been updated in the relevant `doc.js` file

Whatevs. If you think it will be helpful, sure.

Backward compatible. Move fast & break nothing.

### DIFF
--- a/src/js/components/Button/Button.js
+++ b/src/js/components/Button/Button.js
@@ -93,11 +93,7 @@ const getIconColor = (paths = [], theme, colorProp, kind) => {
 };
 
 const tipString = tip => {
-  if (typeof tip === 'string') {
-    return tip;
-  }
-  return undefined;
-};
+  return typeof tip === 'string' ? tip : undefined;
 
 const Button = forwardRef(
   (
--- a/src/js/components/Button/Button.js
+++ b/src/js/components/Button/Button.js
@@ -92,10 +92,16 @@ const getIconColor = (paths = [], theme, colorProp, kind) => {
   return result[1] || undefined;
 };
 
+const tipString = tip => {
+  if (typeof tip === 'string') {
+    return tip;
+  }
+  return '';
+};
+
 const Button = forwardRef(
   (
     {
-      a11yTitle,
       active,
       align = 'center',
       'aria-label': ariaLabel,
@@ -122,6 +128,7 @@ const Button = forwardRef(
       size,
       tip,
       type = 'button',
+      a11yTitle = tipString(tip), // can't alphabetize before tip defined
       as,
       ...rest
     },
--- a/src/js/components/Button/Button.js
+++ b/src/js/components/Button/Button.js
@@ -92,9 +92,6 @@ const getIconColor = (paths = [], theme, colorProp, kind) => {
   return result[1] || undefined;
 };
 
-const tipString = tip => {
-  return typeof tip === 'string' ? tip : undefined;
-
 const Button = forwardRef(
   (
     {
@@ -124,7 +121,8 @@ const Button = forwardRef(
       size,
       tip,
       type = 'button',
-      a11yTitle = tipString(tip), // can't alphabetize before tip defined
+      // can't alphabetize a11yTitle before tip is defined
+      a11yTitle = typeof tip === 'string' ? tip : undefined,
       as,
       ...rest
     },
--- a/src/js/components/Button/Button.js
+++ b/src/js/components/Button/Button.js
@@ -96,7 +96,7 @@ const tipString = tip => {
   if (typeof tip === 'string') {
     return tip;
   }
-  return '';
+  return undefined;
 };
 
 const Button = forwardRef(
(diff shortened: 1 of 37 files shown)


src/js/components/Button/Button.js
```diff
@@ -92,10 +92,16 @@ const getIconColor = (paths = [], theme, colorProp, kind) => {
   return result[1] || undefined;
 };
 
+const tipString = tip => {
+  if (typeof tip === 'string') {
+    return tip;
+  }
+  return '';
```
ShimiSun, line 99: I think it will be cleaner to use `undefined` instead of an empty string.
With the current solution, we have an empty string of `aria-label` that is leaking to the DOM unnecessarily. This change may clean up the snapshots. 
Let me know your thoughts on this.

ShimiSun (commented):
Thank you for this PR and this enhancement! 
I dropped a comment, let me know what you think.

Also, Button newest version is ButtonKind, and that is the Button that is actually being used for the HPE DS button. we'll need to enable the fix for this version of the kind Button as well.

src/js/components/Button/Button.js
samuelgoff: Hi @ShimiSun, good call! I had originally tried to return `false`, but this didn't produce the expected result. That certainly did clean things up.

The only "ButtonKind" reference I could find in the codebase was StyledButtonKind, so presumably this is what you're referring to? I added the same assignment there.

Please let me know if you need anything else to make this work. Thanks!

ShimiSun: THANK YOU! It really cleaned up the PR, the number of changed files is so much more reasonable now :)

Yea, StyledButtonKind sounds like the right place, I don't see any changes there though, what am I missing? 

Also, one last request, we try to keep PRs as lean as possible to their content, can you please remove all the changes of empty spaces removals that are affecting the snapshots/README unnecessary changes? It will remove the clutter from the PR.

Thanks again for working on this.

samuelgoff: > THANK YOU! It really cleaned up the PR, the number of changed files is so much more reasonable now :)

No problem!

> Yea, StyledButtonKind sounds like the right place, I don't see any changes there though, what am I missing?

Line 276 in the same file? That lives inside of `<StyledButtonKind />`
 
> Also, one last request, we try to keep PRs as lean as possible to their content, can you please remove all the changes of empty spaces removals that are affecting the snapshots/README unnecessary changes? It will remove the clutter from the PR.

I can certainly do this. However, I should point out that these files (and changes) were generated by running the scripts in the [Grommet contribution guidelines](https://github.com/grommet/grommet/blob/master/CONTRIBUTING.md), not by any change I introduced from my side. Let me see if I can make this work for you.

> Thanks again for working on this.

No problem. I'm glad I'm able to help.

samuelgoff: Hey @ShimiSun, I believe all the line-break issues have been cleaned up.

src/js/components/Button/Button.js
```diff
@@ -92,10 +92,16 @@ const getIconColor = (paths = [], theme, colorProp, kind) => {
   return result[1] || undefined;
 };
 
+const tipString = tip => {
+  if (typeof tip === 'string') {
+    return tip;
+  }
+  return undefined;
+};
```
ShimiSun, line 100: what about
```suggestion
  return typeof tip === 'string' ? tip : undefined;
```

ShimiSun (commented):
Looking good, dropped one small comment.

src/js/components/Button/Button.js
samuelgoff: hey @ShimiSun, good suggestion!

src/js/components/Button/Button.js
```diff
@@ -92,10 +92,12 @@ const getIconColor = (paths = [], theme, colorProp, kind) => {
   return result[1] || undefined;
 };
 
+const tipString = tip => {
```
ShimiSun, line 95: Aren't you missing a `}`?

Also, I think we can remove the function and just make it a constant variable.

samuelgoff: ahh, that's what I get for approving GitHub suggestions directly. You're right, now that the assignment is a ternary, it's short enough to inline in the default assignment. This change has been made.

ShimiSun (approved):
Thank you for your contribution @samuelgoff! Hoping we'll see more contributions from you in the future!

PR #5157, opened 2021-04-09
Fix SSR warnings with an isomorphic useLayoutEffect

### LINKED ISSUE #4765: Warning when rendering application using SSR frameworks like 'Razzle'
When I started to put simple application page together with Server Side Rendering React based boilerplate project ( razzle )

it gives me following error

Warning: useLayoutEffect does nothing on the server, because its effect cannot be encoded into the server renderer's output format. This will lead to a mismatch between the initial, non-hydrated UI and the intended UI. To avoid this, useLayoutEffect should only be used in components that render exclusively on the client. See https://fb.me/react-uselayouteffect-ssr for common fixes.

### Expected Behavior

expected that components will handle these for server side rendered applications

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

oranoran:
#### What does this PR do?
Fix SSR warnings with an isomorphic useLayoutEffect, including an eslint rule to enforce usage of the isomorphic hook.

Might want to read #4765 and the blog at https://medium.com/@alexandereardon/uselayouteffect-and-ssr-192986cdcf7a
Then look at the fix:
* src/js/utils/use-isomorphic-layout-effect.js
* .eslintrc
Everything else is trivial.

Build, test, lint
Checked that the storybook is stable for relevant components.
Used a fixed build of Grommet instead of the current stable version in a next.js project and saw that nothing is broken and the warning is no longer issued
In the Grommet code, tried to wrongly import useLayoutEffect from 'react', and verified that eslint fails with an appropriate error message.

Check that an affected component such as InfiniteScroll is not broken, in an SSR and a non-SSR use case. Verify that the warning is gone.
You can repeat the lint test by trying to import useLayoutEffect directly from 'react' and verifying that linting fails.

n/a

#4765

n/a

no

yes

backwards compatible

### DIFF
--- a/src/js/utils/use-isomorphic-layout-effect.js
+++ b/src/js/utils/use-isomorphic-layout-effect.js
@@ -1,5 +1,5 @@
 /* eslint-disable no-restricted-imports */
-import React, { useEffect } from 'react';
+import { useLayoutEffect as vanillaUseLayoutEffect, useEffect } from 'react';
 
 /**
  * A substitute for React's useLayoutEffect, which does not generate warnings on
@@ -10,4 +10,6 @@ import React, { useEffect } from 'react';
  * @see https://github.com/grommet/grommet/issues/4765
  */
 export const useLayoutEffect =
-  typeof window !== 'undefined' ? React.useLayoutEffect : useEffect;
+  typeof window !== 'undefined' ? vanillaUseLayoutEffect : useEffect;
+
+export default useLayoutEffect;
--- a/src/js/utils/use-isomorphic-layout-effect.js
+++ b/src/js/utils/use-isomorphic-layout-effect.js
@@ -1,0 +1,13 @@
+/* eslint-disable no-restricted-imports */
+import React, { useEffect } from 'react';
+
+/**
+ * A substitute for React's useLayoutEffect, which does not generate warnings on
+ * SSR. It is named useLayoutEffect so that all eslint rules applying to the
+ * original useLayoutEffect would also apply to it.
+ * This solution was suggested by Alex Reardon.
+ * @see https://medium.com/@alexandereardon/uselayouteffect-and-ssr-192986cdcf7a
+ * @see https://github.com/grommet/grommet/issues/4765
+ */
+export const useLayoutEffect =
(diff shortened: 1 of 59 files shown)


src/js/utils/use-isomorphic-layout-effect.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable no-restricted-imports */
+import React, { useEffect } from 'react';
```
bojan88, line 2: Is it possible not to import the whole `React`? Like `import { useEffect, useLayoutEffect } from 'react';`

src/js/utils/use-isomorphic-layout-effect.js
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * A substitute for React's useLayoutEffect, which does not generate warnings on
+ * SSR. It is named useLayoutEffect so that all eslint rules applying to the
+ * original useLayoutEffect would also apply to it.
+ * This solution was suggested by Alex Reardon.
+ * @see https://medium.com/@alexandereardon/uselayouteffect-and-ssr-192986cdcf7a
+ * @see https://github.com/grommet/grommet/issues/4765
+ */
+export const useLayoutEffect =
```
bojan88, line 12: What do you think about exporting this as default? For me it makes more sense based on the module name.

src/js/utils/use-isomorphic-layout-effect.js
oranoran: Will change this.

src/js/utils/use-isomorphic-layout-effect.js
oranoran: I like the non-default export because it makes usage similar to that of the original useLayoutEffect (which is typically imported by name).
However I don't see any harm in adding a default export too, will add.

ericsoderberghp (approved):

MikeKingdom:
Looks good to me @oranoran . Thanks for your contribution!

MikeKingdom (approved):

ShimiSun:
Toda Raba @oranoran !! Hope we'll see more contributions from you 💟

PR #5158, opened 2021-04-09
Fixed Box to always render ThemeContext.Provider

### LINKED ISSUE #5146: FormField loses focus on validation
This seems to be a regression from 2.16.3 to 2.17.1. The problem does not happen in 2.16.3

When trying to enter text in a required TextInput in a FormField and the field is marked invalid, entering the first character causes the field to switch to valid but the character never appears and the focus shifts elsewhere. You have to click back in the input and before you can enter more text.
 
### Expected Behavior

The focus should not change when the field switches between valid and invalid.

### Actual Behavior

The field loses focus and the text is lost (although the field appears to be valid at this point.)

https://codesandbox.io/s/long-shadow-mjs3i?file=/src/AddForm.js
```jsx
  <Form onSubmit={onSubmit}>
    <FormField
      label="Description"
      name="description"
      htmlFor="description"
      required
      component={TextInput}
      placeholder="Enter a description for the volume"
    />
    <Button type="submit" primary label="Submit" />
  </Form>
```

### Steps to Reproduce

Click on Submit. The field is marked invalid since it is required:
image

Click in the input and try to enter something (e.g 'a')
image

The red background and validation message disappear but the character you typed does not appear. The field still has the focus border but further typing is ignored.

You must click in the field again and continue typing.

Backspacing to remove the text clears the field, but the field doesn't get marked invalid once you clear all text.

image

### Your Environment

- Grommet version: 2.17.1

MikeKingdom:
Currently Box only inserts a `<ThemeContext.Provider>` around the content if it needs to supply a background and/or a value for `dark` based on the background. However if the background changes (like when a validation error causes `FormField` to set the background of it's content `Box` to red or back to no background) this addition or removal of the `ThemeContext.Provider` causes the content in the Box to get re-constructed, losing any state. This manifested in issue #5146 

#### What does this PR do?
Change Box to always render a `<ThemeContext.provider>` for a consistent component hierarchy. This allows the child component specified by the user for the box to maintain state and render more efficiently.

Box.js

Storybook and Jest tests

Any FormField  using the HPE theme and marked required with a TextInput child and a submit button. (See #5146 ) 

Fixes #5146 

No

Yes

Backwards compatible

### DIFF
--- a/src/js/components/Box/Box.js
+++ b/src/js/components/Box/Box.js
@@ -109,26 +109,26 @@ const Box = forwardRef(
       });
     }
 
+    const themeProviderValue = { ...theme };
+
     if (background || theme.darkChanged) {
-      let dark = backgroundIsDark(background, theme);
+      const dark = backgroundIsDark(background, theme);
       const darkChanged = dark !== undefined && dark !== theme.dark;
       if (darkChanged || theme.darkChanged) {
-        dark = dark === undefined ? theme.dark : dark;
-        contents = (
-          <ThemeContext.Provider value={{ ...theme, dark, background }}>
-            {contents}
-          </ThemeContext.Provider>
-        );
+        themeProviderValue.dark = dark === undefined ? theme.dark : dark;
+        themeProviderValue.background = background;
       } else if (background) {
         // This allows DataTable to intelligently set the background of a pinned
         // header or footer.
-        contents = (
-          <ThemeContext.Provider value={{ ...theme, background }}>
-            {contents}
-          </ThemeContext.Provider>
-        );
+        themeProviderValue.background = background;
       }
     }
+
+    contents = (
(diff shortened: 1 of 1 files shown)


ericsoderberghp (approved):

shc023:
Hello @MikeKingdom @ShimiSun , this commit seems to be causing major performance issues for 2.17.3 - when a `Box` re-renders, any descending `styled-components` within it now gets re-rendered even if they're contained in a memoized parent that's wrapped in `React.memo`, because it updates the `theme` for the `styled-components` theme context even if there's no value change.

Here's a minimum repro. You can change the installed grommet version to `2.17.1` and see that the problem goes away.
https://stackblitz.com/edit/grommet-demo-xizjuy?devtoolsheight=33&file=index.js

PR #5161, opened 2021-04-12
Enhance DataTable multiple pinning

### LINKED ISSUE #5130: Enhance DataTable to support pinning multiple columns
### Expected Behavior

If multiple columns are pinned, they will stay fixed to their position if there is horizontal scroll.

### Actual Behavior

Columns scroll to the left until they hit the left edge of the DataTable which causes multiple pinned columns to stack on top of each other.

View when screen === 'small':

https://codesandbox.io/s/multi-pin-columns-e6758?file=/components/TableExample.js

### Steps to Reproduce

1.  View table when screen === 'small'
2. Scroll horizontally
3. Notice how the 3 left columns are pinned but they become stacked on each other.

### Your Environment

- Grommet version: 2.17.1
- Browser Name and version: Mac Chrome
- Operating System and version (desktop or mobile): Mac OS Mojave

leossantiago:
#### What does this PR do?

Enhancement to `DataTable` pinning mechanism. 

The current behavior is to set `position: sticky;` and `left: 0` to all pinned columns, leading to the reported case in #5130 where pinned columns get on top of each other.

This fix is based on filtering which columns have the `pin` property and then calculating the position offset each pinned column should have. This takes pinned column order into consideration and only considers pinned column widths, as users are not limited to pin a sequence of columns on the left side (i.e. users can pin the first and third columns, skipping the second).

Manual testing as this involves `getBoundingClientRect`, but existing tests can confirm DataTable still works as intended.

Introduced a new story (`Visualizations/DataTable/Multiple pins` ) based on the provided example on the issue.

Closes #5130 

Peek 2021-04-12 01-19

No.

I'd consider it backward compatible as it doesn't affect overall usage. This will only affect cases where multiple pinned columns were set.


src/js/components/DataTable/DataTable.js
```diff
@@ -273,6 +273,40 @@ const DataTable = ({
       ? { style: { minWidth: '100%' } }
       : undefined;
 
+  const pinnedColumns = columns
+    .map(
+      (pinnedColumn, index) =>
+        pinnedColumn.pin && { property: pinnedColumn.property, id: index },
```
taysea, line 279: Do we need to return `id` at all? I think we can just leverage `index` on line `289`

src/js/components/DataTable/DataTable.js
```diff
@@ -273,6 +273,40 @@ const DataTable = ({
...
+    )
+    .filter(n => n);
+
+  const getPinnedOffset = () => {
+    const pinnedPropertiesOrder = [];
+    const pinnedColumnsProperties = {};
+    const firstRowCells = headerRef?.current?.firstChild?.cells || [];
+
+    if (firstRowCells.length !== 0) {
+      pinnedColumns.forEach(({ property, id }, index) => {
```
taysea, line 289:
```suggestion
      pinnedColumns.forEach(({ property }, index) => {
```
Think we don't need `id` and can just use `index` in this function.

src/js/components/DataTable/DataTable.js
taysea: We could also probably just return `pinnedColumn.property` and create an array of strings instead of an array of objects with just one key.

src/js/components/DataTable/stories/MultiplePin.js
```diff
@@ -0,0 +1,331 @@
...
+  return (
+    <>
+      <Box>ResponsiveContext size: {size}</Box>
+      <Box>
+        Columns `id`, `Pool Name` and `Group Name` will become pinned columns
+        once the current ResponsiveContext size is `small`
+      </Box>
+      <Box height="medium" overflow="auto">
+        <DataTable
+          data={data}
```
taysea, line 288:
```suggestion
          background={{ pinned: 'background' }}
          data={data}
```
Small suggestion but just adding a background color to the pinned columns so when the data scrolls behind it doesn't become unreadable.

src/js/components/DataTable/stories/MultiplePin.js
```diff
@@ -0,0 +1,331 @@
...
+          columns={[
+            {
+              property: 'id',
+              header: 'Id',
+              primary: true,
+              render: datum => datum.id.slice(datum.id.length - 5),
+              pin: size === 'small',
```
taysea, line 295:
```suggestion
              pin: true,
```
Wondering if it'd be better to set the pin to `true` and then add a fixed width to the containing Box (`width="large"`) so people consuming the example won't have to resize their window to see the functionality.

taysea: Would need adjusting on the two columns below this also.

src/js/components/DataTable/DataTable.js
```diff
@@ -273,6 +273,40 @@ const DataTable = ({
...
+        if (firstRowCells[id]) {
+          pinnedPropertiesOrder.push(property);
```
taysea, line 291: If pinnedColumns becomes an array of strings, I think it would match `pinnedPropertiesOrder` and we wouldn't need this additional array.

taysea (commented):
Nice work here. Dropped some initial comments on a couple places where we might be able to simplify the `getPinnedOffset` function

leossantiago:
This took longer than expected.

@taysea just pushed a commit addressing some refactors to the Multiple Pins story, and also reworked the multiple pinning logic. As I mentioned previously, this also covers cases where the pinned columns aren't after each other.

While reworking it, I discovered a bug related to datatable not computing pinnedOffset properly after the first render in some situations, making the pinned columns override each other just as they used to behave. Fixed it by wrapping the logic into an `useEffect` hook.

src/js/components/DataTable/DataTable.js
```diff
@@ -173,6 +176,40 @@ const DataTable = ({
...
+                    pinnedColumnsProperties[pinnedProperties[index - 1]].width,
+            };
+          }
+        });
+      }
+      return pinnedColumnsProperties;
+    };
+
+    setPinnedOffset(getPinnedOffset());
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
taysea, line 210: Can you add a comment about why we need to ignore exhaustive-deps warning? If dependency array is not desired, can we leave the `[]` out entirely?
```
useEffect(() => {...});
```

src/js/components/DataTable/DataTable.js
```diff
@@ -173,6 +176,40 @@ const DataTable = ({
...
   }, [footerRef, headerRef, placeholder]);
 
+  useEffect(() => {
+    const getPinnedOffset = () => {
+      const pinnedProperties = columns
+        .map(pinnedColumn => pinnedColumn.pin && pinnedColumn.property)
+        .filter(n => n);
+
+      const pinnedColumnsProperties = {};
+      const firstRowCells = headerRef?.current?.firstChild?.cells || [];
```
taysea, line 186: Just for my own understanding, is the `?` syntax the equivalent of saying the following?
```
headerRef && headerRef.current && headerRef.current.firstChild ...
```

src/js/components/DataTable/DataTable.js
leossantiago: We can try a workaround on that, but just removing the dependency array is leading to exhaustive-deps warnings. Probably because of line 209: `setPinnedOffset(getPinnedOffset());`

image


We can also set `columns` as a dependency. Might be clever considering we want this to update if columns are updated as well. What do you think?

src/js/components/DataTable/DataTable.js
leossantiago: Pretty much that. That was introduced as code was breaking when trying to access data from `headerRef.current` while it was undefined. Just realized this is kinda gone once we moved the logic to an useEffect hook. I'd suggest we go with one of those two approaches, in that case:

`headerRef.current.firstChild.cells || []`
or
`headerRef.current.firstChild.cells ?? []`

src/js/components/DataTable/DataTable.js
taysea: Following up from offline discussion. Recommending we try `useMemo` for `pinnedOffset` instead to potentially overcome the 'maximum depths' issue.

src/js/components/DataTable/DataTable.js
```diff
@@ -152,7 +152,41 @@ const DataTable = ({
...
+            left:
+              tableIndex === 0
+                ? 0
+                : pinnedColumnsProperties[pinnedProperties[index - 1]].left +
+                  pinnedColumnsProperties[pinnedProperties[index - 1]].width,
+          };
+        }
+      });
+      return pinnedColumnsProperties;
+    }
```
taysea, line 184:
```suggestion
    }
    return undefined;
```
Would it make sense to return undefined if we didn't set anything for `pinnedColumnsProperties` in this function? Effectively just reseting the state value to its default? If so, I think we could remove the eslint-disable comment

src/js/components/DataTable/DataTable.js
```diff
@@ -151,6 +151,43 @@ const DataTable = ({
...
+  // multiple pinned columns offset
+  const [pinnedOffset, setPinnedOffset] = useState();
+
+  const getPinnedOffset = useCallback(() => {
+    const pinnedProperties = columns
+      .map(pinnedColumn => pinnedColumn.pin && pinnedColumn.property)
+      .filter(n => n);
+
+    const pinnedColumnsProperties = {};
+    const firstRowCells = headerRef.current.firstChild.cells || [];
```
ericsoderberghp, line 163: This is feeling like DataTable is reaching inside Header a bit too much. I'm wondering if we should move this inside Header and Cell, perhaps something like:
`<Header onWidths={([n,n,n]) => {}} />` and `<Cell onWidth={(width) => {}} />`
?

ericsoderberghp (commented):
I like the general idea of calculating the header cell widths and then passing in the appropriate offset. I had one suggestion on how to structure the calculations.

src/js/components/DataTable/DataTable.js
leossantiago: I'm always open to suggestions, but I'm not sure if I'm following this one, sorry 😄 can you go a little further on that?

ericsoderberghp: So, rather than have DataTable know that the Header's first child contains cells, which is knowing more about the Header than DataTable should, have the DataTable pass an `onWidths` to Header, indicating that the DataTable wants to be told the Header cell widths. When `onWidths` is passed to Header, Header passes an `onWidth` to each Cell. When a Cell has an `onWidth`, it reports its width, possible via a `useLayoutEffect()`. Header merges these into an array that it passes back up to DataTable.

The motivation is to not have DataTable know about the internals of the DOM structure of Header, to separate concerns. I am thinking this is just a few lines in each file. But, if going this route appears to be too complicated, I'm willing to consider other approaches.

leossantiago:
Updated the logic to separate responsibilities between cells, header, and DataTable.  However, just noticed there are some conflicts caused by #5177 . I'll address that ASAP.

ericsoderberghp (commented):
Just a few naming suggestions. Functionality looks good.

leossantiago:
Done! Let me know if you have any other suggestions 😄

ericsoderberghp (approved):

Issue #5162, opened 2021-04-13
[TS] DateInput type definition is missing disabled prop


Shannor:
### Expected Behavior

Should be able to disable the DateInput while using typescript

### Actual Behavior

If using typescript you will get an error that disabled doesn't exist on DateInput

### URL, screen shot, or Codepen exhibiting the issue
 Here is a link with someone stating the solution https://github.com/grommet/grommet/issues/4438

### Steps to Reproduce
Use the DateInput and try to disable. 

### Your Environment

- Grommet version:  2.17.1
- Browser Name and version:
- Operating System and version (desktop or mobile):

jcfilben:
Looks like this issue can be closed since #5165 was merged

Shannor:
Cool. Closing now. Thanks!

PR #5165, opened 2021-04-14
[TS] DateInput type definition is missing disabled prop

jaygiang:
#### What does this PR do?
Addresses issue: https://github.com/grommet/grommet/issues/5162
 - Adds disabled prop type to DateInputProps
 - Update docs
 - Adds test
`src/js/components/DateInput/index.d.ts`
Jest testing
In typescript project, use `disabled` props in `DateInput` component. There should be no error saying that disabled doesn't exist.
n/a
 https://github.com/grommet/grommet/issues/5162
n/a
Yes, it has been updated
n/a
Backwards compatible


britt6612:
@jaygiang would you mind running `yarn test -update` locally then commit the changes up to this PR. Then I think the jest tests will pass.

src/js/components/DateInput/index.d.ts
```diff
@@ -17,6 +17,7 @@ export interface DateInputProps {
   name?: string;
   onChange?: (event: { value: string | string[] }) => void;
   value?: string | string[];
+  disabled?: boolean;
```
ShimiSun, line 20: I think we might be missing the underlying JSX.IntrinsicElements, since DateInput should be using `JSX.IntrinsicElements['input']` then `disabled` should already be included as part of it. Could you please take a look at whether it is hiding somewhere or we should add the JSX.IntrinsicElements['input'] manually?

ShimiSun (commented):
Thanks Jay!! Dropped a comment

src/js/components/DateInput/index.d.ts
jaygiang: Thanks Shimi! I've added the JSX.IntrinsicElements['input'] here https://github.com/grommet/grommet/pull/5165/commits/f4a7f51ce21b2de562e55c5e47b70b4ad55abf71

jaygiang:
@britt6612 I ran `yarn test -update` and have pushed up the commits https://github.com/grommet/grommet/pull/5165/commits/3bca29d47c8a32e8c547331a2e2f01068788d463

ShimiSun (commented):
Hi @jaygiang, I think this PR shouldn't have any snapshot changes. You might want to update your upstream and clean up this PR. 
If it doesn't work, you can start fresh and file a new PR with only the TS file change.

src/js/components/DateInput/index.d.ts
ShimiSun: Great! I'm guessing that line #20 is no longer needed if we are using the `JSX.IntrinsicElements['input']` now, since `disabled` is a prop of `JSX.IntrinsicElements['input']` by default, does that work?

britt6612: @jaygiang this should be fine once you remove the `disabled?: boolean;`

britt6612 (changes requested):
Just take out the `disabled?: boolean;` should be good to go 👍

src/js/components/DateInput/index.d.ts
jaygiang: @britt6612 @ShimiSun Thanks for your help! I have removed `disabled?: boolean;`, but I'm not sure why it's still not passing.

src/js/components/DateInput/doc.js
```diff
@@ -59,6 +59,9 @@ export const doc = DateInput => {
       PropTypes.string,
       PropTypes.arrayOf(PropTypes.string),
     ]).description('The date or date range value(s) in ISO8601 format.'),
+    disabled: PropTypes.bool.description(
+      'Whether the field should look disabled.',
```
britt6612, line 63:
```suggestion
      'Whether the Datenput is disabled.',
```

britt6612: sorry didnt see before can we maybe change the description so it matches the other components..

ShimiSun (commented):
I don't think we should add the `disable` prop to the documentation, it's enough to adapt the typescript file for this fix. We are not documenting all of the `input` types of our props unless we are modifying them in any way.

jaygiang:
@ShimiSun I have removed the `disabled` prop from the docs.

@britt6612 After removing the `disabled` prop from docs, I ran `yarn test -update`, but it updates snapshots from 28 different component files. I ran this previously and have pushed up all those files, which have now been removed: https://github.com/grommet/grommet/pull/5165#issuecomment-819941283 . Do you think something could be wrong with my test configuration? I'll take a closer look too. Thanks!

britt6612:
> @ShimiSun I have removed the `disabled` prop from the docs.
> 
> @britt6612 After removing the `disabled` prop from docs, I ran `yarn test -update`, but it updates snapshots from 28 different component files. I ran this previously and have pushed up all those files, which have now been removed: [#5165 (comment)](https://github.com/grommet/grommet/pull/5165#issuecomment-819941283) . Do you think something could be wrong with my test configuration? I'll take a closer look too. Thanks!

@jaygiang Id be happy to screen share with you next week if that works.. Im not sure why on your side its not only updating the certain snapshots.

britt6612 (approved):

ShimiSun (approved):

ShimiSun (approved):

ShimiSun:
Thank you for the contribution @jaygiang!

Issue #5172, opened 2021-04-18
`CheckBox` `fill` property is broken


versoworks:
Using Grommet v2.17.1. Setting the `fill` property on a `CheckBox` doesn't perform as designed, and also raises an error in the console;

`Warning: Received 'true'  for a non-boolean attribute 'fill'.`

See the [Codesandbox](https://codesandbox.io/s/elated-proskuriakova-p86yu) example.

britt6612:
@versoworks @leossantiago 
The `fill` prop was added on the stable branch you will not see these changes until the next release. So this codesandbox would not be up to date can you maybe check stable branch ?

leossantiago:
@britt6612 even though the codesandbox might not be up to date, I was able to reproduce both issues with the master branch. Thanks for the heads up! 😄

ShimiSun:
Related to #5153.

ShimiSun:
Fixed and available on grommet stable, and on future release grommet 2.17.2

PR #5173, opened 2021-04-19
Fix CheckBox fill prop behavior and type warnings

leossantiago:
#### What does this PR do?

Addresses CheckBox `fill` property behavior issues and `type` warnings while using the same prop. 

- `fill` prop behavior

CheckBox component can have the checkbox itself and a label, both wrapped on a div Container. Right now, setting the `fill` prop doesn't result in any changes, where it should expand this Container to the limits of the parent component. 
After some debugging, I ended up discovering that this happens due to fill styling not being properly applied:

https://github.com/grommet/grommet/blob/adfc3091a933958aa5762c9bd493dbe059fb4e61/src/js/components/CheckBox/StyledCheckBox.js#L46-L49

Since `fillStyle` isn't used anywhere else, we can incorporate the `justify-content` prop to it, and call this function if `fill` prop is set.

- `fill` prop type warnings

The warnings mentioned on the issue can easily be associated with the `fill` prop from CheckBox. However, it's related to `StyledCheckBoxContainer`, an internal component that represents the wrapping div of CheckBox. As seen below, it's created based on `styled.label`:

https://github.com/grommet/grommet/blob/adfc3091a933958aa5762c9bd493dbe059fb4e61/src/js/components/CheckBox/StyledCheckBox.js#L41

For some reason, `label` is required to be `true` or `false`, but with the type of string, so, not an issue related to Grommet. Based on that, I added a bypass to that, alongside a Note for future reference so we can update it once it's fixed on `styled-components` end.

I ended up not going further with this as it's not the point of #5172, but those changes made me consider: 
1. Enhancing fill prop to support different alignment options besides `space-between`;
2. Use Box component as base for `StyledCheckBoxContainer` instead of `styled.label`.

Would love to hear our community's inputs on those. 

`src/js/components/CheckBox/CheckBox.js`
`src/js/components/CheckBox/StyledCheckBox.js`

Manual testing and updating existing unit tests. Don't think there's a need to add new ones as the prop usage is already covered.

```js
import React from "react";
import { grommet, Box, CheckBox, Grommet } from "grommet";

export default function App() {
  return (
    <Grommet theme={grommet}>
      <Box fill background="accent-1" pad="small">
        <CheckBox fill reverse label="Check me out" />
      </Box>
    </Grommet>
  );
}
```

-

#5172 

No.

Breaking change, but fixes the property to work as intended.

### DIFF
--- a/src/js/components/CheckBox/CheckBox.js
+++ b/src/js/components/CheckBox/CheckBox.js
@@ -195,9 +195,7 @@ const CheckBox = forwardRef(
     return (
       <StyledCheckBoxContainer
         aria-label={a11yTitle}
-        // Note: StyledCheckBoxContainer is based on `styled.label`, which for
-        // some reason requires the `fill` prop to be of type string.
-        fill={fill && 'true'}
+        fillProp={fill}
         reverse={reverse}
         {...removeUndefined({ htmlFor: id, disabled })}
         checked={checked}
--- a/src/js/components/CheckBox/CheckBox.js
+++ b/src/js/components/CheckBox/CheckBox.js
@@ -195,7 +195,9 @@ const CheckBox = forwardRef(
     return (
       <StyledCheckBoxContainer
         aria-label={a11yTitle}
-        fill={fill}
+        // Note: StyledCheckBoxContainer is based on `styled.label`, which for
+        // some reason requires the `fill` prop to be of type string.
+        fill={fill && 'true'}
         reverse={reverse}
         {...removeUndefined({ htmlFor: id, disabled })}
         checked={checked}
--- a/src/js/components/CheckBox/StyledCheckBox.js
+++ b/src/js/components/CheckBox/StyledCheckBox.js
@@ -8,6 +8,7 @@ const fillStyle = () => `
       height: 100%;
       max-width: none;
       flex: 1 0 auto;
+      justify-content: space-between;
     `;
 
 const disabledStyle = `
@@ -43,10 +44,7 @@ const StyledCheckBoxContainer = styled.label`
   flex-direction: row;
   align-items: center;
   user-select: none;
-  ${props =>
-    props.fill
-      ? fillStyle() && `justify-content: space-between;`
-      : 'width: fit-content;'}
+  ${props => (props.fill ? fillStyle() : 'width: fit-content;')}
   ${props =>
     (props.pad || props.theme.checkBox.pad) &&
     edgeStyle(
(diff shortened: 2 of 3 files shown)


britt6612:
https://storybook.grommet.io/?path=/story/input-checkbox-inside-a-formfield--inside-form-field 
You can see the change in which the `fill` prop here

britt6612:
@leossantiago I do see the error now: 
Screen Shot 2021-04-19 at 9 27 58 AM

src/js/components/CheckBox/StyledCheckBox.js
```diff
@@ -8,6 +8,7 @@ const fillStyle = () => `
       height: 100%;
       max-width: none;
       flex: 1 0 auto;
+      justify-content: space-between;
```
britt6612, line 11: Thank you this is a nice clean up

src/js/components/CheckBox/CheckBox.js
```diff
@@ -195,7 +195,9 @@ const CheckBox = forwardRef(
     return (
       <StyledCheckBoxContainer
         aria-label={a11yTitle}
-        fill={fill}
+        // Note: StyledCheckBoxContainer is based on `styled.label`, which for
+        // some reason requires the `fill` prop to be of type string.
+        fill={fill && 'true'}
```
britt6612, line 200: I think this is good!

src/js/components/CheckBox/StyledCheckBox.js
```diff
@@ -43,10 +44,7 @@ const StyledCheckBoxContainer = styled.label`
   flex-direction: row;
   align-items: center;
   user-select: none;
-  ${props =>
-    props.fill
-      ? fillStyle() && `justify-content: space-between;`
-      : 'width: fit-content;'}
+  ${props => (props.fill ? fillStyle() : 'width: fit-content;')}
```
britt6612, line 47: This is a nice clean up.. I would maybe add a comment to `fillstyle` that it includes the `justify-content: space-between;` just incase it ends up being used again..

britt6612:
@leossantiago Looks good! Thank you for this! I just dropped one comment!

src/js/components/CheckBox/CheckBox.js
ericsoderberghp: If we're trying to avoid a property leaking into the DOM, we should rename it with something like `fillProp`. That allows StyledCheckBoxContainer to have access to the value without react automatically matching it to a DOM attribute.

leossantiago:
@britt6612 and @ericsoderberghp just pushed some changes addressing your suggestions! Thank you! 😄 

I'll resolve both comments. And thank you @ericsoderberghp for the heads up on the leaking props. It looks way cleaner now.

britt6612 (approved):
Better now its not getting leaked to `dom` thank you : 😊

ShimiSun (approved):
Thank you @leossantiago 💟

PR #5174, opened 2021-04-19
Text-Input can get ref

andrewcovenant:
#### What does this PR do?
It allows the text input to get ref as a prop
N/A
Manual testing
Does not need
When developing a custom component based on Grommet. Would be nice to have a ref prop on the component, In our case the typescript asks for that prop to be exported.
No
No
Not breaking anything


ShimiSun:
Why was this PR closed? What was the final resolution?

PR #5177, opened 2021-04-19
made select column pay attention to table pin property

### LINKED ISSUE #5070: DataTable onSelect does not pin header checkbox
### Expected Behavior

If caller provides `pin` on DataTable while also using `onSelect` or `select`, the checkbox in the header of the table should remain fixed with the rest of the header cells.

### Actual Behavior

The checkbox scrolls away and does not stay pinned/

https://user-images.githubusercontent.com/12522275/111372635-215e3c00-8658-11eb-8f75-c406f2cd35f3.mov

### Steps to Reproduce

1. Create DataTable
2. Apply `pin`
3. Apply `onSelect={() => {}}` (to display checkboxes)
4. Scroll

### Your Environment

- Grommet version: Stable
- Browser Name and version: Chrome Version 89.0.4389.82
- Operating System and version (desktop or mobile): Mac OS

MikeKingdom:
When a DataTable has a pinned header and selectable rows, the select checkbox column header cell (and footer cell) doesn't stay pinned on scroll like the other cells.

#### What does this PR do?

This PR changes the select checkbox column to stay pinned like other header/footer cells in a DataTable that has the `pin` property set. The background styling of the header and footer cell for the select checkbox now also comes from the theme `pinned` header and footer styling.

DataTable/Header.js

Manual testing via storybook and Jest.

See the simple example from issue #5070 

Fixes #5070 

No

Yes

Backwards compatible

### DIFF
--- a/src/js/components/DataTable/Footer.js
+++ b/src/js/components/DataTable/Footer.js
@@ -8,29 +8,7 @@ import { TableCell } from '../TableCell';
 
 import { Cell } from './Cell';
 import { StyledDataTableCell, StyledDataTableFooter } from './StyledDataTable';
-import { normalizeBackgroundColor } from './buildState';
-
-const calcBackground = (backgroundProp, pin, theme) => {
-  let background;
-  if (backgroundProp) background = backgroundProp;
-  else if (
-    pin.length > 0 &&
-    theme.dataTable.pinned &&
-    theme.dataTable.pinned.footer
-  ) {
-    background = theme.dataTable.pinned.footer.background;
-    if (!background.color && theme.background) {
-      // theme context has an active background color but the
-      // theme doesn't set an explicit color, repeat the context
-      // background explicitly
-      background = {
-        ...background,
-        color: normalizeBackgroundColor(theme),
-      };
-    }
-  } else background = undefined;
-  return background;
-};
+import { calcPinnedBackground } from './buildState';
 
 const Footer = forwardRef(
   (
@@ -61,7 +39,12 @@ const Footer = forwardRef(
           )}
           {(selected || onSelect) && (
             <StyledDataTableCell
-              background={calcBackground(background, pin, theme)}
+              background={calcPinnedBackground(
+                background,
+                pin,
+                theme,
+                'footer',
+              )}
               pin={pin}
             />
           )}
--- a/src/js/components/DataTable/Footer.js
+++ b/src/js/components/DataTable/Footer.js
@@ -45,6 +45,7 @@ const Footer = forwardRef(
                 theme,
                 'footer',
               )}
+              context="footer"
               pin={pin}
             />
           )}
--- a/src/js/components/DataTable/Footer.js
+++ b/src/js/components/DataTable/Footer.js
@@ -1,4 +1,5 @@
-import React, { forwardRef } from 'react';
+import React, { forwardRef, useContext } from 'react';
+import { ThemeContext } from 'styled-components';
 
 import { defaultProps } from '../../default-props';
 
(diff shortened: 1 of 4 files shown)


src/js/components/DataTable/Footer.js
```diff
@@ -25,34 +49,43 @@ const Footer = forwardRef(
...
-              border={border}
-              context="footer"
-              column={column}
-              datum={footerValues}
-              pad={pad}
-              pin={pin.length ? pin : undefined}
-              primaryProperty={primaryProperty}
+  ) => {
+    const theme = useContext(ThemeContext) || defaultProps.theme;
+    const pin = tablePin ? ['bottom'] : [];
```
taysea, line 54: Should it be `: undefined` if not tablePin?

src/js/components/DataTable/Header.js
```diff
@@ -88,6 +88,28 @@ const StyledContentBox = styled(Box)`
   ${props => props.extend}
 `;
 
+const calcBackground = (backgroundProp, pin, theme) => {
```
taysea, line 91: Any chance we could extract this to a reusable function that also takes `tableContext` as a parameter? then you could say `theme.dataTable.pinned[tableContext]` and re-use across Footer also

src/js/components/DataTable/Footer.js
MikeKingdom: calcBackground checks pin.length so it's expecting an array. The body of calcBackground was extracted from the original logic as-is.

src/js/components/DataTable/Header.js
MikeKingdom: I thought about it although it seems a little odd and obfuscated. Seemed like I had to go look at the function to figure out what it actually did if I had that many args

MikeKingdom: Done

src/js/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
```diff
@@ -3197,7 +3197,7 @@ exports[`DataTable custom theme 1`] = `
         class="StyledTable__StyledTableRow-sc-1m3u5g-2 "
       >
         <td
-          class="c3"
+          class="c3 "
```
ShimiSun, line 3200: Can we try to understand the root cause of this extra space and clean it up?

src/js/components/DataTable/Header.js
```diff
@@ -297,27 +298,16 @@ const Header = forwardRef(
                   </Box>
                 );
               }
-              const pin = [];
-              if (tablePin) pin.push('top');
-              if (columnPin) pin.push('left');
+              const cellPin = [...pin];
+              if (columnPin) cellPin.push('left');
```
ShimiSun, line 302: wouldn't `columnPin` always has a value of at least an empty array? Should this be refined to asking about the array's length?

MikeKingdom: columnPin is a boolean property passed in at line 296 of DataTable.js

src/js/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
MikeKingdom: The select column changed to use `StyledDataTableCell` like the other header cells rather than just `TableCell`. If you look at line 3204 which is the first `column` you see it has the same space after the class name so I'm guessing its how `StyledDataTableCell` converts the pin values into CSS (and pre-existing.) I'll poke at it a little but it doesn't seem like it matters.

That does raise this question though:
Currently/pre-existing the select header cell isn't affected by theme.dataTable.header.extend like other header cells.
Should it be?

ericsoderberghp: It should be pulling stuff from the theme.

MikeKingdom: The extra space on 3200 above (and all other existing variations on TableCell) is because TableCell passes along a className prop to StyledTableCell which ends up causing this space when className is undefined. Doesn't seem worth messing with.

MikeKingdom: The select header cell now pays attention to theme.dataTable.header.extend like other header cells

src/js/components/DataTable/Header.js
```diff
@@ -139,10 +136,16 @@ const Header = forwardRef(
...
-            <TableCell
-              background={background || cellProps.background}
+            <StyledDataTableCell
+              background={
+                calcPinnedBackground(backgroundProp, pin, theme, 'header') ||
+                cellProps.background
+              }
               plain="noPad"
               size="auto"
+              context="header"
```
taysea, line 146: Good catch.

taysea (approved):
LGTM. Good catch with the `context="header"` on the Header CheckBox.

ericsoderberghp (approved):

PR #5179, opened 2021-04-20
Allow DateInput value to initialize as []

### LINKED ISSUE #5164: DateInput with range of dates doesn't work if value starts as undefined
### Expected Behavior

DateInput with desired format of mm/dd/yyyy - mm/dd/yyyy should be able to be initialized with no value. and then when a user interacts, they can select a range.

### Actual Behavior

When no value is set, the user is only ever able to select the first date of the range. After which, the Calendar closes and the end date cannot be selected.

This is because DateInput checks the type of the incoming `value` prop. If it's not an array, then it assumes this DateInput is only aiming to support single date selection. User should be able to flag `range` as a prop on DateInput to opt into the range of dates behavior OR Calendar should support an empty array `[]` for dates.

https://codesandbox.io/s/dateinput-range-with-undefined-value-at-start-ei9sj?file=/index.js

### Steps to Reproduce

1. Click into DateInput
2. Select date
3. Calendar closes, cannot select a range.

### Your Environment

- Grommet version: Stable

taysea:
#### What does this PR do?
DateInput uses a singular `value` prop to indicate how the date or dates should be passed to the underlying Calendar. This enhances DateInput to allow caller to initialize value as an empty array `[]` which is relevant if they want the DateInput to be a range of dates but they don't want any dates to be selected to start.

Some small changes were also made to the utils function that decides when the placeholder value is shown or not to ensure the placeholder is still shown when the array is empty.

Tested in storybook on DateInput range format story. Updated value to `[]` in useState initialization.

Added jest test.

In storybook.

DateInput with desire for range had to be initialized with a value, if not only the first date of the range was able to be selected. Caller could not start with an empty range.

Closes #5164 

No.

Fixed bug in DateInput to allow value to be initialized as `[]`, which allows range of dates to start empty.

Backwards compatible.

### DIFF
--- a/src/js/components/DateInput/DateInput.js
+++ b/src/js/components/DateInput/DateInput.js
@@ -92,7 +92,9 @@ const DateInput = forwardRef(
         id={inline && !format ? id : undefined}
         range={range}
         date={range ? undefined : value}
-        dates={range ? [value] : undefined}
+        // when caller initializes with empty array, dates
+        // should be undefined
+        dates={range && value.length ? [value] : undefined}
         onSelect={
           disabled
             ? undefined
--- a/src/js/components/DateInput/DateInput.js
+++ b/src/js/components/DateInput/DateInput.js
@@ -92,8 +92,8 @@ const DateInput = forwardRef(
         id={inline && !format ? id : undefined}
         range={range}
         date={range ? undefined : value}
-        // when caller initializes with empty array, dates
-        // should be undefined
+        // when caller initializes with empty array, dates should be undefined 
+        // allowing the user to select both begin and end of the range
         dates={range && value.length ? [value] : undefined}
         onSelect={
           disabled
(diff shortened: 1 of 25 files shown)


ShimiSun (commented):
Jest isn't showing love to this test

image

taysea:
> Jest isn't showing love to this test
> 
> image

Looking into it, thanks!

src/js/components/DateInput/DateInput.js
```diff
@@ -92,7 +92,9 @@ const DateInput = forwardRef(
         id={inline && !format ? id : undefined}
         range={range}
         date={range ? undefined : value}
-        dates={range ? [value] : undefined}
+        // when caller initializes with empty array, dates
+        // should be undefined
```
halocline, line 96: I found myself asking "why should this be undefined?" Suggesting we state the desired behavior "allow user to select both start and end"

```suggestion
        // when caller initializes with empty array, dates should be undefined 
        // allowing the user to select both begin and end of the range
```

halocline (approved):
Minor comment. All else looks good.

ShimiSun (approved):

PR #5181, opened 2021-04-20
Changed Select to fix an issue with onChange + valueLabel

ericsoderberghp:
#### What does this PR do?

Changed Select to fix an issue with onChange + valueLabel

If the caller specifies a `valueLabel`, there is no `inputRef` so we shouldn't be triggering the native change event.

Select.js

Added a unit test for this. Verified that it fails without the changes.
Testing in grommet-designer.

add `valueLabel` and `onChange`.

no

yes

backwards compatible


taysea (approved):
LGTM. Good catch with valueLabel vs inputRef.

PR #5182, opened 2021-04-21
Type object added as a type on row grid prop

andrewcovenant:
#### What does this PR do?
Adds an additional type on the row property.

By comparing the Grid Documentation (scroll down to row and check the recommended types) vs the Grid/stories/Responsive.js and checking the "const rows" what is passed as a value.

The default ones

Create a reusable Grid component on top of the grommet on, in TypeScript, and assign an object to the row prop.

We have built a custom Responsive Grid component on top of the Grommet one but using TypeScript when we are passing an object on the row property, the compiler will give us "Object not assignable to type...etc". Because the one in Grommet is built in Javascript, it won't check for the type and the tests are not failing.

We have built a custom Responsive Grid component on top of the Grommet one but using TypeScript when we are passing an object on the row property, the compiler will give us "Object not assignable to type...etc". Because the one in Grommet is built in Javascript, it won't check for the type and the tests are not failing.

[Screenshot](https://ibb.co/VVDfznh)

No

Yes

Backward compatible

### DIFF
--- a/src/js/components/Grid/index.d.ts
+++ b/src/js/components/Grid/index.d.ts
@@ -98,6 +98,26 @@ export interface GridProps {
     | 'medium'
     | 'large'
     | 'xlarge'
+    | {
+      count?: 'fit' | 'fill' | number;
+      size?:
+        | 'xsmall'
+        | 'small'
+        | 'medium'
+        | 'large'
+        | 'xlarge'
+        | 'full'
+        | '1/2'
+        | '1/3'
+        | '2/3'
+        | '1/4'
+        | '2/4'
+        | '3/4'
+        | 'flex'
+        | 'auto'
+        | string
+        | string[];
+    }
     | string;
   tag?: PolymorphicType;
 }


andrewcovenant:
This PR was re-opened as discussed with @ShimiSun 
Old PR [here](https://github.com/grommet/grommet/pull/5062)

src/js/components/Grid/index.d.ts
```diff
@@ -98,6 +98,26 @@ export interface GridProps {
     | 'medium'
     | 'large'
     | 'xlarge'
+    | {
+      count?: 'fit' | 'fill' | number;
```
taysea, line 102: Reading through the discussion this thread, it doesn't seem like "rows" should accept an object with `count` and `size`. https://github.com/grommet/grommet/pull/5062#discussion_r594831535

@ShimiSun do you have any more context on why this should be supported?

ShimiSun: Good point @taysea , @andrewcovenant why did you think to add this type? It's not documented on the docs https://v2.grommet.io/grid#rows? would that even work and what would be the use case for that?

taysea: Confirming that Grommet does not currently support `{count: …, size: ….}` on row prop , but CSS grid does support the behavior trying to be achieved. It would be a separate grommet enhancement (combination of code enhancement, docs updates, and typescript updates) to allow Grommet Grid rows to know how to interpret an object the same way we’ve told columns how to do it.

andrewcovenant: @ShimiSun & @taysea, the problem is simple. Type object is not understood by the row prop, thus TypeScript will trigger a warning.

As for "size" on row prop we need it like air. Even in the documentation the example uses size: https://storybook.grommet.io/?path=/story/layout-grid-responsive-grid--responsive-grid

We can remove "count".

ShimiSun: If `size` on rows is performing as expected we should add it to the docs, and make sure its type is supported.
If `count` isn't supported in grommet, we should remove it from type support until we do support it. Thoughts?

taysea: Hi Andrew, can you provide a link to a CodeSandbox that is just JS that demonstrates that row prop knows what to do with an object? I am unable to reproduce the functionality you're describing and don't think that Grommet Grid currently supports that, so adding typescript support wouldn't make sense.

With regards to the storybook example you linked, an object is not passed to the row prop. What's being demonstrated is that there is an object `const rows` with keys that correspond to ResponsiveContext breakpoints (small, medium, ...). Later down in the code, you'll see `let rowVal` and what is happening here is the ResponsiveContext value (driven by screen width) is determining which key value pair from the rows object should be used. For example, when screen is medium, the value from `rows.medium` is passed to row prop, so it will receive a value of `['xsmall', 'xsmall']`.

Let me know if that helps clarify things!

taysea: If you just want to have a fixed size for your rows, you can do something like `rows="medium"`

andrewcovenant: @taysea , @ShimiSun : https://www.filesharesite.com/files/202105/1622204967Video_showing_lack_of_objectmp4.mp4.html

I have recorded a video showing exactly the problem and the fix

andrewcovenant:
@taysea @ShimiSun 

https://www.filesharesite.com/files/202105/1622204967Video_showing_lack_of_objectmp4.mp4.html

I have recorded a video showing exactly the problem and the fix

andrewcovenant:
Closed because of no response

PR #5184, opened 2021-04-21
Changed RadioButtonGroup to add defaultValue

ericsoderberghp:
#### What does this PR do?

Changed RadioButtonGroup to add defaultValue.
Aligns with TextInput, TextArea, Select, DateInput, etc.

RadioButtonGroup.js

Added a unit test

manually

no

yes

backwards compatible

### DIFF
--- a/src/js/components/RadioButtonGroup/RadioButtonGroup.js
+++ b/src/js/components/RadioButtonGroup/RadioButtonGroup.js
@@ -17,6 +17,7 @@ const RadioButtonGroup = forwardRef(
   (
     {
       children,
+      defaultValue,
       disabled,
       name,
       onChange,
@@ -42,7 +43,11 @@ const RadioButtonGroup = forwardRef(
         : { disabled, ...o },
     );
 
-    const [value, setValue] = formContext.useFormInput(name, valueProp, '');
+    const [value, setValue] = formContext.useFormInput(
+      name,
+      valueProp,
+      defaultValue || '',
+    );
 
     const [focus, setFocus] = useState();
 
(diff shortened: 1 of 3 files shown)


jcfilben (commented):
Looks good, tested with storybook and functionality is working as expected. Shouldn't `defaultValue` be added to the RadioButtonGroup docs though?

ericsoderberghp:
> Looks good, tested with storybook and functionality is working as expected. Shouldn't `defaultValue` be added to the RadioButtonGroup docs though?

We haven't been adding all standard and straightforward DOM attributes where we don't do anything notable to our documentation. For instance, we describe `defaultValue` for Select and DateInput because we have something unique to say but for TextInput and TextArea we don't since they aren't special in any way.

src/js/components/RadioButtonGroup/RadioButtonGroup.js
```diff
@@ -17,6 +17,7 @@ const RadioButtonGroup = forwardRef(
   (
     {
       children,
+      defaultValue,
```
ShimiSun, line 20: I think we need to add `defaultValue` to doc.js.

ericsoderberghp: Should we add it to TextInput and TextArea too?

ShimiSun: I think we should, otherwise, how would users be aware of this optional setting?

ericsoderberghp: It's a standard DOM attribute. So, just like you can put an `id` on any component, we don't document it.

ShimiSun (approved):

Issue #5004, opened 2021-02-25
Need a new prop to Form so that Form can trigger validation after first time Form is rendered


shen6653:
### Expected Behavior

When Form is used in Wizard step, if user types something wrong, they go back to previous step then go back the same step again, they are expecting to see a validation error.

### Actual Behavior
No validation error is displayed.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

ericsoderberghp:
Instead of adding a new prop, Mandy, Yash, Scott, Mike, and I decided on tying it to when validation is "blur" or "change" and only on fields that have a value.  This aligns with the overall design approach when disabling the submit button if the form cannot be submitted. In that approach, we want invalid fields to always be indicated as such in some way. Required fields already have an annotation indicating that they are required. So, it's only fields with invalid data in them that need validation messages.

PR #5187, opened 2021-04-24
Run validations on form when component (Formfield) mounts #5049

ypinneka:
**Continuation of PR#5049 as the old branch had some issues**

What does this PR do?
Run validations on component mount.
Story: Let's say there are forms in two pages, user first finishes form in first page & visits the form in second page, while entering a field, he needs to copy a field value from first page form, so he returns to page one & comes back to page two. Now, page two loads with application preserving already entered field values, but this time there are no validations. By this PR, application can trigger validations on component load.

Form.js

In Process. Created a PR for feedback on current implementation

What are the relevant issues?
#5004

Yes

Yes

Not a breaking change


ericsoderberghp (approved):

src/js/components/Form/Form.js
```diff
@@ -80,6 +80,51 @@ const Form = forwardRef(
...
+          const nextValidationResults = {
+            errors: nextErrors,
+            infos: nextInfos,
+            valid: buildValid(nextErrors),
+          };
+          if (onValidate) onValidate(nextValidationResults);
+          return nextValidationResults;
+        });
+      }
+    }, []);
```
halocline, line 126: When preparing for the next Grommet release, I see ESLint throwing a warning that this useEffect is missing dependencies in the dependency array.

@ypinneka Can you confirm that this useEffect is only supposed to run only on initial mount?

I think we are trying to include all dependencies in the dependency array and handle conditions within the effect. Any insights you can share about the desired intent would be helpful.

cc: @ericsoderberghp

halocline: Just saw https://github.com/grommet/grommet/pull/5195 which is related.

PR #5188, opened 2021-04-25
Fix for missing code in Grid Pagination storybook

bojan88:
Signed-off-by: Bojan Durdevic <bojan88@gmail.com>

#### What does this PR do?
The code is missing for Pagination -> Grid example. This PR is fixing the issue.

I couldn't fine any issues opened for this.

Before:
image

---

After:
image

No

No

### DIFF
--- a/src/js/components/Pagination/stories/Grid.js
+++ b/src/js/components/Pagination/stories/Grid.js
@@ -1,6 +1,13 @@
 import React, { useState } from 'react';
 
-import { Box, Card, Grid, Grommet, Pagination, Text } from 'grommet';
+import {
+  Box,
+  Card,
+  Grommet,
+  Pagination,
+  Text,
+  Grid as GrommetGrid,
+} from 'grommet';
 import { grommet } from 'grommet/themes';
 
 const data = [];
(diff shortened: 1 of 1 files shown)


jcfilben (approved):

src/js/components/Pagination/stories/Grid.js
```diff
@@ -17,7 +24,7 @@ const CardResult = ({ item }) => (
   </Card>
 );
 
-const PaginatedGrid = () => {
+export const Grid = () => {
```
ericsoderberghp, line 27: I think we should keep the use of `Grid` to refer to the grommet component, as that will work well if cutting and pasting parts of the code and be less confusing.

bojan88: That's a good point. I'll try to figure out what exactly is the problem here.
What do you think about renaming this story as a last resort solution?

ericsoderberghp: I think we can do something similar to this: https://github.com/grommet/grommet/blob/master/src/js/components/DropButton/stories/typescript/Calendar.tsx

bojan88: Thanks Eric. This works great.

ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution!

PR #5190, opened 2021-04-26
Keep the errors/infos format consistent 

ypinneka:
**Continuation of PR#5187**

#### What does this PR do?
Needed to keep the errors/infos format consistent even when validation is run on component mount. In PR#5187, I did not filter any errors or infos that we don't have any validations.

Form.js

Ran all the current test cases

Check for returned validation results object

#5004 

Not a breaking change


ypinneka:
@ericsoderberghp  I have missed this in my last PR https://github.com/grommet/grommet/pull/5187

src/js/components/Form/Form.js
```diff
@@ -94,6 +94,22 @@ const Form = forwardRef(
       return valid;
     };
 
+    // Remove any errors that we don't have any validations for anymore.
+    const filterErrorValidations = errors => {
+      const nextErrors = errors;
+      return Object.keys(nextErrors)
+        .filter(n => !validations.current[n] || nextErrors[n] === undefined)
+        .map(n => delete nextErrors[n]);
```
ericsoderberghp, line 102: I don't think we want `map` here, since nothing is depending on the return value. I suspect `forEach` would be better.

ericsoderberghp (commented):
Can you also add a unit test for this scenario, where a field is dynamically removed?

ypinneka:
> Can you also add a unit test for this scenario, where a field is dynamically removed?

We already have this case covered in form-test-uncontrolled.js, `dynamicly removed fields using blur validation
  don't keep validation errors`

src/js/components/Form/Form.js
ypinneka: done

ericsoderberghp:
> > Can you also add a unit test for this scenario, where a field is dynamically removed?
> 
> We already have this case covered in form-test-uncontrolled.js, `dynamicly removed fields using blur validation don't keep validation errors`

Why didn't the test fail before these changes? Are we not testing with pre-validation errors + dynamic fields + asserting on the details of the event?

ypinneka:
> > > Can you also add a unit test for this scenario, where a field is dynamically removed?
> > 
> > 
> > We already have this case covered in form-test-uncontrolled.js, `dynamicly removed fields using blur validation don't keep validation errors`
> 
> Why didn't the test fail before these changes? Are we not testing with pre-validation errors + dynamic fields + asserting on the details of the event?

I was not checking valid flag for `validate on mount` case earlier, adding now. The changes for dynamic fields are in place for change/blur, so tests were passing

ericsoderberghp (approved):

PR #5191, opened 2021-04-27
Pagination update on page prop change

bojan88:
#### What does this PR do?
Update pagination page state on page prop change. This will allow users to update the page from the parent component.

Tested manually and added 1 unit test.

Maybe editing one of the stories to verify this.

I tested pagination components in 3 other UI frameworks.
They are updating the page on prop change. Also, they are not calling `onChange` on prop change.
One of the drawbacks for this change is an excess component re-render on click. I think we can leave it as is since it's not impacting performance a lot, but saving bytes with simple logic.

I didn't find any relevant issues.

Maybe as a bug fix

### DIFF
--- a/src/js/components/Pagination/Pagination.js
+++ b/src/js/components/Pagination/Pagination.js
@@ -1,4 +1,4 @@
-import React, { forwardRef, useContext, useState } from 'react';
+import React, { forwardRef, useContext, useEffect, useState } from 'react';
 import styled, { ThemeContext } from 'styled-components';
 import { defaultProps } from '../../default-props';
 import { Box } from '../Box';
@@ -176,6 +176,10 @@ const Pagination = forwardRef(
       separator: control === 'more-prev' || control === 'more-next',
       ...navProps[control],
     }));
+
+    useEffect(() => {
+      setActivePage(pageProp || 1);
+    }, [pageProp]);
 
     return (
       <StyledPaginationContainer {...theme.pagination.container} {...rest}>
--- a/src/js/components/Pagination/Pagination.js
+++ b/src/js/components/Pagination/Pagination.js
@@ -42,6 +42,10 @@ const Pagination = forwardRef(
       Math.min(pageProp, totalPages) || 1,
     );
 
+    useEffect(() => {
+      setActivePage(pageProp || 1);
+    }, [pageProp]);
+
     /* Define page indices to display */
     const beginPages = getPageIndices(1, Math.min(numberEdgePages, totalPages));
     const endPages = getPageIndices(
@@ -176,10 +180,6 @@ const Pagination = forwardRef(
       separator: control === 'more-prev' || control === 'more-next',
       ...navProps[control],
     }));
-
-    useEffect(() => {
-      setActivePage(pageProp || 1);
-    }, [pageProp]);
 
     return (
       <StyledPaginationContainer {...theme.pagination.container} {...rest}>
--- a/src/js/components/Pagination/__tests__/Pagination-test.js
+++ b/src/js/components/Pagination/__tests__/Pagination-test.js
@@ -347,14 +347,15 @@ describe('Pagination', () => {
   });
 
   test(`should change the page on prop change`, () => {
-    const { container, rerender, getByText } = render(
+    const { container, rerender } = render(
       <Grommet>
         <Pagination numberItems={NUM_ITEMS} page={1} />
       </Grommet>,
     );
 
-    expect(getByText('1').className).not.toBe(getByText('3').className);
-    expect(getByText('2').className).toBe(getByText('3').className);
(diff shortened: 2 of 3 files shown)


halocline:
👀

halocline (approved):
Looks good. Thank you for your contribution.

src/js/components/Pagination/__tests__/Pagination-test.js
```diff
@@ -345,4 +345,25 @@ describe('Pagination', () => {
...
   });
+
+  test(`should change the page on prop change`, () => {
+    const { container, rerender, getByText } = render(
+      <Grommet>
+        <Pagination numberItems={NUM_ITEMS} page={1} />
+      </Grommet>,
+    );
+
+    expect(getByText('1').className).not.toBe(getByText('3').className);
```
taysea, line 356: Just a small suggestion to align with the convention of other tests in this file. We typically don't reference `className` In this case, maybe you could do something like the test above:
```
const desiredPage = '1';
const activePage = container.querySelector(`[aria-current="page"]`)
      .innerHTML;
expect(activePage).toEqual(desiredPage);

File: src/js/components/Pagination/Pagination.js [hunk 2]
```diff
@@ -177,6 +177,10 @@ const Pagination = forwardRef(
       ...navProps[control],
     }));
 
+    useEffect(() => {
```
taysea, line 180: For readability, I'm wondering if we should move this up to line 44 so it is next to where `activePage` is created.

taysea (commented):
Just a couple cleanup comments but functionality seems to be working well. I created an example where a Button outside of pagination component was incrementing the page state value.

src/js/components/Pagination/__tests__/Pagination-test.js
bojan88: Thanks. I made it a bit more concise, hope you don't mind. Let me know please if you think I should make it exactly as posted above.

src/js/components/Pagination/Pagination.js
bojan88: My reasoning to move them at the end was to easily extend it latter if needed, and add dependencies declared above.
But I agree it makes more sense to have ti close to `activePage`

taysea (approved):
Nice, thanks for incorporating the feedback!

ericsoderberghp:
Thanks for your contribution!

mfferreira:
Thanks for this! The `page` prop value update not updating the active page was driving me nuts already.

bojan88:
@mfferreira I don't quite understand your comment. `page` prop should be updating internal `activePage` in `Pagination` component. Maybe you should open an issue with a better explanation. I'll be happy to work on that, just mention me in the issue comments.

B-Esmaili:
When this change should be released?

ShimiSun:
I'm thinking the upcoming release will be in the next couple of weeks. 
@taysea, @halocline  Is there a workaround we can provide? Otherwise, we may want to re-prioritize the upcoming release.

B-Esmaili:
@halocline What I want to do is changing the page from a higher-order-component. I have already filed an [issue ](https://github.com/grommet/grommet/issues/5232)regarding the issue and I have included a link to code-sandbox link demonstrating the issue.

taysea:
@B-Esmaili I believe Pagination is working as expected. I have tested locally in grommet storybook with the following setup:
```
const [page, setPage] = useState(1);
return (
<>
<Pagination numberItems={237} page={page} />
          <Button
            label="Increment Pagination"
            onClick={() => setPage(page + 1)}
          />
</>
```

The external Button is controlling the value of the pagination. Notice as I click, the active page increases and the internal state of pagination is changing as well:

https://user-images.githubusercontent.com/12522275/117837968-bfbce700-b22e-11eb-9ada-d58c1e438fdc.mov


https://user-images.githubusercontent.com/12522275/117837984-c2b7d780-b22e-11eb-8642-eff91335ef2b.mov

Can you explain how this is not meeting your use case?

B-Esmaili:
@taysea thanks for concerning the issue, did you see my [sandbox](https://codesandbox.io/s/bold-flower-6ic42)? And yes it is working now in the latest version, but did you test it in v2.17.1?

taysea:
> @taysea thanks for concerning the issue, did you see my [sandbox](https://codesandbox.io/s/bold-flower-6ic42)? And yes it is working now in the latest version, but did you test it in v2.17.1?

This fix is currently only on our stable branch, so it will be available in the next release. If you'd like to point your project stable you can try out the fix with these steps: https://github.com/grommet/grommet/wiki/What-is-grommet-stable-and-how-to-use-it%3F

B-Esmaili:
@taysea Thanks for the advice, I'll do it accordingly.

PR #5192, opened 2021-04-27
feat: adds CurrencyInput component

alansouzati:
#### What does this PR do?
adds a new component called CurrencyInput

storybook examples and components/CurrencyInput
manual and automated
open storybook examples and tinker with the 3 examples added
yes
yes
backwards compatible

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
   "dependencies": {
     "grommet-icons": "^4.5.0",
     "hoist-non-react-statics": "^3.2.0",
+    "locale-currency": "^0.0.2",
     "markdown-to-jsx": "^6.11.4",
     "polished": "^3.4.1",
     "prop-types": "^15.7.2",
--- a/src/js/components/CurrencyInput/CurrencyInput.js
+++ b/src/js/components/CurrencyInput/CurrencyInput.js
@@ -1,0 +1,136 @@
+import React, { forwardRef, useCallback, useMemo } from 'react';
+import { getCurrency } from 'locale-currency';
+
+import { TextInput } from '../TextInput';
+
+const currentLocale =
+  typeof window === 'undefined'
+    ? 'en-US'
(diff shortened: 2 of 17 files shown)


package.json
```diff
@@ -51,6 +51,7 @@
   "dependencies": {
     "grommet-icons": "^4.5.0",
     "hoist-non-react-statics": "^3.2.0",
+    "locale-currency": "^0.0.2",
```
ShimiSun, line 54: image

Doesn't look too bad, yet another package to maintain.

ericsoderberghp:
- I like having the ability for intelligent numeric separators. I think we should have this capability for numbers in general, not just for currency.
- It would be nice to have the onChange provide both the text and numeric values, so the caller doesn’t also have to parse the text to extract the number.
- I’d be more inclined to explore adding this to MaskedInput, to see how that looks.
- we’re a bit reluctant to add more input components, as that incurs additional maintenance costs (minor)

ShimiSun (commented):
Thank you for this PR, it drove a good conversation! 
In addition to eric's comments above, I'm adding some comments regarding the deployment/functionality

1. I think there is a small bug when numbers are entered after the dot, as you add a number it keeps increasing the value in a decimal value instead of typing the number.
2. On the "Simple" example, it is not clear that the user should skip the typing of the dollar sign.
3. I think we must have an example of each input component inside of a Form.

alansouzati:
No worries, I will have this component on our end. Thanks for considering it.

package.json
alansouzati: this package is a simple map of locale to currency, if you are concerned about the extra dependency I think there are maps available out there

```
{
   "pt-BR": "BRL",
  "pt": "EUR",
  "en-UK": "EUR",
  "en-US": "USD",
  "en": "USD"
}
```

Issue #4895, opened 2021-01-24
Grommet fails to run with snowpack


jeferson-sb:
I have a pretty simple application with Grommet, Styled Components and Snowpack installed
But when I start the app I'm getting the following message:

![image](https://user-images.githubusercontent.com/30840709/105618861-d0a23380-5dca-11eb-9298-f7523513a892.png)

I tried to install a different version of styled-components (above 5.1) but it also doesn't work.
Snowpack with styled-components alone seems to work fine though. 

### Expected Behavior

The application should launch without any issues.

### Actual Behavior

I get a `Uncaught TypeError: (0 , _styledComponents.default) is not a function` error message

### URL, screen shot, or Codepen exhibiting the issue

Codesandbox: https://codesandbox.io/s/grommet-snowpack-repro-o3zks

### Your Environment

- Grommet version: 2.16.2
- Browser Name and version: Microsoft Edge 87
- Operating System and version (desktop or mobile): Desktop Windows 10

ShimiSun:
Were you able to make it work with only snowpack & styled-components, to make sure this issue is caused by using grommet?

jeferson-sb:
@ShimiSun Yes, it does work with only snowpack & styled-components
I have a example [here](https://codesandbox.io/s/snowpack-styled-components-react-jdx3w?file=/src/components/Button/styles.js)

garrettlr:
@ShimiSun @jeferson-sb can also confirm this issue, and that importing styled-components directly works fine.

ShimiSun:
> The root cause of the issue is likely that grommet mixes require and import statements improperly:

@garrettlr thanks for the inputs! Do you have any resources to learn about the issue? why did you get the impression it causes the problem?

garrettlr:
@ShimiSun give me an hour or two and I'll find a few github posts that helped, but here's my process: 

1. Start up any, react + snowpack app - a simple "hello world" is enough. 
2. Bring in Grommet + dependencies, then import that into your app. 
3. At this point, you should see the above error. 
4. (the magic) pick any random Grommet component, Comment out the 'if' statement from the above snippet, directly import only that file, and it will import properly.

wiltsu:
`Uncaught TypeError: (0 , _styledComponents.css) is not a function`
I'm having almost the same issue...

rabaut:
Thank you for doing that @garrettlr ! I'll be using your fork until this is fixed.

woss:
[...]

PR #5202, opened 2021-04-30
BoxSnowpack test component -- Revert before next release

jcfilben:
#### What does this PR do?
This is a test component to be used for testing a snowpack project using Grommet. It will be removed after testing and will never be officially released.

The component is exactly the same as the Box component but it doesn't have `BoxDoc = require('./doc').doc(Box);`. I think that the require statements in the component files are causing problems with snowpack. The goal of this component is to test if removing the require statement will solve the problems with using grommet and snowpack as discussed in [this issue](https://github.com/grommet/grommet/issues/4895) or if there are other problems that need to be fixed as well to get grommet working in a snowpack project.

I also included a doc.js file for this component to see how the documentation for BoxSnowpack will behave without the require statement. Originally I thought that the require statement was being used by the grommet-site for the component docs. However, the code that imports the docs for each component is `import { doc, themeDoc } from 'grommet/components/Grommet/doc';` so the documentation isn't coming from the require statement with the <component>.js file so I am curious to see if the documentation for BoxSnowpack will work normally without the require statement.

Related to #4895 

No

No

Backwards compatible

### DIFF
--- a/src/js/components/BoxSnowpack/BoxSnowpack.js
+++ b/src/js/components/BoxSnowpack/BoxSnowpack.js
@@ -1,0 +1,169 @@
+import React, {
+  Children,
+  forwardRef,
+  useContext,
+  useMemo,
+  useState,
+} from 'react';
+
+import { ThemeContext } from 'styled-components';
+import { defaultProps } from '../../default-props';
+import { backgroundIsDark } from '../../utils';
+import { Keyboard } from '../Keyboard';
+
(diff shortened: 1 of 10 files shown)


jcfilben:
DateInput snapshot tests were failing because it is a new month. For now I just updated the snapshots but eventually these tests should be rewritten so that the snapshots don't fail at the start of every new month. The UI test differences are also related to the month change

PR #5212, opened 2021-05-03
Changed List to add onOrder

ericsoderberghp:
#### What does this PR do?

Adds `onOrder` property to List.
This adds controls to re-order the data items. Buttons are added for accessible access. Drag and drop is added for mouse convenience.

Check out the story to see how it works

Just a story so far, still need unit tests

storybook

Related to https://github.com/grommet/hpe-design-system/issues/1595

yes

eventually

backwards compatible


src/js/components/List/List.js
```diff
@@ -262,18 +350,128 @@ const List = React.forwardRef(
...
+                    },
+                    onDrop: () => {
+                      if (orderingData) {
+                        onOrder(orderingData);
+                      }
+                    },
+                    ref: dragging === index ? draggingRef : undefined,
+                  };
+
+                  const Up = theme.list.icons.up;
```
britt6612, line 392: This should be added to the docs under the theme section so users are aware they can change these.
`list.icons.up`

src/js/components/List/List.js
```diff
@@ -39,6 +40,9 @@ const StyledList = styled.ul`
 
 const StyledItem = styled(Box)`
   ${props => props.onClick && `cursor: pointer;`}
+  ${props =>
+    props.draggable &&
+    `cursor: move;`}
```
britt6612, line 45: Looking ahead at designs for ordering of items did we want the cursor to be `cursor: grab`?

MikeKingdom: Eric and I talked about this and decided to use the `move` cursor instead. Typically, `move` is used more for moving items around (look at github's backlog page as an example.) Hand and grab are more for when you're manipulating something that will change it's properties like edit, drop in a trashcan, etc.

src/js/components/List/List.js
```diff
@@ -66,6 +70,43 @@ const normalize = (item, index, property) => {
...
+  const result = array.slice(0);
+  const tmp = result[source];
+  if (source < target)
+    for (let i = source; i < target; i += 1) result[i] = result[i + 1];
+  else for (let i = source; i > target; i -= 1) result[i] = result[i - 1];
+  result[target] = tmp;
+  return result;
+};
+
+const getPrimaryContent = (item, index, primaryKey) => {
```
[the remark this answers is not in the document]
MikeKingdom: Certainly. totally agree.

src/js/components/List/List.js
```diff
@@ -105,32 +148,74 @@ const List = React.forwardRef(
...
+    
+    
+    if (active >= 0) {
+      let activeId;
+      // We have an item that is 'focused' within the list. This could
+      // be the list item or one of the up/down ordering buttons.
+      // We need to figure out an id of the thing that will be shown as active
+      if (onOrder) {
+        // figure out which arrow button will be the active one.
+        const btnId = (active % 2) ? 'MoveDown' : 'MoveUp';
```
ericsoderberghp, line 167: grommet tends to avoid compressing variable names, suggest 'btnId' => 'buttonId'.

src/js/components/List/List.js
```diff
@@ -66,6 +70,43 @@ const normalize = (item, index, property) => {
...
+    } else {
+      primaryContent = normalize(item, index, primaryKey);
+    }
+  }
+
+  return primaryContent;
+};
+
+
+const chooseKey = (item, index, primaryContent) => {
```
ericsoderberghp, line 97: grommet convention would call these 'getKey' and 'getItemId'

src/js/components/List/List.js
```diff
@@ -93,6 +135,7 @@ const List = React.forwardRef(
     const theme = useContext(ThemeContext);
     const [active, setActive] = useState();
```
ericsoderberghp, line 136: Can we had a comment here explaining what 'active' means, since it has some interesting nuances in it when ordering?

src/js/components/List/List.js
```diff
@@ -105,32 +148,74 @@ const List = React.forwardRef(
...
-                  event.persist();
-                  const adjustedEvent = event;
-                  adjustedEvent.item = data[active];
-                  adjustedEvent.index = active;
-                  onClickItem(adjustedEvent);
+                  if (onOrder) {
+                    const index = Math.trunc(active / 2);
+                    if (active % 2) {
+                      onOrder(reorder(data, index, index + 1));
+                      setActive(Math.min(active + 2, data.length * 2 - 2));
```
ericsoderberghp, line 188: Maybe a comment here explaining that we are adjusting `active` so that the same control on the moved item remains active after the reorder?

ericsoderberghp (commented):
Nice. Just a few naming suggestions.

ericsoderberghp:
I approve, I just don't get a button to declare that. :)

src/js/components/List/__tests__/List-test.js
```diff
@@ -21,6 +21,7 @@ describe('List', () => {
     const { container, getByText } = render(
       <Grommet>
         <List
+          aria-label="List"
```
halocline, line 24: Is their value to adding Jest tests for the various onOrder functionality? Thinking about:
- keyboard interactions
- getByText, mouseover up button, click button, confirm reorder occurred ...
- something to provide developer feedback if they are combining onOrder with paginate and/or onClickItem (test would confirm the feedback, console warning or the like would need to be implemented) --> this is documented, but does something like this improve the developer experience? --> is this overthinking?

src/js/themes/base.js
```diff
@@ -997,6 +997,10 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         pad: { horizontal: 'medium', vertical: 'small' },
         // extend: undefined,
       },
+      icons: {
+        down: FormDown,
+        up: FormUp,
+      },
```
halocline, line 1003: add these as types to `base.d.ts`?

halocline (commented):
Functionality, behavior, accessibility, implementation, clean API -- all look great.

- Question re: value of hardening with tests
- Theme type definitions for up/down icons

src/js/themes/base.js
MikeKingdom: Good catch.

src/js/components/List/__tests__/List-test.js
MikeKingdom: I implemented the tests for keyboard and mouse onOrder functionality.
I didn't put in a warning for the onClickItem since it seems reasonable to still specify onClickItem even though you temporarily specify onOrder to let it be reordered (e.g. the example in the figma)

halocline (approved):

PR #5213, opened 2021-05-04
Add `badge` prop to Button

### LINKED ISSUE #5193: Enhance Button to have `badge` prop
### Expected Behavior

Enhance Button with a badge prop to allow "badge" behavior out of the box. Right now this behavior can be achieved with Stack but it requires a heavy bit of JSX, handling via a prop would be convenient as this is a common pattern. Some use cases to consider:
Screen Shot 2021-04-27 at 10 51 15 AM
Screen Shot 2021-04-27 at 10 50 24 AM

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

taysea:
#### What does this PR do?
This PR adds `badge` as a prop to button. The badge is an indicator that appears on the top-right corner of the Button.

```
`badge` can take the following forms:
badge={2} // typeof number
badge={{background?: BackgroundType, max?: number, value?: number | boolean}}
badge={Custom JSX}
badge // boolean
```

To keep consistent dimensions, the badge has a min-width and min-height of the theme's baseSpacing, but will grow to fit the content when necessary. Adjustments to width/height are handled in useEffects that reference a `ref` placed on the badge container.

Note: Snapshot changes are merely re-ordering of classes, but the values have not changed.

src/js/components/Button/Button.js

Created Badge story and added jest tests.

In local storybook.

Closes #5193 

Updated here.

Enhanced Button to have `badge` prop

Backwards compatible.


src/js/themes/base.js
```diff
@@ -397,6 +397,13 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       // extend: undefined,
     },
     button: {
+      badge: {
```
britt6612, line 400: add these to `base.d.ts`

taysea: Thank you! good reminder

ShimiSun (commented):
The layout doesn't look great on Chromatic 
image
https://www.chromatic.com/test?appId=5d9774839a6eff00203f5cbf&id=6090986e7cf1220039a9d51d

ericsoderberghp:
> The layout doesn't look great on Chromatic
> image
> https://www.chromatic.com/test?appId=5d9774839a6eff00203f5cbf&id=6090986e7cf1220039a9d51d
> 
> @ericsoderberghp thoughts? The reason the badge is not on the icon directly is because it is offset from the button corner (not icon). wondering if we want to rethink the offset approach?

The tricky part is having visibility into where to find the right "corner". Different icons have different visible corners. And, a Button with children rendering is even harder to figure out. I'm open to exploring to see if we can find something though.

src/js/components/Button/Button.js
```diff
@@ -343,9 +370,87 @@ const Button = forwardRef(
...
+      let value;
+      if (typeof badgeProp === 'number') value = badgeProp;
+      else if (typeof badgeProp === 'object') value = badgeProp.value;
+
+      let badge;
+      if (value || typeof badgeProp === 'boolean') {
+        if (value) {
+          badge = (
+            <Text color="text-strong" size="small" ref={badgeRef}>
+              {value > max ? `${max}+` : value}
```
halocline, line 414: Are we okay here if max is not defined by neither the badgeProp nor the theme?

src/js/components/Button/doc.js
```diff
@@ -44,6 +45,22 @@ export const doc = Button => {
...
+      PropTypes.shape({
+        background: backgroundPropType,
+        max: PropTypes.number,
+        value: PropTypes.number,
+      }),
+    ])
+      .description(
+        `An indicator to show on the top-right of the Button. For accessibility,
+         add an aria-label that provides a useful message to screen readers. 
+         For example, "2 unread notifications".`,
```
halocline, line 61: Are we encouraging callers to use `aria-label` over `a11yTitle`? I know we are supporting `aria-label` but `a11yTitle` is the documented prop.

Also, would this be more clear to reference "add an a11yTitle to the Button...?" If I wasn't as familiar, I might assume that was something I should call from the badge.
```suggestion
        `An indicator to show on the top-right of the Button. For accessibility,
         supplement the badge by adding an a11yTitle to the Button which 
         provides a useful message to screen readers. For example, "2 unread 
         notifications".`,
```

src/js/components/Button/doc.js
```diff
@@ -92,7 +109,9 @@ with plain Buttons.`,
     href: PropTypes.string.description(
       'If specified, the button will behave like an anchor tag.',
     ),
-    icon: PropTypes.element.description('Icon element to place in the button.'),
+    icon: PropTypes.element.description(`Icon element to place in the button. 
+    For accessibility with screen readers, if using just an icon (no label), 
+    add an aria-label that describes the icon.`),
```
halocline, line 114: Similar question/comment related to `aria-label` vs `a11yTitle` plus specificity as to where to add.

src/js/components/Button/doc.js
```diff
@@ -251,6 +270,30 @@ export const themeDoc = {
...
+    defaultValue: 'brand',
+  },
+  'button.badge.max': {
+    description: `The max number to display in the badge. If max is 9 and badge 
+    has value of 10, 9+ will be displayed.`,
+    type: 'number',
+    defaultValue: undefined,
+  },
+  'button.badge.size.medium': {
+    description: `The minimum width and height of the badge.`,
```
halocline, line 293: Minimum when containing content? What if badge={true}?

src/js/components/Button/index.d.ts
```diff
@@ -17,6 +17,10 @@ import { TipProps } from '../Tip';
 export interface ButtonProps {
   a11yTitle?: A11yTitleType;
   alignSelf?: AlignSelfType;
+  badge?:
```
halocline, line 20: JSX.Element as well?

src/js/components/Button/stories/Badge.js
```diff
@@ -0,0 +1,55 @@
...
+          <Box background="status-critical" pad="xsmall" round="xsmall">
+            <Alert size="small" />
+          </Box>
+        }
+        hoverIndicator
+        {...props}
+      />
+      <Button
+        aria-label="23 Unread Notifications"
+        icon={<Notification />}
```
halocline, line 43: Similar comment to @ShimiSun about positioning on icon only buttons is odd, but don't currently have any immediate recommendations, unless we were to conditionally move the Stack inside of the styledButtonResult or set the guiding child differently.

src/js/components/Button/stories/Badge.js
```diff
@@ -0,0 +1,55 @@
...
+        badge
+        hoverIndicator
+        {...props}
+      />
+    </Box>
+  </Grommet>
+);
+
+export default {
+  title: 'Controls/Button/Badge',
```
halocline, line 54: Is there any merit to including / augmenting a Typescript story implementation as well? Would that help proactively catch type errors?

src/js/components/Button/index.d.ts
taysea: Good catch.

src/js/components/Button/stories/Badge.js
taysea: Great idea. Going to make this story a `.tsx` instead.

src/js/components/Button/Button.js
```diff
@@ -343,9 +370,87 @@ const Button = forwardRef(
...
+    // if badge has a number, we want to scale the badge to fit.
+    // width will also be used to know how far to horizontally offset the badge
+    useEffect(() => {
+      if (badgeRef && badgeRef.current && typeof badgeProp !== 'boolean') {
+        setBadgeWidth(getDimension('width', badgeProp, badgeRef, theme));
+      }
+    }, [badgeProp, theme]);
+
+    const [badgeHeight, setBadgeHeight] = useState(defaultBadgeDimension);
+    // if badge has custom JSX, we want to scale the badge to fit.
```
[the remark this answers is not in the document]
taysea: Yes, agreed. we don't need.

src/js/components/Button/Button.js
```diff
@@ -92,12 +96,34 @@ const getIconColor = (paths = [], theme, colorProp, kind) => {
   return result[1] || undefined;
 };
 
+const getDimension = (dimension, badgeProp, badgeRef, theme) => {
```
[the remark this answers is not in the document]
taysea: Good suggestion. I'll change.

src/js/components/Button/Button.js
taysea: Yes, we check if there is a value before comparing to max. If so, value > undefined will resolve to false

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Greater_than#comparing_boolean_null_undefined_nan

src/js/components/Button/Button.js
```diff
@@ -343,9 +370,87 @@ const Button = forwardRef(
...
-        return <Tip content={tip}>{styledButtonResult}</Tip>;
+        styledButtonResult = <Tip content={tip}>{styledButtonResult}</Tip>;
       }
-      return <Tip {...tip}>{styledButtonResult}</Tip>;
+      styledButtonResult = <Tip {...tip}>{styledButtonResult}</Tip>;
+    }
+
+    const defaultBadgeDimension =
+      typeof badgeProp === 'boolean'
+        ? // empty badge should be smaller
```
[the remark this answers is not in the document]
taysea: Yes, this is aesthetic but I've added a comment now to capture that.

src/js/components/Button/Button.js
```diff
@@ -343,9 +370,87 @@ const Button = forwardRef(
...
+          `${parseInt(theme.button.badge.size.medium.replace('px', ''), 10) /
+            2}px`
+        : theme.button.badge.size.medium;
+
+    const [badgeWidth, setBadgeWidth] = useState(defaultBadgeDimension);
+    // if badge has a number, we want to scale the badge to fit.
```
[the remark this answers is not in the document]
taysea: Done.

src/js/components/Button/Button.js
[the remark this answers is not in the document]
taysea: Done.

src/js/components/Button/Button.js
```diff
@@ -343,9 +370,87 @@ const Button = forwardRef(
...
+    // width will also be used to know how far to horizontally offset the badge
+    useEffect(() => {
+      if (badgeRef && badgeRef.current && typeof badgeProp !== 'boolean') {
+        setBadgeWidth(getDimension('width', badgeProp, badgeRef, theme));
+      }
+    }, [badgeProp, theme]);
+
+    const [badgeHeight, setBadgeHeight] = useState(defaultBadgeDimension);
+    // if badge has custom JSX, we want to scale the badge to fit.
+    // height will also be used to know how far to horizontally offset the badge
```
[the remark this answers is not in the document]
taysea: Done.

src/js/components/Button/doc.js
taysea: Done. Yes, a11yTitle is the one we document so we should use that.

src/js/components/Button/doc.js
taysea: Good clarification. Adding that.

src/js/components/Button/index.d.ts
taysea: Done.

src/js/components/Button/stories/Badge.js
taysea: Done.

src/js/components/Button/stories/Badge.js
taysea: Took a new approach here now that allows caller to specify target: 'contents' to indicate they want the offset relative to the button contents as opposed to the outer edges of the button.

src/js/components/Button/doc.js
taysea: Updated to a11yTitle since this is what we document.

src/js/components/Button/doc.js
taysea: Done.

src/js/components/Button/Button.js
```diff
@@ -343,9 +370,87 @@ const Button = forwardRef(
...
+    const defaultBadgeDimension =
+      typeof badgeProp === 'boolean'
+        ? // empty badge should be smaller
+          `${parseInt(theme.button.badge.size.medium.replace('px', ''), 10) /
+            2}px`
+        : theme.button.badge.size.medium;
+
+    const [badgeWidth, setBadgeWidth] = useState(defaultBadgeDimension);
+    // if badge has a number, we want to scale the badge to fit.
+    // width will also be used to know how far to horizontally offset the badge
```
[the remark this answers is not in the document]
taysea: Done.

src/js/components/Button/Button.js
```diff
@@ -92,12 +96,80 @@ const getIconColor = (paths = [], theme, colorProp, kind) => {
...
 };
 
+const getBadgeDimension = (dimension, badgeProp, badgeRef, theme) => {
+  if (
+    typeof badgeProp === 'number' ||
+    (typeof badgeProp === 'object' && badgeProp.value)
+  ) {
+    const borderWidth = badgeProp.border
+      ? parseInt(
+          theme.global.borderSize[badgeProp.border.size].replace('px', ''),
```
ericsoderberghp, line 106: Can we use `parseMetricToNum()` instead of `replace()` and `parseInt`?

src/js/components/Button/Button.js
```diff
@@ -92,12 +96,80 @@ const getIconColor = (paths = [], theme, colorProp, kind) => {
...
+          10,
+        ) * 2
```
ericsoderberghp, line 108: Why the `* 2` here?

src/js/components/Button/Button.js
```diff
@@ -92,12 +96,80 @@ const getIconColor = (paths = [], theme, colorProp, kind) => {
...
+        borderWidth,
+      parseInt(theme.button.badge.size.medium.replace('px', ''), 10) +
+        borderWidth,
+    )}px`;
+  }
+  return `${badgeRef.current.getBoundingClientRect()[dimension]}px`;
+};
+
+const getBadge = (badgeProp, badgeRef, badgeHeight, badgeWidth, theme) => {
+  const max = badgeProp.max || theme.button.badge.max;
```
ericsoderberghp, line 129: I'm not sure why `max` would be in the theme. Any reason to not just have a default value for the property?

src/js/components/Button/Button.js
```diff
@@ -131,6 +203,7 @@ const Button = forwardRef(
     const theme = useContext(ThemeContext) || defaultProps.theme;
     const [focus, setFocus] = useState();
     const [hover, setHover] = useState(false);
+    const badgeRef = useRef();
```
ericsoderberghp, line 206: Maybe call this "badgeContentRef", since it doesn't refer to the entire badge?

src/js/components/Button/Button.js
```diff
@@ -92,12 +96,80 @@ const getIconColor = (paths = [], theme, colorProp, kind) => {
...
+        height={badgeHeight}
+        justify="center"
+        round
+        width={badgeWidth}
+      >
+        {badge}
+      </Box>
+    );
+    // caller has provided their own JSX and we will just render that
+  } else badge = <Box ref={badgeRef}>{badgeProp}</Box>;
```
ericsoderberghp, line 162: Something seems inconsistent that `badgeRef` here refers to the whole badge but when teh value is a number, it's just the text content.

src/js/components/Button/Button.js
```diff
@@ -258,6 +331,70 @@ const Button = forwardRef(
...
+    const defaultBadgeDimension =
+      typeof badgeProp === 'boolean' ||
+      (badgeProp && badgeProp.value && typeof badgeProp.value === 'boolean')
+        ? // empty badge should be smaller. this value was chosen as a default
+          // after experimenting with various values
+          `${parseInt(theme.button.badge.size.medium.replace('px', ''), 10) /
+            2}px`
+        : theme.button.badge.size.medium;
+
+    const [badgeWidth, setBadgeWidth] = useState(defaultBadgeDimension);
```
ericsoderberghp, line 343: Would it work to manage this as `badgeDimension`, which would be `[width, height]`? Having them separate results in a bit more duplication. Just a thought.

src/js/components/Button/Button.js
```diff
@@ -92,12 +96,80 @@ const getIconColor = (paths = [], theme, colorProp, kind) => {
   return result[1] || undefined;
 };
 
+const getBadgeDimension = (dimension, badgeProp, badgeRef, theme) => {
```
ericsoderberghp, line 99: There might be enough code here to warrant a separate Button/Badge.js file.

src/js/components/Button/Button.js
```diff
@@ -343,9 +481,36 @@ const Button = forwardRef(
...
       }
-      return <Tip {...tip}>{styledButtonResult}</Tip>;
+      styledButtonResult = <Tip {...tip}>{styledButtonResult}</Tip>;
+    }
+
+    // if the caller doesn't specify that they want
+    // the badge relative to the button content, set the badge
+    // relative to the outer edge of the button
+    if (badgeProp && badgeProp.target !== 'contents') {
+      const badge = getBadge(
```
ericsoderberghp, line 493: There's a fair amount of code duplication here. Wondering if there's a way to restructure things to avoid that?

src/js/components/Button/StyledButtonKind.js
```diff
@@ -78,8 +78,17 @@ const basicStyle = props => css`
   ${padStyle(props)}
   ${fontStyle(props)}
 
-  > svg {
+  // when button has badge, the SVG won't necessarily
+  // be the direct descendant
+  ${
+    props.badge
+      ? `
+  svg {
```
ericsoderberghp, line 86: What if the Button has an `<svg />` child? Would it be possible to put this inside a StyledBadge?

src/js/components/Button/Button.js
```diff
@@ -258,6 +331,70 @@ const Button = forwardRef(
...
+        ? 3.5
+        : 2;
+    const verticalOffset = `-${parseInt(badgeHeight.replace('px', ''), 10) /
+      divisor}px`;
+    const horizontalOffset = `-${parseInt(badgeWidth.replace('px', ''), 10) /
+      divisor}px`;
+
+    // set the badge relative to the button content
+    // as opposed to outer edge of button
+    if (badgeProp && badgeProp.target === 'contents') {
```
ericsoderberghp, line 374: Would it be possible to do this more automatically by checking if the Button has a border or background to decide where to align to?

src/js/components/Button/doc.js
```diff
@@ -251,6 +271,37 @@ export const themeDoc = {
...
+    type: 'string | object',
+    defaultValue: undefined,
+  },
+  'button.badge.max': {
+    description: `The max number to display in the badge. If max is 9 and badge 
+    has value of 10, 9+ will be displayed.`,
+    type: 'number',
+    defaultValue: undefined,
+  },
+  'button.badge.size.medium': {
```
ericsoderberghp, line 299: I'm wondering if we should prepare for eventually having `size` affect the Text size? Since we're going to flex the size a bit based on the content anyway, I'm wondering if we should structure the theme to have `button.badge.text.size` and `button.badge.container.pad`, and move `background` and `border` into `button.badge.container`.

ericsoderberghp (commented):
good progress, a few comments

src/js/components/Button/Button.js
taysea: Border width applies on both left and right

src/js/components/Button/Button.js
taysea: We could set a default value for the property. I guess my mindset was by default there would be no max, but a caller could specify in their theme that across their app they wanted a max. In Design System case, we would want max to be 9, so that anything greater would be 9+. If you're comfortable with grommet having a default max value, we could set max to 9 by default here.

src/js/components/Button/Button.js
ericsoderberghp: wondering if either a comment or renaming "borderWidth" might help. "widthUsedByBorders" is a bit long, but gets at it

src/js/components/Button/Button.js
taysea: Good idea. let me clean that up.

src/js/components/Button/Button.js
taysea: Yes, let me think about how to minimize duplication.

src/js/components/Button/Button.js
ericsoderberghp: this feels like the right 80% default case

src/js/components/Button/Button.js
taysea: Initially, I intended to clone the badgeProp and pass the ref directly onto the badgeProp (which would be more aligned with the concept of "content", but I was running into issue where the badgeProp would need to be built using `forwardRef` to properly receive the ref and didn't think it would be reasonable to except the caller to implement with that. For that reason, I wrapped the badgeProp in a box.

src/js/components/Button/Button.js
taysea: I think I like the idea of that, will try it out locally and see if it feels more readable.

src/js/components/Button/StyledButtonKind.js
taysea: Good thought, let me play around with that.

taysea: Following on here, I think this adjustment should be fine and it makes sense for it to live in `basicStyle`. The intent is that the selector applies to an svg inside the Button. When user does the following, the svg is a direct descendant of the Button, so the original `> svg` selector works:
```
<Button icon={<Add />} />
```

However, when the user adds badge, Stack is introduced into the Button contents, meaning that the svg is no longer the direct descendant of Button, so we need to adjust the selector:
```
<Button icon={<Add />} badge={2} />
```
This is affecting `basicStyle` meaning it will apply when the user is relying on Button props like `icon` and `label`. However, it will take no effect on something where caller has chosen to provide children to Button (because providing children to Button results in the button using `plainStyle`. See below for rendering of the following code:
```
<Button badge={2}>
   <Box background="skyblue" height="xsmall" width="xsmall">
      <MailOption />
  </Box>
</Button>
```
Screen Shot 2021-05-05 at 11 50 30 AM

src/js/components/Button/Button.js
taysea: Applied this change in recent commit.

src/js/components/Button/doc.js
taysea: I've included this in my latest changes. let me know if it aligns with what your were envisioning.

ericsoderberghp: Sounds good for text size, should this `button.badge.size.medium` go in `button.badge.container.width: { min : '24px' }`?

src/js/components/Button/Badge.js
```diff
@@ -0,0 +1,161 @@
...
+        pad =
+          parseMetricToNum(
+            theme.global.edgeSize[pad.horizontal] || pad.horizontal,
+          ) * 2;
+      } else pad = 0;
+    }
+
+    // if content is tall/wide, let badge grow to fit. otherwise,
+    // make sure it's at least badge.size.medium dimensions
+    return `${Math.max(
```
ericsoderberghp, line 49: This function feels like we're doing stuff the browser is already doing. Would it be possible to just set `min-width` and then adjust the offset via `useLayoutEffect()` where we can measure how big it actually is?

src/js/components/Button/Badge.js
```diff
@@ -0,0 +1,161 @@
...
+  // offset the badge so it overlaps content. when badge has content,
+  // offset should be 50%. when badge is empty, offset by a smaller amount to
+  // keep the badge closer to the content. this value was chosen as a
+  // reasonable default after testing with various grommet icons.
+  const divisor =
+    typeof content === 'boolean' || (content && content.value === true)
+      ? 3.5
+      : 2;
+  const offset = {
+    horizontal: `-${parseMetricToNum(width) / divisor}px`,
```
ericsoderberghp, line 101: You can shoot me if you like for this but ...

What about not adding `offset` to Stack and just doing the offsetting directly in the DOM via `element.style.`, calculating the offset distance via `useLayoutEffect()` within Badge.js?

src/js/components/Button/doc.js
taysea: It affects height as well. I was mirroring the approach of Avatar

src/js/components/Button/Badge.js
```diff
@@ -0,0 +1,151 @@
...
+            );
+          }
+        } else {
+          // caller has provided custom JSX
+          setWidth(`${contentRef.current.getBoundingClientRect().width}px`);
+          setHeight(`${contentRef.current.getBoundingClientRect().height}px`);
+        }
+      }
+    };
+    window.addEventListener('resize', onResize);
```
ericsoderberghp, line 70: Why are we tying this to a window resize event? I think we should do this on every layout but only set the state variables if they change.

ericsoderberghp (commented):
Looking good. One comment.

src/js/components/Button/Badge.js
taysea: When we resize and hit a responsive breakpoint, the width of the badge can change (because pad is responsive etc.) and so we want to recalculate the width since the offset value is reliant on the dimensions. Otherwise, I was seeing a case where the padding shrunk on small breakpoint but the width wasn't recalculated so the offset of the badge was incorrect.

ericsoderberghp: Ah, thanks. Sounds like a good candidate for a comment. :)

ericsoderberghp (approved):

src/js/components/Button/doc.js
```diff
@@ -44,6 +45,23 @@ export const doc = Button => {
...
+        background: backgroundPropType,
+        max: PropTypes.number,
+        value: PropTypes.oneOfType([PropTypes.bool, PropTypes.number]),
+      }),
+    ])
+      .description(
+        `An indicator to show on the top-right of the Button. For accessibility,
+         supplement the badge by adding an a11yTitle to the Button which 
+         provides a useful message to screen readers. For example, "2 unread 
+         notifications".`,
```
ShimiSun, line 62: From this description, it is not clear to me what is the behavior for the different types. 
i.e. what is the key/value meaning of the badge object, such as `max`.

PR #5216, opened 2021-05-04
Improved TextInput custom suggestions example

### LINKED ISSUE #5206: selecting a suggestion in TextInput doesn't close the suggestions drop
### Expected Behavior
After selecting a suggestion, I expect no more suggestions will be presented.

### Actual Behavior
The selected suggestion is still presented in the suggestions drop, after selecting it for the 1st time.
Only after selecting that suggestion for the 2nd time, the drop closes.

This can be experienced in the official Grommet Storybook https://storybook.grommet.io/?path=/story/input-textinput-custom-suggestions--custom-suggestions

### Steps to Reproduce

1. enter 'a' in the text input 
2. select 1st suggestion (it should still be displayed)
3. select it again

### Your Environment

- Grommet version: 2.17.1
- Browser Name and version: Chrome 90
- Operating System and version (desktop or mobile): desktop MacOS Catalina

MikeKingdom:
#### What does this PR do?
Updates the TextInput Custom Suggestions Storybook example to only update the `suggestions` property when new suggestions should be displayed.

With PR #4920 the TextInput will pop up the drop containing the suggestions if the input has focus and the `suggestions` property has changed. This was a problem for this example because it was updating the suggestions when the user chooses a suggestion from the drop which causes that drop to stay open.

CustomSuggestions.tsx

The Custom Suggestions example in storybook

The Custom Suggestions example in storybook

Fixes #5206 

No.

It might be worth mentioning changing the TextInput `suggestions` property will open the drop if the input has focus.

Backwards compatible


src/js/components/TextInput/stories/typescript/CustomSuggestions.tsx
```diff
@@ -97,38 +120,15 @@ export const CustomSuggestions = () => {
     [],
   );
 
-  const onSuggestionsOpen = useCallback(() => setSuggestionOpen(true), []);
-  const onSuggestionsClose = useCallback(() => setSuggestionOpen(false), []);
+  const onSuggestionsOpen = useCallback(() => {
```
ericsoderberghp, line 123: Since TextInput doesn't attempt to open the drop unless there are suggestions, why do we need to `setSuggestions` here?

MikeKingdom: The suggestions are out of date when the onChange sets value and the drop is often not open. This event gets triggered right after the onChange and since we know the drop is open, we can update the suggestions here without causing it to open. It's basically the key to the fix.

Issue #5201, opened 2021-04-30
If Value in select is set to a string this is not displayed


britt6612:
If "value" is a string, it is no longer being displayed, though it is being set correctly.
I did some digging on this and found that the change was linked with this PR:
https://github.com/grommet/grommet/pull/3870
### Expected Behavior
The value should be displayed 

### Actual Behavior
The Value is no longer being displayed after 2.11.3

There is nothing being displayed on the `select` 
### URL, screen shot, or Codepen exhibiting the issue

Here is a  codesandbox that reproduces the problem..
If you switch versions back to before 2.11.3 then this is working
https://codesandbox.io/s/bold-elgamal-hqw19?file=/src/App.js

### Steps to Reproduce

look at codesandbox

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

britt6612:
closed via https://github.com/grommet/grommet/pull/5217

PR #5217, opened 2021-05-04
Update Select docs to clarify value, labelKey

taysea:
#### What does this PR do?
Updating Select docs to clarify intended usage of value and labelKey. 

src/js/components/Select/doc.js

Updated here.

No.

Backwards compatible.


britt6612:
This would close issue #5201

britt6612 (approved):

src/js/components/Select/doc.js
```diff
@@ -121,7 +121,8 @@ export const doc = Select => {
       PropTypes.func,
     ]).description(
       `When the options array contains objects, this property indicates how
-      to determine the label of each option. If a string is
+      to determine the label of each option. The label is what is displayed 
+      in the Select button and its dropdown options. If a string is
```
ericsoderberghp, line 125:
```suggestion
      visually for the user, both to indicate the selected value and in the options list. If a string is
```

src/js/components/Select/doc.js
```diff
@@ -214,10 +215,12 @@ export const doc = Select => {
           PropTypes.number,
         ]),
       ),
-    ]).description(`Currently selected value. This can be an array
-      when multiple. Passing an element allows the caller to control how
-      the value is rendered. Passing an element is deprecated. Instead,
-      use the 'valueLabel' property.`),
+    ]).description(`Currently selected value. This should be an option from
```
ericsoderberghp, line 218: It actually should be one of the `options` objects if `valueKey` has `reduce: true`. Perhaps here we can just refer to needing to match whatever the `valueKey` results in for the selected option. We might need to improve the description of `valueKey` to be clearer. I think it's so-so at the moment.

taysea:
This is very helpful, I feel like this direction clarifies a lot. Only point of clarification I still am wondering about is, when the options are objects, should what is stored in Select's value never include the label portion of the option object? or is that up to the caller to decide. Wondering because in this example, it still works for the value to contain the label portion of the option: https://codesandbox.io/s/async-haze-9mvsx?file=/src/Select.js

ericsoderberghp:
> This is very helpful, I feel like this direction clarifies a lot. Only point of clarification I still am wondering about is, when the options are objects, should what is stored in Select's value never include the label portion of the option object? or is that up to the caller to decide. Wondering because in this example, it still works for the value to contain the label portion of the option: https://codesandbox.io/s/async-haze-9mvsx?file=/src/Select.js

What is being stored in "selectedName" is really the "selectedOption". It is initialized with `OPTIONS[0]` and then updated with the `event.option` in `onSelect`. Since valueKey doesn't set `reduce`, then `value` should be passed one of the `options`. This is what "Otherwise, the value(s) should match the selected options." is intending to communicate.

taysea:
> > This is very helpful, I feel like this direction clarifies a lot. Only point of clarification I still am wondering about is, when the options are objects, should what is stored in Select's value never include the label portion of the option object? or is that up to the caller to decide. Wondering because in this example, it still works for the value to contain the label portion of the option: https://codesandbox.io/s/async-haze-9mvsx?file=/src/Select.js
> 
> What is being stored in "selectedName" is really the "selectedOption". It is initialized with `OPTIONS[0]` and then updated with the `event.option` in `onSelect`. Since valueKey doesn't set `reduce`, then `value` should be passed one of the `options`. This is what "Otherwise, the value(s) should match the selected options." is intending to communicate.

Got it, I see the connection now. Maybe then something along the lines of: "Otherwise, the value should be one (or multiple, when multiple = true) of the options from the options array."

britt6612:
@ericsoderberghp I think this is very helpful for the definitions just wondering if we can also add a few comments to the code.. I would not of known this behavior without reading this.. There are a few comments already but we can elaborate a little more to be clear.

ericsoderberghp:
> @ericsoderberghp I think this is very helpful for the definitions just wondering if we can also add a few comments to the code.. I would not of known this behavior without reading this.. There are a few comments already but we can elaborate a little more to be clear.

What about just pasting these in the code as well, so they are available there too?

britt6612:
> > @ericsoderberghp I think this is very helpful for the definitions just wondering if we can also add a few comments to the code.. I would not of known this behavior without reading this.. There are a few comments already but we can elaborate a little more to be clear.
> 
> What about just pasting these in the code as well, so they are available there too?

sounds good I will update this PR with those changed!

ericsoderberghp (approved):

Issue #5144, opened 2021-04-05
FileInput -- Add prop to limit number of files


meekrhino:
The `FileInput` component cannot be operated in a controlled mode, so currently there is no way to limit the number of flies allowed in the list in `multiple` mode.  The easiest solution to this would be to add a new prop for `maximumCount` or similar.  This is distinct from `aggregateThreshhold` which only limits the files displayed, not the number in the list.

### Expected Behavior

There is no way to stop a user from adding any number of files.

### Actual Behavior

A prop to limit the number of files allowed.

### URL, screen shot, or Codepen exhibiting the issue

[The Grommet example Storybook](https://storybook.grommet.io/?path=/story/input-fileinput-custom--custom)

### Steps to Reproduce

1.  Add many files to the list

### Your Environment

- Grommet version: 2.17.1
- Browser Name and version: (any)
- Operating System and version (desktop or mobile): Windows 10 Home -- 19041.867

g4rry420:
Can I work on this issue ?

britt6612:
@g4rry420 of course! 👍 Feel free to file a PR!

PR #5159, opened 2021-04-09
Limit files selected by user in FileInput

g4rry420:
#### What does this PR do?
Added maximumCount prop to limit the number of files selected by the user. I

#### Where should the reviewer start?
components => FileInput.js

#### What testing has been done on this PR?
yarn test-update

#### How should this be manually tested?
Updated a storybook

#### Any background context you want to provide?
N/A

#### What are the relevant issues?
#5144 

#### Screenshots (if appropriate)
N/A

#### Do the grommet docs need to be updated?
Yes, I already did.

#### Should this PR be mentioned in the release notes?
N/A

#### Is this change backwards compatible or is it a breaking change?
Backwards compatible

PR #5218, opened 2021-05-05
Limit files

g4rry420:
#### What does this PR do?
Added `max` option to `multiple` to limit the number of files selected by the user.

components => FileInput.js

N/A

Updated a storybook

#5144 
N/A

Yes

N/A

Backwards compatible


g4rry420:
As discussed in #5159 , I added the `max` option in `multiple` prop. But still in console, files that are over the limit are still being shown as being added.This is because we passing `onChange` with `event` and we haven't removed files from `event.target.files`. Needed your thoughts, how can we achieve this. I was also reading [this](https://stackoverflow.com/questions/1696877/how-to-set-a-value-to-a-file-input-in-html) on stackoverflow about setting value to fileinput.

g4rry420:
> 
> 
> As discussed in #5159 , I added the `max` option in `multiple` prop. But still in console, files that are over the limit are still being shown as being added.This is because we passing `onChange` with `event` and we haven't removed files from `event.target.files`. Needed your thoughts, how can we achieve this. I was also reading [this](https://stackoverflow.com/questions/1696877/how-to-set-a-value-to-a-file-input-in-html) on stackoverflow about setting value to fileinput.

I came up with this solution in line 263 to 267. Not sure, if this is the best one.

src/js/components/FileInput/doc.js
```diff
@@ -51,6 +51,9 @@ export const doc = FileInput => {
         number of files will be shown.`,
       )
       .defaultValue(undefined),
+    maximumCount: PropTypes.number
```
taysea, line 54: It doesn't seem like this is being used. We should remove and update the multiple docs to show that max is an option

src/js/components/FileInput/FileInput.js
```diff
@@ -81,6 +81,7 @@ const FileInput = forwardRef(
     const [hover, setHover] = React.useState();
     const [dragOver, setDragOver] = React.useState();
     const aggregateThreshold = (multiple && multiple.aggregateThreshold) || 10;
+    const maxCount = multiple && multiple.max;
```
taysea, line 84: Maybe it would be more aligned with prop name to just call this variable `max`

g4rry420: Done

src/js/components/FileInput/doc.js
g4rry420: Done

src/js/components/FileInput/doc.js
```diff
@@ -42,18 +42,17 @@ export const doc = FileInput => {
...
         aggregateThreshold: PropTypes.number,
+        max: PropTypes.number,
       }),
     ])
       .description(
         `Whether to allow multiple files. If an object is supplied,
         'aggregageThreshold' indicates the maximum number of individual
         files to show. Above this, only a single message describing the
-        number of files will be shown.`,
+        number of files will be shown.And if the object 'max' is supplied.
```
taysea, line 52: wording suggestion to align with aggregateThreshold:

If an object is supplied, 'max' indicates the maximum number of files that can be added. If the user has reached the max, no additional files will be able to be added.

g4rry420: Done

src/js/components/FileInput/doc.js
```diff
@@ -42,13 +42,17 @@ export const doc = FileInput => {
...
         aggregateThreshold: PropTypes.number,
+        max: PropTypes.number,
       }),
     ])
       .description(
         `Whether to allow multiple files. If an object is supplied,
         'aggregageThreshold' indicates the maximum number of individual
         files to show. Above this, only a single message describing the
-        number of files will be shown.`,
+        number of files will be shown.And if an object is supplied,
```
jcfilben, line 52:
```suggestion
        number of files will be shown. If an object is supplied, 
```

jcfilben (commented):
Looks good, just fix lint error and see comment below about documentation

g4rry420:
@jcfilben Thank you fixing the eslint issue. But there is one more problem, Currently, if the maximum are exceed only then event  will become empty. But suppose, the maxmum limit is `5` for file input and user selects `6` files in the one go, then the console will still show the 6th file.
Current solution for event doesn't handle all use cases.

jcfilben:
@g4rry420, good point. Currently the `Multiple.js` story is logging in the console each entry in `event.target.files` which is a FileList. FileList types are read-only so we are unable to update the list once the user has entered files into the FileInput. Instead we are creating a different list, `files` which contains the actually files that are being displayed and entered into the FileInput. Here are some different possibilities:
1. Prevent the event from happening if a user attempts to select all at once more files than the max and show an error message. 
2. Change the storybook story to have the FileList be `files` instead of `event.target.files`.

ericsoderberghp:
> @g4rry420, good point. Currently the `Multiple.js` story is logging in the console each entry in `event.target.files` which is a FileList. FileList types are read-only so we are unable to update the list once the user has entered files into the FileInput. Instead we are creating a different list, `files` which contains the actually files that are being displayed and entered into the FileInput. Here are some different possibilities:
> 
> 1. Prevent the event from happening if a user attempts to select all at once more files than the `max` and show an error message.
> 2. Change the storybook story to have the FileList be `files` instead of `event.target.files`.

I think the storybook story should match the value we would be giving to a containing Form, and that should be truncated to be at most `max` entries.

jcfilben (approved):

src/js/components/FileInput/FileInput.js
```diff
@@ -396,16 +397,46 @@ const FileInput = forwardRef(
...
+              fileList.length >= max - files.length
+            ) {
+              const newFileList = [];
+              for (let i = 0; i < max - files.length; i += 1) {
+                newFileList.push(fileList[i]);
+              }
+              // if the maximum value is exceeded then
+              // remove files from the event also.
+              if (!newFileList.length) {
+                /* eslint-disable-next-line no-param-reassign */
```
taysea, line 423: @ericsoderberghp are we good with this eslint-disable?

ericsoderberghp: I don't think we should touch the `event.target.value`. The way this component works, we're guiding callers to use the `files` sent to `onChange` instead of `event.target.value` because the `event.target.value` for a file input is read-only.

src/js/components/FileInput/README.md
```diff
@@ -55,12 +55,16 @@ Custom messages for FileInput. Used for accessibility by screen
 Whether to allow multiple files. If an object is supplied,
         'aggregageThreshold' indicates the maximum number of individual
         files to show. Above this, only a single message describing the
-        number of files will be shown.
+        number of files will be shown. If an object is supplied, 
+        'max' indicates the maximum number of files that can be added.
+        If the user has reached the max, no additional files will 
+        be able to be added.
```
ericsoderberghp, line 61: I'm wondering if we can prevent adding new files by disabling the "browse" button when the max is reached? I'm also wondering if we should then add a message indicating that the maximum number of files has been reached?

vavalos5:
Something I really dislike about some FileInputs is when there is not type of description of conceptual help text that tells me prior to attaching files that I can only attach "5" files, only to receive an error message after. I. think for this case it would be good to have contextual help below the FileInput like "Attach a maximum of 5 files" as we would see "File maximum size is 32MB" for example. Now if the user surpasses that number, then an error would pop up below indicating the problem: "Exceeds maximum of 5 files. Please select a maximum of 5" ; the 6th item would not be attached, but it would let the user know again that only 5 could be selected and they can revise the ones that have already been attached.

ericsoderberghp:
Good comments @vavalos5 ! So, we at least need to ensure that the caller has the context to be able to set the validations right or we should have the FileInput handle this automatically, which I think would be preferred. @jlaluces123 is working on something in this area for handling a maximum file size.

vavalos5:
I think we should have both. The context would be visual to the user and the FileInput API would then do it's thing

g4rry420:
Its a good idea to add info messages for the user. I see that we have `messages` prop for the screen readers users. Do you guys think we should let user to pass additional prop to this `messages` object for info and error messages too ? Like, I can add `infoMessage` and `errorMessage` in this prop for custom messages. Otherwise, we will put our own message for max file length. Let me know your thoughts.

g4rry420:
Some of the changes that I made after few commits:
- when there are maximum files are reached `browse` button will be disabled.
- if user has passed `multiple.max` prop then we have our own info message of `Attach a maximum of 5 files only.` However, user can override this message by passing `messages.maxFilesMessage` in props.
- user can also the styles of this custom by passing `props.theme.fileInput.message.maxFileMessage`.

However, docs are not updated these changes. First, need your thoughts on these.

src/js/components/FileInput/stories/Multiple.js
```diff
@@ -8,7 +8,9 @@ export const Multiple = () => (
     <Box fill align="center" justify="start" pad="large">
       <Box width="medium">
         <FileInput
-          multiple
+          multiple={{
```
taysea, line 11: Getting an error in the "Multiple" story:
Screen Shot 2021-07-28 at 9 17 03 AM

src/js/components/FileInput/stories/Custom.js
```diff
@@ -18,6 +18,12 @@ const customTheme = {
         horizontal: '8px',
       },
     },
+    message: {
+      maxFileMessage: {
```
taysea, line 22: Can we just call this `maxFile` since Message is slightly redundant?

g4rry420: Sure

src/js/components/FileInput/stories/Multiple.js
g4rry420: I resolve this one.

src/js/components/FileInput/FileInput.js
```diff
@@ -49,7 +49,8 @@ const Message = styled(Text)`
   ${(props) =>
     props.theme.fileInput &&
     props.theme.fileInput.message &&
-    props.theme.fileInput.message.extend};
+    props.theme.fileInput.message.extend &&
+    props.theme.fileInput.message.maxFileMessage};
```
taysea, line 53:
```suggestion
    props.theme.fileInput.message.maxFile};
```

src/js/components/FileInput/FileInput.js
```diff
@@ -80,6 +81,7 @@ const FileInput = forwardRef(
     const [hover, setHover] = React.useState();
     const [dragOver, setDragOver] = React.useState();
     const aggregateThreshold = (multiple && multiple.aggregateThreshold) || 10;
+    const max = multiple && multiple.max;
```
MikeKingdom, line 84: this is a bit awkward since you use it in numeric expressions and it could be boolean and such this way. How about something like `const max = multiple?.max || Number.POSITIVE_INFINITY`. You might be able to still default to false but that makes for some confusing logic where it's used.

src/js/components/FileInput/FileInput.js
```diff
@@ -158,32 +160,104 @@ const FileInput = forwardRef(
...
+                        margin="small"
+                        onClick={() => {
+                          inputRef.current.click();
+                          inputRef.current.focus();
+                        }}
+                        label={format({
+                          id: 'fileInput.browse',
+                          messages,
+                        })}
+                        disabled={max === files.length}
```
MikeKingdom, line 228: same

src/js/components/FileInput/FileInput.js
```diff
@@ -158,32 +160,104 @@ const FileInput = forwardRef(
...
+                        kind={theme.fileInput.button}
+                        label={format({
+                          id: 'fileInput.browse',
+                          messages,
+                        })}
+                        onClick={() => {
+                          inputRef.current.click();
+                          inputRef.current.focus();
+                        }}
+                        disabled={max === files.length}
```
MikeKingdom, line 212: I think this is kinda accidentally working when files.length = 0 and max isn't defined or is basically coming out `false`.

src/js/components/FileInput/FileInput.js
```diff
@@ -206,6 +280,7 @@ const FileInput = forwardRef(
                         inputRef.current.click();
                         inputRef.current.focus();
                       }}
+                      disabled={max === files.length}
```
MikeKingdom, line 283: same

src/js/components/FileInput/FileInput.js
```diff
@@ -221,198 +296,165 @@ const FileInput = forwardRef(
                         id: 'fileInput.browse',
                         messages,
                       })}
+                      disabled={max === files.length}
```
MikeKingdom, line 299: same

src/js/components/FileInput/FileInput.js
```diff
@@ -221,198 +296,165 @@ const FileInput = forwardRef(
...
+                if (!existing) nextFiles.push(filesToAdd[i]);
+              }
+              setFiles(nextFiles);
+              setDragOver(false);
+              if (onChange) onChange(event, { files: nextFiles });
+            }}
+          />
+        </ContentsBox>
+        {max && (
+          <Message {...theme.fileInput.message.maxFile}>
```
MikeKingdom, line 451: Why the special theme for maxFile? Could it be made more generic for all messages or maybe error messages at least?

src/js/components/FileInput/FileInput.js
```diff
@@ -221,198 +296,165 @@ const FileInput = forwardRef(
...
+            {messages?.maxFile
```
MikeKingdom, line 452: this needs to be done with `format` like line 382 and others. This message also needs to be added to default.json.

MikeKingdom:
Has some size issues:
image

src/js/components/FileInput/FileInput.js
```diff
@@ -49,7 +49,8 @@ const Message = styled(Text)`
   ${(props) =>
     props.theme.fileInput &&
     props.theme.fileInput.message &&
-    props.theme.fileInput.message.extend};
+    props.theme.fileInput.message.extend &&
+    props.theme.fileInput.message.maxFile};
```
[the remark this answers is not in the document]
g4rry420: I will do that. Thank you.

src/js/components/FileInput/FileInput.js
g4rry420: Can you please suggest a better solution ?

src/js/components/FileInput/FileInput.js
g4rry420: Can you suggest how can I add a default maxFile in `default.json`, since I also required `max` prop value ?

src/js/components/FileInput/FileInput.js
g4rry420: Shall I also add one more node for the `error` messages also ?

src/js/components/FileInput/FileInput.js
g4rry420: Adding `Number.POSITIVE_INFINITY` returns true everytime for conditons. This is for `Simple` story.
grommet

src/js/components/FileInput/FileInput.js
```diff
@@ -221,198 +295,165 @@ const FileInput = forwardRef(
...
+              for (let i = 0; i < filesToAdd.length; i += 1) {
+                // avoid duplicates
+                const existing =
+                  nextFiles.filter(
+                    (file) =>
+                      file.name === fileList[i].name &&
+                      file.size === fileList[i].size,
+                  ).length > 0;
+                if (!existing) nextFiles.push(filesToAdd[i]);
+              }
```
MikeKingdom, line 442: This code seems to be trying to avoid making nextFiles larger than max. I think it has an interesting corner-case of when one of the new files is a duplicate of one of the existing files and a max is defined. How about instead just slicing the result to keep it under max. For example:
```jsx
let nextFiles = multiple ? [...files] : [];
for (let i = 0; i < fileList.length; i += 1) {
  // avoid duplicates
  const existing =
  nextFiles.filter(
    (file) =>
      file.name === fileList[i].name &&
      file.size === fileList[i].size,
  ).length > 0;
  if (!existing) nextFiles.push(fileList[i]);
}
if (nextFiles.length > max) {
  nextFiles = nextFiles.slice(0, max);
}
```
This also assumes max was defaulted to `Number.POSITIVE_INFINITY`.

src/js/components/FileInput/FileInput.js
MikeKingdom: I'm talking about our default messages file for localization. I think you're thinking of the default properties values.
In [src/js/languages/default.json](https://github.com/grommet/grommet/blob/master/src/js/languages/default.json) between lines 10 and 11 add
```
"maxFile": "Attach a maximum of {max} files only.",
```
Note there is no $ in this one.

Then here you can do:
```jsx
<Message {...theme.fileInput.message}>
  {format({
    id: 'fileInput.maxFile',
    messages,
    values: { max },
   })}
</Message>
```

This will get the text from default.json (or let it be overridden by a `maxFiles` in the `messages` property if present) and replace the `{max}` token in the text with the value passed in for `max`.

src/js/components/FileInput/FileInput.js
g4rry420: I will refactor this one. Thank you.

src/js/components/FileInput/FileInput.js
```diff
@@ -80,6 +80,7 @@ const FileInput = forwardRef(
     const [hover, setHover] = React.useState();
     const [dragOver, setDragOver] = React.useState();
     const aggregateThreshold = (multiple && multiple.aggregateThreshold) || 10;
+    const max = multiple?.max;
```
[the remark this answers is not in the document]
g4rry420: Will do that. Thank you

g4rry420:
@MikeKingdom  Thank you for your thorough review. If you didn't see any other issues then we can  move forward to add docs.

MikeKingdom (commented):
Looks good and seems to work well. Thanks!

src/js/components/FileInput/FileInput.js
```diff
@@ -221,198 +295,146 @@ const FileInput = forwardRef(
...
+              if (nextFiles.length > max) {
+                nextFiles = nextFiles.slice(0, max);
+              }
+              setFiles(nextFiles);
+              setDragOver(false);
+              if (onChange) onChange(event, { files: nextFiles });
+            }}
+          />
+        </ContentsBox>
+        {multiple?.max && (
```
ericsoderberghp, line 428: This should be integrated with FormField `validate` and `help` rather than adding the message directly here.

Take a look at how https://github.com/grommet/grommet/pull/5457 is handling a `maxSize` for FileInput.

g4rry420: I have made the changes but I am not sure about the following things:-
-  how are we gonna display the `maxFile` message when the `FileInput` is not wrapped inside the `FormField` i.e in other stories.
- User will still saw the validate error message even when we remove `required` prop from the FormField. I know this is because of this `if (fileList?.length && fileList.length <= max) return;`. But is it possible if I can know if the user has added `required` prop or not ?

If you see any other issue, please let me know about it.

MikeKingdom: I'm not sure of the answer to this yet. The `required` prop and validations from the component don't really play well together. Typically both should get executed in order and the first one that fails wins. But this doesn't seem to be how the form does it.

Also, given that the browse button even let you add more than the limit, I don't see the validator or message ever being useful. It will never show except this odd empty field case.

g4rry420: Then, what do you guys think is the good approach to tackle this ?

MikeKingdom: After talking to Eric, what we want to do is remove the code that truncates the list of files that comes from the Browse dialog. Just let the user add however many files they choose then let the validation show that is over the limit.

The the other change is to use remove the `useFormField` call and replace it with a new version of useFormInput that can provide the validation function and error text. This will have to wait for #5457 which adds this ability. This will make `required` still work correctly plus any other validation added to the `FormField`

src/js/components/FileInput/FileInput.js
```diff
@@ -80,11 +80,31 @@ const FileInput = forwardRef(
...
     const inputRef = useForwardedRef(ref);
     const controlRef = useRef();
     const removeRef = useRef();
     const RemoveIcon = theme.fileInput.icons.remove;
 
+    formContext.useFormField(
+      multiple?.max
+        ? {
+            name,
+            validate: (fileList) => {
```
MikeKingdom, line 93: PR #5457 basically added more capability to useFormInput and uses it in FileInput to provide a max file size validation. The `validate` option can be an array of validate functions (first failing function wins.) This PR needs to be adjusted to basically provide both the file count limit and maxSize validation functions in an array to `validate`.

Feel free to reach out if you want my help with that.

g4rry420:
I have made changes for the `multiple.max`  prop. One think I am still not sure about, Currently, the validate message will only appear for the users who wrap `FileInput` into the `FormField`, not for the other users who are using `FileInput` without `FormField`. Even, if we wanted to display the error message for the latter case, then I think, we have to somehow know the name of the parent component to display the message.
 Another thing, as per discussion [above](https://github.com/grommet/grommet/pull/5218#issuecomment-880057642) , I think if we show error message prior to attaching files to the user that be better UX.
 Please, let me know your thoughts.

src/js/components/FileInput/FileInput.js
MikeKingdom: > After talking to Eric, what we want to do is remove the code that truncates the list of files that comes from the Browse dialog. Just let the user add however many files they choose then let the validation show that is over the limit.
> 
> Then the other change is to use remove the `useFormField` call and replace it with a new version of useFormInput that can provide the validation function and error text. This will have to wait for #5457 which adds this ability. This will make `required` still work correctly plus any other validation added to the `FormField`

This code is still disabling browse based on `files.length >= max`. The decision was to let users pick a bunch of files and just rely on the validation to mark it invalid. If you disable the browse then the user wouldn't be able to correct it.

MikeKingdom (commented):
You're right the validation message won't show if the field isn't in a FormField, but that's what we settled on for maxSize also. If the developer wants to have a treatment for validations then we would expect them to either use FormField or implement something that similarly uses the FormContext to get any validations and display them as appropriate.

Note that it's probably worth tweaking the lines that add the `CircleAlert` in the input when maxSize is hit to instead add that icon when max files or maxSize is hit.

g4rry420:
@MikeKingdom  Correct me if I am wrong, the current implemented logic of FileInput in this pr doesn't add up the files more than the `max` prop. So, I believe I have change this too to later scenario when a user clicks `onSubmit`. We will show the error message to user and this error icon will be at the files which are more than the specified in the `max` prop. Is this the way we wanted to implemeent, right ? Please, let me know.

fileInput_limitFiles

MikeKingdom:
> @MikeKingdom Correct me if I am wrong, the current implemented logic of FileInput in this pr doesn't add up the files more than the `max` prop. So, I believe I have change this too to later scenario when a user clicks `onSubmit`. We will show the error message to user and this error icon will be at the files which are more than the specified in the `max` prop. Is this the way we wanted to implemeent, right ? Please, let me know.
> 
> fileInput_limitFiles

No. Part of what we're suggesting is remove the slice() code that is preventing the user from adding more than `max` files. Then instead let the validation mark it as invalid and supply the error message to FormField or whatever else is listening. Then the user can remove whichever files they want from the list. This is preferable to just chopping off the extra files from the end of the list.

g4rry420:
@MikeKingdom I have added the changes. Please, let me know if I made any mistakes.

MikeKingdom:
> @MikeKingdom I have added the changes. Please, let me know if I made any mistakes.

Thanks @g4rry420 . I'll take a look!

src/js/components/FileInput/FileInput.js
```diff
@@ -357,7 +377,8 @@ const FileInput = forwardRef(
                   align="center"
                   direction="row"
                 >
-                  {maxSize && file.size > maxSize && <CircleAlert />}
+                  {(maxSize || max) &&
+                    (file.size > maxSize || index >= max) && <CircleAlert />}
```
MikeKingdom, line 381: I suggest this for readability and to avoid a corner-case:
```jsx
{((maxSize && file.size > maxSize) ||
  (max && index >= max)) && <CircleAlert />}
```
The corner case is if `max` is passed in as an empty string and there's a `maxSize` then `index >= max` is actually true and you can get an erroneous CircleAlert icon on the items.

MikeKingdom (commented):
Looks great! Thanks @g4rry420 

There's just one small corner case I noticed. Also it would be helpful to add a Jest test case or two to check the `max` behavior

MikeKingdom (approved):
Thanks @g4rry420 !

g4rry420:
@MikeKingdom 
One thing, When we include `multiple.max` alongside of `maxSize`, and if both of `max` and `maxSize` prop validation failed in the `formfield` then we will only see the validate error message of `maxSize`  only. Is this intended ?
maxSizeAndmaxCount

MikeKingdom:
> @MikeKingdom
> One thing, When we include `multiple.max` alongside of `maxSize`, and if both of `max` and `maxSize` prop validation failed in the `formfield` then we will only see the validate error message of `maxSize` only. Is this intended ?

Yes. That is intended. In general for validations, in some situations if one check fails, other checks may just confuse the issue.

taysea:
Right now if I upload 5 files and press submit, then I get the validation error. I think validation error should only appear if I upload > max # of files.

Screen Shot 2021-09-16 at 1 49 57 PM

src/js/components/FileInput/FileInput.js
```diff
@@ -103,29 +104,48 @@ const FileInput = forwardRef(
...
-                values: { maxSize: formatBytes(maxSize), numOfInvalidFiles },
-              });
+              return message;
+            }
+          : '',
+        max
+          ? () => {
+              const fileList = [...files];
+              let message = '';
+              if (fileList.length >= max) {
```
taysea, line 138:
```suggestion
              if (fileList.length > max) {
```

taysea: If I upload 5 files and the max is 5, I would expect this to be accepted.

PR #5220, opened 2021-05-05
Add Columns component to handle responsive layout scenarios

ericsoderberghp:
#### What does this PR do?

Add Columns component to handle responsive layout scenarios.

This is an incomplete pull request intended to solicit feedback before adding comprehensive tests.

review the various stories under Columns at different resolutions to see the behavior, then take a look at index.d.ts to see the interface.

Just storybook so far

storybook

https://github.com/grommet/hpe-design-system/issues/1497

yes

yes

backwards compatible


src/js/components/ColumnCenter/doc.js
```diff
@@ -0,0 +1,27 @@
...
+    .availableAt(getAvailableAtBadge('ColumnCenter', 'Layout'))
+    .description('Centered column layout.')
+    .usage(
+      `import { ColumnCenter } from 'grommet';
+<ColumnCenter />`,
+    )
+    .intrinsicElement('div');
+
+  DocumentedColumnCenter.propTypes = {
+    gutter: PropTypes.oneOfType([
```
taysea, line 16: I don't have an issue with `gutter`, but wondering if `gap` be more aligned with Grommet props? Understand though if we'd like to be more specific in this context.

src/js/components/ColumnCenter/ColumnCenter.js
```diff
@@ -0,0 +1,50 @@
+import React, { Children, cloneElement, forwardRef, useContext } from 'react';
+import { ThemeContext } from 'styled-components';
+
+import { defaultProps } from '../../default-props';
+import { parseMetricToNum } from '../../utils';
+import { Grid } from '../Grid';
+import { Box } from '../Box';
+
+const ColumnCenter = forwardRef(
```
taysea, line 9: This component name doesn't really make sense to me/I'm not sure I fully understand why it needs to be separate from the general "Columns" component. Is there any way to have a single component where the caller flags what kind of behavior they want? or based on the props they provide, Columns knows what to do?

src/js/components/Columns/stories/MainAside.js
```diff
@@ -0,0 +1,63 @@
...
+      <Main gap="medium">
+        <Header border="bottom">
+          <Box direction="row" align="center">
+            <Heading size="small">Welcome</Heading>
+          </Box>
+          <Menu label="actions" />
+        </Header>
+        <Columns
+          columns={[
+            undefined,
```
taysea, line 33: In this scenario, what if I wanted the columns to wrap one over the other. Is that supported?

src/js/components/ColumnCenter/doc.js
ericsoderberghp: Good comment. I forgot to remove `ColumnCenter`, an older idea in this PR. I was looking more for `Columns`. I've removed ColumnCenter to reduce confusion. Sorry about that.

src/js/components/Columns/index.d.ts
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+import { MarginType } from '../../utils';
+
+type widthType = 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | string;
+export interface ColumnsProps {
+  columns?: {
+    hide?: boolean;
+    layer?: boolean;
+    responsive?: { hide?: boolean; layer?: boolean };
```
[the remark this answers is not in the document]
ericsoderberghp: Interesting. Do you have a particular example screen in mind?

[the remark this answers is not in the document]
ericsoderberghp: I had been contemplating that wrapping two columns to one might be handled by `<Box direction="row-responsive" />`. I'm trying to trade-off having Columns be useful for typical cases but not having it become a kitchen-sync page layout component. So, I've been considering how to use it in combination with other elements, like Header, Sidebar, etc. I am wanting to keep exploring though.

src/js/components/Columns/doc.js
```diff
@@ -0,0 +1,53 @@
...
+    ).description(`
+      How each child should be handled.
+      'hide'
+        true - indicates that the child should be hidden on initial render
+        false - indicates that the child shouldn't be hideable,
+      'layer' indicates that the child should be shown in a layer.
+      'width' indicates that the child should have a fixed width.
+      'responsive' indicates how the child should behave when the responsive
+      size is small.
+      Control visibility of children via <Columns.ControlButton child={0} />.
```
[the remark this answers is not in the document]
ericsoderberghp: Good idea. I had 'Toggle' in a prior version. I'll bring it back. I liked having "Button" in the name as it helped reinforce what it was doing.

src/js/components/Columns/stories/MainAside.js
taysea: I guess I envisioned that typically when we have 2 column layouts and we hit a small breakpoint, we wrap the second column under the first one

src/js/components/Columns/stories/MainAside.js
```diff
@@ -0,0 +1,63 @@
...
+            <Box alignSelf="end">
+              <Columns.ToggleButton child={1} icon={<Next />} />
+            </Box>
+          </Box>
+          <Box
+            as="aside"
+            pad={{ horizontal: 'medium' }}
+            background="background-contrast"
+          >
+            <Columns.ToggleButton child={1} icon={<Previous />} />
```
britt6612, line 52: So in mobile view would these go underneath each other ?

ericsoderberghp:
Archiving for now.

Issue #5204, opened 2021-04-30
Chromatic snapshots number deviation


jcfilben:
### Expected Behavior
The chromatic snapshots were around 320 before the latest release

<img width="863" alt="Screen Shot 2021-04-30 at 2 23 32 PM" src="https://user-images.githubusercontent.com/54560994/116761172-27279b00-a9d4-11eb-9dba-9b7697925ab9.png">

### Actual Behavior
The snapshot numbers jumped to 406 and include tests that have `parameters = { chromatic: { disable: true }, };` specified

<img width="859" alt="Screen Shot 2021-04-30 at 2 23 48 PM" src="https://user-images.githubusercontent.com/54560994/116761205-3c042e80-a9d4-11eb-94fc-da34506b71d0.png">

ShimiSun:
@jcfilben GREAT catch!! we need to take a look into this... it doesn't look right.

jcfilben:
I had a discussion with a member of the Chromatic team and their response was:

> We're in the process of rolling out some new features. So right now, we are showing all stories but not snapshotting disabled stories.
> Those will only have a canvas tab and not be tested. And it doesn't impact billing.
> ﻿
> ﻿Although they count as tests right now, which is a bug. The team is working on that fix.

So the additional tests from stories with `chromatic: { disabled: true }` are not actually being tested. They will still show up as tests but there isn't anything that can be done about this on our side.

jcfilben:
<img width="824" alt="Screen Shot 2021-05-26 at 1 44 03 PM" src="https://user-images.githubusercontent.com/54560994/119721516-72dd3100-be28-11eb-9a1e-d46b70cfa521.png">

Looks like the issue was fixed by the chromatic team and the number of snapshots has returned to normal

PR #5223, opened 2021-05-05
Upgrade Chromatic Version - test if snapshots numbers are affected

jcfilben:
#### What does this PR do?
I am upgrading to the latest version of chromatic to test if has any effect on the # of chromatic snapshots. Currently chromatic: { disable: true } is being ignored, I am seeing if updating to the latest version will fix this.

No

No

Backwards compatible


jcfilben:
Closing, see issue #5204, waiting to upgrade chromatic once they fix bug where stories with `chromatic: { disable: true }` show up as tests.

PR #5229, opened 2021-05-06
adds import for defaultProps to DataTable

DanielBaird:
Signed-off-by: Daniel Baird <daniel.baird@jcu.edu.au>

#### What does this PR do?
Like other Grommet components, DataTable will default to `defaultProps.theme` if there's no Theme context available. However defaultProps isn't imported. This PR adds that missing import.

Line 12 is the change; line 104 is the reference to defaultProps

I ran the standard tests, which passed (except for a date-related one because the test doesn't allow for my +10 timezone. I'll PR a fix for that later)

You could use a DataTable outside of a ThemeContext -- it will fail without this change.

I hit this bug while updating all my dependencies, when I accidentally included multiple instances of styled-components in my app. That broke ThemeContext enough that my DataTable tried to fall back to (the non-existent) defaultProps.

Nope

Not unless you list every bugfix.

Compatible.

### DIFF
--- a/src/js/components/DataTable/DataTable.js
+++ b/src/js/components/DataTable/DataTable.js
@@ -9,6 +9,8 @@ import React, {
 } from 'react';
 import { ThemeContext } from 'styled-components';
 
+import { defaultProps } from '../../default-props';
+
 import { useLayoutEffect } from '../../utils/use-isomorphic-layout-effect';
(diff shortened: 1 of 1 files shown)


ShimiSun (approved):
Thank you Daniel! 💟

PR #5230, opened 2021-05-07
add back in scroll for select

### LINKED ISSUE #5231: Select Scroll keyboard
The scroll is not working when you use keyboard on `select`
### Expected Behavior
The container on `select` should scroll up or down when you use keyboard arrows 

### Actual Behavior
The container does not move

https://user-images.githubusercontent.com/42451602/117394327-31093c80-aeb3-11eb-97e3-9308e3805e13.mov

### Steps to Reproduce

1. Go to many options example on storybook  2.
2. Tab into the `select` component 
3. Press `enter` to open the drop
4. Use down arrow to see that the container will not scroll as you get to the bottom

### Your Environment

- Grommet version: 2.17.2

britt6612:
#### What does this PR do?
This PR adds a  `useEffect` that looks for the position of the  `activeIndex` to bring the behavior of scroll for keyboard
selectContainer.js
keyboard on storybook 

any of longer options  in `select` for example many options use down arrow and the scroll will appear when you  get to the bottom of the container
**many options story** use down arrow and the scroll will appear when you  get to the bottom of the container
This code was previously here in `select` in  V2.12.0  however this was taken out because of the new behavior that  was added into `infiniteScroll` at the time this took care of the keyboard scroll. However when `infiniteScroll` behavior changed to `setState` for  `showProp` this no longer updated the `show` after the first render. 

https://github.com/grommet/grommet/blob/master/src/js/components/InfiniteScroll/InfiniteScroll.js#L156

closes #5231 
no
yes
backwards compatible


src/js/components/Select/SelectContainer.js
```diff
@@ -126,6 +131,29 @@ const SelectContainer = forwardRef(
       return undefined;
     }, [keyboardNavigation]);
 
+    // make sure options are visible in scroll
+    useEffect(() => {
```
taysea, line 135: Something doesn't seem quite right. When I get past 30 or so here, it stops scrolling/keeping the activeIndex in view: https://deploy-preview-5230--sad-tereshkova-173d07.netlify.app/?path=/story/input-select-many-options--many-options

taysea: Don't see that occurring on TextInput with suggestions: https://deploy-preview-5230--sad-tereshkova-173d07.netlify.app/?path=/story/input-textinput-suggestions--suggestions

src/js/components/Select/SelectContainer.js
```diff
@@ -126,6 +131,29 @@ const SelectContainer = forwardRef(
...
+                0,
+                buttonNode.offsetTop -
+                  (optionsNode.getBoundingClientRect().height -
+                    buttonNode.getBoundingClientRect().height),
+              );
+            } else if (isNodeBeforeScroll(buttonNode, optionsNode))
+              optionsNode.scrollTo(0, buttonNode.offsetTop);
+          }
+        }
+      }, 50);
```
ericsoderberghp, line 153: Can we add a comment explaining why `50` was chosen as the timeout duration?

ericsoderberghp (approved):

taysea (approved):

PR #5234, opened 2021-05-07
fix aria-label in drop button

### LINKED ISSUE #5189: Select component a11yTitle not forward to DropButton
### Expected Behavior
When an `a11yTitle` prop is passed to the Select component, I expect that it will propagate to the rendered button so that a screen reader user has context of the control.  Currently, a screen reader will read a select as "Open Drop, Button" even when `a11yTitle` is passed in.

See below

### Steps to Reproduce
1. Visit https://v2.grommet.io/select
2. Add an a11yTitle prop
3. Use a screen reader or view the HTML output to see that the rendered button does not have a custom `aria-label`.

### Your Environment
Latest grommet

- Grommet version: Latest
- Browser Name and version: Any
- Operating System and version (desktop or mobile): Any
Screen Shot 2021-04-25 at 6 42 24 PM
Screen Shot 2021-04-25 at 6 42 29 PM

britt6612:
#### What does this PR do?
This PR adds to the  `a11ylabel` in  `dropbutton` 
It was set as `Open Drop` so  for the component `select` that uses  `dropButton` the  aria-label was set to `Open Drop` 
DropButton.js
Screen reader with storybook

Also if you look at the `select`  snapshots they were updated correctly.. For instance 
```
    const { container } = render(
      <Grommet>
        <Select options={['one', 'two', 'three']} a11yTitle="test" />
      </Grommet>,
    );
  });

```
As you can see  above in this test its trying to pass  `a11yTitle="test"` this was not correct  in the snapshots  before this was  being printed as       
 `aria-label="Open Drop"`
Screen reader with storybook
closes  #5189 
no
sure
backwards compatible


src/js/components/Select/__tests__/__snapshots__/Select-test.js.snap
```diff
@@ -2412,7 +2412,7 @@ exports[`Select basic 1`] = `
 }
 
 <button
-  aria-label="Open Drop"
+  aria-label="Select"
```
taysea, line 2415: Why did this one change to Select?

src/js/components/DropButton/DropButton.js
```diff
@@ -9,7 +9,7 @@ const defaultDropAlign = { top: 'top', left: 'left' };
 const DropButton = forwardRef(
   (
     {
-      a11yTitle = 'Open Drop',
+      a11yTitle,
```
taysea, line 12:
```suggestion
      a11yTitle = 'Open Drop',
```
Can we leave this here as a default prop? instead of adding it below.

src/js/components/Select/__tests__/__snapshots__/Select-test.js.snap
britt6612: This is this test 

```
  test('basic', () => {
      <Select id="test-select" options={['one', 'two']} a11yTitle="Select" />,
    );
    expect(component.toJSON()).toMatchSnapshot();
  });

```

The allytitle was set to `select` so that's correct now

src/js/components/DropButton/DropButton.js
britt6612: yes looks cleaner!

taysea (approved):
LGTM!

ericsoderberghp (approved):

PR #5235, opened 2021-05-08
Missing target fix when valueLabel prop used in Select component

bojan88:
#### What does this PR do?
1. Fixes https://github.com/grommet/grommet/issues/4668
2. Fix from 518373864c2608df9faf4a6de8900ef5a55d8f20 to trigger change wasn't working when `valueLabel` is set, so this PR is fixing that too 

Tested with `valueLabel` and deprecated React components as options.
In case of React components difference is that after this change `event.target` will be present, but `event.target.value` will be an empty string. I gave up on fixing that because of the deprecation

https://github.com/grommet/grommet/issues/4668

No breaking changes


ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution!

Issue #3976, opened 2020-04-19
A trivial function imported from polished is adding 10kb to bundlesize


avin-kavish:
The rgba function, which essentially does nothing is the only import from `polished` and adds 10kb to the bundle.

It's used in situations like, 

```js
import { rgba } from 'polished';

const colors = {
  active: rgba(221, 221, 221, 0.5),
}
```

which can simply be replaced with the equivalent string, `rgba(221, 221, 221, 0.5)`.

https://bundlephobia.com/result?p=polished@3.5.1

![image](https://user-images.githubusercontent.com/48435155/79688486-4097f200-826c-11ea-8a94-969b08a1f0fe.png)

ShimiSun:
Thank you for the analysis!! we'll need to take a look at that.

avin-kavish:
No worries. Actually, I missed one situation where `rgba` seems to be needed very much.  https://github.com/grommet/grommet/blob/ab402dd96c0e402d1cfd8de2de9892b3b8ab11f3/src/js/themes/base.js#L802

ShimiSun:
Yea, it is helpful when we need the opacity to translate.
Any other alternatives?

avin-kavish:
Hmm... do you know the return format of this function? `normalizeColor('border', props.theme)`

ShimiSun:
Adding @ericsoderberghp to see if there is additional info that we should account for.

@avin-kavish feel free to share a PR if you already have something in mind.

g4rry420:
Can I work on this issue ?

taysea:
Closed by https://github.com/grommet/grommet/pull/5255

PR #5255, opened 2021-05-14
Remove polished package

g4rry420:
- components => StyledRangeInput.js
- themes => base.js, dark.js, 
- Select => stories => theme.js

yarn test-update

Using the storybooks

As mentioned in #3976 , `polished` is not much useful and adds 10kb to bundlesize. The only thing where polished's `rgba` was useful is in `StyledRangeInput.js` but similar logic can still applied through the already written utils function `getRGBA` which is helpful for our case.

#3976 

No

No

### DIFF
--- a/src/js/components/Select/stories/theme.js
+++ b/src/js/components/Select/stories/theme.js
@@ -1,8 +1,6 @@
-import { rgba } from 'polished';
-
 import { CaretDownFill, CaretUpFill, FormCheckmark } from 'grommet-icons';
 
-import { normalizeColor } from '../../../utils';
+import { normalizeColor, getRGBA } from '../../../utils';
 
 import { SearchInput } from './components/SearchInput';
 
@@ -12,7 +10,7 @@ export const theme = {
       selected: 'neutral-3',
       border: '#e0e0e0',
       focus: '#2196F3',
-      gray: rgba(0, 0, 0, 0.54),
+      gray: getRGBA(0, 0, 0, 0.54),
     },
     control: {
       border: {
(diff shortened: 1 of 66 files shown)


src/js/components/Select/stories/theme.js
```diff
@@ -12,7 +10,7 @@ export const theme = {
       selected: 'neutral-3',
       border: '#e0e0e0',
       focus: '#2196F3',
-      gray: rgba(0, 0, 0, 0.54),
+      gray: 'rgba(0, 0, 0, 0.54)',
```
taysea, line 13: confirming the strings around this is intended?

src/js/themes/base.js
```diff
@@ -50,7 +48,7 @@ const lightColors = [
 const focusColor = accentColors[0];
 
 const colors = {
-  active: rgba(221, 221, 221, 0.5),
+  active: 'rgba(221, 221, 221, 0.5)',
```
taysea, line 51: Same as above, did the previous rgba without string around still resolve to a string value?

taysea (commented):
Thanks for looking into this, just left a couple questions

src/js/components/Select/stories/theme.js
g4rry420: Yes, they are. Here, we are using polished's `rgba`.  But, we can use `rgba` provided by css.

taysea (approved):
I believe this looks good. Snapshot changes are just regenerated class names by styled components but no values seem to have shifted, which is fine.

ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution!

Issue #5247, opened 2021-05-11
DateInput does not behave as a controlled component


andrewthamcc:
### Expected Behavior

The `value` prop on the `DateInput` component passes an ISO date string to the component. I would expect that updating the prop would also update the display on the component.

### Actual Behavior

Updating the `value` prop and there is no change to the UI. However, when opening the `DateInput` the calendar displays the correct date that the displayed value should be.

### URL, screen shot, or Codepen exhibiting the issue

Simple CodeSandbox  where two buttons will change the ISO dates passed to each input.

https://codesandbox.io/s/grommet-v2-template-forked-8yrp8?file=/index.js

### Steps to Reproduce

View the CodeSandbox above, thank you.

### Your Environment

- Grommet version: 2.17.2

PR #5258, opened 2021-05-16
Update DateInput TextValue on Value prop change

andrewthamcc:
This PR fixes the behaviour described in the current open issue: https://github.com/grommet/grommet/issues/5247

The display value of the `DateInput` component should update when the value prop of the component changes. 

#### What does this PR do?

Adds a `useEffect` hook in `DateInput.js` to update the `textValue` which is the displayed value of the text iput.

View file `src/js/components/DateInput/DateInput.js`

NA

This could be manually tested by creating a storybook story that updates the passed in date value similar to the codesandbox presented in the issue above.

NA

NA

fixes the issues found here: https://codesandbox.io/s/grommet-v2-template-forked-8yrp8?file=/index.js

No

No

Backwards

### DIFF
--- a/src/js/components/DateInput/DateInput.js
+++ b/src/js/components/DateInput/DateInput.js
@@ -68,6 +68,16 @@ const DateInput = forwardRef(
     const [textValue, setTextValue] = useState(
       schema ? valueToText(value, schema) : undefined,
     );
+
+    // updates the text value if the value prop changes
+    useEffect(() => {
+      if (schema) {
+        setTextValue(valueToText(value, schema));
+      } else {
+        setTextValue(undefined);
+      }
+    }, [value, schema]);
+
     // We need to distinguish between the caller changing a Form value
     // and the user typing a date that he isn't finished with yet.
     // To track this, we keep track of the internalValue from interacting


ericsoderberghp:
Thanks for working on this. :)
I noticed in the Format inline story that if I select a date from the calendar and then try to adjust it by typing that it doesn't behave the same as the stable version. For instance, if I attempt to erase the last character, the whole value is cleared.

andrewthamcc:
> Thanks for working on this. :)
> I noticed in the Format inline story that if I select a date from the calendar and then try to adjust it by typing that it doesn't behave the same as the stable version. For instance, if I attempt to erase the last character, the whole value is cleared.

@ericsoderberghp thanks for taking a look. A fix for this could be to pass `value` as a dependency to the useEffect hook so it only updates the text value if there is a proper ISO time value and not `undefined`.  A problem I'm having with this is with the format range. I feel like It would be better to check if the `DateInput` was initialized as a range instead of:
```javascript
range = Array.isArray(value)
```

Initialing a state with `const [isRange, setIsRange] = useState(Array.isArray(value))` generates a lint error with `setIsRange` not used. And attempting to use useEffect hook with an empty dependency array also gives me a warning:
```react
const [isRange, setIsRange] = useState(false)

useEffect(() => (setIsRange(Array.isArray(value)), []) // asks value to be passed as a dependency but we only want the initial state of the component
```

`range` is currently changing with every update of `value`. While it appears to function correctly it's strange that props passed to `Calendar` would jump from undefined to an actual value and also switch boolean values:
```react
       range={range}
        date={range ? undefined : value}
        // when caller initializes with empty array, dates should be undefined
        // allowing the user to select both begin and end of the range
        dates={range && value.length ? [value] : undefined}
``` 

any thoughts?

andrewthamcc:
I also just noticed on the stable version that if the user attempts to clear the Range Format input with the keyboard an error is generated: `TypeError: Cannot read property '0' of undefined`

I will log another issue for this.

taysea:
@andrewthamcc seems like the reason the whole maskedinput clears is because the onChange in the MaskedInput is setting the value to undefined, then your useEffect is converting that to the textvalue. What about adjusting the condition to something like the following so we only touch the text value if it isn't an empty string? :
```
// updates the text value if the value prop changes
    useEffect(() => {
      if (schema && value !== '') {
        setTextValue(valueToText(value, schema));
      } else if (!schema) {
        setTextValue(undefined);
      }
    }, [value, schema]);
```

andrewthamcc:
> @andrewthamcc seems like the reason the whole maskedinput clears is because the onChange in the MaskedInput is setting the value to undefined, then your useEffect is converting that to the textvalue. What about adjusting the condition to something like the following so we only touch the text value if it isn't an empty string? :
> 
> ```
> // updates the text value if the value prop changes
>     useEffect(() => {
>       if (schema && value !== '') {
>         setTextValue(valueToText(value, schema));
>       } else if (!schema) {
>         setTextValue(undefined);
>       }
>     }, [value, schema]);
> ```

I agree, I have this in my comment above. However, I'm noticing that there are various issues with the DateInput component. This is noticeable on the Format Range and Format stories. 

The Format Story was last updated for double digit yearly values. While it appears to be working correctly keyboard functionality to enter a date no longer works. I had also opened an issue for the Format Range https://github.com/grommet/grommet/issues/5261

ericsoderberghp:
I've put up a separate pull request that I think addresses some of the issues with this one: #5330

PR #5263, opened 2021-05-17
Enhance Button badge docs

taysea:
#### What does this PR do?
Enhancing Button badge docs to elaborate what happens when a user provides an object of shape:
```
{
background: ,
value: ,
max: 
}
```

Updated here.

No.

Backwards compatible.

### DIFF
--- a/src/js/components/Button/README.md
+++ b/src/js/components/Button/README.md
@@ -163,10 +163,16 @@ function
 
 **badge**
 
-An indicator to show on the top-right of the Button. For accessibility,
-         supplement the badge by adding an a11yTitle to the Button which 
-         provides a useful message to screen readers. For example, "2 unread 
-         notifications".
+An indicator to show on the top-right of the Button. When badge is a 
+         boolean, the badge will be a circle with background color 'brand'. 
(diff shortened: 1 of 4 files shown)


src/js/components/Button/README.md
```diff
@@ -163,10 +163,16 @@ function
...
 
-An indicator to show on the top-right of the Button. For accessibility,
-         supplement the badge by adding an a11yTitle to the Button which 
-         provides a useful message to screen readers. For example, "2 unread 
-         notifications".
+An indicator to show on the top-right of the Button. When badge is a 
+         boolean, the badge will be a circle with background color 'brand'. 
+         When badge is a number, that number will appear inside the badge. When 
+         badge is an object, background refers to the background color of the 
+         badge, value refers to the number that will appear inside the badge,
```
ShimiSun, line 170: Are we assuming the value must be a number?

taysea: Updating to include description that value can be a boolean, like the doc type shows

ShimiSun (approved):

PR #5264, opened 2021-05-18
fixes the scroll item into view

g4rry420:
Signed-off-by: gurkiran_singh <gurkiransinghk@gmail.com>

#### What does this PR do?
Fixes the scroll into view for the select element as mentioned in #4288 

components => SelectContainer.js

yarn test-update

For the long list of Select options, key press the element name which is at the bottom or middle of the select options and it will come into view.

No

#4288 

No


src/js/components/Select/SelectContainer.js
```diff
@@ -279,6 +279,18 @@ const SelectContainer = forwardRef(
...
+              if (
+                optionsRef.current.childNodes[
+                  index
+                ].childNodes[0].childNodes[0].textContent
+                  .charAt(0)
+                  .toLowerCase() === event.key.toLowerCase()
+              ) {
+                optionsRef.current.childNodes[index].scrollIntoView(true);
+              }
+            }
```
jcfilben, line 293: This works with the Select/Simple story but not the Select/ManyOptions story. I swapped the array in both stories to be a list of letters in the alphabet so that it was easier to test. In the ManyOptions story I got the following error: `Uncaught TypeError: Cannot read property 'childNodes' of undefined`. I think the code is too specific to one use case of select.

src/js/components/Select/__tests__/Select-test.js
```diff
@@ -444,6 +444,7 @@ describe('Select', () => {
 
   test('select an option with keypress', () => {
     const onChange = jest.fn();
+    window.HTMLElement.prototype.scrollIntoView = function() {};
```
jcfilben, line 447: Not sure why the scrollIntoView statements were added to some of the tests. I don't think these tests should change because of this PR

g4rry420:
I am closing this as it has been already solved in #5230

PR #5265, opened 2021-05-18
Changed Grid to add width and height

ericsoderberghp:
#### What does this PR do?

Changed Grid to add width and height, same as Box.
This makes it easier to use Grid to drive layout without having to wrap with a Box to constrain its dimensions.

anywhere

added unit tests
tested with grommet-designer locally

use it

yes, automatic

yes

backwards compatible

### DIFF
--- a/src/js/components/Grid/__tests__/Grid-test.js
+++ b/src/js/components/Grid/__tests__/Grid-test.js
@@ -311,7 +311,7 @@ describe('Grid', () => {
   });
 
   test('width', () => {
-    const component = renderer.create(
+    const { container } = render(
       <Grommet>
         <Grid width="xsmall" />
         <Grid width="small" />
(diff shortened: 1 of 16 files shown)


MikeKingdom (commented):
Looks good to me. I like the refactoring.

taysea (approved):

src/js/components/Grid/__tests__/Grid-test.js
```diff
@@ -309,4 +309,54 @@ describe('Grid', () => {
     const tree = component.toJSON();
     expect(tree).toMatchSnapshot();
   });
+
+  test('width', () => {
+    const component = renderer.create(
```
ShimiSun, line 314: Good tests. 
I think we are trying to avoid the usage of `renderer.create` from `react-test-renderer` library when adding new tests. We should use `{ render } from '@testing-library/react'` instead.

ShimiSun (approved):

PR #5267, opened 2021-05-18
Feat/diagram animation

### LINKED ISSUE #3210: Diagram to add animation prop
I wonder if we should build this into Diagram? We could add `animation: true|false|draw|pulse` to both Diagram as a whole but also to individual objects in `connections`.

_Originally posted by @ericsoderberghp in https://github.com/grommet/grommet/pull/3208_

jlaluces123:
Work to add animation prop to Diagram with different parameters like true, false, draw, and pulse.

#### What does this PR do?
Resolves #3210 
- 3210 calls for animation prop at the component level as well as the connection level

- Check out Diagram's `index.d.ts` and `doc.js` files
- Run `yarn storybook`
- Experiment around with animations on Diagram's `Progressing` story

Manual testing on Progressing story for Diagram

Run yarn storybook and go to Diagram/Progressing

Completed Items:
[X] Component-level animation prop now available to use (ie. `<Diagram animation />`)
[X] Disabled chromatic snapshots
[X] Hoist the animation prop in `doc.js` to sibling-level of connections prop to be more accurate
[X] Implement connections-level animation usage (ex. having certain connections display specific animations)

#3210 

Yes:

1. New Animation prop for Diagram
2. Theme.Diagram.Animation prop for further customization

No

Will be backwards compatible


src/js/components/Diagram/StyledDiagram.js
```diff
@@ -1,12 +1,90 @@
...
+      }
+      to {
+        ${bounds[1]};
+      }
+    `;
+    return css`
+    ${keyframes`${animationTransition}`}
+    ${normalizeTiming(
+      animation.duration,
+      (theme.global.animation[animation.type]
```
ShimiSun, line 52: `theme.global.animation` and all of its sub-variations, should be added to the theme props on doc.js

src/js/components/Diagram/StyledDiagram.js
```diff
@@ -1,12 +1,90 @@
...
 
 import { defaultProps } from '../../default-props';
 
+const PULSE_SIZES = {
+  xsmall: 1.001,
+  small: 1.01,
+  medium: 1.1,
+  large: 1.5,
+  xlarge: 2,
+};
```
ShimiSun, line 11: Are these the same sizes you'll see on Box's pulse? If the code is repeated we might want to add it to utils and reuse it.

ShimiSun (commented):
Since we are using animation on the storybook example, there is no longer a need to take a Chromatic snapshot of it since it doesn't capture animation.
https://www.chromatic.com/test?appId=5d9774839a6eff00203f5cbf&id=60a42437f4542b003b86bac3
You can disable the chromatic snapshot for the story (look how it is done in other stories)

src/js/components/Diagram/StyledDiagram.js
jlaluces123: Totally agree here, will be adding it on!

src/js/components/Diagram/StyledDiagram.js
ShimiSun: Sounds like a good plan! Thanks!

src/js/components/Diagram/doc.js
```diff
@@ -64,6 +68,16 @@ export const themeDoc = {
     type: 'string | {dark: string, light: string}',
     defaultValue: 'accent-1',
   },
+  'global.animation': {
```
ShimiSun, line 71: we should split it to `global.animation.duration`, `global.animation.draw`

src/js/components/Diagram/doc.js
```diff
@@ -64,6 +68,16 @@ export const themeDoc = {
...
+    description: 'The animation configuration for Diagram.',
+    type: 'object',
+    defaultValue: `{
+      duration: '1s',
+      draw: {
+        duration: '0.1s',
```
ShimiSun, line 77: It is not clear how `duration` didders from `draw.duration`

src/js/components/Diagram/stories/Progressing.js
```diff
@@ -72,7 +72,7 @@ export const Progressing = () => {
               ))}
             </Box>
           </Box>
-          <Diagram connections={connections} />
+          <Diagram animation="draw" connections={connections} />
```
ShimiSun, line 75: I think this should transit slower.

ShimiSun: See if you can leverage `global.duration` to make the behavior transit slower when using `draw`

jlaluces123: With the newest implementation, I also allowed the user to specify a duration via an object for the animation prop:

`animation={{ type: 'draw', duration: 2000 }}`

And after testing this, it slowed down the animation to what we expected in our previous meeting. Next, I planned on trying out the `global.duration` stuff in the Progressing story that I've been using to test so far!

src/js/components/Diagram/StyledDiagram.js
jlaluces123: After testing the functionalities moving many of the overlapping helper functions to utils, I think it's safe to resolve this one 😄

src/js/components/Diagram/doc.js
jlaluces123: Looked into this today via a codesandbox, and it seems that you can specify a "catch-all" duration via `duration: "1s"` and for specific animations (in this case draw), you can also specify a specific duration for it as well.

src/js/components/Diagram/doc.js
jlaluces123: In reference to the next comment on the custom theme durations, I believe specifying here in the document that you can customize the attributes of the specific animations that are available to Diagram might relieve the confusion here. 

It took me a little bit of experimenting myself to realize that if you specify a `global.animation.duration` that will be a catch-all duration for any of the animations available for Diagram, and then `global.animation.draw` would be an object where you can specify specific attribute customizations to only that draw animation. I can aim to make this clearer in the `doc.js` for this component and I'll push it up to get some more thoughts.

ShimiSun: Since `global.animation.draw.duration` is specific to the diagram behavior only, we should have a theme prop on the diagram that provides `duration` of itself and it can fall back to the global animation value in case its `undefined`

src/js/utils/animations.js
```diff
@@ -0,0 +1,144 @@
+import { css, keyframes } from 'styled-components';
```
ShimiSun, line 1: We should make sure StyledBox is also using this util functions of `animation` and it works as expected without backward compatibility issues.

jlaluces123: Removed all the helper functions in StyledBox.js that were already in utils and all of them still work as intended!

src/js/components/Diagram/doc.js
jlaluces123: Updated the doc.js example, so now it's easy to understand that you can specify a duration under `global.animation`

src/js/components/Diagram/doc.js
```diff
@@ -18,6 +18,16 @@ export const doc = Diagram => {
     connections: PropTypes.arrayOf(
       PropTypes.shape({
         anchor: PropTypes.oneOf(['center', 'vertical', 'horizontal']),
+        animation: PropTypes.oneOfType([
```
halocline, line 21: I think I am missing a few details on the intent/scope of this PR, so I am going to ask a handful of clarifying questions. We may want to schedule a call if it is easier to talk through these.

Based on the the issue description, I was expecting animation settings to be available in two forms:
- Specified at the Diagram component level (assuming the prop values would apply to all connections)
- Specified at the each connection level (my assumption is that anything specified at this level would take precedence over the component level for that connection)

Is this expectation in line with the intent of this PR?

My subsequent comments may not be relevant if my expectation is off base.

halocline, line 21: In addition to this definition at the connection level, I was expecting documentation for `animation` to also be shown as a sibling to `connections`.

src/js/components/Diagram/doc.js
```diff
@@ -64,6 +74,11 @@ export const themeDoc = {
     type: 'string | {dark: string, light: string}',
     defaultValue: 'accent-1',
   },
+  'global.animation': {
+    description: 'The animation configuration for Diagram.',
```
halocline, line 78: Does this global animation configuration apply to component in addition to Diagram? If so, should this description be modified?

src/js/components/Diagram/StyledDiagram.js
```diff
@@ -1,12 +1,39 @@
...
+          animation: ${animationItemStyle(props.animation, props.theme)};
+        `}
+`;
 
 const StyledDiagram = styled.svg`
   max-width: 100%;
   width: 100%;
   height: 100%;
 
+  ${props => props.animation && animationStyle}
```
halocline, line 36: Per the documentation in doc.js, it looks like animation values can be set as part of a connection. Is that so? If so, where is that getting applied?

src/js/components/Diagram/doc.js
```diff
@@ -18,6 +18,16 @@ export const doc = Diagram => {
...
       PropTypes.shape({
         anchor: PropTypes.oneOf(['center', 'vertical', 'horizontal']),
+        animation: PropTypes.oneOfType([
+          PropTypes.bool,
+          PropTypes.oneOf(['pulse', 'draw']),
+          PropTypes.shape({
+            type: PropTypes.oneOf(['pulse', 'draw']),
+            delay: PropTypes.number,
+            duration: PropTypes.number,
+            size: PropTypes.string,
```
halocline, line 28: What does size affect? What are possible values?

src/js/components/Diagram/stories/Progressing.js
```diff
@@ -72,13 +72,20 @@ export const Progressing = () => {
               ))}
             </Box>
           </Box>
-          <Diagram connections={connections} />
+          <Diagram
+            animation={{ type: 'draw', duration: 2000 }}
```
halocline, line 76: This prop is being applied at the component level, however the documentation above is referencing as property part of connections.

halocline, line 76: Will this support things like delay and size?

Are there values are/are not acceptable by these? For example and delay of 3000 ms?

halocline (commented):
I fear that there's a mis-match between the work included here and my understanding of the scope/intent. Perhaps we can talk through the intended scope and any areas which were the source of my mis-understanding?

Also, it seems like there could be a decent amount of new functionality involved in this PR: new prop, new settings, etc. I am wondering how we might supplement this functionality with test coverage? What is the best way to test animation? Can this be done effectively in a Jest test? If not, what alternatives might we have?

src/js/components/Diagram/stories/Progressing.js
jlaluces123: Specified where delay works as well as added the size and a comment to show that size currently is only applicable to the pulse animation

src/js/components/Diagram/doc.js
jlaluces123: Added the possible values as well as specified in the code comment that pulse is the only animation currently that uses size

src/js/components/Diagram/StyledDiagram.js
```diff
@@ -1,12 +1,44 @@
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
 
 import { defaultProps } from '../../default-props';
+import { animationObjectStyle } from '../../utils/animations';
+
+const animationItemStyle = (item, theme) => {
```
taysea, line 6: Is `item` an animation type? if so, would it be more clear to call it `type` or `animationType`?

src/js/components/Diagram/stories/Progressing.js
```diff
@@ -72,13 +72,17 @@ export const Progressing = () => {
               ))}
             </Box>
           </Box>
-          <Diagram connections={connections} />
+          <Diagram animation connections={connections} />
```
taysea, line 75: Is the intent for the animation to re-draw everything each time a new connection is added? does the functionality support only "drawing" the new connections?

src/js/utils/animations.js
```diff
@@ -0,0 +1,139 @@
...
+    return ['opacity: 1;', 'opacity: 0;'];
+  }
+  if (type === 'jiggle') {
+    const deg = JIGGLE_SIZES[size];
+    return [`transform: rotate(-${deg}deg);`, `transform: rotate(${deg}deg);`];
+  }
+  if (type === 'pulse') {
+    return ['transform: scale(1);', `transform: scale(${PULSE_SIZES[size]})`];
+  }
+  if (type === 'rotateRight') {
```
taysea, line 52: What happens if I give my diagram an animation type of rotateRight? I understand it's not documented but since the component relies on this code will any errors result? does the diagram rotate?

src/js/components/Diagram/doc.js
taysea: I would agree here. It may be helpful to convert your Progressing.js story into a .tsx to make sure the type definitions are all accounted for

src/js/components/Diagram/stories/Progressing.js
jlaluces123: PR and documentation has been updated to be more accurate on where the Animation prop work is in terms of component level vs connection level.

src/js/utils/animations.js
jlaluces123: Now Fixed: issue where if you gave Diagram any of the animations available the Diagram would carry it out even though it wasn't intended for the component. Handled it now in StyledDiagram.js. Just returns `Not available for Diagram` to be more explicit.

src/js/components/Diagram/doc.js
jlaluces123: Resolving now after our discussion and PR content update to help clarify.

src/js/components/Diagram/stories/Progressing.js
jlaluces123: I think maybe as an enhancement down the line we could have a prop to serve as an option to either re-draw the lines on each new connection or keep all the connections made until a reload of the Diagram. Can definitely see use cases for both.

src/js/components/Diagram/Diagram.js
```diff
@@ -211,21 +221,37 @@ const Diagram = forwardRef(({ connections, ...rest }, ref) => {
...
+                {...cleanedRest}
+                stroke={normalizeColor(colorName, theme)}
+                strokeWidth={strokeWidth}
+                strokeLinecap={round ? 'round' : 'butt'}
+                strokeLinejoin={round ? 'round' : 'miter'}
+                fill="none"
+                d={d}
+              />
+            );
+          }
```
ShimiSun, line 253: This code should not be duplicated just for assigning the `animation` prop

src/js/components/Diagram/README.md
```diff
@@ -13,11 +13,35 @@ import { Diagram } from 'grommet';
...
+
+Animation to be used by entire Diagram
+
+```
+boolean
+pulse
+draw
+{
+  type: 
+    pulse
```
ShimiSun, line 26:
```suggestion
    pulse | draw
```

src/js/components/Diagram/doc.js
```diff
@@ -15,9 +15,35 @@ export const doc = Diagram => {
...
+            delay: PropTypes.number,
+            duration: PropTypes.number,
+            size: PropTypes.oneOf([
+              'xsmall',
+              'small',
+              'medium',
+              'large',
+              'xlarge',
+            ]),
+          }),
```
ShimiSun, line 45: This code doesn't have to be dulicated.

src/js/components/Diagram/index.d.ts
```diff
@@ -6,8 +6,32 @@ export type DiagramConnectionAnchor = 'center' | 'vertical' | 'horizontal';
...
+  animation?:
+    | boolean
+    | 'pulse'
+    | 'draw'
+    | {
+        type?: 'pulse' | 'draw';
+        // looking into it, but delay seems to only work when delay <= 1000
+        delay?: number;
+        duration?: number;
+        // currently, the only animation Diagram has that utilizes size is for "pulse"
```
ShimiSun, line 18: This comment should be placed on doc.js "size is only applicable when using `pulse`"

src/js/components/Diagram/index.d.ts
```diff
@@ -6,8 +6,32 @@ export type DiagramConnectionAnchor = 'center' | 'vertical' | 'horizontal';
...
+      | 'pulse'
+      | 'draw'
+      | {
+          type?: 'pulse' | 'draw';
+          // looking into it, but delay seems to only work when delay <= 1000
+          delay?: number;
+          duration?: number;
+          // currently, the only animation Diagram has that utilizes size is for "pulse"
+          size?: 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge';
+        };
```
ShimiSun, line 34: Shouldn't be duplicated

src/js/components/Diagram/stories/Progressing.js
ShimiSun: This information could be helpful for the user and should be added to the docs.

src/js/components/Diagram/stories/Progressing.js
```diff
@@ -72,13 +76,20 @@ export const Progressing = () => {
               ))}
             </Box>
           </Box>
-          <Diagram connections={connections} />
+          <Diagram
+            animation={{ type: 'draw', duration: 2000 }}
+            connections={connections}
```
ShimiSun, line 81: The connections are being drawn too fast, we should transit it slower.

src/js/components/Diagram/stories/Progressing.js
```diff
@@ -51,7 +51,11 @@ export const Progressing = () => {
 
   if (topRow.length >= 3) {
     connections.push(
-      connection('3', '5', { anchor: 'horizontal', color: 'brand' }),
+      connection('3', '5', {
```
ShimiSun, line 54: Let's add a better description to the title of the story, explaining to the user about the animation that is being used. one sentence should be sufficient.

src/js/components/Diagram/doc.js
ShimiSun: Top priority - clean up `global.animation.draw.duration`, this prop doesn't make sense to be placed on global if its only specific to Diagram. Move it to Diagram theme instead and account for it.

src/js/components/Diagram/Diagram.js
```diff
@@ -215,6 +225,7 @@ const Diagram = forwardRef(({ connections, ...rest }, ref) => {
             <path
               // eslint-disable-next-line react/no-array-index-key
               key={index}
+              animation={animation || undefined}
```
ShimiSun, line 228: wouldn't `animation` be undefined as the default? If so, this should work.
```suggestion
              animation={animation}
```

src/js/components/Diagram/index.d.ts
```diff
@@ -4,10 +4,24 @@ import { ColorType } from '../../utils';
 
 export type DiagramConnectionAnchor = 'center' | 'vertical' | 'horizontal';
 export type DiagramConnectionType = 'direct' | 'curved' | 'rectilinear';
+export type DiagramAnimationType =
+  | boolean
+  | 'pulse'
+  | 'draw'
+  | {
+      type?: 'pulse' | 'draw';
+      // looking into it, but delay seems to only work when delay <= 1000
```
ShimiSun, line 13: What are the findings? The comment shouldn't be placed here.

src/js/components/Diagram/Diagram.js
jlaluces123: Tested this out locally and everything looks good.

src/js/components/Diagram/index.d.ts
jlaluces123: Just tried out a delay again on the story I was working on and it works as long as you don't run into common things like giving a delay longer than the duration or timer. Pushed up a commit getting rid of the comment on the delay as well as the redundant comment left on the `size` attribute since it is included on the `doc` already.

src/js/components/Diagram/stories/Progressing.js
jlaluces123: Progressing story now shows the animation transiting slower.

src/js/components/Diagram/doc.js
jlaluces123: Recent commits now solve this. Now they're under a variable called `ANIMATION_PROP` that get plugged in at both the connection level and component level

src/js/components/Diagram/index.d.ts
jlaluces123: Now consolidated into a variable called `DiagramAnimationType` that gets plugged into both the component level, and connection level

src/js/components/Diagram/index.d.ts
jlaluces123: Recent commit now has this comment in `doc.js` for `size` attribute's description.

src/js/components/Diagram/doc.js
jlaluces123: Updated this into `diagram.animation` now for more accuracy so it only relates to items related to Diagram's animations.

src/js/components/Diagram/doc.js
jlaluces123: This prop was now moved to `diagram.animation` to keep it specific to diagram's behavior only.

src/js/components/Diagram/StyledDiagram.js
```diff
@@ -1,12 +1,80 @@
...
+  }
+  return '';
+};
+
+const availableAnimations = [true, 'draw', 'pulse'];
+
+const animationStyle = props => {
+  const animationType = props.animation.type || props.animation;
+
+  if (!availableAnimations.includes(animationType)) {
```
taysea, line 24: We don't frequently return messages -- nothing inherently wrong here but just breaks a little form our typical pattern. Wondering instead if `animationStyle` should only be called if `availableAnimation.includes(animationType)`. This check should go below wherever animationStyle is being called

src/js/components/Diagram/StyledDiagram.js
```diff
@@ -1,12 +1,80 @@
...
+    }
+    return '';
+  });
 
 const StyledDiagram = styled.svg`
   max-width: 100%;
   width: 100%;
   height: 100%;
 
+  ${props =>
```
taysea, line 75: Nice!

taysea: Maybe add a comment about why this needs to happen before the props.animation below? (otherwise connection level animations won't work correctly)

src/js/components/Diagram/doc.js
```diff
@@ -3,6 +3,23 @@ import { describe, PropTypes } from 'react-desc';
 import { colorPropType } from '../../utils/prop-types';
 import { getAvailableAtBadge } from '../../utils/mixins';
 
+const ANIMATION_PROP = PropTypes.oneOfType([
```
taysea, line 6: Naming recommendation to match with other `prop-types` in utils: `animationPropType`

src/js/components/Diagram/stories/Progressing.js
```diff
@@ -39,7 +39,7 @@ export const Progressing = () => {
   useEffect(() => {
     const timer = setInterval(() => {
       dispatch();
-    }, 2000);
+    }, 3000);
```
taysea, line 42: Is there a reason we had to increase this 1s?

src/js/components/Diagram/stories/Progressing.js
```diff
@@ -51,14 +51,22 @@ export const Progressing = () => {
...
+      }),
     );
   }
 
   return (
     <Grommet theme={grommet}>
       <Box align="start" pad="large">
-        <Text> Adding and removing nodes</Text>
+        <Text>
+          {' '}
```
taysea, line 66:
```suggestion
```

taysea (commented):
Just a few more comments, but looking good

src/js/components/Diagram/StyledDiagram.js
jlaluces123: Great point and thanks for the tip, I definitely agree with this, trying it out now.

src/js/components/Diagram/stories/Progressing.js
jlaluces123: For this story as a whole, I wanted to extend the duration of the animations to show a smoother transit. Increasing the timer here allows for longer durations on animations to better showcase to users the animations available on Diagram.

src/js/components/Diagram/stories/Progressing.js
jlaluces123: Must've bugged this one in the linter check somehow but this got sorted out now.

src/js/components/Diagram/StyledDiagram.js
jlaluces123: Added the comment, after looking up what was recommended for commenting inside a styled component but general consensus seemed to just comment it like normal CSS. Let me know if it looks okay in the newest commit I pushed.

src/js/components/Diagram/StyledDiagram.js
jlaluces123: For `animationStyle` I implemented it when we're calling `${props => props.animation && ...}` but as for `connectionStyle` I did take out the return message and left it as an empty string but left the `if statement` since the check has to happen inside the `connections.map`. Let me know your thoughts.

src/js/components/Diagram/StyledDiagram.js
```diff
@@ -38,14 +33,16 @@ const animationStyle = props => {
...
 
+const availableAnimations = [true, 'draw', 'pulse'];
+
 const connectionStyle = (connections, theme) =>
   connections.map((connection, index) => {
     if (connection !== undefined && connection.props.animation !== undefined) {
       const { type } = connection.props.animation;
       const animationType = type || connection.props.animation;
 
       if (!availableAnimations.includes(animationType)) {
```
taysea, line 44: Similar to the above comment -- should we only call connectionStyle if the animationType is included in the available animations array?

jlaluces123: Gonna take a minute, but I'll see what I can come up with!

src/js/components/Diagram/README.md
```diff
@@ -68,6 +109,24 @@ svg
...
   
+**diagram.animation**
+
+The animation configuration for Diagram. Expects `object`.
+
+Defaults to
+
+```
+{
+      duration: '1s',
```
ShimiSun, line 120: Why do we need this `duration` property within Diagram? Is it actually being used?
If it is used we should consider using `global.animation.duration` instead. any reason you considered separating it from the global value?

jlaluces123: I originally thought if we'd put the `duration` property within Diagram, a user might want to only affect just the duration of all the Diagram's they use. After giving it more thought I think it does seem redundant to put a base duration for Diagram and instead make it available at `global.animation.duration`.

src/js/components/Diagram/index.d.ts
```diff
@@ -4,10 +4,22 @@ import { ColorType } from '../../utils';
 
 export type DiagramConnectionAnchor = 'center' | 'vertical' | 'horizontal';
 export type DiagramConnectionType = 'direct' | 'curved' | 'rectilinear';
+export type DiagramAnimationType =
+  | boolean
```
taysea, line 8: Did you try these out in a .tsx storybook example to make sure all the typings are working as expected? asking from past experience of it not working as I expect hahaa

jlaluces123: The .tsx migration went well, gave the animation's arguments unexpected results and everything functions as expected.

taysea (approved):
Nice job on this!

src/js/components/Diagram/StyledDiagram.js
```diff
@@ -1,12 +1,82 @@
...
 
 import { defaultProps } from '../../default-props';
+import { animationObjectStyle } from '../../utils/animations';
+
+const animationItemStyle = (animationType, theme) => {
+  if (typeof animationType === 'string') {
+    return animationObjectStyle({ type: animationType }, theme);
+  }
+  if (typeof animationType === 'object') {
+    return animationObjectStyle(animationType, theme);
```
halocline, line 11: What happens if I call diagram with something like the following?
```
<Diagram animation={{ duration: 5000 }} />
```

With boolean, we default to 'draw'. Should we also default to 'draw' in this case? If so, should we account for this case in `animationStyle()`? Or is it on the caller to figure out?

halocline (commented):
The interface to apply animations to both Diagram as well as connections aligned with my expectations, was clear and intuitive. Nice work.

I tried to break things and had a hard time finding anything. The only situation to consider is where `animation={{ duration: 5000 }}` is passed without a 'type'.

Other considerations:
- One suggestion for tweaking/clarifying the story.
- Should the "Animation" story be refactored to remove the animation being executed in the custom theme account for this work?


Other than that, I think this is ready for approval.

src/js/components/Diagram/StyledDiagram.js
jlaluces123: I'll test this out tonight as well, I actually didn't think of this! I think by default if not given a type in the object it should default to draw as it is implied that you want an animation by default.

halocline (approved):
Nice work, Jonathan.

src/js/components/Diagram/StyledDiagram.js
ShimiSun: What are the findings of this thread?

src/js/utils/animations.js
```diff
@@ -0,0 +1,150 @@
...
+
+    const defaultDuration = () => {
+      if (theme.diagram.animation !== undefined) {
+        return normalizeTiming(
+          theme.diagram.animation[animation.type]
+            ? theme.diagram.animation[animation.type].duration
+            : undefined,
+          theme.diagram.animation.duration,
+        );
+      }
```
ShimiSun, line 138: I don't think that the `diagram` specific code should sit on a generic utils file. The diagram logic should be performed on the diagram itself and sent to utils for the calculation.

Regardless, I don't think we should have a separate duration for each type of animation, but instead one duration for all types of animation to keep the API simple. Is Box allowing different duration for each type of animation?

src/js/components/Diagram/StyledDiagram.js
jlaluces123: Both the connection level and component level animations will default to `draw` now if given an object but no specific `type` like the example Matt highlights in the thread.

ShimiSun: I think that if the type of `animation` is undefined it should default to how it is currently behaving.

jlaluces123: Just to clarify @ShimiSun, so should I keep it where it defaults to `draw` or if `animation` is undefined, nothing happens and the connection will instantly show up?

ShimiSun: As long as current users wouldn't have an added animation on the default behavior I think we should be good. Only if users specify `animation` prop it will default animation behavior to `draw`.

src/js/components/Diagram/doc.js
```diff
@@ -54,6 +76,18 @@ export const doc = Diagram => {
 };
 
 export const themeDoc = {
+  'diagram.animation': {
+    description: 'Any animation specific configuration for Diagram.',
```
ShimiSun, line 80: "Any animation" is not so accurate, It's only applicable for 'draw' and 'pulse'.

src/js/components/Diagram/doc.js
```diff
@@ -64,6 +98,13 @@ export const themeDoc = {
     type: 'string | {dark: string, light: string}',
     defaultValue: 'accent-1',
   },
+  'global.animation': {
+    description: 'The animation configuration for Diagram.',
+    type: 'string',
+    defaultValue: `{
+      duration: '1s'
+    }`,
```
ShimiSun, line 106: `type` is contradicting the `defaultValue`, is it a string or an object?

src/js/components/Diagram/stories/Animated.js
```diff
@@ -106,7 +90,7 @@ export const Animated = () => {
                 ))}
               </Box>
             </Box>
-            <Diagram connections={connections} />
+            <Diagram animation={{ duration: 3000 }} connections={connections} />
```
ShimiSun, line 93: For the example we might want to demonstrate an example that is more detailed and add in the `type`

src/js/components/Diagram/stories/Progressing.js
```diff
@@ -51,14 +51,22 @@ export const Progressing = () => {
...
   }
 
   return (
     <Grommet theme={grommet}>
       <Box align="start" pad="large">
-        <Text> Adding and removing nodes</Text>
+        <Paragraph>
+          Adding and removing nodes with animated connections. The animation
+          &apos;draw&apos; is applied to the entire diagram, however, the last
+          connection receives its own animation &apos;pulse&apos;.
```
ShimiSun, line 68:
```suggestion
          connection receives its own animation type of &apos;pulse&apos;.
```

src/js/components/Diagram/stories/Animated.js
jlaluces123: Totally agree, added the `type` prop back in

src/js/components/Diagram/doc.js
jlaluces123: Fixed this to be `type: object`

src/js/components/Diagram/doc.js
jlaluces123: Updated the text for this theme prop, please let me know what you think

src/js/components/Carousel/__tests__/__snapshots__/Carousel-test.js.snap
```diff
@@ -121,8 +121,8 @@ exports[`Carousel basic 1`] = `
   -ms-flex-direction: column;
   flex-direction: column;
   opacity: 1;
-  -webkit-animation: hftlBD 1s 0s forwards;
-  animation: hftlBD 1s 0s forwards;
+  -webkit-animation: hftlBD 0s 0s forwards;
```
ShimiSun, line 124: @halocline had been exploring this snapshot changes with the following insights:
>The snapshot changes for Carousel now represent the intended logic.
L158 specifies an animation duration of ‘0’, which is passed to Box’s animation on L164.
The value of 0 is evaluated (incorrectly) in the normalizeTiming function here. Instead of the function interpreting the value as 0 , it is evaluating it as a boolean and therefore interprets 0 as falsey. As consequence, normalizeTiming() had been returning the defaultTiming of 1s instead of the intended 0s.
Jonathan’s PR modifies the normalizeTiming() function and correctly interprets 0 as a numeric value and correctly returns 0s. This is the source of Carousel’s snapshot changes.

jlaluces123: Yup! This is exactly what @halocline and I discussed!

ericsoderberghp (approved):

PR #5270, opened 2021-05-19
Add options prop to Grommet to support configurations

### LINKED ISSUE #5239: Layer places `id` onto multiple DOM nodes
### Expected Behavior

`id` should only be placed on a single DOM node.

### Actual Behavior

`id` is being placed on 2 DOM nodes (StyledContainer, StyledLayer). This is causing errors on Design System accessibility tests that flag when multiple elements have the same id tag.

### Steps to Reproduce

1. Create a Layer with `id`
2. Run accessibility test like `axe`
3. Error of `duplicate-id`

### Your Environment

- Grommet version: 2.17.2
- Browser Name and version: Chrome on Mac

britt6612:
#### What does this PR do?
This PR adds a opt in for users to new behavior that we would eventually want to move towards `layer` having `id` on only one component in dom
layer.js
storybook
storybook look in the dom you can also run 
* Run accessibility test like axe
* Error of duplicate-id
`id` is being placed on 2 DOM nodes (StyledContainer, StyledLayer). This is causing errors on Design System accessibility tests that flag when multiple elements have the same id tag.
closes #5239 
auto
yes
backwards compatible


ericsoderberghp:
I think the opt-in has to be in the caller’s code, rather than the theme.

britt6612:
> I think the opt-in has to be in the caller’s code, rather than the theme.

Are you thinking more along the lines of adding a `prop` ?

ericsoderberghp:
> > I think the opt-in has to be in the caller’s code, rather than the theme.
> 
> Are you thinking more along the lines of adding a `prop` ?

Yes, perhaps something like `<Grommet options={{ singleId: true }} />`, which we would leverage for cleaning up both Layer and Select in how they manage `id`s.

britt6612:
> > > I think the opt-in has to be in the caller’s code, rather than the theme.
> > 
> > 
> > Are you thinking more along the lines of adding a `prop` ?
> 
> Yes, perhaps something like `<Grommet options={{ singleId: true }} />`, which we would leverage for cleaning up both Layer and Select in how they manage `id`s.

awe yes! Good thinking let me explore this option!

src/js/components/Layer/stories/Round.js
```diff
@@ -6,6 +6,7 @@ import { FormClose } from 'grommet-icons';
 
 const customTheme = deepMerge(grommet, {
   layer: {
+    intelligentID: true,
```
taysea, line 9: We can remove this now
```suggestion
```

src/js/components/Layer/stories/Round.js
```diff
@@ -61,6 +62,7 @@ export const RoundLayer = () => {
           position={position}
           onClickOutside={onClose}
           onEsc={onClose}
+          id="new layer"
```
taysea, line 65:
```suggestion
```
we can remove this since you've already demo'd the new behavior in a different story.

src/js/components/Grommet/doc.js
```diff
@@ -21,6 +21,14 @@ export const doc = Grommet => {
     full: PropTypes.bool
       .description('Whether to take the whole viewport.')
       .defaultValue(false),
+    options: PropTypes.object
+      .description(
+        `Opt in that would assure that the id is placed on
```
taysea, line 26: It seems like this is being set up to potentially allow other options/specifications in the future. What about something like (probably needs some refinement):

Any other options to apply globally throughout your application. If singleId is true, any `id` placed on Layer will only apply to the DOM node associated with the visual portion of the Layer and any `id` placed on Select will only apply to its underlying input. This opt-in flag is intended to resolve accessibility considerations that `id` should be unique and placed on a single DOM node while maintaining backwards compatibility with the existing behavior.

taysea (commented):
This direction looks good, but have a use case to consider:

Right now in DS site, we use `id` on SkipLinks (built with Layer) to identify the SkipLinks DOM element and `.focus()` it on a route change so the first tab press will open the SkipLinks. This is necessary because NextJS doesn't reset the tab focus placement to the SkipLinks on a route change. Therefore, we do:
```
on route change completion
document.querySelector('#skip-links').focus()
```

With this new behavior where the `id` is only on StyledContainer, the SkipLinks don't properly re-receive focus with the above snippet. It seems we need the focus on StyledLayer (or because StyledContainer is inside StyledLayer which has tabIndex={-1} it can't receive focus). If we opt-in to the `singleId`, we would need to adjust our code to be:
```
document.querySelector('#skip-links').parentElement.focus()
```
Is that okay? or should we not have to do that.

src/js/components/Grommet/doc.js
taysea: ^^ or if this PR isn't going to address Select, then maybe remove the Select stuff I mention above.

src/js/components/Grommet/index.d.ts
```diff
@@ -7,7 +7,8 @@ export interface GrommetProps {
   containerTarget?: HTMLElement;
   cssVars?: boolean;
   dir?: 'rtl';
-  full?: boolean | 'min';
+  full?: boolean;
```
taysea, line 10:
```suggestion
  full?: boolean | 'min';
```

src/js/components/Grommet/Grommet.js
```diff
@@ -50,6 +51,7 @@ const Grommet = forwardRef((props, ref) => {
     full,
     containerTarget = typeof document === 'object' ? document.body : undefined,
     theme: themeProp,
+    options = {},
```
ericsoderberghp, line 54: I think this will cause the context to be updated every time. If we create a `const defaultOptions = {};` outside of `Grommet`, I think that will avoid that.

src/js/components/Grommet/doc.js
```diff
@@ -25,6 +25,14 @@ export const doc = Grommet => {
         control scrolling.`,
       )
       .defaultValue(false),
+    options: PropTypes.object
+      .description(
+        `Additional configurations that will be applied throughout the 
+        application. If options.layer.singleId === true, id on Layer will be 
+        placed on StyledLayer and '{id}__container' will be placed on
```
ShimiSun, line 32: I think the terminology of StyledLayer is confusing to the reader since users that are not familiar with the source code wouldn't understand what it is referring to.

taysea: Good feedback. Let me work on the wording

src/js/components/Grommet/README.md
```diff
@@ -69,6 +69,17 @@ boolean
 min
 ```
 
+**options**
+
+Additional configurations that will be applied throughout the
```
taysea, line 74: @ShimiSun can you read this and see if it's more clear?

ShimiSun: I think it's better, but I still find it too technical and less descriptive.

For example:
> If options.layer.singleId === true

What is this prop `layer.singleId` and where do we control it (theme/prop)? 
Should we really use this code to explain the functionality or can we use words instead?
What is the purpose of this `options` prop and how can the user leverage it?

src/js/components/Grommet/index.d.ts
```diff
@@ -8,6 +8,7 @@ export interface GrommetProps {
   cssVars?: boolean;
   dir?: 'rtl';
   full?: boolean | 'min';
+  options?: object;
```
ShimiSun, line 11: I think this should be replaced with the actual expected structure of the object that `options` can accept.

src/js/components/Grommet/doc.js
```diff
@@ -25,6 +25,14 @@ export const doc = Grommet => {
         control scrolling.`,
       )
       .defaultValue(false),
+    options: PropTypes.object
+      .description(
+        `Additional configurations that will be applied throughout the 
+        application. If options.layer.singleId === true, Layer id will be 
+        placed on the outermost Layer div and '{id}__container' will be placed 
+        on the div that creates the visible portion of the Layer.`,
```
taysea, line 33: Additional component configurations that will be applied throughout the application. For backwards compatibility, some aspects of a component, such as id placement, cannot be changed within all of Grommet. This prop provides users with the ability to opt-in to newer functionality. If layer.singleId = true, Layer id will be placed on a single div that wraps all the Layer elements.

taysea: options is a way for the Layer to assign a unique id to a single DOM node. currently, it supports only unique id for Layer.

ShimiSun (approved):

PR #5272, opened 2021-05-19
Normalize theme.checkBox.toggle.color

### LINKED ISSUE #5271: Specifying a color string in `theme.checkBox.toggle.color` does not work
### Expected Behavior

Specifying a color name or string for `theme.checkBox.toggle.color` should result in the checkbox toggle being displayed in that color.

### Actual Behavior

Specifying a color string does not work.

The workaround is to specify a color spec instead.

https://codesandbox.io/s/blissful-violet-3t6lq?file=/src/App.js

### Your Environment

- Grommet version: 2.17.2
- Browser Name and version: Firefox 88.0.1
- Operating System and version (desktop or mobile): OSX 10.14.6

jasnross:
#### What does this PR do?

This PR fixes #5271 by removing the light/dark switch and relying on normalizeColor to normalize the color spec for light/dark instead.

Manual testing by adjusting the CheckBox story to use color strings.

1. Run the storybook
2. Check the custom toggle story: http://localhost:9001/?path=/story/input-checkbox-custom-toggle--custom-toggle
3. Update src/js/components/CheckBox/stories/CustomToggle.js to use color strings instead of objects
4. Confirm results are the same

#5271 

Yes... the docs currently say the following:

image

But I believe it should instead say `string | { 'dark': string, 'light': string }`

It looks like it is displaying the `defaultValue` instead of the `type` from the docs here: https://github.com/grommet/grommet/blob/master/src/js/components/CheckBox/doc.js#L158-L162

I wasn't sure how to go about updating it to show the type value instead though.

This is backward compatible


halocline (approved):
This all looks good by me.

The docs and typescript types are all defined correctly. However, I think @jasnross's question about displaying type in addition to values / defaultValue warrants a discussion.

There is a Chromatic snapshot flagged on a Carousel story, but that seems to be unrelated.

ericsoderberghp (approved):

Issue #5145, opened 2021-04-05
Enhance RangeInput with thumb Label and dynamic track color based on a prop value


q-oh:
This is not an issue. Either Grommet doesn't support or I can't seem to figure out.
Currently looking to add the following:
1. Thumb label
2. Dynamic track color based on a prop value (range input value)

Image: 
![image](https://user-images.githubusercontent.com/81991460/113619180-8250b280-960d-11eb-80e1-85e8c9ea4390.png)

### Expected Behavior

- Be able to add an element/string/number to a prop which will generate a label for a thumb
- Be able to add a series of background colors and their corresponding threshold so that track color could reflect based on the range input value.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-forked-puh94?file=/index.js

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: "^2.14.0"
- Browser Name and version:
- Operating System and version (desktop or mobile):

ericsoderberghp:
Look at how we support gradients with Chart to see if we can do something similar.
https://storybook.grommet.io/?path=/story/visualizations-chart-gradient--gradient-charts
https://v2.grommet.io/chart#color

q-oh:
Hi @ericsoderberghp - would it be possible to get an ETA? Trying to plan for our project release. Thanks

ericsoderberghp:
> Hi @ericsoderberghp - would it be possible to get an ETA? Trying to plan for our project release. Thanks

This isn't at the top of the grommet team's work list: https://github.com/orgs/grommet/projects/10?fullscreen=true

Would you be interested in contributing the changes for it?

g4rry420:
@ericsoderberghp  My apologies. I am not sure if the issues moved to `Grommet/Design System Backlog ` can be picked by anyone or not. Still, I find this issue interesting and have worked on `Dynamic track color based on a prop value` part. Wanted to know your thoughts if this is how you wanted to implemented. For the passing the colors, I used used the `rangeInput.track.colors` as an array of objects.


https://user-images.githubusercontent.com/61521805/118433105-40089f80-b6a8-11eb-9e50-58602f8478a9.mp4

halocline:
Closed by #5275

robustdj:
Would it be possible to make the colors fixed? Right the track color will change depending on the value. I have a `RangeInput` such that I want the colors to remain there regardless of the value

[...]

PR #5275, opened 2021-05-21
Feat: Dynamic color for track and thumb Label

g4rry420:
#### What does this PR do?
Added the dynamic color for track and also added thumb label positon.

components => RangeInput

storybooks => RangeInput => MultiColorTrack.js

#5145 

Yes, I did.

backwards compatible


ericsoderberghp:
Thanks for working on this issue. I have a few broad thoughts:
- I like the idea of adding the colors to the track.
- I think the track coloring needs to be done at the RadioButtonGroup property level instead of the theme level. The reason for that is that the value can be variable, it could be 0-10, 0-100, 20-80, etc. So, the caller needs to be able to control the connection between value ranges and specific colors.
- There is an issue with how the gradient is being rendered. The underlying CSS gradient is fading away too soon:
Screen Shot 2021-05-26 at 5 30 15 PM
- I think we should handle the label aspect in a separate pull request as I think it is orthogonal to the track coloring.
- I think we should consider the API surface for the label such that it can eventually be positioned: below thumb, above thumb, right of the bar, etc.
- There is a connection to a separate pull request to allow the user to type the value in. https://github.com/grommet/grommet/pull/4986 I'm wondering if we should consider that as a primary use case, since it's generally more usable.

g4rry420:
> * There is an issue with how the gradient is being rendered. The underlying CSS gradient is fading away too soon:

I agree with your mentioned points and will work on them. 
- For the color gradient as an issue, I was keeping it intentional as if the current value of the `RangeInput` hasn't reached at the provided value, then, it will show faded.
- For implementating it using RadioButtonGroup, are you agree with the prop name of `colors` ?
Please let me know if anything else you wanted to add.

ericsoderberghp:
> * For the color gradient as an issue, I was keeping it intentional as if the current value of the `RangeInput` hasn't reached at the provided value, then, it will show faded.

I don't think that's the correct visual behavior, as it feels like the dot is encompassed by the same unset color on either side of it. I think the semantic way to use this is to set color values to be a range, anything between A and B is one color, anything between B and C is another color, always solid. You can change A, B, and C to suite your situation, but the colors should always be solid.

> * For implementating it using RadioButtonGroup, are you agree with the prop name of `colors` ?
>   Please let me know if anything else you wanted to add.

Oops, I meant to say "RangeInput", not "RadioButtonGroup". I think the property should be named `color` and support a single string for a color name that will be used for the left side of the thumb, or an array of objects containing `value` and `color` keys. This would align with how `color` works for `Chart`:
```
color: PropTypes.oneOfType([
      colorPropType,
      PropTypes.arrayOf(
        PropTypes.shape({
          color: colorPropType,
          value: PropTypes.number,
        }),
      ),
    ])
```

g4rry420:
The way I see the `normaliseColor` function is that it takes care of the theme modes supplied to a specific element i.e `dark` and `light` modes and it also takes care of the Grommet specific color such as `brand`, `accent-3` and converts them into the hex format in the end result.
The `getRGBA` further checks this hex format color and see if this can be converted into `rgba` value using `canExtractRGBArray(color)` function and if this function returns true it further converts the hex value into rgba.

Now, all of these works well, these functions above conveys the Grommet conventions around the color property as specified by @halocline above.But there is only one case where we are getting `color` prop result as `undefined` - because of  `html color name` such as `red`, `coral`, etcetera. As I further looked I found that when we pass `color="red"` and when it comes around the 
`canExtractRGBArray(color)` it returns false and further the resultant value becomes `undefined`.

Other than that, all the Grommet conventions all working well. I have modified and added the stories also. I have also changed the proptypes and typescript definitions. @halocline  Please let me know if you find any issues in my last commit.

src/js/components/RangeInput/stories/Multi.js
```diff
@@ -0,0 +1,83 @@
...
+            min={0}
+            max={10}
+            step={1}
+            value={value}
+            onChange={onChange}
+            color={[
+              { value: 3, color: '#FF0000', opacity: 0.5 },
+              { value: 7, color: { light: 'accent-3', dark: 'brand' } },
+              { value: 10, color: '#00FF00' },
+            ]}
```
halocline, line 60: 👍  I think this demonstrates the idea of the functionality and options well.

src/js/components/RangeInput/stories/Multi.js
```diff
@@ -0,0 +1,83 @@
+import React from 'react';
+
+import { Box, Button, Grommet, RangeInput } from 'grommet';
+
+import { Add, Subtract } from 'grommet-icons';
+
+const rangeInputTheme = {
```
halocline, line 7: Wondering if we can keep the story entirely focused on the "multi-color track" aspect by removing this custom theming?

src/js/components/RangeInput/stories/Modes.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+
+import { Box, Grommet, RangeInput } from 'grommet';
+
+import { Volume } from 'grommet-icons';
+
+// Type annotations can only be used in TypeScript files.
+// Remove ': ThemeType' if you are not using Typescript.
+const customThemeRangeInput = {
```
halocline, line 9: Thinking we want to focus on the `color` prop aspect. Consider removing the custom theme so that developers are able to focus on the relevant aspect of the example.

halocline (commented):
Thank you for continuing to advance this @g4rry420. I appreciate the additional tests.

I've left some comments around:
- Simplifying the prop type definitions.
- Since "stories" are intended as a teaching tool, simplifying the story implementations to focus solely on the relevant feature we want to showcase, in this case the `color` prop, and removing any extras which might crowd the example.
- Ensuring the `color` prop works on its own without any custom theme specified.

src/js/components/RangeInput/index.d.ts
```diff
@@ -1,5 +1,16 @@
 import * as React from 'react';
-import { A11yTitleType } from '../../utils';
+import { A11yTitleType, ColorType } from '../../utils';
+
+interface ModesI {
+  light: ColorType;
+  dark: ColorType;
+}
```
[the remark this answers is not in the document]
g4rry420: Sure, I will

src/js/components/RangeInput/stories/Multi.js
g4rry420: Sure, will do

src/js/components/RangeInput/stories/Modes.js
g4rry420: Sure, will do

g4rry420:
@halocline I have made the changes as you specified.  But, there is one thing I am concerned in regards to the UX. Currently, the `color` prop that we have added works for the `lower` track of the rangeInput. But if the user wants to specify the color for the `upper` track then they have to add that through the theme. Do you think if its a good idea to have a prop for the `upper` track of rangeInput ? Please, let me know your thoughts.

halocline:
> @halocline I have made the changes as you specified. But, there is one thing I am concerned in regards to the UX. Currently, the `color` prop that we have added works for the `lower` track of the rangeInput. But if the user wants to specify the color for the `upper` track then they have to add that through the theme. Do you think if its a good idea to have a prop for the `upper` track of rangeInput ? Please, let me know your thoughts.

I like the question and love the UX focus, @g4rry420!!!  I was wondering the same the other day...

My inclination is that the "upper" is the negative space and thus shouldn't vary in appearance and meaning, which would make it suitable for the theme. That said, if that inclination is short-sighted, how might we alter the API surface to deliver "ease" and communicate "intuitiveness" to the developer, while also providing the flexibility for "upper" and "lower"?

g4rry420:
@halocline I agree with your views. I don't see any usecases where the `upper` track of color is varying differently. Lets keep it the way it is now.

halocline (commented):
Two items I see which we need to resolve:

1. We need to account for a case where `theme.rangeInput.track.upper` AND `theme.rangeInput.track.upper` are both `undefined`. An easy way to demo this is to modify the `Single.js` story by removing the custom theme entirely. In this case, the `color` prop is not being honored. 

In this scenario, these [values are undefined](https://github.com/grommet/grommet/blob/9736b2feaae7330e5a10bdb22a9b242a3add5b5c/src/js/themes/base.js#L1261), and [this block](https://github.com/grommet/grommet/blob/9736b2feaae7330e5a10bdb22a9b242a3add5b5c/src/js/components/RangeInput/StyledRangeInput.js#L41) ends up determining the track color even though `color` is specified.

Thinking about the correct approach here. We could put in an additional check `&& !props.color` but I believe this will not apply the proper upper track color without some 
additional refactoring. ...

2. `color` does not current accept valid HTML color names. Explanation [detailed here](https://github.com/grommet/grommet/pull/5275#issuecomment-907655806). I think we should fix this prior to merging.

g4rry420:
I think I have found the solution to the problem where `color` can also accept HTML color names.  Check the `utils/colors.js` file and `Single.js` story. In my opinion, this will works well with backwards compatible. Please, let me know your thoughts on this.  Reference - [https://stackoverflow.com/questions/1573053/javascript-function-to-convert-color-names-to-hex-codes/24390910](https://stackoverflow.com/questions/1573053/javascript-function-to-convert-color-names-to-hex-codes/24390910)

halocline:
> I think I have found the solution to the problem where `color` can also accept HTML color names. Check the `utils/colors.js` file and `Single.js` story. In my opinion, this will works well with backwards compatible. Please, let me know your thoughts on this. Reference - https://stackoverflow.com/questions/1573053/javascript-function-to-convert-color-names-to-hex-codes/24390910

Thank you @g4rry420! I'll check it out!

halocline:
There were a couple of unintended RangeInput snapshot changes in the most recent commit. I will correct those on Monday.

Other than that, the previous 2 outstanding issues have now been addressed.

src/js/components/RangeInput/index.d.ts
```diff
@@ -1,13 +1,20 @@
 import * as React from 'react';
 import { A11yTitleType } from '../../utils';
 
+interface ColorI {
```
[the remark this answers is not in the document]
taysea: Wondering if for clarity, we just call this `ColorInterface`?

src/js/components/RangeInput/StyledRangeInput.js
```diff
@@ -51,21 +56,78 @@ const trackColorStyle = (props) => {
...
+    : getRGBA(
+        normalizeColor(props.theme.global.colors.control, props.theme),
+        props.theme.rangeInput.track.opacity || 1,
+      );
+
+  if (
+    typeof props.color === 'string' ||
+    (typeof props.color === 'object' && !Array.isArray(props.color))
+  ) {
+    // lowerTrackColor = getRGBA(normalizeColor(props.color, props.theme), 1);
```
britt6612, line 88: can we delete this line ?

halocline: done

britt6612: thanks Im testing everything else out now!

britt6612 (approved):
Looks Good well just need to update this PR 
https://github.com/grommet/grommet-site/pull/304/files

jcfilben (approved):

ericsoderberghp (approved):

src/js/components/RangeInput/StyledRangeInput.js
```diff
@@ -51,21 +56,77 @@ const trackColorStyle = (props) => {
...
+          normalizeColor(color, props.theme),
+          opacity || 1,
+        )} ${valuePercentage}%,`;
+      } else {
+        result += `${getRGBA(
+          normalizeColor(color, props.theme),
+          opacity || 1,
+        )} ${thumbPosition},`;
+        result += `${upperTrackColor} ${thumbPosition}, ${upperTrackColor})`;
+        break;
```
robustdj, line 122: Would it be possible to make this part configurable? I have a `RangeInput` such that I don't want the different sections of the slider to change color depending on the value. I always want the colors to be there regardless of the value

PR #5278, opened 2021-05-22
Changed Grommet to add full='min'

ericsoderberghp:
#### What does this PR do?

Changed Grommet to add full='min'.

I've seen `<Grommet style={{ minHeight: '100vh' }} />` so much that it seemed a good idea to build this into grommet as `<Grommet full="min" />`.

doc.js

Added unit test

automatic

yes

backwards compatible


halocline:
> I've seen `<Grommet style={{ minHeight: '100vh' }} />` so much that it seemed a good idea to build this into grommet as `<Grommet full="min" />`.
> 

I think this approach is definitely useful, if not the default for which many will reach.

Typically, what I see happening is folks running into is applying `full` --> which then introduces a horizontal scroll (if the user's OS shows scrollbars by default and/or is enabled) --> then they add something like a `width: 'auto'` to compensate which can introduce double scroll bars once the content grows --> then leads to ... and so on. This will eliminate that "whack-a-mole" experience.

One other related item for which I was trying to come up with an elegant solution, is related to implementing the Global Header/Footer Service.  The `full='min'` / `min-height: 100vh` accounts for the header/footer service, but results in excess height (e.g. 100vh + header height + footer height). Ideally, the heights of the header/footer would be accounted for, but as a third-party service injected at run-time I was trying to think of a way to elegantly handle...

Example:
Screen Shot 2021-05-24 at 2 24 35 PM
Screen Shot 2021-05-24 at 2 24 44 PM

halocline (approved):

ShimiSun (approved):

PR #5284, opened 2021-05-26
Fix keyboard behavior of RadioButtonGroup

taysea:
#### What does this PR do?
This PR fixes the keyboard behavior for RadioButtonGroup and ensures that the onChange event is always called when the checked radio button changes.

The keyboard event handlers were adjusted to ensure that the native `onChange` event is always called, since it is called every time a user clicks with mouse. Previously, the event was not always called with keyboard navigation.

When in a FormField, the formfield receives focus indicator and the radiobuttons receive hover treatment so it is clear which is active.

In order for the interaction to work correctly on Firefox, a tabindex is toggled between the active radiobutton vs the others to ensure that pressing tab will leave the radiobuttongroup as opposed to navigating between them.

The expected behavior is that (view WCAG demo: https://www.w3.org/TR/2017/WD-wai-aria-practices-1.1-20170628/examples/radio/radio-1/radio-1.html)
- When the RadioButtonGroup receives focus, if nothing has been selected, the first radiobutton should receive focus.
- Pressing space would select the first radiobutton, pressing down arrow should select the second option.
- Using up/now arrows should navigate within the radiobuttons
- Pressing tab should leave the radiobuttongroup

src/js/components/RadioButtonGroup/RadioButtonGroup.js

Tested on Safari, Chrome, and Firefox on Mac with mouse and keyboard.

Navigate through radiobuttons in both the standalone radiobuttongroup and in a form:

Closes https://github.com/grommet/hpe-design-system/issues/899

No.

Fixed RadioButtonGroup keyboard interaction to align with WCAG standards.

Backwards compatible.

### DIFF
--- a/src/js/components/Form/__tests__/__snapshots__/Form-test-uncontrolled.js.snap
+++ b/src/js/components/Form/__tests__/__snapshots__/Form-test-uncontrolled.js.snap
@@ -3022,7 +3022,7 @@ exports[`Form uncontrolled reset clears select, checkbox, radiobuttongroup 2`] =
           role="radiogroup"
         >
           <label
-            class="StyledRadioButton__StyledRadioButtonContainer-g1f6ld-0 hnWETW"
+            class="StyledRadioButton__StyledRadioButtonContainer-g1f6ld-0 hHFdON"
             for="test-radiobuttongroup-one"
           >
             <div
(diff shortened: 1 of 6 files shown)


ericsoderberghp (commented):
Can we add a comment where we define `focus` in RadioButtonGroup where we explain how the focus processing works? Particularly, that it is not strictly aligned to the `activeElement` in the DOM but combines that, when `focusIndicator === true` but when `focusIndicator === false`, it indicates that we apply the active/hover styling to it.

taysea:
> Can we add a comment where we define `focus` in RadioButtonGroup where we explain how the focus processing works? Particularly, that it is not strictly aligned to the `activeElement` in the DOM but combines that, when `focusIndicator === true` but when `focusIndicator === false`, it indicates that we apply the active/hover styling to it.

The focus does strictly align with `activeElement` in the DOM (in that the radiobutton with focus is active DOM element). However, when focusIndicator is set to false we apply to hover treatment to give visual indication to keyboard users. I will add more comments to clarify this.

ericsoderberghp (approved):

PR #5285, opened 2021-05-27
Changed MaskedInput to add textAlign

ericsoderberghp:
#### What does this PR do?

Changed MaskedInput to add textAlign. This is the same property as with TextInput.

Also fixed TextInput to style textAlign before `extend`.

doc.js

added a unit test
tested with grommet-designer

automatic

yes

backwards compatible

### DIFF
--- a/src/js/components/TextInput/StyledTextInput.js
+++ b/src/js/components/TextInput/StyledTextInput.js
@@ -32,8 +32,8 @@ const StyledTextInput = styled.input`
     disabledStyle(
       props.theme.textInput.disabled && props.theme.textInput.disabled.opacity,
     )}
-  ${props => props.theme.textInput && props.theme.textInput.extend};
   ${props => props.textAlign && textAlignStyle}
+  ${props => props.theme.textInput && props.theme.textInput.extend};
 `;
 
 StyledTextInput.defaultProps = {};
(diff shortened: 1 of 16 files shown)


src/js/components/TextInput/StyledTextInput.js
```diff
@@ -32,8 +32,8 @@ const StyledTextInput = styled.input`
     disabledStyle(
       props.theme.textInput.disabled && props.theme.textInput.disabled.opacity,
     )}
-  ${props => props.theme.textInput && props.theme.textInput.extend};
   ${props => props.textAlign && textAlignStyle}
+  ${props => props.theme.textInput && props.theme.textInput.extend};
```
ShimiSun, line 36: Switching between those lines (35 + 36) might cause a backward compatibility issue.
Regardless, I thought that component props should suppress the theme props, so it would make sense that `props.textAlign` will suppress `props.theme.textInput` IMO.

ShimiSun (commented):
Dropped one comment, otherwise looks good.

src/js/components/TextInput/StyledTextInput.js
ericsoderberghp: I think `extend` should always come last. If you look at every other usage in Styled*.js in grommet, it comes last. This was the one place it didn't, which didn't look right.

taysea (commented):
All looks good here as long as we are okay with the TextInput change that Shimi noted. Agreed that `extend` should be the last item in the Styled* file

ShimiSun (approved):

PR #5288, opened 2021-05-27
Remove duplicate focus styling on CheckBox/CheckBoxGroup inside FormField

taysea:
#### What does this PR do?
When CheckBox or CheckBoxGroup is within a FormField, the focus styling should appear on the formfield. When rendered outside of a FormField, the focus styling should appear on the CheckBox itself. However, a keyboard user still needs to be aware that an element is active inside the FormField, especially in a CheckBoxGroup. When in a FormField, the formfield receives focus indicator and the CheckBoxes receive hover treatment so it is clear which is active.

Additionally, the theme was enhanced to support `checkBox.hover.background.color` which mirrors the behavior of `radioButton.hover.background.color`. Once merged, we will leverage this new theme object in grommet-theme-hpe as opposed to `extend` like we have to do now and will remove where we manually remove box-shadow from checkbox (which is causing issues for the stand alone checkbox outside of form field.)

This leverages similar logic to what was recently applied to RadioButton Group: https://github.com/grommet/grommet/pull/5284.

There has been discussion about if arrow keys should be able to navigate within the CheckBoxGroup as opposed to tab. However, this PR does not address any changes to keyboard behaviors. It purely address the styling that should occur when a checkbox or Checkboxgroup is contained in a FormField. That being said, if keyboard interaction is adjusted in the future, this styling would still be able to be leveraged there.

src/js/components/CheckBoxGroup/CheckBoxGroup.js

Test in storybook Custom Form story, standalone checkbox/checkboxgroup stories, and checkboxes in DataTable.

Same as above.

Closes https://github.com/grommet/grommet-theme-hpe/issues/151

Yes, updated here.

Fixed duplicate focus styling that occurred when CheckBox/CheckBoxGroup was contained in FormField.

Backwards compatible.

### DIFF
--- a/src/js/components/CheckBox/CheckBox.js
+++ b/src/js/components/CheckBox/CheckBox.js
@@ -35,6 +35,7 @@ const CheckBox = forwardRef(
       disabled,
       fill,
       focus: focusProp,
+      focusIndicator = true,
       id,
       label,
       name,
@@ -83,6 +84,12 @@ const CheckBox = forwardRef(
       checked,
       disabled,
       focus,
+      // when contained in a FormField, focusIndicator = false,
+      // so that the FormField has focus style. However, we still
+      // need to visually indicate when a CheckBox is active.
+      // If focus = true but focusIndicator = false,
+      // we will apply the hover treament.
+      focusIndicator,
       reverse,
       toggle,
       indeterminate,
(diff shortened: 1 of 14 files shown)


ericsoderberghp (approved):

halocline:
👀

halocline (approved):
Looks good.

[This story](http://localhost:9001/?path=/story/input-checkbox-inside-a-formfield--inside-form-field) briefly threw me off because the field + checkbox are receiving dual focus. However, I noticed that in this example, the direct child of the FormField in a Box instead of an input. Wondering if this story should be updated?

PR #5289, opened 2021-05-28
Webpack v5 upgrade

### LINKED ISSUE #5250: Upgrade webpack devDependencies to ^5.37
### Expected Behavior
webpack devDependencies should point to the latest ^5.37

### Actual Behavior
```
"webpack": "^4.39.3",
"webpack-cli": "^3.2.3"
```

### Steps to Reproduce

1. Go to grommet/package.json
2. Upgrade webpack

### Your Environment

- Grommet version: 2.17.2

jcfilben:
#### What does this PR do?
Upgrades Webpack from v4 to v5

Summary of changes + explaination:
- Bumped Node version from 10 to 12. ts-loader v9 requires at least v12 of Node
- Added `@storybook/builder-webpack5` and `@storybook/manager-webpack5"` so that storybook works with webpack 5. More info [here](https://storybook.js.org/blog/storybook-for-webpack-5/).
- Removed `clean-webpack-plugin`. It is now built in to Webpack and is used by specifying `output: { clean: true }` (line 26 of `webpack.config.babel.js`)
- Upgraded `ts-loader` to v9 to be compatible with Webpack 5

package.json

yarn test, yarn build, yarn storybook

^ above commands

Closes #5250

No

No

Backwards compatible


jcfilben:
The netlify deployment is failing because it is using "node v10.24.1" and this PR changes the required node version to >= 12. I am wondering if I should open a separate PR that bumps the node version?

src/js/components/InfiniteScroll/InfiniteScroll.js
```diff
@@ -177,7 +177,7 @@ const InfiniteScroll = ({
       }
     }, 100);
     return () => clearTimeout(timer);
-  }, [renderPageBounds, show, step]);
+  }, [renderPageBounds, show, step, scrollShow]);
```
ericsoderberghp, line 180: I'm concerned that will have unintended side effects, as this effect calls `setScrollShow`, which will cause `scrollShow` to change. Since this is already a warning today. I think we should decouple this from the webpack upgrade pull request, so that it can get more attention and not hold the upgrade up.

ericsoderberghp:
> The netlify deployment is failing because it is using "node v10.24.1" and this PR changes the required node version to >= 12. I am wondering if I should open a separate PR that bumps the node version?

I believe this is controlled by an environment variable in the Netlify configuration. So, I don't think we need a pull request for it. Do you believe our current master branch will work with Node 12 as is? If so, we can bump Netlify to version 12 first, now.

jcfilben:
Ok I see, I think node v12 will work fine with the master branch as is. I have already been using v12 locally an have not had any problems

src/js/components/InfiniteScroll/InfiniteScroll.js
jcfilben: Ok makes sense I removed this change in the latest commit

jcfilben:
Looks like other people are experiencing the same failure with build-storybook that we are getting in the netlify build 
```
/opt/build/repo/node_modules/webpack/lib/util/MapHelpers.js:17
const value = map.get(key);
TypeError: Cannot read property 'get' of undefined
```
https://github.com/storybookjs/storybook/issues/13893
I am trying to work through the suggestions in this thread but so far not having much luck. I have a few more things to try but I think if I can't get the suggestions in the thread to fix the issue then we should wait to do the upgrade until storybook officially supports webpack 5

ShimiSun:
I just checked and Netlify is using Node Version 14 
image

Is it possible the node version you are referring to is a definition that needs to be adjusted on the circleci yaml file?

jcfilben:
In the circleci config.yml file node is already set to version 14 
`docker: image: circleci/node:14`

ShimiSun:
Hi @jcfilben ! Nice progress!

I think we should link the existing storybook issue (of not being compatible with webpack5) to this PR, and mention it as a blocker for now, and track its progress with this PR.
Until storybook will solve this issue, I think we should split this PR into commits that can go on stable and help with the preliminary work of upgrading to webpack5. So any work that is done on this PR, and could also work with our current webpack version could be split to separate PR and merge separately until we finalize the work for the actual webpack5 upgrade. 
Let me know what you think.

jcfilben:
@ShimiSun I agree, I will link the storybook issue to the webpack issue on grommet and go ahead and close this PR. I will open a new PR(s) for and updates that can be done before the webpack 5 upgrade.

src/js/components/InfiniteScroll/InfiniteScroll.js
ShimiSun: I've created a separate issue for this warning as it seems to come back on other PR's as well.

Issue #5197, opened 2021-04-28
Testing - Refactor the usage of `renderer.create` to `render`


ShimiSun:
### Expected Behavior
In our grommet testing, we shouldn't use `renderer.create` from 
```
import renderer from 'react-test-renderer';
```
Instead, we should use `render` by 
```
import { render } from '@testing-library/react';
```

### Actual Behavior
Some tests are still using the legacy API.

### Steps to Reproduce

1. Go to tests that use the older API
2. Replace it with the better practice API suggested above.

### Your Environment

- Grommet version: stable
- Browser Name and version:N/A
- Operating System and version (desktop or mobile): N/A

g4rry420:
I am working on it.

g4rry420:
Shall I also remove package `react-test-renderer` ?

g4rry420:
Some of the test are using utility functions such as `findAllByType`. I believe those tests should remain the same. Because if I change them then I also have to change the utility functions, which might be used in other areas of repository. Need your inputs.

g4rry420:
I have made the changes but the pr is a big one. Not sure if I should still make request for it or not ?

PR #5291, opened 2021-05-29
Refactor Button tests to use `render` @testing-library function

Lukazovic:
#### What does this PR do?

`src/js/components/Button/__tests__/Button-test.js`

`yarn test`

Run: `yarn test`

No.

No.

Backwards compatible.

### DIFF
--- a/src/js/components/Button/__tests__/Button-test.js
+++ b/src/js/components/Button/__tests__/Button-test.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import renderer from 'react-test-renderer';
 
 import 'jest-styled-components';
 import 'jest-axe/extend-expect';
@@ -8,7 +7,6 @@ import 'regenerator-runtime/runtime';
 import { axe } from 'jest-axe';
 import { Add, Next } from 'grommet-icons';
 import { cleanup, fireEvent, render } from '@testing-library/react';
-import { findAllByType } from '../../../utils';
 import { Grommet, Button, Text } from '../..';
 
 describe('Button', () => {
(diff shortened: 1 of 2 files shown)


ShimiSun (approved):

src/js/components/Button/__tests__/Button-test.js
```diff
@@ -116,41 +113,39 @@ describe('Button', () => {
...
   test('fill', () => {
-    const component = renderer.create(
+    const { container } = render(
       <Grommet>
-        <Button>
-          <Button fill />
-          <Button fill={false} />
-          <Button fill="horizontal" />
-          <Button fill="vertical" />
-        </Button>
```
jcfilben, line 149: Good catch here, I agree that it doesn't make sense for the 4 Buttons to be wrapped within a Button for this test. I think this is a good change that simplifies the test.

jcfilben (approved):
Looks good! All the changes look right to me

src/js/components/Button/__tests__/Button-test.js
Lukazovic: Thanks!

I've noticed there are some other Button tests we can improve but I tried not to do it in the same PR to not make it hard to understand.

jcfilben: Yeah good call not doing those other changes in this PR, but if you think some of the other tests could be improved definitely feel free to file a separate PR with those changes

Lukazovic: Cool, I'll do it soon

ShimiSun:

PR #5296, opened 2021-06-01
Added selected styling to custom child in Select

### LINKED ISSUE #5154: "Selected" style is not applied to Select options when using custom children
Hello! I'm using the Select input and passing in a custom `children` argument to render a custom label. When I render the custom label, the "hover" and "active" styles from the Select input get applied automatically, so I get the theme's hover/active background/color for free, which is great; but the "selected" style from Select does not seem to get applied to my custom label.

### Expected Behavior

I think the selected style should be applied to the custom option, like the active and hover styles are. (that is, if i'm reading correctly, i think this [OptionBox](https://github.com/grommet/grommet/blob/master/src/js/components/Select/SelectContainer.js#L34), or at least its styles, should get applied to custom children instead of just to default children.)

### Actual Behavior

The active and hover styles are applied to custom options, but the selected style is not.

I know that `selected` is passed as a prop to the children function, so I can use that to apply a style to my custom option; but I feel like since the active and hover styles come for free, it would be nice and consistent if the selected one did as well.

https://codesandbox.io/s/grommet-v2-template-forked-tw8t4?file=/index.js

### Steps to Reproduce

There are 2 Selects here; the first one does not use custom children and the second one does. 
1. In both Selects, hover and active styles work identically
2. In the Select on the left, selected style works as expected
2. In the Select on the right, selecting an option does not apply the selected style to that option

### Your Environment

- Grommet version: 2.16.2
- Operating System and version (desktop or mobile): osx 11.2.1

jcfilben:
#### What does this PR do?
When you have custom children in Select the hover styling is applied but the selected styling is not applied. This PR adds the selected styling to custom children when they are selected.

SelectContainer.js

Tested with this storybook story
```javascript
import React from 'react';

import { Box, Grommet, Select, Text, Paragraph } from 'grommet';

export const TEST = () => {
  return (
        <Box gap="small" width="small">
            <Select options={options} />
            <Select options={options}>{(option) => <Text>{option}</Text>}</Select>
            <Select options={options}>{(option) => option}</Select>
            <Select options={options}>{(option) => <Paragraph>{option}</Paragraph>}</Select>
        </Box>
    </Grommet>
  );
};

export default {
  title: 'Input/Select/TEST',
};
```

See above

Closes #5154 

No

Yes

Backwards compatible

### DIFF
--- a/src/js/components/Select/SelectContainer.js
+++ b/src/js/components/Select/SelectContainer.js
@@ -382,11 +382,15 @@ const SelectContainer = forwardRef(
                   // as an option Button kind property.
                   let child;
                   if (children)
-                    child = children(option, index, options, {
-                      active: optionActive,
-                      disabled: optionDisabled,
-                      selected: optionSelected,
-                    });
+                    child = (
+                      <OptionBox selected={optionSelected}>
+                        {children(option, index, options, {
+                          active: optionActive,
+                          disabled: optionDisabled,
+                          selected: optionSelected,
+                        })}
+                      </OptionBox>
+                    );
                   else if (theme.select.options)
                     child = (
                       <OptionBox
@@ -398,6 +402,7 @@ const SelectContainer = forwardRef(
                         </Text>
                       </OptionBox>
                     );
+
                   // if we have a child, turn on plain, and hoverIndicator
 
                   return (
--- a/src/js/components/Select/SelectContainer.js
+++ b/src/js/components/Select/SelectContainer.js
@@ -30,11 +30,8 @@ const OptionsBox = styled.div`
   outline: none;
 `;
 
-const OptionBox = styled(Box)`
-  ${props => props.selected && selectedStyle}
-`;
-
 const SelectOption = styled(Button)`
+  ${props => props.selected && props.textComponent && selectedStyle}
   display: block;
   width: 100%;
 `;
@@ -381,27 +378,30 @@ const SelectContainer = forwardRef(
                   // Determine whether the label is done as a child or
                   // as an option Button kind property.
                   let child;
-                  if (children)
+                  let textComponent = false;
+                  if (children) {
+                    child = children(option, index, options, {
+                      active: optionActive,
+                      disabled: optionDisabled,
+                      selected: optionSelected,
+                    });
+                    if (
+                      typeof child === 'string' ||
+                      (child.props &&
+                        child.props.children &&
+                        typeof child.props.children === 'string')
+                    )
+                      textComponent = true;
+                  } else if (theme.select.options) {
                     child = (
-                      <OptionBox selected={optionSelected}>
-                        {children(option, index, options, {
-                          active: optionActive,
-                          disabled: optionDisabled,
-                          selected: optionSelected,
-                        })}
-                      </OptionBox>
-                    );
-                  else if (theme.select.options)
-                    child = (
-                      <OptionBox
-                        {...selectOptionsStyle}
-                        selected={optionSelected}
-                      >
+                      <Box {...selectOptionsStyle}>
                         <Text {...theme.select.options.text}>
                           {optionLabel(index)}
                         </Text>
-                      </OptionBox>
+                      </Box>
                     );
+                    textComponent = true;
+                  }
 
                   // if we have a child, turn on plain, and hoverIndicator
 
@@ -427,6 +427,7 @@ const SelectContainer = forwardRef(
                       onClick={
                         !optionDisabled ? selectOption(index) : undefined
                       }
+                      textComponent={textComponent}
                     >
                       {child}
                     </SelectOption>
@@ -442,11 +443,11 @@ const SelectContainer = forwardRef(
                 disabled
                 option={emptySearchMessage}
               >
-                <OptionBox {...selectOptionsStyle}>
+                <Box {...selectOptionsStyle}>
                   <Text {...theme.select.container.text}>
                     {emptySearchMessage}
                   </Text>
-                </OptionBox>
+                </Box>
               </SelectOption>
             )}
           </OptionsBox>
--- a/src/js/components/Button/StyledButton.js
+++ b/src/js/components/Button/StyledButton.js
@@ -199,7 +199,7 @@ const StyledButton = styled.button`
   ${props => !props.plain && basicStyle(props)}
   ${props => props.primary && primaryStyle(props)}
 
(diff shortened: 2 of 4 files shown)


src/js/components/Select/SelectContainer.js
```diff
@@ -382,11 +382,15 @@ const SelectContainer = forwardRef(
                   // as an option Button kind property.
                   let child;
                   if (children)
-                    child = children(option, index, options, {
-                      active: optionActive,
-                      disabled: optionDisabled,
-                      selected: optionSelected,
-                    });
+                    child = (
+                      <OptionBox selected={optionSelected}>
```
ericsoderberghp, line 386: Should we also be passing `{...selectOptionsStyle}`?

src/js/components/Select/SelectContainer.js
```diff
@@ -382,11 +382,15 @@ const SelectContainer = forwardRef(
...
+                        {children(option, index, options, {
+                          active: optionActive,
+                          disabled: optionDisabled,
+                          selected: optionSelected,
+                        })}
+                      </OptionBox>
+                    );
                   else if (theme.select.options)
                     child = (
                       <OptionBox
```
ericsoderberghp, line 396: Wondering if we should reduce duplication and do `{child && <OptionBox ...>{child}</OptionBox>}` inside `<SelectOption />`?

src/js/components/Select/SelectContainer.js
```diff
@@ -398,6 +402,7 @@ const SelectContainer = forwardRef(
                         </Text>
                       </OptionBox>
                     );
+
                   // if we have a child, turn on plain, and hoverIndicator
 
                   return (
```
ericsoderberghp, line 408: Would it make any sense to add `selected` as a prop to `Button` and then do that styling, in the same we do `active` styling? I'm not sure, so just wondering.

jcfilben: If the styling for a selected option is done in Button it changes the behavior when a selected option is hovered over the background color changes to the `active` color. Current behavior for Select is that when a selected option is hovered over the background color doesn't change.

src/js/components/Select/SelectContainer.js
jcfilben: I don't think we should pass `{...selectOptionsStyle}` because this adds padding to each option and if the user is passing a custom child I don't think grommet should automatically add padding on it.

src/js/components/Select/SelectContainer.js
jcfilben: Since we are passing different props `selected={optionSelected}` in the first case and `selected={optionSelected}` and `{...selectOptionsStyle}` in the second case is it possible to reduce duplication?

ericsoderberghp: not in a significantly cleaner way

src/js/components/Select/SelectContainer.js
ericsoderberghp: I'm wondering if we should reconsider that Button behavior. Is there anywhere else where we use `<Button selected />`? I noticed that there is currently no `selected` property documented for Button. I was wondering if we should add it, and have its behavior be what you mention, that hover doesn't affect its styling.

src/js/components/Select/SelectContainer.js
ericsoderberghp: Hmm... tricky to know what the theme might have in `theme.select.options.container`. Agreed on padding being a problem. Not so sure though if it has border or background. I think if we add `selected` to `Button`, this aspect isn't an issue.

src/js/components/Select/SelectContainer.js
jcfilben: I didn't find other places where `<Button selected />` is used. The places where I thought it might come up (like on Calendar) are not doing the selected styling on the button, similar to how Select is currently set up.

It makes sense to me to change the selected behavior on button so that the hover isn't changing the background. I can go ahead and do that and add the selected prop to the documentation.

src/js/components/Select/SelectContainer.js
jcfilben: `theme.select.options.container` has the following properties set: 
```
options: {
        container: {
          align: 'start',
          pad: 'small',
        },
```
But as you said, if we are doing the selected styling in the Button then this won't matter

ericsoderberghp: I was thinking we just add `${props => props.selected && selectedStyle}` to Button. That would leave the select options container styling to always be supplied unless the caller passes a children render function, keeping the `plain` nature of that path.

src/js/components/Select/SelectContainer.js
ericsoderberghp: OK. I think we might consider not adding it to the documentation to start with, and just consider this an internal usage for now. Then, if demand arises, we can add it. I say this in the spirit of not growing our published API surface unless there are concrete needs for it.

jcfilben:
I changed Button so that it is handling the selected styling. I realized that in storybook we have stories where when an option is selected and the selection is indicated in a different way than the background color. For example the Many Options story.

**Here is the Many Options story with `selected` styling being handled by Button:**
Screen Shot 2021-06-03 at 3 58 41 PM
**Here is the Many Options story with `selected` styling being handled by Select:**
Screen Shot 2021-06-03 at 3 59 02 PM

If button is handling the selected styling it will always force the background color to change

ericsoderberghp:
Agreed on wanting the children renderer to be able to control the select styling, as shown in the Many Options example with CheckBoxes. But, I see the same problem you showed for the Button approach with the approach in this PR: https://deploy-preview-5296--sad-tereshkova-173d07.netlify.app/?path=/story/input-select-many-options--many-options

I'm wondering if we should tweak the approach to say that if the children function returns a string, we will include the select styling, but otherwise we will leave it as is. I'm trying to think of another approach to allow both types of behavior, but haven't thought of something clean yet. Any suggestions?

jcfilben:
> I see the same problem you showed for the Button approach with the approach in this PR

Good catch I didn't notice that before. In the latest commit I updated Select to handle selected styling through Button and I fixed the issue with the Many Options story by only applying selected styling to custom children if they only contain text.

ericsoderberghp (approved):
Approved, but I think we should give this a little time on `stable` before releasing, so I'm delaying the merge until after the pending release.

PR #5302, opened 2021-06-03
Added List Editable story

ericsoderberghp:
Added List Editable story, to demonstrate an issue with List.

For now, not expecting to merge this, just building the story for filing an issue.


ericsoderberghp:
Figured out the problem. When passing `data={[str1, str2, ...]}`, List uses the strings as the underlying component `key`. When editing, the string changes, and thus the key changes, and thus react thinks it's a different thing and renders fresh elements. The solution is to convert it to something like `data={[{ id: 1, datum: str1 }, ...]} primaryKey="id"` so that changing the datum string doesn't affect the key and the TextInput maintains focus.

Issue #5306, opened 2021-06-04
DataTable sorting is not case-insensitive


samuelgoff:
If I have a dataset that includes mixed case (some ALL CAPS, some Initial Caps, some lowercase), the sorting process should ignore case.

### Expected Behavior

If I have a dataset that includes mixed case (some ALL CAPS, some Initial Caps, some lowercase), I expect “a” and “A” to appear before “z” and “Z”, regardless of case.

### Actual Behavior

If I have a dataset that includes mixed case (some ALL CAPS, some Initial Caps, some lowercase), the sorting process will put initial-capped items alphabetically, then lowercase items alphabetically. This makes sense if you’re a computer, not if you’re a human.

### URL, screen shot, or Codepen exhibiting the issue

https://jx4k9.sse.codesandbox.io/datatable
https://codesandbox.io/s/mutable-snowflake-jx4k9?file=/src/DataTable.js
![Screen Shot 2021-06-04 at 4 35 08 PM](https://user-images.githubusercontent.com/2007084/120866793-3f9c4f80-c556-11eb-924f-7ecc4b53dcb0.png)

### Steps to Reproduce

1. go to https://codesandbox.io/s/mutable-snowflake-jx4k9?file=/src/DataTable.js
2. click on `DataTable`
3. click on `Foods and Drinks` column header to sort alphabetically
4. scroll to bottom
5. note that "apple cider" and "vanilla wafers" appear **_after_** the list that includes (and ends with) "Vanilla Ice Cream"

### Your Environment

- Grommet version: 2.17.2
- Browser Name and version: Safari Version 14.1.1 (16611.2.7.1.4)
- Operating System and version (desktop or mobile): macOS Big Sur 11.4

PR #5307, opened 2021-06-05
DataTable case-insensitive sorting

samuelgoff:
Addresses #5306 : When `DataTable` is provided string data including mixed case (some ALL CAPS, some Initial Caps, and/or some lowercase), this improves sorting to ignore differences in capitalization.

#### What does this PR do?
- fixes a missing dependency in `InfiniteScroll.js` preventing me from committing/pushing any other changes
- improves `DataTable` sorting logic to ignore differences in capitalization when data being compared is string
- improve the test data included in the `DataTable` stories that feeds into the sorting story & proves that it works as expected

- `yarn storybook`
- `src/js/components/DataTable/buildState.js`

- `yarn test`
- `yarn storybook`

- run `yarn storybook`
- click `DataTable` under `visualizations`
- click on `Sort`
- note the default view applies a sort descending and both lowercase and InitialCapped names are properly included in descending sort order
- click on `Name with Extra` column header to toggle sort to ascending
- note the resulting view applies a sort ascending and both lowercase and InitialCapped names are properly included in ascending sort order

When mixed-case text is included in a list, users expect sorting to ignore differences in capitalization. Otherwise, the capitalized dataset gets alphabetized, followed by a lowercase dataset. This can be extremely confusing and frustrating to a user, since with larger mixed-case datasets, users often won't even realize there is an additional set of lowercase results which appear off-screen. The users' attention is focused on the (capitalized) alphabetical sequence, and when they look for "b" to appear after "A" but before "C" and don't find it, this is completely outside of their expectations.

**BEFORE:**
Screen Shot 2021-06-04 at 4 35 08 PM

**AFTER (ascending):**
Screen Shot 2021-06-05 at 2 00 43 PM

**AFTER (descending):**
Screen Shot 2021-06-05 at 2 00 49 PM

Nope.

I would recommend it.

backwards compatible


src/js/components/InfiniteScroll/InfiniteScroll.js
```diff
@@ -177,7 +177,7 @@ const InfiniteScroll = ({
       }
     }, 100);
     return () => clearTimeout(timer);
-  }, [renderPageBounds, show, step]);
+  }, [renderPageBounds, scrollShow, show, step]);
```
ShimiSun, line 180: Unless you are thinking this fix is important and related to this PR functionality, here is a quote from https://github.com/grommet/grommet/pull/5289#discussion_r641842233

>I'm concerned that will have unintended side effects, as this effect calls setScrollShow, which will cause scrollShow to change.

Also linking to https://github.com/grommet/grommet/issues/5317

samuelgoff: Hi @ShimiSun, this change is only related in that the pre-commit hook prevented me from pushing unless I either fixed the linting error or used the `--no-verify` flag while pushing. As your pre-commit hook specifically says, "This is ill-advised since the commit is broken." I'm happy to revert that change, if you prefer to address this by other means.

I've just reverted that change & pushed. It should be ready to review now, please.

src/js/components/DataTable/buildState.js
```diff
@@ -84,12 +84,34 @@ export const filterAndSortData = (data, filters, onSearch, sort) => {
...
-      if (datumValue(d1, property) > datumValue(d2, property)) return before;
-      if (datumValue(d1, property) < datumValue(d2, property)) return after;
-      return 0;
+      const d1Val = datumValue(d1, property);
+      const d2Val = datumValue(d2, property);
+      const d1IsString = typeof d1Val === 'string';
+      const d2IsString = typeof d2Val === 'string';
+      const d1ValProc = d1IsString ? d1Val.toLowerCase() : d1Val;
+      const d2ValProc = d2IsString ? d2Val.toLowerCase() : d2Val;
+      switch (true) {
```
MikeKingdom, line 97: this is a really odd use of switch to do multiple bits of logic.  We should look at a cleaner way to do this. This is very hard to figure out compared to the original.

src/js/components/DataTable/buildState.js
```diff
@@ -84,12 +84,34 @@ export const filterAndSortData = (data, filters, onSearch, sort) => {
...
+    const after = sortAsc ? -1 : 1;
     result.sort((d1, d2) => {
-      if (datumValue(d1, property) > datumValue(d2, property)) return before;
-      if (datumValue(d1, property) < datumValue(d2, property)) return after;
-      return 0;
+      const d1Val = datumValue(d1, property);
+      const d2Val = datumValue(d2, property);
+      const d1IsString = typeof d1Val === 'string';
+      const d2IsString = typeof d2Val === 'string';
+      const d1ValProc = d1IsString ? d1Val.toLowerCase() : d1Val;
```
MikeKingdom, line 95: why do you do toLowerCase when you're already going to use { sensitivity: 'base' } ?

src/js/components/DataTable/buildState.js
samuelgoff: Hey @MikeKingdom, the rule of thumb I was following is that when you have more than 3 cases in an if/then, a switch is generally considered preferable. That said, I see your point, and I realized there was a way to improve & simplify the logic a bit. I've pushed the change, so please review.

src/js/components/DataTable/buildState.js
samuelgoff: D'oh! Good call. Fixed.

src/js/components/DataTable/buildState.js
```diff
@@ -84,11 +84,26 @@ export const filterAndSortData = (data, filters, onSearch, sort) => {
...
+    const after = sortAsc ? -1 : 1;
     result.sort((d1, d2) => {
-      if (datumValue(d1, property) > datumValue(d2, property)) return before;
-      if (datumValue(d1, property) < datumValue(d2, property)) return after;
+      const d1Val = datumValue(d1, property);
+      const d2Val = datumValue(d2, property);
+      const d1IsString = typeof d1Val === 'string';
+      const d2IsString = typeof d2Val === 'string';
+      if (d1IsString && d2IsString) {
+        if (sortAsc) {
```
MikeKingdom, line 96: Thanks Samuel. This helps. I'd also consider simplifying to something like:
```es6
if (typeof d1Val === 'string' && typeof d2Val === 'string') {
  const result = d1Val.localeCompare(d2Val, undefined, { sensitivity: 'base' });
  return sortAsc ? result : -result;
}
```
It might be a little confusing but entertaining to do
```es6
return before * d1Val.localeCompare(d2Val, undefined, { sensitivity: 'base' });
```
:)

MikeKingdom (commented):
Thanks Samuel. Looks good to me. I'll others get the final word.

ericsoderberghp (dismissed):

src/js/components/DataTable/stories/data.js
```diff
@@ -70,6 +70,27 @@ for (let i = 0; i < 40; i += 1) {
 }
 
 export const DATA = [
+  {
+    name: 'zoey',
+    location: '',
```
ericsoderberghp, line 75: Is there a reason we're not filling out the other properties with more interesting values? I ask because the other stories sharing this data are looking a little off.

src/js/components/DataTable/buildState.js
samuelgoff: Hey @MikeKingdom, I'm good with the former suggestion, but I agree: the latter option might be a little confusing, so I'd prefer to avoid it. I've pushed the change. Thanks!

src/js/components/DataTable/stories/data.js
samuelgoff: Hi @ericsoderberghp, I think my point in adding these items to the data was to demonstrate case-insensitive sorting (or lack thereof). Prior to this, only initial-capped values were used, so this was a blind-spot. If the data is "perfect", then little bugs can creep in that don't become obvious until they encounter real-world edge cases. If'd you'd like to see more interesting values so that other stories look better, let's please make sure we don't eliminate some of the intentional flaws I introduced by this change.

ericsoderberghp: That's fine. In that case, maybe we should create a custom `data` value for this story.

samuelgoff: Hey @ericsoderberghp, okay. I've just pushed the change. Please review.

ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution :)

Issue #5266, opened 2021-05-18
Persist List/DataTable active row background color


saggiemilva:

ericsoderberghp:
One idea for this would be to preserve the active row across focus/blur but clear it when `data` changes.

PR #5308, opened 2021-06-07
Fix: Persisting Active Style on List and DataTable

jlaluces123:
Addresses #5266 : When selecting a row from a list or table, the active styling on the item remains when a user opens a modal / layer detail, but when closing the modal, the active styling disappears, preventing users from seeing where they last left off.

##### List Component:
- Persists the active styling during an `onBlur` and `onMouseOut` 
- Updates to the `onClickItem` story to match the scenario found in issue #5266
- `lastActive` pointer to keep track of the latest active item / row selected to persist active styling

##### DataTable Component:
- Updates to the `clickable` story to match scenario found in issue #5266
- In Progress: persisting the active style during `onBlur` and `onMouseLeave`

##### List Component:
1. In `components/List/List.js` check out the `clickProps` to see how the active styling is applied / persisting depending on the events
2. `yarn storybook`
3. Navigate to `List's onClickItem` story
4. Click on any list item, close the modal and watch how the active styling remains until you hover other list items

##### DataTable Component
1. In `components/DataTable/Body.js` check out the `clickProps (onMouseEnter, onFocus, [...])` to see how the active styling is applied / persisting depending on the events
2. `yarn storybook`
3. Navigate to `DataTable's Clickable` story
4. Work in Progress: click on any row, close the modal and watch how the active styling remains until you hover other list items

Manual testing via the onClickItem story.

1. `yarn storybook`
2. Navigate to `List's onClickItem` story
3. Click on any list item, close the modal and watch how the active styling remains until you hover other list items
4. Repeat Steps 2 and 3 for `DataTable's Clickable` story

- #5266

No.

No.


src/js/components/DataTable/DataTable.js
```diff
@@ -164,6 +164,7 @@ const DataTable = ({
   const footerRef = useRef();
   const [headerHeight, setHeaderHeight] = useState();
   const [footerHeight, setFooterHeight] = useState();
+  // const [lastActive, setLastActive] = useState();
```
ShimiSun, line 167: we try not to have code comments on our code base, to avoid clutter.

ShimiSun (commented):
Some of the auto spacing on this PR isn't aligned with our original code base guidelines, it will be helpful if the changed code won't show spacing and braces changes if not necessary.

ericsoderberghp:
I'm noticing a couple of things about the behavior in Storybook that don't look quite right:

List

I can tab to the List and see the focus outline
I can use the arrow keys to move the active item indicator up and down
I can use the enter key to show the layer
I can use the esc key to close the layer and the focus and active indicator look correct
But, if I now tab to remove focus from the List, the active list item indicator remains. I think the active item indicator shouldn't be shown if the List doesn't have focus.

DataTable

If I tab to the DataTable, I don't see the focus indicator, but I should.
I can use the arrow keys to move the active row indicator up and down.
I can use the enter key to show the modal
I can use the esc key to close the modal and the active indicator looks correct, but I still don't see the focus indicator.
If I tab away from the DataTable, the active row indicator remains. I think the active row indicator shouldn't be shown if the DataTable doesn't have focus.

src/js/components/DataTable/Body.js
```diff
@@ -44,6 +48,7 @@ const Row = memo(
       <StyledDataTableRow
         ref={rowRef}
         size={size}
+        tabIndex={onClickRow ? -1 : undefined}
```
ericsoderberghp, line 51: Why is this needed? I wasn't expecting us to be giving focus to the row.

src/js/components/DataTable/Body.js
```diff
@@ -218,10 +250,15 @@ const Body = forwardRef(
...
                 <Row
                   key={index}
+                  tableRef={ref}
+                  lastActive={lastActive}
+                  setLastActive={setLastActive}
                   rowRef={rowRef}
                   primaryValue={primaryValue}
                   isSelected={isSelected}
                   isRowExpanded={isRowExpanded}
+                  itemFocus={itemFocus}
```
ericsoderberghp, line 260: I didn't see where this was used by Row. Do we need to pass it here?

src/js/components/DataTable/StyledDataTable.js
```diff
@@ -103,12 +108,18 @@ const StyledDataTableBody = styled(TableBody)`
     overflow: auto;
   `}
 
-  &:focus {
-    ${focusStyle({ skipSvgChildren: true, forceOutline: true })}
-  }
+  ${props =>
+    props.itemFocus &&
```
ericsoderberghp, line 112: I don't think the StyledDataTableBody ever received `itemFocus` as a property.

src/js/components/DataTable/StyledDataTable.js
```diff
@@ -103,12 +108,18 @@ const StyledDataTableBody = styled(TableBody)`
     overflow: auto;
   `}
 
-  &:focus {
```
ericsoderberghp, line 106: Why did we not keep focus at the table body?

src/js/components/DataTable/Body.js
```diff
@@ -54,13 +59,37 @@ const Row = memo(
...
                 onClickRow(adjustedEvent);
+                // putting focus on tbody for WCAG guideline
+                tableRef.current.focus();
+              }
+            : undefined
+        }
+        onMouseEnter={
+          onClickRow
+            ? () => {
+                setLastActive(undefined);
```
ericsoderberghp, line 70: I'm thinking we might want to simplify this and put more logic in the Body than the Row, the Body is the one keeping track of the active state. For example, the Body can manage `lastActive` as it sees `active` change.

src/js/components/DataTable/StyledDataTable.js
```diff
@@ -83,6 +83,11 @@ const StyledDataTableRow = styled(TableRow)`
     `
     cursor: pointer;
   `}
+
+  &:focus {
```
ericsoderberghp, line 87: Just wondering why we are changing the pattern and putting focus on the row instead of the body.

src/js/components/List/List.js
```diff
@@ -232,10 +232,8 @@ const List = React.forwardRef(
             (onClickItem || onOrder) && data && data.length
               ? () => {
                   const min = onOrder ? 1 : 0;
-                  const max = onOrder ? (data.length * 2) - 2 : data.length - 1;
-                  setActive(
-                    active >= min ? Math.min(active + 1, max) : min,
-                  );
+                  const max = onOrder ? data.length * 2 - 2 : data.length - 1;
```
ericsoderberghp, line 235: Why are all of these syntax changes being included? They don't seem related to this active state change.

src/js/components/List/List.js
```diff
@@ -22,11 +22,13 @@ const StyledList = styled.ul`
   padding: 0;
   ${genericStyles}
 
+  // Customizes to make table have a focus border color of green
```
ericsoderberghp, line 25:
```suggestion
  // Customizes to make list have a focus border color of green
```

src/js/components/List/List.js
```diff
@@ -220,22 +222,23 @@ const List = React.forwardRef(
                 }
               : undefined
           }
+          onTab={() => setActive(undefined)}
```
ericsoderberghp, line 225: Why do we need the `onTab` handler? I would expect `onBlur` to be sufficient.

src/js/components/List/List.js
```diff
@@ -357,13 +366,19 @@ const List = React.forwardRef(
...
                       listRef.current.focus();
                     },
-                    onMouseOver: () => setActive(index),
-                    onMouseOut: () => setActive(undefined),
+                    onMouseOver: () => {
+                      setLastActive(undefined);
+                      setActive(index);
+                    },
+                    onMouseOut: () => {
+                      setActive(lastActive);
```
ericsoderberghp, line 374: This doesn't seem right. I would think we would leave `lastActive` alone, to be used in the next `onFocus`, but we wouldn't want to `setActive` when the mouse goes away.

src/js/components/List/List.js
```diff
@@ -357,13 +366,19 @@ const List = React.forwardRef(
...
+                    },
                     onFocus: () => {
                       setActive(index);
                       setItemFocus(true);
                     },
                     onBlur: () => {
+                      setLastActive(active);
```
ericsoderberghp, line 381: I would have expected `onFocus` to set both `active` and `lastActive` and `onBlur` to just clear `active`. Why are we setting last active at `onBlur` time?

ericsoderberghp (commented):
I've added some some comments to List for now. It looks like there are still unresolved comments in DataTable from prior reviews.

src/js/components/List/List.js
jlaluces123: I agree with this idea, moving the `lastActive` to `onFocus` makes more sense here

src/js/components/List/List.js
jlaluces123: From what I gathered, since `onBlur` is only part of the `clickProps` events, `onBlur` and `onFocus` never get triggered when doing keyboard navigation which is why I decided to use the `onTab` prop available for Keyboard as a way to remove the active indicator upon losing focus of the List or DataTable.

src/js/components/List/List.js
jlaluces123: As soon as the modal / Layer opens, `onMouseOut` is triggered and upon closing, I chose to `setActive`  to what `lastActive` had stored as the way to keep an item active when a modal / Layer closes.

src/js/components/DataTable/Body.js
jlaluces123: I played around with this and the tabIndex was needed in order to use the `onFocus` and `onBlur` properties but since we don't want to put focus on the row, I set it to `-1` to counteract that.

src/js/components/DataTable/Body.js
jlaluces123: Got rid of this here after trimming the unused code as it was incorrectly placed, this was supposed to be put on `StyledDataTableBody`

src/js/components/DataTable/StyledDataTable.js
jlaluces123: Resolved now, accidentally placed `itemFocus` on the Row component instead of the Table Body

src/js/components/DataTable/StyledDataTable.js
jlaluces123: This is back in place now but utilizing the same approach as List.js where it'll look like:

```js
&:focus {
  ${props => props.tabIndex >= 0 && focusStyle ...}
}
```

src/js/components/DataTable/StyledDataTable.js
jlaluces123: As part of the trimming down I've done, this will be removed as well. I believe where we check for `props.active` that mimics the same functionality.

ShimiSun:
Closing as this PR has an updated version #5371

Issue #5198, opened 2021-04-28
Investigate PR merge despite ESLint warnings


halocline:
### Expected Behavior

CI checks should catch any ESLint warnings which are present on a PR prior to merging.

### Actual Behavior

https://github.com/grommet/grommet/pull/5187 snuck by despite having ESLint warnings for missing useEffect dependencies.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

taysea:
Looking into our CI scripts. It looks like the lint command we run is lint (as opposed to lint-fix which would throw an error on any linting output). lint-fix should run in pre-commits but if a commit is pushed up with --no-verify we may miss this. Wondering if we should run `lint-fix` in CI instead.

taysea:
Bumping this as it occurred prior to the `v2.27.0` release.

jcfilben:
Confirm that eslint on circle ci is flagging warnings in addition to errors

PR #5309, opened 2021-06-08
Disabled missing dependency warning to InfiniteScroll useLayoutEffect

halocline:
#### What does this PR do?

~Resolves~ Disables a linter warning for missing dependency in InfiniteScroll useLayoutEffect dependency array and provides a descriptive comment.

Closes https://github.com/grommet/grommet/issues/5317

InfiniteScroll.js

Manual running of `yarn test`.

Also, test the story "Select/Many options" to test the keyboard scroll behavior addressed by this PR #5230.

Manual running of `yarn test`. Prior to this commit, a warning would be called out. After, warning has been resolved.

Also, test the story "Select/Many options" to test the keyboard scroll behavior addressed by this PR #5230.

Ran into lint warning while preparing for release. Would like to resolve warning for a clean release.

#5230

Related: https://github.com/grommet/grommet/issues/5198
https://github.com/grommet/grommet/issues/5317

No.

No.

Backwards compatible.


taysea (approved):

ericsoderberghp:
I have this vague recollection that adding that dependency caused some kind of unintended loop in some situation. I’m not remembering the details though. I’m wondering if it would be safer to release without adding it rather than add it without taking some time to understand why it was omitted.

halocline:
> I have this vague recollection that adding that dependency caused some kind of unintended loop in some situation. I’m not remembering the details though. I’m wondering if it would be safer to release without adding it rather than add it without taking some time to understand why it was omitted.

I will switch this to a draft PR until edge cases can be tested and/or the dependency can be ruled as being problematic. If the latter, this PR should be modified to include a comment explaining why the dependency is not included.

ShimiSun:
Related to #5317

src/js/components/InfiniteScroll/InfiniteScroll.js
```diff
@@ -177,7 +177,7 @@ const InfiniteScroll = ({
       }
     }, 100);
     return () => clearTimeout(timer);
-  }, [renderPageBounds, show, step]);
+  }, [renderPageBounds, show, step, scrollShow]);
```
ShimiSun, line 180: Why is this PR marked as draft?

halocline: Marked as draft per this discussion: https://github.com/grommet/grommet/pull/5309#issuecomment-857232042. There was a concern that including dependency could result in an infinite loop, so want to hold off merging until edge cases are investigated.

halocline: I believe the if statement on Line 160 accounts for and addresses the concern. I also have not been able to reproduce any cases invoking the infinite loop.

Regardless, if the decision is made to not include `scrollShow` as a dependency, then a comment should be added so that the rationale is documented.

halocline:
I was unable to create a situation where the infinite loop was invoked. However, because the dependency is being called within a timer makes the outcome non-deterministic. After reviewing with @MikeKingdom, thinking is best to not include the dependency and disable the warning.

MikeKingdom:
Basically agree with what Matt said. It's hard to tell what would happen and don't really want any extra renders. The way the timer is created unconditionally right now means you could end up with multiple timer events outstanding even if the conditions inside the function result in a noop. It's hard to tell. Also I was wondering if this is a candidate for the intersection observer API instead.

ericsoderberghp (approved):

ShimiSun (approved):

PR #5315, opened 2021-06-10
Fix Carousel showed slide when mounted

### LINKED ISSUE #5093: Carousel shows different image on mount

DolceTriade:
#### What does this PR do?
Provides a fix for #5093 by building on #5115, which seems abandoned.

Following the insights of the issue reporter, reversing the Carousel children fixes the behavior for some cases (initialChild = 0). However, passing any other value as initialChild results in the same problem.

The problem itself is related to children animations. That's how Carousel navigation is handled, by changing their visibility. This fix solves the problem by reordering the children based on which one is the activeIndex, and also changing z-Index and opacity for the other children.

src/js/components/Carousel/Carousel.js

Manual testing and the existing tests (that were updated)

```
const CarouselModal = ({ close }) => (
  <Layer onClickOutside={close}>
    <Carousel controls="arrows" initialChild={0}>
      <Image fit="cover" src="//v2.grommet.io/assets/Wilderpeople_Ricky.jpg" />
      <Image fit="cover" src="//v2.grommet.io/assets/IMG_4245.jpg" />
      <Image fit="cover" src="//v2.grommet.io/assets/IMG_4210.jpg" />
    </Carousel>
  </Layer>
);

export const Simple = () => {
  const [showModal, setShowModal] = useState(false);

  return (
    <Grommet theme={grommet}>
      <Button label="Show Modal Carousel" onClick={() => setShowModal(true)} />

      {showModal && <CarouselModal close={() => setShowModal(false)} />}
    </Grommet>
  );
};
```

Closes #5093

No.

Backwards compatible, but it might break some snapshots as children's order might differ a little bit.

### DIFF
--- a/src/js/components/Carousel/Carousel.js
+++ b/src/js/components/Carousel/Carousel.js
@@ -127,7 +127,8 @@ const Carousel = ({
   );
 
   const selectors = [];
-  const wrappedChildren = Children.map(children, (child, index) => {
+  const wrappedChildren = [];
+  Children.map(children, (child, index) => {
     selectors.push(
       <Button
         // eslint-disable-next-line react/no-array-index-key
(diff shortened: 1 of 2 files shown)


DolceTriade:
This doesn't work. nvm.

Issue #4068, opened 2020-05-11
Grommet Accessibility Audit


taysea:
# Purpose of this Issue
The purpose of this issue is to identify keyboard/ARIA accessibility improvements that can be made in Grommet to better align the library with [WCAG standards](https://www.w3.org/TR/WCAG21/). Grommet does a great job of building in accessibility, but there are some areas for improvement.

## What is accessibility?
Accessibility is the practice of making your websites usable by as many people as possible. Some users can navigate only via keyboard, or screen reader, etc. You might also think of accessibility as treating everyone the same, and giving them equal opportunities, no matter what their ability or circumstances.

Source: https://developer.mozilla.org/en-US/docs/Learn/Accessibility/What_is_accessibility

## How were these improvements identified?
Grommet components were assessed for their keyboard navigability and success with Voiceover. Their behavior was compared side by side with the recommendations from WCAG. Typically, searching for [component] + wcag keyboard returns links to the appropriate section of w3 site.

For example, searching range "input wcag keyboard" returns this helpful resource as a top result: https://www.w3.org/TR/wai-aria-practices/examples/slider/slider-1.html

The behavior of the above slider example (via mouse, keyboard only, and screen reader) can be compared side by side with the Grommet component to identify any discrepancies. Additionally, there is often direct guidance for what the keyboard behavior should be.

Additionally, for some FormField label things, the WAVE chrome browser plugin was used. This identifies certain [ARIA](https://www.w3.org/TR/wai-aria-practices/) and color contrast errors.

## Grommet Component Analysis Results
The following are the areas of improvement that need to be address and have tests written for.

### FormField
- Providing a label on a FormField and on its contained Checkbox leads to an orphaned label on the formfield.

### Meter
- [ ] Similar to Chart, screen reader users should be able to identify what's in the chart without seeing it.
- Currently, Voiceover doesn't have access to read chart text information; just read out as "image" with no content information
- Recommendation, some sort of aria attribute should explain the value.

ShimiSun:
Excellent analysis @taysea ! thank you for this contribution.
We will work on adding accessibility tests as well to enforce the best practices.

eddieshowcase:
I would like to add the this list of items that need some review or attention:

### Tabs

- Tab components should respond to use of arrow keys for navigation (not just rely on 'tab' key)
- https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-2/tabs.html

### Accordion

[...]

PR #5319, opened 2021-06-10
Fix/calendar-accessibility

amandadupell:
#### What does this PR do?
- added correct tabIndex to allow users with screen readers to tab between dates instead of reading aloud the entire calendar

- had to update tests because i added a aria-label

- test manually with the simple calendar story in a new tab and using VoiceOver screen reader

- in response to the issue on this [accessibility issue](https://github.com/grommet/grommet/issues/4068)


ericsoderberghp:
Does this cause the user to have to tab through all days in order to get past the Calendar? For instance, if I have a Form with an inline DateInput Calendar and the user tabs from the field before the Calendar into the Calendar, that they will have to tab through all dates to get to the next input in the Form? If so, I think we should consider approaches like recently done for RadioButtonGroup. I think @MikeKingdom has the details.

src/js/components/Calendar/Calendar.js
```diff
@@ -124,6 +124,8 @@ const CalendarDay = ({
   <StyledDayContainer sizeProp={size} fillContainer={fill}>
     <CalendarDayButton fill={fill} {...buttonProps}>
       <StyledDay
+        tabIndex={0}
+        aria-label={buttonProps.a11yTitle}
```
ShimiSun, line 128: We should make sure that if the user is trying to pass the common `aria-label` prop it will propagate as well.

MikeKingdom:
> Does this cause the user to have to tab through all days in order to get past the Calendar? For instance, if I have a Form with an inline DateInput Calendar and the user tabs from the field before the Calendar into the Calendar, that they will have to tab through all dates to get to the next input in the Form? If so, I think we should consider approaches like recently done for RadioButtonGroup. I think @MikeKingdom has the details.

The approach used in the RadioButtonGroup was to use role="group". PR #5209. This allows the screen reader to be a little more clear that the other components are part of that group. The group does need some form of label like aria-label or aria-labeledby="<some id of another element>"

For the List order controls (PR #5212 ) we set role="listbox" and aria-activedescendant="<id of active move control or row>" and the controls or list row are managed separate from focus. The list has the keyboard focus and the row or controls get marked active as the user uses arrow keys or clicks to move around. That way tab only moves focus in or out of the list rather than having to tab through each row or move control.

amandadupell:
this PR covers the following issues with calendar:
- on click for the arrow buttons will say 'moved to' and the month the calendar has moved focus to
- the calendar is now a grid/table and when a user gets to the table, it tells them the month that the table represents and how many rows and columns make up that month
- when a user moves around in the grid of dates, the date is read aloud to them

ericsoderberghp:
This has a lot more files changed in it probably should. I'd recommend connecting with Matt to get the prettier syntax changes in first in a separate pull request.

halocline:
> This has a lot more files changed in it probably should. I'd recommend connecting with Matt to get the prettier syntax changes in first in a separate pull request.

Agree. Prettier PR in progress. I will connect with @amandadupell to coordinate.

src/js/components/Calendar/__tests__/Calendar-test.js
```diff
@@ -303,7 +303,7 @@ describe('Calendar', () => {
       </Grommet>,
     );
     // select date greater than January 1st
-    fireEvent.click(getByLabelText('Fri Jan 03 2020'));
+    fireEvent.click(getAllByLabelText('Fri Jan 03 2020')[0]);
```
taysea, line 306: why do the tests need to use `getAllByLabelText` now?

src/js/components/Calendar/Calendar.js
taysea: agreed with this comment, unless the `aria-label` will already be placed naturally on line 125? with {...buttonProps}

MikeKingdom: The other question is how will a screenreader deal with a11yTitle being on both the CalendarDayButton and the StyledDay? Will it read it twice in some cases?

src/js/components/Calendar/__tests__/Calendar-test.js
amandadupell: this was an old change i had pushed up by accident, had it changed when i was using tab Index on the day

src/js/components/Calendar/Calendar.js
amandadupell: i've removed this from the StyledDay as i think taylor is right about {...buttonProps}; here is a screenreader demo with that approach: 

https://user-images.githubusercontent.com/38971752/124791561-294d2f00-df1a-11eb-8dc7-68458c7ad3a6.mov

ericsoderberghp (approved):
Beautiful code.

MikeKingdom (approved):
Thanks!

ericsoderberghp:
Thanks for your contribution!

PR #5327, opened 2021-06-16
Refactor: improve button tests

Lukazovic:
#### What does this PR do?

Improve `Button` tests.

I noticed almost every button test is a snapshot test, which is not wrong, but I think we can improve it. I mean, if I have to test the `fill` prop I can write tests to make sure the prop is being applied correctly, but with a snapshot, I have to check if the `width` and `height` are correct.

Another good example of improvement is the [`href` test](https://github.com/grommet/grommet/blob/master/src/js/components/Button/__tests__/Button-test.js#L233) which aims to ensure the button will be rendered as a link. But to make sure this is happening, nowadays we have to check if it is being rendered correctly on snapshots. We can write a test to make sure this button will be rendered as a link, and also add validation to make sure the `href` prop is being passed correctly, without even check the snapshot. And this is what I did.

`src/js/components/Button/__tests__/Button-test.js`

```bash
yarn test
# or
yarn jest src/js/components/Button/__tests__/Button-test.js
```

- `afterEach`

I removed the `afterEach(cleanup)` since this is applied automatically by `testing-library`.

> Please note that this is done automatically if the testing framework you're using supports the afterEach global and it is injected to your testing environment (like mocha, Jest, and Jasmine). If not, you will need to do manual cleanups after each test.

ref: https://testing-library.com/docs/react-testing-library/api/#cleanup

- `user-event`

I added `user-event` lib to improve some tests. This may be useful to a lot of components test.

> `user-event` is a companion library for Testing Library that provides more advanced simulation of browser interactions than the built-in `fireEvent` method.

ref: https://testing-library.com/docs/ecosystem-user-event/

No.

No.

No.


jcfilben (approved):

src/js/components/Button/__tests__/Button-test.js
```diff
@@ -3,39 +3,39 @@ import React from 'react';
...
   test('should have no accessibility violations', async () => {
-    const { container, getByText } = render(
+    const { container } = render(
       <Grommet>
         <Button a11yTitle="Test button" label="Test" onClick={() => {}} />
       </Grommet>,
     );
 
-    fireEvent.click(getByText('Test'));
+    fireEvent.click(screen.getByText('Test'));
```
ShimiSun, line 22: Can you please add a few words of why you think your enhancement is better than the previous approach?

Lukazovic: In this case, I don't think we'll have a difference.

Importing `screen` directly from the `'@testing-library/react'` will only make it easier since we won't have to get `getByText`, `getByRole`, etc, from the `render` function. And it follows the official docs example: https://testing-library.com/docs/react-testing-library/example-intro#full-example

ShimiSun:
@Lukazovic Thank you for the PR buddy! 💟 
I was trying to resolve conflicts but it seems that the functionality is relying on package we no longer import after the merge with master
>grommet/src/js/components/Button/__tests__/Button-test.js
  11:23  error  Unable to resolve path to module '@testing-library/user-event'  import/no-unresolved
  
 Can you please address this lint error on your PR?

Lukazovic:
> @Lukazovic Thank you for the PR buddy!

Thanks 🤗 

> I was trying to resolve conflicts but it seems that the functionality is relying on package we no longer import after the merge with master

It looks like the [`@testing-library/user-event`](https://testing-library.com/docs/ecosystem-user-event/) is missing, I'll add it back.

ShimiSun (approved):

PR #5328, opened 2021-06-16
Changed DateInput tests to remove react-test-renderer

ericsoderberghp:
#### What does this PR do?

Changed DateInput tests to remove react-test-renderer. This is being done to prepare for a fix to a DateInput issue, updating the tests first.

test

unit test, review snapshots

review snapshot changes

no

no

backwards compatible

### DIFF
--- a/src/js/components/DateInput/__tests__/DateInput-test.js
+++ b/src/js/components/DateInput/__tests__/DateInput-test.js
@@ -1,5 +1,4 @@
 import React from 'react';
-import renderer from 'react-test-renderer';
 import 'jest-styled-components';
 import { cleanup, fireEvent, render } from '@testing-library/react';
 import { axe } from 'jest-axe';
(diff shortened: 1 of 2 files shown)


src/js/components/DateInput/__tests__/DateInput-test.js
```diff
@@ -369,7 +351,8 @@ describe('DateInput', () => {
 
   test(`buttonProps should pass props to Button 
   when not inline and no format`, () => {
-    const component = renderer.create(
+    window.scrollTo = jest.fn();
```
jcfilben, line 354: Why was `window.scrollTo` included here? It is not affecting the test or snapshot

ericsoderberghp: Drop calls it via `setFocusWithoutScroll`. In the older `renderer` approach, we weren't going as far rendering, so this didn't show up. You can see how we do the same thing in Drop-test, Select-test, etc.

jcfilben: Ah okay thanks for the explanation

jcfilben (approved):

jlaluces123 (commented):
Just as a general question, since there are some components that already moved away from using `react-test-renderer` like this one, are there plans to fully migrate away from it down the line?

ericsoderberghp:
> Just as a general question, since there are some components that already moved away from using `react-test-renderer` like this one, are there plans to fully migrate away from it down the line?

Yes. We're not in a huge rush, but that's the direction. And, it's better to update the tests like this without having any corresponding changes to the component in the same PR. So, I have a DateInput fix coming and wanted to get the tests updated first.

PR #5330, opened 2021-06-17
Changed DateInput to fix an issue with setting the value

ericsoderberghp:
#### What does this PR do?

Changed DateInput to fix an issue with setting the value.

This changes how we determine when the caller, or the containing Form, changes the `value` vs. when the value changes due to the user typing.

Enhanced a story to include a "Today" button.
Added a unit test.

Use the Today button on the DateInput  Format inline story

#5247 

no

yes

backwards compatible

### DIFF
--- a/src/js/components/DataChart/DataChart.js
+++ b/src/js/components/DataChart/DataChart.js
@@ -1,9 +1,17 @@
-import React, { forwardRef, useMemo, useRef, useState } from 'react';
+import React, {
+  forwardRef,
+  useContext,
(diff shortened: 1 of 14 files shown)


jcfilben (approved):

taysea (approved):
This looks good -- maybe a separate issue (unless you think it relates here), but when using `format="mm/dd/yyyy-mm/dd/yyyy` and trying to type in a value, the second value in the format doesn't allow me to type anything once I press delete.

Issue #5312, opened 2021-06-09
DataChart does not display details for large data sets


GautamSatish:
Data chart doe not display details when the chart large number of points.

### Expected Behavior

DataChart should display details for the larger data sets or the limit should be documented so that data sets can be sampled down to required size.

### Actual Behavior

The vertical marker does not show up, probably because the points are too close to each other?

### URL, screen shot, or Codepen exhibiting the issue
CodeSandbox link: https://codesandbox.io/s/perf-data-j0rhn?file=/src/App.js

### Steps to Reproduce

1.  The sandbox renders sensor values at 30 intervals, 180 intervals and so on. Except for 30 intervals, the details do not appear for the rest of the charts.

### Your Environment

- Grommet version: 2.17.2
- Browser Name and version: Same across Chrome and Firefox
- Operating System and version (desktop or mobile): Does not matter

jcfilben:
Looks like the limit is 120, you can see that in this [codesandbox](https://codesandbox.io/s/perf-data-forked-4qxlm?file=/src/App.js) if you add one more field to the json data (bringing total to 121) the details will no longer display. I am still looking to see if it is possible to expand this limit or if should just be added to the documentation.

PR #5331, opened 2021-06-17
Changed DataChart to allow detail to work for more data points.

ericsoderberghp:
#### What does this PR do?

Changed DataChart to allow detail to work for more data points.

The old way was using the outer `pad` to drive the thickness of the detail segments. This sort of worked, as long as half the pad could be determined. But, when the thickness of the Chart elements was `hair`, the half size of that became 0, which meant there was no pad, but then no width to the Detail segments. This new approach separates the outer `pad` from the thickness of the detail segments.

Detail.js

Changed the Detail story to have 130 data points.

#5312 

no

yes

backwards compatible

### DIFF
--- a/src/js/components/DataChart/DataChart.js
+++ b/src/js/components/DataChart/DataChart.js
@@ -1,9 +1,17 @@
-import React, { forwardRef, useMemo, useRef, useState } from 'react';
+import React, {
+  forwardRef,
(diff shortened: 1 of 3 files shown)


jcfilben (approved):
Looks good tested with data set of 180

PR #5332, opened 2021-06-17
Refactor Markdown, SelectMultiple and ResponsiveContext tests to remove react-test-renderer

Lukazovic:
#### What does this PR do?

`src/js/components/Markdown/__tests__/Markdown-test.js`
`src/js/components/Select/__tests__/SelectMultiple-test.js`
`src/js/contexts/ResponsiveContext/__tests__/ResponsiveContext-test.js`

`yarn test`

Run: `yarn test`

No.

No.

Backwards compatible.

### DIFF
--- a/src/js/contexts/ResponsiveContext/__tests__/__snapshots__/ResponsiveContext-test.js.snap
+++ b/src/js/contexts/ResponsiveContext/__tests__/__snapshots__/ResponsiveContext-test.js.snap
@@ -12,8 +12,8 @@ exports[`ResponsiveContext basic 1`] = `
 }
 
 <div
-  className="c0"
+  class="c0"
 >
-  medium
+  small
 </div>
 `;
--- a/src/js/contexts/ResponsiveContext/__tests__/__snapshots__/ResponsiveContext-test.js.snap
+++ b/src/js/contexts/ResponsiveContext/__tests__/__snapshots__/ResponsiveContext-test.js.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`ResponsiveContext basic 1`] = `
+exports[`ResponsiveContext when viewport width is 768px should return small 1`] = `
 .c0 {
   font-size: 18px;
   line-height: 24px;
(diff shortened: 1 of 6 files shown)


src/js/contexts/ResponsiveContext/__tests__/__snapshots__/ResponsiveContext-test.js.snap
```diff
@@ -12,8 +12,8 @@ exports[`ResponsiveContext basic 1`] = `
 }
 
 <div
-  className="c0"
+  class="c0"
 >
-  medium
+  small
```
jcfilben, line 17: I'm curious why this is changing from medium to small

Lukazovic: @jcfilben me too. It might have some difference between the `react-test-renderer` and `testing-library` render methods but to be honest, I don't know the reason

Lukazovic: hey @jcfilben, I think I understood a little bit of what is going on.

So this test renders with the `<Grommet />` context, debugging a bit of the code I understood why the `size` is "small".

The code is in this file: [`src/js/components/Grommet/Grommet.js`](https://github.com/grommet/grommet/blob/master/src/js/components/Grommet/Grommet.js)

When the component is rendered the size is "small" because of this line:
```js
  const responsive =
    stateResponsive || // undefined
    deviceResponsive(userAgent, theme) || // userAgent = undefined => returns undefined
    theme.global.deviceBreakpoints.tablet; // "medium"
```

Then so far the size is "medium". But when the component is mounted, it triggers the [`useEffect`](https://github.com/grommet/grommet/blob/master/src/js/components/Grommet/Grommet.js#L98) function (which is expected). On the `useEffect` a new `stateResponsive` is set based on the value of the `document.body.clientWidth` that, in the test, is equal to 0. So the `stateResponsive` (size) becomes "small":
```jsx
  useEffect(() => {
    const onResize = () => {
      setResponsive(getBreakpoint(document.body.clientWidth, theme));
    };
    window.addEventListener('resize', onResize);
    onResize();
    return () => {
      window.removeEventListener('resize', onResize);
    };
  }, [theme]);
```
A screenshot with some `console.log` to help to understand the flow:
Screenshot from 2021-06-19 10-57-51

Lukazovic: I am going to improve the `ResponsiveContext` test suite to assert all the available sizes.

Lukazovic: I tested as suggested on this issue https://github.com/facebook/react/issues/15321 and in the docs: https://reactjs.org/docs/test-renderer.html#testrendereract. And it worked, it trigged the `useEffect` and the size changed to "small":

Screenshot from 2021-06-19 11-51-12

jcfilben: Good findings and explanation! Thank you

jcfilben (approved):

ericsoderberghp (approved):

PR #5348, opened 2021-06-20
Refactor Grommet tests to remove react-test-renderer

Lukazovic:
#### What does this PR do?

`src/js/components/Grommet/__tests__/Grommet-test.js`

`yarn test`

Run: `yarn test`

No.

No.

Backwards compatible.


src/js/components/Grommet/__tests__/Grommet-test.js
```diff
@@ -129,9 +126,8 @@ describe('Grommet', () => {
     /* eslint-enable max-len */
   ].forEach(ua => {
     test(`ssr rendering ${ua.substring(0, 25)}`, () => {
-      const component = renderer.create(<SSRTester ua={ua} />);
-      expect(component.toJSON()).toMatchSnapshot();
-      component.unmount();
+      const {container} = render(<SSRTester ua={ua} />);
+      expect(container.firstChild).toMatchSnapshot();
     });
```
jcfilben, line 131: I think this test could be improved similar to what was done in the Responsive context tests, right now it is returning small for each option

Lukazovic: Yeah, I agree. I'll try to improve it when I have some free time.

I even tried to run these tests with them wrapped in an `act()` and it returned "small" for each option as well. I think it is happening something similar to Responsive Context tests.
Screenshot from 2021-06-20 16-22-07

src/js/components/Grommet/__tests__/Grommet-test.js
```diff
@@ -129,9 +122,11 @@ describe('Grommet', () => {
...
   ].forEach(ua => {
     test(`ssr rendering ${ua.substring(0, 25)}`, () => {
-      const component = renderer.create(<SSRTester ua={ua} />);
-      expect(component.toJSON()).toMatchSnapshot();
-      component.unmount();
+      act(() => {
+        const { container } = render(<SSRTester ua={ua} />);
+
+        expect(container.firstChild).toMatchSnapshot();
+      });
```
Lukazovic, line 129: Hey, @jcfilben. I just improved the tests.

As I mentioned in the ResponsiveContext tests when we render with testing-library render it triggers the `useEffect` before starting the assertions. And as I also mentioned the `useEffect` changes the `size` based on the `document.body.clientWidth`. 

But in these specific tests, we want to test the `userAgent` so I decided to wrap the `render` in an `act` and test the snapshots inside this `act` so it doesn't trigger the `useEffect` which means we are testing the `userAgent`.

What do you think?

jcfilben: Nice solution, looks good to me

jcfilben (approved):

ericsoderberghp (approved):

PR #5349, opened 2021-06-21
Fix required FormField validation when value is an empty array

### LINKED ISSUE #5333: DateInput with initial value of [] does not allow required validation in FormField
### Expected Behavior

If a DateInput has a value of [] (indicating the user will be selecting a range of dates), the FormField should interpret this as an empty value so that required validation will still flag the input.

### Actual Behavior

When value of dateinput is initialized as [], it doesn't receive required validation ([] is interpretted as a value).

https://codesandbox.io/s/calendar-with-disabled-weekends-forked-jxbft?file=/index.js:604-606

### Steps to Reproduce

1. Submit the form
2. Date Range doesn't receive required validation.

### Your Environment

- Grommet version: 2.17.3
- Browser Name and version: Mac Chrome Version 91.0.4472.101 (Official Build) (x86_64)
- Operating System and version (desktop or mobile): Mac OS Mojave

Lukazovic:
#### What does this PR do?

Fix/Add required FormField validation when value is an empty array (`[]`).

Closes #5333

`src/js/components/Form/Form.js` and on `DateInput` storybook.

I created a test case on `src/js/components/Form/__tests__/Form-test-uncontrolled.js` to cover the bug described in #5333

Manual running: 

```bash
yarn test
```

I also simplified a validation on the same file: instead of doing `value2 === undefined || value2 === '' || value2 === false` we can simply do `!value2` that means the same thing and also validates when the value is `null`.

#5333

No.

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/Form/Form.js
+++ b/src/js/components/Form/Form.js
@@ -339,11 +339,7 @@ const Form = forwardRef(
           let result;
           if (
             required &&
-            // false is for CheckBox
-            (value2 === undefined ||
-              value2 === '' ||
-              value2 === false ||
-              (Array.isArray(value2) && !value2.length))
+            (!value2 || (Array.isArray(value2) && !value2.length))
           ) {
             result = messages.required;
           } else if (validateArg) {
--- a/src/js/components/Form/Form.js
+++ b/src/js/components/Form/Form.js
@@ -339,7 +339,11 @@ const Form = forwardRef(
           let result;
           if (
             required &&
-            (!value2 || (Array.isArray(value2) && !value2.length))
+            // false is for CheckBox
+            (value2 === undefined ||
+              value2 === '' ||
+              value2 === false ||
+              (Array.isArray(value2) && !value2.length))
           ) {
             result = messages.required;
           } else if (validateArg) {
--- a/src/js/components/Form/Form.js
+++ b/src/js/components/Form/Form.js
@@ -340,7 +340,10 @@ const Form = forwardRef(
           if (
             required &&
             // false is for CheckBox
-            (value2 === undefined || value2 === '' || value2 === false)
+            (value2 === undefined ||
+              value2 === '' ||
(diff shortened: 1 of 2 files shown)


src/js/components/Form/Form.js
```diff
@@ -339,8 +339,7 @@ const Form = forwardRef(
           let result;
           if (
             required &&
-            // false is for CheckBox
-            (value2 === undefined || value2 === '' || value2 === false)
+            (!value2 || (Array.isArray(value2) && !value2.length))
```
taysea, line 342: is there a reason we needed to remove the existing logic as opposed to just adding `(Array.isArray(value2) && !value2.length)` as one more condition? just worried we may be losing something. otherwise this is looking good.

Lukazovic: @taysea there's not a good reason, to be honest, I just thought the logic could be simplified.

This verification:
```js
if (!value) {
  // do something
}
```
is equal to:
```js
if (value === null || value === undefined || value === '' || value === false) {
  // do something
}
```

But I can bring back the old logic, what do you think?

ericsoderberghp: We still need to accommodate when `value` is set to 0. That will not be correctly handled by `!value`.

taysea: ^ Good catch, thanks for that. In this case, I would recommend leaving the existing logic and just adding your array condition as another `||` @Lukazovic

Lukazovic: Good catch @ericsoderberghp! I didn't think about that, thanks.

I'll update it, @taysea. Thanks

taysea (approved):
Looks good! Thanks for taking this on.

ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution!

Issue #2976, opened 2019-03-25
refactor documentation to remove react-desc


ericsoderberghp:
We can have issues in our documentation due to `react-desc` trying to combine `propTypes` with documentation and not always working well on either side. For example, we cannot reference propTypes from react-desc in other components due to how they are decorated. And, we need the ability to provide richer example syntaxes that don't work with the more simplified propType derived approach.

A suggestion is to have two files per component: `propTypes.js` and `doc.js`. The former would just have the `propTypes` and could be reference from other components. The latter would change the current documentation by removing the programatic prop definitions and instead support a richer `syntaxExamples` per each property, similar to what we do in `grommet-site`. The README.md files per component would be generated from `doc.js` with better syntax examples that aligned with grommet-site.

oorestisime:
How about we just write the proptypes in the component directly?
Should remove some friction of working on grommet (having to open multiple files for the same component currently the X.js, doc.js, index.d.ts and maybe propTypes.js)

atanasster:
Great to see some thiught given finally to take out react-desc. It has been truly a problem for code completion, debugging and overall complexity.

I agree the propTypes should be in the component file and remove the complexities introduced with doc.js files because of react-desc once and for all.

However i suggest completely removing react-desc - not just for prop types, but also for documenting. There are much better choices out there like styleguidist, docz etc, that would also support a future move to typescript.

harunurhan:
Having only `index.d.ts`, even going further by having `TypeScript` typings inside the component files (does not require migrating whole codebase to `TypeScript`) would be much better, IMHO.

Because:
- `TypeScript` is more expressive and powerful when it comes to defining types, compared to `prop-types`, even to `flow`.
- It has tooling for generating docs.
- You get build time static type checking which will help to catch bugs earlier.
- Will solve problems that are caused by inconsistent typings (different types for `prop-types` and `TypeScript`).

ericsoderberghp:
Two years later ...

- I think we should remove `react-desc` entirely.
- I think we should out the propTypes in a separate file, that makes the typescript typings and prop types consistent. I suggest this because I think there is enough complexity in some of the components typings that adding them into the component file would make it a bit unwieldy.
- I think all of the documentation should be in moved to the grommet-site repository. This will allow more customization of how complex types are described but just describing them in markup, rather than building a more complex JSON structure just to then have to convert that to markup later.
[...]

PR #5351, opened 2021-06-21
Removed react-desc from Anchor component

jcfilben:
#### What does this PR do?
This PR changes the Anchor component so that it no longer uses `react-desc` or exports the documentation. This is part of an effort to remove `react-desc` from Grommet and move the documentation to the grommet-site.

Notes:
- the `utils/general-prop-types` file was added as a copy of the `utils/prop-types` file but it uses `PropTypes` from `prop-types` instead of from `react-desc` and it doesn't include `.description` blocks. I added a new file instead of modifying the utils/prop-types file so that the components that still use react-desc are not affected. Once all components no longer use `react-desc` we can consolidate into one utils/prop-types file.

- the `generate-readme` file and readme snapshots can be removed once `react-desc` is removed from all components.

Anchor.js

Tested prop types with storybook (set the wrong type for a prop and check that warning is correctly output into console)

#2976 

Done in https://github.com/grommet/grommet-site/pull/221

No

Backwards compatible


ericsoderberghp (approved):

tools/generate-readme.js
```diff
@@ -37,23 +37,30 @@ const FOLDER = path.resolve('src/js/components');
...
+    const readmeContent = themeDoc
+      ? `${replaceHoc(DocumentedComponent.toMarkdown())}\n${toMarkdown(
+          themeDoc,
+        )}`
+      : `${replaceHoc(DocumentedComponent.toMarkdown())}`;
+    del(readmeDestination).then(() =>
+      fs.writeFileSync(readmeDestination, readmeContent),
+    );
+  } catch (ex) {
+    // don't make changes to README.md if doc.js file isn't found
```
ShimiSun, line 64: Are we intentionally hiding the exception? shouldn't we print it to the console in case something will go wrong?

ShimiSun (commented):
I might be missing something, but on the new structure suggested on this PR, where is Anchor's `themeDoc` being exported?

tools/generate-readme.js
jcfilben: As components have react-desc removed and no longer have doc files, the catch block will trigger. In this case nothing went wrong, it was expected that as react-desc is removed components will no longer have doc files in grommet. So I don't think printing to console is the right thing to do since it will make it look like something went wrong. This block was just added to maintain backwards compatibility for components that still use react-desc. Once react-desc is removed it can be removed as well so it is just temporary.

jcfilben:
Anchor's themeDoc is no longer being exported from grommet and is instead in the Anchor file on grommet-site. See PR grommet/grommet-site#221

tools/generate-readme.js
ShimiSun: Please add this detailed response as a comment on the catch block to avoid future confusion.

ShimiSun:
> Anchor's themeDoc is no longer being exported from grommet and is instead in the Anchor file on grommet-site. See PR [grommet/grommet-site#221](https://github.com/grommet/grommet-site/pull/221)

I'm not sure I'm comfortable with props of grommet core and theme props on the site. It will add complexity for future contributors, enhancements, and bug fixes. What was the motivation for that?

ericsoderberghp:
> > Anchor's themeDoc is no longer being exported from grommet and is instead in the Anchor file on grommet-site. See PR [grommet/grommet-site#221](https://github.com/grommet/grommet-site/pull/221)
> 
> I'm not sure I'm comfortable with props on grommet core and theme props on the grommet-site. It will add complexity for future contributors, enhancements, and bug fixes. What was the motivation for that?

There is no such thing as "prop-types" for themes. So, the only thing `themeDoc` was there for was documentation, which is being removed from grommet and put in grommet-site instead, which is what renders it already anyway. Agreed that grommet-site now has to be updated when any property or theme changes are made, to document the changes.

ShimiSun:
> So, the only thing `themeDoc` was there for was documentation, which is being removed from grommet and put in grommet-site instead, which is what renders it already anyway. Agreed that grommet-site now has to be updated when any property or theme changes are made, to document the changes.

Let's say a contributor is adding a new theme prop, the functionality will be added to grommet core, the types will be added to prop-types, the typescript types will be added to `base.d.ts` (also on grommet core), but the description, type, and example of the new theme prop will need to be added to the grommet-site on a separate PR. 
Sounds like we are opening a window to be forgetful about new theme props that should be edited/added per grommet core changes, with that, we're also making the contribution process more cumbersome.
I'm wondering why not consolidate all the documentation into one file, i.e. doc.js.

jcfilben:
Correct me if I am wrong @ericsoderberghp but I think the intention of moving the documentation to the site was that it would enable us down the line to have more flexibility with how the docs are being rendered. Right now the docs are staying in a similar format to how they were on grommet core, but moving them to the site will allows us to make changes down the line such as adding multiple examples per each type of prop, displaying parts of the documentation in markdown, having multiple descriptions for a prop when there are different types, and other changes we may want to do. There is more discussion about this on #2976

jcfilben:
Also related note: moving the documentation to the grommet-site will enable us to remove all the require statements from grommet and make sure we are aligned with ES6 module rules

jcfilben:
New Anchor documentation has been merged here: https://github.com/grommet/grommet-site/pull/224

The new documentation direction doesn't affect the changes here, they are still relevant and this PR is ready to be reviewed/merged since https://github.com/grommet/grommet-site/pull/224 has been merged

ericsoderberghp (approved):

PR #5355, opened 2021-06-22
Changed DataTable to refactor how cells are styled

ericsoderberghp:
#### What does this PR do?

Changed DataTable to refactor how cells are styled.
This fixes some issues with custom styling grouped row expand/collapse.

The essence of the refactoring is to calculate the appropriate cell properties in `cellProps` up from in DataTable, via `normalizeCellProps()`. From there, DataTable passes the appropriate flavor of `cellProps` down to Header, Footer, Body, and GroupedBody. These ultimately get passed to Cell and TableCell. The only munging along the way is to handle per-row styling at the row level via `normalizeRowCellProps()`. Previously, we were re-calculating the cell style props for every cell, which was inefficient and also ended up with manipulations in multiple places. Hopefully, this new approach makes it clearer to understand and more consistent, especially with theming and caller prop styling.

Along with this TableCell has been enhanced to set `theme.dark` if a background color is used. This fixes an issue where a DataTable header had a dark custom background color but the group ExpanderCell wasn't handling that when deciding what color to make its icon.

I'd start with DataTable.js and then Cell.js to see the key changes.
Then, I'd review buildState.js to see how the props are handled.

Storybook, enhanced the DataTable Styled story to include groups, which would have demonstrated some of the issues fixed.

Storybook

no

yes

backwards compatible


jcfilben (commented):
In the DataTable/Styled story my assumption would be that the 'Eric' row would have 'accent-2' as its background color based on this line `rowProps={{ Eric: { background: 'accent-2', pad: 'small' } }}` but instead this appears to be overridden by the background.body prop so the Eric row has 'light-1' as its background color, is that the intended behavior?

src/js/components/TableCell/TableCell.js
```diff
@@ -125,31 +125,48 @@ const TableCell = forwardRef(
...
-          <Box
-            {...mergedProps}
-            align={align}
-            justify={verticalAlignToJustify[verticalAlign]}
-          >
-            {children}
-          </Box>
-        )}
-      </StyledTableCell>
+      <ThemeContext.Provider value={themeProviderValue}>
```
MikeKingdom, line 144: I think this is OK. I was trying to figure out if there are any issues with themeProviderValue basically changing each render. Probably not and it's good that it always includes this context so we don't run into what we saw in #5146

MikeKingdom (approved):
Nice improvements

ericsoderberghp:
> In the DataTable/Styled story my assumption would be that the 'Eric' row would have 'accent-2' as its background color based on this line `rowProps={{ Eric: { background: 'accent-2', pad: 'small' } }}` but instead this appears to be overridden by the background.body prop so the Eric row has 'light-1' as its background color, is that the intended behavior?

Great catch! Fixed. Changed Styled story to pre-expand custom styled row so that Chromatic will capture this in the future.

src/js/components/TableCell/TableCell.js
ericsoderberghp: Good thoughts. I've reworked the code to avoid redoing the ThemeContext unless there's something to put in it.

src/js/components/DataTable/Header.js
```diff
@@ -23,36 +23,37 @@ import {
...
     color,
     font,
     gap, // gap is used for space between header cell elements only
+    pad, // covered by cellProps
     units,
     ...rest
   } = theme.dataTable.header;
 
-  const cellProps = { background, border };
+  // const cellProps = { background, border };
```
jcfilben, line 44: Should this line just be removed instead of commented out?

src/js/components/DataTable/Cell.js
```diff
@@ -5,17 +5,15 @@ import { defaultProps } from '../../default-props';
 
 import { Text } from '../Text';
 import { StyledDataTableCell } from './StyledDataTable';
-import { datumValue, normalizeBackgroundColor } from './buildState';
+import {
+  datumValue,
+  // normalizeBackgroundColor,
```
jcfilben, line 10: Can this comment be removed

src/js/components/TableCell/TableCell.js
```diff
@@ -125,49 +125,58 @@ const TableCell = forwardRef(
...
         themeProviderValue.dark = dark === undefined ? theme.dark : dark;
         themeProviderValue.background = cellProps.background;
       } else if (cellProps.background) {
         // This allows DataTable to intelligently set the background of a pinned
         // header or footer.
+        themeProviderValue = { ...theme };
         themeProviderValue.background = cellProps.background;
       }
+
+      if (themeProviderValue) {
```
MikeKingdom, line 170: I don't think you should do this conditionally or we could easily run into something like #5146. If the color changes or light vs dark theme etc, you could cause a new hierarchy which causes the components within this cell to get re-created, losing any state or focus, etc.

ericsoderberghp: good catch, reverted

MikeKingdom (approved):
LGTM!

jcfilben (commented):
The `pad` and `border` props in `rowProps` have a different look than previously. Using the code `Michael: { pad: 'large', border: '1px solid black' }` in rowProps before this would show up like this:
Screen Shot 2021-06-23 at 4 00 12 PM

Now it is showing up like this:
Screen Shot 2021-06-23 at 3 49 52 PM

src/js/components/DataTable/Header.js
ericsoderberghp: good catch, removed

src/js/components/DataTable/Cell.js
ericsoderberghp: good catch, removed

ericsoderberghp:
> The `pad` and `border` props in `rowProps` have a different look than previously. Using the code `Michael: { pad: 'large', border: '1px solid black' }` in rowProps before this would show up like this:

That looks correct to me. The column order was changed in the story but the row cells seem to be picking up the right styling. Perhaps I'm missing something?

jcfilben:
> That looks correct to me. The column order was changed in the story but the row cells seem to be picking up the right styling. Perhaps I'm missing something?

Nope your right, I just realized that the column on the left for the grouping is what is pushing things over and the border and padding are correct

jcfilben (approved):

PR #5356, opened 2021-06-22
Add a global messages property to Grommet 

### LINKED ISSUE #5337: Add I18N messages to Grommet
Add a global messages property to `Grommet`.

```jsx
<Grommet messages={ {
  messages: { ... },
  format: ({ id } ) => { ... },
}} />
```

Each component that currently has its own `messages` property can use those definitions first. If not defined, it will use the format function to get the text. The default format function will look in this messages passed to Grommet.

This will require messages to have unique keys so the default format function will need to first look for things like "menu.openMenu" rather than just "openMenu".

All the default messages will move to a base messages file (similar to the theme's base definitions.)

MikeKingdom:
#### What does this PR do?
This adds a `messages` property to `Grommet`. This is similar to how you can provide
messages to individual components, but this allows a developer to provide all messages
needed for grommet components in one place. The individual components will pick up
their messages from this global setting, If a developer passes a `messages` property to
individual components, those values will be used instead.

The message object passed to `Grommet`  has a similar hierarchical structure to the
theme. For example:
```javascript
{
  messages: {
    grommet: {
      menu: {
        openMenu: "Open Menu",
        closeMenu: "Close Menu",
      },
    },
  },
}
```
The grommet components access their messages with a hierarichical  key from a `format`
function they get from `MessageContext`. For example:
```jsx
          format({ id: 'grommet.menu.openMenu', messages })
```
The extra `messages` member passed in the options allows the component to pass any
`messages` property that was passed to the component. Messages will be resolved from
that component first with the last part of the key (e.g. `openMenu`). This allows the
message property specified to the component to work the same as it did before and
override any message specified at the global level.

This change also allows the messages to come from a `format` function passed in the
 Grommet `messages` property. This makes it easier to use packages like `react-intl` and
 `react-i18next` which allow the developer to use a format function to reference messages
dynamically (sometimes even remotely,) One advantage to this is it's easy to specify all
their messages in fewer bundles (easier for localizers) and access them by simply specifying
a single `format` function at the top level of the application. 

For example, with `react-intl`:
```jsx
  const intl = useIntl();
  ...
   <Grommet messages={{ format: opts => intl.formatMessage({ id: opts.id }) }}>
    ...
   </Grommet>
```
In this case the grommet messages would be defined in the bundle they provided
to `react-intl` when  it was initialized. This bundle would typically look more like this:
```json
{
  "grommet.menu.openMenu": "Open Menu"
}
```

MessageContext.js

Storybook, Jest test cases added

Go to the new Grommet "Messages" example in the storybook.

Fixes #5337 

Yes

Yes

Backward compatible


ericsoderberghp:
Since we are already in a "grommet" context. It would seem cleaner to me to have it be: `messages= {{ menu: { ... }, ... }}`, rather than having "messages" and "grommet" layers in the object. Are those layers needed?

ericsoderberghp:
Would we be able to pass any parameters via `format({ id: 'grommet.menu.openMenu', messages })`? For example, when we have `<Select multiple />`, maybe the message when the drop is closed could be "5 selected" instead of "multiple". We'd need a way to pass the "5" to the formatter.

src/js/components/FileInput/__tests__/__snapshots__/FileInput-test.js.snap
```diff
@@ -1239,7 +1239,7 @@ exports[`FileInput messages 1`] = `
       <span
         class="c4 "
       >
-        Drag and drop
+        Drop file here or
```
ericsoderberghp, line 1242: Any idea why this changed?

ericsoderberghp (commented):
Code looks good. One small question. Two comments overall that I put outside of this review.

MikeKingdom:
> Since we are already in a "grommet" context. It would seem cleaner to me to have it be: `messages= {{ menu: { ... }, ... }}`, rather than having "messages" and "grommet" layers in the object. Are those layers needed?

I debated about this a bit and felt like it should be namespaced. But after seeing it more in context I think I agree. I was originally worried that people would be forced to do something like
 ```jsx
messages={{ format: opts => intl.formatMessage({ id: `grommet.${opts.id}`}) }}
```
to keep things separate from their messages, but I'm sure we can let them decide if that's needed.

**Comment by MikeKingdom** [comment]
> Would we be able to pass any parameters via `format({ id: 'grommet.menu.openMenu', messages })`? For example, when we have `<Select multiple />`, maybe the message when the drop is closed could be "5 selected" instead of "multiple". We'd need a way to pass the "5" to the formatter.

Yes, which is why I kept the options object arg rather than any simple parameters. The idea was we could pass a `values` member in opts which would work seemlessly with stuff based on FormatJS like react-intl. For example:

The message would be defined as something like:
```jsx
select: {
   multiple: '{total} selected'
}
```
And the code would do
```jsx
format({
  id: 'select.multiple',
  values: { total: 5 },
})
```
And for `react-intl` all they would need to do is:
```jsx
<Grommet messages={{ format: opts => intl.formatMessage(opts) }}>
```

src/js/components/FileInput/__tests__/__snapshots__/FileInput-test.js.snap
MikeKingdom: Yes. This was a case where the way FileInput was written it takes the messages object passed in and only uses that as a whole. It no longer has whatever was in defaultProps.messages. So if you pass in a sparse messages object (ie. missing some messages) it was using the hardcoded default rather than the message in the default props. This was the only case where the hardcoded was different than default props.

The way this works, the default always comes from default.json (although the format function supports a defaultMessage member similar to intl.formatMessage()). And personally I thought that hidden default was not very useful :)

MikeKingdom:
Question: where should we document the format and contents of default.json?

ericsoderberghp:
> > Would we be able to pass any parameters via `format({ id: 'grommet.menu.openMenu', messages })`? For example, when we have `<Select multiple />`, maybe the message when the drop is closed could be "5 selected" instead of "multiple". We'd need a way to pass the "5" to the formatter.
> 
> Yes, which is why I kept the options object arg rather than any simple parameters. The idea was we could pass a `values` member in opts which would work seemlessly with stuff based on FormatJS like react-intl. For example:
> 
> The message would be defined as something like:
> 
> ```js
> select: {
>    multiple: '{total} selected'
> }
> ```
> 
> And the code would do
> 
> ```js
> format({
>   id: 'select.multiple',
>   values: { total: 5 },
> })
> ```
> 
> And for `react-intl` all they would need to do is:
> 
> ```js
> <Grommet messages={{ format: opts => intl.formatMessage(opts) }}>
> ```

Makes sense. Would be good to add to documentation.

ericsoderberghp:
> Question: where should we document the format and contents of default.json?

I'd assume in the description of the Grommet `messages` prop.

I think we'll need typescript typings, like we do for the theme prop.

MikeKingdom:
BTW, I don't think the `messages` property documented on Accordion actually does anything.

MikeKingdom:
The failed UI test is a drop shadow showing on the edge from the Skip Links div and seems unrelated to this change

src/js/components/Grommet/__tests__/Grommet-test.js
```diff
@@ -129,4 +130,36 @@ describe('Grommet', () => {
       });
     });
   });
+
+  test('messages', () => {
+    const component = renderer.create(
```
ShimiSun, line 135: Can we try to stay away from using `renderer.create` when creating new tests? Please see more details on https://github.com/grommet/grommet/issues/5197

src/js/components/Grommet/__tests__/Grommet-test.js
```diff
@@ -129,4 +130,36 @@ describe('Grommet', () => {
...
+      </Grommet>,
+    );
+    expect(component.toJSON()).toMatchSnapshot();
+  });
+
+  test('message format function', () => {
+    const messages = {
+      "test.label": "My Label",
+    };
+    const component = renderer.create(
```
ShimiSun, line 156: Can we try to stay away from using `renderer.create` when creating new tests? Please see more details on https://github.com/grommet/grommet/issues/5197

src/js/languages/default.json
```diff
@@ -0,0 +1,48 @@
+{
```
ShimiSun, line 1: Broad question, how would this file work when the grommet user isn't using English?

src/js/contexts/MessageContext/MessageContext.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
+
+import defaultMessages from '../../languages/default.json';
+
+// opts:
```
ShimiSun, line 5:
```suggestion
// prop options:
```
I think?!

src/js/contexts/MessageContext/MessageContext.js
```diff
@@ -0,0 +1,46 @@
...
+//   id: message id
+//   messages: (optional) an object of message overrides
+//   values: (optional) currently unused but in the future
+//     will be an object with substitution values for 
+//     positional variables in the message text.
+//   defaultMessage: (optional) default message to use if
+//     the message isn't found elsewhere.
+export const format = (opts, messages) => {
```
ShimiSun, line 13: Can we rename `opts` to `options` to be more explicit? Or am I missing another hidden meaning?

src/js/components/__tests__/__snapshots__/README-test.js.snap
```diff
@@ -19120,17 +19237,7 @@ boolean
...
-  \\"fullScreen\\": \\"full screen\\",
-  \\"progressMeter\\": \\"video progress\\",
-  \\"scrubber\\": \\"scrubber\\",
-  \\"openMenu\\": \\"open menu\\",
-  \\"pauseButton\\": \\"pause\\",
-  \\"playButton\\": \\"play\\",
-  \\"volumeDown\\": \\"volume down\\",
-  \\"volumeUp\\": \\"volume up\\"
-}\`.
+Custom messages. Used for accessibility by screen readers.
```
ShimiSun, line 19240: Why did we remove the default messages? Is it breaking backward compatibility? Should we add a reference to the new messages object?

src/js/components/Grommet/__tests__/Grommet-test.js
MikeKingdom: Certainly. Latest merge brought in the cleanup of this.

src/js/components/Grommet/__tests__/Grommet-test.js
MikeKingdom: Fixed

src/js/contexts/MessageContext/MessageContext.js
MikeKingdom: It's not a prop. Just an options argument to the function

src/js/components/__tests__/__snapshots__/README-test.js.snap
MikeKingdom: The components don't specify a default for their messages object anymore since the messages come from the overall grommet messages default. This doesn't break backward compatibility since if they don't specify a messages property to the component they still get the same text. It just comes from a different place internally. The doc.js does still tell you what the default message text is. If you have some thoughts about pointing them toward grommet's overall messages that might help? It's more like the theme now really.

src/js/languages/default.json
MikeKingdom: It would still be the default. This isn't dependent on browser locale. Eric and I talked a bit about whether grommet itself would have localizations for it's messages and it's our expectation that whomever had an app that was being localized would at that point provide localized versions of the messages. BTW, one rule of I18N/L10N is to avoid having any mixed languages. If we provided any localizations that got triggered by browser locale or such, that could result in grommet messages being in one language while the app's content was maybe in another language. Best to make sure the app controls all that.

src/js/contexts/MessageContext/MessageContext.js
MikeKingdom: Can if you prefer. I tend to use opts kinda like how we tend to abbreviate properties as props. I think I adopted this one from intl.formatMessage() but full spelling is certainly fine.

ericsoderberghp (approved):

ShimiSun (approved):

PR #5371, opened 2021-06-24
Fix: Active Style to persist on List

jlaluces123:
Addresses #5266 : When selecting a row from a list, the active styling on the item remains when a user opens a modal / layer detail, but when closing the modal, the active styling disappears, preventing users from seeing where they last left off.

#### What does this PR do?
- Persists the active styling during an `onBlur` and `onMouseOut` for List Component
- Updates to List's `onClickItem` story to match the scenario found in issue #5266

1. In `components/List/List.js` view the `clickProps` to see how the active styling is applied / persisting depending on the events as well as the `onFocus` and `onBlur` attributes given to `<StyledList>` 
2. `yarn storybook`
3. Navigate to `List's onClickItem` story
4. Click on any list item (or navigate via keyboard by tabbing to the table and pressing enter on an item), close the modal (via pressing `escape`, clicking `close`, or clicking outside the modal) and watch how the active styling remains until you hover other list items

Manual testing via the onClickItem story.

1. `yarn storybook`
2. Navigate to `List's onClickItem` story
3. Click on any list item (or navigate via keyboard after tabbing to the table and press enter), close the modal and watch how the active styling remains until you hover other list items

- #5266

No.

No.


ShimiSun:
Removing the Task Force project as we try to assign issues to it and not PRs. It will help us to maintain fewer duplications on the board.

src/js/components/List/List.js
```diff
@@ -357,13 +366,22 @@ const List = React.forwardRef(
...
                     },
-                    onMouseOver: () => setActive(index),
-                    onMouseOut: () => setActive(undefined),
-                    onFocus: () => {
+                    onMouseOver: () => {
+                      setLastActive(undefined);
                       setActive(index);
+                    },
+                    onMouseOut: () => {
+                      console.log('Mouse Out');
```
ericsoderberghp, line 374: need to remove console logging

src/js/components/List/List.js
```diff
@@ -357,13 +366,22 @@ const List = React.forwardRef(
...
+                      setActive(lastActive);
```
ericsoderberghp, line 375: Why are we setting `active` during `onMouseOut`?

I would have expected that we would set `active` as we already do. Then, we would `setLastActive(active)` during `onBlur`. Then, in `onFocus` we would `setActive(lastActive)`. That feels like it would mirror the behavior we are shooting for and be simple enough. But, I wonder if I'm missing something?

src/js/components/DataTable/Body.js
```diff
@@ -165,6 +198,7 @@ const Body = forwardRef(
               }
             : undefined
         }
+        onTab={() => setActive(undefined)}
```
ericsoderberghp, line 201: I'm still unclear why we need `onTab` and can't do everything via `onFocus` and `onBlur`.

src/js/components/DataTable/Body.js
```diff
@@ -216,6 +251,10 @@ const Body = forwardRef(
               return (
                 <Row
                   key={index}
+                  tableRef={ref}
```
ericsoderberghp, line 254: Is this really the "table" ref? It looks more like the "tbody" ref.

src/js/components/List/List.js
jlaluces123: After applying `onFocus and onBlur` directly to the `<StyledList>` the intended logic you described is now back in place.

src/js/components/DataTable/Body.js
jlaluces123: Renamed this to `tableBodyRef` to be more accurate in the newly-separated PR for DataTable in https://github.com/grommet/grommet/pull/5397

src/js/components/List/List.js
```diff
@@ -225,17 +227,17 @@ const List = React.forwardRef(
...
             (onClickItem || onOrder) && data && data.length
               ? () => {
                   const min = onOrder ? 1 : 0;
-                  const max = onOrder ? (data.length * 2) - 2 : data.length - 1;
-                  setActive(
-                    active >= min ? Math.min(active + 1, max) : min,
-                  );
+                  const max = onOrder ? data.length * 2 - 2 : data.length - 1;
+                  setActive(active >= min ? Math.min(active + 1, max) : min);
+                  setLastActive(active);
```
ericsoderberghp, line 240: I don't think we need to `setLastActive` here.

src/js/components/List/List.js
```diff
@@ -245,7 +247,19 @@ const List = React.forwardRef(
             as={as || 'ul'}
             itemFocus={itemFocus}
             tabIndex={onClickItem || onOrder ? 0 : undefined}
-            onBlur={onOrder ? () => setActive(undefined) : undefined}
+            onBlur={
+              onOrder
```
ericsoderberghp, line 251: I think we should always `setLastActive` in `onBlur`, irrespective of whether `onOrder` is being used.

src/js/components/List/List.js
```diff
@@ -245,7 +247,19 @@ const List = React.forwardRef(
...
+                ? () => {
+                    setActive(undefined);
+                  }
+                : () => {
+                    if (active) setLastActive(active);
```
ericsoderberghp, line 256: If `active` is 0, this won't be true. I think it's fine to always `setLastActive(active)` irrespective of what `active` is.

src/js/components/List/List.js
```diff
@@ -245,7 +247,19 @@ const List = React.forwardRef(
...
+                    setActive(undefined);
+                  }
+            }
+            onFocus={() => {
+              if (lastActive) setActive(lastActive);
```
ericsoderberghp, line 261: If `lastActive` is zero, this won't be true. I think we can always `setActive(lastActive)` irrespective of what `lastActive` is.

src/js/components/List/List.js
```diff
@@ -357,13 +371,17 @@ const List = React.forwardRef(
                       // accessibility guidelines that focus remains on `ul`
                       listRef.current.focus();
                     },
-                    onMouseOver: () => setActive(index),
+                    onMouseOver: () => {
+                      setLastActive(undefined);
```
ericsoderberghp, line 375: I don't think we need to `setLastActive` here.

src/js/components/List/List.js
```diff
@@ -357,13 +371,17 @@ const List = React.forwardRef(
...
+                      setActive(index);
+                    },
                     onMouseOut: () => setActive(undefined),
                     onFocus: () => {
                       setActive(index);
                       setItemFocus(true);
                     },
                     onBlur: () => {
+                      setLastActive(active);
```
ericsoderberghp, line 384: I don't think we need to `setLastActive` here.

src/js/components/List/stories/onClickItem.js
```diff
@@ -25,14 +25,28 @@ for (let i = 0; i < 40; i += 1) {
...
+            console.log('=== Modal Opened ===');
+            setClicked(event.item);
+          }}
         />
 
+        {show && (
+          <Layer
+            onEsc={() => setShow(false)}
+            onClickOutside={() => setShow(false)}
+          >
```
ericsoderberghp, line 46: Can we add something inside the Layer indicating what the active object is? I think that will reinforce what's going on.

jlaluces123: To clarify, would you like to display what the current active item was before the Layer opened inside the Layer?

ericsoderberghp: No, I was wanting to only show the detail of the item inside the Layer. I think we can remove the previous behavior of showing it below the List.

jlaluces123: Ahh gotcha! Will do

src/js/components/List/List.js
jlaluces123: Also removed this from `onUp` as well and everything still works as intended!

src/js/components/List/List.js
jlaluces123: Updated so now it doesn't have to check for `onOrder`, works as intended still.

src/js/components/List/List.js
jlaluces123: After removal, functionality works as intended

src/js/components/List/List.js
jlaluces123: After removal, functionality works as intended

src/js/components/List/List.js
jlaluces123: After replacing, works as intended with the new updates to its own `onFocus`

src/js/components/List/List.js
jlaluces123: After logging the relationship between `active and lastActive` , I updated this code chunk to handle what happens when `active` is read as `falsy`.

Now when you click / press enter on the first entry in the `onClickItem` story, remove the table's focus, and then regain the table's focus, the first entry still remains active as intended. 

This also solves the bug where:
- clicking / pressing enter on an item, clicking off the table / removing table focus, regaining table focus, and then holding down a click on another item would show the focus style on the `lastActive` instead of the item the user is currently clicking down on.

src/js/components/DataTable/Body.js
jlaluces123: Now moved to the DataTable-specific PR, but is now removed

src/js/components/List/List.js
```diff
@@ -245,7 +243,15 @@ const List = React.forwardRef(
             as={as || 'ul'}
             itemFocus={itemFocus}
             tabIndex={onClickItem || onOrder ? 0 : undefined}
-            onBlur={onOrder ? () => setActive(undefined) : undefined}
+            onFocus={() =>
+              !active && active !== 0
```
ericsoderberghp, line 247: So, if `active === undefined`, we are going to `setActive(active)`? That seems odd. Can we not just do `onFocus={() => setActive(lastActive)}`? Why do we need to check `active`?

ericsoderberghp:
Can you merge this with `master` again? I think that will remove most of the syntax changes.

src/js/components/List/List.js
jlaluces123: After implementing `onFocus={() => setActive(lastActive)}` the issue that would arise was:

1. You selected / clicked on an item
2. You tab out of the table / click out of the table
3. You hold down a click on a different item and the active styling would show up on the previous item (in this case `lastActive`)

And one of the issues that we realized on the most recent deployment preview was the fact that clicking the first entry (or selecting it via keyboard) would not retain its active styling after closing the modal as the code would read the `0th` index as `undefined`, thus messing up the entire chain of `onFocuses` and `onBlurs`. After logging out the exact orders and values of what `active` and `lastActive` were, doing the check here eliminates `undefined` getting passed down the chain of functions. While it looks out of place, when logging out the chain of events that occurs when a user clicks, closes the modal, and clicks / tabs out of the table, the check is necessary to pass the correct indices of what is active or what was lastActive.

ericsoderberghp: OK. Can you add a brief comment to the code then? This could be helpful detail down the road.

jlaluces123: No problem, will do!

MikeKingdom:
Has this been tested with onOrder and keyboard interaction with onOrder? `active` behaves differently in that case since it's tracking the up/down controls rather than the entire row.

jlaluces123:
> Has this been tested with onOrder and keyboard interaction with onOrder? `active` behaves differently in that case since it's tracking the up/down controls rather than the entire row.

Not yet, I will actively test this today as well to ensure everything is accounted for.

jlaluces123:
@MikeKingdom after comparing this branch with the deployed storybook for List's `Order` story, the only change I saw was when navigating via keyboard, if you were to "select" a button (the up or down controls), make the list lose focus, and then gain focus again by tabbing into the list, the control you previously selected would be the first item that gets focused whereas the deployed storybook would start from the top of the list (first focus goes to the table, pressing down would be the down control for the very first item in the table) every time.

ericsoderberghp (approved):

MikeKingdom:
> @MikeKingdom after comparing this branch with the deployed storybook for List's `Order` story, the only change I saw was when navigating via keyboard, if you were to "select" a button (the up or down controls), make the list lose focus, and then gain focus again by tabbing into the list, the control you previously selected would be the first item that gets focused whereas the deployed storybook would start from the top of the list (first focus goes to the table, pressing down would be the down control for the very first item in the table) every time.

That sounds broken although one trick with this is the list rows need a consistent `key` attribute. You can see in the code how we try to make this happen, but it depends a bit on how the data is passed in by the developer. If they key value changes for a particular item when it changes position, it will instantiate a new jsx element, losing any state of focus of the old one. This is why React warns you to not use an index for a key value :)

MikeKingdom:
The netlify preview of storybook only partially works. stories below the datachart fail to load so it's impossible to check the list stories

jlaluces123:
> The netlify preview of storybook only partially works. stories below the datachart fail to load so it's impossible to check the list stories

Oh does this mean you're unable to view the List's stories in this case?

> That sounds broken although one trick with this is the list rows need a consistent `key` attribute. You can see in the code how we try to make this happen, but it depends a bit on how the data is passed in by the developer. If they key value changes for a particular item when it changes position, it will instantiate a new jsx element, losing any state of focus of the old one. This is why React warns you to not use an index for a key value :)
>
> Note that we also don't let the up/down buttons get focus and instead handle click and keyboard events at the List level. The list itself should be the only thing that gets focus when you tab

After checking to make sure, tabbing only focuses on the List and then to get to the controls, you have to use up/down keys or mouse. And I just wanted to clarify for your comment, could you point to where in the code I would need to take a look at?

MikeKingdom:
> Oh does this mean you're unable to view the List's stories in this case?

Correct and Eric is looking at that issue (sounds like webpack issue)
 
> After checking to make sure, tabbing only focuses on the List and then to get to the controls, you have to use up/down keys or mouse. And I just wanted to clarify for your comment, could you point to where in the code I would need to take a look at?

That sounds more correct than what I though you were describing. It's definitely only intended to keep focus on the List itself. The active highlight should stay on the same item's up or down control as you use the keyboard or mouse to move items with that control.

jlaluces123:
> The active highlight should stay on the same item's up or down control as you use the keyboard or mouse to move items with that control.

Gotcha, I know you mentioned the story is unviewable via the deploy preview but to check on my end, I selected an up control, tabbed out of the list, regained focus on the list and saw:

1. The table is focused with the green border
2. The active styling is still on the up control for the correct item it corresponds to.

Would this be the correct behavior we're looking for in the `Order` story? If so, I think it's just a matter of time so you can view it on your end too.

MikeKingdom:
> Gotcha, I know you mentioned the story is unviewable via the deploy preview but to check on my end, I selected an up control, tabbed out of the list, regained focus on the list and saw:
> 
> 1. The table is focused with the green border
> 2. The active styling is still on the up control for the correct item it corresponds to.
> 
> Would this be the correct behavior we're looking for in the `Order` story? If so, I think it's just a matter of time so you can view it on your end too.

Sounds pretty good and thanks for checking.

MikeKingdom (approved):
Thanks!

PR #5397, opened 2021-06-29
Fix: Active Style to persist on DataTable

jlaluces123:
Addresses #5266 : When selecting a row from a DataTable, the active styling on the item remains when a user opens a modal / layer detail, but when closing the modal, the active styling disappears, preventing users from seeing where they last left off.

#### What does this PR do?
- Persists the active styling during an `onBlur` for DataTable Component
- Updates to DataTable's `Clickable` story to match the scenario found in issue #5266

1. In `components/DataTable/Body.js` view how the `<StyledDataTableBody>` handles `onFocus and onBlur`
2. In `components/DataTable/StyledDataTable.js` view how `StyledDataTableRow` and `StyledDataTableBody` apply styling changes
2. `yarn storybook`
3. Navigate to `DataTable's Clickable` story
4. Click on any row (or navigate via keyboard by tabbing to the table and pressing enter on a row), close the modal (via pressing `escape`, clicking `close`, or clicking outside the modal) and watch how the active styling remains until you hover other list items

Manual testing via the Clickable story.

1. `yarn storybook`
2. Navigate to `DataTable's Clickable` story
3. Click on any row (or navigate via keyboard after tabbing to the table and press enter), close the modal and watch how the active styling remains until you hover other list items

- #5266

No.

No.


ericsoderberghp:
List.js is still in this PR

ericsoderberghp:
I'm not seeing the "Clickable" DataTable story in the storybook preview. Can you investigate why it's not there? Perhaps a merge with master is needed?

jlaluces123:
> I'm not seeing the "Clickable" DataTable story in the storybook preview. Can you investigate why it's not there? Perhaps a merge with master is needed?

Yeah I just checked it out and 

> I'm not seeing the "Clickable" DataTable story in the storybook preview. Can you investigate why it's not there? Perhaps a merge with master is needed?

Yeah I was thinking the same thing, I’ll be able to merge this in a bit but Mike also brought up a good point in the List PR that the deployment previews only seem to partially load the storybook. Were you able to test the DataTable changes in previous iterations of my work on DataTable?

ericsoderberghp:
I'm working on: https://github.com/grommet/grommet/issues/5421

jlaluces123:
> I'm working on: #5421, which should fix the typescript stories

Just checked out the issue, should I just be on the lookout when the issue is resolved to continue testing this PR?

src/js/components/DataTable/StyledDataTable.js
```diff
@@ -82,7 +81,9 @@ const StyledDataTableRow = styled(TableRow)`
     props.onClick &&
     `
     cursor: pointer;
-  `}
+  `}  
+
+  // Could remove: changes nothing UI-wise
```
taysea, line 86: Should we remove this styling block and comment?

src/js/components/DataTable/StyledDataTable.js
```diff
@@ -103,12 +104,11 @@ const StyledDataTableBody = styled(TableBody)`
...
+  // Needed to give outline styling
   &:focus {
-    ${focusStyle({ skipSvgChildren: true, forceOutline: true })}
-  }
-
-  &:focus:not(:focus-visible) {
-    ${unfocusStyle({ skipSvgChildren: true, forceOutline: true })}
+    ${props =>
+      props.tabIndex >= 0 &&
+      focusStyle({ forceOutline: true, skipSvgChildren: true })}
```
taysea, line 111: Noticing that with mouse click, the focus outline is appearing around the datatable body. We made [recent changes](https://github.com/grommet/grommet/pull/5148) so that focus outline would only appear on keyboard interaction. Is there anyway to have the body receive focus without causing the focus outline on mouse click?

src/js/components/DataTable/StyledDataTable.js
jlaluces123: I haven't ran through all of the other DataTable stories with this removed in a while so I will look into this today to see if it really doesn't affect anything else

src/js/components/DataTable/stories/typescript/Clickable.tsx
```diff
@@ -140,19 +140,43 @@ const DATA: RowType[] = [
...
-  </Grommet>
-);
+export const ClickableDataTable = () => {
+  const [show, setShow] = React.useState(false);
+  const [clicked, setClicked] = React.useState({});
+
+  return (
+    <Grommet theme={grommet}>
+      <Box align="center" pad="large">
+        {/* eslint-disable no-alert */}
```
taysea, line 150: why do we need this eslint-disable?

jlaluces123: This will be trimmed in the next commit, totally forgot to get rid of this as the first iteration of this story was using an `alert()`

src/js/components/DataTable/StyledDataTable.js
jlaluces123: Just to clarify, so we would like the DataTable body to only receive the focus outline via keyboard navigation? If so, this should be a quick fix I can look into.

src/js/components/DataTable/StyledDataTable.js
```diff
@@ -107,8 +110,8 @@ const StyledDataTableBody = styled(TableBody)`
     ${focusStyle({ skipSvgChildren: true, forceOutline: true })}
   }
 
-  &:focus:not(:focus-visible) {
-    ${unfocusStyle({ skipSvgChildren: true, forceOutline: true })}
+  &:not(:focus-visible) {
+    ${unfocusStyle()}
```
taysea, line 114: Behavior with keyboard and mouse looks good, but just curious why we needed to adjust this at all. What happens if we leave it as it was before?

taysea (commented):
I think the behavior here looks good for both mouse and keyboard. Seems like the only remaining item is the note around a styling block that might not be having any effect on the UI. If we are confident in removing it, we can. otherwise, i might be inclined to leave it for now

jlaluces123:
> I think the behavior here looks good for both mouse and keyboard. Seems like the only remaining item is the note around a styling block that might not be having any effect on the UI. If we are confident in removing it, we can. otherwise, i might be inclined to leave it for now

After running through all of the stories and comparing them to the one currently live, I didn't notice anything UI-wise change or behavior wise. Removed for now.

taysea (approved):

ericsoderberghp (approved):

Issue #5421, opened 2021-07-06
Storybook build failures


ericsoderberghp:
Storybook fails to build many stories

### Expected Behavior

No errors or warnings in the browser console

### Actual Behavior

Many messages like:
![Screen Shot 2021-07-06 at 12 51 38 PM](https://user-images.githubusercontent.com/11637956/124658683-e9724300-de58-11eb-8637-27cde958c2fa.png)

### Steps to Reproduce

Open https://storybook.grommet.io/?path=/story/all--all
Check browser console

ericsoderberghp:
This appears to be restricted to typescript stories.

ericsoderberghp:
Looks like it's related to updating `ts-loader` to v9, which doesn't support webpack v4.

PR #5422, opened 2021-07-06
Downgrade ts-loader to v8.3 since v9 doesn't support webpack v4

ericsoderberghp:
#### What does this PR do?

Downgrade ts-loader to v8.3 since v9 doesn't support webpack v4

package.json

storybook

storybook should have no browser warnings about typescript stories

#5421 

no

no

backwards compatible

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
     "simple-git": "^2.40.0",
     "styled-components": "^5.3.0",
     "tarball-extract": "^0.0.6",
-    "ts-loader": "^9.2.3",
+    "ts-loader": "^8.3.0",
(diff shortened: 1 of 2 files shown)


ericsoderberghp:
I'm wondering how all of the chromatic story changes made it through when we upgraded ts-loader earlier.

jcfilben:
> I'm wondering how all of the chromatic story changes made it through when we upgraded ts-loader earlier.
Looks like this happened when we ran out of chromatic testing snapshots near the end of the month

jcfilben (approved):

halocline (approved):

PR #5423, opened 2021-07-06
Fix DataTable header styling when sortable

taysea:
#### What does this PR do?
When a DataTable header cell is sortable, the padding should be applied to the underlying button instead of the cell. This adjusts the logic to restore that functionality which was lost in https://github.com/grommet/grommet/pull/5355

Tested locally in storybook with hpe theme and sortable table.

Same as above.

Closes https://github.com/grommet/hpe-design-system/issues/1711

No.

No (currently this regression is only living on stable I think).

Backwards compatible.


src/js/components/DataTable/Header.js
```diff
@@ -55,6 +55,9 @@ const buttonStyle = ({ theme }) => {
   const [layoutProps, , iconProps] = separateThemeProps(theme);
 
   if (layoutProps) {
+    // if cell is sortable, we want pad to be applied
+    // to the button instead of the cell
+    layoutProps.pad = theme.dataTable.header.pad;
```
ericsoderberghp, line 60: I think we want to pass `cellProps.pad` to StyledHeaderCellButton. It is already accounting for styling coming from the props and the theme and falling back to theme.table styling as needed.

ericsoderberghp:
It looks like we are applying the pad twice, once to the `<button />` and once to the `<th />`, which looks like a bit too much padding.

Screen Shot 2021-07-06 at 5 57 35 PM

Screen Shot 2021-07-06 at 5 57 54 PM

taysea:
> It looks like we are applying the pad twice, once to the `<button />` and once to the `<th />`, which looks like a bit too much padding.
> 
> Screen Shot 2021-07-06 at 5 57 35 PM
> 
> Screen Shot 2021-07-06 at 5 57 54 PM

Thanks for this -- let me look into it. the goal would be not to do this and just apply once.

ericsoderberghp:
Screen Shot 2021-07-07 at 5 05 12 PM

Controlled story header checkbox doesn't look right here. Same with rowDetails. Otherwise, looking better.

taysea:
> Screen Shot 2021-07-07 at 5 05 12 PM
> 
> Controlled story header checkbox doesn't look right here. Same with rowDetails. Otherwise, looking better.

Thank you -- the remaining snapshot changes are due to the padding being placed internally on the button as opposed to the whole cell. I'm wondering why these were never changed previously but given the intent of the styling, this feels like it's applying things correctly now so maybe the changes are okay?

ericsoderberghp (approved):

PR #5424, opened 2021-07-06
Applying new Prettier defaults to files - batch #5

halocline:
#### What does this PR do?

Batch # 5

Incrementally applying [new Prettier defaults](https://prettier.io/docs/en/options.html#arrow-function-parentheses). Applying in batches for easier review.

Applying remaining Prettier syntax fixes to unblock PRs which are held up from review.

Manual review of syntax changes, plus commit test suite.

Manual review of syntax changes

https://prettier.io/docs/en/options.html#arrow-function-parentheses

no

no

Backwards compatible.

### DIFF
--- a/src/js/components/List/List.js
+++ b/src/js/components/List/List.js
@@ -18,12 +18,12 @@ import {
 
 const StyledList = styled.ul`
   list-style: none;
-  ${props => !props.margin && 'margin: 0;'}
+  ${(props) => !props.margin && 'margin: 0;'}
   padding: 0;
   ${genericStyles}
 
   &:focus {
-    ${props =>
+    ${(props) =>
       props.tabIndex >= 0 &&
       focusStyle({ forceOutline: true, skipSvgChildren: true })}
   }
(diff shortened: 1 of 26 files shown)


src/js/components/List/List.js
```diff
@@ -166,23 +162,25 @@ const List = React.forwardRef(
...
+
     if (active >= 0) {
       let activeId;
       // We have an item that is 'focused' within the list. This could
       // be the list item or one of the up/down ordering buttons.
       // We need to figure out an id of the thing that will be shown as active
       if (onOrder) {
         // figure out which arrow button will be the active one.
-        const buttonId = (active % 2) ? 'MoveDown' : 'MoveUp';
+        const buttonId = active % 2 ? 'MoveDown' : 'MoveUp';
```
taysea, line 177: Interesting that the parentheses are lost in this case, just confirming that's part of the new version rules?

src/js/components/List/List.js
```diff
@@ -232,10 +229,8 @@ const List = React.forwardRef(
             (onClickItem || onOrder) && data && data.length
               ? () => {
                   const min = onOrder ? 1 : 0;
-                  const max = onOrder ? (data.length * 2) - 2 : data.length - 1;
-                  setActive(
-                    active >= min ? Math.min(active + 1, max) : min,
-                  );
+                  const max = onOrder ? data.length * 2 - 2 : data.length - 1;
```
taysea, line 232: seems like removal of parentheses for calculations is consistent across files, but just want to confirm it is expected.

taysea (commented):
I believe all looks good, just confirming one type of change.

src/js/components/List/List.js
halocline: Confirmed. I believe this is done by the `eslint-plugin-prettier` and the rule no-mixed-operators. Lots of folks on Google wondering why their mathematical operations loose their parens...

src/js/components/List/List.js
halocline: Confirmed. I believe this is done by the `eslint-plugin-prettier`.

taysea (approved):

ericsoderberghp (approved):

Issue #4955, opened 2021-02-11
grommet to supports showing a link at the end of table scroll


mgh5:
This is to support displaying a "Load more..." option at the end of table scrolling.  

The current behavior supports infinite scroll; this will be an added feature.  Need the ability to disable the link after it has been clicked a number of times.

jennifer-manriquez:
Hi, I'm new collaborating on OS projects, is anyone already working on this? I would like to help with this enhancement :)

jennifer-manriquez:
Hi again! If it is okay I would like to try to make a new property loadMore in the DataTable component, which would take a boolean value that if true displays a button when the currently available items are rendered. If the button is clicked, then the button would get eliminated and the onMore function could be called allowing the rendering of more items.

ericsoderberghp:
Hi Jennifer! Sorry we missed your comment from a few days ago. This is exactly the right way to start contributing, asking about what the design and API surface might look like and having a discussion on these aspects first.

I'm in favor of the general concept:
- When the user scrolls to the end of the available data, she is presented with a Button. When she clicks the Button, more data is loaded.
- The caller can decide whether to turn this behavior on or not.

I think there are some design details to consider:
- What happens immediately after the Button is pressed but before the new data arrives?
- What if it takes 10 seconds to retrieve the new data, what sort of feedback is given to the user?

I think we should start by working out the design details first and then proceed to the API surface.

jennifer-manriquez:
Hi! Addressing your questions. Maybe there could be a change in the button’s state displaying that the information is being loaded, maybe show an icon, the button could even get disabled in the meantime. I was thinking it could be something like this: 
![DesignIdea2](https://user-images.githubusercontent.com/56976945/123215325-ef961600-d48d-11eb-8ece-7974c5a46b4c.gif)


I would also be glad to know what your take is on those subjects.

ericsoderberghp:
We've tended to not want to change the Button. How about replacing the Button with a Spinner? Here are some visuals: https://designer.grommet.io/?id=DataTable-load-more-eric-soderberg-hpe-com

ericsoderberghp:
For option 1, maybe `onMore={{ function: () => {}, buttonProps: true | {} }}`.
For option 2, maybe `moreButton=true | {}`, which would align a bit closer on the term "more" with "onMore".

I'm inclined more towards the second option as the "on" prefix to "onMore" feels like it value should always be a function.

jennifer-manriquez:
[...]

PR #5427, opened 2021-07-07
DataTable to support showing a button at the end to load more data

jennifer-manriquez:
This is to support displaying a "Load more..." option at the end of DataTable. This addresses [#4955](https://github.com/grommet/grommet/issues/4955). 

#### What does this PR do?

By adding the prop moreButton (which takes a boolean value), this provides a button that retrieves more data when clicked. This is an alternative to loading more data with the infiniteScroll default behavior. This also gives the user a feedback if 10 seconds pass without retrieving data and lets them ask again. 

`yarn test`
Manual test with storybook, story: More Button. 

 DataTable to supports showing a link at the end of table scroll [#4955](https://github.com/grommet/grommet/issues/4955). 
Design aspects discussed in issue. 

Yes. Already updated.

Backwards compatible.

### DIFF
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -9,6 +9,7 @@ import React, {
 } from 'react';
 import { ThemeContext } from 'styled-components';
 import { defaultProps } from '../../default-props';
+import { AnnounceContext } from '../../contexts/AnnounceContext';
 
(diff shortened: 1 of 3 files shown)


jcfilben (commented):
The table headers don't look right in the storybook DataTable examples
Screen Shot 2021-07-07 at 4 14 25 PM

PR #5432, opened 2021-07-09
DataTable to show button at the end to load more

jennifer-manriquez:
This is to support displaying a "Load more..." option at the end of DataTable. This addresses [#4955](https://github.com/grommet/grommet/issues/4955). I had recently submitted a PR, for this issue, however I had inadvertently inserted an error in the headers that I couldn't find easily, so I went back to the parent commit to bring in again my changes without it, I wasn't aware at that point that doing so would close the PR but I see why it is a good measure. Here is the code without that error :) 

#### What does this PR do?

By adding the prop moreButton (which takes a boolean value), this provides a button that retrieves more data when clicked. This is an alternative to loading more data with the infiniteScroll default behavior. This also gives the user a feedback if 10 seconds pass without retrieving data and lets them ask again. 

Run `yarn test` 
Manual test with storybook, story: More Button. 

Manual test with storybook. Change load more function to take longer than 10 seconds to view the unretrieved data state.  

 DataTable to supports showing a link at the end of table scroll [#4955](https://github.com/grommet/grommet/issues/4955). 
Design aspects discussed in issue. 

Yes. Already updated.

Backwards compatible.


vavalos5:
I think this approach is definitely the way to go! Especially since we like to keep mobile first approach in mind. A big button at the bottom of a table can be easily accessible to a user on a mobile device.

Also, [after reading this article](https://uxdesign.cc/ui-cheat-sheet-pagination-infinite-scroll-and-the-load-more-button-e5c452e279a8) it confirms that using one of these in the button is the most common:  ‘Load more’, ‘Show more results’ or ‘More results’.

Great job! 🎉

ericsoderberghp:
I'm wondering if putting the button below the footer is the right approach. For instance, in this situation: https://storybook.grommet.io/?path=/story/visualizations-datatable-fill-and-pin--fill where the table body is what's scrolling but the header and footer remain pinned. In that scenario, I think I would expect the load more button to be at the bottom of the table body but above the footer.

vavalos5:
Something I really dislike about some FileInputs is when there is not type of description of conceptual help text that tells me prior to attaching files that I can only attach "5" files, only to receive an error message after. I. think for this case it would be good to have contextual help below the FileInput like "Attach a maximum of 5 files" as we would see "File maximum size is 32MB" for example. Now if the user surpasses that number, then an error would pop up below indicating the problem: "Exceeds maximum of 5 files. Please select a maximum of 5" ; the 6th item would not be attached, but it would let the user know again that only 5 could be selected and they can revise the ones that have already been attached.

jennifer-manriquez:
Hi! I made a recent commit, in which I placed the button above the footer. Also I tested how it would look on the other stories with these results. In the last one you can see the styled story, when onMore is declared, the render marker is visible as a white space. 
NewGrommetTests. 
I tried taking it off, but then this approach wouldn't work as it needs the render marker, or other modifications should be done that might be a little more from what I can do at this point.

src/js/components/DataTable/DataTable.js
```diff
@@ -314,6 +320,80 @@ const DataTable = ({
...
+          label="load more"
+          onClick={() => setMoreButtonState('spinner')}
+        />
+      );
+    }
+    if (moreButton && moreButtonState === 'spinner') {
+      return (
+        <Box gap="small" direction="row">
+          <Spinner />
+          <Text weight="bold">Loading...</Text>
```
taysea, line 377: Do we want the text of "Loading..."? What if it is desired to just have a spinner. Would it be better for this text to be drive my a `messages` prop or something? If so, if the message was undefined then the text would not render

src/js/components/DataTable/DataTable.js
```diff
@@ -314,6 +320,80 @@ const DataTable = ({
...
+        </Box>
+      );
+    }
+    if (moreButton && moreButtonState === 'unretrieved') {
+      return (
+        <>
+          <Text> Retrieving data is taking too long. Try again.</Text>
```
taysea, line 384: Similar comment as above, can we make this message customizable for the caller?

src/js/components/DataTable/DataTable.js
```diff
@@ -314,6 +320,80 @@ const DataTable = ({
...
+  }, [data.length]);
+
+  if (moreButton === true && !onMore) {
+    console.warn(
+      'You need to provide an function in the onMore field to use moreButton',
+    );
+  }
+
+  const memoizedmoreButtonState = useMemo(() => {
+    if (moreButton && moreButtonState === 'button') {
```
taysea, line 365: Feels like having all of these button states handled as strings introduces opportunity for things to break. What about creating some constant and referencing here?

```
const moreButtonStates = {
   button: 'button',
   spinner: 'spinner',
   unretrieved: 'unretrieved'
}

...

moreButton && moreButtonState === moreButtonStates.button
```

src/js/components/DataTable/DataTable.js
```diff
@@ -314,6 +320,80 @@ const DataTable = ({
...
+      );
+    }
+    if (moreButton && moreButtonState === 'unretrieved') {
+      return (
+        <>
+          <Text> Retrieving data is taking too long. Try again.</Text>
+          <Button
+            label="load more"
+            margin="small"
+            display="block"
```
taysea, line 388: Why is display "block" needed?

src/js/components/DataTable/DataTable.js
```diff
@@ -314,6 +320,80 @@ const DataTable = ({
...
+          <Text weight="bold">Loading...</Text>
+        </Box>
+      );
+    }
+    if (moreButton && moreButtonState === 'unretrieved') {
+      return (
+        <>
+          <Text> Retrieving data is taking too long. Try again.</Text>
+          <Button
+            label="load more"
```
taysea, line 386:
```suggestion
            label="Load More"
```

src/js/components/DataTable/DataTable.js
```diff
@@ -413,6 +494,30 @@ const DataTable = ({
               setRowExpand={setRowExpand}
             />
           )}
+          {onMore && moreButton ? (
+            <StyledDataTableBody>
```
taysea, line 498: Is it correct for this to introduce another TableBody inside the same table? Or should this just be added as a row in the Body?

taysea (commented):
Awesome work on this! Just left a few comments.

jcfilben:
Closing for now, but feel free to re-open this with the feedback addressed

Issue #5321, opened 2021-06-14
Allow caller to add a custom `onClick` function  to FileInput remove file button


taysea:
### Expected Behavior

Allow a user to provide some custom `onClick` handler to the remove button of FileInput, such as "onClick, open a confirmation dialog about removing the file". This would be useful in cases where file uploads/validations take a long time and removing the file automatically could cause time consuming work to need to occur again.

### Actual Behavior

No way to access any extra handler functions for the file remove button.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: stable
- Browser Name and version: chrome mac os
- Operating System and version (desktop or mobile): mac os mojave

DouglasPds:
Hi, Can I try to implement this?

jcfilben:
Closing since #5711 was merged and allows you to have a custom function for confirming removal of a file with the `confirmRemove` prop

PR #5433, opened 2021-07-12
Adding a prop to the InputFile to pass a function to the remove button in onClick

DouglasPds:
#### What does this PR do?
The new prop in FileInput receives a function and passes it to be executed in onClick in the remove button.

FileInput.js

Passing a function to onClickRemove, and see that be executed when clicking on the remove button

Some doubts. The function passed to the remove button need to override the onClick that removes the file? Or be executed inside the onClick with the other commands?

#5321 

Yes

Is compatible

### DIFF
--- a/src/js/components/FileInput/FileInput.js
+++ b/src/js/components/FileInput/FileInput.js
@@ -67,6 +67,7 @@ const FileInput = forwardRef(
       multiple,
       name,
       onChange,
+      onClickRemove,
       pad,
       value: valueProp,
       ...rest
@@ -247,6 +248,7 @@ const FileInput = forwardRef(
                 icon={<RemoveIcon />}
                 hoverIndicator
                 onClick={(event) => {
+                  if (onClickRemove) onClickRemove();
                   event.stopPropagation();
                   if (onChange) onChange(event, { files: [] });
                   setFiles([]);
@@ -328,6 +330,7 @@ const FileInput = forwardRef(
                   icon={<RemoveIcon />}
                   hoverIndicator
                   onClick={(event) => {
+                    if (onClickRemove) if (!onClickRemove()) return;
                     event.stopPropagation();
                     const nextFiles = [...files];
                     nextFiles.splice(index, 1);
--- a/src/js/components/FileInput/stories/Simple.js
+++ b/src/js/components/FileInput/stories/Simple.js
@@ -3,11 +3,16 @@ import React from 'react';
 import { Box, Grommet, FileInput } from 'grommet';
 import { grommet } from 'grommet/themes';
 
+function confirmationDialog() {
+  return window.confirm('Do you really want to delete this file');
+}
+
 export const Simple = () => (
   <Grommet full theme={grommet}>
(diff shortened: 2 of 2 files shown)


src/js/components/FileInput/FileInput.js
```diff
@@ -328,6 +330,7 @@ const FileInput = forwardRef(
                   icon={<RemoveIcon />}
                   hoverIndicator
                   onClick={(event) => {
+                    if (onClickRemove) if (!onClickRemove()) return;
```
wenderjean, line 333: It should be the same as `if (onClickRemove && !onClickRemove()) return;`? :)

src/js/components/FileInput/FileInput.js
```diff
@@ -247,6 +248,7 @@ const FileInput = forwardRef(
                 icon={<RemoveIcon />}
                 hoverIndicator
                 onClick={(event) => {
+                  if (onClickRemove) onClickRemove();
```
hjJunior, line 251: WDUT passing the file to `onClickRemove` function? So the callback can decides according with the file

src/js/components/FileInput/FileInput.js
```diff
@@ -67,6 +67,7 @@ const FileInput = forwardRef(
       multiple,
       name,
       onChange,
+      onClickRemove,
```
hjJunior, line 70: Suggestion lets rename it and call as `canRemove` or something like this, where can accept a boolean or a function that returns a boolean

taysea: Recommending `onRemove` instead. I think just accepting a function is sufficient.

src/js/components/FileInput/FileInput.js
taysea: Each `onRemove` would be specific to an individual file. However, I could see there being value to having access to that file data.

To be consistent with other handlers I think it would be appropriate to call the onRemove with a signature of something like:
```
const adjustedEvent = event;
adjustedEvent.file = file;
onRemove(adjustedEvent)
```

src/js/components/FileInput/FileInput.js
ericsoderberghp: I think the function will want to know more about which file is being removed.

ericsoderberghp, line 333: I think we should also enable working with other forms of confirmation, such as opening a Layer. In that case, perhaps the call signature could be something like: `onRemove(file, () => { ... do the remove stuff ... })`

taysea: @DouglasPds will this functionality be possible with your current implementation? Or will updates to the code be needed

taysea:
Closing this PR because we have implemented `confirmRemove` prop instead: https://github.com/grommet/grommet/pull/5711

The use case that was driving this enhancement centered around being able to render some kind of confirmation to the user when they attempt to remove a file. To be more specific to this need, we've added `confirmRemove` prop that allows caller to decide how to render this confirmation.

PR #5434, opened 2021-07-12
fix(Header): aligns ref support with Box

Tbhesswebber:
#### What does this PR do?

This PR adds support for forwarding refs to the Box component (which forwards refs itself)

`/components/Header/Header.js`

I can add tests if necessary

Run `yarn storybook` and then expand the `Header` section.  Click on `Ref` and click the `Change Color` button.

In our codebase, we have primary, secondary, and tertiary headers.  Each wraps this header, but under certain circumstances, we need to be able to perform various updates to them (e.g. - when another header is present the primary disappears on scroll to free up screen real estate, etc)

N/A

No.  This makes the statement of "a Box with the following properties" more accurate

No

Backwards compatible

### DIFF
--- a/src/js/components/Header/stories/typescript/Ref.tsx
+++ b/src/js/components/Header/stories/typescript/Ref.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+
+import { Avatar, Anchor, Nav, Grommet, Header, Button } from 'grommet';
+import { grommet } from 'grommet/themes';
+
+const gravatarLink =
+  '//s.gravatar.com/avatar/b7fb138d53ba0f573212ccce38a7c43b?s=80';
+
+export const Ref = () => {
+  const ref = React.createRef<HTMLDivElement>();
+  const [color, setColor] = React.useState('');
+
+  const setRandomColor = () => {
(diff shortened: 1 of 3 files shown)


src/js/components/Header/stories/typescript/Ref.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+const gravatarLink =
+  '//s.gravatar.com/avatar/b7fb138d53ba0f573212ccce38a7c43b?s=80';
+
+export const Ref = () => {
+  const ref = React.createRef<HTMLDivElement>();
+  const [color, setColor] = React.useState('');
+
+  const setRandomColor = () => {
+    const red = Math.floor(Math.random() * 255).toString(16);
```
MikeKingdom, line 14: We should avoid random since it makes hard to have any sort of automated test for storybook.

MikeKingdom (commented):
@Tbhesswebber Thanks for adding this.

Could you add a jest test case for this?

Tbhesswebber:
@MikeKingdom - I goofed a git command and accidentally deleted this branch.  New PR is https://github.com/grommet/grommet/pull/5439

PR #5444, opened 2021-07-14
Allow aria-label to be used where a11yTitle is accepted

taysea:
#### What does this PR do?
Leverages work started in #5137 but resolves all merge conflicts. Adds tests where appropriate.

Added tests and reviewed snapshots.

Add `aria-label` to a component such as Menu. Should see `aria-label` in the DOM.

No.

Yes, improved support for `aria-label`

Backwards compatible.

### DIFF
--- a/src/js/components/Chart/Chart.js
+++ b/src/js/components/Chart/Chart.js
@@ -19,7 +19,6 @@ const Chart = React.forwardRef(
   (
     {
       a11yTitle,
-      'aria-label': ariaLabel,
       bounds: propsBounds,
       color,
       dash,
@@ -648,7 +647,7 @@ const Chart = React.forwardRef(
       <StyledChart
         ref={containerRef}
         id={id}
-        aria-label={ariaLabel || a11yTitle}
+        aria-label={a11yTitle}
         viewBox={viewBox}
         preserveAspectRatio="none"
         width={size === 'full' ? '100%' : size[0]}
--- a/src/js/components/Chart/Chart.js
+++ b/src/js/components/Chart/Chart.js
@@ -19,6 +19,7 @@ const Chart = React.forwardRef(
   (
     {
       a11yTitle,
+      'aria-label': ariaLabel,
       bounds: propsBounds,
       color,
       dash,
@@ -647,7 +648,7 @@ const Chart = React.forwardRef(
       <StyledChart
         ref={containerRef}
         id={id}
-        aria-label={a11yTitle}
+        aria-label={ariaLabel || a11yTitle}
         viewBox={viewBox}
         preserveAspectRatio="none"
         width={size === 'full' ? '100%' : size[0]}
--- a/src/js/components/DataChart/__tests__/__snapshots__/DataChart-test.js.snap
+++ b/src/js/components/DataChart/__tests__/__snapshots__/DataChart-test.js.snap
@@ -6971,34 +6971,6 @@ exports[`DataChart pad 1`] = `
 </div>
 `;
 
-exports[`DataChart renders a11yTitle and aria-label 1`] = `
-.c0 {
-  font-size: 18px;
-  line-height: 24px;
-  box-sizing: border-box;
-  -webkit-text-size-adjust: 100%;
-  -ms-text-size-adjust: 100%;
-  -moz-osx-font-smoothing: grayscale;
-  -webkit-font-smoothing: antialiased;
-}
-
-.c1 {
-  position: relative;
-  grid-area: charts;
-}
-
-<div
-  class="c0"
->
-  <div
-    class="c1"
-  />
-  <div
-    class="c1"
-  />
-</div>
-`;
-
 exports[`DataChart size 1`] = `
 .c0 {
   font-size: 18px;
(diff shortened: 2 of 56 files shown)


src/js/components/Image/Image.js
```diff
@@ -2,9 +2,21 @@ import React, { forwardRef, useState } from 'react';
...
+  (
+    {
+      a11yTitle,
+      'aria-label': ariaLabel,
+      fallback,
+      onError,
+      opacity,
+      fill,
+      src,
+      ...rest
```
jcfilben, line 14: Since we have `...rest` here do we need to explicitly have aria-label?

src/js/components/MaskedInput/MaskedInput.js
```diff
@@ -161,6 +161,7 @@ const MaskedInput = forwardRef(
   (
     {
       a11yTitle,
+      'aria-label': ariaLabel,
```
jcfilben, line 164: MaskedInput also has `...rest` so wondering if we need `aria-label` here

src/js/components/RadioButton/RadioButton.js
```diff
@@ -18,6 +18,7 @@ const RadioButton = forwardRef(
   (
     {
       a11yTitle,
+      'aria-label': ariaLabel,
```
jcfilben, line 21: RadioButton has `...rest`, same comment as above

src/js/components/Text/Text.js
```diff
@@ -12,6 +12,7 @@ const Text = forwardRef(
       tip,
       // can't alphabetize a11yTitle before tip is defined
       a11yTitle = typeof tip === 'string' ? tip : undefined,
+      'aria-label': ariaLabel,
       ...rest
     },
```
jcfilben, line 17: Text has `...rest`, same comment as above

src/js/components/TextArea/TextArea.js
```diff
@@ -9,6 +9,7 @@ const TextArea = forwardRef(
   (
     {
       a11yTitle,
+      'aria-label': ariaLabel,
```
jcfilben, line 12: TextArea has `...rest`

src/js/components/TextInput/TextInput.js
```diff
@@ -69,6 +69,7 @@ const TextInput = forwardRef(
   (
     {
       a11yTitle,
+      'aria-label': ariaLabel,
```
jcfilben, line 72: TextInput has `...rest`

jcfilben (commented):
Nice job, I just added some comments about not needing to have `aria-label` explicitly added if the component already has `...rest`

taysea:
> Nice job, I just added some comments about not needing to have `aria-label` explicitly added if the component already has `...rest`

Thanks! Sometimes the ...rest happened before the `aria-label` so it was needed, but in the cases you mentioned it is not. reverting these changes but leaving the tests to make sure they do carry through to DOM.

src/js/components/DataChart/__tests__/__snapshots__/DataChart-test.js.snap
```diff
@@ -6971,6 +6971,34 @@ exports[`DataChart pad 1`] = `
...
+
+.c1 {
+  position: relative;
+  grid-area: charts;
+}
+
+<div
+  class="c0"
+>
+  <div
```
ericsoderberghp, line 6993: Where's the `aria-label`?

ericsoderberghp:
Can we use something like https://testing-library.com/docs/queries/byrole/ to explicitly validate the attribute we expect, instead of just relying on snapshots?

src/js/components/DataChart/__tests__/__snapshots__/DataChart-test.js.snap
taysea: hmmm seems like something isn't applying correctly to DataChart. maybe we can look at that separately from this PR.

src/js/components/Grid/__tests__/Grid-test.js
```diff
@@ -16,10 +16,11 @@ describe('Grid', () => {
...
 
-  test('a11yTitle renders', () => {
+  test('a11yTitle and aria-label renders', () => {
     const { container, getByLabelText } = render(
       <Grommet>
         <Grid a11yTitle="My Grid" />
+        <Grid aria-label="My Other Grid" />
       </Grommet>,
     );
     const gridWithLabel = getByLabelText('My Grid');
```
jcfilben, line 26: I think we should add another assertion for "My Other Grid"

src/js/components/DataChart/__tests__/__snapshots__/DataChart-test.js.snap
ericsoderberghp: If we remove DataChart from this PR, I'm OK with separating it. I'd rather not go forward with a snapshot we know isn't right.

jcfilben (approved):

src/js/components/Chart/Chart.js
```diff
@@ -19,7 +19,6 @@ const Chart = React.forwardRef(
   (
     {
       a11yTitle,
-      'aria-label': ariaLabel,
```
jcfilben: Oops, looks like you removed the aria-label on Chart instead of DataChart

taysea: Thanks, I noticed chart had `...rest` so I removed and I've included the DataChart stuff on my latest push.

ericsoderberghp (approved):

PR #5452, opened 2021-07-16
Refactored DataTable GroupedBody.js to use InfiniteScroll

### LINKED ISSUE #5283: DataTable groupBy should use InfiniteScroll
### Expected Behavior

groupBy should leverage InfiniteScroll so grouped data isn't all rendered at once.

### Actual Behavior

InfiniteScroll is not being used with grouped data. Therefore, when a group has many items (such as 500 in a group), the application is slowed down.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: stable
- Browser Name and version: macos chrome
- Operating System and version (desktop or mobile): mac os

MikeKingdom:
DataTables normally render their table body within an `InfiniteScroll`. However if they have a `groupBy` attribute, the table body is not within an `InfiniteScroll`.

#### What does this PR do?
Refactors the `GroupedBody` component in `DataTable` to put their body content within and `InfiniteScroll` so they are more scalable, especially for large groups of rows which are currently hidden or simply below the current scroll position.

This required refactoring GroupedBody to generate the set of group headers and items ahead of time so they could be cleanly handed to InfiniteScroll. I also took this opportunity to combine a lot of duplicated code for rendering the rows rather than having separate copies for group header rows and group member/body rows.
 
I discovered a few addition issues with DataTable that were pre-existing and might deserve attention:
- Changing the data causes the groupState to be regenerated because 'groups' got updated in DataTable. This means if you try to use `groupBy={ {onExpand: ().... } }`  to lazy-add the items in the group when the user expands it, it won't expand on first click. There really isn't a good way to support this with the current implementation, especially since the group won't have an expander unless at least 2 items were already in the group in the initial data.
- if a column is 'aggregate' and not all data is in the data attribute (e.g. some was added lazily by onExpand or onMore) the aggregate column won't include those items in its calculation.

GroupedBody.js

Storybook and jest tests.

Use storybook to try the various GroupBy examples for DataTable.

Fixes #5283 

No

Yes

Backwards compatible


ericsoderberghp (approved):
Looks good. I checked it out locally and changed the story to use `data` with 400 records and it seemed to be working well. Nice to see no snapshot changes.

MikeKingdom:
@ericsoderberghp Thanks!

PR #5453, opened 2021-07-16
Add tip as value for truncate on Text

taysea:
#### What does this PR do?

When truncate="tip", if the text has been truncated, the entirety of the text will be placed in a Tip.

src/js/components/Text/Text.js

Adjusted Tip.js story to demo `truncate="tip"`.

Checkout deploy Tip.js under Text stories.

Minimize implementation effort to be able to place truncated content into a tip.

Related to: https://github.com/grommet/hpe-design-system/issues/1503
Related to: https://github.com/grommet/hpe-design-system/pull/1734

Updated here.

Yes. Added `truncate="tip"` to Text.

Backwards compatible.

### DIFF
--- a/src/js/components/Text/README.md
+++ b/src/js/components/Text/README.md
@@ -213,10 +213,13 @@ string
 
 Restrict the text to a single line and truncate with ellipsis if it
 is too long to all fit. For truncate to be applied, Text needs to be 
-contained within a layout component (such as Box or a generic div).
+contained within a layout component (such as Box or a generic div). If 
+truncate = tip, the full text content will be placed in a Tip that will appear 
+on hover.
 
 ```
 boolean
+tip
 ```
 
 **weight**
--- a/src/js/components/Text/Text.js
+++ b/src/js/components/Text/Text.js
@@ -1,29 +1,51 @@
-import React, { forwardRef } from 'react';
+import React, { forwardRef, useEffect, useState } from 'react';
 
 import { StyledText } from './StyledText';
 import { Tip } from '../Tip';
+import { useForwardedRef } from '../../utils';
 
 const Text = forwardRef(
   (
     {
+      children,
       color,
       tag,
       as,
-      tip,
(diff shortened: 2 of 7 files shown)


ericsoderberghp (approved):

ericsoderberghp:
Need to look into why the Button:Badge chromatic test is failing. I've run it twice and it fails in the same place.

MikeKingdom:
> Need to look into why the Button:Badge chromatic test is failing. I've run it twice and it fails in the same place.

Seems like it has something to do with how the ref if being passed around. Badge seems like it's miscalculating contentRef width

ericsoderberghp:
> > Need to look into why the Button:Badge chromatic test is failing. I've run it twice and it fails in the same place.
> 
> Seems like it has something to do with how the ref if being passed around. Badge seems like it's miscalculating contentRef width

I think it's because the Chromatic snapshot is being taken before the `useLayoutEffect` is running the initial `onResize` to stretch the badge if needed. Is there a way to delay the snapshot?

ericsoderberghp:
Might want to try something like:
```
Badge.parameters = {
  chromatic: { delay: 50 }, // 50 is empirical
};
```

MikeKingdom:
> Might want to try something like:
> 
> ```
> Badge.parameters = {
>   chromatic: { delay: 50 }, // 50 is empirical
> };
> ```

I can still make it happen locally. There are cases where nothing causes another render and no amount of waiting will help since nothing is coming.

Basically, useForwardedRef sets the .current of the passed in ref (in this case Badge's copy) in a useEffect which is called after Badge's useLayoutEffect. If nothing causes Text to get re-rendered then Badge won't ever see the real .current.

Making Badge use useEffect there instead does work, but that basically just points out the way useForwardedRef no works in Text is not backward compatible with anyone else passing a ref to Text.

MikeKingdom:
Replaced by PR #5462

Issue #5320, opened 2021-06-14
Add `maxSize` prop to FileInput


taysea:
### Expected Behavior

Allow caller to specify a `maxSize` for the file being uploaded. Some things to consider would be to allow, in a form, a validation message to be presented to the user such as "File must be < 2MB." (or some provided validation). Demo here of what that might look like: https://codesandbox.io/s/fileinput-max-size-85sv9?file=/index.js

### Actual Behavior

Currently, there is no simple way for a caller to limit the accepted file size in FileInput.

### URL, screen shot, or Codepen exhibiting the issue

Demo of what the interaction might look like to limit file size and provide some feedback to caller when in a form.

https://codesandbox.io/s/fileinput-max-size-85sv9?file=/index.js
### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: stable
- Browser Name and version: mac os chrome
- Operating System and version (desktop or mobile): mac os mojave

PR #5457, opened 2021-07-19
Max Size prop added, story changes, and docs updated for new prop

jlaluces123:
#### What does this PR do?
Adds the `maxSize` prop to `FileInput`, allowing users to create a limit on how large the files given to `FileInput` can be.

- Check out the `maxSize` prop at  `index.d.ts` file for `FileInput` and its `doc.js`
- `yarn storybook` and navigate to FileInput's `Multiple` or `Simple` stories
- (Work in Progress): see how `maxSize` can be used in the code snippets of those stories and how its error messages work

Manual testing

- `yarn storybook` and navigate to FileInput's `Multiple` or `Simple` stories
- Add a file that is bigger than `maxSize` on the story
- (Work In Progress): see the error that pops up when `maxSize` is exceeded

Issue #5320 

Yes, updates to adding `maxSize` prop to `FileInput`

No

Will be backwards compatible


src/js/components/FileInput/FileInput.js
```diff
@@ -120,6 +128,20 @@ const FileInput = forwardRef(
       }
     }
 
+    const parseBytes = (bytes) => {
```
ericsoderberghp, line 131: I think we should call this `normalizeBytes` and pull it outside of the component, perhaps around line 18.

ericsoderberghp: On second thought, maybe we should omit the object value to start with. We can always add it in later as needed.

src/js/components/FileInput/FileInput.js
```diff
@@ -85,6 +86,13 @@ const FileInput = forwardRef(
     const removeRef = useRef();
     const RemoveIcon = theme.fileInput.icons.remove;
 
+    const { error, info } = formContext.useFormField({
```
ericsoderberghp, line 89: I think this wants to be something like:
```
formContext.useFormField(maxSize ? { name, validate: (fileList) => {
  for (let i = 0; i < fileList.length; i += 1) {
    const file = fileList[i];
    if (file.size > maxSize) {
      return `${file.name} is too large`;
    }
  }
  return undefined;
}} : {});
```

src/js/components/FileInput/FileInput.js
jlaluces123: Gotcha, I'll leave it commented out where you recommended and down the line we can reintroduce it!

src/js/components/FileInput/doc.js
```diff
@@ -18,6 +18,13 @@ export const doc = (FileInput) => {
       .description('Whether the control is disabled.')
       .defaultValue(undefined),
     id: PropTypes.string.description('The id attribute of the input.'),
+    maxSize: PropTypes.oneOfType([
+      PropTypes.number,
+      PropTypes.shape({
+        unit: PropTypes.string,
+        size: PropTypes.number,
+      }),
+    ]).description('Specifies a file size limit for the FileInput'),
```
jcfilben, line 27: Doc should be updated on grommet-site repo, see https://github.com/grommet/grommet-site/pull/248

src/js/components/FileInput/FileInput.js
```diff
@@ -85,6 +87,46 @@ const FileInput = forwardRef(
...
+    formContext.useFormField(
+      maxSize
+        ? {
+            name,
+            validate: (fileList) => {
+              let numOfFiles = 0;
+              let message;
+              for (let i = 0; i < fileList.length; i += 1) {
+                const file = fileList[i];
+                if (file.size > maxSize) {
```
MikeKingdom, line 99: maxSize id documented as a number or object which specifies a number and a string for units. This seems to be expecting a number of bytes. How does the object get converted to bytes? How will it appear in validation messages then?

src/js/languages/default.json
```diff
@@ -8,6 +8,11 @@
     "dropPrompt": "Drop file here or",
     "dropPromptMultiple": "Drop files here or",
     "files": "files",
+    "maxSizeSingle": "File exceeds size limit. Select a file with max size of {maxSize}.",
+    "maxSizeMultiple": {
+      "singular": "1 File exceeds size limit. Select files with a max size of {maxSize}.",
+      "plural": "{numOfFiles} Files exceed size limit. Select files with a max size of {maxSize}."
```
MikeKingdom, line 14: I used `largeFiles` above but if you stick with this, drop the Of (ie. `numFiles`).
Also these are very computer-ish sounding and *max* isn't a word.  How about something more like:
```
"maxSizeSingle": "The file is too large. Select a file no larger than {maxSize} bytes.",
"maxSizeMultiple": {
   "singular": "One file is too large. Select files which are no larger than {maxSize} bytes.",
    "plural": "{numOfFiles} files are too large. Select files which are no larger than {maxSize} bytes.`
},
```
The units and value for maxSize may need to be more friendly depending on if they can specify units with the value.

src/js/components/FileInput/FileInput.js
```diff
@@ -312,10 +368,15 @@ const FileInput = forwardRef(
...
                     theme.global.input.weight || theme.global.input.font.weight
                   }
-                  truncate
                   {...theme.fileInput.label}
+                  alignSelf="center"
                 >
-                  {file.name}
+                  <Box direction="row">
+                    {maxSize && file.size > maxSize ? <CircleAlert /> : null}
+                    <Text truncate margin={{ left: 'xsmall' }}>
```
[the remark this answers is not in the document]
jlaluces123: Ahh gotcha, yes in this code chunk it's definitely still up in the air since I will be implementing something like this once a submit happens instead of just having a conditional render here. But I will get this implemented for now with `gap="xsmall"` thanks for the heads up 👍

src/js/components/FileInput/FileInput.js
jlaluces123: Down the line and in response to the `PropTypes` comment as well, currently this will only accept maxSize in bytes but down the line I will be looking into adding a more verbose way to specify maxSize to look something like:

```js
const maxSize = {{
  unit: 'mb',
  size: '5'
}}
```

But if in our first pass we won't be looking to go into it I would be removing the human-readable format for byte parsing before anything else gets published later.

src/js/languages/default.json
jlaluces123: I definitely liked the naming of `largeFiles` instead so this will be getting implemented with the next push. Thanks for the help on this!

ericsoderberghp:
So, I took a look at how the current Form validations work and have this proposal for how to refactor to accommodate the changes needed in this pull request:
- change `useFormInput` to take an object argument, add `validate` to it
- change `validations[name] = func` to be `validations[name].field = func` and add `validations[name].input = func`
- change `validate()` in Form.js and other places to understand new structure to `validations[name]`
Let's try this approach and see how it holds up.

ericsoderberghp:
Might be interesting to review the API surface and behavior for alignment with and possible use of https://developer.mozilla.org/en-US/docs/Web/API/notification

src/js/languages/default.json
```diff
@@ -8,6 +8,11 @@
     "dropPrompt": "Drop file here or",
     "dropPromptMultiple": "Drop files here or",
     "files": "files",
+    "maxSizeSingle": "The file is too large. Select a file no larger than {maxSize} bytes.",
```
taysea, line 11: Maybe an enhancement for later but any way to convert bytes into MB, GB, etc. where applicable since a large number of bytes may be unfamiliar to users especially since file browsers convert the values

src/js/components/__tests__/__snapshots__/README-test.js.snap
```diff
@@ -8479,6 +8479,14 @@ The id attribute of the input.
 string
 \`\`\`
 
+**maxSize**
+
+Specifies a file size limit for the FileInput
```
taysea, line 8484:
```suggestion
Specifies a file size limit in bytes for the FileInput
```

taysea: ^^ Updates to docs should be filed on grommet-site now, but leaving this comment here as a suggestion to the wording

src/js/languages/default.json
jlaluces123: Absolutely agree. I know we discussed an enhancement being the ability to specify our `maxSize` in some sort of shape that would look like:

```js
const maxSize = {
  unit: 'MB',
  size: 5
} // ==> 5MB
```

But I will try and see if I can implement a function to parse the current bytes implementation into a more readable format inside the message for now.

MikeKingdom: I would recommend just always specifying the maxSize in bytes and let the code decide how to best format it for display.

jlaluces123: I implemented a function to parse the bytes into a readable format and adjusted the messaging accordingly.

src/js/components/FileInput/FileInput.js
```diff
@@ -85,6 +86,48 @@ const FileInput = forwardRef(
     const removeRef = useRef();
     const RemoveIcon = theme.fileInput.icons.remove;
 
+    const formatBytes = (size) => {
+      const units = ['B', 'KB', 'MB', 'GB', 'TB'];
+      const factor = 1000;
+      let index = 0;
+      let num = size;
+      while (num >= factor) {
```
MikeKingdom, line 94: Probably should add a check for `index < units.length - 1` to be safe, however unlikely 1 PB+ is.

src/js/components/FileInput/FileInput.js
```diff
@@ -85,6 +86,48 @@ const FileInput = forwardRef(
     const removeRef = useRef();
     const RemoveIcon = theme.fileInput.icons.remove;
 
+    const formatBytes = (size) => {
```
MikeKingdom, line 89: this doesn't depend on any outside variable/parameters so it should be moved above the component declaration

src/js/components/RangeInput/RangeInput.js
```diff
@@ -10,7 +10,10 @@ const RangeInput = forwardRef(
   ) => {
     const formContext = useContext(FormContext);
 
-    const [value, setValue] = formContext.useFormInput(name, valueProp);
+    const [num, setNum] = formContext.useFormInput({
```
MikeKingdom, line 13: Why not keep this `[value, setValue]`?

MikeKingdom (commented):
Getting closer!

src/js/components/RangeInput/RangeInput.js
jlaluces123: Ah I believe this is a mistake on my part when I was debugging the hook changes, will revert this back!

MikeKingdom (commented):
Thanks Jonathan. It's looking good.

Were you still planning on fixing the gap between the circle icon and filename?
There are some issues with how longer file names overflow which I'm guessing is pre-existing?

jlaluces123:
> Thanks Jonathan. It's looking good.
> 
> Were you still planning on fixing the gap between the circle icon and filename?
> There are some issues with how longer file names overflow which I'm guessing is pre-existing?

Awesome, glad it's really coming along now! I will be going over the UI changes but if memory serves me correct, I believe the truncation of longer filenames was working before my changes so I will get that looked at as well as the gap between the circle icons and filenames!

MikeKingdom (approved):
Looks great. Thanks!

src/js/components/FileInput/FileInput.js
```diff
@@ -86,6 +99,36 @@ const FileInput = forwardRef(
     const removeRef = useRef();
     const RemoveIcon = theme.fileInput.icons.remove;
 
+    const [files, setFiles] = formContext.useFormInput({
+      name,
+      value: valueProp,
+      initialValue: [],
+      validate: () => {
```
ericsoderberghp, line 106: I think we can be a bit leaner and do something like: `validate: maxSize ? () => {...} : undefined`

src/js/components/FileInput/FileInput.js
```diff
@@ -86,6 +99,36 @@ const FileInput = forwardRef(
...
+        const fileList = [...files];
+        let message = '';
+        if (maxSize) {
+          const numOfFiles = fileList.filter(
```
ericsoderberghp, line 110: Perhaps renaming "numOfFiles" to "numOfInvalidFiles" would be a bit clearer.

src/js/components/FileInput/FileInput.js
```diff
@@ -309,15 +352,24 @@ const FileInput = forwardRef(
...
-                  truncate
+                <Box
                   {...theme.fileInput.label}
+                  gap="xsmall"
+                  align="center"
+                  direction="row"
                 >
-                  {file.name}
-                </Label>
+                  {maxSize && file.size > maxSize ? <CircleAlert /> : null}
```
ericsoderberghp, line 361:
```suggestion
                  {maxSize && file.size > maxSize && <CircleAlert />}
```

src/js/components/FileInput/FileInput.js
```diff
@@ -309,15 +352,24 @@ const FileInput = forwardRef(
...
+                  <Label
+                    weight={
+                      theme.global.input.weight ||
+                      theme.global.input.font.weight
+                    }
+                    alignSelf="center"
```
ericsoderberghp, line 367: I don't think we need this `alignSelf="center"`

ericsoderberghp (approved):

Issue #5221, opened 2021-05-05
CheckBox and CheckBoxGroup should accept children


Tbhesswebber:
In terms of intent, a raw HTML checkbox group is a radio group that allows multiple selections. Given this, the API for the two should be the same when it comes to rendering.

### Expected Behavior

In our forms, we need non-dropdown interfaces for multiple and single selections - i.e. checkbox groups and radio groups.  To support this, we have the following component for the radio button (it's not a radio group because we have a custom radio group that uses a Grid rather than a Box).

```ts
interface RadioChipProps extends Omit<RadioButtonProps, "label"> {
  label: string;
}

function RadioChip({
  name,
  label,
  ...radioProps
}: RadioChipProps): JSX.Element {
  return (
    <RadioButton name={name} {...radioProps} value={label}>
      {({ checked }: { checked: boolean }) => (
        <Chip
          background={!checked ? Colors.white : Colors.primary}
          label={label}
          fill
        />
      )}
    </RadioButton>
  );
}
```

I didn't reference the docs before then writing the following for a CheckBox, which I expected to work with some minor tweaks once I did reference the docs.

```ts
interface CheckBoxChipProps extends Omit<CheckBoxProps, "label"> {
  label: string;
}

function CheckBoxChip({
  name,
  label,
  ...checkBoxProps
}: CheckBoxChipProps): JSX.Element {
  return (
    <CheckBox name={name} {...checkBoxProps}>
      {({ checked }: { checked: boolean }) => (
        <Chip
          background={!checked ? Colors.white : Colors.primary}
          label={label}
          fill
        />
      )}
    </CheckBox>
  );
}
```

### Actual Behavior

The browser throws the following error (in addition to an error about a function not being a valid React child):

![image](https://user-images.githubusercontent.com/28069638/117159553-60725900-ad8e-11eb-99bf-d338044dcbd2.png)

### Your Environment

- Grommet version: ^2.17.2 
- React version: ^17.0.2

jcfilben:
Closing since #5460 was merged

PR #5460, opened 2021-07-19
CheckBox and CheckBoxGroup can accept children now

g4rry420:
Signed-off-by: gurkiran_singh <gurkiransinghk@gmail.com>

#### What does this PR do?
`CheckBox` and `CheckBoxGroup` can accept children now

Components => CheckBox and CheckBoxGroup

Added story `Children` in both `CheckBox` and `CheckBoxGroup`

Using Storybook

N/A

#5221 

N/A

Yes, I already did that

backwards compatiable


jcfilben:
Docs should be updated on grommet-site repo since CheckBox and CheckBoxGroup docs have been transitioned to the new format. See https://github.com/grommet/grommet-site/issues/233 for more info

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -103,7 +104,9 @@ export const CheckBoxGroup = forwardRef(
               onChange={(event) =>
                 onCheckBoxChange(event, valueOption, optionProps)
               }
-            />
+            >
+              {children ? (state) => children(option.value, state) : null}
+            </CheckBox>
```
jcfilben, line 109: This doesn't work when options is an array of objects: `options={[{label: 'asc', disabled: true},{label: 'desc'}]}`

g4rry420: I can solve this by passing `options[index]` into the children.
One more thing for the array of objects that you have shown in above: `options={[{label: 'asc', disabled: true},{label: 'desc'}]}`.When I put this in options its showing me a bug by checking both the checkboxes when I select one. Correct me if I am wrong I think this is because we didn't pass `value` prop along with label. I am not sure if the checkboxes are supposed perfectly when only `label` prop with array of objects is passed. Below is the screenshot of above `options` object.

jcfilben: I think it is fine to have the object in the options array not include a 'value' prop but then you need to include a 'valueLabel' prop in CheckBoxGroup. See the [CheckBoxGroup/Form Uncontrolled](https://storybook.grommet.io/?path=/story/input-checkboxgroup-form-uncontrolled--form-uncontrolled) and [CheckBoxGroup/Object](https://storybook.grommet.io/?path=/story/input-checkboxgroup-object-options--object-options) Options stories for examples

g4rry420: okay, Thanks

jcfilben (approved):
Looks good to me

src/js/components/CheckBox/CheckBox.js
```diff
@@ -191,8 +192,14 @@ const CheckBox = forwardRef(
             if (onChange) onChange(event);
           }}
         />
-        {visual}
-        {hidden}
```
ericsoderberghp, line 195: Why are we removing `hidden` as well? I was expecting only `visual` to be superseded.

src/js/components/CheckBox/CheckBox.js
```diff
@@ -191,8 +192,14 @@ const CheckBox = forwardRef(
...
+        {children ? (
+          children({ checked })
```
ericsoderberghp, line 196: might the callee want `indeterminate` as well?

src/js/components/CheckBoxGroup/doc.js
```diff
@@ -13,6 +13,14 @@ export const doc = CheckBoxGroup => {
     .intrinsicElement('div');
 
   DocumentedCheckBoxGroup.propTypes = {
+    children: PropTypes.func.description(
+      `Function that will be called to render the visual representation.
+      It will be passed an object indicating whether the button is checked. It
```
ericsoderberghp, line 18:
```suggestion
      It will be called for each option and passed the option, and an object indicating whether the option is checked. It
```

src/js/components/CheckBox/CheckBox.js
g4rry420: okay, I will add it.

src/js/components/CheckBox/CheckBox.js
g4rry420: Willl do that.

src/js/components/CheckBox/CheckBox.js
```diff
@@ -191,7 +192,7 @@ const CheckBox = forwardRef(
             if (onChange) onChange(event);
           }}
         />
-        {visual}
+        {children ? children({ checked, indeterminate }) : <>{visual}</>}
```
ericsoderberghp, line 195: Why is there a Fragment around `visual`?
```suggestion
        {children ? children({ checked, indeterminate }) : visual}
```

src/js/components/CheckBox/doc.js
```diff
@@ -21,6 +21,14 @@ export const doc = (CheckBox) => {
     checked: PropTypes.bool
       .description('Same as React <input checked={} />')
       .defaultValue(false),
+    children: PropTypes.func.description(
```
ericsoderberghp, line 24: This change will now need to be made in https://github.com/grommet/grommet-site/blob/master/src/screens/CheckBox.js

src/js/components/CheckBoxGroup/doc.js
```diff
@@ -13,6 +13,15 @@ export const doc = CheckBoxGroup => {
     .intrinsicElement('div');
 
   DocumentedCheckBoxGroup.propTypes = {
+    children: PropTypes.func.description(
```
ericsoderberghp, line 16: https://github.com/grommet/grommet-site/blob/master/src/screens/CheckBoxGroup.js

ericsoderberghp (commented):
very close

ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution!

PR #5463, opened 2021-07-21
add a type of semi circle to meter

### LINKED ISSUE #5454: Add type to Meter of "semicircle"
### Expected Behavior

It would be nice to have a semicircle meter (half a circle) to support a pattern like this:
Screen Shot 2021-07-16 at 10 46 22 AM

This gauge type of meter is fairly common to convey a % of a whole, particularly with status.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

britt6612:
#### What does this PR do?
This PR add `semicircle` as an option for `type` of meter
meter/circle.js

just visual 
I will add a test once we decide if this is the correct route we want to go..
storybook 
changed the `circle` story to use `type=semicircle`
When adding just half of the `circle` I found that the path was not able to go from 270degrees to 90degree. 
Some of the functions for example `translateEndAngle`  using a start angle at 270 and end angle at 90 the math was not being calculated correctly. Instead of writing out a new function for the new angles I went ahead and used a start angle of 0 and an end angle of 180. Then I rotated this to look like it was from 270 degrees to 90 degree angle. 

`translateEndAngle` takes in `startAngle, anglePer, value` and the start angle of 270 was throwing off the calculations. 

As of now the ends are a little off I believe this is because for the  

```
        strokeDasharray={type === 'semicircle' ? circumference : 0}
        strokeDashoffset={type === 'semicircle' ? circumference / 2 : 0}
```

The numbers are about 5px off. This could be because of the rounding.. 

closes #5454 
yes
yes
backwards compatible


ericsoderberghp:
Great start! A few thoughts:
- I think we need to adjust the height to be (width / 2) earlier on.
- I think we should fix the math rather than use the transform, perhaps by setting the angle to start at?

src/js/components/Meter/stories/Circle.js
```diff
@@ -22,11 +22,7 @@ export const Circle = () => {
   return (
     <Grommet theme={grommet}>
       <Box align="center" pad="large">
-        <Meter
-          type="circle"
-          background="light-2"
-          values={[{ value, color: value > 50 ? 'accent-2' : 'accent-1' }]}
-        />
+        <Meter type="semicircle" background="light-2" value={50} max={100} />
```
ericsoderberghp, line 25: How about we merge Circle and Bar to have a single story for Type, where we show different types all together?

ericsoderberghp:
Math ideas for semicircle:
- start angle should be 270, end angle should be 90
- center Y should be `width` instead of `width / 2`
- `anglePer` should be `180 / max`
- we should use something like `angle % 360` instead of `Math.max(360, angle)` in `arcCommands()`, just let the angle loop around
- resulting viewBox should have `height = width / 2`
- we'll have to draw an arc instead of using `<circle />` for the background track

Hopefully this gives some ideas on things to work from.

src/js/components/Meter/stories/Circle.js
```diff
@@ -22,11 +22,7 @@ export const Circle = () => {
   return (
     <Grommet theme={grommet}>
       <Box align="center" pad="large">
-        <Meter
-          type="circle"
-          background="light-2"
-          values={[{ value, color: value > 50 ? 'accent-2' : 'accent-1' }]}
-        />
+        <Meter type="semicircle" background="light-2" value={90} max={100} />
```
ericsoderberghp, line 25: I don't think we should be removing the "circle" type here. I think we should have a separate story?

src/js/components/Meter/Circle.js
```diff
@@ -19,11 +25,13 @@ const Circle = forwardRef((props, ref) => {
       : parseMetricToNum(theme.global.edgeSize[thickness] || thickness);
   const mid = width / 2;
   const radius = width / 2 - height / 2;
-  const anglePer = 360 / max;
+  const anglePer = type === 'semicircle' ? 180 / max : 360 / max;
```
ericsoderberghp, line 28:
```suggestion
  const anglePer = (type === 'semicircle' ? 180 : 360) / max;
```

src/js/components/Meter/Circle.js
```diff
@@ -36,7 +44,13 @@ const Circle = forwardRef((props, ref) => {
 
       let endAngle;
       if (startValue + value >= max) {
```
ericsoderberghp, line 46: I think this block can just be:
```
      if (startValue + value >= max) {
        endAngle = type === 'semicircle' ? 90 : 360;
      } else {
        endAngle = translateEndAngle(startAngle, anglePer, value);
      }
```

src/js/components/Meter/Circle.js
```diff
@@ -145,6 +159,8 @@ const Circle = forwardRef((props, ref) => {
         strokeWidth={height}
         strokeLinecap={round ? 'round' : 'square'}
         fill="none"
+        strokeDasharray={type === 'semicircle' ? circumference : 0}
```
ericsoderberghp, line 162: I don't think we want this. We should use `arcCommands` to add a path in the semicircle case, instead of using `<circle />`.

britt6612: > Math ideas for semicircle:
> 
> * start angle should be 270, end angle should be 90
> * center Y should be `width` instead of `width / 2`
> * `anglePer` should be `180 / max`
> * we should use something like `angle % 360` instead of `Math.max(360, angle)` in `arcCommands()`, just let the angle loop around
> * resulting viewBox should have `height = width / 2`
> * we'll have to draw an arc instead of using `<circle />` for the background track
> 
> Hopefully this gives some ideas on things to work from.

So of course I saw this after I pushed up some changes! I think I had thought of it a little different but I agree with the
> * start angle should be 270, end angle should be 90
> * center Y should be `width` instead of `width / 2`
> * `anglePer` should be `180 / max`
> * resulting viewBox should have `height = width / 2`

I think the 2 things I would like to discuss further today are:   
> * we'll have to draw an arc instead of using `<circle />` for the background track
> * we should use something like `angle % 360` instead of `Math.max(360, angle)` in `arcCommands()`, just let the angle loop around

Just to set up our discussion 👆🏻

src/js/components/Meter/stories/Circle.js
britt6612: yes I had it only for testing purposes but I  added `semicircle` story then I can 
> *  merge Circle and Bar to have a single story for Type

src/js/components/Meter/Circle.js
```diff
@@ -129,23 +121,46 @@ const Circle = forwardRef((props, ref) => {
       startAngle = endAngle;
     });
 
+  const d1 = arcCommands(width / 2, width, radius, 270, 90);
```
ericsoderberghp, line 124: We shouldn't be building this path if we're not doing a "semicircle"

src/js/utils/graphics.js
```diff
@@ -13,7 +13,7 @@ export const polarToCartesian = (centerX, centerY, radius, angleInDegrees) => {
 export const arcCommands = (centerX, centerY, radius, startAngle, endAngle) => {
   // handle that we can't draw a complete circle
   let normalizedEndAngle = endAngle;
-  if (endAngle - startAngle >= 360) {
+  if (endAngle > startAngle && endAngle - startAngle >= 360) {
```
taysea, line 16: Can we add a comment around what this logic does? especially since now we're updating it

src/js/utils/graphics.js
```diff
@@ -36,4 +36,6 @@ export const arcCommands = (centerX, centerY, radius, startAngle, endAngle) => {
 };
 
 export const translateEndAngle = (startAngle, anglePer, value) =>
-  Math.min(360, Math.max(0, startAngle + anglePer * value));
+  Math.max(0, startAngle + anglePer * value) % 360;
+// Math.min(360, Math.max(0, startAngle + anglePer * value));
```
taysea, line 40: Remove unused code. Also can you add a comment around what this is doing since we're changing the logic? will be helpful for future

src/js/components/Meter/stories/Circle.js
taysea: Think we still need to remove "type=semicircle" from here

src/js/utils/graphics.js
britt6612: aw yes, good catch! Didn't realize that was still in there

src/js/components/Meter/Circle.js
```diff
@@ -129,23 +120,49 @@ const Circle = forwardRef((props, ref) => {
       startAngle = endAngle;
     });
 
+  let SemiCirclePath;
```
ericsoderberghp, line 123: How about doing these more consistently with each other? For example:
```
let track;
if (type === 'semicircle') {
  const d1 = ...
  track = <path ... />;
} else {
  track = <circle ... />;
}

ericsoderberghp (approved):

PR #5468, opened 2021-07-22
add values to format messageContext

### LINKED ISSUE #5459: Use `format` for Calendar announcements
In TextInput, we use `format()` from `MessageContext` to localize the strings we pass to `announce()`. We should do the same for Calendar.

britt6612:
#### What does this PR do?
This PR adds to the `MessageContext` in order for the `values` to be read for the announce messages
MessageContext.js
Calendar.js
I changed  `calendar` in order to add the `format` to:
```
           announce(
                  format({
                    id: 'calendar.previous',
                    messages,
                    values: {
                      date: previousMonth.toLocaleDateString(locale, {
                        month: 'long',
                        year: 'numeric',
                      }),
                      name: 14,
                    },
                  }),
                );
```

use the screen reader on calendar using the next and previous buttons
  `values: (optional) currently unused but in the future
  will be an object with substitution values for
  positional variables in the message text.`

We needed to add values 
closes #5459 
no
sure
backwards compatible

### DIFF
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -10,6 +10,7 @@ import React, {
 import { ThemeContext } from 'styled-components';
 import { defaultProps } from '../../default-props';
 import { AnnounceContext } from '../../contexts/AnnounceContext';
+import { MessageContext } from '../../contexts/MessageContext';
 
 import { Box } from '../Box';
 import { Button } from '../Button';
@@ -172,6 +173,7 @@ const Calendar = forwardRef(
       firstDayOfWeek = 0,
       header,
       locale = 'en-US',
+      messages,
       onReference,
       onSelect,
       range,
@@ -184,6 +186,7 @@ const Calendar = forwardRef(
   ) => {
     const theme = useContext(ThemeContext) || defaultProps.theme;
     const announce = useContext(AnnounceContext);
+    const { format } = useContext(MessageContext);
 
     // set activeDate when caller changes it, allows us to change
     // it internally too
@@ -586,10 +589,16 @@ const Calendar = forwardRef(
               onClick={() => {
                 changeReference(previousMonth);
                 announce(
-                  `Moved to ${previousMonth.toLocaleDateString(locale, {
-                    month: 'long',
-                    year: 'numeric',
-                  })}`,
+                  format({
+                    id: 'calendar.previous',
+                    messages,
+                    values: {
+                      date: previousMonth.toLocaleDateString(locale, {
+                        month: 'long',
+                        year: 'numeric',
+                      }),
+                    },
+                  }),
                 );
               }}
             />
@@ -603,10 +612,16 @@ const Calendar = forwardRef(
               onClick={() => {
                 changeReference(nextMonth);
                 announce(
-                  `Moved to ${nextMonth.toLocaleDateString(locale, {
-                    month: 'long',
-                    year: 'numeric',
-                  })}`,
+                  format({
+                    id: 'calendar.next',
+                    messages,
+                    values: {
+                      date: previousMonth.toLocaleDateString(locale, {
+                        month: 'long',
+                        year: 'numeric',
+                      }),
+                    },
+                  }),
                 );
               }}
             />
@@ -786,19 +801,31 @@ const Calendar = forwardRef(
                 onPreviousMonth: () => {
                   changeReference(previousMonth);
                   announce(
-                    `Moved to ${previousMonth.toLocaleDateString(locale, {
-                      month: 'long',
-                      year: 'numeric',
-                    })}`,
+                    format({
+                      id: 'calendar.previous',
+                      messages,
+                      values: {
+                        date: previousMonth.toLocaleDateString(locale, {
+                          month: 'long',
+                          year: 'numeric',
+                        }),
+                      },
+                    }),
                   );
                 },
                 onNextMonth: () => {
                   changeReference(nextMonth);
                   announce(
-                    `Moved to ${previousMonth.toLocaleDateString(locale, {
-                      month: 'long',
-                      year: 'numeric',
-                    })}`,
+                    format({
+                      id: 'calendar.next',
+                      messages,
+                      values: {
+                        date: previousMonth.toLocaleDateString(locale, {
+                          month: 'long',
+                          year: 'numeric',
+                        }),
+                      },
+                    }),
                   );
                 },
                 previousInBound: betweenDates(previousMonth, bounds),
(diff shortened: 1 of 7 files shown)


src/js/components/Calendar/Calendar.js
```diff
@@ -172,6 +173,7 @@ const Calendar = forwardRef(
       firstDayOfWeek = 0,
       header,
       locale = 'en-US',
+      messages,
```
MikeKingdom, line 176: be sure to document this and add it to the types

src/js/contexts/MessageContext/MessageContext.js
```diff
@@ -32,11 +32,26 @@ export const format = (options, messages) => {
...
     }
   });
+
   const message =
     (options.messages ? options.messages[baseId] : undefined) ||
     messageObj ||
     options.defaultMessage;
-  return message;
+
+  const values = options.values ? options.values : undefined;
```
MikeKingdom, line 41: This doesn't really do anything except maybe convert null, 0 or '' to undefined.
You could just do
```
const { values } = options;
Also message could technically be undefined at this point (which is why I try to use the newer ? operator in the suggestion below.)

File: src/js/components/Calendar/Calendar.js (hunk 1 above)
Comment (britt6612) [reply src/js/components/Calendar/Calendar.js:176]: good catch Ill add that

File: src/js/contexts/MessageContext/MessageContext.js [hunk 3]
```diff
@@ -32,11 +32,26 @@ export const format = (options, messages) => {
...
+
+  let newMessage;
+  if (values) {
+    newMessage = message.split(' ');
```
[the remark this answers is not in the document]
britt6612: your right bad assumption! l'll check this out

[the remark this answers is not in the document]
jlaluces123: Hey @MikeKingdom , could you provide an example of this as well so I can get a better view of what we might not be accounting for?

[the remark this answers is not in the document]
MikeKingdom: Surely. Ideographic languages like Japanese, Chinese, Korean, etc, often don't use spaces to separate words or phrases. Here's what the "Moved to {date}" string would be translated to in Japanese:
`"{date}に移動しました"`
There are no spaces there not to mention the date appears first (hence the need for substitution rather than concatenation in the code).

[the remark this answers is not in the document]
MikeKingdom: And one thought that just occurred to me is it might be safest to not replace tokens we think we find but which don't have a corresponding thing in `values`. My example code would just replace them with an empty string.

src/js/components/Calendar/doc.js
```diff
@@ -100,6 +100,18 @@ to disable the previous and next buttons.
...
+      next: PropTypes.string,
+    })
+      .description(
+        `Custom messages for Calendar. Used for accessibility by screen
+          readers.`,
+      )
+      .defaultValue({
+        previous: 'Moved to {date}',
+        next: 'Moved to {date}',
+      }),
```
jcfilben, line 114: doc changes should be done on grommet-site since the Calendar doc has been updated to the new format. See https://github.com/grommet/grommet-site/pull/230

src/js/components/Calendar/Calendar.js
```diff
@@ -786,19 +801,31 @@ const Calendar = forwardRef(
...
                   announce(
-                    `Moved to ${previousMonth.toLocaleDateString(locale, {
-                      month: 'long',
-                      year: 'numeric',
-                    })}`,
+                    format({
+                      id: 'calendar.next',
+                      messages,
+                      values: {
+                        date: previousMonth.toLocaleDateString(locale, {
```
MikeKingdom, line 823: This should be nextMonth?

src/js/components/Calendar/Calendar.js
```diff
@@ -603,10 +612,16 @@ const Calendar = forwardRef(
...
                 announce(
-                  `Moved to ${nextMonth.toLocaleDateString(locale, {
-                    month: 'long',
-                    year: 'numeric',
-                  })}`,
+                  format({
+                    id: 'calendar.next',
+                    messages,
+                    values: {
+                      date: previousMonth.toLocaleDateString(locale, {
```
MikeKingdom, line 619: this should be nextMonth?

src/js/contexts/MessageContext/MessageContext.js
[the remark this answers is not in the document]
britt6612: So if we don't have a corresponding `values`  we should not replace them with an empty string ?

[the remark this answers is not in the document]
MikeKingdom: That would be my suggestion I think simply so that if our way of processing substitutions tries to replace something the developer didn't intend as a parameter it would just leave it there. That kind of thing seems more likely with backend error content and such that we wouldn't control (although probably not the case with Grommet itself. However. this also helps people notice if they have a message where they missed providing a value.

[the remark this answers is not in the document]
MikeKingdom: BTW. I would still recommend something more like my suggested code above for parsing the substitutions.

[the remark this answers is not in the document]
britt6612: got it thanks for clarifying!/l

src/js/contexts/MessageContext/MessageContext.js
```diff
@@ -32,11 +32,24 @@ export const format = (options, messages) => {
...
     messageObj ||
     options.defaultMessage;
-  return message;
+
+  const { values } = options;
+
+  let newMessage = message;
+  const tokens = message?.match(/\{(.+?)\}/g);
+  tokens?.forEach((token) => {
+    const name = token;
```
MikeKingdom, line 46: If you want to use `name` instead of `token` should probably just change line 45 to
```jsx
tokens?.forEach(name => {
```
and delete line 46

src/js/languages/default.json
```diff
@@ -1,4 +1,8 @@
 {
+  "calendar": {
+    "previous": "{Moved to {date}",
```
MikeKingdom, line 3: Looks like an extra { crept in.

src/js/contexts/MessageContext/MessageContext.js
britt6612: `token` is fine when I was `console.logging` everything I changed some names to break it up but its good now!

jlaluces123 (approved):
I pulled down `MessageContext.js` to test on `FileInput` and everything works as intended. Great work!

MikeKingdom (approved):

ericsoderberghp (approved):

PR #5473, opened 2021-07-26
Pass along RadioButton focus state to children.

### LINKED ISSUE #5472: RadioButtonGroup does not pass focus state to children components
### Expected Behavior

Custom children components for `RadioButtonGroup` should receive focus state along with hover and checked so they can be styled appropriately

### Actual Behavior

Focus state is not passed down to the children components if children is defined.

The line in question is https://github.com/grommet/grommet/blob/b001ef3a3ef668edf2c71dd32258408f9dd1f793/src/js/components/RadioButton/RadioButton.js#L85

And this state is passed here when using `RadioButtonGroup`: https://github.com/grommet/grommet/blob/b001ef3a3ef668edf2c71dd32258408f9dd1f793/src/js/components/RadioButtonGroup/RadioButtonGroup.js#L172-L173

The issue is observed in the story which uses children:

https://storybook.grommet.io/?path=/story/input-radiobuttongroup-children--children

The children have no styling for focus state and thus inaccessible by keyboard users. 

### Steps to Reproduce

N/A

### Your Environment

- Grommet version: 2.17.1
- Browser Name and version: Chrome 91

srallen:
Signed-off-by: Sarah Allen <sarah@zooniverse.org>

#### What does this PR do?

This fixes #5472 by including the `RadioButton` focus state as part of the state object that gets passed to children by the `RadioButtonGroup`.

The `RadioButtonGroup` children story is now updated to demonstrate that the focus state is now available and the children can now be styled based on that state like it is for hover and checked.

Manual testing via storybook. I didn't see an existing unit test for testing hover with children, so I didn't add one for focus. Happy to do so if requested.

As noted above, the story is now updated.

As of right now, if someone uses `RadioButtonGroup` and defines custom children, then the child components can be inaccessible for keyboard users if there's no focus state styling.

#5472 

Backwards compatible.

### DIFF
--- a/src/js/components/RadioButton/RadioButton.js
+++ b/src/js/components/RadioButton/RadioButton.js
@@ -82,7 +82,7 @@ const RadioButton = forwardRef(
             })}
           />
           {children ? (
-            children({ checked, hover })
+            children({ checked, focus: focus && focusIndicator, hover })
           ) : (
             <StyledRadioButtonBox
               focus={focus && focusIndicator}
--- a/src/js/components/RadioButtonGroup/stories/Children.js
+++ b/src/js/components/RadioButtonGroup/stories/Children.js
@@ -16,13 +16,14 @@ export const Children = () => {
           gap="xsmall"
           options={['asc', 'desc']}
           value={value}
-          onChange={event => setValue(event.target.value)}
+          onChange={(event) => setValue(event.target.value)}
         >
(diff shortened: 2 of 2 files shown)


ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution!

PR #5474, opened 2021-07-27
Don't apply role="gridcell" to blank days or role="row" to blank week

### LINKED ISSUE #5436: Fix Calendar Accessibility with Blank Dates
Related to #4068 

### Expected Behavior

When a user interacts with a calendar with "showAdjacentDays" = false, the blank days should not be a part of the table and/or there needs to be proper context given to those with screen readers.

### Actual Behavior

With a calendar with "showAdjacentDays" = false, a user needs to go through the blank dates in the table. The screen reader reads "blank" with these dates, which is confusing and provides no helpful information for context.

### Video exhibiting the issue

https://user-images.githubusercontent.com/38971752/125327043-b1fc0e80-e310-11eb-9b96-c1b2166bc3db.mov

### Steps to Reproduce

1. Using VoiceOver for Mac or another screen reader, open the story for Show Adjacent Days Calendar in Storybook
2. Use keyboard commands to navigate to the days in the table for the "false" state and interact with the blank spots

### Your Environment

- Grommet version: v2.17.3
- Operating System and version (desktop or mobile): Macbook Pro Laptop

taysea:
#### What does this PR do?
Removes use of role="gridcell" on blank Calendar days and role="row" from weeks that are completely blank to improve experience of screen readers.

Removing "gridcell" on blank days ensures a screen reader user doesn't need to navigate over these days.

The removal of role="row" on blank weeks is needed for screen reader to properly calculate number of rows in a table. If not, a completely blank week will still be included in the row count but not accessible by screen reader which is confusing. For example "Row 1 of 6" where only 5 rows are able to be accessed because on week is empty.

Tested with VoiceOver locally.

Test on showAdjacentDays story with VoiceOver and arrow keys. Should not be able to access blank days.

Closes #5436 

No.

Yes. Improved behavior for screen reader users by removing blank days from navigable Calendar days.

Backwards compatible.

### DIFF
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -640,6 +640,8 @@ const Calendar = forwardRef(
     let day = new Date(displayBounds[0]);
     let days;
     let firstDayInMonth;
+    // if a week is filled with all blank days, we don't want to set role="row"
+    let blankWeek = false;
 
     while (day.getTime() < displayBounds[1].getTime()) {
       if (day.getDay() === firstDayOfWeek) {
@@ -657,7 +659,6 @@ const Calendar = forwardRef(
       if (!showAdjacentDays && otherMonth) {
         days.push(
           <StyledDayContainer
-            role="gridcell"
             key={day.getTime()}
             sizeProp={size}
             fillContainer={fill}
@@ -665,6 +666,15 @@ const Calendar = forwardRef(
             <StyledDay sizeProp={size} fillContainer={fill} />
           </StyledDayContainer>,
         );
+
+        if (
+          weeks.length === 5 &&
+          /* If the length days array is less than the current getDate()
+          we know that all days in the array are from the next month. */
+          days.length < day.getDate()
+        ) {
+          blankWeek = true;
+        }
       } else if (
         /* Do not show adjacent days in 6th row if all days
         fall in the next month */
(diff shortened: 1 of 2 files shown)


amandadupell (approved):
looks good to me!!

src/js/components/Calendar/Calendar.js
```diff
@@ -640,6 +640,8 @@ const Calendar = forwardRef(
     let day = new Date(displayBounds[0]);
     let days;
     let firstDayInMonth;
+    // if a week is filled with all blank days, we don't want to set role="row"
```
halocline, line 643: Feel like this comment may be more helpful where the logic is implemented on L793 and this comment could either be eliminated or read
```suggestion
    // a week containing only blank days
```
if you feel like it is helpful to be defined.

src/js/components/Calendar/Calendar.js
```diff
@@ -776,11 +786,14 @@ const Calendar = forwardRef(
           );
         }
       }
-
       day = addDays(day, 1);
     }
     weeks.push(
-      <StyledWeek role="row" key={day.getTime()} fillContainer={fill}>
+      <StyledWeek
+        role={!blankWeek ? 'row' : undefined}
```
halocline, line 793:
```suggestion
        // if a week contains only blank days, for screen reader accessibility we don't want to set role="row"
        role={!blankWeek ? 'row' : undefined}
```

halocline (approved):
Functionality is very nice.

Minor suggestion regarding placement of code comments. Soft opinion.

src/js/components/Calendar/Calendar.js
taysea: Included this in recent push.

ericsoderberghp (approved):

Issue #5437, opened 2021-07-12
Fix DateInput Accessibility for Input Field and Calendar Pop-up


amandadupell:
Related to #4068 

### Suggestions

A user needs to be alerted of a new layer popping open, like how the calendar does when the user navigates to the input field. The user can select from the table of dates when the calendar pops up, but the connection between the input field and the pop-up calendar may be confusing to someone using a screen reader. 

### Expected Behavior

When a user goes to enter a date in the input field, it should allow a user to navigate to the table of dates and use the arrow keys to select a date. A user should be able to navigate back to the input field easily. The user should be alerted that the calendar has been opened and closed based on the toggled view.

### Actual Behavior

When a user goes to enter a date in the input field, the calendar is opened and the user is not alerted. The user is unable to use keyboard commands to get back to the input field. Any buttons under the input field are also unable to be accessed by the screen reader.

### Video exhibiting the issue

https://user-images.githubusercontent.com/38971752/125334004-e1af1480-e318-11eb-84d6-9679516d42e6.mov

### Steps to Reproduce

1. Using VoiceOver for Macbook or an other screen reader, open the Form DateInput story in Storybook
2. Navigate to the input field. The Calendar will open and you can navigate to the table of dates
3. Once in the table, try to get back to the input field or to the button

### Your Environment

- Grommet version: v2.17.3
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): MacBook Pro

britt6612:
closing this issue both were addressed in the PRs that were merged

PR #5477, opened 2021-07-27
Improve DateInput keyboard accessibility

taysea:
#### What does this PR do?

Updates the Drop in DateInput Calendar to not trapFocus in order to allow the keyboard user to move backward to the input if desired. However, if the user leaves both the input and Calendar, we want to close the Calendar so an additional mechanism/effect has been added to see if focus has left both of these elements.

Tested locally in "Format" DateInput storybook example. 

In that example, tab to the DateInput. Once the calendar opens, tab to items within the Calendar. Shift + tab back to the input. Tab through the Calendar until you are past the Calendar. Calendar should now be closed.

Related to #5437 

No.

Improved keyboard navigation of DateInput.

Backwards compatible.

### DIFF
--- a/src/js/components/DateInput/DateInput.js
+++ b/src/js/components/DateInput/DateInput.js
@@ -3,6 +3,7 @@ import React, {
   useContext,
   useEffect,
   useMemo,
+  useRef,
   useState,
 } from 'react';
 import { ThemeContext } from 'styled-components';
@@ -15,7 +16,7 @@ import { DropButton } from '../DropButton';
 import { FormContext } from '../Form';
 import { Keyboard } from '../Keyboard';
 import { MaskedInput } from '../MaskedInput';
-import { useForwardedRef } from '../../utils';
+import { containsFocus, useForwardedRef } from '../../utils';
 import {
   formatToSchema,
   schemaToMask,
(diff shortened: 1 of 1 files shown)


britt6612 (approved):
LGTM! Nice work

jcfilben (approved):
Looks good to me!

halocline (approved):

taysea:
Converting to draft to look into behavior raised by @amandadupell where Calendar Grid isn't first thing navigated to when DateInput exists on a page with other tabbable elements:

https://user-images.githubusercontent.com/12522275/127396380-c7a116b0-c0ae-44be-a79c-3196f638663c.mov

halocline:
👀  Functionality looks good. I'm reviewing the logic and playing with variable naming a bit.

amandadupell (commented):
seems to be working as expected with opening and closing; some odd behavior with selecting a date but we can discuss on monday

ericsoderberghp:
It seems like quite a bit of state management. I'm wondering if we can simplify this a bit by doing things within the existing event handlers and using `document.activeElement` rather than `useState` and `useEffect`.

Or, perhaps we should only open the drop when the input already has focus and the user clicks on it or presses the down arrow or the space bar?

taysea:
> It seems like quite a bit of state management. I'm wondering if we can simplify this a bit by doing things within the existing event handlers and using `document.activeElement` rather than `useState` and `useEffect`.
> 
> Or, perhaps we should only open the drop when the input already has focus and the user clicks on it or presses the down arrow or the space bar?

I like the behavior that the Drop opens when the user first enters the input -- either via click or tab focus. But I will look to see if `document.activeElement` could be leveraged more over state management/effects

Issue #5483, opened 2021-07-29
Anchor weight should accept 'inherit'.


Zodiase:
### Expected Behavior

![image](https://user-images.githubusercontent.com/1815174/127429713-8e33a819-296a-4600-b9a2-70f0a7581d6d.png)

By "any browser font-weight definition" I expect `inherit` to be expected.
I need this setting for a plug-and-play component using an anchor that always inherits the contextual font-size, font-weight and text color.

### Actual Behavior

I'm discouraged by both the typing and the props check.
`weight` is limited to only `normal`, `bold` and numeric values. per https://github.com/grommet/grommet/blob/6fda568ed39a73497cb25fd3a5ecbd092bff7685/src/js/components/Anchor/propTypes.js#L26-L29 and https://github.com/grommet/grommet/blob/6fda568ed39a73497cb25fd3a5ecbd092bff7685/src/js/components/Anchor/index.d.ts#L32

I've managed to bypass the type check by using `weight={'inherit' as 'normal'}`.
But the props check, even tho not behavior breaking, is throwing annoying errors in tests.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-forked-v15kv?file=/index.js

### Steps to Reproduce

1. Have an anchor.
2. Add a weight prop to it with the value `"inherit"`.

### Your Environment

I don't think these matter but...

- Grommet version: ^2.15.0.
- Browser Name and version: Chrome, 92.0.4515.107
- Operating System and version (desktop or mobile): macOS 11.1.

PR #5484, opened 2021-07-29
Expand accepted value types for anchor weight.

Zodiase:
#### What does this PR do?

Resolve https://github.com/grommet/grommet/issues/5483 by adding all acceptable values for font-weight to anchor weight.

See https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight.

There are only three files among which one is the test file. 

I've followed the current test cases for font weight and added more to cover the new acceptable values.

Make sure all acceptable values in https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight do not throw type errors or prop-type check errors.

https://github.com/grommet/grommet/issues/5483

The docs promised "any browser font-weight definition" so it's already generous enough.

Sure.

I don't think this breaks anything.
However I do think the `'inherit' | 'initial' | 'revert' | 'unset'` part is universal to all override-able styles and should be a defined constant if there isn't already one somewhere.

### DIFF
--- a/src/js/components/Anchor/index.d.ts
+++ b/src/js/components/Anchor/index.d.ts
@@ -29,7 +29,16 @@ export interface AnchorProps {
     | 'xlarge'
     | 'xxlarge'
     | string;
-  weight?: 'normal' | 'bold' | number;
+  weight?:
+    | 'normal'
+    | 'bold'
+    | 'lighter'
+    | 'bolder'
+    | 'inherit'
+    | 'initial'
+    | 'revert'
+    | 'unset'
+    | number;
 }
 
 type aProps = Omit<JSX.IntrinsicElements['a'], 'color'>;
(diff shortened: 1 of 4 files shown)


jcfilben (commented):
Looks like the test is failing, make sure the snapshot is updated. Run `yarn test-update`

halocline (approved):
Thank you, @Zodiase.

Chromatic snapshot change looks appropriate.

src/js/components/Anchor/index.d.ts
```diff
@@ -29,7 +29,16 @@ export interface AnchorProps {
     | 'xlarge'
     | 'xxlarge'
     | string;
-  weight?: 'normal' | 'bold' | number;
+  weight?:
```
ericsoderberghp, line 32: We don't typically explicitly type for all possible browser CSS values and instead just use "string" for that case. That helps direct the user to using the typical grommet values while still allowing less frequently used ones.

However, if there's a way to leverage the default typing somehow via some indirection, like "these values plus any default values", that would be fine too. Not sure that we can do that though.

jcfilben (approved):

jcfilben:
I pushed up changes so that weight just specifies `'normal' | 'bold' | string | number` instead of listing out all the possible browser CSS values for weight.

The other changes are lint changes from merging with master.

PR #5491, opened 2021-07-30
fix: updates DateInput click behavior

### LINKED ISSUE #5481: DateInput Calendar doesn't always open when input is clicked
### Expected Behavior

Calendar should always open when the input is clicked.

### Actual Behavior

After the input is opened, then date selected and close, the user can't get the calendar to open again without clicking somewhere outside the DateInput and then back into it.

https://storybook.grommet.io/?path=/story/input-dateinput-format--format

### Steps to Reproduce

1. Click into input (Calendar will open)
2. Click again (Calendar will close)
3. Click again (Calendar won't open)

### Your Environment

- Grommet version: Stable
- Browser Name and version: Version 92.0.4515.107

PauloE314:
#### What does this PR do?
It fixes a bug in the DateInput component (when the date input was clicked 3 times, the calendar didn't open until the user click in somewhere out).

DateInput.js

I've tested on a local storybook.

Run the storybook and check the behavior.

Closes #5481.

No.

Backwards compatible.


jcfilben (approved):
Looks good!

src/js/components/DateInput/DateInput.js
```diff
@@ -177,6 +177,9 @@ const DateInput = forwardRef(
               setOpen(true);
               if (onFocus) onFocus(event);
             }}
+            onMouseDown={({ target }) => {
```
wenderjean, line 180: Is it the kinda behavior we're able to cover with tests?

PauloE314: Yes, I think so

taysea: Typically, we try to avoid `onMouseDown` to achieve a click behavior because a user may mouse down and drag away from the input and decide not to click. Is there a reason we can't do this with `onClick` handler?

PauloE314: I've tried implement the solution with onClick, but this method is called after onFocus, so, when onFocus 'opens' the calendar, the onClick method is called and immediately closes it. This could be fixed passing the open/close logic only for the onClick method, but in this case, the calendar would no longer appear when only the onFocus method is called (when the user uses 'tab', for example).

ericsoderberghp: Can the `onClick` call `setOpen(true)` instead of `setOpen(!open)`?

PauloE314: This could be a solution, but this "toggle" behavior would be lost. I thought of put some logic with React states to achieve the same result, but without using the `onMouseDown` event. What you guys think about it?

ericsoderberghp: I'm wondering if we need the toggle behavior. Why would clicking on the input indicate that the user wants the Calendar Drop to be closed?

PauloE314: I thought it was a expected feature, so I tried make it work correctly, but that would be a nice solution.

taysea (approved):
This looks good. Thanks!

ericsoderberghp:
Thanks for your contribution!

Issue #5445, opened 2021-07-14
Menu incorrectly includes menu button when flipped vertically due to space availability


maitlandmarsh:
### Expected Behavior

When using drop alignment for a menu the menu button itself is not included in the drop content.
Expected behavior is that this is true whether the menu is opened in the expected direction or is flipped due to a lack of available space.

### Actual Behavior

The menu button is included in the drop content when the menu is flipped due to a lack of available space.

### URL, screen shot, or Codepen exhibiting the issue

This can be seen with the Menu Storybook
Looking at https://storybook.grommet.io/?path=/story/controls-menu-children--children we can see that with no drop align provided the menu button is included in the drop content.
![image](https://user-images.githubusercontent.com/33257834/125685254-cdd8d321-d75f-4f6f-acf7-5110d78518bd.png)

Compare this with https://storybook.grommet.io/?path=/story/controls-menu-simple--simple where we see that the button is not included.
![image](https://user-images.githubusercontent.com/33257834/125682573-ce65aea3-9717-4dcb-8a80-16fadf53642a.png)

If you shrink the browser window enough that the internal calculation done by DropContainer finds it necessary to dynamically flip the menu drop direction, you will see that the menu button is unexpectedly included.
![image](https://user-images.githubusercontent.com/33257834/125682937-fa699e71-c773-4728-9c97-fb3b5b6966b3.png)

The content going off the top of screen doesn't matter for this example. Below is a screenshot from a Storybook that has been modified to move the menu away from the top of the page, better showing the behavior:
![image](https://user-images.githubusercontent.com/33257834/125685445-d9714c00-03e6-4916-9783-c0bb536c940b.png)

### Steps to Reproduce

1. Set either the dropAlign or dropProps.align property on a Menu to align top at bottom 
  (e.g. `dropProps={{ align: { top: 'bottom' } }}` or `dropAlign={{ top: 'bottom' }}`)
2. Reduce the available screen space below the menu such that the menu is forced to flip vertically.

### Your Environment

https://storybook.grommet.io/

- Grommet version: 2.17.3
- Browser Name and version: Chrome (91.0.4472.124)
- Operating System and version (desktop or mobile): Windows 10

maitlandmarsh:
[...]

PR #5497, opened 2021-08-02
Fix: Remove duplicate menu button when menu forcefully flipped

jlaluces123:
This PR looks to solve issue #5445.

#### What does this PR do?
[Still for review] Introduces a boolean to be added to the conditional rendering check of `controlMirror` to see if the menu has shrunk and is being forced to flip vertically, removing the duplicate menu button.

- Review new boolean `shrunk` on `Menu.js`
- `yarn storybook` and checkout `Menu/Simple` story
- Shrink available window space so the menu is forcefully flipped
- Click on the menu and see how no duplicate menu button is present

Manual testing by observing UI changes on storybook (`Menu/Simple`)

Please follow steps in the `Where should the reviewer start?` section.
 
#5445

None needed

Nope, just a simple bug fix.

Backwards compatible 👍


src/js/components/Menu/Menu.js
```diff
@@ -352,7 +353,8 @@ const Menu = forwardRef((props, ref) => {
                   );
                 })}
               </Box>
-              {alignControlMirror === 'bottom' || align.bottom === 'bottom'
+              {(shrunk !== true && alignControlMirror === 'bottom') ||
+              align.bottom === 'bottom'
```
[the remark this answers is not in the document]
jlaluces123: Just wanted to get more knowledge on this, is there an example I could try out to see if I might've introduced a "side effect" or unexpected behavior?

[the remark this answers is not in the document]
MikeKingdom: Any time you modify the contents of a passed in object you risk some bad bug/side-effect. Basically the caller is passing a reference to an object and you don't know if it's OK to modify the contents of the object. So as a general rule, don't modify passed in data. Best practice is to copy (deep copy/clone in this case) to a new object which is safe to manipulate.

In this case an example side-effect could be you accidentally modify the theme itself for all other consumers of the theme context in the same app.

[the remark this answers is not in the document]
jlaluces123: All resolved in our meeting! Originally thought I was setting `align.bottom = bottom` instead of the comparison `===`. 😄 resolving this thread.

src/js/components/Menu/Menu.js
```diff
@@ -88,6 +88,7 @@ const Menu = forwardRef((props, ref) => {
   // when there's not enough space below DropButton. This state
   // is modified on /Drop/DropContainer.js.
   const [alignControlMirror, setAlignControlMirror] = useState();
+  const shrunk = alignControlMirror === align.top;
```
taysea, line 91: I am wondering about this variable name. Wondering if something like `alignTop` would make more sense?

src/js/components/Menu/Menu.js
```diff
@@ -352,7 +353,8 @@ const Menu = forwardRef((props, ref) => {
                   );
                 })}
               </Box>
-              {alignControlMirror === 'bottom' || align.bottom === 'bottom'
+              {!shrunk &&
```
taysea, line 356: Maybe add a comment
// if align.top is defined, don't show the controlMirror when the window height has shrunk

src/js/components/Menu/Menu.js
jlaluces123: Would `initialAlignTop` be a good name for this since we made this boolean to check if there was an align.top when starting?

src/js/components/Menu/Menu.js
jlaluces123: Added in for next commit! 😄

taysea (approved):
Looks good, thank you!

ericsoderberghp (approved):

PR #5506, opened 2021-08-03
fix height and size for semimeter

### LINKED ISSUE #5504: Meter semicircle is too small
The semicircle Meter isn't being drawn at the right scale. Compare:

https://storybook.grommet.io/?path=/story/visualizations-meter-semi-circle--semi-circle

with 

https://storybook.grommet.io/?path=/story/visualizations-meter-circle--circle

Both are size medium. The `<svg />` is the right viewbox size. But the arc paths look too small, both in radius and thickness.

britt6612:
#### What does this PR do?
fixes the `height` and `size` for `semicircle` 

circle.js
storybook
The height should be *  2 since it is half the circle and the radius is updated
closes #5504 
no
no
backwards compatible


src/js/components/Meter/Circle.js
```diff
@@ -13,12 +13,17 @@ const Circle = forwardRef((props, ref) => {
   const theme = useContext(ThemeContext);
   const width =
     size === 'full' ? 288 : parseMetricToNum(theme.global.size[size] || size);
-  const height =
-    type === 'pie'
-      ? width / 2
-      : parseMetricToNum(theme.global.edgeSize[thickness] || thickness);
+  let height;
```
ericsoderberghp, line 16: Can we rename this variable to "strokeWidth", changing the passed in prop to "strokeWidth"?

src/js/components/Meter/Circle.js
```diff
@@ -13,12 +13,17 @@ const Circle = forwardRef((props, ref) => {
...
+  if (type === 'pie') {
+    height = width / 2;
+  } else if (type === 'semicircle') {
+    height =
+      2 * parseMetricToNum(theme.global.edgeSize[thickness] || thickness);
```
ericsoderberghp, line 21: Why does this need to be different for "semicircle"? I was expecting the strokeWidth to be the same as for "circle" and "bar".

src/js/components/Meter/Circle.js
```diff
@@ -13,12 +13,17 @@ const Circle = forwardRef((props, ref) => {
...
+  } else
+    height = parseMetricToNum(theme.global.edgeSize[thickness] || thickness);
   const mid = width / 2;
-  const radius = width / 2 - height / 2;
+  const radius =
```
ericsoderberghp, line 25: I would think we want the "radius" to be the same, since the arc should have the same radius.

src/js/components/Meter/Circle.js
```diff
@@ -13,12 +13,17 @@ const Circle = forwardRef((props, ref) => {
...
-      : parseMetricToNum(theme.global.edgeSize[thickness] || thickness);
+  let height;
+  if (type === 'pie') {
+    height = width / 2;
+  } else if (type === 'semicircle') {
+    height =
+      2 * parseMetricToNum(theme.global.edgeSize[thickness] || thickness);
+  } else
+    height = parseMetricToNum(theme.global.edgeSize[thickness] || thickness);
   const mid = width / 2;
```
ericsoderberghp, line 24: How about we call this "centerX" and add a "centerY" and then use that as the first two arguments to arcCommands()?

src/js/components/Meter/Circle.js
```diff
@@ -129,7 +134,7 @@ const Circle = forwardRef((props, ref) => {
         strokeWidth={height}
         fill="none"
         {...strokeProps(background, theme)}
-        strokeLinecap="round"
+        strokeLinecap={round ? 'round' : 'square'}
```
ericsoderberghp, line 137: Why are we changing this?

ericsoderberghp (commented):
line 130: `const d1 = arcCommands(width / 2, width, radius, 270, 90);` should have "width / 2" or "centerY" as the second argument.

line 164: `viewBox={`0 0 ${width} ${width}`}` should be viewBox={`0 0 ${width} ${viewBoxHeight}`}

src/js/components/Meter/Circle.js
britt6612: Screen Shot 2021-08-03 at 1 29 17 PM
This is an option in props and I was not taking it into account 
https://v2.grommet.io/meter#round

src/js/components/Meter/Circle.js
britt6612: Im not sure if I follow why we would need `centerX` and `centerY`
`width/2` is passed in as the first 2 parameters for the arch

ericsoderberghp: I was trying to be more explicit to make it clear what we were using. For instance, the first two arguments to arcCommands are the center x,y coordinates for the arc. Using "mid" sometimes, but not always, makes the code a bit harder to follow.

src/js/components/Meter/Circle.js
```diff
@@ -101,15 +101,20 @@ const Circle = forwardRef((props, ref) => {
         }
         pathCaps.unshift(pathCap);
       } else {
-        const y = type === 'semicircle' ? width : width / 2;
-        const d = arcCommands(width / 2, y, radius, startAngle, endAngle);
+        const d = arcCommands(
+          width / 2,
+          width / 2,
```
ericsoderberghp, line 106: This should be "width" for the semicircle case, part of why I think "centerY" would make it clearer.

ericsoderberghp: Also, to catch this, I think we should have unit tests for both "round" and "not round" cases for each type.

src/js/components/Meter/Circle.js
britt6612: oh okay I see what you were trying to do I was exactly sure since we were using mid for both but yes I can add this so its more explicit what were passing in

src/js/components/Meter/Circle.js
britt6612: So I have `centerY` which is width/2 for the `semicircle` this is working on my end if I just put the width then I get the below picture.. 
Im not sure if Im missing something that you are seeing on your end but with the lastest changes the `semicircle` looks correct. 

Screen Shot 2021-08-03 at 8 02 01 PM

PR #5513, opened 2021-08-05
Remove component docs and react-desc

### LINKED ISSUE #5353: Tracking removal of react-desc
### Expected Behavior
`react-desc` should be removed from grommet and the documentation for each component should be contained in the grommet-site repo.

Original discussion about removing `react-desc` #2976 

### Actual Behavior
Grommet is using `react-desc` for documentation and the docs are in the grommet repo

### Steps:
In the steps below replace <component> with the name of the component you are changing. Pull requests should be submitted per component.

**Grommet-site**
Ensure the component has been transitioned to the new documentation format on the grommet site. This must be done BEFORE the doc is removed on grommet. See https://github.com/grommet/grommet-site/issues/233 to determine which components are using the new documentation format and can safely have their doc removed from grommet

Note: as of 7/29 all components have been transitioned to the new format for documentation on grommet-site

**Grommet**
Example pull request: https://github.com/grommet/grommet/pull/5492

1. Create a new file `src/js/components/<component>/propTypes.js`
2. Add `import PropTypes from 'prop-types';` to the propTypes file.
3. If necessary add any imports needed from `../../utils/general-prop-types'` to the propTypes file.
4. In the propTypes file add the propTypes that are defined in `src/js/components/<component>/doc.js`.
5. Export the propTypes in the propTypes file as `<component>Type`. (ex. AnchorType).
6. In the `src/js/components/<component>/<component>.js` file add `import { <component>Type } from './propTypes';`
7. Remove code referencing the component documentation from the `<component>.js` file.
  Example for the Anchor component remove the following:
    ```
    let AnchorDoc;
    // eslint-disable-next-line global-require
    AnchorDoc = require('./doc').doc(Anchor);
    }
    const AnchorWrapper = AnchorDoc || Anchor;
    
    export { AnchorWrapper as Anchor };
    ```
    replace with:
    ```
    <component>.propTypes = <component>Type;
    export { <component> };
    ``` 
9. Remove `src/js/components/<component>/doc.js`.
10. Replace the contents of the `src/js/components/<component>/README.md` file with 
    ```
    ## <component>
    Documentation for this component: https://v2.grommet.io/<component>
    ```
11. Run `yarn test-update`. The snapshots for `README-test` and `components-test` should be updated.

## Components
List for tracking which components have had `react-desc` removed from them:

- [x] Accordion
- [x] AccordionPanel
- [x] Anchor
- [x] Avatar
- [x] Box
- [x] Button
- [x] Calendar
- [x] Card, CardBody, CardFooter & CardHeader (should be done together)
- [x] Carousel
- [x] CheckBoxGroup
- [x] Chart
- [x] CheckBox
- [x] Clock
- [x] Collapsible
- [x] DataChart
- [x] DataTable
- [x] DateInput
- [x] Diagram
- [x] Distribution
- [x] Drop
- [x] DropButton
- [x] FileInput
- [x] Footer
- [x] Form
- [x] FormField
- [x] Grid
- [x] Grommet
- [x] Header
- [x] Heading
- [x] Image
- [x] InfiniteScroll
- [x] Keyboard
- [x] Layer
- [x] List
- [x] Main
- [x] Markdown
- [x] MaskedInput
- [x] Menu
- [x] Meter
- [x] Pagination
- [x] Paragraph
- [x] Nav
- [x] RadioButton
- [x] RadioButtonGroup
- [x] RangeInput
- [x] RangeSelector
- [x] RoutedAnchor (special case, not on grommet-site)
- [x] RoutedButton (special case, not on grommet-site)
- [x] Select
- [x] Sidebar
- [x] SkipLinks
- [x] Spinner
- [x] Stack
- [x] Table, TableBody, TableCell, TableFooter, TableHeader, & TableRow (should be done together)
- [x] Tab & Tabs (should be done together)
- [x] Text
- [x] TextArea
- [x] TextInput
- [x] Tip
- [x] Video
- [x] WorldMap

jcfilben:
#### What does this PR do?
Removes doc files from components, updates component readme files, and removes react-desc. Type checking is now done with prop-types instead of through react-desc

Storybook and jest

Storybook and jest

Closes #5353 

No

Backwards compatible


ericsoderberghp:
Looking good to me. Thanks!

ericsoderberghp:
I think we should look into why this increased the bundle size. I would have expected the opposite.

jcfilben:
Yeah, looking into the bundle size right now

jcfilben:
Update:
The difference in bundles size is caused by the addition of the propTypes.js files. The doc.js files were removed, which previously contained the propTypes, but the doc.js files were excluded from the /dist folder and the propTypes.js file are included.

We have the option of also excluding the propTypes.js files but I'm thinking that we should not do this because then there wouldn't be a warning if a users uses the incorrect propType (this is the current behavior except for generic proptypes that are shared across components which are checked). I think we should include the propTypes in the /dist so that users are notified if they use the wrong type for a prop.

@halocline mentioned that we could potentially change the webpack config to split grommet.min.js and a propTypes.min.js into two different entry points. Or do something where bundled development and production environments differently using a babel plugin that strips out propTypes conditionally. I am planning to look into these options.

halocline:
> Update:
> The difference in bundles size is caused by the addition of the `propTypes.js` files. The `doc.js` files were removed, which previously contained the propTypes, but the `doc.js` files were excluded from the `/dist` folder and the `propTypes.js` files are included.
> 

I'd like to look into the bundle diffs more. I believe the `doc.js` files are still being bundled, but obscured when evaluating the minified files. Do you see where they are being excluded?

By substituting each `doc.js` file with a `propTypes.js` we are reducing our unbundled size ~3.5 KB * ~60 components. It seems odd that size of the `grommet.min.js.map` is increasing by ~150 KB.

I think we should dig in here a bit more.

jcfilben:
I think the reason the doc.js files are not included in the minified file is because they are imported using require statements only when `process.env.NODE_ENV !== 'production'`. I tested importing a propType.js file this way and it was no longer being included in the minified version.

ericsoderberghp:
If we are sure that the size increase is appropriate for adding the propTypes, I am OK with increasing the bundle size limit. What we typically catch with this check is when grommet-icons are incorrectly imported. Which is very helpful.

jcfilben:
I did some further investigation and the way the propTypes are setup now on grommet, where they are not being included in the `grommet.min.js` file but still in the `/dist` folder, is the way to go in my opinion. 

I was initially confused and I thought that prop type checking was not occurring at all since the `doc.js` files were not included in the bundle which lead me to believe if we want prop type checking we would need to include the `propTypes.js` files in `grommet.min.js`. However, prop checking is occurring when the `doc.js` files are not included in grommet.min.js but only when a project is run in development mode and not when a project is in production mode.

The prop type checking only when in development mode is achieved with the following code:
```javascript
let BoxDoc;
  BoxDoc = require('./doc').doc(Box); // eslint-disable-line global-require
}
const BoxWrapper = BoxDoc || Box;

export { BoxWrapper as Box };
```

Having the prop types checked only in development mode makes sense to me and will allows us to keep the bundle size the same. So I think we should have a similar approach with the `propTypes.js` files as the `doc.js` files. We should include them in the `/dist` folder but not in `grommet.min.js`. The key is finding a way to do this without using require statements, since those are not allowed in ES6 modules. I am looking into using this library https://www.npmjs.com/package/babel-plugin-transform-react-remove-prop-types to achieve this.

jcfilben:
I updated the propTypes for each component to be wrapped with `if (process.env.NODE_ENV !== 'production')` so that in production mode the bundle size is reduced. The propTypes file is imported to the component regardless of being in production or development mode to avoid using `require`.

halocline:
> I updated the propTypes for each component to be wrapped with `if (process.env.NODE_ENV !== 'production')` so that in production mode the bundle size is reduced. 

Did the [babel-plugin approach](https://www.npmjs.com/package/babel-plugin-transform-react-remove-prop-types) not work in the same fashion? As expected? Other rationale?

src/js/components/Accordion/Accordion.js
```diff
@@ -67,6 +67,6 @@ const Accordion = forwardRef(
   },
 );
 
-Accordion.propTypes = AccordionType;
+Accordion.propTypes = AccordionPropType;
```
ericsoderberghp, line 70: A naming comment ...
Since we are assigning this to "propTypes", it might be slightly more aligned to call this "AccordionPropTypes".

jcfilben:
> Did the babel-plugin approach not work in the same fashion? As expected? Other rationale?

The babel-plugin did not work to remove the propTypes. I tried different configuration options with it but it only removed the `<component>.propTypes = <component>PropType;` from the component file, and therefore didn't reduce the overall bundle size or remove the propTypes.js files from the bundle. I think the plugin is expecting that the propTypes are contained within the same file they are used in.

src/js/components/Accordion/Accordion.js
jcfilben: Yeah that makes sense, I will go back and update

halocline:
> > Did the babel-plugin approach not work in the same fashion? As expected? Other rationale?
> 
> The babel-plugin did not work to remove the propTypes. I tried different configuration options with it but it only removed the `<component>.propTypes = <component>PropType;` from the component file, and therefore didn't reduce the overall bundle size or remove the propTypes.js files from the bundle. I think the plugin is expecting that the propTypes are contained within the same file they are used in.

Thank you for the explanation @jcfilben!

ericsoderberghp (approved):

PR #5517, opened 2021-08-09
Reset DateInput field value after state change.

AlbertoTonegari:
#### What does this PR do?
This PR adds the possibility to reset the DateInput field on state change.

From `DateInput.js`

I have added 2 unit tests for the new functionality.

I added 1 story called ResetDate.js and tested the functionality there.

After reviewing `DateInput.js` and `ResetDate.js` the functionality should be tested by opening the `ResetDate.js` story.

-

The issue is that on state change, there is no possibility to clear the `DateInput` field.

Here is a codesandbox showcasing the issue that the DateInput field can't be cleared.
https://codesandbox.io/s/recursing-star-3tnfk

-

I don't think so.

Yes, absolutely.

From my knowledge, I think it's backwards compatible.


src/js/components/DateInput/DateInput.js
```diff
@@ -105,17 +109,17 @@ const DateInput = forwardRef(
           disabled
             ? undefined
             : (nextValue) => {
-                let normalizedValue;
```
ericsoderberghp, line 108: I'm wondering why these changes are being added. Do you have the latest `prettier` installed?

src/js/components/DateInput/DateInput.js
```diff
@@ -74,8 +74,12 @@ const DateInput = forwardRef(
     useEffect(() => {
       if (
         schema &&
-        value &&
-        ((Array.isArray(value) && value[0]) || !Array.isArray(value))
+        (value || value?.length === 0) &&
```
ericsoderberghp, line 77: Could this be simplified to `value !== undefined`?

src/js/components/DateInput/DateInput.js
```diff
@@ -74,8 +74,12 @@ const DateInput = forwardRef(
...
+        (
+          (Array.isArray(value) && value[0])
+          || !Array.isArray(value)
+          || value?.length === 0
```
ericsoderberghp, line 81: I think was checked on line 77 already. I suspect you mean `value[0]?.length === 0`? In which case, I think line 79 could be: `(Array.isArray(value && value[0] !== undefined) || !Array.isArray(value)`

ericsoderberghp (commented):
Thanks for finding and including a fix. One other thing, can you file an issue in GitHub for this as well? Should take but a moment https://github.com/grommet/grommet/issues/new

src/js/components/DateInput/DateInput.js
AlbertoTonegari: Hello, thank you for the review.
Unfortunately, this will not work because I want to reset the date when I pass an `[ ]` as a value. And in this case `value[0] !== undefined` will be false.

src/js/components/DateInput/DateInput.js
AlbertoTonegari: Thank you for the review.
I have simplified that expression to the following:
 ```
![undefined, null].includes(value) &&
            Array.isArray(value) || !Array.isArray(value)
```
What do you think @ericsoderberghp ?

File: src/js/components/DateInput/DateInput.js (hunk 1 above)
Comment (AlbertoTonegari) [reply src/js/components/DateInput/DateInput.js]: Thank you for the review.
Yes, I don't know why it has done that.
I have reverted those changes.

File: src/js/components/DateInput/DateInput.js [hunk 4]
```diff
@@ -74,8 +74,8 @@ const DateInput = forwardRef(
     useEffect(() => {
       if (
         schema &&
-        value &&
-        ((Array.isArray(value) && value[0]) || !Array.isArray(value))
+        ![undefined, null].includes(value) &&
+          (Array.isArray(value) || !Array.isArray(value))
```
ericsoderberghp, line 78: Won't this logic always return true, since the value is either an Array or not an Array?

I'm thinking `undefined !== value` might be enough. That would align with how `undefined` is what determines whether an input component is controlled or not. In other words, I don't think we need to also allow for `null`.

AlbertoTonegari: You are absolutely right. 😅
I have modified it.

ericsoderberghp:
Thanks for your contribution!

PR #5518, opened 2021-08-09
move theme option out of back ticks

britt6612:
#### What does this PR do?
This PR moved a block of  code outside of backticks 
styledDataTable.js
storybook
add an option in the extend to pass theme this was not working before
https://github.com/grommet/grommet-theme-hpe/pull/202
When trying to add an option for the `background` to look for `theme.dark ?` this was not  working correctly
Need this to work to fix an issue in Hpe theme
no
no
backwards compatible

### DIFF
--- a/src/js/components/DataTable/StyledDataTable.js
+++ b/src/js/components/DataTable/StyledDataTable.js
@@ -155,8 +155,7 @@ const StyledDataTableCell = styled(TableCell)`
   ${(props) =>
     props.pin &&
     props.pin.length > 0 &&
-    `
-    position: sticky;
+    `position: sticky;
     ${props.pin
       .map(
         (p) =>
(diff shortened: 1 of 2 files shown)


ericsoderberghp:
Why didn't the prior code work?

britt6612:
> Why didn't the prior code work?

I was trying to pass through `theme` through `extend` to do something like this:

```
      extend: ({ theme }) => `
        background { theme.dark ? 'some color' : 'another color' ]
        }
      `,

```

I was finding that it was not recognizing `theme` I could tried adding `props.theme &&` before it looked for `props.theme.dataTable` but this also was not working. So I moved this block outside of the backtick of `sticky` and it was working correctly.. I couldn't exactly pin point as to why it was not recognizing theme prop being passed in..

taysea:
> > Why didn't the prior code work?
> 
> I was trying to pass through `theme` through `extend` to do something like this:
> 
> ```
>       extend: ({ theme }) => `
>         background { theme.dark ? 'some color' : 'another color' ]
>         }
>       `,
> ```

@ericsoderberghp a pure CSS string would work fine such as `background: red;` but the function extend notation was not. Any thoughts on why that might be the case?

ericsoderberghp:
Not off the top of my head. But, just looking at the code it seems like they should be identical. I think it would be worth understanding a bit more to know why.

britt6612:
> Not off the top of my head. But, just looking at the code it seems like they should be identical. I think it would be worth understanding a bit more to know why.

I'll investigate a little more.. see if I can pin point exactly why this isn't working properly as it is when it is stand alone

britt6612:
@ericsoderberghp 

So I kept playing around with the block of code within the backticks and I do not have a solid answer as to why this is not working but after also doing the same thing with something a little more simple like `anchor` extend and doing 

```
  ${(props) =>
    props.hasIcon &&
    ` border: dashed;
     ${props.theme.anchor.extend}
     `}
```

I believe what it comes down to is that when `props` is not properly getting passed down because of the backticks.

taysea (approved):

ericsoderberghp (approved):

PR #5524, opened 2021-08-11
Fix options layer bug

### LINKED ISSUE #5516: Layer breaks if it isn't include within a Grommet block
It looks like https://github.com/grommet/grommet/pull/5270 changed `Layer` so that it can no longer be used outside of a `<Grommet>` block. Since that previously was not the case, this is definitely a breaking change in the API surface, and so should have resulted in a major version bump if you are using semvar properly.

I've had to lock my project to `2.17.2` because `2.17.4` breaks a bunch of my tests.

### Expected Behavior

Layer should be able to be used outside of a Grommet block (ie, in tests)

### Actual Behavior

`TypeError: Cannot read property 'layer' of undefined`

### Steps to Reproduce

Write a test (or some code) that includes a layer outside of a Grommet block

### Your Environment

- Grommet version: 2.17.4
- Browser Name and version: Chrome: 92.0.4515.107
- Operating System and version (desktop or mobile): MacOS 11.3

britt6612:
#### What does this PR do?
This PR adds default values for both `OptionsContext` and `RootsContext` 
RootsContext.js
OptionsContext.js

Storybook
Take out `<Grommet/>` wrapper to ensure that `Layer`  is still being rendered correctly
There was no default value there was only `undefined` being passed through.
__This default value can be helpful for testing components in isolation without wrapping them.__
The default value needs to match what is being passed to the  `useContext` for instance `OptionsContext` would expect an object so the default value needs to be an object
closes #5516 
no
no
backwards compatible

### DIFF
--- a/src/js/components/FocusedContainer.js
+++ b/src/js/components/FocusedContainer.js
@@ -45,15 +45,14 @@ export const FocusedContainer = ({
 
   useEffect(() => {
     const timer = setTimeout(() => {
-      if (!hidden && trapFocus && roots && roots[0] !== null) {
+      if (!hidden && trapFocus && roots && roots[0]) {
         roots.forEach(makeNodeUnfocusable);
       }
     }, 0);
 
     return () => {
       // remove trap and restore ability to focus on the last root only
-      if (roots && roots[0] !== null)
-        makeNodeFocusable(roots[roots.length - 1]);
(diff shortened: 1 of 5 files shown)


MikeKingdom (approved):
Looks great. Thanks for doing this :)

taysea (approved):
Looks good. Thank you!

ericsoderberghp (approved):

PR #5525, opened 2021-08-12
Add object to EmptyMessages PropTypes

### LINKED ISSUE #5512: Select "emptySearchMessage" PropTypes should accept "object"/"JSX.Element"
### Expected Behavior

Property "emptySearchMessage" of "Select" component should accept "string" and "object"/"JSX.Element" as it will allow to customize the rendered node for empty result.

### Actual Behavior

Following error happen in console : 
```
Warning: Failed prop type: Invalid prop `emptySearchMessage` of type `object` supplied to `Select`, expected `string`.
```

https://codesandbox.io/s/grommet-v2-template-forked-3cqg8?file=/index.js

### Steps to Reproduce

1. Go to the link
2. See error in console

### Your Environment

- Grommet version: 2.17.4
- Browser Name and version: Firefox 90.0.2
- Operating System and version (desktop or mobile): Desktop Windows 10

britt6612:
#### What does this PR do?
This PR adds an object to the prop types
select/doc.js
storybook
add an object for `emptySearchMessage`
closes #5512 
already done
no
backwards compatible


ericsoderberghp:
- Looks like some unexpected DataTable changes were included.
- need to update index.d.ts as well
- "object" doesn't seem correct here, I think `PropTypes.node` might be more accurate?

Also, can you look into why we pass `option` to `SelectOption`? I can't see where it does anything but it is passed the `emptySearchMessage` value and we should see where that might end up.

britt6612:
> * Looks like some unexpected DataTable changes were included.
> * need to update index.d.ts as well
> * "object" doesn't seem correct here, I think `PropTypes.node` might be more accurate?
> 
> Also, can you look into why we pass `option` to `SelectOption`? I can't see where it does anything but it is passed the `emptySearchMessage` value and we should see where that might end up.

Yes I can look into that this afternoon!

britt6612:
> * Also, can you look into why we pass option to SelectOption? I can't see where it does anything but it is passed the emptySearchMessage value and we should see where that might end up.

So I looked into `option` as to where it may be going however since the `SeclectOption` is disabled this would not really end up somewhere. I did look at if a user was keeping a state  value `onChange={({ option }) => setValue(option)}` however since this is disabled this would never be included in the `onChange`

britt6612:
added to docs here 
https://github.com/grommet/grommet-site/pull/300

ericsoderberghp (approved):

PR #5526, opened 2021-08-12
fix extra box gap for empty child

### LINKED ISSUE #5523: Box gap is applied for empty children
Conditional rendering is not playing well with `gap`.
### Expected Behavior
If a `Box` has children which may be `null`, a gap shouldn't be applied for them.

### Actual Behavior
Extra gap is added to the layout if there are "missing" items.

https://codesandbox.io/s/ecstatic-mclaren-okuu1?file=/src/Box.js

vujevits:
Conditional rendering does not play well for Box with the `gap` property, it's applied for empty (null / <></>). This PR resolves this issue.

#### What does this PR do?
Checking if children are null with a helper, instead of relying it being truthy, which will always(?) be true if the result is coming from a render function.

None

The codesandbox example

Fixes #5523
No

Not aware of breaking changes

### DIFF
--- a/src/js/components/Box/Box.js
+++ b/src/js/components/Box/Box.js
@@ -5,6 +5,7 @@ import React, {
   useMemo,
   useState,
 } from 'react';
+import ReactDOMServer from "react-dom/server";
 
 import { ThemeContext } from 'styled-components';
 import { defaultProps } from '../../default-props';
@@ -88,7 +89,7 @@ const Box = forwardRef(
       contents = [];
       let firstIndex;
       Children.forEach(children, (child, index) => {
-        if (child) {
+        if (!isChildNull(child)) {
           if (firstIndex === undefined) {
             firstIndex = index;
           } else {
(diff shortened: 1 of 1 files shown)


src/js/components/Box/Box.js
```diff
@@ -164,6 +165,10 @@ const Box = forwardRef(
   },
 );
 
+const isChildNull = (children) => {
+  return !Boolean(ReactDOMServer.renderToStaticMarkup(children));
```
ericsoderberghp, line 169: Hmm... this feels like it could be rather heavy, rendering all children to string all of the time, especially if the children are complex.

I'm wondering if instead of doing it this way, we should explore a different means of adding the gap other than inserting extra elements into the DOM. For instance, using styled-components to include a child selector that adds a margin to all children except the last one?

vujevits: yes, I’m also worried about the performance implications, but haven’t really tested that aspect.

I’m happy to try your suggested approach.

jcfilben:
Closing due to inactivity, feel free to open

PR #5529, opened 2021-08-16
Add Typescript type for Button kind

### LINKED ISSUE #5527: Button doesn't support `kind` in Typescript
### Expected Behavior

`kind` prop on Button is not openly documented but needs to be part of the typescript definitions in order for typescript teams to use.

### Actual Behavior

Button prop `kind` not supported.

### Steps to Reproduce

1. Create Typescript project with HPE theme
2. Create <Button kind="toolbar" />
3. See error

### Your Environment

- Grommet version: Stable
- Browser Name and version: Mac Chrome

taysea:
#### What does this PR do?

Adds typescript support for `kind` on Button. While this is not a documented prop, we do use it the grommet-theme-hpe, so teams using typescript need it.

src/js/components/Button/index.d.ts

Tested in Custom Button tsx storybook example, before and after

Screen Shot 2021-08-16 at 2 57 47 PM

Screen Shot 2021-08-16 at 2 57 59 PM

Create tsx story and test `kind` prop.

Closes #5527 

No (unless we want to start documenting kind)

Yes, backwards compatible.


src/js/components/Button/index.d.ts
```diff
@@ -38,6 +38,7 @@ export interface ButtonProps {
   href?: string;
   target?: '_self' | '_blank' | '_parent' | '_top' | string;
   icon?: JSX.Element;
+  kind?: string | ButtonType;
```
halocline, line 41: Is this intended to be `ButtonKindType`?

```suggestion
  kind?: string | ButtonKindType;
```

src/js/components/Button/index.d.ts
```diff
@@ -11,7 +11,7 @@ import {
   Omit,
   PolymorphicType,
 } from '../../utils';
-
+import { ButtonKindType } from 'src/js/themes/base';
```
halocline, line 14: Rationale for explicit path versus relative here (e.g. `'../../themes/base'`)?

taysea: I think it auto-imported. let me update to relative.

src/js/components/Button/index.d.ts
taysea: Looking at the way we have done it in Pagination, I think it should actually be ButtonType. We can discuss but I am updating import statements to match.

src/js/components/Button/index.d.ts
```diff
@@ -38,6 +38,7 @@ export interface ButtonProps {
   href?: string;
   target?: '_self' | '_blank' | '_parent' | '_top' | string;
   icon?: JSX.Element;
+  kind?: string | ButtonThemeType;
```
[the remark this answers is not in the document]
taysea: After some discussion with @halocline, we've decided to simplify the type definition to just "string" for now since this is the typescript use case we are aiming to solve.

Internally in grommet, there are a few places where we assign `kind` to a theme object. However, in order to support this in a typescript project (where someone can define their own custom `kind` in their theme), there would need to be additional enhancements to our Button typing. Given that `kind` is not even a documented prop and just something we've used within grommet-theme-hpe, I think we should consider the added support of a button type a separate enhancement (and at that point want to consider documenting `kind` as well.

ericsoderberghp (approved):

halocline (approved):

PR #5533, opened 2021-08-18
Accessibility enhancements for RangeSelector

### LINKED ISSUE #5500: RangeSelector does not pass all jest-axe accessibility checks
### Expected Behavior

RangeSelector should not have accessibility violations and pass this test:
```
    const { container } = render(
      <Grommet>
        <RangeSelector values={[20, 30]} />
      </Grommet>,
    );

  });
```

### Actual Behavior

Running latest `jest-axe` (`^5.0.0`) which relies on `axe-core` (`^4.2.0` rule set) catches the following accessibility error:
Screen Shot 2021-08-02 at 1 30 45 PM

**Background**

- Current RangeSelector places an `aria-label="Lower Bounds"` and `aria-label="Upper Bounds"`on the corresponding thumb's `div`.
- According to [Dequeue](https://dequeuniversity.com/rules/axe/4.2/aria-allowed-attr?application=axeAPI) and [W3.org](https://www.w3.org/TR/html-aria/#dfn-no-corresponding-role), the use of any `aria-*` attribute on a `div` should be accompanied by a `role` attribute to better communicate with assistive technologies.
- `role="slider"` may be the most appropriate (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role)
  - If using `role="slider"`, also note advisement for use of `aria-valuemin`, `aria-valuemax`, and `aria-valuenow`.

### Steps to Reproduce

- Modify `package.json` to use `"jest-axe": "^5.0.1"`
- `yarn install`
- `yarn test`

### Your Environment

- Grommet version: 2.17.4
- Browser Name and version:

jcfilben:
#### What does this PR do?
In jest-axe v5 the RangeSelector accessibility pass is failing because the use of any aria-* attribute on a div should be accompanied by a role attribute. This PR adds the `role="slider"` attribute and `aria-valuemin`, `aria-valuemax`, and `aria-valuenow` as advised in https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_slider_role

Upgrade to jest-axe v5
run yarn test
accessibility test for RangeSelector should pass

Closes #5500 

no

no

backwards compatible

### DIFF
--- a/src/js/components/RangeSelector/RangeSelector.js
+++ b/src/js/components/RangeSelector/RangeSelector.js
@@ -195,6 +195,10 @@ const RangeSelector = forwardRef(
         />
         <EdgeControl
           a11yTitle={format({ id: 'rangeSelector.lower', messages })}
+          role="slider"
+          aria-valuenow={lower}
+          aria-valuemin={min}
+          aria-valuemax={max}
           tabIndex={0}
           ref={ref}
           color={color}
@@ -239,6 +243,10 @@ const RangeSelector = forwardRef(
         />
         <EdgeControl
           a11yTitle={format({ id: 'rangeSelector.upper', messages })}
+          role="slider"
+          aria-valuenow={upper}
+          aria-valuemin={min}
+          aria-valuemax={max}
           tabIndex={0}
           color={color}
           direction={direction}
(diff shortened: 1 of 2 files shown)


halocline (approved):
Thank you, @jcfilben.

ericsoderberghp (approved):

PR #5535, opened 2021-08-19
borders: Support passing arrays to allow per side border config.

DolceTriade:
BorderType suggests that it supports arrays to allow configuring each
side of the border independently of the other sides, however, when
actually trying to use an array of border objects, we find that they are
not respected. This commit adds support for arrays by recursively
merging each border entry and returns that. Added a storybook entry to
demonstrate this.

Check the storybook for "CustomBorders"

Storybook + my site.

Storybook

With this change, the grommet docs should now be more correct :)

It is not breaking as grommet suggests that this functionality is already supported.

### DIFF
--- a/src/js/components/Table/stories/CustomBorders.js
+++ b/src/js/components/Table/stories/CustomBorders.js
@@ -1,0 +1,68 @@
+import React from 'react';
+
+import {
+  Box,
+  Grommet,
+  Table,
+  TableBody,
+  TableCell,
+  TableFooter,
+  TableHeader,
+  TableRow,
+  Text,
+} from 'grommet';
+import { grommet } from 'grommet/themes';
+import { data, columns } from './data';
+
+export const Default = () => (
+  <Grommet theme={grommet}>
+    <Box align="center" pad="large">
+      <Table caption="Default Table">
+        <TableHeader>
+          <TableRow>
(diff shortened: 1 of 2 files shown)


jcfilben:
@DolceTriade I think this could be useful to add to grommet. In the description you mention that the documentation already suggests that this is supported. Which part of the documentation did you find this in? Looking through the Table documentation I wasn't able to find anything.

Also BorderTypes is currently defined in the propTypes file as:
```
const borderTypes = [
  PropTypes.bool,
  PropTypes.oneOf(sides),
  PropTypes.shape({
    color: PropTypes.oneOfType([
      PropTypes.string,
      PropTypes.shape({
        dark: PropTypes.string,
        light: PropTypes.string,
      }),
    ]),
    side: PropTypes.oneOf(sides),
    size: PropTypes.oneOfType([PropTypes.oneOf(sizes), PropTypes.string]),
  }),
];
```
Which doesn't include an array.

jcfilben:
It looks like the index.d.ts file was not aligned with the documentation or propTypes file. If we want to add this enhancement we need to update the documentation on grommet-site and the propTypes.

britt6612:
closing due to this being accomplished [here](https://github.com/grommet/grommet/pull/5774)

PR #5545, opened 2021-08-24
Fix prop type error in MaskedInput Story

britt6612:
#### What does this PR do?
This PR fixes a prop type error that was in the MaskedInput story
MaskedInput/customBox
storybook
run the story
```
react.development.js:220 Warning: Failed prop type: The prop `target` is marked as required in `Drop`, but its value is `undefined`.

```

There was an error because the ref was not yet defined so it was trying to pass `undefined` to `target` prop
no
no
backwards compatible

### DIFF
--- a/src/js/components/MaskedInput/stories/CustomBox.js
+++ b/src/js/components/MaskedInput/stories/CustomBox.js
@@ -4,7 +4,15 @@ import { grommet } from 'grommet/themes';
 
 export const CustomBoxMaskedInput = () => {
   const [value, setValue] = React.useState('');
+  const [dropProps, setDropProps] = React.useState({});
   const boxRef = React.useRef();
+
+  // Update the DropProps target once ref exists
+  React.useEffect(() => {
+    const nextDropProps = { ...dropProps };
+    nextDropProps.target = boxRef.current;
+    setDropProps(nextDropProps);
+  }, [boxRef]);
 
   return (
     <Grommet full theme={grommet}>
@@ -22,7 +30,7 @@ export const CustomBoxMaskedInput = () => {
         <Box flex width="medium" gap="medium">
           <MaskedInput
             plain
-            dropProps={{ target: boxRef.current }}
+            dropProps={dropProps}
             mask={[
               {
                 length: [1, 4],


ericsoderberghp:
hmm... why did we need this change? It seems like a fairly typical use of a `ref`.

britt6612:
> hmm... why did we need this change? It seems like a fairly typical use of a `ref`.

so on the first render the ref was coming back as `undefined` therefore it was throwing an error that `undefined` was being passed to `target` with was a `required` prop.

ericsoderberghp:
> > hmm... why did we need this change? It seems like a fairly typical use of a `ref`.
> 
> so on the first render the ref was coming back as `undefined` therefore it was throwing an error that `undefined` was being passed to `target` with was a `required` prop.

That feels like something we should handle within grommet. I don't think we should expect callers to have to add the extra state management to use a `ref`. We should treat an `undefined` ref in this scenario the same as if the caller didn't pass `dropProps`.

britt6612:
> > > hmm... why did we need this change? It seems like a fairly typical use of a `ref`.
> > 
> > 
> > so on the first render the ref was coming back as `undefined` therefore it was throwing an error that `undefined` was being passed to `target` with was a `required` prop.
> 
> That feels like something we should handle within grommet. I don't think we should expect callers to have to add the extra state management to use a `ref`. We should treat an `undefined` ref in this scenario the same as if the caller didn't pass `dropProps`.

Thats fair.. I can look at another approach.

britt6612:
> > > hmm... why did we need this change? It seems like a fairly typical use of a `ref`.
> > 
> > 
> > so on the first render the ref was coming back as `undefined` therefore it was throwing an error that `undefined` was being passed to `target` with was a `required` prop.
> 
> That feels like something we should handle within grommet. I don't think we should expect callers to have to add the extra state management to use a `ref`. We should treat an `undefined` ref in this scenario the same as if the caller didn't pass `dropProps`.

Thats fair.. I can look at another approach.

britt6612:
That feels like something we should handle within grommet. I don't think we should expect callers to have to add the extra state management to use a ref. We should treat an undefined ref in this scenario the same as if the caller didn't pass dropProps.

If I'm reading this correctly you would want to treat `undefined` as the default which is the `inputRef.current` however what this story is trying to display is the `drop` being under the whole parent `box` therefore they needed that `box.ref` in order to align the drop correctly. 
Below is the picture as if we were to treat `undefined` as the `inputRef.current`(notice the drop is not aligned to the far left )
Screen Shot 2021-08-24 at 4 43 23 PM

Whereas the image below the drop is aligned all the way to the far left once the `box.ref` is defined. Therefore it needed that parent `box` ref 

Screen Shot 2021-08-24 at 4 44 45 PM

ericsoderberghp:
I'm not saying we shouldn't pass the Box's ref, just that the caller shouldn't have to wrap the use of it in a `useState`.

Issue #5449, opened 2021-07-15
DataTable with select does not allow pinning the CheckBox column


LiveFreeAndRoam:
Support a method to pin the CheckBox column.

Horizontal scrolling causes the CheckBox column to scroll, rather than be pinned in place.

The CheckBox column scrolls:

https://codesandbox.io/s/datatable-select-horizontal-scrolling-epq16?file=/src/App.js

This was also mentioned in Slack: https://grommet.slack.com/archives/C09QQEG69/p1626320307320600

### Steps to Reproduce

See codesandbox.

### Your Environment

- Grommet version: 2.17.3
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): Any

PR #5547, opened 2021-08-24
Pin checkbox column for selectable DataTables

MikeKingdom:
For DataTables which are selectable and have one or more pinned columns, there is no way
to pin the checkbox column. The result is difficult to use and confusing since the checkbox column
scrolls out of view horizontally.

#### What does this PR do?
Whenever a DataTable that is selectable has any pinned columns, this automatically pins the checkbox column also.
However, if no columns are pinned, the checkbox column won't be pinned either.

DataTable.js

Storybook DataTable examples and jest tests.

In the Storybook examples for the DataTable, visit the Multiple pins example. If you scroll the table horizontally, the checkbox column will be pinned along with the other pinned columns.

#5449 

No

Yes

This should be backwards compatible.


src/js/components/DataTable/DataTable.js
```diff
@@ -160,20 +160,20 @@ const DataTable = ({
...
   const onHeaderWidths = useCallback(
     (columnWidths) => {
-      const pinnedProperties = columns
+      const hasSelectColumn = Boolean(select || onSelect);
+      let pinnedProperties = columns
         .map((pinnedColumn) => pinnedColumn.pin && pinnedColumn.property)
         .filter((n) => n);
-
+      if (hasSelectColumn && pinnedProperties.length > 0) {
+        pinnedProperties = ['_select', ...pinnedProperties];
```
MikeKingdom, line 168: I'd love to find a cleaner solution for the checkbox column than giving it the internal name `_select` if anyone has an idea.

taysea:
Screen Shot 2021-08-30 at 10 27 12 AM

The text seems to show up behind the pinned checkboxes -- should there be some default background color when pinned / give caller ability to theme or style via props?

MikeKingdom:
> Screen Shot 2021-08-30 at 10 27 12 AM
> 
> The text seems to show up behind the pinned checkboxes -- should there be some default background color when pinned / give caller ability to theme or style via props?

Great question. I think the checkbox cell should do this similar to the other cells:
```jsx
background={
     (pinnedOffset?._grommetDataTableSelect &&  cellProps.pinned.background) ||
     cellProps.background
}
```

ericsoderberghp (approved):

Issue #5548, opened 2021-08-25
Responsive Context is not updated when the scroll bar appears


bdeguigne:
If the scroll bar appears a few second after the first load, the responsive context is not updated. This is because the "resize" event listener does not trigger the event when the scroll bar appears or disappears.

### Expected Behavior

Responsive Context should be updated when the scroll bar appears or disappears

### Actual Behavior

Responsive Context is not updated when the scroll bar appears or disappears

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-issue-3xpo2?file=/src/App.js

### Steps to Reproduce

1. Resize the codeSandBox window width to : **784px**, the responsive context return : **"small"**
2. Reload the codeSandBox window 
3. The responsive context now return : **"medium"**

### Your Environment

The problem does not occur in macOS, because the scroll bar does not move the content.

- Grommet version: 2.17.4
- Browser Name and version: Chrome (Latest)
- Operating System and version (desktop or mobile): Windows 10

PR #5559, opened 2021-08-27
Fix : Responsive Context is not updated when the scroll bar appears

bdeguigne:
If the scroll bar appears a few second after the first load, the responsive context is not updated. This is because the "resize" event listener does not trigger the event when the scroll bar appears or disappears.

Creation of an custom hook "useScrollbarDetector" which detects the scrollbar that  appears  or disappears and then triggers the Grommet responsive context update method.

This solution uses the Resize Observer API and can be improve with this package : [resize-observer-polyfill](
https://www.npmjs.com/package/resize-observer-polyfill), for a better browser compatibility.

#### What does this PR do?
Responsive Context is updated when the scroll bar appears or disappears

src/js/utils/useScrollbarDetector.js

manual testing 

Follow the instructions in the "Steps to Reproduce" part on [the issue](https://github.com/grommet/grommet/issues/5548)

- Closes https://github.com/grommet/grommet/issues/5548

no


src/setupTests.js
```diff
@@ -1,6 +1,20 @@
...
+        // do nothing
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    unobserve() {
+        // do nothing
+    }
+}
+
+window.ResizeObserver = ResizeObserver;
```
jcfilben, line 20: I am confused about the purpose of this section, why does it need to be included?

jcfilben (commented):
Hey @bdeguigne sorry for the long wait on the review. I normally work on a mac and had to bring out my old windows laptop to reproduce the issue.

jcfilben:
Closing due to inactivity. I think we should explore an approach where we use `resizeObserver` directly in Grommet.js instead of calling the custom `useScrollBarDetector` hook.

Issue #5299, opened 2021-06-03
TextInput placeholder will not removed when it is JSX.Element


gavin99587:
TextInput placeholder can not removed when it is JSX.Element, while when the placeholder is a simple string ,it works as
 expected

<img width="464" alt="截屏2021-06-03 17 45 02" src="https://user-images.githubusercontent.com/11565497/120624600-71001700-c493-11eb-9496-1716ca0588b1.png">

if the placeholder is  a JSX.Element,when i change the value through a function  named handleBtnAllTap, the placeholder will not disappear, but is the placeholder is a simple string,it works as expected.

    <Box width={blockWrap.width} margin={{ top: '24px' }}>
          <Form
            value={formValue}
            onChange={(v) => setFormValue(v)}
            onSubmit={handleFormSubmit}
          >
            <Box direction="column">
              <Box direction="row" align={'center'} justify={'between'}>
                <FormFieldWrap
                  name="amount"
                  label={
                    <Text size="xsmall" color={CustomColor.c697281}>
                      Amount
                    </Text>
                  }
                >
                  <TextInput
                    name="amount"
                    type="text"
                    placeholder={
                      <Text size="small" color={'#697281'}>
                        I am placeholder
                      </Text>
                    }
                  />
                </FormFieldWrap>
                <Anchor size="small" onClick={handleBtnAllTap}>
                  All
                </Anchor>
              </Box>
            </Box>
          </Form>
        </Box>

1. 2. 3.

### Your Environment

- Grommet version:2.17.1
- Browser Name and version: chrome
- Operating System and version (desktop or mobile):

loongluangkhot:
Hey @ShimiSun, i'm a new contributor here, and i'd like to work on this issue! Wondering how do i self-assign / if you could assign this issue to me?

jcfilben:
@loongluangkhot feel free to go ahead and work on this issue

g4rry420:
@loongluangkhot Are you still working on this issue ? If not, Can I take this one ?

loongluangkhot:
@g4rry420 pls go ahead and do it. Sorry for super late reply!

jcfilben:
Closing this issue since #5571 has been merged.

PR #5571, opened 2021-09-02
Fix: TextInput placeholder bug

g4rry420:
#### What does this PR do?
Fixes the TextInput placeholder bug when added value to TextInput using a function

components => TextInput

Reproduce the code in the mentioned and test this

N/A

#5299 

No

Backwards compatible


src/js/components/TextInput/TextInput.js
```diff
@@ -300,6 +300,13 @@ const TextInput = forwardRef(
         !value,
     );
 
+    useEffect(() => {
+      if (!value || !showStyledPlaceholder) return;
+      setShowStyledPlaceholder(
+        placeholder && typeof placeholder !== 'string' && !value,
+      );
```
jcfilben, line 307: I'm wondering if this could be simplified to
```suggestion
      if (showStyledPlaceholder){
        setShowStyledPlaceholder(
          placeholder && typeof placeholder !== 'string' && !value,
        );
      }
```

jcfilben:
For reference here is the code I tested with
```javascript
import React from 'react';

import { Box, Button, Grommet, TextInput, Text } from 'grommet';

export const Simple = () => {
  const [value, setValue] = React.useState("");
  const handleClick = () => {
    setValue(100.5);
  };

  return (
      <Box pad="small">
        <TextInput
          name="amount"
          type="text"
          placeholder={
            <Text size="small" color={"#697281"}>
              I am placeholder
            </Text>
          }
          value={value}
          onChange={(event) => setValue(event.target.value)}
        />
        <Button size="small" label="Click to Change Value" onClick={handleClick} />
      </Box>
    </Grommet>
  );
};
```

src/js/components/TextInput/TextInput.js
jcfilben: When I test, I am not seeing a difference between having `showStyledPlaceholder && value` vs just `showStyledPlaceholder` in the if statement, even when I set value=""

g4rry420:
```
      if (showStyledPlaceholder){
        console.log('I am fired')
        setShowStyledPlaceholder(
          placeholder && typeof placeholder !== 'string' && !value,
        );
      }
```

https://user-images.githubusercontent.com/61521805/131904581-a0dabbc7-1917-4f69-a0cd-1e9dadb88ebb.mp4

With only `showStyledPlaceholder`, the setShowStyledPlaceholder is fired when on the first mount of the component and when the value in the TextInput becomes empty. Please, let me your thoughts.

jcfilben: Thanks for sharing the video, I see now why we should have `value` in the if statement. I think we should include it

jcfilben (approved):

ericsoderberghp:
Something about the overall approach of using `useState` for `showStyledPlaceholder` feels heavier than it needs to be.

I'm wondering if we can replace the `useState` with something like:
```
const showStyledPlaceholder = useMemo(
  () =>
    placeholder &&
    typeof placeholder !== 'string' &&
    !(inputRef.current && inputRef.current.value) &&
    !value,
  [inputRef, placeholder, value],
);
```
and then remove the `useEffect` and places where were calling `setShowStyledPlaceholder`.

g4rry420:
@ericsoderberghp  This is a nice suggestion and it works well with the usecases. Thanks

ericsoderberghp (approved):

PR #5572, opened 2021-09-02
Add defaultProps import

mecirmartin:
Signed-off-by: Martin Mecir mecir.martin@gmail.com

Related to #5229

What does this PR do?

Like other Grommet components, DataTable will default to defaultProps.theme if there's no Theme context available. However defaultProps isn't imported in the Body component inside DataTable component. This PR adds that missing import.

Line 14 is the change; line 158 is the reference to defaultProps

You could use a DataTable outside of a ThemeContext -- it will fail without this change.

I hit this bug while trying to reproduce the Datatable example from `grommet-sandbox` in separate project. I used it as standalone component and that probably broke ThemeContext enough that my DataTable tried to fall back to (the non-existent) defaultProps.

Nope

Not unless you list every bugfix.

Compatible.


jcfilben (approved):

MikeKingdom (approved):
LGTM :)

ericsoderberghp (approved):

ericsoderberghp:
Thanks for your contribution!

PR #5573, opened 2021-09-02
add prop orderAlign for list order

### LINKED ISSUE #5567: Enhance List to align text
As of now there is not a way to align the text for List 
### Expected Behavior
You may want the context to align start to the index
It is align-center now

### Actual Behavior
Screen Shot 2021-08-31 at 2 46 47 PM

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

britt6612:
#### What does this PR do?
This PR adds a `orderAlign` prop 
list.js
storybook
Changed the one-story  `order` to use this prop 
THIS STORY SHOULD NOT BE MERGED JUST CHAGED IT FOR TESTING PURPOSES
When there was longer content  within the list it was being `aligned in the center` this makes to hard to follow horizontally 

**I added the one prop to only adjust the alignment of the index and the content. If we want to also add something for the user to control the `controls` alignment then maybe we can discuss and expand on the prop or leave it how it is.**
closes #5567 
none
yes I will update docs in grommet site
yes a new prop
 backwards compatible

### DIFF
--- a/src/js/components/List/List.js
+++ b/src/js/components/List/List.js
@@ -132,6 +132,7 @@ const List = React.forwardRef(
       step = paginate ? 50 : undefined,
       onClickItem,
       onMore,
+      orderAlign,
       ...rest
     },
     ref,
@@ -488,7 +489,7 @@ const List = React.forwardRef(
                   boxProps = { ...boxProps, ...itemProps[index] };
                 }
 
-                return (
+                return !orderAlign ? (
                   <StyledItem
                     key={key}
                     tag="li"
@@ -502,6 +503,24 @@ const List = React.forwardRef(
                   >
                     {onOrder && <Text>{index + 1}</Text>}
                     {content}
+                    {orderControls}
+                  </StyledItem>
+                ) : (
+                  <StyledItem
+                    key={key}
+                    tag="li"
+                    flex={false}
+                    pad={pad || theme.list.item.pad}
+                    background={adjustedBackground}
+                    border={adjustedBorder}
+                    {...boxProps}
+                    {...clickProps}
+                    {...orderProps}
+                  >
+                    <Box direction="row" align={orderAlign} flex {...boxProps}>
+                      {onOrder && <Text>{index + 1}</Text>}
+                      {content}
+                    </Box>
                     {orderControls}
                   </StyledItem>
                 );
--- a/src/js/components/List/index.d.ts
+++ b/src/js/components/List/index.d.ts
@@ -2,6 +2,7 @@ import * as React from 'react';
 import {
   A11yTitleType,
   AlignSelfType,
+  AlignType,
   GridAreaType,
   MarginType,
   PadType,
@@ -53,6 +54,7 @@ export interface ListProps<ListItemType> {
     | ((event: React.MouseEvent) => void)
     | ((event: { item?: ListItemType; index?: number }) => void);
   onOrder?: (orderedData: ListItemType[]) => void;
+  orderAlign?: AlignType;
   pad?: PadType;
   paginate?: boolean | PaginationType;
   primaryKey?: string | ((item: ListItemType) => React.ReactElement);
--- a/src/js/components/List/propTypes.js
+++ b/src/js/components/List/propTypes.js
@@ -61,6 +61,7 @@ if (process.env.NODE_ENV !== 'production') {
     onMore: PropTypes.func,
     onClickItem: PropTypes.func,
     onOrder: PropTypes.func,
+    orderAlign: PropTypes.string,
     pad: PropTypes.oneOfType([padPropType]),
     paginate: PropTypes.oneOfType([PropTypes.bool, PropTypes.object]),
     primaryKey: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
--- a/src/js/components/List/stories/Order.js
+++ b/src/js/components/List/stories/Order.js
@@ -4,9 +4,14 @@ import { Grommet, Box, List } from 'grommet';
 import { grommet } from 'grommet/themes';
 
 const locations = [
-  'Boise',
-  'Fort Collins',
-  'Los Gatos',
+  `Lorem ipsum dolor sit amet, consectetur adipiscing elit,
+   sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.`,
+  `Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+   eiusmod tempor incididunt ut labore et dolore magna aliqua.`,
+  `Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+   eiusmod tempor incididunt ut labore et dolore magna aliqua.
+   Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
+  eiusmod tempor incididunt ut labore et dolore magna aliqua.`,
   'Palo Alto',
   'San Francisco',
 ];
--- a/src/js/components/__tests__/__snapshots__/components-test.js.snap
+++ b/src/js/components/__tests__/__snapshots__/components-test.js.snap
@@ -519,6 +519,7 @@ Object {
       "onClickItem": [Function],
       "onMore": [Function],
       "onOrder": [Function],
+      "orderAlign": [Function],
       "pad": [Function],
       "paginate": [Function],
       "primaryKey": [Function],


src/js/components/List/List.js
```diff
@@ -504,6 +505,24 @@ const List = React.forwardRef(
...
+                    tag="li"
+                    flex={false}
+                    pad={pad || theme.list.item.pad}
+                    background={adjustedBackground}
+                    border={adjustedBorder}
+                    {...boxProps}
+                    {...clickProps}
+                    {...orderProps}
+                  >
+                    <Box direction="row" align={orderAlign} flex {...boxProps}>
```
halocline, line 520: Won't `{...boxProps}` always overwrite what is specified in `direction` and `align`? Should it precede these props?

src/js/components/List/List.js
```diff
@@ -132,6 +132,7 @@ const List = React.forwardRef(
       step = paginate ? 50 : undefined,
       onClickItem,
       onMore,
+      orderAlign,
```
halocline, line 135: Thinking on naming here. Also wondering if we are committing to something quite narrow here. What if I want the ability to align my "up/down" controls? Should we be considering that case at this time?

halocline: Are callers going to want to be able to affect any other Box props?

Would something like
```
order?: { index?: { align?: AlignType } }
```
provide more flexibility long term? This interface could be extended to allow for additional styling such as allowing for pad or allowing for control of the up/down controls, without crowding the top-level List API. 

Wondering aloud here to solicit others thoughts...

src/js/components/List/List.js
britt6612: Your right changing now

src/js/components/List/List.js
taysea: It's a good point, I think if we plan to enhance to support control alignment in the future, then starting with a more flexible prop name would be valid. I was torn between simplicity with the given use case vs flexibility.

src/js/components/List/stories/Order.js
```diff
@@ -4,9 +4,14 @@ import { Grommet, Box, List } from 'grommet';
 import { grommet } from 'grommet/themes';
 
 const locations = [
-  'Boise',
-  'Fort Collins',
-  'Los Gatos',
+  `Lorem ipsum dolor sit amet, consectetur adipiscing elit,
```
halocline, line 7: I'm not seeing the new prop being tested here.

halocline (commented):
A Jest snapshot test for this new prop would be nice.

src/js/components/List/List.js
halocline: The tension is real 😉

britt6612: Yes I also was thinking the same:

I also thought about expanding the `itemProps` to allow for this but my hesitation there was I would assume to be consistant a user would not want to only `align` a certain row however I could be wrong so open to suggestions.. 
otherwise, they would have to do something like  ` itemProps={{ global: { align: 'start' }}}` so that did not feel like the best approach.. 

Also we could do another Prop maybe  `Align table` something and then that would accept the object of the content and the index as well as the controls ..
`AlignTable: {controls: {align: 'start'}}`

src/js/components/List/stories/Order.js
britt6612: just added sorry forgot to push that part up

britt6612:
> A Jest snapshot test for this new prop would be nice.

I was kinda waiting until we decided on a clear direction before adding test. I will add one though 👍

ericsoderberghp:
I'm wondering if we should add `itemProps: BoxTypes`, which would allow anything Boxy. It wouldn't be tied to `onOrder` and could be used anytime.

britt6612:
> I'm wondering if we should add `itemProps: BoxTypes`, which would allow anything Boxy. It wouldn't be tied to `onOrder` and could be used anytime.

Only one way to find out!! Ill play around with that and see

halocline:
> I'm wondering if we should add `itemProps: BoxTypes`, which would allow anything Boxy. It wouldn't be tied to `onOrder` and could be used anytime.

I like this direction. May be a bit buried, but I think it provides a great deal of flexibility.

britt6612:
Just to make sure we are on the same page @ericsoderberghp 
As a caller if I wanted to change the `IndexAlignment` I would  do something like:

```
<List
  ...
  itemProps={{ 
      orderIndex: {
        align: 'start',
      },
  }}
...
</List>
```


or were you thinking it would be like this?

```
<List
  ...
  itemProps={{ 
    boxTypes: {
      orderIndex: {
        align: 'start',
      },
    },
  }}
...
</List>
```

ericsoderberghp:
I was thinking just `<List itemProps={{ align: 'start' }} />`

britt6612:
> I was thinking just `<List itemProps={{ align: 'start' }} />`

I was thinking we would have a prop name for changing the index & content than another prop name for the controls. So that is why I was thinking something like `IndexAlignment`  than a `ControlsAlignment` 

`IndexAlignment` would be passed to a `box` that contained the `index` and content
Screen Shot 2021-09-03 at 10 26 26 AM

Than the `ControlsAlignment` would be any `boxprops` that would then be passed the the `box` containing `controls`
Screen Shot 2021-09-03 at 10 27 29 AM

For what we need in the design system is disregarding the `controls` I was just adding flexibility for grommet

britt6612:
So I put the `StyledItem {...itemProps} />`
and updated the story you can see with not tying it to `onOrder` then this styling would effect the whole content as a whole including the controls

britt6612:
@halocline @taysea 
We decided to have one prop that would affect the whole alignment including controls and would not be tied to `onOrder` wanted to get your thoughts on naming. as of now, I have `generalItemProps`

halocline:
> We decided to have one prop that would affect the whole alignment including controls and would not be tied to `onOrder` wanted to get your thoughts on naming. as of now, I have `generalItemProps`

Throwing out ideas:

- `itemsProps`
  - Pros: Will be presented directly adjacent to `itemProps` in documentation, which will help folks directly compare contrast the purpose of each (e.g. `itemProps` allows for targeting a specific item, whereas `itemsProps` applies to all items).
  - Con: Visually similar which may lead to some initial confusion.

Eliminating my other ideas because I don't like them. ☹️

src/js/components/List/List.js
```diff
@@ -499,6 +500,7 @@ const List = React.forwardRef(
                     {...boxProps}
                     {...clickProps}
                     {...orderProps}
+                    {...generalItemProps}
```
halocline, line 503: If I specify both `itemProps` and `generalItemProps` but they have a conflict (e.g. border, background, pad), which styling wins? Which should win?

britt6612:
> > We decided to have one prop that would affect the whole alignment including controls and would not be tied to `onOrder` wanted to get your thoughts on naming. as of now, I have `generalItemProps`
> 
> Throwing out ideas:
> 
> * `itemsProps`
>   
>   * Pros: Will be presented directly adjacent to `itemProps` in documentation, which will help folks directly compare contrast the purpose of each (e.g. `itemProps` allows for targeting a specific item, whereas `itemsProps` applies to all items).
>   * Con: Visually similar which may lead to some initial confusion.
> * `itemContainerProps`, `itemContainer`, `itemsContainer`
> 
> Eliminating my other ideas because I don't like them. ☹️

I like the (e.g. `itemProps` allows for targeting a specific item, whereas `itemsProps` applies to all items).
However I could see where that could lead to problems I like the `itemContainerProps` its better in the fact that it doesn't  cause confusion within the props.

src/js/components/List/List.js
britt6612: good question..
I guess if someone is wanting to apply `itemProps` they are going to want it to a specific index so wondering if that should win.. I could see someone setting `background: 'background'` for the whole list then doing something like `itemProps: { index 1: background: 'contrast'}`

britt6612: Sounds good I’ll look into that!

ericsoderberghp:
I feel like `itemsProps` is too subtly different and will thus be more confusing.

Here are some other ideas I've thought of:

**keep it all within `itemProps` but add a special key** designed to not conflict with potential data item keys, perhaps something like '_' or '*'. This feels a little too code-y to me.

**keep it all within `itemProps` but increase the depth** of the object by something like: `itemProps={{ default: {}, primaryKeys: { <primary_key>: {} }}`. This does have the advantage in that it reinforces that the item identifiers are the primary keys. But, I'm wondering if it's starting to feel a bit too deep as an property value object.

**new `defaultItemProps` property** I think the term "default" has some reasonable precedent, as in "defaultValue" and "value".

halocline:
> I feel like `itemsProps` is too subtly different and will thus be more confusing.
> 
I agree and we can eliminate this.

> **keep it all within `itemProps` but increase the depth** of the object by something like: `itemProps={{ default: {}, primaryKeys: { <primary_key>: {} }}`. This does have the advantage in that it reinforces that the item identifiers are the primary keys. But, I'm wondering if it's starting to feel a bit too deep as an property value object.
> 
Initially I was attracted to this, but am now weary about the depth which introduces more friction in discoverability and learning curve.

> **new `defaultItemProps` property** I think the term "default" has some reasonable precedent, as in "defaultValue" and "value".
👍  This would be my vote. `defaultItemProps` = defaults set to all items --> `itemProps` = ability to apply specific treatments to specific items. This seems allow for clarity and ease of explanation.

britt6612:
Changed name to `defaultItemProps` once approved I will revert storybook changes back just left them for preview to test

src/js/components/List/index.d.ts
```diff
@@ -44,6 +45,7 @@ export interface ListProps<ListItemType> {
   border?: BorderType;
   data?: ListItemType[];
   gridArea?: GridAreaType;
+  generalItemProps?: BoxTypes;
```
taysea, line 48:
```suggestion
  defaultItemProps?: BoxTypes;
```

taysea (commented):
I would expect for the single line items to be aligned with the controls and only multi-line items would align more vertically. Just looks a bit strange for the single line items to be horizontally above the controls:

Screen Shot 2021-09-08 at 1 31 36 PM

britt6612:
> I would expect for the single line items to be aligned with the controls and only multi-line items would align more vertically. Just looks a bit strange for the single line items to be horizontally above the controls:
> 
> Screen Shot 2021-09-08 at 1 31 36 PM

Yeah I ran this by Vicky and she was okay with the single lines being aligned start as well that way its overall consistant

src/js/components/List/index.d.ts
britt6612: thanks for catching

taysea (approved):
Approved, just need to remove story before merging!

src/js/components/List/List.js
```diff
@@ -499,6 +500,7 @@ const List = React.forwardRef(
                     {...boxProps}
                     {...clickProps}
                     {...orderProps}
+                    {...defaultItemProps}
```
halocline, line 503: Please check on the order in which this should be placed. I believe `defaultItemProps` will trump `itemProps` in this implementation, however, I believe we want the opposite.

halocline (changes requested):

src/js/components/List/List.js
britt6612: Yes I checked seemed like the ‘itemProps’ took over the ‘defaultItemProps’ I added an example in the story can you check there and make sure we’re on the same page?

britt6612: my mistake your correct..

halocline (commented):
Looking good, thanks for updating.

Since this is a new prop, should we have a snapshot test accompanying it?

britt6612:
> Since this is a new prop, should we have a snapshot test accompanying it?

Yes now that we have agreed on the name I will add test

src/js/components/List/stories/Order.js
```diff
@@ -16,7 +22,12 @@ export const Order = () => {
   return (
     <Grommet theme={grommet} role="application">
       <Box align="center" pad="large">
-        <List data={ordered} onOrder={setOrder} />
+        <List
+          itemProps={{ 0: { background: 'pink' } }}
```
ericsoderberghp, line 26: I think adding `itemProps` and `defaultItemProps` to this story sort of distract from the focus of `onOrder`. I'm thinking we should just leave this story as was. We've got the unit test to cover snapshot testing.

ericsoderberghp (commented):
Looking good overall. Just one comment.

src/js/components/List/stories/Order.js
britt6612: yes sorry that was my intentions to take this out once got approved

halocline (commented):
Does `.../List/propTypes.js` need a definition for `defaultItemProps`?

src/js/components/List/propTypes.js
```diff
@@ -57,6 +57,7 @@ if (process.env.NODE_ENV !== 'production') {
       PropTypes.oneOfType([PropTypes.string, PropTypes.shape({})]),
     ),
     children: PropTypes.func,
+    defaultItemProps: PropTypes.shape({}),
```
halocline, line 60: In the `index.d.ts` file we are supporting BoxTypes. I would think we'd want our propType and TypeScript definitions to have parity (i.e. be the same). Should this be `BoxPropTypes`?

britt6612: Let me test out make sure!

halocline (approved):

ericsoderberghp (approved):

Issue #5429, opened 2021-07-07
Fix Calendar Accessibility with Selected Dates


amandadupell:
Related to #4068 and #5319 

### Expected Behavior

When the user selects a date or a range of dates, the selected date(s) should be relayed to the user via the screen reader in some way.

### Actual Behavior

When a user selects a date or a range of dates, nothing is relayed to the user that a date was selected or a change occurred.

### Video exhibiting the issue

https://user-images.githubusercontent.com/38971752/124836905-ce372e80-df51-11eb-8342-59026d6c7538.mov

### Steps to Reproduce

1. Use MacBook Pro VoiceOver or another screen reader and open the simple story for Calendar in Storybook
2. Using arrow key commands (control + option + arrow keys), navigate to the table of dates
3. Choose any date by selecting with the space bar or enter key

### Your Environment

- Grommet version: 2.17.3
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): MacBook Pro Desktop

PR #5575, opened 2021-09-03
Fix/cal-select-accessibility: updated label to include selected date/s/range

amandadupell:
#### What does this PR do?

Includes selected date/dates/range for calendar so a user with a screen reader is aware if a date/s is selected upon entering the grid

Calendar

Manual testing all stories and ensuring tests still pass as expected

Storybook (Multiple, Range, Simple stories are a good place to start but you can test any of the stories)

#5429 

No

No

Backwards compatible


britt6612 (approved):
Look Good! Nice Work

src/js/components/Calendar/__tests__/__snapshots__/Calendar-test.js.snap
```diff
@@ -17230,7 +17277,11 @@ exports[`Calendar select dates 1`] = `
         </div>
       </div>
       <div
-        aria-label="January 2020"
+        aria-label="
+                January 2020;
+                Currently selected 
+  2020-01-12 ,Range: 2020-01-08 ,2020-01-10
```
taysea, line 17283: I'm not sure I understand the difference between "currently selected" vs "range" -- is this which date is active vs which dates have been selected previously for range?

amandadupell: there isn't a difference between currently selected and range. in this test, as long as i'm understanding it correctly, january 12 is selected and the range from january 8th to january 10th is also selected. so something like this situation: 

Screen Shot 2021-09-10 at 5 08 39 PM

so maybe i could better phrase it so it's clear that all of the read aloud dates are selected? what do you think?

taysea: Got it -- maybe it could read "2020-01-08 through 2020-01-10, 2020-01-12"? 

Unless other calendars have common convention of saying "Range:" to screen reader users, but my though was to keep it more conversational if possible.

taysea (approved):

ericsoderberghp (approved):

PR #5580, opened 2021-09-07
Accessible Select - DRAFT

### LINKED ISSUE #5501: Select does not pass all jest-axe accessibility checks
### Expected Behavior

Select should not have accessibility violations and pass this test:
```
    const { container } = render(
      <Grommet>
        <Select options={['one', 'two', 'three']} a11yTitle="test" />
      </Grommet>,
    );

  });
```

### Actual Behavior

Running latest `jest-axe` (`^5.0.0`) which relies on `axe-core` (`^4.2.0` rule set) catches the following accessibility error:
Screen Shot 2021-08-02 at 1 49 46 PM

**Background**

- Current Select places an `<input />` nested within a `<button />`
Screen Shot 2021-08-02 at 1 54 06 PM
- According to [Dequeue & WCAG 4.1.2](https://dequeuniversity.com/rules/axe/4.2/nested-interactive?application=axeAPI) "interactive controls must not have focusable descendants." 
- Why it matters: "Focusable elements with an interactive control ancestor (any element that accepts user input such as button or anchor elements) are not announced by screen readers and create an empty tab stop. That is, you could tab to the element but the screen reader will not announce its name, role, or state."

### Steps to Reproduce

- Modify `package.json` to use `"jest-axe": "^5.0.1"`
- `yarn install`
- `yarn test`

### Your Environment

- Grommet version: 2.17.4
- Browser Name and version:

jcfilben:
#### What does this PR do?
Currently Select is built with a DropButton and a TextInput. Although the TextInput is readonly, this still causes an accessibility issue. Button cannot have any focusable descendants.

This PR adds an option for an accessible Select that is built using Text instead of TextInput. This PR is still a work in progress. Some of the things I am still working on:
- The original Select has a default width. I can't figure out where this is coming from. TextInput doesn't normally have a set width and when using Text the width of the Select stretches/shrinks to fit the text unless wrapped in a box with a set width.
- When a long option is selected in the original Select you can scroll to see the rest of the text. In the Select with Text if Text is wrapped with a box with a set width, you can scroll but a scroll bar shows up whereas in the original select no scroll bar shows.

Select.js

Storybook
`yarn test`

Storybook Select/Simple story
Try wrapping <SelectText> with <Box width="small"> (Lines 374-382 in Select.js) to see scroll bar in Select with long option
`yarn test` after upgrading to `"jest-axe": "^5.0.1"` (Original select will fail accessibility test. Accessible Select will pass)

Closes #5501 

Yes

Yes

Backwards compatible


ericsoderberghp:
First, I wanted to clarify that the `accessible` property is just there for this exploration and not something we would eventually publish. Am I right?

Second, not having the actual value in a DOM form input element (select, input, textarea, etc.) means it won't work for uncontrolled forms. I think we might need to have a hidden `<select />` or `<input />` somewhere that takes care of this.

jcfilben:
> First, I wanted to clarify that the accessible property is just there for this exploration and not something we would eventually publish. Am I right?

Yes the `accessible` prop is just there for testing right now to make is easy to compare the original select to the select built using text.

> Second, not having the actual value in a DOM form input element (select, input, textarea, etc.) means it won't work for uncontrolled forms. I think we might need to have a hidden <select /> or <input /> somewhere that takes care of this.

I will look into adding a hidden input to make select work for uncontrolled forms

jcfilben:
**Update**
An explanation of what I have tried so far:

1. Setting the width of the Select with a hidden TextInput
This would solve the width issue but not the scrollbar issue. A hidden input needs to be included anyway if TextInput is switched to Text so that Select would work correctly with uncontrolled forms, as Eric mentioned above. The idea was to use this hidden input to set the size of the Select so that it matched the original Select that was built with TextInput. The problem was when `type="hidden"` was added to the TextInput it no longer had any width.

2. Using a Stack
This method would allow us to keep the TextInput and fix the accessibility issue by having the DropButton over the TextInput instead of having the TextInput as a child of DropButton. Everything about this solution worked except when you selected a long option you could no longer horizontally scroll to see it. In a Stack only one of the children can be interacted with. In this case the DropButton is the one you can interact with so you can't interact with the TextInput to scroll.

jcfilben:
After discussing with @halocline I attempted a potential solution where the hidden input discussed above was wrapped in a Box with a set width. The width of this Box would be determined by the font-size, font-family, and font-weight  in an effort to replicate how the native HTML input determines width. In addition to correctly setting the width on the Box we would also need to hide the scrollbar but still allow for scrolling to replicate the behavior of our current Select and ensure backwards compatibility. Although this method is possible it would require different solutions per browser and could make the Select component more difficult to maintain, as well as open the door for potential problems with Select.

As it is currently built, the Select component is following accessibility guidelines. Jest-axe is flagging it because the component is built using a TextInput within a DropButton. The motivation for this is that according to Dequeue and WCAG,  "interactive controls must not have focusable descendants". **Jest-axe does not take into account that the TextInput is read only and has a tab index of -1. So in practice our Select as it is currently built is accessible and following WCAG guidelines.**

**With this being considered, I think the potential for problems that may arise from changing the structure of Select and the obfuscated code that would be required to achieve the correct sizing and scroll bar styling outweighs the benefit.**

Instead of changing the Select component I am changing the pull request to disable the 'nested-interactive' rule from the Select accessibility tests and include a comment explaining why the rule is being disabled.

ericsoderberghp:
That is a beautiful description 🥇

jcfilben:
I am going to close this PR and open a new one that upgrades the jest-axe version and disables the nested-interactive rule on the Select accessibility tests since I can't do one and not the other without failing tests.

PR #5581, opened 2021-09-07
remove box container from Markdown table cells

### LINKED ISSUE #5486: Markdown italics causes linebreak
### Expected Behavior

Italics should not cause a linebreak, should just be part of the text.

### Actual Behavior

When italics are used around some text in a sentence, it forces a line break at the end.

https://designer.grommet.io/markdown-table?id=Demonstration-edgar-circenis-hpe-com

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.17.4

MikeKingdom:
Currently Markdown uses `TableCell` for any `<td>` elements generated from the markdown content.
`TableCell` wraps the cell content in a `Box`. However, since `Box` is `display:flex` and `direction=column` so
causes any inline content like `<em>`, `<a>`, `<b>`, etc, to end up `display:block` rather than being inline.

#### What does this PR do?
This PR removes the `Box` wrapper for `<td>` cells in markdown content so that any other interesting inline
elements in the cells continue to be inline.

From what we can tell, we don't need the `Box` wrapper here.

Markdown.js

Storybook, Jest and some handcrafted examples.

Go to the Markdown Simple example in the Storybook and verify the table cells with other markup
continue to be inline as desired.

Fixes #5486 

No

Yes

Should be backwards compatible although it may be hard to predict what interesting markdown people create.

### DIFF
--- a/src/js/components/Markdown/Markdown.js
+++ b/src/js/components/Markdown/Markdown.js
@@ -31,7 +31,7 @@ const GrommetMarkdown = ({ components, options, theme, ...rest }) => {
       img: { component: Image },
       p: { component: Paragraph },
       table: { component: Table },
-      td: { component: TableCell },
+      td: { component: TableCell, props: { plain: true } },
       tbody: { component: TableBody },
       tfoot: { component: TableFooter },
       th: { component: TableCell },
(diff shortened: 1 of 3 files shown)


ericsoderberghp (approved):

halocline (approved):

taysea (approved):
LGTM.

PR #5582, opened 2021-09-07
Fix: rel attribute added to Button

### LINKED ISSUE #5537: Button typedef does not allow `rel` attribute
Type definition of Button component is `declare const Button: React.FC<ButtonExtendedProps>;`. That includes some anchor props, but not `rel`
```ts
  href?: string;
  target?: '_self' | '_blank' | '_parent' | '_top' | string;
```

So TypeScript compiler throws an error when creating
```tsx
<Button href="https://example.com" target="_blank" rel="noopener">external</Button>
```

### Expected Behavior

Button component should allow attribute `rel` (commonly accompanying `target="_blank"`)

### Actual Behavior

Typescript compiler throws `Property 'rel' does not exist on type 'IntrinsicAttributes & ButtonExtendedProps & { children?: ReactNode; }'.  TS2322`

https://codesandbox.io/s/grommet-v2-template-forked-rc3g2?file=/index.tsx
(open and wait a bit for typescript to detect `rel` as an error)

### Your Environment

- Grommet version: v2.17.4
- TypeScript version: 4.3.5

jlaluces123:
#### What does this PR do?
Aims to resolve #5537 by adding the `rel` attribute (in the anchor tag context) to Button.

`Button/index.d.ts` to view the `RelType` list and see how it was added to Button

Manual testing by adding `rel` to a `Button` component to make sure TypeScript no longer complains about the attribute

Add a `rel` attribute to any `Button` as you would an anchor tag and take notice that TypeScript does not give a warning / error

#5537
Yes, will be in its own PR under the grommet-site repo

No need

Backwards compatible


src/js/components/Button/index.d.ts
```diff
@@ -14,6 +14,22 @@ import {
 
 import { TipProps } from '../Tip';
 
+export type RelType =
```
ericsoderberghp, line 17: Isn't `rel` part of `JSX.IntrinsicElements['anchor']`? I'd like to find a way to leverage the built in intrinsic element props, rather than having to hard code them ourselves.

jlaluces123: Ahh definitely agree, going to tinker around with this to see what I can find!

ericsoderberghp:
I think we should also extend one of the typescript stories to include a `href` and `rel` properties and verify that typescript is happy.

ericsoderberghp (approved):

PR #5587, opened 2021-09-09
Changed Markdown to support refs

ericsoderberghp:
#### What does this PR do?

Changed Markdown to support refs

Had to add a wrapping `<div />` to be able to tie a ref in. :(

grommet-designer, where it was discovered

no

yes

backwards compatible for the user and caller. Any DOM snapshot based tests will have an additional `<div />` wrapping Markdown content.


jcfilben (approved):
Looks good to me

src/js/components/Markdown/Markdown.js
```diff
@@ -15,36 +15,44 @@ import { TableHeader } from '../TableHeader';
...
+      options && options.overrides,
+    );
 
-  return <Markdown options={{ ...options, overrides }} {...rest} />;
-};
+    // We have to wrap in a div because markdown-to-jsx is a
+    // functional component that doesn't handle refs.
+    return (
+      <div ref={ref}>
+        <Markdown options={{ ...options, overrides }} {...rest} />
```
MikeKingdom, line 51: I tried setting options.wrapper= props => <div ref={ref} {...props} />. It definitely keeps the one <div> wrapper. I didn't verify the ref worked as expected.

It's hard to tell from the docs if that would potentially get used somewhere other than the toplevel.

MikeKingdom (approved):
Cool. I like it.

Issue #5557, opened 2021-08-26
DateInput/Calendar arrow keys move active date offscreen but doesn't move with it


taysea:
### Expected Behavior

If user uses up/down arrow and leaves the current month, the month should change to the previous/next month to keep the active date in view. Similar to behavior here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

### Actual Behavior

When user pressed up/down and leaves the current month, the active date becomes out of view.

### URL, screen shot, or Codepen exhibiting the issue

http://localhost:9001/?path=/story/visualizations-calendar-simple--simple

### Steps to Reproduce

1. Tab to Calendar
2. Press up
3. Active date is not in view.

### Your Environment

- Grommet version: Stable
- Browser Name and version: Mac Chrome
- Operating System and version (desktop or mobile): MacOS Big Sur

PR #5588, opened 2021-09-10
Fix/cal-5557: added keyboard switching view

amandadupell:
#### What does this PR do?

Improved the keyboard to keep active date in view when using the arrow keys

Calendar.js

Tested across multiple Calendar and DateInput stories

Test across Calendar and DateInput stories

#5557 

#### Video of implemented behavior

https://user-images.githubusercontent.com/38971752/132919959-4b2f9f3a-f017-49f3-b012-bc032c8b7297.mov

No

No

Backwards compatible


taysea (commented):
Overall the behavior looks good. The only thing that was a little strange in the experience was when a date from the next month is visible in my current calendar (but greyed out), when I arrow to it, it already scrolls the calendar to make the next month active. This had the effect of me thinking I might've pressed the down arrow key twice on accident.

My thought is that if a date is visible in the current view, even if it is from the next month, we shouldn't change the active month. This should only happen when the arrow key is moving the active date out of view.

amandadupell:
updated behavior:

https://user-images.githubusercontent.com/38971752/133688375-a2e11aa4-13c3-4ead-b15d-a36153f0902e.mov

src/js/components/Calendar/Calendar.js
```diff
@@ -902,14 +902,30 @@ const Calendar = forwardRef(
               event.preventDefault();
               event.stopPropagation(); // so the page doesn't scroll
               setActive(addDays(active, -7));
+              if (!betweenDates(addDays(active, -7), displayBounds)) {
```
taysea, line 905: nice!

taysea:
For some reason in some cases it's not going up to the next month in this case

https://user-images.githubusercontent.com/12522275/133689228-1e0f3efc-58cc-402d-8b75-6df1c18f7257.mov

taysea (approved):
LGTM! Good job.

ericsoderberghp (approved):

PR #5127, opened 2021-04-01
[TS] Add TypeScript unit testing support

alexmirrington:
#### What does this PR do?

This is a proof-of-concept PR related to #5111 that adds _very strict_ TypeScript unit testing support, without forcing the rest of the `grommet` code base to use strict type checks. This means that we can gradually convert existing unit tests from `*-test.js` to `*-test.tsx` files, and fix any issues that TypeScript raises in the process. This should allow us to find and fix existing TypeScript issues quicker, as well as decrease opportunities for TS-related bugs in the future.

#### Where should the reviewer start?

The most important change is in `jest.config.js`, where I specify we want to use the new `ts-jest` dependency for all `tsx?` tests, but use existing pipelines for all `.jsx?` tests.

Secondly, there have been some changes to the `tsconfig.json` file, to support the creation of a more strict `tsconfig.tests.json` file that only applies to files in `**/__tests__` directories. The removed fields are to allow the `string: true` field in the `tsconfig.tests.json` file, and should all be consistent `tsconfig` [defaults](https://www.typescriptlang.org/tsconfig). The reviewer should check that all removed fields are consistent with `tsconfig` [defaults](https://www.typescriptlang.org/tsconfig). 

#### What testing has been done on this PR?

All existing unit tests still pass despite the newly added TypeScript support, both in my local `pre-commit` hook and on CI. See manual testing below for more details.

#### How should this be manually tested?

The reviewer should test that converting a `.js` test file for a few components to `.tsx` should cause tests to fail, since a bunch of new strict checks have been put in place. Failures may range from implicit `any` types through to grommet bugs caused my missing/incorrect prop types. In my testing, I used the `Accordion` and `DataTable` components, however I'd suggest the reviewer tries a few other components too. You can also change some correct prop interfaces to check that errors are correctly identified both in your code editor and when running `yarn test`.

#### Any background context you want to provide?

All context is described here: #5111

#### Do the grommet docs need to be updated? Should this PR be mentioned in the release notes?

No, it is internal to Grommet.

#### Is this change backwards compatible or is it a breaking change?

Backwards compatible

Issue #5111, opened 2021-03-28
[TS] Add TypeScript unit testing pipeline


alexmirrington:
## Motivation

In the last few weeks, I've noticed a lot of issues outlining discrepancies between actual component bahaviour (outlined in `.js` files) and the types defined in `index.d.ts` files. I propose either migrating existing tests to `.ts` files, or adding TypeScript-specific tests alongside existing tests to help improve the Grommet TypeScript experience for users.

I'd like this issue to be a thread for initial discussion/proposals, and if/when we decide on a way forward I'll log a separate issue to track migration/implementation progress.

## Related Issues

Here is a list ot the types of TypeScript issues that are being raised recently:

- Missing/incorrect typings:
  - #5108
  - #5107 
  - #5117 
- Generics for prop types
  - #5087 
  - #5041
- Storybook TS
  - #5121

oskari:
There really is much to desire for the TypeScript experience. Even new features have issues when using with TS. Grommet really should have first-class TypeScript support.

ShimiSun:
@alexmirrington THANK YOU! As we are not field experts on Typescript, we are always welcome to feedback and suggestions from the community of TS specialists. 
Grommet v2.17.0 that was released recently, had massive TS improvements thanks to the community pushing for it, and we need the community's support to help us improve in the future as well.
Feel free to file any POC PRs that demonstrate how we can improve the Typescript dev experience, so the discussion could be more proactive and we can improve the overall TS experience within Grommet.

alexmirrington:
@ShimiSun What are your thoughts on an achievable direction for improving the grommet-ts experience? There are a few directions to head, some of which are more difficult/time consuming than others:

1. First class TypeScript support, converting all existing `*.js` files to `*.ts`
  - This is by far the best option in terms of pure TS experience, but time consuming to implement and will mak eit difficult for people less experienced with TypeScript to contribute
2. Convert unit tests to TypeScript, and run them with `ts-jest` so that when type checks fail, so do unit tests.  
  - I personally think this is the most achievable option to start out, since many of the current issues we are seeing with TS are things that could have very easily been picked up by failing tests. I'm happy to start work on drafting a sample PR if you think this is a good direction to head.
3. Add TypeScript stories in Storybook to make it more transparent to grommet users/contributors how we see TypeScript being used with Grommet.
  - At the moment, everyone seems to have different ideas on the best way to use types with Grommet since there's no real documentation surrounding it.

Keen to hear your thoughts on this!

[...]

PR #5589, opened 2021-09-12
[TS] Add TypeScript unit testing support

alexmirrington:
#### What does this PR do?

This is a proof-of-concept PR related to #5111 that adds **strict** TypeScript unit testing support, without forcing the rest of the `grommet` code base to use strict type checks. This means that we can gradually convert existing unit tests from `*-test.js` to `*-test.tsx` files, and fix any issues that TypeScript raises in the process. This should allow us to find and fix existing TypeScript issues quicker, as well as decrease opportunities for TS-related bugs in the future.

The most important change is in `package.json`, where I specify we want to use the new `ts-jest` dependency for all `tsx?` tests, but use existing pipelines (`babel-jest`) for all `.jsx?` tests.

Secondly, there have been some changes to the `tsconfig.json` file, to support the creation of a more strict `tsconfig.tests.json` file that only applies to files in `**/__tests__` directories. The removed fields are to allow the `strict: true` field in the `tsconfig.tests.json` file, and are all consistent with `tsconfig` [defaults](https://www.typescriptlang.org/tsconfig), meaning type strictness has not been affected for the main `grommet` codebase, only tests. The reviewer should double check that all removed fields are consistent with `tsconfig` [defaults](https://www.typescriptlang.org/tsconfig).

All existing unit tests still pass despite the newly added TypeScript support, both in my local `pre-commit` hook and on CI. See manual testing below for more details.

The reviewer should test that converting a `.js` test file for a few components to `.tsx` should cause tests to fail, since a bunch of new strict checks have been put in place. Failures may range from implicit `any` types through to grommet bugs caused my missing/incorrect prop types. In my testing, I used the `Accordion` and `DataTable` components, however I'd suggest the reviewer tries a few other components too. You can also change some correct prop interfaces to check that errors are correctly identified both in your code editor and when running `yarn test`. In some cases such as the `Anchor` component, there are not type errors at all, and conversion to typescript is as simple as renaming the test file from `.js` to `.tsx` and updating the test spec.

All context is described here: #5111
This PR is an updated version of #5127, as we have had a few major upgrades to our dependencies since then.

#### Do the grommet docs need to be updated? Should this PR be mentioned in the release notes?

No, it is internal to Grommet.

Backwards compatible

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -17,9 +17,9 @@
       "grommet/utils": ["src/js/utils"]
     },
     "lib": ["dom", "es2017"],
-    "types": [],
+    "types": ["jest"],
     "typeRoots": ["./node_modules/@types/", "./src/@types/"]
   },
-  "exclude": ["node_modules"],
+  "exclude": ["node_modules", "**/__tests__"],
   "include": ["src/**/*"]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,13 +6,10 @@
     "moduleResolution": "node",
     "outDir": "./dist/",
     "sourceMap": true,
-    "noImplicitAny": false,
     "module": "esnext",
     "target": "es6",
     "jsx": "react",
-    "strictPropertyInitialization": false,
     "declaration": true,
-    "strict": false,
     "baseUrl": "./",
     "paths": {
       "grommet": ["src/js"],
--- a/tsconfig.tests.json
+++ b/tsconfig.tests.json
@@ -1,0 +1,13 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    /* Strict Type-Checking Options */
+    "strict": true /* Enable all strict type-checking options: noImplicitAny, strictNullChecks, strictFunctionTypes, strictPropertyInitialization, noImplicitThis, alwaysStrict */,
+    "noUnusedLocals": true /* Report errors on unused locals. */,
+    "noUnusedParameters": true /* Report errors on unused parameters. Requires TypeScript version 2.0 or later.*/,
+    "noImplicitReturns": true /* Report error when not all code paths in function return a value. */,
+    "noFallthroughCasesInSwitch": true /* Report errors for fallthrough cases in switch statement. */
+  },
+  "exclude": ["node_modules"],
+  "include": ["**/__tests__"]
+}
--- a/tsconfig.tests.json
+++ b/tsconfig.tests.json
@@ -1,12 +1,11 @@
 {
   "extends": "./tsconfig.json",
   "compilerOptions": {
-    /* Strict Type-Checking Options */
-    "strict": true /* Enable all strict type-checking options: noImplicitAny, strictNullChecks, strictFunctionTypes, strictPropertyInitialization, noImplicitThis, alwaysStrict */,
-    "noUnusedLocals": true /* Report errors on unused locals. */,
-    "noUnusedParameters": true /* Report errors on unused parameters. Requires TypeScript version 2.0 or later.*/,
-    "noImplicitReturns": true /* Report error when not all code paths in function return a value. */,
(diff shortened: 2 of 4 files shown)


tsconfig.json
```diff
@@ -6,23 +6,20 @@
     "moduleResolution": "node",
     "outDir": "./dist/",
     "sourceMap": true,
-    "noImplicitAny": false,
     "module": "esnext",
     "target": "es6",
     "jsx": "react",
-    "strictPropertyInitialization": false,
     "declaration": true,
-    "strict": false,
```
jcfilben, line 15: I'm curious why we are removing these settings

tsconfig.tests.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noImplicitReturns": true,
+    "noFallthroughCasesInSwitch": true
```
jcfilben, line 8: Why these specific options, are they just rules you think should be enforced or is it a recommended pattern?

jcfilben (commented):
Nice job! I converted some of the test files to .tsx files and looks like the js tests and tsx tests are running as expected. I just had a question about the tsconfig files

tsconfig.json
alexmirrington: There are two reasons:
- These settings are the defaults for `.tsconfig` files, so removing them has not impact on the strictness of the Typescript checks used for the main part of the repo
- When extending `.tsconfig.yaml` in `.tsconfig.tests.yaml`, it's important to note that only ["files, include and exclude from the inheriting config file overwrite those from the base config file"](https://www.typescriptlang.org/tsconfig#extends). What this means for us is that including `strict`, `noUnusedLocals`, `noUnusedParameters` etc. in the base `tsconfig` stops us from overriding these values in the extended `tsconfig`, contrary to the intuition we usually have from OOP.

tsconfig.tests.json
alexmirrington: We could definitely add more to be more strict, but these were some that I have found to be useful in some of my own TypeScript projects. If I recall correctly they were adapted from a default strict tsconfig. Here's what they do:

```
    /* Strict Type-Checking Options */
    "strict": true /* Enable all strict type-checking options: noImplicitAny, strictNullChecks, strictFunctionTypes, strictPropertyInitialization, noImplicitThis, alwaysStrict */,
    "noUnusedLocals": true /* Report errors on unused locals. */,
    "noUnusedParameters": true /* Report errors on unused parameters. Requires TypeScript version 2.0 or later.*/,
    "noImplicitReturns": true /* Report error when not all code paths in function return a value. */,
    "noFallthroughCasesInSwitch": true /* Report errors for fallthrough cases in switch statement. */,
```

tsconfig.json
jcfilben: Thank you for the detailed explanation!

tsconfig.tests.json
jcfilben: Great, thanks for providing detail for each rule

jcfilben (approved):
Looks good to me

alexmirrington:
FYI: The force push was to resolve `yarn.lock` locally and rebase onto the current master.

ericsoderberghp (approved):
I'm good with this as a building block to start with. I like how the Anchor test needs no changes. It will be interesting to see how large the changes are elsewhere as we go.

ericsoderberghp:
BTW, thanks for an excellent Description for this pull request.

alexmirrington:
@ericsoderberghp You're welcome! I'll create a separate issue to start tracking test migration progress. I'll save most of the discussion for that ticket, however the way I would recommend we start is to try converting tests to `.tsx` at first, and if there are a small amount of TS errors they can be fixed in the same PR as the test migration, otherwise separate issues can be created prior to migrating the test files to avoid massive PRs.

PR #5590, opened 2021-09-13
Fix: Introducing Select onBlur and onFocus

### LINKED ISSUE #5511: Impossible to detect onBlur for closed Select
### Expected Behavior

We are using onBlur event to detect when users have touched a form input.
Thus, we can start field validation as soon as user touched the input.

It works well with TextInput, etc...
However for "Select", "onBlur" is not given to the top level component (StyledSelectDropButton) but on the "SelectTextInput".

Thus, when we use tab to go through fields of a form, we wan not detect that user touched the select.

### Actual Behavior

No onBlur handler called when tabbing through "Select" components.

https://codesandbox.io/s/grommet-v2-template-forked-m02r7?file=/index.js

### Steps to Reproduce

Just click on "Click here than tab", tab, and go to console.
=> Only "TextInput onBlur" is displayed.

### Your Environment

- Grommet version: 2.17.4
- Browser Name and version: Firefox 90.0.2
- Operating System and version (desktop or mobile): Dekstop Windows 10

jlaluces123:
#### What does this PR do?
This PR aims to:
- Resolve #5511
- Introduces `onBlur` and `onFocus` handlers

`Select.js`

1. Took codesandbox linked in #5511 and made a new story with it to test inside grommet repo

Following the instructions of what was posted on #5511:
1. Copy the codesandbox linked in #5511 thread to a new storybook inside grommet repo
2. Run `yarn storybook` and navigate to the newly made story
3. Click on the `Click here then tab` text
4. Tab to Select and watch the `Select onBlur` message get logged to the console

#5511

No

No

Backwards compatible


jcfilben (approved):

src/js/components/Select/Select.js
```diff
@@ -128,6 +130,8 @@ const Select = forwardRef(
     }, [value, valueKey]);
     // search input value
     const [search, setSearch] = useState();
+    // focus state
+    const [focus, setFocus] = useState();
```
ericsoderberghp, line 134: Why do we need to add this? I don't think DropButton or Button take a `focus` property.

jlaluces123: Ahh I think this was just part of when I was "brute-forcing" my way by following how other components implemented their own focus states. After reviewing the code this is most definitely not necessary and will be removed in the upcoming update.

src/js/components/Select/Select.js
```diff
@@ -295,6 +297,12 @@ const Select = forwardRef(
           open={open}
           alignSelf={alignSelf}
           focusIndicator={focusIndicator}
+          onFocus={(event) => {
```
ericsoderberghp, line 300: I think this can just be `onFocus={onFocus}`, similar for `onBlur`

ericsoderberghp (approved):

PR #5568, opened 2021-09-01
Fix missing effect dependency in Notification

taysea:
#### What does this PR do?

Lint was calling out missing dependency in a useEffect in notifications. This adds it in.

#### Where should the reviewer start?
src/js/components/Notification/Notification.js

#### What testing has been done on this PR?
Tested toast notification in storybook.

#### What are the relevant issues?
Related to https://github.com/grommet/grommet/issues/5198

#### Do the grommet docs need to be updated?
No.

#### Should this PR be mentioned in the release notes?
No.

#### Is this change backwards compatible or is it a breaking change?
Backwards compatible.

PR #5593, opened 2021-09-13
Fixes Select dropdown rendering issue

### LINKED ISSUE #5586: Select Dropdown rendered outside of Drop
### Expected Behavior
A `Select` options dropdown should be confined to the `Drop` it is rendered inside.

### Actual Behavior

A `Select` options dropdown is rendered 'on top' of the Drop. This results in a floating dropdown while the Select itself is no longer visible inside the Drop.

https://codesandbox.io/s/interesting-maxwell-sh15l?file=/src/App.js

### Steps to Reproduce

1. Open Sandbox
2. Open the Select dropdown
3. Scroll down inside of the Drop until the Select is hidden
### Your Environment

- Grommet version: 2.17.5

PauloE314:
#### What does this PR do?

It fixes a very subtle bug on the Select component dropdown. Sometimes, the Select dropdown content was rendered on the top left corner of the screen during a fraction of a second (it was basically a blink) before it be rendered in the correct place.

Drop > DropContainer.js

The tests where done mainly in a local storybook. I've created a custom storybook page with some basic code just to make debugging easier. 

Run a local storybook, or a custom React project, and run the code specified in #5586. 

Closes #5586.

No.

Backwards compatible.

### DIFF
--- a/src/js/components/Drop/DropContainer.js
+++ b/src/js/components/Drop/DropContainer.js
@@ -5,6 +5,8 @@ import React, {
   useMemo,
   useRef,
 } from 'react';
+import useLayoutEffect from 'use-isomorphic-layout-effect';
+
 import { ThemeContext } from 'styled-components';
 
 import { FocusedContainer } from '../FocusedContainer';
@@ -63,7 +65,7 @@ const DropContainer = forwardRef(
       [portalContext, portalId],
     );
     const dropRef = useRef();
-    useEffect(() => {
+    useLayoutEffect(() => {
       const notifyAlign = () => {
         const styleCurrent = (ref || dropRef).current.style;
         const alignControl = styleCurrent.top !== '' ? 'top' : 'bottom';
--- a/package.json
+++ b/package.json
@@ -51,7 +51,8 @@
     "grommet-icons": "^4.6.2",
     "hoist-non-react-statics": "^3.2.0",
     "markdown-to-jsx": "^7.1.3",
-    "prop-types": "^15.7.2"
+    "prop-types": "^15.7.2",
+    "use-isomorphic-layout-effect": "^1.1.1"
   },
   "devDependencies": {
     "@babel/cli": "^7.14.8",
--- a/yarn.lock
+++ b/yarn.lock
@@ -12778,7 +12778,7 @@ use-composed-ref@^1.0.0:
   dependencies:
     ts-essentials "^2.0.3"
 
-use-isomorphic-layout-effect@^1.0.0:
+use-isomorphic-layout-effect@^1.0.0, use-isomorphic-layout-effect@^1.1.1:
   version "1.1.1"
   resolved "https://registry.yarnpkg.com/use-isomorphic-layout-effect/-/use-isomorphic-layout-effect-1.1.1.tgz#7bb6589170cd2987a152042f9084f9effb75c225"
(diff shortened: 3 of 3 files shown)


halocline:
Thank you, @PauloE314.  Would [using this implementation](https://github.com/grommet/grommet/blob/7eaf82c2acac88c7df08dfa99eae00db56ca8542/src/js/utils/use-isomorphic-layout-effect.js) achieve the same fix?

PauloE314:
Oh, sure, @halocline.

halocline:
@kysley Are you able to confirm this implementation will resolve https://github.com/grommet/grommet/issues/5586?

PauloE314:
I've tested a lot and the bug stopped occurring. My conclusion was that the `place` method wasn't being called fast enough, but, as `useLayoutEffect` is called before the browser be able to render anything, that would be a nice solution.

kysley:
> 
> 
> @kysley Are you able to confirm this implementation will resolve #5586?

hey @halocline, perhaps the sandbox I provided wasn't quite clear- and my "fix" was misconstrued as the problem.

What I was trying to highlight is the Dropdown of the Select is "on top" of the `Drop` it is rendered inside. It seems to be a layering issue, since both of the components are rendered with a Portal. I mocked up something ugly in paint to try to illustrate ;) The green border being the `Drop` and the red being the dropdown content from the `Select`. You can see that the Dropdown content spills outside of the Drop's boundary, rather than being on the same z-index as the rest of the `Drop`'s content.

image

halocline:
> You can see that the Dropdown content spills outside of the Drop's boundary, rather than being on the same z-index as the rest of the `Drop`'s content.

That's what I wanted to confirm! Thank you for replying @kysley.

PauloE314:
Sorry for this, @kysley, when I opened the sandbox and clicked the select, the first thing a saw was a fast bug with the select options (they were appearing on the top of the screen during a really short amount of time), so I through the issue was related to it. Maybe I've found a new bug, so I think it's good to keep the little changes made, please leave your opinions about this.

Now that I understood the issue, I can work on it correctly.

kysley:
No problem @PauloE314! I think the bug you found here is valid since I had to write a quick workaround in the sandbox. The fix seems appropriate as well, accessing a ref's computed style _after_ its been committed to the DOM.

amandadupell (commented):
from the history of this PR, it looks like this PR is addressing a different bug than the one outlined in #5568 so i think it might be nice to update exactly what issue this is solving to avoid confusing it with the other bug : )

britt6612:
> from the history of this PR, it looks like this PR is addressing a different bug than the one outlined in #5568 so i think it might be nice to update exactly what issue this is solving to avoid confusing it with the other bug : )

Its issue #5586...

PauloE314:
Yeah, @amandadupell. Even after a week testing ways to fix the real bug (from the issue #5586), unfortunately, I still couldn't find a solution for that, sorry, guys.

src/js/components/Drop/DropContainer.js
```diff
@@ -63,7 +64,7 @@ const DropContainer = forwardRef(
       [portalContext, portalId],
     );
     const dropRef = useRef();
-    useEffect(() => {
+    useLayoutEffect(() => {
```
ericsoderberghp, line 67: I like `useLayoutEffect()` for re-calculating `place()`. But, it seems expensive to be adding and removing event listeners on each layout. Instead, I'm wondering if we should separate out `place()` via `useCallback()` and call it from the listeners as well as the layout effect instead.

PR #5597, opened 2021-09-15
change viewbox height

### LINKED ISSUE #5596: Meter type=circle size=full has svg error
When specifying `type="circle" size="full"` to Meter

### Expected Behavior

Shouldn't produce a browser error in the console

### Actual Behavior

This error is displayed in the chrome javascript  console.

image

### Steps to Reproduce

```jsx
<Meter
   type="circle"
   background="light-2"
   size="full"
   values={[{ value: 50, color: 'accent-1' }]}
/>
```

This is because Circle.js uses 100% for `viewBoxHeight` and `viewBox` doesn't seem to allow the % spec.
```jsx
let viewBoxHeight;
  if (type === 'semicircle') {
    viewBoxHeight = width / 2;
  } else if (size === 'full') {
    viewBoxHeight = '100%';
  } else viewBoxHeight = width;

  return (
    <StyledMeter
      ref={ref}
      viewBox={`0 0 ${width} ${viewBoxHeight}`}
      width={size === 'full' ? '100%' : width}
      height={viewBoxHeight}
```
### Your Environment

- Grommet version: stable 9/14/2021 (post 2.17.5 but before 2.17.6)
- Browser Name and version: chrome latest
- Operating System and version (desktop or mobile): windows

britt6612:
#### What does this PR do?
This PR changes the way that the height is set
circle.js
storybook
```
<Meter
   type="circle"
   background="light-2"
   size="full"
   values={[{ value: 50, color: 'accent-1' }]}
/>
```
closes #5596
no
sure
 backwards compatible


src/js/components/Meter/Circle.js
```diff
@@ -147,12 +147,17 @@ const Circle = forwardRef((props, ref) => {
     viewBoxHeight = '100%';
   } else viewBoxHeight = width;
 
+  let viewBoxSize;
```
ericsoderberghp, line 150: It doesn't look like we're ever setting `viewBoxSize`?

britt6612: yes sorry, I committed the wrong change. 

I think the problem here was before the `viewBox={0 0 ${width} ${width}}`
This was changed so that when `size="full"` it was set to '100%' this should only be applied to the height

britt6612: also saw that when `size="full"` for `semimeter` was not working correctly so I added that check

src/js/components/Meter/Circle.js
```diff
@@ -145,12 +145,14 @@ const Circle = forwardRef((props, ref) => {
     viewBoxHeight = width / 2;
   } else if (size === 'full') {
     viewBoxHeight = '100%';
+  } else if (size === 'full' && type === 'semicircle') {
+    viewBoxHeight = '100%';
```
MikeKingdom, line 149: You'll never hit this because of 144-146.
It's also weird because this is called viewBoxHeight but it's not used for the viewBox but is used for height.
And for semicircle don't you want the view box to have a height of width/2 even for `full`?

How about something like:
```jsx
const viewBoxHeight = (type === 'semicircle') ? width / 2 : width;

return (
<StyledMeter
  ref={ref}
  viewBox={`0 0 ${width} ${viewBoxHeight}`}
  width={size === 'full' ? '100%' : width}
  height={size === 'full' ? '100%' : viewBoxHeight}
  {...rest}
>
  {track}
  {paths}
  {pathCaps}
</StyledMeter>
```

britt6612: we would also want to check for `full` correct ? to then use '100%' 
otherwise is they pass `size="full"` then the meter will look smaller than `medium`

britt6612: I agree! I wasn't entirely sure if the `semicircle` even when full should be `width/2`

src/js/components/Meter/Circle.js
```diff
@@ -140,19 +140,14 @@ const Circle = forwardRef((props, ref) => {
...
-  } else if (size === 'full') {
-    viewBoxHeight = '100%';
-  } else viewBoxHeight = width;
+  const viewBoxHeight = type === 'semicircle' ? width / 2 : width;
 
   return (
     <StyledMeter
       ref={ref}
-      viewBox={`0 0 ${width} ${viewBoxHeight}`}
+      viewBox={`0 0 ${width} ${width}`}
```
MikeKingdom, line 148: this still looks wrong to me

MikeKingdom (commented):
Thanks Brittany. That seems to make alot more sense to me. It's hard to know without being able to test this easily.

Given how the problem happened, it seems like we could use at least a couple more Jest test cases.

britt6612:
> Thanks Brittany. That seems to make alot more sense to me. It's hard to know without being able to test this easily.
> 
> Given how the problem happened, it seems like we could use at least a couple more Jest test cases.

Yes I was looking through the `Meter` tests and saw that there are snapshots for `size` however `full` was not included in this so I think this should be added.

MikeKingdom (approved):
Awesome. Thanks!

ericsoderberghp (approved):

PR #5603, opened 2021-09-20
Fix: reversed animation on loop for autoplay carousel

### LINKED ISSUE #5585: Carousel animation reverses when looping from last slide back to first slide
### Expected Behavior

When a carousel is set to `autoplay` and it returns from the last slide to the first slide, the animation should play in the same direction as the previous animation. For example, if the other slides slide into view from `right --> left`, the first slide should also slide in from `right --> left`.

It's a little visually jarring for the direction to suddenly change.

### Actual Behavior

The animation for the first slide is reversed (E.G. it slides in from `left --> right` instead of `right --> left`) compared to the others.

See it here: https://codesandbox.io/s/grommet-v2-template-forked-8z72t?file=/index.js

### Steps to Reproduce

1. Setup an autoplay carousel with at least 2 slides
2. Observe the animations that play

### Your Environment

- Grommet version: 2.17.5
- Browser Name and version: Safari Version 14.1.2 (16611.3.10.1.6), Chrome 92.0.4515.159 (Official Build) (x86_64)
- Operating System and version (desktop or mobile): macOS Big Sur 11.5.2

jlaluces123:
#### What does this PR do?
This PR aims to:
- Resolve #5585
- When carousel is set to `play` of any duration, when the last slide plays (thus looping back to the beginning of the carousel), the animation is consistent (`slideLeft`).

`Carousel.js` line 151 to see newest change in animation parsing logic

Comparison of animation functionality for all Carousel components. The direct issue can be seen being solved in Carousel's storybook titled `Without Controls`

In the `Without Controls` story for Carousel, you can verify the animation is no longer doing a `slideRight` once the loop has started over again.

Closes #5585
No
No
Backwards compatible


jcfilben (commented):
This looks good!

However, potentially something to discuss/think about is if we should give the user the option to specify the animation through a prop on Carousel.

src/js/components/Carousel/Carousel.js
```diff
@@ -147,7 +147,8 @@ const Carousel = ({
     if (index === activeIndex) {
       if (priorActiveIndex !== undefined) {
         animation = {
-          type: priorActiveIndex < activeIndex ? 'slideLeft' : 'slideRight',
+          type:
+            play || priorActiveIndex < activeIndex ? 'slideLeft' : 'slideRight',
```
ericsoderberghp, line 151: Just wondering why we would only tie this to `play`? I would think that `play` just means automatically advance. So, I would think we could just make this "slideRight".

ericsoderberghp:
> This looks good!
> 
> However, potentially something to discuss/think about is if we should give the user the option to specify the animation through a prop on Carousel. For example, I could see users potentially wanting to use 'fadeIn' instead of the slide animations

I like it as a future possibility, but I think we can wait for a specific request down the road to help inform what we should do in this regard.

src/js/components/Carousel/Carousel.js
jcfilben: I think when `play={false}` and you are manually clicking through the buttons on the left or right to change the image it makes sense if you click the left button to have the animation 'slideLeft' instead of always being 'slideRight'

ericsoderberghp: Ah, makes sense. Thanks.

ericsoderberghp (approved):

PR #5606, opened 2021-09-20
update text weight proptypes

britt6612:
#### What does this PR do?
This PR adds `bolder` to the `weight` for Text proptypes
proptypes
storybook
storybook should be no warning
design-system
#### Screenshots (if appropri
Screen Shot 2021-09-20 at 4 14 26 PM
ate)

backwards compatible


britt6612:
We are using `bolder` for the highlighted words in the search for the design-system

Screen Shot 2021-09-20 at 5 42 58 PM

ericsoderberghp:
Can that just be "bold"? I'm trying to stick to the typography guidance using the HPE Design System. Just wondering.

halocline:
> Can that just be "bold"? I'm trying to stick to the typography guidance using the HPE Design System. Just wondering.

I feel that we are sticking to the typography guidance "[For digital experiences we use Light, Regular, Medium, and Bold weights at a variety of scales to compliment the content in a given design](http://localhost:3000/foundation/typography#metrichpe-styles)." I feel that 'bolder' is saying is step up one font weight, but still adheres to the available font-weights which are dictated by the theme. In this particular context, the component is inheriting whatever weight is provided, then emphasizing a particular portion of text by stepping up to the next weight.

Soft opinion.

From a Grommet perspective, 'lighter' and 'bolder' are common CSS font-weight values.

Soft opinion.

src/js/components/Text/propTypes.js
```diff
@@ -81,7 +81,7 @@ if (process.env.NODE_ENV !== 'production') {
     ]),
     truncate: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf(['tip'])]),
     weight: PropTypes.oneOfType([
-      PropTypes.oneOf(['normal', 'bold']),
+      PropTypes.oneOf(['normal', 'bold', 'bolder']),
```
halocline, line 84: Recommend we support the [common keyword font-weight values](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#syntax). We have 'normal', 'bold', and numerics, so just need to add 'lighter' and 'bolder'.

```suggestion
      PropTypes.oneOf(['normal', 'bold', 'bolder', 'lighter']),
```

halocline (commented):
`index.d.ts` definition needed as well

Grommet-site docs needed as well

taysea:
I am in support of adding this in Grommet for general users, but I am wondering if there are any other cases in designs for Design System where we use `bolder`? If this is a one-off use case for `bolder`, then I might be inclined to say we should use `bold` for Design System. Also a soft opinion.

src/js/components/Text/propTypes.js
```diff
@@ -81,7 +81,7 @@ if (process.env.NODE_ENV !== 'production') {
     ]),
     truncate: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf(['tip'])]),
     weight: PropTypes.oneOfType([
-      PropTypes.oneOf(['normal', 'bold']),
+      PropTypes.oneOf(['normal', 'bold', 'bolder', 'lighter']),
```
ericsoderberghp, line 84: What about index.d.ts?

britt6612: I'll add

britt6612:
here is repo for grommet-site 
https://github.com/grommet/grommet-site/pull/314

ericsoderberghp (approved):

halocline (approved):

PR #5614, opened 2021-09-22
Changed Markdown to fix an issue with property placement

ericsoderberghp:
#### What does this PR do?

Fixes an issue where Markdown couldn't support `onClick`. This was due to `...rest` being placed on the Markdown component instead of its wrapping div. Since the Markdown wrapper is a Fragment, it was ignoring most of `...rest`. 

Markdown.js

Integrated with grommet-designer

no

yes

backwards compatible


amandadupell (commented):
looks good to me! my only thought is if there needs to be a new example of this behavior for Markdown in Storybook to show how we support the onClick now, just a thought!

ericsoderberghp:
> looks good to me! my only thought is if there needs to be a new example of this behavior for Markdown in Storybook to show how we support the onClick now, just a thought!

I thought about that. But, this is a pretty obscure usage. It doesn't really make sense for most contexts to have the Markdown component be interactive, just something unique to grommet-designer.

jcfilben (approved):
Looks good to me!

amandadupell (approved):

MikeKingdom (approved):
LGTM!

PR #5621, opened 2021-09-23
Change radiobutton theme background color to accept symbolic colors

### LINKED ISSUE #5618: For certain attributes, we cannot use theme colors.
### Expected Behavior

When building a custom theme, I should be able to use the theme defined colors for other components

### Actual Behavior

With certain component styling attributes, I cannot use those predefined colors. 

For example, if I were to define a color `'light-2': '#F3F3F3'` and then use it in the `radioButton.check.background.color` it will not work.

This:
Screen Shot 2021-09-22 at 2 22 23 PM

Produces:
Screen Shot 2021-09-22 at 2 43 44 PM

Where the `radioButton.check.background.color` is `'#FFFFFF'`  and not the expected `'#F3F3F3'`

MikeKingdom:
#### What does this PR do?

Change RadioButton to recognize symbolic colors in
theme.radioButton.check.background.color.

StyledRadioButton.js

Storybook with a modified RadioButton/Theme example. The previous version of that theme used the color 'red' which is a valid CSS color so it didn't expose the issue.

Storybook RadioButton/Theme

Fixes #5618 

No

Sure

Backwards compatible

### DIFF
--- a/src/js/components/RadioButton/StyledRadioButton.js
+++ b/src/js/components/RadioButton/StyledRadioButton.js
@@ -102,8 +102,10 @@ Object.setPrototypeOf(StyledRadioButtonIcon.defaultProps, defaultProps);
 
 const StyledRadioButtonBox = styled.div`
   background-color: ${(props) =>
-    props.theme.radioButton.check.background &&
-    props.theme.radioButton.check.background.color};
+    normalizeColor(
+      props.theme.radioButton.check.background?.color,
+      props.theme,
+    )};
   ${(props) => props.focus && focusStyle()};
   ${(props) => props.theme.radioButton.check.extend};
 `;
--- a/src/js/components/RadioButton/stories/Theme.js
+++ b/src/js/components/RadioButton/stories/Theme.js
@@ -18,7 +18,7 @@ const theme = deepMerge(grommet, {
         light: 'neutral-1',
       },
       background: {
-        color: 'red',
+        color: 'brand',
       },
     },
     icon: {


britt6612 (approved):

britt6612:
also noticed we had this as a test 

```
  test('background-color themed', () => {
    const customTheme = {
      radioButton: {
        check: {
          background: {
            color: 'red',
          },
        },
      },
    };

    const { container } = render(
      <Grommet theme={customTheme}>
        <RadioButton name="test" />
      </Grommet>,
    );

  });


```

do you think we should change that value or leave as is? Im good either way since the `brand` is added to the storybook but would be nice in the test.

MikeKingdom:
> also noticed we had this as a test
> 
> ```
>   test('background-color themed', () => {
>     const customTheme = {
>       radioButton: {
>         check: {
>           background: {
>             color: 'red',
>           },
>         },
>       },
>     };
> 
>     const { container } = render(
>       <Grommet theme={customTheme}>
>         <RadioButton name="test" />
>       </Grommet>,
>     );
> 
>     expect(container.firstChild).toMatchSnapshot();
>   });
> ```
> 
> do you think we should change that value or leave as is? Im good either way since the `brand` is added to the storybook but would be nice in the test.

Thanks for taking a look @britt6612 . I debated a little on this and figured I'd keep that test case as an example of setting the background to something typical from a CSS color spec. Not sure that's a big deal but seems useful. We could add a separate test case for a symbolic color to cover both cases if you think that is important to have in addition to what chromatic will see from Storybook. Thoughts?

britt6612:
I was also debating I think an additional test would be the best since one of us may over look the story if something similar happens.

britt6612 (approved):
Thanks for adding the test 😊

jcfilben (approved):

taysea (approved):
LGTM.

ericsoderberghp (approved):

Issue #5604, opened 2021-09-20
PropertyList - draft grommet component


ericsoderberghp:
Create a draft pull request including the index.d.ts file to enable discussing the API surface.

britt6612:
These are now ready for review!

PR #5623, opened 2021-09-23
NEW NameValueList Component Box approach

britt6612:
#### What does this PR do?
Provides a first pass proposal and prototype examples for <NameValueList />.

`NameValueList` is essentially the parent `box` for `NameValuePair` 
`NameValuePair` was created in a separate file to keep code clean.

`NameValuePair` contains `Boxs` that are determining the direction for the `name text` and `value text`

To keep things simple on the users part I added a `layout` prop that can be leveraged to make up the different variants of `NameValueList`

## First Approach

After exploring this approach I found that it may be useful to have a `grid` layout. This would keep things more consistent. I wanted to get this out there though to start the conversation. 

**NameValueList API** 

```

export interface NameValueListProps {
  a11yTitle?: string,
  data?: array,
  layout?: string,
  pairProps?: boxProps,
  nameProps?: textProps,
  listProps?: boxProps,
  valueProps?: textProps,
  align?: object 

```

TODOs:

- Gather feedback
- Documentation
- Unit tests
- Stories
- mobile view (responsiveness)
- accessibility 
- grid layout (@taysea ) is going to be pushing this porition up here soon! 

NameValueList.js
storybook
storybook
[Figma file](https://www.figma.com/file/Ny7rYK5gI5VO4PtZsq6vGC/HPE-Name-Value-Pair?node-id=148%3A1936)
issue #5604 
Closes https://github.com/grommet/hpe-design-system/issues/1904

eventually
yes minor release
backwards compatibl

### DIFF
--- a/src/js/components/NameValueList/NameValueList.js
+++ b/src/js/components/NameValueList/NameValueList.js
@@ -1,0 +1,44 @@
+import React, { forwardRef } from 'react';
+import { Box } from '../Box';
+import { NameValuePair } from './NameValuePair';
+
+const NameValueList = forwardRef(
+  (
+    {
+      a11yTitle,
+      data,
+      layout = 'vertical',
+      pairProps,
+      nameProps,
+      listProps,
+      valueProps,
+      align,
+      ...rest
+    },
+    ref,
+  ) => (
+    <Box
+      ref={ref}
+      gap={layout === 'horizontal' ? 'large' : 'medium'}
+      direction={layout === 'horizontal' ? 'row' : 'column'}
+      {...listProps}
+      {...rest}
+    >
+      {data.map((datum) => (
+        <NameValuePair
+          // direction can be vertical(default) stacked, or horizontal(inline)
+          layout={layout}
+          align={align}
+          data={datum}
+          nameProps={nameProps}
+          pairProps={pairProps}
+          valueProps={valueProps}
+        />
+      ))}
+    </Box>
+  ),
+);
+
+NameValueList.displayName = 'NameValueList';
+
+export { NameValueList };
--- a/src/js/components/NameValueList/NameValuePair.js
+++ b/src/js/components/NameValueList/NameValuePair.js
@@ -1,0 +1,64 @@
+import React from 'react';
+import { Box } from '../Box';
+import { Text } from '../Text';
+
+export const NameValuePair = ({
+  data,
+  layout,
+  pairProps,
+  nameProps,
+  valueProps,
+  align,
+}) => {
+  // based on the layout prop we would style accordingly
+  // Should we base gap on theme ?
+  // const theme = useContext(ThemeContext) || defaultProps.theme;
+  const layoutDirection = layout === 'vertical' ? 'row' : 'column';
+  return (
+    <Box
+      direction={layoutDirection}
+      gap={layout === 'vertical' ? 'large' : 'small'}
+      {...pairProps}
+    >
+      <Box>
+        <Box
+          direction="row"
+          gap="small"
+          width="small"
+          // user can overid the set props with theme
+          {...nameProps?.container}
+        >
+          {data.nameVisual}
+          <Text
+            textAlign={align?.name}
+            weight="bold"
+            color="text-strong"
+            {...nameProps?.text}
+          >
+            {data.name}
+          </Text>
+        </Box>
+      </Box>
+      <Box fill>
+        <Box
+          direction="row"
+          gap="small"
+          width={{ max: 'medium' }}
+          flex
+          justify={align?.value}
+          {...valueProps?.container}
+        >
+          {/* how should we name the "visual" that can accompany? */}
+          {data.valueVisual}
+          <Text
+            textAlign={align?.value}
+            color="text-strong"
+            {...valueProps?.text}
+          >
+            {data.value}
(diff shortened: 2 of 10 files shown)


halocline:
@britt6612 would it be possible to provide an updated definition of the prop types and their possible values? That'd be helpful to familiarize myself with the interface.

e.g.

```
export interface NameValueListProps {
  a11yTitle?: string,
  data?: array,   // would expect this to be an array of objects containing either 'key' and 'value' or 'name' and 'value'
  layout?: string, // what values are accepted here?
  pairProps?: boxProps, // adding a comment description of this prop would be helpful
  nameProps?: textProps, // adding a comment description of this prop would be helpful
  listProps?: boxProps, // adding a comment description of this prop would be helpful
  valueProps?: textProps, // adding a comment description of this prop would be helpful
  align?: object // what is this aligning?

```

I see some of this exists in the Story and will check out those examples more.

halocline:
Wondering about `layout` vs. `direction`... 

Wondering about decoupling list direction and pair direction...  would decoupling provide the caller more flexibility by being able to have one prop controlling the list's direction (row or column) and another to control the pair's direction (row or column)? This would align closer with existing Grommet Box terminology as well as allow for 4 "layout" combinations instead of the current 3 (e.g. vertical, horizontal, stacked).

britt6612:
> @britt6612 would it be possible to provide an updated definition of the prop types and their possible values? That'd be helpful to familiarize myself with the interface.
> 
> e.g.
> 
> ```
> export interface NameValueListProps {
>   a11yTitle?: string,
>   data?: array,   // would expect this to be an array of objects containing either 'key' and 'value' or 'name' and 'value'
>   layout?: string, // what values are accepted here?
>   pairProps?: boxProps, // adding a comment description of this prop would be helpful
>   nameProps?: textProps, // adding a comment description of this prop would be helpful
>   listProps?: boxProps, // adding a comment description of this prop would be helpful
>   valueProps?: textProps, // adding a comment description of this prop would be helpful
>   align?: object // what is this aligning?
> ```
> 
> I see some of this exists in the Story and will check out those examples more.

yes doing now!

britt6612:
> Wondering about `layout` vs. `direction`...
> 
> Wondering about decoupling list direction and pair direction... would decoupling provide the caller more flexibility by being able to have one prop controlling the list's direction (row or column) and another to control the pair's direction (row or column)? This would align closer with existing Grommet Box terminology as well as allow for 4 "layout" combinations instead of the current 3 (e.g. vertical, horizontal, stacked).
> 
> Or do I specify these in the `listProps.direction` and `pairProps.direction` if I want that control?

So I added the `layout` with more of the Design-System in mind so that way other teams would be able to easier pass in which layout they would like and it would take care of everything for them however with keeping the community in mind I added the `{...listProps}` and the `{...pairProps}` that way they can override the direction and have which ever layout they desire.

halocline:
RE: alignment - would be good to see the definitions for possible values.

Can I have a 'end' aligned 'name'? In the 'stacked' scenario, can I align centered? Should I be able to?

Wondering about the tradeoffs between distinct `align` vs. `pairProps.align` since `pairProps` is defined as `boxProps`?
Same question for the tradeoffs between distinct `layout` vs. `listProps.direction` since `listProps` is defined as `boxProps`?

Part of me likes the convenience of having `layout` or `direction` at the top-level API for convenience, but also like the minimal surface approach as well. In the current implementation, if I were to specify:
```
  <NameValueList data={data} layout="horizontal" listProps={{ direction: 'column' }} />
```
which would win?

britt6612:
> RE: alignment - would be good to see the definitions for possible values.
> 
> Can I have a 'end' aligned 'name'? In the 'stacked' scenario, can I align centered? Should I be able to?
> 
> Wondering about the tradeoffs between distinct `align` vs. `pairProps.align` since `pairProps` is defined as `boxProps`?
> Same question for the tradeoffs between distinct `layout` vs. `listProps.direction` since `listProps` is defined as `boxProps`?
> 
> Part of me likes the convenience of having `layout` or `direction` at the top-level API for convenience, but also like the minimal surface approach as well. In the current implementation, if I were to specify:
> 
> ```
>   <NameValueList data={data} layout="horizontal" listProps={{ direction: 'column' }} />
> ```
> 
> which would win?

Im going to add some more examples for the `align` and how that would look those are great points that you brought up..

So something like this 

```

<NameValueList data={data} layout="horizontal" listProps={{ direction: 'column' }} />

```

The `listProps={{ direction: 'column' }} ` would win in this case which allows the user to create which every layout they wanted to but also keeping the 3 layouts of 'vertical', 'stacked' and 'horizontal' easy to achieve with just passing one prop as opposed to passing `listProps={{ direction: 'column' }} ` and `pairProps={{direction: 'column'}}` to achieve a `stacked` look

britt6612:
> RE: Can I have a 'end' aligned 'name'? In the 'stacked' scenario, can I align centered? Should I be able to?

This is actually a good point I have a set `width` for the `name` so you would not be able to align I think this is something I ran into that's why we brought up the idea of having a grid but then maybe having a prop that controls the width of the name so users can control that aspect of it.

britt6612:
Going to close this PR in favor of the Grid approach. We decided that was better for taking care of the responsiveness.

Issue #5440, opened 2021-07-13
TextInput with suggestions Drop will close when TextInput is re-selected


haysclark:
There is a bug when using a TextInput with a suggestions Drop. If the user clicks on the TextInput field to show the suggestions, a 2nd click of the TextInput will close the suggestions. The 2nd mouse click triggers the Drops onClickOutside() event, which subsequently closes the suggestions.

Use case: The user selects the TextInput to start a search, but then tries to move their cursor via a mouse click.

### Expected Behavior

TextInput suggestions should stay open as long as the TextInput has focus.

### Actual Behavior

TextInput suggestions show up on initial click when receiving focus; however, re-selecting the TextField causes suggestions to close despite the TextField never loosing focus.

### Examples

The [Grommet Default Suggestion story](https://storybook.grommet.io/?path=/story/input-textinput-default-suggestion--default-suggestion) and [Grommet Custom Suggestion story](https://storybook.grommet.io/?path=/story/input-textinput-custom-suggestions--custom-suggestions) serve as a nice starting point
for demonstrating the issue.

- https://storybook.grommet.io/?path=/story/input-textinput-default-suggestion--default-suggestion
- https://storybook.grommet.io/?path=/story/input-textinput-custom-suggestions--custom-suggestions

### Steps to Reproduce

1. navigate to one of the examples above
1. click on TextInput and type "a" to show names
1. click TextInput a 2nd time
1. notice suggestions have closed

### Your Environment

- Grommet version: 2.17.2
- Browser Name and version: All browsers (using Chrome)
- Operating System and version (desktop or mobile): All systems, using OSX

novykh:
@haysclark you can check this PR https://github.com/grommet/grommet/pull/5625 until they decide to merge it.

Issue #5426, opened 2021-07-07
TextInput suggestions dropdown issue


ischoe:
Hi, i use the Textinput with suggestions from api calls. 
As my limit of results is always 5, the dropdown is not opening after i doubleclick inside to remove the value and type something new. I figured out that there is this suggestionsAtClose check, which might be the reason for this. 
```
if (focus && !showDrop && suggestions && suggestions.length && (!suggestionsAtClose || suggestionsAtClose.length !== suggestions.length)) {
      openDrop();
    }
```

You can also see the issue in the storybook : 
https://storybook.grommet.io/?path=/story/input-textinput-default-suggestion--default-suggestion

- type in sug, the dropdown is open
- double click inside and delete
- type again sug => no dropdown opens anymore.
=> You need to click outside and click inside again to make it work.

I am using the latest grommet version 2.17.3

novykh:
@ischoe please check this PR https://github.com/grommet/grommet/pull/5625 to see if it works fine for you. Maybe show some love in order to merge it

PR #5625, opened 2021-09-24
Text input suggestion update.

novykh:
#### What does this PR do?
Fix suggestions drop handling

The main issue (for https://github.com/grommet/grommet/issues/5426) here is the simplified solution that checks the `suggestionsAtClose.length` in order to show or not the drop - which especially on dynamic suggestions prop causes issues, since although there might be the same length of suggestions, the suggestions are different and should be shown.
The fix lies on reseting the `suggestionsAtClose` on select and on input value change.

Also, (for https://github.com/grommet/grommet/issues/5440) we should check if the reason that the drop closes is clicking on the input itself.

Added a failing test on selecting a suggestion closes the drop. Changing the value of the input, should bring up the suggestions even when the length is the same as before.

https://github.com/grommet/grommet/issues/5440 Clicking on the input again should not close the drop.
https://github.com/grommet/grommet/issues/5426 On the [story](https://storybook.grommet.io/?path=/story/input-textinput-suggestions--suggestions) write "1 suggestion", select the suggestion, remove some chars - suggestions don't come back up.

https://github.com/grommet/grommet/issues/5440
https://github.com/grommet/grommet/issues/5426

No

Maybe?

Yes it is

### DIFF
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -481,6 +481,11 @@ const TextInput = forwardRef(
                     if (onChange) onChange(event);
                   }
             }
+            onClick={() => {
+              if (suggestions && !showDrop) {
+                openDrop();
+              }
+            }}
           />
         </Keyboard>
 
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -145,11 +145,21 @@ const TextInput = forwardRef(
       if (onSuggestionsOpen) onSuggestionsOpen();
     }, [announce, messages, format, onSuggestionsOpen, suggestions]);
 
-    const closeDrop = useCallback(() => {
-      setSuggestionsAtClose(suggestions); // must be before closing drop
-      setShowDrop(false);
-      if (onSuggestionsClose) onSuggestionsClose();
-    }, [onSuggestionsClose, suggestions]);
+    const closeDrop = useCallback(
+      (event) => {
+        if (
+          event &&
+          event.type === 'mousedown' &&
+          event.target === inputRef.current
+        ) {
+          return;
+        }
+        setSuggestionsAtClose(suggestions); // must be before closing drop
+        setShowDrop(false);
+        if (onSuggestionsClose) onSuggestionsClose();
+      },
+      [onSuggestionsClose, suggestions, inputRef],
+    );
 
     // Handle scenarios where we have focus, the drop isn't showing,
     // and the suggestions change. We don't want to open the drop if
@@ -478,14 +488,10 @@ const TextInput = forwardRef(
                     // placeholder only appears when there is no value
                     setValue(event.target.value);
                     setActiveSuggestionIndex(resetSuggestionIndex);
+                    setSuggestionsAtClose();
                     if (onChange) onChange(event);
                   }
             }
-            onClick={() => {
-              if (suggestions && !showDrop) {
-                openDrop();
-              }
-            }}
           />
         </Keyboard>
 
--- a/src/js/components/TextInput/__tests__/TextInput-test.tsx
+++ b/src/js/components/TextInput/__tests__/TextInput-test.tsx
@@ -95,7 +95,7 @@ describe('TextInput', () => {
     }, 50);
   });
 
-  test('opens drop on focus after selection of suggestion', (done) => {
+  test('opens drop on typing again after selection of suggestion', (done) => {
     const onChange = jest.fn();
     const onFocus = jest.fn();
     const { getByTestId } = render(
--- a/src/js/components/TextInput/__tests__/TextInput-test.tsx
+++ b/src/js/components/TextInput/__tests__/TextInput-test.tsx
@@ -95,6 +95,49 @@ describe('TextInput', () => {
     }, 50);
   });
 
+  test('opens drop on focus after selection of suggestion', (done) => {
+    const onChange = jest.fn();
+    const onFocus = jest.fn();
+    const { getByTestId } = render(
+      <TextInput
+        data-testid="test-input"
+        id="item"
+        name="item"
+        suggestions={['test']}
+        onChange={onChange}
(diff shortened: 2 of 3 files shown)


src/js/components/TextInput/TextInput.js
```diff
@@ -260,13 +270,15 @@ const TextInput = forwardRef(
...
       inputRef.current.focus();
       inputRef.current.value = suggestion; // needed for uncontrolled cases
-      closeDrop();
+
       if (handleSuggestionSelect) {
         if (event.persist) event.persist();
         const adjustedEvent = event;
         adjustedEvent.suggestion = suggestion;
         handleSuggestionSelect(adjustedEvent);
+        setSuggestionsAtClose();
```
jcfilben, line 279: Why is `setSuggestionsAtClose()` being called here? And why is suggestionsAtClose being set to an empty value?

src/js/components/TextInput/TextInput.js
```diff
@@ -480,6 +492,7 @@ const TextInput = forwardRef(
                     // placeholder only appears when there is no value
                     setValue(event.target.value);
                     setActiveSuggestionIndex(resetSuggestionIndex);
+                    setSuggestionsAtClose();
```
jcfilben, line 495: Same question as before, why is `setSuggestionsAtClose()` being called here? And why is suggestionsAtClose being set to an empty value?

jcfilben (commented):
Thanks for taking these two issues on with this pull request.
I'm a little confused about the setSuggestionsAtClose() calls. It seems like suggestionsAtClose should always reflect the full list of items in suggestions.

src/js/components/TextInput/TextInput.js
novykh: Halo thanks for reviewing. My understanding is that `suggestionsAtClose` are needed in order to avoid re-opening the drop when select a value or blur out of the input and the drop closes. That means they are valid only for the current value in the input (since just focusing back, doesn't change the suggestions). As soon as we change the value though, we need to open the drop again and present the suggestions to the user. 
Check [this](https://storybook.grommet.io/?path=/story/input-textinput-custom-suggestions--custom-suggestions) for example: Type `alan` -> one suggestion comes up -> focus again, select the text and just replace it with `j`, drop won't open even though the suggestion (again just one) is completely different.

jcfilben: I don't think we should be setting suggestionsAtClose to and empty value since that doesn't accurately reflect the intended value of the variable. Based on this comment:
```
// Handle scenarios where we have focus, the drop isn't showing,
// and the suggestions change. We don't want to open the drop if
// the drop has been closed by onEsc and the suggestions haven't
// changed. So, we remember the suggestions we are showing when
// the drop was closed and only re-open it when the suggestions
// subsequently change.
```
I think this is the intended use of suggestionsAtClose

novykh: I'm sorry but maybe you haven't check the issue: https://github.com/grommet/grommet/issues/5426 and also I guess you haven't done what I wrote on my previous comment.
I will leave that PR here, if there is anyone else that wants to review it feel free, otherwise close it.
Chreerz

jcfilben: Hi @novykh I did look at the issue and test using the Custom suggestions story you mentioned above. I was just thinking that setting the `suggestionsAtClose` variable might not be the best solution since it doesn't accurately reflect the intended value of the variable. Instead I think we could do something like this to achieve the desired behavior:
```javascript
onChange={
  readOnly
    ? undefined
    : (event) => {
        // when TextInput is not contained in a Form, no re-render
        // will come from this onChange and remove the placeholder
        // so we need to update state to ensure the styled
        // placeholder only appears when there is no value
        if (suggestions && focus && !showDrop) {
          openDrop();
        }
        setValue(event.target.value);
        setActiveSuggestionIndex(resetSuggestionIndex);
      }
}
onClick={() => {
  if (suggestions && !showDrop) {
    openDrop();
  }
}}
```

novykh: Hey, ok maybe it's my personal opinion. But since this variable exists only "for when the drop will close" in order to hijack it let's say, and not re-open. By the moment the user changes the value, those suggestions are obsolete. I will try your suggestion, since it looks smaller and I don't want to change the way you've built things. But the data will be obsolete when checking the length of `suggestionsAtClose` vs `suggestions` since the length might be the same, but the suggestions will be different. Do you mind if I open later a new PR with your suggestion, and mention you for visibility?

jcfilben: The suggestionsAtClose variable exists so that if the suggestions change and the TextInput still has focus the drop will reopen if it was closed with 'escape'. And yes you can open a new PR and mention me on it

novykh: OK one month passed by, but I'm back.
@jcfilben I tried your solution, and after selecting a suggestion (by clicking it) the drop closes, thus when you click the input again, it re-opens. This is good, but it's not covering the whole issue here.
So to clarify here is an example where you use the TextInput for searching stuff:

1. type `a`
2. the suggestions drop opens with
```
all are great!
all are great?
```
3. you select one of them with your keyboard
4. the drop closes but you are still focused on the input.
5. you backspace and remove the punctuation character in the end
5.i. should show the 2 same suggestions again
6. nothing happens because when you selected there were 2 suggestion and again now there are 2 suggestions

So, given that people are using the suggestions prop in order to always display suggestions as the user types, I don't understand why aren't we supposed to reset the `suggestionsAtClose` after the user enters a new value - in the example above "removing a character".

You can check your thoughts against the test I added on this PR. Thank you for your time, but since I see that after all this debate there is not a common understanding on the fix needed I don't feel I can continue with this. You can go ahead and close the PR. Cheers

jcfilben:
Closing since https://github.com/grommet/grommet/pull/5860 was merged and solved the issue

PR #5626, opened 2021-09-24
NEW NameValueList - Grid Approach

taysea:
#### What does this PR do?

This is a first pass to explore the API structure for a NameValueList component. This is an "Approach 2" to the other NameValueList approach started in https://github.com/grommet/grommet/pull/5623.

The layout of this approach relies on Grid component. Where the Grid becomes single column once size of `small` is reached.

Proposed API is as follows:
```
align?: {
    name?: AlignType;
    value?: AlignType;
  };
  columns?: GridColumnsType;
  data?: {
    name?: string;
    nameIcon?: React.ReactElement;
    value?: string;
    valueIcon?: React.ReactElement;
  }[];
  direction?: { list?: 'column' | 'row'; pair?: 'column' | 'row' };
  gap?: GapType | { row?: GapType; column?: GapType };
  nameProps?: BoxTypes;
  valueProps?: BoxTypes;
```

Things to consider:
- [ ] Expected data structure (not sure about `nameIcon` / `valueIcon`)

To-do:
- [ ] Add in any additional accessibility attributes
- [ ] Add / Split up stories
- [ ] Add tests
- [ ] Add propTypes, etc. files

src/js/components/NameValueList/NameValueList.js

View the storybook deploy: 

https://www.figma.com/file/Ny7rYK5gI5VO4PtZsq6vGC/HPE-Name-Value-Pair?node-id=148%3A1936

Related to https://github.com/grommet/grommet/issues/5604

### DIFF
--- a/src/js/components/Grid/index.d.ts
+++ b/src/js/components/Grid/index.d.ts
@@ -34,6 +34,13 @@ export type GridSizeType =
   | string
   | string[];
 
+export type GridColumnsType =
+  | GridSizeType
+  | GridSizeType[]
+  | {
+      count?: 'fit' | 'fill' | number;
+      size?: GridSizeType;
+    };
 export interface GridProps {
   a11yTitle?: A11yTitleType;
   alignSelf?: AlignSelfType;
@@ -42,13 +49,7 @@ export interface GridProps {
   areas?: { name?: string; start?: number[]; end?: number[] }[] | string[][];
   as?: PolymorphicType;
   border?: BorderType;
-  columns?:
-    | GridSizeType
-    | GridSizeType[]
-    | {
-        count?: 'fit' | 'fill' | number;
-        size?: GridSizeType;
-      };
+  columns?: GridColumnsType;
   fill?: FillType;
   gap?: GapType | { row?: GapType; column?: GapType };
   gridArea?: GridAreaType;
--- a/src/js/components/NameValueList/NameValueList.js
+++ b/src/js/components/NameValueList/NameValueList.js
@@ -28,6 +28,7 @@ const NameValueList = forwardRef(
 
     return (
       <Grid
+        as="dl"
         ref={ref}
         columns={columnsProp || columns}
         gap={gap || theme.nameValueList.gap}
--- a/src/js/components/NameValueList/NameValueList.js
+++ b/src/js/components/NameValueList/NameValueList.js
@@ -1,0 +1,54 @@
+import React, { forwardRef, useContext } from 'react';
+import { ThemeContext } from 'styled-components';
+import { Grid } from '../Grid';
+import { ResponsiveContext } from '../../contexts/ResponsiveContext';
+import { NameValuePair } from './NameValuePair';
+
+const NameValueList = forwardRef(
+  (
+    {
+      a11yTitle,
+      align,
+      columns: columnsProp,
+      data,
+      direction = { list: 'column', pair: 'row' },
+      gap,
+      nameProps,
+      valueProps,
+      ...rest
+    },
+    ref,
+  ) => {
+    const size = useContext(ResponsiveContext);
+    const theme = useContext(ThemeContext);
+
+    let columns;
+    if (size === 'small' || direction.list === 'row') columns = 'medium';
+    else if (direction.list === 'column') columns = ['small', 'medium'];
+
+    return (
+      <Grid
+        ref={ref}
+        columns={columnsProp || columns}
+        gap={gap || theme.nameValueList.gap}
+        fill={direction.list === 'row'}
+        {...rest}
+      >
+        {data.map((datum) => (
+          <NameValuePair
+            key={datum?.name}
+            align={align}
+            data={datum}
+            direction={direction.pair}
+            nameProps={nameProps}
+            valueProps={valueProps}
+          />
+        ))}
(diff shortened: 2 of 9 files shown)


MikeKingdom:
value should probably be able to be a react component. I can imagine various cases where that's needed.

taysea:
> value should probably be able to be a react component. I can imagine various cases where that's needed.

The intention of "valueIcon" is to allow the value to be prefixed with an icon or visual element. Do you anticipate other needs for value to be something more complex?

MikeKingdom:
> > value should probably be able to be a react component. I can imagine various cases where that's needed.
> 
> The intention of "valueIcon" is to allow the value to be prefixed with an icon or visual element. Do you anticipate other needs for value to be something more complex? @MikeKingdom

Yes, definitely. From something as simple as value plus action links to something more complex to display the value with more formatting, meters, etc. Kinda endless, hence I'd allow something generic

taysea:
> Yes, definitely. From something as simple as value plus action links to something more complex to display the value with more formatting, meters, etc. Kinda endless, hence I'd allow something generic

Good points, I would agree. I will make adjustments.

src/js/components/Property/index.d.ts
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+import { TextType } from 'grommet';
```
britt6612, line 2: think this is the wrong import

src/js/components/PropertyList/index.d.ts
```diff
@@ -0,0 +1,33 @@
+import { BoxTypes } from 'grommet';
```
britt6612, line 1: this import looks incorrect

src/js/components/PropertyList/PropertyList.js
```diff
@@ -0,0 +1,45 @@
...
+import { ResponsiveContext } from '../../contexts/ResponsiveContext';
+import { PropertyListContext } from './PropertyListContext';
+
+const PropertyList = forwardRef(
+  (
+    {
+      a11yTitle,
+      align,
+      columns: columnsProp,
+      direction = { list: 'column', property: 'row' },
```
britt6612, line 13: would this make more sense to be 
` direction = { name: 'column', value: 'row' },`

then we would have to change the names below as well
just a little taken back there's `list` and `property` but then `nameProp` and `valueProps`

src/js/components/Property/Property.js
```diff
@@ -0,0 +1,77 @@
...
+        {children}
+      </Box>
+    );
+
+  return (
+    <Container {...containerProps}>
+      {direction === 'column-reverse' ? value : undefined}
+      {name}
+      {direction !== 'column-reverse' ? value : undefined}
+    </Container>
```
britt6612, line 73: This looks a little complicated to be in the return could we do something like 
```
  const first = direction === 'column-reverse' ? value : undefined;
  const second = direction !== 'column-reverse' ? value : undefined;

```
then pass 
```
  return (
    <Container {...containerProps}>
      {first}
      {name}
      {second}
    </Container>
  );
```

maybe with better naming but so its not in the render I was confused at first when looking at thiis

britt6612:
one thing to look at is the screen between `medium` and `smalll` there is a larger gap I believe this is because of the size of the grids.. this is on a 1400px wide screen

Screen Shot 2021-10-04 at 9 43 06 PM

src/js/components/PropertyList/PropertyList.js
taysea: The direction prop is regarding the overall `list` vs the direction of individual properties, so the `property` key determines direction of name & value.

src/js/components/Property/Property.js
taysea: I like the idea of `first` `second`, we can probably mirror what's happening in Button.js

taysea:
> one thing to look at is the screen between `medium` and `small` there is a larger gap I believe this is because of the size of the grids.. this is on a 1400px wide screen.. wonder if we should change too use size `small` for columns in grid
> 
> Screen Shot 2021-10-04 at 9 43 06 PM

Yes, this is a tricky side effect of Grid and the way column width is determined on responsive with the `1fr` -- would be interested to hear others thoughts on how to overcome this: https://github.com/grommet/grommet/blob/master/src/js/components/Grid/StyledGrid.js#L100

taysea:
Write tests for NameValueList:
- [ ] Should render correct width of name
- [ ] Should render correct width of value
- [ ] Should render pairs in a grid when layout="grid"
- [ ] Should render pairs in a column when layout="column" (default)
- [ ] Should render correct alignment of name
- [ ] Should render correct alignment of value
- [ ] Should render correct gap between rows and columns

Write test for NameValuePair:
- [ ] Should render name when name is typeof string
- [ ] Should render name when name is typeof JSX Element
- [ ] Should render value when provided as child of type string
- [ ] Should render value when provided as child of type number
- [ ] Should render value when provided as child of type JSX Element
- [ ] Should support custom theming

halocline:
@jcfilben We are also right at our bundle size limit and this component will likely tip us over. Let's connect and discuss options.

ericsoderberghp:
Let's up the bundlesize threshold to 121 kb as part of this pull request. Then, I think we'll be good to go.

PR #5628, opened 2021-09-25
add hasLabel and hasIcon prop to StyledButtonKind

### LINKED ISSUE #5613: iconProp and labelProp extend props in theme for Button
I would like to extend css for specific conditions when the icon and label props has been passed to the Grommet Button, but observed that there are no Grommet specific props available. It would be very helpful to me to have these props.

### Expected Behavior

There is iconProp and labelProp available for extending CSS.

### Actual Behavior

The props are not available.

image

### Steps to Reproduce

Console log out the extend props.

- Grommet version: 2.16.2
- Operating System and version (desktop or mobile): Mac desktop

lleBit:
#### What does this PR do?
add missing hasIcon and hasLabel to StyledButtonKind. Additionally, sets the line-height to 0 when there is only an icon.

Button.js

Manual + added a unit test.

1. Console log out the props under StyledButtonKind.
2. Console log out the button extend props in the jest test.

I need these props to style the theme of a button that only has an icon

hasIcon and hasLabel is accessible to StyledButton already.
closes #5613 

I don't think so

maybe no

non-breaking changes

image


src/js/components/Button/__tests__/Button-test.js
```diff
@@ -569,4 +569,18 @@ describe('Button', () => {
 
     expect(container.firstChild).toMatchSnapshot();
   });
+
+  test(`line-height should be zero for icon only`, () => {
+    const { getByRole } = render(
+      <Grommet
+        theme={{
+          button: { default: {} },
+        }}
```
jcfilben, line 578: Button should not have default defined in the theme. Default should only be defined if using button kind.

```suggestion
```

lleBit: @jcfilben - Thanks for catching that! Removed the theme.

jcfilben (commented):
Screen Shot 2021-09-27 at 2 03 17 PM
This change doesn't look great. The Buttons with text are not aligned with the Button with only an icon. I'm wondering if we should remove this:

```
${(props) =>
    props.hasIcon &&
    !props.hasLabel &&
    `
    line-height: 0;
  `}
```

Is there a reason we want to set the line-height to 0 when a Button only contains an icon?

lleBit:
> Is there a reason we want to set the line-height to 0 when a Button only contains an icon?

@jcfilben - I copied that code from StyledButton.js. I figured it would make it consistent with the behavior that already exists.

It's not a big deal since I can override it on my side of the code. StyledButton.js and StyledButtonKind.js wouldn't behave the same, but maybe that's what you guys want. Let me know.

halocline:
> > Is there a reason we want to set the line-height to 0 when a Button only contains an icon?
> 
> @jcfilben - I copied that code from StyledButton.js. I figured it would make it consistent with the behavior that already exists.
> 
> It's not a big deal since I can override it on my side of the code. StyledButton.js and StyledButtonKind.js wouldn't behave the same, but maybe that's what you guys want. Let me know.

Thank you for checking for parity between the two. In this case though, I do not think the line-height adjustment is needed (it was added to [fix an alignment issue particular to StyledButton](https://github.com/grommet/grommet/pull/2770), but that issue is not present in StyledButtonKind).

Let's revert that adjustment which should cleanup the snapshot changes. Everything else is looking great.

lleBit:
> Let's revert that adjustment which should cleanup the snapshot changes. Everything else is looking great.
@halocline 

Sounds good. I just removed it and also the test. I'm not sure how to test this since it's a prop that gets passed to StyledButtonKind. Thank you!!

halocline:
> Sounds good. I just removed it and also the test. I'm not sure how to test this since it's a prop that gets passed to StyledButtonKind. Thank you!!

I think in this particular case I would be comfortable without a supporting test, because it is more of internal implementation detail. @jcfilben thoughts?

jcfilben (approved):
Looks good!

Issue #5598, opened 2021-09-15
Color is not applied to Spinner when hpe theme is set 


jonka5:
### Expected Behavior

The color set in the color property of the Spinner component should override the theme setting.

### Actual Behavior

It always displays the color which is set in the theme and ignores the color set in the Spinner component.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-spinner-vw28f?file=/index.js

### Use Case

In a colored button (login/submit/etc.) I want a white spinner and on different backgrounds (white/colored) I want different colored spinner.

### Your Environment

- Grommet version: 2.17.5
- Browser Name and version: any
- Operating System and version (desktop or mobile): windows

PR #5631, opened 2021-09-26
Fix Spinner color being override by theme

jeferson-sb:
#### What does this PR do?
Prevent the Spinner color prop (when set) from being override by theme setting.

`src/js/components/Spinner/Spinner.js`

`yarn test`

`yarn storybook`

See #5598 

No

No

backwards compatible

### DIFF
--- a/src/js/components/Spinner/Spinner.js
+++ b/src/js/components/Spinner/Spinner.js
@@ -44,6 +44,7 @@ const Spinner = forwardRef(
     const {
       size: sizeThemeProp,
       color: colorThemeProp,
+      border: borderThemeProp,
       ...themeProps
     } = theme.spinner.container;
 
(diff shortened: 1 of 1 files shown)


src/js/components/Spinner/Spinner.js
```diff
@@ -53,6 +54,21 @@ const Spinner = forwardRef(
...
 
+    const defaultBorder = [
+      { side: 'all', color: 'background-contrast', size },
+      { side: 'top', color, size },
+    ];
+    const spinnerBorder = Array.isArray(borderThemeProp)
+      ? borderThemeProp.map((borderSide) => ({
+          ...borderSide,
+          color:
+            borderSide.side === 'all'
```
wenderjean, line 65: Are we able to cover this with an automated test?

jeferson-sb: @wenderjean I'm not entirely sure, but this seems to be covered on this spec: https://github.com/grommet/grommet/blob/master/src/js/components/Spinner/__tests__/Spinner-test.js#L77

jcfilben:
I think if border is defined in the theme and a color is given through the spinner prop the spinner should still keep the border styling coming from the theme but just change the color.

src/js/components/Spinner/Spinner.js
jcfilben: We should have an automated test with a theme and the color prop defined on spinner

jeferson-sb:
@jcfilben oh yeah, you're right, I mixed up the spinner size with border-size by accident

amandadupell (approved):
nice work! lgtm

jcfilben (approved):

Issue #5619, opened 2021-09-23
DateInput onChange not triggered when controlled


taysea:
### Expected Behavior

When DateInput value changes in a controlled use case, the maskedinput should display those dates.

### Actual Behavior

Dates do not appear in the maskedinput when state changes in controlled case, onChange is not fired but when you open the Calendar, the dates are selected properly.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/dateinput-controlled-5rkud

### Steps to Reproduce

1. Click Change date
2. Nothing is in the masked input
3. Click into the DateInput to open the calendar and see that the correct dates have been selected even though they aren't appearing in the input itself.

### Your Environment

- Grommet version: 2.17.5
- Browser Name and version: Chrome Version 93.0.4577.82 (Official Build) (x86_64)
- Operating System and version (desktop or mobile): MacOS Big sur

Spectre-ak:
Can I work on this?

halocline:
> Can I work on this?

It looks like this issue already has a [PR in review](https://github.com/grommet/grommet/pull/5632).

PR #5632, opened 2021-09-27
Date input value bug

g4rry420:
#### What does this PR do?
Resolved dateInput value bug when it has been called from the outside function.

components => DateInput => utils.js

None

Test using a below storybook example

This happens because of the way `every` array method works in js. It will still return true for the empty array. More info can be found [here](https://stackoverflow.com/questions/34137250/why-does-array-prototype-every-return-true-on-an-empty-array/34137289)

#5619 

No

Backwards Compatible


g4rry420:
For testing purpose, you can use this code:
```
import React from 'react';

import {
  Box,
  Form,
  Grommet,
  DateInput,
  FormField,
  Button,
  TextInput,
} from 'grommet';
import { hpe } from 'grommet-theme-hpe';

export const Test = () => {
  const [value, setValue] = React.useState([]);
  const [value1, setValue1] = React.useState('');

  const onChange = (event) => {
    console.log('onChange in Test', nextValue);
    setValue(nextValue);
  };

  return (
    <Grommet theme={hpe}>
      <Box pad="small">
        <Form>
          <FormField>
            <DateInput
              value={value}
              format="mm/dd/yyyy-mm/dd/yyyy"
              onChange={onChange}
            />
          </FormField>
          <FormField>
            <TextInput value={value1} />
          </FormField>
          <Button
            type="submit"
            label="Change Date"
            primary
            onClick={() => {
              setValue([
                '2020-07-31T15:24:26.256Z',
                '2020-08-07T15:24:26.256Z',
              ]);
              setValue1('Text Value');
            }}
          />
        </Form>
      </Box>
    </Grommet>
  );
};

export default {
  title: 'Input/DateInput/Test',
};

```

jcfilben (commented):
Typing directly into the DateInput in the above example no longer works

g4rry420:
I have made the changes but not sure, if this is the correct approach. Please, let me know about it.

jcfilben (commented):
Now I am able to make changes to the Date after setting it with the 'change date' button but if the date input is empty I can't type directly into it. See video below. I think this has to do with the comment above the use affect function about using the textToValue and valuesAreEqual functions to determine if a user is typing a date and hasn't finished yet or if a caller is changing the value


https://user-images.githubusercontent.com/54560994/135496314-a8551c10-8361-4248-aca6-425140b4ada9.mp4

g4rry420:
Thank you for explaining with the video. I just rushed into finishing it without testing.
I believe my latest commit will resolve this issue. Please, let me know if you find any other issues.

jcfilben (approved):
Nice solution, this looks good to me

halocline:
> Nice solution, this looks good to me

@jcfilben , do you know if we have a Jest test in place to cover `date should be displayed when value is controlled`? Would this be valuable?

jcfilben:
> @jcfilben , do you know if we have a Jest test in place to cover date should be displayed when value is controlled? Would this be valuable?

We don't currently have a test for this but I do think that would be a good thing to add

jcfilben (approved):
Thanks for adding a test!

amandadupell (approved):
lgtm!

Issue #5605, opened 2021-09-20
Enhance `truncate="tip"` on Text to also support `dropProps`


taysea:
### Expected Behavior

When user applies truncate="tip", they should be able to also apply `dropProps` that affect the tip alignment (similar to the tip component).

### Actual Behavior

When user applies truncate="tip", they have no control over the drop alignment.
https://storybook.grommet.io/?path=/story/type-text-tip--tip

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: Stable
- Browser Name and version: Chrome Mac Version 93.0.4577.82 (Official Build) (x86_64)
- Operating System and version (desktop or mobile): MacOS Big Sur

PR #5637, opened 2021-09-28
Text to accept dropProps when truncate='tip'

g4rry420:

#### What does this PR do?
Text Component to accept dropProps when truncate='tip'

components => Text

None

Check the storybook changes

N/A

#5605

Not sure

Not sure

Backwards compatible

### DIFF
--- a/src/js/components/Text/Text.js
+++ b/src/js/components/Text/Text.js
@@ -51,7 +51,11 @@ const Text = forwardRef(
 
     if (tip) {
       if (typeof tip === 'string') {
-        return <Tip content={tip}>{styledTextResult}</Tip>;
+        return (
+          <Tip content={tip} dropProps={rest?.dropProps}>
+            {styledTextResult}
+          </Tip>
+        );
       }
       return <Tip {...tip}>{styledTextResult}</Tip>;
     }
--- a/src/js/components/Text/stories/Tip.js
+++ b/src/js/components/Text/stories/Tip.js
@@ -8,7 +8,12 @@ export const Tip = () => (
   <Grommet theme={grommet}>
     <Box align="center" pad="medium" gap="xlarge">
       <Box width="small">
-        <Text truncate="tip">{alphabet}</Text>
+        <Text
+          truncate="tip"
+          dropProps={{ align: { left: 'right', top: 'bottom' } }}
+        >
+          {alphabet}
+        </Text>
       </Box>
       <Text
         tip={{ dropProps: { align: { left: 'right' } }, content: 'tooltip' }}


src/js/components/Text/Text.js
```diff
@@ -51,7 +51,11 @@ const Text = forwardRef(
 
     if (tip) {
       if (typeof tip === 'string') {
-        return <Tip content={tip}>{styledTextResult}</Tip>;
+        return (
+          <Tip content={tip} dropProps={rest?.dropProps}>
```
taysea, line 55: Should we pull out `dropProps` as a documented prop where we mentioned it only take affect when `tip` is applied? and then just have this as  `dropProps={dropProps}`?

g4rry420: Correct me if I am wrong, you meant to say that, we should pull out `dropProps` separately rather than using it as `rest?.dropProps` ?

jcfilben: Yes, I think that is what @taysea is saying. We should add the prop on Text and use it directly instead of `rest?.dropProps`. Also we will need to add the prop to the documentation on [grommet-site](https://github.com/grommet/grommet-site) and specify that it should only be used when there is a `tip`.

jcfilben (approved):
Looks good to me! Just need to make a PR for grommet-site documentation

ericsoderberghp:
I am wondering whether `dropProps` is the right name for this property. My hesitation is that the `tip` is a particular type of Drop that could be associated with a Text component but there might be other kinds of Drops down the road. I'm wondering if we should rename it to something more connected with `tip`. One idea would be `tipDropProps`, another would be to make `tip` richer with something like: `tip={{ content: 'my tip content', dropProps: {}}}`. Just wanting to initiate a discussion for now.

g4rry420:
In my opinion, rather than making a new Prop, we can make `tip` richer. It already accepts the `{
  content: node | string,
  dropProps: {},
  plain: boolean
}` as per the docs. We only have to made changes for the condition, when `truncate='tip'`. Please, let me know you thoughts.

ericsoderberghp:
> In my opinion, rather than making a new Prop, we can make `tip` richer. It already accepts the `{ content: node | string, dropProps: {}, plain: boolean }` as per the docs. We only have to made changes for the condition, when `truncate='tip'`. Please, let me know you thoughts.

Are you suggesting `<Text truncate="tip" tip={{ dropProps: { ... } }} />`? If so, that sounds reasonable to me.

taysea (approved):
Great job, thanks for adding this!

Issue #5599, opened 2021-09-19
[TS] Migrate tests to .tsx


alexmirrington:
### Overview and Motivation

This is a ticket to aid the migration of `*.js` tests to `*.tsx`. The general motivation here is to improve TypeScript support in Grommet so that when type checks fail, so do unit tests. For more details about why and how we are doing this, please refer to the related Issues and PRs below.

### Related Issues and PRs

Addition of TS unit testing support:

- Issue: #5111
- PR: #5589

### Contributing

 Since we have close to 75 components now and the tests for each range from simple to quite complex, It makes sense to migrate a single component at a time to `*.tsx` files, taking a test-driven development (TDD) approach.

- Pick a single component at a time to work on test migrations for.
- Create a draft PR while you work on the components tests. Mention this issue in the PR to make sure multiple people don't double up on a single component.
- Rename the tests for that component from `*-test.js` to `*-test.tsx`
- Rename the `*-test.js.snap` files for that component to `*-test.tsx.snap`
- Run the test suite for that component with `yarn test`. No test snapshots should fail, but some tests might fail due to type check failures.
- If migrating tests causes only a few Typescript errors (_e.g._ missing field on prop types), feel free to fix those issues in the same PR as the test migration **including the test failure and an explanation about how the changes fix the test failure.** If there are a lot of new errors, create separate issues for each error and fix them in a separate PR. Update the test migration branch with these changes until the tests pass.

### Roadmap

Test migration completed for:

- [x] Accordion #5609 
- [x] AccordionPanel #5610
- [x] Anchor #5600
- [x] Avatar #5602
- [x] Box #5611
- [x] Button #5645
- [x] Calendar #5656
- [x] Card #5612
- [x] ~~CardBody~~ No tests to migrate
- [x] ~~CardFooter~~ No tests to migrate
- [x] ~~CardHeader~~ No tests to migrate
- [x] Carousel #5643
- [x] Chart https://github.com/grommet/grommet/pull/6192
- [x] CheckBox #5644
- [x] CheckBoxGroup #5653
- [x] Clock #5655
- [x] Collapsible #5686
- [x] DataChart #6377
- [x] DataTable
- [x] DateInput #5696
- [x] Diagram #5759 
- [x] Distribution #6378
- [x] Drop #5714 
- [x] DropButton #6379
- [x] FileInput #5696
- [x] Footer #5703
- [x] Form
- [x] FormField
- [ ] Grid
- [ ] Grommet
- [x] Header #5701
[...]

PR #5645, opened 2021-10-01
[TS] Migrate Button tests to .tsx

anderoonies:
#### What does this PR do?

This PR migrates Button component tests from *.js to *.tsx to allow ts-jest. A few changes to the tests **and index.d.ts** had to be made to support this, and are justified below.

- Ensure all tests are currently passing
- Confirm the updated type for `Button` in `index.d.ts` is correct

- Manually ran and passed all component tests locally

- Confirm the updated `Button` type doesn't produce type errors

Test changes for types:
- Previously `console.warn = jest.fn()` was being used to prevent the `Error` from popping up in tests, but that later required `console.warn.mockReset()`. This is a type error, since `mockReset()` doesn't exist on `console.warn`. This could be resolved with casting `console.warn as jest.Mock<typeof console.warn>`, but the cast was producing an `eslint` error. Mocking the implementation to an empty thunk accomplishes the same goal of not throwing the error, but without the type issue.
- `._values` is not a typed property on `CSSStyleDeclaration` (returned from `getComputedStyle`). I replaced this with directly accessing the `.cursor` style, which works.
- **The type of `Button` is changed** to provide typed props to a functional child. Previously, 
```tsx
<Button>
{({disabled}) => { `${disabled}` }
</Button>
```
was unable to infer the type of `disabled`. Explicitly typing `Button`'s `children` with:
```ts
declare const Button: React.FC<
  ButtonExtendedProps &
    anchorType & {
      children?:
        | React.ReactNode
        | (({
            disabled,
            hover,
            focus,
          }: {
            disabled: boolean;
            hover: boolean;
            focus: boolean;
          }) => React.ReactNode);
    }
>;
```
fixes this.

#5599

N/A

No

Maybe, the public types are updating

This could be a breaking change for typed usage of `Button` that makes improper use of a functional child's props as `any`


jcfilben (approved):
Looks good!

ericsoderberghp (approved):

ericsoderberghp:
needs some merge conflict resolution

jcfilben (approved):

PR #5650, opened 2021-10-02
updated Image component

### LINKED ISSUE #5495: Image component keeping previous props
I have in an Image component 'src' and 'fallback' props. Since as first time the src is undefined (''), when a value is assigned to the src, it does not change display. I had to create two Image components, each with a different key as turnaround to the issue. I will show.

### Expected Behavior

I would expect that I have only to declare only one Image component with a `src={image_source}` props that will be assigned dynamically and the component should be capable to reload itself or recognize that there´s a valid src and display it.

### Actual Behavior

As of now, it seems that it keeps the values of initial props (I don't understand how) and do not respond to dynamic changes to the props.

### Steps to Reproduce

To reproduce, 
create a component that after 2 seconds it will assign a valid value to the source, it will maintain the value in the fallback:

<Image {...{
    fill: 'horizontal',
    fallback: fallback_img, 
    src: img_src,
}}/>

Instead if you do as the next part, it will work as expected: 
const [MainImage, setMainImage] = useState<undefined | ReactElement>()
if (img_src) {
      setMainImage(
        <Image {...{
          key: 'main_image_with_src',
          fallback: fallback_img, 
          src: img_src,
        }}/>
      )
    } else {
      setMainImage(
        <Image {...{
          key: 'main_image_no_src',
          fill: 'horizontal',
          fallback: fallback_img, 
          src: img_src,
        }}/>
      )
    }
return (<>{ MainImage }</>)

### Your Environment

- Grommet version: 2.17.4
- Browser Name and version: chrome Version 92.0.4515.107 (Official Build) (64-bit)
- Operating System and version (desktop or mobile): ubuntu 20.0.4, desktop, Node 14

Hope it's clear. Thanks! :)

Spectre-ak:
Fixes #5495

fixed image component so that it'll update src when props are updated

#### What does this PR do?
Fixes #5495 issue, when props are updated the image src still uses old props. 
by directly using the updated image component
Tested locally and as per the instructions at #5495 

Manually testing can be done by creating an Image component and setting the initial src to undefined("") and changing it to a valid one after few seconds.

none
#5495 
no
no
yes, backward compatible

### DIFF
--- a/src/js/components/Image/Image.js
+++ b/src/js/components/Image/Image.js
@@ -4,16 +4,23 @@ import { ImagePropTypes } from './propTypes';
 
 const Image = forwardRef(
   ({ a11yTitle, fallback, onError, opacity, fill, src, ...rest }, ref) => {
-    const [imageMissing, setImageMissing] = useState(false);
-    const handleError = (event) => {
-      if (onError) {
-        onError(event);
-      }
-      setImageMissing(true);
+    const [isFallbackInUse, setFallbackFlag] = useState(false);
+
+    const handleError = (event)=>{
+      if(!isFallbackInUse)
+        event.target.src=fallback;
+      setFallbackFlag(true);
     };
+
+    const handleOnLoad = () => {
+      setFallbackFlag(false);
+    };
+
     const extraProps = {
-      onError: (onError || fallback) && handleError,
+      onError: handleError,
+      onLoad : handleOnLoad,
     };
+    
     return (
       <StyledImage
         aria-label={a11yTitle}
@@ -22,7 +29,7 @@ const Image = forwardRef(
         ref={ref}
         opacityProp={opacity}
         fillProp={fill}
-        src={!imageMissing ? src : fallback}
+        src={src}
       />
     );
   },


src/js/components/Image/Image.js
```diff
@@ -4,16 +4,24 @@ import { ImagePropTypes } from './propTypes';
...
-        onError(event);
-      }
-      setImageMissing(true);
+    const [isFallbackInUse, setFallbackFlag] = useState(false);
+    
+    /* eslint-disable no-param-reassign */
+    const handleError = (event)=>{
+      if(!isFallbackInUse)
+        event.target.src=fallback;
+      setFallbackFlag(true);
```
jcfilben, line 13:
```suggestion
    const handleError = (event)=>{
      if (onError)
        onError(event);
      if(!isFallbackInUse && fallback && fallback !== ""){
        // eslint-disable-next-line no-param-reassign
        event.target.src=fallback;
        setFallbackFlag(true);
      }
```
I think we should still include `onError`. We should also check to make sure fallback is defined and it is not an empty string.

src/js/components/Image/Image.js
```diff
@@ -4,16 +4,24 @@ import { ImagePropTypes } from './propTypes';
...
     };
+
+    const handleOnLoad = () => {
+      setFallbackFlag(false);
+    };
+
     const extraProps = {
-      onError: (onError || fallback) && handleError,
+      onError: handleError,
```
jcfilben, line 21: Can we keep this as `onError: (onError || fallback) && handleError`?

src/js/components/Image/__tests__/Image-test.js
```diff
@@ -96,5 +96,5 @@ test('Image onError', () => {
     fireEvent(getByAltText('test'), new Event('error'));
   });
 
-  expect(onError).toHaveBeenCalledTimes(1);
+  expect(onError).toHaveBeenCalledTimes(0);
```
jcfilben, line 99:
```suggestion
  expect(onError).toHaveBeenCalledTimes(1);
```
We should still expect onError to be called 1 time.

jcfilben (commented):
Nice work so far, just a few comments/questions. I am also thinking we should include a jest test that tests that at first the fallback image is shown and then after a second passes the updated image should be shown instead.

jcfilben:
For reference here is the story I tested with. It jumps between the fallback image, the regular image and a fallback image again.
```javascript

import { Grommet, Image, Box, Button } from 'grommet';

export const Test = () => {
    const [imgSrc, setImgSrc] = useState("");
  return (
    <Grommet theme={grommet}>
      <Box width="medium" gap="medium">
        <Image
          fill="horizontal"
          fallback="//v2.grommet.io/assets/IMG_4245.jpg"
          src={imgSrc}
        />
        <Button label="Update Image" onClick={() => {setTimeout(() => {
            setImgSrc('https://v2.grommet.io/img/stak-hurrah.svg');
          }, 1000)
          setTimeout(() => {
            setImgSrc('');
          }, 5000)}} />
      </Box>
    </Grommet>
  );
};

export default {
  title: 'Media/Image/Test',
};
```

PR #5654, opened 2021-10-03
[TS] Migrate WorldMap tests to .tsx

anderoonies:
#### What does this PR do?

WorldMap tests from .js to .tsx

Confirm tests pass

Full `yarn test`, CI run from below

Running `yarn test`

I added `@ts-ignore` to the two `data-testid` that were being used. Doesn't seem appropriate to cast these or update the allowed types.

#5599

No

No

Backwards compatible


src/js/components/WorldMap/__tests__/WorldMap-test.tsx
```diff
@@ -69,6 +69,7 @@ describe('WorldMap', () => {
         <WorldMap
           places={[
             {
+              // @ts-ignore
               'data-testid': 'Sydney',
```
alexmirrington, line 73: Is `data-test-id` reflected in the test snapshots at all or can we remove it entirely? Would prefer removal over `// @ts-ignore` if it's not affecting the test output at all and it's just left from some legacy code.

src/js/components/WorldMap/__tests__/WorldMap-test.tsx
```diff
@@ -104,6 +105,7 @@ describe('WorldMap', () => {
         <WorldMap
           continents={[
             {
+              // @ts-ignore
```
alexmirrington, line 108: Same feedback as above.

src/js/components/WorldMap/__tests__/WorldMap-test.tsx
anderoonies: It is in the snapshots, and I think it's needed to select the target for the events. Tests break without it

jcfilben: 'data-testid' is being used in the query statements for the tests but I am thinking we can remove it and change the query statements. For example the following query uses 'data-testid':
`fireEvent.mouseOver(getByTestId('Sydney'));`
I think we could change it to something like this:
`fireEvent.mouseOver(screen.getByRole('button', {name: /Sydney/i}));`

anderoonies:
Good point, I’ll update that and generate a new snapshot, since I believ
the testid is present in the snapshot

anderoonies:
@jcfilben @alexmirrington updated to use `getByLabelText`, which selects on the `aria-label` property that's on the places. let me know if that works!

jcfilben (approved):
Great, thanks for updating this!

ericsoderberghp (approved):

Issue #5403, opened 2021-06-29
Checkbox focus state is stuck if the checkbox is disabled


MLRichmond:
### Expected Behavior

Focus should still respond appropriately after a checkbox has been disabled and subsequently re-enabled. 

### Actual Behavior

Focus borders become locked in place once a checkbox has been disabled. 

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/unruffled-browser-hewpg?file=/src/CheckBox.js

### Steps to Reproduce

1. Add a checkbox
2. Click the checkbox and disable it in the on-click (so before another control can acquire focus)
3. The checkbox focus border is now stuck, even if the checkbox is subsequently re-enabled.

See code sandbox. Other controls, such as button, do not seem to have this defect.

### Your Environment

- Grommet version: 17.3
- Browser Name and version: Chrome 91.0.4472.101
- Operating System and version (desktop or mobile): Desktop

Spectre-ak:
Hi, I've created a PR for this issue [Fixes Checkbox focus state when checkbox is disabled](https://github.com/grommet/grommet/pull/5658)

PR #5658, opened 2021-10-04
Fixes Checkbox focus state is stuck if the checkbox is disabled

Spectre-ak:
Updated CheckBox component to turn off the focus when checkbox is disabled

#### What does this PR do?
Fixes [Checkbox focus state is stuck if the checkbox is disabled](https://github.com/grommet/grommet/issues/5403) issue

Can start by creating a checkbox and then disable it and watching on focus

Tested locally and also on the sample code provided on the issue page

None
#5403 
none
no
no
yes, backwards compatible

### DIFF
--- a/src/js/components/CheckBox/CheckBox.js
+++ b/src/js/components/CheckBox/CheckBox.js
@@ -66,6 +66,11 @@ const CheckBox = forwardRef(
       initialValue: defaultChecked,
     });
 
+    if (disabled) {
+      /* eslint-disable no-param-reassign */
+      focusProp = false; 
+    }
+
     const [focus, setFocus] = useState(focusProp);
     useEffect(() => setFocus(focusProp), [focusProp]);
 


src/js/components/CheckBox/CheckBox.js
```diff
@@ -66,6 +66,11 @@ const CheckBox = forwardRef(
       initialValue: defaultChecked,
     });
 
+    if (disabled) {
+      /* eslint-disable no-param-reassign */
+      focusProp = false; 
+    }
+
     const [focus, setFocus] = useState(focusProp);
     useEffect(() => setFocus(focusProp), [focusProp]);
```
jcfilben, line 75:
```suggestion
    const [focus, setFocus] = useState(focusProp);
    useEffect(() => {
      // don't allow checkbox to have focus when disabled
      if (disabled && focusProp) setFocus(false);
      else setFocus(focusProp);
    }, [disabled, focusProp]);
```
I'm thinking we should do something like this. Then we don't have to re-assign a function parameter. Also useEffect should be used for code that happens during lifecycle of the component, such as removing focus when disabled switches from false to true.

Spectre-ak: Yes sure. Using useEffect we won't be re-assigning function param. 
Thanks for the suggestion. I've made the changes as suggested.

jcfilben (approved):
Look good!

ericsoderberghp (approved):

Issue #5647, opened 2021-10-02
Control margin of Tabs without changing margin of Tab content


lleBit:
Currently, there is no way to set the margin of the Tabs header independently of the Tab content.
I would like to be able to align the Tabs header without affecting the content.

### Expected Behavior
There should be a Tabs prop like headerMargin to be able to set the header margin separately.

### Actual Behavior
There is only a margin prop which sets the margin of the content.

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/s/determined-bhaskara-ndddw?file=/src/Tabs.js
This kind of describes the issue if you inspect the margin.

### Steps to Reproduce
1. set the Tabs margin
2. notice that the Tab content margin is constrained within that margin.

### Your Environment
- Grommet version: 2.16.2

lleBit:
I'm open to submitting a PR for this

lleBit:
https://github.com/grommet/grommet/pull/5661

jcfilben:
Just wanted to check back in and see if there was still interest from anyone in this enhancement

PR #5661, opened 2021-10-04
add headerMargin prop to Tabs

lleBit:
#### What does this PR do?

This PR adds option to control header alignment via the headerMargin prop, without changing the content margins (panel margins)

Tabs.js

Manually adjusted margins using headerMargin 

Adjust headerMargin for Tabs (using small, medium, large, or px)

I would like to be able to control the header margin separately from the content. This allows me to control the margins of the header alignment, while leaving the content alignment alone.

Currently, I'm using a workaround invisible border to do this.

Yes. Add the headerMargin.

Probably yes.

This change will not break anything.

### DIFF
--- a/src/js/components/Tabs/Tabs.js
+++ b/src/js/components/Tabs/Tabs.js
@@ -19,6 +19,7 @@ const Tabs = forwardRef(
       justify = 'center',
       messages,
       responsive = true,
+      headerMargin,
       ...rest
     },
     ref,
@@ -107,6 +108,7 @@ const Tabs = forwardRef(
           wrap
           background={theme.tabs.header.background}
           gap={theme.tabs.gap}
+          margin={headerMargin}
           {...tabsHeaderStyles}
         >
           {tabs}


jcfilben:
This behavior can currently be achieved with the following:
```javascript
tabs: {
  header: {
    extend: css `
      margin: 10px;
    `
  },
}
```
Not sure if we want to include the `headerMargin` prop since it can already be adjusted though the tabs.header.extend in the theme. I'm curious of your thoughts on this @ericsoderberghp

lleBit:
> This behavior can currently be achieved with the following:
> 
> ```js
> tabs: {
>   header: {
>     extend: css `
>       margin: 10px;
>     `
>   },
> }
> ```
> 
> Not sure if we want to include the `headerMargin` prop since it can already be adjusted though the tabs.header.extend in the theme. I'm curious of your thoughts on this @ericsoderberghp

This would set ALL the tabs to have the same margin, right? Also, I think you wouldn't be able to use the Grommet defined 'small, medium, large...' etc...  Am I wrong?

jcfilben:
> This would set ALL the tabs to have the same margin, right? Also, I think you wouldn't be able to use the Grommet defined 'small, medium, large...' etc... Am I wrong?

Yes, this would define the margin for all <Tabs>. If you only wanted to set the margin on one Tab heading ThemeContext.Extend could be used.

And no the T-shirt sizes ('small', 'medium', 'large', etc.) could not be used with this method.

lleBit:
> > This would set ALL the tabs to have the same margin, right? Also, I think you wouldn't be able to use the Grommet defined 'small, medium, large...' etc... Am I wrong?
> 
> Yes, this would define the margin for all . If you only wanted to set the margin on one Tab heading ThemeContext.Extend could be used.
> 
> And no the T-shirt sizes ('small', 'medium', 'large', etc.) could not be used with this method.

If possible, I would like to be able to set the header margins individually and additionally also use the T-shirt sizes.

src/js/components/Tabs/Tabs.js
```diff
@@ -107,6 +108,7 @@ const Tabs = forwardRef(
           wrap
           background={theme.tabs.header.background}
           gap={theme.tabs.gap}
+          margin={headerMargin}
```
britt6612, line 111: not sure if we would want this to be a `prop` could we maybe do something in the theme similar to `theme.tabs.header.margin`

britt6612: @ericsoderberghp curious to hear your thoughts around adding a prop vs theme in my opinion if we can achieve this behavior in theme that would be the best place.

lleBit: I think we should allow it to be themed but the prop would be helpful for individually aligning each tabs group separately.

britt6612: @lleBit good point to individually align. In that case this looks good I would just see if you can add a test reflecting this new prop

lleBit: @britt6612 - Can we do both options?

Also, I'm not sure the best way to test this as it's just a margin control on a component, but I'll take a look

halocline:
Thanks everyone for your inputs and feedback on this request.

Hi @lleBit, I've assembled a [CodeSandbox](https://codesandbox.io/s/tabs-header-margin-zfbfq?file=/src/App.js) demonstrating how you can achieve your desired implementation (e.g. two or more variants of tabs header margins, plus ability to apply those margins using t-shirt sizing.) 

I've provided two approaches. One, which uses `<ThemeContext.Extend />` and is probably more suited for scenarios where you only need to modify the theme in a one-off scenario. The second approach applies `styled-components` to `Tabs` and allows you to have a second variant of your tabs implementation. This approach is useful if you plan on reusing each variant throughout your application.

After discussion, our thinking is the following:

- For user experience, it is important to have consistency throughout applications.  In general, it seems that tabs are something that folks would want to be re-usable and consistent, and that the need for one-off implementations should be rare. The theme is a great place to achieve that consistency.
- Your use case is certainly valid (e.g. one tabs layout/styling for a page-level tabs, and a second tabs layout for content-detail tabs), however I think it can be achieved without needing to alter the props API. In this case, we'd like to err on leaving as is until we hear greater demand for this feature from the community.

Can you please verify that the [CodeSandbox](https://codesandbox.io/s/tabs-header-margin-zfbfq?file=/src/App.js) demonstrates how you can achieve the two "flavors" of tabs you desire and meets your use case?

lleBit:
> Can you please verify that the [CodeSandbox](https://codesandbox.io/s/tabs-header-margin-zfbfq?file=/src/App.js) demonstrates how you can achieve the two "flavors" of tabs you desire and meets your use case?

@halocline - Thank you for providing those examples! I've tried those options before and confirmed again that it doesn't match my exact use case. While it does allow you to use the pre-defined t-shirt sizes, the padding will not adjust itself as the window resizes.

I've been using the header: border: {size" xlarge} setting in the theme as a workaround at the moment, and 'size' does adjust itself when the window resizes. Is there a workaround for this?

I attached a screenshot so you can see my current workaround.

Screen Shot 2021-11-03 at 5 18 51 PM

halocline:
> > Can you please verify that the [CodeSandbox](https://codesandbox.io/s/tabs-header-margin-zfbfq?file=/src/App.js) demonstrates how you can achieve the two "flavors" of tabs you desire and meets your use case?
> 
> @halocline - Thank you for providing those examples! I've tried those options before and confirmed again that it doesn't match my exact use case. While it does allow you to use the pre-defined t-shirt sizes, the padding will not adjust itself as the window resizes.
> 

Hi @lleBit - I've modified the examples to demonstrate how you can achieve the responsive edge styles (i.e. change margin based on viewport width). The implementation could be refined, but should convey the concept.

Can you please checkout the modifications in the [CodeSandbox](https://codesandbox.io/s/tabs-header-margin-zfbfq?file=/src/App.js) and see if they now are meeting your needs? 🤞 

Lastly, if we were to add a prop to adjust margin, we'd probably want something less acute, so that we would be solving for header styling in general. We'd probably do something like a `headerProps` prop which would accept any `boxProps`.

lleBit:
> Lastly, if we were to add a prop to adjust margin, we'd probably want something less acute, so that we would be solving for header styling in general. We'd probably do something like a `headerProps` prop which would accept any `boxProps`.

Hi @halocline. This solution works as a workaround. Adding breakpoint as a prop didn't work well with TypeScript, but I was able to utilize a media query to adjust the padding at the breakpoint.

Though I can live without this change, I still think we should have the ability to easily control the header padding/margins and the content padding/margins easily with props if needed.

Is the final decision to wait until there's more of a need for this?

Screenshot
image

halocline:
> Though I can live without this change, I still think we should have the ability to easily control the header padding/margins and the content padding/margins easily with props if needed.
> 
> Is the final decision to wait until there's more of a need for this?

I think it would be good to modify this PR (or create a new issue) to propose "Enhance Tabs by allowing caller to more easily control the header and panel content padding and margins." Along with the proposal would be the proposed [backwards compatible] props/theme API surface and designs/use cases such as the one you provided above. I think there is value in your proposal and the main thing is getting the API surface correct. This will provide something concrete for folks to react to.

P.S. Thank you for contributing and advocating for your and the community's needs. Super valuable and much appreciated. 🙏

lleBit:
> > Though I can live without this change, I still think we should have the ability to easily control the header padding/margins and the content padding/margins easily with props if needed.
> > Is the final decision to wait until there's more of a need for this?
> 
> I think it would be good to modify this PR (or create a new issue) to propose "Enhance Tabs by allowing caller to more easily control the header and panel content padding and margins." Along with the proposal would be the proposed [backwards compatible] props/theme API surface and designs/use cases such as the one you provided above. I think there is value in your proposal and the main thing is getting the API surface correct. This will provide something concrete for folks to react to.
> 
> P.S. Thank you for contributing and advocating for your and the community's needs. Super valuable and much appreciated. 🙏

Awesome! I will create a new issue for this suggestion and close this PR. Thank you!

lleBit:
I just updated the issue https://github.com/grommet/grommet/issues/5647 since the intention was the same.

PR #5665, opened 2021-10-05
Fixed bug for focus on first day of month when selecting range

Kishore-abhimanyu:
Fixed bug for focus on first day of month when selecting range


Kishore-abhimanyu:
Please review the PR?

jcfilben:
Closing since this issue was fixed in #5784

PR #5666, opened 2021-10-05
Add pad option to theme for rangeInput

britt6612:
#### What does this PR do?
adds an option to the theme for `pad` RangeInput

StyledRangeInput.js
storybook
storybook 
In the design-System, we need to add padding to RangeInput when it is wrapped in FormField. 

In FormField there is an option that adds `pad` for when it wraps around RangeInput
however here we are having the pad be undefined  
https://github.com/grommet/grommet-theme-hpe/blob/master/src/js/index.js#L680
so we need another way to leverage the theme in order to put the pad.

https://github.com/grommet/hpe-design-system/issues/1909
yes I will update the docs 
yes
backwards compatible


src/js/components/FormField/FormField.js
```diff
@@ -32,12 +32,7 @@ const grommetInputNames = [
   'FileInput',
   'RadioButtonGroup',
 ];
-const grommetInputPadNames = [
-  'CheckBox',
-  'CheckBoxGroup',
-  'RadioButtonGroup',
-  'RangeInput',
-];
+const grommetInputPadNames = ['CheckBox', 'CheckBoxGroup', 'RadioButtonGroup'];
```
taysea, line 35: It seems like removing this here is causing snapshot changes for existing storybooks since base.js cause `formField.content.pad = 'small'` by default. Can we leave RangeInput in this array but just add the pad option like you've done below to enable themes like grommet-theme-hpe to specify pad directly on the RangeInput since we override the formfield.content.pad to undefined?

britt6612: opps this was not meant to be changed 
part of me playing around with things

PR #5667, opened 2021-10-05
Align DateInput behavior with native HTML date picker

taysea:
#### What does this PR do?

Completes work started by https://github.com/grommet/grommet/pull/5550 to align DateInput with [native HTML date picker behavior](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date). Some of the changes from #5550 were reverted in a later DateInput PR that was open concurrently, so this re-aligns with the desired behavior for keyboard. Additionally, when DateInput is clicked, the calendar should not open right away, instead the user clicks "space" to open the drop. This is consistent with other inputs and native HTML date picker behavior.

I have also updated the accessibility messages to align with the behavior and confirmed they were announced properly with VoiceOver.

Storybook with DateInput

Storybook.

No.

Yes. Improved DateInput keyboard and mouse behavior.

Changes when DateInput calendar opens, but aligns it with native HTML as was desired by #5550

### DIFF
--- a/src/js/components/DateInput/DateInput.js
+++ b/src/js/components/DateInput/DateInput.js
@@ -201,10 +201,11 @@ const DateInput = forwardRef(
               }
             }}
             onFocus={(event) => {
-              openCalendar();
+              announce(
+                formatMessage({ id: 'dateInput.openCalendar', messages }),
+              );
               if (onFocus) onFocus(event);
             }}
-            onClick={openCalendar}
           />
         </Keyboard>
       </FormContext.Provider>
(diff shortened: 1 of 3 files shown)


britt6612 (approved):

amandadupell (approved):
looks good to me. checked voiceover and spacebar new capability.

ericsoderberghp (approved):

Barnyclark:
On the native HTML date picker, the calendar shows when the icon is clicked - is there a way to enable this behaviour on the Grommet date input?

PR #5668, opened 2021-10-05
Add alwaysShowMask to MaskedInput

### LINKED ISSUE #5480: [Feature] Create permanent maskings for MaskedInput
Currently, the `MaskedInput` component accepts a prop `mask` that takes an array of objects that identify whether the mask is either a string or regex matcher and what the behavior should be for each.  This is sufficient for most use-cases, but misses a large subset in which the mask should _always_ be in a particular position if there is a value in the field.

These cases include, but are not limited to:

- currencies that leverage the ISO code instead of a symbol (100 INR or 1500 USD)
- percentages (25%)
- almost anything that leverages a specific unit (kg or GB)

While there are workarounds for this, none of them provide the context in as clear a way as a masked input does in that the context is immediately after the letters being typed. (Check out [this Slack thread](https://grommet.slack.com/archives/C09QQEG69/p1627495571044700?thread_ts=1627489921.044400&cid=C09QQEG69) for some of the workarounds...)

This proposal is to add a property similar to `fixed` to the possible elements within a mask object - I'll refer to that property as `permanent`.

### Expected Behavior

```
const mask = [
  { permanent: "https://" },
  { regex: /^.*$/, placeholder: "my-subdomain" },
  { permanent: ".foobar.com" }
];
```

The usage of permanent would mean that the mask is there whenever the typed value is not undefined.

#### Examples with above mask

**Case: User hasn't typed anything**
Rendered Text: (placeholder) `https://my-subdomain.foobar.com`
Value: `undefined`

**Case: User types ba**
Rendered Text: `https://ba.foobar.com`
Value: `"ba"`

**Case: User deletes all text**
Rendered Text: (placeholder) `https://my-subdomain.foobar.com`
Value: `undefined`

### Actual Behavior

Currently, the closest that you can get with this is to use `fixed` with very constrained rules surrounding the fixed value, which means that you have to have a very constrained definition of what the mask can be.

MikeKingdom:
#### What does this PR do?
This is an early experiment to add `alwaysShowMask` to MaskedInput.
The goal is to give people a chance to look at the behavior in the Storybook/MaskedInput/URL example. The second example there is based on the original request #5480 to have a permanent style of fixed sections.

Please ignore the details in the code at the moment, including `console.log`s :) and any test failures.

With this property set, the value of the input will always try to show all parts of the mask. Fixed sections will always be their fixed value and non-fixed parts will be their inputted values or the placeholder for that chunk if possible.

When you focus the input it tries to select the first part this is just a placeholder.
When you move the caret left or right it tries to only allow that caret in the non-fixed sections.

It still has some edge case issues, especially on last non-fixed sections on regarding backspace/delete. This is partially given to how it maps existing text to the mask.
One improvement idea is to select a placeholder section if the caret is moved to the start of the placeholder chunk.

Any feedback welcome on whether this seems useful.

Storybook/MaskedInput/URL

Fixes #5480 

Yes

Yes

Backwards compatible

### DIFF
--- a/src/js/components/MaskedInput/MaskedInput.js
+++ b/src/js/components/MaskedInput/MaskedInput.js
@@ -23,7 +23,7 @@ import {
 } from './StyledMaskedInput';
 import { MaskedInputPropTypes } from './propTypes';
 
-const parseValue = (mask, value) => {
+const parseValue = (mask, value, alwaysShowMask) => {
   // break the value up into mask parts
   const valueParts = []; // { part, beginIndex, endIndex }
   let valueIndex = 0;
@@ -49,9 +49,10 @@ const parseValue = (mask, value) => {
         matching += 1;
       }
 
+      console.log('fixed ', item.fixed, matching, valueIndex, value.length);
       if (matching > 0) {
         let part = value.slice(valueIndex, valueIndex + matching);
-        if (valueIndex + matching < value.length) {
+        if (valueIndex + matching - 1 < value.length) {
           // matched part of the fixed portion but there's more stuff
           // after it. Go ahead and fill in the entire fixed chunk
           part = item.fixed;
@@ -135,6 +136,21 @@ const parseValue = (mask, value) => {
       }
     }
   }
+  console.log('VALUE PARTS', value, valueParts);
+  if (alwaysShowMask) {
+    while (maskIndex < mask.length) {
+      console.log('NEED to ADD parts');
+      const item = mask[maskIndex];
+      const part = item.fixed || item.placeholder || '_';
+      valueParts.push({
+        part,
+        beginIndex: valueIndex,
+        endIndex: valueIndex + part.length - 1,
+      });
+      valueIndex += part.length;
+      maskIndex += 1;
+    }
+  }
   return valueParts;
 };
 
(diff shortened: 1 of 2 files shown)


taysea (commented):
I think the behavioral piece of this that I find confusing is that when I see the placeholder, I interpret that as the format I need to start typing my input as, so when the fixed portion suddenly appears after typing, it is a little confusing. If I can't ever edit a certain portion of the input, I would almost prefer it to be persistent and in text color as opposed to placeholder.

For example, if I were going to enter my site as `abc.foobar.com` and see the `https://` in the placeholder, I might think that I need to type the https, and so I type `h`, then I see the placeholder get filled in. To me, that's a little bit confusing. I'm more familiar with a state prefix/suffix like this, where the country code is input and fixed. Not sure how to handle that for fixed sections in the middle of inputs though ..

Screen Shot 2021-10-06 at 9 51 03 AM

ericsoderberghp:
> For example, if I were going to enter my site as `abc.foobar.com` and see the `https://` in the placeholder, I might think that I need to type the https, and so I type `h`, then I see the placeholder get filled in. To me, that's a little bit confusing.

Agreed on this.

Also, I cannot backspace to undo. So, if I type "h" and then realize I meant to type "j", my first instinct is to hit backspace to start over but backspace doesn't do anything.

I'm wondering whether instead of having a complete placeholder that then gets replaces on the first character we should not have a placeholder and just show "https://.foobar.com" with the text entry caret between the "/." characters? That might be clearer as far as text entry goes, I just get to edit there. But, the text entry caret might be too visually subtle to understand what's going on when focus arrives.

I'm also wondering whether in the case the issue wants to address we should just recommend `<Text>https://</Text><TextInput /><Text>.foobar.com</Text>`.

I'm suspecting we won't be able to find a solution that combines a simple implementation with a good user experience.

MikeKingdom:
We're choosing not to pursue this for now since it will take a lot more development and design to do this appropriately.

PR #5689, opened 2021-10-15
Feature: Video controls customization

### LINKED ISSUE #5531: Video - extendable controls
Allow adding the ability to resize a Video from a thumbnail to a larger size within the browser, such as the full width of the browser window, similar to how other embedded videos work. This allows the user to transition from deciding whether they want to watch a video to focusing on watching it, without having to make it full screen.

Suggest achieving this by adding something like `<Video controls={[]} />` which would add additional controls to the current controls drop list. The caller would then add a control that allows the user to initiate it. The caller would then change the layout of the page as needed, especially sizing and placing the container of the Video as desired.

jlaluces123:
#### What does this PR do?
This PR aims to:
- Add the ability to add new controls to the video control list
- Extend the `controls` prop to be able to order, pick, and add their own controls

`index.d.ts` and `Video.js` to view the changes structurally to allow for this functionality

Testing via a new storybook titled `Theater mode`

Resolves #5531 

Yes, will be updated to highlight new structural capabilities of `controls` prop

Yes

Backwards compatible


src/js/components/Video/stories/typescript/ControlsBelow.tsx
```diff
@@ -25,7 +25,13 @@ const SimpleVideo = props => (
   </Grommet>
 );
 
-export const ControlsBelow = () => <SimpleVideo controls="below" />;
+export const ControlsBelow = () => (
+  <SimpleVideo
+    controls={{
+      position: 'below',
+    }}
+  />
```
jcfilben, line 33:
```suggestion
  <SimpleVideo controls='below' />
```
We should be able to leave this as controls='below' for backwards compatibility. Right now if we do that it shows the controls 'over'.

src/js/components/Video/stories/typescript/TheaterMode.tsx
```diff
@@ -0,0 +1,51 @@
...
+        />
+      </Box>
+    </Grommet>
+  );
+};
+
+export const CustomizableControls = () => <TheaterMode />;
+
+CustomizableControls.storyName = 'Theater mode';
+
```
jcfilben, line 48:
```suggestion

CustomizableControls.parameters = {
  chromatic: { disable: true },
};

```
I think we can disable this story from chromatic testing

jcfilben (approved):
This looks good to me!
We will need to update the documentation on grommet-site

src/js/components/Video/Video.js
```diff
@@ -48,7 +48,7 @@ const Video = forwardRef(
       alignSelf,
       autoPlay,
       children,
-      controls = 'over',
```
ericsoderberghp, line 51: Why can't we preserve the existing default here? The caller can pass 'over', so we have to be prepared to handle it anyway.

src/js/components/Video/Video.js
```diff
@@ -48,7 +48,7 @@ const Video = forwardRef(
...
+      controls,
```
ericsoderberghp, line 51: I think we should do `controls: controlsProp` here and then use `[controls, setControls]` for the state below. "newControls" is a bit too time sensitive to the where we are today. In 9 months, it will seem odd.

src/js/components/Video/Video.js
```diff
@@ -80,6 +80,26 @@ const Video = forwardRef(
     const containerRef = useRef();
     const scrubberRef = useRef();
     const videoRef = useForwardedRef(ref);
+    const [newControls, setNewControls] = useState({
+      position: 'over',
```
ericsoderberghp, line 84: I think we should leverage the passed in controls prop more here:
`useState(typeof controls === 'string' ? { position: controls, items: [...] } : controls)`

src/js/components/Video/Video.js
```diff
@@ -80,6 +80,26 @@ const Video = forwardRef(
...
+      items: ['volume', 'reduceVolume', 'fullScreen'],
+    });
+
+    useLayoutEffect(() => {
```
ericsoderberghp, line 88: I don't think this needs to use `useLayoutEffect()`. There's nothing about the contents that depends on the DOM render state.
```suggestion
    useEffect(() => {
```

src/js/components/Video/Video.js
```diff
@@ -254,10 +274,71 @@ const Video = forwardRef(
         onClick: () => showCaptions(caption.active ? -1 : 0),
       }));
 
+      const buttonTable = {
```
ericsoderberghp, line 277: This doesn't look like a "Table". Can we call it "buttonProps"?
```suggestion
      const buttonProps = {
```

src/js/components/Video/Video.js
```diff
@@ -254,10 +274,71 @@ const Video = forwardRef(
...
+        },
+      };
+
+      let newItem;
+      const controlsArr =
+        newControls.items &&
+        newControls.items.map((item) => {
+          if (typeof item === 'string') {
+            // Transforming item into PascalCase for lookup on Icons
+            const newString = item.charAt(0).toUpperCase() + item.slice(1);
```
ericsoderberghp, line 306: This feels a little fragile to me. Can we instead create an object to map them formally at the top of this file?

src/js/components/Video/Video.js
```diff
@@ -254,10 +274,71 @@ const Video = forwardRef(
...
+
+            const Icon = Icons[newString];
+            const buttonTitle = buttonTable[item].a11yTitle
+              ? buttonTable[item].a11yTitle
+              : `video.${item}`;
```
ericsoderberghp, line 311: Why are we prefixing the title with "video."? I don't think that will make much sense to a screen reader.

src/js/components/Video/Video.js
```diff
@@ -254,10 +274,71 @@ const Video = forwardRef(
...
+          close: false,
+        },
+        volume: {
+          a11yTitle: 'video.volumeUp',
+          onClick: volume <= 1 - VOLUME_STEP ? louder : undefined,
+          close: false,
+        },
+      };
+
+      let newItem;
```
ericsoderberghp, line 300: I don't think we need "newItem" outside the `map()` function scope.

src/js/components/Video/Video.js
```diff
@@ -254,10 +274,71 @@ const Video = forwardRef(
...
+      const controlsArr =
```
ericsoderberghp, line 301: We try to avoid both abbreviations and type suffixes in variable names. Instead of "controlsArr", I think "controlsMenuItems" will be more descriptive of the contents and its purpose.

src/js/components/Video/Video.js
ericsoderberghp, line 277: For all of these, I think we can include the icon property here as well, instead of calculating it and then adding it later on.

src/js/components/Video/Video.js
```diff
@@ -254,10 +274,71 @@ const Video = forwardRef(
...
+          if (typeof item === 'string') {
+            // Transforming item into PascalCase for lookup on Icons
+            const newString = item.charAt(0).toUpperCase() + item.slice(1);
+
+            const Icon = Icons[newString];
+            const buttonTitle = buttonTable[item].a11yTitle
+              ? buttonTable[item].a11yTitle
+              : `video.${item}`;
+
+            newItem = {
```
ericsoderberghp, line 313: Instead of creating an object and returning it. I think we can just do: `return buttonTable[item];`?

src/js/components/Video/__tests__/Video-test.js
```diff
@@ -67,12 +67,12 @@ describe('Video', () => {
   });
 
   test('controls below renders', () => {
-    const { container } = render(<App controls="below" />);
```
ericsoderberghp, line 70: Let's leave these tests alone, to ensure we haven't broken existing behavior, and then add new tests for the new style of property value.

src/js/components/Video/index.d.ts
```diff
@@ -11,7 +11,28 @@ export interface VideoProps {
...
+            | 'configure'
+            | 'fullScreen'
+            | 'pause'
+            | 'play'
+            | 'reduceVolume'
+            | 'volume'
+          ),
+          {
+            icon?: React.ReactNode;
+            onClick?: () => {};
```
ericsoderberghp, line 32: Let's also accept "a11yTitle" or "title" here as well.

src/js/components/Video/stories/typescript/TheaterMode.tsx
```diff
@@ -0,0 +1,55 @@
...
+          fit="cover"
+          controls={{
+            position: 'below',
+            items: [
+              'volume',
+              'reduceVolume',
+              {
+                icon: <Monitor />,
+                onClick: () =>
+                  setSource(
```
ericsoderberghp, line 29: I'm wondering if leaving the source alone but changing the Box's dimensions might be a better approach, and potentially cause less flicker when toggling the mode. Maybe start with the Box having 'small' width and then unsetting `width` and setting `fill="horizontal"` for theater mode. Can you try that?

ericsoderberghp (commented):
Good progress

src/js/components/Video/Video.js
```diff
@@ -222,8 +252,8 @@ const Video = forwardRef(
     }, [videoRef]);
 
     let controlsElement;
-    if (controls) {
-      const over = controls === 'over';
+    if (controls && controls.position) {
```
ericsoderberghp, line 255:
```suggestion
    if (controls?.position) {
```

src/js/components/Video/Video.js
```diff
@@ -80,6 +80,36 @@ const Video = forwardRef(
...
+    const [controls, setControls] = useState(
+      typeof controlsProp === 'string' || typeof controlsProp === 'boolean'
+        ? {
+            position: controlsProp,
+            items: ['volume', 'reduceVolume', 'fullScreen'],
+          }
+        : controlsProp,
+    );
+
+    useEffect(() => {
```
ericsoderberghp, line 92: I'm thinking we can simplify this whole function a bit down to just:
```
setControls({
  items: controlsProp?.items || ['volume', 'reduceVolume', 'fullScreen'],
  position: (typeof controlsProp === 'string' && controlsProp) || controlsProp?.position || 'over',
});
```

src/js/components/Video/index.d.ts
```diff
@@ -11,7 +11,29 @@ export interface VideoProps {
...
+        position?: false | 'over' | 'below';
+        items?: [
+          (
+            | 'closedCaption'
+            | 'configure'
+            | 'fullScreen'
+            | 'pause'
+            | 'play'
+            | 'reduceVolume'
+            | 'volume'
```
ericsoderberghp, line 28: I suspect these names were selected based on the icon name. But, I don't think these are quite right as control names. Instead, I think we should use: captions, fullScreen, play, pause, volume. I don't think we need "configure" as a built in control. I think we should combine the "volume" into a single name here but then expand that to multiple controls automatically internally.

src/js/components/Video/Video.js
```diff
@@ -254,10 +284,68 @@ const Video = forwardRef(
...
+        },
+      };
+
+      const controlsMenuItems =
+        controls.items &&
+        controls.items.map((item) => {
+          if (typeof item === 'string') {
+            return buttonProps[item];
+          }
+          return {
```
ericsoderberghp, line 336: How about just returning `item`? This will allow the caller to set other properties on the Button as needed. I think this is what Menu does.

src/js/components/Video/Video.js
```diff
@@ -399,7 +447,7 @@ const Video = forwardRef(
     }
 
     let style;
-    if (rest.fit === 'contain' && controls === 'over') {
+    if (rest.fit === 'contain' && controls && controls.position === 'over') {
```
ericsoderberghp, line 450:
```suggestion
    if (rest.fit === 'contain' && controls?.position === 'over') {
```

ericsoderberghp (commented):
Also, I think we can remove "small_video.mp4"

src/js/components/Video/Video.js
```diff
@@ -80,6 +80,24 @@ const Video = forwardRef(
     const containerRef = useRef();
     const scrubberRef = useRef();
     const videoRef = useForwardedRef(ref);
+    const [controls, setControls] = useState(
```
ericsoderberghp, line 83: I think we should combine this `useState` and the subsequent `useEffect` into a single `useMemo`. We aren't using `setControls` anywhere else and the effect only depends on the `controlsProp`.

ericsoderberghp (commented):
Just one more small comment

src/js/components/Video/Video.js
jlaluces123: Just switched it over to useMemo! Went over all the behavioral functionalities again and everything checked out!

src/js/components/Video/stories/typescript/TheaterMode.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { useState } from 'react';
+
+import { Box, Grommet, Video, grommet } from 'grommet';
+import { Monitor } from 'grommet-icons/icons/Monitor';
+
+const TheaterMode = (props) => {
+  const [theaterMode, setTheaterMode] = useState(false);
+
+  return (
+    <Grommet theme={grommet}>
```
ericsoderberghp, line 10: Need to remove `<Grommet />`, like in other stories. This might help fix the fact that currently the video is showing full height but small width, which looks weird.

Screen Shot 2021-12-15 at 9 26 52 AM

src/js/components/Video/stories/typescript/TheaterMode.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { useState } from 'react';
+
+import { Box, Video } from 'grommet';
+import { Monitor } from 'grommet-icons/icons/Monitor';
+
+const TheaterMode = (props) => {
+  const [theaterMode, setTheaterMode] = useState(false);
+  const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
```
ericsoderberghp, line 8: This looks like a hack. There should be a way to get the desired behavior without having to resort to this approach. You might want to remove the `fill="vertical"` on the Box, and potentially the `flex={true}`.

ericsoderberghp (approved):

ShimiSun:
Hi! Thanks for this fun feature @jlaluces123! 
Is there any work in progress regarding the updated documentation of the new `controls` customization? 
It will be great to let the users know about this feature.

Issue #5304, opened 2021-06-04
DateInput: wrong date roll over


lozenge0:
### Expected Behavior

When you input a bad date such as the 30th of February any year, the DateInput should flag this as a bad date input and force the user to change the date to something valid or change the date displayed in the DateInput to the corrected date i.e. the rolled over date

### Actual Behavior

When you input a bad date such as the 30th of February any year, the DateInput will automatically roll over the outputted date into the next month and allows a user to click away closing the calendar but keeping the wrong date in the TextInput

### URL, screen shot, or Codepen exhibiting the issue

[choosing the wrong date](https://ibb.co/bRfFYts)

[clicking away and still displaying the wrong date](https://ibb.co/vYG8BZJ)

### Steps to Reproduce

1. click into the DateInput and type the date 31/02/2020
2. click away from the DateInput and the date is wrong but the underlying value will be a correct date 

### Your Environment

- Grommet version: 2.17.2
- Browser Name and version: Chrome version 90.0.4430.93
- Operating System and version (desktop or mobile): Windows 1909 desktop

PR #5715, opened 2021-10-21
Fix/date-input-5304: added more sanity checking for dates, aligned with current behavior

amandadupell:
#### What does this PR do?

Add more sanity checking for DateInput. Checks for edge case of leap years and the month of February. Approaches in the same way as other invalid dates (just doesn't select it in calendar, leaves as is)

DateInput utils file

Passes existing and manual testing for different formats/range/etc

Storybook

#5304

### DIFF
--- a/src/js/components/RangeInput/RangeInput.js
+++ b/src/js/components/RangeInput/RangeInput.js
@@ -1,8 +1,15 @@
-import React, { forwardRef, useContext, useState } from 'react';
+import React, {
+  forwardRef,
+  useContext,
+  useState,
+  useCallback,
+  useEffect,
+} from 'react';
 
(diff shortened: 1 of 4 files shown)


amandadupell:
For this fix, we are just preventing the user from inputting an invalid date. We are already doing some sanity checks for dates, so this fix aligns with our current way of handling invalid dates, now just factoring in leap years and the dates for February. This also applies for ranges.

If a user inputs an invalid date, it will not select a date (like it was before this fix) and will alert the user that a date is still required in the form. See below:
#### Not selecting date anymore
Screen Shot 2021-10-21 at 2 56 45 PM

#### Informs user that a date is still required for form
Screen Shot 2021-10-21 at 2 56 34 PM

PR #5716, opened 2021-10-21
Feat/datatable onupdate

### LINKED ISSUE #5679: DataTable - Better support for lazy loading in grouped datasets
### Expected Behavior

DataTable should allow user to provide information about all potential groups for a given column instead of needing to rely on knowing all the data upfront.

DataTable should also accept an `onMore` in the groupBy prop similar to the `onExpand` that allows more data to be fetched.

This is particularly needed for tables that may be rendering large data sets (1+ million records).

### Actual Behavior

Caller is required to provide DataTable with all of the data upfront to easily handle grouping data.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: latest
- Browser Name and version: chrome Version 94.0.4606.61 (Official Build) (x86_64)

Tasks TODO:
- [x] Group selection needs some improvements
- [x] Test with a real use case/remote data like SpaceX
- [x] Add test cases

MikeKingdom:
#### What does this PR do?
This allows lazy loading by adding an `onUpdate` callback which gets called when something changes which could affect which data is shown. It gets called when a group is expanded or contracted, when a sortable header is clicked on and when more data is required because the user scrolled farther. The presence of `onUpdate` causes the `DataTable` to not attempt to filter, sort or group the data and it will no longer call `onMore`. The intent is `onUpdate` can pass any sort criteria, what groups are expanded and what the maximum number of items are needed to a backend API which can efficiently determine the current set of data and update the `data` property as appropriate. `onUpdate` is called with an argument containing the following information:
```jsx
{
  expanded: [groupKey1, groupKey2], // an array of keys of groups which are currently expanded
  sort: { property: 'propertyName', direction: 'asc' | 'desc' }, // the data property to sort by and direction
  count: number. // the maximum number of items currently needed in the table
  show: number, // an index of the first item to make visible in the table
}
```

For grouped tables that use `onUpdate`, the developer should include group header items in the `data` property on the `DataTable`. It's also required to pass an array of key values which identify which items in `data` are group headers. The `primaryKey` property tells DataTable which attribute in an item contains its key. Any items whose key value matches one of the keys in the `groupBy.expandable`  array will be rendered with an expander control. `groupBy.expand` can also be used to tell DataTable which groups should be expanded.

A simple `DataTable` definition would look like.
```jsx
  const getData = ({expandable, expand, count, step}} => {
    ...
  }

<DataTable
   primaryKey="id"
   columns={columns}
   data={data}
   sortable
   replace
   groupBy={{
      expandable: ['Boise', 'Fort Collins', 'Palo Alto'],
      expand: [],
      property: 'location',
    }}
    onUpdate={opts => {
        setExpand(opts.expanded);
        setData(getData(opts));
    }}
</DataTable>
```

TODO:
* Group selection needs some improvements
* count needs a way to be decreased when possible?

DataTable.js

Storybook DataTable examples

* The DataTable/OnUpdate example from Storybook
* The DataTable/Grouped SpaceX example
* *NEW* The DataTable/Ungrouped SpaceX example

Fixes #5679

Yes

Yes

Backwards compatible


src/js/components/DataTable/stories/OnUpdate.js
```diff
@@ -7,89 +8,219 @@ import { grommet } from 'grommet/themes';
...
+  );
+  
+  const onSelect = useCallback((selected, row) => {
+    const groupUpdates = { };
+    if (row?.location) {
+      // this is a member of a group. Update the group selection state
+      const memberKeys = groupMap[row.location].members.map(({id}) => id);
+      const selectedMembers = selected.filter((s) => memberKeys.includes(s));
+      if (selectedMembers.length === 0) {
+        groupUpdates[row.location] = 'none';
```
taysea, line 163: Since we are referencing `none`, `some`, etc. often as strings, would it be good to extract this as a constant:
```
const selected = {
   none: 'none', 
   some: 'some',
   all: 'all'
};
```
Then reference as `selected .none`?

Not necessary but just to avoid string typos

src/js/components/DataTable/stories/OnUpdate.js
```diff
@@ -0,0 +1,252 @@
+/* eslint-disable no-param-reassign */
+import React, { useCallback, useMemo, useState } from 'react';
```
taysea, line 2: When I expand a group, the groups below it are no longer rendering.

https://user-images.githubusercontent.com/12522275/140793136-70a66324-4796-46e5-a348-d2f94c6eb2d7.mov

:

taysea: It could be that the story just needs overflow="auto"?

MikeKingdom: Its looking like InfiniteScroll has a defect where it calculates the index of the lastPage wrong if items.length is a multiple of step. It's getting 1 too large at the boundary causing it to not call onMore when needed. I'll submit a separate issue and PR for that after running it by Matt also

src/js/components/DataTable/GroupedBody.js
```diff
@@ -88,16 +99,25 @@ export const GroupedBody = forwardRef(
                   : 'body',
               isSelected,
               onChange: () => {
-                if (isSelected) {
-                  onSelect(selected.filter((s) => s !== primaryValue));
-                } else onSelect([...selected, primaryValue]);
+                const nextSelected = isSelected ?
```
taysea, line 102: nice cleanup

src/js/components/DataTable/Header.js
```diff
@@ -182,26 +189,46 @@ const Header = forwardRef(
...
-                    selected.length === data.length
+                    totalSelected === data.length
                       ? 'unselect all'
                       : 'select all'
                   }
                   checked={
-                    selected.length > 0 &&
-                    data.length > 0 &&
-                    selected.length === data.length
+                    groupBy?.select ?
```
taysea, line 197: In the `onUpdate` story, I'm not able to click the checkbox in the datatable header to select all the data.

taysea (commented):
Feels like we're getting very close here!

src/js/components/DataTable/Header.js
```diff
@@ -151,6 +152,12 @@ const Header = forwardRef(
       ? [...pin, 'left']
       : pin;
 
+    const totalSelectedGroups = /* onUpdate && */ groupBy?.select ?
```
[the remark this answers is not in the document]
MikeKingdom: Yup, but I should remove it :) Recognition of the groupBy.select doesn't need to be only for onUpdate.

src/js/components/DataTable/Header.js
MikeKingdom: Good catch. I somehow broke this (unitialized variable.)

src/js/components/DataTable/DataTable.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-nested-ternary */
```
ericsoderberghp, line 1: Do we still need to disable this? I skimmed the file and didn't see any nested ternaries.

src/js/components/DataTable/DataTable.js
```diff
@@ -236,6 +242,16 @@ const DataTable = ({
     else direction = 'asc';
     const nextSort = { property, direction, external };
     setSort(nextSort);
+    if (onUpdate) {
+      onUpdate({
+        expanded: Object.keys(groupState).filter(
```
ericsoderberghp, line 247: Since we do this several times, it might be a little cleaner to create a helper function, to enable these three lines to be one where it's needed.
```
const filterExpandedGroups = (groupState) => groupState.filter(...);
```
just a thought, not required for approval.

src/js/components/DataTable/DataTable.js
```diff
@@ -390,14 +442,24 @@ const DataTable = ({
               cellProps={cellProps.body}
               columns={columns}
               data={!paginate ? adjustedData : items}
-              onMore={onMore}
+              onMore={onUpdate ? () => {
+                if (adjustedData.length === limit) {
+                  onUpdate({
+                    expanded: [],
```
ericsoderberghp, line 448: I'm wondering if we should even include `expanded` when we aren't doing any grouping?

src/js/components/DataTable/buildState.js
```diff
@@ -173,33 +173,45 @@ export const buildFooterValues = (columns, data) => {
...
     data.forEach((datum) => {
+      const key = datumValue(datum, primaryProperty);
+      const isGroup = (key && groupBy.expandable?.includes(key));
+
       const groupByProperty = groupBy.property ? groupBy.property : groupBy;
-      const groupValue = datumValue(datum, groupByProperty);
-      if (!groupMap[groupValue]) {
-        const group = { data: [], datum: {}, key: groupValue };
+      const groupValue = isGroup ? key : datumValue(datum, groupByProperty);
+      if (!groupMap[isGroup ? key : groupValue]) {
```
ericsoderberghp, line 187: I think `groupValue` already is `key` when `isGroup`.
```suggestion
      if (!groupMap[groupValue]) {
```

src/js/components/DataTable/buildState.js
```diff
@@ -173,33 +173,45 @@ export const buildFooterValues = (columns, data) => {
...
     // include any aggregate column values across the data for each group
-    columns.forEach((column) => {
-      if (column.aggregate) {
-        result.forEach((group) => {
-          const { datum } = group;
-          datum[column.property] = aggregateColumn(column, group.data);
-        });
-      }
-    });
+    if (!groupBy.expandable) {
```
ericsoderberghp, line 201: Can we include a comment about why we don't do this when `expandable`?

src/js/components/DataTable/index.d.ts
```diff
@@ -102,8 +105,14 @@ export interface DataTableProps<TRowType = any> {
   onClickRow?: (event: MouseClick<TRowType> | KeyPress<TRowType>) => void;
   onMore?: () => void;
   onSearch?: (search: string) => void;
-  onSelect?: (select: (string | number)[]) => void;
+  onSelect?: (select: (string | number)[], datum: TRowType) => void;
   onSort?: (sort: { property: string; direction: 'asc' | 'desc' }) => void;
+  onUpdate?: (datatableState: {
+    sort: { property: string; direction: 'asc' | 'desc' };
```
ericsoderberghp, line 111: Can we make `sort` and `expanded` optional, an only include them if the properties to enable them are present we we call `onUpdate`? I'm wanting to keep the interface simpler for the typical case.

src/js/components/DataTable/stories/SpaceX.js
```diff
@@ -0,0 +1,196 @@
...
+            if (opts.sort) setSort(opts.sort);
+          }}
+          step={20}
+        />
+      </Box>
+    </Grommet>
+  );
+};
+
+SpaceX.storyName = 'SpaceX Grouped';
```
ericsoderberghp, line 192: I think we should disable this story from Chromatic since the data can change as more rockets are launched.

src/js/components/DataTable/stories/SpaceXUngrouped.js
```diff
@@ -0,0 +1,115 @@
...
+            if (opts.sort) setSort(opts.sort);
+          }}
+          step={20}
+        />
+      </Box>
+    </Grommet>
+  );
+};
+
+SpaceXUngrouped.storyName = 'SpaceX Ungrouped';
```
ericsoderberghp, line 111: I think we should disable this story for Chromatic as well.

src/js/components/DataTable/DataTable.js
MikeKingdom: Nope. Good catch

src/js/components/DataTable/DataTable.js
MikeKingdom: removed

src/js/components/DataTable/buildState.js
MikeKingdom: very true. Thanks

src/js/components/DataTable/buildState.js
MikeKingdom: done

src/js/components/DataTable/index.d.ts
MikeKingdom: Certainly. Is it OK if the expand/sort member is in the object but is undefined or just shouldn't be there? A little messier code to not include it at all.

src/js/components/DataTable/DataTable.js
MikeKingdom: I think because it's so simple and sometimes its groupState and sometimes nextGroupState I might leave it if that's cool.

ericsoderberghp:
Also need to update bundlesize.maxSize in package.json.

src/js/components/DataTable/DataTable.js
```diff
@@ -363,21 +403,38 @@ const DataTable = ({
               ref={bodyRef}
               cellProps={cellProps.body}
               columns={columns}
-              groupBy={groupBy.property ? groupBy.property : groupBy}
+              groupBy={typeof groupBy === 'string' ?
+                { property: groupBy} :
```
taysea, line 407:
```suggestion
                { property: groupBy } :
```
just seems like missing lint thing

taysea (approved):
all behaviors in the storybook look good!

ericsoderberghp (approved):

Issue #5735, opened 2021-10-25
Enhance Stories to Use Storybook Theming Functionality


jcfilben:
The ability to view Storybook stories with different themes was added in https://github.com/grommet/grommet/pull/5674.

We need to make adjustments to the stories component by component to take advantage of the theming ability. The Button stories are currently using this enhancement and can be viewed on [storybook](https://storybook.grommet.io/?path=/story/controls-button-active--active). While on a Button story you can change the theme by selecting the 'Theme' button in the top bar above the story and then selecting a theme from the drop down. Notice the stories changes style when you switch themes. We want to add this feature to all the stories.

For each story within a component if the story...
1. Is wrapped with `<Grommet theme={grommet}>...</Grommet>`
  -- remove `<Grommet theme={grommet}>` and replace with the following comment:
      `// Uncomment <Grommet> lines when using outside of storybook`
      `// <Grommet theme={...}>`
  -- comment out `</Grommet>`
  -- remove `import { Grommet }` and `import { grommet }`
2. Uses a custom theme
  -- create a CustomThemed folder directly under the `component/stories` folder
  -- move the story into the CustomThemed folder
  -- change the story export so that it shows up in a 'Custom Themed' folder within storybook. (Example: `export default { title: 'Controls/Button/Custom Themed/Active' };`)

For examples take a look at https://github.com/grommet/grommet/pull/5674. The `Button/stories/Basic.js` file shows an example of a story wrapped with `<Grommet theme={grommet}>`. The `Button/stories/CustomThemed/Active.js` shows an example of a story that uses a custom theme.

- [x] Accordion #5843 
- [x] Anchor #5737 
- [x] Avatar #5840 
- [x] Box #5962
- [x] Button https://github.com/grommet/grommet/pull/5674
- [x] Calendar #5843 
- [x] Card #5843 
- [x] Carousel #5843
- [x] CheckBoxGroup #5740 
- [x] Chart #5764 
- [x] CheckBox #5740 
- [x] Clock #5767 
- [x] Collapsible #5870 
- [x] DataChart #5765 
- [x] DataTable #5870 
- [x] DateInput #5740
- [x] Diagram #5962
- [x] Distribution #5962
- [x] Drop #5769 
- [x] DropButton #5737 
- [x] FileInput #5740 
- [x] Footer #5962
- [x] Form #5962
- [x] FormField #5962
- [x] Grid #5919
- [x] Header #5772 
- [x] Heading #5768 
- [x] Image #5962
[...]

PR #5738, opened 2021-10-26
Add ability to specify Grommet `full` with themed storybook

jcfilben:
#### What does this PR do?
In stories that use the new theme structure we need the ability to specify the `full` prop on Grommet

See the DropButton/Menu story in #5737 
The story doesn't look right without Grommet `full` prop.

I tested by adding 
```
Menu.args = {
  full: true,
};
```
to the story and checking that the story looked like the original when it is not wrapped with `<Grommet>`

Related to #5735 

No

No

Backwards compatible


storybook/preview.js
```diff
@@ -15,6 +15,7 @@ export const decorators = [
     useEffect(() => {
       setState(context.globals.theme);
     }, [context.globals.theme]);
+    const full = context.parameters?.args?.full;
```
ericsoderberghp, line 18:
```suggestion
    const full = context.parameters?.args?.full || 'min';
```

ericsoderberghp (approved):

Issue #5675, opened 2021-10-08
Grommet Carousel Transition


matsawkaHPE:
Expected Behavior:
When clicking Next or Previous on the Grommet Carousel there would be a smooth transition. Something like Bootstrap, where the current image slides off as the new image slides in.

Actual Behavior
The current image disappears and the new images slides in, which makes the transition feel abrupt especially for large hero carousels. 

Steps to Reproduce
This is the standard out of the box behavior of the carousel as you can see with the grommet example:
https://v2.grommet.io/carousel

Your Environment
- Grommet version: 2.18.0
- Browser Name and version: all 
- Operating System and version: all

halocline:
Notes:
- Abruptness occurs because `visibility` is set to 'hidden' and which does not allow the fadeOut animation to display https://github.com/grommet/grommet/blob/45198d0d6878f79d8e82f0e08ee4190df6d3d839/src/js/components/Carousel/Carousel.js#L162
- Looks like this was a regression introduced by https://github.com/grommet/grommet/pull/5326 when trying to prevent non-active slides from being visible onMount.

kevingill55:
That's right yeah. I can start taking a look at this

ericsoderberghp:
Original requestor is asking for a visual transition that does a whole image slide as opposed to a fade out.

halocline:
Additional details from Slack:

"Hello Grommet Team, I have a deep-pull of a bug for you related to the carousel. On the carousel, we have some text - article title and short description - which is coming from a fetch call, the front-end is using some css to limit the text to 2 lines as to not overflow, the css is:
```
overflow: hidden;
text-overflow: ellipsis;
display: -webkit-box;
-webkit-box-orient: vertical;
-webkit-line-clamp: 2;
```

those 5 lines - which is working, except for Safari and iOS
after some head scratching, I did find this deep-pull article: https://bugs.webkit.org/show_bug.cgi?id=207013
is it possible while you are looking at the source code of the carousel to see if you swapped visibility with opacity, that this issue will be resolved? and either have a fix, or some code to override the visibility with opacity if works?
I didn't see any documentation on how to capture a click event on Next/Previous buttons in the documentation"

I have asked for a CodeSandbox to review. We may have recommendation for an alternate approach for how the truncation / overflow prevention is achieved.

PR #5741, opened 2021-10-27
[Superseded by PR#5839] Carousel Component Revisions

kevingill55:
** This PR is superseded by #5839 **

#### What does this PR do?
This changes the `Carousel` component fairly significantly, solving some issues that exist in the current component.

This changes:
- Transitions between slides.
- Adding dimensions to the carousel container (fixed height and width).
- New ways to navigate between slides.

See the Carousel directory under components. See the JS files there.
General code cleanup and organization. Tested in SB with numerous use cases.
Interesting use cases:
- Carousel with only one slide.
- Carousel with too too many slides.
- Spamming the controls (e.g. clicking the next arrow very fast).
- Skipping around slides. Going from the last slide to the first slide with the "Next" button and vice versa.
- General accessibility. Navigating through controls and any clickable child elements.
Carousel stories in Storybook.
- #3536 
- #5675 
- Possibly others?


src/js/components/Carousel/Carousel.js
```diff
@@ -1,242 +1,222 @@
...
+   * Delays the transitions between Carousel slides. This is needed to
+   * avoid users "spamming" the controls which results in jarring animations
+   * and a bad user experience.
+   */
+  useEffect(() => {
+    let transitionTimer;
+    if (inTransition) {
+      transitionTimer = setTimeout(() => {
+        setInTransition(false);
+      }, 500);
```
halocline, line 144: How was this value arrived at? A comment documenting would be helpful.

src/js/components/Carousel/Carousel.js
```diff
@@ -1,242 +1,222 @@
...
-      setIndexes({ activeIndex: index, priorActiveIndex: activeIndex });
-      onChildChange(index);
+  // Allows Carousel slides to be controlled outside the component
+  useEffect(() => {
+    if (activeChild !== undefined && activeChild !== current) {
+      onJumpNavigation(current, activeChild, false);
     }
-  };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [activeChild]);
```
halocline, line 166: Are there downsides to including 'current' and 'onJump...' as dependencies here?

src/js/components/Carousel/Carousel.js
```diff
@@ -1,242 +1,222 @@
...
-        onChild(index);
-      }
-    },
-    [onChild],
-  );
+  /**
+   * Delays the transitions between Carousel slides. This is needed to
+   * avoid users "spamming" the controls which results in jarring animations
+   * and a bad user experience.
+   */
```
halocline, line 138: Thank you for including comments accompanying useEffects throughout. Very helpful.

src/js/components/Carousel/Carousel.js
```diff
@@ -1,242 +1,222 @@
...
-  const [indexes, setIndexes] = useState({
-    activeIndex: activeChild !== undefined ? activeChild : initialChild,
+  const noContainer = !fill && !height && !width;
+  const numSlides = children.length;
+  const firstChildRef = useRef(null);
+  const [current, setCurrent] = useState(initialChild);
+  const [previous, setPrevious] = useState(undefined);
+  const [containerProps, setContainerProps] = useState({
+    heightProp: height,
+    widthProp: width,
```
halocline, line 76: I think we can get away with:
```suggestion
    height,
    width,
```

If more readable, we tend to rename the destructured prop being passed by the caller to [variable_name]Prop to make it clear where the value is originating from. This is more of a convention.

```suggestion
    height: heightProp,
    width: widthProp,
```

src/js/components/Carousel/Carousel.js
```diff
@@ -1,242 +1,222 @@
...
+        } else setContainerProps({});
+      }
+    } else {
+      setContainerProps({
+        heightProp: fill ? '100%' : height,
+        widthProp: fill ? '100%' : width,
       });
     }
-  }
+  }, [noContainer, fill, height, width]);
```
halocline, line 132: Any reason not to include the full dependency list here?

src/js/components/Carousel/CarouselChild.js
```diff
@@ -1,70 +1,51 @@
...
+      if (index !== current && index !== previous) return;
+
+      if (index === current) {
+        setDisplay('block');
+        if (direction === 'next') setAnimation(slideLeftCurrent);
+        else setAnimation(slideRightCurrent);
+      } else {
+        setTimeout(() => {
+          setDisplay('none');
+        }, 600);
```
halocline, line 31: How was this timeout value determined?

src/js/components/Carousel/CarouselChild.js
```diff
@@ -1,70 +1,51 @@
...
+      // on mount, do not render the first slide with an animation.
+      if (previous === undefined) return;
+      if (index !== current && index !== previous) return;
+
+      if (index === current) {
+        setDisplay('block');
+        if (direction === 'next') setAnimation(slideLeftCurrent);
+        else setAnimation(slideRightCurrent);
+      } else {
+        setTimeout(() => {
```
halocline, line 29: Will this timeout need to be cleared on unmount?

src/js/components/Carousel/CarouselChild.js
```diff
@@ -1,70 +1,51 @@
...
-         * This check will only be false onMount of the component. It ensures
-         * the initial active slide of the Carousel renders with no animation.
-         */
-        setAnimation({
-          type:
-            play || priorActiveIndex < activeIndex ? 'slideLeft' : 'slideRight',
-          size: 'xlarge',
-          duration: theme.carousel.animation.duration,
-        });
+    useEffect(() => {
```
halocline, line 18: Can we add a brief comment explaining what this useEffect is doing?

halocline (commented):
Looking very nice.

Initial pass focussing on API, behavior, Carousel and CarouselChild implementations.

Next will be reviewing CarouselControls and StyledCarousel implementations. Then stories and tests afterward.

src/js/components/Carousel/CarouselChild.js
kevingill55: Nope, I'll clear it manually good catch.

src/js/components/Carousel/CarouselChild.js
kevingill55: It's the length of the animation time, the current value in the theme is 1 second which seems a bit long but we can use that for now (or ideally update that value?).

src/js/components/Carousel/Carousel.js
kevingill55: I've seen in the past that when I spread those props of `height` and `width` to the Styled Box those values are read by the Box and overwrite what is happening the in the Styled Box css logic. But gonna look into this again.

src/js/components/Carousel/CarouselChild.js
ericsoderberghp: Anytime we have a hard-coded timer value, we should have a comment explaining how it was arrived at. Something like "empirically determined based on manual interaction in consultation with designer" would be fine.

src/js/components/Carousel/Carousel.js
kevingill55: It should just be the value of the animation duration, see my other comment this should be `600` since animations are now 600ms.

src/js/components/Carousel/Carousel.js
kevingill55: Yeah we only want to subscribe for `activeChild` updates. If current is modified and it runs those comparisons in the `useEffect` it will see that `activeChild !== current` and it will update the Carousel to the `activeChild` index even though it didn't get modified.

src/js/components/Carousel/Carousel.js
kevingill55: I think it's fully covered - what's missing?

src/js/components/Carousel/Carousel.js
ericsoderberghp: We have gotten burned turning off exhaustive-deps in the past. It would be good to see if we can find a way to deliver the behavior we want without needing to disabled this.

src/js/components/Carousel/Carousel.js
halocline: I believe `containerProps` object used on L114-115.

src/js/components/Carousel/__tests__/Carousel-test.tsx
```diff
@@ -1,139 +1,157 @@
...
+    );
 
-    fireEvent.focus(button!);
-    expect(onFocus).toHaveBeenCalledTimes(1);
+    /**
+     * - Currently on "Slide One"
+     * - Simulating click on the 3rd selector button
+     * - Expecting "Slide Three" to be visible
+     */
+    const thirdSelector = screen.getAllByRole('button')[2];
```
halocline, line 96: If we add a11yTitles to the selectors, we can be more explicit and select by the label text instead.

src/js/components/Carousel/__tests__/Carousel-test.tsx
```diff
@@ -1,139 +1,157 @@
...
-      </Grommet>,
+      <Carousel>
+        <div>Slide One</div>
+        <div>Slide Two</div>
+      </Carousel>,
     );
-
+    expect(getSlideOne()).toBeVisible();
+    expect(getSlideTwo()).not.toBeVisible();
     expect(container.firstChild).toMatchSnapshot();
```
halocline, line 23: I think we are attempting to migrate to the following:
```suggestion
    expect(asFragment()).toMatchSnapshot();
```

will also need 
```
const { asFragment } = render(...)

```

src/js/components/Carousel/__tests__/Carousel-test.tsx
```diff
@@ -1,139 +1,157 @@
...
+    render(
+      <Carousel controls="arrows">
+        <div>Slide One</div>
+        <div>Slide Two</div>
+        <div>Slide Three</div>
+      </Carousel>,
     );
-    expect(container.firstChild).toMatchSnapshot();
-
-    fireEvent.keyDown(getByTestId('test-carousel'), {
```
halocline, line 47: I think we should have test coverage for keyboard interactions, right and left arrow keys.

src/js/components/Carousel/propTypes.js
```diff
@@ -1,31 +1,42 @@
...
 if (process.env.NODE_ENV !== 'production') {
-  PropType = {
+  CarouselPropType = {
     ...genericProps,
     activeChild: PropTypes.number,
-    fill: PropTypes.bool,
-    play: PropTypes.number,
     initialChild: PropTypes.number,
+    fill: PropTypes.bool,
+    controls: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
```
halocline, line 11: Should we accept any strings or enumerate the accepted values?

src/js/components/Carousel/propTypes.js
```diff
@@ -1,31 +1,42 @@
...
+    direction: PropTypes.string,
   };
 }
 export const CarouselChildPropTypes = CarouselChildPropType;
+
+let CarouselControlsPropType;
+if (process.env.NODE_ENV !== 'production') {
+  CarouselControlsPropType = {
+    current: PropTypes.number.isRequired,
+    controls: PropTypes.oneOfType([PropTypes.string, PropTypes.bool])
```
halocline, line 33: any string or enumerated?

src/js/components/Carousel/propTypes.js
halocline: If we revert to 
```
controls: PropTypes.oneOfType([
      PropTypes.bool,
      PropTypes.oneOf(['arrows', 'selectors']),
    ]),
```

then I don't think we need to touch `index.d.ts` -- otherwise we'll want these types to be in sync.

src/js/components/Carousel/__tests__/Carousel-test.tsx
```diff
@@ -1,139 +1,157 @@
...
-      .getElementById('test-carousel')!
-      .querySelector('button');
-
-    expect(button).not.toBeNull();
-
-    fireEvent.click(button!);
-    expect(setActiveSlide).toBeCalledWith(0);
+    expect(container.firstChild).toMatchSnapshot();
+    expect(setActiveSlide).toHaveBeenCalledWith(1);
   });
```
halocline, line 156: Should we have a test for interactive/clickable child?

src/js/components/Carousel/stories/InteractiveChild.js
```diff
@@ -0,0 +1,55 @@
...
+        fill
+        align="center"
+        justify="center"
+        background="lavender"
+        gap="small"
+      >
+        <Text weight="bold" size="xlarge">
+          Slide 1
+        </Text>
+        <Button label="Button" />
```
halocline, line 18: Would an onClick with an alert or console log be valuable?

src/js/components/Carousel/stories/Custom.js
```diff
@@ -1,44 +0,0 @@
-import React from 'react';
-import { Attraction, Car, TreeOption } from 'grommet-icons';
-
-import { Grommet, Box, Carousel } from 'grommet';
-
-const customTheme = {
```
halocline, line 6: Should we have a custom theme directory and story? Something that swaps out icons, icon color, and animation duration?

```
carousel: {
      icons: {
        current: Subtract,
        next: Next,
        previous: Previous,
        // color: { dark: undefined, light: undefined },
      },
      animation: {
        duration: 1000,
      },
      disabled: {
        icons: {
          // color: { dark: undefined, light: undefined },
        },
      },
    },
```

src/js/components/Carousel/CarouselControls.js
```diff
@@ -0,0 +1,89 @@
...
+  const selectorColor = normalizeColor(
+    theme.carousel.icons.color || 'control',
+    theme,
+  );
+
+  return (
+    <>
+      {/* Previous Arrow */}
+      {arrowNavigation && (
+        <StyledControl offset="left" fill="vertical">
```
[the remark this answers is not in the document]
kevingill55: Good catch

src/js/components/Carousel/__tests__/Carousel-test.tsx
```diff
@@ -1,139 +1,157 @@
...
-import { Image } from '../../Image';
+
+const getSlideOne = () => screen.getByText('Slide One');
+const getSlideTwo = () => screen.getByText('Slide Two');
+const getSlideThree = () => screen.getByText('Slide Three');
 
 describe('Carousel', () => {
   test('basic', () => {
     const { container } = render(
-      <Grommet>
```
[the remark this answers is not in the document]
kevingill55: I see I can add it for consistency reasons. I'm not sure what it's adding other than applying styles I suppose but that shouldn't change the functionality of the component itself - which is what I'm expecting we want to be testing.

I'll add it back in but curious to the approach in general, maybe I'm missing some other context it provides.

src/js/components/Carousel/propTypes.js
kevingill55: Right, I can add those values back in.

src/js/components/Carousel/propTypes.js
kevingill55: The latter 😄

src/js/components/Carousel/stories/InteractiveChild.js
kevingill55: It could be, I think the fact you can clearly hover and click the button should be good. But, depending on the convention of the Grommet storybook I am open to adding whatever works best.

src/js/components/Carousel/stories/Custom.js
halocline: Write as typescript

src/js/components/Carousel/Carousel.js
```diff
@@ -1,242 +1,222 @@
...
-          focus={focus}
-          onFocus={(event) => {
-            setFocus(true);
-            if (onFocus) onFocus(event);
+    <Keyboard
+      onLeft={() => onPrevious(current, inTransition)}
+      onRight={() => onNext(current, inTransition)}
+    >
+      <StyledCarouselContainer {...containerProps}>
+        <div
```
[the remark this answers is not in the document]
kevingill55: I'll follow up if the StyledCarouselContainer is even necessary or if we can just use the native height and width of the `Box` component

src/js/components/Carousel/Carousel.js
kevingill55: Using animation duration from theme

src/js/components/Carousel/Carousel.js
kevingill55: Rewriting this useEffect to adhere to lint rule 👍

src/js/components/Carousel/Carousel.js
kevingill55: Same comment above about styled component

src/js/components/Carousel/CarouselChild.js
kevingill55: Using animation duration from theme

src/js/components/Carousel/CarouselChild.js
kevingill55: Find a better name for the `animationTimer` variable (maybe)

src/js/components/Carousel/CarouselChild.js
kevingill55: Add "shortcut" comment, remove in depth explanation

halocline (commented):
Last couple items that I see:
- [x] Add to all stories.
```
  // <Grommet theme={grommet}>
  ...
  // </Grommet>
  ```
- [x] Add to typescript storty
```
```
- [x] `index.d.ts` needs the `continuous` prop added

**Review by halocline — APPROVED** [review]
Ready for additional 👀 !!!

File: src/js/components/Carousel/Carousel.js [hunk 20]
```diff
@@ -1,242 +1,260 @@
...
+import { CarouselPropTypes } from './propTypes';
+import { Keyboard } from '../Keyboard';
 import { ThemeContext } from '../../contexts';
+import { defaultProps } from '../../default-props';
 
-import { Box } from '../Box';
-import { Button } from '../Button';
-import { Keyboard } from '../Keyboard';
-import { Stack } from '../Stack';
+const handleSelectorNavigation =
```
ericsoderberghp, line 19: What's the rationale for moving these functions outside of the component? It feels to me like there is almost twice as much code with the current approach, given the number of arguments being called with.

ericsoderberghp (commented):
Looking good. Just one minor comment.

src/js/components/Carousel/Carousel.js
kevingill55: These are just closures to try to DRY things up. I think this makes the code cleaner rather than just inlining these fairly complex functions. Ideally they could live in a utility file and just be imported into the functional component but I don't see that pattern used anywhere else - so I decided to just leave them above.

I think closures help with the readability and performance of these functions with so many arguments, but I'd be interested to see if you had other ideas. The only "extra" code it adds (from what I can see) is just instantiating the first part of the closure in the component. I'm open to exploring other ideas as well so if you have any recommendations please let me know!

src/js/components/Carousel/propTypes.js
```diff
@@ -1,31 +1,50 @@
...
+    fill: PropTypes.bool,
+    continuous: PropTypes.bool,
     controls: PropTypes.oneOfType([
-      PropTypes.bool,
       PropTypes.oneOf(['arrows', 'selectors']),
+      PropTypes.bool,
     ]),
+    onChild: PropTypes.func,
+    play: PropTypes.number,
   };
```
jcfilben, line 18: Should `width` and `height` props be included here?

src/js/components/Carousel/stories/Multiple.js
```diff
@@ -46,13 +46,14 @@ const View2 = () => {
...
-  <Grommet>
-    <Carousel>
-      <View0 />
-      <View1 />
-      <View2 />
-    </Carousel>
-  </Grommet>
+  // Uncomment <Grommet> lines when using outside of storybook
+  // <Grommet theme={grommet}>
+  <Carousel controls="arrows" showProgress>
```
jcfilben, line 51: What is the `showProgress` prop?

src/js/components/Carousel/stories/ControlVariations.js
```diff
@@ -0,0 +1,105 @@
...
+import { Attraction, Car, TreeOption, Cloud, Amazon } from 'grommet-icons';
+import { Box, Text } from 'grommet';
+import { Carousel } from '../Carousel';
+
+export const ControlVariations = () => (
+  // Uncomment <Grommet> lines when using outside of storybook
+  // <Grommet theme={grommet}>
+  <Box direction="row" align="center" justify="center" fill="horizontal">
+    <Box
+      height="500px"
```
jcfilben, line 11: Can we use t-shirt sizing here?

src/js/components/Carousel/__tests__/Carousel-test.tsx
```diff
@@ -1,139 +1,249 @@
...
+    render(
+      <Grommet>
+        <Carousel initialChild={1} a11yTitle="test-carousel">
+          <div>Slide One</div>
+          <div>Slide Two</div>
+        </Carousel>
+      </Grommet>,
+    );
+    const carousel = screen.getByLabelText('test-carousel');
+    fireEvent.keyDown(carousel, {
```
jcfilben, line 175: Can we use `userEvent` here instead of `fireEvent`?

jcfilben (commented):
Issues with responsive behavior when the Carousel doesn't have a set width/height

https://user-images.githubusercontent.com/54560994/141532455-4c2ad98c-ad47-499d-b93b-315a4c489046.mp4

src/js/components/Carousel/__tests__/Carousel-test.tsx
kevingill55: This is the one case where you wanna use `fireEvent` instead of `userEvent` https://testing-library.com/docs/guide-events/#alternatives

src/js/components/Carousel/stories/Multiple.js
kevingill55: Left over prop, good catch!

src/js/components/Carousel/Carousel.js
ericsoderberghp: So, if I look at `handleOnNext`:
- the actual code is lines 61-69 (9 lines)
- the function parameters are defined in lines 53-60 (8 lines)
- calling this function to set up `onNext` is done in lines 138-145 (8 lines)
If instead we just put the actual code directly where `onNext` is defined at line 138, we'd still only have a single copy of the code (DRY) but we be able to omit the two 8 line parameter definition blocks. So, less code.
Furthermore, we should probably be using `useCallback` to ensure we don't re-define the function on each render.

kevingill55: Makes sense! Addressed this

src/js/components/Carousel/Carousel.js
```diff
@@ -1,242 +1,213 @@
...
 }) => {
+  const containerRef = React.createRef();
   const theme = useContext(ThemeContext) || defaultProps.theme;
-  const [focus, setFocus] = useState();
-  const timerRef = useRef();
-
-  const [indexes, setIndexes] = useState({
-    activeIndex: activeChild !== undefined ? activeChild : initialChild,
+  const numSlides = children.length;
+  const noContainer = !fill && (!height || !width);
```
ericsoderberghp, line 32: `noContainer` doesn't seem quite accurate as a name. Don't we always include `StyledCarouselContainer`?

src/js/components/Carousel/Carousel.js
```diff
@@ -1,242 +1,213 @@
...
+  const animationDuration =
```
ericsoderberghp, line 33: Since this is included as a dependency for effects, I think we should `useMemo()` to avoid re-running the effects when not needed.

src/js/components/Carousel/Carousel.js
```diff
@@ -1,242 +1,213 @@
...
+      ? play
+      : theme.carousel.animation.duration;
+
+  const [current, setCurrent] = useState(activeChild || initialChild);
+  const [previous, setPrevious] = useState(undefined);
+  const [activeChildState, setActiveChildState] = useState(activeChild);
+  const [direction, setDirection] = useState(undefined);
+  const [inTransition, setInTransition] = useState(false);
+  const [containerProps, setContainerProps] = useState({
+    containerHeight: undefined,
```
ericsoderberghp, line 44: Can these just be called `height` and `width`? Since they are already in `containerProps`, they already have a context for the "container".

src/js/components/Carousel/StyledCarousel.js
```diff
@@ -0,0 +1,82 @@
...
+
+const StyledCarouselContainer = styled(Box)`
+  position: relative;
+  display: block;
+  overflow: hidden;
+  height: 100%;
+  ${(props) => props.height && heightStyle(props.height, props.theme)}
+  ${(props) => props.width && widthStyle(props.width, props.theme)}
+  ${(props) =>
+    props.inTransition &&
```
ericsoderberghp, line 13: Since `containerHeight` and `containerWidth` are already undefined when not `inTransition`, do we still need to test for `props.inTransition` here?

src/js/components/Carousel/index.d.ts
```diff
@@ -3,19 +3,24 @@ import {
...
+  HeightType,
   MarginType,
+  WidthType,
 } from '../../utils';
 
 export interface CarouselProps {
   a11yTitle?: A11yTitleType;
   activeChild?: number;
   alignSelf?: AlignSelfType;
+  continuous?: boolean;
```
ericsoderberghp, line 15: Can we rename this property to be "wrap"? I think that conveys the behavior a bit better than "continuous". I think "continuous" feels more like it's associated with "play", which isn't as true here.

src/js/components/Carousel/Carousel.js
kevingill55: These props are only meant to be processed in the styled component. Since it is a styled `Box`, if we name these `height` and `width` then the values will be overwritten.

src/js/components/Carousel/Carousel.js
kevingill55: This just indicates that no container / dimensions have been specified and therefore the Carousel will expand its container to the size of the child. I can rename to `noDimensions` maybe?

src/js/components/Carousel/Carousel.js
ericsoderberghp: I think Box can handle us passing an explicit height and width already.

src/js/components/Carousel/CarouselControls.js
```diff
@@ -0,0 +1,115 @@
...
+  return (
+    <>
+      {/* Previous Arrow */}
+      {arrowNavigation && (
+        <StyledControl offsetProp="left" fill="vertical">
+          <Button
+            plain
+            hoverIndicator
+            fill="vertical"
+            a11yTitle={`Go to slide ${current}`}
```
MikeKingdom, line 43: This should go in languages/default.json and you should use MessageContext.js to get the message so it can be localized.

src/js/components/Carousel/CarouselControls.js
```diff
@@ -0,0 +1,115 @@
...
+      )}
+
+      {/* Next Arrow */}
+      {arrowNavigation && (
+        <StyledControl offsetProp="right" fill="vertical">
+          <Button
+            plain
+            hoverIndicator
+            fill="vertical"
+            a11yTitle={`Go to slide ${current + 2}`}
```
MikeKingdom, line 67: This should go in languages/default.json and you should use MessageContext.js to get the message so it can be localized.

src/js/components/Carousel/CarouselControls.js
```diff
@@ -0,0 +1,115 @@
...
+          fill="horizontal"
+          gap="xsmall"
+          pad={{ vertical: 'xsmall' }}
+        >
+          {Array.from(Array(numSlides).keys()).map((_, index) => (
+            <Button
+              plain
+              hoverIndicator
+              key={`control-${index + 1}`}
+              a11yTitle={`Jump to slide ${index + 1}`}
```
MikeKingdom, line 98: This should go in languages/default.json and you should use MessageContext.js to get the message so it can be localized.

src/js/components/Carousel/stories/Autoplay.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { Attraction, Car, TreeOption } from 'grommet-icons';
+
+import { Text, Box } from 'grommet';
+import { Carousel } from '../Carousel';
+
+export const Autoplay = () => (
```
MikeKingdom, line 7: I get a flash every so often at the end of a transition where it renders the child off to the left outside the container then immediately re-renders it correctly. Browser is chrome on windows.

src/js/components/Carousel/stories/Multiple.js
```diff
@@ -46,13 +46,14 @@ const View2 = () => {
...
-  <Grommet>
-    <Carousel>
-      <View0 />
-      <View1 />
-      <View2 />
-    </Carousel>
-  </Grommet>
+  // Uncomment <Grommet> lines when using outside of storybook
+  // <Grommet theme={grommet}>
+  <Carousel controls="arrows">
```
MikeKingdom, line 51: The sizing is pretty odd for t he middle items. It scrolls in cut-off then jumps in height when the transition is done.

src/js/components/Carousel/Carousel.js
```diff
@@ -1,242 +1,218 @@
...
+      } else {
+        setContainerProps({
+          containerHeight: undefined,
+          containerWidth: undefined,
+        });
+      }
     }
-    return () => {};
-  }, [activeIndex, play, children, lastIndex, onChildChange]);
+  }, [inTransition, noDimensions]);
```
MikeKingdom, line 123: This line complains about missing dependency on containerRef

src/js/components/Carousel/Carousel.js
```diff
@@ -1,242 +1,218 @@
...
+  // Handles auto-playing Carousel slides
+  useEffect(() => {
+    let playTimer;
+    if (play) {
+      playTimer = setInterval(() => {
+        onNext();
+      }, play);
+    }
+    return () => clearTimeout(playTimer);
+  }, [current, inTransition, play, onNext]);
```
MikeKingdom, line 149: don't need current or inTransition here although you're basically getting that since onNext's useCallback is dependent on both. The result is this useEffect destroys and re-creates the timer each transition.

src/js/components/Carousel/Carousel.js
```diff
@@ -1,242 +1,218 @@
...
-      timerRef.current = timer;
-
-      return () => {
-        clearTimeout(timer);
-      };
+      } else {
+        setContainerProps({
+          containerHeight: undefined,
+          containerWidth: undefined,
+        });
```
MikeKingdom, line 120: This is causing the flash I see in chrome on windows at the end of a transition.

It looks visibly like the container ends up bigger briefly and the child is drawn much farther left briefly and then the container goes back to the expected size and the child ends up back where it was before.

Is this really needed?

MikeKingdom:
Replaced by #5839

PR #5762, opened 2021-10-28
Fix Card to be backwards compatible

taysea:
#### What does this PR do?
In a [recent PR](https://github.com/grommet/grommet/pull/5736), we introduced changes that were not backwards compatible. The issue was rooted in a side-effect of styled components `forwardedAs` prop and how it handles creating classes for `styled` HOCs.

Essentially, by abstracting Card to go from being just a Box to a `styled(Box)`, we introduce another layer of styled components which resulted in the Box props applied to a Card getting lost. Styled components isn't set up to know how to merge/layer so many levels of chained styled components. Suddenly when we updated the DS site, everything looked like this (yikes):

Screen Shot 2021-10-28 at 3 26 58 PM

Therefore, this PR takes a new approach where the desired styling for Card is passed as `kind` on Box. Similar to what we do with Button. `kind` takes a structure of `{ hover: theme.card.hover.container, ...theme.card.container}`. Then, StyledBox handles the `kind` styles.

src/js/components/Card/Card.js

Tested on Clickable story.

Also, I mirrored what the setup in the DS site was (which caused backwards compatibility issues) and confirmed this would work correctly:

```
// StyledConentCard (as it existed in DS)
const StyledNewCard = styled(Card)`
  &:hover {
    background: lavender;
  }
  transition: all 0.2s ease-in-out;
`;

const theme = deepMerge(grommet, {
  card: {
    hover: {
      container: {
        elevation: 'large',
      },
    },
  },
});
export const Simple = () => (
  <Grommet theme={theme}>
    <Box pad="medium" gap="medium">
      <Link href="/" passHref>
        <StyledNewCard
          width="medium"
          forwardedAs="a"
          background="gold"
          pad="25px"
          style={{ textDecoration: 'none' }}
        >
          This is equivalent to what we have as ContentCard in Design System.
        </StyledNewCard>
      </Link>
    </Box>
  </Grommet>
);
export default {
  title: 'Layout/Card/Simple',
};
```

https://user-images.githubusercontent.com/12522275/139345446-d00eb5b5-2d3c-4f18-a4e4-9bdf7e7a71b5.mov

Created a styled component that is a `styled(Card)` and pass Box props on the card. All of those styles should appear.

Yes, need to move the extend outside of the hover block.

Yes.

Backwards compatible.


src/js/components/Box/StyledBox.js
```diff
@@ -194,6 +194,11 @@ const interactiveStyle = css`
     ${(props) =>
       props.hoverIndicator &&
       getHoverIndicatorStyle(props.hoverIndicator, props.theme)}
+    ${(props) =>
+      props.context === 'card' &&
+      props.onClick &&
+      props.theme.card.hover?.container?.elevation &&
```
ericsoderberghp, line 200: This feels like Box knows too much about Card. Would it be possible to have Card pass `hoverIndicator` as a property to Box with it pulling the value from `theme.card.hover?`?

taysea: This feels like a fair approach. Let me try that out.

src/js/components/Box/StyledBox.js
```diff
@@ -191,14 +191,22 @@ const interactiveStyle = css`
...
+      getHoverIndicatorStyle(props.kind.hover, props.theme)}
     ${(props) =>
       props.hoverIndicator &&
       getHoverIndicatorStyle(props.hoverIndicator, props.theme)}
   }
 `;
 
 // NOTE: basis must be after flex! Otherwise, flex overrides basis
-const StyledBox = styled.div`
+const StyledBox = styled.div.withConfig({
```
ericsoderberghp, line 204: Can we just rename `kind` to `kindProp` to prevent it leaking? It feels like we do this elsewhere.

taysea: I will give this a try

ericsoderberghp (approved):

PR #5773, opened 2021-11-01
Add support to gradient on Meter component

leossantiago:
#### What does this PR do?

As the title says, introduces gradients as color options to both Circle and Bar Meters. This was achieved by mapping an array of objects containing values and colors, similar to how the `<Chart />` component handles that. This proposal is a product of the discussion started in #5020. Following some offline discussions with @taysea, we decided to open a new PR as the solution completely deviates from the original proposal. 

For a single value, the gradient can be set using the `color` property. For multiple values, the `values` property should be used, each having its color (solid or gradient). Gradient arrays can also receive a single object containing linearGradient-related props.

```js
        color={[
          {
            gradientUnits: 'userSpaceOnUse',
            gradientTransform: 'rotate(90)',
          },
          { value: 0, color: 'purple' },
          { value: 50, color: 'white' },
          { value: 100, color: 'blue' },
        ]}
```

`src/js/components/Meter/Bar.js` and `src/js/components/Meter/Circle.js`

Manual testing and storybook.

`src/js/components/Meter/stories/Gradient.js`

image

Yes

Backwards compatible.


src/js/components/Meter/Circle.js
```diff
@@ -48,10 +54,46 @@ const Circle = forwardRef((props, ref) => {
...
-      const stroke = strokeProps(
+
+      let stroke = strokeProps(
         someHighlight && !highlight ? background : colorName,
         theme,
       );
+
+      const useGradient = color && Array.isArray(color);
+
+      if (useGradient) {
```
jcfilben, line 65:
```suggestion
      if (color && Array.isArray(color)) {
```
Since we only use this variable one I think we don't have to explicitly define it

ericsoderberghp:
For the circle case, I would expect the gradient to follow the stroke path, as opposed to being more of an overall background mask. Here are some examples: https://gist.github.com/mbostock/4163057

src/js/components/Meter/Bar.js
```diff
@@ -60,15 +62,49 @@ const Bar = forwardRef((props, ref) => {
...
         }
 
+        let stroke = strokeProps(
+          someHighlight && !highlight ? background : colorName,
+          theme,
+        );
+
+        const useGradient = color && Array.isArray(color);
+
+        if (useGradient) {
```
jcfilben, line 72:
```suggestion
        if (color && Array.isArray(color)) {
```
Since we only use this variable one I think we don't have to explicitly define it

jcfilben (commented):
Can we add a jest test for gradient with circular meter and bar meter?

leossantiago:
@ericsoderberghp should we give users the ability to choose between both cases? I do like what you suggested and will start working on it, but I'm wondering if we should also support it as if it was a whole circle.

---

@jcfilben we definitely can! I Will work on that. 


Thanks you two for the suggestions 😄

jcfilben:
Hi @leossantiago I'm wondering what the use case would be for the 'background mask' version of the meter gradient where it shows the full gradient within the colored portion of the meter.

For the stroke version (where only part of the gradient is show based on how full the meter is) it makes sense that the colors are helping represent the value of the meter. For example if you have a Green, Yellow, Red gradient meter that is representing the fire danger for the day. If the meter is 20% full and the fire danger is relatively low the meter will only show the green color. In this case it would be confusing to show the full gradient when the value is at 20%. I'm just wondering how you see the 'background mask' version of the meter gradient being used.

leossantiago:
Hey @jcfilben ! Honestly, I didn't have a proper use case. Just thought that since gradients have the property `gradientUnits` which handles that, it could be nice to give users the control over that (gradient filling the full length would be the default behavior). 

We can drop that, and make it a bit cleaner if needed.

By the way, loved the example you came up with 😄 

I'm working on Eric's request. Hopefully we'll be able to test this out soon.

Let me know if you have any suggestions on SVG properties support.

jcfilben:
Yeah I am thinking since we don't have a need for the case where the full gradient takes up whatever the value of the meter is, let's just do the gradient filling the full length to keep things simple for now.

leossantiago:
Aye. Sounds good. Do you think it's worth turning `gradientTransform` into a prop of the Meter? Perhaps we could consider only rotations. That would remove the need of passing a properties object inside the colors array.

jcfilben:
I think if we give people the option to transform the gradient it shouldn't be a prop on Meter and instead should just be contained within the color array. My thinking is that the 80% use case is for people to use Meter without a gradient. If they are using a gradient they are most likely want the gradient to go left to right on a Bar meter and for a circle have the gradient start at the top and work its way around the circle clockwise. This should be the default behavior. My thought is that it won't be a super common use case for people to want to use a gradient and then transform the gradient to deviate from the default, so there shouldn't be a separate prop for this. Let me know if that sounds okay or if you think otherwise.

jcfilben:
Closing due to inactivity, feel free to re-open

PR #5778, opened 2021-11-02
Initial tags implementation

### LINKED ISSUE #5777: Tags - initial implementation
Implementation of [#1919](https://github.com/grommet/hpe-design-system/issues/1919)

MikeKingdom:
#### What does this PR do?
Implements a Tag component.

[Figma](https://www.figma.com/file/8zrInotU4JK9LWP4WyZy25/HPE-Design-System-Tags?node-id=0%3A1)

Tag.js

Stprybook

Storybook Tag/Simple
Fixes #5777 

image

Yes

Yes

Backwards compatible


src/js/components/Tag/propTypes.js
```diff
@@ -0,0 +1,20 @@
...
+  genericProps,
+} from '../../utils/general-prop-types';
+
+let PropType = {};
+if (process.env.NODE_ENV !== 'production') {
+  PropType = {
+    ...genericProps,
+    as: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
+    background: backgroundPropType,
+    name: PropTypes.string,
```
ericsoderberghp, line 13: Doesn't match typescript here. I'm wondering if we should start with just string for name and string | number for value in both cases?

src/js/components/Tag/propTypes.js
```diff
@@ -0,0 +1,20 @@
...
+  backgroundPropType,
+  genericProps,
+} from '../../utils/general-prop-types';
+
+let PropType = {};
+if (process.env.NODE_ENV !== 'production') {
+  PropType = {
+    ...genericProps,
+    as: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
+    background: backgroundPropType,
```
ericsoderberghp, line 12: missing border shown in typescript, but, I'm wondering if we should omit this for the initial version?

src/js/components/Tag/propTypes.js
```diff
@@ -0,0 +1,20 @@
...
+    name: PropTypes.string,
+    value: PropTypes.string,
```
ericsoderberghp, line 14: value should be required

src/js/components/Tag/stories/Simple.js
```diff
@@ -0,0 +1,78 @@
...
+        direction="row"
+        align="start"
+      >
+        <Box gap="medium" align="start">
+          <Tag
+            name="name that is much longer and longer and longer"
+            value="value"
+            onRemove={onRemove}
+          />
+          <Tag name="name" onRemove={onRemove} />
```
ericsoderberghp, line 27: I don't think it should be legal to not have a value.

src/js/themes/base.js
```diff
@@ -1436,6 +1436,61 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
+        border: true,
+        direction: 'row',
+        round: 'large',
+        pad: {
+          horizontal: `${baseSpacing / 4 - borderWidth}px`,
+        },
+      },
+      gap: 'xsmall',
+      name: { },
+      padding: {
```
ericsoderberghp, line 1482: What drives putting something inside `tag.container` vs. just at `tag`?

ericsoderberghp (commented):
Some initial comments to get started with.

src/js/themes/base.js
MikeKingdom: > What drives putting something inside `tag.container` vs. just at `tag`?

`tag.container` is applied to the very outside component which is either a Box or a Button depending on if its onClick-able or not. It has the border styling/shape and some basic padding.

The rest of the stuff in tag ends up on the `name:value` content and remove button.

Any thoughts on how to structure this?

src/js/themes/base.js
```diff
@@ -1462,6 +1462,61 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         // extend: undefined,
       },
     },
+    tag: {
```
britt6612, line 1465: overall looks good can we also add types to the `base.d.ts` file

src/js/themes/base.js
```diff
@@ -1462,6 +1462,61 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
+            radius: `${baseSpacing}px`, // 24px
+          },
+          pad: {
+            vertical: `${baseSpacing / 4 + borderWidth}px`, // 8px
+            horizontal: `${baseSpacing + borderWidth * 4}px`, // 32px,
+          },
+        }, 
+        */
+      },
+      container: {
```
britt6612, line 1497: maybe add a comment for users to see it takes `boxProps`

src/js/components/Tag/index.d.ts
```diff
@@ -0,0 +1,40 @@
...
+} from '../../utils';
+
+export interface TagProps {
+  a11yTitle?: A11yTitleType;
+  alignSelf?: AlignSelfType;
+  gridArea?: GridAreaType;
+  margin?: MarginType;
+  background?: BackgroundType;
+  // focusIndicator?: boolean;
+  // hoverIndicator?: BackgroundType | boolean;
```
britt6612, line 19: are you keeping these?

src/js/components/Tag/__tests__/Tag-test.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  test('onClick', () => {
+    const onClick = jest.fn();
+    const {container } = render(
+      <Grommet>
+        <Tag value="Test" onClick={onClick} />
+      </Grommet>,
+    );
+
+    fireEvent.click(screen.getByRole('button'));
```
jcfilben, line 57: Can we use userEvent instead of fireEvent here?

jcfilben (commented):
Functionality looks good!
I just added some minor comments around testing and storybook

src/js/themes/base.js
ericsoderberghp: How is `theme.tag.container.pad` different from `theme.tag.padding`? I can't tell by looking at the theme. I'm wondering if we can simplify it to:
```
tag: {
  name: { },
  value: { },
  gap:
  separator:
  pad:
  round:
  background:
  border:
},
```
or whether we should go towards:
```
tag: {
  name: { },
  value: { },
  container: {
    gap:
    separator:
    pad:
    round:
    background:
    border:
  },
},
```

MikeKingdom:
When the remove button is there, it needs a smaller amount of padding between the right side of it and the border. The name/value piece has a separate padding that is always around it, whether the button is there or not. What would be the best way to accomplish that? It seemed less generic to piecemeal each of the attributes we currently use and just let it use whatever was defined per grouping.

What do you think is the easiest way to split it up?
________________________________

halocline:
> When the remove button is there, it needs a smaller amount of padding between the right side of it and the border. The name/value piece has a separate padding that is always around it, whether the button is there or not. What would be the best way to accomplish that? It seemed less generic to piecemeal each of the attributes we currently use and just let it use whatever was defined per grouping. What do you think is the easiest way to split it up?

This might be another instance where we treat 'remove' as a state of the tag and adjust styling based on that state. When in remove state, the container (padding in this case) can be adjusted as desired.

I can see the tag being the 'container' itself or being more explicit with `tag.container`. No strong opinion here.

```
tag: {
    name: { },
    value: { },
    gap:
    separator:
    pad: '6px 10px',
    round:
    background:
    border:
    remove: {
        container: {
            pad: 6px 8px 6px 10px,
        },
        round:     
     },
},
```

src/js/themes/base.js
MikeKingdom: Good point. Thank!

MikeKingdom:
I ended up simplifying the  theme pieces to basically Eric's first suggestion with `remove`. So it now looks like:
```jsx
tag: {
  background: undefined,
  border: true,
  round: 'large',
  name: { },
  pad: {  
    horizontal: 'small',
    vertical: 'xsmall',
  },
  remove: {
    margin: { right: 'xsmall'},
  },
  separator: ' : ',
  value: {
    weight: 600,
  },
},
```
Part of these changes adjusted how the `name : value` was structured so it behaves better when things are too large.

src/js/themes/base.js
```diff
@@ -1462,6 +1462,23 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         // extend: undefined,
       },
     },
+    tag: {
+      // background: undefined,
+      border: true,
+      round: 'large',
+      name: { },
```
taysea, line 1469:
```suggestion
      // name: undefined,
```
Should this be undefined if we don't have anything default in the theme?

MikeKingdom: Whatever folks prefer. I put it as an empty object to make it clear it was like the value one but I suppose it would match some other examples.

ericsoderberghp:
I think we need to change package.json to increase bundlesize.maxSize to be "122 kB"

src/js/components/Tag/index.d.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export interface TagProps {
+  a11yTitle?: A11yTitleType;
+  alignSelf?: AlignSelfType;
+  gridArea?: GridAreaType;
+  name?: string;
+  value: string | number;
+  onClick?: (...args: any[]) => any;
+  onRemove?: (...args: any[]) => any;
+  size?: 'small' | 'medium' | 'large';
```
ericsoderberghp, line 20: I don't think we are handling `size` correctly. I think it ends up in `...rest` which is correctly propagated to the StyledTagButton but then incorrectly to the container Box in the onRemove case. I'm actually inclined to remove `size` for this initial version. We can add it later.

MikeKingdom: Definitely wasn't going to expose size yet so I think I just need to remove it from index.d.ts

src/js/components/Tag/Tag.js
```diff
@@ -0,0 +1,75 @@
...
+      <Box width={{ min: 'min-content' }} pad={theme.tag.pad}>
+        <Text>
+          {name && <Text {...theme.tag.name}>{name}</Text>}
+          {name && value ? theme.tag.separator : '' }
+          {value && <Text {...theme.tag.value}>{value}</Text>}
+        </Text>
+      </Box>
+    );
+
+    return onRemove || !onClick ? (
```
taysea, line 37: If I'm reading this correctly, this allows Tag to have `onRemove` and `onClick` which would create a nested interactive element. Do we want to allow this?

src/js/components/Tag/__tests__/Tag-test.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { render, screen } from '@testing-library/react';
+import { axe } from 'jest-axe';
+import userEvent from '@testing-library/user-event';
+
+import 'jest-styled-components';
+import 'jest-axe/extend-expect';
+import 'regenerator-runtime/runtime';
+import '@testing-library/jest-dom';
+
+// import { Add, Next } from 'grommet-icons';
```
taysea, line 11:
```suggestion
```
Should we remove if not used?

src/js/themes/base.d.ts
```diff
@@ -1352,6 +1352,15 @@ export interface ThemeType {
       };
     };
   };
+  tag?: {
+    border?: BorderType;
```
taysea, line 1356: missing `background` in this

taysea (commented):
Looking good! Just a few small comments

src/js/components/Tag/index.d.ts
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import {
+  A11yTitleType,
+  AlignSelfType,
+  BackgroundType,
+  GridAreaType,
+  MarginType,
+  // Omit,
```
britt6612, line 8: can we remove?

src/js/components/Tag/Tag.js
MikeKingdom: We don't allow both. onClick will be ignored if onRemove is specified.

onClick is only used at line 61. If they specify onRemove it will do the static label plus a remove button. Same thing if neither onClick or onRemove are specified.

If only onClick is specified it will render a StyledTagButton with the onClick callback included.

I've written the component doc for grommet-site to say you can't use both. This may deserve a console.warn similar to what we do in a couple other components though

src/js/components/Tag/__tests__/Tag-test.tsx
MikeKingdom: Definitely. Thanks

src/js/themes/base.d.ts
MikeKingdom: Good catch

src/js/components/Tag/index.d.ts
MikeKingdom: Done

src/js/components/Tag/index.d.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export interface TagProps {
+  a11yTitle?: A11yTitleType;
+  alignSelf?: AlignSelfType;
+  gridArea?: GridAreaType;
+  name?: string;
+  value: string | number;
+  onClick?: (...args: any[]) => any;
+  onRemove?: (...args: any[]) => any;
+  as?: PolymorphicType;
```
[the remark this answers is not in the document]
MikeKingdom: Done

taysea (approved):
Nice job!

src/js/components/Tag/__tests__/Tag-test.tsx
```diff
@@ -0,0 +1,75 @@
...
+  });
+
+  test('basic', () => {
+    const { container } = render(
+      <Grommet>
+        <Tag name="Name" value="Value" />
+      </Grommet>,
+    );
+
+    expect(container.firstChild).toMatchSnapshot();
```
jcfilben, line 45:
```suggestion
    expect(asFragment()).toMatchSnapshot();
```
Can we switch the snapshots to use asFragment() instead of container to follow the testing best practices guidance?

jcfilben (commented):
Just a minor testing comment, otherwise looks good

ericsoderberghp (approved):

Issue #5672, opened 2021-10-07
Background is not applied to pinned header in DataTable


BittencourtMH:
### Expected Behavior

The value of the `background` property should be applied to all cells of the table.

### Actual Behavior

When the header is pinned, the value of the `background` property is not applied to the header.

### URL, screen shot, or Codepen exhibiting the issue

![Screen Shot 2021-10-07 at 16 26 44](https://user-images.githubusercontent.com/28353395/136449673-d8c5cf87-35a1-40e2-9acf-6200fa16fc87.png)

### Steps to Reproduce

The code below is based on the example given on the documentation page:

```
<DataTable
  background="light-6"
  columns={[
  {
    property: 'name',
    header: <Text>Name</Text>,
    primary: true,
  },
  {
    property: 'percent',
    header: 'Complete',
    render: datum => (
      <Box pad={{ vertical: 'xsmall' }}>
        <Meter
          values={[{ value: datum.percent }]}
          thickness="small"
          size="small"
        />
      </Box>
    ),
  },
  ]}
  data={[
    { name: 'Alan', percent: 20 },
    { name: 'Bryan', percent: 30 },
    { name: 'Chris', percent: 40 },
    { name: 'Eric', percent: 80 },
  ]}
  pin
/>
```

### Your Environment

- Grommet version: 2.18.0
- Browser Name and version: Firefox 93.0
- Operating System and version (desktop or mobile): Ubuntu 18.04.6 LTS

PR #5783, opened 2021-11-04
Fix: Background not applied to pinned header

jlaluces123:
This PR aims to fix the issue where the `background` prop when `pin` is enabled does not work as intended (not correctly changing the background color as expected).

#### What does this PR do?
This PR:
- Allows background color when `pin` is active to actually change the background color accordingly
- Introduces hierarchy rules under the hood for picking the correct background value from prop or theme

Hierarchy rules:
- `background="onlyString"` overrides every background styling
- `background={{ header: "someColor", footer: "someColor" }}` overrides what's in theme
- `theme.dataTable.pinned.[header | footer].background` overrules `theme.dataTable.[header | footer].background`
- `theme.dataTable.[header | footer].background` has least priority / last one in the list  

- `normalizeCellProps()` in `DataTable/buildState.js`

- Manual testing going through the combinations listed in the hierarchy rules above

- Make sure `pin` prop is enabled (used the `DataTable/Fill and Pin story`)
- Try any of the combinations in the hierarchy list and view what the code chooses

Solves #5672 
No
Yes, involves how background color is determined in terms of prop values vs. theme values

Backwards compatible

### DIFF
--- a/src/js/components/DataTable/buildState.js
+++ b/src/js/components/DataTable/buildState.js
@@ -248,10 +248,12 @@ export const normalizeCellProps = (props, theme) => {
         props?.[propName]?.[context] ||
         // if the propName is used without context, it applies to all contexts
         (tableContextNames.every((n) => !props?.[propName]?.[n]) &&
-          props?.[propName]) ||
+          props?.[propName]);
+      if (value !== undefined) result[context][propName] = value;
+
+      let themeValue =
         theme?.dataTable?.[context]?.[propName] ||
         theme?.table?.[context]?.[propName];
-      if (value !== undefined) result[context][propName] = value;
 
       // pinned case
       value =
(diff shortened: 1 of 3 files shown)


src/js/components/DataTable/buildState.js
```diff
@@ -248,38 +248,43 @@ export const normalizeCellProps = (props, theme) => {
         props?.[propName]?.[context] ||
         // if the propName is used without context, it applies to all contexts
         (tableContextNames.every((n) => !props?.[propName]?.[n]) &&
-          props?.[propName]) ||
+          props?.[propName]);
+      if (value !== undefined) result[context][propName] = value;
+
+      let themeValue =
```
MikeKingdom, line 254: I'm not seeing how the theme values make it into the cell when it's not pinned. It seems like this gets dropped now.

It seems like the above value with the theme values still needs to get put in the non-pinned part of the results like it did before. You could still grab the first part without the theme piece and use it at 264 and just tack on the theme pieces (pinned version first)  here directly instead of tacking it on at line 287.

src/js/components/DataTable/buildState.js
```diff
@@ -248,38 +248,43 @@ export const normalizeCellProps = (props, theme) => {
...
+        value !== undefined &&
+        propName === 'background' &&
+        theme.background &&
+        value.opacity &&
+        !value.color
+      )
+        // theme context has an active background color but the
+        // theme doesn't set an explicit color, repeat the context
+        // background explicitly
+        value.color = normalizeBackgroundColor(theme);
```
MikeKingdom, line 279: I know this is pre-existing, but this looks like a bad side-effect that could actually be modifying the theme. Since value could be a reference directly into the theme, I'm wondering if this is actually modifying that part of the theme for everyone. This type of side-effect can cause some really odd and hard to find bugs.

src/js/components/DataTable/stories/Fill.js
```diff
@@ -42,9 +43,7 @@ export const Fill = () => (
         step={10}
         fill
         pin
-        background={{
-          pinned: { color: 'background-contrast' },
-        }}
+        background={{ body: 'light-1' }}
```
MikeKingdom, line 46: interesting test case but we may not want the example to promote setting the background this way

PR #5784, opened 2021-11-04
Fix Calendar hover styling

### LINKED ISSUE #5634: Calendar 1st day of month gets hover styling when clicking on other dates
### Expected Behavior
Hover styling should only be applied when mouse is over the date

### Actual Behavior
If `range` is true or you can select multiple dates on the Calendar, when clicking a date the hover styling is applied to the first day of the month even if you are selecting a different date.

Can be seen in the following stories:
https://storybook.grommet.io/?path=/story/visualizations-calendar-active-date--active-date
https://storybook.grommet.io/?path=/story/visualizations-calendar-dual--dual
https://storybook.grommet.io/?path=/story/visualizations-calendar-multiple--multiple
https://storybook.grommet.io/?path=/story/visualizations-calendar-range--range

### Steps to Reproduce
In one of the above stories select a date that is not the first of the month and see the hover styling being applied to the first

### Your Environment
- Grommet version: 2.17.5

taysea:
#### What does this PR do?

Updates how `active` calendar date is adjusted when a date is selected / when days grid receives focus. Previously, the firstDayInMonth was incorrectly receiving active treatment upon selection of a date OR if dates were already selection then the old date was receiving active treatment upon selection of a new date.

This approach instead only places the active date on the first day on the month when the Calendar receives focus via keyboard. Otherwise, the active date will always be the most recently selected date.

Tested in Calendar stories.

1. Click on a day on Calendar
   - confirm that hover styling is not on the first day of month
   - confirm that if you now press arrow key that the active date moves to the appropriate adjacent day
2. Tab to calendar days via keyboard
   - confirm that first day in month receives active treatment
   
Closes #5634 

No.

Yes. Fixed Calendar active day styling.

Backwards compatible.


jcfilben (approved):
Looks good to me!

ericsoderberghp (approved):

PR #5785, opened 2021-11-04
Resolve Tip memoryleak

taysea:
#### What does this PR do?

Resolves a memory leak that was occurring from an invalid setState in an effect cleanup due to recent Tip changes. We need to define the setState function as a separate const that is referenced in the effect and its cleanup, otherwise it is viewed as two separate functions.
 
src/js/components/Tip/Tip.js

Check console after clicking button in "themed" tip story.

Same as above, ensure no console errors.

Thread with explanation on solution: https://stackoverflow.com/a/47387145

BEFORE:
Screen Shot 2021-11-04 at 2 06 05 PM

AFTER:
Screen Shot 2021-11-04 at 2 12 21 PM

No.

No, functionality hadn't been released.

Backwards compatible.


halocline (approved):
Hah. I was just going to investigate why we just started seeing this. Thank you.

jcfilben (approved):
Nice fix!

ericsoderberghp (approved):

PR #5788, opened 2021-11-05
Name value list add margin

britt6612:
#### What does this PR do?
changes gap to `medium`
I changed to add `margin: { bottom: 'xxsmall' } `
I did this to take place of `gap` because 
base.js
storybook
snapshots
designs call for `medium` 

I changed to add `margin: { bottom: 'xxsmall' } `
I did this to take place of `gap` because 
`gap` when I added `gap` that then becomes a `div` in between the name and value
In which was causing accessibility warning that 
```
<dl> elements must only directly contain properly-ordered <dt> and <dd> groups,.
```

Therefore I added `margin: { bottom: 'xxsmall' } `


src/js/themes/base.js
```diff
@@ -1047,7 +1047,7 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       // extend: undefined,
     },
     nameValueList: {
-      gap: { column: 'large', row: 'small' },
+      gap: { column: 'medium', row: 'small' },
```
taysea, line 1050: I don't think we want to adjust the gap

src/js/components/NameValuePair/NameValuePair.js
```diff
@@ -25,7 +25,12 @@ const NameValuePair = ({ children, name: nameProp }) => {
   let name;
   if (typeof nameProp === 'string' || typeof nameProp === 'number')
     name = (
-      <Text as="dt" textAlign={nameAlign} {...theme.nameValuePair.name}>
+      <Text
+        as="dt"
+        margin={column ? { bottom: theme.nameValuePair.margin } : 'undefined'}
```
taysea, line 30:
```suggestion
        margin={column ? theme.nameValuePair.margin : 'undefined'}
```
margin has margin.bottom, so we should just put it directly here

src/js/components/NameValueList/stories/Layout.js
```diff
@@ -1,15 +1,22 @@
...
+} from 'grommet';
 import { grommet } from 'grommet/themes';
 import { data, metricData } from './data';
 
 export const Layout = () => (
   <Grommet theme={grommet}>
     <Box pad="small" gap="large">
       <>
-        <Text weight="bold" size="2xl">
+        <Heading level={2} weight="bold" s>
```
taysea, line 17: Since there isn't an `h1` on this page, I don't think it's semantically correct to have `h2`s -- wondering if we can just leave this story as is

src/js/themes/base.js
britt6612: Ill take out this shouldnt be here

src/js/components/NameValueList/stories/Layout.js
britt6612: okay thats fine we can leave it as is

taysea (commented):
Since this is more focused on `margin`, can we update the PR title?

taysea (commented):
Looks good, just need to update base.d.ts!

EDIT: already covered by TextProps

taysea (approved):
LGTM!

src/js/themes/base.js
```diff
@@ -1060,6 +1060,9 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         // any text props
         color: 'text',
         weight: 'bold',
+        margin: {
```
ericsoderberghp, line 1063: It seems a little hidden to me that this is only valid when in a column context. As the theme is structured, I would expect this to be applied all of the time. Given how we are using this. I'm wondering if we should instead place this under something like `theme.nameValuePair.column.gap` or `theme.nameValuePair.column.name.margin`. I have a slight preference for the former, as I think it aligns better with `theme.nameValueList.gap`.

britt6612: good point I made it so that within the `NameValuePair.js` this is under the `name` 
```
  margin={column ? theme.nameValuePair.margin : 'undefined'}

```
which by looking at the theme this does not be as appropriate. I originally leaned towards `theme.nameValuePair.column.gap` however since this is only acting like a `gap` and really a  `margin` under the hood I think something like `theme.nameValuePair.column.name.margin` would be appropriate

taysea (approved):

src/js/components/NameValuePair/NameValuePair.js
```diff
@@ -25,7 +25,21 @@ const NameValuePair = ({ children, name: nameProp }) => {
   let name;
   if (typeof nameProp === 'string' || typeof nameProp === 'number')
     name = (
-      <Text as="dt" textAlign={nameAlign} {...theme.nameValuePair.name}>
+      <Text
+        as="dt"
+        margin={
+          column
```
ericsoderberghp, line 31: I'm wondering if we should check the theme too. Something feels a bit odd about passing `margin={{ bottom: undefined }}`.

ericsoderberghp (approved):

PR #5789, opened 2021-11-05
NameValueList - Allow user to define different gap depending on pairProp direction

taysea:
#### What does this PR do?
Adds `theme.nameValueList.pair.column.gap` to theme to allow custom specification of gap for when a pair is in a column.

src/js/components/NameValueList/NameValueList.js

Tested in storybook examples, updated snapshots and confirmed changes are expected.

Storybook, inspect. OR update storybook example locally.

Row/Desktop:
Screen Shot 2021-11-05 at 9 49 20 AM

Column/Mobile:
Screen Shot 2021-11-05 at 9 49 31 AM

Yes.

No. Component hasn't been released.

Backwards compatible.


src/js/components/NameValueList/stories/typescript/Themed.tsx
```diff
@@ -7,6 +7,12 @@ import { data } from '../data';
 // Type annotations can only be used in TypeScript files.
 // Remove ': ThemeType' if you are not using Typescript.
 const customTheme: ThemeType = deepMerge(grommet, {
+  nameValueList: {
+    gap: {
+      column: 'xlarge',
+      row: 'xlarge',
```
britt6612, line 13: 👍

britt6612 (approved):

britt6612:
well just need to add to the site if the naming is what we decide on I can do this since Im adding the `margin`

src/js/themes/base.d.ts
```diff
@@ -1151,7 +1151,16 @@ export interface ThemeType {
     };
   };
   nameValueList?: {
-    gap?: { row?: GapType; column?: GapType };
+    gap?: { 
+      pairDirectionColumn?: {
```
ericsoderberghp, line 1155: This doesn't seem consistent with our pattern elsewhere of keying off of state at the root of the component, for example `theme.button.active`. Did you consider something like `theme.nameValueList.column.gap` and `theme.nameValueList.grid.gap`?

halocline: I also like the idea of keeping the state at the root, then setting attributes specifically for each particular state.

ericsoderberghp (approved):

britt6612 (approved):

kevingill55 (commented):
This looks good on my end! One comment about whether or not we should add the `Themed` storybook example to a Custom Themed directory (to adhere to our new storybook theming conventions)

britt6612:
> This looks good on my end! One comment about whether or not we should add the `Themed` storybook example to a Custom Themed directory (to adhere to our new storybook theming conventions)

good point pushing this change up now

Issue #3552, opened 2019-11-15
useEventHandler custom hook suggestion


ShimiSun:
`onResize` depends on `width` and `height` variables from state. So whenever these variables will change, our effect would re-run and add/remove the handler again.. and again.
If it's not much of a problem, I can move the handler into the effect.

I suggested above that we should extract this as a custom hook `useEventHandler` based on this example: https://codesandbox.io/s/z64on3ypm
I did the same thing with `onResize` as in the above example.
There multiple components in which we would be able to use this hook.

_Originally posted by @husseyexplores in https://github.com/grommet/grommet/pull/3458_

ericsoderberghp:
Which components are you referring to for `onResize` here?

husseyexplores:
I can chime in, although I haven't contributed to this repo in a while.

This comment was posted in the context of `Diagram.js` component. [Here](https://github.com/grommet/grommet/blob/master/src/js/components/Diagram/Diagram.js#L79) is the link.

I assumed that this was a fairly common use-case (adding handlers) and many components will be able to use this hook. But, I may be wrong here.

g4rry420:
@husseyexplores  Free feel to hop on PR, if you wanted to suggest anymore usecases.

PR #5792, opened 2021-11-08
Refactor: Added useEventListener custom hook

g4rry420:

#### What does this PR do?
Added useEventListener custom hook, so that our effect don't run over and over again.

utils => use-event-listener.js

Pre-written tests has been passed successfully.

By making sure the behaviour of the refactored componenets remains same.

I have only found a couple of components where this custom hook is useful, if there are any other, please let me know.

#3552 

N/A

No

Backwards compatible.

### DIFF
--- a/src/js/components/Chart/Chart.js
+++ b/src/js/components/Chart/Chart.js
@@ -1,4 +1,4 @@
-import React, { useContext, useEffect, useMemo, useState } from 'react';
+import React, { useContext, useMemo, useState } from 'react';
 
 import { ThemeContext } from 'styled-components';
 import { useLayoutEffect } from '../../utils/use-isomorphic-layout-effect';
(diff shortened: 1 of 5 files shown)


src/js/components/Chart/Chart.js
```diff
@@ -190,19 +191,16 @@ const Chart = React.forwardRef(
...
-      return undefined;
-    }, [containerRef, needContainerSize]);
+    const onResize = () => {
+      const { parentNode } = containerRef.current;
+      const rect = parentNode.getBoundingClientRect();
+      setContainerSize([rect.width, rect.height]);
+    };
+
+    if (needContainerSize) {
+      // eslint-disable-next-line react-hooks/rules-of-hooks
```
ericsoderberghp, line 201: We shouldn't be disabling the rules-of-hooks.

ericsoderberghp:
I think we should look into addressing the specific issue for the component identified in the issue first.

g4rry420:
I am closing this, Since, the custom hook is only applicable to 2-3 components. Feel free to reopen, if you think more components can be included.

PR #5797, opened 2021-11-09
align icon to be start with multi line text

britt6612:
#### What does this PR do?
This PR changes a story to display how to align an`icon` to be `align: start` with the text `value` when `value` is mult-line
story/customValue.js
Storybook
View story
It was brought up that when value is multi-line the icon needs to be `align: start`  in the current story we have now it would be `align: center` which is what we do not want. 

We want it to look as follows: 
However the size of the `icon` doesn't line up with the `line-height` of the `text` in order to accomplish a better alignment I added the `margin={{ top: 'xsmall' }}` 

This isn't quite "out of the box" working like a user would want this is more of a work around but if we want to open another issue to explore how we can better align our grommet-icons with `text` Im happy to explore that route.

Screen Shot 2021-11-09 at 12 34 23 PM
Designs when value is multi-line with icon
no
no
yes


src/js/components/NameValueList/stories/data.js
```diff
@@ -19,7 +19,7 @@ export const metricData = {
 };
 
 export const statusData = {
-  Power: 'Ok',
+  Power: 'Management for campus, branch, remote, and data center network',
```
ericsoderberghp, line 22: In the spirit of keeping our stories real, this long string doesn't seem to be appropriate for "Power". Perhaps we can just add another pair?

britt6612: yes I can add another pair that is more of a "real" story.. I was not really wanting to change the story just wanted to  put context behind why the `icon` does not align with `text`

kevingill55 (approved):
This looks good to me. Seems like a tricky concept to try and standardize. Ideally, people would have more flexibility over the size of their icons but I think Grommet only supports t-shirt sizes for icons for now.

ericsoderberghp (approved):

PR #5800, opened 2021-11-10
enhance checkbox to align label only if it is a string and not a cust…

britt6612:
#### What does this PR do?
This PR changes the `props.theme.checkBox.label.align` to only apply to `labels` that are strings so any custom `labels` with not receive this alignment
styledCheckBox.js
storybook 
I needed to pass 
`label={label}` to check if `label` was a string which is why the snapshot changes are occurring 

https://github.com/grommet/grommet-theme-hpe/pull/222

> For example, if the caller does label={<Text size="large">my fancy label</Text>} this would not take the alignment from the theme

yes


kevingill55 (approved):
This works well! Is this already documented or is this already the expected behavior?

taysea (approved):

ericsoderberghp (approved):

PR #5805, opened 2021-11-16
Update DateInput to avoid timezone conflicts

### LINKED ISSUE #5539: DateInput test is not passing for other time zones
except for a date-related one because the test doesn't allow for my +10 timezone.
### Expected Behavior
all tests should pass

no errors
### Actual Behavior
Different time zones fail the `dateinput` test
Screen Shot 2021-08-24 at 8 20 34 AM

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

taysea:
[Preview behavior](https://deploy-preview-5805--sad-tereshkova-173d07.netlify.app/?path=/story/input-dateinput-format-inline--format-inline)

#### What does this PR do?

This PR updates DateInput / Calendar to support all timezones. The intended behavior is:
1. If defaultValue or value is provided with a timezone, all returned values will maintain that timezone and the calendar selected styling with correspond to that timezone.
2. If defaultValue or value is provided _without_ a timezone, all returned values will just be YYYY-MM-DD.
3. If defaultValue or value is provided as undefined / [], all returned values will be in UTC relative to local timezone.
 
TO-DO:
- [ ] Selected date is off when in a timezone pre-PT

src/js/components/DateInput/__tests__/DateInput-test.tsx

Testing behavior using this modified story: https://deploy-preview-5805--sad-tereshkova-173d07.netlify.app/?path=/story/input-dateinput-format-inline--format-inline

Use the story above and interact with each one via clicking, typing values, pasting full date value (as it happens in the jest test)

try changing your computer timezone to something before and after PT.

Closes #5539 

No.

No.

backwards compatible.

### DIFF
--- a/src/js/components/DateInput/DateInput.js
+++ b/src/js/components/DateInput/DateInput.js
@@ -28,6 +28,9 @@ import {
 } from './utils';
 import { DateInputPropTypes } from './propTypes';
 
+const timestampRegExp = new RegExp(/T.*/);
+const getTimestamp = (date) => new Date(date).toISOString().split('T')[1];
+
 const DateInput = forwardRef(
   (
     {
@@ -62,6 +65,21 @@ const DateInput = forwardRef(
       initialValue: defaultValue,
     });
 
+    let timestamp = false;
+    if (Array.isArray(defaultValue)) {
+      if (timestampRegExp.test(defaultValue[0])) {
+        timestamp = getTimestamp(defaultValue[0]);
+      }
+    } else if (
+      typeof defaultValue === 'string' &&
+      timestampRegExp.test(defaultValue)
+    ) {
+      timestamp = getTimestamp(defaultValue);
+    } else if (Array.isArray(value) && timestampRegExp.test(value[0])) {
+      timestamp = getTimestamp(value[0]);
+    } else if (typeof value === 'string' && timestampRegExp.test(value)) {
+      timestamp = getTimestamp(value);
+    }
     // do we expect multiple dates?
     const range = Array.isArray(value) || (format && format.includes('-'));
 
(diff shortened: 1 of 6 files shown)


jcfilben (commented):
This fixes the issue with the 'dates initialized with empty array' test, but I am still getting failing tests for 'type format inline', 'type format inline short', 'type format inline range', and 'type format inline range partial' when I change my timezone. (Tested with Central Africa Time)

src/js/components/DateInput/DateInput.js
```diff
@@ -28,6 +28,9 @@ import {
 } from './utils';
 import { DateInputPropTypes } from './propTypes';
 
+const timestampRegExp = new RegExp(/T.*/);
```
MikeKingdom, line 31: I don't think this is really any different than checking indexOf('T').

src/js/components/DateInput/DateInput.js
```diff
@@ -62,6 +65,21 @@ const DateInput = forwardRef(
       initialValue: defaultValue,
     });
 
+    let timestamp = false;
+    if (Array.isArray(defaultValue)) {
```
MikeKingdom, line 69: could pull the array vs string out into a function and simplify this to a simple defaultValue vs value decision.

ericsoderberghp:
Given #2, that not supplying a timezone should use the local timezone, in Calendar.js shouldn't `normalizeForTimezone()` not be using `localTimezoneToUTC()` when no timezone is supplied?

src/js/components/DateInput/utils.js
```diff
@@ -140,17 +142,31 @@ export const textToValue = (text, schema, valueProp, range) => {
...
-      date = `${date.split('T')[0]}T${valueDate.split('T')[1]}`;
+    // match time and timezone of supplied value
+    let h;
+    let m;
+    let s;
+    let ms;
+    if (timestamp) {
+      const mockDate = new Date(
+        `${
+          new Date(parts.y, parts.m - 1, parts.d).toISOString().split('T')[0]
```
ericsoderberghp, line 153: I wonder if part of the problem is using this type of Date constructor, where the year, month, and day "The parameter values are all evaluated against the local time zone, rather than UTC." according to MDN.

Instead, I'm thinking we should build the ISO string directly such that, when combined with the timestamp, are then correctly interpreted by the `Date` constructor at line 151.

ericsoderberghp (commented):
I'd start with this comment. I think we need to work more in ISO strings when dealing with the text value.

src/js/components/DateInput/utils.js
taysea: Yes I think we are running into issues when having everything rooted in local time zone and then trying to combine with other timezones. I will explore.

taysea:
Closing in favor of https://github.com/grommet/grommet/pull/5837

PR #5806, opened 2021-11-16
Fix InfiniteScroll lastPage calculation

### LINKED ISSUE #5794: InfiniteScroll doesn't call onMore enough...calculates lastPage wrong at step boundaries
### Expected Behavior

onMore should get called when not enough items have been loaded (ie. step is too small)

### Actual Behavior

onMore doesn't get called enough times to fill up the visible area,

The problem seems to be how InfiniteScroll calculates `lastPage`. Currently it does:

```jsx
  // the last page we have items for
  const lastPage = useMemo(
    () => Math.floor(items.length / step),
    [items.length, step],
  );
```
`lastPage` is a page index. If `step` is 5 and `items.length` is 15 it calculates `lastPage` as 3 when it should still be 2.

The above code should be:
```jsx
// the last page we have items for
  const lastPage = useMemo(
    () => Math.max(0, Math.ceil(items.length / step) - 1),
    [items.length, step],
  );
```

### Steps to Reproduce
This was observed with the DataTable onUpdate changes.

Define a datatable which has plenty of vertical height.
Define onMore to return 5 extra items on each call.
Set step to 5
Set data to something with 5 items.
onMore should get called and set data to 10 items
onMore may get called once more and set data to 15 items
onMore won't continue to get called.

### Your Environment

- Grommet version: stable 2.17.3+
- Operating System and version (desktop or mobile): windows 10

MikeKingdom:
#### What does this PR do?
InfiniteScroll calculates its lastPage index slightly wrong. This fixes that calculation.

The test snapshot changes are cases where InfiniteScroll really shouldn't have put
a marker div. Most of the cases are pagination enabled (where it doesn't need to
detect that the list is scrolled to the end). The change to the InfiniteScroll snaphot
is a case where it had a number of items that was exactly a multiple of the step
(the main bug this fixes.)

InfiniteScroll.js

Storybook plus some special situations with DataTable onUpdate.

See #5794 

Fixes #5794 

No

Yes

This is backwards compatible

### DIFF
--- a/src/js/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
+++ b/src/js/components/DataTable/__tests__/__snapshots__/DataTable-test.js.snap
@@ -19342,7 +19342,7 @@ exports[`DataTable select 2`] = `
 `;
 
 exports[`DataTable should apply pagination styling 1`] = `
-.c19 {
+.c18 {
   display: inline-block;
   -webkit-flex: 0 0 auto;
   -ms-flex: 0 0 auto;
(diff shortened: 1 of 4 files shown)


ericsoderberghp (approved):

taysea (approved):

halocline (approved):
Thank you, Mike.

PR #5809, opened 2021-11-17
Update Accordion tests to use best testing practices

jcfilben:
#### What does this PR do?
Updated tests for the component to follow the testing guidelines//best practices that are being added to the contributing guide. We can use the Accordion tests as a reference for people to look at.

I also added `eslint-plugin-testing-library` to catch errors related to using react testing library. For now I only enabled it for the Accordion test file.

Note: switching from `container.firstChild` to `asFragment()` is causing the majority of snapshot changes

Accordion-test.tsx

Related to grommet/hpe-design-system#1959

No

No

Backwards compatible

### DIFF
--- a/src/js/components/Accordion/__tests__/Accordion-test.tsx
+++ b/src/js/components/Accordion/__tests__/Accordion-test.tsx
@@ -4,7 +4,8 @@ import 'jest-axe/extend-expect';
 import 'regenerator-runtime/runtime';
 
 import { axe } from 'jest-axe';
-import { render, fireEvent, act } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 
 import { Accordion, AccordionPanel, Box, Grommet } from '../..';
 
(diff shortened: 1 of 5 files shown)


halocline (approved):

src/js/components/Accordion/__tests__/Accordion-test.tsx
```diff
@@ -26,20 +27,20 @@ describe('Accordion', () => {
 
     const results = await axe(container);
     expect(results).toHaveNoViolations();
-    expect(container).toMatchSnapshot();
+    expect(asFragment()).toMatchSnapshot();
   });
 
   test('no AccordionPanel', () => {
-    const { container } = render(
+    const { asFragment } = render(
```
taysea, line 34: can `asFragment` come from "screen"? Or it needs to get defined in this way?

jcfilben: Only queries (`getByRole`, `getByLabelText`, etc.) can come from screen so `asFragment` can't come from screen and has to be defined like that

taysea: Thanks!

taysea (approved):
Thanks!

ericsoderberghp (approved):

PR #5820, opened 2021-11-24
add dependencies

britt6612:


halocline:
@britt6612 - I looked into this a bit. I think we are going to have to phase in the dependency upgrades in batches of like updates (as opposed to all at once). The babel dependencies are particularly problematic and it looks like we are going to need to alter configs to accommodate.

Perhaps we can touch base next week?

Also, we should run `yarn build && yarn build-storybook` to ensure the builds succeed prior to commits.

britt6612:
closing PR going to update in groups for dependencies

PR #5827, opened 2021-12-01
add or for ts in theme for pagination

britt6612:
#### What does this PR do?
This PR fixes a TS error that was with `pagination` 
base.d.ts
storybook

The error was 
```
{ background: { color: string; }; color: string; }' is not assignable to type 'ButtonKindType & { [key: string]:
```
It seemed that since there was a & with the `{ [key: string]: ButtonKindType }` it was flagging that it should of been included but we do not always want this so I added an OR option

https://stackoverflow.com/questions/57145172/how-to-create-an-optional-object-key-string-in-typescript/57145386
no
no
backwards compatible

### DIFF
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -177,12 +177,13 @@ interface ButtonType {
   primary?: ButtonKindType;
   secondary?: ButtonKindType;
   option?: ButtonKindType;
-  active?: ButtonKindType & {
-    default?: ButtonKindType;
-    primary?: ButtonKindType;
-    secondary?: ButtonKindType;
-    [key: string]: ButtonKindType;
-  };
+  active?:
+    | (ButtonKindType & {
+        default?: ButtonKindType;
+        primary?: ButtonKindType;
+        secondary?: ButtonKindType;
+      })
+    | { [key: string]: ButtonKindType };
   disabled?: ButtonKindType & { opacity?: OpacityType };
   hover?: ButtonKindType & {
     default?: ButtonKindType;
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -178,7 +178,7 @@ interface ButtonType {
   secondary?: ButtonKindType;
   option?: ButtonKindType;
   active?:
-    | (ButtonKindType & {
+    (ButtonKindType & {
         default?: ButtonKindType;
         primary?: ButtonKindType;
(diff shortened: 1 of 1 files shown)


jcfilben (approved):
Looks good!

amandadupell (approved):
lgtm!

src/js/themes/base.d.ts
```diff
@@ -177,12 +177,13 @@ interface ButtonType {
...
   secondary?: ButtonKindType;
   option?: ButtonKindType;
-  active?: ButtonKindType & {
-    default?: ButtonKindType;
-    primary?: ButtonKindType;
-    secondary?: ButtonKindType;
-    [key: string]: ButtonKindType;
-  };
+  active?:
+    | (ButtonKindType & {
```
halocline, line 181: How is this OR affecting the type evaluation? It feels out of place.
```suggestion
    (ButtonKindType & {
```

britt6612: good catch I did not mean to add that OR there

britt6612:
In order to get this error you need do the following:
1. pull down master branch
2. go through and `rm -rf node_modules/ && rm -rf yarn.lock && yarn cache clean && yarn`
3. run `yarn build` 
4. run `yarn build-storybook` 
5. you should get something similar to 

Screen Shot 2021-12-01 at 11 01 33 AM

halocline (commented):
Hey @britt6612 - The reason the `[key: string]` was added in the first place was to accommodate a scenario like the following where a caller's theme may have a custom named button which they'd like to reference ([more context](https://github.com/grommet/grommet/pull/5722/files#r735952176)).

I confirmed this modification fixes the typing for `active` in both the pagination/customTheme story as well as the scenario below - @britt6612 while we are in here, should we also apply the same typing to `hover` as well?

```
const customTheme2 = { 
  button: { 
    default: {}, 
    bright: { 
      color: 'text-strong', 
      border: { 
        color: 'skyblue', 
        width: '2px', 
      }, 
    }, 
    active: { 
      bright: { 
        background: { 
          color: '#CA9CEA', 
        }, 
        border: { 
          color: 'transparent', 
        }, 
        color: 'text', 
      }, 
    }, 
  }, 
  pagination: { 
    button: 'bright', 
  }, 
};
```

halocline (approved):
Thank you @britt6612

Issue #5370, opened 2021-06-24
Form validation not occurring on Form load/mount when input values are manually set


canyongrace:
### Expected Behavior

Background: When dynamically adding and removing items from a form, we can't use formValues (controlled form) due to issue #5248. The alternative for our use case is to set the value of a TextInput field (part of a FormField) manually (although technically I believe this is still controlled). We expect the FormField validation to run immediately when the Form is mounted.

### Actual Behavior

The FormField validate function does not appear to get called when an initial TextInput value is set. On load/mount of the form, if the input value is incorrect, no error is displayed. Yet if the user edits the value, the validate function will suddenly start executing and display the error, even if they re-type the same value as the default.

This behavior is inconsistent for the user experience and doesn't let them see an error when the form values are pre-populated. For example, in a wizard-type scenario, this could allow the user to move past a screen without correcting invalid values.

### URL, screen shot, or Codepen exhibiting the issue

See the following codesandbox for a simple, reproducible example. The incorrect value of 'a', set by default, doesn't display an error. But if the user types additional letters and deletes them, suddenly an error is displayed for 'a'. As previously stated, we would expect the error to be displayed at the beginning as well.

https://codesandbox.io/s/flamboyant-beaver-v04st?file=/src/App.js

### Steps to Reproduce

See codesandbox explanation

### Your Environment

- Grommet version: 2.17.3
- Browser Name and version: Firefox 89.0.1
- Operating System and version (desktop or mobile): Windows 10

cc: @ypinneka @shreyashah7

canyongrace:
In addition, it appears if an element is removed and re-added (say the user checks a box to enable functionality), Form validation does not occur on any pre-existing values in the input fields. There is no way for the caller to trigger the validation method manually.

@ypinneka @shreyashah7

ericsoderberghp:
Now that https://github.com/grommet/grommet/issues/5248 has been addressed, is this still an issue?

canyongrace:
Hi @ericsoderberghp. From our perspective, this is still an issue. I discussed with @taysea last week.

When using the Form 'value' prop, validation works properly. But we're not using the Form value prop in this scenario. The reason is that our use case involves the user dynamically adding and removing form fields. If we were using the Form value prop, we'd have to manage both adding and removing formValues - adding a new one with the correct index or removing one with the correct index and renaming the keys.

[...]

PR #5834, opened 2021-12-06
Add Form validation onMount for controlled input forms

halocline:
#### What does this PR do?

- Closes https://github.com/grommet/grommet/issues/5370 by ensuring validation is run 'onMount' for "controlled input forms."
- Consolidates duplicate validation logic and the number of places from which it is called.
- Attempts to improve readability by modifying naming and apply consistently throughout.

- Review [original issue](https://github.com/grommet/grommet/issues/5370)
  - Desire is to run form validation onMount, however that was not working for controlled input forms.
  - The source of the issue is that a controlled input's value is not available to the Form onMount; the value isn't available until second render. This PR proposes a solution to treat the second render as the mount event for the onMount validation logic.
- Form.js
  - Modified to run validation logic:
    1) onMount (simulated mount event)
    2) in response to user input from onChange, onBlur, onSubmit
    3) in response to user input attempting to resolve an existing error

- All existing Storybook stories
- Added Jest tests for onMount validation in controlled form and controlled input scenarios

- The Jest test 'validate on mount - controlled input' would be a good one.

- [x] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority)) -- using previously implemented testing approach of `queryByAllText` so as not to change too much all at once. Recommend a focussed issue at re-viewing and re-writing Form tests to new standard.
- [x] `userEvent` is used in place of `fireEvent`. **Not Applicable in this PR**
- [x] `asFragment()` is used for snapshot testing. **Not Applicable in this PR**

https://github.com/grommet/grommet/issues/5370

No.

Yes.

Backwards compatible.


src/js/components/Form/Form.js
```diff
@@ -165,16 +174,26 @@ const Form = forwardRef(
...
     const [touched, setTouched] = useState(defaultTouched);
-    const [validationResults, setValidationResults] = useState(
-      defaultValidationResults,
-    );
+    const [validationResults, setValidationResults] = useState({
+      errors: errorsProp,
+      infos: infosProp,
+    });
+    const validationResultsRef = useRef({});
+    // Simulated onMount state. Consider Form to be mounted once it has
```
halocline, line 188: Open to alternate ideas here. Played with a number of approaches, but landed here.

src/js/components/Form/Form.js
```diff
@@ -165,29 +174,27 @@ const Form = forwardRef(
...
-    // potential click events before running the onBlur validation.
-    // otherwise, click events like reset, etc. may not be registered.
-    // for a detailed scenario/discussion,
-    // see: https://github.com/grommet/grommet/issues/4863
-    // the value of pendingValidation is the name of the FormField
-    // awaiting validation.
-    const [pendingValidation, setPendingValidation] = useState(undefined);
-
+    const [validationResults, setValidationResults] = useState({
+      errors: errorsProp,
```
taysea, line 184: Did we delete `defaultValidationResults`? or is it still being used somewhere else. I didn't see it deleted but seems like we're not using it

src/js/components/Form/Form.js
taysea: I am okay with this approach

taysea (commented):
Just one small comment

src/js/components/Form/Form.js
halocline: Being used by `onReset`, L526-ish.

src/js/components/Form/Form.js
```diff
@@ -165,29 +174,27 @@ const Form = forwardRef(
...
+    const validationResultsRef = useRef({});
+    // Simulated onMount state. Consider Form to be mounted once it has
+    // accounted for values originating from controlled inputs (available
+    // at second rendering).
+    const [mounted, setMounted] = useState(false);
     useEffect(() => {
-      setPendingValidation(undefined);
-      setValidationResults({ errors: errorsProp, infos: infosProp });
-    }, [errorsProp, infosProp]);
+      if (!mounted) setMounted('mounting');
```
MikeKingdom, line 193: What's are stance on changing types of data? This goes from boolean to string to boolean. This can be hard to understand and have subtle bugs if you have something like `if (!mounted)` .
Would it be better to stick to strings?

MikeKingdom (commented):
Looks pretty nice.

src/js/components/Form/Form.js
halocline: Good call. I agree. Addressed.

src/js/components/Form/Form.js
ericsoderberghp: hmm... My initial reaction is that I'm wondering if there's a more semantic way of attacking this, rather than relying on the number of renderings. I'll see if this sticks after reviewing more.

src/js/components/Form/Form.js
```diff
@@ -165,29 +174,27 @@ const Form = forwardRef(
...
-    // see: https://github.com/grommet/grommet/issues/4863
-    // the value of pendingValidation is the name of the FormField
-    // awaiting validation.
-    const [pendingValidation, setPendingValidation] = useState(undefined);
-
+    const [validationResults, setValidationResults] = useState({
+      errors: errorsProp,
+      infos: infosProp,
+    });
+    const validationResultsRef = useRef({});
```
ericsoderberghp, line 187: I'm not sure why we need both `validationResults` and `validationResultsRef`. They both seem to refer to the same thing "validation results", but one is state and the other is a ref. How do they differ? When should one used vs. another? Why doesn't it work to just have the state?

Perhaps a comment above each of them explaining their purpose and how they are related?

src/js/components/Form/Form.js
```diff
@@ -31,7 +31,7 @@ const stringToArray = (string) => {
   return undefined;
 };
 
-const getValueForName = (name, value) => {
+const getFieldValue = (name, value) => {
```
ericsoderberghp, line 34: I like the naming changes and the test changes. I think there are enough of them that it might be easier to digest the changes if they were separated out into a separate pull request.

src/js/components/Form/Form.js
```diff
@@ -165,29 +174,27 @@ const Form = forwardRef(
...
-      setPendingValidation(undefined);
-      setValidationResults({ errors: errorsProp, infos: infosProp });
-    }, [errorsProp, infosProp]);
+      if (!mounted) setMounted('mounting');
+      else if (mounted === 'mounting') setMounted('mounted');
+    }, [mounted]);
+    // `pendingValidation` is the name of the FormField awaiting validation.
+    const [pendingValidation, setPendingValidation] = useState(undefined);
 
     const validations = useRef({});
```
ericsoderberghp, line 199: Should this be `validationRulesRef`? I'm feeling like the terms "validation" and "validation rule" aren't always being used consistently. But, perhaps there's a perspective I'm missing.

src/js/components/Form/Form.js
ericsoderberghp: If I'm understanding the need for this, it's that Form renders something like a TextInput and the TextInput calls `useFormInput` with the controlled `value` but Form isn't able to put that in its `valueState` right away and instead waits for a `useEffect` to put it in later, after the first render. Have I got that right?

ericsoderberghp: This approach seems vulnerable to a Form being rendered in multiple steps. For instance, the caller renders the form with the data it has from an API request but then sees that one of the values requires another API request to get the second chunk of data. The user is shown the initial form and a Spinner as it waits to load the rest of the form. This is a bit contrived but I'm wondering if the basic pattern could still occur.

ericsoderberghp: I'm wondering if we can queue up the initial controlled values inside Form in a ref and let the Form handle both `setValueState` and running mount time validations in an effect of its own, instead of relying on an effect inside `useFormInput`? So:
- TextInput calls `useFormInput` with the controlled `value`
- this `useFormInput` adds the value to a ref inside Form, call it `inputValuesRef`.
- Form has an effect that checks if there are any values in `inputValuesRef`, adds them to `valueState`, runs any mount time validations, and clears `inputValuesRef`.
I'm not 100% sure, just contemplating. Thoughts?

MikeKingdom:
Resolved by #5881

PR #5838, opened 2021-12-08
DateInput Calendar Icon Opens Drop when Clicked

### LINKED ISSUE #5706: Add option to show Calendar on DateInput when the icon is clicked
### Expected Behavior

If the format prop is specified, clicking on the calendar icon in the DateInput should open the calendar.

### Actual Behavior

If the format prop is not set clicking the icon shows the calendar (as expected). However, when the format prop is set clicking the calendar icon does not show the calendar. The only way to show the calendar is by pressing space - however, this also causes the page to scroll down.

This can be seen in the docs:
- With format prop: https://storybook.grommet.io/?path=/story/input-dateinput-european--european
- Without format prop: https://storybook.grommet.io/?path=/story/input-dateinput-simple--simple

### Steps to Reproduce

Add the DateInput component and click on the calendar icon.

### Your Environment

This is behaviour on the current stable branch following this commit: https://github.com/grommet/grommet/pull/5667

jcfilben:
#### What does this PR do?
Adds the ability to click the DateInput icon to open and close the drop. This more closely aligns with the [native html date input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) behavior.

DateInput.js

Tested with existing DateInput stories

DateInput stories

Closes #5706 

No

Yes (DateInput icon fixed to open and close drop when format prop is specified)

Backwards compatible (If we consider it a bug that this behavior wasn't happening before)


jcfilben:
Note on Chromatic Snapshot Changes:
The form story is changing by 1px because the padding around the MaskedInput is set to 11px by default. Previously the icon was contained within the MaskedInput so the padding was applied to the right of the icon. Now the icon is outside of the MaksedInput so the padding is applied between the input and icon. The icon had a pad of 'small' or 12px applied to the right side to create space between the icon and border. Going from 11 to 12px creates a slight change in the snapshot.

The other snapshot changes are caused by the icon being wrapped in a button. This slightly alters the positioning of the icon which is reflected in the snapshot. This can be fixed by wrapping the icon in a box instead of a button but I opted for a button for accessibility reasons.

src/js/components/DateInput/DateInput.js
```diff
@@ -56,6 +59,7 @@ const DateInput = forwardRef(
       (theme.dateInput.icon && theme.dateInput.icon.size) || 'medium';
     const { useFormInput } = useContext(FormContext);
     const ref = useForwardedRef(refArg);
+    const calendarIconRef = createRef();
```
taysea, line 62: Should this be `useRef()`?

src/js/components/DateInput/DateInput.js
```diff
@@ -172,42 +176,51 @@ const DateInput = forwardRef(
...
+              }}
+              onFocus={(event) => {
+                announce(
+                  formatMessage({ id: 'dateInput.openCalendar', messages }),
+                );
+                if (onFocus) onFocus(event);
+              }}
+            />
+            <Button
+              onClick={open ? () => closeCalendar() : () => openCalendar()}
```
taysea, line 218: You'll need to double check this locally but if closeCalendar / openCalendar are already callback functions, can we just do `open ? closeCalendar : openCalendar`?

taysea (commented):
A few small comments but looking good!

src/js/components/DateInput/DateInput.js
```diff
@@ -172,42 +177,51 @@ const DateInput = forwardRef(
...
+              }}
+              onFocus={(event) => {
+                announce(
+                  formatMessage({ id: 'dateInput.openCalendar', messages }),
+                );
+                if (onFocus) onFocus(event);
+              }}
+            />
+            <Button
+              onClick={open ? closeCalendar : openCalendar}
```
taysea, line 218: Every time I click, it is opening the calendar instead of opening and closing -- maybe removing the callback wasn't the right call.

Can we do:
```
haven't tried this locally
() => open ? closeCalendar() : openCalendar()

OR the original approach

open ? () => closeCalendar() : () => openCalendar()
```

taysea: Or is having it always open on click the desired behavior? That seems to be the behavior in the native HTML: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date

if that's the case, we could just to onClick={() =>openCalendar}

jcfilben: This was something I was debating. Even though the native HTML date input always opens on click I was leaning towards having the button open and close the drop.

jcfilben: When I try `() => open ? closeCalendar() : openCalendar()` I get the following error: `Too many re-renders. React limits the number of renders to prevent an infinite loop.`

Going to go with `open ? () => closeCalendar() : () => openCalendar()`, this appears to be working correctly

ericsoderberghp: What about: `open ? closeCalendar : openCalendar`?

jcfilben: `open ? closeCalendar : openCalendar` was resulting in the calendar only opening instead of opening and closing

src/js/components/DateInput/DateInput.js
```diff
@@ -44,6 +46,7 @@ const DateInput = forwardRef(
       name,
       onChange,
       onFocus,
+      plain,
```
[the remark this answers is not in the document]
jcfilben: When the DateInput is contained within a FormField it adds the plain prop which was passed down to the MaskedInput through `...rest`. Now plain is always set to true on MaskedInput and the border is on the Box container. I extracted the plain prop so that I could use it to disable the border on the Box when plain is true (See line 199).

src/js/components/DateInput/DateInput.js
jcfilben: After changing the onClickOutside function `open ? closeCalendar : openCalendar` works fine. Switching to use this and pushing up changes now.

ericsoderberghp (approved):

PR #5840, opened 2021-12-09
enable avatar storybook theme

matheushahnn:
It is a ticket related to https://github.com/grommet/grommet/issues/5735 
it enabled to switch theme in Avatar component in storybook

### DIFF
--- a/src/js/components/Avatar/stories/CustomThemed/Stacked.tsx
+++ b/src/js/components/Avatar/stories/CustomThemed/Stacked.tsx
@@ -2,8 +2,7 @@ import React from 'react';
 
 import { Favorite, UserFemale, UserNew } from 'grommet-icons';
 
-import { Avatar, Box, Grommet, Stack } from 'grommet';
-import { grommet } from 'grommet/themes';
+import { Avatar, Box, Stack } from 'grommet';
 
 export const Stacked = () => {
   const shimi = '//s.gravatar.com/avatar/b7fb138d53ba0f573212ccce38a7c43b?s=80';
@@ -50,7 +49,7 @@ export const Stacked = () => {
   );
 
   return (
-    <Grommet theme={grommet}>
+    <Box>
       <Box align="center" gap="medium" pad="large" background="dark-1">
         {/* Nested Avatars */}
         <Stack anchor="bottom-right">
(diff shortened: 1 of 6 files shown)


matheushahnn:
In the story Avatar/Basic the Favorite Icon disappears when selecting hpe theme 
image

Is that expected?

jcfilben:
The Favorite Icon disappears in the Basic story while using the hpe theme because it uses "accent-4" and "accent-2" which are not defined in the [hpe theme](https://github.com/grommet/grommet-theme-hpe/blob/master/src/js/themes/colors.js).

I think we should change the colors on this story to something that is defined in all the themes. We could do something like this: 
```javascript
<Avatar background="dark-4">
  <Favorite color="light-2" />
</Avatar>
```

src/js/components/Avatar/stories/CustomThemed/Stacked.tsx
```diff
@@ -76,10 +75,10 @@ export const Stacked = () => {
...
         <GroupedGravatarRTL />
       </Box>
-    </Grommet>
+    </Box>
   );
 };
 
 export default {
-  title: 'Visualizations/Avatar/Stacked',
+  title: 'Visualizations/Avatar/Custom Themed/Stacked',
```
jcfilben, line 83: This story shouldn't go in the custom themed folder since it doesn't use a custom theme. Instead it looks like we need to change the colors like in the Basic story since the colors for some of the Avatar's are not defined in the hpe theme.

matheushahnn:
@jcfilben  Thank you for the review :D

src/js/components/Avatar/stories/typescript/Stacked.tsx
```diff
@@ -43,14 +42,14 @@ export const Stacked = () => {
       <Avatar background="accent-2" margin={{ left: 'medium' }}>
```
jcfilben, line 42: There are still some colors in this story that aren't available in the hpe theme (any of the "accent-*" colors)

src/js/components/Avatar/stories/typescript/Stacked.tsx
```diff
@@ -67,7 +66,7 @@ export const Stacked = () => {
         {/* Notification */}
         <Stack anchor="top-right">
           <Avatar src={shimi} />
-          <Box pad="xsmall" round background="accent-4" responsive={false} />
+          <Box pad="xsmall" round background="black-4" responsive={false} />
```
jcfilben, line 69:
```suggestion
          <Box pad="xsmall" round background="light-1" responsive={false} />
```
I don't think black-4 is a color

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

Issue #4559, opened 2020-10-01
How to improve the performance of multiple TextInput in a parent component?


johnliu55tw:
So I was designing a form that allows users to dynamically add/remove entries, where an entry contains a `Select` and `TextInput` component:

![image](https://user-images.githubusercontent.com/5047724/94774706-0624aa80-03f1-11eb-970b-85ddfd9ce4dd.png)

The naive implementation is to create a parent component that keeps the state of each entry and passing the required states and several callbacks as props to the children component, so that the parent can control the children and track the state change. However, the performance is bad with just around ten entries created. Input is slightly lagged, and you won't see characters coming out one-by-one when holding a key. It just freezes when a key is pressed, and then a bunch of characters coming up after released the key.

Here's the code example. Click Add icon to add more entries:
https://codesandbox.io/s/dynamic-input-state-in-parent-qy9pc?file=/src/App.js

I have another design that keeps the state of `TextInput` and `Select` inside the children component. The parent component uses an **instance variable** instead of state to track children's state, so that children state changes won't trigger rendering. This design does improve the performance, but I'm just a few months into React so I'm not sure this the correct way to use React.

I'm wondering how can I improve the performance? The performance is acceptable when using pure React (with only `div`, `input`, `button`...).

ShimiSun:
@IanKBovard can you please take a look at this one?

johnliu55tw:
I've found a simple solution: Use `React.memo`!

Actually before I post this, I've tried using `React.memo` on the children component `Entry` in the real code but it doesn't' work (performance is still bad). I decided to try it again in Codesandbox and now it's working!

https://codesandbox.io/s/dynamic-input-state-in-parent-with-pure-entry-7b2fs?file=/src/App.js
See the `PureEntry` component.

It turns out I have a `ThemeContext.Extend` wrapped around the parent component in the real code, which would cause **every child component to re-render even if it has `React.memo` wrapped around it**. Codesandbox is giving me errors while using `ThemeContext.Extend` so I exclude it in the example code, but this saves me ;)

I still need a custom theme, so I moved `ThemeContext.Extend` into the children component so it won't cause everything to re-render. It's working totally fine.

Let me know if I still need to keep this issue open, Thanks!

BTW I love Grommet! I'm just several months into frontend development and Grommet makes my web app looks like it's designed by a professional UI designer! Totally enjoy the development with Grommet :+1:

johnliu55tw:
[...]

PR #5848, opened 2021-12-15
Upgraded eslint and eslint-config-airbnb to latest major versions

halocline:
#### What does this PR do?

- Closes https://github.com/grommet/hpe-design-system/issues/2096
- Upgrade `eslint` and `eslint-config-airbnb` to latest major versions.

**NOTE:** The airbnb config introduced rules which were turned on by default which weren't present previously. This raised ~1,000 eslint errors. Each rule and associated remedies are listed out below and each is contained in a separate commit.

- **Dependency upgrade** ([commit](https://github.com/grommet/grommet/commit/a81697f8b1f9a42cd4faf79d6b13256a4a5cdd98)) --> 
    - Alphabetized rule set.
    - Disabled all new rules causing errors. Each rule was then addressed in the following commits.
    - Disabled `default-param-last` ([docs](https://eslint.org/docs/rules/default-param-last))
- **no-unsafe-optional-chaining** ([commit](https://github.com/grommet/grommet/commit/bb354fa2dbf56015c42e157d34839823bb82ec6e), [docs](https://eslint.org/docs/rules/no-unsafe-optional-chaining)) --> Enabled rule. Addressed each instance individually, either remedying or disabling for the instance.
- **prefer-regex-literals** ([commit](https://github.com/grommet/grommet/pull/5848/commits/eaf8bb357c7db8cdfe5e104a9bb444a7c24548d7), [docs](https://eslint.org/docs/rules/prefer-regex-literals)) --> Enabled rule. Remedied each instance.
- **react/jsx-no-constructed-context-values** ([commit](https://github.com/grommet/grommet/pull/5848/commits/73bae1df45f26cb0f84d009ab42a87322cbf7c60), [docs](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-constructed-context-values.md)) --> Enabled rule. Remedied each instance.
- **react/no-unstable-nested-components** ([commit](https://github.com/grommet/grommet/pull/5848/commits/d42974ecf93660289c75fe09f9c4cbab44ed94c3), [docs](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unstable-nested-components.md)) --> Enabled for Grommet codebase. Disabled for stories. Remedied codebase instance.

- package.json
- .eslintrc
- Each individual commit

All automated linter and unit tests pass.

N/A - no new functionality, no new tests.

https://github.com/grommet/hpe-design-system/issues/2096#issuecomment-987081797

https://github.com/grommet/hpe-design-system/issues/2096

No.

No.

Backwards compatible.


.eslintrc
```diff
@@ -62,7 +49,25 @@
           }
         ]
       }
-    ]
+    ],
+    "no-useless-concat": 0,
+    // react/function-component-definition affects 600+ instances. Temporarily disabled.
+    // ToDo: Enable gradually and apply component by component.
```
britt6612, line 55: also think we should discuss a little more if this is the correct direction we want to go.

src/js/components/Form/Form.js
```diff
@@ -571,7 +571,10 @@ const Form = forwardRef(
           }
         }}
       >
-        <FormContext.Provider value={{ useFormField, useFormInput }}>
+        <FormContext.Provider
+          // eslint-disable-next-line react/jsx-no-constructed-context-values
```
britt6612, line 575: should we add a TODO comment?

britt6612:
> react/no-unstable-nested-components (commit, docs) --> Enabled for Grommet codebase. Disabled for stories. Remedied codebase instance.

I like this approach there were a lot of changes with the stories so I wasn't sure if we wanted to change those since they were in the stories. Another good topic we can discuss but I agree with this approach of `Enabled for Grommet codebase and disable for stories`

Nice work just left a couple comments thanks!

.eslintrc
halocline: Agree.

src/js/components/Form/Form.js
halocline: This one is a quite a bit more involved. It also is linked to some potential performance wins which we have documented in an existing issue.

src/js/contexts/MessageContext/MessageContext.js
```diff
@@ -25,6 +25,7 @@ export const format = (options, messages) => {
   // their bundles that way and don't need to pass the messages
   // themselves in this property, just the format function.
   const idParts = options.id?.split('.') || [];
+  // eslint-disable-next-line no-unsafe-optional-chaining
   const baseId = idParts[idParts?.length - 1];
```
jcfilben, line 29: Any reason not to do `const baseId = idParts[(idParts?.length || 0) - 1];`

.eslintrc
jcfilben: I don't think this rule feels super important to go back and turn on since it is just defining a code style for writing function components in the project versus a rule that is preventing us from bad patterns. Since it is an open source project this rule feels a little too restrictive without providing much benefit. Definitely curious to hear other people's opinions though.

src/js/components/Form/Form.js
ericsoderberghp: I think changing it will improve performance.

Is this one really a lot more involved? Can we create the FormContext value via `useMemo`?

.eslintrc
halocline: Thanks for the input and perspective @jcfilben.

src/js/components/Grommet/Grommet.js
```diff
@@ -134,11 +140,12 @@ const Grommet = forwardRef((props, ref) => {
     theme.global.deviceBreakpoints.tablet;
 
   const grommetRef = useForwardedRef(ref);
+  const getRootsContext = useCallback((root) => [root], []);
```
ericsoderberghp, line 143: This one seems a bit odd as there would ideally be a way to have something that doesn't change across renders if grommetRef.current doesn't change, but we're going to call this function on every render here. So, we are going to be creating a new Array on each render. I think something that hits the spirit of this more would be to `const [rootsContextValue, setRootsContextValue] = useState()` and then call `setRootsContextValue` within the StyledGrommet `ref`, something like `ref={(node) => setRootsContextValue([node])}`.

src/js/components/Layer/LayerContainer.js
```diff
@@ -243,11 +243,16 @@ const LayerContainer = forwardRef(
       );
     }
 
-    if (theme.layer.background) {
+    const themeContext = useMemo(() => {
```
ericsoderberghp, line 246: I'm wondering if we should be clearer in naming this to be `themContextValue` since it isn't really a ThemeContext.

src/js/components/Form/Form.js
halocline: In theory, I do think changing it to comply with the eslint rule would likely lead to performance gains, but not enough to address other performance issues it has.

Today, if anything changes in the form's context, the entire Form and all children (FormFields, inputs) are all re-rendered. This is because the form's value and its validation status are important to all the field/inputs.

As example, a TextInput which is validated onChange will affect the value held by the entire form, will trigger form validation, and thus the form's overall context. All subcomponents dependent on that context are then re-rendered due to a single key stroke. Add lots of fields to the form, which are all getting re-rendered with every keystroke eventually becomes problematic.

Ideally, only the form and input being validated is re-rendered. I think to solve for this will require rearchitecting, possibly by splitting up the contexts. Refactoring the hooks driving the context will be quite invasive and I am not fully confident that it would address the "all children re-rendering" on each onChange.

Here's the [issue and some related discussion](https://github.com/grommet/grommet/issues/4559#issuecomment-727025196).

src/js/contexts/MessageContext/MessageContext.js
halocline: I debated this one and am open to persuasion. My thinking was that if we did that, we are then taking a negative index of our array and that becomes undefined which leaves us in pretty much the same situation, taking a NaN index of an array which is also undefined.

I kind of felt like adding the `||` provided a false sense of security without improving the situation. I opted for disabling the linter rule to at least serve as a 'caution flag' if we needed to re-visit this logic.

src/js/components/Form/Form.js
halocline: @ericsoderberghp - I will take a look again, but I do think it is more invasive. I'll double check.

.eslintrc
halocline: @MikeKingdom voiced similar thoughts, so it sounds like we have a consensus here.

I will update so that it is simply disabled.

src/js/contexts/MessageContext/MessageContext.js
jcfilben: Yeah that makes sense, I agree switching to `(idParts?.length || 0)` does provide a false sense of security and doesn't improve anything besides making lint happy. I'm good with leaving the rule disabled (or if it isn't too complicated maybe changing the logic so that if the idParts array is empty we just return to avoid an index out of bounds issue).

src/js/components/Layer/LayerContainer.js
halocline: Addressed.

src/js/components/Form/Form.js
halocline: I did revisit this and was able to memoize the FormContext value here. Additionally, I did some testing with React DevTools Profiler and found that:
1) Reduced the number of commits from 18 to 6.
2) Eliminated the re-rendering of all child inputs after the onChange event. Reduced from 27ms to 2ms.
3) Reduced total render time for equivalent form update from 40ms to 12ms.

[Gist with Profiler Data](https://gist.github.com/halocline/36060bbda2f1ae6ece6b9d35826c0d5f) - Upload to React Dev Tools Profiler - Generated from FireFox

**Before memo**
Screen Shot 2021-12-17 at 2 51 58 PM

**After memo**
Screen Shot 2021-12-17 at 2 52 33 PM

src/js/components/Grommet/Grommet.js
halocline: Addressed.

ericsoderberghp (approved):

Issue #5844, opened 2021-12-13
Calendar component chevron aria label not informative enough


snsw-vw:
Currently, if a user navigates the calendar component through a screen reader, the chevron in the calendar component only says the name of the next/prev month, instead of what the chevron actually does. Even though a 'messages' prop is provided, the label defined in that prop is only announced upon a user clicking the chevron, not when the user navigates to the chevron. Therefore, the functionality of the chevrons may not be clear enough to users using voice-over.

### Expected Behavior

The chevron labels defined in the 'messages' prop are used as aria-label when the user navigates to the chevron.
e.g. If the 'messages' prop is { next: go to {date}, prev: go to {date} }, the chevrons should have these label with or without user clicking them.

### Actual Behavior

The labels defined in the 'messages' prop is only used as the aria-label when the user clicks one of the chevrons.

### URL, screen shot, or Codepen exhibiting the issue

Currently the chevron label in static is rendered with aria-label of the month name even if the message is 'go to {date}'
<button aria-label="January 2022" type="button" class="StyledButton-sc-323bzc-0 iWlgRU"><svg aria-label="Next" viewBox="0 0 24 24" class="StyledIcon-sc-ofa7kd-0 dNHImd"><path fill="none" stroke="#000" stroke-width="2" d="m7 2 10 10L7 22"></path></svg></button>

### Steps to Reproduce

1. Render a calendar component with customized 'messages' prop
2. Turn on voice over and navigate to the calendar chevron
3. The voiceover will only announce that it is a button and the name of the next/prev month.

### Your Environment

- Grommet version: 2.17.5
- Browser Name and version: Microsoft Edge 96.0.1054.34
- Operating System and version (desktop or mobile): Mac OS Monterey 12.0.1

jcfilben:
Closing since #5849 was merged

PR #5849, opened 2021-12-15
Calendar component chevron aria label not informative enough 

matheushahnn:
#### What does this PR do?
It add aria-label in chevrons when messages is given 

Starts by Calendar

Add messages in Calendar on storybook 
`messages={{ next: 'go to {date}', previous: 'go to {date}', }}`
and run `yarn storybook`
turn on voice over and focus in chevrons to listen the message

https://github.com/grommet/grommet/issues/5844

Yes, it is

### DIFF
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -669,10 +669,23 @@ const Calendar = forwardRef(
           </Box>
           <Box flex={false} direction="row" align="center">
             <Button
-              a11yTitle={previousMonth.toLocaleDateString(locale, {
-                month: 'long',
-                year: 'numeric',
-              })}
+              a11yTitle={
+                messages?.previous
+                  ? format({
+                      id: 'calendar.previous',
+                      messages,
+                      values: {
+                        date: previousMonth.toLocaleDateString(locale, {
+                          month: 'long',
+                          year: 'numeric',
+                        }),
+                      },
+                    })
+                  : previousMonth.toLocaleDateString(locale, {
+                      month: 'long',
+                      year: 'numeric',
+                    })
+              }
               icon={<PreviousIcon size={size !== 'small' ? size : undefined} />}
               disabled={!betweenDates(previousMonth, bounds)}
               onClick={() => {
@@ -692,10 +705,23 @@ const Calendar = forwardRef(
               }}
             />
             <Button
-              a11yTitle={nextMonth.toLocaleDateString(locale, {
-                month: 'long',
-                year: 'numeric',
-              })}
+              a11yTitle={
+                messages?.next
+                  ? format({
+                      id: 'calendar.next',
+                      messages,
+                      values: {
+                        date: nextMonth.toLocaleDateString(locale, {
+                          month: 'long',
+                          year: 'numeric',
+                        }),
+                      },
+                    })
+                  : nextMonth.toLocaleDateString(locale, {
+                      month: 'long',
+                      year: 'numeric',
+                    })
+              }
               icon={<NextIcon size={size !== 'small' ? size : undefined} />}
               disabled={!betweenDates(nextMonth, bounds)}
               onClick={() => {
@@ -864,7 +890,7 @@ const Calendar = forwardRef(
                       onClick: () => {
                         selectDate(dateString);
                         announce(
-                          `Selected 
+                          `Selected
                           ${formatToLocalYYYYMMDD(dateString)}`,
                           'assertive',
                         );
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -55,7 +55,7 @@ const activeDates = {
 
 const timeStamp = /T.*/;
 
-const formatSelectedDatesString = (date) => `Currently selected
+const formatSelectedDatesString = (date) => `Currently selected 
   ${date?.map((item) => {
     let dates;
     if (!Array.isArray(item)) {
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -55,7 +55,7 @@ const activeDates = {
 
 const timeStamp = /T.*/;
 
-const formatSelectedDatesString = (date) => `Currently selected 
+const formatSelectedDatesString = (date) => `Currently selected
   ${date?.map((item) => {
     let dates;
     if (!Array.isArray(item)) {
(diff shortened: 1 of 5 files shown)


src/js/components/Calendar/Calendar.js
```diff
@@ -630,10 +630,23 @@ const Calendar = forwardRef(
...
-                month: 'long',
-                year: 'numeric',
-              })}
+              a11yTitle={
+                messages?.previous
+                  ? format({
+                      id: 'calendar.previous',
+                      messages,
+                      values: {
+                        date: previousMonth.toLocaleDateString(locale, {
```
taysea, line 639: Seeing this block repeated twice, wondering if there is an opportunity to pull this out as a variable and reference it in both places. Soft opinion here, would also be fine just leaving as is.

```
previousMonth.toLocaleDateString(locale, {
    month: 'long',
    year: 'numeric',
})
```

src/js/components/Calendar/Calendar.js
```diff
@@ -630,10 +630,23 @@ const Calendar = forwardRef(
           </Box>
           <Box flex={false} direction="row" align="center">
             <Button
-              a11yTitle={previousMonth.toLocaleDateString(locale, {
-                month: 'long',
-                year: 'numeric',
-              })}
+              a11yTitle={
+                messages?.previous
```
MikeKingdom, line 634: this shouldn't be checking messages passed it. It should be letting format figure it out. Ie. if messages wasn't passed in, it should still be getting it from the global messages.

src/js/components/Calendar/Calendar.js
```diff
@@ -653,10 +666,23 @@ const Calendar = forwardRef(
               }}
             />
             <Button
-              a11yTitle={nextMonth.toLocaleDateString(locale, {
-                month: 'long',
-                year: 'numeric',
-              })}
+              a11yTitle={
+                messages?.next
```
MikeKingdom, line 670: same

src/js/components/Calendar/Calendar.js
matheushahnn: So, that is why I check if messages.previous was passed.
If the Calendar is rendered without messages it will use the default message in the else condition: `previousMonth.toLocaleDateString(locale, {
                      month: 'long',
                      year: 'numeric',
                    })`  which is the current master behavior.
If I set messages.next and messages.previous the aria-label keeps in the default e.g.: "Moved to November 2021", so when we set the prop it will use the messages as aria-label

src/js/components/Calendar/Calendar.js
matheushahnn: If I keep this method it won't update the aria-label with the messages given.
I thought to create this method:
`const getChevronAriaLabel = (type) => {
      let month = nextMonth;
      if (type === 'previous') {
        month = previousMonth;
      }
      if (messages && messages[type]) {
        return format({
          id: `calendar.${type}`,
          messages,
          values: {
            date: month.toLocaleDateString(locale, {
              month: 'long',
              year: 'numeric',
            }),
          },
        });
      }

      return month.toLocaleDateString(locale, {
        month: 'long',
        year: 'numeric',
      });
    };`
    
What do you think? Type should be equals to 'next'or 'previous'

src/js/components/Calendar/Calendar.js
MikeKingdom: Is there a reason you don't want the default to have the 'Moved to' text? That seems to be a better default than what it used to be.

matheushahnn: I'm not sure if I get yor point.
If you don't set the messages prop it will keep the 'Moved to'
If you set the messages the aria-label will be what you set. 
So when you turn on voice over and focus in the chevron you will heard the same message when you click.
The current master changes the onclick event but not the focus.

MikeKingdom: > If you don't set the messages prop it will keep the 'Moved to'

Only for the announce during onClick. The aria-label will still be just the date.

> If you set the messages the aria-label will be what you set. So when you turn on voice over and focus in the chevron you will heard the same message when you click. 

Which is great but I'd think that's also what we want when the messages prop isn't set so apps which aren't passing messages to Calendar itself get the better experience.

taysea: If I'm following this correctly, I think we can just do the following:
```
a11yTitle={
   format({
       id: 'calendar.previous',
        messages,
        values: {
           date: previousMonth.toLocaleDateString(locale, {
               month: 'long',
               year: 'numeric',
           }),
        },
    })
}
```

`format` function will be able to determine if the caller has passed a custom `message` or if we should rely on the default.

matheushahnn: It worked! Thank you all 😀

src/js/components/Calendar/Calendar.js
```diff
@@ -55,7 +55,7 @@ const activeDates = {
 
 const timeStamp = new RegExp(/T.*/);
 
-const formatSelectedDatesString = (date) => `Currently selected
```
matheushahnn, line 58: I will undo this change

src/js/components/Calendar/Calendar.js
```diff
@@ -692,9 +698,15 @@ const Calendar = forwardRef(
               }}
             />
             <Button
-              a11yTitle={nextMonth.toLocaleDateString(locale, {
-                month: 'long',
-                year: 'numeric',
+              a11yTitle={format({
+                id: 'calendar.next',
```
amandadupell, line 702: this id is specific to the onClick button announcement to a user that the calendar UI has moved to the next month. with your current implementation, if a user does not provide a messages prop, the button label is "Moved to {date}" and the user is no longer accurately informed when the UI changes when they click the button. this does not align with the expected screen reader experience. i'm proposing the following: 

- change the `default.json` language file to include new id values for the calendar:
````
"calendar": {
    "previousMove": "Moved to {date}",
    "previous": "Go to {date}",
    "nextMove": "Moved to {date}",
    "next": "Go to {date}"
},
````
- update the onClick announce format id to use `'calendar.nextMove'` instead of `'calendar.next'`
- update the onClick announce format id to use `'calendar.previousMove'` instead of `'calendar.previous'`

with these changes, the user experience is this:
- when navigating to the previous arrow button, the screen reader will read "Go to [previous month]"
- when navigating to the next arrow button, the screen reader will read "Go to [next month]"
- when the user presses the previous arrow button, the screen reader will announce "Moved to [previous month]"
- when the user presses the next arrow button, the screen reader will announce "Moved to [next month]"

let me know if you have questions, thank you!

amandadupell (changes requested):
one comment about aligning behavior better with screen reader experience

amandadupell (approved):
looks good to me now! thank you!

ericsoderberghp:
Thanks for your contribution!

snsw-vw:
Thanks folks!

PR #5856, opened 2021-12-17
Change button icon alignment for small icons

jcfilben:
#### What does this PR do?
Fixes an alignment issue when using small Icons within Buttons. Previously all icons in buttons had the css prop `vertical-align: bottom`. We need this prop for medium and large size icons to ensure Buttons with just an icon don't add additional vertical height internally. However this is not needed with small buttons and it causes the icon to look off center.

Previously:
Screen Shot 2021-12-17 at 3 39 24 PM

What Icons look like with the fix in this PR:
Screen Shot 2021-12-17 at 3 39 08 PM

Tested with the following story as well as looking at other Button stories and DateInput stories with the hpe theme.
```javascript
import React from 'react';
import { Close } from 'grommet-icons';

import { Box, Button, Text } from 'grommet';

export const Test = () => (
  <Box align="center" pad="large" gap="medium">
    <Text>Button with small icon</Text>
    <Button icon={<Close size="small" />} onClick={() => {}} primary />
    <Text>Button with medium icon</Text>
    <Button icon={<Close size="medium" />} onClick={() => {}} primary />
  </Box>
);

export default {
  title: `Controls/Button/Test`,
};
```

This issue became much more noticeable after the recent DateInput updates. The hpe theme uses the small icon for DateInput which resulted in the icon looking vertically off center.

Screen Shot 2021-12-17 at 4 00 03 PM

No

Yes "Fixed vertical alignment of small icons in Buttons"

Backwards compatible


src/js/components/Button/StyledButtonKind.js
```diff
@@ -80,13 +80,16 @@ const basicStyle = (props) => css`
...
-  ${props.badge
-    ? `
-  svg {
+  ${(props.badge &&
+    `svg {
     vertical-align: bottom;
-  }`
-    : `> svg {
+  }`) ||
+  (props.icon?.props?.size !== 'small' &&
```
ericsoderberghp, line 87: This feels a bit fragile. The icon could have `size="13px"`.

One idea: `useLayoutEffect` to measure the actual height of the icon after rendering and adjust as needed.

Another idea: change the `<button />` to use `display: flex` and align items to the center.

I'm trying to think if there's a way to manage the layout without `vertical-align` and needing to have CSS selectors  for child elements.

jcfilben: Sounds good, looking into it...

ericsoderberghp (approved):

Issue #5859, opened 2021-12-20
DateInput in FormField Gets Double Focus


jcfilben:
### Expected Behavior
The FormField should get focus and then if 'tab' is pressed focus should be removed from the FormField and placed on the Calendar Button.

### Actual Behavior
The FormField gets focus. When 'tab' is pressed the Calendar Button gets focus in addition to the FormField.

### URL, screen shot, or Codepen exhibiting the issue
<img width="273" alt="Screen Shot 2021-12-20 at 10 12 28 AM" src="https://user-images.githubusercontent.com/54560994/146806212-c5e66b4b-c9e4-4c60-b4c4-9dd4435d468c.png">

### Steps to Reproduce
https://storybook.grommet.io/?path=/story/input-dateinput-form--date-form&globals=theme:hpe
1. 'tab' to the DateInput FormField
2. hit 'tab' again and you will see the double focus

### Your Environment
- Grommet version: stable branch

PR #5862, opened 2021-12-21
Fix: DateInput in FormField Gets Double Focus

matheushahnn:
#### What does this PR do?
When changed the focus inside the FormField it will check if the focus changed, if the focus continues in FormField and the focus is in a button it will remove the focus in the FormField Input, it will apply the focus again otherwise (I try to find why it happens and only find at Input/CheckBoxGroup)

Start for FormField.js

1. Open storybook and go to Inout/DateInput/Form
2. Focus in the input and press "tab" again to change the focus to calendar icon.

https://github.com/grommet/grommet/issues/5859

image

Yes, it is


jcfilben (commented):
This fixes the DateInput issue but it causes Select not to get focus when it is in a FormField because select is a button. I think `event.target.type === 'button'` is a bit too broad.

matheushahnn:
> This fixes the DateInput issue but it causes Select not to get focus when it is in a FormField because select is a button. I think `event.target.type === 'button'` is a bit too broad.

You're right, sorry about that. I tried something different now, what do you think?

jcfilben (approved):
Nice solution, this looks good to me!

src/js/utils/DOM.js
```diff
@@ -64,18 +64,28 @@ export const containsFocus = node => {
   return !!element;
 };
 
-export const getFirstFocusableDescendant = element => {
+export const isFocusable = (element) => {
+  const tagName = element.tagName.toLowerCase();
+  return tagName === 'input' || tagName === 'select';
```
ericsoderberghp, line 69: This feels a bit fragile to me. For instance, it doesn't cover `<textarea />`.

matheushahnn: I agree, I just got this logic from the getFirstFocusableDescendant and split in a different method.
I can add `<textarea />` do you think need to create for some different element? 
Or I can create a different method, like:
```
export const checkTagName = (element, tagList=["input", "select", "textarea"]) => {
  const tagName = element.tagName.toLowerCase();
  return tagList.includes(tagName);
})
```
What do you think?

ericsoderberghp: Ah, I didn't realize this was just being copied. I'm OK with just adding "textarea" for this pull request.

src/js/utils/DOM.js
```diff
@@ -64,18 +64,28 @@ export const containsFocus = node => {
...
   for (let i = 0; i < children.length; i += 1) {
     const child = children[i];
-    const tagName = child.tagName.toLowerCase();
-    if (tagName === 'input' || tagName === 'select') {
+    if (isFocusable(child)) {
       return child;
     }
   }
   return undefined;
 };
```
ericsoderberghp, line 82: Can you add a comment above this function explaining its use and basic logic? I'm not thinking something elaborate, just a sentence or two.

src/js/utils/DOM.js
```diff
@@ -64,18 +64,38 @@ export const containsFocus = node => {
   return !!element;
 };
 
-export const getFirstFocusableDescendant = element => {
+/**
+ * Check if the element.tagName is an input, select or textarea
+ * @param {Element} element
+ * @returns boolean
+ */
```
jcfilben, line 71:
```suggestion
// Check if the element.tagName is an input, select or textarea
```
I think we can just have this as a normal comment since we don't use JSDoc style comments anywhere else

src/js/utils/DOM.js
```diff
@@ -64,18 +64,38 @@ export const containsFocus = node => {
...
+export const isFocusable = (element) => {
+  const tagName = element.tagName.toLowerCase();
+  return tagName === 'input' || tagName === 'select' || tagName === 'textarea';
+};
+
+/**
+ * Get the first element that can receive focus
+ * @param {Element} element
+ * @returns Element | undefined
+ */
```
jcfilben, line 81:
```suggestion
// Get the first element that can receive focus
```

jcfilben (approved):

ericsoderberghp (approved):

Issue #5865, opened 2021-12-22
In the Select Component dropdown the search text is persisting after close dropdown option and again open the dropdown


matheushahnn:
If I search some text and click out of the option the search keeps the text but the filter doesn't work.

### Expected Behavior
The Select should either clear the search text when the drop closes or it should let the text persist and filter results.

### Actual Behavior
The search is not cleared and the options are not filtered

### URL, screen shot, or Codepen exhibiting the issue

URL: https://codesandbox.io/s/grommet-v2-template-forked-wgs3l?file=/index.js

<img width="600px" src="https://user-images.githubusercontent.com/7703431/147092103-ddb438b2-4864-41c3-98fd-8e38b83073b4.png" />

<img width="600px" src="https://user-images.githubusercontent.com/7703431/147092162-a310f57a-0a04-4f1a-a433-b49566e74e75.png" />

<img width="600px" src="https://user-images.githubusercontent.com/7703431/147092125-dc3f33ac-7e70-4bea-a922-d571a0101f3a.png" />

### Steps to Reproduce

1. Click in the select and search by one 
2. Click out of options, it will close the dropdown
3. Open the dropdown again

### Your Environment
- Grommet version: 2.16.2
- Browser Name and version: chrome
- Operating System and version (desktop or mobile): desktop

This issue was discussed in https://github.com/grommet/grommet/issues/5830

PR #5866, opened 2021-12-22
fix(Select): clean search field when close dropdown

matheushahnn:
#### What does this PR do?
It cleans search field in select component when closes without selection any option

Start on Select component 

1. Start storybook
2. Open Input/Select/Search
3. Open the first dropdown and search by option 3
4. Click out of dropdown, it should close
5. Open again the dropdown and the search field must be empty

https://github.com/grommet/grommet/issues/5865

image
image

Yes, it is.


src/js/components/Select/Select.js
```diff
@@ -172,7 +172,8 @@ const Select = forwardRef(
     const onRequestClose = useCallback(() => {
       setOpen(false);
       if (onClose) onClose();
-    }, [onClose]);
+      setSearch();
+    }, [onClose, setSearch]);
```
jcfilben, line 176:
```suggestion
    }, [onClose]);
```
I don't think we need to include setSearch in the dependencies array since the setSearch itself isn't changing

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

PR #5869, opened 2021-12-23
Fix: TextInput cursor jumping to end

### LINKED ISSUE #5682: TextInput loses caret position inside Tab controls
### Expected Behavior

If you move the caret to the center of a text string and type, the caret should remain at the position of the data entry.

### Actual Behavior

If you move the caret to the center of a text string and type, the new character is inserted correctly but the caret jumps to the end of the string.

https://codesandbox.io/s/grommet-v2-template-forked-zssn2?file=/index.js

### Steps to Reproduce

1.  Enter text in a textInput that is inside a Tab control
2. Move the caret to the middle of the string
3. Enter additional characters

### Your Environment
- Grommet version: 2.18

jlaluces123:
#### What does this PR do?
This PR aims to solve the issue of TextInput's cursor jumping to the end when the cursor was not positioned there.

Changes made in `TextInput.js`

In issue #5682, I made a new story using the submitter's example. With the new changes in place the behavior described in the issue is no longer prevalent (cursor jumping to end of the text)

- Make a new storybook using the submitter's codesandbox (https://codesandbox.io/s/grommet-v2-template-forked-zssn2?file=/index.js)
- Type into TextInput
- Move cursor around what you typed and then try typing again
- The cursor should no longer jump to the end of the string you typed

Resolves #5682

No

Possibly as small bug fix to TextInput

Backwards Compatible

### DIFF
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -116,6 +116,7 @@ const TextInput = forwardRef(
 
     const [focus, setFocus] = useState();
     const [showDrop, setShowDrop] = useState(false);
+    const [cursor, setCursor] = useState(null);
 
     const handleSuggestionSelect = useMemo(
       () => (onSelect && !onSuggestionSelect ? onSelect : onSuggestionSelect),
@@ -176,6 +177,12 @@ const TextInput = forwardRef(
         closeDrop();
       }
     }, [closeDrop, showDrop, suggestions]);
+
+    useEffect(() => {
+      // log inputRef's input type to console
+      if (inputRef.current?.type === 'text')
+        inputRef.current.setSelectionRange(cursor, cursor);
+    }, [inputRef, cursor, value]);
 
     const valueSuggestionIndex = useMemo(
       () =>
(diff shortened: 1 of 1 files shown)


jcfilben:
Looking into this more... still trying to understand how the cursor issue is related to having a custom theme

jcfilben:
While this does solve the issue, it feels like more of a workaround than solving the root of the problem (Note that this issue is also happening with TextArea). The TextInput cursor works correctly when we have it inside of a Tab but don't wrap with Grommet. It works correctly when wrapped with Grommet if we don't have a custom theme. It also works if the TextInput or TextArea doesn't have `value` or `onChange` set. I think this issue might require additional investigation into why the interaction between a custom theme and a TextInput or TextArea within a Tab causes the cursor to behave differently. Once we understand more about the issue we can determine if the root of the issue is solvable or if we need to have code within the TextInput and TextArea components to control the cursor behavior.

jcfilben:
Closing since I think we need to do a little more investigation on the root cause of the issue

MikeKingdom:
Noted in #5682 also, but this seems to be because the theme object is technically a different object every render. The fix is to `useMemo` the theme object or move it outside all renders.

Issue #5854, opened 2021-12-17
Use better `key` for `List` with custom renderer functions?


Zodiase:
### Expected Behavior

I'm hoping I could specify `key` values without affecting rendering.

This is not the API I'd expect but just to illustrate one possibility and explain what I'm talking about:
```
<List primaryKey="_id">
    {(item) => <Box>{item.title}</Box>}
</List>
```
With this hopefully the rendered React Virtual DOM would be:
```
<li key="item-id-1">
    <Box>Item title 1</Box>
</li>
<li key="item-id-2">
    <Box>Item title 2</Box>
</li>
```

### Actual Behavior

I could use just `primaryKey` with an attribute name, which would use the attribute value as `key`. But it poses a couple problems:

1. If I use `primaryKey="_id"`, the `key` will be unique, but it will also be rendered. I don't need the `_id` to be rendered.
2. If I use `primaryKey="title"`, "title" being the attribute I'm trying to render, "title" is not guaranteed to be unique so can't be used as `key`.

I could also use the custom renderers but it seems in that case the `List` will always use `index` as the `key`, which is not ideal.

### URL, screen shot, or Codepen exhibiting the issue

![image](https://user-images.githubusercontent.com/1815174/146513111-c969512d-0600-47b2-838e-e1d0764c8341.png)

### Steps to Reproduce

1. Use `List` with a custom renderer.
2. Try to use some ID attribute as the `key`.
3. See the `key` props of the list items are always the index numbers.

### Your Environment

- Grommet version: 2.19.1
- Browser Name and version: N/A
- Operating System and version (desktop or mobile): N/A

jcfilben:
HI @Zodiase, not sure if I am following what you are trying to do here. The documentation for the `primaryKey` prop says, "When a string is supplied, it indicates the property in a data item object to use to get the primary content. If a function is supplied, it will be called with the current data item object and should return a React element that will be rendered as the primary content.". The `primaryKey` and `secondaryKey` should both be used to for the "primary content" and "secondary content" being displayed in your list. You can take a look at this [List/SecondaryKey](https://storybook.grommet.io/?path=/story/visualizations-list-secondary-key--secondary-key) story to see an example.

It sounds like you are wanting to attach a `key` prop to each item in the list, `primaryKey` shouldn't be used for this. Instead in your render function for each item in the list you can attach a key there based on the index.

Zodiase:
@jcfilben 

[...]

PR #5872, opened 2022-01-02
[Enhance] List component to specify key for the List items

g4rry420:

#### What does this PR do?
I have enhanced the List component so that the users can specify `key` to the List Item. I have utilised the `primaryKey` prop for this.

components => List

Pre-written has been passed successfully

Using storybook

Didn't write any tests for it.

I see two ways to resolve this issue:
1. To add new prop so that users can specify propertyName from `data` object for the key or maybe they can add func to this prop so that it geneartes key on their end.
2. To enhance the existing prop - `primaryKey` so that it can somehow recognize the propertyName from `data` object.

I have choosen the 2nd option for this enhancement.
These are potential cases of what the key will and with what conditions.

Suppose, we have data object of:
```
const data = [
  { id: 1, location: 'Boise'},
  { id: 2, location: 'Fort Collins'},
  { id: 3, location: 'Los Gatos'},
  { id: 4, location: 'Palo Alto'},
  { id: 5, location: 'San Francisco'},
];
```

1. when there is function added to `primaryKey` prop
    Then, `content` will be generated from that func and key will be same that is given to returned element of func.
    For eg, if our primaryKey func is:
```
  const primaryFunc = (item) => {
    return <Box key={item.id}>{item.location}</Box>
  };
```
The `key` for List Item will be same given to returned element of func. ie, `key={item.id}`.

2. when the primaryKey = "location" prop
   then the content = `location` property and the key = `location` property.

3.  when the primaryKey = "location, id" prop
    then the content = `location` property and key = `id` property (given that these are properties do exist in data object)

4. when there is children but no primaryKey prop
     then its default generated by `getKey` func ie, index values from 0 to n (n - total number of items in `data` array)

5. when there is children and the primaryKey = 'id'
	then key will be property described in primaryKey, ie, `id`

Above mentioned cases works for the Array of objects but if there is array of simple strings such as:
```
const data = [
  'Boise',
  'Fort Collins',
  'Los Gatos',
  'Palo Alto',
  'San Francisco',
];
```
Then, the key will be same as the value of the array.
I think, this area needs a improvement and in my opinion, in this case, we should let users specify the function for this.

Nevertheless, I am open to any new suggestions and feedback. Please, let me know your thoughts.

#5854 

Yes

Backwards compatible


src/js/components/List/List.js
```diff
@@ -277,16 +277,40 @@ const List = React.forwardRef(
...
                   );
                 } else if (primaryKey) {
                   if (typeof primaryKey === 'function') {
-                    itemId = primaryKey(item, index);
-                    content = itemId;
+                    content = primaryKey(item, index);
+                    if (!content?.key)
+                      console.error(
+                        `You must add key to your element in primaryKey func.`,
+                      );
```
jcfilben, line 299: I think in the case where there is a primaryKey function but the user doesn't specify a key in the function it shouldn't result in an error and instead the key should default to index values from 0 to n

src/js/components/List/List.js
```diff
@@ -277,16 +277,40 @@ const List = React.forwardRef(
...
+                    itemId = content.key;
+                  } else if (
+                    primaryKey.match(/^(.+),(.+)$/) &&
+                    primaryKey.split(',').length === 2
+                  ) {
```
jcfilben, line 304: I'm still thinking about this section. I'm questioning if the pattern of having `primaryKey="displayValue, id"` is intuitive for users. We are deviating from the original purpose of primaryKey, I'm wondering if it would make more sense to have a separate prop for key instead of expanding the primaryKey prop to have multiple purposes and only use primaryKey as the key if a key value is not specifically given through the `key` prop.

g4rry420: I agree with your points. Thank you for suggesting them. I will make the changes. What do you suggest should be the name of the prop ?  Maybe `listKey` ?

ericsoderberghp: I'm wondering if this is an infrequent enough scenario that we can just rely on the primaryKey as function approach?

jcfilben: I’m thinking we should have a prop `itemKey`. I was leaning towards itemKey over listKey to convey to the user that the key is being applied to each item in the list not something that is being applied to the list as a whole. `itemKey` could be a string or a function.

Example data: 
const data = [
  { id: 1, location: 'Boise'},
  { id: 2, location: 'Fort Collins'},
  { id: 3, location: 'Los Gatos'},
  { id: 4, location: 'Palo Alto'},
  { id: 5, location: 'San Francisco'},
];

1. The user sets `itemKey` to a string `itemKey=“id”` 
The key should be equal to the id property for each item (ex 1, 2, 3…)
2. The `itemKey` prop is a function `itemKey={(item) => (key-${item.id})}` 
The key should be “key-<id property>” (ex. key-1, key-2, key-3…)
3. The user doesn’t set the `itemKey` prop but they do set `primaryKey`
 The `primaryKey` property should be used for the key (I believe this is how the key is currently determined)
4. The user doesn’t set `itemKey` or `primaryKey`
 We should default to the current behavior of using the index for the key (ex 0, 1, 2…)

@ericsoderberghp any thoughts on this direction?

jcfilben: Oops I didn't see your comment above before I posted mine @ericsoderberghp

jcfilben: I agree, I think we could just support setting the key within a `primaryKey` function and not support setting the key like this: `primaryKey="displayValue, id"`. Then we wouldn't need to add an additional prop.

g4rry420: If we are going through the `primarKey` prop way, I thought this way `primaryKey="displayValue, id"`, works well, but if you guys have a better way, I am happy to implement that.

jcfilben: I think we can drop support for `primaryKey="displayValue, id"` and just allow users to set the key using a `primaryKey` function ex. `primaryKey={(item) => (<Box key={item.id}>{item.location}</Box>)}`

ericsoderberghp: Hmm... I don't think we can use a React element as a `key` property: https://reactjs.org/docs/lists-and-keys.html#keys
Given that, and at the risk of being harangued with "I told you so" comments, I'm going back to @jcfilben 's excellent suggestion of adding an optional `itemKey` property that is optional and can be a string or a function as she outlined. Apologies.

src/js/components/List/List.js
```diff
@@ -319,7 +320,15 @@ const List = React.forwardRef(
...
-                const key = getKey(item, index, itemId);
+                if (itemKey) {
+                  if (typeof itemKey === 'function') {
+                    itemId = itemKey(item);
+                  } else {
+                    itemId = normalize(item, index, itemKey);
+                  }
+                }
+
+                const key = itemId || getKey(item, index, itemId);
```
jcfilben, line 331:
```suggestion
                const key = itemKey ? itemId : getKey(item, index, itemId);
```
I think we should change this so that we only set key to itemId when we have an `itemKey` defined. Otherwise we get an error in the case where we don't have an `itemKey` but we have a `primaryKey` that is a function and it tries to set the key to the `primaryKey` function.

jcfilben (commented):
This is coming along nicely!
A few things:
Can you add a jest test that tests `itemKey`?
Also can you add `itemKey` to the docs on [grommet-site](https://github.com/grommet/grommet-site)?

src/js/components/List/__tests__/List-test.js
```diff
@@ -527,4 +527,36 @@ describe('List onOrder', () => {
...
+          data={[
+            { a: 'one', b: 1 },
+            { a: 'two', b: 2 },
+          ]}
+          primaryKey="a"
+          itemKey="b"
+        />
+      </Grommet>,
+    );
+    expect(container.firstChild).toMatchSnapshot();
```
jcfilben, line 544: The snapshots don't show the key. Can we change the tests to just check what key got applied to the list item

g4rry420: Correct me if I am wrong, after researching, I don't think there is a good way to test which key got applied to the list.
https://stackoverflow.com/questions/69758600/how-can-i-assert-which-value-has-been-passed-to-a-li-key-attribute-in-react-wi

However, if we can specify `id={key}` prop to the `StyledItem`, then it might be possible to capture the value of key in the snapshots. Please let me know your thoughts.

jcfilben: I took some time to think about this more and I agree that there isn't a good way to test this. While we could do something to be able to access and test the key prop I am thinking that testing for specific prop values isn't something we want to be doing in the first place. I found this great article on why this type of testing isn't a good practice: https://kentcdodds.com/blog/testing-implementation-details.

Anyway sorry for the back and forth on this. I think we can remove the tests and then this PR will be ready to go!

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

Issue #5873, opened 2022-01-02
theme documentation: invalid attribute name for toast time configuration


lausek:
The documentation for notifications states that the attribute `notification.toast.time` must be changed in order to alter the toast timeout. [Docs](https://v2.grommet.io/notification#notification.toast.time)
However, the actual attribute name is `notification.time`.

https://github.com/grommet/grommet/blob/690cb49abc22031bd80913008b44624d632c0f77/src/js/components/Notification/Notification.js#L23

### Your Environment

- Grommet version: 2.19.1

PR #5874, opened 2022-01-04
Fix/toast-time-prop5873: updated notifications code to make time prop specific to toast

amandadupell:
#### What does this PR do?

This PR updates the code to match the Grommet documentation for the Notification component. The time prop is specific to the toast notification. This could change as more types of notifications are introduced. Updated in the component, `base.js` and `base.d.ts`. 

`Notification.js`

Updated tests to match best testing protocols.

View the toast notification story in Storybook

- [X] `userEvent` is used in place of `fireEvent`.

#5873 

No

No

Backwards compatible

### DIFF
--- a/src/js/components/Notification/Notification.js
+++ b/src/js/components/Notification/Notification.js
@@ -20,10 +20,13 @@ const Notification = ({ message, onClose, status, title, toast }) => {
   }, [onClose]);
 
   useEffect(() => {
-    const timer = setTimeout(close, theme.notification.toast.time);
+    const timer = setTimeout(
+      close,
+      theme.notification.toast.time || theme.notification.time,
+    );
 
     return () => clearTimeout(timer);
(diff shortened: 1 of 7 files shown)


jcfilben (approved):
Looks good!

src/js/components/Notification/Notification.js
```diff
@@ -20,10 +20,10 @@ const Notification = ({ message, onClose, status, title, toast }) => {
   }, [onClose]);
 
   useEffect(() => {
-    const timer = setTimeout(close, theme.notification.time);
+    const timer = setTimeout(close, theme.notification.toast.time);
```
ericsoderberghp, line 23: This would be a breaking change. I think we need to change it to be backwards compatible.
```suggestion
    const timer = setTimeout(close, theme.notification.toast.time || theme.notification.time);
```

ericsoderberghp (approved):

PR #5875, opened 2022-01-05
fix when empty string is given for value for calendar

britt6612:
#### What does this PR do?
This PR checks for if someone is passing an empty string for value in `dateinput`
DateInput.js
simple story in DateInput

The `timeStamp` function was checking if a string was provided however if a user passed in an empty string then the `timeStamp` should be `undefined` 
This was blocking the release currently the simple story was showing the wrong date when clicked 
no
no
backwards compatible


jcfilben (commented):
This did fix the DateInput/Simple story but I am still seeing the incorrect date issue on the DropButton/Calendar story

ericsoderberghp (approved):

britt6612:
> This did fix the DateInput/Simple story but I am still seeing the incorrect date issue on the DropButton/Calendar story

yes I need to add to `calendar` as well pushing up now

PR #5876, opened 2022-01-05
fix when empty string in Calendar

britt6612:
#### What does this PR do?
This PR checks for if someone is passing an empty string for value in `date` in Calendar
Calendar.js
dropdown calendar story

The `timeStamp` function was checking if a string was provided however if a user passed in an empty string then the `timeStamp` should be `undefined` 
This was blocking the release currently the simple story was showing the wrong date when clicked 
no
no
backwards compatible


britt6612:
same solution in DateInput we just also need in Calendar

taysea (approved):

jcfilben (approved):

ericsoderberghp (approved):

PR #5885, opened 2022-01-11
Fix icon styling for plain buttons

### LINKED ISSUE #5884: Button styling is causing changes to button dimensions / icon alignment
### Expected Behavior

Icon should be vertically centered in a Button and the Button should retain its dimensions regardless of what theme is used.

### Actual Behavior

It seems that https://github.com/grommet/grommet/pull/5856 introduced some issues with icon alignment in Buttons (as can be seen in HPE themed Tag with onRemove: https://storybook.grommet.io/?path=/story/type-tag-onremove--on-remove&globals=theme:hpe).

https://storybook.grommet.io/?path=/story/type-tag-onremove--on-remove&globals=theme:hpe

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: Stable
- Browser Name and version: Chrome Version 97.0.4692.71 (Official Build) (x86_64)

jcfilben:
#### What does this PR do?
This PR fixes an issue with the alignment of icons in plain buttons.

storybook

View the Tag/onRemove story with the hpe theme

Closes #5884

Before:
Screen Shot 2022-01-10 at 5 22 20 PM

After:
Screen Shot 2022-01-12 at 11 38 20 AM

No

Yes

Backwards compatible

### DIFF
--- a/src/js/components/Button/StyledButtonKind.js
+++ b/src/js/components/Button/StyledButtonKind.js
@@ -204,12 +204,19 @@ const fillStyle = (fillContainer) => {
   return undefined;
 };
 
-const plainStyle = css`
+const plainStyle = (props) => css`
   outline: none;
   border: none;
   padding: 0;
   text-align: inherit;
   color: inherit;
+  ${props.icon &&
+  `
+    > svg {
+      display: flex;
+      align-self: center;
+    }
+  `}
 `;
 
 const StyledButtonKind = styled.button.withConfig({
@@ -229,7 +236,7 @@ const StyledButtonKind = styled.button.withConfig({
   text-transform: none;
 
   ${genericStyles}
-  ${(props) => props.plain && plainStyle}
+  ${(props) => props.plain && plainStyle(props)}
   // set baseline activeStyle for all buttons including plain buttons
   // buttons with kind will have active styling overridden by kindStyle
   // if they have specific state styles


src/js/components/Button/StyledButtonKind.js
```diff
@@ -204,12 +204,19 @@ const fillStyle = (fillContainer) => {
...
 };
 
-const plainStyle = css`
+const plainStyle = (props) => css`
   outline: none;
   border: none;
   padding: 0;
   text-align: inherit;
   color: inherit;
+  ${props.icon &&
```
ericsoderberghp, line 213: hmm... this feels like the opposite of "plain", where we tend to turn off styling. What if the caller wants to create a custom kind of badge that is an SVG styled to be in the upper right corner?

taysea: I'm not following this use case so much -- previously we had svg styling in `plainStyle` that mimicked what happened in `basicStyle` which we lost in the recent changes. I would expect this change to restore the styling we had previously supported

ericsoderberghp: Ah, interesting. OK.

taysea: We had it implemented differently (using vertical-align bottom as opposed to display flex) but that styling did exist in both `plainStyle` and `basicStyle` -- we can discuss more later in the week if you're not feeling like this is the right approach

taysea (commented):
I would've expected some snapshot changes here if we had a test that covered this scenario. Do you think you could add a test for this case?

ericsoderberghp (approved):

jcfilben:
> I would've expected some snapshot changes here if we had a test that covered this scenario. Do you think you could add a test for this case?

Yes I will go back and add a test for this. 

I was also thinking it might be helpful to have some chromatic snapshots with the hpe theme since some of the issues we have been seeing lately were not noticeable in the grommet theme but caused issues with the hpe theme

PR #5890, opened 2022-01-12
Fix double focus for Select with search in FormField

### LINKED ISSUE #5887: FormField with Select + Search receives double focus
### Expected Behavior

When Select search input is active element, it should be the only element with focus indicator.

### Actual Behavior

For certain themes (such as HPE), focus indicator is being shown on FormField and Select search input.

double-focus

https://design-system.hpe.design/components/select?q=select#select-with-search

### Steps to Reproduce

1. Open Search
2. See double focus.

### Your Environment

- Grommet version: Stable
- Browser Name and version: Chrome Version 97.0.4692.71

jcfilben:
#### What does this PR do?
This fixes an issue where Select would receive double focus when it had search capability and was contained in a Form. This issue was introduced in https://github.com/grommet/grommet/pull/5862 in an effort to fix a double focus issue on DateInput. I changed FormField back to the way it was handling focus before #5826 and added a case to handle DateInput focus separately. 

Tested with the following story:
```javascript
import { FileInput, Form, FormField, Select, DateInput } from 'grommet'

const defaultOptions = ['EMEA', 'Asia/Pacific', 'Americas', 'Polar Regions'];
// Escaping regular expression special characters: [ \ ^ $ . | ? * + ( )
const getEscapedText = (text) => text.replace(/[-\\^$*+?.()|[\]{}]/g, '\\$&');
// Create the regular expression with escaped special characters.
const formatSearchExpression = (text) => new RegExp(getEscapedText(text), 'i');

export const Test = () => {
  const [selected, setSelected] = useState('');

  const onSearch = (text) => {
    const exp = formatSearchExpression(text);
    setOptions(defaultOptions.filter((option) => exp.test(option)));
  };

  return (
      <Form>
        <FormField
          htmlFor="select-with-search__input"
          name="select-with-search"
          label="Location"
          help="Type to filter the location options"
        >
          <Select
            id="select-with-search"
            name="select-with-search"
            placeholder="Select location"
            searchPlaceholder="Search locations"
            options={options}
            value={selected}
            onChange={({ option }) => setSelected(option)}
            onSearch={(text) => onSearch(text)}
            clear
          />
        </FormField>
        <FormField>
          <DateInput format="mm/dd/yyyy" />
        </FormField>
      </Form>
  );
};

export default {
  title: 'Input/Select/Test',
};
```

Ensure focus is working correctly with Select + search in a Form and that DateInput focus in a Form is also working correctly

#5887 

No

Yes (Fixed an issue where Select with search in a FormField had double focus)

Backwards compatible

### DIFF
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -458,8 +458,11 @@ const FormField = forwardRef(
         {...outerProps}
         style={outerStyle}
         onFocus={(event) => {
-          setFocus(containsFocus(formFieldRef.current));
-          if (isDateInputComponent) {
+          if (!isDateInputComponent) {
+            setFocus(containsFocus(formFieldRef.current));
+          } else {
+            // if FormField contains a DateInput component check if the calendar
+            // button is in focus before assigning focus to the FormField
             const formDescendants =
               formFieldRef.current.getElementsByTagName('*');
             let buttonChild;
@@ -468,6 +471,7 @@ const FormField = forwardRef(
               if (child.tagName.toLowerCase() === 'button') buttonChild = child;
             }
             if (buttonChild === document.activeElement) setFocus(false);
+            else setFocus(containsFocus(formFieldRef.current));
           }
           if (onFocus) onFocus(event);
         }}
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -8,7 +8,7 @@ import React, {
 import styled, { ThemeContext } from 'styled-components';
 import { defaultProps } from '../../default-props';
 
-import { shouldKeepFocus } from '../../utils/DOM';
+import { containsFocus } from '../../utils/DOM';
 import { focusStyle } from '../../utils/styles';
 import { parseMetricToNum } from '../../utils/mixins';
 import { useForwardedRef } from '../../utils/refs';
@@ -262,6 +262,7 @@ const FormField = forwardRef(
     // fileinput handle
     // use fileinput plain use formfield to drive the border
     let isFileInputComponent;
+    let isDateInputComponent;
     if (
       children &&
       Children.forEach(children, (child) => {
@@ -271,6 +272,12 @@ const FormField = forwardRef(
           'FileInput'.indexOf(child.type.displayName) !== -1
         )
           isFileInputComponent = true;
+        if (
+          child &&
+          child.type &&
+          'DateInput'.indexOf(child.type.displayName) !== -1
+        )
+          isDateInputComponent = true;
       })
     );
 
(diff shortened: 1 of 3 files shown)


src/js/components/FormField/FormField.js
```diff
@@ -451,7 +458,17 @@ const FormField = forwardRef(
         {...outerProps}
         style={outerStyle}
         onFocus={(event) => {
-          setFocus(shouldKeepFocus());
+          setFocus(containsFocus(formFieldRef.current));
+          if (isDateInputComponent) {
```
taysea, line 462: Can we add a comment around what this block is doing? If i'm following correctly, we are removing formfield focus from the formfield when the dateinput calendar button is in focus?

Also wondering if we should structure as
```
if (!isDateInputComponent) setFocus(containsFocus...)
else {
  const formDescendants
  ...
}

File: src/js/components/FormField/FormField.js [hunk 2]
```diff
@@ -271,6 +272,12 @@ const FormField = forwardRef(
           'FileInput'.indexOf(child.type.displayName) !== -1
         )
           isFileInputComponent = true;
+        if (
+          child &&
+          child.type &&
+          'DateInput'.indexOf(child.type.displayName) !== -1
```
MikeKingdom, line 278: Is there a way we can manage this more generically such that FormField doesn't have to specifically know about DateInput and FileInput? This is a type of coupling that causes maintenance issues. It also means if someone creates their own component with similar issues it will be a problem if in a FormField.

src/js/components/FormField/FormField.js
```diff
@@ -451,7 +458,21 @@ const FormField = forwardRef(
...
+            setFocus(containsFocus(formFieldRef.current));
+          } else {
+            // if FormField contains a DateInput component check if the calendar
+            // button is in focus before assigning focus to the FormField
+            const formDescendants =
+              formFieldRef.current.getElementsByTagName('*');
+            let buttonChild;
+            for (let i = 0; i < formDescendants.length; i += 1) {
+              const child = formDescendants[i];
+              if (child.tagName.toLowerCase() === 'button') buttonChild = child;
```
MikeKingdom, line 471: This looks easy to break if we change what types of components/html make up a DateInput

src/js/components/FormField/FormField.js
jcfilben: I'm currently exploring what other ways we can handle this so that we don't have to have specific code to handle the DateInput case

jcfilben: I opened a new PR with an alternate solution that is a bit more generic and doesn't rely on specific code to handle the DateInput case https://github.com/grommet/grommet/pull/5895

jcfilben:
Closing in favor of https://github.com/grommet/grommet/pull/5895

Issue #5878, opened 2022-01-05
MaskedInput with options won't allow for continued typing after match


mattoni:
### Expected Behavior

Should be able to continue typing after a match in a MaskedInput

### Actual Behavior

The dropdown options, when matched, stop additional input from happening. For example, if I have an input to match any number, but options of 30, 45, and 60, I can't input 301 though I can input 311.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-forked-l16op?file=/index.js

### Steps to Reproduce

1. Enter 311 into the box, it works
2. Enter 301 into the box, observe you can only input 30.

### Your Environment

- Grommet version: 2.19.1
- Browser Name and version: 
- Operating System and version (desktop or mobile):

jcfilben:
Note: the MaskedInput is only preventing continued typing when the options are strings. If the options are numbers this issue isn't happening, see this codesandbox: https://codesandbox.io/s/grommet-v2-template-forked-rivgk?file=/index.js

matheushahnn:
I would like to help with this bug, but I didn't get the expected behavior. 
Should be able to type a different key from the regex and options? I mean, it doesn't look right to type 31 if the options are 30, 45, 60 and 120.

mattoni:
Basically I'm trying to use this as an input with some suggestions in the dropdown, that follows a pattern. e.g. I want to allow arbitrary number of minutes to be input, but also provide 'quick' suggestions via the dropdown.

PR #5894, opened 2022-01-14
fix: maskedInput not allowing to continue typing 

matheushahnn:
#### What does this PR do?
This PR removes the code that checks the options and the input value

Start by maskedInput storybook

Created unit test to this scenario.

Type in a maskedInput with options one option and continue typing

https://github.com/grommet/grommet/issues/5878


src/js/components/MaskedInput/MaskedInput.js
```diff
@@ -72,27 +72,6 @@ const parseValue = (mask, value) => {
 
       maskIndex += 1;
       found = true;
-    } else if (item.options) {
```
matheushahnn, line 75: I haven't found side effects, let me know if I missed some test.

jcfilben (commented):
In the [MaskedInput/Filter](https://storybook.grommet.io/?path=/story/input-maskedinput-filtered--filtered) story, previously if you selected a poet from the list it would then show you options to select a poem by that poet, but now the poem options no longer show up.

I am wondering if allowing the user to continue typing beyond one of the options should be an opt-in behavior

matheushahnn:
> In the [MaskedInput/Filter](https://storybook.grommet.io/?path=/story/input-maskedinput-filtered--filtered) story, previously if you selected a poet from the list it would then show you options to select a poem by that poet, but now the poem options no longer show up.
> 
> I am wondering if allowing the user to continue typing beyond one of the options should be an opt-in behavior

I created the ignoreOptions attribute. When it passed true it will ignore the options. When false or not passed it will behave as usual.

src/js/components/MaskedInput/__tests__/MaskedInput-test.js
```diff
@@ -191,6 +191,55 @@ describe('MaskedInput', () => {
...
+        mask={[
+          {
+            options: ['aaa', 'aba', 'abb'],
+            regexp: /\w$/,
+          },
+        ]}
+        onChange={onChange}
+      />,
+    );
+    expect(container.firstChild).toMatchSnapshot();
```
jcfilben, line 210:
```suggestion
```
I don't think we need a snapshot here since it isn't capturing any behavior we are trying to specifically test for in this test

src/js/components/MaskedInput/__tests__/MaskedInput-test.js
```diff
@@ -191,6 +191,55 @@ describe('MaskedInput', () => {
...
+
+    fireEvent.change(getByTestId('test-input'), { target: { value: 'abbb' } });
```
jcfilben, line 212: Can we use `userEvent` instead of `fireEvent` and `getByRole` instead of `getByTestId`? See https://github.com/grommet/grommet/blob/master/CONTRIBUTING.md#testing-your-code for more info

src/js/components/MaskedInput/__tests__/MaskedInput-test.js
```diff
@@ -191,6 +191,55 @@ describe('MaskedInput', () => {
...
+          {
+            ignoreOptions: true,
+            options: ['aaa', 'aba', 'abb'],
+            regexp: /\w$/,
+          },
+        ]}
+        onChange={onChange}
+      />,
+    );
+    expect(container.firstChild).toMatchSnapshot();
```
jcfilben, line 235:
```suggestion
```
Same comment as above, I don't think we need a snapshot here

src/js/components/MaskedInput/__tests__/MaskedInput-test.js
```diff
@@ -191,6 +191,55 @@ describe('MaskedInput', () => {
...
+
+    fireEvent.change(getByTestId('test-input'), { target: { value: 'abbb' } });
```
jcfilben, line 237: Same comment as above,
Can we use `userEvent` instead of `fireEvent` and `getByRole` instead of `getByTestId`?

jcfilben (commented):
Can you add the `ignoreOptions` prop to the documentation on the [grommet-site](https://github.com/grommet/grommet-site) repository?

src/js/components/MaskedInput/__tests__/MaskedInput-test.js
matheushahnn: Sure, sorry about that.

matheushahnn:
> Can you add the `ignoreOptions` prop to the documentation on the [grommet-site](https://github.com/grommet/grommet-site) repository?

Done https://github.com/grommet/grommet-site/pull/344
I just don't know where write more details about this prop.

jcfilben (commented):
Thanks for updating the tests and adding the ignoreOptions prop to the docs!

Something else I noticed while testing was if you have the following mask for inputing a time
```
{
  length: [1, 2],
  options: ['1', '2', '3', '4', '5', '6', '7', '8', '9' ],
  ignoreOptions: true,
  regexp: /^1[1-2]$|^[0-9]$/,
  placeholder: 'hh',
},
{ fixed: ':' },
{
  length: 2,
  options: ['00', '15', '30', '45'],
  ignoreOptions: true,
  regexp: /^[0-5][0-9]$|^[0-9]$/,
  placeholder: 'mm',
}
```
and you want to type the time "10:00" after typing a 1 and then a 0 it automatically inserts a colon between them so you get "1:0" instead 0f "10". I think this happens because 0 happens to be the first number in one of the options for minutes. This doesn't happen when trying to type "12:00".

matheushahnn:
> 

In this case, the regexp is not considering the "0", it is only allowed to type "11" and "12". 
Instead of `regexp: /^1[1-2]$|^[0-9]$/,` try this one `regexp: /^1[0-2]$|^[0-9]$/,`

```
{
  length: [1, 2],
  options: ["1", "2", "3", "4", "5", "6", "7", "8", "9"],
  ignoreOptions: true,
  regexp: /^1[0-2]$|^[0-9]$/,
  placeholder: "hh"
},
{ fixed: ":" },
{
  length: 2,
  options: ["00", "15", "30", "45"],
  ignoreOptions: true,
  regexp: /^[0-5][0-9]$|^[0-9]$/,
  placeholder: "mm"
}
```

jcfilben:
> In this case, the regexp is not considering the "0", it is only allowed to type "11" and "12".
Instead of regexp: /^1[1-2]$|^[0-9]$/, try this one regexp: /^1[0-2]$|^[0-9]$/,

Thanks for the explanation!

jcfilben (approved):
Looks good!

matheushahnn:
@jcfilben Can I merge it?

jcfilben:
@matheushahnn we are planning to do a patch release of Grommet this Thursday so we are waiting to merge any additional pull requests until after the patch release

ericsoderberghp:
I think the functionality makes sense. But, I think the name "ignoreOptions" isn't quite right. On the surface, "ignoreOptions" would seem to indicate that the "options" should be ignored, as in not even considered. Instead, I think something like "requireOption" or "restrictToOptions" seems a bit more aligned to the behavior.

src/js/components/MaskedInput/MaskedInput.js
```diff
@@ -72,7 +72,7 @@ const parseValue = (mask, value) => {
...
+    } else if (item.options && item.restrictToOptions) {
```
ericsoderberghp, line 75: This looks like it will be a breaking change unless we test explicitly for `item.restrictToOptions !== false`.

matheushahnn:
@jcfilben Hey, can I try to fix this error in unit tests?

jcfilben:
> @jcfilben Hey, can I try to fix this error in unit tests?

Yes! It looks like an issue with storybook. I would try just updating the branch with master and see if that re-triggers the storybook run correctly

ericsoderberghp (approved):

PR #5895, opened 2022-01-14
FormField Focus - Alternate Solution

### LINKED ISSUE #5887: FormField with Select + Search receives double focus

jcfilben:
#### What does this PR do?
This fixes an issue where Select with search receives double focus when it is contained within a FormField. This is an alternate solution to the one proposed in #5890. This solution relies less on knowing the specific component that is the child of FormField and handling focus based on that, but is instead a more general approach.

**Note:**
The chromatic snapshot change is happening because of the code I added on line 277 in FormField.js that sets isFileInputComponent to true when it is defined like this: ` <FormField label="File" name="file" component={FileInput} />`
```
if (component && component.displayName === 'FileInput' && !isFileInputComponent) {
  isFileInputComponent = true;
}
```
Previously the style of a FileInput within a FormField would change based on how it was written:
```javascript
 <FormField label="File" name="file" component={FileInput} />
```
Would result in this:
Screen Shot 2022-01-14 at 12 15 35 PM

Whereas this:
```javascript
 <FormField>
  <FileInput />
</FormField>
```
Would result in this: (Now both options will show up like this)
Screen Shot 2022-01-14 at 12 14 51 PM

Compare this PR to the approach in #5890 

Tested with the following story:
```javascript
import { FileInput, Form, FormField, Select, DateInput } from 'grommet'

const defaultOptions = ['EMEA', 'Asia/Pacific', 'Americas', 'Polar Regions'];
// Escaping regular expression special characters: [ \ ^ $ . | ? * + ( )
const getEscapedText = (text) => text.replace(/[-\\^$*+?.()|[\]{}]/g, '\\$&');
// Create the regular expression with escaped special characters.
const formatSearchExpression = (text) => new RegExp(getEscapedText(text), 'i');

export const Test = () => {
  const [selected, setSelected] = useState('');

  const onSearch = (text) => {
    const exp = formatSearchExpression(text);
    setOptions(defaultOptions.filter((option) => exp.test(option)));
  };

  return (
      <Form>
        <FormField
          htmlFor="select-with-search__input"
          name="select-with-search"
          label="Location"
          help="Type to filter the location options"
        >
          <Select
            id="select-with-search"
            name="select-with-search"
            placeholder="Select location"
            searchPlaceholder="Search locations"
            options={options}
            value={selected}
            onChange={({ option }) => setSelected(option)}
            onSearch={(text) => onSearch(text)}
            clear
          />
        </FormField>
        <FormField>
          <DateInput format="mm/dd/yyyy" />
        </FormField>
      </Form>
  );
};

export default {
  title: 'Input/Select/Test',
};
```

Closes #5887

No

Yes

Backwards compatible

### DIFF
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -458,8 +458,11 @@ const FormField = forwardRef(
         {...outerProps}
         style={outerStyle}
         onFocus={(event) => {
-          setFocus(containsFocus(formFieldRef.current));
-          if (isDateInputComponent) {
+          if (!isDateInputComponent) {
+            setFocus(containsFocus(formFieldRef.current));
+          } else {
+            // if FormField contains a DateInput component check if the calendar
+            // button is in focus before assigning focus to the FormField
             const formDescendants =
               formFieldRef.current.getElementsByTagName('*');
             let buttonChild;
@@ -468,6 +471,7 @@ const FormField = forwardRef(
               if (child.tagName.toLowerCase() === 'button') buttonChild = child;
             }
             if (buttonChild === document.activeElement) setFocus(false);
+            else setFocus(containsFocus(formFieldRef.current));
           }
           if (onFocus) onFocus(event);
         }}
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -468,6 +468,9 @@ const FormField = forwardRef(
             if (isFocusable(child)) {
               focusableDescendants += 1;
             }
+            // FormField descendant could be focusable. If descendant is
+            // active element in DOM, don't show focus on FormField.
+            // e.g. DateInput allows Calendar button to receive focus
             if (focusableDescendants >= 2 && child === document.activeElement) {
               setFocus(false);
             }
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -8,7 +8,7 @@ import React, {
 import styled, { ThemeContext } from 'styled-components';
 import { defaultProps } from '../../default-props';
 
-import { containsFocus } from '../../utils/DOM';
+import { containsFocus, isFocusable } from '../../utils/DOM';
 import { focusStyle } from '../../utils/styles';
 import { parseMetricToNum } from '../../utils/mixins';
 import { useForwardedRef } from '../../utils/refs';
@@ -262,7 +262,6 @@ const FormField = forwardRef(
     // fileinput handle
     // use fileinput plain use formfield to drive the border
     let isFileInputComponent;
-    let isDateInputComponent;
     if (
       children &&
       Children.forEach(children, (child) => {
@@ -272,14 +271,16 @@ const FormField = forwardRef(
           'FileInput'.indexOf(child.type.displayName) !== -1
         )
           isFileInputComponent = true;
-        if (
-          child &&
-          child.type &&
-          'DateInput'.indexOf(child.type.displayName) !== -1
-        )
-          isDateInputComponent = true;
       })
     );
+
+    if (
+      component &&
+      component.displayName === 'FileInput' &&
+      !isFileInputComponent
+    ) {
+      isFileInputComponent = true;
+    }
 
     if (!themeBorder) {
       contents = (
@@ -458,20 +459,18 @@ const FormField = forwardRef(
         {...outerProps}
         style={outerStyle}
         onFocus={(event) => {
-          if (!isDateInputComponent) {
-            setFocus(containsFocus(formFieldRef.current));
-          } else {
-            // if FormField contains a DateInput component check if the calendar
-            // button is in focus before assigning focus to the FormField
-            const formDescendants =
-              formFieldRef.current.getElementsByTagName('*');
-            let buttonChild;
-            for (let i = 0; i < formDescendants.length; i += 1) {
-              const child = formDescendants[i];
-              if (child.tagName.toLowerCase() === 'button') buttonChild = child;
+          setFocus(containsFocus(formFieldRef.current));
+          const formDescendants =
+            formFieldRef.current.getElementsByTagName('*');
+          let focusableDescendants = 0;
+          for (let i = 0; i < formDescendants.length; i += 1) {
+            const child = formDescendants[i];
+            if (isFocusable(child)) {
+              focusableDescendants += 1;
             }
-            if (buttonChild === document.activeElement) setFocus(false);
-            else setFocus(containsFocus(formFieldRef.current));
+            if (focusableDescendants >= 2 && child === document.activeElement) {
+              setFocus(false);
+            }
           }
           if (onFocus) onFocus(event);
         }}
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -8,7 +8,7 @@ import React, {
 import styled, { ThemeContext } from 'styled-components';
 import { defaultProps } from '../../default-props';
 
-import { containsFocus, isFocusable } from '../../utils/DOM';
+import { containsFocus, shouldKeepFocus } from '../../utils/DOM';
 import { focusStyle } from '../../utils/styles';
 import { parseMetricToNum } from '../../utils/mixins';
 import { useForwardedRef } from '../../utils/refs';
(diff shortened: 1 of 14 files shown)


src/js/components/FormField/FormField.js
```diff
@@ -451,7 +459,19 @@ const FormField = forwardRef(
...
+          setFocus(containsFocus(formFieldRef.current));
+          const formDescendants =
+            formFieldRef.current.getElementsByTagName('*');
+          let focusableDescendants = 0;
+          for (let i = 0; i < formDescendants.length; i += 1) {
+            const child = formDescendants[i];
+            if (isFocusable(child)) {
+              focusableDescendants += 1;
+            }
+            if (focusableDescendants >= 2 && child === document.activeElement) {
```
taysea, line 471: maybe adding a small comment around why `2` is the condition would be helpful for folks in future.

suggestion but feel free to alter
```
// FormField descendant could be focusable. If descendant is
// active element in DOM, don't show focus on FormField.
// e.g. DateInput allows Calendar button to receive focus
```

src/js/components/FormField/FormField.js
```diff
@@ -451,7 +459,19 @@ const FormField = forwardRef(
...
+              setFocus(false);
```
taysea, line 472: I would expect a snapshot change if we had a test case covering this, could be add one here -- maybe using Select + Search or DateInput calendar button as the use case?

taysea:
Overall, I like this approach better

src/js/components/FormField/FormField.js
jcfilben: Added a comment in the latest push

src/js/components/FormField/FormField.js
jcfilben: I added a test in the most recent commit for testing focus with DateInput within a FormField

jcfilben:
> Do you have any insight on why a snapshot changed? seems like everything is just shifted down by a pixel or two?

I updated the PR description with an explanation on the chromatic snapshot change

src/js/components/FormField/FormField.js
```diff
@@ -451,7 +459,22 @@ const FormField = forwardRef(
...
         style={outerStyle}
         onFocus={(event) => {
-          setFocus(shouldKeepFocus());
+          setFocus(containsFocus(formFieldRef.current));
+          const formDescendants =
+            formFieldRef.current.getElementsByTagName('*');
+          let focusableDescendants = 0;
+          for (let i = 0; i < formDescendants.length; i += 1) {
+            const child = formDescendants[i];
+            if (isFocusable(child)) {
```
MikeKingdom, line 468: I like this generic. It seems more generic and more likely to work with random components.

One question I have is how does this work with Drops which may not be a descendent? Is this an issue?

src/js/components/FormField/FormField.js
```diff
@@ -451,7 +459,22 @@ const FormField = forwardRef(
...
+              focusableDescendants += 1;
+            }
+            // FormField descendant could be focusable. If descendant is
+            // active element in DOM, don't show focus on FormField.
+            // e.g. DateInput allows Calendar button to receive focus
+            if (focusableDescendants >= 2 && child === document.activeElement) {
```
MikeKingdom, line 474: The activeElement check weeds out things that may not actually be visible (e.g. `display:none`) or may be `disabled`?

src/js/utils/DOM.js
```diff
@@ -67,7 +67,12 @@ export const containsFocus = (node) => {
 // Check if the element.tagName is an input, select or textarea
 export const isFocusable = (element) => {
   const tagName = element.tagName.toLowerCase();
-  return tagName === 'input' || tagName === 'select' || tagName === 'textarea';
+  return (
+    tagName === 'input' ||
```
MikeKingdom, line 71: I think relying on tagName misses a lot of rich or custom components. For example a RangeSelector has two little controls, one of which may have the focus. Those controls are just `div`s with `tabindex="0"`. If a RangeSelector is somehow one of the descendents then it won't get counted. Focusable things include those tags plus basically anything with `tabindex` (even if it is "-1").

src/js/components/FormField/FormField.js
jcfilben: When focus transfers to content in the Drop this is handled by `setFocus(containsFocus(formFieldRef.current));` (line 462 in FormField.js) so it is not an issue that the Drop isn't included in formDescendants. (Note: I tested with Select and DateInput, both are built using Drop and focus correctly transfers.)

src/js/components/FormField/FormField.js
jcfilben: Yes, if an element is disabled or if it has `visibility: 'hidden'` it will still show up in the `focusableDescendants` list but the `child === document.activeElement` check makes sure the focus isn't incorrectly removed from the FormField if it contains a disabled or invisible child. Focus will only be removed from the FormField if there is another focusable descendant and that element has focus

src/js/utils/DOM.js
jcfilben: Currently looking into if there is a better way to handle this

src/js/components/FormField/FormField.js
MikeKingdom: Thanks. Sounds good.

src/js/components/FormField/FormField.js
MikeKingdom: Great. Thanks.

src/js/utils/DOM.js
MikeKingdom: I think adding the tabindex check OR'd with the above tags is promising.

jcfilben: Updated in latest commit

src/js/components/FormField/FormField.js
```diff
@@ -451,7 +459,47 @@ const FormField = forwardRef(
...
+            if (
+              (children &&
+                children.type &&
+                grommetInputNames.includes(children.type.displayName)) ||
+              (!children &&
+                component &&
+                grommetInputNames.includes(component.displayName)) ||
+              (!children && !component)
+            )
+              isGrommetInputComponent = true;
```
MikeKingdom, line 483: I'm not sure if I'm reading this right, but it's trying to decide if the component or child of the formfield is one of the known Grommet input style components but does that mean if the child is a Box that contains a grommet input component it won't notice? And if we add some other grommet input type component we'll have to realize we need to update FormField? It's not clear to me why it really matters if it's a grommet component (ie. that distinction is really complicating this code alot.)

src/js/utils/DOM.js
```diff
@@ -67,25 +67,13 @@ export const containsFocus = (node) => {
...
-export const shouldKeepFocus = () => {
-  const element = document.activeElement;
-  if (isFocusable(element)) return true;
-  return !!getFirstFocusableDescendant(element);
+  return (
+    tagName === 'input' ||
+    tagName === 'select' ||
+    tagName === 'textarea' ||
+    tagName === 'button' ||
+    element.tabIndex >= 0
```
MikeKingdom, line 75: Why exclude -1? -1 takes it out of the tab order but you can still get focus on it a couple different ways.

jcfilben: Including -1 results in too many elements being marked as focusable. For example, with the Select component 9 elements will be marked as focusable including Boxes, Icons, Svgs, and others because they have a tabIndex of -1.

src/js/components/FormField/FormField.js
jcfilben: If a FormField contains one of the grommet input components, the input components are set up to not have a `focusIndicator` and instead have the focus on the FormField. So for grommet input components even though they have a focusable descendant we don’t want to remove focus from the FormField. For all other components (including custom components) we should handle focus normally and remove focus from the FormField when one of the focusable descendants has focus.

If an input component is wrapped with a Box it won’t be flagged as a grommet input component and focus will be on the FormField. (Tested with Select wrapped in a Box)

As far as updating FormField.js when we add a new grommet input component this is something we would have to do anyway because `grommetInputNames` is being used to determine padding and in other places in FormField.js

taysea: @jcfilben -- regarding your point re: Select wrapped in Box, are you saying that if I implemented the following, I would still see the double focus on Select search and formfield?

```
<FormField>
   <Box>
      <Select onSearch={() => {}} />
   </Box>
</FormField>
```

jcfilben: Sorry I should have wrote that the Select will have focus and not the FormField in this case. If you have Select wrapped in a Box it wouldn't be flagged as a grommet input component and focus would be handled in this statement: `if(focusableDescendants.length >= 1 && !isGrommetInputComponent && focusableDescendants.includes(document.activeElement)) setFocus(false)` (See line 494 of FormField). The select would count as a focusable descendant and it would be the active element when tabbed to so focus on FormField would be set to false as we would see the focusIndicator on Select like this after tabbing to the field:

Screen Shot 2022-01-20 at 1 58 37 PM

taysea: Gotcha, thanks for clarifying. Something I'm noticing here (but a separate issue from what you're doing) is the FormField and Select both are having their border now (creating a double border) -- not sure if we have an issue filed around it. It seems to be an effect of only searching the direct children of the formfield

jcfilben: The double border shown in the screenshots above is just a result of wrapping Select in a Box within the FormField. If we have `<FormField> <Select /> </FormField>` there is code in the FormField.js file that removes the border of Select so we don't have 2 borders. If you are wrapping Select in a Box within FormField I think the expectation would be that you are doing some custom thing and I think it is reasonable to still show the Select border

src/js/components/FormField/FormField.js
```diff
@@ -451,7 +459,44 @@ const FormField = forwardRef(
...
+            that has focus. */
+            if (
+              (focusableDescendants.length >= 2 &&
+                isGrommetInputComponent &&
+                child === document.activeElement) ||
+              (focusableDescendants.length >= 1 &&
+                !isGrommetInputComponent &&
+                focusableDescendants.includes(document.activeElement))
+            )
+              setFocus(false);
```
MikeKingdom, line 498: I think you can actually remove all this if you change `containsFocus` like noted below. I tested the result and it seems to work well

src/js/utils/DOM.js
```diff
@@ -67,25 +67,13 @@ export const containsFocus = (node) => {
...
+  );
```
MikeKingdom, line 76:
```jsx
return (
  element.tabIndex >= 0 ||
  ((tagName === 'input' ||
    tagName === 'select' ||
    tagName === 'textarea' ||
    tagName === 'button') &&
    element.tabIndex !== -1)
  );
```

MikeKingdom (approved):
Awesome. Thanks!

taysea (approved):

ericsoderberghp (approved):

PR #5906, opened 2022-01-21
add gap to theme for button

britt6612:
#### What does this PR do?
Adds the ability to change the gap through the theme 
button.js
tested it using a value in the theme 

Design System needs a way to change `gap` to `6px` for buttons with Icons
yes
sure
backwards compatible


src/js/components/Button/Button.js
```diff
@@ -276,7 +276,7 @@ const Button = forwardRef(
           direction="row"
           align="center"
           justify={justify || (align === 'center' ? 'center' : 'between')}
-          gap={gap}
+          gap={theme.button.gap || gap}
```
taysea, line 279:
```suggestion
          gap={gap || theme.button.gap}
```
should it be like this? theme would be fallback if gap prop is undefined

src/js/themes/base.js
```diff
@@ -418,6 +418,7 @@ export const generate = (baseSpacing = 24, scale = 6) => {
           },
         },
       },
+      // gap: undefined,
```
taysea, line 421: Should the default `small` from Button.js get moved here to the theme?

britt6612: I was debating on this as well and removing from the prop I just wasn't sure

taysea (approved):
LGTM!

ericsoderberghp (approved):

PR #5910, opened 2022-01-24
add position prop for notifications

britt6612:
#### What does this PR do?
This PR adds a `position` prop to notification
This is for when a user is using a `Toast` it allows them to have the `Toast` in the `top` or `top-right`
notification.js
storybook

In Design-System site we have a placement [guidance](http://localhost:3000/templates/toast-notifications#placement) 
This states Place `Toast` either top-center or top-right of the screen.
In order for a user to be able to put in the `top-right` we need to provide a prop. 
Needed for Design system
yes
yes
 backwards compatible


src/js/components/Notification/stories/TitleAndMessage.js
```diff
@@ -18,6 +18,7 @@ const TitleAndMessageNotification = () => {
       {visible && (
         <Notification
           toast
+          position="top-right"
```
taysea, line 21: instead of adjusting this story, can we add a jest test that catches the position in a snapshot?

britt6612: Yes good point ill adjust

taysea: Jest test looks good -- can we remove from here now?

taysea (commented):
Just one small comment

src/js/components/Notification/stories/TitleAndMessage.js
britt6612: yes I will update

taysea (approved):

src/js/components/Notification/Notification.js
```diff
@@ -10,7 +10,15 @@ import { Text } from '../Text';
 
 import { NotificationType } from './propTypes';
 
-const Notification = ({ message, onClose, status, title, toast }) => {
+const Notification = ({
+  message,
+  onClose,
+  id,
+  position,
```
ericsoderberghp, line 17:
```suggestion
  position = 'top',
```

src/js/components/Notification/Notification.js
```diff
@@ -72,8 +80,10 @@ const Notification = ({ message, onClose, status, title, toast }) => {
         role="log"
         modal={false}
         onEsc={onClose}
+        id={id}
         responsive
         plain
+        position={position === 'top-right' ? 'top-right' : 'top'}
```
ericsoderberghp, line 86: Rather than default here, I think we should default in the initial declaration and take whatever the caller passes in. So, "bottom-right" would work. From a grommet perspective, I think we should be this flexible.

src/js/components/Notification/__tests__/Notification-test.js
```diff
@@ -45,4 +47,15 @@ describe('Notification', () => {
     userEvent.click(screen.getByRole('button'));
     expect(onClose).toBeCalled();
   });
+
+  test(`should apply correct position`, () => {
```
ericsoderberghp, line 51: Can we add a test covering using the `position` property?

src/js/components/Notification/Notification.js
britt6612: I was going back and forth with whether or not to just allow all of the properties. I'll reconfigure

ericsoderberghp (approved):

Issue #5889, opened 2022-01-12
DateInput Impossible to change the Icon


WolfCSGO:
### Expected Behavior
The Icon rendered when a format is pass as a prop should be customisable as before via props.

### Actual Behavior
In version 2.19, the icon of the DateInput can be customised via inputProps because the icon was inside the MakedInput component and with inputProps={{ icon: <OtherIcon /> }}, it can be changed.
Due to a recent change in version 2.20, the Icon is separated from the MaskedInput in a Button and it's not possible to change. If you pass an icon in inputProps, you get two icons, one of the MaskedInput and another from the Button.

### URL, screen shot, or Codepen exhibiting the issue
Last version:
https://codesandbox.io/s/dateinput-icon-bug-emnc2

Version 2.19: 
https://codesandbox.io/s/dateinput-icon-bug-v2-19-h153w

PR #5911, opened 2022-01-25
[Bug] - DateInput icon issue

g4rry420:

#### What does this PR do?
Resolves the two icons in the dateinput

components => DateInput.js

None

Using the code in the codesandbox of the #5889 

N/A

N/A

#5889

No

Backwards compatibe


jcfilben:
Implementation looks good!

I am debating if we should have an `icon` prop for DateInput. I realize we need to allow setting the icon through `inputProps` to maintain backwards compatibility. I am wondering if we should mark setting the icon through `inputProps` as deprecated and encourage changing the DateInput through an `icon` prop since the icon is no longer within the input that inputProps is referring to. Curious to hear what you think @ericsoderberghp or if anyone else has thoughts on this

g4rry420:
Good idea, I will make the changes

src/js/components/DateInput/DateInput.js
```diff
@@ -185,7 +188,7 @@ const DateInput = forwardRef(
           id={id}
           dropProps={{ align: { top: 'bottom', left: 'left' }, ...dropProps }}
           dropContent={calendar}
-          icon={<CalendarIcon size={iconSize} />}
+          icon={icon || <CalendarIcon size={iconSize} />}
```
jcfilben, line 191: We should still allow the icon to be set through input props.
```suggestion
          icon={icon || MaskedInputIcon || <CalendarIcon size={iconSize} />}
```

src/js/components/DateInput/DateInput.js
```diff
@@ -247,7 +250,7 @@ const DateInput = forwardRef(
             <Button
               onClick={open ? closeCalendar : openCalendar}
               plain
-              icon={<CalendarIcon size={iconSize} />}
+              icon={icon || <CalendarIcon size={iconSize} />}
```
jcfilben, line 253:
```suggestion
              icon={icon || MaskedInputIcon || <CalendarIcon size={iconSize} />}
```

src/js/components/DateInput/DateInput.js
```diff
@@ -97,6 +98,8 @@ const DateInput = forwardRef(
       schema ? valueToText(normalizedDate, schema) : undefined,
     );
 
+    const { icon: MaskedInputIcon, ...restOfInputProps } = inputProps || {};
+
```
jcfilben, line 102: We should add a comment that changing the icon with inputProps is deprecated and also put a warning in the console noting this as well.
```suggestion
    // Setting the icon through `inputProps` is deprecated.
    // The `icon` prop should be used instead.
    const { icon: MaskedInputIcon, ...restOfInputProps } = inputProps || {};
    if (MaskedInputIcon) {
      console.warn(
        `Customizing the DateInput icon through inputProps is deprecated. 
Use the icon prop instead.`,
      );
    }

```

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

PR #5914, opened 2022-01-26
Tag size

MikeKingdom:
#### What does this PR do?
Add a `size` property to Tag

Tag.js

Storybook example and jest testcase

Storybook Type/Tag/Size example

Greenlake Marketplace designs want a smaller tag so they don't overwhelm and distract from the content on their product cards.

image

Yes

Yes

Backwards compatible


src/js/components/Tag/Tag.js
```diff
@@ -11,23 +11,27 @@ import { Text } from '../Text';
...
-    <Box width={{ min: 'min-content' }} pad={theme.tag.pad}>
-      <Text>
-        {name && <Text {...theme.tag.name}>{name}</Text>}
-        {name && value ? theme.tag.separator : ''}
-        {value && <Text {...theme.tag.value}>{value}</Text>}
+    <Box
+      width={{ min: 'min-content' }}
+      pad={theme.tag.size?.[size]?.pad || theme.tag.pad}
+    >
+      <Text size={size}>
```
ericsoderberghp, line 31: Why do we need the size both here and in the contained Text elements? I would think one or the other would suffice. Maybe just keeping this one?

src/js/components/Tag/Tag.js
```diff
@@ -11,23 +11,27 @@ import { Text } from '../Text';
...
+        {name && <Text {...theme.tag.name} size={size}> {name}</Text>}
+        {name && value ? <Text size={size}>{theme.tag.separator}</Text> : ''}
+        {value && <Text {...theme.tag.value} size={size}>{value}</Text>}
```
ericsoderberghp, line 34: I think omitting `size` here would make it easier for theme creators to have a different size for name or value, if we just kept the size at line 31.

src/js/components/Tag/__tests__/Tag-test.tsx
```diff
@@ -72,4 +72,14 @@ describe('Tag', () => {
 
     expect(container.firstChild).toMatchSnapshot();
   });
+
+  test('size', () => {
+    const { container } = render(
+      <Grommet>
+        <Tag size="small" name="Name" value="Value" />
```
ericsoderberghp, line 79: Can we include all sizes?

src/js/components/Tag/Tag.js
```diff
@@ -50,8 +54,8 @@ const Tag = forwardRef(({ name, value, onRemove, onClick, ...rest }, ref) => {
           plain
           hoverIndicator
           focusIndicator
-          icon={<FormClose />}
-          round={theme.tag.round}
+          icon={<FormClose {...theme.tag.size?.[size]?.icon} />}
```
ericsoderberghp, line 57: Shouldn't this be `theme.tag.size?.[size]?.icon?.size`?

src/js/themes/base.js
```diff
@@ -1489,6 +1489,23 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         margin: { right: 'xsmall' },
       },
       separator: ' : ',
+      size: {
+        xsmall: { 
+          pad: { horizontal: 'small' },
```
ericsoderberghp, line 1494: Does this padding align with Button for its sizes?

src/js/components/Tag/Tag.js
MikeKingdom: Without it here you get a span with a line-height of medium which makes it bigger than the contained text

src/js/components/Tag/Tag.js
MikeKingdom: I tried this but the name and value didn't inherit it

src/js/components/Tag/Tag.js
MikeKingdom: We could do size={theme.tag.size?.[size]?.icon?.size} but the above lets you specify any icon property. Whichever you prefer...

src/js/components/Tag/__tests__/Tag-test.tsx
MikeKingdom: surely

src/js/themes/base.js
MikeKingdom: Tag has less horizontal padding (that's how it was spec'd in the figma). Button also calculates it as something like ${baseSpacing / 4 - borderWidth} etc rather than using tshirt sizes which we could do if needed, especially once the Figma details are decided.

src/js/components/Tag/stories/Size.js
```diff
@@ -0,0 +1,20 @@
...
+import { Box, Tag, Text } from 'grommet';
+
+export const Size = () => (
+  <Box pad="large" direction="row" gap="medium" wrap>
+    {['xsmall', 'small', 'medium', 'large', 'xlarge'].map(size=>(
+      <Box gap="medium" align="start" key={size}>
+        <Text size={size} weight="bold">{size}</Text>
+        <Tag name="Name" value="Value" size={size}/>
+        <Tag value="Value" size={size}/>
+        <Tag value="Value" size={size} onRemove={()=>{}}/>
```
taysea, line 12: `onRemove` for size "xsmall" seems bad for accessibility. The clickable target is only 12x12px. Wondering if we shouldn't support `onRemove` for this size OR at least advise away from it for Design System related patterns.

src/js/components/__tests__/__snapshots__/components-test.js.snap
```diff
@@ -2081,6 +2081,39 @@ Object {
...
+                "horizontal": "medium",
+                "vertical": "xsmall",
+              },
+            },
+            "xsmall": Object {
+              "icon": Object {
+                "size": "small",
+              },
+              "pad": Object {
+                "horizontal": "small",
```
taysea, line 2113: This horizontal pad looks disproportionally large than the horizontal pad on other sizes -- is that intentional?

src/js/components/Tag/stories/Size.js
MikeKingdom: Agree. I don't think xsmall makes sense to have in a case where you want onRemove. I bigger target may make xsmall not really any smaller than small if it does have onRemove. Either way these values were a guess just to get some discussion started :)

src/js/components/__tests__/__snapshots__/components-test.js.snap
MikeKingdom: It does look a little big but xsmall is too small so it probably needs to be a pixel value (which I think we'd do anyway to be more like how button's padding is done.)

src/js/components/Tag/stories/Size.js
MikeKingdom: Changed the remove button for xsmall to be like the small one. It seems to fit fine and looks decent.

src/js/themes/base.js
MikeKingdom: I tried to align them to be more consistent. They are a little different than button but that seems reasonable since they better fit the original Tag "medium" design.

src/js/components/Tag/Tag.js
britt6612: I like that this way allows you to specify any icon property just for our open-source mentality to not be to opinionated

src/js/components/Tag/Tag.js
```diff
@@ -11,23 +11,27 @@ import { Text } from '../Text';
 
 import { StyledRemoveButton, StyledTagButton } from './StyledTag';
 
-const Tag = forwardRef(({ name, value, onRemove, onClick, ...rest }, ref) => {
+const Tag = forwardRef(
+  ({ name, value, size, onRemove, onClick, ...rest }, ref) => {
```
britt6612, line 15: can we add `size` to the grommet docs

src/js/components/Tag/stories/Size.js
britt6612: the `xsmall` remove looks better

PR #5926, opened 2022-02-01
Feat/position notification

britt6612:
#### What does this PR do?
This PR adds a `position` object that can be passed in with `toast`
This is for when a user is using a `Toast` it allows them to have the `Toast` in the `top` or `top-right`
notification.js
storybook

In Design-System site we have a placement [guidance](http://localhost:3000/templates/toast-notifications#placement) 
This states Place `Toast` either top-center or top-right of the screen.
In order for a user to be able to put in the `top-right` we need to provide a prop. 
Needed for Design system
yes
yes
 backwards compatible


src/js/components/Notification/index.d.ts
```diff
@@ -1,12 +1,27 @@
 import * as React from 'react';
 
 export type StatusType = 'critical' | 'warning' | 'normal' | 'unknown';
+export type PositionType =
```
ericsoderberghp, line 4: Can we just import `LayerPositionType` and re-use it here.

britt6612: yes good call

ericsoderberghp (approved):

Issue #5922, opened 2022-01-29
DateInput assumes a range when format does not specify one (occurs when the format uses dash as a separator)


sbytnar:
v2.20.1. 'DateInput format="yyyy-mm-dd"' does not specify a range, but the component allows the user to select a range.

### Expected Behavior

A format specifying a single date should not allow the user to select a range of days.

### Actual Behavior

The user can select a range.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-forked-74ml3

### Steps to Reproduce

There are two DateInput components on the codesandbox page.
When the DateInput uses '/' as a separator, the user can only choose one date. (Correct.)
When the DateInput uses '-' as a separator, the user can choose a range of dates. (Incorrect detection of the user specifying a date range.)

### Your Environment

- Grommet version: v2.20.1
- Browser Name and version: Chrome, Safari
- Operating System and version (desktop or mobile): MacOS desktop

negreirosleo:
@jcfilben I'm on it, soon I will create the PR :)

valeporti:
hello, hello, could we reopen this one?
I think it's necessary 🤔 . Either let more formats or let a "formatFn" props to let the user choose the format function of his/her choice (ex date-fns formatDate), or choose the separator string when two dates for exemple (instead of only '-').

jcfilben:
Hey @valeporti thanks for sharing your ideas, this issue is still open and the Grommet team is open to an enhancement for this behavior

PR #5930, opened 2022-02-03
Fix DateInput assuming a range on a non range format

negreirosleo:
#### What does this PR do?
It makes the DateInput recognize a date format `mm-dd-yyyy`as a single date and not range

The Reviewer shoud start

I tested it on the storybook and also wrote tests for it.

Try to create a DateInput with a `mm-dd-yyyy` and check that its working as a range 

- [X] `userEvent` is used in place of `fireEvent`.

https://github.com/grommet/grommet/issues/5922

No

Yes

It could be a breaking change if anyone is relying on this buggy behaviour.

### DIFF
--- a/src/js/components/DateInput/DateInput.js
+++ b/src/js/components/DateInput/DateInput.js
@@ -84,8 +84,13 @@ const DateInput = forwardRef(
     // normalize value based on timestamp vs user's local timezone
     const normalizedDate = normalizeForTimezone(value, timestamp);
 
+    // Check if format is on a range format
+    // 10 characters is the largest non range format (mm/dd/yyyy)
+    // eg. mm-dd-yyyy (non-range) mm/dd/yyyy-mm/dd/yyyy (range)
+    const isFormatRange = format && format.length > 10 && format.includes('-');
+
     // do we expect multiple dates?
-    const range = Array.isArray(value) || (format && format.includes('-'));
+    const range = Array.isArray(value) || isFormatRange;
 
     // parse format and build a formal schema we can use elsewhere
     const schema = useMemo(() => formatToSchema(format), [format]);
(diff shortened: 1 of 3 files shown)


src/js/components/DateInput/DateInput.js
```diff
@@ -84,8 +84,13 @@ const DateInput = forwardRef(
     // normalize value based on timestamp vs user's local timezone
     const normalizedDate = normalizeForTimezone(value, timestamp);
 
+    // Check if format is on a range format
+    // 10 characters is the largest non range format (mm/dd/yyyy)
+    // eg. mm-dd-yyyy (non-range) mm/dd/yyyy-mm/dd/yyyy (range)
+    const isFormatRange = format && format.length > 10 && format.includes('-');
```
jcfilben, line 90: Checking the format length wouldn't flag this format as a range: 'm/d-m/d' since it is less than 10 characters long.
```suggestion
    const isFormatRange = format && format.includes('-') && format.includes('/');
```
I wonder if we should do something like and just not support '-' as a separator when choosing a range of dates since it  might be confusing anyway (we would need to make note of this in the docs). 'm-d-yy-m-d-yy' types of formats are difficult to read and interpret as a range.

negreirosleo: maybe the range could be a boolean property, and the format is just a regular single format

jcfilben: > maybe the range could be a boolean property, and the format is just a regular single format
I think this could cause backwards compatibility issues since we are currently using the format prop to include range formats

ericsoderberghp: What about detecting the range condition by looking for duplication of placeholders? For instance, two blocks of 'mm', 'dd', or 'yyyy' would indicate a range scenario.

jcfilben:
Hi @negreirosleo just wanted to check in on this PR and see if this was something you are still interested in working on or if you had any questions on it. Looks like this is pretty close to the finish line!

jcfilben:
Closing due to inactivity. Feel free to re-open

PR #5933, opened 2022-02-08
Update DateInput and Calendar logic for normalization

taysea:
#### What does this PR do?
This PR intends to provide improved logic for DateInput/Calendar when the defaultValue is undefined. In these instances, we want the `onChange` handlers to return dates in UTC (relative to the user's local timezone), but we want the active date in the calendar and the text in the dateinput to reflect the user's local timezone. For example, if I am in Athens, Greece (UTC+2), and I click on February 7, then onChange should call with Feb 6 10pm (since that is the relative UTC time).

The current approach introduces a `normalize` variable that is set based on the timestamp. Right now it is a ref, but I think it needs to be changed to a state variable (in case the initial value is being fetched from a server and there is some delay).

The `formatToLocalYYYYMMDD` and `normalizeForTimezone` functions should only take into account localOffsets when `normalize` is true. Otherwise, we don't want to make adjustments to the time because we want to operate in UTC.

In timezones ahead of UTC, this approach is working. However, for timezones before UTC (in instances where no timezone is specified), upon clicking on a date in the Calendar, the previous date momentarily highlights and then the correct date becomes selected.

### DIFF
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -55,31 +55,35 @@ const activeDates = {
 
 const timeStamp = /T.*/;
 
-const formatSelectedDatesString = (date) => `Currently selected
+const formatSelectedDatesString = (date, normalize) => `Currently selected
   ${date?.map((item) => {
     let dates;
     if (!Array.isArray(item)) {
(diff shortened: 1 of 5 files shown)


taysea:
Closing in favor of https://github.com/grommet/grommet/pull/5942

PR #5934, opened 2022-02-08
Improve Select screenreader and keyboard navigation experience

### LINKED ISSUE #5917: Select has differing screenreader output compared to native select
### Expected Behavior

When focusing on a [native select input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select) with a screenreader (Voiceover in my case), the screenreader announces to the user that they are on a collapsed menu pop-up; remaining focused and allowing the screenreader to continue reading gives the user instructions on how to open a list of options. ("To display a list of options, press...")

Initial readout of select input on focus
image

Additional info as screenreader continues while focused on element
image

### Actual Behavior

My expectation is that the Grommet would have the same readout as the native select input, in order to meet accessibility standards. However, when focusing on a grommet select input, I get a less specific announcement about being on a drop button; remaining focused on the Select gives instructions on how to click the button, but never announces that the button in question corresponds to a menu that will allow me to select a list of options. ("To click this button, press...")

Initial readout of select input on focus
image

Additional info as screenreader continues while focused on element
image

I think it would be nice if the Grommet select mimicked the behavior of the native select in screenreader use cases. Our team's accessibility audit has determined that this is a critical issue that prevents screenreader users from being able to successfully submit a form that has a select input in it, so we've had to pause adoption of the Grommet Select :( I'm not personally super-experienced with accessibility issues but I can empathize with how the screenreader output of Grommet's Select input would be confusing to the user.

(I just used the MDN select input linked above and the Grommet select input in the docs for this testing)

### Your Environment

- Operating System and version (desktop or mobile): OSX 11.4, Voiceover

### LINKED ISSUE #5918: Select menu atypically uses left and right arrow keys to navigate when using a screenreader
### Expected Behavior

When using on a [native select input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select) with a screenreader (Voiceover in my case), I am able to open the list of options by pressing the down arrow key, and then am able to navigate through the options by using the up and down arrow keys.

(here's a gif of me using voiceover with the native select linked above - it's not super informative, but i'm using the arrow keys described above to navigate, and the enter key to select an option)
selectnative

### Actual Behavior

My expectation is that the Grommet would have the same behavior as the native select input, in order to meet accessibility standards. However, when using the Grommet Select input with Voiceover, the down arrow key does not open the menu, the Enter key does instead; and the up and down arrow keys do nothing when the menu is open, I use the left and right arrow keys to navigate through the options instead.

(again, this gif might not be super informative lol - but first i try to open the menu using down arrow key, and get a "Landmark not found" message; then I open it with enter and attempt to navigate using up and down arrow keys, and then use left and right to navigate and enter to select an option)
selectGrommet

I think it would be nice if the Grommet select mimicked the behavior of the native select in screenreader use cases. Our team's accessibility audit has determined that this is a critical issue that prevents screenreader users from being able to successfully submit a form that has a select input in it, so we've had to pause adoption of the Grommet Select :(

(I just used the MDN select input linked above and the Grommet select input in the docs for this testing)

### Your Environment

- Operating System and version (desktop or mobile): OSX 11.4, Voiceover

jcfilben:
#### What does this PR do?
Improves how the screenreader reads and interacts with the Select component. 
- Options can now be selected with either 'Enter' or 'Space' (previously only 'Enter' worked)
- Options are read when navigating with up and down arrows (previously they were not read out)
- The user is notified when an option is selected and when revisiting a selected option.
- When an option is selected and the drop is closed and the select has focus the aria-label, a11yTitle, or placeholder is read along with the selected option.

Notes: 
- There is an issue in Chrome with voice over where once an option is selected, when navigating to other options the screen reader reads out the selected option. This issue is present in other Select components when using a screenreader (see https://mui.com/components/selects/). Firefox and Safari do not have this issue.
- The 'Clear' button is not currently accessible. I am planning to open a separate issue for this.

Snapshot Changes:
- This following section is being removed from snapshots because I am setting the `focusIndicator={false}` in SelectOption (line435 of SelectContainer.js). Previously we were not setting focus on the options which is why they were not being read out by the screenreader. Now that we are setting focus I removed the focusIndicator to stay consistent with how options were displayed before. Visually the Select looks the same as it did before.
```
.c5:focus {
  outline: none;
  box-shadow: 0 0 2px 2px #6FFFB0;
}
.c5:focus > circle,
.c5:focus > ellipse,
.c5:focus > line,
.c5:focus > path,
.c5:focus > polygon,
.c5:focus > polyline,
.c5:focus > rect {
  outline: none;
  box-shadow: 0 0 2px 2px #6FFFB0;
}
.c5:focus::-moz-focus-inner {
  border: 0;
}
```

- Some of the aria-labels are changing. See [languages/default.json](https://github.com/grommet/grommet/pull/5934/files#diff-ecd05461ba0fd4b2a00e66e9708f8d8a59bf3dde54716424f8aadaa0a4750430) and [Select.js line 307](https://github.com/grommet/grommet/pull/5934/files#diff-cdd95f77ccbcbeefc239c0ad1f72eec210c2e09b16719e18af4a75e71e68e73bR307) for relevant code. The aria label change was made so that the screenreader will notify the user what the aria-label or placeholder for the select is in addition to the currently selected item or items when the Select gets focus.

- `aria-posinset` is increasing by one in the snapshots. Previously the indexing was starting at 0 so when there was two options the screenreader would read "option name (0 of 2)" for the first and "option name (1 of 2)" for the second. Now it reads "option name (1 of 2)" for the first and "option name (2 of 2)".

See how the current Select is working with a screenreader & keyboard navigation.

Tested on Chrome, Firefox, and Safari with voice over

Test with a screenreader on the Select storybook stories

Closes #5918 
Closes #5917 

No

Yes

Backwards compatible

### DIFF
--- a/src/js/components/Select/Select.js
+++ b/src/js/components/Select/Select.js
@@ -101,6 +101,9 @@ const Select = forwardRef(
     const formContext = useContext(FormContext);
     const { format } = useContext(MessageContext);
 
+    // Determine if the Select is opened with the keyboard. If so,
+    // focus should be set on the first option when the drop opens
+    // see set initial focus code in SelectContainer.js
     const [usingKeyboard, setUsingKeyboard] = useState();
 
     const onMouseDown = () => setUsingKeyboard(false);
--- a/src/js/components/Select/Select.js
+++ b/src/js/components/Select/Select.js
@@ -100,6 +100,20 @@ const Select = forwardRef(
     const inputRef = useRef();
     const formContext = useContext(FormContext);
     const { format } = useContext(MessageContext);
+
+    const [usingKeyboard, setUsingKeyboard] = useState();
+
+    const onMouseDown = () => setUsingKeyboard(false);
+    const onKeyPress = () => setUsingKeyboard(true);
+    useEffect(() => {
+      document.addEventListener('mousedown', onMouseDown);
+      document.addEventListener('keydown', onKeyPress);
+      return () => {
+        document.removeEventListener('mousedown', onMouseDown);
+        document.removeEventListener('keydown', onKeyPress);
+      };
+    }, []);
+
     // value is used for what we receive in valueProp and the basis for
     // what we send with onChange
     // When 'valueKey' sets 'reduce', the value(s) here should match
@@ -338,6 +352,7 @@ const Select = forwardRef(
               search={search}
               setSearch={setSearch}
               selected={selected}
+              usingKeyboard={usingKeyboard}
               value={value}
               valueKey={valueKey}
             >
--- a/src/js/components/Select/Select.js
+++ b/src/js/components/Select/Select.js
@@ -291,7 +291,8 @@ const Select = forwardRef(
         <StyledSelectDropButton
           ref={ref}
           a11yTitle={
-            `${placeholder}${
+            // let title = ariaLabel || a11yTitle || placeholder || undefined;
+            `${ariaLabel || a11yTitle || placeholder || undefined}${
               value
                 ? format({
                     id: 'select.selected',
@@ -299,10 +300,10 @@ const Select = forwardRef(
                     values: { currentSelectedValue: value },
                   })
                 : ''
-            }` ||
-            ariaLabel ||
-            a11yTitle
+            }`
           }
+          aria-expanded={Boolean(open)}
+          aria-haspopup="listbox"
           id={id}
           disabled={disabled === true || undefined}
           dropAlign={dropAlign}
@@ -317,8 +318,6 @@ const Select = forwardRef(
           onOpen={onRequestOpen}
           onClose={onRequestClose}
           onClick={onClick}
-          aria-haspopup="listbox"
-          aria-expanded={open}
           dropContent={
             <SelectContainer
               clear={clear}
--- a/src/js/components/Select/Select.js
+++ b/src/js/components/Select/Select.js
@@ -290,7 +290,19 @@ const Select = forwardRef(
       <Keyboard onDown={onRequestOpen} onUp={onRequestOpen}>
         <StyledSelectDropButton
           ref={ref}
-          a11yTitle={ariaLabel || a11yTitle}
+          a11yTitle={
+            `${placeholder}${
+              value
+                ? format({
+                    id: 'select.selected',
+                    messages,
+                    values: { currentSelectedValue: value },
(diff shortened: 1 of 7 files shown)


britt6612:
@MikeKingdom tagged you since you have a windows thought it would be nice to get your feedback from using voice over

MikeKingdom:
> @MikeKingdom tagged you since you have a windows thought it would be nice to get your feedback from using voice over

I only have NVDA. Does that work?

britt6612:
> NVDA

yes that works I figured you had sometime different so just want to verify on different devices if it still reads out what it should

src/js/components/Select/Select.js
```diff
@@ -100,6 +100,20 @@ const Select = forwardRef(
     const inputRef = useRef();
     const formContext = useContext(FormContext);
     const { format } = useContext(MessageContext);
+
+    const [usingKeyboard, setUsingKeyboard] = useState();
```
britt6612, line 104: wondering if we can add a comment to just say something simple of us having to do this to have better alignment with the HTML select accessibility

src/js/components/Select/Select.js
```diff
@@ -100,6 +100,20 @@ const Select = forwardRef(
...
+
+    const onMouseDown = () => setUsingKeyboard(false);
+    const onKeyPress = () => setUsingKeyboard(true);
+    useEffect(() => {
+      document.addEventListener('mousedown', onMouseDown);
+      document.addEventListener('keydown', onKeyPress);
+      return () => {
+        document.removeEventListener('mousedown', onMouseDown);
```
britt6612, line 112: nice these work good now!

britt6612 (approved):
Screen ready and keyboard is much better on my end

src/js/components/Select/Select.js
jcfilben: Added comment for this in latest commit

ericsoderberghp (approved):

MikeKingdom (approved):
It looks good. It seemed to work well in NVDA in chrome on windows. Mainly when I got confused or in trouble it was me not understanding some of the NVDA keyboard commands (left/right arrows do interesting things.)

amandadupell (approved):
looks good to me!!

PR #5942, opened 2022-02-10
Improve logic for when to normalize date in DateInput

### LINKED ISSUE #5920: DateInput selects previous Day on timezone UTC+1
### Expected Behavior

The date selected in the calendar should appear in the DateInputField.

### Actual Behavior

It gets the previous day of the selected date on the first try.

Behavior depends on the timezone and version of grommet. Error occurred in UTC+1 and grommet 2.20.1.
Grommet 2.18.0 works fine.
https://codesandbox.io/s/dateinput-vbv6g?file=/src/App.js

https://user-images.githubusercontent.com/74107477/151588876-c56b9cfc-c026-44f4-a594-3d654d1113f2.mov

### Your Environment

- Grommet version: 2.20.1
- Browser Name and version: chrome latest
- Operating System and version (desktop or mobile): CodeSandbox

taysea:
#### What does this PR do?
When DateInput value starts as undefined, the `onChange` logic should be in UTC relative to the local timezone of the user. In these instances, we don't want to perform the subsequent normalization logic. This PR introduces a state variable in dateinput that is passed ot the normalization functions. When it is true, we proceed with normalization. When it is false, we don't normalize because the date is already in UTC with respect to the timezone.

DateInput.js

Format Inline story in timezones before UTC and after (for example, pacific time and eastern european.

With FormatInline story in various timezones.

N/A did not touch tests aside from fixing a test that did not have `defaultValue` in ISO8601 format.

Closes #5920 

No.

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -261,6 +261,10 @@ const Calendar = forwardRef(
       };
     }, []);
 
+    // whether or not we should normalize the date based on the timestamp.
+    // will be set to false if the initial timestamp is undefined (meaning
+    // a user did not provide a defaultValue or value). in this case, we
+    // will just rely on the UTC timestamp and don't need to normalize.
     const [normalize, setNormalize] = useState(normalizeProp);
 
     // set activeDate when caller changes it, allows us to change
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -55,31 +55,35 @@ const activeDates = {
 
 const timeStamp = /T.*/;
 
-const formatSelectedDatesString = (date) => `Currently selected
+const formatSelectedDatesString = (date, normalize) => `Currently selected
   ${date?.map((item) => {
     let dates;
     if (!Array.isArray(item)) {
-      dates = `${formatToLocalYYYYMMDD(item)} `;
+      dates = `${formatToLocalYYYYMMDD(item, normalize)} `;
     } else {
       const start =
-        item[0] !== undefined ? formatToLocalYYYYMMDD(item[0]) : 'none';
+        item[0] !== undefined
+          ? formatToLocalYYYYMMDD(item[0], normalize)
+          : 'none';
       const end =
-        item[1] !== undefined ? formatToLocalYYYYMMDD(item[1]) : 'none';
+        item[1] !== undefined
+          ? formatToLocalYYYYMMDD(item[1], normalize)
+          : 'none';
       dates = `${start} through ${end}`;
     }
 
     return dates;
   })}`;
 
-const getAccessibilityString = (date, dates) => {
+const getAccessibilityString = (date, dates, normalize) => {
   if (date && !Array.isArray(date)) {
-    return `Currently selected ${formatToLocalYYYYMMDD(date)};`;
+    return `Currently selected ${formatToLocalYYYYMMDD(date, normalize)};`;
   }
   if (date && Array.isArray(date)) {
-    return formatSelectedDatesString(date);
+    return formatSelectedDatesString(date, normalize);
   }
   if (dates?.length) {
-    return formatSelectedDatesString(dates);
+    return formatSelectedDatesString(dates, normalize);
   }
 
   return 'No date selected';
@@ -225,6 +229,7 @@ const Calendar = forwardRef(
       header,
       locale = 'en-US',
       messages,
+      normalize,
       onReference,
       onSelect,
       range,
@@ -487,11 +492,15 @@ const Calendar = forwardRef(
         // output date with no timestamp if that's how user provided it
         let adjustedDate;
         if (!range) {
-          nextDate = formatDateToPropStructure(selectedDate, timestamp);
+          nextDate = formatDateToPropStructure(
+            selectedDate,
+            timestamp,
+            normalize,
+          );
           if (datesProp) {
             datesProp.forEach((d) => {
               if (!timeStamp.test(d)) {
-                adjustedDate = formatToLocalYYYYMMDD(nextDate);
+                adjustedDate = formatToLocalYYYYMMDD(nextDate, normalize);
                 if (d === adjustedDate) {
                   nextDate = undefined;
                 } else {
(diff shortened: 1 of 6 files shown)


src/js/components/DateInput/DateInput.js
```diff
@@ -152,9 +153,21 @@ const DateInput = forwardRef(
                 // clicking an edge date removes it
                 else if (range) normalizedValue = [nextValue, nextValue];
                 else normalizedValue = nextValue;
+                let nextNormalize = normalize;
+                if (timestamp === undefined) {
```
ericsoderberghp, line 157: Maybe a brief comment explaining that timestamp will be undefined if no defaultValue or value have been passed in, indicating that we should stay local if the user first picks a date via the Calendar.

src/js/components/DateInput/DateInput.js
```diff
@@ -152,9 +153,21 @@ const DateInput = forwardRef(
...
+                  nextNormalize = false;
+                  setNormalize(nextNormalize);
```
ericsoderberghp, line 159: Do we need to do something similar in the `onChange` for MaskedInput, when the user enters the first date by typing it?

ericsoderberghp (commented):
The basic direction seems to be holding up. It does feel a bit complicated, but I'm not sure how to simplify it.

src/js/components/DateInput/DateInput.js
taysea: Hmm good thought. let me see.

src/js/components/Calendar/utils.js
```diff
@@ -139,15 +140,20 @@ export const getTimestamp = (date) =>
 // can create undesired results. The standardizes the input value
 // for internal calculations
 // Reference: https://www.ursahealth.com/new-insights/dates-and-timezones-in-javascript
-export const normalizeForTimezone = (value, timestamp) => {
+
+// default this to true
+export const normalizeForTimezone = (value, timestamp, normalize = true) => {
```
MikeKingdom, line 145: So if normalize is false this degenerates to just convert the value toISOString()

MikeKingdom (commented):
I think it looks good. I'm trying to think of how to simplify this in general but it's hard to. I'd rather see the normalize flag not go so deep through so many calls and isolate it more to the initial look at defaultValue and value.

Really it feels like the crux of the problem is caring about timezone at all when all we're trying to do is choose a day/month/year. That seems to boil down to how we accept an iso8601 string with all that info in it and try to maintain that format through the api even though we don't really use the time aspect or intentionally convert between timezones. Changing that would probably be breaking.

src/js/components/Calendar/utils.js
taysea: yes, I can add a comment so that's documented.

> I think it looks good. I'm trying to think of how to simplify this in general but it's hard to. I'd rather see the normalize flag not go so deep through so many calls and isolate it more to the initial look at defaultValue and value.

Agreed it's not the cleanest to have to pass `normalize` in so many calls, but seemed like the most reliable solution for the time being. Eric and I had explored passing this "normalize" concept in the `timestamp` variable potentially (as opposed to adding another parameter to certain functions) but this caused issues because of the way timestamp is used to append to the date in various places.

> Really it feels like the crux of the problem is caring about timezone at all when all we're trying to do is choose a day/month/year. That seems to boil down to how we accept an iso8601 string with all that info in it and try to maintain that format through the api even though we don't really use the time aspect or intentionally convert between timezones. Changing that would probably be breaking.

Yes, difficult to not introduce breaking changes.

src/js/components/Calendar/Calendar.js
```diff
@@ -254,6 +261,8 @@ const Calendar = forwardRef(
       };
     }, []);
 
+    const [normalize, setNormalize] = useState(normalizeProp);
```
ericsoderberghp, line 264: Can we add a comment above this briefly explaining what `normalize` means and how it is used?

src/js/components/Calendar/Calendar.js
```diff
@@ -596,15 +576,49 @@ const Calendar = forwardRef(
...
-          // for internal calculations, adjust based on timestamp
-          // and user's local timezone
-          setDate(normalizeForTimezone(nextDate, timestamp));
+          setDate(nextDate);
         } else if (date && Array.isArray(date)) {
           setDate(nextDates);
         }
         setActive(new Date(selectedDate));
         if (onSelect) {
+          // output date with no timestamp if that's how user provided it
```
ericsoderberghp, line 585: Can we move all of this conversion logic into a separate utility function? I think that would make this callback code a bit easier to follow.

src/js/components/DateInput/DateInput.js
```diff
@@ -68,22 +68,22 @@ const DateInput = forwardRef(
...
-    } else if (typeof value === 'string' && value.length) {
-      timestamp = getTimestamp(value);
-    }
+      if (typeof value === 'string' && value.length) return getTimestamp(value);
+      return undefined;
+    }, [defaultValue, value]);
 
-    // normalize value based on timestamp vs user's local timezone
-    const normalizedDate = normalizeForTimezone(value, timestamp);
+    // whether or not we should normalize the date based on the timestamp
```
ericsoderberghp, line 82: maybe beef up this comment just a bit to explain when it is set false?

ericsoderberghp (commented):
Am I right that this set of changes is just carrying into Calendar the sorts of things we were working on within DateInput already?

Just a few comments, nothing major.

taysea:
> Am I right that this set of changes is just carrying into Calendar the sorts of things we were working on within DateInput already?
> 

Yes, you are exactly correct.

src/js/components/DateInput/DateInput.js
```diff
@@ -233,12 +249,32 @@ Use the icon prop instead.`,
               onChange={(event) => {
                 const nextTextValue = event.target.value;
                 setTextValue(nextTextValue);
+
+                let localTimestamp;
+                // get the UTC timestamp relative to the user's timezone
+                // once a date it complete
```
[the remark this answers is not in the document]
taysea: thanks, good catch.

src/js/components/DateInput/DateInput.js
taysea: Done!

src/js/components/Calendar/Calendar.js
taysea: Done.

src/js/components/Calendar/Calendar.js
taysea: I separated this out into its own function. Additionally, I was able to remove blocks of this logic that were never being entered (confirmed via manual testing and coverage analysis in jest).

MikeKingdom (approved):
Looks great. Thanks!

ericsoderberghp (approved):
nice improvements!

Issue #5913, opened 2022-01-25
Validation trigger at FormField level


dcatoday:
### Expected Behavior

It would be great if we could tell individual FormFields what validation trigger to use.  I am currently trying to validate on change for a password field to help the user fill it out while other fields are done on blue.

### Actual Behavior

Right now we set the validation trigger at the Form level so all fields share that logic.

We are only able to set at Form level.

### Steps to Reproduce
1. Create form
2. Add FormField
3. Cannot set validation trigger at Formfield level

### Your Environment

- Grommet version: 2.20.0
- Browser Name and version: Chrome 
- Operating System and version (desktop or mobile): windows desktop

ericsoderberghp:
Right now, the grommet Form validation is very much centered on the Form as opposed to the FormField. I am not opposed to exploring what this issue is asking for. But, I'm noting that it will likely involve a significant refactoring of how Form validation is done, which is something we've considered for other reasons as well.

g4rry420:
Can I work on this issue ?

PR #5943, opened 2022-02-11
Trigger validation using Form Field

g4rry420:

#### What does this PR do?
This pr provides the functionality to trigger validation using the Form Field.
Please, feel free to point out any mistakes.

Form.js, FormField.js

Test using the `TriggerValidationUsingFormField.js` storyName. I have tested these scenario:
- When component is mounted, 
- One FormField 's validateOn is different than another
- Basic functionality of validate on Form level also works

Using the `TriggerValidationUsingFormField.js`

#5913 

Yes

backwards compatible


src/js/components/Form/Form.js
```diff
@@ -455,6 +481,12 @@ const Form = forwardRef(
                 setTouched(nextTouched);
               }
 
+              if (!currentlySelectedElement[name]) {
+                setCurrentlySelectedElement({
```
MikeKingdom, line 485: I'm not really seeing how this is any different than `touched`. It seems to mirror it. Why the separate flag?

src/js/components/Form/stories/TriggerValidationUsingFormField.js
```diff
@@ -0,0 +1,97 @@
...
+export const TriggerValidationUsingFormField = () => (
+  <Grommet full theme={grommet}>
+    <Box fill align="center" justify="center">
+      <Box width="medium">
+        <Form
+          onReset={(event) => console.log(event)}
+          onSubmit={({ value }) => console.log('Submit', value)}
+        >
+          <FormField
+            label="First Name"
```
MikeKingdom, line 16: It would help to add some text to this story or maybe to the field labels themselves that say what it is doing (ie validate on blur etc). Otherwise, it's hard to figure out what the example is trying to portray

MikeKingdom (commented):
A couple of small comments, otherwise I think the functionality is good.

src/js/components/Form/Form.js
g4rry420: Yes, you are right. We don't need that, while iterating through different approaches to tackle this, I had added it then. But, I will remove it.

src/js/components/Form/stories/TriggerValidationUsingFormField.js
g4rry420: Sure, I will make the story descriptive.

MikeKingdom (approved):
Thanks!

g4rry420:
Any updates on this pr @ericsoderberghp @ShimiSun ?

ericsoderberghp:
Apologies for such a long delay in responding to this. Thanks for the new story. I tried interacting with it in the following manner:
- click on the Blur field
- tab to the Submit field - no validation is shown for the Blur field, I would have expected the "required" validation to be shown
- tab to the Change field - no validations are shown, which seems correct
- shift tab back to the Submit field - no validation issues are shown for the Change field, I was expecting it to show "required"
- shift tab back to the Blur field - no validations are shown, which seems correct
- type "a" - the "must be >1 character" validation shows up, this seems like it is being shown based on a change rather than a blur, which doesn't seem correct.

I didn't attempt to sift through the code to pinpoint where the issues might lie. But, I figured this testing might be helpful to start with.

g4rry420:
Thank you for pointing out these issues.

-  I have resolved the "required" validation not showing on "blur" validateOn value.
- For the same bug on the "onChange" validateOn value, it is currently exists on live deployment too https://storybook.grommet.io/?path=/story/input-form-validate-on-change--validate-on-change. I am not sure if this is something, we should take a look into more. If so, please tell me.

One more thing I notice, Suppose, there are 3 Form fields wrapped around the form component, if the user specifies the Form level `validate ="change"` prop and also specifies the Form Field level `validateOn="blur"` prop only on 1 form field, we would expect that the other two form fields automatically gets `change` validation behaviour. Is this the correct behaviour ?

g4rry420:
@ericsoderberghp It's been a long time since this pr is reviewed. I will really appreciate your's and Grommet maintainers efforts to take some time out of your busy schedule to review this pr.

> One more thing I notice, Suppose, there are 3 Form fields wrapped around the form component, if the user specifies the Form level validate ="change" prop and also specifies the Form Field level validateOn="blur" prop only on 1 form field, we would expect that the other two form fields automatically gets change validation behaviour. Is this the correct behaviour ?

Apart from this scenario, I don't see any issues with this pr. Please, let me know if you find any unexpected behaviour. Thanks :)
cc @ShimiSun

jcfilben:
Hi @g4rry420 sorry for the long delay here. Thanks for reaching out to get some eyes on this again! I think this is super close to the finish line.

The behavior looks good to me, I'm in the process of digging into the code a bit more to make sure things look good there.

> One more thing I notice, Suppose, there are 3 Form fields wrapped around the form component, if the user specifies the Form level validate ="change" prop and also specifies the Form Field level validateOn="blur" prop only on 1 form field, we would expect that the other two form fields automatically gets change validation behaviour. Is this the correct behaviour ?

Yes, I think that is the correct behavior. If validation is defined at the FormField level then that should override the Form level but only for that field. The other fields should align with the Form level validation.

src/js/components/Form/Form.js
```diff
@@ -536,8 +560,21 @@ const Form = forwardRef(
               validateArg,
               required,
             );
+
+            // priority is given to validateOn prop on formField, if it is
+            // undefined, then we will use the validate prop value of Form.
+            // The reason we don't want to add validateOn = "submit" here is
+            // because we don't have to trigger validation of "submit" field
```
g4rry420, line 567:
```suggestion
            // The reason we don't want to add validateOn = "submit" here is
            // because we don't want to trigger validation of "submit" field
```

g4rry420:
@jcfilben Above discussed behavior is also implemented. Please, let me know if you see any issues :)

jcfilben:
> @jcfilben Above discussed behavior is also implemented. Please, let me know if you see any issues :)

Tested this out and storybook and so far it is looking good! Still working my way through the code. In the meantime I think it would be nice to have some unit tests for this.

jcfilben (commented):
Thanks for adding tests! Just a minor comment on the story. Also looks like the bundlesize needs increased. This can be done in the package.json file. Just bump it up slightly to 144 kB.

jcfilben (approved):
Looks good!

ericsoderberghp:
I think the new story should be omitted from Chromatic snapshot evaluation, since it's there just for interaction. We can take care of this post-merge.

PR #5944, opened 2022-02-11
Restored arrow key functionality to Menu component

### LINKED ISSUE #5935: Allow arrow key navigation in Menu
### Expected Behavior
You should be able to navigate in the Menu component with the up and down arrow keys.

### Actual Behavior
You can't navigate with arrow keys, you can only tab through the Menu.

See Menu stories on storybook

### Steps to Reproduce
Try using arrow keys to navigate on a Menu

jcfilben:
#### What does this PR do?
Restores arrow key navigation to the Menu component

Tested with existing Menu stories

Closes #5935

No

Yes

Backwards compatible

### DIFF
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -1,9 +1,11 @@
 import React, {
+  useRef,
   forwardRef,
   useCallback,
   useContext,
   useMemo,
   useState,
+  useEffect,
 } from 'react';
 import styled, { ThemeContext } from 'styled-components';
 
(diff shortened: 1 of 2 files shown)


britt6612:
Screen Shot 2022-02-11 at 12 52 30 PM
can we get rid of this we may need to add something to the styled menu

britt6612:
I know previously that `menu` opened with the `down arrow` this seems strange to me I would expect that  you can also hit enter this can be a separate PR but I wanted to get your thoughts

jcfilben:
> can we get rid of this we may need to add something to the styled menu

Yes, I got rid of the outline in the latest commit

> I know previously that menu opened with the down arrow this seems strange to me I would expect that you can also hit enter this can be a separate PR but I wanted to get your thoughts

I agree, you should be able to open the select with 'enter'. I fixed this in the latest commit

src/js/components/Menu/Menu.js
```diff
@@ -26,6 +28,10 @@ const ContainerBox = styled(Box)`
     width: 100%;
   }
```
britt6612, line 30: maybe just add a comment just a suggestion

britt6612:
One small comment the rest looks good!

taysea (approved):

Issue #5882, opened 2022-01-07
Add Arrows to Diagram (Enhancement)


britt6612:
In the Diagram component, it would be nice to have an option to have `arrows` for the connections.
There are some instances where having an arrow would be good to have in Diagram.
<img width="333" alt="Screen Shot 2022-01-07 at 10 19 26 AM" src="https://user-images.githubusercontent.com/42451602/148581745-39cc014e-e993-4cdf-9bff-5ae229f78be9.png">

PR #5950, opened 2022-02-15
[WIP]: Added Arrow for Diagram Component

g4rry420:

NOTE: This PR is still work in progress.

#### What does this PR do?
This pr added the functionality to display the arrow in the Diagram Component. To achieve that, I have considered two types of arrows:
1. Filled
2. Unfilled

Currently, I am displaying a single arrow using the `marker` for all the connections provided. However, the only disadvantage of using this approach I found is if the different connections have different colors as per `Progressing` story.
We can resolve this problem by adding arrow property in the connections prop and looping through that and creating a separate `marker` for each connection.
Need your thoughts, on the approach.

I have also added two new properties to the connections prop and below are the usecases of these:
1. elementPosition: This prop describes if the target element is at the `top` or `bottom` position from the `from Element`. This prop  helps us in the decrease the line length of the connections which I found was over the Diamond diagram in the `Animated` story.
2. decreaseLineLength: As name describes, this props helps in decreasing the line of the connection.

Diagram.js

I have tested this using storybook

Using storybook

Didn't added tests yet

#5882 

UnFilled: 
Screenshot 2022-02-14 204853

Filled:
Screenshot 2022-02-13 170518

FilledTwo:
Screenshot 2022-02-13 170448

Between, Filled and FilledTwo, I am not sure which is good from design perceptive.
However, if none of them match design requirements, please feel free to point out the mistakes :)

Yes

backwards compatible


src/js/components/Diagram/index.d.ts
```diff
@@ -17,10 +17,16 @@ export type DiagramAnimationType =
...
   animation?: DiagramAnimationType;
+  arrow?: {
+    fill?: boolean;
+    color?: ColorType;
+  };
   connections: {
     anchor?: DiagramConnectionAnchor;
     animation?: DiagramAnimationType;
     color?: ColorType;
+    decreaseLineLength?: Number;
```
ericsoderberghp, line 28: I don't think the caller should have to figure this out. Instead, I think as part of measuring where the `fromTarget` and `toTarget` are on screen, in combination with `anchor` indicating how to draw the lines, we can calculate the appropriate edge location. For instance, if the center of the fromTarget is 100,100, the anchor is horizontal, and the size of the fromTarget is 20x20, we would draw the line starting at 110,100.

src/js/components/Diagram/index.d.ts
```diff
@@ -17,10 +17,16 @@ export type DiagramAnimationType =
...
+    elementPosition?: 'top' | 'bottom';
```
ericsoderberghp, line 29: I'm not sure what `elementPosition` drives. Is it which end of the line to put the arrowhead on? If so, I think we should tie this more to the targets. For example: `arrow: 'from'|'to'|true;`, with `true` meaning put an arrowhead on both ends.

src/js/components/Diagram/index.d.ts
```diff
@@ -17,10 +17,16 @@ export type DiagramAnimationType =
 
 export interface DiagramProps {
   animation?: DiagramAnimationType;
+  arrow?: {
```
ericsoderberghp, line 20: Hmm... there's a lot of potential variety in arrowhead styling. I think we should have our first pass at arrows not include any styling customization. We should pick one way to start with, use the color for the connection, and use the thickness for the connection to drive both the thickness of the arrowhead lines as well as their length. So, thinner lines get smaller arrowheads. I think we should start with something along these lines and consider further arrowhead customization at a later time.

ericsoderberghp (commented):
Sorry about the long delay in responding to this pull request. We are definitely still interested in it. I've started here with some comments around the API surface. I think we should iron that out first and then proceed with the implementation details. Thanks so much for your work so far on this.

src/js/components/Diagram/index.d.ts
g4rry420: Screenshot 2022-03-21 203413

If I don't use the `decreaseLineLength`, then shows over the target element. The `decreaseLineLength` is being used alongside the `elementPosition` prop which represents the targetElement (i.e, if it is at top or bottom) and based on this prop we provide the -ve or +ve value to decrease the line length. Please, check line number 46-48 for reference.

```
    const lineLength = ['top'].includes(elementPosition)
      ? decreaseLineLength
      : -decreaseLineLength;
```

However, I didn't quite understand how to do by your mentioned way by modifying into fromTarget values. Even if we did that, I imagine it'll be a breaking change ?
Please, let me know if I am missing something.

src/js/components/Diagram/index.d.ts
g4rry420: As mentioned above this prop is used differently, https://github.com/grommet/grommet/pull/5950#discussion_r831675683,
However, I haven't added the two way arrow thing, If you want I can add that too. Please, let me know.

src/js/components/Diagram/index.d.ts
g4rry420: Sounds, good.
Can you please, tell me know which arrowhead you are particularly refering to in these screenshots ? https://github.com/grommet/grommet/pull/5950#issue-1138842626

ericsoderberghp: Let's start with just the open arrow point, with just two legs, like is shown in the screenshot in the comment above.

src/js/components/Diagram/index.d.ts
ericsoderberghp: Regarding `elementPosition` If I have toTarget A centered at location 100,100 and fromTarget B centered at 200,200, each with height and width 20, I'm not sure what setting `elementPosition: 'top'` indicates. Does that mean the line should terminate at the top of B and the top of A?

My suggestion is to have a way to indicate which side of each target to orient the line from. In my example above, I think we need a way to describe whether to terminate the line at 90,100 for A or 100,110 for A. Same for B. Perhaps something like:
```
connections: [
  {
    from: { target: A, anchor: 'horizontal' },
    to: { target: B, anchor: 'vertical', arrow: true },
    type: 'rectilinear',
  }
],

```
which, given the coordinates and dimensions above, would draw from 90,100 to 200,100 to 200, 190.

src/js/components/Diagram/index.d.ts
ericsoderberghp: I would like to ensure that the API surface will allow for defining where the arrows are.

src/js/components/Diagram/Diagram.js
```diff
@@ -36,13 +43,36 @@ const COMMANDS = {
         `Q ${fromPoint[0] + offset},${midPoint[1] + offset} ` +
         `${midPoint[0] + offset},${midPoint[1] + offset} `;
     }
-    cmds += `T ${toPoint[0] + offset},${toPoint[1] + offset}`;
+    const lineLength = ['top'].includes(elementPosition)
+      ? decreaseLineLength
+      : -decreaseLineLength;
```
ShimiSun, line 48: This code block is triplicated. 
Can you please optimize it to a constant length value and reuse it across the file?

src/js/components/Diagram/Diagram.js
```diff
@@ -250,6 +299,49 @@ const Diagram = forwardRef(({ connections, ...rest }, ref) => {
       connections={paths}
       {...rest}
     >
+      <defs>
+        <marker
+          id="filledTwo"
+          markerWidth="10"
+          markerHeight="10"
+          refX="2"
+          refY="6"
```
ShimiSun, line 308: Can we add a few words on what do these magic numbers symbolize? Also, it might be more maintainable to extract them to a const variable and reuse the const across the page.

src/js/components/Diagram/index.d.ts
g4rry420: I have added the `arrow: 'from'|'to'|true;` options

src/js/components/Diagram/index.d.ts
g4rry420: @ericsoderberghp I have added your specified approach, but seems like it doesn't to work with the `horizontal` anchor when we have to select it from `COMMANDS`. Please, take a look and let me know if I am missing something.

src/js/components/Diagram/Diagram.js
```diff
@@ -39,10 +45,22 @@ const COMMANDS = {
     cmds += `T ${toPoint[0] + offset},${toPoint[1] + offset}`;
```
g4rry420, line 45: https://github.com/grommet/grommet/pull/5950#discussion_r842227385
How do we know which anchor should have priority ? The initial `anchor` or `anchor` defined inside the `fromTarget` and `toTarget` ?

PR #5951, opened 2022-02-15
Changed DataChart to fix an issue with passing Chart props per property when using areas

ericsoderberghp:
#### What does this PR do?

Changed DataChart to fix an issue with passing Chart props per property when using areas

DataChart.js

enhanced unit test and Stacked Areas story

Stacked Areas story

no updates to test approach

no

yes

backwards compatible

### DIFF
--- a/src/js/components/DataChart/DataChart.js
+++ b/src/js/components/DataChart/DataChart.js
@@ -560,6 +560,8 @@ const DataChart = forwardRef(
             return prop
               .map((cProp, j) => {
                 const pProp = cProp.property || cProp;
+                const propRest = typeof cProp === 'object' ? cProp : {};
+                delete propRest.property;
                 return (
                   <Chart
                     // eslint-disable-next-line react/no-array-index-key
@@ -570,6 +572,7 @@ const DataChart = forwardRef(
                     {...seriesStyles[pProp]}
                     {...chartProps[i]}
                     {...chartRest}
+                    {...propRest}
                     {...offsetProps}
                     type={type === 'areas' ? 'area' : 'bar'}
                     size={size}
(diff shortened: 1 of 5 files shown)


jcfilben:
In the stacked areas story when you click 'usage' or 'bonus' it disappears from the legend

https://user-images.githubusercontent.com/54560994/154148663-ccab4383-99f1-45c0-810f-6da57520f646.mp4

jcfilben:
If we set the opacity like this in chart property: `[{ property: 'usage', opacity: 'strong' }, ... ]` do we want the opacity to stay the same regardless of "focusing" on the other property (in this case "bonus")? In cases where we don't specify the opacity through `property: [{ opacity: ... }]` the opacity for the unselected option would change to be lighter. Notice below "bonus" is selected in the legend and the opacity for "usage" is still strong. Is this the behavior we want?
Screen Shot 2022-02-16 at 10 30 20 AM

ericsoderberghp:
> If we set the opacity like this in chart property: `[{ property: 'usage', opacity: 'strong' }, ... ]` do we want the opacity to stay the same regardless of "focusing" on the other property (in this case "bonus")? In cases where we don't specify the opacity through `property: [{ opacity: ... }]` the opacity for the unselected option would change to be lighter. Notice below "bonus" is selected in the legend and the opacity for "usage" is still strong. Is this the behavior we want?
> Screen Shot 2022-02-16 at 10 30 20 AM

For now, yes. we use opacity to help highlight the active data set. If the user wants to set their own opacity, we don't have another mechanism to show that. This is an existing behavior if you have multiple charts in a DataChart.

jcfilben (approved):

src/js/components/DataChart/DataChart.js
```diff
@@ -574,6 +574,8 @@ const DataChart = forwardRef(
             return prop
               .map((cProp, j) => {
                 const pProp = cProp.property || cProp;
+                const { property, ...propRest } =
+                  typeof cProp === 'object' ? cProp : {};
```
MikeKingdom, line 578: So would this basically be:
```jsx
const { property: pProp, ...propRest } =
  typeof cProp === 'object' ? cProp : { property: cProp };
```

PR #5952, opened 2022-02-15
Proposal for banner Notification

taysea:
#### What does this PR do?

Adds `banner` prop to Notification and theming.

### DIFF
--- a/src/js/components/Notification/Notification.js
+++ b/src/js/components/Notification/Notification.js
@@ -16,8 +16,17 @@ import { Text } from '../Text';
 
 import { NotificationType } from './propTypes';
 
-const Notification = ({ message, onClose, id, status, title, toast }) => {
-  const autoClose = toast?.autoClose === undefined ? true : toast.autoClose;
+const Notification = ({
+  banner,
+  message,
+  onClose,
+  id,
+  status,
+  title,
+  toast,
+}) => {
+  let autoClose = toast?.autoClose === undefined ? true : toast.autoClose;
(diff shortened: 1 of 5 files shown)


ericsoderberghp:
Why is the `banner` property needed? Or, what's the difference between `<Notification />` and `<Notification banner />`?

britt6612:
After looking into the accessibility aspect of the notifications I found that with a `Toast` we are using the [`role: log` ](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/log_role)
```
In contrast to other types of live region, a log is sequentially ordered and new information is only added to the end of the log. When this role is added to an element, the browser will send out an accessible log event to assistive technology products which can then notify the user about it.

```

This makes sense for a Toast because `A log is a type of live region where new information is added in meaningful order and old information may disappear.` Since a Toast notification is not critical I think this makes sense.

That being said a Banner notification is more critical and should use the `role: alert` this should be used for time-sensitive, information. More about this [alert](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role)

taysea:
> Why is the `banner` property needed? Or, what's the difference between `<Notification />` and `<Notification banner />`?

I hadn't realize the default Notification was intended to be used for banner. In the [latest commit](https://github.com/grommet/grommet/pull/5952/commits/abab6aaf01b227787a6b8b84ebeec0b8df00616d), I show what we would need to change to support the Design System designs while keeping the component backwards compatible for the open source community.

PR #5958, opened 2022-02-17
Enhance Notifications to allow for actions and added theming 

taysea:
#### What does this PR do?

Enhancements to Notification to support `actions`as well as additional theming capabilities such as the directional layout of the title/message.

src/js/components/Notification/Notification.js

Storybook -- staged PR has "Banner" storybook example added for review purposes, ensuring that we will be able to achieve HPE Design System designs. This storybook example can be removed after review is complete.

Added jest tests.

With Banner storybook example.

Related to [#2352](https://github.com/grommet/hpe-design-system/issues/2352)

Yes.

Yes. Enhanced Notification to support `onClick`, `href`, truncation, and directional layout of title/message.

Backwards compatible.


src/js/components/Notification/Notification.js
```diff
@@ -16,8 +17,56 @@ import { Text } from '../Text';
...
+      right: right || horizontal,
+    };
+  }
+
+  return [textStyle, closeButtonStyle];
+};
+
+const Notification = ({
+  message,
+  href,
```
ericsoderberghp, line 60: Why do we need to add `href`?

src/js/components/Notification/Notification.js
```diff
@@ -45,38 +94,74 @@ const Notification = ({ message, onClose, id, status, title, toast }) => {
   ]);
 
   const { icon: CloseIcon } = theme.notification.close;
-  const { icon: StatusIcon, color } = theme.notification[status];
+  const { icon: StatusIcon, background, color } = theme.notification[status];
   const { color: closeIconColor } = theme.notification.close;
 
+  const { direction, truncate } = toast
```
ericsoderberghp, line 100: will we fall back to base values if not specified under `theme.notification.toast`?

src/js/components/Notification/Notification.js
```diff
@@ -45,38 +94,74 @@ const Notification = ({ message, onClose, id, status, title, toast }) => {
...
+    ? theme.notification.toast
+    : theme.notification;
+
+  const TextWrapper = direction === 'row' ? Text : Box;
```
ericsoderberghp, line 104: Any reason to not always put inside a `Box`?

src/js/components/Notification/Notification.js
```diff
@@ -45,38 +94,74 @@ const Notification = ({ message, onClose, id, status, title, toast }) => {
...
+  let textPad;
+  let closeButtonPad;
+  if (onClose) [textPad, closeButtonPad] = adaptThemeStyle(pad, theme);
+  else textPad = pad;
+
   let content = (
     <Box
       {...theme.notification.container}
       {...(toast ? { ...theme.notification.toast.container } : {})}
+      background={
```
ericsoderberghp, line 123: what about building something like `callerProps` that captures properties passed in and then just expanding `{...callerProps}`?

src/js/components/Notification/Notification.js
```diff
@@ -45,38 +94,74 @@ const Notification = ({ message, onClose, id, status, title, toast }) => {
...
+              ) : (
+                <Text {...theme.notification.message} truncate={truncate}>
+                  {message}
+                </Text>
+              )}
+            </TextWrapper>
+          </Box>
+        </Box>
+      </Wrapper>
+      <Box pad={closeButtonPad}>
```
ericsoderberghp, line 154: Can we remove this Box entirely if `!onClose`?

src/js/components/Notification/Notification.js
```diff
@@ -45,38 +94,74 @@ const Notification = ({ message, onClose, id, status, title, toast }) => {
...
+      {/* separate from onClose button to avoid nested interactive elements */}
+      <Wrapper href={href} onClick={onClick}>
+        <Box direction="row" pad={textPad} flex>
+          <Box {...theme.notification.iconContainer}>
+            <StatusIcon color={color} />
+          </Box>
+          <Box {...theme.notification.textContainer}>
+            <TextWrapper {...textWrapperProps}>
+              {title && <Text {...theme.notification.title}>{title}</Text>}
+              {message && title && direction === 'row' && <Text> </Text>}
```
ericsoderberghp, line 142: Why do we have this `<Text> </Text>`?

src/js/components/Notification/Notification.js
```diff
@@ -45,38 +94,74 @@ const Notification = ({ message, onClose, id, status, title, toast }) => {
...
       {...(toast ? { ...theme.notification.toast.container } : {})}
+      background={
+        !toast && background
+          ? background
+          : theme.notification.container.background
+      }
+      // let internal box control pad so clickable region includes pad
+      pad={undefined}
       direction="row"
+      gap={onClose ? 'small' : undefined}
```
ericsoderberghp, line 131: Can we always leave `gap` and rely on only one child if `!onClose`?

src/js/components/Notification/Notification.js
```diff
@@ -16,8 +17,56 @@ import { Text } from '../Text';
...
+const Wrapper = ({ href, onClick, ...rest }) =>
+  href || onClick ? (
+    <Box flex>
+      <Button href={href} onClick={onClick} {...rest} />
+    </Box>
+  ) : (
+    <Fragment {...rest} />
+  );
+
+const adaptThemeStyle = (value, theme) => {
```
ericsoderberghp, line 29: I'm OK with the basic concept here.

src/js/components/Notification/Notification.js
```diff
@@ -16,8 +17,56 @@ import { Text } from '../Text';
 
 import { NotificationType } from './propTypes';
 
-const Notification = ({ message, onClose, id, status, title, toast }) => {
-  const autoClose = toast?.autoClose === undefined ? true : toast.autoClose;
+const Wrapper = ({ href, onClick, ...rest }) =>
```
ericsoderberghp, line 20: Can we name this more explicitly, like IconTextContainer?

src/js/components/Notification/Notification.js
taysea: This isn't handling properties passed in -- it's just using theme values. I had destructured `background` above in the file

src/js/components/Notification/Notification.js
taysea: Yes.

src/js/components/Notification/Notification.js
taysea: Will look into that.

taysea: that works.

src/js/components/Notification/Notification.js
taysea: We need the whole thing to be in text for the truncation to work properly (otherwise the title and message won't truncate as a single string, which made for strange wrapping behavior). However, we can leave the whole thing nested in another Box as it was prior.

src/js/components/Notification/Notification.js
taysea: Yes we will, that is what is seen under https://deploy-preview-5958--sad-tereshkova-173d07.netlify.app/?path=/story/visualizations-notification-status--status -- since base.js doesn't define `theme.notification.toast.direction` or t`heme.notification.toast.truncate`.

src/js/themes/base.d.ts
```diff
@@ -1039,8 +1039,13 @@ export interface ThemeType {
   };
   notification?: {
     container?: BoxProps;
+    direction?: 'column' | 'row';
+    separator?: string;
```
ericsoderberghp, line 1043: Wondering if we can get away with not adding `separator` yet.

src/js/components/Notification/index.d.ts
```diff
@@ -4,7 +4,7 @@ import { LayerPositionType } from '../Layer';
 export type StatusType = 'critical' | 'warning' | 'normal' | 'unknown';
 
 export interface NotificationProps {
```
ericsoderberghp, line 6: Need to add `href` and `onClick`?

src/js/components/Notification/Notification.js
```diff
@@ -45,39 +94,85 @@ const Notification = ({ message, onClose, id, status, title, toast }) => {
...
+                )}
+                {message}
+              </TextWrapper>
+            </Box>
+          </Box>
+        </Box>
+      </IconTextContainer>
+      {onClose && (
+        <Box pad={closeButtonPad}>
+          <Box {...theme.notification.textContainer}>
```
ericsoderberghp, line 166: It seems strange to have nested Boxes here, can they be merged?

src/js/components/Notification/Notification.js
```diff
@@ -45,39 +94,85 @@ const Notification = ({ message, onClose, id, status, title, toast }) => {
...
-            <Paragraph {...theme.notification.message}>{message}</Paragraph>
-          )}
+      {/* separate from onClose button to avoid nested interactive elements */}
+      <IconTextContainer href={href} onClick={onClick}>
+        <Box direction="row" pad={textPad} flex>
+          <Box {...theme.notification.iconContainer}>
+            <StatusIcon color={color} />
+          </Box>
+          <Box {...theme.notification.textContainer}>
+            <Box>
```
ericsoderberghp, line 152: Can we merge the nested Boxes?

ericsoderberghp (commented):
a few quick comments

src/js/themes/base.js
```diff
@@ -19,6 +19,7 @@ import { Volume } from 'grommet-icons/icons/Volume';
 import { VolumeLow } from 'grommet-icons/icons/VolumeLow';
 import { base as iconBase } from 'grommet-icons/themes/base';
 
+import { CircleInformation } from 'grommet-icons';
```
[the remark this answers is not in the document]
taysea: Good catch, thank you.

src/js/themes/base.d.ts
taysea: That's fine. I'll revert to the previous approach and add a comment since it seemed that was causing confusion on the first review.

src/js/components/Notification/Notification.js
taysea: The only reason I hadn't was for backwards compatibility. The `pad` on the outer box is technically coming from `theme.notification.container` (being applied here since we now separate the icon+text and button). so presumably someone could've had something defined for both theme objects and we would be reducing it down to just one of the `pad`s.

ericsoderberghp: So the `textContainer` might also have a `pad`, and we need to account for both?

taysea: Yes

src/js/components/Notification/Notification.js
```diff
@@ -16,8 +17,56 @@ import { Text } from '../Text';
 
 import { NotificationType } from './propTypes';
 
-const Notification = ({ message, onClose, id, status, title, toast }) => {
-  const autoClose = toast?.autoClose === undefined ? true : toast.autoClose;
+const IconTextContainer = ({ href, onClick, ...rest }) =>
+  href || onClick ? (
+    <Box flex>
```
ericsoderberghp, line 22: Any reason we can't have `<Box flex />` always be the IconTextContainer? If not, can we add a comment here explaining why?

src/js/themes/base.d.ts
```diff
@@ -1039,8 +1039,12 @@ export interface ThemeType {
   };
   notification?: {
     container?: BoxProps;
+    direction?: 'column' | 'row';
+    truncate?: boolean;
     toast?: {
```
ericsoderberghp, line 1044: I think we should support the background color for toast working similar to how direction and truncate work. As in, if not defined for toast they will use what the base notification has under the status keys. So, `theme.notification.toast.critical.background` would take precedence for toast before `theme.notification.critial.background` but if it isn't defined under `toast`, then the second one would be applied. I realize this means explicitly turning off the background for the HPE theme, but I think it mirrors general grommet patterns better.

src/js/components/Notification/Notification.js
taysea: We don't need it for the text to span the available width, so was trying to avoid adding another `div` to the DOM in "non-click" cases. I can add a comment.

src/js/themes/base.d.ts
taysea: That's fair, let me look at making those adjustments.

src/js/components/Notification/Notification.js
```diff
@@ -45,39 +90,92 @@ const Notification = ({ message, onClose, id, status, title, toast }) => {
...
+        key={action.label}
+        // create space between first anchor and text content
+        margin={
+          index === 0 && (message || title) ? { left: 'small' } : undefined
+        }
+        {...action}
+      />
+    ));
+
+  const MessageWrapper = direction !== 'row' ? Paragraph : Text;
```
ericsoderberghp, line 126: Wondering if we should just call this `Message`? Since we don't tend to think of "Text" being a "wrapper" for its string content.

src/js/components/Notification/Notification.js
```diff
@@ -45,39 +90,92 @@ const Notification = ({ message, onClose, id, status, title, toast }) => {
...
+
+  let actions;
+  let message = messageProp;
+
+  if (actionsProp)
+    actions = actionsProp.map((action, index) => (
+      <NotificationAnchor
+        key={action.label}
+        // create space between first anchor and text content
+        margin={
```
ericsoderberghp, line 119: Wondering if we can just put a "&nbsp;" after `{message}` below and avoid having to muck with the `margin` here.

src/js/components/Notification/Notification.js
```diff
@@ -45,39 +90,92 @@ const Notification = ({ message, onClose, id, status, title, toast }) => {
...
+      {/* separate from onClose button to allow "onClick" in the future and 
+        avoid nested interactive elements */}
+      <Box direction="row" pad={textPad} flex>
+        <Box {...theme.notification.iconContainer}>
+          <StatusIcon color={color} />
+        </Box>
+        <Box {...theme.notification.textContainer}>
+          <TextWrapper>
+            {title && <Text {...theme.notification.title}>{title}</Text>}
+            {message && title && direction === 'row' && <Text> </Text>}
```
ericsoderberghp, line 160: The DOM would be cleaner if we inserted `<>&nbsp;</>` instead.

I have a feeling you had this before but I didn't like it?

src/js/components/Notification/__tests__/Notification-test.js
```diff
@@ -155,4 +159,60 @@ describe('Notification', () => {
...
+
+    expect(asFragment()).toMatchSnapshot();
+    userEvent.click(
+      screen.getByRole('button', { name: 'Show Toast Notification' }),
+    );
+    expect(screen.getByText('Toast title')).toBeInTheDocument();
+    expect(onOpen).toHaveBeenCalled();
+  });
+
+  test('actions', () => {
```
ericsoderberghp, line 206: need a `global` test?

src/js/themes/base.js
```diff
@@ -1082,14 +1083,21 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
     notification: {
+      direction: 'column',
       container: {
         // any box props
         pad: { horizontal: 'small', vertical: 'xsmall' },
         background: {
           color: 'background-front',
         },
       },
+      global: {
```
ericsoderberghp, line 1094: wondering if we should default `global` to `direction: 'row'`?

src/js/components/Notification/Notification.js
taysea: To my eye, the space came across a bit too small -- I can screenshot a version of each side by side. Might just be my opinion

src/js/components/Notification/Notification.js
taysea: Don't think I had before, let me try that.

src/js/themes/base.js
taysea: Makes sense, will add.

src/js/components/Notification/Notification.js
taysea: With some margin
Screen Shot 2022-03-02 at 9 39 45 AM

With just a space (feel like it looks squished)
Screen Shot 2022-03-02 at 9 40 12 AM

ericsoderberghp: agreed with your margin

src/js/components/Notification/Notification.js
```diff
@@ -48,36 +93,107 @@ const Notification = ({ message, onClose, id, status, title, toast }) => {
...
+    return undefined;
+  }, [global, toast]);
+
+  let direction;
+  if (kind && theme.notification[kind].direction)
+    direction = theme.notification[kind].direction;
+  else direction = theme.notification.direction;
+
+  let background;
+  if (kind && theme.notification[status][kind]?.background)
```
ericsoderberghp, line 108:
```suggestion
  if (kind && theme.notification?.[status]?.[kind]?.background)
```

src/js/components/Notification/Notification.js
```diff
@@ -48,36 +93,107 @@ const Notification = ({ message, onClose, id, status, title, toast }) => {
...
+    background = theme.notification[status][kind]?.background;
+  else if (theme.notification[status].background)
```
ericsoderberghp, line 110:
```suggestion
  else if (theme.notification?.[status].background)
```

src/js/components/Notification/Notification.js
```diff
@@ -48,36 +93,107 @@ const Notification = ({ message, onClose, id, status, title, toast }) => {
...
+  }, [global, toast]);
+
+  let direction;
+  if (kind && theme.notification[kind].direction)
+    direction = theme.notification[kind].direction;
+  else direction = theme.notification.direction;
+
+  let background;
+  if (kind && theme.notification[status][kind]?.background)
+    background = theme.notification[status][kind]?.background;
```
ericsoderberghp, line 109:
```suggestion
    background = theme.notification[status][kind].background;
```

src/js/components/Notification/Notification.js
ericsoderberghp, line 108: Will this work to allow turning off the background by setting `background: undefined` in the theme?

src/js/components/Notification/__tests__/Notification-test.js
```diff
@@ -155,4 +159,78 @@ describe('Notification', () => {
...
+        },
+      },
+    };
+
+    const Test = () => {
+      const [visible, setVisible] = useState(false);
+      return (
+        <Grommet theme={theme}>
+          <TestNotification status="critical" />
+          <Button
```
ericsoderberghp, line 191: Do we need the button interaction for testing toast notifications? Can we just render the Notification?

src/js/components/Notification/__tests__/Notification-test.js
```diff
@@ -155,4 +159,78 @@ describe('Notification', () => {
...
+          actions={[{ href: '/some-link', label: 'Renew Subscription' }]}
+        />
+      </Grommet>,
+    );
+
+    const link = screen.getByRole('link', { name: 'Renew Subscription' });
+    expect(link).toHaveAttribute('href', '/some-link');
+  });
+
+  test('global', () => {
```
ericsoderberghp, line 227: Also need a test for `actions`. Just ensuring we have test coverage for all properties.

src/js/themes/base.d.ts
```diff
@@ -1054,22 +1059,32 @@ export interface ThemeType {
     };
     critical?: {
       icon?: React.ReactNode;
+      background?: BackgroundType;
       color?: ColorType;
```
ericsoderberghp, line 1063: What about `theme.notification.critical.toast.background` and `theme.notification.global.background`?

ericsoderberghp:
Also, can we rename the "Banner" story to "Global" and not have the page level one?

src/js/components/Notification/__tests__/Notification-test.js
taysea: actions is being tested in the test above.

src/js/components/Notification/__tests__/Notification-test.js
taysea: True, we don't need it. I have simplified the story.

src/js/components/Notification/Notification.js
taysea: You mean to explicitly set "toast" background (for example) to undefined and not fallback to other backgrounds?

src/js/components/Notification/__tests__/Notification-test.js
ericsoderberghp: duh :)

src/js/components/Notification/Notification.js
ericsoderberghp: yes, need a way to turn off default behavior. I thought we had an example of this elsewhere, perhaps not though.

src/js/themes/base.d.ts
taysea: Updated.

src/js/components/Notification/Notification.js
ericsoderberghp: For example, if we have `theme.notification.container.background: 'blue'` but we don't wan't for only critical toast notifications, we could do `theme.notification.critical.toast.container.background: undefined`.

ericsoderberghp: Well, I just skimmed through StyledButtonKind.js and didn't find an example like this. Maybe I'm just dreaming. I'm OK not worrying about it for now. We'll have some time on the stable branch to find any tricky things we don't cover before merging.

ericsoderberghp (approved):

PR #5960, opened 2022-02-18
Page and PageContent Components

jcfilben:
#### What does this PR do?
Adds Page and PageSection Components

Note: This PR is still a work in progress. I am still working on the theme structure, adding tests and creating storybook examples (note the storybook examples currently in this PR and not meant to be published), just wanted to get something out there for people to start looking at.

Tested with "Left Column" and "Narrow" storybook stories

Tests not added yet

Closes https://github.com/grommet/hpe-design-system/issues/2334

Yes

Yes

Backwards compatible


src/js/components/PageSection/PageSection.js
```diff
@@ -0,0 +1,27 @@
...
+        </Box>
+      </Box>
+    );
+  }
+  return (
+    <Box background={background} {...pageContext} {...rest}>
+      {children}
+    </Box>
+  );
+};
```
halocline, line 22: Could also do something like this. Set up the base, then add the wrapper if needed. That way if we need to modify the base, we do it in a single location.

```suggestion
  let contents = <Box background={background} {...pageContext} {...rest}>
      {children}
    </Box>;
  if (full) {
    contents = <Box background={background}>
        {contents}
      </Box>;
  }
  return contents;
};
```

src/js/components/Page/stories/CustomThemed/LeftColumn.js
```diff
@@ -0,0 +1,91 @@
...
+    <Box background="#35bd55">
+      <Heading>Global Header</Heading>
+    </Box>
+    <Page kind="testKind">
+      <PageSection>
+        <Header>
+          <Heading>Heading</Heading>
+        </Header>
+      </PageSection>
+      <PageSection full background="pink">
```
halocline, line 42: Theoretically, this should work for a background image as well. For peace of mind, can we do a background image as well?

src/js/components/Page/Page.js
```diff
@@ -0,0 +1,36 @@
...
+      width: theme.page[kind].width,
+      pad: theme.page[kind].pad[size],
+    }),
+    [theme, size, kind],
+  );
+
+  return (
+    <PageContext.Provider value={value}>
+      <Box
+        overflow="horizontal"
```
MikeKingdom, line 25: Seems like some of this stuff should be overrideable from props. This one seems a bit limiting.

src/js/components/Page/stories/CustomThemed/LeftColumn.js
jcfilben: Yes, I added a background image to the narrow story in the latest commit

src/js/components/Page/Page.js
halocline: Could we do something like:

```
const Page = ({ kind, overflow = "horizontal", ...rest }) => {
    ...
    <PageContext.Provider value={value}>
      <Box
        overflow={overflow}
     ...
```

That way we'd still get a smart default (since we are putting a min-width constraint), but the default can still be overridden.

jcfilben: I got rid of the overflow="horizontal" in the latest commit, but the user can set this by passing props. I also added an example of the user overriding the default `fill: 'horizontal'` prop in the Page/Narrow story.

src/js/themes/base.js
```diff
@@ -1142,6 +1142,44 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         color: 'status-unknown',
       },
     },
+    page: {
+      wide: {
+        alignSelf: 'center',
+        width: {
+          min: '336px', // 336 + 24 (margin) + 24 (margin) = 384 (e.g. 'medium')
```
britt6612, line 1149: wondering if we should use a t-shirt size here for grommet purposes then use the `336` specifically in the hpe theme open to discussion

src/js/components/Page/Page.js
halocline: I see that now. 

I think there is still value in having a "smart default" to accommodate the min-width scenario. But perhaps that could come from the theme as well.

src/js/components/Page/PageContext.js
```diff
@@ -0,0 +1,3 @@
+import { createContext } from 'react';
+
+export const PageContext = createContext({});
```
taysea, line 3:
```suggestion
export const PageContext = createContext({});

```

taysea: need new line at end of file i think

src/js/themes/base.js
jcfilben: Update the theme in latest commit to use t-shirt sizing

src/js/components/Page/PageContext.js
jcfilben: Looks like lint is removing the extra line when I try to add it

jcfilben: Fixed

src/js/components/PageSection/PageSection.js
```diff
@@ -0,0 +1,22 @@
+import React, { useContext } from 'react';
+import { Box } from '../Box';
+import { PageContext } from '../Page';
+import { PageSectionPropTypes } from './propTypes';
+
+const PageSection = ({ children, fullBackground, ...rest }) => {
```
taysea, line 6: Nice, I like this component and that the PageContext styling is handled for the user.

src/js/themes/base.d.ts
```diff
@@ -1073,6 +1073,19 @@ export interface ThemeType {
       color?: ColorType;
     };
   };
+  page?: {
+    [key: string]: {
```
taysea, line 1077: I'm not sure if the ? is needed, but just wondering

```suggestion
    [key: string]?: {
```

src/js/components/Page/stories/CustomThemed/LeftColumn.js
```diff
@@ -0,0 +1,92 @@
...
+      },
+      margin: { vertical: 'small', horizontal: 'small' },
+    },
+  },
+};
+
+export const LeftColumn = () => (
+  <Grommet theme={customTheme}>
+    <Box background="#35bd55">
+      <Heading>Global Header</Heading>
```
taysea, line 35: Should this be a heading? or just plain text

src/js/components/Page/stories/CustomThemed/LeftColumn.js
```diff
@@ -0,0 +1,92 @@
...
+    </Box>
+    <Page kind="customKind">
+      <PageSection>
+        <Header>
+          <Heading>Heading</Heading>
```
taysea, line 40: I'm finding these sections really large and it's making it hard for me to understand the page structure
```suggestion
          <Heading size="small">Heading</Heading>
```

src/js/components/Page/stories/CustomThemed/LeftColumn.js
```diff
@@ -0,0 +1,92 @@
...
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer
+          commodo gravida tincidunt. Nunc fringilla blandit tortor, id accumsan
+          nisi dictum quis. Aenean porttitor at mi id semper. Donec mattis
+          bibendum leo, interdum ullamcorper lectus ultrices vel. Fusce nec enim
+          faucibus nunc porta egestas. Fusce dapibus lobortis tincidunt.
+        </Paragraph>
+      </PageSection>
+    </Page>
+    <Box background="#35bd55">
+      <Heading>Global Footer</Heading>
```
taysea, line 83: same as comment on global header

src/js/themes/base.d.ts
jcfilben: Nope, it causes an error when added

src/js/components/Page/stories/CustomThemed/LeftColumn.js
jcfilben: The stories in this PR aren't meant to be published to storybook, they were more for just testing. I'm working on creating storybook examples for page that we can actually publish. I will push these up soon.

src/js/components/Page/Page.js
```diff
@@ -0,0 +1,43 @@
...
+import { PageContext } from './PageContext';
+import { ResponsiveContext } from '../../contexts/ResponsiveContext';
+import { PagePropTypes } from './propTypes';
+
+const Page = ({ kind, ...rest }) => {
+  const size = useContext(ResponsiveContext);
+  const theme = useContext(ThemeContext);
+
+  const value = useMemo(
+    () => ({
```
ericsoderberghp, line 13: What about structuring the theme to be `theme.page[kind][size].pad`? I think this would align a bit better with how we structure the theme namespace with state before property, so we can put all properties appropriate for a state together, as opposed to put all states for a property together.

This would also allow passing `theme.page[kind][size]` as the value for the PageContext, which would allow adding other properties to the theme and have them get passed along.

src/js/components/Page/Page.js
```diff
@@ -0,0 +1,43 @@
...
+      margin: theme.page[kind]?.margin?.[size],
+    }),
+    [theme, size, kind],
+  );
+
+  return (
+    <PageContext.Provider value={value}>
+      <Box
+        width={
+          theme.page[kind].width?.min
```
ericsoderberghp, line 27: Can we just pass `theme.page[kind].width`? It will handle whether it's a string or an object.

src/js/components/PageSection/index.d.ts
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
+import { BoxProps } from '../Box/index';
+import { BackgroundType } from '../../utils';
+
+export interface PageSectionProps {
+  fullBackground?: BackgroundType;
```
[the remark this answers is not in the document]
jcfilben: I have `fullBackground` as a separate prop because I wanted to allow setting the background on the PageContent through background in addition to setting a background that isn't confined by the page kind width. For example in the below picture the image is being set with the `fullBackground` prop and the pink background is being set with the `background` prop. So if we do `background={{ fill: "horizontal" }}` this would only allow setting either the full background or the constrained background not both.

Screen Shot 2022-02-18 at 1 52 05 PM

src/js/components/Page/Page.js
jcfilben: Yes, changed in latest commit

src/js/components/Page/Page.js
jcfilben: I ended up removing the width from this box, realized it wasn't needed here

src/js/components/PageSection/PageSection.js
```diff
@@ -0,0 +1,22 @@
...
+      {children}
+    </Box>
+  );
+  if (fullBackground) {
+    return <Box background={fullBackground}>{contents}</Box>;
+  }
+  return contents;
+};
+
+PageSection.displayName = 'PageSection';
```
ericsoderberghp, line 19: I am wondering about the term "Section". HTML5 has `<section />`, https://developer.mozilla.org/en-US/docs/Web/HTML/Element/section. But this doesn't seem to quite line up with it. Or, is the expectation that we would render this in the DOM as `<section />`? If not, I'm trying to think of another name that would convey the use. I think 80% of the time there will be only one, but sometimes there might be multiple for a Page. I was OK with `PageContent`, other ideas?

src/js/components/PageSection/index.d.ts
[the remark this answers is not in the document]
ericsoderberghp: Hmm... do we have an actual design example where we would want both? I'm thinking that would be very rare, and that the caller could always fill the PageSection with a `<Box fill background={} />` in that case.

src/js/components/PageSection/PageSection.js
jcfilben: Yeah I could see how the name PageSection could cause confusion with the section element. This was something I hadn't thought of. I agree that the majority of the time there will only be one "PageSection". I'm good with switching the name to PageContent

jcfilben:
> Hmm... do we have an actual design example where we would want both? I'm thinking that would be very rare, and that the caller could always fill the PageSection with a <Box fill background={} /> in that case.

Good point, I don't have a real world design example of where this might happen and agree it would be a rare case. I'll go back and remove the `fullBackground` prop and have something where the user can set the background to fill within the background prop `background={{ fill: "horizontal" }}`

src/js/components/PageSection/PageSection.js
ericsoderberghp: I was actually wondering if the 80% case would be only one PageContent and whether we could allow a Page to have a single PageContent built in automatically, perhaps as a property that defaults to do that but you could turn it off, thoughts?

halocline: I like this 👆 .

jcfilben: Hmm actually I'm thinking it might make sense to create a Page with separate PageContent sections for the header, main content, and footer so the 80% case might not always be one PageContent within Page.

ericsoderberghp: I'm not sure that's an 80% case though. If I look at the GLCP designs, they don't have that background variation, which I think is the main driver of having multiple sections.

jcfilben: Yeah that makes sense. I'll play around with having a PageContent built in automatically and a prop that you can switch this behavior on/off. I will push something up soon

MikeKingdom: The marketplace main page has a full width banner area that has a background image (no margins). Then the main content area below the banner is background-dark although the content adheres to max width of what would be `wide`. Not that the background in this case is full even though the content is limited width. I'm not really seeing the point of the background width being restricted since I'd just do that via whatever my content layout box is. See https://www.figma.com/file/RFHKQmOmyn5pvY2nqa1WIs/VAS-Marketplace?node-id=3799%3A49344

jcfilben: I pushed up some changes. I removed the `fullBackground` prop and now you can set `background={{ fill: 'horizontal' }}`. The Page has a PageContent section by default unless you set the `customizeContent` prop to true (still not quite sure about naming for the customizeContent prop).

With the design shared above that would be possible by setting `background={{ fill: 'horizontal', image: '...' }}` on the PageContent that is wrapping the "HPE GreenLake Marketplace" section. This would create the full background with the content maintaining its padding and restricted width.

I think it is useful to have the option to set the background of the PageContent without it horizontally filling the Page area. If you have a box within the PageContent and you set a background on that box its width won't be as wide as the PageContent since PageContent has some padding.

src/js/components/Page/Page.js
```diff
@@ -0,0 +1,42 @@
...
+      alignSelf: theme.page[kind]?.alignSelf,
+      fill: 'horizontal',
+      width: theme.page[kind]?.width,
+      ...theme.page[kind][size],
+    }),
+    [theme, size, kind],
+  );
+
+  let content = <PageContent {...contentProps}>{children}</PageContent>;
+  if (customizeContent) content = children;
```
ericsoderberghp, line 24: Maybe simplify to not remove PageContent but only add it when needed?
`const content = customizeContent ? children : <PageContent ...>{children}</PageContent>;`

src/js/components/Page/index.d.ts
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import { BoxProps } from '../Box/index';
+
+export interface PageProps {
+  contentProps?: BoxProps;
+  customizeContent?: boolean;
```
ericsoderberghp, line 6: Agreed on wanting a better name for this. One idea:
- `plain` - since you could put any children you want and they would fill the Page width, aligns with Button, default false

src/js/components/Page/index.d.ts
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import { BoxProps } from '../Box/index';
+
+export interface PageProps {
+  contentProps?: BoxProps;
```
ericsoderberghp, line 5: What's a scenario where we would need `contentProps`? I'm wondering if this is infrequent enough that callers should just use their own PageContent.

src/js/components/PageContent/index.d.ts
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import { BoxProps } from '../Box/index';
+import { ColorType } from '../../utils';
+
+export interface PageContentProps {
+  background?:
```
ericsoderberghp, line 6: use `BackgroundType` from utils

src/js/components/PageContent/index.d.ts
```diff
@@ -0,0 +1,27 @@
...
+        repeat?: 'no-repeat' | 'repeat' | string;
+        size?: 'cover' | 'contain' | string;
+        light?: string;
+        fill?: 'horizontal';
+      };
+}
+
+export interface PageContentExtendedProps
+  extends PageContentProps,
+    Omit<BoxProps, 'background'> {}
```
ericsoderberghp, line 23: Actually, can we skip this `Omit` and just accept all BoxProps?

ericsoderberghp (commented):
still working on the API surface, but a little bit of code comments too

src/js/components/PageContent/index.d.ts
jcfilben: I wanted to allow passing `background={{ fill: 'horizontal' }}` which is why I omitted the background prop from BoxProps since it doesn't include `fill`

src/js/components/Page/index.d.ts
jcfilben: Yeah that's reasonable, I'll remove contentProps

src/js/components/Page/Page.js
```diff
@@ -0,0 +1,41 @@
...
+import { PageContext } from './PageContext';
+import { ResponsiveContext } from '../../contexts/ResponsiveContext';
+import { PagePropTypes } from './propTypes';
+import { PageContent } from '../PageContent';
+
+const Page = ({ children, plain, kind, ...rest }) => {
+  const size = useContext(ResponsiveContext);
+  const theme = useContext(ThemeContext);
+
+  const value = useMemo(
```
ericsoderberghp, line 13: very minor, but can we re-label this to be `contextValue`? I'm so used to seeing `value` in Form and *Input contexts.

src/js/components/PageContent/index.d.ts
ericsoderberghp: Is there a way to `Extend` rather than `Omit`? I'm just wondering how to avoid duplicating the BackgroundType and missing changes to it here.

src/js/index.d.ts
```diff
@@ -40,6 +40,8 @@ export * from './components/Markdown';
 export * from './components/MaskedInput';
 export * from './components/Menu';
 export * from './components/Meter';
+export * from './components/Page';
+export * from './components/PageSection';
```
ericsoderberghp, line 44: So, is it "Content" or "Section"?

ericsoderberghp (commented):
just a couple of small items

src/js/index.d.ts
jcfilben: Oops is should be Content. Going back and fixing

src/js/components/Page/Page.js
```diff
@@ -0,0 +1,35 @@
...
+import { PagePropTypes } from './propTypes';
+
+const Page = ({ kind, ...rest }) => {
+  const size = useContext(ResponsiveContext);
+  const theme = useContext(ThemeContext);
+
+  const contentValue = useMemo(
+    () => ({
+      alignSelf: theme.page[kind]?.alignSelf,
+      fill: 'horizontal',
```
ericsoderberghp, line 15: Why is this 'fill' needed here? I would think PageContent could set it itself?

ericsoderberghp (approved):

Issue #5939, opened 2022-02-09
Clear button on Select is not accessibile


jcfilben:
### Expected Behavior
The user should be able to navigate to the clear button on Select using the keyboard.

### Actual Behavior
You cannot reach the clear button using the keyboard.

### URL, screen shot, or Codepen exhibiting the issue
See Select/Clear story https://storybook.grommet.io/?path=/story/input-select-clear--clear

### Steps to Reproduce
Try navigating with the keyboard to the clear button on the above story

### Your Environment
- Grommet version: 2.20.1

PR #5972, opened 2022-03-01
Fix/select-clear-access5939: allow for clear button to be accessed by keyboard

amandadupell:
#### What does this PR do?

Allow access to clear button via keyboard navigation

SelectContainer.js

Ensuring tests are consistent and not breaking

Storybook - Select - Clear Story

Didn't change existing tests

#5939 

No

No

Compatible


jcfilben (commented):
Nice progress, keyboard navigation works well except when using a screenreader. If you are using voiceover and navigating with the keyboard you can't get down to the last option. I think this may be because the clear button is within the listbox but isn't an 'option'

amandadupell:
> Nice progress, keyboard navigation works well except when using a screenreader. If you are using voiceover and navigating with the keyboard you can't get down to the last option. I think this may be because the clear button is within the listbox but isn't an 'option'

Could you upload a video of that experience? Here's what I'm seeing on my end:

https://user-images.githubusercontent.com/38971752/156670723-22f37fe2-22dc-4c4a-8e17-7f701ab022b6.mov

src/js/components/Select/SelectContainer.js
```diff
@@ -109,11 +105,14 @@ const SelectContainer = forwardRef(
       // need to wait for Drop to be ready
       const timer = setTimeout(() => {
         const optionsNode = optionsRef.current;
+        const clearButton = clearRef.current;
         if (onSearch) {
           const searchInput = searchRef.current;
           if (searchInput && searchInput.focus) {
             setFocusWithoutScroll(searchInput);
           }
+        } else if (clear && clearButton && clearButton.focus) {
```
jcfilben, line 114: I think we should only set focus on the clear button if it is at the top. If `clear.position = "bottom"` we should just set focus on the first option.

jcfilben:
When I am on the first option it looks like the clear button still has a focus indicator

Screen Shot 2022-03-04 at 3 18 05 PM

src/js/components/Select/SelectContainer.js
```diff
@@ -109,11 +105,17 @@ const SelectContainer = forwardRef(
...
```
jcfilben, line 114:
```suggestion
        } else if (clear && clearButton && clearButton.focus && clear.position !== 'bottom') {
```

src/js/components/Select/SelectContainer.js
```diff
@@ -109,11 +105,17 @@ const SelectContainer = forwardRef(
...
+          setFocusWithoutScroll(clearButton);
+          if (clear.position === 'bottom') {
+            setActiveIndex(0);
+          }
```
jcfilben, line 118: I think we can remove this and let the code below handle setting active index below
```suggestion
```

src/js/components/Select/SelectContainer.js
```diff
@@ -260,23 +263,42 @@ const SelectContainer = forwardRef(
...
         }
+        if (
+          clear &&
+          clear.position === 'bottom' &&
+          clearButton &&
+          nextActiveIndex >= options.length
+        ) {
+          setActiveIndex(options.length + 1);
+        }
+        console.log(activeIndex);
```
jcfilben, line 274: Don't forget to remove these

src/js/components/Select/SelectContainer.js
```diff
@@ -260,23 +263,42 @@ const SelectContainer = forwardRef(
           setActiveIndex(nextActiveIndex);
           setKeyboardNavigation(true);
         }
+        if (
+          clear &&
+          clear.position === 'bottom' &&
+          clearButton &&
+          nextActiveIndex >= options.length
+        ) {
+          setActiveIndex(options.length + 1);
```
jcfilben, line 272:
```suggestion
          setActiveIndex(options.length);
          setFocusWithoutScroll(clearButton);
```

amandadupell: yes i think `option.length` makes sense

as far as the `setFocusWithoutScroll`: i had this code implemented and i thought it would have worked but it still doesn't set the focus on the button. this was the code snippet i was still struggling with, as i thought doing `setFocusWithoutScroll` would handle it similarly to when the button is at the beginning.

i'll make the suggested changes from above and push so it's updated.

src/js/components/Select/SelectContainer.js
```diff
@@ -36,16 +36,12 @@ const SelectOption = styled(Button)`
   width: 100%;
 `;
 
-const ClearButton = ({ clear, onClear, name, theme, setFocus }) => {
+const ClearButton = ({ clear, onClear, name, theme, innerRef }) => {
```
ericsoderberghp, line 39: I think this can just use the standard `forwardRef()` from React. That would be a more consistent pattern to follow.

ericsoderberghp (commented):
Just one small comment.

ericsoderberghp (approved):

Issue #5964, opened 2022-02-24
Menu accessibility


Tbhesswebber:
Menus should be accessible.  In general, all menus should have proper roles, but I think it makes sense to enable a menu to accept an `a11yTitle` that is the name of the menu, using heuristics to name the aspects of the menu (e.g. "Open Primary Menu" on the button when closed should become "Primary Menu" on the menu and "Close Primary Menu" on the button when open in case the menu is a persistent one)

### Expected Behavior

Menus should have a role of menu by default and either intelligently apply aria attributes or allow greater out of the box support for them. TypeScript also shouldn't error when you override any of these values.

### Actual Behavior

I was working on writing a Cypress test and discovered that my search for an element with `role=menu` was failing when trying to target the opened menu. In attempting to set the role attribute on the menu, I discovered that TypeScript errors when you attempt to pass HTML attributes in the `dropProps` prop despite applying them correctly. 

### URL, screen shot, or Codepen exhibiting the issue

![image](https://user-images.githubusercontent.com/28069638/155434156-8712ba0d-d869-4746-bd6f-6b5278913565.png)

![image](https://user-images.githubusercontent.com/28069638/155434203-d14492b5-a7f0-4673-ae3a-0f8ce08f177c.png)

![image](https://user-images.githubusercontent.com/28069638/155434395-2a2cd8e0-dc12-4b0e-989c-6fe129d3d4c2.png)

If a repro is valuable, I can add it, but we're sprinting toward a release and I don't really have a ton of time to spare right now.

### Steps to Reproduce

1. Use a `Menu` in a TypeScript environment
2. Inspect the element and note the lack of a `role`
3. Try to apply the role and cry because TypeScript pulls out the red underline of fury

### Your Environment

- Grommet version: 2.20.1
- Browser Name and version: Chrome (latest)
- Operating System and version (desktop or mobile): Desktop MacOS

PR #5982, opened 2022-03-07
Improve Menu accessibility

jeferson-sb:
#### What does this PR do?
Add aria role and state attributes to Menu component to improve screen reader's experience, following specs of [WAI-ARIA](https://www.w3.org/TR/wai-aria-practices-1.1/#menu)

`src/js/components/Menu/Menu.js`

`yarn test`

`yarn storybook`
choose a Menu story and turn on your screen reader

Related to #5964 

backwards compatible


src/js/components/Menu/Menu.js
```diff
@@ -351,11 +357,12 @@ const Menu = forwardRef((props, ref) => {
 
                   return (
                     // eslint-disable-next-line react/no-array-index-key
-                    <Box key={index} flex={false}>
+                    <Box key={index} flex={false} role="none">
```
jcfilben, line 360: Do we need role="none" here?

jeferson-sb: This is because `Box` is the parent of each menuitem since a `menuitem` is not directly below a `menu` the accessibility tree might display a bit differently but I don't think it will change how it's announced.

jcfilben: Thanks for the explanation!

jcfilben (approved):

PR #5986, opened 2022-03-09
Change Notification Global story to not customize theme

ericsoderberghp:
#### What does this PR do?

Change Notification Global story to not customize theme

However, it looks like the `hpe` theme hasn't been published yet, so there's no background coloring.
Also, I'm wondering if the `base` theme should have some default background coloring, as that feels more consistent, like just using our `status-*` colors with some opacity.

Global.js

Global story

Global story

no test changes

Screen Shot 2022-03-09 at 10 48 03 AM

no

no

backwards compatible

### DIFF
--- a/src/js/components/Notification/__tests__/__snapshots__/Notification-test.js.snap
+++ b/src/js/components/Notification/__tests__/__snapshots__/Notification-test.js.snap
@@ -244,8 +244,7 @@ exports[`Notification global 1`] = `
   display: flex;
   box-sizing: border-box;
   max-width: 100%;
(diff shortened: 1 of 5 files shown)


MikeKingdom (approved):
Looks good. Are these really different colors than status-*? maybe simply because they are background with text potentially over them rather than foreground like an icon?

jcfilben (approved):
Looks good
I'm in favor of the base theme having default background coloring depending on the status

taysea (approved):
LGTM!

Issue #5967, opened 2022-02-25
DataTable placeholder should be visible below header


taysea:
### Expected Behavior

If placeholder is provided to DataTable, it should be visible and retain proper styling underneath the DataTable header.

### Actual Behavior

The content around the placeholder doesn't always respect the placeholders height / shrinking the DataTable can cause the placeholder to render behind the header. 

### URL, screen shot, or Codepen exhibiting the issue

One example is when DataTable is wrapped in a Box with overflow="auto", the placeholder is hidden behind a very small scrollable region:
https://codesandbox.io/s/datatable-placeholder-bug-i8d8j?file=/index.js

When shrinking browser window, the placeholder is not fully visible:
<img width="685" alt="Screen Shot 2022-02-25 at 2 33 35 PM" src="https://user-images.githubusercontent.com/12522275/155816115-dd32c6a2-debf-41ea-88fa-7a9e9543facc.png">

### Steps to Reproduce

1. Codesanbox example above.

### Your Environment

- Grommet version: 2.21.0
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): MacOS

MikeKingdom:
The placeholder is implemented via a `<caption>` . Normally the caption is rendered above the table but it is styled to be `position: absolute`. As a result, it doesn't contribute to the size calculation of the table. That means any table that doesn't specify a size may not leave space for the placeholder.

MikeKingdom:
Fixed by https://github.com/grommet/grommet/pull/5987

PR #5987, opened 2022-03-10
Fix DataTable placeholder to be visible

MikeKingdom:
Fix DataTable placeholder to be visible when the table height is sized by it's content but there is no content yet.

In cases where the DataTable has no items and the table has no size or constraint that causes it to be taller than it's natural size, a `placeholder` string or component may not be visible. The placeholder doesn't normally contribute to the DataTable's height since it is `position: absolute`. 

#### What does this PR do?

**UPDATE**:  Instead of attempting to fix table heights to account for a placeholder overlay it instead puts the placeholder in the `tbody` if there is no data. The makes sure the table is at least tall enough to show the placeholder. If there are items in the table then it still shows the placeholder as an overlay like before. This does have a small downside still that if the number of items are too small compared to the height of the placeholder, the placholder can still get clipped.

~~Make the DataTable at least tall enough to include any header, footer and placeholder. The `<table>` s minimum height is set to account for a placholder if present.~~

~~This change doesn't try to make the placeholder visible somehow if the table has content and the top portion of the table is scrolled out of the visible region. I believe that is a different and orthogonal feature that should be considered separately.~~

The placeholder continues to be implemented via a `<caption>` to preserve compatibility,

This PR does also remove a `placeholder` attribute that was leaking into the `<tbody>` DOM.

DataTable.js

Storybook, Jest and several manual test cases.

Modify the DataTable Placeholder storybook example to not include any data items in the table.

No change.

#5967 

No

Yes.

Backwards compatible.


MikeKingdom:
Looks like the Placeholder storybook example isn't getting stretched like it used to. I'll have to look closer at that

src/js/components/DataTable/DataTable.js
```diff
@@ -206,6 +208,19 @@ const DataTable = ({
...
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useLayoutEffect(() => {
+    if (placeholder && placeholderRef.current) {
+      const nextHeaderHeight = headerRef.current
+        ? headerRef.current.getBoundingClientRect().height : 0;
+      const nextFooterHeight = footerRef.current
+        ? footerRef.current.getBoundingClientRect().height : 0;
+      const nextMinHeight = nextHeaderHeight + nextFooterHeight +
+        placeholderRef.current.getBoundingClientRect().height;
+      if (minHeight !== nextMinHeight) setMinHeight(nextMinHeight);
```
ericsoderberghp, line 220: Rather than cause a re-render, I'm wondering if we should just set the `min-height` directly on the `placeholderRef.current` here.

src/js/components/DataTable/DataTable.js
```diff
@@ -206,6 +208,19 @@ const DataTable = ({
...
   });
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useLayoutEffect(() => {
+    if (placeholder && placeholderRef.current) {
+      const nextHeaderHeight = headerRef.current
+        ? headerRef.current.getBoundingClientRect().height : 0;
+      const nextFooterHeight = footerRef.current
+        ? footerRef.current.getBoundingClientRect().height : 0;
+      const nextMinHeight = nextHeaderHeight + nextFooterHeight +
```
ericsoderberghp, line 218: Why do we include the header and footer heights here but then also pass them as `top` and `bottom` to StyledPlaceholder? As with my other comment for the other layout effect, I'm wondering if we should render the StyledPlaceholder and then have a single layout effect that sets top, bottom, and min-height directly on the `placeholderRef.current`

src/js/components/DataTable/DataTable.js
```diff
@@ -206,6 +208,19 @@ const DataTable = ({
     if (nextScrollOffset !== scrollOffset) setScrollOffset(nextScrollOffset);
   });
 
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
ericsoderberghp, line 211: I feel like we've gotten in trouble for doing this in other places and am wondering if we find a way to avoid this.

MikeKingdom: Yup. I'll look at this more. In this case I do want it run after every render and it will only cause a re-render if it has to change the min height. Dependencies on this prevent that causing the size to be wrong on a resize. I think that is also true and existing defect with the existing layout effect that is updating the placeholder.top and .bottom based on the header and footer.

src/js/components/DataTable/DataTable.js
MikeKingdom: I'll look at that. Really we're wanting this element to be relative to the body and without a resize handler I think we're better off making this parented to the body, but I think that means it must be in a `<tr><td>`

src/js/components/DataTable/DataTable.js
MikeKingdom: I'll test that idea

src/js/components/DataTable/DataTable.js
```diff
@@ -490,6 +503,13 @@ const DataTable = ({
               setRowExpand={setRowExpand}
             />
           )}
+          {placeholder && (!items || items.length === 0) && (
+            <tbody>
```
ericsoderberghp, line 507: Will this result in two `tbody` elements in the DOM, since Body is also inserting one?

MikeKingdom: This work in progress does, but I was going to eliminate that if you think this idea is worth it. The extra empty tbody didn't really cause any visual issues but I wouldn't leave it there

MikeKingdom: :) Yup. It's basically columns.length + 1 for onSelect I think

ericsoderberghp:
Looking good. Can we also add a unit test for the no-items + placeholder case?

MikeKingdom:
> Looking good. Can we also add a unit test for the no-items + placeholder case?

Certainly. Good plan.

taysea (approved):
LGTM once we have the unit test

ericsoderberghp (approved):

PR #5996, opened 2022-03-15
Changed Box to fix an issue with desired height using an object

ericsoderberghp:
#### What does this PR do?

Changed Box to fix an issue with desired height using an object.

The prior code was awkwardly using `height={{ min, ..., max: ..., width: ... }}` rather than `height={{ min, ..., max: ..., height: ... }}`. It looks like a cut/paste issue.

utils function

Added a unit test.
Also verified with grommet-designer.

Followed existing test patterns.

yes

yes

backwards compatible


MikeKingdom (approved):
LGTM

PR #5997, opened 2022-03-15
Fix Calendar to account for Daylight Savings

jcfilben:
#### What does this PR do?
If the current date is within daylight savings and you select a date that is outside of daylight savings the date will be off by 1. This was causing some of the Calendar tests to fail after March 13th (when daylight savings began).

No
Yes
Backwards compatible


src/js/components/Calendar/utils.js
```diff
@@ -150,7 +158,17 @@ export const normalizeForTimezone = (value, timestamp, normalize = true) => {
   let localOffset = 0;
   if (normalize) {
     if (timestamp && typeof timestamp === 'string') {
-      hourDelta = parseInt(timestamp?.split(':')[0], 10);
+      const day = Array.isArray(value) ? undefined : new Date(value);
+      const today = new Date();
+      if (
+        day &&
+        !inDaylightSavings(day) &&
```
ericsoderberghp, line 165: What about the other direction, where `day` is inDaylightSavings but `today` isn't?

src/js/components/Calendar/utils.js
```diff
@@ -150,7 +158,17 @@ export const normalizeForTimezone = (value, timestamp, normalize = true) => {
...
+        day.getTimezoneOffset() > today.getTimezoneOffset()
+      ) {
+        hourDelta = parseInt(timestamp?.split(':')[0], 10) - 1;
```
ericsoderberghp, line 168: Rather than repeat the `parseInt()`, what about calculating `hourDelta` like before but then just adding or subtracting one from it if needed in the conditionals?

src/js/components/Calendar/utils.js
ericsoderberghp: Do we have to check whether `inDaylightSavings(day) !== inDaylightSavings(today)`?

yarn.lock
```diff
@@ -33,10 +33,10 @@
   dependencies:
     "@babel/highlight" "^7.16.7"
 
-"@babel/compat-data@^7.13.11", "@babel/compat-data@^7.16.4", "@babel/compat-data@^7.16.8", "@babel/compat-data@^7.17.0":
```
britt6612, line 36: why are we updating `yarn.lock` ?

jcfilben: Oops I accidentally pulled in the yarn lock changes I will go back and remove

src/js/components/Calendar/utils.js
```diff
@@ -150,7 +158,22 @@ export const normalizeForTimezone = (value, timestamp, normalize = true) => {
   let localOffset = 0;
   if (normalize) {
     if (timestamp && typeof timestamp === 'string') {
+      const day = Array.isArray(value) ? undefined : new Date(value);
+      const today = new Date();
       hourDelta = parseInt(timestamp?.split(':')[0], 10);
+      if (
+        day &&
+        !inDaylightSavings(day) &&
```
ericsoderberghp, line 166: Do we have to check whether inDaylightSavings(day) !== inDaylightSavings(today)?

jcfilben: this line `day.getTimezoneOffset() > today.getTimezoneOffset()` covers that, it one of the days (either day or today) is in daylight savings and the other is not the timezoneOffset will be different. If they are both in or outside of daylight savings the timezoneOffset will be the same

ericsoderberghp: Thanks. Can you add that to a comment? We seem to have trouble remembering some of these important details.

britt6612:
Screen Shot 2022-03-16 at 4 54 45 PM
need to look more into why this is failing

src/js/components/Calendar/utils.js
jcfilben: Yeah I'll add a comment

jcfilben:
> Screen Shot 2022-03-16 at 4 54 45 PM
> 
> need to look more into why this is failing

Marking as a draft so I can look into why this test is sometimes failing locally but works on circleci

britt6612:
sounds good I will look on my end as well

britt6612 (approved):
Looks good tested it in multiple different time zones and checks out

ericsoderberghp (approved):

Issue #6002, opened 2022-03-16
Move over more relevant DS examples to storybook


britt6612:
- [x] https://design-system.hpe.design/components/stack
- [x] https://design-system.hpe.design/components/tabs

PR #6003, opened 2022-03-16
Add Stack and Tab examples from Design System & cleaned up Tabs stories

jcfilben:
#### What does this PR do?
Add Stack and Tab examples from Design System & cleaned up Tabs stories

- Replaced Stack/Simple story with design system example
- Removed Tabs/AlignControls, alignControls prop was added in another story
- Removed Tabs/Plain, wasn't adding much value only showed how to use the plain prop
- Changed Tabs/Responsive to have a more realistic number of tabs while still showing responsive behavior.
- Replaced Tabs/Rich with an example of Tabs with Icons from the design system
- Changed Tabs/Scrollable to be a more realistic example
- Removed Tabs/Uncontrolled, we show an example of uncontrolled tabs in another story
- Changed Tabs/Controlled to a design system example

storybook

https://github.com/grommet/grommet/issues/6002

no

no
backwards compatible


britt6612:
Screen Shot 2022-03-16 at 4 50 11 PM
circle looks wide

britt6612:
other than the stack story LGTM

britt6612 (approved):

PR #6004, opened 2022-03-16
Fix/focus-range-4947: remove double focus when RangeInput is in FormField

### LINKED ISSUE #4947: Remove double focus on some FormField components
### Expected Behavior
Each FormField component should have a single focus on the Formfield itself rather than both the input field and the FormField itself. Similarly to how TextInput works inside FormField.

The issue should be fixed for:
* RadioButton
* CheckBox
* Group components
* RangeInput
* Select

### Actual Behavior
Each FormField component has two focus indicators, one for the FormField and the other for the input field.

https://user-images.githubusercontent.com/6320236/107573065-44558600-6baa-11eb-832f-4b3e77ae2328.mov

### Steps to Reproduce

1. https://storybook.grommet.io/?path=/story/input-form-custom--custom 
2. See the double focus behavior of the components listed above.

### Your Environment

- Grommet version: 2.16.3

amandadupell:
#### What does this PR do?

If there is a RangeInput in a FormField, only the FormField should have focus, but the RangeInput will have the hover behavior to show it is active.
If the RangeInput is not in a FormField or another focused element, it will have focus.

`RangeInput.js`

Manual Testing with Test story in Storybook

Can test in the Test story in Storybook or in the Custom Formfield Story

Used a similar approach as #4930 

closes #4947 

Not in FormField:
Screen Shot 2022-03-17 at 2 38 08 PM

In FormField:
Screen Shot 2022-03-17 at 2 38 15 PM

Maybe? To mention the focusIndictor prop similarly to in #4930 

No

Backwards compatible


taysea:
For elements like checkbox, checkboxgroup, etc. that let the formfield have the focus indicator instead of having it on themselves, we have given the element its hover treatment. I wonder if we should do the same for rangeinput? because it's a little bit confusing what it means when there's just the green underline / could be easy to miss.

britt6612:
> Details

this is only the case for hpe-theme correct? I think for grommet we can have the green outline to match the other components or are you thinking to also change this here

yarn.lock
```diff
@@ -33,10 +33,10 @@
   dependencies:
     "@babel/highlight" "^7.16.7"
 
-"@babel/compat-data@^7.13.11", "@babel/compat-data@^7.16.4", "@babel/compat-data@^7.16.8", "@babel/compat-data@^7.17.0":
```
britt6612, line 36: we shoudnt have any yarn lock changes

britt6612 (approved):
This looks good to me code wise I would just add that we make a separate Ticket to add a hover background for RangeInput like Taylor mentioned how we have for `checkbox` and other components so we can have the same behavior in the DS site. Something similar to [this](https://github.com/grommet/grommet/blob/master/src/js/components/CheckBox/StyledCheckBox.js#L28). Id be okay with doing that in a separate PR

jcfilben:
> This looks good to me code wise I would just add that we make a separate Ticket to add a hover background for RangeInput like Taylor mentioned how we have for `checkbox` and other components so we can have the same behavior in the DS site. Something similar to [this](https://github.com/grommet/grommet/blob/master/src/js/components/CheckBox/StyledCheckBox.js#L28). Id be okay with doing that in a separate PR


I think hover styling when focused & in a formfield was already added in StyledRangeInput.js in this PR
```
 ${(props) =>
      props.focus &&
      !props.focusIndicator &&
      css`
        box-shadow: 0px 0px 0px 2px
          ${normalizeColor(
            props.theme.rangeInput.thumb.color || 'control',
            props.theme,
          )};
      `}

**Review by jcfilben — APPROVED** [review]
Looks good!

File: src/js/components/RangeInput/stories/Test.js [hunk 2]
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
```
jcfilben, line 1: Actually can we remove this story? It is helpful for testing but I don't think we need to add it to storybook

amandadupell:
> > This looks good to me code wise I would just add that we make a separate Ticket to add a hover background for RangeInput like Taylor mentioned how we have for `checkbox` and other components so we can have the same behavior in the DS site. Something similar to [this](https://github.com/grommet/grommet/blob/master/src/js/components/CheckBox/StyledCheckBox.js#L28). Id be okay with doing that in a separate PR
> 
> I think hover styling when focused & in a formfield was already added in StyledRangeInput.js in this PR
> 
> ```
>  ${(props) =>
>       props.focus &&
>       !props.focusIndicator &&
>       css`
>         box-shadow: 0px 0px 0px 2px
>           ${normalizeColor(
>             props.theme.rangeInput.thumb.color || 'control',
>             props.theme,
>           )};
>       `}
> ```

yeah, this applies hover styling to the RangeInput when in a FormField!

britt6612:
> > This looks good to me code wise I would just add that we make a separate Ticket to add a hover background for RangeInput like Taylor mentioned how we have for `checkbox` and other components so we can have the same behavior in the DS site. Something similar to [this](https://github.com/grommet/grommet/blob/master/src/js/components/CheckBox/StyledCheckBox.js#L28). Id be okay with doing that in a separate PR
> 
> I think hover styling when focused & in a formfield was already added in StyledRangeInput.js in this PR
> 
> ```
>  ${(props) =>
>       props.focus &&
>       !props.focusIndicator &&
>       css`
>         box-shadow: 0px 0px 0px 2px
>           ${normalizeColor(
>             props.theme.rangeInput.thumb.color || 'control',
>             props.theme,
>           )};
>       `}
> ```

I meant the whole background grey that we have when we hover in `checkbox` this is just the hover color for the `rangeinput` thumb
Screen Shot 2022-03-17 at 2 02 35 PM

src/js/components/RangeInput/stories/Test.js
amandadupell: done!

src/js/components/RangeInput/StyledRangeInput.js
```diff
@@ -246,6 +246,16 @@ const StyledRangeInput = styled.input`
             )};
         }
       `}
+    ${(props) =>
+      props.focus &&
+      !props.focusIndicator &&
+      css`
+        box-shadow: 0px 0px 0px 2px
```
taysea, line 253: Would it be possible to extract this as a `hoverStyle()` function so it's not copy and pasted from the lines above?

taysea (approved):

ericsoderberghp (approved):

PR #6008, opened 2022-03-21
Fix/toast-position-5995

### LINKED ISSUE #5995: Toast misaligned in smaller screens
### Expected Behavior
The should appear centered in all screen widths.

### Actual Behavior
The toast appears left-aligned in smaller screen widths. It is correctly positioned in the center in larger screens.

image

### Steps to Reproduce
- Go to the https://v2.grommet.io/notification
- Open a Toast and resize the browser

amandadupell:
#### What does this PR do?

Fixes align behavior for toast container to ensure toast is always centered, regardless of screen width.

`base.js`

Visual/storybook testing

Storybook

closes #5995 

https://user-images.githubusercontent.com/38971752/159335436-0d8243bd-9d0e-4b2d-8a7d-11fb6f299a33.mov

Added new prop to theme `base.js` so need to update grommet docs to include this

No

Compatible


taysea:
This is looking good, but if I apply a position to the toast such as "top-right" then it still appears a little off on smaller screens:

Screen Shot 2022-03-21 at 11 33 29 AM

jcfilben (commented):
I think if the user sets a position on the toast notification we should use the specified position even in small screens instead of defaulting to center

src/js/components/Notification/Notification.js
```diff
@@ -63,9 +64,13 @@ const Notification = ({
...
   const autoClose =
     toast && toast?.autoClose === undefined ? true : toast.autoClose;
+  const size = useContext(ResponsiveContext);
   const theme = useContext(ThemeContext) || defaultProps.theme;
   const [visible, setVisible] = useState(true);
-  const position = useMemo(() => (toast && toast?.position) || 'top', [toast]);
+
+  const position = useMemo(() => 
+    (size === 'small') ? 'top' : (toast && toast?.position) || 'top', 
+    [size, toast]);
```
ShimiSun, line 73: A use case that will be tricky with this logic, is what would happen when the user defines an extended t-shirt sizing range (custom breakpoints), such as "xsmall"? 
https://storybook.grommet.io/?path=/story/utilities-responsivecontext-custom-breakpoints--custom-breakpoints

taysea: Good comment ^^ was thinking the same. (noticed we implemented this logic in NameValueList and will likely have to think of a better approach/mirror what we land on here)

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

PR #6012, opened 2022-03-22
Allow header cells to take full height when DataTable is paginated

### LINKED ISSUE #5968: DataTable Header cell hover indicator doesn't fill full cell height
### Expected Behavior

The clickable and hover indicated region of a header cell should always be the full height of the row. So, if there are some header cells with single line text and others that wrap to multiple lines, the clickable and hover indicator should take up the height of the tallest cell.

### Actual Behavior

When you resize your window, causing text in header cells to wrap, the header cells whose text remains as a single line does not take up the full height of the cell. Then, when you hover on them, the hoverindicator isn't filling the whole cell.

https://design-system.hpe.design/components/datatable#sorting

Screen_Recording_2022-02-25_at_3_24_54_PM_AdobeCreativeCloudExpress

### Steps to Reproduce

1. Open the table in the link above ^^ so that the table header cells all render in a single line.
2. Hover over the cells (note how they take full height)
3. Shrink your window width until some of the cells wrap to two lines
4. hover over the header cells (note how the single line header cells do not take full height)

### Your Environment

- Grommet version: 2.21.0
- Browser Name and version: Chrome 98
- Operating System and version (desktop or mobile): MacOS Monterey

taysea:
#### What does this PR do?
Allows fill prop to be passed through on paginated datatables. This will allow header cells to take the full height of the row. Previously, fill was explicitly set to undefined for paginated data tables which kept the table height at "auto" (meaning the table could never have a defined height).

DataTable.js

Added "Test" storybook. Resize the window until the labels wrap then hover over the DataTable headers. The hoverindicator should fill the full height of the header for both single line and multiline headers.

N/A didn't touch.

Closes #5968

No.

Backwards compatible.

### DIFF
--- a/src/js/components/DataTable/DataTable.js
+++ b/src/js/components/DataTable/DataTable.js
@@ -360,7 +360,7 @@ const DataTable = ({
     <Container {...containterProps}>
       <OverflowContainer {...overflowContainerProps}>
         <StyledDataTable
-          fillProp={!paginate ? fill : undefined}
+          fillProp={!paginate ? fill : true}
           {...paginatedDataTableProps}
           {...rest}
         >
--- a/src/js/components/DataTable/DataTable.js
+++ b/src/js/components/DataTable/DataTable.js
@@ -359,11 +359,7 @@ const DataTable = ({
   return (
     <Container {...containterProps}>
       <OverflowContainer {...overflowContainerProps}>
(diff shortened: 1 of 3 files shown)


taysea:
Closing this because the `undefined` is required for proper overflow behavior on Firefox.

Issue #5928, opened 2022-02-02
Not memoizing certain props for Drop causes unexpected scrolling on re-render


phinch:
Hi! This isn't exactly a bug per se, but it was kind of strange behavior with the Drop component that took me a while to debug on my end, so figured I'd share in case there is something that can be done about it.

### Expected/Actual Behavior

The two modes of behavior can be found [in this codesandbox](https://codesandbox.io/s/grommet-v2-template-forked-vv7jz?file=/index.js). There's more explanation within the CodeSandbox; but basically, if the drop content has scroll, not memoizing the `align` object prop passed into the `Drop` component causes the scroll of the Drop to reset to the top when it is re-rendered; when `align` is memoized (or not provided), this does not happen. I believe (but have not fully confirmed) that this is due to [this dependency array](https://github.com/grommet/grommet/blob/master/src/js/components/Drop/DropContainer.js#L295-L305) triggering the associated `useEffect` on each render, since `align` is an object and does not pass a shallow equality test when it is not memoized. I also encountered similar behavior with `onClickOutside` not being call-backified, which is why I think more strongly that it's connected to that dependency array.

I believe the memoized behavior is what is expected, so there's nothing inherently broken about the Drop. I just feel like this behavior is a little bit cryptic, and is a common pitfall since `align` is commonly a simple object that might not be frequently memoized. There also isn't really any documentation that I could find that explains that the consumer should take care to memoize some/all props being passed into the Drop - instead, I had to dig through the source code to find why my usage of the component was behaving unexpectedly. I think it would be cool if this Drop API were more intuitive, either through a code change or a documentation change - but I also can see the argument for just assuming that the user is memoizing and callbackifying props that are passed to Grommet more indiscriminately than I was. 

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-forked-vv7jz?file=/index.js

### Steps to Reproduce

Steps are outlined in the CodeSandbox above

### Your Environment

- Grommet version: 2.19.0
- Browser Name and version:
- Operating System and version (desktop or mobile):

ericsoderberghp:
Closed via documentation updates.

PR #6017, opened 2022-03-23
Fix/drop-memo-5928

amandadupell:
#### What does this PR do?

Updates stories for Drop and DropButton to use useMemo to define `align` prop to avoid re-renders. Aligns with documentation changes.

Storybook stories for Drop

Story testing

Storybook

#5928 

PR is open to update docs here: https://github.com/grommet/grommet-site/pull/360


src/js/components/Drop/DropContainer.js
```diff
@@ -292,10 +293,8 @@ const DropContainer = forwardRef(
         }
       };
     }, [
-      align,
```
MikeKingdom, line 295: I think removing these from the dependency array will cause other defects since the code inside will continue to use old values if either of those dependencies actually change. 

I'm guessing that really the  caller shouldn't be passing in objects that are constructed again each render or this very large useEffect function needs to be refactored so it doesn't have to depend on those props.

amandadupell: @MikeKingdom agreed about not removing from dependency array, i just asked @taysea about that and we were looking at how to memoize the props at the component level. i was looking at useMemo for the align prop, but still unsure if this makes the most sense since it relies on an empty dependency array

would refactoring require abstracting the logic depending on the align prop from that useEffect?

MikeKingdom: The short answer is it is too late to memoize within here if it's coming from a prop. If the prop coming in is always a different instance on each render, your memoize will re-memorize it for the same reason it was causing the useEffect to get re-triggered, even when the contains of the object in the prop are identical to before. Either the caller has to memoize it or the code here has to be refactored so that the useEffect doesn't have a copy of it. In practice that is a bit difficult, especially with this involved of a useEffect function

MikeKingdom: I tested a version of this where I flattened the align prop and used that as the dependencies and it worked fine. There's probably a cleaner way to do this or break down the useEffect into standalone bits. But the general idea is object props like this in dependencies need to really be flattened so that they don't unnecessarily retrigger the function involved.
My hack:

```jsx
  const alignLeft = align?.left;
  const alignRight = align?.right;
  const alignTop = align?.top;
  const alignBottom = align?.bottom;
  ...
  // replace align.xxx in the function with the copy above
  ...
, [ alignLeft, alignRight, alignTop, alignBottom, ... ]
```

ericsoderberghp:
Can we just document that callers should memo-ize any objects they pass in? And, ensure we do that ourselves within grommet code when we call Drop? I believe we tackle the same problem elsewhere the same way, rather than always munging what the caller passes in just to avoid this.

amandadupell:
> Can we just document that callers should memo-ize any objects they pass in? And, ensure we do that ourselves within grommet code when we call Drop? I believe we tackle the same problem elsewhere the same way, rather than always munging what the caller passes in just to avoid this.

I'm thinking that this might be the best approach. @MikeKingdom's approach solves this issue but it might not be necessary if we choose to better document the expectations of what a user should pass in and ensure that our examples are including this as best practice.

should i make changes within the grommet docs and our design system to outline this? @ericsoderberghp

MikeKingdom:
> Can we just document that callers should memo-ize any objects they pass in? And, ensure we do that ourselves within grommet code when we call Drop? I believe we tackle the same problem elsewhere the same way, rather than always munging what the caller passes in just to avoid this.

I've been on the fence about this kind of thing but I run into it enough and it results in pretty obscure issues that end up in products. It has made me more careful about what I do totally inline. However how often do you do things like `<Something pad={{ vertical: 'huge'}}>`? If that has a subtle render-related thing that's easy to stumble into.

src/js/components/Drop/stories/CustomThemed/Lazy.js
```diff
@@ -22,6 +22,23 @@ const LazyDrop = () => {
   const bottomLeftTargetRef = useRef();
   const bottomRightTargetRef = useRef();
 
+  const alignBottomLeft = useMemo(() => {
```
MikeKingdom, line 25: Eric mentioned this in the site example review, but since this is really a constant that doesn't depend on anything else it should outside the component function. It doesn't need memoizing really. It just needs to be actually constant.

ericsoderberghp (approved):

Issue #5978, opened 2022-03-04
Opening DateInput with 'space' shouldn't cause page to scroll down


jcfilben:
### Expected Behavior
Page shouldn't scroll down when opening DateInput by pressing 'space' key

### Actual Behavior
The page scrolls down

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/s/grommet-v2-template-forked-jy1h4z?file=/index.js

### Steps to Reproduce
In the codesandbox tab to DateInput and press space to open

### Your Environment

- Grommet version: 2.21.0
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile):

PR #6022, opened 2022-03-24
Prevent DateInput from scrolling down the page when opening with space

jcfilben:
#### What does this PR do?
If you open a DateInput by pressing space it will scroll down the page.

Tested with the following code:
```javascript
<Box width="medium" gap="medium">
  <DateInput format="dd/mm/yyyy" value={value} onChange={onChange} />
</Box>
<Box width="medium" height="xlarge" background ="orange" />
```
I put a colored box below the DateInput so you can see the page scroll down

Closes https://github.com/grommet/grommet/issues/5978

no

no

backwards compatible


taysea (approved):
LGTM

PR #6024, opened 2022-03-25
Enhance sticky header

britt6612:
#### What does this PR do?
This PR adds a `sticky` prop to header that will take a string `scrollup` this will show the header when a user scrolls up and hide the header when the user scrolls down.

ToDo List
- [x]  Add propTypes
- [x] Add ts

header.js
storybook
storybook 
Box BackGround story 
I will add a new story for header
need to add jest tests 

This enhancement came up in the DS site
https://github.com/grommet/hpe-design-system/issues/1353
closes https://github.com/grommet/hpe-design-system/issues/1353
none
yes 
I will need to make a PR on the grommet site
yes
 backwards compatible


src/js/components/Header/Header.js
```diff
@@ -1,19 +1,92 @@
...
+      if (Math.abs(scrollY - lastScrollY) < threshold) {
+        ticking = false;
+        return;
+      }
+      if (scrollY === 0) {
+        setScrollUp(false);
+        ticking = false;
+      } else setScrollUp(scrollY < lastScrollY);
+      // calcualte left, right, width, height
+      setHeight(`${containerRef.current.getBoundingClientRect().height}px`);
```
ericsoderberghp, line 39: Rather than storing these in separate state variables, how about putting them in a single object? For example:
`const [stickyStyles, setStickyStyles] = useState();`

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,92 @@
...
-    as="header"
-    direction="row"
-    flex={false}
-    justify="between"
-    gap="medium"
-    {...rest}
-    ref={ref}
-  />
-));
+  return (
```
ericsoderberghp, line 76: I think we still need to render a Box at the right height when showing the sticky header as a fixed DOM element.

src/js/components/Header/Header.js
britt6612: sounds good Ill clean this up the `width` was changing on me so I was going to look closer to that today

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,92 @@
-import React from 'react';
-
+import React, { useEffect, useState } from 'react';
 import { Box } from '../Box';
+import { useForwardedRef } from '../../utils';
+
+const Header = React.forwardRef(({ sticky, ...rest }, ref) => {
```
taysea, line 5: Are we going to support "sticky" as a boolean? such that adding `sticky` would make it always sticky, regardless of scroll position?

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,92 @@
...
+  const [left, setLeft] = useState();
+  const [scrollUp, setScrollUp] = useState(false);
+
+  useEffect(() => {
+    //   // threshold inital value set to 0
+    const threshold = 0;
+    //   // number of pixels the document is currently scrolled
+    let lastScrollY = window.pageYOffset;
+    //   // run event listener callback once
+    let ticking = false;
```
taysea, line 21: what does `ticking` mean?

src/js/components/Header/Header.js
ericsoderberghp: But, we allowed for this down the road if we decide to go that way

src/js/components/Header/Header.js
britt6612: So I had the `requestAnimationFrame` to calculate the new offset once the page is rendered after the scroll is complete. 
The ticking flag was to make sure to just run the event listener callback once in each `requestAnimationFrame` 
I know it's not the best word so I changed it to `runEventListener`

open to suggestions

src/js/components/Header/Header.js
britt6612: agree I think after we get some feedback on how this is it would be easy to implement the `sticky` boolean

src/js/components/Header/Header.js
britt6612: fixed

src/js/components/Header/Header.js
britt6612: I cleaned this section up

britt6612:
I can add `sicky="scrollup"` to our `TS` as well as `proptypes` if everyone is good with that naming.

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,86 @@
-import React from 'react';
-
+import React, { useEffect, useState } from 'react';
 import { Box } from '../Box';
+import { useForwardedRef } from '../../utils';
+
+const Header = React.forwardRef(({ sticky, ...rest }, ref) => {
+  const containerRef = useForwardedRef(ref);
+  const [stickyStyles, setStickyStyles] = useState();
```
ericsoderberghp, line 7: I'm wondering if we can simplify this further by removing `scrollUp` as state and just using `stickStyles`, setting it to `undefined` when we don't need it.

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,86 @@
...
+        runEventListener = true;
+      }
+    };
+
+    window.addEventListener('scroll', onScroll);
+
+    return () => window.removeEventListener('scroll', onScroll);
+  }, [containerRef, scrollUp]);
+
+  const stickyStyle = {
```
ericsoderberghp, line 52: Can we add these extra keys, like `position`, etc. when we call `setStickStyles` and then just use that object directly, rather than creating `stickyStyle` as a separate object?

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,86 @@
...
+    position: 'fixed',
+    zIndex: '1',
```
ericsoderberghp, line 54: I wonder if "1" is always going to be correct. For instance, if a Drop is open in the main content that is scrolling under the header, is there a chance the Drop will render above the Header? We should test this.

taysea: Good call, Grommet's Layer and Drop have zIndex set to 20 by default

src/js/components/Header/Header.js
britt6612: good thinking I can try this

src/js/components/Header/Header.js
britt6612: yes 👍

src/js/components/Header/Header.js
britt6612: good thinking Ill test this out with the Drop then if that is the case I can use the 20 default

ericsoderberghp: I think we want to make sure that it is above the Drop, not just equal to it. And, we might want to put this in the theme, especially if the Drop is.

britt6612: sounds good Ill handle this portion of it in the theme I believe Drop is in the Global.Drop do you have a preference where this would live?
`global.header.zIndex`
`theme.header.sticky.zIndex`

ericsoderberghp: I'm thinking `theme.header.sticky.zIndex`

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,83 @@
...
+
+    const updateScrollDir = () => {
+      const scrollY = window.pageYOffset;
+      if (Math.abs(scrollY - lastScrollY) < threshold) {
+        runEventListener = false;
+        return;
+      }
+      if (scrollY === 0) {
+        runEventListener = false;
+        setStickyStyles();
```
halocline, line 27: The end result is the same for the state value, but I think supplying undefined or null is more clear in the intent.

```suggestion
        setStickyStyles(undefined);
```

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,83 @@
...
+      }
+      if (scrollY < lastScrollY) {
+        setStickyStyles({
+          width: `${containerRef.current.getBoundingClientRect().width}px`,
+          height: `${containerRef.current.getBoundingClientRect().height}px`,
+          left: `${containerRef.current.getBoundingClientRect().left}px`,
+          right: `${containerRef.current.getBoundingClientRect().right}px`,
+          zIndex: `${theme.header?.sticky?.zIndex}`,
+          position: 'fixed',
+          top: '0',
```
halocline, line 37: Let me know when you've added animation, or if you'd like to pair up on it.

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,83 @@
...
+        });
+      } else setStickyStyles();
```
halocline, line 39:
```suggestion
      } else setStickyStyles(undefined);
```

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,83 @@
...
+    window.addEventListener('scroll', onScroll);
+
+    return () => window.removeEventListener('scroll', onScroll);
+  }, [containerRef, stickyStyles, theme.header?.sticky?.zIndex]);
+
+  return (
+    <>
+      {sticky === 'scrollup' && (
+        <Box
+          height={stickyStyles && stickyStyles.height}
```
halocline, line 61: a bit cleaner?

```suggestion
          height={stickyStyles?.height}
```

src/js/components/Header/propTypes.js
```diff
@@ -0,0 +1,9 @@
+import PropTypes from 'prop-types';
+
+let PropType = {};
+if (process.env.NODE_ENV !== 'production') {
+  PropType = {
+    sticky: PropTypes.oneOf(['scrollup']),
```
halocline, line 6: Saw the earlier thread about whether a boolean should be included. Fine with this direction based on the discussion.

src/js/components/Header/stories/Sticky.js
```diff
@@ -0,0 +1,81 @@
...
+  PageContent,
+  Paragraph,
+} from 'grommet';
+
+export const Sticky = () => (
+  // Uncomment <Grommet> lines when using outside of storybook
+  // <Grommet theme={...}>
+  <Page kind="narrow">
+    <PageContent background="light-3">
+      <Header sticky="scrollup" background="light-4" pad="small">
```
halocline, line 19: We'd likely have the background fully bleed across.

```suggestion
    <PageContent background={{color: "light-3", fill: "horizontal"}}>
      <Header sticky="scrollup" background="light-4" pad={{vertical: "small"}}>
```

src/js/components/Header/stories/Sticky.js
```diff
@@ -0,0 +1,81 @@
...
+          <Anchor label="Profile" href="#" />
+        </Nav>
+      </Header>
+      <Paragraph>
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer commodo
+        gravida tincidunt. Nunc fringilla blandit tortor, id accumsan nisi
+        dictum quis. Aenean porttitor at mi id semper. Donec mattis bibendum
+        leo, interdum ullamcorper lectus ultrices vel. Fusce nec enim faucibus
+        nunc porta egestas. Fusce dapibus lobortis tincidunt.
+      </Paragraph>
```
halocline, line 32: If we are going to have a lot of text, would it make sense for us to use it? Just an idea.

```suggestion
      <Paragraph>
        To maximize screen real-estate, the Header on this page scrolls out of view as the user moves down the page. However if the user scrolls upwards, the Header is revealed and fixed atop the window. On long pages this behavior allows easy access to the Header's content, such as navigation or menus, while preventing the Header from obscuring content on mobile devices or in smaller windows.
      </Paragraph>
```

src/js/themes/base.js
```diff
@@ -912,6 +912,11 @@ export const generate = (baseSpacing = 24, scale = 6) => {
     grommet: {
       // extend: undefined
     },
+    header: {
+      sticky: {
+        zIndex: '1000',
```
halocline, line 917: Can we add a comment for why this value is what it is and how we arrived at the value? 1000 seems a bit excessive.

Reference zIndex of Drop and Layer.

halocline (commented):
With zIndex being a factor, I think it would be worthwhile to show a story with a page Header combined with a Layer, such as a side-drawer layer. Where is the Header revealed when a Layer is open? Atop the Layer? Below? If the Layer is a scrollable region, do scroll events within it trigger the Header reveal?

src/js/components/Box/stories/Background.js
```diff
@@ -1,83 +1,95 @@
...
-          'url(https://images.unsplash.com/photo-1487088678257-3a541e6e3922?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=2167&q=80)',
-      }}
-    >
-      image + color
-    </Box>
-    <Box background="dark-1" pad="medium">
-      <Box background="#FFFFFF08" pad="small">
-        low opacity on dark background
+  <>
+    <Header background="light-4" pad="small">
```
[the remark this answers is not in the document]
britt6612: Yes this was temporary

src/js/components/Header/propTypes.js
britt6612: yes we discussed that if we want to enhance this more we will add a boolean but for now we are going to just have the `scrollup`

src/js/themes/base.js
britt6612: yes It is a little excessive Drop and Layer are both 20 in the theme so I can change this to 50. I wanted to test out Global Header with this and see how that would layout

src/js/components/Header/stories/Sticky.js
halocline: I was hoping for something more like:
Screen Shot 2022-03-29 at 3 47 19 PM

halocline: This will expose some challenges with PageContent + sticky Header.

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,83 @@
...
+  }, [containerRef, stickyStyles, theme.header?.sticky?.zIndex]);
+
+  return (
+    <>
+      {sticky === 'scrollup' && (
+        <Box
+          height={stickyStyles?.height}
+          width={stickyStyles && stickyStyles.width}
+        />
+      )}
```
halocline, line 64: What is this providing for us?

britt6612: Eric had mentioned having `Box` in the dom 
I think we still need to render a Box at the right height when showing the sticky header as a fixed DOM element.

src/js/components/Header/stories/Sticky.js
```diff
@@ -0,0 +1,93 @@
...
+  PageContent,
+  Paragraph,
+} from 'grommet';
+import { Power, User } from 'grommet-icons';
+
+export const Sticky = () => (
+  // Uncomment <Grommet> lines when using outside of storybook
+  // <Grommet theme={...}>
+  <Page kind="narrow">
+    <Header direction="column" sticky="scrollup" background="light-4">
```
taysea, line 20: This feels awkward to me. I'm also not sure it's best practice to have nested `<header>` tags in the DOM. It would make more sense more me if line22 with the second header was just implemented as a box with appropriate props.

britt6612: good point that would require us coping most all of the `props` from Header to a Box so not sure about that either but would make more sense than having 2 `Headers` in the dom but also more of a chance for users to forget and not get the right behavior

ericsoderberghp: It's tough, because we want half of the Header containing the PageContent and half of it within the PageContent. Some other options:
- make Header smarter such that includes it's own PageContent via something like `<Header page />`?
- Use `as={Box}` for the inner Header
- ???

britt6612: adding `as={Box}` would be a lot easier then adding all of the other styles

halocline: > * make Header smarter such that includes it's own PageContent via something like `<Header page />`?

I had been thinking something along these lines ☝️ , where Header would consume PageContext in the same manner that PageContent does.

britt6612:
I made changes in which the user would have to pass `sticky={{ type: 'scrollup', fill: true }}` outside of PageContent if they want the Header too expand the whole page. This is shown in the Sticky story. If they want the header to be inline with PageContent then they can just pass `sticky={{ type: 'scrollup', }}` This is shown in the test story.

There is some `Box` clean up I can do but want to make sure this direction is good before I start code clean up.

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,141 @@
...
+          left: `${containerRef.current.getBoundingClientRect().left}px`,
+          right: `${containerRef.current.getBoundingClientRect().right}px`,
+          zIndex: `${theme.header?.sticky?.zIndex}`,
+          position: 'fixed',
+          transition: 'top 0.6s',
+          top: '0',
+        });
+      } else
+        setStickyStyles({
+          top: `-${containerRef.current.getBoundingClientRect().height}px`,
```
taysea, line 44: I feel like sometimes when I'm scrolling down, I don't see the animation. Do we need to add the transition back here?

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,141 @@
...
+            width={stickyStyles && stickyStyles.width}
+            background={background}
+            style={stickyStyles}
+            {...rest}
+            ref={containerRef}
+          >
+            <Box
+              align="center"
+              direction="row"
+              fill
```
taysea, line 84:
```suggestion
              fill="horizontal"
```

taysea: Would it just be cleaner to import `PageContent` and have this be:

<PageContent {...defaultHeaderProps} {...rest} />

taysea: Feels like we're repeating logic and work that PageContent already knows how to do

src/js/components/Header/Header.js
britt6612: oh really Ill take a look

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,141 @@
...
-    as="header"
-    direction="row"
-    flex={false}
-    justify="between"
-    gap="medium"
-    {...rest}
-    ref={ref}
-  />
-));
+const Header = React.forwardRef(({ sticky, background, ...rest }, ref) => {
```
taysea, line 7: Wondering if the signature of `sticky` should be:

sticky: 'scrollup' | { type: 'scrollup', fill: 'horizontal' }

src/js/components/Header/Header.js
britt6612: that's a possibility if we like this direction I can clean that up

ericsoderberghp:
Rather than add `fill`, I'm wondering if we should just have a separate `PageHeader` that always has `PageContent`? Then, we could use `Header` for the `Layer` case.

britt6612:
> 

So would the `PageHeader` have the sticky prop as well similar to Header

ericsoderberghp:
> > Rather than add `fill`, I'm wondering if we should just have a separate `PageHeader` that always has `PageContent`? Then, we could use `Header` for the `Layer` case.
> 
> So would the `PageHeader` have the sticky prop as well similar to Header. Would you rather I separate the two PRs or add a new PageHeader or have it all in here?

Yes, the real difference at the moment would be the that PageHeader would use PageContent

src/js/components/Header/Header.js
taysea: i've attached a video, notice when I scroll down how the header abruptly disappears. In order to see this, you have to be scrolled down the page a bit (not right at the top).

https://user-images.githubusercontent.com/12522275/161622911-fc6d478c-cff3-4d3d-b025-8a517d3196ae.mov

taysea: And then just a general question, is the position still "fixed" when we scroll down or what does `top` do we we don't have position fixed?

britt6612: I updated this let me know what you think

src/js/components/Header/Header.js
```diff
@@ -26,21 +26,19 @@ const Header = React.forwardRef(({ sticky, background, ...rest }, ref) => {
         runEventListener = false;
         setStickyStyles(undefined);
```
taysea, line 27: When I first scroll down, the header abruptly disappears (position is not fixed yet then changes to fixed). I think this is caused by a similar issue to the other comment I had made previously that you resolved by always having position fixed and transition when the user has scrolled down (whether they are scrolling up or down).

That being said is there any reason we are not just always having the transition, position fixed, etc. when a sticky header is desired? Feels like we need that to be persistent in order for the animation to work and I'm not seeing the benefit of removing the sticky styles when we are scroll position 0.

Notice how the header abruptly disappears on my first scroll down but after that the transition works as expected (when I'm not at scrollY = 0), but then when I get back to scrollY = 0 at the end the header abruptly comes in.

https://user-images.githubusercontent.com/12522275/161646529-de011523-12de-4b17-832f-04b2daee677e.mov

britt6612: I agree I changed the logic around this I think the transition is better now. @ericsoderberghp I would like to hear your thoughts around this direction as well. 

I found that this is not working when in a `Layer` because of the `window.addEventListener` is not being triggered since it is a modal trying to solve for this now if we are good with the transition on the `Header`

ericsoderberghp: Agreed that the transitions shouldn't be removed without having fully run.

One potential reason to not use position fixed all of the time is that could prevent the Header's size from being influenced by its container. For instance, if, for whatever reason, the caller turns off `flex={false}` for the Header, such that a Box that contains it determines it's height. Or, perhaps the caller puts the Header inside a Grid area (we should test this scenario, since the added Box might not have the `gridArea` set) and the height is controlled by the Grid row.

Would it work to not add the sticky version until scrolling up and then delay removing the sticky version until after the transition finishes when scrolling down or reaching the top?

ericsoderberghp: that’s what I’d recommend trying

taysea:
Noticing that the animation doesn't work when I scroll down (it just abruptly disappears):

https://user-images.githubusercontent.com/12522275/163480783-73dfaa0c-947c-44c5-9b29-5fc3bdc993cc.mov

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,62 @@
...
+    const updateScrollDir = () => {
+      // target is its scroll parent
+      const target = findScrollParent(containerRef.current);
+      const container = containerRef.current;
+
+      if (target && sticky === 'scrollup') {
+        const nextScrollTop =
+          target === document ? window.pageYOffset : target.scrollTop;
+        if (scrollTop - nextScrollTop <= 0) {
+          container.style.top = `-${
```
taysea, line 20: Think we need transition here still (and maybe position sticky? not sure on that) in order for the transition to work when user scrolls down.

src/js/components/Header/stories/Sticky.js
```diff
@@ -0,0 +1,157 @@
...
+        </Grid>
+        <Paragraph>
+          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer
+          commodo gravida tincidunt. Nunc fringilla blandit tortor, id accumsan
+          nisi dictum quis. Aenean porttitor at mi id semper. Donec mattis
+          bibendum leo, interdum ullamcorper lectus ultrices vel. Fusce nec enim
+          faucibus nunc porta egestas. Fusce dapibus lobortis tincidunt.
+        </Paragraph>
+      </PageContent>
+      <Box pad="large" align="center">
```
taysea, line 101: Could we potentially move this higher up on the story?

src/js/components/Header/Header.js
britt6612: aw yeah I was trying to reset those but looks like I would have to put it back for it to work

src/js/components/Header/stories/Sticky.js
britt6612: yeah no problem!

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,61 @@
...
-));
+const Header = React.forwardRef(({ sticky, ...rest }, ref) => {
+  const theme = useContext(ThemeContext);
+  const containerRef = useForwardedRef(ref);
+  useEffect(() => {
+    let scrollTop = 0;
+    const updateScrollDir = () => {
+      // target is its scroll parent
+      const target = findScrollParent(containerRef.current);
+      const container = containerRef.current;
```
taysea, line 14: More just a variable naming suggestion. Would it make sense to rename:
`container` --> `header`
`containerRef` --> `headerRef`

?

britt6612: yeah that seems approprite

britt6612: done

taysea (approved):
Nice work! : )

jcfilben (commented):
Minor comments on the story, functionality looks good

src/js/components/Header/stories/Sticky.js
```diff
@@ -0,0 +1,157 @@
...
+export const Sticky = () => {
+  const [open, setOpen] = useState(false);
+  const onOpen = () => setOpen(true);
+  const onClose = () => setOpen(undefined);
+
+  // eslint-disable-next-line no-unused-vars
+  const onSubmit = ({ value, touched }) => {
+    // Your submission logic here
+  };
+  return (
```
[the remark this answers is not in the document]
britt6612: done

jcfilben (approved):
Looks good!

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,59 @@
...
+          header.style.top = '0px';
+          header.style.zIndex = `${theme.header?.sticky?.zIndex}`;
+          header.style.transition = 'top 0.6s';
+        }
+        scrollTop = nextScrollTop;
+      }
+    };
+
+    updateScrollDir();
+    window.addEventListener('resize', updateScrollDir);
```
ericsoderberghp, line 34: Can we wrap these addEventListeners() inside a test for `sticky === 'scrollup'`? I think that would make our code a bit more efficient.

src/js/themes/base.d.ts
```diff
@@ -820,6 +820,9 @@ export interface ThemeType {
   grommet?: {
     extend?: ExtendType;
   };
+  header?: {
+    zIndex?: string;
```
ericsoderberghp, line 824: I think this should be `header?: { sticky?: { zIndex?: string; } }`

ericsoderberghp (commented):
A couple of small things.

src/js/themes/base.d.ts
britt6612: aw yes you are correct

src/js/components/Header/Header.js
britt6612: good thoughts doesn't need to run otherwise

src/js/themes/base.d.ts
britt6612: done

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,61 @@
...
+    };
+
+    if (sticky === 'scrollup') {
+      console.log('hello');
+      updateScrollDir();
+      window.addEventListener('resize', updateScrollDir);
+      window.addEventListener('scroll', updateScrollDir, true);
+    }
+    return () => {
+      window.removeEventListener('resize', updateScrollDir);
```
ericsoderberghp, line 40: I think we need the tests here too, to align with how they are added above

src/js/components/Header/Header.js
```diff
@@ -1,19 +1,61 @@
...
+          header.style.top = '0px';
+          header.style.zIndex = `${theme.header?.sticky?.zIndex}`;
+          header.style.transition = 'top 0.6s';
+        }
+        scrollTop = nextScrollTop;
+      }
+    };
+
+    if (sticky === 'scrollup') {
+      console.log('hello');
```
ericsoderberghp, line 34: remove :)

src/js/components/Header/Header.js
britt6612: your to fast pushed up now

src/js/components/Header/Header.js
britt6612: done

ericsoderberghp (approved):

PR #6026, opened 2022-03-25
fix(range): enables range wheel event override

### LINKED ISSUE #6025: Range input scroll behavior should be optional
### Expected Behavior

Developers should be able to opt-out of the range-input's scroll behavior.

### Actual Behavior

Range inputs always have the scroll behavior even if you attempt to override via props or any other method.

https://codesandbox.io/s/grommet-rangeinput-scroll-bug-rkvucc?file=/src/App.tsx

### Steps to Reproduce

Skipping this section due to the repro - will definitely fill it out if needed

### Your Environment

- Grommet version: 2.21.0
- Operating System and version (desktop or mobile): macOS

### Additional Context

Additional context can be found in [this slack conversation](https://grommet.slack.com/archives/C09QQEG69/p1648134314121579)

Tbhesswebber:
#### What does this PR do?

Enables a user to opt out of the mouse wheel behavior by setting the `onWheel` event on a range input. 

[This slack conversation](https://grommet.slack.com/archives/C09QQEG69/p1648134314121579) and then the RangeInput test file

Added tests for both the overridden and default scroll behavior

Manual testing can be done with the story in storybook

^ I used the same patterns already in the file

This feels to me like it's a less ideal solution than adding a toggle prop for `overrideScrollBehavior` (or something similar)

Closes #6025

Yes

Yes

Backwards compatible

### DIFF
--- a/src/js/components/RangeInput/RangeInput.js
+++ b/src/js/components/RangeInput/RangeInput.js
@@ -36,11 +36,11 @@ const RangeInput = forwardRef(
     },
     ref,
   ) => {
-    const themeContext = useContext(ThemeContext) || defaultProps.theme;
+    const theme = useContext(ThemeContext) || defaultProps.theme;
     const formContext = useContext(FormContext);
     const [focus, setFocus] = useState(focusProp);
 
-    const isScrollEnabled = themeContext?.rangeInput?.wheel !== false;
+    const isScrollEnabled = theme?.rangeInput?.wheel !== false;
 
     const [value, setValue] = formContext.useFormInput({
       name,
--- a/src/js/components/RangeInput/RangeInput.js
+++ b/src/js/components/RangeInput/RangeInput.js
@@ -6,6 +6,9 @@ import React, {
   useEffect,
 } from 'react';
 
+import { ThemeContext } from 'styled-components';
+import { defaultProps } from '../../default-props';
+
 import { FormContext } from '../Form/FormContext';
 import { StyledRangeInput } from './StyledRangeInput';
 import { RangeInputPropTypes } from './propTypes';
@@ -33,8 +36,11 @@ const RangeInput = forwardRef(
     },
     ref,
   ) => {
+    const themeContext = useContext(ThemeContext) || defaultProps.theme;
     const formContext = useContext(FormContext);
     const [focus, setFocus] = useState(focusProp);
+
+    const isScrollEnabled = themeContext?.rangeInput?.wheel !== false;
 
     const [value, setValue] = formContext.useFormInput({
       name,
@@ -77,6 +83,9 @@ const RangeInput = forwardRef(
 
     const handleOnWheel = (event) => {
       const newValue = parseFloat(value);
+      if (!isScrollEnabled) {
+        return;
+      }
       if (event.deltaY < 0) {
         setRangeInputValue(newValue + step);
       } else {
@@ -86,8 +95,9 @@ const RangeInput = forwardRef(
     // This is to make sure scrollbar doesn't move
     // when user changes RangeInput value.
     const handleMouseOver = () =>
-      setScroll({ x: window.scrollX, y: window.scrollY });
-    const handleMouseOut = () => setScroll({ x: null, y: null });
+      isScrollEnabled && setScroll({ x: window.scrollX, y: window.scrollY });
+    const handleMouseOut = () =>
+      isScrollEnabled && setScroll({ x: null, y: null });
 
     return (
       <StyledRangeInput
--- a/src/js/components/RangeInput/__tests__/RangeInput-test.tsx
+++ b/src/js/components/RangeInput/__tests__/RangeInput-test.tsx
@@ -7,7 +7,6 @@ import 'regenerator-runtime/runtime';
 
 import { Grommet } from '../../Grommet';
 import { RangeInput } from '..';
-import { ThemeContext } from '../../../contexts/ThemeContext';
 import { ThemeType } from '../../../themes';
 
 describe('RangeInput', () => {
@@ -162,7 +161,7 @@ describe('RangeInput', () => {
     const onChange = jest.fn();
     const { getByDisplayValue } = render(
       <Grommet theme={theme}>
-          <RangeInput min={0} max={10} step={1} value={5} onChange={onChange} />
+        <RangeInput min={0} max={10} step={1} value={5} onChange={onChange} />
       </Grommet>,
     );
 
--- a/src/js/components/RangeInput/__tests__/RangeInput-test.tsx
+++ b/src/js/components/RangeInput/__tests__/RangeInput-test.tsx
@@ -7,6 +7,8 @@ import 'regenerator-runtime/runtime';
 
 import { Grommet } from '../../Grommet';
 import { RangeInput } from '..';
+import { ThemeContext } from '../../../contexts/ThemeContext';
+import { ThemeType } from '../../../themes';
 
 describe('RangeInput', () => {
   test('should have no accessibility violations', async () => {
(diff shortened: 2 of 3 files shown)


jcfilben (commented):
This looks good! Thanks for including jest tests for this.

I think the storybook story is good for testing but we don't need to commit it as a storybook example

jcfilben (approved):

ericsoderberghp:
I'm fine with having the capability to disable the mouse wheel behavior for changing the range input. However, I don't think we should handle this by using the presence of the `onWheel` property. Everywhere else in grommet, if the caller passes in an event handler, we make sure we call our internal event handler function as well as the passed in one. I don't think we ever skip the internal behavior.

I think it would be more consistent to handle this via a theme property since I would expect you would want all instances of RangeInput to behave consistently across an app. Perhaps something like: `theme.rangeinput.wheel = false`. Thoughts?

jcfilben:
> I think it would be more consistent to handle this via a theme property since I would expect you would want all instances of RangeInput to behave consistently across an app. Perhaps something like: theme.rangeinput.wheel = false. Thoughts?

Yeah this is a good point, that makes sense to me

jcfilben:
Hey @Tbhesswebber just wanted to check in and see if this is something you are still interested in working on. I think this pr just needs changed so that the mouse wheel behavior can be controlled in the theme and then this will be ready to go! If you are no longer interested in working on this I think we can go ahead and close the pull request.

Tbhesswebber:
Ah, sorry! I never saw any kind of notification that there were responses after yours looking for input. I can definitely refactor to that.

Do you know of an existing pattern for turning off events in the global theme that I could use as a model?

jcfilben:
> Ah, sorry! I never saw any kind of notification that there were responses after yours looking for input. I can definitely refactor to that.
>
> Do you know of an existing pattern for turning off events in the global theme that I could use as a model?

No worries! I can't think of an example off the top of my head where we control an event with a theme prop, but I think we could just create a boolean theme prop, `theme.rangeinput.wheel` and set it to true by default in the base theme (see `themes/base.js`). We would also need to add this prop to the typescript definitions in `themes/base.d.ts`. Then in the RangeInput file in `handleMouseOver` and `handleMouseOut` check if `theme.rangeinput.wheel` is true, if so `setScroll` otherwise do nothing.

Tbhesswebber:
> No worries! I can't think of an example off the top of my head where we control an event with a theme prop, but I think we could just create a boolean theme prop, `theme.rangeinput.wheel` and set it to true by default in the base theme (see `themes/base.js`). We would also need to add this prop to the typescript definitions in `themes/base.d.ts`. Then in the RangeInput file in `handleMouseOver` and `handleMouseOut` check if `theme.rangeinput.wheel` is true, if so `setScroll` otherwise do nothing.

Works for me - thanks for being so thorough! 

Since the default behavior is currently to have scrolling on, is it preferable to check that the property is not false rather than explicitly true so as to avoid a breaking change in behavior for anyone who depends on it and is not extending the base theme?

jcfilben:
> Since the default behavior is currently to have scrolling on, is it preferable to check that the property is not false rather than explicitly true so as to avoid a breaking change in behavior for anyone who depends on it and is not extending the base theme?

Yes! Good call

Tbhesswebber:
@jcfilben @ericsoderberghp - In putting together a partner PR for the docs repo, I realized that (almost?) everything else in a theme is specifically a property rather than a control.  Is `theme.<component>.<event_name>` an existing pattern in other components? If not, would it be worthwhile to follow a pattern like `theme.<component>.events.<event_name>`?

jcfilben:
> @jcfilben @ericsoderberghp - In putting together a partner PR for the docs repo, I realized that (almost?) everything else in a theme is specifically a property rather than a control. Is theme.<component>.<event_name> an existing pattern in other components? If not, would it be worthwhile to follow a pattern like theme.<component>.events.<event_name>?

This is a good note, we do have a couple places where we have event related props in the theme. One example is `carousel.animation`. I think I feel comfortable moving forward with `theme.rangeinput.wheel` since we do have a similar pattern elsewhere in the theme.

Tbhesswebber:
Great! Thanks, @jcfilben. This should be good for another round of review, then

src/js/components/RangeInput/__tests__/RangeInput-test.tsx
```diff
@@ -138,4 +140,35 @@ describe('RangeInput', () => {
...
+
+  test('onWheel overrides should not call onChange', () => {
+    const theme: ThemeType = { rangeInput: { wheel: false } };
+    const onChange = jest.fn();
+    const { getByDisplayValue } = render(
+      <Grommet>
+        <ThemeContext.Extend value={theme}>
+          <RangeInput min={0} max={10} step={1} value={5} onChange={onChange} />
+        </ThemeContext.Extend>
+      </Grommet>,
```
jcfilben, line 168:
```suggestion
    const theme: ThemeType = { rangeInput: { wheel: false } };
    const onChange = jest.fn();
    const { getByDisplayValue } = render(
      <Grommet theme={theme}>
          <RangeInput min={0} max={10} step={1} value={5} onChange={onChange} />
      </Grommet>,
```
Just a minor simplification so we no longer need to import ThemeContext

src/js/components/RangeInput/__tests__/RangeInput-test.tsx
```diff
@@ -7,6 +7,8 @@ import 'regenerator-runtime/runtime';
 
 import { Grommet } from '../../Grommet';
 import { RangeInput } from '..';
+import { ThemeContext } from '../../../contexts/ThemeContext';
```
jcfilben, line 10: I think we can remove this import now
```suggestion
```

jcfilben (approved):
This looks good to me!

jcfilben:
@ericsoderberghp just tagging to let you know this is ready for your eyes. Github won't let me re-tag you on this one for review

src/js/components/RangeInput/RangeInput.js
```diff
@@ -86,8 +95,9 @@ const RangeInput = forwardRef(
     // This is to make sure scrollbar doesn't move
     // when user changes RangeInput value.
     const handleMouseOver = () =>
-      setScroll({ x: window.scrollX, y: window.scrollY });
-    const handleMouseOut = () => setScroll({ x: null, y: null });
+      isScrollEnabled && setScroll({ x: window.scrollX, y: window.scrollY });
```
ericsoderberghp, line 98: It would be slightly more efficient to check before even assigning  the handler:
`onWheel={isScrollEnabled ? handleOnWheel : undefined}`

src/js/components/RangeInput/RangeInput.js
```diff
@@ -33,9 +36,12 @@ const RangeInput = forwardRef(
     },
     ref,
   ) => {
+    const themeContext = useContext(ThemeContext) || defaultProps.theme;
```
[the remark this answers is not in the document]
Tbhesswebber: Ah, okay.  I had assumed based on the naming on the next line that adding the `Context` suffix was the standard in Grommet.  Fixing now

src/js/components/RangeInput/RangeInput.js
Tbhesswebber: Should I also prevent adding the event listener to the window if I'm switching to go this route? I was trying to make a change with the smallest footprint possible since I'm not familiar with the codebase or patterns within

ericsoderberghp: yes

taysea: @Tbhesswebber just following up here -- are you willing to incorporate the changes suggested above?

kid-icarus:
Hi folks, I rebased this against master, addressed the last bits of PR feedback and re-opened the PR at https://github.com/grommet/grommet/pull/7130

@ericsoderberghp would you mind giving a review when you get a chance? This seems like the quickest way for me to work around https://github.com/grommet/grommet/issues/7129

jcfilben:
Superseded by https://github.com/grommet/grommet/pull/7130

PR #6033, opened 2022-03-30
Add ability to pass in a11yTitle in dropProps for Menu

### LINKED ISSUE #5964: Menu accessibility

amandadupell:
#### What does this PR do?

Adds an optional a11yTitle prop to DropProps to allow users to specify an a11yTitle that way

`Menu.js`

closes #5964 

No

Yes

Compatible


jcfilben (commented):
What if we have a case where Menu is given an `a11yTitle` and an `a11yTitle` is passed through drop props? Currently it looks like the dropProps `a11yTitle` will override the menu `a11yTitle`

jcfilben (approved):
Looks good!

src/js/components/Drop/index.d.ts
```diff
@@ -8,6 +8,7 @@ import {
 } from '../../utils';
 
 export interface DropProps {
+  a11yTitle?: string;
```
britt6612, line 11: can we use the a11yTitle type here ?

src/js/components/Menu/stories/typescript/Children.tsx
```diff
@@ -10,9 +10,12 @@ const MenuWithChildren = (props) => (
       { label: 'Launch', onClick: () => {} },
       { label: 'Abort', onClick: () => {} },
     ]}
+    dropProps={{
+      a11yTitle: 'Test',
```
britt6612, line 14: if we are going to keep this story can we add something else other than `Test` or remove

britt6612 (approved):

src/js/components/Menu/Menu.js
```diff
@@ -332,11 +330,7 @@ const Menu = forwardRef((props, ref) => {
               {alignControlMirror === 'top' && align.top === 'top'
                 ? controlMirror
                 : undefined}
-              <Box
-                overflow="auto"
-                role="menu"
-                a11yTitle={ariaLabel || a11yTitle}
-              >
+              <Box overflow="auto" role="menu" a11yTitle={dropA11y || a11y}>
```
ericsoderberghp, line 333: We are already passing `dropProps` to the DropButton at line 312. I think anything passed via `dropProps` should be handled by Drop itself. In this particular case, I think we should be changing DropContainer.js to set the `aria-label` on StyledDrop around line 331.

Issue #6032, opened 2022-03-30
Allow Drop to be placed inline in the DOM instead of in a React portal


ericsoderberghp:
A React portal places content outside of the current react root and is handy for Layer. However, sometimes it would be preferable for a Drop to be placed as a child of its controlling target, such as a Menu. This would improve the keyboard interaction behavior, as tabbing through the menu would proceed to the next component in the tab order after the Menu. It also would keep the DOM a bit cleaner and work better with more focused snapshot testing.

See https://github.com/grommet/grommet/issues/5896

PR #6034, opened 2022-03-30
Changed Drop to add container inline

ericsoderberghp:
#### What does this PR do?

Changed Drop to add `container="inline"`. This allows the Drop to be rendered within its parent as opposed to in a React portal.
See the issue for some detail around when this might be useful.

Inline.js story

Added a new Inline story and added a jest test.

Inline story

no changes to the unit test patterns already in the file

#6032 

yes

yes

backwards compatible

### DIFF
--- a/src/js/components/Drop/stories/Inline.js
+++ b/src/js/components/Drop/stories/Inline.js
@@ -1,0 +1,49 @@
+import React, { useEffect, useRef, useState } from 'react';
+
+import { Box, Drop } from 'grommet';
+
+const align = { top: 'bottom', left: 'left' };
+
+const InlineDrop = () => {
+  const targetRef = useRef();
+
+  // trigger re-render so we have the targetRef
+  const [, setShowDrop] = useState(false);
+  useEffect(() => {
+    setShowDrop(true);
+  }, []);
+
+  return (
+    // Uncomment <Grommet> lines when using outside of storybook
+    // <Grommet theme={...}>
+    <Box fill align="center" justify="center">
+      <Box
+        background="dark-3"
+        pad="medium"
+        align="center"
+        justify="start"
+        ref={targetRef}
+      >
+        Target
+        {targetRef.current && (
+          <Drop container="inline" align={align} target={targetRef.current}>
+            <Box pad="large">Drop Contents</Box>
+          </Drop>
+        )}
+      </Box>
+    </Box>
+    // </Grommet>
+  );
+};
+
+export const Inline = () => <InlineDrop />;
+Inline.parameters = {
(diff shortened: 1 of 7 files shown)


MikeKingdom (commented):
I do get an exception in story Drop/Multiple when I click on the "Layer" button:
image

MikeKingdom (approved):

jcfilben (approved):
Looks good to me

MikeKingdom (approved):
LGTM

src/js/components/Drop/stories/Inline.js
```diff
@@ -0,0 +1,49 @@
...
+        ref={targetRef}
+      >
+        Target
+        {targetRef.current && (
+          <Drop container="inline" align={align} target={targetRef.current}>
+            <Box pad="large">Drop Contents</Box>
+          </Drop>
+        )}
+      </Box>
+    </Box>
```
ShimiSun, line 34: Wondering if there is a reason for not using interactive elements in this story. This could help the user to examine how tabbing works in the case of an inline drop.

src/js/components/Drop/stories/Inline.js
```diff
@@ -0,0 +1,49 @@
...
+      <Box
+        background="dark-3"
+        pad="medium"
+        align="center"
+        justify="start"
+        ref={targetRef}
+      >
+        Target
+        {targetRef.current && (
+          <Drop container="inline" align={align} target={targetRef.current}>
```
ShimiSun, line 29: In this story example, the `container` is the prop and 'inline' is the value. Is this the expected usage? 
There are no details about a `container` prop. 
The prop `inline` should be a boolean according to the specified types.

src/js/components/Drop/stories/Inline.js
ShimiSun: I've filed a POC PR to demonstrate how tabbing would work with the suggested inline prop https://github.com/grommet/grommet/pull/6049

ShimiSun:
Looking deeper into the functionality and the purpose of allowing inline rendering of drop, I've applied `<Select dropProps: {{ inline: true}} ... />` to the Select on "All Components" story, and although the drop is placed 'inline' to the owning element on the DOM, the keyboard tabbing navigation still places the user at the end of the page rather than to the next interactive element.
Please let me know if I should use the prop in a different way or whether there should be more work to be done around the tabbing enablement. Here is a POC to play with https://github.com/grommet/grommet/pull/6049

IMO, I don't see a good functional reason to allow `inline` rendering of the Drop if it's not solving the keyboard skipping to the end of the page.

Issue #6021, opened 2022-03-24
Video accessibility


amandadupell:
### Expected Behavior

There are multiple aspects of the video component that appear to be broken or not accessible:

- [x] 1. User should be able to scrub along the timeline of the video to change the time position
- [x] 2. User plays or pauses the video via the extended "controls" prop. They are able to pause or play with any of the two controls. For example, the video will be playing and the user can go to the extended controls and hit the play icon. This will pause the video.

- [x] 3. The user cannot toggle captions on/off even if there is a designated `<track />` element with a captions file and "captions" is added to controls.

### Actual Behavior

There are multiple aspects of the video component that appear to be broken or not accessible:
1. User is unable to use their keyboard to scrub along the timeline of the video
2. There is a disconnect when controls include pause/play in the pop-up menu
3. Captions are not being supported even when including a `<track />` element with a captions/subtitles file (.vtt, .srt, etc.)

### URL, screen shot, or Codepen exhibiting the issue

[Codebox](https://codesandbox.io/s/laughing-franklin-9ge06h?file=/index.js)

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #6035, opened 2022-03-30
Enhanced Video to allow moving along video timeline with keyboard

jcfilben:
#### What does this PR do?
Adds the ability to move forward and backwards in a video with the left and right arrow keys. The default amount of time to move forward/backward in the video is 5 seconds but this can be customized with the `skipInterval` prop.

Tested with the existing video storybook stories and customizing the time with the `skipInterval` prop.

https://github.com/grommet/grommet/issues/6021

Yes, I will add a PR for the grommet doc changes

Yes

Compatible


src/js/components/Video/StyledVideo.js
```diff
@@ -11,18 +11,18 @@ const FIT_MAP = {
...
   max-width: 100%;
   z-index: 1;
-  ${props => props.fit && fitStyle} ::cue {
-    background: ${props => props.theme.video.captions.background};
+  ${(props) => props.fit && fitStyle} ::cue {
+    background: ${(props) => props.theme.video.captions.background};
   }
 
-  ${props => props.theme.video && props.theme.video.extend};
+  ${(props) => props.theme.video && props.theme.video.extend};
```
britt6612, line 25: are these changes intended these linting ?

jcfilben: Those were just changes from `yarn lint`

src/js/components/Video/Video.js
```diff
@@ -210,6 +212,16 @@ const Video = forwardRef(
       [duration, videoRef],
     );
 
+    const seekForward = useCallback(() => {
+      setInteracting(true);
+      videoRef.current.currentTime += skipInterval || 5;
```
britt6612, line 217: Is 5 just the standard number? Just want to know how you decided on 5

src/js/components/Video/StyledVideo.js
```diff
@@ -35,6 +35,14 @@ const StyledVideoContainer = styled.div`
   overflow: hidden;
   position: relative;
   ${genericStyles};
+  &:focus {
```
britt6612, line 38: I like the idea of the line for the focus just thinking is someone didn't want this or wanted a horizontal line

src/js/components/Video/Video.js
jcfilben: I saw other video players using 5 seconds but I couldn't find any info on if there is a standard for this

src/js/components/Video/StyledVideo.js
```diff
@@ -35,6 +35,14 @@ const StyledVideoContainer = styled.div`
...
+    outline-offset: 0px;
+    outline: ${(props) => {
+      const color = normalizeColor('focus', props.theme);
+      if (color) return `2px solid ${color};`;
```
MikeKingdom, line 42: Should this come from utils/styles.js focusStyles?

src/js/components/Video/Video.js
MikeKingdom: Netflix and I think YouTube tend to use 10s, at list on their skip forward and backward buttons.

halocline: Might be nice to have the default specified in the theme, so that callers can set their own defaults as opposed to each instance.

jcfilben: If we have the default set in the theme should we still have the `skipInterval` prop to override the default?

src/js/components/Video/StyledVideo.js
jcfilben: Yes, I will go back and change that

src/js/components/Video/Video.js
halocline: yes, I would keep the prop -- nice to have ability to override and not lock folks in

jcfilben: Sounds good, I will make these changes

src/js/themes/base.js
```diff
@@ -1682,6 +1682,7 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       // scrubber: { track: { color: undefined } },
       scrubber: {
         color: 'light-4',
+        interval: 10,
```
halocline, line 1685: Needed in base.d.ts as well?

halocline, line 1685: Commented out Line 1682 looks suspect. Can we address the `scrubber.track.color` properties? Consolidate under the active scrubber property + get 100% alignment here and  `base.d.js`.

src/js/components/Video/stories/typescript/ControlsBelow.tsx
```diff
@@ -4,7 +4,7 @@ import { Box, Video } from 'grommet';
 
 const SimpleVideo = (props) => (
   <Box align="center" pad="large">
-    <Video {...props}>
+    <Video skipInterval={2} {...props}>
```
halocline, line 7: I am having troubles accessing the player controls after I play the video. They are available on first render, but if I play, they get hidden and I am no longer able to access them with the keyboard. Am I doing something wrong?

britt6612 (approved):
Looks good

src/js/components/Video/propTypes.js
```diff
@@ -36,6 +36,7 @@ if (process.env.NODE_ENV !== 'production') {
       volumeDown: PropTypes.string,
       volumeUp: PropTypes.string,
     }),
+    skipInterval: PropTypes.number,
```
halocline, line 39: While we are in here, can we make sure that the propTypes and index.d.ts definitions are at parity and easy to compare?
- first level of properties, can we alphabetize?
- messages properties, can we alphabetize?
- `controls.items` is missing enum prop type

I think that's all.

src/js/components/Video/Video.js
```diff
@@ -455,53 +469,56 @@ const Video = forwardRef(
...
-            setPercentagePlayed((video.currentTime / video.duration) * 100);
-            if (onTimeUpdate) onTimeUpdate(event);
-          }}
-          onVolumeChange={(event) => {
-            setVolume(videoRef.current.volume);
-            if (onVolumeChange) onVolumeChange(event);
-          }}
-          autoPlay={autoPlay || false}
-          loop={loop || false}
+      <Keyboard onRight={seekForward} onLeft={seekBackward}>
```
halocline, line 472: I noted an issue regarding the lack of ability to access the controls after video play. Would it be appropriate to add an `onTab` here with a callback that sets `setInteracting(true)`?

halocline (commented):
Looking good. Noted a few items for tidying up.

src/js/components/Video/Video.js
jcfilben: Yes I think that is a good idea. I will add that

src/js/components/Video/stories/typescript/ControlsBelow.tsx
jcfilben: resolve in latest commit

halocline (approved):
Thanks for the cleanups and adding the focus / interacting events.

Outside of this PR's scope, the controls' focus gets cutoff -- do we have an issue for that?

jcfilben:
> Outside of this PR's scope, the controls' focus gets cutoff -- do we have an issue for that?

Screen Shot 2022-04-05 at 9 29 56 AM

Just want to confirm that you are talking about how you can't see the focus indicator when interacting with keyboard on the controls menu. If so then no we do not have an issue for this but I can create one

jcfilben:
Issue created: https://github.com/grommet/grommet/issues/6054

ericsoderberghp (approved):

PR #6037, opened 2022-03-30
add disabled to controls in which the video is already playing or paused

britt6612:
#### What does this PR do?
This PR disables `play` button in the control menu when the video is already playing as well as disables the pause button in control menu when the Video is already paused
video.js
storybook 
add `prop` to Video in storybook 

```
controls={{ position: "below", items: ["pause", "play"] }}

```

User plays or pauses the video via the extended "controls" prop. They are able to pause or play with any of the two controls. For example, the video will be playing and the user can go to the extended controls and hit the play icon. This will pause the video.
#6021 
no
maybe
backwards compatible

### DIFF
--- a/src/js/components/Video/Video.js
+++ b/src/js/components/Video/Video.js
@@ -313,6 +313,7 @@ const Video = forwardRef(
             id: 'video.pauseButton',
             messages,
           }),
+          disabled: !playing,
           onClick: playing ? pause : play,
         },
         play: {
@@ -321,6 +322,7 @@ const Video = forwardRef(
             id: 'video.playButton',
             messages,
           }),
+          disabled: playing,
           onClick: playing ? pause : play,
         },
       };
--- a/src/js/components/Video/Video.js
+++ b/src/js/components/Video/Video.js
@@ -314,7 +314,7 @@ const Video = forwardRef(
             messages,
           }),
           disabled: !playing,
-          onClick: playing ? pause : play,
+          onClick: pause,
         },
         play: {
           icon: <Icons.Play color={iconColor} />,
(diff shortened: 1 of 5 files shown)


britt6612:
Need to look at TS issue

jcfilben (commented):
I think we need to add disabled to the typescript definition for items. Right now it is just has icon, a11yTitle, and onClick
```
items: PropTypes.arrayOf(
  PropTypes.oneOfType([
    PropTypes.string,
    PropTypes.shape({
      icon: PropTypes.element,
      a11yTitle: PropTypes.string,
      onClick: PropTypes.func,
    }),
  ]),
),
```

src/js/components/Video/stories/typescript/TheaterMode.tsx
```diff
@@ -25,9 +25,9 @@ const TheaterMode = (props) => {
               a11yTitle: 'Toggle theater mode',
             },
             'fullScreen',
+            'pause',
```
jcfilben, line 28:
```suggestion
```
I don't think we need to include the pause button in this example. Or if we do include it we should also have the play button.

britt6612: aw yes no this was me testing TS

jcfilben (approved):
Looks good

src/js/components/Video/Video.js
```diff
@@ -313,6 +313,7 @@ const Video = forwardRef(
             id: 'video.pauseButton',
             messages,
           }),
+          disabled: !playing,
```
ericsoderberghp, line 316: This seems to conflict with the code in the next line that checks for `playing` and calls `play` if not.

PR #5823, opened 2021-11-30
Fix: Anchor invalid "as" prop

jlaluces123:
#### What does this PR do?
This PR extends Anchor's `as` type-checking to include `PropType.elementType` to solve `invalid prop 'as' supplied to Anchor` console error.

#### Where should the reviewer start?
`Anchor/propTypes.js`

#### What testing has been done on this PR?
Manually tested

#### How should this be manually tested?
- Adding `React-Router-Dom` package to project
- Setting up an Anchor to use the Link component from react-router-dom in any story ( `as={Link}` )
- View story's console and ensured no console errors were shown

#### Do Jest tests follow these best practices?

- [X] `screen` is used for querying.
- [X] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [X] `userEvent` is used in place of `fireEvent`.
- [X] `asFragment()` is used for snapshot testing.

#### What are the relevant issues?
Resolves #5803 

#### Do the grommet docs need to be updated?
Possibly, to add an example using a component

#### Should this PR be mentioned in the release notes?
Yes
#### Is this change backwards compatible or is it a breaking change?
Backwards compatible

PR #6038, opened 2022-03-30
Make "as" prop on Box accept any React element

dherault:
#### What does this PR do?
Very similarly to https://github.com/grommet/grommet/pull/5823, this PR enhances Box to accept any React element for the `as` prop.
When working with legacy code or `styled-components`, I stumbled upon the need for this extension.
Before this PR, a warning like this would appear, but the code would still work like a charm.
image

At the tests

Extensive testing 🧐

Pass a React element such as an old school component as the `as` prop to a Box

They update automatically, right?
Yes.
This PR does not change Grommet's behavior, but simply removes a useless warning on a specific edge case.

### DIFF
--- a/src/js/components/Box/__tests__/Box-test.tsx
+++ b/src/js/components/Box/__tests__/Box-test.tsx
@@ -486,10 +486,35 @@ describe('Box', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
-  test('as', () => {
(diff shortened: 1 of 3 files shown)


jcfilben (approved):
Looks good, thanks for including tests!

ericsoderberghp (approved):

Issue #5896, opened 2022-01-15
Tabbing beyond the content of a Drop jumps the tab focus to the top of the page


phinch:
### Expected Behavior 

It's not clear to me what the globally expected behavior should be - but I would generally expect that if I have a DropButton in a page, then tabbing when the drop is open should allow me to tab through content within the drop, and when there is no more content to tab through, tab should either close the drop, or do nothing.

### Actual Behavior

Tabbing beyond the content of a drop seems to cause the tab order to return to the top of the page, rather than doing either of those things. This can have negative impacts for users who are trying to navigate a page using a keyboard - in my team's use case, we have a Select input inside of a form (which uses DropButton under the hood). When the user is using a keyboard to navigate and fill out the form, which is a use case we are required to cover to meet accessibility standards, when the user tabs within the Select options list, they are taken completely out of the form and it is very hard for them to continue navigating via keyboard.

### URL, screen shot, or Codepen exhibiting the issue

I put together some simple examples here - the issue can also be reproduced on the Grommet website examples for DropButton and Select.
https://codesandbox.io/s/grommet-v2-template-forked-j4c80?file=/index.js

### Steps to Reproduce

1. Navigate to the Select input using tab and open the menu by pressing Enter.
2. Press Tab while the menu is open
3. Observe that Tab does not move focus to the next item on the page; additionally observe that other keyboard interactions, like Escape to close the menu, no longer work

### Your Environment

- Grommet version: 2.19.0

eatyourgreens:
We've run into this same problem. We're setting `dropProps` to `trapFocus: false` on the Select, so that you can tab out of it while it is open. Unfortunately, we then find that tabbing out of the open menu leaves the menu open and sends focus to the first focusable element on the page:
https://github.com/zooniverse/front-end-monorepo/pull/2838#pullrequestreview-883406277

We've looked at tab behaviour for HTML `<select>` elements in different browsers:
- Chrome and Safari (and presumably Edge): An open menu traps focus and Tab is ignored. I think this it what Grommet's `Select` is emulating with `trapFocus: true`.
- Firefox: Tabbing out of an open menu selects the focussed option and sends keyboard focus to the next focusable element in the page (usually the next form input.) We'd really like to be able to do this with the Grommet `Select` too.

phinch:
[...]

PR #6049, opened 2022-04-04
Updated All Components Select to use inline drop POC

ShimiSun:
#### What does this PR do?
This PR is a POC for https://github.com/grommet/grommet/pull/6034

Looking deeper into the functionality of allowing Drop to be rendered inline to its parent component on the DOM, I've applied `<Select dropProps: {{ inline: true}} ... />` to the Select on "All Components" story, and although the drop is placed 'inline' to the owning element on the DOM, the keyboard navigation issue is still not resolved, and the tabbing takes the user to the end of the page rather than to the next interactive element.

I think this PR is valid and the All Components Select story should use an inline drop, however, the inline prop should solve the keyboard navigation issue which currently is still an issue.

IMO, I don't see a good functional reason to allow inline rendering of the Drop if it's not solving the keyboard tabbing that is skipping to the end of the page.

The story
storybook
It's a POC

POC for the Drop inline functionality

#6032 
#5896 
Nope
Nope
backwards compatible

### DIFF
--- a/src/js/all/stories/AllComponents.js
+++ b/src/js/all/stories/AllComponents.js
@@ -123,6 +123,7 @@ const Components = () => {
     </Box>,
     <Box key="input" gap="small">
       <Select
+        dropProps={{ inline: true }}
         placeholder="Select"
         options={['One', 'Two']}
         onChange={() => {}}


ericsoderberghp:
I think we need to investigate why entering Tab when the drop is open doesn't do what is intended. I suspect it is because `trapFocus` is true, which isn't what we want for the scenario we are exploring.

jcfilben:
Proof of concept no longer needed

Issue #5677, opened 2021-10-08
Accessibility Calendar. Click arrow to select day


denant0:
Using Calendar, Low contrast grey button against the white background is still the only visual means of tracking keyboard focus inside the calendar.  What can be done so that when you press the keys, the text is read?

### Expected Behavior

![изображение](https://user-images.githubusercontent.com/1597545/136600618-1631c46b-6b7d-4b4c-8cbf-d08d8d50a239.png)

jcfilben:
When using a screenreader the dates are correctly read out when navigating, this was fixed in https://github.com/grommet/grommet/pull/5319. 

I agree that the grey button against the white background isn't enough of a visual indicator for tracking keyboard focus. I think we should have the focus indicator on the date when navigating with the keyboard rather than having the focus indicator on the whole calendar.

Ojas-Sri:
Hi,

I have made some progress in addressing this issue. Below are 2 screenshots for the same , please suggest if any changes are needed:

(The calendar will be in focus when mouse clicks on the calendar area. When the keyboard navigates to other days, the focus moves from calendar to the particular day. If the user clicks again via mouse pointer , focus resets to calendar)

![PR (11)](https://user-images.githubusercontent.com/80472243/161683492-f7c76be9-963a-423d-a03d-d1b2077a9ff2.gif)

![PR (12)](https://user-images.githubusercontent.com/80472243/161683613-cd970580-366c-447d-8788-8c7b83eb1d47.gif)

jcfilben:
Hi @Ojas-Sri nice work on this and thanks for sharing what you are thinking
This behavior looks good to me. I just had a few questions
- If a user is navigating by keyboard and tabs to the calendar will the focusIndicator be around the whole calendar or on the first day of the month? I'm thinking it should be on the first day of the month in this case
- When the focusIndicator is on a specific date it looks like the calendar still has an outline around it. I think if we move the focus indicator to a date the calendar shouldn't have an outline

Ojas-Sri:
Hi @jcfilben,

Thanks for the comments, I have addressed both of them. Screenshot of my implementation attached below.

![PR (2)](https://user-images.githubusercontent.com/80472243/162043958-a5b0c317-5adf-4d2f-a201-7999ac816f07.gif)


![PR (3)](https://user-images.githubusercontent.com/80472243/162044138-8cedee4a-1b73-479f-8c58-3382d39dc13a.gif)


[...]

PR #6057, opened 2022-04-08
Fix for #5677 : Improved accessibility of Calendar 

Ojas-Sri:
Added functionality for the focus to be on the day, when a user navigates using a keyboard.

#### What does this PR do?

**In code**
- This PR introduces `focus` prop to the `StyledDayContainer`.

 **For the User**
- If the user navigates to the `Calendar` component using `tab` key, the first day of the month will be in focus rather than the entire `Calendar`
- If the user decides navigates the calendar using the arrow key, the focus will shift to the active day.
- If the user decides to return to mouse pointer, the entire `Calendar` would come in focus on Mouse Down ( & selection)

- The reviewer could start from file `src/js/components/Calendar/StyledCalendar.js` , where I have added the required props.
- Further, the reviewer can refer file `src/js/components/Calendar/Calendar.js` at line 1002, 1004 and thereon.

- I have tested it using the storybook. Screenshots of the functionality below : 

- The tester could use tab keys to reach the `Calendar` Component, once in focus, the arrow keys could be used to navigate.

#5677 

PR (2)
PR (3)

- Don't think so.

- Not sure

- It should be.

### DIFF
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -173,12 +173,18 @@ const CalendarDay = ({
   children,
   fill,
   size,
+  focus,
   isInRange,
   isSelected,
   otherMonth,
   buttonProps = {},
 }) => (
-  <StyledDayContainer role="gridcell" sizeProp={size} fillContainer={fill}>
+  <StyledDayContainer
+    role="gridcell"
+    sizeProp={size}
+    fillContainer={fill}
+    focus={focus}
+  >
     <CalendarDayButton fill={fill} {...buttonProps}>
       <StyledDay
         disabledProp={buttonProps.disabled}
@@ -832,6 +838,7 @@ const Calendar = forwardRef(
               }}
               isInRange={inRange}
               isSelected={selected}
+              focus={!focus && active?.getTime() === day.getTime()}
               otherMonth={day.getMonth() !== reference.getMonth()}
               size={size}
               fill={fill}
@@ -944,6 +951,7 @@ const Calendar = forwardRef(
             onUp={(event) => {
               event.preventDefault();
               event.stopPropagation(); // so the page doesn't scroll
+              setFocus(false);
               setActive(addDays(active, -7));
               if (!betweenDates(addDays(active, -7), displayBounds)) {
                 changeReference(addDays(active, -7));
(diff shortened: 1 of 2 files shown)


Ojas-Sri:
`yarn test` is failing for the `Calendar` Component, as the snapshots are not matching. Have provided a snip below for reference : 


image



Please advise.

jcfilben:
Thanks for the detailed description and screenshots!
I'm thinking when the user is navigating with a mouse and hovering over dates we don't need to show the focus indicator on the date. We only need to show the focus indicator on the date if the user is navigating by keyboard.

The snapshot changes seem fine, looks like the class name is changing which is fine since we are changing the focus indicator behavior which affects the css styling. You can run `yarn test -u` to update the snapshots and push them up and then we can make sure they all look good.

Ojas-Sri:
> `yarn test` is failing for the `Calendar` Component, as the snapshots are not matching. Have provided a snip below for reference :
> 
> image
> 
> Please advise.

Updated snapshots: 
image

Ojas-Sri:
> Thanks for the detailed description and screenshots! I'm thinking when the user is navigating with a mouse and hovering over dates we don't need to show the focus indicator on the date. We only need to show the focus indicator on the date if the user is navigating by keyboard.
> 
> The snapshot changes seem fine, looks like the class name is changing which is fine since we are changing the focus indicator behavior which affects the css styling. You can run `yarn test -u` to update the snapshots and push them up and then we can make sure they all look good.


Hi @jcfilben ,

If the user is navigating using a mouse then the dates are getting a hover Indicator , while the focus stays on the calendar box.

However, there is an edge case too. Say if the user is navigating using the keyboard and suddenly switches to using a mouse , then the first pressed day (with mouse down) would be in focus momentarily but the focus would return to the calendar box ultimately.

Did I answer your question?

jcfilben:
When I am navigating with the mouse the date I am hovering on is getting both the hover styling (grey background) and the focus indicator (green outline). I think if we are navigating with the mouse we only need to show the hover styling and not show the focus indicator.
Screen Shot 2022-04-13 at 4 44 17 PM

jcfilben (approved):
Looks good!

Ojas-Sri:
> Looks good!

Hi @jcfilben , there is a bug I noticed in my code. When the user selects outside the calendar ,then comes back to the calendar the day appear in focus.

Requesting some time to fix this bug. My apologies for the inconvenience caused.

Regards
Ojas

jcfilben:
> Hi @jcfilben , there is a bug I noticed in my code. When the user selects outside the calendar ,then comes back to the calendar the day appear in focus.
>
> Requesting some time to fix this bug, will commit to this PR with the fix. My apologies for the inconvenience caused.
>
> Regards
> Ojas

Good catch, I appreciate your attention to detail! And no inconvenience caused :)
I'm going to mark this PR as a draft and then you can mark it 'ready for review' once the bug is fixed and the PR is ready for us to look at again.

PR #6058, opened 2022-04-10
Include Grid props in NameValueList props type

conartist6:
#### What does this PR do? 
It broadens the type for NameValueList's props to include Grid's props. This makes the types match what the code actually does.

I have verified that the the fix works by editing the source in `node_modules`

See above

Here is where the rest of the props to `NameValueList` get passed to `Grid`
https://github.com/grommet/grommet/blob/d34c1e8826b7632aefa488bf5409472603875242/src/js/components/NameValueList/NameValueList.js#L62

None

Yes: https://github.com/grommet/grommet-site/pull/370

No

Backwards compatible

### DIFF
--- a/src/js/components/NameValueList/index.d.ts
+++ b/src/js/components/NameValueList/index.d.ts
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { AlignType, WidthType } from '../../utils';
+import { GridProps } from '../Grid'
 export interface NameValueListProps {
   align?: AlignType;
   layout?: 'column' | 'grid';
@@ -16,9 +17,9 @@ export interface NameValueListProps {
   };
 }
 export interface NameValueListExtendedProps
-  extends NameValueListProps,
+  extends NameValueListProps, GridProps,
     Omit<JSX.IntrinsicElements['dl'], keyof NameValueListProps> {}
 
-declare const NameValueList: React.FC<NameValueListProps>;
+declare const NameValueList: React.FC<NameValueListExtendedProps>;
 
 export { NameValueList };


jcfilben:
I am supportive of adding GridProps to the NameValueList prop types. Can you share the use case and what grid props you are using on NameValueList? It would be helpful to know which grid props are useful on NameValueList as this may affect how we want to document them.

jcfilben:
I am supportive of adding GridProps to the NameValueList prop types. Can you share the use case and what grid props you are using on NameValueList? It would be helpful to know which grid props are useful on NameValueList as this may affect how we want to document them.

britt6612:
> I am supportive of adding GridProps to the NameValueList prop types. Can you share the use case and what grid props you are using on NameValueList? It would be helpful to know which grid props are useful on NameValueList as this may affect how we want to document them.

You can pass any `grid` props on the NameValueList since we have the `...rest` on that Grid so we should include the types for the typescript.

conartist6:
It's `columns` that I need. I use `columns={['auto', 'auto']}` to ensure that the entire `NameValueList` element takes up no more space than necessary in the layout.

Here's what my usage looks like:

Screen Shot 2022-04-11 at 5 46 01 PM

And you can see the whole app at https://regex-playground-rho.vercel.app/

conartist6:
I believe the `NameValueList` `propTypes` would still need to be updated as well.

jcfilben:
Thanks for sharing the use case. Also I misspoke when I said prop types. I think we are fine just updating the typescript types. This is how components like Accordion handle it. Accordion can be passed Box props which are included in the typescript types but not explicitly listed in the prop types file.

jcfilben (approved):

ericsoderberghp (approved):

PR #6063, opened 2022-04-12
Changed styling for sortable datatable

### LINKED ISSUE #5965: DataTable sorting looks bad for end-aligned column or columns with units
A sortable end-aligned column leaves a gap (for the sort icon, which is normally _invisible_) and causes the label to not align with the row content.  It would work a lot better if the sort controls were always _visible_ instead of only appearing when hovered.  Even a light gray control that turns black upon hover would be better and would help to explain the awkward alignment.

A sortable column where units are employed places the sort indicator (the gap) between the column header and the units.  It should place this after the units.

When you have both end-alignment and units (which is common for numeric fields) it looks particularly bad.

Grommet Designer example: https://designer.grommet.io/?id=DataTable-Sorting-edgar-circenis-hpe-com

image

amandadupell:
#### What does this PR do?

Changes the styling for a sortable DataTable with units, as requested in this issue #5965. Moved the icon for sorting to be after the units and added align for when columns wrap, didn't change hover/non-hover behavior.

DataTable.js

Wrote story to show new styling

Storybook

closes #5965 

Icon Placement
Screen Shot 2022-04-12 at 11 53 34 AM

Wrapping
Screen Shot 2022-04-12 at 12 01 27 PM

No

Yes?

Compatible


taysea:
Are we planing on discussing the icon visibility that is noted in the original issue? I do agree with the original issue that end aligned column headers always look mis-aligned when there is the invisible space for the sort icon. Any changes for this would need to be opt-in for backwards compatibility but I think it's worth a discussion.

jcfilben (commented):
In this example can we change the columns with numbers (count, percent used, consumption) to be right aligned? Note sure if the design system site explicitly has guidance on this, but it seems to be the pattern we are following in other examples.
Screen Shot 2022-04-13 at 5 14 53 PM

Also not sure if it would make more sense to have the sort icon on the left if a column is right aligned or if the icon should always be on the right regardless of if the column is right or left aligned.

amandadupell:
if we wanted to change the styling around the icon being "invisible", we could make this change in the hpe theme [here](https://github.com/grommet/grommet-theme-hpe/blob/31e52ccc1aa211be75e63eed8d73d1f03580a42a/src/js/themes/hpe.js#L519)

src/js/themes/base.js
```diff
@@ -776,6 +776,7 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         units: {
           color: 'text-xweak',
           margin: { left: 'xsmall' },
+          alignSelf: 'end',
```
taysea, line 779: Just confirming I follow we this is here -- does this cover what was previously being handled by `align="baseline"`?

amandadupell: the `align="baseline"` was not impacting the styling of the units but i believe it was trying to achieve the results accomplished through aligning the units at the end of the container so when wrapped, the units are end aligned

jcfilben (commented):
I am good with these changes. I think for now we can leave the hpe theme as is or create a separate issue for it since it might need some design decisions and could potentially be a breaking change in the hpe theme.

ericsoderberghp (approved):

Issue #6062, opened 2022-04-12
Breaking change from `@types/react@18`


kristofferaas:
`@types/react@18` removed implicit children prop from type `React.FC`, from now on `children` should now be explicitly defined.
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210

From a quick exploration; extending `BoxExtendedProps` instead of `BoxProps` on affected components should resolve the issue :)

### Expected behaviour

Allowed to pass children props when using React 18 and TypeScript.

### Current behaviour

TypeScript complains that `Property 'children' does not exist on type X`.

https://codesandbox.io/s/fragrant-waterfall-6921n6?file=/src/App.tsx

### Steps to Reproduce

Create a new project with React 18, `@types/react@18` and TypeScript.

### Your Environment

- Grommet version: 2.22.0
- Browser Name and version:
- Operating System and version (desktop or mobile):

jcfilben:
Related to https://github.com/grommet/grommet/issues/5947

Issue #5947, opened 2022-02-15
React 18


jcfilben:
As of Nov. 15 2021 React 18 is in beta. The official v18 release is expected soon. The react team has stated, "Since concurrency in React 18 is opt-in, there are no significant out-of-the-box breaking changes to component behavior. You can upgrade to React 18 with minimal or no changes to your application code". For more information on React 18 see this article: https://reactjs.org/blog/2021/06/08/the-plan-for-react-18.html.

Currently Grommet is compatible with React ^16.6.1 and ^17.0.0.  Grommet is planning to adopt React 18 but not use any of the opt-in features that would cause breaking changes. This will allow us to continue supporting React ^16.6.1 and ^17.0.0 in addition to React 18.0.0.

Action Items:

- [ ] Test upgrading to the React 18 beta with Grommet. Confirm that storybook and unit tests are all working correctly.
- [ ] Upgrade to React 18 when it is officially released.
- [ ] Upgrade @testing-library/react from v12 to v13

jcfilben:
Update: React 18 has officially been released https://reactjs.org/blog/2022/03/29/react-v18.html

PR #6066, opened 2022-04-13
React 18

jcfilben:
#### What does this PR do?
Upgrades to React v18

- Anyone consuming Grommet should be able to use React v16, v17, or v18. The breaking changes in react 18 are within our testing files.

Important changes in React 18:
**ReactDOM.render is no longer supported**
You are still able to use ReactDOM.render but you will get a warning that it is deprecated. `createRoot` should be used instead of render. Render was being used in the following places in grommet:
- In the unit tests we are using render from `@testing-library/react`. As of v13.0.0 of `@testing-library/react` when you use render from testing-library it is actually using createRoot under the hood.
- Storybook is loading stories with render. There is a pre-release branch of storybook that supports `createRoot` but it looks like they still need to iron out some issues so we are sticking with the official release version of storybook for now. This means that when you run storybook everything should work as before but there will be warnings in the console on each story. **We will need to make a note of this in the release notes if it is not fixed before then**.
- RoutedAnchor and RoutedButton use `render`. This means if these components are used in a React 18 application they will generate a warning but still function. Since these components are deprecated I felt this was fine.

**Updates to TypeScript definitions**
The most notable change is that the `children` prop now needs to be explicitly defined. A list of the typescript changes can be seen here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210. I have updated the relevant type definitions to include children. It is possible that I may have missed updating some of the types so something we will want to keep an eye on as we move closer to the release this month.

Other Notes:
- I removed the `userAgent` test from Grommet-test.js. With the testing library update this test's document.body's client width was always returning 0 so the ResponsiveContext would always receive a size of small. I ensured that the `userAgent` code in Grommet was working as expected and it was. I played around with the test but in the end felt that there wasn't a good way to write the test that was effectively testing the `userAgent`. I think if we want to have this kind of test it is better suited to an end-to-end type of testing environment instead of testing with jsdom (which acknowledges that it shouldn't be used for this kind of testing).

Helpful links:
How to upgrade to react 18 - https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html
React's changelog - https://github.com/facebook/react/blob/main/CHANGELOG.md

`yarn test`
`yarn storybook`
`yarn build`

Closes https://github.com/grommet/grommet/issues/5947
Closes https://github.com/grommet/grommet/issues/6062

Yes
Yes
Backwards compatible


src/js/components/Footer/index.d.ts
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
-import { BoxProps } from '../Box/index';
```
britt6612, line 2: I believe we have the same in `Header` in the `index.d.ts` can we update that ?

src/js/components/Grommet/__tests__/Grommet-test.js
```diff
@@ -15,35 +14,6 @@ const TestAnnouncer = ({ announce }) => {
...
-        value: 1000,
-      },
-    },
-  },
-};
-
-const SSRTester = ({ ua }) => (
-  <Grommet theme={customBreakpointsTheme} userAgent={ua}>
-    <ResponsiveContext.Consumer>
-      {(size) => <Heading>{`Received size ${size} for ${ua}`}</Heading>}
```
britt6612, line 42: should we add this issue of testing to our end-to-end issue we have on the DS backlog? Just so we know when we start that we can account for this.

src/js/components/NameValueList/index.d.ts
```diff
@@ -3,6 +3,7 @@ import { AlignType, WidthType } from '../../utils';
 import { GridProps } from '../Grid';
 export interface NameValueListProps {
   align?: AlignType;
+  children?: React.ReactNode;
```
britt6612, line 6: more of a question in some of the files you added the ` children?: React.ReactNode;`
as well as adding `children` to the omit` Omit<ButtonType, 'icon' | 'size' | 'children'> {}`  just trying to understand why this is included in certain files?

jcfilben: I am omitting the `children` prop and then redefining it in certain cases if the `children` prop that was given through the intrinsic element or interface has a different type definition from what we need.

src/js/components/Grommet/__tests__/Grommet-test.js
jcfilben: Yeah I will add this to the end-to-end testing issue in our backlog

src/js/components/Footer/index.d.ts
jcfilben: Yeah, Header is using `BoxTypes` which includes the `children` prop but to keep Header consistent with the other files I updated it to use `BoxExtendedProps`

src/js/components/NameValueList/index.d.ts
britt6612: thank you for clarifying

src/js/components/Footer/index.d.ts
britt6612: yeah that's what I meant to add in the `BoxExtendedProps` for Header thank you 😄

amandadupell (approved):
changes make sense based on looking into react 18 upgrades and explanations, tests look good!

ericsoderberghp (approved):

Issue #6067, opened 2022-04-14
Is there a way to disable check boxes of grommet data table


mahanthi-arun-kumar:
On checking one of check boxes grommet data table , I need to disabled the other check boxes !

### Expected Behavior

On checking one of check boxes grommet data table , I need to disabled the other check boxes !

### Actual Behavior

Currently there seems like no prop to disable  check boxes of grommet data table

### URL, screen shot, 
<img width="1280" alt="Screenshot 2022-04-13 at 3 50 12 PM" src="https://user-images.githubusercontent.com/48381857/163324298-266a45f3-fcfa-4893-89b3-06816ff796f9.png">
or Codepen exhibiting the issue

### Steps to Reproduce

1. Go to grommet data table having check boxes
2. Try to disable check boxes by prop  

### Your Environment

- Grommet version:2.21.0
- Browser Name and version:100.0.4896.75 (Official Build) (x86_64)
- Operating System and version (desktop or mobile): desktop

PR #6070, opened 2022-04-14
Changed DataTable to add disabled

ericsoderberghp:
#### What does this PR do?

Changed DataTable to add `disabled` which takes an array similar to `select` to specify which rows should be disabled from interaction. The two interactions that can be disabled are selection, when using `onSelect`, and row clicking, when using `onClickRow`.

index.d.ts

Added unit test.
Enhanced Clickable and On Select stories to have a disabled row ("Matt").

stories mentioned above

new test is consistent with existing tests

#6067 

yes

yes

backwards compatible


src/js/components/DataTable/Body.js
```diff
@@ -182,7 +184,9 @@ const Body = forwardRef(
               }
             : undefined
         }
+        // TODO: skip disabled
```
jcfilben, line 187: Should we create an issue to change the arrow key behavior for disabled options to skip them?

jcfilben (approved):
Looks good, we will need to update the docs on grommet-site to include the `disabled` prop

jcfilben (commented):
When you are using a screenreader there isn't any indication that a row is disabled. When you try to interact with a disabled row/checkbox it make a noise that indicates you were able to interact with the row/checkbox, but nothing will happen.

ericsoderberghp:
> When you are using a screenreader there isn't any indication that a row is disabled. When you try to interact with a disabled row/checkbox it make a noise that indicates you were able to interact with the row/checkbox, but nothing will happen.

Any suggestions on what should happen?

jcfilben:
When you can click on a row I think we need to add `aria-disabled` to StyledDataTableRow for the rows that are disabled. When there is a disabled checkbox we should add `aria-disabled` to the Cell that contains the checkbox. Ideally the screen reader should read 'disabled' or 'dimmed' and then the name of the cell and then when you attempt to interact with it it shouldn't make the selected noise.

PR #6072, opened 2022-04-15
Fixed captions in Video component

jcfilben:
#### What does this PR do?
This PR fixes the video captions so they will display correctly when the caption button is clicked. The user can also specify multiple caption tracks for different languages and pass in a label to the track element to name different caption tracks for the video.

Added captions to the control below story. Also tested by adding multiple captions tracks to one video and adding labels to the track element.

^see above

https://github.com/grommet/grommet/issues/6021
No
Yes
Backwards compatible


src/js/components/Video/Video.js
```diff
@@ -161,9 +161,10 @@ const Video = forwardRef(
         const { textTracks } = video;
         if (textTracks.length > 0) {
           if (textTracks.length === 1) {
-            const active = textTracks[0].mode === 'showing';
+            const track = textTracks[0];
+            const active = track.mode === 'showing';
             if (!captions || !captions[0] || captions[0].active !== active) {
-              setCaptions([{ active }]);
+              setCaptions([{ label: track.label, active }]);
```
britt6612, line 167: maybe add a comment as to even if there is one `track` we still want to grab that  `label` if there  is one

britt6612 (approved):
Besides adding a comment looks good!  
Just took us a while to understand logic so adding a comment or  2 can help in the future!

src/js/components/Video/Video.js
jcfilben: Added a comment here as well as adding comments a few other places in the code.

src/js/components/Video/Video.js
```diff
@@ -278,13 +285,21 @@ const Video = forwardRef(
...
           <Icons.ClosedCaption color={iconColor} />
         ),
         label: caption.label,
         active: caption.active,
-        onClick: () => showCaptions(caption.active ? -1 : 0),
+        a11yTitle: caption.label || 'video.captions',
+        onClick: () => {
+          showCaptions(caption.active ? -1 : index);
+          // set the currently selected captions to active
+          captions[index].active = !captions[index].active;
```
taysea, line 298: Is there a reason we aren't using the `setCaptions` function here?

jcfilben: Updated this section to use `setCaptions` in latest commit

ericsoderberghp (approved):

PR #6074, opened 2022-04-17
checkbox toggle bugfix

### LINKED ISSUE #6073: RTL issue with CheckBox toggle variant
### Expected Behavior

Toggle should animate to left int `rtl` mode.

### Actual Behavior

Toggle animates to right and gets out of bounding frame.

Here is a link to [sandbox](https://codesandbox.io/s/intelligent-clarke-3n2q45?file=/src/App.js).

### Your Environment

- Grommet version: latest
- Browser Name and version: Chrome 99
- Operating System and version (desktop or mobile): Ubuntu Desktop

B-Esmaili:
#### What does this PR do?

Fixes #6073.

`yarn test` & storybook manual test

- [ x] `screen` is used for querying.
- [ x] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [ x] `userEvent` is used in place of `fireEvent`.
- [ x] `asFragment()` is used for snapshot testing.

#6073
No
Maybe


jcfilben (commented):
Nice work, only thing is that it looks like the margin on the toggle is placed on the wrong side when we use `dir='rtl'`. In CheckBox.js can we change line 167 to `const side = (!reverse != !theme.dir) ? 'left' : 'right';` so that the margin is placed on the correct side when we use RTL

B-Esmaili:
> const side = (!reverse != !theme.dir) ? 'left' : 'right';



> Nice work, only thing is that it looks like the margin on the toggle is placed on the wrong side when we use `dir='rtl'`. In CheckBox.js can we change line 167 to `const side = (!reverse != !theme.dir) ? 'left' : 'right';` so that the margin is placed on the correct side when we use RTL

Good catch, just made the suggested change.

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

PR #6084, opened 2022-04-22
add OnClickRow to also supoort onSelect

### LINKED ISSUE #5451: DataTable — cannot do `onSelect` and `onClickRow` together
### Expected Behavior

The clickable section of "onClickRow" should not include the selection checkbox so that the user could click on row content and also successfully check/uncheck the box.

### Actual Behavior

The clickable section of the row includes the checkbox, so a user cannot select a checkbox without the `onClickRow` function from being executed.

https://codesandbox.io/s/datatable-onselectonclickrow-7384x?file=/index.js

### Steps to Reproduce

1. Click on Checkbox in a body row
2. The `onClickRow` function is triggered, so user can't make a selection

### Your Environment

- Grommet version: 2.17.3
- Browser Name and version: MacOS Chrome
- Operating System and version (desktop or mobile): MacOS Mojave 10.14.6

britt6612:
#### What does this PR do?
This PR adds a option for the user to have the whole row clickable when using `onSelect` 
DataTable/body.js
storybook
storybook 
Added `onClickRow="onSelect"` to the story

It was brought up that users were not able to do onSelect and onClickRow together 
we decided this would not be something we recommend. If using onSelect then your onClickRow should be set to `onSelect` if user is wanting the whole row to be clickable otherwise `onClickRow` should not be used.
Closes #5451 

yes
yes
changed the keyboard behavior

### DIFF
--- a/src/js/components/DataTable/Body.js
+++ b/src/js/components/DataTable/Body.js
@@ -54,8 +54,7 @@ const Row = memo(
               adjustedEvent.datum = datum;
               adjustedEvent.index = index;
               onClickRow(adjustedEvent);
-            }
-            if (onClickRow === 'onSelect') {
+            } else if (onClickRow === 'onSelect') {
               if (isSelected) {
                 onSelect(selected.filter((s) => s !== primaryValue));
               } else onSelect([...selected, primaryValue]);
@@ -70,6 +69,11 @@ const Row = memo(
         {(selected || onSelect) && (
           <Cell
             pinnedOffset={pinnedOffset?._grommetDataTableSelect}
+            background={
+              (pinnedOffset?._grommetDataTableSelect &&
+                cellProps.pinned.background) ||
+              cellProps.background
+            }
             aria-disabled={isDisabled || !onSelect || undefined}
             column={{
               pin: Boolean(pinnedOffset?._grommetDataTableSelect),
@@ -180,16 +184,17 @@ const Body = forwardRef(
       <Keyboard
         onEnter={(event) => {
           if (
-            (onClickRow && active >= 0 && !disabled) ||
-            !disabled.includes(datumValue(data[active], primaryProperty))
+            onClickRow &&
+            active >= 0 &&
+            (!disabled ||
+              !disabled.includes(datumValue(data[active], primaryProperty)))
           ) {
             if (typeof onClickRow === 'function') {
               event.persist();
               const adjustedEvent = event;
               adjustedEvent.datum = data[active];
               onClickRow(adjustedEvent);
-            }
-            if (onClickRow === 'onSelect') {
+            } else if (onClickRow === 'onSelect') {
               const primaryValue = data[active]?.[primaryProperty];
               if (selected && selected.includes(primaryValue)) {
                 onSelect(selected.filter((s) => s !== primaryValue));
--- a/src/js/components/DataTable/Body.js
+++ b/src/js/components/DataTable/Body.js
@@ -191,7 +191,7 @@ const Body = forwardRef(
       } else onSelect([...selected, primaryValue]);
     };
 
-    const clickedRow =
+    const clickableRow =
       onClickRow &&
       active >= 0 &&
       (!disabled ||
@@ -200,9 +200,9 @@ const Body = forwardRef(
     return (
       <Keyboard
         onEnter={
-          clickedRow
+          clickableRow
             ? (event) => {
-                if (clickedRow) {
+                if (clickableRow) {
                   if (typeof onClickRow === 'function') {
                     event.persist();
                     const adjustedEvent = event;
@@ -217,7 +217,7 @@ const Body = forwardRef(
         }
         // The WCAG recommendation for checkboxes is to select them with "Space"
         onSpace={() => {
-          if (clickedRow) {
+          if (clickableRow) {
             if (onClickRow === 'select') {
               selectRow();
             }
--- a/src/js/components/DataTable/Body.js
+++ b/src/js/components/DataTable/Body.js
@@ -202,6 +202,7 @@ const Body = forwardRef(
             }
           }
         }}
+        // The WCAG recommendation for checkboxes is to select them with "Space"
         onSpace={() => {
           if (
             onClickRow &&
--- a/src/js/components/DataTable/Body.js
+++ b/src/js/components/DataTable/Body.js
@@ -202,6 +202,21 @@ const Body = forwardRef(
             }
           }
         }}
+        onSpace={() => {
+          if (
+            onClickRow &&
+            active >= 0 &&
+            (!disabled ||
+              !disabled.includes(datumValue(data[active], primaryProperty)))
+          ) {
+            if (onClickRow === 'onSelect') {
+              const primaryValue = data[active]?.[primaryProperty];
+              if (selected && selected.includes(primaryValue)) {
+                onSelect(selected.filter((s) => s !== primaryValue));
+              } else onSelect([...selected, primaryValue]);
+            }
+          }
+        }}
         onUp={onClickRow && active ? () => setActive(active - 1) : undefined}
         onDown={
           onClickRow && data.length && active < data.length - 1
--- a/src/js/components/DataTable/Body.js
+++ b/src/js/components/DataTable/Body.js
@@ -45,34 +45,30 @@ const Row = memo(
         size={size}
         active={active}
         aria-disabled={(onClickRow && isDisabled) || undefined}
-        onClick={
-          onClickRow && !isDisabled
-            ? (event) => {
-                // extract from React's synthetic event pool
-                event.persist();
-                const adjustedEvent = event;
-                adjustedEvent.datum = datum;
-                adjustedEvent.index = index;
-                onClickRow(adjustedEvent);
-              }
-            : undefined
-        }
+        onClick={(event) => {
+          if (onClickRow && !isDisabled) {
+            if (typeof onClickRow === 'function') {
+              // extract from React's synthetic event pool
+              event.persist();
+              const adjustedEvent = event;
+              adjustedEvent.datum = datum;
+              adjustedEvent.index = index;
+              onClickRow(adjustedEvent);
+            }
+            if (typeof onClickRow === 'string') {
+              if (isSelected) {
+                onSelect(selected.filter((s) => s !== primaryValue));
+              } else onSelect([...selected, primaryValue]);
+            }
+          }
+        }}
         onMouseEnter={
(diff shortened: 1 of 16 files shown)


src/js/components/DataTable/Body.js
```diff
@@ -45,34 +45,30 @@ const Row = memo(
...
+          if (onClickRow && !isDisabled) {
+            if (typeof onClickRow === 'function') {
+              // extract from React's synthetic event pool
+              event.persist();
+              const adjustedEvent = event;
+              adjustedEvent.datum = datum;
+              adjustedEvent.index = index;
+              onClickRow(adjustedEvent);
+            }
+            if (onClickRow === 'onSelect') {
```
taysea, line 58:
```suggestion
            else if (onClickRow === 'onSelect') {
```
liking need to fix formatting locally w linter.

src/js/components/DataTable/Body.js
```diff
@@ -45,34 +45,30 @@ const Row = memo(
...
-                setActive(index);
-              }
-            : undefined
+          onClickRow && !isDisabled ? () => setActive(index) : undefined
         }
         onMouseLeave={onClickRow ? () => setActive(undefined) : undefined}
       >
         {(selected || onSelect) && (
           <Cell
-            background={
```
taysea, line 71: Why was this removed?

src/js/components/DataTable/Body.js
```diff
@@ -181,20 +178,25 @@ const Body = forwardRef(
...
-                event.persist();
-                const adjustedEvent = event;
-                adjustedEvent.datum = data[active];
-                onClickRow(adjustedEvent);
-              }
-            : undefined
-        }
+        onEnter={(event) => {
+          if (
+            (onClickRow && active >= 0 && !disabled) ||
```
taysea, line 183: Formatting seems to have changed here. It should be:

```
(onClickRow && active >= 0 && (!disabled || !disabled.includes(datumValue(data[active], primaryProperty))

File: src/js/components/DataTable/Body.js [hunk 4]
```diff
@@ -181,20 +178,25 @@ const Body = forwardRef(
...
+            !disabled.includes(datumValue(data[active], primaryProperty))
+          ) {
+            if (typeof onClickRow === 'function') {
+              event.persist();
+              const adjustedEvent = event;
+              adjustedEvent.datum = data[active];
+              onClickRow(adjustedEvent);
+            }
+            if (onClickRow === 'onSelect') {
```
taysea, line 192:
```suggestion
            else if (onClickRow === 'onSelect') {
```
Fix formatting locally w linter

src/js/components/DataTable/stories/typescript/OnSelect.tsx
```diff
@@ -25,16 +25,9 @@ const columns: ColumnConfig<RowType>[] = [
...
-          pad={{ horizontal: 'small', vertical: 'xsmall' }}
-          onClick={() => {}}
-          hoverIndicator
-          fill
-        >
-          <Text truncate weight="bold">
-            {datum.location}
-          </Text>
-        </Box>
+        <Text truncate weight="bold">
```
taysea, line 28: Any reason we changed this?

britt6612: took this out so we can test how it is without having a row clickable since we wouldn't recommend doing this I can put back in

src/js/components/DataTable/Body.js
britt6612: done

src/js/components/DataTable/Body.js
britt6612: fixed

src/js/components/DataTable/Body.js
britt6612: put back in shouldn't have been removed

src/js/components/DataTable/Body.js
britt6612: done

src/js/components/DataTable/Body.js
```diff
@@ -181,20 +182,26 @@ const Body = forwardRef(
...
-            !disabled.includes(datumValue(data[active], primaryProperty)))
-            ? (event) => {
-                event.persist();
-                const adjustedEvent = event;
-                adjustedEvent.datum = data[active];
-                onClickRow(adjustedEvent);
-              }
-            : undefined
-        }
+        onEnter={(event) => {
```
taysea, line 185: The WCAG recommendation for checkboxes is to select them with "Space" (https://www.w3.org/TR/wai-aria-practices-1.1/examples/checkbox/checkbox-1/checkbox-1.html). That is how datatable keyboard selection currently works. So this would be a breaking change to only support "Enter" for selection.

Proposing that we support "Space" also for the selection when `onClickRow="onSelect"`.

britt6612: I'm glad you brought this up I was thinking the same especially since the normal behavior of `checkbox` is a "Space" bar I will add this with a comment

src/js/components/DataTable/Body.js
```diff
@@ -181,20 +182,42 @@ const Body = forwardRef(
...
+              const primaryValue = data[active]?.[primaryProperty];
+              if (selected && selected.includes(primaryValue)) {
+                onSelect(selected.filter((s) => s !== primaryValue));
+              } else onSelect([...selected, primaryValue]);
+            }
+          }
+        }}
+        // The WCAG recommendation for checkboxes is to select them with "Space"
+        onSpace={() => {
+          if (
```
taysea, line 207: Instead of duplicating this code for `onEnter` and `onSpace`, is there any way to create reusable function that both use?

britt6612: yes I pulled this out but left the `If` statement for the first part

src/js/components/DataTable/Body.js
```diff
@@ -179,22 +180,45 @@ const Body = forwardRef(
...
+            (!disabled ||
+              !disabled.includes(datumValue(data[active], primaryProperty)))
+          ) {
+            if (typeof onClickRow === 'function') {
+              event.persist();
+              const adjustedEvent = event;
+              adjustedEvent.datum = data[active];
+              onClickRow(adjustedEvent);
+            } else if (onClickRow === 'onSelect') {
+              selectCheckBox();
```
taysea, line 205: naming suggestion: `selectRow()`

src/js/components/DataTable/Body.js
```diff
@@ -179,22 +180,45 @@ const Body = forwardRef(
...
-                event.persist();
-                const adjustedEvent = event;
-                adjustedEvent.datum = data[active];
-                onClickRow(adjustedEvent);
-              }
-            : undefined
-        }
+        onEnter={(event) => {
+          if (
+            onClickRow &&
```
taysea, line 194: Could we potentially pull this out into a variable and place before the return:

```
const clickedRow = onClickRow && active >= 0 &&
            (!disabled ||
              !disabled.includes(datumValue(data[active], primaryProperty)))
```

And then use that in the onEnter and onSpace as 

```
if (clickedRow) {
...
}

File: src/js/components/DataTable/Body.js (hunk 8 above)
Comment (britt6612) [reply src/js/components/DataTable/Body.js:205]: I like that better

**Review by taysea — APPROVED** [review]
LGTM!

File: src/js/components/DataTable/index.d.ts [hunk 10]
```diff
@@ -105,7 +105,9 @@ export interface DataTableProps<TRowType = any> {
   step?: number;
 
   // Events
-  onClickRow?: (event: MouseClick<TRowType> | KeyPress<TRowType>) => void;
+  onClickRow?:
+    | 'onSelect'
```
ericsoderberghp, line 109: The term "onSelect" seems a little odd to me here. Like it will call any `onSelect` handler when the row is clicked. I'm wondering if we should use "select" here, indicating that if the user clicks the row, it will be treated as initiating a selection.

src/js/components/DataTable/Body.js
```diff
@@ -179,22 +180,41 @@ const Body = forwardRef(
...
     const [lastActive, setLastActive] = React.useState();
 
+    const selectRow = () => {
+      const primaryValue = data[active]?.[primaryProperty];
+      if (selected && selected.includes(primaryValue)) {
+        onSelect(selected.filter((s) => s !== primaryValue));
+      } else onSelect([...selected, primaryValue]);
+    };
+
+    const clickedRow =
```
ericsoderberghp, line 190: The term "clickedRow" seems to imply that the row was clicked. Perhaps "clickableRow" or "clickEnabled"?

src/js/components/DataTable/Body.js
```diff
@@ -179,22 +180,41 @@ const Body = forwardRef(
...
-            !disabled.includes(datumValue(data[active], primaryProperty)))
-            ? (event) => {
-                event.persist();
-                const adjustedEvent = event;
-                adjustedEvent.datum = data[active];
-                onClickRow(adjustedEvent);
-              }
-            : undefined
-        }
+        onEnter={(event) => {
```
ericsoderberghp, line 198: Can we only bind an event handler if the condition is true, rather than always binding a function and then testing whether it should do anything? For example: `onEnter={clickedRow ? (event) => { if ... } : undefined`

src/js/components/DataTable/Body.js
```diff
@@ -45,24 +45,24 @@ const Row = memo(
...
-                // extract from React's synthetic event pool
-                event.persist();
-                const adjustedEvent = event;
-                adjustedEvent.datum = datum;
-                adjustedEvent.index = index;
-                onClickRow(adjustedEvent);
-              }
-            : undefined
-        }
+        onClick={(event) => {
```
ericsoderberghp, line 48: Can we only bind a handler if needed? For example: `onClick={(onClickRow && !isDisabled) ? (event) => {} : undefined}`

src/js/components/DataTable/Body.js
britt6612: I can see that I like `clickableRow`

src/js/components/DataTable/Body.js
britt6612: done

src/js/components/DataTable/Body.js
britt6612: done

src/js/components/DataTable/index.d.ts
britt6612: that's fair and makes sense

ericsoderberghp (approved):

PR #3533, opened 2019-11-07
Changed DataTable to fix issues with header layout

ericsoderberghp:
#### What does this PR do?

Changed DataTable to fix issues with header layout

#### Where should the reviewer start?

Header.js

#### What testing has been done on this PR?

All DataTable stories across Safari, Chrome, and Firefox.
Also manually toggled properties in the Tunable story and changed the containing Box `fill` and `align` properties as well.

#### How should this be manually tested?

storybook

#### Any background context you want to provide?

https://grommet.slack.com/archives/C04LMHN59/p1573141859413900

#### Do the grommet docs need to be updated?

no

#### Should this PR be mentioned in the release notes?

yes

#### Is this change backwards compatible or is it a breaking change?

Despite seeming like it might not be, so far the testing indicates that it is backwards compatible.

PR #6086, opened 2022-04-22
Fixed hover indicator to fill cell on DataTable header

jcfilben:
#### What does this PR do?
Fixes an issue with DataTable where if the columns in the table are `sortable` and there are columns that have names that span multiple lines and columns names that are only on one line, the one line column header button will not fill the vertical space in the header. See screenshots below for a visual.

Notes:
1. This PR changes `height: auto` to `height: 100%` in `StyledDataTable.js`. We previously had height set to 100% but it was changed to auto in https://github.com/grommet/grommet/pull/3533 to fix an issue in FireFox where having a DataTable with the `groupBy` prop set, wrapped with `<Box fill>` resulted in this:
Screen Shot 2022-04-22 at 1 56 37 PM

See [this related slack thread](https://grommet.slack.com/archives/C04LMHN59/p1573141859413900) for more info.
The change from `height: auto` to `height: 100%` was made over 2 years ago and it looks like something either changed on our end with Grommet or with FireFox because I am no longer seeing this issue in FireFox when I set height back to 100%.

2. This fix does not work for FireFox, the sortable button in the header will not fill the cell if other cells wrap multiple lines. The fix is working in both chrome and safari.

[StyledDataTable.js](https://github.com/grommet/grommet/compare/master...jcfilben:fix/dataTable-hover?expand=1#diff-3c607ce17d65f5520f60a0b3ad2154f7d906796eb228bd0a0ebb076090a9286b)

Tested with the existing DataTable stories as well as the code provided in this slack thread https://grommet.slack.com/archives/C04LMHN59/p1573141859413900.

Closes https://github.com/grommet/grommet/issues/5968

Before:
Screen Shot 2022-04-22 at 1 48 16 PM
After:
Screen Shot 2022-04-22 at 1 48 50 PM

No

Yes

Backwards compatible

### DIFF
--- a/src/js/components/DataTable/Header.js
+++ b/src/js/components/DataTable/Header.js
@@ -372,6 +372,7 @@ const Header = forwardRef(
                 );
               }
 
+              // content should fill any available space in cell
               content = (
                 <Box flex="grow" fill={onResize ? 'vertical' : false}>
                   {content}
@@ -418,9 +419,6 @@ const Header = forwardRef(
                   </Box>
                 );
               }
-
-              // content should fill any available space in cell
-              // content = <Box flex="grow">{content}</Box>;
 
               const cellPin = [...pin];
               if (columnPin) cellPin.push('left');
--- a/src/js/components/DataTable/Header.js
+++ b/src/js/components/DataTable/Header.js
@@ -419,7 +419,6 @@ const Header = forwardRef(
                   </Box>
                 );
               }
-
               const cellPin = [...pin];
               if (columnPin) cellPin.push('left');
 
--- a/src/js/components/DataTable/Header.js
+++ b/src/js/components/DataTable/Header.js
@@ -345,7 +345,11 @@ const Header = forwardRef(
               }
 
               // content should fill any available space in cell
-              content = <Box flex="grow">{content}</Box>;
+              content = (
+                <Box flex="grow" fill={onSort ? 'vertical' : false}>
+                  {content}
+                </Box>
+              );
 
               if (search || onResize) {
                 const resizer = onResize ? (
--- a/src/js/components/DataTable/Header.js
+++ b/src/js/components/DataTable/Header.js
@@ -13,7 +13,7 @@ import { defaultProps } from '../../default-props';
 import { Box } from '../Box';
 import { Button } from '../Button';
 import { CheckBox } from '../CheckBox';
-import { TableCell } from '../TableCell';
+import { TableCell, verticalAlignToJustify } from '../TableCell/TableCell';
 import { Text } from '../Text';
 
 import { Resizer } from './Resizer';
@@ -325,11 +325,8 @@ const Header = forwardRef(
 
               if (verticalAlign || columnVerticalAlign) {
                 const vertical = verticalAlign || columnVerticalAlign;
-                let justify = 'center';
-                if (vertical === 'bottom') justify = 'end';
-                if (vertical === 'top') justify = 'start';
                 content = (
-                  <Box height="100%" justify={justify}>
+                  <Box height="100%" justify={verticalAlignToJustify[vertical]}>
                     {content}
                   </Box>
                 );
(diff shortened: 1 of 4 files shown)


taysea:
> 3. If the size prop is set on the DataTable the sortable button in the header will not fill the cell if other cells wrap multiple lines.

I notice in the inspector when DataTable has a size, the CSS changes to `height: auto` as opposed to `height: 100%`. Is there any way to have the overall table height still come through as `100%` in this case? Seems like that should fix it.

jcfilben:
Update: I added `height: fit-content` to the header when the `size` prop is defined. This fixed the issue where if the `size` prop is set the sortable button in the header will not fill the cell. Now the sortable button is filling the cell regardless of the `size` prop being defined or not.

src/js/components/DataTable/Header.js
```diff
@@ -345,7 +345,14 @@ const Header = forwardRef(
               }
 
               // content should fill any available space in cell
-              content = <Box flex="grow">{content}</Box>;
+              content = (
+                <Box
+                  flex="grow"
+                  fill={onSort && sortable !== false ? 'vertical' : false}
```
taysea, line 351: Wouldn't we still want the content to fill even if it's not sortable? For example for cases like: https://storybook.grommet.io/?path=/story/visualizations-datatable-vertical-align--vertical-align

taysea: Or just a broader question, can we just remove that adjustment all together? Does it work with just the other CSS modifications you've made?

jcfilben: Yes I will update this so that fill is always true. 
I'm going to go back looking into the vertical align story more, in this PR when I view the story with the hpe theme it is not respecting the vertical alignment and instead just aligning center. Marking as a draft while I look into this...

jcfilben: > Or just a broader question, can we just remove that adjustment all together? Does it work with just the other CSS modifications you've made?

Yeah, was actually able to remove in latest commit

jcfilben: Removing `fill` caused an issue with the resizable cells where the cell wasn't stretching when it was resized so I added back the fill prop when the columns are resizable `fill={onResize ? 'vertical' : false}`.

src/js/components/DataTable/Header.js
```diff
@@ -310,6 +323,18 @@ const Header = forwardRef(
                 );
               }
 
+              if (verticalAlign || columnVerticalAlign) {
+                const vertical = verticalAlign || columnVerticalAlign;
+                let justify = 'center';
```
taysea, line 328: Can we reuse `verticalAlignToJustify` from TableCell by chance?

src/js/components/DataTable/Header.js
```diff
@@ -347,7 +373,11 @@ const Header = forwardRef(
               }
 
               // content should fill any available space in cell
-              content = <Box flex="grow">{content}</Box>;
+              content = (
+                <Box flex="grow" fill={onResize ? 'vertical' : false}>
```
taysea, line 377: Can you add a comment of why we only do `fill` when it's onResize?

taysea: I'm not sure I follow why

src/js/components/DataTable/Header.js
jcfilben: Switched this to use `verticalAlignToJustify` in latest commit

src/js/components/DataTable/Header.js
jcfilben: I added a comment, I also realized this should be `fill={onResize || search ? 'vertical' : false}` so I update this in the latest commit. The reason `onResize` and `search` need the `fill` prop is later in the code if `onResize` or `search` is true content will get wrapped with another Box (see line 398 in Header.js), they need to be wrapped with another Box to have the correct behavior when resizing or typing in the search. If we don't set `fill` on the original Box it won't expand to fill the full space if wrapped with another Box.

britt6612: Your new comment looks great helps me understand

britt6612 (approved):
Everything works as expected are we going to worry about it not working in firefox again or ?

taysea (approved):
Nice work.

taysea:
> Everything works as expected are we going to worry about it not working in firefox again or ?

From my understanding this PR isn't breaking anything for Firefox, it just leaves the behavior as is/doesn't enhance it. Can you confirm @jcfilben.

britt6612:
> > Everything works as expected are we going to worry about it not working in firefox again or ?
> 
> @britt6612 From my understanding this PR isn't breaking anything for Firefox, it just leaves the behavior as is/doesn't enhance it. Can you confirm @jcfilben.

maybe I misunderstood

ericsoderberghp (approved):

Issue #5954, opened 2022-02-16
DropContainer "place" method use wrong boundaries


LeComptoirDesPharmacies:
With a custom `ContainerTargetContext` on a box with `transform` css property, the `Drop` position is wrongly computed.

Indeed, `Drop` fixed position (width, top, bottom, left) is computed using the `place` method of `DropContainer`. (See: https://github.com/grommet/grommet/blob/master/src/js/components/Drop/DropContainer.js#L92) which is using clientRect bounds.

However, according to CSS documentation (https://developer.mozilla.org/fr/docs/Web/CSS/position#fixed), a fixed element will position itself according to viewport only if there is no ancestor with [transform](https://developer.mozilla.org/fr/docs/Web/CSS/transform), [perspective](https://developer.mozilla.org/fr/docs/Web/CSS/perspective) or [filter](https://developer.mozilla.org/fr/docs/Web/CSS/filter) css property.

### Expected Behavior

Drop should be positioned next to its target.

### Actual Behavior

Drop position get a wrong offset which is linked tot he offset of the ancestor with "transform" property.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-forked-sgxv28?file=/index.js

### Steps to Reproduce

1. Go to the sandbox
2. Click on "Fancy Selector"
3. See that Drop open with a large top offset

4. Remove the `transform` property
5. Drop is well positionned

### Your Environment

- Grommet version: 2.20.1
- Browser Name and version: Firefox 97.0
- Operating System and version (desktop or mobile): Windows 10

PR #6097, opened 2022-04-28
Fix drop container transform

JrDuComptoirDesPharmacies:
#### What does this PR do?

Fixes the drop container position when it inlude in a transformed parent position (fixed position is not computed according viewport)

Before the fix: https://codesandbox.io/s/adoring-currying-ztj374
After the fix: https://codesandbox.io/s/grommet-v2-template-forked-vojosq

See in `src/js/components/Drop/DropContainer.js` in place method

Maual testing through firefox/chrome using codesandbox from issue https://github.com/grommet/grommet/issues/5954

Click on the Fancy Selector and control the drop position

Description of the fixed position according to transformed parent:
https://developer.mozilla.org/fr/docs/Web/CSS/position#fixed

Description of way to find the containing block used for fixed element position computation :
https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block

Similar to this issue: https://github.com/floating-ui/floating-ui/issues/1577

Grommet solution inspired of this code: https://github.com/floating-ui/floating-ui/blob/a3299dc9c56a7c95183dfb3463f0542db4e945e2/packages/dom/src/utils/getOffsetParent.ts#L22

See https://github.com/grommet/grommet/issues/5954

**Before**:
image

**After**:
image

No

Yes

Backwards compatible


src/js/components/Drop/DropContainer.js
```diff
@@ -30,6 +30,49 @@ const preventLayerClose = (event) => {
...
+const getContainingBlock = (element) => {
+  const isIE = window.navigator.userAgent.toLowerCase().includes('trident');
+  if (isIE && element instanceof window.HTMLElement) {
+    // In IE 9, 10 and 11 fixed elements containing block is always established
+    // by the viewport
+    const css = window.getComputedStyle(element);
+    if (css.position === 'fixed') {
+      return null;
+    }
+  }
```
jcfilben, line 47: Not sure if we need to include this since internet explorer's end of life is June 15, 2022.

src/js/components/Drop/DropContainer.js
```diff
@@ -30,6 +30,49 @@ const preventLayerClose = (event) => {
...
+    // create a containing block.
+    // https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block
+    if (
+      (css.transform ? css.transform !== 'none' : false) ||
+      (css.perspective ? css .perspective !== 'none' : false) ||
+      (css.backdropFilter ? css.backdropFilter !== 'none' : false) ||
+      css.contain === 'paint' ||
+      ['transform', 'perspective'].includes(css.willChange) ||
+      (isFirefox && css.willChange === 'filter') ||
+      (isFirefox && (css.filter ? css.filter !== 'none' : false))
```
jcfilben, line 67: Based on https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block it looks like it doesn't matter what browser just if filter has a value
```suggestion
      (css.filter ? css.filter !== 'none' : false)
```

src/js/components/Drop/DropContainer.js
JrDuComptoirDesPharmacies: you're right

src/js/components/Drop/DropContainer.js
JrDuComptoirDesPharmacies: I tested (https://codesandbox.io/s/silly-sara-2o04hm?file=/index.js) and it still works

jcfilben (approved):
Looks good!

AntoineDuComptoirDesPharmacies:
Hi @ericsoderberghp,

Any news about this PR ?
Do not hesitate to tell if you want us to change something.

Yours faithfully,
LCDP

ericsoderberghp:
I think I at least partly understand what's going on here. We are skipping over parents that have some sort of CSS transformation so that we can adjust the drop position by the appropriate offset of the transformation. Two thoughts:
- what if the immediate parent and the great-grandparent both have transformations? I think we will be missing the second one since we just stop at the grandparent?
- the code to get computed style and look through various CSS properties feels a bit fragile and I'm wondering if there's a cleaner way to accomplish it

JrDuComptoirDesPharmacies:
Thank you for your reply. To answer your two thoughts:
- The principle is to try to find the first parent that has a transformation and position the component from there. More distant relatives are not taken into account for the position calculation. Here's an example that shows multiple nested transforms aren't a problem: https://codesandbox.io/s/quizzical-shadow-h5q68z?file=/index.js
- For the second point, many projects rely on this type of detection to identify the containing block (https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block), here are some examples :
    - https://github.com/floating-ui/floating-ui/blob/3dd6b42489b2beaeb97379769a6e4d63333ee171/packages/dom/src/utils/is.ts
    - https://github.com/sonnat/sonnat-ui/blob/a2c2582f4548c7ff354da56799c072924f376ff1/lib/utils/is.ts

Yours faithfully,
LCDP

GeryDuComptoirDesPharmacies:
Hello,

Someone available to merge this PR ?

Best regards,
Géry

GeryDuComptoirDesPharmacies:
Hi,
@ericsoderberghp still waiting for your approval.

Can you tell if changes are needed please ?

Best regards,
Géry

jonscottclark:
Hello :) I have been testing this PR today, and it did help us solve a particular use case where we have a Drop being positioned incorrectly when `inline === true`. Thank you to @JrDuComptoirDesPharmacies for your work on this!

I also have some suggestions which improve on the solution.

1. When the containingBlock is `overflow: auto`, we need to factor in the scrollLeft and scrollTop of that container, otherwise the placement is incorrect if the containingBlock has been scrolled at all.

In `DropContainer.js`, when setting the final position values we need to use these new calculations:

```js
container.style.left = `${left - viewportOffsetLeft + (containingBlock?.scrollLeft ?? 0);

// ...

if (top !== '') {
  container.style.top = `${top - viewportOffsetTop +
    (containingBlock?.scrollTop ?? 0)}px`;
}
if (bottom !== '') {
  container.style.bottom = `${viewportOffsetBottom - bottom -
    (containingBlock?.scrollTop ?? 0)}px`;
}

2. The Drop does not update its position if the horizontal position of the target has changed, i.e. if it is within a containingBlock that has been scrolled to the left or right. (This is somewhat unrelated to the initial changes, but it does fix a positioning bug):

In `DropContainer.js`:

```diff
let scrollParents;
+ let scrollParentsHorizontal;

const addScrollListeners = () => {
  scrollParents = findScrollParents(dropTarget);
  + scrollParentsHorizontal = findScrollParents(dropTarget, true);
  scrollParents.forEach((scrollParent) =>
    scrollParent.addEventListener('scroll', place),
  );
  + scrollParentsHorizontal.forEach((scrollParent) =>
    + scrollParent.addEventListener('scroll', place),
  + );
};
```

These changes work when the `inline` prop is `true` or `false`. We found better performance and control over layering when `inline: true`.

AntoineDuComptoirDesPharmacies:
Hi @jonscottclark,

Thank you for this additional use case, it seems logic to realign the drop when the horizontal scroll is changing too.

I tried to modify the CodeSandbox to represent your use case without success.
Could you Fork and update this CodeSandbox so we can see the problem please ? 
[https://codesandbox.io/s/grommet-v2-template-forked-sgxv28?file=/index.js](https://codesandbox.io/s/grommet-v2-template-forked-sgxv28?file=/index.js)

Once done, we will compare the behavior with and without your fix and we will add your changes to the Pull Request.

Thanks in advance !
Yours faithfully,
LCDP

jonscottclark:
@AntoineDuComptoirDesPharmacies Good idea, I'll attempt to make a minimum reproduction of the situation I'm describing.

jonscottclark:
@AntoineDuComptoirDesPharmacies 

Here's the simplified scenario: https://codesandbox.io/s/grommet-v2-template-forked-fqhfx5?file=/index.js

We have a very wide Box with items in it - scroll horizontally to see all of the items.

After clicking a button the incorrect "top" positioning is as expected - the currently broken positioning that your PR already fixes.

Now try clicking on "Open Drop 1", then start scrolling vertically and horizontally. You will see the necessity of checking for `containingBlock.scroll<Top|Left>`! :)

If you change `inline` to `false`, you lose the horizontal scroll updating which is what I address in the 2nd item, however, I'm not super happy with the implementation (more resize listeners) - but it does what we need it to! Open to suggestions on how to improve that

AntoineDuComptoirDesPharmacies:
@jonscottclark Thank you for this perfect reproducer, we can notice that the bug you outlined is resolved with your suggested modifications.

We have just added your modifications to the current PR.

Yours faithfully,
LCDP

jonscottclark:
@AntoineDuComptoirDesPharmacies test snapshot looks like it can be safely updated

The Chromatic step is failing due to the appId not being recognized:

```
Authenticating with Chromatic
    → Connecting to https://index.chromatic.com
    → No app with code '9q99in2ygnh' found

**Comment by AntoineDuComptoirDesPharmacies** [comment]
Thank you for the ping @jonscottclark.

We merged the current master branch in the PR so tests pass now.

Yours faithfully,
LCDP

**Comment by jonscottclark** [comment]
Hey @ericsoderberghp , friendly ping to see if you have time to look at this again. Thank you in advance :)

**Comment by jonscottclark** [comment]
*bump* 🫣

**Comment by AntoineDuComptoirDesPharmacies** [comment]
Hi @ericsoderberghp,

Using absolute positioning is not always possible or the easiest solution.
Here is one of our use case which is hard to solve with absolute : 
- We have a slider which is offscreen (left) and come into the screen when user click on a special button
- In this slider, we have Select Components that user can open to search in with a Drop for options display
- When user close the slider by swiping left (phone, tablet), the slider close but Select remain open

In this context, keeping default portal container our Select Drop will remain in the middle of the screen.
A solution is to create a portal into the slider so the Drop content will follow when user swipe left to close the slider.
However, as slider have a "transform" CSS property to smoothly enter/leave the screen, the Drop position is wrong.

**Comment by AntoineDuComptoirDesPharmacies** [comment]
@jonscottclark, what is the commit number you want integrated in the feature branch ?
Because I rebased the branch and republish on the 23th of march through the following version : [2.31.0-fix-drop-position.1](https://www.npmjs.com/package/@lcdp/grommet/v/2.31.0-fix-drop-position.1)

Yours faithfully,
LCDP

**Comment by GeryDuComptoirDesPharmacies** [comment]
Hello, 

We have this PR in our dashboard since last year, we want to get rid of it.

So tell us can you merge it or not ? We spend a whole year to merge one file change, really guys ??

@ericsoderberghp @jcfilben @jonscottclark 

You are doing a great work on grommet but please be more reactive.

**Comment by ericsoderberghp** [comment]
I'm wondering if there is a slightly simpler way to handle this by having the caller provide a reference to the transformed element explicitly rather than having grommet have to figure it out programmatically? For instance: `dropProps={{ transformContainer: <ref or id> }}` How does that sound?

**Comment by AntoineDuComptoirDesPharmacies** [comment]
In fact it would remove the need of going through parent tree in order to find the transform container.

But, in my opinion, it will add a lot of pain to developpers finding the node by hand, creating React `ref`, passing `ref` through the DOM tree to the last children.
Moreover, the transform container may be generated by third party library without the capability to add a `ref` on it.

To finish with, it will be less maintainable for developpers because if one day someone add a new transform container inside the existing tree, the drop will be bad positioned again.

**Comment by ericsoderberghp** [comment]
Those are good points about potentially not having access to the code to create a ref or an id. I'm still reluctant to have to run through the parent tree in this way on every call to `place()`. Since this is a rare situation, perhaps we can put it behind something that callers can opt-in to. For example, via either https://v2.grommet.io/grommet#options or something like `<Box align={{ checkContainingBlock: true }} />`. Thoughts?

**Comment by AntoineDuComptoirDesPharmacies** [comment]
Good idea, we added `drop.checkContainingBlock` option in the PR in order to opt-in this tree discovery algorithm. 👍

**Comment by jonscottclark** [comment]
@AntoineDuComptoirDesPharmacies I apologize for the delay in testing this again.

I like the global opt-in behaviour. It leaves room to test this further and to potentially optimize it (or move the opt-in to a narrower scope in future).

There have been more requests from my teams to address the visual bug that would be resolved by this fix so I will be testing the latest set of commits against our desired use case, but it really LGTM at this point!

**Comment by AntoineDuComptoirDesPharmacies** [comment]
No problem @jonscottclark , we all have our own planning and your commits were a good addition to this pull request.
Note that the part fixing horizontal scroll is out of this new opt-in option scope and will be enabled for everyone as it allow to fix scrolling without costly code.

**Comment by GeryDuComptoirDesPharmacies** [comment]
Sorry, to bother you guys. This issue still in our dashboard, @ericsoderberghp could you have a look and close this issue if the PR is good for you please ?

Been a while we are working to get it merged.

**Comment by GeryDuComptoirDesPharmacies** [comment]
Hi everyone,

Any news about when this PR will be merge please ?

@ericsoderberghp @jonscottclark

File: src/js/components/Drop/DropContainer.js [hunk 3]
```diff
@@ -202,7 +235,29 @@ const DropContainer = forwardRef(
...
+          const viewportOffsetLeft = containingBlockRect?.left ?? 0;
+          const viewportOffsetTop = containingBlockRect?.top ?? 0;
+          const viewportOffsetBottom =
+            containingBlockRect?.bottom ?? windowHeight;
+
+          const containerOffsetLeft = containingBlock?.scrollLeft ?? 0;
+          const containerOffsetTop = containingBlock?.scrollTop ?? 0;
+
+          container.style.left = `${left - viewportOffsetLeft +
+            containerOffsetLeft}px`;
```
jcfilben, line 259: I'm wondering if we can structure this code so it is a bit more contained and is only run when `dropOptions.checkContainingBlock = true`. For example:

```
if (dropOptions?.checkContainingBlock && getContainingBlock(container)) {
  ...
} else {
  ...
}
```

src/js/components/Drop/DropContainer.js
```diff
@@ -231,19 +288,28 @@ const DropContainer = forwardRef(
       };
 
       let scrollParents;
+      let scrollParentsHorizontal;
```
jcfilben, line 291: From what I was seeing the scroll horizontal issue was only happening when it was within a containing block that has been transformed. Can we do the scrollParentsHorizontal code only if `dropOptions.checkContainingBlock = true`?

jcfilben (commented):
Sorry for the long wait on this! Just a few small comments otherwise I think this looks good

src/js/components/Drop/DropContainer.js
jonscottclark: (quick/lazy response from me without testing explicitly)

I think even for a case where `getContainingBlock` returns `null` (it has traversed all parents), or the `checkContainingBlock` option is not enabled, we still want the Drop positioning to respond to horizontal scrolling.

For example, say there's a box with `overflow: auto` that has been given an explicit width (that is more than the viewport width) based on the number of children inside of it, and it scrolls horizontally only. Even if we haven't opted in to this new behaviour, we don't get Drop positioning updates when this container is scrolled.

src/js/components/Drop/DropContainer.js
jonscottclark: imo, `containingBlockRect = containingBlock?.getBoundingClientRect();` is the expensive operation and it only gets run if the option is enabled.

All of the nullish coalescing in the extra variables that depend on `containingBlockRect` is not a performance hit. Maybe this is just a stylistic suggestion but I prefer the current style, defining each value in a single `const` declaration

jonscottclark:
@AntoineDuComptoirDesPharmacies or @GeryDuComptoirDesPharmacies , could you please rebase against the main grommet branch and publish a test version to the npm registry? We will do some additional testing on our end and share the results.

jonscottclark:
Currently testing...

1. `yarn add grommet@npm:@lcdp/grommet@2.32.0-fix-drop-position.1`

2. Updated our root `<Grommet>` container with `options={{ drop: { checkContainingBlock: true } }}`

Will update with some results!

jonscottclark:
@AntoineDuComptoirDesPharmacies Could you perhaps publish a version of `@lcdp/grommet` that is up to date with the main branch here, just so I am certain I'm not seeing unrelated issues during testing?

jonscottclark (approved):
I tested this locally by cloning the [LDCP fork](https://github.com/LeComptoirDesPharmacies/grommet/tree/fix-drop-container-transform), rebasing against the `grommet` main branch (only conflicts were package.json), running a local build and then linking the `/dist` output locally to our project. I found that one change was needed to make this PR work perfectly for some real-world use cases 🎉

See here that we reference `dropTarget`, (the prop): https://github.com/grommet/grommet/pull/6097/files#diff-a8c7783b048816fb6a81c4d4d49839bb165a273f34164cc9fa580226391aa51eR294-R295

We need to instead first check if `dropTarget?.current` is defined, like we do in the `place()` function: https://github.com/grommet/grommet/pull/6097/files#diff-a8c7783b048816fb6a81c4d4d49839bb165a273f34164cc9fa580226391aa51eR146

If we move L146 out of `place()` and put it at L96, then rename `dropTarget` to `target` in `addScrollListeners()`, everything works 🙌

This works with `inline: true` and `inline: false`.

This is with `{ drop: { checkContainerBlock: true }}`. With this option off, the positioning is okay with `inline: false`, but with `inline: true` there's erroneous positioning due to the omission of the containing block's position - for the reasons I list here: https://github.com/grommet/grommet/pull/6097#issuecomment-1425158981

So it seems this work actually contains two things:

- A fix to the scroll listeners not always sending a valid element to `findScrollParents`
- The opt-in behaviour to calculate drop positioning of possible containing blocks

For what it's worth, and to continue my rambling, we will 100% be opting in to this behaviour (and hopefully it becomes a default), and also using `inline: true` for all of our Drop instances (so that we can tell rendered Drop content to layer beneath some ancestor sticky elements - with `inline: false` the Drop content will always go overtop of everything)

Thank you, LDCP folks for the initial PR and continued attention, and Grommet folks for the work you do 🤝

src/js/components/Drop/DropContainer.js
jcfilben: No performance concerns here, rather I just wanted to organize the code this way for readability. Since it is an opt in behavior and not the default it would be nice (but not critical) to have it separated a bit more in the code.

src/js/components/Drop/DropContainer.js
jcfilben: I tested with the [code found here](https://codesandbox.io/s/grommet-v2-template-forked-hv4hf6?file=/index.js)m which I think is the situation you are describing. I tested with both a horizontal scroll case and a vertical scroll case. In both situations the drop stays in view when I scroll horizontally or vertically both in the current version of grommet and when testing using the code in this PR. 

Previously in this thread I mentioned that I was seeing the horizontal scroll issue when the drop was within a containing block that had been transformed. I realized that the fix for this was not actually coming from the scrollParentsHorizontal code. When I commented that code out and left the containing block code in I got the same behavior upon scrolling horizontally so the drop is out of view. 

Maybe I am missing the correct test case to see the issue that the scrollParentsHorizontal code is fixing.

jonscottclark:
@GeryDuComptoirDesPharmacies I should clarify that I approved the PR (instead of saying Request Changes) because it works for me assuming this comment is implemented:

> See here that we reference dropTarget, (the prop): [#6097 (files)](https://github.com/grommet/grommet/pull/6097/files#diff-a8c7783b048816fb6a81c4d4d49839bb165a273f34164cc9fa580226391aa51eR294-R295)

> We need to instead first check if dropTarget?.current is defined, like we do in the place() function: [#6097 (files)](https://github.com/grommet/grommet/pull/6097/files#diff-a8c7783b048816fb6a81c4d4d49839bb165a273f34164cc9fa580226391aa51eR146)

> If we move L146 out of place() and put it at L96, then rename dropTarget to target in addScrollListeners(), everything works 🙌

Just didn't want that point to get lost in the text :) Thanks

GeryDuComptoirDesPharmacies:
@jonscottclark We will implement what you ask for.
Actually we are a little bit overwork. In the end of the month, it will be done.

Thanks for your time.

GeryDuComptoirDesPharmacies:
Hi,

Review is done, let's merge it this time please 🙏🏾
Let's don't waste time as we have rebase from master.

@ericsoderberghp @jonscottclark @jcfilben 

Thanks.

jcfilben:
Hi @GeryDuComptoirDesPharmacies I am still seeing issues with the drop positioning behavior.


https://github.com/grommet/grommet/assets/54560994/68735cf8-9723-4633-9129-0632235607a6

Since the primary issue this PR is trying to solve is the containing block issue I would be okay moving forward with only the containing block code and removing the `scrollParentsHorizontal` code.

jonscottclark:
@jcfilben is the issue you mention the seemingly slow updating of the drop position?

jcfilben:
> @jcfilben is the issue you mention the seemingly slow updating of the drop position?

Yes

jcfilben:
> @jcfilben Okay - I see there's a lot of stuff happening in the `place()` function that is related to top/bottom positioning but not the same equivalent for `left`, so I suppose that needs a bit more attention in order to support the horizontal scrolling case.
> 
> I'm okay with removing the `scrollParentsHorizontal` support from the scope of this PR so it can progress, and I can possibly open a PR to add support for `scrollParentsHorizontal` later, as trying to add that functionality to this PR will delay it even further.
> 
> If you agree, then perhaps we can get a check from @ericsoderberghp once it's removed since everything else is looking great in this PR.
> 
> cc @GeryDuComptoirDesPharmacies

Yes, I'm supporting of removing the `scrollParentsHorizontal` code and handling that issue in a separate PR

AntoineDuComptoirDesPharmacies:
@jcfilben @jonscottclark 
O.K to us, the PR have been updated with horizontal scroll fix removed.

Yours faithfully,
LCDP

GeryDuComptoirDesPharmacies:
Hi everyone, 

Everything is fine now. Can we merge please ?

@jcfilben @jonscottclark @ericsoderberghp

src/js/components/Drop/DropContainer.js
```diff
@@ -53,13 +84,16 @@ const DropContainer = forwardRef(
...
+    // dropOptions was created to preserve backwards compatibility
+    const { drop: dropOptions } = useContext(OptionsContext);
     const portalContext = useContext(PortalContext);
     const portalId = useMemo(() => portalContext.length, [portalContext]);
     const nextPortalContext = useMemo(
       () => [...portalContext, portalId],
       [portalContext, portalId],
     );
     const dropRef = useForwardedRef(ref);
+    const target = dropTarget?.current || dropTarget;
```
jcfilben, line 96: Why does this need to be defined outside of the useEffect?

AntoineDuComptoirDesPharmacies: It do not need to be defined outside.
I followed the recommendations of @jonscottclark here : https://github.com/grommet/grommet/pull/6097#pullrequestreview-1563661513

I just moved it into the useEffect to save some computation.
Yours faithfully,
LCDP

jcfilben (approved):
Looks good!

halocline (approved):

taysea (approved):

jcfilben:
Looks like there are some merge conflicts that need to be resolved before merging

AntoineDuComptoirDesPharmacies:
@jcfilben conflict have been resolved

JrDuComptoirDesPharmacies:
🥳

Issue #6088, opened 2022-04-22
List — Add `disabled` prop that takes array of disabled items


taysea:
### Expected Behavior

List should allow for `disabled` prop that takes an array of items that should be disabled from interaction (similar to DataTable `disabled` https://github.com/grommet/grommet/pull/6070).

TBD: Should the array contain the index of the List item to be disabled (similar to List [itemProps](https://v2.grommet.io/list#itemProps) OR the primaryValue of the list item when a primaryKey is supplied?

-----------------------------
Desired behavior:

- [x] Add prop allowing array of indexes to be supplied by caller.
- [x] Disabled items should be included in the list length and are visually styled as disabled at the specified indices.
- [x] Because disabled items are in the list length and a visually perceivable, we need to replicate this behavior for screen reader users. 
  - As example, consider presenting a product on a product detail page. That product may have attributes such as color and size. The color and sizes options are presented in a list so that the user knows what options are available. However, at certain points in time, a particular color or size may be out of stock and the interface will want to display that item as disabled, indicating that that item is normally carried, but currently out of stock.
  - [x] Ensure that disabled items are included in the "item X of Y" count. Identify list items as disabled to screen readers.`aria-disabled="true` [is preferred](Identify list items as disabled to screen readers.) over the `disabled` attribute as it includes the item in the count, allows for focus, and announces "disabled" or "dimmed" depending on the browser and screen reader.
- [x] onClick and onEnter events should be prevented for disabled items.
- [x] Treatment of disabled items when List is in onOrder state:
  - [x] Disabled items should be included in the onOrder list.
  - [x] Disabled items should retain disabled styling and `aria-disabled`, however, the items may be re-ordered.
  - [x] Disabled items should be able to be re-ordered. Disabled indices should be updated to reflect the new order.
    - As example, consider a list of rules which are to be applied in a certain order, such as a lead quality filter regulating the amount of leads being distributed to a sales team. As circumstances change, the order of the rules may want to be modified and/or certain rules may want to be temporarily disabled.  

### Actual Behavior

Not currently supported.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #6111, opened 2022-05-04
List - add disabled prop

halocline:
#### What does this PR do?

Closes https://github.com/grommet/grommet/issues/6088

Add `disabled` prop to List.

- [x] Add prop allowing array of indexes to be supplied by caller.
- [x] Disabled items should be included in the list length and are visually styled as disabled at the specified indices.
- [x] Because disabled items are in the list length and a visually perceivable, we need to replicate this behavior for screen reader users. 
  - As example, consider presenting a product on a product detail page. That product may have attributes such as color and size. The color and sizes options are presented in a list so that the user knows what options are available. However, at certain points in time, a particular color or size may be out of stock and the interface will want to display that item as disabled, indicating that that item is normally carried, but currently out of stock.
  - [x] Ensure that disabled items are included in the "item X of Y" count. Identify list items as disabled to screen readers.`aria-disabled="true` [is preferred](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-disabled) over the `disabled` attribute as it includes the item in the count, allows for focus, and announces "disabled" or "dimmed" depending on the browser and screen reader.
- [x] onClick and onEnter events should be prevented for disabled items.
- [x] Treatment of disabled items when List is in onOrder state:
  - [x] Disabled items should be included in the onOrder list.
  - [x] Disabled items should retain disabled styling and `aria-disabled`, however, the items may be re-ordered.
  - [x] Disabled items should be able to be re-ordered. Disabled indices should be updated to reflect the new order.
    - As example, consider a list of rules which are to be applied in a certain order, such as a lead quality filter regulating the amount of leads being distributed to a sales team. As circumstances change, the order of the rules may want to be modified and/or certain rules may want to be temporarily disabled.  

stories/List/Disabled

Jest tests added.

stories/List/Disabled
- [ ] basic disabled
- [ ] basic disabled with screen reader
- [ ] disabled + onClick
- [ ] disabled + onClick + keyboard
- [ ] disabled + onClick + screen reader
- [ ] disabled + onOrder
- [ ] disabled + onOrder + keyboard
- [ ] disabled + onOrder + screen reader

- [x] `userEvent` is used in place of `fireEvent`. **userEvent.keyboard** not behaving as expected.

https://github.com/grommet/grommet/issues/6088

Yes.

Yes.

Backwards compatible.


jcfilben (commented):
While using a screenreader I am able to select disabled options and the screen reader will read "list item dimmed, selected (2 of 5)"

src/js/components/List/List.js
```diff
@@ -52,6 +53,17 @@ const StyledItem = styled(Box)`
   &:focus {
     ${unfocusStyle({ forceOutline: true, skipSvgChildren: true })}
   }
+  ${(props) => {
+    let disabledStyle;
```
britt6612, line 57: Wondering if we should have a hover over if it is `disabled` on DataTable the `disabled` story the hover behavior is not there http://localhost:9001/?path=/story/visualizations-datatable-onselect--on-select-data-table

britt6612, line 57: Wondering if we should have a hover over if it is `disabled` on DataTable the `disabled` story the hover behavior is not there http://localhost:9001/?path=/story/visualizations-datatable-onselect--on-select-data-table

src/js/themes/base.d.ts
```diff
@@ -1012,13 +1012,11 @@ export interface ThemeType {
...
     item?: {
       background?: BackgroundType;
-      border?:
-        | string
-        | {
-            side?: string;
-            color?: ColorType;
-            size?: string;
-          };
+      border?: BorderType;
```
britt6612, line 1015: 👍

britt6612, line 1015: 👍

src/js/components/List/List.js
halocline: Mmmm... good call. I think we want the active styling when using keyboard (because I need to know which item I am currently on), but agree that it is unnecessary when mouseover.

britt6612: I agree with the using `Keyboard` 👍

halocline: addressed.

halocline:
> While using a screenreader I am able to select disabled options and the screen reader will read "list item dimmed, selected (2 of 5)". My expectation would be that I shouldn't be able to select disabled items

Addressed.

src/js/components/List/List.js
britt6612: nice much better!

jcfilben (approved):
This looks good to me, thanks for adding jest tests!
One thing to note is on Chrome and Firefox while using voice over, options that are not disabled are read as selected even if we haven't selected them. This is the current behavior on the master branch and I think fixing this falls outside the scope of this PR, so I am fine with leaving this as is for now. Just something to note that we should fix later.

britt6612 (approved):

ericsoderberghp (approved):

PR #6125, opened 2022-05-12
Improving Accessibility of Calendar (#5677)

Ojas-Sri:
Added functionality for the focus to be on the day, when a user navigates using a keyboard.

#### What does this PR do?

**In code**
- This PR introduces `focus` prop to the `StyledDayContainer`.
- This PR also updated snapshots for `Calendar` component using `yarn test -u`

 **For the User**
- If the user navigates to the `Calendar` component using `tab` key, the first day of the month will be in focus rather than the entire `Calendar`
- If the user decides navigates the calendar using the arrow key, the focus will shift to the active day.
- If the user decides to return to mouse pointer, the entire `Calendar` would come in focus on Mouse Down ( & selection)

- The reviewer could start from file `src/js/components/Calendar/StyledCalendar.js` , where I have added the required props.
- Further, the reviewer can refer file `src/js/components/Calendar/Calendar.js` at line 1010, 1012 and thereon.

- I have tested it using the storybook. Screenshots of the functionality below : 

PR (23)

- The tester could use tab keys to reach the `Calendar` Component, once in focus, the arrow keys could be used to navigate.

#5677 

PR (24)

- Don't think so.

- Not sure

- It should be.

### DIFF
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -173,12 +173,18 @@ const CalendarDay = ({
   children,
   fill,
   size,
+  focus,
   isInRange,
   isSelected,
   otherMonth,
   buttonProps = {},
 }) => (
-  <StyledDayContainer role="gridcell" sizeProp={size} fillContainer={fill}>
+  <StyledDayContainer
+    role="gridcell"
+    sizeProp={size}
+    fillContainer={fill}
+    focus={focus}
+  >
     <CalendarDayButton fill={fill} {...buttonProps}>
       <StyledDay
         disabledProp={buttonProps.disabled}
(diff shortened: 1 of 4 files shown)


jcfilben (approved):
Looks good!

ericsoderberghp:
This behavior seems inconsistent with how other grommet components work. For example, List and DataTable tab puts focus on the container and the arrow keys change the active item. Select and TextInput with suggestions leave focus on the input and use the arrow keys to change the active option. I think the main theme of the issue was in making the currently active date more easily visible than just with the grey background. I think we should explore that aspect, and apply it to all components consistently, rather than handling focus as is currently being done here.

When I interacted with the Calendar story and tabbed to show the first date of the month, I was then expecting to continue tabbing through the dates. It seemed surprising that the tab key didn't advance the focus to the next date in this mode.

The outline being shown is being clipped for the dates that are on the edges of the container.

jcfilben:
> This behavior seems inconsistent with how other grommet components work. For example, List and DataTable tab puts focus on the container and the arrow keys change the active item. Select and TextInput with suggestions leave focus on the input and use the arrow keys to change the active option. I think the main theme of the issue was in making the currently active date more easily visible than just with the grey background. I think we should explore that aspect, and apply it to all components consistently, rather than handling focus as is currently being done here.

Based on accessibility info found here: https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html, I think the expected accessible behavior is for the focus indicator to be on the date instead of the Calendar container. I realize this isn't how we are handling things in other grommet components, but this seems like the correct behavior for Calendar.

> When I interacted with the Calendar story and tabbed to show the first date of the month, I was then expecting to continue tabbing through the dates. It seemed surprising that the tab key didn't advance the focus to the next date in this mode.

For the tab behavior, only the active date should be in the tab index. The other dates should be reachable with the arrow keys. If all the dates were in the tab index, it would take a long time to tab through them to get to the next item on the page. (See example in the link above to see how tabindex is handled there).

> The outline being shown is being clipped for the dates that are on the edges of the container.

Good catch, I agree this should be fixed.

jcfilben:
Waiting on research from https://github.com/grommet/hpe-design-system/issues/2664 before moving forward on this.

Issue #6117, opened 2022-05-06
Tabs - Enhance to allow responsive behavior and improved accessibility


taysea:
### Expected Behavior

Caller should have ability to opt-in to responsive Tabs behavior via a theme flag or prop (TBD). Desired behavior is that when there is not enough horizontal space to display all Tabs, then:
- Only the Tabs that can fit should be displayed
- Left/right arrow controls show appear on the sides of the Tabs (similar to pagination controls) allowing the user to move in steps of 3 through any remaining tabs when clicked
- Left arrow should be disabled when there are no more tabs to view to the left
- Right arrow should be disable when there are no more tabs to view to the right

Keyboard behavior:
- When keyboard tab reaches Tabs, focus is placed on active Tab.
- Left/right arrows should move between tabs one-by-one (follows best practice: https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-2/tabs.html)
- Pressing "Enter" should select and change the active Tab
- Tabbing again should leave Tabs and move onto next page content

### Actual Behavior

Not currently possible.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #6127, opened 2022-05-13
Improve keyboard accessibility of Tabs

jcfilben:
#### What does this PR do?
- When keyboard tab reaches Tabs, focus is placed on active Tab. (Other tabs are not in the tabIndex)
- The other tabs are reached with the left and right arrow keys on the keyboard.
- Pressing "Enter" should select and change the active Tab.
- The 'home' key ('fn' + 'left arrow' on mac)  should move focus to the first Tab.
- The 'end' key ('fn' + 'right arrow' on mac) should move focus to the last Tab.

Additional notes:
This PR adds 'onEnd' and 'onHome' to Keyboard. In addition to being used in this PR I believe there are other components that can utilize these to align with the recommended accessible keyboard support.

I changed Tab to use the ref approach [outlined here](https://gist.github.com/gaearon/1a018a023347fe1c2476073330cc5509) instead of using forward ref. It seemed like this was the better approach when a parent has multiple children each with their own ref.

Tested with existing Tab stories

https://github.com/grommet/grommet/issues/6117

Yes, this PR adds 'onEnd' and 'onHome' to Keyboard

Yes

Backwards compatible

### DIFF
--- a/src/js/components/Tab/Tab.js
+++ b/src/js/components/Tab/Tab.js
@@ -1,4 +1,4 @@
-import React, { forwardRef, useContext, useEffect, useState } from 'react';
+import React, { useContext, useEffect, useState } from 'react';
 import { ThemeContext } from 'styled-components';
 
 import { defaultProps } from '../../default-props';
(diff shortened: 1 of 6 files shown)


taysea:
I noticed this unexpected behavior when I do the following on the Controlled Tabs story:
1. Tab onto the Tabs (first tab is active)
2. Right arrow, right arrow (focus is on billing)
3. Press Tab (would expect it to move off the Tabs, but instead it moves back to the first tab)

src/js/components/Tab/Tab.js
```diff
@@ -12,194 +12,184 @@ import { normalizeColor } from '../../utils';
...
+        if (onMouseOver) onMouseOver();
+      }}
+      onBlur={() => {
+        setFocus(undefined);
+        if (onMouseOut) onMouseOut();
+      }}
+      // ensure focus outline is not covered by hover styling
+      // of adjacent tabs
+      style={focus && { zIndex: 1 }}
+      tabIndex={active ? 0 : -1}
```
taysea, line 178: Nice w roving tab index. Wondering if this will have any effect on people depending on how they've written e2e tests. That being said, I think we should consider the existing behavior a "bug" since it didn't follow existing accessibility best practices so we should be fine wrt backwards compatibility.

src/js/components/Tabs/Tabs.js
```diff
@@ -93,37 +98,62 @@ const Tabs = forwardRef(
...
-        ref={ref}
-        as={Box}
-        role="tablist"
-        flex={flex}
-        responsive={responsive}
-        {...rest}
-        background={theme.tabs.background}
-      >
-        <StyledTabsHeader
+      <Keyboard onLeft={onLeft} onRight={onRight} onHome={onHome} onEnd={onEnd}>
```
taysea, line 125: Nice.

src/js/components/Tab/Tab.js
taysea: Also, do we need to pass through `0` or will a button automatically be in the Tab focus? Not sure but could we do `tabIndex={!active && -1}` or something like that? Haven't tried locally -- was just trying to avoid adding more to the DOM if tab-index=0 is redundant

jcfilben:
> I noticed this unexpected behavior when I do the following on the Controlled Tabs story:
> 
> 1. Tab onto the Tabs (first tab is active)
> 2. Right arrow, right arrow (focus is on billing)
> 3. Press Tab (would expect it to move off the Tabs, but instead it moves back to the first tab)

Good catch this should be fixed in the latest push.

src/js/components/Tab/Tab.js
jcfilben: I could change this to `tabIndex={active ? undefined : -1}`.

`tabIndex={!active && -1}` or variations of this attempt to pass a boolean as the value.

jcfilben: > Nice w roving tab index. Wondering if this will have any effect on people depending on how they've written e2e tests. That being said, I think we should consider the existing behavior a "bug" since it didn't follow existing accessibility best practices so we should be fine wrt backwards compatibility.

Yeah this change could definitely cause some failing tests since the keyboard behavior is changing. I agree that it is a fix, just need to make it clear in the release notes that the keyboard behavior of Tabs is changing

taysea (approved):
I am fine passing tabIndex 0 through, I think it aids readability in the code. So good leaving as is unless others feel like we should sift it out as undefined.

amandadupell (approved):
tested with VO and keyboard experience; looks good to me! aligns with expected behavior for tabs

ericsoderberghp:
Is there any reason we aren't just letting the tab key traverse across the tabs directly? I ask because:
- we expect there to be a limited number of tabs, so not onerous to go through them
- Airbnb, Apple, and GitHub all work this way

jcfilben:
> Is there any reason we aren't just letting the tab key traverse across the tabs directly? I ask because:
>
> we expect there to be a limited number of tabs, so not onerous to go through them
> Airbnb, Apple, and GitHub all work this way

I changed the Tab keyboard behavior to align with the web accessibility initiative's recommendation for how keyboard behavior for Tabs should be handled. More information here: https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.

I agree that it wouldn't be too cumbersome for the user to tab through the Tabs, but I just wanted to align with the wai recommendation for keyboard accessibility.

Issue #6056, opened 2022-04-07
FileInput not updating when file removed


JackGoldsworth:
### Expected Behavior

When removing a file from the FileInput list, event.target.files should not be undefined, and should instead be the updated file list.

### Actual Behavior

event.target.files is undefined when an onChange event occurs during the removing of a file.

### URL, screen shot, or Codepen exhibiting the issue

![image](https://user-images.githubusercontent.com/38666839/162278945-5e1df6d6-afe2-4a5a-a917-73ec332fe7f6.png)

https://codesandbox.io/s/grommet-v2-template-forked-kgdlcg?file=/index.js

### Steps to Reproduce

1. Create a file input with multiple files.
2. In the onChange function, log the output of event.target.files.
3. Add files and you'll see that the files are logged.
4. Remove files and you'll see undefined.

### Your Environment

I don't believe there are any relevant environment details that need to be shared.

- Grommet version: 2.22.0
- Browser Name and version: Chrome Version 99.0.4844.84
- Operating System and version (desktop or mobile): Windows 10

PR #6136, opened 2022-05-18
Enhance FileInput to Update event.target.files on Remove

amandadupell:
#### What does this PR do?

Allows for a user to access `event.target.files` in the `onChange` function when removing a file. Previously, this was `undefined` because we were passing in the `event` from the remove button's `onClick`, not the input's `onChange` (`onClick` does not have a `files` value, only the FileInput's `onChange` does). This PR aligns the `onChange` function so that the `event.target.files` is accurately adjusted on removal and allows a user to access the FileList.

FileInput.js

Tested with Storybook. Updated the FileInput/Multiple story to `console.log(event.target.files)`.

Storybook

Included link to a resource I looked into. Also, this method of using `dispatchEvent` can be seen in other Input components (see MaskedInput, Select, RangeInput).

#6056 

Screen Shot 2022-05-18 at 4 07 46 PM

Yes, need to outline that `event.target.files` will correspond to the most recent event/action.

Yes.

Compatible.


jcfilben (commented):
If I add two files at the same time and then remove one file, `event.target.files` is a FileList containing both files. I was expecting there to only be one file in the FileList.

amandadupell:
> If I add two files at the same time and then remove one file, `event.target.files` is a FileList containing both files. I was expecting there to only be one file in the FileList.

This is what I'm seeing when I do that:
https://user-images.githubusercontent.com/38971752/169171171-50bd40f1-bc02-4444-9eb9-e161015ba86d.mov

amandadupell:
after talking with jessica, we discussed the difference with `event.target.files` and our `files` array we allow users to access. with this PR, `event.target.files` is tied to the most recent event as it relates to the input component. 

for example, each time a user goes to add files to the input, `event.target.files` will reflect the most recent event/action. removing will then correlate to this new event as well. this will differ from just accessing the `files` array.

jcfilben:
> after talking with jessica, we discussed the difference with event.target.files and our files array we allow users to access. with this PR, event.target.files is tied to the most recent event as it relates to the input component.
> 
> for example, each time a user goes to add files to the input, event.target.files will reflect the most recent event/action. removing will then correlate to this new event as well. this will differ from just accessing the files array.

I like this direction, it makes sense to me that `event.target.files` is only showing files related to the event, since we already have access the full list of files through `files`.

My only uncertainty is when we remove files whether `event.target.files` should reflect the updated file list (this is currently how the PR is handling it) or should it contain the file or files that were removed since that is the file or files tied to the event action?

@taysea, @ericsoderberghp any opinions on this?

JackGoldsworth:
@jcfilben As the original reporter of this issue, I expected the event.target.files list to contain the updated file list. So if you have a list of 5 files and remove one of them, you would receive a list of the 4 remaining files.

JackGoldsworth:
@amandadupell Fair enough. This brings up another issue however, which is the fact that you can't access { files } in typescript (Unless I'm doing something wrong). Should I create another issue for this?

src/js/components/FileInput/FileInput.js
```diff
@@ -239,8 +238,32 @@ const FileInput = forwardRef(
...
+      }
+
+      if (onChange) {
+        // We are using dispatchEvent to align the onChange event from
+        // the input, instead of the onClick event from the remove button.
+        const nativeInputValueSetter = Object.getOwnPropertyDescriptor(
+          window.HTMLInputElement.prototype,
+          'files',
+        ).set;
+        nativeInputValueSetter.call(inputRef.current, dt.files);
```
ericsoderberghp, line 262: Shouldn't we be doing this irregardless of whether `onChange` is provided? That way a test program could introspect the DOM to get the current value.

src/js/components/FileInput/FileInput.js
```diff
@@ -239,8 +238,32 @@ const FileInput = forwardRef(
...
-      if (nextFiles.length === 0) inputRef.current.value = '';
+
+      // Need to have a way to track the files other than an array
+      // since inputRef.current.files is a read-only FileList
+      // https://stackoverflow.com/a/64019766
+      /* eslint-disable no-undef */
+      const dt = new DataTransfer();
+      const curFiles = inputRef.current.files;
+      if (index === 'all' || nextFiles.length === 0)
+        inputRef.current.value = '';
```
ericsoderberghp, line 249: Why do we change the `value` in this case but not in the case where there are files?

ericsoderberghp (commented):
Overall, I like the direction this is heading. Just had a few small comments.

src/js/components/FileInput/FileInput.js
amandadupell: i can move this out of the if statement based on `onChange`

src/js/components/FileInput/FileInput.js
amandadupell: this resets `event.target.files` so we're only doing this in the case where a user wants to remove all of the files or there aren't any files anymore

amandadupell:
> @amandadupell Fair enough. This brings up another issue however, which is the fact that you can't access { files } in typescript (Unless I'm doing something wrong). Should I create another issue for this?

Thank you for pointing this out! Please create a new issue covering this and include a codesandbox that reproduces the error you're seeing in Typescript. Someone will pick this up from out backlog moving forward.

JackGoldsworth:
@amandadupell Done, the issue can be found below if anyone is immediately interested.
#6143

amandadupell:
@ericsoderberghp i spoke with jessica yesterday about aligning this PR with the native html `<input type="file">` behavior. in working on this and researching a bit more, there isn't any defined behavior for removing files anf how `event.target.files` should be updated. ultimately, it's left up to the developer to decide on how to best remove files one by one. the only "defined" behavior is when a user clears all of the files in the input, in which `event.target.files` should be set to an empty string. this is currently how we handle removing all of our files or reaching an empty array of files. 

https://stackoverflow.com/questions/1703228/how-can-i-clear-an-html-file-input-with-javascript
https://stackoverflow.com/questions/3144419/how-do-i-remove-a-file-from-the-filelist

considering this, i would be interested if this PR approach is a viable solution for how we decide to handle removing a file and how `event.target.files` shows this. since there is no defined behavior for removing a file for the input html element, i think we would just need to clearly define how we decide to handle this so users understand what `event.target.files` reflects.

jcfilben (approved):
This looks good to me. I think since there isn't a native way to remove a single file, showing the remaining file list seems like a reasonable behavior for `event.target.files` when we remove a file.

ericsoderberghp (approved):

PR #6137, opened 2022-05-18
Responsive Tabs

jcfilben:
#### What does this PR do?
Changes the responsive behavior for Tabs. Previously tabs would wrap if there is not enough room. This PR changes the behavior so if there isn't enough room to show the tabs, only the tabs that fit will be displayed and left and right arrows will be displayed on either side of the Tabs header and can be used to view additional tabs. A number can be passed to the `step` prop to define the number of tabs the left and right arrows will move by. The default step is 1.

Notes:

- I reduced the jest coverage threshold. Testing the responsive tabs is better suited for an end to end environment since it is dependent on screen size. I made note of this in [our end-to-end tracking issue](https://github.com/grommet/hpe-design-system/issues/2344).
 
Tested with Tabs/Responsive story

Closes https://github.com/grommet/grommet/issues/6117

Yes, need to add `step` prop to docs

Yes

Backwards compatible


ericsoderberghp:
> Adds the `scroll` prop to Tabs. If `scroll={true}` and there isn't enough room on the screen to show all the tabs, only that Tabs that fit will show and left are right arrows will be displayed on either side of the Tabs header and can be used to scroll left or right to see additional tabs. An interval can be passed to scroll like so: `scroll={{ interval: 2 }}` to define the number of tabs the left and right arrows will move by.

Some thoughts on the API surface ...
- From a language standpoint, I would expect `scroll={true}` to mean that we would show a scrollbar to allow scrolling through the tabs.
- If our current behavior is to wrap when we run out of room for the tabs, I'm wondering if we should consider this a bug and just have the new behavior be the default. I'm wondering if the current wrapping behavior would ever be desirable?
- From a language standpoint, I would expect `interval` to be related to "time". I wonder if "step" would be a better term.

taysea:
This is coming together really nicely.

A couple of behavioral/stylistic comments.
1. The focus indicator is getting cut off:
Screen Shot 2022-05-20 at 1 08 36 PM

2. When I tab to the "right" control and then it becomes disabled, the focus is lost. I might expect it gets placed onto the last tab potentially?

https://user-images.githubusercontent.com/12522275/169603794-bd00df00-4c7d-46a4-b562-edfc63c9fc7e.mov

src/js/components/Tabs/Tabs.js
```diff
@@ -93,37 +180,145 @@ const Tabs = forwardRef(
...
         >
-          {activeContent}
-        </StyledTabPanel>
-      </StyledTabs>
+          <Box direction={overflow && scroll ? 'row' : 'column'}>
+            {overflow && scroll && (
+              <Button
+                a11yTitle="Previous Tab"
+                icon={<Previous />}
+                disabled={disableLeftArrow}
```
taysea, line 222: Could we avoid the state variables for this and update arrow state and just do:
```suggestion
                disabled={isVisible(tabRefs[0].current)}
```

taysea: same for the "Next Tab" button

src/js/components/Tabs/Tabs.js
```diff
@@ -93,37 +180,145 @@ const Tabs = forwardRef(
...
+                      }
+                      scrolledToIndex = i;
+                      tabRefs[i].current.scrollIntoView({ behavior: 'smooth' });
+                      break;
+                    }
+                  }
+
+                  setDisableRightArrow(false);
+                  if (scrolledToIndex === 0) {
+                    // wait for scroll animation to finish
```
taysea, line 244: Do we need to let the animation finish? I think I expected it to become disabled more immediately

src/js/components/Tabs/Tabs.js
```diff
@@ -93,37 +180,145 @@ const Tabs = forwardRef(
...
+                disabled={disableLeftArrow}
+                onClick={() => {
+                  let scrolledToIndex;
+                  for (let i = 0; i < tabRefs.length - 1; i += 1) {
+                    if (
+                      !isVisible(tabRefs[i].current) &&
+                      isVisible(tabRefs[i + 1].current)
+                    ) {
+                      if (scroll && scroll.interval) {
+                        i =
```
taysea, line 231: Could we just do:
`i = Math.max(i - (scroll.interval - 1), 0)`

src/js/components/Tabs/Tabs.js
```diff
@@ -93,37 +180,145 @@ const Tabs = forwardRef(
...
+          {...rest}
+          background={theme.tabs.background}
         >
-          {activeContent}
-        </StyledTabPanel>
-      </StyledTabs>
+          <Box direction={overflow && scroll ? 'row' : 'column'}>
+            {overflow && scroll && (
+              <Button
+                a11yTitle="Previous Tab"
```
taysea, line 220: I'm wondering if these buttons should even be in the tab flow for a keyboard user because for example:
1. if the active tab is Tab 1
2. then I tab to the "Next" button and advance so I'm seeing Tab 10 (for example)
3. I want to select that, I have to shift+tab to back to the tabs
4. Tab 1 is active so now I have to use arrow keys to move to Tab 10

Propose that we either mimic the behavior shown in this [Angular responsive tabs example](https://material.angularjs.org/latest/demo/tabs) OR make it so "shift+tab" from the "Next" button places focus on the farthest right visible tab and "tab" from the "Previous button takes you to the farthest left visible tab?

Demo of current behavior:
https://user-images.githubusercontent.com/12522275/169605670-3fe98fcd-40da-473b-867e-794ebf8cff1a.mov

src/js/components/Tabs/Tabs.js
```diff
@@ -39,6 +57,74 @@ const Tabs = forwardRef(
...
+          updateArrowState();
+        };
+
+        if (
+          overflow === undefined ||
+          disableLeftArrow === undefined ||
+          disableRightArrow === undefined
+        ) {
+          onResize();
+        }
```
taysea, line 111: When I refresh the page, the responsive controls aren't visible until I resize the page. Do we need to always call `onResize` on the initial render (outside of this condition)

https://user-images.githubusercontent.com/12522275/169605873-3f400d21-2bfc-4e49-9380-0094cf8f3cb8.mov

?

src/js/components/Tabs/Tabs.js
```diff
@@ -39,6 +57,74 @@ const Tabs = forwardRef(
...
     /* eslint-enable no-param-reassign */
 
+    const tabRefs = React.Children.map(children, () => React.createRef());
+
+    // check if tab is in view
+    function isVisible(element) {
+      const tabRect = element?.getBoundingClientRect();
+      const headerRect = targetRef.current?.getBoundingClientRect();
+      return (
+        tabRect.left >= headerRect.left - 1 &&
```
taysea, line 67: Why do we need the -1 and +1?

src/js/components/Tabs/Tabs.js
```diff
@@ -27,8 +40,13 @@ const Tabs = forwardRef(
     const { format } = useContext(MessageContext);
     const { activeIndex: propsActiveIndex, onActive } = rest;
     const [activeIndex, setActiveIndex] = useState(rest.activeIndex || 0);
+    const [focusIndex, setFocusIndex] = useState(activeIndex);
     const [activeContent, setActiveContent] = useState();
     const [activeTitle, setActiveTitle] = useState();
+    const [disableLeftArrow, setDisableLeftArrow] = useState();
+    const [disableRightArrow, setDisableRightArrow] = useState();
+    const [overflow, setOverflow] = useState(undefined);
+    const targetRef = useRef();
```
taysea, line 49: Can we add a comment of why this is called `targetRef`?

src/js/components/Tabs/Tabs.js
```diff
@@ -39,6 +57,74 @@ const Tabs = forwardRef(
     delete rest.onActive;
     /* eslint-enable no-param-reassign */
 
+    const tabRefs = React.Children.map(children, () => React.createRef());
+
+    // check if tab is in view
+    function isVisible(element) {
```
taysea, line 63:
```suggestion
    const isVisible = (element) => {
```
Feel like this is more aligned with our conventions

src/js/components/Tabs/Tabs.js
```diff
@@ -93,37 +180,145 @@ const Tabs = forwardRef(
...
+
+    const onRight = () => {
+      if (focusIndex < tabRefs.length - 1) {
+        tabRefs[focusIndex + 1].current.focus();
+        setFocusIndex(focusIndex + 1);
+      }
+    };
+
+    const onHome = () => {
+      if (focusIndex !== 0 && tabRefs.length > 0) tabRefs[0].current.focus();
```
taysea, line 198: Do we need to `setFocusIndex` here too?

src/js/components/Tabs/Tabs.js
```diff
@@ -93,37 +180,145 @@ const Tabs = forwardRef(
...
+    };
+
+    const onEnd = () => {
+      if (focusIndex !== 0 && tabRefs.length > 0)
+        tabRefs[tabRefs.length - 1].current.focus();
```
taysea, line 203: `setFocusIndex` here too?

src/js/components/Tabs/Tabs.js
jcfilben: `disabled={isVisible(tabRefs[0].current)}` will only check once if the tab is visible or not and then never update this value. I used a state variable so that disabled would update as we are moving through the tabs.

src/js/components/Tabs/Tabs.js
jcfilben: If the button becomes disabled before the animation finishes, the animation will stop where it is and never continue. I'll try and mess around with this more though and see if there is a way around this because I agree, ideally the button could be disabled immediately and the animation would continue.

src/js/components/Tabs/Tabs.js
jcfilben: Agreed, I'll make this change

src/js/components/Tabs/Tabs.js
jcfilben: Yeah I like this! I will update to use this instead

src/js/components/Tabs/Tabs.js
jcfilben: I'm supportive of removing the left and right arrows from the tabIndex. It feels like they add an extra element that doesn't provide much utility for keyboard only users.

src/js/components/Tabs/Tabs.js
jcfilben: Hmm I haven't been able to reproduce this behavior. Is it happening every time you are refreshing the page or is it only happening sometimes?

src/js/components/Tabs/Tabs.js
jcfilben: I found that I needed to add the -1 and +1 so that there was a little leniency when calculating if an element was in view. Without the -1 and +1 a tab could be fully in view but `isVisible` would return false.

src/js/components/Tabs/Tabs.js
jcfilben: I'm thinking of changing the name to `headerRef` to make it more clear that this ref is for the tabs header container

src/js/components/Tabs/Tabs.js
jcfilben: Oops, yes that should be in there

src/js/components/Tabs/Tabs.js
jcfilben: Fixed

src/js/components/Tabs/Tabs.js
jcfilben: Fixed

src/js/components/Tabs/Tabs.js
jcfilben: Fixed

src/js/components/Tabs/Tabs.js
jcfilben: Changed name to `headerRef` in latest push

src/js/components/Tabs/Tabs.js
jcfilben: Added a comment in the code about this

src/js/components/Tabs/Tabs.js
jcfilben: Set `tabIndex={-1}` for the left and right arrow in the latest commit

src/js/components/Tabs/Tabs.js
jcfilben: Fixed

src/js/components/Tabs/Tabs.js
jcfilben: I adjusted this code so that the button gets disabled as soon as the animation finishes instead of waiting half a second before being disabled. But I couldn't find a way around having to wait for the animation to finish.

jcfilben:
> 1. The focus indicator is getting cut off:
> 
> Screen Shot 2022-05-20 at 1 08 36 PM

Unfortunately I couldn't find a good way around this. I am setting `overflow='hidden'` on the tabs header container which allows for the tabs that don't fit on the screen to be hidden, but it has the unfortunate side effect of cutting off the focus indicator.

src/js/components/Tabs/Tabs.js
taysea: Not sure if this would achieve the same thing but what if instead of `>=`, we just do `>`. So it'd be:

`tabRect.left > headerRect.left`?

Or does that cause the same issue.

src/js/components/Tabs/Tabs.js
```diff
@@ -93,37 +185,150 @@ const Tabs = forwardRef(
...
+                    if (
+                      !isVisible(tabRefs[i].current) &&
+                      isVisible(tabRefs[i - 1].current)
+                    ) {
+                      if (scroll && scroll.step) {
+                        i = Math.min(i + (scroll.step - 1), tabRefs.length - 1);
+                      }
+                      scrolledToIndex = i;
+                      tabRefs[i].current.scrollIntoView({ behavior: 'smooth' });
+                      break;
```
taysea, line 301: Is there any way we can implement this logic without using `break` -- I think break is generally discouraged bc it creates multiple ways a loop can be exited

src/js/components/Tabs/Tabs.js
```diff
@@ -93,37 +185,150 @@ const Tabs = forwardRef(
...
+                    }
+                  }
+
+                  setDisableLeftArrow(false);
```
taysea, line 305: Can we be sure we want to set it to `false`, would it be more reliable to `updateArrowStates`?

ericsoderberghp:
> > If our current behavior is to wrap when we run out of room for the tabs, I'm wondering if we should consider this a bug and just have the new behavior be the default. I'm wondering if the current wrapping behavior would ever be desirable?
> 
> I can't think of a case where the wrapping would be desirable. My only hesitation with making this the default behavior is that we have a storybook story [Tabs/Responsive](https://storybook.grommet.io/?path=/story/controls-tabs-responsive--responsive) that is advertising that this is how we have been handling responsive tabs. So it feels like something that we have been considering a feature up to this point.

I'm still inclined to make the new changes the default behavior and not support wrapping. I know we have a story showing wrapping, but the new behavior is a better design and I don't think we are removing the ability to use the component.

src/js/components/Tabs/Tabs.js
jcfilben: unfortunately `tabRect.left > headerRect.left` does cause the same issue

jcfilben:
> I'm still inclined to make the new changes the default behavior and not support wrapping. I know we have a story showing wrapping, but the new behavior is a better design and I don't think we are removing the ability to use the component.

Ok that sounds fair, I will move forward with this direction

ericsoderberghp:
Also, the previous/next Button controls aren't working for me on the latest Safari.

ericsoderberghp:
I'm wondering whether clipping tabs is the right approach. In this screenshot, the number of the last tab is clipped in such a way that it looks like the tab label is "Tab":
Screen Shot 2022-05-23 at 2 28 51 PM

If the tab label was "old trees", you might think the tab was called "old". I'm wondering if we should never clip tab labels and just omit rendering any tabs that we can't fit, relying on the previous/next controls to indicate when there is more.

jcfilben:
> If the tab label was "old trees", you might think the tab was called "old". I'm wondering if we should never clip tab labels and just omit rendering any tabs that we can't fit, relying on the previous/next controls to indicate when there is more.

This was something Beza and I discussed, the reason for showing the clipped tab was to avoid having extra whitespace in the header or having the tab header change size. Willing to go back and reconsider this though, especially since not clipping could solve the cut off focusIndicator issue

src/js/components/Tabs/Tabs.js
jcfilben: That makes sense, I will go back and update this to not use `break`

src/js/components/Tabs/Tabs.js
jcfilben: I think it is safe to set this to false here because we know that the arrows will only show up if not all the tabs can fit on the screen. If we are able to press the right arrow we know we will move to the right by some amount and if we move to the right by any amount the left arrow should be enabled. The reason I didn't use `updateArrowStates` here is because we would need to wait for the animation to finish and I didn't want there to be a delay if we could avoid it.

jcfilben:
Just chatted with Beza and Kenny, the consensus was to move forward with the cut off tabs. Some of the points brought up in our discussion:

- Having cut off tabs reinforces to the user that there is more to see
- If a tab title is cut off and a user clicks on that tab, the tab will scroll so that the whole tab title is visible, which will help ease possible confusion a user may have over cut off tabs.
- We are encouraging short names for tabs, so it would be relatively uncommon for a tab to be cut off in such a way that it causes confusion to the user
- If we only show tabs that can fit there will be extra whitespace between the tab titles and the right arrow. This will also create an extra step for keyboard users when navigating and force them to use the left and right buttons, instead of being able to reach all the tabs with left/right arrows on the keyboard.

ericsoderberghp:
OK, thanks for verifying

ericsoderberghp:
Now, we have to figure out how to address the focus clipping issue ...

src/js/components/Tabs/Tabs.js
```diff
@@ -19,6 +30,7 @@ const Tabs = forwardRef(
       justify = 'center',
       messages,
       responsive = true,
+      step = 1,
```
taysea, line 33: What are we recommending the step be for HPE Design System? Would it make sense to align the default value to that as opposed to requiring teams to add the prop each time?

ericsoderberghp: Or, set the default in the theme?

src/js/components/Tabs/Tabs.js
```diff
@@ -103,19 +234,154 @@ const Tabs = forwardRef(
...
+                  // wait for scroll animation to finish
+                  const checkVisible = setInterval(() => {
+                    if (isVisible(tabRefs[tabRefs.length - 1].current)) {
+                      setDisableRightArrow(true);
+                      clearInterval(checkVisible);
+                    }
+                  }, 100);
+                  setTimeout(() => {
+                    clearInterval(checkVisible);
+                  }, 500);
```
taysea, line 378: can you add a comment about how these timeout values were determined?

src/js/components/Tabs/Tabs.js
```diff
@@ -103,19 +234,154 @@ const Tabs = forwardRef(
...
-          alignSelf={alignControls}
-          flex={false}
-          wrap
-          background={theme.tabs.header.background}
-          gap={theme.tabs.gap}
-          {...tabsHeaderStyles}
-        >
-          {tabs}
-        </StyledTabsHeader>
+        <Box direction={overflow ? 'row' : 'column'} {...tabsHeaderStyles}>
```
taysea, line 237: The height once the overflow controls appear is `48px` while the default view when there's enough space for all the Tabs is `38px`. This causes a jump in the UI. Is it possible to eliminate the 10px shift?

src/js/components/Tabs/Tabs.js
```diff
@@ -93,37 +185,150 @@ const Tabs = forwardRef(
...
+                }}
+              />
+            )}
+            <StyledTabsHeader
+              ref={headerRef}
+              as={Box}
+              direction="row"
+              justify={justify}
+              alignSelf={alignControls}
+              flex={!!scroll}
```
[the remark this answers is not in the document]
jcfilben: Looks like you were looking at an old version of the file. Currently this is `flex={flex}`. We have a `flex` prop on Tabs so I think we need to leave `flex={flex}` instead of having flex be false or undefined.

src/js/components/Tab/Tab.js
```diff
@@ -12,194 +12,206 @@ import { normalizeColor } from '../../utils';
 import { StyledTab } from './StyledTab';
 import { TabPropTypes } from './propTypes';
 
-const Tab = forwardRef(
```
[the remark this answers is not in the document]
jcfilben: I changed Tab to use the ref approach [outlined here](https://gist.github.com/gaearon/1a018a023347fe1c2476073330cc5509) instead of using forward ref. It seemed like this was the better approach when a parent has multiple children each with their own ref. So instead of using forwardRef, the refs are passed through TabsContext.

src/js/components/Tabs/Tabs.js
jcfilben: Are you testing this with the responsive story or with different code? I am getting a height of 50px both when there is enough space for all the tabs and when the tabs overflow and the arrows show up. I am also not seeing a jump in the UI.

jcfilben: Oh okay so I was testing with the HPE theme, looks like the height shift is happening in the grommet theme. The reason for this is the left and right arrow buttons have a height of 48px which causes the height of the container to increase when we show the left and right arrows. I can see if I can adjust the left and right buttons.

taysea: Perfect yes I was testing on the Grommet theme

src/js/components/Tabs/Tabs.js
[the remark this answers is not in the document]
jcfilben: Oops I was looking at the wrong section of code. So currently this is `flex={!!overflow}` we need flex to be true if the content is overflowing so that the tabs fit between the left and right arrow buttons.

src/js/components/Tab/Tab.js
```diff
@@ -12,194 +12,206 @@ import { normalizeColor } from '../../utils';
...
+      aria-expanded={active}
+      disabled={disabled}
+      {...rest}
+      onClick={onClickTab}
+      onMouseOver={onMouseOverTab}
+      onMouseOut={onMouseOutTab}
+      onFocus={() => {
+        setFocus(true);
+        setFocusIndex(index);
+        if (onMouseOver) onMouseOver();
```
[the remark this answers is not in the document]
jcfilben: Fixed in latest commit

src/js/components/Tab/Tab.js
```diff
@@ -12,194 +12,206 @@ import { normalizeColor } from '../../utils';
...
+      }}
+      onBlur={() => {
+        setFocus(undefined);
+        setFocusIndex(-1);
+        if (onMouseOut) onMouseOut();
+      }}
+      // ensure focus outline is not covered by hover styling
+      // of adjacent tabs
+      style={focus && { zIndex: 1 }}
```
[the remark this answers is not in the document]
jcfilben: Fixed in latest commit

src/js/components/Tabs/Tabs.js
jcfilben: Added comments for this in latest commit

src/js/components/Tabs/Tabs.js
jcfilben: I changed the button arrows to contain an icon wrapped with a box which allowed for more control over the padding around the icon. Let me know if this looks okay

taysea:
I'm unable to use the arrow keys to move between the tabs, instead I have to tab through each one. This doesn't meet the accessibility best practice.

jcfilben:
> I'm unable to use the arrow keys to move between the tabs, instead I have to tab through each one. This doesn't meet the accessibility best practice.

Eric and I chatted about this. The plan is to have the arrow key behavior as an opt-in option in the theme and have the default keyboard behavior still be tabbing through the tabs. Adding the arrow key behavior will be done in a separate pull request.

src/js/components/Tabs/Tabs.js
```diff
@@ -103,19 +238,183 @@ const Tabs = forwardRef(
...
+                  // the tab name is very long.
+                  const checkVisible = setInterval(() => {
+                    if (isVisible(tabRefs[0].current)) {
+                      setDisableLeftArrow(true);
+                      clearInterval(checkVisible);
+                    }
+                  }, 100);
+                  setTimeout(() => {
+                    clearInterval(checkVisible);
+                  }, 500);
```
taysea, line 305: Why does this interval differ from the left arrow?

src/js/themes/base.js
```diff
@@ -1646,6 +1646,11 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       panel: {
         // extend: undefined,
       },
+      step: {
+        small: 1,
```
taysea, line 1650: Why do we only step by 1 on a small screen?

jcfilben: If you are on a phone and the tabs move by a step of 3, with longer tab names you could potentially miss seeing one of the tabs. So for small screens where fewer tabs can fit, it felt better to only move by one tab at a time

src/js/components/Tab/Tab.js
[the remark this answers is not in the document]
ericsoderberghp: I'm fine with the `ref` via context approach. But, I think we still need the `forwardRef` to allow callers to connect a ref to individual Tab components.

src/js/components/Tabs/Tabs.js
```diff
@@ -103,19 +238,179 @@ const Tabs = forwardRef(
...
-        </StyledTabsHeader>
+        <Box direction={overflow ? 'row' : 'column'} {...tabsHeaderStyles}>
+          {overflow && (
+            <Button
+              a11yTitle="Previous Tab"
+              disabled={disableLeftArrow}
+              // removed from tabIndex, button is redundant for keyboard users
+              tabIndex={-1}
+              onClick={() => {
+                let scrolledToIndex;
```
ericsoderberghp, line 249: There seems to be a certain amount of overlap between this code and `moveIntoView()`.  Would it be possible to combine them somehow?

It seems a little odd to me here that we're looping through all of the tabs here. Can we do something a simple as "if the left or right edge of the Tab we want to be visible isn't, scroll so it is by the amount that it isn't?

src/js/components/Tab/Tab.js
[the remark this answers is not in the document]
jcfilben: Fixed in latest commit.

src/js/components/Tabs/Tabs.js
jcfilben: Consolidated code in latest commit

ericsoderberghp:
In the current version, the previous and next buttons don't seem to be working for me. I can tab focus through. This is on Safari.

ericsoderberghp:
If I use the tab key to focus through the tabs, the last tab has the right side of the focus indicator hidden. If I go the other way, the first tab correctly shows the full focus indicator.

Screen Shot 2022-06-09 at 2 34 38 PM

jcfilben:
> In the current version, the previous and next buttons don't seem to be working for me. I can tab focus through. This is on Safari.

Hmm I just tested with Safari, Chrome, and Firefox. They arrow keys are working in all of them for me. The only difference is in Safari the 'smooth' scroll behavior isn't supported so the tab isn't sliding in or out like in the other browsers. I will push up changes soon based on the review comments and then if the arrow keys are still not working in Safari on your end I will investigate a little deeper why it is working differently on my end vs on your end.

jcfilben:
> I'm wondering if we should accommodate track pad interaction, since we are showing some tabs partially and mimicking scroll behavior.

This was something Beza and I discussed. Her reasons for not wanting track pad scrolling were that a user could scroll quickly and miss a tab. I could revisit this topic with her though

jcfilben:
> If I use the tab key to focus through the tabs, the last tab has the right side of the focus indicator hidden. If I go the other way, the first tab correctly shows the full focus indicator.

Fixed in latest commit

src/js/components/Tabs/Tabs.js
```diff
@@ -39,6 +57,159 @@ const Tabs = forwardRef(
     delete rest.onActive;
     /* eslint-enable no-param-reassign */
 
+    const tabRefs = React.Children.map(children, () => React.createRef());
```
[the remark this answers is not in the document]
jcfilben: Yes, fixed in latest commit

ericsoderberghp:
> > I'm wondering if we should accommodate track pad interaction, since we are showing some tabs partially and mimicking scroll behavior.
> 
> This was something Beza and I discussed. Her reasons for not wanting track pad scrolling were that a user could scroll quickly and miss a tab. I could revisit this topic with her though

Hmm... How is that different from any scroll situation, where the user can scroll quickly and skip over something? Perhaps the user knows they want the last tab and just want to get to it quickly, without having to step through each one to get there?

ericsoderberghp:
> > If I use the tab key to focus through the tabs, the last tab has the right side of the focus indicator hidden. If I go the other way, the first tab correctly shows the full focus indicator.
> 
> Fixed in latest commit

For me, Firefox and Safari still have this problem. On Edge, I don't see a problem with the right edge of Tab 12 but I do see it with the left edge of Tab 1 when going backwards.

ericsoderberghp:
> > In the current version, the previous and next buttons don't seem to be working for me. I can tab focus through. This is on Safari.
> 
> Hmm I just tested with Safari, Chrome, and Firefox. They arrow keys are working in all of them for me. The only difference is in Safari the 'smooth' scroll behavior isn't supported so the tab isn't sliding in or out like in the other browsers. I will push up changes soon based on the review comments and then if the arrow keys are still not working in Safari on your end I will investigate a little deeper why it is working differently on my end vs on your end.

Edge and Firefox are working for me. Safari still isn't.

jcfilben:
> Hmm... How is that different from any scroll situation, where the user can scroll quickly and skip over something? Perhaps the user knows they want the last tab and just want to get to it quickly, without having to step through each one to get there?
> 
> I was noticing that when I had a small window size, it was a bit annoying to use the arrows to get back to the first tab after having clicked next through them all to review them.

I just chatted with Beza about this. She still feels that scrolling through the tabs is not a good experience. But she said we could allow the option to allow scrolling as a toggle in the theme and for the hpe theme set it so that we don't allow scrolling.

jcfilben:
An update on the Safari issue where the arrow keys don't scroll the select:

There is an issue in v15.5 of safari where if a container has `overflow: 'hidden'` and the scroll behavior is set to 'smooth', scrolling will fail without giving any warnings or errors. (See related issue: https://bugs.webkit.org/show_bug.cgi?id=238497).

A pull request has been merged on the webkit repository to fix this (https://github.com/WebKit/WebKit/pull/1387). This fix is not yet included in the latest version of Safari. The latest version of safari uses webkit `v7613.2.7.1.8` the fix is in webkit `v7614.1.16.1`.

Should we add a work around for this in the meantime? Or go ahead a move forward knowing it should be fixed on Safari's end soon?

jcfilben:
> I'm still seeing focus indicator clipping on Edge, Firefox, and Safari. Screen Shot 2022-06-13 at 2 00 48 PM

Looking into it... right now I am only seeing the clipping issue on Safari

ericsoderberghp:
I wonder if our timing calculations are too precise for particular environments and not adaptive enough? I will pull the branch and experiment with them a bit.

src/js/components/Tabs/Tabs.js
```diff
@@ -29,16 +42,226 @@ const Tabs = forwardRef(
...
+    useEffect(() => {
+      // if the active tab isn't visible scroll to it
+      if (
+        overflow &&
+        tabRefs &&
+        tabRefs[activeIndex].current &&
+        !isVisible(tabRefs[activeIndex].current)
+      )
+        scrollToIndex(activeIndex, true);
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
ericsoderberghp, line 207: Why do we need to disable this? More often than not, not having exhaustive dependencies has led to bugs. In the few cases where we are sure we don't want them all, we put a fairly descriptive comments explaining why.

jcfilben: My reasoning for disabling this is that we only want to scroll to the active index if it is set initially, or if it is changed. If we include all the dependencies every time we tried to scroll this would force us to scroll us back so that the activeIndex is in view.

src/js/components/Tab/Tab.js
```diff
@@ -175,23 +203,23 @@ const Tab = forwardRef(
...
-        // of adjacent tabs
-        style={focus && { zIndex: 1 }}
       >
         <StyledTab
           as={Box}
           disabled={disabled}
           plain={plain}
           {...withIconStyles}
           {...tabStyles}
+          style={{
```
ericsoderberghp, line 220: We should put this inside StyledTab itself, instead of using the `style` property here.

src/js/components/Tabs/Tabs.js
```diff
@@ -29,16 +42,226 @@ const Tabs = forwardRef(
...
     /* eslint-enable no-param-reassign */
 
+    const tabRefs = useMemo(
+      () => React.Children.map(children, () => React.createRef()),
+      [children],
+    );
+
+    // check if tab is in view
+    const isVisible = useCallback(
+      (element) => {
```
ericsoderberghp, line 76: I'm wondering if the code where we call this would be a bit simpler if we passed `index` here, instead of `element`. It would align more with `scrollToIndex` as well.

src/js/components/Tabs/Tabs.js
```diff
@@ -29,16 +42,226 @@ const Tabs = forwardRef(
...
+      },
+      [headerRef],
+    );
+
+    const updateArrowState = useCallback(() => {
+      setDisableLeftArrow(isVisible(tabRefs[0].current));
+      setDisableRightArrow(isVisible(tabRefs[tabRefs.length - 1].current));
+    }, [tabRefs, isVisible]);
+
+    const scrollToIndex = useCallback(
```
ericsoderberghp, line 100: Wondering if we should call this just `scrollTo`?

src/js/components/Tabs/Tabs.js
ericsoderberghp: An example of a potential issue is that `tabRefs` is re-built via `useMemo` if the `children` change. If we don't have `tabRefs` in the dependency list here, this effect will use the old one incorrectly.

src/js/components/Tabs/Tabs.js
```diff
@@ -29,16 +42,226 @@ const Tabs = forwardRef(
...
+      // if the focus index is changing make sure the previous and next arrows
+      // are in sync
+      if (
+        overflow &&
+        focusIndex !== -1 &&
+        !isVisible(tabRefs[focusIndex].current)
+      )
+        scrollToIndex(focusIndex, true);
+      else if (overflow && focusIndex !== -1) {
+        // If the browser scrolled the focused item into view and
```
ericsoderberghp, line 220: Do we still need this block? Since I would assume `scrollToIndex` and `isVisible` would already be taking care of this situation.

src/js/components/Tabs/Tabs.js
```diff
@@ -53,12 +276,15 @@ const Tabs = forwardRef(
         return {
           activeIndex,
           active: activeIndex === index,
+          index,
+          tabsContextRef: tabRefs[index],
```
ericsoderberghp, line 280: I think the key/name should just be `ref`. For example, we don't say `tabsContextIndex`. If the consumer needs to rename it, it can do that as needed when using the context value.

src/js/components/Tabs/Tabs.js
jcfilben: This block is used in case where the browser handles the scroll into view of focused items. In this case `scrollTo` won't be called because the tab will be visible so we need to handle scrolling a few pixels further to show the focusIndicator

src/js/components/Tabs/Tabs.js
jcfilben: Change scrollToIndex to scrollTo in latest commit

src/js/components/Tabs/Tabs.js
jcfilben: Change to pass index in latest commit

src/js/components/Tab/Tab.js
jcfilben: Fixed in latest commit

src/js/components/Tabs/Tabs.js
jcfilben: I went back and removed the `eslint-disable-next-line react-hooks/exhaustive-deps`, I realized that now that we are using `useMemo` for the tabRefs we no longer run into the issue of it trying to force us to scroll back to the activeIndex tab

src/js/components/Tabs/Tabs.js
ericsoderberghp: I would expect `isVisible()` to take care of that already, such that we would call `scrollTo` at line 206 if any scrolling might be needed. In other words, I would expect `isVisible` of true to mean we don't need to scroll and `isVisible` of false meaning we do need to scroll.

ericsoderberghp: Maybe we should rename `isVisible` to `needsScroll`?

taysea (approved):

Issue #6134, opened 2022-05-18
Notification Component is missing 'id' prop


utkalsb:
Expected Behavior: 
Id prop should be set for all notifications type

Actual Behavior: 
id prop is set only when notification type is toast. 
![image](https://user-images.githubusercontent.com/29633479/168935982-a25fb624-6291-4195-8917-1739309a8799.png)

halocline:
In addition to `id` mentioned above ☝️ , other HTML attributes and React props should be passed to Notification similar to how other Grommet components receive them via a spread prop.

PR #6139, opened 2022-05-20
Changed Notification to pass through additional properties

ericsoderberghp:
#### What does this PR do?

Changed Notification to pass through additional properties.

These changes were necessary for grommet-designer:
- don't crash on arbitrary values for `status`, which occurs when the user is editing the status
- pass through additional props, which is necessary for `onClick` and `style`.

Notification.js

added a unit test
built and tested in a local instance of grommet-designer

no

yes

backwards compatible


halocline:
Is it possible for this PR to resolve [this issue](https://github.com/grommet/grommet/issues/6134) as well?

src/js/components/Notification/Notification.js
```diff
@@ -173,6 +173,7 @@ const Notification = ({
       pad={undefined}
       direction="row"
       gap="small"
+      {...rest}
```
jcfilben, line 176: Adding this will address https://github.com/grommet/grommet/issues/6134
```suggestion
      id={toast ? undefined : id}
      {...rest}
```

jcfilben (approved):
Looks good!

Issue #6143, opened 2022-05-20
Can't access { files } in the onChange method for FileInput in TypeScript


JackGoldsworth:
### Expected Behavior

Like the example in the storybook, you should be able to access the files attribute from within the onChange method for the FileInput element.

### Actual Behavior

It will not allow you to access the files attribute because onChange complains about the function type definition.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-forked-wijew6

### Steps to Reproduce

1. Setup a FileInput element.
2. Have the onChange function only include the event as a parameter.
3. Add { files } to the parameters to the onChange function, and you will get a type definition error.

### Your Environment

Grommet version: 2.23.0
Browser Name and version: Firefox Version 100.0.2
Operating System and version (desktop or mobile): Windows 10

PR #6145, opened 2022-05-23
Fix onChange type for FileInput

jcfilben:
#### What does this PR do?
Fixes the typescript definition for onChange in FileInput

Tested with the following code in a typescript story:
```typescript
const onChange = (event, { files }) => {
  console.log(files);
);
```
and
```typescript
const onChange = ({ target: { files } }) => {
  console.log(files);
};
```

with `<FileInput onChange={onChange} />`

Test an `onChange` function with FileInput in a typescript file

Closes https://github.com/grommet/grommet/issues/6143

No
Maybe
Backwards compatible

### DIFF
--- a/src/js/components/FileInput/index.d.ts
+++ b/src/js/components/FileInput/index.d.ts
@@ -22,10 +22,11 @@ export interface FileInputProps {
   };
   multiple?: boolean | { aggregateThreshold?: number; max?: number };
   name?: string;
+  onChange?: (event?: any, { files }?: { files: any }) => void;
   renderFile?: (...args: any[]) => void;
 }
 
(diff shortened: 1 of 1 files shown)


src/js/components/FileInput/index.d.ts
```diff
@@ -22,10 +22,11 @@ export interface FileInputProps {
   };
   multiple?: boolean | { aggregateThreshold?: number; max?: number };
   name?: string;
+  onChange?: (event?: any, { files }?: { files: any }) => void;
```
britt6612, line 25: is there anything more specific than `any`

jcfilben: Good comment, just updated this to make the types more specific

ericsoderberghp (approved):

Issue #6133, opened 2022-05-17
Select search text is cleared (but doesn't call onSearch) after select when closeOnChange=false


kariwhitcomb:
### Expected Behavior

Using Select component with multiple and onSearch specified and closeOnChange=false.  User types into search field, then selects an entry.  Select drop should stay open (because closeOnChange=false) and search text should remain as it was.

### Actual Behavior

Search text is cleared.  onSearch isn't called, so the options are still filtered, but the text is gone.  So there is no way for user to clear the search other than to close and reopen the drop.

### URL, screen shot, or Codepen exhibiting the issue
Issue in seen in multi-select story book: https://storybook.grommet.io/?path=/story/input-select-search--search

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.23.0
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): Mac

PR #6147, opened 2022-05-23
Fixed Select to continue showing search term after an option is selected

jcfilben:
#### What does this PR do?
If we have a Select with `closeOnChange={false}` and `search` and a user searches then selects an option, the search term will clear but the list of options will still be filtered based on the search term.

This PR changes the behavior so that the search term is not cleared when an option is selected.

Tested with the Select/Search story

Closes https://github.com/grommet/grommet/issues/6133

No
Yes
Backward compatible

### DIFF
--- a/src/js/components/Select/Select.js
+++ b/src/js/components/Select/Select.js
@@ -245,7 +245,6 @@ const Select = forwardRef(
           }
           onChange(adjustedEvent);
         }
-        setSearch();
       },
       [closeOnChange, onChange, onRequestClose, setValue, triggerChangeEvent],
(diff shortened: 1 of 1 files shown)


ericsoderberghp (approved):
gotta love the size of this change :)

britt6612:
> #6133

think this was the easiest thing Ive ever reviewed 👍

britt6612:
> gotta love the size of this change :)

think this was the easiest thing Ive ever reviewed 👍

britt6612 (approved):

Issue #6148, opened 2022-05-24
PageHeader - Accept a prop to to allow PageHeader layout to switch from row oriented to column oriented at specified breakpoints


halocline:
Modify Grommet PageHeader to accept a prop controlling whether `actions` maintains its position to the right of the title, or is placed under the title & subtitle at narrow breakpoints (e.g. when primary action is present or not - see screenshots for reference) (Grommet).

**Expected**
A page header's "primary" action should always be prominent and visible since it is the primary objective for the user. The page header layout, specifically the placement of the `actions` region, should adjust so that all page header regions are visible on smaller width screens and the primary action maintains prominence.

**Actual**
By default, PageHeader keeps "actions" in the second column of the Grid at all responsive breakpoints. This is desired if I would like actions to collapse into an overflow menu. However, in cases where I have an action which is the primary purpose of the screen, I do not want this action to be placed within the overflow and hidden from the user. I also do not want the page title and primary action to compete for real-estate on the same row when on a small screen. Instead, I would like the actions to be placed below the page's title/subtitle.

PR #6157, opened 2022-05-27
PageHeader - Added responsive prop

halocline:
#### What does this PR do?

Closes https://github.com/grommet/grommet/issues/6148

Proposes a `responsive` prop which allows the PageHeader to switch from a row-oriented, two-column layout to a single column presenting all page header regions at specified responsive breakpoints. This is useful in cases where the page header contains a primary action, which as a primary page action should always be visible instead of collapsing into an overflow menu.

`responsive` is used across a variety of Grommet components as a boolean indicating that the component should take on special presentation for smaller view ports. Typically the presentation is adjusted by scaling ([Box](https://v2.grommet.io/box#responsive), [Grid](https://v2.grommet.io/grid#responsive), [Heading](https://v2.grommet.io/heading#responsive), [Paragraph](https://v2.grommet.io/paragraph#responsive)) or by change in layout ([Drop](https://v2.grommet.io/drop#responsive), [Layer](https://v2.grommet.io/layer#responsive)).

If needed in the future, I could see `responsive` props being extended to accept an object which allows more precise control by the caller to affect specific responsive attributes to apply/not apply (e.g. layout direction, margin, pad, gap, font-size, etc.).

PageHeader.js
[Storybook/PageHeader/Responsive]

Storybook

Storybook

No Jest tests added

https://github.com/grommet/grommet/issues/6148

**Responsive = false**
Screen Shot 2022-05-27 at 9 58 31 AM

**Responsive = true**
Screen Shot 2022-05-27 at 9 58 37 AM

Yes.

No. Should only announce NEW PageHeader

Backwards compatible

### DIFF
--- a/src/js/components/PageHeader/PageHeader.js
+++ b/src/js/components/PageHeader/PageHeader.js
@@ -9,12 +9,37 @@ import { Paragraph } from '../Paragraph';
 import { ResponsiveContext } from '../../contexts/ResponsiveContext';
 
 const PageHeader = forwardRef(
-  ({ actions, children, gridProps, parent, subtitle, title, ...rest }, ref) => {
+  (
+    {
+      actions,
+      children,
+      gridProps: gridPropsProp,
+      parent,
+      responsive,
+      subtitle,
+      title,
+      ...rest
+    },
+    ref,
+  ) => {
     const theme = useContext(ThemeContext);
     const breakpoint = useContext(ResponsiveContext);
 
-    const { areas, columns, gap, rows } =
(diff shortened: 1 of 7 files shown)


taysea (approved):
LGTM!

ericsoderberghp:
I completely agree with wanting to provide the behaviors. I'm less sure about `responsive` as the property name to drive it. But, looking at how we've used that property elsewhere, it's not as though we're super consistent with its meaning. It seems to basically imply "a reasonable alternate behavior in responsive situations". So, I'm OK with it here.

ericsoderberghp (approved):

Issue #6153, opened 2022-05-26
Allow to reverse icon in DateInput


marutzen:
### Expected Behavior

When pass inputProps with { reverse: false } should change icon to left.

### Actual Behavior

reverse props is not working, keeping the icon in the right side.

### URL, screen shot, or Codepen exhibiting the issue

![image](https://user-images.githubusercontent.com/43305032/170547217-b15e3080-3eba-4035-afab-6afa80370e57.png)

### Steps to Reproduce

No steps needed.

### Your Environment

- Grommet version: 2.23.0
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): desktop

joaoGabriel55:
Hello @marutzen I think that this prop is not available for this component. Access this doc link:

https://v2.grommet.io/dateinput

But I think that could be a good addition. What do you think @jcfilben?

britt6612:
So this `reverse` prop is part of the InputProps which is `Any properties to pass on to the underlying MaskedInput when there is a format.` so I think this is a bug it should be 
`inputProps={{ reverse: true }}` but this does not work

joaoGabriel55:
> So this `reverse` prop is part of the InputProps which is `Any properties to pass on to the underlying MaskedInput when there is a format.` so I think this is a bug it should be `inputProps={{ reverse: true }}` but this does not work

I did not know this. Good to know, thanks.

britt6612:
no problem! There is some missing behavior with this though

joaoGabriel55:
I can take a look at that

britt6612:
sounds great @joaoGabriel55 thank you

jcfilben:
Closing, #6160 has been merged

PR #6160, opened 2022-05-27
Allowing to reverse icon in DateInput

joaoGabriel55:
#### What does this PR do?

Fixed issue https://github.com/grommet/grommet/issues/6153 which consists of allowing to reverse icon in DateInput.

Just access:

http://localhost:9001/?path=/story/input-dateinput-format--format

Unit test to Match Snapshot.

It is not necessary

No

No

#### Screenshots

image

Yes, to add the new DateInput prop, `reverse`

No

No


jcfilben (commented):
Just a few notes:
- We should support both passing the reverse prop through `inputProps` and having the `reverse` prop on DateInput. Ideally just using the `reverse` prop is preferred but we need to support passing reverse through `inputProps` for backwards compatibility since we supported this in the past. See https://github.com/grommet/grommet/pull/6160/files for a similar discussion about the icon prop on DateInput.

- When the icon is on the left side of DateInput the tab order is messed up. The focus moves first to the input and then the icon. It should move to the icon first and then the input since that is how it is visually ordered.

amandadupell (approved):
looks good to me! tab order goes from icon to input, following logical order. allows for `reverse` prop and passing in via `inputProps` for backwards compatibility

src/js/components/DateInput/stories/Format.js
```diff
@@ -12,9 +13,30 @@ export const Format = () => {
...
       </Box>
+      <Box width="medium" margin={{ vertical: 'small' }}>
+        <DateInput format="m/d/yy" value={value} reverse onChange={onChange} />
+      </Box>
+      <Box width="medium" margin={{ vertical: 'small' }}>
+        <DateInput
+          format="m/d/yy"
+          value={value}
+          icon={<ClockIcon />}
+          inputProps={{ reverse: true }}
```
jcfilben, line 27: I think we should encourage using the `reverse` prop over setting reverse through input props. Can we change the story examples to use `reverse`?

src/js/components/DateInput/DateInput.js
```diff
@@ -111,6 +112,13 @@ Use the icon prop instead.`,
       );
     }
 
+    const reverse = reverseProp || restOfInputProps.reverse;
+
+    const calendarDropdownAlign = {
+      top: 'bottom',
+      ...(reverse ? { right: 'right' } : { left: 'left' }),
+    };
```
jcfilben, line 120: I think the alignment of the calendar should stay the same even if we use the reverse prop. The reverse prop should only affect the icon.

jcfilben (commented):
Coming along nicely, just a few comments

src/js/components/DateInput/stories/Format.js
joaoGabriel55: Yes, that makes sense

src/js/components/DateInput/DateInput.js
joaoGabriel55: If I don't do this the Input will be this way:

image

Would that be the expected behavior?

joaoGabriel55: Very nice tip

ericsoderberghp (approved):

PR #6162, opened 2022-06-01
Enhance Menu to allow grouped items

taysea:
#### What does this PR do?

[Deploy Preview](https://deploy-preview-6162--sad-tereshkova-173d07.netlify.app/?path=/story/controls-menu-grouped--grouped)

This PR enhances `items` prop on Menu to allow for grouped items by passing an array of menu item arrays.

Looking for feedback on:
- [x] API surface
- [x] Theme structure
- [x] How separator is implemented as 'top' when index of group is not the first one

```
API

<Menu
   items={[
      [
         { label: 'View Details' },
         { label: 'Edit Permissions' },
      ],
      [
         { label: 'Delete' },
      ],
    ]}
/>
```

**Theme**
- Adds `theme.menu.group.container` --> any Box props for an individual group
- Adds `theme.menu.group.separator` --> Color and size of separator

Related to https://github.com/grommet/hpe-design-system/issues/2621

Yes.

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -288,6 +288,103 @@ const Menu = forwardRef((props, ref) => {
     </Box>
   );
 
+  const menuItem = (item, index) => {
+    // Determine whether the label is done as a child or
+    // as an option Button kind property.
+    const child = !theme.button.option ? (
+      <Box
+        align="start"
+        pad="small"
+        direction="row"
+        gap={item.gap}
+        justify={item.justify}
+      >
+        {item.reverse && item.label}
+        {item.icon}
+        {!item.reverse && item.label}
+      </Box>
+    ) : undefined;
+
+    // if we have a child, turn on plain, and hoverIndicator
+    return (
+      // eslint-disable-next-line react/no-array-index-key
+      <Box key={index} flex={false} role="none">
+        <Button
+          ref={(r) => {
+            buttonRefs.current[index] = r;
+          }}
+          role="menuitem"
+          onFocus={() => {
+            setActiveItemIndex(index);
+          }}
+          active={activeItemIndex === index}
+          focusIndicator={false}
+          plain={!child ? undefined : true}
+          align="start"
+          justify={item.justify}
+          kind={!child ? 'option' : undefined}
+          hoverIndicator={!child ? undefined : 'background'}
+          {...(!child
+            ? item
+            : {
+                ...item,
+                gap: undefined,
+                icon: undefined,
+                label: undefined,
+                reverse: undefined,
+              })}
+          onClick={(...args) => {
+            if (item.onClick) {
+              item.onClick(...args);
+            }
+            if (item.close !== false) {
+              onDropClose();
+            }
+          }}
+        >
+          {child}
+        </Button>
+      </Box>
+    );
+  };
+
+  let menuContent;
+  if (items.length && Array.isArray(items[0])) {
+    menuContent = items.map((group, i) => {
+      const groupHeading = group.find((item) => item.heading);
+
+      return (
+        <Box
+          {...theme.menu.group?.container}
+          border={
+            i > 0
+              ? { side: 'top', ...theme.menu.group?.container?.border }
+              : undefined
+          }
+        >
+          {groupHeading && (
+            <Box {...theme.menu.group?.heading?.container}>
+              <Text
+                id={groupHeading?.id}
+                // don't have screenreader announce this text element directly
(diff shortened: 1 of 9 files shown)


ericsoderberghp:
Thanks for publishing early for feedback! Can you add an index.d.ts? I find that's a bit easier to review for API surface changes than prop types, especially with object details rather than just `object`.

I'm good with the array of array of objects, where the objects are 1-1 Button props.

I've been debating whether the heading approach should be built into Menu vs. using DropButton with a custom Drop. There's a line somewhere where adding something that's used rarely but can be done with a more generic approach, like DropButton, makes sense.

If we do go the heading approach, I'm wondering if we should be more structural and do something like:
`items={[{ heading: string, items: [{}, ...]}, ...]}`

taysea:
> I've been debating whether the heading approach should be built into Menu vs. using DropButton with a custom Drop. There's a line somewhere where adding something that's used rarely but can be done with a more generic approach, like DropButton, makes sense.
> 

I feel like building the headings into Menu could be a valuable addition, especially from the standpoint that you get to leverage all the other menu theming benefits, keyboard/accessibility behavior, etc. with this. While I don't think headings are the majority case, I do think they're a valid case.

If we decide not to add headings in, we might want to include a detailed code example of how to achieve both the visuals and the accessibility behavior.

> If we do go the heading approach, I'm wondering if we should be more structural and do something like: `items={[{ heading: string, items: [{}, ...]}, ...]}`

This was also a structure Matt and I ideated on, could be work exploring again to leave items being 1-1 with Button props if that's desirable.

src/js/themes/base.js
```diff
@@ -1049,6 +1049,13 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         },
         // any drop props
       },
+      group: {
+        container: {
+          pad: {
```
ericsoderberghp, line 1054: Can we put the `border : { side: 'top' }` here?

But, I can see where this doesn't work so well as we don't want that on the first group. I'm wondering if we should structure this more as `group.separator: { color, size }` ?

src/js/components/Menu/Menu.js
```diff
@@ -288,6 +300,92 @@ const Menu = forwardRef((props, ref) => {
...
+  let menuContent;
+  if (itemCount && Array.isArray(items[0])) {
+    menuContent = items.map((group, groupIndex) => (
+      <Box
+        // eslint-disable-next-line react/no-array-index-key
+        key={groupIndex}
+        {...theme.menu.group?.container}
+        border={
+          groupIndex > 0
+            ? { side: 'top', ...theme.menu.group?.container?.border }
```
ericsoderberghp, line 372: See my comments on base.js theme related to this

src/js/components/Menu/Menu.js
```diff
@@ -288,6 +300,92 @@ const Menu = forwardRef((props, ref) => {
...
+            : undefined
+        }
+      >
+        {/* index needs to be its index in the entire menu as if
```
ericsoderberghp, line 376: very minor, but can we put the comment within the code block at like 378, where we can use `//`

src/js/themes/base.js
taysea: Agree, I was wondering the same and like that idea. Also, I've seen other grouped menus implement the separator as its own `div` all together as opposed to using border on one side of the box. I went the top border approach to try to avoid adding another div to the DOM -- are you good with that?

src/js/components/Menu/Menu.js
```diff
@@ -288,6 +300,92 @@ const Menu = forwardRef((props, ref) => {
...
+          groupIndex > 0
+            ? { side: 'top', ...theme.menu.group?.container?.border }
+            : undefined
+        }
+      >
+        {/* index needs to be its index in the entire menu as if 
+        it were a flat array */}
+        {group.map((item, j) => {
+          let itemIndex = j;
+          for (let k = 0; k < groupIndex; k += 1) {
```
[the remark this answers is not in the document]
taysea: Done.

src/js/components/Menu/Menu.js
taysea: Done.

src/js/components/Menu/Menu.js
taysea: Done.

src/js/themes/base.js
taysea: Done.

ericsoderberghp (approved):

halocline (approved):

PR #6172, opened 2022-06-09
Allow Button to accept elementType for as

taysea:
#### What does this PR do?

Allows Button `as` to accept elementType. This is important for cases when someone uses a router like react-router-dom and wants to do something like:

```
<Button
   as={Link}
   to="/home" (to is the prop for react-router-dom link)
   label="Home"
/>
```

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/Button/__tests__/Button-test.tsx
+++ b/src/js/components/Button/__tests__/Button-test.tsx
@@ -444,14 +444,17 @@ describe('Button', () => {
   });
 
   test('as', () => {
-    const { container } = render(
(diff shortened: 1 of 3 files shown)


ericsoderberghp:
Do we need anything for index.d.ts?

taysea:
> Do we need anything for index.d.ts?

I don't believe so. Index.d.ts already has `PolymorphicType` defined which includes `React.ComponentType<any>;`

ericsoderberghp (approved):

Issue #6151, opened 2022-05-26
MaskInput clearing the whole value when entering some invalid data like below


venkatesh-elchuri:
### Expected Behavior
Mask Input should not clear the whole input value, 
My scenario 

1. I would like to allow only numbers in the input  by passing regex in mas options.
2. Entered number like 123
3. After I came to initial position and entered some spl characters like +,_,),( , Then the whole input value becomes empty .
4. Actually it should be display +123 or _123 like that , if need some invalid entry message. 

### Actual Behavior
If it was a invalid we can show the invalid entry message , But clearing whole value making consumer confuse.

### URL, screen shot, or Codepen exhibiting the issue
![CPT2205261449-942x517](https://user-images.githubusercontent.com/30321668/170458736-2ddda2b5-bf90-4b7e-8fe5-f0b7fb0a53e0.gif)

https://codesandbox.io/s/grommet-text-input-forked-rp8u8p?file=/src/App.js:173-193

In example , I entered 123, After that i came to initial position and tried to enter  "+", Then it become empty

**Is This expected behavior ?**

- Grommet version: Latest
- Browser Name and version: Chrome Latest
- Operating System and version (desktop or mobile): desktop

venkatesh-elchuri:
@ericsoderberghp Can you please look into it ?

ericsoderberghp:
The issue is arising because the component in `parseValue()` gives up as soon as it encounters something that doesn't match the mask. Trying to think about how to more gracefully handle situations like this. Another scenario to consider is entering "12" and then attempting to add a "." between the "1" and "2". Currently, this causes the value to change to "1 ".

PR #6174, opened 2022-06-10
Changed MaskedInput to not clear entire value when mask doesn't match

ericsoderberghp:
#### What does this PR do?

Changed MaskedInput to not clear entire value when mask doesn't match.

Instead of pruning the value down to only the initial part that matches, this change restores it to the prior value. Neither is ideal but I think this behavior is a bit better and aligns more with the existing behavior when an invalid character is entered at the end, with the preceding value remaining.

MaskedInput.js

MaskedInput stories, especially Size + units
DateInput stories with formatting

same as above

preserved existing unit test structure

#6151 

no

yes

backwards compatible

### DIFF
--- a/src/js/components/DateInput/__tests__/DateInput-test.tsx
+++ b/src/js/components/DateInput/__tests__/DateInput-test.tsx
@@ -762,10 +762,9 @@ describe('DateInput', () => {
     fireEvent.change(getByPlaceholderText('mm/dd/yyyy-mm/dd/yyyy'), {
       target: { value: '07//2020-07/27/2021' },
     });
-    expect(onChange).toHaveNthReturnedWith(3, []);
     // cannot check snapshot here as it will be relative to the current date
 
-    expect(onChange).toHaveBeenCalledTimes(3);
+    expect(onChange).toHaveBeenCalledTimes(2);
   });
 
   test('type format inline range partial without timezone', () => {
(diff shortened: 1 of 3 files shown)


src/js/components/DateInput/__tests__/DateInput-test.tsx
```diff
@@ -762,10 +762,9 @@ describe('DateInput', () => {
     fireEvent.change(getByPlaceholderText('mm/dd/yyyy-mm/dd/yyyy'), {
       target: { value: '07//2020-07/27/2021' },
     });
-    expect(onChange).toHaveNthReturnedWith(3, []);
```
jcfilben, line 765: Could this assertion just be changed to `expect(onChange).toHaveNthReturnedWith(3, ['07/']);` instead of being removed?

jcfilben: Ah okay I see

jcfilben (approved):
Looks good!

taysea:
This feels like a frustrating behavior where if (for example) I am inputting a phone number and I realize that one digit is wrong I can't just adjust that one, I have to delete it and start over.

https://user-images.githubusercontent.com/12522275/173131717-8be0b142-99b2-4655-a4f3-a67bbd798bfb.mov

I think I would expect it to behave more like this:

https://user-images.githubusercontent.com/12522275/173132094-79cb55bd-70e3-4df5-8f03-abff03d321d1.mov

https://bl00mber.github.io/react-phone-input-2.html

That being said, I understand that could be tricky to implement.

ericsoderberghp:
Great feedback on it being better but not ideal. So, I've enhanced it just a bit. If the user is removing characters such that the value is invalid, we let them. But, if the user attempts to add characters that would be invalid, we ignore them. Some more interactive testing would be appreciated.

PR #6177, opened 2022-06-13
Add accordion.hover.background to theme

### LINKED ISSUE #6176: Accordion - Add hover indicator
https://www.figma.com/file/UvTBUQdhk07wV79nijsrtr/HPE-Accordion-Component?node-id=0%3A1

Please add the hover state as shown in the designs above to the Accordion component in grommet.

taysea:
#### What does this PR do?

Adds `accordion.hover.background` to theme.

AccordionPanel

Adjusted Custom story and added jest test.

Closes #6176 

Yes.

Yes.

Backwards compatible.

### DIFF
--- a/src/js/themes/base.js
+++ b/src/js/themes/base.js
@@ -346,6 +346,7 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         // margin: undefined
       },
       hover: {
+        // background: undefined,
         color: { dark: 'light-4', light: 'dark-3' }, // deprecated
         heading: {
           color: { dark: 'light-4', light: 'dark-3' },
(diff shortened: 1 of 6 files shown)


src/js/themes/base.js
```diff
@@ -346,6 +346,7 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         // margin: undefined
       },
       hover: {
+        // background: undefined,
```
ericsoderberghp, line 349: My only thought is whether we should consider the lack of hover indicator a bug and add a default here.

ericsoderberghp (commented):
So good. One comment.

src/js/themes/base.js
taysea: I don't think I'd consider it a bug because grommet theme has a hover behavior already that changes the heading color of the text.

ericsoderberghp (approved):

Issue #6179, opened 2022-06-14
Change Box to use CSS gap


ericsoderberghp:
We should change Box to use https://www.w3.org/TR/css-align-3/#gaps instead of inserting an element to control it.

PR #6181, opened 2022-06-14
Changed Box to use row-gap and column-gap when not border='between'.

ericsoderberghp:
#### What does this PR do?

Changed Box to use row-gap and column-gap when not border='between'.

StyledBox.js

cruising through stories

#6179 

no

yes

backwards compatible


src/js/components/Box/Box.js
```diff
@@ -201,6 +212,14 @@ const Box = forwardRef(
         elevationProp={elevation}
         fillProp={fill}
         focus={focus}
+        gap={
+          boxOptions?.cssGap &&
+          gap &&
+          gap !== 'none' &&
+          border !== 'between' &&
+          border?.side !== 'between' &&
+          gap
```
jcfilben, line 221: It looks like we check for gap twice
```suggestion
          boxOptions?.cssGap &&
          gap &&
          gap !== 'none' &&
          border !== 'between' &&
          border?.side !== 'between'
```

ericsoderberghp: We want to return the actual `gap` value if all prior checks are true.

jcfilben: Oops, that makes sense

src/js/components/Menu/__tests__/Menu-test.js
```diff
@@ -140,12 +140,6 @@ describe('Menu', () => {
         />
       </Grommet>,
     );
-
-    const firstItem = getByText('Item 1');
```
taysea, line 144: Any reason why we changed this test?

taysea (approved):
Just one test question.

src/js/components/Menu/__tests__/Menu-test.js
ericsoderberghp: Hmm, I think it was because it couldn't work for the CSS approach, since there is no "StyledBoxGap" in that case. I suspect it's a holdover from the first version that didn't have the Grommet `options` flag. Now that I look at it, I'm wondering if we need it though, since we already have the snapshot check.

ericsoderberghp: It appears to be the only test where we are doing a white-box selector on StyledBoxGap as a class.

jcfilben: I'm fine with removing this section of the test since we have the snapshot to cover this.

PR #6198, opened 2022-06-21
Added id prop to typescript prop definitions

### LINKED ISSUE #6149: Id prop not working with Chart component - Typescript React
Id prop seems like it's working visually, in map of Charts it changes the style of each of my Charts as it should  but I get an error first. 
Apologies if I'm missing something here, I definitely could be. as I'm new to Typescript.  

I posted it here on StackOverflow first. 

https://stackoverflow.com/questions/72333202/cant-use-built-in-prop-of-grommet-chart-component-with-react-typescript

Thanks!!

amandadupell:
#### What does this PR do?

Adds the `id` prop to the typescript prop definitions.

`index.d.ts`

Closes #6149 

No

Yes

Compatible

### DIFF
--- a/src/js/components/Chart/index.d.ts
+++ b/src/js/components/Chart/index.d.ts
@@ -25,6 +25,7 @@ export interface ChartProps {
         color?: ColorType;
         opacity?: 'weak' | 'medium' | 'strong' | boolean | number;
       };
+  id?: string;
   dash?: boolean;
   gap?: GapType;
   onClick?: (...args: any[]) => any;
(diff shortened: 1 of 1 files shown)


ericsoderberghp:
Instead of doing this one property at a time, should we be using something like: `Omit<JSX.IntrinsicElements['svg'], keyof ChartProps>`?

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

Issue #6188, opened 2022-06-16
Menu - Adjust grouped menu separator implementation to accommodate new design direction


halocline:
Based on the new design direction as result of: https://github.com/grommet/hpe-design-system/issues/2640, I believe Menu's group separator implementation needs to be adjusted slightly to allow for the separator to respect the drop's padding.

PR #6201, opened 2022-06-22
Adjust menu grouped separator

jcfilben:
#### What does this PR do?
Adjusts the Menu grouped separator to have the same amount of padding/margin as the menu options.

- [ ] `userEvent` is used in place of `fireEvent`. (Note I had to use firEvent instead of userEvent for the down arrow. User Event wasn't opening the menu)

Closes https://github.com/grommet/grommet/issues/6188

Yes

No
Backwards compatible

### DIFF
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -368,19 +368,17 @@ const Menu = forwardRef((props, ref) => {
         // eslint-disable-next-line react/no-array-index-key
         key={groupIndex}
       >
-        <Box pad={theme.menu.group.separator.pad}>
-          <Box
-            border={
-              groupIndex > 0
-                ? {
-                    side: 'top',
-                    color: theme.menu.group?.separator?.color,
-                    size: theme.menu.group?.separator?.size,
-                  }
-                : undefined
-            }
-          />
-        </Box>
+        {groupIndex > 0 && (
+          <Box pad={theme.menu.group.separator.pad}>
+            <Box
+              border={{
+                side: 'top',
+                color: theme.menu.group?.separator?.color,
+                size: theme.menu.group?.separator?.size,
+              }}
+            />
+          </Box>
+        )}
         <Box {...theme.menu.group?.container}>
           {group.map((item) => {
             // item index needs to be its index in the entire menu as if
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -367,31 +367,30 @@ const Menu = forwardRef((props, ref) => {
       <Box
         // eslint-disable-next-line react/no-array-index-key
         key={groupIndex}
-        {...theme.menu.group?.container}
       >
-        <Box
-          margin={{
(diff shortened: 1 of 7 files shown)


src/js/components/Menu/Menu.js
```diff
@@ -368,16 +368,22 @@ const Menu = forwardRef((props, ref) => {
...
-                color: theme.menu.group?.separator?.color,
-                size: theme.menu.group?.separator?.size,
-              }
-            : undefined
-        }
       >
+        <Box
+          margin={{
+            horizontal: 'small',
+            bottom: index === 0 ? 'none' : 'small',
```
taysea, line 375: The container of the group already has vertical padding. It doesn't feel like we need to add this bottom margin also.

src/js/components/Menu/Menu.js
```diff
@@ -368,16 +368,22 @@ const Menu = forwardRef((props, ref) => {
...
-            ? {
-                side: 'top',
-                color: theme.menu.group?.separator?.color,
-                size: theme.menu.group?.separator?.size,
-              }
-            : undefined
-        }
       >
+        <Box
+          margin={{
```
taysea, line 373: I've seen issues with margin crop up in the past where on areas of restricted width, the margin isn't respected. I'm wondering if we should implement as padding potentially:
```
<Box pad={{ horizontal: 'small' }}><Box border="top" /></Box>

File: src/js/components/Menu/Menu.js [hunk 3]
```diff
@@ -368,16 +368,22 @@ const Menu = forwardRef((props, ref) => {
...
+            horizontal: 'small',
```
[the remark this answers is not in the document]
jcfilben: Yeah that makes sense, I added a `menu.separator.pad` prop to the theme

src/js/components/Menu/Menu.js
jcfilben: Good call, I updated this in the latest commit

src/js/components/Menu/Menu.js
jcfilben: Fixed in latest commit

src/js/components/Menu/Menu.js
```diff
@@ -367,25 +367,30 @@ const Menu = forwardRef((props, ref) => {
...
-        {group.map((item) => {
-          // item index needs to be its index in the entire menu as if
-          // it were a flat array
-          const currentIndex = index;
-          index += 1;
-
-          return menuItem(item, currentIndex);
-        })}
+        <Box pad={theme.menu.group.separator.pad}>
+          <Box
```
taysea, line 372: Are we missing tests for a grouped menu? I would've expected snapshots to update

jcfilben: We had a test for grouped menu but the menu items weren't showing in the snapshot. I changed the test so that the menu items are showing.

taysea (approved):
Nice work

ericsoderberghp (approved):

Issue #6197, opened 2022-06-21
Enhancement: Improve layout for bar charts


thomas-friedrich:
Our design includes a bar chart with 2 bars that should be displayed like this:
<img width="400" alt="image" src="https://user-images.githubusercontent.com/84544729/174848247-83b31699-5e2a-46d4-a49b-d2482e44739b.png">

However, grommet tries to evenly distribute the data points along the x-axis, starting at the beginning and renders the 2 bars like this:
<img width="400" alt="image" src="https://user-images.githubusercontent.com/84544729/174848399-56ee09b3-6714-4a93-8a6d-43d18329c720.png">

We'd like more control over the layout in DataCharts with bars and either provide a margin for the x-axis where the bars should start or pick a different layout to display the bars around the center of the chart.

halocline:
Related [Slack thread](https://grommet.slack.com/archives/C09QQEG69/p1655494770951099) with recommendations for how to achieve the desired rendering.

thomas-friedrich:
The suggested `pad={{ horizontal: 'xlarge' }}` in the Slack thread, doesn't work 100% because the bar, label and the details tooltip are not lining up (in fact, you have to hover next to bar to show the details).
<img width="380" alt="image" src="https://user-images.githubusercontent.com/84544729/174899017-a7340534-905a-49a5-a53c-23921c388bf3.png">

jcfilben:
It looks like we need to support an option to allow having the x-axis bars not start at the far left and end at the far right and instead allow options to be centered with equal space between. Or we need to improve the `pad` prop to adjust the `detail` alignment.

PR #6206, opened 2022-06-27
Changed DataChart to fix an issue with detail padding.

ericsoderberghp:
#### What does this PR do?

Changed DataChart to fix an issue with detail padding.

code

Added a "Pad" story

"Pad" story

no changes to unit tests, not sure how to test detail interaction via jest

#6197 

Screen Shot 2022-06-27 at 8 15 47 AM

no

yes

backwards compatible


jcfilben (commented):
Code looks good

Just a clarification question. It looks like this is working well when the `chart.thickness` is equal to the `pad.horizontal` value. If we wanted a chart with a 'xlarge' pad but with 'small' thickness for the bars is there a way to achieve this with the `detail` correctly aligned? Or would this be a separate issue/PR?

ericsoderberghp:
> Code looks good
> 
> Just a clarification question. It looks like this is working well when the `chart.thickness` is equal to the `pad.horizontal` value. If we wanted a chart with a 'xlarge' pad but with 'small' thickness for the bars is there a way to achieve this with the `detail` correctly aligned? Or would this be a separate issue/PR?

Good idea. I've updated it to handle pad and thickness better.

jcfilben (approved):
Looks good!

PR #6207, opened 2022-06-27
Implement Paragraph maxLines

MikeKingdom:
Many designs have situations where they put a paragraph of text on a card or some other space-restricted area. Often that content comes from a dynamic source like external product descriptions, etc. This designs typically would like to truncate this content at a certain number of lines. Our current `<Text>` component can truncate a single line with an ellipsis, but we have no convenient solution for truncating something that is multiple lines with an ellipsis in an effecient way.

#### What does this PR do?
This adds a `maxLines` property to `<Paragraph>`. When the `<Paragraph>` gets constrained by a parent container or its max width restriction the content will get cleanly truncated with an ellipsis using an efficient browser css-supported mechanism much like `<Text truncate>`

image

Storybook Paragraph->Maxlines

Storybook and jest

Storybook

Yes

Yes

Backwards compatible

### DIFF
--- a/src/js/components/Paragraph/StyledParagraph.js
+++ b/src/js/components/Paragraph/StyledParagraph.js
@@ -21,8 +21,18 @@ const fontFamily = css`
   font-family: ${(props) => props.theme.paragraph.font.family};
 `;
 
+const maxlinesStyle = (props) =>
+  props.maxLines &&
+  css`
+    display: -webkit-box;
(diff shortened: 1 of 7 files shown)


jcfilben (approved):
Looks good!

britt6612 (approved):
LGTM
Lets just add to docs once this is merged 👍

PR #6211, opened 2022-06-28
Typescript - include info as status type on Notification

taysea:
#### What does this PR do?

Adds `info` as a StatusType type definition for Notification.

src/js/components/Notification/index.d.ts

Locally on TS storybook.

https://github.com/grommet/grommet-site/pull/396

Yes.

Backwards compatible.


jcfilben (commented):
Looks like we also need to add 'info' to `status` in Notification's propTypes and add it to the docs

ericsoderberghp (approved):

Issue #6044, opened 2022-04-01
DateInput daylight savings time


jcfilben:
### Expected Behavior
DateInput should show the day you selected on the Calendar

### Actual Behavior
If you select a date in daylight savings time (mid March - mid November) and then select a date outside of daylight savings (late November - early March) the DateInput will show 1 day before what you actually selected. This is happening across timezones.

### URL, screen shot, or Codepen exhibiting the issue
This can be seen in the [DateInput/European](https://storybook.grommet.io/?path=/story/input-dateinput-european--european) story

### Steps to Reproduce

1. Open the DateInput and select March 30th
2. Select March 2nd (date displayed will be March 1st)

### Your Environment

- Grommet version: 2.22.0
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile):

danielm-hpe:
I mentioned something *similar* in slack, but it's enough different that it may not be related.  I think it was likely related to the normalization behavior introduced in https://github.com/grommet/grommet/pull/5942

- in the browser it works OK
- in jest tests on CI/CD it works OK
- in jest tests locally the dates in the onChange are OK, but the display value is off by 1.  If the timezone is explicitly set to UTC in the local test run by doing something like `TZ=UTC yarn test`, then the tests pass

I may have a different issue based on this information, but I created a minimum repro that demonstrates the browser behavior and test behavior:
https://github.com/danielm-hpe/grommet-dateinput-repro

halocline:
Potentially related detail: https://grommet.slack.com/archives/C09QQEG69/p1654906323407089

kimkscott:
Thanks @halocline for adding a pointer to my question in slack.  I don't know if it is related, but sometimes when I pass a UTC iso timestamp to DateInput in the value property,  the component doesn't show the date according to the browser system’s time zone.  The question in slack has a screenshot of this.  I was thinking it might be related to the calculations in  normalizeForTimezone() in Calendar/utils.

Issue #6043, opened 2022-04-01
Calendar daylight savings handling not working in certain timezones


jcfilben:
### Expected Behavior
The date that is click on the calendar should be what is displayed as selected. 

### Actual Behavior
In certain timezones (eastern standard time is one) when you select a date in daylight savings time (mid March - mid November) and then select a date outside of daylight savings (late November - early March) the Calendar will show 1 day before what you actually selected. 

### URL, screen shot, or Codepen exhibiting the issue
This can be seen on the Calendar/Simple story

### Steps to Reproduce

1. Change timezone to eastern standard time
2. Select March 30th on the Calendar
3. Select March 2nd on the Calendar (date displayed will be March 1st)

### Your Environment

- Grommet version: 2.22.0
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile):

halocline:
Crux of the issue -

Right now Calendar’s approach conflates UTC offsets with timezones. Specifically, in order to properly handle standard vs. daylight savings time (DST), we need to know the timezone which contains the information needed to adjust for DST.

Currently, Calendar’s implementation uses the UTC offset (‘timestamp’) as proxy for timezone. This works for most cases except for when DST information is needed. We cannot derive the DST from offset. Many timezones may reference the same offset. Once using offset, we cannot know which timezone should be mapped to, timezone holds DST information and is not available, and therefore we cannot handle DST situations properly. 

To remedy this, Calendar needs to store and use timezone information. Additionally, allowing a caller to specify a timezone should be a consideration. This would be needed in scenarios where a user sits in one timezone, but is scheduling an activity in another timezone.

PR #6214, opened 2022-06-30
Update Calendar to rely on Date object

taysea:
#### What does this PR do?

Previously, Calendar could not handle Daylight Savings Time properly because it relied on ISOstring to infer timezone, however ISOstring only contains information related to UTC offset, but the actual timezone information is lost. [See comment for more detail](https://github.com/grommet/grommet/issues/6043#issuecomment-1171532041).

This PR refactors Calendar rely on Date object as opposed to ISO strings.
- Calendar accepts ISOstrings as input --> normalizes the inputs as Date objects --> all logic is handled as Date objects --> only normalizes output supplied to onSelect as last step, returning dates as ISOstrings.
- Additionally, to reduce complexity, consolidate logic, and improve control flow, this PR makes the following adjustments:
  - Internally, `date` and `dates` both feed an internal variable called `value`. API surface remains the same. All logic consider `value` to be either a Date object, an array of Date objects, a nested array of Date objects, or undefined.
- Adds timezone test coverage to Jest scripts.

Closes https://github.com/grommet/grommet/issues/6043
Closes https://github.com/grommet/grommet/issues/6044

Yes.

Backwards compatible + some bug fixes


src/js/components/Calendar/Calendar.js
```diff
@@ -276,75 +240,12 @@ const Calendar = forwardRef(
...
-      setDate(normalizedDate);
-    }, [normalizedDate]);
+    const [date, setDate] = useState(dateProp);
 
-    // set dates when caller changes it, allows us to change it internally too
-    const [dates, setDates] = useState(normalizedDates);
-    useEffect(() => {
-      setDates(normalizedDates);
-    }, [normalizedDates]);
+    const [dates, setDates] = useState(datesProp);
```
MikeKingdom, line 245: If the dateProp changes on a future render this won't get updated and it will continue to use the original value unless something else calls setDates. Is that intended?

taysea: Good callout -- we'll need to look into this. Still exploring behavior here in general, thanks for the note.

src/js/components/Calendar/Calendar.js
```diff
@@ -165,6 +165,21 @@ const buildDisplayBounds = (reference, firstDayOfWeek) => {
...
 
+const getOutputFormat = (dates) => {
+  let date;
+  if (typeof dates === 'string') {
+    date = dates;
+  } else if (Array.isArray(dates)) {
+    [date] = dates;
+  }
+  let result = 'date timezone';
+  const regex = /T/;
```
MikeKingdom, line 176: Seems a bit of overkill to use regex here instead of indexOf

jcfilben (commented):
It looks like I am no longer able to change the date on Calendars that don't have `onSelect` defined. (See Calendar/First on Sunday and Daylight Savings stories)

taysea:
> It looks like I am no longer able to change the date on Calendars that don't have `onSelect` defined. (See Calendar/First on Sunday and Daylight Savings stories)

Good point -- the docs read that not specifying `onSelect` makes the component read only, so we are aligning to that: https://v2.grommet.io/calendar#onSelect

we should confirm, as a team, that that's desired behavior.

jcfilben:
> Good point -- the docs read that not specifying onSelect makes the component read only, so we are aligning to that: https://v2.grommet.io/calendar#onSelect
> 
> we should confirm, as a team, that that's desired behavior.

Ah okay that makes sense. I think that behavior is reasonable, we should just to be sure to call this out in the release notes since before we weren't aligning to this

ericsoderberghp (approved):
nice cleanup

src/js/components/DateInput/__tests__/__snapshots__/DateInput-test.tsx.snap
```diff
@@ -5474,6 +5464,28 @@ exports[`DateInput controlled format inline without timezone 2`] = `
...
+              <div
+                class="c15"
+                role="row"
+              >
+                <div
+                  class="c16"
+                  role="gridcell"
+                >
+                  <button
+                    aria-label="Sun Aug 09 2020"
```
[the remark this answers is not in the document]
taysea: Looking into this more -- it's strange because I'm unable to reproduce this on storybook.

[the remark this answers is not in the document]
taysea: Fixed in recent commit!

jcfilben (approved):
Looks good!

MikeKingdom (approved):
Looks good. Thanks!

Issue #6130, opened 2022-05-14
Form sets state after onChange debounce delay without ensuring component is still mounted


simleb:
The `Form` component, when using `onChange` validation (`validate="change"`) debounces validation (using a hardcoded delay, see https://github.com/grommet/grommet/issues/6129) using a [function that attempts to set some state][1] without checking that the form is still mounted. This bit me while testing and forced me to use a delay to ensure the debounced action was finished before ending the test.

[1]: https://github.com/grommet/grommet/blob/9b022dd427f6ae18705c06b0584e152ab273676b/src/js/components/Form/Form.js#L518-L523

### Expected Behavior

When the form dismounts, either cancel all debounced actions or use a ref to ensure the component is still mounted before attempting to set its state.

### Actual Behavior

Debounced action tries to set state after the form component has been dismounted.

### URL, screen shot, or Codepen exhibiting the issue

N/A
### Steps to Reproduce

```typescript
<Form validate="change">
  <FormField validate={(_val: string) => undefined}>
    <TextInput placeholder="type here" />
  </FormField>
</Form>
```
1. Mount form component
1. Send a change event
1. Destroy form component less than 500ms after the event

### Your Environment

Jest/Enzyme

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

Issue #6129, opened 2022-05-14
FormField onChange validation delay is hardcoded to 500ms


simleb:
`onChange` validation is debounced to 500ms and there is no option to override it.

Beyond the fact that there is no universally accepted "correct delay" so it's fair to let developers override the default, it's also troublesome because [`Form` attempts to set state][0] after the debounce delay without checking that the component is still mounted (I wrote it as a [separate issue][4]).

[0]: https://github.com/grommet/grommet/blob/382858daa1c393b478428f7296688bbbd84e2e95/src/js/components/Form/Form.js#L517-L523
[4]: https://github.com/grommet/grommet/issues/6130

I found [this PR fixing a similar issue for `Select`][1] but the code has changed since then and while [`debounceDelay` remains in the theme][2], it appears to be unused.

[1]: https://github.com/grommet/grommet/pull/2489
[2]: https://github.com/grommet/grommet/blob/512ec0dbb9856a9d50fad16d6f02af8493de1d10/src/js/themes/base.js#L222-L223

### Expected Behavior

`onChange` validation delay should be customizable, either at the theme level or as a prop on the form (or ideally both with the prop taking precedence).

If validation ends up being [customizable at the `FormField` level][3], then a prop at that level would be appropriate too.

[3]: https://github.com/grommet/grommet/issues/5913

The docs should be updated to reflect the change.

### Actual Behavior

`onChange` validation delay is hardcoded at 500ms.

### URL, screen shot, or Codepen exhibiting the issue

https://github.com/grommet/grommet/blob/382858daa1c393b478428f7296688bbbd84e2e95/src/js/components/FormField/FormField.js#L479-L484

### Steps to Reproduce

N/A

### Your Environment

N/A

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #6226, opened 2022-07-08
Cancel debounce function after FormField unmounts

jcfilben:
#### What does this PR do?
Prevents the debounce function from executing if the FormField is unmounted.

Tested with the following story and console log statements in the debounce function
```
import { Box, Form, FormField, TextInput, Button } from 'grommet';

export const TEMP = () => {
  const [showForm, setShowForm] = useState(true);

  return (
      <Box width="medium">
        {showForm && <MyForm setShowForm={setShowForm} />}
        <Button
          label="Show form"
          onClick={() => {
            setShowForm(true);
          }}
        />
      </Box>
    </Box>
  );
};

const MyForm = ({ setShowForm }) => (
  <Form
    validate="change"
    onChange={(val) => {
      console.log(`onChange ${val}`);
      setShowForm(false);
    }}
  >
    <FormField
      name="myField"
      validate={(val) => {
        console.log(`validate ${val}`);
      }}
    >
      <TextInput name="myField" placeholder="type here" />
    </FormField>
  </Form>
);

TEMP.storyName = 'TEMP';
TEMP.args = {
  full: true,
};

export default {
  title: 'Input/Form/TEMP',
};
```

Closes https://github.com/grommet/grommet/issues/6130
Closes https://github.com/grommet/grommet/issues/6129

It might be good to add the `debounceDelay` theme prop to the FormField doc. I will create a PR for this.
Maybe
Backwards compatible

### DIFF
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -139,7 +139,6 @@ const debounce = (func, wait, fieldRef) => {
   return function executedFunction(...args) {
     const later = () => {
       if (fieldRef.current) {
-        console.log('function executed');
         timeout = null;
         func(...args);
       }
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -134,12 +134,15 @@ const Input = ({ component, disabled, invalid, name, onChange, ...rest }) => {
   );
 };
 
-const debounce = (func, wait) => {
+const debounce = (func, wait, fieldRef) => {
   let timeout;
   return function executedFunction(...args) {
     const later = () => {
-      timeout = null;
-      func(...args);
+      if (fieldRef.current) {
+        console.log('function executed');
+        timeout = null;
+        func(...args);
+      }
     };
     clearTimeout(timeout);
     timeout = setTimeout(later, wait);
@@ -474,14 +477,19 @@ const FormField = forwardRef(
                 event.persist();
                 if (onChange) onChange(event);
                 if (contextOnChange) {
-                  const debouncedFn = debounce(() => {
-                    contextOnChange(event);
-                    // A half second (500ms) debounce can be a helpful starting
-                    // point. You want to give the user time to fill out a
-                    // field, but capture their attention before they move on
-                    // past it. 2 second (2000ms) might be too long depending
-                    // on how fast people type, and 200ms would be an eye blink
-                  }, 500);
+                  const debouncedFn = debounce(
+                    () => {
+                      contextOnChange(event);
+                      // A half second (500ms) debounce can be a helpful
+                      // starting point. You want to give the user time to
+                      // fill out a field, but capture their attention before
+                      // they move on past it. 2 second (2000ms) might be too
+                      // long depending on how fast people type, and 200ms
+                      // would be an eye blink
+                    },
+                    500,
+                    formFieldRef,
+                  );
                   debouncedFn();
                 }
               }
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -4,6 +4,7 @@ import React, {
   forwardRef,
   useContext,
   useState,
+  useEffect,
 } from 'react';
 import styled, { ThemeContext } from 'styled-components';
 import { defaultProps } from '../../default-props';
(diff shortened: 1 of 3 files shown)


src/js/components/FormField/FormField.js
```diff
@@ -474,14 +476,19 @@ const FormField = forwardRef(
...
+                    () => {
+                      contextOnChange(event);
+                      // A half second (500ms) debounce can be a helpful
+                      // starting point. You want to give the user time to
+                      // fill out a field, but capture their attention before
+                      // they move on past it. 2 second (2000ms) might be too
+                      // long depending on how fast people type, and 200ms
+                      // would be an eye blink
+                    },
+                    500,
```
halocline, line 489: Wondering if this delay should be something the caller can configure? Especially since we don't know what kind of validation function might be called. There is a [theme prop](https://github.com/grommet/grommet/blob/f8c7358c43895ac8bb10e91ada33a5356a3adedd/src/js/themes/base.js#L223), but seems to have been abandoned...

halocline (approved):
Tested. Cancelation works as expected. Nice job.

src/js/components/FormField/FormField.js
jcfilben: Yes, I agree that this should be configurable. Since `debounceDelay` is not being used currently I think we could change it from '300' to '500' and then use it here

src/js/components/FormField/FormField.js
```diff
@@ -474,14 +476,19 @@ const FormField = forwardRef(
...
-                    // A half second (500ms) debounce can be a helpful starting
-                    // point. You want to give the user time to fill out a
-                    // field, but capture their attention before they move on
-                    // past it. 2 second (2000ms) might be too long depending
-                    // on how fast people type, and 200ms would be an eye blink
-                  }, 500);
+                  const debouncedFn = debounce(
+                    () => {
+                      contextOnChange(event);
+                      // A half second (500ms) debounce can be a helpful
```
ericsoderberghp, line 482: Can we move this comment into `base.js`, since that is where the value is being set?

src/js/components/FormField/FormField.js
```diff
@@ -474,14 +476,19 @@ const FormField = forwardRef(
...
                 if (contextOnChange) {
-                  const debouncedFn = debounce(() => {
-                    contextOnChange(event);
-                    // A half second (500ms) debounce can be a helpful starting
-                    // point. You want to give the user time to fill out a
-                    // field, but capture their attention before they move on
-                    // past it. 2 second (2000ms) might be too long depending
-                    // on how fast people type, and 200ms would be an eye blink
-                  }, 500);
+                  const debouncedFn = debounce(
```
ericsoderberghp, line 479: Can we remove the `Fn` suffix? It doesn't seem consistent with grommet's naming convention to include typing in variable names.

src/js/components/FormField/FormField.js
```diff
@@ -134,12 +134,14 @@ const Input = ({ component, disabled, invalid, name, onChange, ...rest }) => {
   );
 };
 
-const debounce = (func, wait) => {
+const debounce = (func, wait, fieldRef) => {
```
ericsoderberghp, line 137: Instead of making debounce() know about the ref, what about changing it to be a hook that will clear the timer automatically when the component is unmounted?

src/js/components/FormField/FormField.js
jcfilben: Yup, done

src/js/components/FormField/FormField.js
jcfilben: Done

src/js/components/FormField/FormField.js
jcfilben: Good idea, I updated to use a hook in the latest commit

ericsoderberghp: A suggestion to tighten it up a bit:
```
const useDebounce = () => {
  const [func, setFunc] = useState();

  useEffect(() => {
    const timer = setTimeout(() => func(), theme.global.debounceDelay);
  }, [func]);
  
  return setFunc;
};

...
const debounce = useDebounce();
...
if (contextOnChange) debounce(() => contextOnChange(event)));
```
untested, but perhaps you get the idea?

src/js/components/FormField/FormField.js
```diff
@@ -473,17 +473,7 @@ const FormField = forwardRef(
...
-                    contextOnChange(event);
-                    // A half second (500ms) debounce can be a helpful starting
-                    // point. You want to give the user time to fill out a
-                    // field, but capture their attention before they move on
-                    // past it. 2 second (2000ms) might be too long depending
-                    // on how fast people type, and 200ms would be an eye blink
-                  }, 500);
-                  debouncedFn();
-                }
+                if (contextOnChange) setDebounced(() => contextOnChange);
```
ericsoderberghp, line 476: We seem to have lost the event argument to `contextOnChange()`

jcfilben: Fixed in latest commit

ericsoderberghp (commented):
looking good, one simple comment

ericsoderberghp (approved):

Issue #5446, opened 2021-07-14
Extend Grommet Theme Typings


Kevin-Carlos:
I was looking for a way to do something similar to `styled-components` and how you can override their `DefaultTheme` to get proper typings for custom theming.

See: https://styled-components.com/docs/api#create-a-declarations-file

I wasn't able to find anything similar for grommet nor a similar issue, however, I could have easily missed one!

Being able to have proper type access to our custom colors, etc, would definitely enhance the development experience instead of remembering particular strings off the top of our heads though!

PR #6234, opened 2022-07-16
[Typescript] Support focus shadow as a string

ShimiSun:
#### What does this PR do?
Supports the theme's `focus.shadow` to accept a string.

theme file with types
local tests.
locally.
N/A

On src/js/utils/styles.js we use:
```
return `
      outline: none;
      box-shadow: ${focus.shadow};
    `;
```
and since `box-shadow` accepts the type of string, it [should be accepted as a valid input type](https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow) as well.

#5446 
Where are the global theme types are documented? the new type should be added there.
yep, as a TS fix.
 backward compatible

### DIFF
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -368,7 +368,8 @@ export interface ThemeType {
         color?: ColorType;
         size?: string;
       };
-      shadow?: {
+      shadow?: 
+        string | {
         color?: ColorType;
         size?: string;
       };


jcfilben:
> Do the grommet docs need to be updated?
Where are the global theme types are documented? the new type should be added there.

Some of the global theme props are documented on the [grommet theme wiki](https://github.com/grommet/grommet/wiki/Grommet-v2-theming-documentation) but this is out of date and doesn't include the latest props. I am working on creating a global theme props page on grommet-site that will include documentation for all of the global theme props.

ericsoderberghp (approved):

PR #6237, opened 2022-07-19
Fix TS definition on Menu items

taysea:
#### What does this PR do?

Fixes Menu `items` typescript definition to use ButtonExtendedProps which includes intrinsic properties.

#### Where should the reviewer start?
src/js/components/Menu/index.d.ts

#### What testing has been done on this PR?
TS example

#### Do Jest tests follow these best practices?

- [ ] `screen` is used for querying.
- [ ] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [ ] `userEvent` is used in place of `fireEvent`.
- [ ] `asFragment()` is used for snapshot testing.

#### Should this PR be mentioned in the release notes?
Yes. Fixed Menu `items` typescript definition.

#### Is this change backwards compatible or is it a breaking change?
Backwards compatible.

PR #6241, opened 2022-07-20
fix: menu types and added 'button' prop for menu theme

ShimiSun:
#### What does this PR do?
1. Suggests a clean way for generic styles on Menu's item-buttons via theme. Applying the styles (on both default + mirror rendering) helps declutter lots of inline styles per Menu item.

2. Improve types of Menu (item type, theme.menu.drop type, etc), the existing Stroybook Menu 'Themed' example is throwing an error due to the `onClick` of the Menu item:
image

3. Extending Menu 'Themed' story with the new prop

Menu.js

storybook - check the enhanced story 
storybook
N/A

The new prop will help clean up a lot of inline styles within Menu items.
Yes.

Yes, will be helpful for Menu users and TS users.
 backward compatible


src/js/components/Menu/Menu.js
```diff
@@ -353,6 +354,7 @@ const Menu = forwardRef((props, ref) => {
               onDropClose();
             }
           }}
+          {...theme.menu.button}
```
jcfilben, line 357: It seems like this should be above some of the other props in the button instead of at the end so that if we pass button props through `items` they don't get overridden by what is defined in the theme.

src/js/components/Menu/index.d.ts
```diff
@@ -31,7 +31,7 @@ export interface MenuProps {
   dropProps?: DropProps;
   gridArea?: GridAreaType;
   icon?: boolean | React.ReactNode;
-  items: ButtonProps[] | ButtonProps[][];
+  items: ButtonProps[] | ButtonProps[][] | ButtonType[] | ButtonType[][];
```
jcfilben, line 34: I think this can just be
```suggestion
  items: ButtonExtendedProps[] | ButtonExtendedProps[][];
```
like in https://github.com/grommet/grommet/pull/6237/files

src/js/themes/base.d.ts
```diff
@@ -232,6 +232,7 @@ interface ButtonType {
       };
     };
   };
+  style?: Partial<CSSStyleDeclaration>;
```
jcfilben, line 235: Can you give an example of something that we would want to pass through style that wouldn't be achievable through button props? Just trying to get a better idea of how this could be used

src/js/components/Menu/Menu.js
ShimiSun: I agree. If the caller used a specific style per item I would expect it to take precedence over a theme styling. Fixed.

src/js/components/Menu/index.d.ts
ShimiSun: Yep, I tested it and it did the trick.

jcfilben (commented):
Trying to pass certain props through `menu.button`, like `justify` isn't working well because of how child is defined (line 306)

src/js/themes/base.d.ts
```diff
@@ -1025,7 +1026,8 @@ export interface ThemeType {
   };
   menu?: {
     background?: BackgroundType;
-    drop?: DropProps;
+    button?: ButtonType;
```
ericsoderberghp, line 1029: I'm wondering if this should be called `item`, since the Menu is a DropButton already. It might align better with the `items` property.

src/js/themes/base.d.ts
ericsoderberghp, line 235: This feels like heading in a new direction for something that feels like we've used `extend` for in other places. What the rationale here?

src/js/components/Menu/stories/CustomThemed/Themed.tsx
```diff
@@ -60,10 +60,16 @@ const customBreakpoints: ThemeType = {
...
   },
   menu: {
+    background: 'light-3',
     icons: {
       down: FormDown,
       up: FormUp,
     },
+    button: {
+      color: 'brand',
+      style: { margin: '6px' },
```
ericsoderberghp, line 70: Can you share a visual on why `margin` and not `pad` somewhere? Is the idea to create space around the menu items? If so, I wonder if `theme.menu.itemContainer: BoxType` might be more consistent, like `theme.notification.textContainer`.

src/js/components/Menu/stories/CustomThemed/Themed.tsx
```diff
@@ -60,10 +60,16 @@ const customBreakpoints: ThemeType = {
...
+    },
+    extend: `border: 3px solid gray`,
```
ericsoderberghp, line 72: Why `extend` here instead of just `theme.menu.border`?

src/js/themes/base.d.ts
ShimiSun: We still need `extend` when styles are driven from theme props, but if the theme uses component props it will also need to accommodate any `styles` of the component as well.

src/js/components/Menu/stories/CustomThemed/Themed.tsx
ShimiSun: I used `margin` just for the easy example, I'll replace it with a font style to get closer to my use case where font-weight should behave differently on the menu button.

src/js/components/Menu/stories/CustomThemed/Themed.tsx
ShimiSun: Right! That would be nice if we had `theme.menu.border` :) It could be a nice enhancement request but not pressing on my end.

src/js/themes/base.d.ts
ShimiSun: Good observation. That does make sense to me, I'll refine it.

jcfilben (approved):
Looks good!

src/js/components/Menu/Menu.js
```diff
@@ -309,7 +310,7 @@ const Menu = forwardRef((props, ref) => {
         pad="small"
         direction="row"
         gap={item.gap}
```
ericsoderberghp, line 312:
```suggestion
        gap={item.gap || theme.menu.item?.gap}
```

src/js/components/Menu/Menu.js
```diff
@@ -336,6 +337,7 @@ const Menu = forwardRef((props, ref) => {
           justify={item.justify}
           kind={!child ? 'option' : undefined}
           hoverIndicator={!child ? undefined : 'background'}
+          {...theme.menu.item}
```
ericsoderberghp, line 340: This looks like the theme would override `justify={item.justify}` above. I'mm wondering if we should more generally merge properties, via something like:
```
const defaultItemProps = { align: 'start' };
...
<Button {{ ...defaultItemProps, ...theme.menu.item, ...item }} \>
```
Does that make any sense, or am I missing something?

ShimiSun: Agreed.

ericsoderberghp (approved):

PR #6243, opened 2022-07-20
Fix Select activeRef tracking

### LINKED ISSUE #6208: Unable to select middle option in a select component
If I scroll all the way to the bottom of the Select and then try to quickly scroll up, I'm unable to select middle options as options keep scrolling away. I can only select like the first ten or the last ten from this list, the middle ones it's like bugging and not letting me select.

### Expected Behavior

Select shouldn't scroll up when selecting middle options.

### Actual Behavior

Select keeps scrolling

https://user-images.githubusercontent.com/664555/176014736-c68b0073-49c2-4cd2-8125-de8acdae2bb5.mov

### Steps to Reproduce

- Scroll all the way to the bottom of the Select and then try to quickly scroll up.

### Your Environment

- Grommet version: 2.24
- Browser Name and version: Chrome Version 102.0.5005.115 (Official Build) (x86_64) for MacOS

### LINKED ISSUE #6238: Select - Odd autoscroll behavior in Select
**Expected Behavior**

When onMore occurs and adds items to `options` the user's cursor should remain at the item index in the list on which they currently have focus.

**Actual Behavior**

An unexpected autoscroll occurs after an initial onMore is called, more data is loaded, and then the user moves their cursor.  I will create a minimal code example which reproduces this experience and add it to this issue.

These are the steps which trigger the scroll behavior shown in the video below:
- Open modal containing a Select input.
- Click on Select dropdown to reveal the first set of options. `options` is initially populated with a list of 50 items.
- Scroll down through the options to approximately the 75% point in the scrollbar to trigger the onMore call.  onMore triggers a call to the API, retrieving the next 50 items adding them to `options`. Note: I did replicate the same behavior with an array, taking the next slice to remove the API call from the possible variables.
- Stop scrolling.  (in the video, the cursor is on the cursor is on "Hawaii" when this occurs at ~00:08).
- Note: The height of the scroll control reduces as the next 50 items are provided as options at 00:09.
- All is fine as long as the cursor remains on the currently active item (e.g. "Hawaii"). 
- However, as soon as the cursor moves onto another item (00:12), which would become the next active index, the unexpected autoscroll occurs, skipping about 40 items and resolving at an item called "athena-f-2004200" at 00:14.

https://user-images.githubusercontent.com/1756948/179840901-03d32e29-2e37-458b-8d74-6d50458b1840.mov

### LINKED ISSUE #6248: Select - custom theme option hover background state no longer respected
### Expected Behavior

Select's option hover background should apply the value specified in [global.hover.background](global.hover.background).

### Actual Behavior

Select alway applies 'active' color and no longer respect's the global.hover.background property.

[Snapshot change occurred in this PR](https://github.com/grommet/grommet/pull/5934/files#diff-22d4dd713b3d80e2619b333f14558eab4e03e5a48b609b833b84b29b5e5067ca) (the styles specified lines 2127-2130 should still remain), but the regression was not noticed in review.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: v2.21+

taysea:
#### What does this PR do?

Improves tracking of active item. Previously we were keying off of `activeIndex` which doesn't always align with the elements index in the select options (because infinite scroll loads options in steps). Instead, we track the active value in a ref and focus that.

We only ever want one option to be highlighted at once. I've simplified the `keyboardNavigation` logic to more simply handle the trade off of keyboard and mouse movement.

In order for the underlying button to receive the `theme.global.hover.background` treatment correctly for both mouse interactions and keyboard interactions (the styling should be consistent regardless of if user is interacting via mouse or keyboard). Adjusted SelectOption styled component to apply this instead of relying on hoverIndicator (which would only apply on :hover).

The snapshot changes are related to the active styling being handled now by `activeStyle` as opposed to `hoverStyle`. The visual experience for the user remains unchanged.

Tested with voice over and compared new behavior with behavior enhanced in https://github.com/grommet/grommet/pull/5934. All behavior is retained.

Previously flagged keyboard errors have been resolved.

Closes #6208 
Closes #6238
Closes #6248

No.

Yes.

Backwards compatible.


src/js/components/Select/__tests__/__snapshots__/Select-test.js.snap
```diff
@@ -2040,6 +2040,11 @@ exports[`Select applies custom global.hover theme to options 2`] = `
   color: #000000;
 }
 
+.c0:hover {
```
taysea, line 2043: I believe this is a fix to behavior that previously should've been in the snapshot.

ericsoderberghp (dismissed):

halocline (changes requested):
The keyboard behavior seems off.

- Go to 'many options' story.
- Tab focus on input
- Spacebar to open
- Down arrow as far as you can
- Active index gets stuck at Option 33 and then keyboard controls stop working

Alternatively:
- Open drop with mouse
- Scroll to index far down (e.g. 60)
- Try using arrow keys

src/js/components/Select/SelectContainer.js
```diff
@@ -86,22 +86,21 @@ const SelectContainer = forwardRef(
...
+    const [activeIndex, setActiveIndex] = useState(usingKeyboard ? 0 : -1);
     const [keyboardNavigation, setKeyboardNavigation] = useState();
     const searchRef = useRef();
     const optionsRef = useRef();
     const clearRef = useRef();
+    const activeRef = useRef();
 
+    // Only update active suggestion index when the mouse actually moves,
+    // not when suggestions are moving under the mouse.
+    const [mouseMovedSinceLastKey, setMouseMovedSinceLastKey] = useState(true);
```
halocline, line 98: A couple of thoughts/questions around this variable. Will connect with you to talk through.

halocline (commented):
Functionality looks good, just want to chat through a couple items to see if we might be able to simplify a bit of the keyboard vs mouse interactions.

- Fixes #6208 & #6238
- Fixes a custom theme hover background state regression as well.

src/js/components/Select/SelectContainer.js
taysea: I've cleaned this up now.

halocline (approved):
Closes #6208, #6238, and #6248.

🙌

src/js/components/Button/StyledButton.js
```diff
@@ -142,6 +142,8 @@ const plainStyle = (props) => css`
 
 const activeButtonStyle = (props) => css`
   ${activeStyle}
+  ${props.activeIndicator &&
```
ericsoderberghp, line 145: It looks like we're adding `activeIndicator` to Button? If so, I think we're missing typescript, propTypes, unit tests, etc. Also, I'm wondering why we need this at the Button API level. I would expect that the presence of the `active` property would be sufficient. In other words, if I indicate that a Button is `active`, then I would expect the Button to indicate that visually some how.

src/js/components/Select/SelectContainer.js
```diff
@@ -483,7 +471,9 @@ const SelectContainer = forwardRef(
                       plain={!child ? undefined : true}
                       align="start"
                       kind={!child ? 'option' : undefined}
-                      hoverIndicator={!child ? undefined : 'background'}
+                      // applies theme.global.hover.background to the active
+                      // option for mouse and keyboard interactions
+                      activeIndicator={!child ? undefined : 'background'}
```
taysea, line 476: @ericsoderberghp the activeIndicator changes are needed for this functionality only (isolated to select) and I don't think should be a documented prop on button.

src/js/components/Button/StyledButton.js
taysea: Details about isolated use in select in the code comment here: https://github.com/grommet/grommet/pull/6243/files#r927979426

src/js/components/Select/SelectContainer.js
taysea: Retaining support for `theme.global.hover.background` is needed for backwards compatibility

taysea: https://github.com/grommet/grommet/pull/6243/files#r927865579

src/js/components/Button/StyledButton.js
taysea: open to other ways of achieving the functionality, but the functionality is needed

src/js/components/Select/SelectContainer.js
ericsoderberghp: Let me see if I've got this right ...
We want to display the same visual as hovering but we don't want that to be tied strictly to mouse hovering but also when traversing via the keyboard. We are already setting `active` to follow the keyboard interaction. We just need a way to indicate that we want an indicator for this active state that mirrors the hover indicator.
If I've got that right ...
Thinking about this purely from the context of Button ...
I would expect that we would use `theme.global.active.background` because we want the `active` indicator styling. `hoverIndicator` can be set to `false` disable hover styling. I would expect `activeIndicator` to be set to `false` to disable the `active` styling. I think it's the explicit connection of `active` with `hover` via `getHoverIndicatorStyle()` in StyledButton when `activeIndicator` is set that seems off to me. A Button can be `active` independently of hovering. So, we want to be able to style `active` and `hover` independently.
Some suggestions ...
`<Button activeIndicator="hover" />` would indicate that the `active` indication should match the `hover` indication.
OR
Put the logic for getting the style from the theme in SelectContainer and pass this through to `activeIndicator` with StyledButton just adding it.
OR
Move the logic into SelectOption where it would just check for `active` in the props and then do the `getHoverIndicatorStyle()`
I'm leaning towards this last one at the moment.

taysea: The last one seems reasonable to me -- that would remove the need to touch Button I think.

ericsoderberghp (approved):

PR #6245, opened 2022-07-21
fix spinner size by taking off pad when to large for size

### LINKED ISSUE #6210: Spinner component 'size' prop doesn't always seem to work as expected (xsmall and arbitrary string size has no affect)
### Expected Behavior

Based on component documentation, the spinner 'size' prop should accept 'xsmall', 'small', 'medium', 'large', 'xlarge' and an arbitrary string like "16px" for size.

### Actual Behavior

It seems that 'xsmall' and an arbitrary string like "16px" does not always work. For whatever reason, this is 100% reproducible for me on all browsers, scenarios. 

Yet, it seems intermittent or perhaps environment/browser based though, as 'xsmall' seems to work for some (worked for Taylor and a few of my devs) but not others (fails 100% for some of my devs too). 

The arbitrary string size doesn't seem to work at all.

I can see this easily on your own sandbox here:

https://storybook.grommet.io/?path=/story/visualizations-spinner-size--size&globals=theme:hpe

... you can see that 'xsmall' is rendered the same as 'small' for me:

Screen Shot 2022-06-28 at 8 44 19 AM

### Steps to Reproduce

https://storybook.grommet.io/?path=/story/visualizations-spinner-size--size&globals=theme:hpe

### Your Environment

   "grommet": "^2.24.0",
        "grommet-icons": "^4.7.0",
        "grommet-theme-hpe": "^3.1.0",

- Browser Name and version: I tried Chrome, Safari, and Vivaldi ... all exhibit the issue.
- Operating System and version (desktop or mobile): Mac OS

britt6612:
#### What does this PR do?
This PR checks to see if the `pad` plus the `border` is larger than the intended size for the spinner and if it is then do not add `pad` so the theme sizes stay true to the size of the spinner.

spinner.js
storybook 
storybook 

Screen Shot 2022-07-21 at 2 12 34 PM
In the image above you can see that `xsmall` and `small` are very close in size with each other. 
The reasoning behind this is because we have `pad` set in our theme to be `small` 
for the `xsmall` you can see that we have a `pad` of `12px` and a border of `1px` so together we have a `box` that is 26px X 26px this is not true to our size in which we have 
`xsmall` to be 18px stated in our theme. 

```
      size: {
        xsmall: `${baseSpacing * 0.75}px`,
        small: `${baseSpacing}px`, // default 24
        medium: `${baseSpacing * 2}px`,
        large: `${baseSpacing * 3}px`,
        xlarge: `${baseSpacing * 4}px`,
      },

```

Screen Shot 2022-07-21 at 2 13 59 PM

Screen Shot 2022-07-21 at 2 14 59 PM

The same issue is also happening to make our `small` be 28 X 28 instead of our default of 24px 

I realize this would not be backwards compatible because it may change the size in which teams are using but I also think that this is a bug since it does not represent our theme sizes. 

Another issue I found was when `size` was not being used as a prop there was inconsistency with the border size. For instance if you use Spinner with `size = 'small'` you get a border of `2px` but if you use Spinner without passing `small` (which is our default) the border size is `1px`  This should be consistent so I added a check if `size` is not given we will assume it is the default `small` 

Screen Shot 2022-07-21 at 2 21 46 PM

closes #6210 
maybe
yes
read above

### DIFF
--- a/src/js/components/Spinner/Spinner.js
+++ b/src/js/components/Spinner/Spinner.js
@@ -61,9 +61,9 @@ const Spinner = forwardRef(
       {
         side: 'all',
         color: 'background-contrast',
-        size: size === undefined ? 'small' : size,
+        size: normalizedSize,
       },
-      { side: 'top', color, size: size === undefined ? 'small' : size },
+      { side: 'top', color, size: normalizedSize },
     ];
     const spinnerBorder = Array.isArray(borderThemeProp)
       ? borderThemeProp.map((borderSide) => ({
@@ -76,24 +76,17 @@ const Spinner = forwardRef(
       : borderThemeProp;
 
     // check the size of the pad plus border
+    // than multiple by 2 in order to compare to it to theme sizes
     let spinnerPadBorder;
     if (!rest.border) {
-      if (size !== undefined) {
-        spinnerPadBorder =
+      spinnerPadBorder =
+        2 *
+        (parseMetricToNum(
+          theme?.global?.edgeSize && theme.global.edgeSize[spinnerPad],
+        ) +
           parseMetricToNum(
-            theme?.global?.edgeSize && theme.global.edgeSize[spinnerPad],
-          ) +
-          parseMetricToNum(
-            theme?.global?.edgeSize && theme.global.borderSize[size],
-          );
-      } else
-        spinnerPadBorder =
-          parseMetricToNum(
-            theme?.global?.edgeSize && theme.global.edgeSize[spinnerPad],
-          ) +
-          parseMetricToNum(
-            theme?.global?.edgeSize && theme.global.borderSize.small,
-          );
+            theme?.global?.edgeSize && theme.global.borderSize[normalizedSize],
+          ));
     }
 
     // children will take prsecedence over theme attributes
@@ -128,7 +121,7 @@ const Spinner = forwardRef(
         {...themeProps}
         // If border plus pad is larger than spinnerSize there should be no pad.
         pad={
-          spinnerPadBorder * 2 > spinnerSize.replace(/px/, '') * 1
+          spinnerPadBorder > spinnerSize.replace('px', '')
             ? 'none'
             : theme.spinner.container.pad
         }
--- a/src/js/components/Spinner/Spinner.js
+++ b/src/js/components/Spinner/Spinner.js
@@ -10,6 +10,7 @@ import { AnnounceContext } from '../../contexts/AnnounceContext';
 import { Box } from '../Box';
 import { defaultProps } from '../../default-props';
 import { SpinnerPropTypes } from './propTypes';
+import { parseMetricToNum } from '../../utils';
 
 const BasicSpinner = ({ ref, size, ...rest }) => (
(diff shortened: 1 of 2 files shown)


src/js/components/Spinner/Spinner.js
```diff
@@ -68,7 +75,28 @@ const Spinner = forwardRef(
         }))
       : borderThemeProp;
 
-    // children will take precedence over theme attributes
+    // check the size of the pad plus border
+    let spinnerPadBorder;
+    if (!rest.border) {
+      if (size !== undefined) {
```
taysea, line 81: same as above, we can simplify this logic if size defaults to `small` -- it seems we document it as that (https://v2.grommet.io/spinner#size) so i'd consider it a bug we're fixing

src/js/components/Spinner/Spinner.js
```diff
@@ -98,6 +126,12 @@ const Spinner = forwardRef(
           typeof borderThemeProp === 'undefined' ? defaultBorder : spinnerBorder
         }
         {...themeProps}
+        // If border plus pad is larger than spinnerSize there should be no pad.
+        pad={
+          spinnerPadBorder * 2 > spinnerSize.replace(/px/, '') * 1
```
taysea, line 131:
```suggestion
          spinnerPadBorder * 2 > spinnerSize.replace('px', '')
```

taysea: Should we just include the `*2` above where we calculate `spinnerPadBorder`? and add a comment about why we multiple by 2.

src/js/components/Spinner/Spinner.js
britt6612: agree I can do that just wanted to get thoughts on this direction

src/js/components/Spinner/Spinner.js
britt6612: yes I can also add

src/js/components/Spinner/Spinner.js
```diff
@@ -68,7 +75,21 @@ const Spinner = forwardRef(
...
+    // than multiple by 2 in order to compare to it to theme sizes
+    let spinnerPadBorder;
+    if (!rest.border) {
+      spinnerPadBorder =
+        2 *
+        (parseMetricToNum(
+          theme?.global?.edgeSize && theme.global.edgeSize[spinnerPad],
+        ) +
+          parseMetricToNum(
+            theme?.global?.edgeSize && theme.global.borderSize[normalizedSize],
```
taysea, line 88: I'm not follow why we say if `!rest.border` --> shouldn't we be always be adding the border width, whether it's coming from a prop or the theme?

britt6612: think I need a separate check if `rest.border` then map through and get that  `size`

britt6612: so I added a check if someone has border size set in their  theme similar to how hpe theme works but if  someone passes  in `border` as a prop we should just let the `pad` be

britt6612:
closing this PR going to set `pad: 'none'` for hpe theme

PR #6250, opened 2022-07-25
[Meter] add ability to reverse the circle meter's direction

callumfrance:
#### What does this PR do?
this PR adds a new prop to the `Meter` component so that its `Circle` variant can have its direction reversed.

snapshots updated; visual inspection using `storybook`

updated storybook too

#### Screenshot
Screen Shot 2022-07-25 at 5 23 11 pm

yes, i have not done this

i think so, idk

should be; the prop is optional


ericsoderberghp:
Can you provide a brief example illustrating when this would be desirable? That would help me understand how to evaluate it and how to document it clearly.

callumfrance:
> Can you provide a brief example illustrating when this would be desirable? That would help me understand how to evaluate it and how to document it clearly.

hey @ericsoderberghp 

The intention behind adding this in is to provide more customisation options to the meter for the purposes of individualised branding and theming. 
i.e. I could see this being something that you set at the theme level to customise how all of your circle meters look.

That is essentially why I created the PR - because I am working with a style guide that has circle meters going anti-clockwise like this, and it would be neat to have this is in place within Grommet.

ericsoderberghp:
Thanks for the additional context. Given that, I think we should first allow controlling it via the theme. Also, I'm wondering if it would make sense for `bar` Meters to also reverse, especially since semicircle ones are as well.

I'm wanting to discuss the behavior a bit more first before delving into the code details.

callumfrance:
it would also be cool to discuss a vertical variant of the `RangeInput` e.g. to fit vertically next to the y-axis of a data visualisation

ericsoderberghp:
A bit more detail on what I'd like to solidify first ...
I think we should add a theme property to control the direction, such as `theme.meter.direction`, which would apply to all types of Meter. Does this address the desired behavior?

callumfrance:
@ericsoderberghp agreed, that's probably a better long-term solution

jcfilben:
Closing due to inactivity feel free to re-open

PR #6256, opened 2022-07-28
feat(tip): add controlled version

buberdds:
#### What does this PR do?

Controlled version of Tip

Tip.js

Manual and Jest

By checking all Tip stories

Tip used to provide a feedback based on interaction with app not mouse events.
Sample use case (included in a new story). I have a "Copy to Clipboard" button. When user clicks it, a 3rd party lib put a string into clipboard. Tip is opened to provide a feedback about successful action.

Yes

Yes

It is backwards compatible

### DIFF
--- a/src/js/components/Tip/Tip.js
+++ b/src/js/components/Tip/Tip.js
@@ -12,75 +12,92 @@ import { Drop } from '../Drop';
 import { useForwardedRef, useKeyboard } from '../../utils';
 import { TipPropTypes } from './propTypes';
 
-const Tip = forwardRef(({ children, content, dropProps, plain }, tipRef) => {
-  const theme = useContext(ThemeContext);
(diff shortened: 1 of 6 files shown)


jcfilben:
Hi @buberdds so based on your use case it sounds like the situation you have might be better for using `DropButton` instead of `Tip`. I'm assuming you only want the feedback about a successful action to show when the button is clicked or selected with the keyboard, not on mouse over or when the button has focus. This makes me think that `Tip` isn't the best solution here and instead you could accomplish this with `DropButton`. I put together this [codesandbox](https://codesandbox.io/s/grommet-v2-template-forked-f6bujt?file=/index.js) example that uses `DropButton`. Let me know what you think.

buberdds:
Ok, I will explore that option, thx.

Issue #6263, opened 2022-08-06
`RadioButton` Custom Icon not passed `theme` as a prop


GFargo:
### Expected Behavior

Similar to the [custom _checked_ icon ](https://github.com/grommet/grommet/blob/master/src/js/components/CheckBox/CheckBox.js#L139)within the CheckBox component, the `StyleRadioButtonIcon` should be [passed the `theme` as a prop here](https://github.com/grommet/grommet/blob/master/src/js/components/RadioButton/RadioButton.js#L103).

The `ThemeType` type should likely also match the `any` value being used by `checkBox.icons.checked`.

### Actual Behavior

Instead there is both a type error within `ThemeType` & type error when trying to pass a custom icon via `radioButton.icons.circle` due to `icon` being undefined.

### URL, screen shot, or Codepen exhibiting the issue

<img width="679" alt="Screen Shot 2022-08-06 at 5 31 33 PM" src="https://user-images.githubusercontent.com/3642037/183266658-44af7f3c-0b80-4274-877f-fc6ccd144265.png">

[Demo of issue](https://codesandbox.io/s/grommet-v2-issue-with-using-custom-circle-icon-in-radiobutton-1nfhcn)

### Steps to Reproduce

1. Create custom theme for `radioButton.icons.circle` that includes a grommet icon, like you would for `checkBox.icons.checked`

e.g. 

```
// Works 🎉 
checkBox: {
  icons: {
    checked: Gremlin,
  }
},

// Doesn't work 😢 
radioButton: {
  icons: {
    circle: Gremlin,
  }
}  
```

2. Pass theme to any component trying to render the grommet `RadioButton` component.

### Your Environment

- Grommet version: `2.25.1`
- Browser Name and version: Chrome Version 103.0.5060.134 (Official Build) (arm64)
- Operating System and version (desktop or mobile): Desktop - MacOS `12.5`

jcfilben:
Closing since #6264 has been merged

PR #6264, opened 2022-08-06
fix: pass `theme` as prop into StyledRadioButtonIcon component

GFargo:
#### What does this PR do?

Resolves issues with `RadioButton` component custom circle icon not being passed `theme` as a prop.

This PR attempts to match what is being done in the `CheckBox` component [here](https://github.com/grommet/grommet/blob/master/src/js/components/CheckBox/CheckBox.js#L139) however, within the `RadioButton` component `theme` is not passed into the `Icon` component [here](https://github.com/grommet/grommet/blob/master/src/js/components/RadioButton/RadioButton.js#L103).

In addition, the `any` type within ThemeType for [checkBox.icons.checked](https://github.com/grommet/grommet/blob/master/src/js/themes/base.d.ts#L589) has been carried over to replace the existing `string` type for [radioButton.icon.circle](https://github.com/grommet/grommet/blob/master/src/js/themes/base.d.ts#L1260).

Basic tests have been added for both `RadioButton` and `CheckBox` components to verify custom icons are properly rendered.

[This codesandbox demo](https://codesandbox.io/s/grommet-v2-issue-with-using-custom-circle-icon-in-radiobutton-1nfhcn?file=/index.js) can be forked to verify fix.

#6263

No

No breaking changes


src/js/themes/base.d.ts
```diff
@@ -368,11 +368,12 @@ export interface ThemeType {
         color?: ColorType;
         size?: string;
       };
-      shadow?: 
-        string | {
-        color?: ColorType;
-        size?: string;
-      };
+      shadow?:
```
GFargo, line 371: This adjustment to formatting was performed automatically via pre-commit hooks.

src/js/themes/base.d.ts
```diff
@@ -1257,7 +1258,7 @@ export interface ThemeType {
       size?: string;
     };
     icons?: {
-      circle?: string;
+      circle?: any;
```
GFargo, line 1261: Eventually it likely makes sense to provide stronger typing for this and things like `checkBox.icons.checked`...

Appears as though there are a few variants inside of the `ThemeType` currently e.g.

```typescript
JSX.Element | string
```
[here](https://github.com/grommet/grommet/blob/master/src/js/themes/base.d.ts#L242)

```typescript
React.ReactNode | Icon
```
[here](https://github.com/grommet/grommet/blob/master/src/js/themes/base.d.ts#L1011-L1014)

```typescript
ReactComponentElement<any>;
```
[here](https://github.com/grommet/grommet/blob/master/src/js/themes/base.d.ts#L1363)


Considering something like `React.ReactNode | Icon` or `React.ReactNode | Icon | string` would allow for both flexibility in implementation as well as more informative type hinting for developers.

jcfilben: I think we can go ahead and use `React.ReactNode | Icon` as our type here

jcfilben (commented):
Thanks for the detailed description and codesandbox example! I just added a minor comment about the typescript type definition. Everything else looks good

src/js/themes/base.d.ts
GFargo: Made these updates @jcfilben, thank you!

jcfilben:
FYI I updated the readme just to trigger a circleci run. Also the readme for RadioButton had outdated content.

jcfilben (approved):
Looks good!

src/js/themes/base.d.ts
ShimiSun: @jcfilben what is the consideration for dropping the type of `string`? wouldn't that be important for backward compatibility?

jcfilben: What `string` value could be passed here? I was thinking we could remove the `string` type here because I wasn't sure that there would be any strings that could be passed here.

GFargo: When playing around with this previously during my debugging I noticed the only strings that appear to work are names of HTML tags e.g. `span`, `div` etc..

However when these are rendered in the DOM they are empty tags and are invisible to the user.  We certainly could keep `string` here to maintain backward compatibility.   That said, anyone who is currently using this is likely writing custom CSS styles to manually insert an icon via `background-image` or something similar.

This in combination that the CSS classes generated by styled components are often unpredictable and done during build/runtime I have a hard time believing anyone is currently using this approach 💭

jcfilben: Thanks for the description, this is helpful. I think since someone could pass a string here even though it may be unlikely, we can keep `string` for backwards compatibility. So `React.ReactNode | Icon | string`

ericsoderberghp (approved):

jcfilben:
@GFargo Thanks for contributing to Grommet!

PR #6268, opened 2022-08-11
Adding testcafe

britt6612:
#### What does this PR do?
This PR adds the base work for the e2e testing files to Grommet. 
We can now run `yarn start` that will bring up another local host from which we will have our testing located. 
We can run `yarn test-e2e` which will run the test cafe tests that we will create 
in the app folder
locally
locally 

Also added packages that we will need for testcafe for example 
`@testing-library/testcafe`
in which we can use the best practices to get elements by using screen for querying

closes https://github.com/grommet/hpe-design-system/issues/2780
no
maybe
 backwards compatible


britt6612:
Adding a command to run localhost and run tests..

jcfilben (commented):
We should exclude the app folder from the distribution (See `build` command line 31 of package.json)

ericsoderberghp:
I'm wondering whether the top level directory should be "app" or "test". "test" might be more indicative of its purpose and prevent us from feeling like its main goal is to be a well rounded "app" example. This might give us more freedom to do oddball things that we want to test without feeling like they have to be good examples for building apps.

britt6612:
> I'm wondering whether the top level directory should be "app" or "test". "test" might be more indicative of its purpose and prevent us from feeling like its main goal is to be a well rounded "app" example. This might give us more freedom to do oddball things that we want to test without feeling like they have to be good examples for building apps.

Im wondering if maybe we should name it Grommet-Tests or component tests then move the `test` folder into that subfolder. That way we don't have 2 folders that are both tests

jcfilben:
> Im wondering if maybe we should name it Grommet-Tests or component tests then move the test folder into that subfolder. That way we don't have 2 folders that are both tests

Maybe something in the name to signify that this space if for end to end testing as apposed to unit testing.

taysea:
> > Im wondering if maybe we should name it Grommet-Tests or component tests then move the test folder into that subfolder. That way we don't have 2 folders that are both tests
> 
> Maybe something in the name to signify that this space if for end to end testing as apposed to unit testing.

`/e2e` ?

ericsoderberghp:
> > > Im wondering if maybe we should name it Grommet-Tests or component tests then move the test folder into that subfolder. That way we don't have 2 folders that are both tests
> > 
> > 
> > Maybe something in the name to signify that this space if for end to end testing as apposed to unit testing.
> 
> `/e2e` ?

would work well with `yarn start-e2e` and `yarn test-e2e`

halocline:
> `/e2e` ?

☝️ was my initial reaction as well.

Would the directory structure then be? (e.g. app and tests nested under e2e):
/coverage
/dist
/e2e
  /app
  /tests
...
/src
/storybook
/tools

ericsoderberghp:
> > `/e2e` ?
> 
> ☝️ was my initial reaction as well.
> 
> Would the directory structure then be? (e.g. app and tests nested under e2e): /coverage /dist /e2e -- /app -- /tests ... /src /storybook /tools

I was thinking paralleling a typical app:
/e2e
-- /public
-- /src
-- /webpack.config.js
-- /test

ericsoderberghp:
Can the e2e test configuration be inside "app", since they go together?

ericsoderberghp (approved):
Seems like a reasonable place to start.

jcfilben (approved):
Looks good!

Issue #6260, opened 2022-08-01
MultiSelect Component


jcfilben:
Create a new MultiSelect component the supports the default version design. Related to https://github.com/grommet/hpe-design-system/issues/2746

## Deliverables
- [ ] draft pull request, ready for feedback

Update 8/5:
This week I worked on building a new MultiSelect component that supports both the default and variant versions of the design. Next week I will be focusing on code consolidation between Select and MultiSelect as well as ensuring support for different types of options that can be passed, custom children, etc.

Update 8/22:
Last week I finished the indes.d.ts and propTypes.js files, added unit testing, and consolidated common code into a utils file. This week I am working on breaking up the component into smaller separate components (SelectValue, SelectClear, SelectOptions, SelectOption) and continue to add more unit tests.

PR #6275, opened 2022-08-12
SelectMultiple Component

jcfilben:
#### What does this PR do?
Adds the SelectMultiple component.

Added storybook stories

Note: still working on adding unit tests

Applicable [WCAG Rules](https://www.w3.org/WAI/WCAG21/quickref/?showtechniques=136%2C322) for this component:
- [ ] 1.3.5 Identify Input Purpose (AA)
- [ ] 1.3.6 Identify Purpose (AAA)
- [ ] 1.4.1 Use of Color (A)
- [ ] 1.4.3 Contrast (Minimum) (AA)
- [ ] 1.4.4 Resize Text (AA)
- [ ] 1.4.6 Contrast (Enhanced) (AAA)
- [ ] 1.4.10 Reflow (AA)
- [ ] 1.4.11 Non-text Contrast (AA)
- [ ] 1.4.12 Text Spacing (AA)
- [ ] 2.1.1 Keyboard (A)
- [ ] 2.1.2 No Keyboard Trap (A)
- [ ] 2.1.3 Keyboard (No Exception) (AAA)
- [ ] 2.4.7 Focus Visible (AA)
- [ ] 3.2.1 On Focus (A)
- [ ] 4.1.1 Parsing (A)
- [ ] 4.1.2 Name, Role, Value (A)
- [ ] 4.1.3 Status Message (AA)

Closes https://github.com/grommet/grommet/issues/6260
Yes
Yes
Backwards compatible


ericsoderberghp (commented):
A few things to start with.

I'm hoping we can try to improve on the complexity currently in Select. For example, requiring `value` to always be an Array. There's also a few of `(...) : (...)` when rendering where each part is rather large. I'm wondering if there's a way to simplify by splitting things up a bit. Just thinking out loud at the moment.

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,687 @@
...
+    // selected option object.
+    // Otherwise, the value(s) should match the selected options.
+
+    const [value, setValue] = formContext.useFormInput({
+      name,
+      value: valueProp,
+      initialValue: defaultValue || '',
+    });
+
+    const valuedValue = useMemo(
```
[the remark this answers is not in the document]
jcfilben: Added a comment to explain in latest commit

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,687 @@
...
+    // to be referenced when filtered by search.
+    useEffect(() => {
+      if (!search) setAllOptions(optionsProp);
+    }, [optionsProp, search]);
+
+    // the option indexes present in the value
+    const optionIndexesInValue = useMemo(() => {
+      const result = [];
+      allOptions.forEach((option, index) => {
+        if (Array.isArray(valuedValue)) {
```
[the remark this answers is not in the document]
jcfilben: Yes, done in latest commit

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,687 @@
...
+          }
+          onChange(adjustedEvent);
+        }
+      },
+      [onChange, setValue, triggerChangeEvent],
+    );
+
+    const SelectIcon = getSelectIcon(icon, theme, open);
+
+    const optionLabel = useCallback(
```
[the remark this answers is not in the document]
jcfilben: Done in latest commit. Should I do this for getOptionLabel as well?

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,687 @@
...
+                      ) !== -1,
+                  )
+                  /* eslint-disable-next-line array-callback-return, 
+                  consistent-return */
+                  .map((i) => {
+                    const iValue =
+                      valueKey && valueKey.reduce ? applyKey(i, valueKey) : i;
+                    const indexOptions = allOptions.indexOf(i);
+                    const iLabel = optionLabel(indexOptions);
+                    if (value.indexOf(iValue) < 5) {
```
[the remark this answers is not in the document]
jcfilben: I moved some of the code here out into a separate function. (See `visibleValue()`)

src/js/components/MultiSelect/MultiSelect.js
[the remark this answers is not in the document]
ericsoderberghp: yep, there were a few in the same pattern

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,687 @@
...
+      >
+        {isValidElement(SelectIcon) ? (
+          SelectIcon
+        ) : (
+          <SelectIcon color={iconColor} size={size} />
+        )}
+      </Box>
+    );
+
+    const defaultSelectTextInput = (
```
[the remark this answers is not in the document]
jcfilben: I changed this in that latest commit

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,687 @@
...
+                        const optionActive = false;
+                        child = children(i, indexOptions, allOptions, {
+                          active: optionActive,
+                          disabled: optionDisabled,
+                          selected: optionSelected,
+                        });
+                      }
+
+                      return (
+                        <Button
```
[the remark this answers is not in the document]
jcfilben: I changed this button to be `SelectOption`. There are a few places we use SelectOption so I am currently looking to see if there is enough similarities in them make some props common for SelectOption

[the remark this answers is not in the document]
ericsoderberghp: Excellent! That's the kind of thing I am hoping we can do more of.

ericsoderberghp:
What about splitting it up into even more discrete components? For example, SelectContainer, SelectValue, SelectClear, SelectOptions, SelectOption. Where each has its own file and properties related solely to its function?

jcfilben:
> What about splitting it up into even more discrete components? For example, SelectContainer, SelectValue, SelectClear, SelectOptions, SelectOption. Where each has its own file and properties related solely to its function?

Sounds good I will try something like this out. Marking this PR as a draft in the meantime while I work on this

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,680 @@
...
+      options: optionsProp,
+      placeholder,
+      plain,
+      replace,
+      searchPlaceholder,
+      size,
+      value: valueProp,
+      valueKey,
+      valueLabel,
+      visibleSelection = false,
```
taysea, line 100:
```suggestion
      showSelected = false,
```
```suggestion
      showSelection = false,
```

throwing out a couple other ideas for this prop name

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,680 @@
...
+      disabled,
+      disabledKey,
+      dropAlign,
+      dropHeight,
+      dropProps,
+      dropTarget,
+      emptySearchMessage,
+      focusIndicator,
+      gridArea,
+      helpContent,
```
taysea, line 73:
```suggestion
     help,
```
would we want to call this `help` to align with prop concept in FormField?

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,680 @@
...
+      children,
+      defaultValue,
+      disabled,
+      disabledKey,
+      dropAlign,
+      dropHeight,
+      dropProps,
+      dropTarget,
+      emptySearchMessage,
+      focusIndicator,
```
taysea, line 71:
```suggestion
      focusIndicator, // internal only from FormField
```
this is just for internal use correct? but not something we would document or encourage

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,680 @@
...
+      onBlur,
+      onChange,
+      onClick,
+      onClose,
+      onFocus,
+      onKeyDown,
+      onMore,
+      onOpen,
+      onSearch,
+      open: propOpen,
```
taysea, line 90:
```suggestion
      open: openProp,
```
more aligned with our typical convention

src/js/components/MultiSelect/MultiSelect.js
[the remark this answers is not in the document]
taysea: proposing we call this `normalizedValue` or `adjustedValue` and adjust the function below to something like `normalizeValue`, `getNormalizedValue`, etc.

(comment was helpful to stay regardless)

src/js/components/MultiSelect/MultiSelect.js
taysea, line 90: When I open the MultiSelect with keyboard, I think I would expect the focus to be placed onto the first focusable element in the drop. Right now, when I open I don't see any keyboard focus

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,680 @@
...
+                  : `${iLabel} not selected`
+              }
+              aria-setsize={value.length}
+              aria-posinset={value.indexOf(iValue)}
+              aria-selected={value.includes(iValue)}
+              aria-disabled={isDisabled(indexOptions)}
+              plain
+              hoverIndicator
+              fill="horizontal"
+              tabIndex="0"
```
taysea, line 272: It would be nice if when the checkbox receives keyboard focus via tabbing that the active checkbox receives the hover treatment so I can still see where focus is. (not sure if this is the right line to be placing this comment).

Video below, I am "shift+tabbing" back through the active options but can't see any indication of where I am until I reach the search:

https://user-images.githubusercontent.com/12522275/185683005-391464be-347b-4f45-b31d-ff7f04eefdfb.mov

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,680 @@
...
+                    onSelectChange(event, {
+                      option: iValue,
+                      value: intermediate.filter((v) => v !== iValue),
+                    });
+                    if (index !== intermediate.length - 1) {
+                      let timer = 0;
+                      // if index is the last one visible
+                      // allow time for the next option to
+                      // become visible
+                      if (index === 4) timer = 200;
```
taysea, line 287: Why 4? This feels brittle potentially

taysea: I'm also not following why we need a timer -- I haven't seen this pattern before of using `getElementById` etc as we do below. Can you elaborate on the purpose of this?

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,680 @@
...
+                  .map((i) => visibleValue(i))}
+              {showA11yDiv && (
+                <Box
+                  height="0px"
+                  width="0px"
+                  overflow="hidden"
+                  // announce when an item is removed from selected options
+                  aria-live="assertive"
+                >
+                  <Text>{showA11yDiv}</Text>
```
taysea, line 398: If this is just for screenreader, can we remove the `Text`?
```suggestion
                  {showA11yDiv}
```

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,680 @@
...
+      visibleValue,
+    ]);
+
+    // After announcing set showA11yDiv to undefined so it won't
+    // be read out again
+    useEffect(() => {
+      if (showA11yDiv !== undefined) {
+        setTimeout(() => {
+          setShowA11yDiv(undefined);
+        }, 1000);
```
taysea, line 432: How did we choose 1000?

taysea (commented):
Nice work on this big component! I noticed there's a lot of considerations for accessibility, thank you!

I've dropped some initial comments on naming/some functionality things I noted. I'll probably have more questions as I can dig deeper into the code but these were just some first things.

src/js/components/MultiSelect/MultiSelect.js
jcfilben: Good suggestions, I like `showSelectedInline` I think it provides the most clarity to what the prop is doing.

src/js/components/MultiSelect/MultiSelect.js
jcfilben: Yes this is just for internal use

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,680 @@
...
+      placeholder,
+      plain,
+      replace,
+      searchPlaceholder,
+      size,
+      value: valueProp,
+      valueKey,
+      valueLabel,
+      visibleSelection = false,
+      width,
```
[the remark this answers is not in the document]
jcfilben: Fixed in latest commit

src/js/components/MultiSelect/MultiSelect.js
jcfilben: Fixed in latest commit. I added a useEffect that puts the focus on the first focusable item when the drop opens

src/js/components/MultiSelect/MultiSelect.js
jcfilben: Good idea, I changed this in the latest commit

src/js/components/MultiSelect/MultiSelect.js
jcfilben: 5 is the maximum number of items that should be shown at the input level based on the designs. I could pull this out as a prop though. For hpe I think we are always recommending this be set to 5, but I could see other users wanting to adjust. I'm unsure if I want to make this a regular prop or theme prop. Right now I'm leaning towards a regular prop but open to other opinions on this.

jcfilben: I have the timer because if our focus is on the last item within the visible selections at the input level but there are still more options selected that are not being shown (because we only show 5 at a time), when we remove the last visible option, we have to wait until the next option is rendered and then once it is visible I move the focus to that option.

src/js/components/MultiSelect/MultiSelect.js
jcfilben: Yes, removed `Text` in latest commit

src/js/components/MultiSelect/MultiSelect.js
jcfilben: Changed in latest commit

src/js/components/MultiSelect/MultiSelect.js
jcfilben: I picked it as a starting point, but after testing more with the screenreader I realized I can reduce the amount of time here

src/js/components/MultiSelect/MultiSelect.js
ericsoderberghp: I could see both a theme value as well as a property for this. I'd probably start with just the theme for now and only add the prop later down the road when a demand surface more.

src/js/components/MultiSelect/MultiSelectContainer.js
```diff
@@ -0,0 +1,640 @@
...
+                pad={{ top: 'xsmall', bottom: 'xsmall' }}
+                direction="row"
+                justify="between"
+                gap="small"
+                fill="horizontal"
+              >
+                {selectClearAll}
+              </Box>
+              <Button onClick={onClose} a11yTitle="Close Select">
+                <FormUp />
```
britt6612, line 436: any reason you didnt do 
`icon={<FormUp />}` within the Button? and take out the `FormUp` as a child?

src/js/components/MultiSelect/MultiSelectValue.js
```diff
@@ -0,0 +1,217 @@
...
+            key={iLabel}
+            id={`selected-${iValue}`}
+          >
+            {children && child ? (
+              child
+            ) : (
+              <CheckBox
+                disabled={iDisabled}
+                label={
+                  <Box alignSelf="center" width="100%" align="start">
```
britt6612, line 135: is the `width` needed?

src/js/components/MultiSelect/index.d.ts
```diff
@@ -0,0 +1,67 @@
...
+import {
+  A11yTitleType,
+  AlignSelfType,
+  GridAreaType,
+  MarginType,
+  PlaceHolderType,
+} from '../../utils';
+
+export interface MultiSelectProps {
+  a11yTitle?: A11yTitleType;
```
britt6612, line 12: Similar to PropTypes can we pull out the identical Types used by both select and mutli-select and then just import them so we only have to go to one place to change for both?

src/js/components/MultiSelect/propTypes.js
```diff
@@ -0,0 +1,99 @@
...
+    ...genericProps,
+    children: PropTypes.func,
+    defaultValue: PropTypes.arrayOf(
+      PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.object,
+        PropTypes.number,
+      ]),
+    ),
+    disabled: PropTypes.oneOfType([
```
britt6612, line 16: can we pull out the identical propTypes used by both select and mutli-select and then just import them so we only have to go to one place to change for both?

src/js/components/MultiSelect/stories/Children.js
```diff
@@ -0,0 +1,109 @@
+import React, { useState } from 'react';
```
britt6612, line 1: can we maybe change one of these stories to a typescript file?

src/js/components/Select/SelectContainer.js
```diff
@@ -137,33 +119,14 @@ const SelectContainer = forwardRef(
...
-            result = disabled.indexOf(index) !== -1;
-          } else {
-            const optionVal = optionValue(index);
-            result = disabled.indexOf(optionVal) !== -1;
-          }
-        }
-        return result;
-      },
-      [disabled, disabledKey, options, optionValue],
+      (index) => checkDisabled(index, disabled, disabledKey, options, valueKey),
```
britt6612, line 128: nice cleanup!

src/js/components/MultiSelect/stories/ObjectOptions.js
```diff
@@ -0,0 +1,64 @@
...
+  },
+];
+
+export const ObjectOptions = () => {
+  const [options, setOptions] = useState(objectOptions);
+
+  return (
+    // Uncomment <Grommet> lines when using outside of storybook
+    // <Grommet theme={...}>
+    <Box fill align="center" justify="start" pad="large">
```
britt6612, line 31:
```suggestion
    <Box fill gap='large' align="center" justify="start" pad="large">
```


can we add `gap` to match layout of other stories?

src/js/components/MultiSelect/MultiSelectContainer.js
```diff
@@ -0,0 +1,640 @@
...
+              </Button>
+            </Box>
+          ) : (
+            <Box
+              pad={{ horizontal: 'xsmall', top: 'xsmall', bottom: 'xsmall' }}
```
britt6612, line 441:
```suggestion
              pad={{ horizontal: 'xsmall', vertical: 'xsmall' }}
```

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,504 @@
...
+      dropProps,
+      dropContent,
+      theme,
+    };
+
+    const displaySelectIcon = SelectIcon && (
+      <Box
+        alignSelf="center"
+        margin={theme.select.icons.margin}
+        style={{ minWidth: 'auto' }}
```
britt6612, line 360: we have a minWidth prop on `Box` did that not work the same?
`width={{ min: 'auto' }}`

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,504 @@
...
+            alignSelf={alignSelf}
+            direction="row"
+            alignContent="start"
+            background={theme.select.background}
+            ref={selectBoxRef}
+            flex={false}
+            callerPlain={plain}
+            width={width}
+          >
+            <Box direction="column" width="100%">
```
britt6612, line 384: the column is the default direction can we just strip that out?

src/js/components/MultiSelect/MultiSelect.js
```diff
@@ -0,0 +1,504 @@
...
+                alignSelf="start"
+                {...dropButtonProps}
+                dropAlign={
+                  dropAlign || { top: 'top', right: 'right', left: 'left' }
+                }
+                dropTarget={dropTarget || selectBoxRef.current}
+              >
+                {selectValue || displayLabelKey ? (
+                  <>
+                    <Box direction="row" width="100%">
```
britt6612, line 396: do we need the `width`?

src/js/components/MultiSelect/MultiSelect.js
britt6612: not sure if `auto` is an option in grommet though

src/js/components/MultiSelect/MultiSelectContainer.js
```diff
@@ -0,0 +1,640 @@
...
+      if (value.length === disabledSelected) return true;
+      return false;
+    }, [value, allOptions, isDisabled, optionValue]);
+
+    const selectClearAll =
+      search === '' || search === undefined ? (
+        <>
+          <Box alignSelf="center">
+            {value.length === 0 ? (
+              <Text size="small">0 selected</Text>
```
britt6612, line 338: something to think about down the line but if someone wanted to change how this text looks should it be in the theme?

src/js/components/MultiSelect/MultiSelectContainer.js
```diff
@@ -0,0 +1,640 @@
...
+            ) : (
+              <Text size="small">
+                {value.length} selected of {options.length}
+              </Text>
+            )}
+          </Box>
+          <Box>
```
britt6612, line 345: why need this `<Box>` can we just use a `<>`?

src/js/components/MultiSelect/MultiSelectContainer.js
```diff
@@ -0,0 +1,640 @@
...
+        >
+          {showSelectedInline ? (
+            <Box
+              direction="row"
+              justify="between"
+              flex={false}
+              pad={{ horizontal: 'small', top: 'xsmall' }}
+            >
+              <Box
+                pad={{ top: 'xsmall', bottom: 'xsmall' }}
```
britt6612, line 427:
```suggestion
                pad={{ vertical: 'xsmall' }}
```

britt6612:
One thing I noticed on the Limited story if you click the select open and pick a few options then close it and reopen it the focus is on the "clear all" which is good and it tells you in voice over to clear all button but once you click `clear all` it's not really clear what you can do next? also the up and down arrows to not work.

src/js/components/MultiSelect/MultiSelect.js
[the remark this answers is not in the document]
jcfilben: Changed to `normalizedValue` and `getNormalizedValue`

src/js/components/MultiSelect/MultiSelectContainer.js
jcfilben: Switched to use `icon={<FormUp />}` in latest push

src/js/components/MultiSelect/MultiSelectValue.js
jcfilben: Nope, removed

src/js/components/MultiSelect/index.d.ts
jcfilben: Good idea, I pull out identical props in the PropTypes and index.d.ts files

src/js/components/MultiSelect/propTypes.js
jcfilben: done

src/js/components/MultiSelect/stories/Children.js
jcfilben: Yes, changed the Children story to be a typescript story

src/js/components/MultiSelect/stories/ObjectOptions.js
jcfilben: Yes, added

src/js/components/MultiSelect/MultiSelectContainer.js
jcfilben: Fixed

src/js/components/MultiSelect/MultiSelect.js
jcfilben: `width={{ min: 'auto' }}` does work, I wasn't thinking when I made the inline style. I switched it in latest push

src/js/components/MultiSelect/MultiSelect.js
jcfilben: Yes, fixed

src/js/components/MultiSelect/MultiSelect.js
jcfilben: Nope, removed

src/js/components/MultiSelect/MultiSelectContainer.js
jcfilben: Good note, I left this alone for now but definitely something we might want to consider later

src/js/components/MultiSelect/MultiSelectContainer.js
jcfilben: Fixed

src/js/components/MultiSelect/MultiSelectContainer.js
jcfilben: Fixed

jcfilben:
> One thing I noticed on the Limited story if you click the select open and pick a few options then close it and reopen it the focus is on the "clear all" which is good and it tells you in voice over to clear all button but once you click clear all it's not really clear what you can do next? also the up and down arrows to not work.

Fixed this in latest push so that when you clear all focus is moved to the first option

src/js/components/MultiSelect/MultiSelect.js
jcfilben: I realized for chrome we don't need a timeout, however safari requires a timeout for this so I left his at 1000. I can add a comment in the code if that would be helpful

src/js/components/MultiSelect/MultiSelect.js
jcfilben: Added `selectMultiple.visibleInline` as a theme prop

src/js/components/SelectMultiple/stories/Variant.js
```diff
@@ -0,0 +1,101 @@
...
+  'Grape',
+  'Melon',
+  'Strawberry',
+  'Kiwi',
+  'Mango',
+  'Raspberry',
+  'Rhubarb',
+];
+
+export const Variant = () => {
```
taysea, line 18: Instead of "Variant" can we call this story "showSelectedInline"

src/js/components/SelectMultiple/stories/Variant.js
```diff
@@ -0,0 +1,101 @@
+import React, { useState } from 'react';
+
+import { Box, SelectMultiple, Text } from 'grommet';
+
+const dummyOptions = [
```
taysea, line 5:
```suggestion
const defaultOptions = [
```

src/js/components/SelectMultiple/stories/Variant.js
```diff
@@ -0,0 +1,101 @@
...
+          setOpen(true);
+          let next = [...valueMultiple];
+          next.sort();
+          const sortedOptions = dummyOptions.filter((i) => !next.includes(i));
+          next = next.filter((i) => dummyOptions.includes(i));
+          sortedOptions.sort();
+          const sortedAllOptions = next.concat(sortedOptions);
+          setOptions(sortedAllOptions);
+        }}
+        onClose={() => {
```
taysea, line 64: I think I imagined that more of this sorting would be handled within the component itself. This feels like a lot of overhead to get the desired behavior consistently across teams.

I guess this could be handled as a separate enhancement, but wondering if there could be a prop like `sort` that is a boolean that would handle the onOpen/onClose sorting happening here.

jcfilben: Yeah I originally envisioned this being handled within the component but when I was implementing it I was feeling like it was making too many assumptions.
It assumes the data should be sorted alphabetically.
Any code that is based on the index of the option will not work correctly. (For example the `disabled` prop allows users to specify which options are disabled based on the index)
Custom children are passed the index of the option so there may be code in there that relies on the index of the option.

It would be nice to have a more convenient way to get the desired behavior, maybe we could have the `sort` prop but just warn that any behavior that utilizes the `index` of the option should not be used when `sort=true`?

ericsoderberghp: I would think the caller would control the sort order of the `options` up front and the SelectMultiple component would honor the relative ordering of those items when placing them in the selected portion.

src/js/components/SelectMultiple/stories/Variant.js
```diff
@@ -0,0 +1,101 @@
...
+          const escapedText = text.replace(/[-\\^$*+?.()|[\]{}]/g, '\\$&');
+
+          // Create the regular expression with modified value which
+          // handles escaping special characters. Without escaping special
+          // characters, errors will appear in the console
+          const exp = new RegExp(escapedText, 'i');
+
+          let next = [...valueMultiple];
+          next = next.filter((o) => exp.test(o));
+          let sortedOptions = dummyOptions.filter((i) => !next.includes(i));
```
ericsoderberghp, line 44: Why are these "sorted" options? They look like just "filtered" options.

src/js/components/SelectMultiple/stories/Variant.js
```diff
@@ -0,0 +1,101 @@
...
+          const sortedAllOptions = next.concat(sortedOptions);
+          setOptions(sortedAllOptions);
+        }}
+        value={valueMultiple}
+        open={open}
+        onOpen={() => {
+          // setValueMultiple(intermediateValue);
+          setOpen(true);
+          let next = [...valueMultiple];
+          next.sort();
```
ericsoderberghp, line 57: Why do we need to `sort()` when opening the drop down?

src/js/components/SelectMultiple/stories/Variant.js
```diff
@@ -0,0 +1,101 @@
...
+  'Melon',
+  'Strawberry',
+  'Kiwi',
+  'Mango',
+  'Raspberry',
+  'Rhubarb',
+];
+
+export const Variant = () => {
+  const [options, setOptions] = useState(dummyOptions.sort());
```
ericsoderberghp, line 19: Why do we need to sort these here? Can `defaultOptions` be sorted correctly to begin with?

src/js/components/SelectMultiple/stories/Variant.js
jcfilben: Yeah that makes sense. I added a `sort` prop in the latest commit. It sorts selected items at the top while respecting the original order of the `options` prop.

src/js/components/SelectMultiple/stories/Variant.js
jcfilben: No longer using this code since `sort` prop was added in latest commit

src/js/components/SelectMultiple/stories/Variant.js
jcfilben: No longer using this code since sort prop was added in latest commit

src/js/components/SelectMultiple/stories/Variant.js
jcfilben: No longer using this code since sort prop was added in latest commit

jcfilben:
> What about splitting it up into even more discrete components? For example, SelectContainer, SelectValue, SelectClear, SelectOptions, SelectOption. Where each has its own file and properties related solely to its function?

Update on this:
I moved some of the code out into separate files. I created a file `SelectMultipleValue.js` that handles displaying options at the input level and `SelectionSummary.js` which includes the code that handles the select all/clear all and summary of number of items selected. 

I left the SelectOption code within the MultiSelect container for now, I was unsure if I was moving too much code out of the MultiSelectContainer and potentially making it more difficult to understand because it would require looking through multiple files or if it would be helpful to move this code out. Open to thoughts/opinions on this.

jcfilben:
FYI planning to change name of `sort` prop to `separateSelected`, just ran out of time today

src/js/components/SelectMultiple/stories/Variant.js
jcfilben: Yup, done

src/js/components/SelectMultiple/stories/Variant.js
jcfilben: Done

src/js/components/Select/utils.js
```diff
@@ -1,7 +1,156 @@
...
+
 export const applyKey = (option, key) => {
   if (option === undefined) return undefined;
   if (typeof key === 'object') return applyKey(option, key.key);
   if (typeof key === 'function') return key(option);
   if (key !== undefined) return option[key];
   return option;
 };
+
+export const DefaultSelectTextInput = forwardRef(
```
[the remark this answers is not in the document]
jcfilben: Moved the components out of the utils file in latest commit

src/js/components/SelectMultiple/stories/typescript/Children.tsx
```diff
@@ -0,0 +1,109 @@
...
+    <Box pad="small" background={state.active ? 'active' : undefined}>
+      {option}
+    </Box>
+  );
+
+  return (
+    // Uncomment <Grommet> lines when using outside of storybook
+    // <Grommet theme={...}>
+    <Box fill align="center" pad="small">
+      <SelectMultiple
```
MikeKingdom, line 65: This example looks odd with the `0 of 10 selected` text when nothing is selected and wasn't clear, especially with the placeholder being the default text color

MikeKingdom (commented):
What does `separateSelected` do? It isn't obvious to me what effect it has.

src/js/components/Select/DefaultSelectTextInput.js
```diff
@@ -0,0 +1,33 @@
+import React, { forwardRef } from 'react';
+import { SelectTextInput } from './StyledSelect';
+
+export const DefaultSelectTextInput = forwardRef(
+  (
+    { a11yTitle, disabled, id, name, placeholder, value, size, theme, ...rest },
+    ref,
+  ) => (
+    <SelectTextInput
+      a11yTitle={a11yTitle}
```
ericsoderberghp, line 10: Can this just be part of `...rest`?

src/js/components/Select/DefaultSelectTextInput.js
```diff
@@ -0,0 +1,33 @@
...
+      // When Select is disabled, we want to show a default cursor
+      // but not have disabled styling come from TextInput
+      // Disabled can be a bool or an array of options to disable.
+      // We only want to disable the TextInput if the control
+      // button should be disabled which occurs when disabled
+      // equals true.
+      defaultCursor={disabled === true || undefined}
+      focusIndicator={false}
+      id={id ? `${id}__input` : undefined}
+      name={name}
```
ericsoderberghp, line 20: Can this just be part of `...rest`?

src/js/components/Select/DefaultSelectTextInput.js
```diff
@@ -0,0 +1,33 @@
...
+      ref={ref}
+      {...rest}
+      tabIndex="-1"
+      type="text"
+      placeholder={placeholder}
```
ericsoderberghp, line 25: Can this just be part of `...rest`?

src/js/components/Select/DefaultSelectTextInput.js
```diff
@@ -0,0 +1,33 @@
...
+      plain
+      readOnly
+      value={value}
```
ericsoderberghp, line 28: these too

src/js/components/SelectMultiple/stories/typescript/Children.tsx
```diff
@@ -0,0 +1,109 @@
...
+      <Box
+        align="center"
+        direction="row"
+        gap="xsmall"
+        pad={{ vertical: 'xsmall', horizontal: 'small' }}
+        margin="xsmall"
+        background="brand"
+        round="large"
+      >
+        <Text size="small">{season}</Text>
```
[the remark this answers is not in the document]
jcfilben: Fixed, changed the background color to 'light-3' instead of 'brand'

src/js/components/SelectMultiple/stories/typescript/Children.tsx
```diff
@@ -0,0 +1,109 @@
...
+                  console.log(option);
+                  return renderSeason(i);
+                })}
+              </>
+            ) : (
+              <Box
+                pad={{ vertical: 'xsmall', horizontal: 'small' }}
+                margin="xsmall"
+              >
+                Select Season
```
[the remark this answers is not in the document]
jcfilben: Good idea, I changed the text to 'text-weak' (same as placeholder)

src/js/components/SelectMultiple/SelectMultipleValue.js
```diff
@@ -0,0 +1,225 @@
...
+            // announce when an item is removed from selected options
+            aria-live="assertive"
+          >
+            {showA11yDiv}
+          </Box>
+        )}
+      </Box>
+      {value && value.length > theme.selectMultiple.visibleInline && (
+        <Box alignSelf="start">
+          <Button
```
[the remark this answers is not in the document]
jcfilben: I added some padding around the button

src/js/components/SelectMultiple/stories/typescript/Children.tsx
jcfilben: I changed this so that when nothing is selected the `0 out of x` doesn't show and instead we just see the placeholder or valueLabel

src/js/components/Select/DefaultSelectTextInput.js
jcfilben: Yes, fixed

src/js/components/Select/DefaultSelectTextInput.js
jcfilben: Fixed

src/js/components/Select/DefaultSelectTextInput.js
jcfilben: Fixed

src/js/components/Select/DefaultSelectTextInput.js
jcfilben: Fixed

src/js/components/SelectMultiple/index.d.ts
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import { BasicSelectProps } from '../Select';
+
+export interface SelectMultipleProps extends BasicSelectProps {
+  defaultValue?: (string | number | object)[];
+  help?: React.ReactNode;
```
taysea, line 6: Can this just be a string of text?

src/js/components/SelectMultiple/index.d.ts
```diff
@@ -0,0 +1,21 @@
...
+  limit?: number;
+  value?: (string | number | object)[];
+  showSelectedInline?: boolean;
+  separateSelected?: boolean;
```
taysea, line 10: `separateSelected` doesn't immediately make sense to me. Thinking out loud with some ideas:
- selectionsTop
- sortSelectedOnClose
- liftSelected
- groupSelected

Also, should the value default to `true` ? or will that cause issues

src/js/components/SelectMultiple/SelectMultipleContainer.js
```diff
@@ -0,0 +1,614 @@
...
+          }
+          return result;
+        };
+        for (let i = 0; i < options.length; i += 1) {
+          if (!isSelected(i) && !originallyDisabled(i)) {
+            newDisabled.push(options[i]);
+          }
+        }
+        if (usingKeyboard)
+          setShowA11yLimit('Selected. Maximum selection limit reached.');
```
taysea, line 303: This string message seems to read weirdly

taysea: Should it just be "Maximum selection limit reached."

src/js/components/SelectMultiple/SelectMultipleContainer.js
```diff
@@ -0,0 +1,614 @@
...
+      usingKeyboard,
+      valueKey,
+    ]);
+
+    // reset showA11yLimit after announcement is read
+    useEffect(() => {
+      if (showA11yLimit !== undefined) {
+        setTimeout(() => {
+          setShowA11yLimit(undefined);
+        }, 2000);
```
taysea, line 327:
```suggestion
        }, 2000); // value chosen based on length of a11yLimit message
```

src/js/components/SelectMultiple/SelectMultipleContainer.js
```diff
@@ -0,0 +1,614 @@
...
+          a11yTitle="Select dropdown"
+        >
+          {showSelectedInline ? (
+            <Box
+              direction="row"
+              justify="between"
+              flex={false}
+              pad={{ horizontal: 'small', top: 'xsmall' }}
+            >
+              <Box
```
taysea, line 354: This whole return block feels a bit repetitive -- should this Box be built into SelectionSummary? Then depending on if `showSelectedInline` or not, the Box props could be altered. Seems like since SelectionSummary should present in a row, that the box defining the row should be part of it.

taysea:
```
let inlineContent = (
           <SelectionSummary
                  allOptions={allOptions}
                  clearRef={clearRef}
                  disabled={disabled}
                  disabledKey={disabledKey}
                  isSelected={isSelected}
                  labelKey={labelKey}
                  limit={limit}
                  onChange={onChange}
                  options={options}
                  search={search}
                  setActiveIndex={setActiveIndex}
                  value={value}
                  valueKey={valueKey}
                  boxProps={showSelectedInline 
                      ? { pad: { vertical: 'xsmall' },
                           direction: "row",
                           justify: "between",
                           gap: "small",
                           fill: "horizontal" }
                       : { .... }
                    }
             />
);

if (showSelectedInline) inlineContent = (
         <Box
              direction="row"
              justify="between"
              flex={false}
              pad={{ horizontal: 'small', top: 'xsmall' }}
            >
               {inlineContent}
               <Button
                icon={<FormUp />}
                onClick={onClose}
                a11yTitle="Close Select"
                plain
              />
         </Box>
);


... 

return ( ....
   <StyledContainer>
         {inlineContent}
         ....

File: src/js/components/SelectMultiple/SelectMultipleContainer.js [hunk 50]
```diff
@@ -0,0 +1,614 @@
...
+                  const optionDisabled = checkDisabled(
+                    index,
+                    disabled,
+                    disabledKey,
+                    options,
+                    valueKey || labelKey,
+                  );
+                  const optionSelected = value.includes(iValue);
+                  const optionActive = activeIndex === index;
+                  const iLabel = getOptionLabel(
```
taysea, line 465:
```suggestion
                  const optionLabel = getOptionLabel(
```
feel like this is a little more aligned w grommet conventions

src/js/components/SelectMultiple/SelectMultipleContainer.js
```diff
@@ -0,0 +1,614 @@
...
+            {options.length > 0 ? (
+              <InfiniteScroll
+                items={options}
+                step={theme.select.step}
+                onMore={onMore}
+                replace={replace}
+                show={activeIndex !== -1 ? activeIndex : undefined}
+              >
+                {(option, index, optionRef) => {
+                  const iValue =
```
taysea, line 452:
```suggestion
                  const optionValue =
```
feel like this is a little more aligned w grommet conventions

src/js/components/SelectMultiple/SelectMultipleContainer.js
```diff
@@ -0,0 +1,614 @@
...
+                    );
+                  }
+
+                  if (!children && !theme.select.options && search) {
+                    if (
+                      typeof iLabel === typeof '' &&
+                      iLabel.toLowerCase().indexOf(search) >= 0
+                    ) {
+                      // code to bold search term in matching options
+                      const boldIndex = iLabel.toLowerCase().indexOf(search);
```
taysea, line 510: How do I test this functionality?

src/js/components/SelectMultiple/SelectMultipleValue.js
```diff
@@ -0,0 +1,225 @@
...
+                      );
+                      setShowA11yDiv(
+                        `Unselected ${iLabel}. Focus moved to 
+                          ${getOptionLabel(
+                            allOptions.indexOf(result),
+                            allOptions,
+                            labelKey || valueKey,
+                          )}`,
+                      );
+                    }, 200);
```
taysea, line 125: add comment for how this value was chosen?

src/js/components/SelectMultiple/SelectMultipleValue.js
```diff
@@ -0,0 +1,225 @@
...
+                  } else if (dropButtonRef.current)
+                    dropButtonRef.current.focus();
+                }
+              }
+            }}
+            key={iLabel}
+            id={`selected-${iValue}`}
+          >
+            {children && child ? (
```
taysea, line 134:
```suggestion
            {child ? (
```
Are children required in order for `child` to receive a value?

src/js/components/SelectMultiple/SelectMultiple.js
```diff
@@ -0,0 +1,529 @@
...
+  getDisplayLabelKey,
+} from '../Select/utils';
+import { DefaultSelectTextInput } from '../Select/DefaultSelectTextInput';
+import { MessageContext } from '../../contexts/MessageContext';
+import { SelectMultiplePropTypes } from './propTypes';
+
+const StyledSelectBox = styled(Box)`
+  ${(props) => !props.callerPlain && controlBorderStyle};
+  ${(props) =>
+    props.theme.select &&
```
ericsoderberghp, line 42: With the way `?.` works, do we need the intervening checks still? Can this just be `${(props) => props.theme.select?.control?.extend}`

src/js/components/SelectMultiple/SelectMultiple.js
```diff
@@ -0,0 +1,529 @@
...
+  getSelectIcon,
+  getIconColor,
+  getDisplayLabelKey,
+} from '../Select/utils';
+import { DefaultSelectTextInput } from '../Select/DefaultSelectTextInput';
+import { MessageContext } from '../../contexts/MessageContext';
+import { SelectMultiplePropTypes } from './propTypes';
+
+const StyledSelectBox = styled(Box)`
+  ${(props) => !props.callerPlain && controlBorderStyle};
```
ericsoderberghp, line 40: I realize this is inherited from prior Select, but I'm wondering if "plainSelect" might be a bit more descriptive rather than "callerPlain" as a property name.

src/js/components/SelectMultiple/SelectMultiple.js
```diff
@@ -0,0 +1,529 @@
...
+    // search input value
+    const [search, setSearch] = useState();
+    // All select option indices and values
+    const [allOptions, setAllOptions] = useState(optionsProp);
+    const [orderedOptions, setOrderedOptions] = useState();
+    // Track changes to options property, except when options are being
+    // updated due to search activity. Allows option's initial index value
+    // to be referenced when filtered by search.
+    useEffect(() => {
+      if (!search) setAllOptions(optionsProp);
```
ericsoderberghp, line 145: Do we always want to update `allOptions` when `separateSelected` changes?

src/js/components/SelectMultiple/SelectMultiple.js
```diff
@@ -0,0 +1,529 @@
...
+    }, [optionsProp, search, separateSelected]);
+
+    // the option indexes present in the value
+    const optionIndexesInValue = useMemo(() => {
+      const result = [];
+      allOptions.forEach((option, index) => {
+        if (
+          normalizedValue &&
+          normalizedValue.length > 0 &&
+          normalizedValue.some((v) => v === applyKey(option, valueKey))
```
ericsoderberghp, line 156: Can these all be simplified to: `normalizedValue?.some(...)`?

src/js/components/SelectMultiple/SelectMultiple.js
```diff
@@ -0,0 +1,529 @@
...
+    const onRequestOpen = useCallback(() => {
+      if (open) return;
+      setOpen(true);
+      if (onOpen) onOpen();
+    }, [onOpen, open]);
+
+    useEffect(() => {
+      if (separateSelected && ((open && search) || !open)) {
+        const selectedOptions = optionsProp.filter((option) => {
+          const iValue =
```
ericsoderberghp, line 176: suggestion: "iValue" -> "optionValue"

src/js/components/SelectMultiple/SelectMultiple.js
```diff
@@ -0,0 +1,529 @@
...
+            valueKey && valueKey.reduce ? applyKey(option, valueKey) : option;
+          return value.includes(iValue);
+        });
+        const unselectedOptions = optionsProp.filter(
+          (i) => !selectedOptions.includes(i),
+        );
+        const nextOrderedOptions = selectedOptions.concat(unselectedOptions);
+        setOrderedOptions(nextOrderedOptions);
```
ericsoderberghp, line 184: Should `orderedOptions` just be calculated via `useMemo()` when any dependencies change, as opposed to needing `useState()` + `useEffect()`?

src/js/components/SelectMultiple/index.d.ts
jcfilben: Yes, added `string` to the type definition

src/js/components/SelectMultiple/index.d.ts
jcfilben: I changed the prop name to `sortSelectedOnClose` and have it default to true

src/js/components/SelectMultiple/SelectMultipleContainer.js
jcfilben: Yeah so this message gets read when an option is selected and the number of selected options is now equal to the limit. I wanted to give some confirmation that the option was actually selected. I was thinking if the message just read 'Maximum selection limit reached' the user might be confused as to whether their option was selected or not.

src/js/components/SelectMultiple/index.d.ts
ericsoderberghp: My one concern with using "sort" is that if the caller has ordered the options to be: [A, C, B, E, D] and the user selects [C, B], the name "sortSelectedOnClose" implies to me that the component will covert the selected to [B, C].

src/js/components/SelectMultiple/SelectMultipleContainer.js
jcfilben: Done

src/js/components/SelectMultiple/SelectMultipleContainer.js
jcfilben: Good idea, I broke this out as suggested

src/js/components/SelectMultiple/SelectMultipleContainer.js
jcfilben: Yes, fixed in latest commit

src/js/components/SelectMultiple/SelectMultipleContainer.js
jcfilben: Yes, fixed in latest commit

src/js/components/SelectMultiple/SelectMultipleContainer.js
jcfilben: When you type in the search you should see substrings that match your search term bolded. I realized that this wasn't working on the grommet theme, but fixed this in the latest commit.

src/js/components/SelectMultiple/SelectMultipleValue.js
jcfilben: Done

src/js/components/SelectMultiple/SelectMultipleValue.js
jcfilben: Yes, I simplified this in the latest commit

src/js/components/SelectMultiple/SelectMultiple.js
jcfilben: Changed to `${(props) => props.theme.select?.control?.extend}` in latest commit

src/js/components/SelectMultiple/SelectMultiple.js
jcfilben: Changed to `plainSelect` in latest commit

src/js/components/SelectMultiple/SelectMultiple.js
jcfilben: No, I separated the `orderedOptions` code so that `allOpitions` doesn't get updated when `sortSelectedOnClose` (previously `separateSelected`) changes.

src/js/components/SelectMultiple/SelectMultiple.js
jcfilben: Done

src/js/components/SelectMultiple/SelectMultiple.js
jcfilben: I played around with using `useMemo` for this but wasn't getting the results I wanted. `orderedOptions` was getting reset to the regular list of options at the incorrect times so that the ordering wasn't reflecting in the list. I did tweak the useEffect slightly to tighten things up.

src/js/components/SelectMultiple/SelectMultiple.js
jcfilben: If I just have `normalizedValue?.some(...)` I get the error: `normalizedValue.some is not a function`. I modified this to `if (normalizedValue.length && normalizedValue?.some(...))`

ericsoderberghp: What about `normalizedValue?.some?.(...)`

src/js/components/Select/Select.js
```diff
@@ -262,22 +228,16 @@ const Select = forwardRef(
...
-      );
-      if (
-        !selectValue &&
-        optionIndexesInValue.length === 1 &&
-        typeof optionLabelKey === 'object'
-      )
-        return optionLabelKey;
-      return undefined;
-    }, [labelKey, allOptions, optionIndexesInValue, selectValue]);
+    const displayLabelKey = useMemo(
```
ericsoderberghp, line 231: As an example of custom hooks, this could be `useDisplayLabelKey(labelKey, allOptions, optionIndexesInValue, selectValue)`. Not required, just illustrating.

src/js/components/SelectMultiple/SelectMultiple.js
jcfilben: Yeah that works! I'll push up that change

src/js/components/Select/StyledSelect.js
```diff
@@ -1,18 +1,74 @@
...
       ? sizeStyle('max-height', props.dropHeight, props.theme)
       : 'max-height: inherit;'};
 
-  ${props =>
+  ${(props) =>
     props.theme.select.container && props.theme.select.container.extend};
 `;
+
+// position relative is so scroll can be managed correctly
+export const OptionsBox = styled.div`
```
ericsoderberghp, line 29: Since this isn't actually a "Box", can we call it something like "OptionsContainer"?

src/js/components/Select/Select.js
jcfilben: Good to know, thanks for the example

src/js/components/Select/StyledSelect.js
jcfilben: Yes, done

src/js/components/Select/SelectContainer.js
```diff
@@ -294,7 +244,13 @@ const SelectContainer = forwardRef(
         let nextActiveIndex = activeIndex + 1;
         while (
           nextActiveIndex < options.length &&
-          isDisabled(nextActiveIndex)
+          checkDisabled(
```
ericsoderberghp, line 247: Suggest "checkDisabled" -> "isDisabled", to align more with "isSelected"

src/js/components/Select/utils.js
```diff
@@ -7,3 +9,96 @@ export const applyKey = (option, key) => {
...
   return option;
 };
+
+export const getOptionLabel = (index, options, labelKey) =>
+  applyKey(options[index], labelKey);
+
+export const getOptionValue = (index, options, valueKey) =>
+  applyKey(options[index], valueKey);
+
+export const checkDisabled = (
```
ericsoderberghp, line 19: If you made this a custom hook, it might make its usage cleaner:
```
const isDisabled = useDisabled(disabled, disabledKey, options, valueKey);
...
if (isDisabled(nextActiveIndex) ...
```
where `useDisabled` would then simply be:
```
const useDisabled(disabled, disabledKey, options, valueKey) = useCallback((index) => {...}, [disabled, disabledKey, options, valueKey]);
```

ericsoderberghp: There's just a lot of lines of code for dependencies in the other useCallbacks() in SelectContainer.js

src/js/components/SelectMultiple/SelectMultipleContainer.js
```diff
@@ -0,0 +1,595 @@
...
+          const searchInput = searchRef.current;
+          if (searchInput && searchInput.focus) {
+            setFocusWithoutScroll(searchInput);
+          }
+        } else if (activeRef.current) {
+          setFocusWithoutScroll(activeRef.current);
+        } else if (optionsRef.current) {
+          setActiveIndex(0);
+        }
+      }, 100);
```
ericsoderberghp, line 100: Can you add a comment indicating why "100" was chosen?

src/js/components/SelectMultiple/SelectMultipleContainer.js
```diff
@@ -0,0 +1,595 @@
...
+                        tabIndex="-1"
+                        checked={optionSelected}
+                        disabled={optionDisabled}
+                      />
+                    );
+                  }
+
+                  if (!children && search) {
+                    if (
+                      typeof optionLabel === typeof '' &&
```
ericsoderberghp, line 482: Why not `typeof optionLabel === 'string'`?

src/js/themes/base.d.ts
```diff
@@ -1395,6 +1395,9 @@ export interface ThemeType {
     searchInput?: ReactComponentElement<any>;
     step?: number;
   };
+  selectMultiple?: {
+    visibleInline?: number;
```
ericsoderberghp, line 1399: Suggest "visibleInline" -> "maxInline" to be a little clearer that this is a maximum. I think "inline" implies "visible".

src/js/components/Select/SelectContainer.js
jcfilben: Done

src/js/components/Select/utils.js
jcfilben: Changed to use `useDisabled` hook

src/js/components/SelectMultiple/SelectMultipleContainer.js
jcfilben: Yes, done

src/js/components/SelectMultiple/SelectMultipleContainer.js
jcfilben: Changed to `typeof optionLabel === 'string'` in latest commit

src/js/themes/base.d.ts
jcfilben: Good idea, I changed this to `maxInline` in the latest commit

ericsoderberghp (approved):

PR #6277, opened 2022-08-13
(Spinner) fix: keep the correct size

buberdds:
#### What does this PR do?
Adds `min-width` css rule to Spinner to keep correct size. Usually, the issue is visible on mobile devices where content can squeeze Spinner component.

Spinner.js

Manual/snapshot

Enable Responsive Design mode in dev tools and validate Storybook

Spinner is not keeping its original size.

#### Screenshots (if a appropriate)
Screenshot 2022-08-13 at 11 42 36

with `min-width`
Screenshot 2022-08-13 at 11 53 16

No

No

backwards compatible

### DIFF
--- a/src/js/components/Spinner/Spinner.js
+++ b/src/js/components/Spinner/Spinner.js
@@ -12,7 +12,7 @@ import { defaultProps } from '../../default-props';
 import { SpinnerPropTypes } from './propTypes';
 
 const BasicSpinner = ({ ref, size, ...rest }) => (
-  <Box height={size} width={size} ref={ref} {...rest} />
+  <Box flex={false} height={size} width={size} ref={ref} {...rest} />
 );
 /**
  * If the user is calling <Spinner>…</Spinner> with children, it will take
--- a/src/js/components/Spinner/__tests__/__snapshots__/Spinner-test.tsx.snap
+++ b/src/js/components/Spinner/__tests__/__snapshots__/Spinner-test.tsx.snap
@@ -26,6 +26,9 @@ exports[`Spinner border renders 1`] = `
   flex-direction: column;
   height: 24px;
   width: 24px;
+  -webkit-flex: 0 0 auto;
+  -ms-flex: 0 0 auto;
+  flex: 0 0 auto;
   padding: 12px;
   border-radius: 100%;
   -webkit-transform: rotate(0deg);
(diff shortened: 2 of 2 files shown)


britt6612:
@buberdds Thank you for submitting this PR. Trying to understand the issue a little more is it the sizes of the spinner in which your adding the `min-width` also can you paste your code as to how you got these spinners? they look strange 
Screen Shot 2022-08-18 at 3 45 26 PM

buberdds:
They way I use Spinner is similar to storybook example. Steps to repro
- src/js/components/Spinner/stories/Size.js
- replace L~10
```
 <Text size={size}>{size}</Text>
```
with

```
        <Text size={size}>
          Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut id ante in nulla tincidunt sollicitudin.
        </Text>
```
- open dev tools and toggle device toolbar (mobile device preview mode)

britt6612:
@buberdds thank you I see the issue sorry I didn't realize this was on a smaller screen makes sense now. The `min-width` I agree with. Can we just change a couple of things? We try to avoid inline `styles` would you mind pulling it out?

```
const SpinnerBox = styled(Box)`
  min-width: ${(props) => props.width};
`;

const BasicSpinner = ({ ref, size, ...rest }) => (
  <SpinnerBox height={size} width={size} ref={ref} {...rest} />
);

```

jcfilben (approved):
Looks good!

src/js/components/Spinner/Spinner.js
```diff
@@ -4,15 +4,19 @@ import React, {
...
 import { defaultProps } from '../../default-props';
 import { SpinnerPropTypes } from './propTypes';
 
+const SpinnerBox = styled(Box)`
+  min-width: ${(props) => props.width};
+`;
+
 const BasicSpinner = ({ ref, size, ...rest }) => (
-  <Box height={size} width={size} ref={ref} {...rest} />
+  <SpinnerBox height={size} width={size} ref={ref} {...rest} />
```
ericsoderberghp, line 19: Any reason we didn't just use `<Box flex={false} />`?

buberdds: The main reason is that I am not fully familiar with API yet. Changing `flex-shrink` to `0` via `flex={false}` will work as well and without additional styled component 👍

britt6612: also didnt think about `flex={false}` Thanks Eric

ericsoderberghp (approved):

PR #6279, opened 2022-08-15
6273 pageheader size

### LINKED ISSUE #6273: PageHeader - Add size prop and theme options
### Expected Behavior

Allow caller to specify PageHeader `size` -- which would affect top/bottom pad, heading size, and subtitle size. These values should be able to be set in the theme `pageHeader.size`.

`pageHeader.size.small.title`, `pageHeader.size.small.subtitle`.. etc

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

taysea:
#### What does this PR do?

Adds `size` prop to PageHeader and related theming of `theme.pageHeader.size.small`, `theme.pageHeader.size.medium`, `theme.pageHeader.size.large` which can set title, subtitle, and pad values for each size.

Discussion:
- [ ] Are `small`, `medium`, and `large` the right sizes to support out of the box? Designs we have seen only really necessitate a "large" variant in additional to the default, but felt that "small" could be reasonable also.

PageHeader.js / base.js

Storybook and unit test.

Use "Size" PageHeader storybook example.

Closes #6273 

Yes.

Yes. Added `size` to PageHeader.

Backwards compatible.

### DIFF
--- a/src/js/components/PageHeader/PageHeader.js
+++ b/src/js/components/PageHeader/PageHeader.js
@@ -8,6 +8,11 @@ import { Grid } from '../Grid';
 import { Paragraph } from '../Paragraph';
 import { ResponsiveContext } from '../../contexts/ResponsiveContext';
 
+const sizeStyle = (size, style, theme) =>
+  size && theme.pageHeader.size[size]?.[style]
+    ? theme.pageHeader.size[size]?.[style]
+    : theme.pageHeader[style];
+
 const PageHeader = forwardRef(
   (
     {
@@ -15,6 +20,7 @@ const PageHeader = forwardRef(
       gridProps: gridPropsProp,
       parent,
       responsive,
+      size,
       subtitle,
       title,
       ...rest
@@ -45,7 +51,7 @@ const PageHeader = forwardRef(
         ref={ref}
         direction="column"
         gap="none"
-        pad={theme.pageHeader.pad}
+        pad={sizeStyle(size, 'pad', theme)}
         {...rest}
       >
         <Grid
@@ -61,14 +67,16 @@ const PageHeader = forwardRef(
           </Box>
           <Box gridArea="title">
             {typeof title === 'string' ? (
-              <Heading {...theme.pageHeader.title}>{title}</Heading>
+              <Heading {...sizeStyle(size, 'title', theme)}>{title}</Heading>
             ) : (
               title
             )}
           </Box>
           <Box gridArea="subtitle">
             {typeof subtitle === 'string' ? (
-              <Paragraph {...theme.pageHeader.subtitle}>{subtitle}</Paragraph>
+              <Paragraph {...sizeStyle(size, 'subtitle', theme)}>
+                {subtitle}
+              </Paragraph>
             ) : (
               subtitle
             )}
--- a/src/js/components/PageHeader/__tests__/PageHeader-test.tsx
+++ b/src/js/components/PageHeader/__tests__/PageHeader-test.tsx
@@ -26,6 +26,24 @@ describe('PageHeader', () => {
     expect(asFragment()).toMatchSnapshot();
   });
 
+  test('size', () => {
+    const { asFragment } = render(
+      <Grommet>
+        <PageHeader
+          title="Grommet"
+          subtitle={`Grommet helps you build responsive and accessible 
+          mobile-first projects for the web with an easy to use component 
+          library.`}
+          actions={<Button label="Get Started" primary />}
+          parent={<Anchor label="Parent Page" />}
+          size="large"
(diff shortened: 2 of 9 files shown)


src/js/components/PageHeader/PageHeader.js
```diff
@@ -8,13 +8,19 @@ import { Grid } from '../Grid';
 import { Paragraph } from '../Paragraph';
 import { ResponsiveContext } from '../../contexts/ResponsiveContext';
 
+const sizeStyle = (size, style, theme) =>
```
halocline, line 11: Minor, but for some reason I don't immediately associate `title`, `subtitle`, `pad` with `style`. Something like property, attribute, item, aspect, feature, etc. might feel more fitting.

```suggestion
const sizeStyle = (size, feature, theme) =>
```

halocline (approved):
`small`, `medium`, `large` seem reasonable to me. Scaling of the font sizes and padding look good.

Minor naming suggestion.

Will need follow on docs PR.

src/js/themes/base.js
```diff
@@ -1321,6 +1321,41 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
       },
+      size: {
+        small: {
+          pad: {
+            top: 'medium',
+            bottom: 'small',
+          },
+          subtitle: {
+            size: 'small',
+            margin: 'none',
```
ericsoderberghp, line 1332: Is `margin: 'none'` needed? I was expecting that "small" and "large" would baseline with the defaults and only adjust specific properties that were different.

src/js/components/PageHeader/stories/Size.js
```diff
@@ -0,0 +1,53 @@
...
+  Form,
+  FormField,
+  PageHeader,
+  Page,
+  PageContent,
+  Select,
+} from 'grommet';
+
+export const Size = () => {
+  const [size, setSize] = useState('medium');
```
ericsoderberghp, line 14: Wondering if we should always show all sizes, similar to how Heading:All and Paragraph:All work. I suggest this partly because Chromatic would detect any changes if we did that, which it wouldn't detect when requiring interaction.

src/js/components/PageHeader/__tests__/PageHeader-test.tsx
```diff
@@ -26,6 +26,24 @@ describe('PageHeader', () => {
...
+    const { asFragment } = render(
+      <Grommet>
+        <PageHeader
+          title="Grommet"
+          subtitle={`Grommet helps you build responsive and accessible 
+          mobile-first projects for the web with an easy to use component 
+          library.`}
+          actions={<Button label="Get Started" primary />}
+          parent={<Anchor label="Parent Page" />}
+          size="large"
```
ericsoderberghp, line 39: Can we do all sizes via unit test? That would cover "small" and explicitly setting "medium" as well.

src/js/components/PageHeader/PageHeader.js
```diff
@@ -8,13 +8,19 @@ import { Grid } from '../Grid';
...
+  size && theme.pageHeader.size[size]?.[style]
```
ericsoderberghp, line 12: This seems a bit fragile for cases like `pad: 0`. How about:
`(size && theme.pageHeader.size[size]?.[style]) ?? theme.pageHeader[style]`
or maybe, to not require replicating all properties:
`{ ...theme.pageHeader[style], ...((size && theme.pageHeader.size[size]?.[style]) ?? theme.pageHeader[style]) }`

src/js/components/PageHeader/__tests__/PageHeader-test.tsx
taysea: Yes, will update.

src/js/themes/base.js
taysea: That's fair. Let me update the logic.

src/js/components/PageHeader/PageHeader.js
taysea: Done.

src/js/components/PageHeader/stories/Size.js
taysea: Done.

src/js/components/PageHeader/__tests__/PageHeader-test.tsx
taysea: Done.

src/js/components/PageHeader/PageHeader.js
taysea: Done.

ericsoderberghp (approved):

PR #6280, opened 2022-08-16
Implement basic Analytics

MikeKingdom:
#### What does this PR do?
Implement the onAnalytics instrumentation in <Grommet>.

This is a follow on to https://github.com/grommet/hpe-design-system/issues/2714

```jsx
<Grommet onAnalytics={(evt) => { ... } } />
```
onAnalytics called with events in situations like:

- Form onSubmit and in a form
  - evt contains touched, the form element, number of failed validations per field, time spent in the form.
- Form open and close. This can help detect form cancel
- Layer open and close.
- Tabs activated
- Button onClick - send button element to track actions and other usages
- URL changes (e.g. pageViews.) Use the MutationObserver trick to track url changes and send events for page views.
- const sendAnalytics = useAnalytics() API available for external use (as well as internal)

Some example data sent to `onAnalytics`:
image

AnalyticsContext.js

closes https://github.com/grommet/hpe-design-system/issues/2818

Not yet

No

Backwards compatible

### DIFF
--- a/src/js/components/Button/Button.js
+++ b/src/js/components/Button/Button.js
@@ -178,11 +178,13 @@ const Button = forwardRef(
         sendAnalytics({
           type: 'buttonClick',
           element: event.target,
-          data: event,
+          event,
+          href,
+          label: typeof label === 'string' ? label: undefined,
         });
         if (onClickProp) onClickProp(event);
       },
-      [onClickProp, sendAnalytics],
+      [onClickProp, sendAnalytics, href, label],
     );
 
     // kindArg is object if we are referencing a theme object
--- a/src/js/components/Button/Button.js
+++ b/src/js/components/Button/Button.js
@@ -5,6 +5,7 @@ import React, {
   useContext,
   useMemo,
   useState,
+  useCallback,
 } from 'react';
 
 import { ThemeContext } from 'styled-components';
@@ -23,6 +24,7 @@ import { Tip } from '../Tip';
 import { Badge } from './Badge';
 import { StyledButton } from './StyledButton';
 import { StyledButtonKind } from './StyledButtonKind';
+import { useAnalytics } from '../../contexts/AnalyticsContext';
 
 // We have two Styled* components to separate
 // the newer default|primary|secondary approach,
@@ -140,7 +142,7 @@ const Button = forwardRef(
       kind: kindArg,
       label,
       onBlur,
-      onClick,
+      onClick: onClickProp,
       onFocus,
       onMouseOut,
       onMouseOver,
@@ -168,6 +170,20 @@ const Button = forwardRef(
         'Button should not have children if icon or label is provided',
       );
     }
+
+    const sendAnalytics = useAnalytics();
+
+    const onClick = useCallback(
+      (event) => {
+        sendAnalytics({
+          type: 'buttonClick',
+          element: event.target,
+          data: event,
+        });
+        if (onClickProp) onClickProp(event);
+      },
+      [onClickProp, sendAnalytics],
+    );
 
     // kindArg is object if we are referencing a theme object
     // outside of theme.button
@@ -375,7 +391,7 @@ const Button = forwardRef(
           href={href}
           kind={kind}
           themePaths={themePaths}
-          onClick={onClick}
+          onClick={onClickProp ? onClick : undefined}
           onFocus={(event) => {
             setFocus(true);
             if (onFocus) onFocus(event);
--- a/src/js/components/Form/Form.js
+++ b/src/js/components/Form/Form.js
@@ -361,12 +361,15 @@ const Form = forwardRef(
         element,
       });
       return () => {
-        sendAnalytics({
-          type: 'formClose',
-          element,
-          errors: analyticsRef.current.errors,
-          elapsed: new Date().getTime() - analyticsRef.current.start.getTime(),
-        });
+        if (!analyticsRef.current.submitted) {
+          sendAnalytics({
+            type: 'formClose',
+            element,
+            errors: analyticsRef.current.errors,
+            elapsed: 
+              new Date().getTime() - analyticsRef.current.start.getTime(),
+          });
+        }
       };
     }, [sendAnalytics, formRef]);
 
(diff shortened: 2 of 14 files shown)


halocline:
👀

halocline:
@MikeKingdom - Getting to look at this now. Going to apply to local DS site and play around.

src/js/components/Button/Button.js
```diff
@@ -169,6 +171,20 @@ const Button = forwardRef(
       );
     }
 
+    const sendAnalytics = useAnalytics();
+
+    const onClick = useCallback(
+      (event) => {
+        sendAnalytics({
```
ericsoderberghp, line 178: Would it be at all helpful to automatically include some key information, such as `label` and `id` so the recipient doesn't need to introspect into the DOM to derive them?

src/js/components/Button/Button.js
```diff
@@ -169,6 +171,20 @@ const Button = forwardRef(
...
+          type: 'buttonClick',
+          element: event.target,
+          data: event,
```
ericsoderberghp, line 181: Just confirming that "event" wouldn't be a more appropriate key name for the event?

src/js/contexts/AnalyticsContext/AnalyticsContext.js
```diff
@@ -0,0 +1,40 @@
...
+      observer = new window.MutationObserver(() => {
+        const url = window.location.href;
+        if (url !== pathInfoRef.current.lastUrl) {
+          pathInfoRef.current.lastUrl = url;
+          sendAnalytics({ type: 'pageView', data: url });
+        }
+      });
+      observer.observe(document, { subtree: true, childList: true });
+    }
+    pathInfoRef.current.observer = observer;
```
ericsoderberghp, line 30: Why do we need to save the `observer` in the `pathInfoRef`?

src/js/components/Form/Form.js
```diff
@@ -334,6 +353,23 @@ const Form = forwardRef(
...
 
+    useEffect(() => {
+      const element = formRef.current;
+      analyticsRef.current = { start: new Date(), errors: {} };
+      sendAnalytics({
+        type: 'formOpen',
+        element,
+      });
+      return () => {
+        sendAnalytics({
```
ericsoderberghp, line 364: I'm wondering if we can optimize to only send this when the form wasn't submitted? It seems a bit redundant to send both "formSubmit" and "formClose" with similar content.

ericsoderberghp (commented):
Nice start. A few quick thoughts.

src/js/components/Button/Button.js
MikeKingdom: I originally did that although I wanted access to the dom element anyway so they could put their analytics path in a `data-` attribute and it's pretty trivial to do `event.element.id` compared to `event.id`. We can certainly add the id if it exists though for convenience. I have more of a concern of putting the label there for button since I'm afraid they'll pass it to adobe or amplitude and accidentally cause issues with their data once they localize (I ran into this with OneView when we defaulted some analytics keys to labels. But it would probably be nice to see it in the events

src/js/components/Button/Button.js
MikeKingdom: probably true at this point. I originally settled on `data` because there are cases where it's not tied to real event.

src/js/contexts/AnalyticsContext/AnalyticsContext.js
MikeKingdom: Probably don't. I originally put it there so consumers would have access to it, but I don't think it's useful

src/js/components/Form/Form.js
MikeKingdom: Great idea.

src/js/contexts/AnalyticsContext/AnalyticsContext.js
```diff
@@ -0,0 +1,40 @@
+import React, { useCallback, useContext, useEffect, useRef } from 'react';
+
+export const AnalyticsContext = React.createContext(() => {});
+
+export const useAnalytics = () => useContext(AnalyticsContext);
+
+export const AnalyticsProvider = ({ onAnalytics, children }) => {
+  const pathInfoRef = useRef({
+    lastUrl: window.location.href,
```
halocline, line 9: In SSR app, window will not be available yet. Can we initialize as undefined?

```suggestion
    lastUrl: undefined,
```

src/js/contexts/AnalyticsContext/AnalyticsContext.js
```diff
@@ -0,0 +1,40 @@
...
+    observer: undefined,
+  });
+
+  const sendAnalytics = useCallback(
+    (data) => onAnalytics && onAnalytics(data),
```
halocline, line 14: It would be valuable to have each analytics event contain a timestamp.

src/js/contexts/AnalyticsContext/AnalyticsContext.js
```diff
@@ -0,0 +1,40 @@
...
+  );
+
+  useEffect(() => {
+    let observer;
+    if (onAnalytics) {
+      observer = new window.MutationObserver(() => {
+        const url = window.location.href;
+        if (url !== pathInfoRef.current.lastUrl) {
+          pathInfoRef.current.lastUrl = url;
+          sendAnalytics({ type: 'pageView', data: url });
```
halocline, line 25: previousUrl would be useful in this event in order to get a sense of from and to.

src/js/components/Form/Form.js
halocline: Agree. I was wanting to be able to differentiate between whether the form closed because of a successful event vs. an abandonment due to errors vs. just didn't need it (e.g. it mounted but I navigated away).

src/js/components/Button/Button.js
```diff
@@ -169,6 +171,20 @@ const Button = forwardRef(
       );
     }
 
+    const sendAnalytics = useAnalytics();
+
+    const onClick = useCallback(
+      (event) => {
+        sendAnalytics({
+          type: 'buttonClick',
+          element: event.target,
```
halocline, line 180: This comes through as the actual target (e.g. the child svg, text, div, etc.), but I find myself wanting to know what the parent button is.

halocline:
Anchor would be a good to have in this first pass.

ericsoderberghp (approved):

PR #6285, opened 2022-08-18
Fix Select to load all options even when clear={true}

### LINKED ISSUE #6267: Select dropdown is not allowing to scroll all the options when clear selection is enabled.
Select dropdown not allowing to scroll all the options when clear selection is enabled.

### Expected Behavior

Select dropdown should load all the options whether clear option is enabled or disabled.

### Actual Behavior

When clear option is enable Select dropdown is not completely scrollable

https://codesandbox.io/s/stupefied-proskuriakova-t9r5xw

### Steps to Reproduce

1. Load some dropdown options for select which has more options that enable scroll in Select.
2. Enable clear selection property for Select
3. Make a selection in the dropdown.
4. Observe you will not be able to scroll through all the options..

### Your Environment

- Grommet version: > 2.21.0
- Browser Name and version: All browsers

taysea:
#### What does this PR do?

- Fixes Select so that when `clear` is used, all Select options still render (original issue filed)
- Improves accessibility behavior of clear button by:
   - Moving ClearButton outside of the options box (clear button is not an option)
   - Improving styling of focused element to apply hover treatment (since focusIndicator={false}) so that keyboard user can track where the active element is
   - Enhanced a11yTitle of Clear Button

To Do: 
- [x] Enhance testing suite to improve "Clear" tests (I have made some enhancements but commented them out due to a timeout error I was receiving)

User should be able to tab to the "Clear Selection" button if it is at the bottom.

In this video, I open the select with "space", use down arrow to move through the options, "enter" to select options", then "tab" to tab through the select options and get to "Clear Selection". I am using VoiceOver on Mac.

https://user-images.githubusercontent.com/12522275/185514749-a8aa33c5-790c-4ead-aca8-d253975a7b5d.mov

SelectContainer

Using Clear story.

Use Clear story.

- [x] `userEvent` is used in place of `fireEvent`. (userEvent was causing timeout, used fireEvent)

Closes #6267 

No.

Yes. Fixed `clear` bug where not all Select options would render.

Backwards compatible.

### DIFF
--- a/src/js/components/Select/SelectContainer.js
+++ b/src/js/components/Select/SelectContainer.js
@@ -75,7 +75,7 @@ const ClearButton = forwardRef(
     const buttonLabel = label || `Clear ${name || 'selection'}`;
     return (
       <StyledButton
-        a11yTitle={`${buttonLabel}, press ${
+        a11yTitle={`${buttonLabel}. Or, press ${
           position === 'bottom' ? 'shift tab' : 'down arrow'
         } to move to select options`}
         fill="horizontal"
--- a/src/js/components/Select/SelectContainer.js
+++ b/src/js/components/Select/SelectContainer.js
@@ -49,6 +49,7 @@ const SelectOption = styled(Button)`
   // visible to keyboard users
   &:focus {
     ${(props) =>
+      props.active &&
       getHoverIndicatorStyle(
         !props.children && !props.theme.select.options
           ? undefined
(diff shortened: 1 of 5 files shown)


jcfilben (commented):
It might be nice if once the clear button receives focus to set the activeIndex to -1 so that we don't have the situation shown below where focus is on the clear button but the "Number One" option still has active styling.
Screen Shot 2022-08-19 at 9 48 44 AM

jcfilben (approved):
Looks good! I looked into the tests you added and was getting the same issue where the tests were timing out. I tried switching from userEvent to fireEvent and that seemed to solve the issue. Occasionally I've run across some cases where userEvent doesn't work as expected or affects the timing, so sometimes fireEvent is needed instead

src/js/components/Select/SelectContainer.js
```diff
@@ -44,27 +45,52 @@ const SelectOption = styled(Button)`
...
+`;
+
+const ClearButton = forwardRef(
+  ({ clear, onClear, name, theme, ...rest }, ref) => {
+    const { label, position } = clear;
+    const align = position !== 'bottom' ? 'start' : 'center';
+    const buttonLabel = label || `Clear ${name || 'selection'}`;
+    return (
+      <StyledButton
+        a11yTitle={`${buttonLabel}, press ${
```
halocline, line 78: This worked out nicely from a screenreader experience.

halocline:
```suggestion
        a11yTitle={`${buttonLabel}. Or, press ${
```

src/js/components/Select/SelectContainer.js
```diff
@@ -296,16 +326,8 @@ const SelectContainer = forwardRef(
           setActiveIndex(nextActiveIndex);
           setKeyboardNavigation(true);
         }
```
halocline, line 328: Should we add a check here that if activeIndex === 0, the up arrow key will allow focus on the clear button?

src/js/components/Select/stories/Clear.js
```diff
@@ -2,7 +2,10 @@ import React, { useState } from 'react';
...
 import { Box, Select } from 'grommet';
 
-const options = ['one', 'two', 'three'];
+const options = [];
+for (let i = 0; i < 500; i += 1) {
+  options.push(`Number ${i}`);
+}
 
 const ClearTop = () => {
   const [value, setValue] = useState();
```
halocline, line 11: can we add `dropHeight="large"` or something similar?

halocline (commented):
Looking good. Just a couple polishing items.

src/js/components/Select/stories/Clear.js
taysea: Done.

src/js/components/Select/SelectContainer.js
taysea: Done.

halocline: Works nicely.

halocline (approved):

ericsoderberghp (approved):

PR #6286, opened 2022-08-19
add formfield.checkbox.pad 

britt6612:
#### What does this PR do?
adds `formField.checkBox.pad` this will add `pad` to checkbox from theme.
FormField
set pad value to formField.checkBox.pad
set pad value to formField.checkBox.pad
Custom story

needed for theme
https://github.com/grommet/grommet-theme-hpe/pull/269
yes
sure
backwards compatible


src/js/components/CheckBox/CheckBox.js
```diff
@@ -50,6 +50,7 @@ const CheckBox = forwardRef(
       onMouseOut,
       onMouseOver,
       pad,
+      plainProp = false,
```
taysea, line 53:
```suggestion
      plain = false,
```

Any reason not to just refer to as plain?

britt6612: yeah in the `formfield` code we have 
```
     if (
            child &&
            child.type &&
            grommetInputNames.indexOf(child.type.displayName) !== -1 &&
            child.props.plain === undefined &&
            child.props.focusIndicator === undefined
          ) {
            return cloneElement(child, {
              plain: true,
              focusIndicator: false,
            });
          }
```
which sets `plain: true` since `checkbox`  would fall under this if statement... so I wasn't sure what else to name it

src/js/components/CheckBox/propTypes.js
```diff
@@ -13,6 +13,7 @@ if (process.env.NODE_ENV !== 'production') {
     name: PropTypes.string,
     onChange: PropTypes.func,
     pad: padPropType,
+    plain: PropTypes.oneOfType([PropTypes.string, PropTypes.bool]),
```
ericsoderberghp, line 16: when would `plain` be a "string"?

src/js/components/FormField/FormField.js
```diff
@@ -225,7 +225,7 @@ const FormField = forwardRef(
             child.props.focusIndicator === undefined
           ) {
             return cloneElement(child, {
-              plain: true,
+              plain: 'noBorder',
```
ericsoderberghp, line 228: I don't see "noBorder" being used anywhere. What does setting plain to this value do?

src/js/components/TextInput/propTypes.js
```diff
@@ -33,7 +33,11 @@ if (process.env.NODE_ENV !== 'production') {
     onSuggestionsOpen: PropTypes.func,
     onSuggestionsClose: PropTypes.func,
     placeholder: PropTypes.node,
-    plain: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf(['full'])]),
+    plain: PropTypes.oneOfType([
+      PropTypes.bool,
+      PropTypes.string,
```
ericsoderberghp, line 38: When would a string other than "full" be used?

src/js/components/CheckBox/propTypes.js
britt6612: So in `FormField.js` on line 220 there is 
```
          if (
            child &&
            child.type &&
            grommetInputNames.indexOf(child.type.displayName) !== -1 &&
            child.props.plain === undefined &&
            child.props.focusIndicator === undefined
          ) {
            return cloneElement(child, {
              plain: 'noBorder',
              focusIndicator: false,
            });
          }

```

in this case I changed the `plain: 'noBorder'` instead of true so that way it was not setting plain on `checkbox`. 
setting `plain: 'noBorder'` also triggers TextInput and Select to have their plain be true to get rid of the double border.

britt6612: This PR also is not the only direction depending on which route we want to go:
we have a plain prop and our theme looks specific if plain is true then take off horizontal pad.
Or we can have in our hpe theme
CheckBox: {
    vertical: ‘xsmall’,
    horizontal: ‘small’
},
Then someone would have to pass `pad={{ vertical: ‘xsmall’ }}` as a prop to `checkbox`

britt6612:
not sure what is going on with the netlify failing but think this is a better option for checkbox pad

britt6612:
closing in favor of 
https://github.com/grommet/grommet/pull/6300.
some how a `Submodule git` was added that cant seem to take out.

PR #6300, opened 2022-08-25
add formField.checkBox.pad as an option to be handeled in formField

britt6612:
#### What does this PR do?
adds `formField.checkBox.pad` this will add `pad` to checkbox from theme.
FormField
set pad value to formField.checkBox.pad
set pad value to formField.checkBox.pad
Custom story

needed for theme
https://github.com/grommet/grommet-theme-hpe/pull/269
yes I will add a PR if this is the direction we go
sure
backwards compatible

### DIFF
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -220,6 +220,7 @@ const FormField = forwardRef(
           if (
             child &&
             child.type &&
+            'CheckBox'.indexOf(child.type.displayName) === -1 &&
             grommetInputNames.indexOf(child.type.displayName) !== -1 &&
             child.props.plain === undefined &&
             child.props.focusIndicator === undefined
@@ -227,6 +228,19 @@ const FormField = forwardRef(
             return cloneElement(child, {
               plain: true,
               focusIndicator: false,
+            });
+          }
+          if (
+            child &&
+            child.type &&
+            'CheckBox'.indexOf(child.type.displayName) !== -1 &&
+            child.props.plain === undefined &&
+            child.props.focusIndicator === undefined
+          ) {
+            return cloneElement(child, {
+              plain: true,
+              focusIndicator: false,
+              pad: formFieldTheme?.checkBox?.pad,
             });
           }
           return child;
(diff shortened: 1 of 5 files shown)


src/js/components/FormField/FormField.js
```diff
@@ -229,6 +230,19 @@ const FormField = forwardRef(
               focusIndicator: false,
```
taysea, line 230:
```suggestion
              focusIndicator: false,
              pad: 'CheckBox'.indexOf(child.type.displayName) !== -1 ? formFieldTheme?.checkBox?.pad : undefined,
```

Does doing this work? And then removing the new 'if' statement below? Just trying to minimize repeated logic if possible

src/js/components/FormField/FormField.js
```diff
@@ -220,6 +220,7 @@ const FormField = forwardRef(
           if (
             child &&
             child.type &&
+            'CheckBox'.indexOf(child.type.displayName) === -1 &&
```
taysea, line 223: When testing my other comment, we would need to remove this line.

src/js/components/FormField/FormField.js
britt6612: let me test on my side!

britt6612: worked good!

taysea:
Can we add a unit test for this?

taysea (approved):
LGTM, thanks for adding the test. I tried the behavior locally and looks good.

britt6612:
> LGTM, thanks for adding the test. I tried the behavior locally and looks good.

thank you for the feedback 😄

ericsoderberghp (approved):

src/js/themes/base.d.ts
```diff
@@ -782,6 +782,9 @@ export interface ThemeType {
       margin?: MarginType;
       pad?: PadType;
     };
+    checkBox: {
+      pad: PadType;
```
lunika, line 786: Should not be optional like other keys ?

lunika: Same for `CheckBox` key.

britt6612: yes you are correct pushing up a fix now sorry for any inconvenience. Thanks for the catch!

Issue #6303, opened 2022-08-29
validate = 'change' does not work for select component 


anshu-kri:
### Expected Behavior

When Form element has validate="change".. The select component inside should validate onChange.. 

### Actual Behavior

When Form element has validate="change".. The select component inside should validate onChange.. It is not validating onChange, instead when there is change on any other field inside that form then the validate is fired.

### URL, screen shot, or Codepen exhibiting the issue

Here's a Codesandbox template that serves as a nice starting point for demonstrating an issue: https://codesandbox.io/s/wonderful-pascal-74dubv?file=/src/App.js

### Steps to Reproduce

1. Select any option in select component.. Notice the validate is not happening 
2. Now type a letter in textInput and validate will happen for both select and textInput component.

### Your Environment

- Grommet version: 2.25.1
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): Desktop (Windows)

paola-machado:
Hi! I would like to work on this, if that's ok

jcfilben:
Hi @paola-machado! Yes absolutely feel free to work on this

anshu-kri:
@jcfilben As this issue has been fixed and merged, When can we expect a release?

jcfilben:
Hi @anshu-kri, we are planning a release for sometime next week. If this fix is needed sooner the [stable branch](https://github.com/grommet/grommet/wiki/What-is-grommet-stable-and-how-to-use-it%3F) can be used.

Marking this issue as closed since #6316 has been merged.

PR #6316, opened 2022-09-06
(Select) fix: trigger a validation for multiple selection

buberdds:
#### What does this PR do?
For Select component with multiple selection `nextValue` is an object (array) and we want to trigger change event in this situation to trigger a validation.

Select.js

Manual and unit tests

https://github.com/grommet/grommet/issues/6303

### DIFF
--- a/src/js/components/Form/__tests__/Form-test-controlled.js
+++ b/src/js/components/Form/__tests__/Form-test-controlled.js
@@ -1,4 +1,5 @@
 import React, { useState } from 'react';
+import userEvent from '@testing-library/user-event';
 
 import 'jest-styled-components';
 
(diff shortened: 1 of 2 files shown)


src/setupTests.js
```diff
@@ -1,3 +1,5 @@
+global.scrollTo = jest.fn();
```
jcfilben, line 1: Can we just add this to the tests that need it instead of the global test setup file?

src/js/components/Form/__tests__/Form-test-controlled.js
```diff
@@ -931,4 +938,83 @@ describe('Form controlled', () => {
     // lastName should not trigger required validation onMount
     expect(screen.queryAllByText('required')).toHaveLength(0);
   });
+
+  test('validate select with multiple selection', async () => {
```
jcfilben, line 942: Thanks for adding a jest test!

jcfilben (approved):
Looks good, thanks for working on this!

ericsoderberghp (approved):

Issue #5354, opened 2021-06-22
Enhancement - Skeletons


taysea:
### Expected Behavior

Explore how we might implement a skeleton view of components. 

### Purpose & Motivation

Ease transition from no content to content by providing users with visual representations of content to be loaded while waiting for data to be returned.
- Provide sense of progress through visual blocking and motion.
- Provide indication for the types of content to be loaded and signal the layout and content's information hierarchy.
- Reduce cognitive load by allowing users' minds to anticipate the types of content to be displayed.

### Primary Needs
- Loading state - provided by caller
- Skeleton representation(s) for content/components to be loaded
- Mechanism(s) to map a component to its skeleton representation
   
### Additional Features to Consider (especially when creating the default)
- style/variant/type (e.g. rectangle, heading, text, paragraph, circle, media, table, chart, ...)
- dimensions/size
- animation styles
- fill color(s)
- duration
- level of customization / composability
- smart defaults
- visual contrast
- announcement for screen readers

### Potential Approaches

Currently have two ideas, both of which include a SkeletonContext internal to Grommet that would allow content components like Text, DataTable, etc. to render themselves as a skeleton.

Approach 1: Add `skeleton` prop to Box/Grid. With this, all children of the Box would render themselves as skeleton. Prop could be boolean, where default animations (or lack of animation, depending on what we decide) and durations would apply or could be an object where user specifies desired animation, duration, number of data to show (ex: in a table, should there be 5 skeleton row, 10?, etc.)

```
// Box.js

<SkeletonContext.Provider value={skeleton}>
    ... TYPICAL RETURN STUFF
</SkeletonContext.Provider>

// Text.js
export const Text = ({...props}) => {
    const {loading, animation} = useContext(SkeletonContext);

    if (loading) return <Skeleton type="text" {...animation} />;

    return (
        ... TYPICAL RETURN STUFF
    )
}

// IMPLEMENTATION

<Box skeleton={{loading: isLoading, animation: { type: 'fadeOut', duration: '3000'}}}>
    <Text>Hey there</Text>
</Box>
```

In the above, when `isLoading` (a state variable controlled by the caller is true, the Text will render in its skeleton flavor.

Approach 2: Add `Skeleton` component that acts as a wrapper around any set of components. With this, all children of the wrapper would render themselves as skeleton while `loading=true`. The component may take its own props to specify desired animation, minimum duration to be shown, etc. and the children inherit those properties.

```
// Text.js
export const Text = ({...props}) => {
[...]

PR #6318, opened 2022-09-12
Skeleton protoype implementation

MikeKingdom:
#### What does this PR do?

This PR implements a basic skeleton feature for Box, Text and Button.

See https://github.com/grommet/grommet/issues/5354

The usage is basically:
```jsx
<Box skeleton={{loading: isLoading, animation: { type: 'fadeOut', duration: '3000'}}}>
    <Text>Hey there</Text>
</Box>
```

TODO:
- [ ] Add the `skeleton` propTypes and typescript to Box (it already supports the property)
- [ ] Make `Heading`, `Paragraph` and `Card` recognize when they are in a skeleton context like `Text` and `Button`.
- [ ] Add tests
- [ ] Add a mechanism for overriding things like the width of a skeletonized Text on a per instance basis.

Skeleton/stories/Simple.js

https://github.com/grommet/grommet/issues/5354

image
image
image
image

Yes

Yes

Backwards compatible


src/js/components/Text/Text.js
```diff
@@ -43,6 +46,10 @@ const Text = forwardRef(
       return () => window.removeEventListener('resize', updateTip);
     }, [textRef, truncate]);
 
+    if (loading) {
+      return <Skeleton ref={ref} type="text" {...rest} />;
```
ericsoderberghp, line 50: Should we include "as" as well?

ericsoderberghp:
Some thoughts on the width aspect ...
What about `<Skeleton fill="horizontal" />` to allow callers to indicate that it should take the width of the container?
For Text, what about defaulting to inserting a `<span />` child with a number of `&nbsp;` characters matching the width of the text child of Text, if any? Just thinking about ways to "do the right thing" out of the box if possible.

src/js/components/Button/Button.js
```diff
@@ -271,6 +274,10 @@ const Button = forwardRef(
       return result;
     }, [active, disabled, kind, kindObj, plain, selected]);
 
+    if (loading) {
+       return <Skeleton ref={ref} type="button" />;
```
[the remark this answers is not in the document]
MikeKingdom: Others originally envisioned this as passing a few different types for shapes like maybe rectangle, circle etc. Basically Button is just picking a shape from the theme here for its skeleton. I chose `kind` to match some other usages of the kind idea.  Thoughts?

src/js/themes/base.js
```diff
@@ -1621,6 +1621,24 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
+      },
+      text: {
+        width: '100px',
+        margin: { vertical: 'xsmall' },
+        colors: {
+          dark: ['background-front', 'border'],
+          light: ['background-back', 'background-front'],
+        },
+      },
+      button: {
```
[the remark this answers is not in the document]
MikeKingdom: Good point. I can see that.

src/js/components/Text/Text.js
MikeKingdom: Sounds good

MikeKingdom:
> Some thoughts on the width aspect ... What about `<Skeleton fill="horizontal" />` to allow callers to indicate that it should take the width of the container? For Text, what about defaulting to inserting a `<span />` child with a number of `&nbsp;` characters matching the width of the text child of Text, if any? Just thinking about ways to "do the right thing" out of the box if possible.

I think it would make sense to attempt to match the text size if there is some. Most of the cases I was testing wouldn't normally have text though and probably need some other mechanism unless they just supply a placeholder. I also looked at just giving the span a background, which is possible, but the span itself tends to be effectively fill=horizontal already, despite text size. I'm not quite sure why the browser does that.

halocline (commented):
I'm liking the approach and Eric's comment about keeping Skeleton generic.

I can see the next steps as focussing on the detail work such as:
- refining sizing approximation
- colors
- how "deep" we want to go (I can see folks wanting to control this)
- animation phases

src/js/themes/base.js
```diff
@@ -1621,6 +1621,24 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
       maxInline: 5,
     },
+    skeleton: {
+      colors: {
+        dark: ['background-back', 'background-front'],
+        light: ['background-front', 'background-back'],
+      },
+      text: {
+        width: '100px',
+        margin: { vertical: 'xsmall' },
```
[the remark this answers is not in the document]
MikeKingdom: Moved the component specifics to the component. The margin is there so multiple <Text> have some gap

ericsoderberghp:
FYI, I'd be supportive of an API direction that we feel would allow for interesting animations but not needing to include that in the first version. Just looking for what the most essential first step is and taking it.

ericsoderberghp:
I think having a reasonable default for `<Box skeleton />` would be handy. With a simple default behavior in the base.js theme file and allowing other themes to override this. I like tying the skeleton styling more to the theme than to setting per property.

jcfilben:
Some notes from testing:
- Button alignSelf=“end” isn’t taken into account (button is shown at the start)

- Skeleton placeholders don't adapt for different screen sizes

- Margin on Heading ins’t included in skeleton layout, resulting in a large vertical different once content has loaded

src/js/components/Skeleton/stories/TableSkeleton.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+import { Box, DataTable } from 'grommet';
+import { columns, DATA } from '../../DataTable/stories/data';
+
+export const TableSkeleton = () => (
+  <Box align="center" pad="large" skeleton>
+    <DataTable columns={columns} data={DATA} step={10} />
```
MikeKingdom, line 8: I hadn't done table yet since I don't think it's a useful example so far. In a real usage you wouldn't have data yet and we wouldn't want the headers to skeletonize I think. To do this correctly I think requires some changes to DataTable

jcfilben: That makes sense, I can go back and remove this story

MikeKingdom:
> Some notes from testing:
> 
> * Button alignSelf=“end” isn’t taken into account (button is shown at the start)
> * Skeleton placeholders don't adapt for different screen sizes
> * Margin on Heading ins’t included in skeleton layout, resulting in a large vertical different once content has loaded

Fix the button skeleton to get the generic props.
Improved Heading skeleton to handle the same props and have the same default margin and such.

britt6612:
some notes about accessibility I found: 

A screen reader should notify a user if an application is loading, busy, gets stuck, or if a process fails.

https://www.w3.org/WAI/GL/wiki/Notification_of_Loading/Busy

When the loading indicator disappears, it conveys a second meaning, which is 'not loading' or 'finished'. However, especially where loads may take more than a few seconds, a user who cannot see the icon disappear needs to be made aware the system is no longer "loading" and is thus available for usage.

There are several ways to do this. If new content takes focus when the loading is completed (for instance after a full page load), then users will understand the system is now available

use aria-live set to "assertive" to immediately surface a loading status to assistive technologies.
provide an SVG title value for the loading icon, which is exposed through the aria-live section.
The completion of the loading state should be conveyed to assistive technologies. A non-visible status message such as "loading complete" could be put in the aria-live section or exposed through a role="status". Alternatively, focus could be set to an appropriate element.

ericsoderberghp:
So, I wired up a local grommet-designer to try out the skeleton functionality. I built a screen with Page, PageContent, PageHeader, Heading, Paragraph, Text, NameValueList, NameValuePair:

Screen Shot 2022-11-01 at 5 33 08 PM

I then set `skeleton={true}` on the PageContent:

Screen Shot 2022-11-01 at 5 34 01 PM

It looks like the PageHeader is showing an extra Paragraph skeleton even though the PageHeader has no `subtitle`. Not sure why but it doesn't look right.

MikeKingdom:
> It looks like the PageHeader is showing an extra Paragraph skeleton even though the PageHeader has no `subtitle`. Not sure why but it doesn't look right.

PageHeader puts a `Paragraph` in the subtitle area if `subtitle` is of type string. So is designer passing an empty string there?

Skeleton always draws a `Paragraph` as the 3 lines of gray boxes (two wide and one not as wide.)

src/js/components/Skeleton/stories/Group.js
```diff
@@ -0,0 +1,68 @@
...
+    pad="medium"
+    gap="large"
+    round="medium"
+    flex="grow"
+    border
+    {...rest}
+  >
+    <Box direction="row" gap="large" justify="between" align="center">
+      <Box direction="row" gap="medium">
+        <Box width="96px" height="96px" background="orange!" round="small">
```
britt6612, line 18: can we just use `Avatar` size `xlarge`?

src/js/components/Skeleton/stories/Group.js
```diff
@@ -0,0 +1,68 @@
...
+            skeleton={{ width: '200px' }}
+          >
+            Compliance
+          </Text>
+          <Text size="small">
+            Acme Company Inc
+          </Text>
+        </Box>
+      </Box>
+      <Button label="Add" reverse icon={<FormAdd />} secondary />
```
britt6612, line 35: I dont see the icon coming through

src/js/components/Skeleton/stories/Group.js
```diff
@@ -0,0 +1,68 @@
...
+    gap="large"
+    round="medium"
+    flex="grow"
+    border
+    {...rest}
+  >
+    <Box direction="row" gap="large" justify="between" align="center">
+      <Box direction="row" gap="medium">
+        <Box width="96px" height="96px" background="orange!" round="small">
+          <Image />
```
britt6612, line 19: should we add a `src` to show an actual image?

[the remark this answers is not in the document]
britt6612: Screen Shot 2022-11-08 at 3 00 25 PM

src/js/components/Skeleton/stories/GridContainer.js
```diff
@@ -0,0 +1,75 @@
+import React, { useEffect, useState } from 'react';
+
+import { Box, Button, Card, Grid, Image, Text} from 'grommet';
+import { FormAdd } from 'grommet-icons';
+
+const labels = [];
+for (let i = 0; i < 10; i += 1) labels.push(`Item ${i}`);
+
+const skeleton = { message: { start: "Loading", end: 'Done' } };
```
britt6612, line 9:
```suggestion
const skeleton = { message: { start: "Loading", end: 'Content Loaded' } };
```

src/js/components/Skeleton/StyledSkeleton.js
```diff
@@ -0,0 +1,38 @@
...
+import {
+  backgroundStyle,
+  edgeStyle,
+  genericStyles,
+  heightStyle,
+  roundStyle,
+  widthStyle,
+} from '../../utils';
+
+export const StyledSkeleton = styled.div`
```
britt6612, line 13: curious why `div` and not `box` ?

src/js/components/Skeleton/__tests__/Skeleton-test.tsx
```diff
@@ -0,0 +1,124 @@
...
+          round="small"
+          colors={{ dark: ['green', 'blue'], light: ['pink', 'orange'] }}
+        />
+      </Grommet>,
+    );
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  // test('Skeleton with theme', () => {
```
britt6612, line 104: we can uncomment now should be good

src/js/components/Skeleton/index.d.ts
```diff
@@ -0,0 +1,32 @@
...
+} from '../../utils';
+
+
+export interface SkeletonProps {
+  a11yTitle?: A11yTitleType;
+  alignSelf?: AlignSelfType;
+  as?: PolymorphicType;
+  colors?: SkeletonColorsType;
+  gridArea?: GridAreaType;
+  margin?: MarginType;
```
britt6612, line 23: do we need `message` prop?

src/js/components/Skeleton/propTypes.js
```diff
@@ -0,0 +1,28 @@
...
+if (process.env.NODE_ENV !== 'production') {
+  PropType = {
+    ...genericProps,
+    as: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.func,
+      PropTypes.element,
+    ]),
+    colors: skeletonColorsPropType,
+    height: heightPropType,
```
britt6612, line 22: margin and message prop

src/js/components/Skeleton/stories/Group.js
MikeKingdom: This is more of just a product image and not really an avatar. Like with Marketplace cards where we have the product icon which is an image. Avatar is intended to be more of a circle.

src/js/components/Skeleton/stories/Group.js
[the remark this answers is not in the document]
MikeKingdom: Probably. Just need to find something suitable. In the skeleton mode I wouldn't expect the image to have a `src` since that data is probably what you're waiting on

src/js/components/Skeleton/stories/GridContainer.js
MikeKingdom: makes sens

src/js/components/Skeleton/StyledSkeleton.js
MikeKingdom: partly because Box has some special interpretation when the skeleton context is present. I started that way but it got more complicated

src/js/components/Skeleton/stories/Group.js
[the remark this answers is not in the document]
britt6612: good point to me just looked like something was missing

src/js/components/Skeleton/__tests__/Skeleton-test.tsx
MikeKingdom: Thanks. Didn't notice that got commented out

src/js/components/Skeleton/index.d.ts
MikeKingdom: I didn't think so since the intent was the main box container which has the context was owning that. I'm less clear about the direct usage of the Skeleton component since it seemed more valuable to just do it via a Box with the skeleton prop and not generally use Skeleton directly. Curious to see how the use model evolves

src/js/components/Skeleton/propTypes.js
MikeKingdom: Wouldn't expect message here as I noted above. Margin is in the genericProps

src/js/components/Skeleton/stories/Group.js
britt6612: okay thats fair

src/js/components/Skeleton/StyledSkeleton.js
britt6612: good to know figured there was some reasoning behind picking `div` over `box`

src/js/components/Skeleton/index.d.ts
britt6612: maybe we can discuss ☝️

britt6612:
> > It looks like the PageHeader is showing an extra Paragraph skeleton even though the PageHeader has no `subtitle`. Not sure why but it doesn't look right.
> 
> PageHeader puts a `Paragraph` in the subtitle area if `subtitle` is of type string. So is designer passing an empty string there?
> 
> Skeleton always draws a `Paragraph` as the 3 lines of gray boxes (two wide and one not as wide.)

I didn't see any changes to PageHeader are we going to address this or?

src/js/components/Paragraph/ParagraphSkeleton.js
```diff
@@ -0,0 +1,28 @@
...
+  const height = data ? data.size : size;
+  return (
+    <Box
+      ref={ref}
+      gap="xsmall"
+      width={{ max: fill ? 'none' : data && data.maxWidth }}
+    >
+      <Skeleton height={height} />
+      <Skeleton height={height} />
+      <Skeleton height={height} width="30%" />
```
britt6612, line 21: why 30%

ericsoderberghp:
> > > It looks like the PageHeader is showing an extra Paragraph skeleton even though the PageHeader has no `subtitle`. Not sure why but it doesn't look right.
> > 
> > 
> > PageHeader puts a `Paragraph` in the subtitle area if `subtitle` is of type string. So is designer passing an empty string there?
> > Skeleton always draws a `Paragraph` as the 3 lines of gray boxes (two wide and one not as wide.)
> 
> I didn't see any changes to PageHeader are we going to address this or?

Ah, yes. grommet-designer is passing `subtitle=""`. Hmm... I would expect that PageHeader wouldn't add a Paragraph for this case. But, I'm OK treating this as a separate issue.

src/js/components/Skeleton/StyledSkeleton.js
taysea: Should we add a comment about this in the code? ^^

src/js/components/Button/Button.js
```diff
@@ -271,6 +274,19 @@ const Button = forwardRef(
       return result;
     }, [active, disabled, kind, kindObj, plain, selected]);
 
+    if (skeleton) {
+       return (
+        <Skeleton
+          ref={ref}
+          height={theme.text[size || 'medium']?.height || size}
+          a11yTitle={a11yTitle}
+          { ...rest }
```
taysea, line 283: It looks like there's extra spaces here that the linter/prettier usually pulls out.

taysea: same with some other spots in the code.

src/js/components/Heading/HeadingSkeleton.js
```diff
@@ -0,0 +1,33 @@
+import React, { forwardRef, useContext } from 'react';
+import { ThemeContext } from 'styled-components';
+import { Skeleton } from '../Skeleton';
+
+
+const HeadingSkeleton = forwardRef(
+  (
+    props,
```
taysea, line 8:
```suggestion
    {
    as,
    level,
    size
    },
```
can we destructure the props?

src/js/components/Paragraph/ParagraphSkeleton.js
taysea: I think so the last line looks like it's trailing off

src/js/components/Paragraph/ParagraphSkeleton.js
```diff
@@ -0,0 +1,28 @@
...
+    </Box>
+  );
+});
+
+ParagraphSkeleton.displayName = 'ParagraphSkeleton';
+
+export { ParagraphSkeleton };
```
taysea, line 28:
```suggestion
export { ParagraphSkeleton };

```
missing new line at end of file

src/js/components/Skeleton/__tests__/Skeleton-test.tsx
```diff
@@ -0,0 +1,124 @@
...
+            Heading
+          </Heading>
+          <Text size="xsmall">Text</Text>
+          <Paragraph size="large">Paragraph</Paragraph>
+          <Button size="large" label="Button" onClick={() => {}} />
+        </Box>
+      </Grommet>,
+    );
+
+    expect(asFragment()).toMatchSnapshot();
```
taysea, line 56: Should we add expects here that:

`expect(screen.queryByText('Heading')).toBeNull();`

src/js/components/Skeleton/stories/CustomThemed/Custom.js
```diff
@@ -0,0 +1,56 @@
...
+
+import { Box, Button, Heading, Grommet, Text } from 'grommet';
+
+export const Custom = () => {
+  const customTheme = {
+    skeleton: {
+      colors: {
+        light: ['#a2a8a8', '#adb9ba'],
+      },
+      round: 'xsmall',
```
taysea, line 11: I'm not seeing the rounding:
Screen Shot 2022-11-09 at 11 21 29 AM

src/js/components/Heading/Heading.js
```diff
@@ -34,6 +46,13 @@ const Heading = forwardRef(
       return () => window.removeEventListener('resize', updateOverflowWrap);
     }, [headingRef, overflowWrapProp]);
 
+    let content = children;
+    if (skeleton) {
+      content = (
+        <HeadingSkeleton level={level} />
```
jcfilben, line 52: Should we be passing `size` here as well? It looks like size is used in the HeadingSkeleton file

src/js/components/Skeleton/index.d.ts
```diff
@@ -0,0 +1,32 @@
...
+  height?: HeightType;
+  pad?: PadType;
+  round?: RoundType;
+  width?: WidthType;
+}
+
+declare const Skeleton: React.FC<SkeletonProps>;
+
+export { Skeleton };
```
taysea, line 32:
```suggestion
export { Skeleton };

```
missing new line

src/js/components/Skeleton/stories/GridContainer.js
```diff
@@ -0,0 +1,75 @@
...
+import { FormAdd } from 'grommet-icons';
+
+const labels = [];
+for (let i = 0; i < 10; i += 1) labels.push(`Item ${i}`);
+
+const skeleton = { message: { start: "Loading", end: 'Content Loaded' } };
+
+const Item = ({ title, ...rest }) => (
+  <Card
+    width="531px"
```
taysea, line 13: can we avoid custom px values in the stories and use tshirt size instead?

src/js/components/Skeleton/README.md
```diff
@@ -0,0 +1,2 @@
+## Text
```
jcfilben, line 1:
```suggestion
## Skeleton
```

src/js/components/Skeleton/stories/GridContainer.js
```diff
@@ -0,0 +1,75 @@
...
+    pad="medium"
+    gap="large"
+    round="medium"
+    flex="grow"
+    {...rest}
+  >
+    <Box direction="row" gap="large" justify="between" align="center">
+      <Box direction="row" gap="medium">
+        <Box width="96px" height="96px" background="brand" round="small">
```
taysea, line 22: Any way to avoid px value and use tshirt size instead?

src/js/components/Skeleton/stories/GridContainer.js
```diff
@@ -0,0 +1,75 @@
...
+          <Image />
+        </Box>
+        <Box>
+          <Text
+            size="large"
+            color="text-strong"
+            weight="bold"
+            skeleton={{ width: '200px' }}
```
taysea, line 30:
```suggestion
            skeleton={{ width: 'small' }}
```
small --> 192px which feels close enough?

src/js/components/Skeleton/stories/Group.js
```diff
@@ -0,0 +1,68 @@
+import React, { useContext } from 'react';
+
+import { Box, Button, Card, Grommet, Image, Text, ThemeContext} from 'grommet';
+import { FormAdd } from 'grommet-icons';
+
+const Item = ({ ...rest }) => (
+  <Card
+    width="531px"
```
taysea, line 8: Same as other comment, can we avoid `px` values?

src/js/components/Skeleton/stories/Group.js
```diff
@@ -0,0 +1,68 @@
...
+      <Box direction="row" gap="medium">
+        <Box width="96px" height="96px" background="orange!" round="small">
+          <Image />
+        </Box>
+        <Box>
+          <Text
+            size="large"
+            color="text-strong"
+            weight="bold"
+            skeleton={{ width: '200px' }}
```
taysea, line 26:
```suggestion
            skeleton={{ width: 'small' }}
```

src/js/components/Text/TextSkeleton.js
```diff
@@ -0,0 +1,25 @@
...
+      height={height}
+      {...theme.text.skeleton}
+      {...rest}
+    />
+  );
+});
+
+TextSkeleton.displayName = 'TextSkeleton';
+
+export { TextSkeleton };
```
taysea, line 25:
```suggestion
export { TextSkeleton };

```

src/js/components/Skeleton/stories/GridContainer.js
```diff
@@ -0,0 +1,75 @@
...
+      <Grid pad="small" gap="small" columns={['medium', 'medium']}>
+        {labels.map((label, index) => (
+          <Item
+            key={label}
+            title={label}
+            skeleton={
+              loading
+                ? {
+                    animation: [
+                      { type: 'fadeIn', delay: index * 200},
```
taysea, line 61: wondering if we can show this looping? where it fades in and out?

src/js/components/Heading/Heading.js
MikeKingdom: Yup. Adding actually `fill` and `...rest`. Just don't need the weight or children, etc.

src/js/components/Heading/HeadingSkeleton.js
MikeKingdom: definitely

src/js/components/Paragraph/ParagraphSkeleton.js
MikeKingdom: What Taylor said :)

src/js/components/Skeleton/stories/GridContainer.js
MikeKingdom: tshirt size here didn't work very well for the example. I've run into that alot with cards in the wild

src/js/components/Skeleton/stories/GridContainer.js
MikeKingdom: I replaced most of the pixel sizes with tshirt sizes. I only kept one of the card widths as pixels

src/js/components/Skeleton/stories/GridContainer.js
MikeKingdom: yup

src/js/components/Skeleton/stories/GridContainer.js
MikeKingdom: I originally tried that but didn't find a decent way to do that. Suggestions?

src/js/components/Skeleton/stories/Group.js
MikeKingdom: I made the GridContainer example do tshirt sizes for the Card but  for the Group and Individual examples it wasn't looking good or representing what we mocked. It's that awkward size between medium and large.

src/js/components/Skeleton/stories/GridContainer.js
```diff
@@ -0,0 +1,70 @@
...
+                : undefined
+            }
+          />
+        ))}
+      </Grid>
+    </Box>
+  );
+};
+
+export default {
```
ericsoderberghp, line 68: Because of the animation, the screen first renders as blank, so there's not much for Chromatic to notice. I think we should disable this test from Chromatic.

src/js/components/Skeleton/stories/Group.js
```diff
@@ -0,0 +1,66 @@
...
+        <Content />
+      </Grommet>
+      <Grommet theme={theme} themeMode="dark">
+        <Content />
+      </Grommet>
+    </>
+  );
+};
+
+export default {
```
ericsoderberghp, line 64: This is also a blank page for Chromatic and should probably be disabled.

ericsoderberghp (commented):
Just a few small things.

I could envision wanting to add `skeleton` to Grid at some point. OK to not have it there now though.

MikeKingdom:
> Just a few small things.
> 
> I could envision wanting to add `skeleton` to Grid at some point. OK to not have it there now though.

Definitely.

ericsoderberghp (approved):

jcfilben:
With v4 of `grommet-theme-hpe` the dark mode skeleton is a lot harder to see than it was before.

Previously (v3.3.0 of hpe theme):
Screen Shot 2022-11-10 at 12 36 14 PM

After (v4.0.0 of hpe theme):
Screen Shot 2022-11-10 at 12 34 43 PM

MikeKingdom:
> With v4 of `grommet-theme-hpe` the dark mode skeleton is a lot harder to see than it was before.

It's using `background-front`. Does background-front need a bit more lightness in v4?

ericsoderberghp:
> > With v4 of `grommet-theme-hpe` the dark mode skeleton is a lot harder to see than it was before.
> 
> It's using `background-front`. Does background-front need a bit more lightness in v4?

I'm inclined to address this as needed within the hpe theme.

I do think it's helpful to ensure we look at the dark theme stuff in a dark theme context, a dark room, with the rest of the OS being dark, etc. I'm wondering if it might be OK in that context. But happy to adjust if needed, just within the hpe theme.

jcfilben (approved):
Looks good!

PR #6327, opened 2022-09-14
added font-variant to global font type

joybh98:
#### What does this PR do?
Adds the ```font-variant``` option to the global font type
```src/js/themes/base.d.ts```
```yarn test```, tests failing

NA
https://github.com/grommet/grommet/issues/6289

### DIFF
--- a/src/js/utils/styles.js
+++ b/src/js/utils/styles.js
@@ -9,6 +9,7 @@ export const baseStyle = css`
   font-size: ${(props) => props.theme.global.font.size};
   line-height: ${(props) => props.theme.global.font.height};
   font-weight: ${(props) => props.theme.global.font.weight};
+  font-variant: ${(props) => props.theme.global.font.variant};
   ${(props) =>
     !props.plain && backgroundStyle(props.theme.baseBackground, props.theme)}
   box-sizing: border-box;
--- a/src/js/utils/styles.js
+++ b/src/js/utils/styles.js
@@ -9,7 +9,10 @@ export const baseStyle = css`
   font-size: ${(props) => props.theme.global.font.size};
   line-height: ${(props) => props.theme.global.font.height};
   font-weight: ${(props) => props.theme.global.font.weight};
-  font-variant: ${(props) => props.theme.global.font.variant};
+  /* check if prop is defined in the theme*/
+  ${(props)=>props.theme.global.font.variant && `
+    font-variant:${props.theme.global.font.variant};
+  `}
   ${(props) =>
     !props.plain && backgroundStyle(props.theme.baseBackground, props.theme)}
   box-sizing: border-box;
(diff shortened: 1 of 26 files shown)

### CODE REVIEW

**Review by jcfilben — COMMENTED** [review]
Thanks for working on this!
I think the following needs added to the `src/js/utils/styles.js` file in baseStyle so that we make sure the font variant we are passing through the theme actually gets applied:
`font-variant: ${(props) => props.theme.global.font.variant};`

After that try running `yarn test-update` to update the testing snapshots

**Comment by joybh98** [comment]
hi @jcfilben added font-variant,updated the tests, but still tests for some components are failing, do let me know what can be done.

And thanks for being patient and taking out the time to look at this PR! Appreciate it

**Comment by jcfilben** [comment]
> hi @jcfilben added font-variant,updated the tests, but still tests for some components are failing, do let me know what can be done.

Hmm I'm not sure why the tests are acting strange. I'm going to try and just update the branch on my end and then push that up to this branch and see if that gets the tests to re-run

**Comment by joybh98** [comment]
@jcfilben checks are passing now ! Thank you ! Do let me know if any other changes and/or tests needs to be updated before merging.

File: src/js/utils/styles.js [hunk 1]
```diff
@@ -9,6 +9,7 @@ export const baseStyle = css`
   font-size: ${(props) => props.theme.global.font.size};
   line-height: ${(props) => props.theme.global.font.height};
   font-weight: ${(props) => props.theme.global.font.weight};
+  font-variant: ${(props) => props.theme.global.font.variant};
```
ericsoderberghp, line 12: What happens if there is no `props.theme.global.font.variant`, do we still include this in the CSS? If so, I'm wondering if we should make these more conditional.

joybh98: @jcfilben could you give some inputs?

jcfilben: I think we could change these to be mode conditional by doing something like this, where we first check if the prop is defined in the theme before adding the css.
```
${(props) => props.theme.global.font.variant && `
    font-variant: ${props.theme.global.font.variant};
`}
```

jcfilben: Strange I'm not sure what is going on in your local environment that is causing the tests to behave oddly. I will go ahead and do the same thing we did last time and update the branch from my end to get the tests to re-run.

If you create any future pull requests it might be worth starting with a fresh fork of grommet to see if that resolves any issues with the tests, but no need to do that for this pull request.

joybh98: @jcfilben noted ! any future PRs will be with a fresh fork! Thanks for taking out the time to update the branch. Let me know if anything else needs to be done for this PR

jcfilben (approved):

ericsoderberghp (approved):

jcfilben:
@joybh98 thanks for contributing to Grommet!

PR #6335, opened 2022-09-16
SelectMultiple - adjust behavior when lazy loading

jcfilben:
#### What does this PR do?
When using the `onMore` prop with SelectMultiple we shouldn't show the 'Select All' button, because without all the options loaded we don't have a good way to select all of them. In addition we shouldn't show the total number of options when lazy loading because the total is unknown.

This PR also adjusts the alignment of the close arrow when using `showSelectedInline`.

Added storybook story SelectMultiple/Long list

no
no, component hasn't been released yet
backwards compatible

### DIFF
--- a/src/js/components/SelectMultiple/SelectMultiple.js
+++ b/src/js/components/SelectMultiple/SelectMultiple.js
@@ -421,10 +421,12 @@ const SelectMultiple = forwardRef(
                         {...rest}
                         tabIndex="-1"
                         type="text"
-                        // eslint-disable-next-line max-len
                         placeholder={
+                          // eslint-disable-next-line no-nested-ternary
                           value.length === 0
                             ? placeholder || selectValue || displayLabelKey
+                            : onMore
+                            ? `${value.length} selected`
                             : `${value.length} selected of ${allOptions.length}`
                         }
                         plain
--- a/src/js/components/SelectMultiple/SelectMultipleContainer.js
+++ b/src/js/components/SelectMultiple/SelectMultipleContainer.js
@@ -330,6 +330,7 @@ const SelectMultipleContainer = forwardRef(
         labelKey={labelKey}
         limit={limit}
         onChange={onChange}
+        onMore={onMore}
         options={options}
         search={search}
         setActiveIndex={setActiveIndex}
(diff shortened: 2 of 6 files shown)


jcfilben:
Marking as a draft. I realized the close icon button alignment is working well for the hpe theme but not for the grommet theme. Going back to see if I can find a solution that works for both

jcfilben:
I tested out different options for the close button icon alignment and found that having a top pad of 'small' works better for the grommet theme and 'xsmall' works better for the hpe theme (see line 353 of SelectMultipleContainer.js).

For now I left the padding as 'small' as it felt visually slightly better than setting padding to 'xsmall'

src/js/components/SelectMultiple/SelectMultiple.js
```diff
@@ -421,10 +421,12 @@ const SelectMultiple = forwardRef(
...
                         tabIndex="-1"
                         type="text"
-                        // eslint-disable-next-line max-len
                         placeholder={
+                          // eslint-disable-next-line no-nested-ternary
                           value.length === 0
                             ? placeholder || selectValue || displayLabelKey
+                            : onMore
+                            ? `${value.length} selected`
                             : `${value.length} selected of ${allOptions.length}`
```
ericsoderberghp, line 430: OK for now, but these should be included in i18n and use MessageContext at some point. Can you file a separate issue for that?

ericsoderberghp (approved):

src/js/components/SelectMultiple/SelectMultiple.js
jcfilben: Yes, I'll create a separate issue

jcfilben: Issue created https://github.com/grommet/grommet/issues/6337

taysea:
Maybe an enhancement separate, but wondering if we should allow the caller to control `totalOptions` (num) and `onSelectAll` (func) to handle large data cases?

ericsoderberghp:
> Maybe an enhancement separate, but wondering if we should allow the caller to control `totalOptions` (num) and `onSelectAll` (func) to handle large data cases?

Good thoughts. I am willing to move forward without for now.

taysea (approved):

PR #6353, opened 2022-09-26
pass a11ytitle to avatar

britt6612:
#### What does this PR do?
This PR looks at if a `src` is passed then it should use the `image` tag. Also expands the props to include a11ytitle. 
Avatar.js
added a test with `a11ytitle`
add a `a11ytitle` 

our current Avatar does not include `a11ytitle` also when it is an image it does not show an image tag in the dom https://storybook.grommet.io/?path=/story/visualizations-avatar-basic--basic
none
yes
yess
 backwards compatible


jcfilben (approved):
Looks good!

src/js/components/Avatar/Avatar.js
```diff
@@ -49,12 +52,18 @@ const Avatar = ({
 
   if (typeof src === 'string') {
     return (
-      <StyledAvatar {...avatarProps} {...rest} background={`url(${src})`} />
+      <StyledAvatar {...avatarProps} {...rest}>
+        <Image a11yTitle={a11yTitle || ariaLabel} fit="contain" src={src} />
```
ericsoderberghp, line 56: It seems inconsistent to put `a11yTitle` in the child `Image` here but in the parent `StyledAvatar` below. I would think they should always be at the same place in the DOM.

ericsoderberghp (commented):
Sorry for the long delay on this one.

src/js/components/Avatar/Avatar.js
britt6612: That's fair to be consistent it should be on the `StyledAvatar`

src/js/components/Avatar/Avatar.js
```diff
@@ -49,12 +52,22 @@ const Avatar = ({
 
   if (typeof src === 'string') {
     return (
-      <StyledAvatar {...avatarProps} {...rest} background={`url(${src})`} />
+      <StyledAvatar
```
ericsoderberghp, line 55: Rather than repeating this block twice, can we build the `content` using the typeof checks and then put that inside a single `StyledAvatar`?

ericsoderberghp (approved):

src/js/components/Avatar/Avatar.js
jcfilben: I think when we have an Image needs an a11yTitle. Also it looks like we are getting an accessibility error for putting the aria-label on a div without a role.

Screen Shot 2022-10-04 at 3 49 24 PM

britt6612: Ill look

Issue #6328, opened 2022-09-14
Radio Button lacks focus indicator 


britt6612:
### Expected Behavior
If you `tab` to a RadioButton I would expect to have the focus on the RadioButton it self

### Actual Behavior
Receives no focus if it is not clicked 

### Steps to Reproduce

https://storybook.grommet.io/?path=/story/input-radiobutton-simple--simple
Tab here and see RadioButton never gets focused 

If you tab in this example the focus goes on the active RadioButton
https://v2.grommet.io/radiobuttongroup

### Your Environment

- Grommet version: latest
- Browser Name and version: Chrome Latest
- Operating System and version (desktop or mobile):

jcfilben:
Hi @avantika-codes you can try running `yarn jest -u` to update snapshots, then put up a pull request and we can take a look at the changes

avantika-codes:
Hi @jcfilben, I have created a PR for this issue, when running yarn test || npm test on my local, no test cases fail. Although the pipeline fails in PR.

PR #6367, opened 2022-09-30
fix/radio-button-focus-indicator

paola-machado:
#### What does this PR do?
Adds style to input when using `tab` and Closes (hopefully 😅) issue #6328 

The issue was that there was no focus style being applied to the radio button when navigating with the keyboard.

On the StyledRadioButton file. Also, related snapshot test were updated.

Snapshot was fixed and all tests are passing. 

`yarn test`

#6328

No

No

Should be backwards compatible.

### DIFF
--- a/src/js/components/RadioButton/StyledRadioButton.js
+++ b/src/js/components/RadioButton/StyledRadioButton.js
@@ -14,8 +14,9 @@ const StyledRadioButtonContainer = styled.label`
   align-items: center;
   user-select: none;
   width: fit-content;
-  ${(props) => props.disabled && disabledStyle} ${(props) =>
-    !props.disabled && 'cursor: pointer;'}
+
+  ${(props) => props.disabled && disabledStyle};
+  ${(props) => !props.disabled && 'cursor: pointer;'};
 
   :hover input:not([disabled]) + div,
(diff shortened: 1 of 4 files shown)


src/js/components/RadioButton/StyledRadioButton.js
```diff
@@ -108,6 +109,10 @@ const StyledRadioButtonBox = styled.div`
     )};
   ${(props) => props.focus && focusStyle()};
```
jcfilben, line 110: The solution you have is working but I'm thinking we may want to utilize the focus prop we have here on line 110 and get that working instead of adding a separate way to trigger focus styling.
To do this I think we just need to change `RadioButton.js` to set focus to true/false when `onFocus` or `onBlur` is triggered. You can take a look at the focus code in CheckBox.js to get an idea of how this is working it that component. Also I think we should default `focusIndicator` to true (we only want to turn off the focus indicator in special cases such as when the component is wrapped in a FormField). Again you can follow pretty much the same pattern CheckBox.js is for handling the focus stuff. Let me know if there are any questions!

jcfilben:
Closing due to inactivity but feel free to re-open

PR #6378, opened 2022-10-02
[TS] Migrate Distribution tests to tsx 

alexmirrington:
#### What does this PR do?

This PR migrates `Distribution` component tests from `*.js` to `*.tsx`.

- Remove test case for undefined values as this is an unsupported type in both `index.d.ts` and `propTypes.js`.

- Ensure all tests are currently passing
- Refer to the background context section below for more information about the initial errors picked up by TypeScript and the fixed implemented in this PR.

- Manually ran and passed unit tests locally for the changed component
- CircleCI tests have passed.

- The reviewer should verify all changes made to tests and ensure that:
    - All test removals are due to unused legacy code and will not cause regressions
    - All tests behave identically and modifications are type-related.

None

Parent tracking issue: #5599

No

No, all fixes are test-related and not component-related.

Backwards compatible.


alexmirrington:
> Can we keep the undefined value test and just add a `ts-ignore` comment? Even though we don't support passing undefined to value I think we still want to keep this test just to make sure that having and undefined value in Distribution doesn't crash the app.

For sure, that makes sense for those that aren't using TypeScript, especially those that might already have nulls and/or undefined values being passed through to the component. Added the test back in and adopted a similar strategy for my other PRs

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

Issue #6369, opened 2022-09-30
Reorder List via drag and drop problem


jonka5:
### Expected Behavior

With onOrder the item should be dropped to where ever the user wants it.

### Actual Behavior

When reordering the list via drag and drop you could only move the item one step up or down.

### URL, screen shot, or Codepen exhibiting the issue

Version 2.25.3 works as expected. Behavior starts to occur in 2.26.0

https://user-images.githubusercontent.com/74107477/193322053-8ae41a3f-dfc8-476d-9679-8443cce474b3.mov

https://codesandbox.io/s/list-with-order-and-action-combined-nmeh4o?file=/src/App.js

### Your Environment

- Grommet version: 2.26.0

MikeKingdom:
PR #6249 seems to have caused this regression

britt6612:
@MikeKingdom can we close this since it was merged?

MikeKingdom:
Yup. Must've forgot a keyword like Fixes :)

PR #6386, opened 2022-10-04
Fix onOrder dragging issues

MikeKingdom:
#### What does this PR do?
This fixes a regression in re-ordering items in List by dragging. (Dragging/re-ordering is enabled when `onOrder` is specified. )

The recently added `pinned` functionality was not using the `orderingData` when trying to figure out the current index of an item. `orderingData` is an temporary ordering of the `data` that only exists while the user is dragging. Once the user  is done dragging, that orderingData is passed to the `onOrder` callback to make the new ordering permanent.

List.js

storybook List/Order and Pinned stories.

#6369 

No

Yes

Backwards compatible


ericsoderberghp (approved):

jcfilben (approved):
Looks good!

Issue #6283, opened 2022-08-17
Feature Request | wen need auto-scroll to selected options in `Select` component.


y-okamoto-1113:
### Expected Behavior
after i selected  an option at `Select` component and reopen the dropdown of `Select` component, dropdown auto-scrolls to the selected option.

### Actual Behavior
auto-scrolling to the selected option is not working.

### Steps to Reproduce
1. click here.
https://v2.grommet.io/select?c=GYVwdgxgLglg9mABAUQB4EMC2AHANgUwAoBKRAbwChFEIEBnKRAbQDd1cR8AaRO-KAGrtOAXUQBeRACV86aADoQfAMpR0UIgHJNxANxVEAJ34hDSQgeoAeZfgLRL1RHGyx64sk011M7XJp5NTHwAExgQTADETVx0QwBzfCjNdGSAI2SIZJDkpMDgZPjkgAtkmGSAK2SAa00RAF9HajYOfA8WzkanJwQAYWL0MESPQjJnV3gketJxAD5efiFWwhc3MGIu7sQQwxcACXwYeOKocQAiACYABivsVDPHAHpZgz0KeqA
2. Click `Select` component.
3. select the "k" option.
4. reopen dropdown options.

after i selected option like 'k' and reopen dropdown options, it doesn't auto-scroll to the selected option(k).

https://user-images.githubusercontent.com/40970030/185215482-4dddc8de-2dc2-4319-bd75-448ede941a84.mov

### Your Environment

- Grommet version: 2.17.5
- Browser Name and version: Chrome  104.0.5112.79 (Official Build) (arm64)
- Operating System and version (desktop or mobile): m1 mac Monetery 12.5

joybh98:
Can i take this ?

joybh98:
cool @jcfilben will start work on it this weekend!

jcfilben:
> cool @jcfilben will start work on it this weekend!

Perfect, just in time for [hacktoberfest](https://hacktoberfest.com/)! Feel free to reach out if there are any questions

PR #6388, opened 2022-10-04
Added modifications to select.js for auto-scroll

joybh98:
#### What does this PR do?
Adds auto scroll option to the `Select` component #6283 
`src/js/components/Select/Select.js` 
`yarn test`, but it may not pass TC's right now as I opened this PR for discussion about my approach.
NA

NA
#6283 
Yes

### DIFF
--- a/src/js/components/Select/Select.js
+++ b/src/js/components/Select/Select.js
@@ -42,6 +42,7 @@ const Select = forwardRef(
       a11yTitle,
       'aria-label': ariaLabel,
       alignSelf,
+      autoScroll,
       children,
       clear = false,
       closeOnChange = true,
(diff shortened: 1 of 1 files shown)


joybh98:
hi @jcfilben, I have opened this PR to discuss the approach i am taking and I am not expecting it to pass test cases, whenever you get time do take a look at it !

Thanks for your patience!

jcfilben:
Hey @joybh98 thanks for opening a PR to get the discussion started on this! I am thinking we will want to have a prop that controls this so that people can opt-in to this behavior.

Once the drop opens we should grab the ref for the selected option that comes first in the list of options (if there are multiple options selected). We should check that the option isn't already in view. If it is not in view then we should scroll to it. I think most people will likely want this scroll to be 'smooth' so that visually they can see that the select is moving them down to the first selected option.

Some things to consider
- What if when we open the drop there is already a search string within the search that the options are being filtered by. In this case our selected option could potentially not be visible at all.
- Will it be clear to a screenreader user where they are within the select? Are there any announcements needed to make this a better experience for screenreader users? Should we consider disabling this for keyboard users

joybh98:
thanks for the input @jcfilben , will start work on them ASAP

joybh98:
@jcfilben can you take a look at my latest push

joybh98:
Can anyone help me with https://github.com/grommet/grommet/pull/6388 ? I have adde the autoScroll prop, and then I am checking when the dropdown is being opened, and then when passing the <StyledSelectDropButton> and the <Box> component, but when checking the storybook, I see the error "autoScroll" is not defined, I have added the prop to the ```propTypes.js```

MikeKingdom:
> Can anyone help me with #6388 ? I have adde the autoScroll prop, and then I am checking when the dropdown is being opened, and then when passing the and the component, but when checking the storybook, I see the error "autoScroll" is not defined, I have added the prop to the `propTypes.js`

There are a number of issues with how you're attempting to do this. The useEffect and useRef can't be done inside a callback. You probably would need to do this within the `SelectContainer`. And my first thought would be you might be able to do it by setting a `show` property on the `InfiniteScroll` in the SelectContainer.

jcfilben:
Closing due to inactivity, feel free to re-open

PR #6391, opened 2022-10-04
Fix file input focus traversal

### LINKED ISSUE #6372: FileInput - tab order and styling
When navigating with the keyboard the tab order of FileInput is unexpected. First our focus is on the 'browse' button hitting tab again moves our focus to the main portion of the FileInput. This can be seen in the video below. It feels strange for focus to move right to left across the screen. Also the main portion of the FileInput gives us context for what the input field is so it should come first in the tab order before the browse button.

Additionally when not using a screenreader the main portion of the FileInput doesn't receive a focusIndicator. When navigating with the keyboard we should visually show a focus indicator on this section.

This issue is reproducable on the FileInput storybook stories: https://storybook.grommet.io/?path=/story/input-fileinput-max-size--max-size

Video of the issue with screenreader output:

https://user-images.githubusercontent.com/54560994/193368791-0fd2f60f-b1dd-4e55-9c74-4228dc80436a.mp4

miqh:
Hello there,

These changes are intended to address #6372 following some discussion over there.

---

#### What does this PR do?
It revises the keyboard navigation focus for the file input component. There's only one focus stop and it corresponds to the actual HTML file input element. Refer to the screen captures I've attached below.

All of the changes are in the file input component implementation.

Manual verification through the file input stories available on Storybook.

By using keyboard navigation on any of the available file input stories.

I haven't included any additional tests to cover the adjusted traversal behaviour, but happy to do so if necessary.

I looked around but didn't spot (or overlooked) any other components which had tests specifically for testing keyboard navigation accessibility.

None.

Refer to #6372

Backward compatibility should not be affected.

---

Closes #6372


src/js/components/FileInput/FileInput.js
```diff
@@ -306,6 +352,7 @@ const FileInput = forwardRef(
                   >
                     {theme.fileInput.button ? (
                       <Button
+                        tabIndex={-1}
```
jcfilben, line 355: Can we add a comment to the places we set the `tabIndex={-1}`?
```suggestion
                        // Button is redundant for keyboard users
                        tabIndex={-1}
```

miqh: @jcfilben, we sure can. I've amended my commit with the extra commentary.

jcfilben (approved):
Looks good!

jcfilben (commented):
One more thing, can we only show the focus style when we are using the keyboard? We do this on other components

miqh:
@jcfilben, sure.

I've brought in the `useKeyboard` hook so that the file input element focus is made conditional on that.

The `Select` implementation is much meatier than `FileInput`, so if this isn't the behaviour you had in mind please point me in the right direction. 🙂

As an aside, thought I'd also mention that the `Anchor` component appears to trigger a focus outline on `mousedown`. This behaviour doesn't look immediately customisable from the `props` available.

jcfilben:
> @jcfilben, sure.
>
> I've brought in the useKeyboard hook so that the file input element focus is made conditional on that.
>
> The Select implementation is much meatier than FileInput, so if this isn't the behaviour you had in mind please point me in > the right direction. 🙂

Nice work, this looks good to me

> As an aside, thought I'd also mention that the Anchor component appears to trigger a focus outline on mousedown. This behaviour doesn't look immediately customisable from the props available.

Yeah I think there are some components that still aren't handling focus in the ideal way. My inclination is that Anchor shouldn't get a focusIndicator unless using the keyboard. I realize we do have a bit of inconsistency across the library with this. I'll bring this up with some other members of the grommet team but feel free to share any more thoughts if you notice other components that aren't handling focus the way you would expect them to.

taysea (approved):
Nice work!

ericsoderberghp (approved):

Issue #6358, opened 2022-09-28
Meter - Support RTL


jcfilben:
Enhance the Meter component to support displaying progress from right to left instead of only supporting left to right.
<img width="397" alt="Screen Shot 2022-09-28 at 11 44 56 AM" src="https://user-images.githubusercontent.com/54560994/192852449-59c2afc5-83f8-426d-a1d9-45dcf5376265.png">

negreirosleo:
@jcfilben May I take this issue?

jcfilben:
> @jcfilben May I take this issue?

Absolutely!

jcfilben:
Added in #6398

PR #6398, opened 2022-10-06
Add RTL support to progress-bar

negreirosleo:
#### What does this PR do?

Adds RTL support to the progress meter,

it works by passing the `rtl` prop to meter:
```js
  <Meter type="bar" rtl value={value} />
```

Or if the `dir` prop on `theme` has the 'rtl' value:

```js
    <Grommet dir="rtl">
        <Meter type="bar" value={value} />
      </Box>
    </Grommet>
```

A story was written for both cases

https://github.com/grommet/grommet/issues/6358

image

Yes, the `rtl` prop should be added to the docs

Yes

It is backwards compatible.


jcfilben (commented):
Nice start on this! A few thoughts...
I am thinking we should allow the direction of meter to be reversed when it is a circle or semicircle not just a bar.
Also I am wondering if we should call the prop `reverse` instead of `rtl`. There could be cases where people want to reverse the direction even with using left to right orientated languages. `reverse` makes the meaning of prop a bit more clear. Anyone using it for right to left purposes will likely set that at the Grommet component level

negreirosleo:
> Nice start on this! A few thoughts... I am thinking we should allow the direction of meter to be reversed when it is a circle or semicircle not just a bar. Also I am wondering if we should call the prop `reverse` instead of `rtl`. There could be cases where people want to reverse the direction even with using left to right orientated languages. `reverse` makes the meaning of prop a bit more clear. Anyone using it for right to left purposes will likely set that at the Grommet component level

In the case of the semicircle, it should rotate counter clockwise right?
also, should we also add it for the vertical direction and remove the dir="rtl" check from the theme?

jcfilben:
> In the case of the semicircle, it should rotate counter clockwise right?

Yes, that's what I'm thinking

> also, should we also add it for the vertical direction and remove the dir="rtl" check from the theme?

I'm a little torn on if we should add it for the vertical direction. I can't think of a use case for someone needing this so maybe we should hold off on adding it for the vertical direction. And setting `dir="rtl"` shouldn't affect the vertical direction meter.

Here is a little more detail of what I'm thinking:
Add a `reverse` prop on the `<Meter>` component. The we leave the `dir="rtl"` as is on the `<Grommet>` component.
When `reverse=true` or `dir="rtl"` the meter should start on the right and move to the left or counterclockwise for circle/semicircle meters.
If `reverse=true` and `dir="rtl"` the meter should start on the left and move to the right or clockwise for circle/semicircle meters.

jcfilben:
Hey @negreirosleo just checking in on this PR to see if you had any questions on this or if you are still interested in working on this

negreirosleo:
@jcfilben sorry taking so long! but I applied the suggestions!

jcfilben:
> @jcfilben sorry taking so long! but I applied the suggestions!

Thanks! I'll take a closer look at the changes tomorrow, but I took a quick scan through the stories and things are looking good there!

jcfilben (commented):
Logic all looks good! I think we just need one storybook story to show the reverse prop. Can the ReverseBarTheme and ReverseCircle stories be removed and unit tests be added for these cases?

negreirosleo:
@jcfilben I'm having a issue where most of the snapshots are failling locally but the one that is failling on the CI is passing here, do you know what this can be?

jcfilben:
> @jcfilben I'm having a issue where most of the snapshots are failing locally but the one that is failling on the CI is passing here, do you know what this can be?

Hey @negreirosleo can you try updating your branch with master locally and then running `rm -rf yarn.lock && rm -rf node_modules && yarn cache clean && yarn install` and then see if the snapshots are still failing? If they are still failing after this and you are using node v18 you can try using node v16

negreirosleo:
that was it, thanks @jcfilben !!

jcfilben (approved):
Thanks for making the changes, this looks good to me!

src/js/components/Meter/Meter.js
```diff
@@ -25,18 +27,26 @@ const Meter = forwardRef(
       thickness = 'medium',
       type = 'bar',
       value,
+      reverse,
```
taysea, line 30:
```suggestion
      reverse: reverseProp,
```

src/js/components/Meter/Meter.js
```diff
@@ -25,18 +27,26 @@ const Meter = forwardRef(
...
+    const theme = useContext(ThemeContext) || defaultProps.theme;
+
     // normalize values to an array of objects
     const values = useMemo(() => {
       if (valuesProp) return valuesProp;
       if (value) return [{ color, value }];
       return [];
     }, [color, value, valuesProp]);
 
+    const reverseDirection =
```
taysea, line 45:
```suggestion
    const reverse =
```

src/js/components/Meter/Meter.js
```diff
@@ -25,18 +27,26 @@ const Meter = forwardRef(
...
+      direction === 'horizontal' &&
+      (theme.dir === 'rtl' || reverse) &&
```
taysea, line 47:
```suggestion
      (theme.dir === 'rtl' || reverseProp) &&
```

src/js/components/Meter/Meter.js
```diff
@@ -25,18 +27,26 @@ const Meter = forwardRef(
...
+      !(theme.dir === 'rtl' && reverse);
```
taysea, line 48:
```suggestion
      !(theme.dir === 'rtl' && reverseProp);
```

src/js/components/Meter/Meter.js
```diff
@@ -49,6 +59,7 @@ const Meter = forwardRef(
           thickness={thickness}
           background={background}
           direction={direction}
+          reverse={reverseDirection}
```
taysea, line 62:
```suggestion
          reverse={reverse}
```

src/js/components/Meter/Meter.js
taysea: Some small naming/code convention suggestions to align with how we handle in other places.

taysea: Can we also move the placement of where this is destructured to be in alphabetical order?

src/js/components/Meter/index.d.ts
```diff
@@ -16,6 +16,7 @@ export interface MeterProps {
   gridArea?: GridAreaType;
   margin?: MarginType;
   max?: number;
+  reverse?: boolean;
```
taysea, line 19: Do we also need to add this to the `propTypes.js` file for Meter?

negreirosleo: I makes sense adding it, so I added!

negreirosleo:
@taysea I applied your suggestions :+1:

taysea (approved):
Nice work!

PR #6400, opened 2022-10-06
Update DataTable groupBy parameters to optional

taysea:
#### What does this PR do?

Errors were being caused for some TS users bc the `groupBy` object notation was requiring all parameters. However, our [storybook example](https://storybook.grommet.io/?path=/story/visualizations-datatable-spacex-grouped--space-x) that is just js doesn't use all the parameters. The TS should mirror the non-typed experience.

src/js/components/DataTable/index.d.ts

Yes. Updated `groupBy` typescript definition.

Backwards compatible.

### DIFF
--- a/src/js/components/DataTable/index.d.ts
+++ b/src/js/components/DataTable/index.d.ts
@@ -91,12 +91,12 @@ export interface DataTableProps<TRowType = any> {
   groupBy?:
     | string
     | {
-        property: string;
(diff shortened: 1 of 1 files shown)


MikeKingdom (approved):
Thanks Taylor

halocline (approved):

ericsoderberghp (approved):

PR #6403, opened 2022-10-06
SelectMultiple - fix object comparison

### LINKED ISSUE #6395: SelectMultiple - pre selected object options not working
### Expected Behavior
Options in `value` should should show as selected

### Actual Behavior
In SelectMultiple if `options` is an array of objects and `value` is initially set to one of the options, the option will not show as selected when the drop is open

https://codesandbox.io/s/grommet-v2-template-forked-ftppqe?file=/index.js

### Steps to Reproduce

1. Go to codesandbox link above
2. Open the drop and see that the 'xyz' option is not checked

### Your Environment

- Grommet version: codesandbox example used v2.26.0 of grommet but I also confirmed this behavior on the stable branch
- Browser Name and version:

jcfilben:
#### What does this PR do?
Fixes an issue in SelectMultiple where objects are not being correctly compared. This PR removes array function instances such as `indexOf` and `includes` which don't work well for arrays containing objects.

Note: marking as a draft, I still need to add jest tests

See the `getOptionIndex` and `arrayIncludes` functions added in the util file.

Tested with existing storybook stories and the following story (see [codesandbox](https://codesandbox.io/s/grommet-v2-template-forked-drqodb?file=/index.js) to see before and after):
```javascript

import { SelectMultiple } from '../SelectMultiple';
const options = [
  { fieldType: 'SYSTEM', value: 'abc', display: 'ABC', other: 123 },
  { fieldType: 'SYSTEM', value: 'xyz', display: 'XYZ', other: 456 },
];
export const Test = () => {
  const [value, setValue] = useState([
    {
      fieldType: 'SYSTEM',
      value: 'abc',
      display: 'ABC',
      other: 123,
    },
  ]);
  return (
    <SelectMultiple
      showSelectedInline
      labelKey="display"
      valueKey="value"
      options={options}
      value={value}
      onChange={({ value }) => setValue(value)}
    />
  );
};

export default {
  title: 'Input/SelectMultiple/Test',
};


```

Closes #6395 

No
Yes
Backwards compatible


src/js/components/Select/utils.js
```diff
@@ -17,7 +17,25 @@ export const getOptionLabel = (index, options, labelKey) =>
 export const getOptionValue = (index, options, valueKey) =>
   applyKey(options[index], valueKey);
 
-export const useDisabled = (disabled, disabledKey, options, valueKey) =>
+export const getOptionIndex = (options, i) => {
+  if (options) {
+    if (typeof i === 'object')
+      return options.findIndex((x) => JSON.stringify(x) === JSON.stringify(i));
```
taysea, line 23: Is there any possibility that the key/value pairs of the object might be ordered differently between `x` and `i`? Wondering if a `deepEquals` type function would be more reliable than JSON.stringify

jcfilben: Good point! Would you be okay if I created a separate issue for this instead of handling it in this PR? I think the JSON.stringify will cover the 80% case, but I agree it would be nice to have a more robust method for object comparison. I'm thinking the `deepEquals` type of function will take a bit of time writing and testing to make sure it is working as expected and feel like it would be better to create a separate issue rather than push back the release

ericsoderberghp: I'm wondering if this is being too strict. For instance, if the object in options has something unrelated to the value, like the label changes. I think we should be using valueKey for comparison whenever possible when we have object options.

jcfilben: That's fair, I will go back and make that change

taysea (approved):
Nice, looks good to me!

ericsoderberghp (approved):

src/js/components/Select/utils.js
ericsoderberghp: that feels better, thanks

taysea:
Related to https://github.com/grommet/grommet/issues/5198

Issue #6393, opened 2022-10-04
Select - When value is an object, the aria-label should be human readable


halocline:
### Expected Behavior

When value is an object, the aria-label should be human readable, such as using the provided labelKey.

### Actual Behavior

Currently the label reads `aria-label="...; Selected: [object Object]`.

This probably involves work regarding the a11yTitle being set on StyledSelectDropButton.

### URL, screen shot, or Codepen exhibiting the issue

Jest test for `Select select an option with complex options 1` shows the aria-label being read as stated above.

[Snapshot test](https://github.com/grommet/grommet/blob/48663b7c6c520e00fa17d916b0a9d43d55c822b2/src/js/components/Select/__tests__/__snapshots__/Select-test.js.snap#L14367)

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: v2.26.0
- Browser Name and version:
- Operating System and version (desktop or mobile):

illiteratewriter:
Can I work on this?

britt6612:
@illiteratewriter of course, feel free to let us know if you need any assistance.

illiteratewriter:
Some snapshots are failing without me changing anything at all. mostly in the DateInput component.  What should I do about those? Is it because I am in a different time zone or something?

jcfilben:
Hey @illiteratewriter someone else also ran into this issue, I am thinking it is timezone related and there are adjustments needed to our tests or directly to the DateInput component. **For now you can make commits with the `--no-verify` flag**. We have the environment set up so that the tests will also run on Github so we can confirm there that the tests are behaving as expected (other than having timezone related issues).

For debugging purposes what timezone are you working in?

PR #6407, opened 2022-10-08
make aria-label readable for Select

illiteratewriter:
#### What does this PR do?

make aria-label readable for Select component

Select component

Check the aria-labels for Select examples in the Storybook

#6393 

no

### DIFF
--- a/src/js/components/Select/Select.js
+++ b/src/js/components/Select/Select.js
@@ -278,19 +278,35 @@ const Select = forwardRef(
 
     const iconColor = getIconColor(theme);
 
+    let ariaLabelGenerated = '';
+    if (value) {
+      if (Array.isArray(value)) {
+        ariaLabelGenerated = `; ${value.length || 0} selected`;
+      } else if (typeof value === 'object' && value !== null) {
+        ariaLabelGenerated = `; Selected: ${inputValue}`;
+      } else {
+        ariaLabelGenerated = format({
+          id: 'select.selected',
+          messages,
+          values: {
+            // eslint-disable-next-line no-nested-ternary
+            currentSelectedValue: valueKey?.reduce
+              ? inputValue
+              : displayLabelKey
+              ? value[displayLabelKey]
+              : value,
+          },
+        });
+      }
+    }
+
     return (
       <Keyboard onDown={onRequestOpen} onUp={onRequestOpen}>
         <StyledSelectDropButton
           ref={ref}
-          a11yTitle={`${ariaLabel || a11yTitle || placeholder || 'Open Drop'}${
-            value
-              ? format({
-                  id: 'select.selected',
-                  messages,
-                  values: { currentSelectedValue: value },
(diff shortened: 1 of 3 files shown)


src/js/components/Select/Select.js
```diff
@@ -278,19 +278,35 @@ const Select = forwardRef(
 
     const iconColor = getIconColor(theme);
 
+    let ariaLabelGenerated = '';
+    if (value) {
+      if (Array.isArray(value)) {
+        ariaLabelGenerated = `; ${value.length || 0} selected`;
```
jcfilben, line 284:
```suggestion
        ariaLabelGenerated = `; ${value.length || 0} ${
          value.length === 1 ? 'option' : 'options'
        } selected`;
```
Can we adjust this so it reads out `<x> options selected` instead of `<x> selected`. This helps clarify in the case where we might have options with numerical names.

jcfilben (commented):
Coming along nicely, just a small comment

src/js/components/Select/Select.js
illiteratewriter: This sounds better. `SelectMultiple` component had `<x> selected` that's why I went with that. Maybe we can make another PR to make ariaLabel consistent in `SelectMultiple` also.

jcfilben: Yes, good idea I think this change could be made in SelectMultiple too

src/js/components/Select/__tests__/__snapshots__/Select-multiple-test.js.snap
```diff
@@ -489,7 +489,7 @@ exports[`Select Controlled multiple 1`] = `
 <button
   aria-expanded="false"
   aria-haspopup="listbox"
-  aria-label="Open Drop; Selected: "
+  aria-label="Open Drop; 0 selected"
```
jcfilben, line 492: Can we change Select so when `multiple` is true the aria-label reads "x options selected" or "1 option selected" if there is only one?

src/js/components/Select/Select.js
```diff
@@ -278,19 +278,37 @@ const Select = forwardRef(
 
     const iconColor = getIconColor(theme);
 
+    let ariaLabelGenerated = '';
+    if (value) {
+      if (Array.isArray(value)) {
+        ariaLabelGenerated = `; ${value.length || 0} ${
```
ericsoderberghp, line 284: Why aren't we using `format()` for all of these cases? As shown, if the `value` is an Array, we won't be using the message catalog to localize the aria label. I was expecting we would introduce more keys in the catalog for the various permutations, allowing simple localization of each permutation.

illiteratewriter:
Hi, I'm sorry but I will not be working on this due to some personal emergencies.

jcfilben:
Thanks for letting us know, hope everything is okay!
I'm going to go ahead and close this pull request

PR #6408, opened 2022-10-08
Changed DateInput to fix an issue with range format and no value

ericsoderberghp:
#### What does this PR do?

Changed DateInput to fix an issue with range format and no value.

DateInput-test change, to see the problem
One character change to production code, but also a new unit test

added a unit test

followed pattern of existing tests

uncovered while using DateInput with grommet-designer

no

yes

backwards compatible


jcfilben (approved):
Looks good!

Issue #6376, opened 2022-10-01
Text & Box alignSelf prop is missing ability to use baseline value


samuelgoff:
### Expected Behavior

`Text` & `Box` `alignSelf` prop is missing ability to use baseline value. User should be able to provide `baseline` as a value to `alignSelf` prop. (Docs accurately reflect that support for this value is missing.)

Ideally, **_all_** supported values should be accessible through `align`, `alignSelf`, and `alignContent` props, support for which is variable.

### Actual Behavior

If the user provides `baseline` to `alignSelf` prop for `Text` or `Box` component, the value in CSS shows up as `undefined`, which gets ignored by the browser.

### URL, screen shot, or Codepen exhibiting the issue
![Screen Shot 2022-10-01 at 3 24 51 PM](https://user-images.githubusercontent.com/2007084/193427118-19ad3fcd-3294-4047-b428-05885012bf9c.png)

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

jcfilben:
Feature added in #6413

PR #6413, opened 2022-10-11
feat: add baseline on alignSelf attribute

matheushahnn:
#### What does this PR do?
Add baseline as one of the alignSelf possible attribute

https://github.com/grommet/grommet/issues/6376

Yes


jcfilben (commented):
Nice work, just a few things...
Can you add a jest test to cover this enhancement? Also can you create a pull request on [grommet-site](https://github.com/grommet/grommet-site) to add 'baseline' to the docs?

matheushahnn:
> Collaborator

Sure thing. What kind of test do you imagine? Just test snapshots?

matheushahnn:
Created the PR on grommet-site: https://github.com/grommet/grommet-site/pull/428

jcfilben:
> Sure thing. What kind of test do you imagine? Just test snapshots?

Yeah i think we can just do a snapshot test to make sure the styling was correctly applied when we use `alignSelf="baseline"`

matheushahnn:
Idk why, but when I add a new Box inside Box/alignSelf test 2 tests fail when rendering, and also I see this message: 

```
Uncaught 'Over 200 classes were generated for component StyledBox with the id of "StyledBox-sc-13pk1d4-0".\n' +
      'Consider using the attrs method, together with a style object for frequently changed styles.\n' +
      'Example:\n' +
      '  const Component = styled.div.attrs(props => ({\n' +
      '    style: {\n' +
      '      background: props.background,\n' +
      '    },\n' +
      '  }))`width: 100%;`\n' +
      '\n' +
      '  <Component />'
```

Have you already seen this?

**NOTE** If I remove one Box (e.g.: `<Box alignSelf="stretch" />` the tests pass

jcfilben:
Just checked and I am getting that same error on my end. I think it is related to how many versions of styled boxes we have in that test file. I think in this case it makes sense to suppress the error in this file because we know that the reason we are over limit is just because of how many test cases we have.

matheushahnn:
I am unsure how to suppress the error in this file, any suggestions?
If I move one test to another file it works, maybe we can split these tests into two files, what do you think?

jcfilben:
Hmm yeah I am not finding any good options for disabling the error for the test file. If possible I think we should avoid splitting this up into two test files. Ideally it would be nice to keep things organized under one file. I'll keep looking to see if it can find any other possible solutions...

matheushahnn:
I mode one style to attrs on StyledBox as the error suggested and now the tests are passing. 
Please, take a look and let me know if it is one possible approach.

halocline:
> I mode one style to attrs on StyledBox as the error suggested and now the tests are passing. Please, take a look and let me know if it is one possible approach.

I think this ☝️ is the right direction. I believe @jcfilben was also exploring this approach.

jcfilben:
Hey @matheushahnn I'm thinking this is probably the direction we will go. I'm going to look at it a bit more on Monday, I'm thinking about if we want to move additional items in styledBox within attrs not just `wrap`

jcfilben:
Update: I chatted with some other members of the team and we are thinking that we do want to use `attrs` for styled box. However, these changes are going to happen in a separate pull request. For this PR I think we can remove the change to styled box, and comment out the test and just add a comment explaining that the test is being commented out until we change styled box to use `attrs`.

matheushahnn:
@jcfilben Followed your instructions 👍

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

Issue #6124, opened 2022-05-11
Storybook Accessibility Violations


jcfilben:
Recently the addon-a11y tool was added to storybook which has flagged a number of accessibility violations. To see the violations go to [storybook](https://storybook.grommet.io/?path=/story/all--all) and below each story click the accessibility tab. It would be great to clean up these accessibility violations. Some fixes may just require changes to story while others will require changes to grommet components. Below is a list of stories with accessibility violations. Feel free to contribute fixes per story or per component.

- [ ] All (need design help - Calendar color contrast & Calendar heading level https://github.com/grommet/grommet/issues/7450)

**Accordion**
- [x] Header https://github.com/grommet/grommet/pull/6383
- [x] CustomThemed/Rich

**Drop**
- [x] https://github.com/grommet/grommet/pull/6384
- (Inline, Overflow, Plain, Simple, Styled)

**Menu**
- [x] Children #6410 

**Pagination**

- [x] (component wide error) `<ul>` and `<ol>` must only directly contain `<li>`, `<script>` or `<template>` elements
- [x] Number edge pages
- [x] Number middle pages
- [x] Simple
- [x] Size

**CheckBox**
- [x] 	Inside a Button
- [x] No label https://github.com/grommet/grommet/pull/6449
	
**FileInput** 
- [x] https://github.com/grommet/grommet/pull/6417
	
**Form**
- [x] Aggreagate validaton https://github.com/grommet/grommet/pull/6469
- [x] Array of Form Fields https://github.com/grommet/grommet/pull/6464
- [x] Controlled
- [x] Controlled Input
- [x] Controlled Input lazy
- [x] Controlled lazy
- [x] Dynamic fields https://github.com/grommet/grommet/pull/6465
- [ ] Field states (need design help)
- [x] Field with children
- [x] Field with component prop
- [ ] Help and error (need design help)
- [x] Typed form
- [x] Uncontrolled
- [x] Validate on blur https://github.com/grommet/grommet/pull/6466
- [x] Validate on change
- [ ] Validate on mount (need design help)
- [x] Custom
- [ ] Field spacing options (need design help)
- [ ] Required label (need design help)
- [x] Themed

**FormField**
- [x] Simple
- [x] Custom

**MaskedInput**
- [x] https://github.com/grommet/grommet/pull/6448

**RangeInput**
- [x] https://github.com/grommet/grommet/pull/6423

**Select**
- [x] (component wide error) Interactive controls must not be nested

**TextArea**
[...]

PR #6417, opened 2022-10-12
Fix Storybook accessibility violations for the file input (#6124)

miqh:
Hello there,

This should fix up the accessibility violations reported across some of the `FileInput` stories.

It seems like using a `label` element, explicit or implicit, appears to be the preferred way to work around such violations. In the interest of preserving the existing story visuals, I opted for `aria-label`. If the former is more desirable, I can update my changes.

---

#### What does this PR do?
Resolve accessibility violations for `FileInput` stories.

Changes are confined to the file input stories reporting violations.

Confirmed the violations no longer appear in the Storybook.

As above.

N/A.

None.

Refer to #6124.

N/A.

Don't think so.

Up to maintainers to decide.

Should have no impact on consumers of the library.

### DIFF
--- a/src/js/components/FileInput/stories/CustomThemed/Custom.js
+++ b/src/js/components/FileInput/stories/CustomThemed/Custom.js
@@ -47,6 +47,7 @@ export const Custom = () => (
     <Box fill align="center" justify="start" pad="large">
       <Box width="medium">
         <FileInput
+          aria-label="Choose files"
           renderFile={(file) => (
             <Box>
               <Text weight="bold" truncate>
--- a/src/js/components/FileInput/stories/MaxFileCount.js
+++ b/src/js/components/FileInput/stories/MaxFileCount.js
@@ -8,6 +8,7 @@ export const MaxFileCount = () => (
       <Form validate="submit">
         <FormField name="fileInput" htmlFor="fileInput" required>
           <FileInput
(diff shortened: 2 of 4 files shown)


jcfilben (commented):
This looks good, can we also add an aria-label to the FileInput/Custom/CustomThemed story as well? I know it was left out from the issue description checklist but just happened to notice it while I was reviewing.

miqh:
@jcfilben, can do.

Until now that you've pointed out, I didn't even notice the theme switcher option in Storybook. My bad!

A quick look after, it seems like there are additional colour contrast violations in the "hpe" theme. What's your preferred approach with resolving these for now given that any colour tweaks will affect component consumers?

jcfilben:
Good catch on the hpe theme! For now let's just leave the hpe ones as they are. I will bring the issue up with other members of the grommet team and we can decide how to move forward on the hpe theme issue, but don't worry about it for now in this pull request.

miqh:
@jcfilben, got it.

Beyond `FileInput`, there are other components which exhibit that colour contrast violation (though I'm sure you're probably already aware). If you reach a consensus on what to do about that, then those violations can hopefully be dealt with in a uniform manner as well.

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

PR #6418, opened 2022-10-12
Correct types for onChange() params of FileInput

undeletable:
#### What does this PR do?

It corrects types definitions for additional params of `onChange` property of `FileInput` component.

`FileInput` component.

Manual testing with the app that's using TypeScript.

Add `FileInput` with `onChange` property to the render of your component. Try to describe params of `onChange` precisely.

https://github.com/grommet/grommet/pull/6145 (the fix is incomplete).

No.

I think yes.

This change might break an app which has incorrect handling of second parameter of `onChange` (e.g., referring to property of `files` which doesn't actually exist, but TS wouldn't warn about this with current type definition).

### DIFF
--- a/src/js/components/FileInput/index.d.ts
+++ b/src/js/components/FileInput/index.d.ts
@@ -24,8 +24,8 @@ export interface FileInputProps {
   name?: string;
   onChange?: (
     event?: React.ChangeEvent<HTMLInputElement>,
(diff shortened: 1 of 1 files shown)


jcfilben (approved):
Looks good!

ericsoderberghp (approved):

Issue #6421, opened 2022-10-13
DateInput - incorrect timezone offset being applied


halocline:
### Expected Behavior

When defaultValue has no timezone, everything is returned without a timezone and is assumed to be local time. The value in state, input, and active calendar date should all display the same value.
![Screen Shot 2022-10-13 at 11 21 45 PM](https://user-images.githubusercontent.com/1756948/195670055-6ae0d208-4cf1-4408-bca7-28739cd1d475.png)

### Actual Behavior

The value in state is not in sync with input and active calendar date.
![Screen Shot 2022-10-13 at 11 23 27 PM](https://user-images.githubusercontent.com/1756948/195670393-efa3ebab-966b-41c2-96c8-70096d4a7a5e.png)

This is a similar issue to what is reported by https://github.com/grommet/grommet/issues/6356.

The fix provided in https://github.com/grommet/grommet/pull/6396 should be applied in a few other locations:
- https://github.com/grommet/grommet/blob/aad44bf01e4b1bff0ef8f34ddcefaf8e841b8921/src/js/components/DateInput/DateInput.js#L48
- https://github.com/grommet/grommet/blob/aad44bf01e4b1bff0ef8f34ddcefaf8e841b8921/src/js/components/DateInput/utils.js#L63
- And possibly: https://github.com/grommet/grommet/blob/aad44bf01e4b1bff0ef8f34ddcefaf8e841b8921/src/js/components/Calendar/utils.js#L21

Since this logic is duplicated in multiple locations, consider moving to a single util function.

```
const adjustedDate = new Date(dateValue);
    // if time is not specified in ISOstring, normalize to midnight
    if (dateValue.indexOf('T') === -1) {
      const offset = adjustedDate.getTimezoneOffset();
      const hour = adjustedDate.getHours();
      adjustedDate.setHours(hour, offset < 0 ? -offset : offset);
    }
    result = adjustedDate;
```

### Steps to Reproduce

1. Use this story: https://storybook.grommet.io/?path=/story/input-dateinput-format-inline--format-inline
2. Specifically scenarios 3 & 5
3. Set OS timezone to IST.

### Your Environment

- Grommet version: 2.27.0
- Browser Name and version: all
- Operating System and version (desktop or mobile): MacOS 12.6

PR #6424, opened 2022-10-14
feat: set new hours considering the offset

matheushahnn:
#### What does this PR do?
It fixes the issue related to timezone on Calendar

Closes https://github.com/grommet/grommet/issues/6421

Test this storybook https://storybook.grommet.io/?path=/story/input-dateinput-format-inline--format-inline

https://github.com/grommet/grommet/issues/6421


matheushahnn:
I still need to change on [grommet/src/js/components/Calendar/utils.js](https://github.com/grommet/grommet/blob/aad44bf01e4b1bff0ef8f34ddcefaf8e841b8921/src/js/components/Calendar/utils.js#L21)

halocline:
@matheushahnn First, thank you for picking this issue up.

> I still need to change on [grommet/src/js/components/Calendar/utils.js](https://github.com/grommet/grommet/blob/aad44bf01e4b1bff0ef8f34ddcefaf8e841b8921/src/js/components/Calendar/utils.js#L21) In fact, I am not sure how it should do in this file, because it is adding days. Do you know where it is failing this utils and how I can reproduce locally?

This is actually performing as expected. I flagged this as an opportunity to consolidate logic as it looked like we are doing something somewhat related to the timezone offset work, but in a slightly different manner. I did not study closely.

If your judgement and recommendation is that this should remain as is, please let us know.

The rest of the work is looking great, thank you.

matheushahnn:
I think we can keep it as it is.
Please, let me know what else I should do in this PR.

src/js/utils/dates.js
```diff
@@ -0,0 +1,11 @@
+export const setHoursWithOffset = (date) => {
```
jcfilben, line 1: It looks like lines 89-95 of Calendar.js could be removed and we could just call this function instead

matheushahnn: Good call! 
Done

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

PR #6427, opened 2022-10-14
Feat: improve formatBytes function

### LINKED ISSUE #5807: Enhance FileInput to recognize user operating system for maxSize calculation
### Expected Behavior

FileInput should recognize what operating system a user is on when determining if `1024` (binary) or `1000` (decimal) should be used as the conversion factor between bytes, KB, etc.

Right now, FileInput uses binary value (1024) of bytes as a conversion factor to determine if a file has exceeded the max size or not. This seems to be the better option (vs decimal conversion factor of 1000) because it ensures that all windows users will experience and accurate comparison between their file browser and Grommet's FileInput. Previously, window's users were being flagged by Grommet's FileInput that their file was too large even though their file browser said it was less than the max size (their file browser used 1024, while grommet used 1000).

While our current solution provides users (regardless of their operating system) with a reasonable experience, it may be nice to enhance FileInput in the future to be slightly more intelligent.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

Isaac-alencar:
#### What does this PR do?
This PR improve `formatBytes` function by making it more dynamic. Based on OS user is on we change the conversion factor
 from 1000 to 1024.

 FYI: 
  Windows assumes that there are 1024 Bytes in a Kilobyte unit, and 1024 Kilobytes in a Megabyte unit etc.
  while Ubuntu and Mac assumes, a 1000 bytes constitute a Kilobyte (KB) unit, 1000 Kilobytes for a Megabyte (MB) and so on.

That is, same file on ubuntu and mac are larger than windows

For example: 
  - 213066 ÷ 1024 = 208072 bites
  - 213066 ÷ 1000 = 213066 bites

So, if you try to attach same file in different OS you're gonna get an error

`src/js/components/FileInput/FileInput.js`

closes #5807


src/js/components/FileInput/FileInput.js
```diff
@@ -17,9 +17,34 @@ import { Text } from '../Text';
...
+  const SI_CONVERSION_FACTOR = 1000;
+  const IEC_CONVERSION_FACTOR = 1024;
+
+  const osConversionFactor = {
+    Win: IEC_CONVERSION_FACTOR,
+    Linux: SI_CONVERSION_FACTOR,
+    Mac: SI_CONVERSION_FACTOR,
+  };
+
+  const currentOS = ['Win', 'Linux', 'Mac'].find(
```
jcfilben, line 30: Tested with a Mac on Chrome but `currentOS` returned `undefined`

Isaac-alencar:
@jcfilben thanks for reviewing this, I'll work to fix it

jcfilben (approved):
This looks good to me, tested on Chrome, Safari, and FireFox on Mac

jcfilben:
@MikeKingdom tagging you for review since you have a windows pc and I'm thinking it would be good to check this on different operating systems

src/js/components/FileInput/FileInput.js
```diff
@@ -17,9 +17,33 @@ import { Text } from '../Text';
...
+  }
+
+  return SI_CONVERSION_FACTOR;
+};
+
 const formatBytes = (size) => {
-  const units = ['B', 'KB', 'MB', 'GB', 'TB'];
-  const factor = 1024;
+  const units =
+    getConversionFactor() === 1000
```
MikeKingdom, line 43: Since you have constants I'd use them (ie. move the factor const outside and do
```jsx
getConversionFactor() === SI_CONVERSION_FACTOR;
```

MikeKingdom (commented):
I'll check it on windows shortly but in the meantime one small comment...

MikeKingdom (commented):
I tested this on windows and it is correct.

However I must note that windows' file explorer uses units of KB, MB etc even though it's technically KiB or MiB. If you're wanting the numbers to match they still won't because windows shows the `Size` detail column always in KB. The numbers won't match between the maxsize message and what windows shows in the detail column because windows will still show sizes in KB even if larger than 1024 KB. For example, windows file explore will show a file as 29,820 KB instead of 29.1 MB, In the property dialog for the file it will show "29.1 MB (30,535,680 bytes)". It's still not saying MiB but it will be technically MiB.

MikeKingdom:
> Said that, do you think is a good approach create a business rule on the code only for windows conversion?

If the goal is to make it was easier to know if your file is too big by looking at the file explorer details column (which is what you'll see when you click Browse to choose the file assuming that file dialog is set to the "details" view) then I'd think for windows it should show in KiB (the 1024 version)

Isaac-alencar:
@MikeKingdom I've done some changes, so when you have a chance, could you review?

src/js/components/FileInput/utils/formatBytes.js
```diff
@@ -0,0 +1,45 @@
...
+};
+
+const windowsFormat = (size) => {
+  const num = Math.ceil(size / IEC_CONVERSION_FACTOR);
+
+  function formatNumber(number) {
+    return number.toString().replace(/(\d)(?=(\d{3})+(?!\d))/g, '$1.');
+  }
+
+  return `${formatNumber(num)} ${'KB'}`;
```
MikeKingdom, line 31: Don't  need the ${}:
```jsx
return `${formatNumber(num)} KB`;
```

MikeKingdom (commented):
Thanks Isaac. I think it needs a couple tweaks on what it does for windows

src/js/components/FileInput/utils/formatBytes.js
```diff
@@ -0,0 +1,45 @@
...
+  let num = size;
+  while (num >= factor && index < units.length - 1) {
+    num /= factor;
+    index += 1;
+  }
+  return `${num.toFixed(1)} ${units[index]}`;
+};
+
+const windowsFormat = (size) => {
+  const num = Math.ceil(size / IEC_CONVERSION_FACTOR);
```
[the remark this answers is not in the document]
Isaac-alencar: The `Math.round()` function returns the value of a number rounded to the nearest integer, whereas `Math.ceil()`  function always rounds up and returns the smaller integer greater than or equal to a given number (like windows does). 

ex:
```js
console.log(Math.ceil(.95));
// expected output: 1

console.log(Math.ceil(4));
// expected output: 4

console.log(Math.ceil(7.004));
// expected output: 8

console.log(Math.ceil(-7.004));
// expected output: -7
```

Isaac-alencar:
@MikeKingdom I send some updates!

src/js/components/FileInput/utils/formatBytes.js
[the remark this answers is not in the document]
MikeKingdom: Sure and that avoids it showing 0 KB. That probably makes sense for the limit.

src/js/components/FileInput/utils/formatBytes.js
```diff
@@ -0,0 +1,45 @@
...
+
+  function formatNumber(number) {
+    return Intl.NumberFormat().format(number);
+  }
+
+  return `${formatNumber(num)} KB`;
```
MikeKingdom, line 31: Looks good and thanks for making the change. It might be nice to simplify this to
```jsx
return `${Intl.NumberFormat().format(num)} KB`;
```
since the formatNumber function is trivial and not used elsewhere.

Isaac-alencar: done!

Isaac-alencar:
Hi @MikeKingdom could you take a look again? I sent some updates 😃

src/js/components/FileInput/utils/formatBytes.js
```diff
@@ -0,0 +1,41 @@
...
+    num /= factor;
+    index += 1;
+  }
+  return `${num.toFixed(1)} ${units[index]}`;
+};
+
+const windowsFormat = (size) => {
+  const num = Math.ceil(size / IEC_CONVERSION_FACTOR);
+
+  return Intl.NumberFormat().format(num);
```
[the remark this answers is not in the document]
Isaac-alencar: yup, I'll change to 
```js
return ${Intl.NumberFormat().format(num)} KB;
```
It is incorrect (technically speaking), but windows don't show to it's users with `KiB`

MikeKingdom (approved):
Looks good. Thanks!

ericsoderberghp (approved):

src/js/components/FileInput/utils/formatBytes.js
```diff
@@ -0,0 +1,41 @@
+const SI_CONVERSION_FACTOR = 1000;
+const IEC_CONVERSION_FACTOR = 1024;
+
+const getCurrentOS = () => {
+  const currentOS = ['Win', 'Linux', 'Mac'].find(
+    (v) => window.navigator.userAgent.indexOf(v) >= 0,
```
taysea, line 6: In the HPE Design System site, I'm getting an error when I try to run locally (with latest Grommet stable) of window is not defined. Think we might need to add some fallback for the case that window isn't defined yet?

Screen Shot 2022-11-03 at 1 44 35 PM

Isaac-alencar: Thanks to share it!. I'll work on it 👍🏻

jcfilben: Looks like a PR for this has already been filed https://github.com/grommet/grommet/pull/6480 to fix this. Appreciate the willingness to work on this though!

PR #6429, opened 2022-10-16
Fix a11y violation in Notification/Global Storybook

julieg18:
#### What does this PR do?

~Fixes accessibility violation by using a div instead of a header in `Notification/Global` storybook~ Fixes accessibility violation by switching a `Box` for a `Main` in `Notification/Global` storybook

`Notification/Global` storybook

`yarn test` was run before committing

Check the accessibility tab in the `Notification/Global` storybook

#6124

image

No

No

No


src/js/components/Notification/stories/Global.js
```diff
@@ -43,10 +43,10 @@ export const Global = () => {
           />
         )}
         <Box width="large" margin="auto" pad="medium" gap="medium">
-          <Header>
+          <Box direction="row" justify="between">
```
jcfilben, line 46: Instead of changing this from `Header` to `Box` I think we should change the `Box` at line 29 to `Main`. This ensures that this container is still correctly marked as a `Header`, it just won't have the banner role.

julieg18: Great idea! Added the header back and added the `Main`.

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

PR #6432, opened 2022-10-18
Changed Box to reduce the generation of common styled component classes

ericsoderberghp:
#### What does this PR do?

Changed Box tests to avoid running into styled-components exception about generating over 200 classes.

Given the nature of Box, where there are many properties that affect styling in a variety of ways, it is natural that a unit test covering all possible permutations would stumble across this exception. The simple solution is to split up the unit tests for Box into separate files.

Box unit tests

unit tests

no test changes, just re-organization

This was uncovered as part of https://github.com/grommet/grommet/pull/6413

no

yes

backwards compatible


jcfilben (approved):
Looks good!

Issue #6333, opened 2022-09-16
DataTable header pin not working with paginate.


rollsover:
Paginated DataTable in Box container not pins header.

### URL, screen shot, or Codepen exhibiting the issue

[CodeSandbox](https://codesandbox.io/s/grommet-v2-not-working-pin-with-paginate-ctth91)

### Steps to Reproduce

See CodeSandbox.

### Your Environment

- Grommet version: 2.23.0
- Browser Name and version: CodeSandbox
- Operating System and version (desktop or mobile): MacOS

jcfilben:
Closing since #6439 has been merged.

PR #6439, opened 2022-10-19
fix: keep DataTable container filled when paginate

matheushahnn:
#### What does this PR do?
It will keep the fill style when DataTable has pagination

Manual test with other storybooks and one unit test

Copy the component from https://codesandbox.io/s/grommet-v2-not-working-pin-with-paginate-ctth91 

https://github.com/grommet/grommet/issues/6333

### DIFF
--- a/src/js/components/DataTable/DataTable.js
+++ b/src/js/components/DataTable/DataTable.js
@@ -352,7 +352,7 @@ const DataTable = ({
   // should remain in its location
   const OverflowContainer = paginate ? Box : Fragment;
   const overflowContainerProps = paginate
-    ? { overflow: { horizontal: 'auto' }, flex: false }
+    ? { overflow: { horizontal: 'auto' }, flex: false, fill }
     : undefined;
 
   // necessary for Firefox, otherwise paginated DataTable will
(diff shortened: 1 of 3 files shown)


jcfilben (commented):
I tested with the following code and I am still seeing the issue
```javascript
<Box margin="large" height="200px" overflow="auto">
  <DataTable
    pin="header"
    columns={columns}
    data={DATA}
    step={5}
    paginate
  />
</Box>
```

matheushahnn:
You're right, thank you for showing me that issue.
Sorry about the delay in responding.
<hr />

There is a comment about not enabling flex on https://github.com/grommet/grommet/blob/83d9e9c4f7d1538ef1dff0c3b3c5cc315ee4b7f7/src/js/components/DataTable/DataTable.js#L355
but if I set as true it will work, so this is one option to keep the header sticky
image

I am investigating another way to consider pagination height in the table content to keep pagination inside scroll.

jcfilben:
> lmk if it is ok, or there is a case that it breaks, please

I think moving the pagination controls into the overflow container may not be the best user experience. I would expect to be able to switch pages without having to scroll through the table. Also if `pin=true` and both the header and footer of the table are pinned it seems really odd that you still have to scroll through the table to get to the pagination controls.

It looks like setting `flex: true` is working well. I think if we just do that and leave the paginated controls outside of the overflow container that could work.

taysea (approved):
Behavior looks good to me, I'm trying to remember why we'd initially decided to do `flex: false` with Pagination PR.

ericsoderberghp (approved):

Issue #6175, opened 2022-06-10
Enhancement - Responsive sizing of Paragraph and Text


halocline:
### Expected Behavior

Heading has the ability to scale down font size at specified responsive breakpoint. Paragraph and Text font sizes should have the ability to scale down as well to maintain proportion in relation to Heading font.

Note: Padding, margins, and gap also have the ability to scale down at specified responsive breakpoint.

### Actual Behavior

Paragraph and Text do not currently have the functionality to scale responsively.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: v2.24.0
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #6451, opened 2022-10-23
Enhancement - #6175: Added option for responsive Paragraph and Text text

georgiabains:
#### What does this PR do?
This PR mainly modifies the Paragraph and Text schemas and components to add a `responsiveBreakpoint` that is used to scale down font sizes for resolutions under that breakpoint. The method follows that of `Heading` quite closely, however I did have to change the schema to group the font sizes within one object, in order to create an ordered array to loop through to determine the appropriate smaller size.

Looking at `base.js`, `StyledParagraph.js` and `StyledText.js`.

I ran `yarn test-update` and `yarn test` before each commit and created Paragraph and Text components in my local host, set up with `responsive={true}` and tested the different possible `size`s. I also tried inputting `size`s that didn't exist, to confirm that text still displayed.

In a localhost, set up a Paragraph and Text component with the property `responsive={true}`. Confirm that resolutions under 768px have a smaller font size. Further testing can be done by changing the `size` property.

N/A

I followed the existing coding structure that the Heading component uses.

[Link to github issue (6175)](https://github.com/grommet/grommet/issues/6175)

N/A

No

I'm afraid I might be a little unfamiliar with Grommet, do users have the ability to directly target theme props? If so then it should be, so that users can update any references to `props.theme.paragraph` and `props.theme.text`.

I believe its a breaking change, as modifying the schema for Text resulted in modifying references to `props.theme.text`, which are used in other components.

### DIFF
--- a/src/js/components/Anchor/StyledAnchor.js
+++ b/src/js/components/Anchor/StyledAnchor.js
@@ -12,7 +12,7 @@ const disabledStyle = `
 const sizeStyle = (props) => {
   if (props.size) {
     const size = props.size || 'medium';
-    const data = props.theme.text[size];
+    const data = props.theme.text.fontSize[size];
     return css`
       font-size: ${data ? data.size : size};
       line-height: ${data ? data.height : 'normal'};
--- a/src/js/components/Button/StyledButton.js
+++ b/src/js/components/Button/StyledButton.js
@@ -30,7 +30,7 @@ const radiusStyle = (props) => {
 
 const fontStyle = (props) => {
   const size = props.sizeProp || 'medium';
-  const data = props.theme.text[size];
+  const data = props.theme.text.fontSize[size];
   return css`
     font-size: ${data.size};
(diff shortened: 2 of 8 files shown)


jcfilben:
Just wanted to give a quick update on this pull request. I have been looking into it, the problem is changing the theme structure to include `fontSize` will cause backwards compatibility issues. However I realize we need the sizes to be grouped together so we can move up/down a size. I'm trying to think of other ways we could do this that won't cause backwards compatibility issues.

georgiabains:
Hey @jcfilben thanks for responding! That solution makes way more sense rather than forcing countless users to update their codebases 😅

I can work on making this update this week 👍

jcfilben:
> Hey @jcfilben thanks for responding! That solution makes way more sense rather than forcing countless users to update their codebases 😅
>
> I can work on making this update this week 👍

Perfect!

Issue #6270, opened 2022-08-11
Toast Notification Timing Adjustable


jcfilben:
### Expected Behavior
Toasts need to support a way to extend timing or set a new time limit. See https://www.w3.org/TR/WCAG21/#timing-adjustable

### Actual Behavior
Currently the time limit is controlled by the theme. To adjust it would require changing the theme. We don't support a good way to extend the time limit of a toast before it is removed from the screen.

grazirs:
Can I work on this?

jcfilben:
Hi @grazirs, great to hear you are interested in working on this! I think there may be some interesting design considerations that come up as you work through this so definitely reach out if needed along the way, but feel free to work on this

jcfilben:
Closing since #6462 has been merged

PR #6462, opened 2022-10-28
Toast notification timing adjustable

grazirs:
#### What does this PR do?
Adds a time property, now the user can set a custom time or use the default controlled by the theme.

`/components/Notification/Notification.js`

A test was added to verify if the custom time is working on the component.
`src/js/components/Notification/__tests__/Notification-test.js`

#6270


src/js/components/Notification/stories/TitleAndMessage.js
```diff
@@ -21,6 +21,7 @@ const TitleAndMessageNotification = () => {
           title="Status Title"
           message="Messages should be at max two lines of text."
           onClose={onClose}
+          time={4000}
```
jcfilben, line 24: Minor comment but I think we can just have `time={4000}` in the Notification/Title story and leave it out of this story so we should both the default behavior and the behavior of a toast when `time` is defined.
```suggestion
```

jcfilben (approved):
Nice work! This looks good to me

ericsoderberghp (approved):

PR #6431, opened 2022-10-17
fix: dropProps should accept inline style

ShimiSun:
#### What does this PR do?
Allows `dropProps` of Tip to accept inline style.

#### Where should the reviewer start?
Tip types
#### What testing has been done on this PR?
local
#### How should this be manually tested?
Use `style` prop inline with `dropProps` of Tip 
#### Do Jest tests follow these best practices?
N/A

#### What are the relevant issues?
#types
#### Do the grommet docs need to be updated?
types
#### Should this PR be mentioned in the release notes?
As a TS issue
#### Is this change backward compatible or is it a breaking change?
backward compatible

PR #6463, opened 2022-10-28
Fix types of drop props in components

ShimiSun:
#### What does this PR do?
It's a follow-up to #6431 
The goal is for `dropProps` to accept `style` as a property. This behavior already exists in DateInput, Tip, but also needs alignment for other components as this PR suggests.

Any ts file
local
Use `style` prop with `dropProps` before/after the fix.
N/A

It's a follow-up to #6431 

Any bucket of TS bugs.
nope
Yes, as a TS fix.
Yes. It extends the type of `dropProps`


jcfilben (approved):
Looks good!

ericsoderberghp (approved):

PR #6465, opened 2022-10-28
Fix axe issues for Form DynamicFields

ShimiSun:
#### What does this PR do?
Fix axe issues for Form DynamicFields
storybook file
local
storybook
N/A
no
#6124 
image (3)
image (2)
no
no
b-c


jcfilben (approved):
Looks good!

ShimiSun:
> Looks good!

Are pending action items needed for this one to get it merged?

src/js/components/Form/stories/DynamicFields.js
```diff
@@ -16,7 +16,7 @@ export const DynamicFields = () => {
           onSubmit={({ value }) => console.log('Submit', value)}
         >
           <FormField label="Name" name="name" required>
-            <TextInput name="name" />
+            <TextInput name="name" aria-label="name" />
```
ericsoderberghp, line 19: Why only this TextInput and not the CheckBox and other TextInput below?

ericsoderberghp, line 19: Also, if this is the standard direction to head, should TextInput automatically set `aria-label` to `name` if `aria-label` isn't set otherwise?

ShimiSun: https://github.com/grommet/grommet/pull/6466#discussion_r1015940492

>I think there is a trade-off for that.
Setting it by default will be good for the average user that forgets to pay attention to aria and accessibility attributes. BUT, for the more savvy accessibility users that would like to use `aria-describedby` or `aria-description` instead of `aria-label`, setting a default of aria-label will clutter and overuse the screen reader content which IMHO is unfair.

ShimiSun: That was the only one that storybook was flagging. I haven't dived into their flagging mechanism or to on other components on this page.

ericsoderberghp (approved):

PR #6466, opened 2022-10-28
Fix axe issues for Form ValidateOnBlur

ShimiSun:
#### What does this PR do?
Fix axe issues for Form ValidateOnBlur
storybook file
local
storybook
N/A
no
#6124 
image (5)
image (4)

no
no
b-c


jcfilben (approved):
Looks good!

src/js/components/Form/stories/ValidateOnBlur.js
```diff
@@ -39,7 +40,7 @@ export const ValidateOnBlur = () => (
         />
 
         <FormField label="Email" name="email" required>
-          <TextInput name="email" type="email" />
+          <TextInput name="email" aria-label="email" type="email" />
```
ericsoderberghp, line 43: How come sometimes we put `aria-label` on the FormField and other times we put it on the TextInput?

ShimiSun: From WCAG docs, Th best practice is to use it on the actual input control https://www.w3.org/WAI/tutorials/forms/labels/

ericsoderberghp: Can we automatically set `aria-label` to `name` if not provided?

ericsoderberghp: Agreed on not wanting to clutter the DOM. Can we check for `aria-describedby` though? In other words, if it appears that the caller is managing aria attributes then we don't add any by default, otherwise we do.

ShimiSun: We can. I suspect it might not be sustainable though, what happens if it is not just `aria-describedby` and its `aria-description` or `aria-labeledBy`? would we still use `aria-label`? There might be other props to consider as well. 
I guess my question is, is it worth adding all of this logic and make the assumption that `aria-label` is the right prop? I need time to reflect on the answer.

ShimiSun: Regardless, I don't think it should block the change in this story.

ericsoderberghp (approved):

src/js/components/Form/stories/ValidateOnBlur.js
ericsoderberghp: OK. We can work on that aspect later.

jcfilben: Issue created to explore this https://github.com/grommet/grommet/issues/6483

PR #6472, opened 2022-10-31
Fixed type of zIndex theme property of Drop

ShimiSun:
#### What does this PR do?
Fixes the zIndex type for drop on the theme object
base.d.ts
local
TS storybook
N/A

Not sure
No
Yes, TS fix
backwards compatible


jcfilben (commented):
Should we add `number` to the other places in `base.d.ts` that use zIndex. I think if we change it in one place it would make sense to change it in the other places as well. I think there are only three other places `zIndex` is defined in `base.d.ts`

ShimiSun:
> Should we add `number` to the other places in `base.d.ts` that use zIndex. I think if we change it in one place it would make sense to change it in the other places as well. I think there are only three other places `zIndex` is defined in `base.d.ts`

I didn't check other places of this prop, their use cases, or how critical their change would be.  I fixed this one because it was breaking locally and it is supported for no TS users.

britt6612:
> > Should we add `number` to the other places in `base.d.ts` that use zIndex. I think if we change it in one place it would make sense to change it in the other places as well. I think there are only three other places `zIndex` is defined in `base.d.ts`
> 
> I didn't check other places of this prop, their use cases, or how critical their change would be. I fixed this one because it was breaking locally and it is supported for no TS users.

That's fair we can go back and assess the other area but do not want to block you on this one.

britt6612 (approved):

Issue #6476, opened 2022-11-03
Select - null value causes crash


jcfilben:
### Expected Behavior
Select should be able to handle the value being set to null

### Actual Behavior
App crashes

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/s/grommet-v2-template-forked-e1nlzs?file=/index.js:495-499

### Steps to Reproduce

See codesandbox above and try clicking on the Select

### Your Environment

- Grommet version: 2.27.0
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #6479, opened 2022-11-03
Changed Select to fix an issue with null value.

ericsoderberghp:
#### What does this PR do?

Changed Select to fix an issue with null value.

The problem comes from `null` being a javascript object. So, `typeof value === 'object'` is `true` when value is `null`

one line change

changed OrderOptions story to start with a `null` value instead of `''`

^^^

no test changes

no

#6476 

no

yes

backwards compatible

### DIFF
--- a/src/js/components/Select/SelectContainer.js
+++ b/src/js/components/Select/SelectContainer.js
@@ -177,7 +177,7 @@ const SelectContainer = forwardRef(
                 return valueValue === optionVal;
               });
             }
-          } else if (valueKey && typeof value === 'object') {
+          } else if (valueKey && value !== null && typeof value === 'object') {
             const valueValue =
               typeof valueKey === 'function'
                 ? valueKey(value)


src/js/components/Select/SelectContainer.js
```diff
@@ -177,7 +177,7 @@ const SelectContainer = forwardRef(
                 return valueValue === optionVal;
               });
             }
-          } else if (valueKey && typeof value === 'object') {
+          } else if (valueKey && value !== null && typeof value === 'object') {
```
MikeKingdom, line 180: Looks good. Kinda surprises me that `typeof undefined is 'undefined'` (literally a string `'undefined'`) but `typeof null` is `'object'`.

MikeKingdom (approved):

src/js/components/Select/SelectContainer.js
ericsoderberghp: Yep. Javascript has improve a lot in the last ten years. But, there are still quirks like this lying around.

PR #6481, opened 2022-11-04
Fix window not defined issue

jcfilben:
#### What does this PR do?
Follow on work to https://github.com/grommet/grommet/pull/6480. It looks like we need to check that the typeof window is not undefined otherwise we still get the window undefined error.

Tested change on DS site and made sure site built okay

No
No
Backwards compatible

### DIFF
--- a/src/js/components/FileInput/utils/formatBytes.js
+++ b/src/js/components/FileInput/utils/formatBytes.js
@@ -2,9 +2,11 @@ const SI_CONVERSION_FACTOR = 1000;
 const IEC_CONVERSION_FACTOR = 1024;
 
 const getCurrentOS = () => {
-  const currentOS = ['Win', 'Linux', 'Mac'].find(
-    (v) => window?.navigator?.userAgent?.indexOf(v) >= 0,
-  );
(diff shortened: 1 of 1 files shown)


src/js/components/FileInput/utils/formatBytes.js
```diff
@@ -2,9 +2,11 @@ const SI_CONVERSION_FACTOR = 1000;
 const IEC_CONVERSION_FACTOR = 1024;
 
 const getCurrentOS = () => {
-  const currentOS = ['Win', 'Linux', 'Mac'].find(
-    (v) => window?.navigator?.userAgent?.indexOf(v) >= 0,
-  );
+  const currentOS = ['Win', 'Linux', 'Mac'].find((v) => {
+    if (typeof window !== 'undefined')
+      return window?.navigator?.userAgent?.indexOf(v) >= 0;
```
britt6612, line 7: since we are wrapping it think we can take out the `?` checks

britt6612 (approved):

taysea (approved):
Strange that the chaining didn't work, but this looks good.

ericsoderberghp (approved):

PR #6491, opened 2022-11-14
fix: makes SelectMultiple accept undefined or null

### LINKED ISSUE #6351: SelectMultiple - Loading enhancement
Enhance SelectMultiple to allow for loading text or spinner to display if data is taking some time to fetch

Screen Shot 2022-09-23 at 2 22 19 PM

BrunoViveiros:
#### What does this PR do?
Makes SelectMultiple accept undefined or null

SelectMultiple component

not sure how to test when receiving undefined value

when passing a undefined value for `options` it should not generate a error

I'm not sure if I should create a new story showing a loading when the options are taking some time to fetch

closes #6351

### DIFF
--- a/src/js/components/SelectMultiple/SelectMultiple.js
+++ b/src/js/components/SelectMultiple/SelectMultiple.js
@@ -81,7 +81,7 @@ const SelectMultiple = forwardRef(
       onOpen,
       onSearch,
       open: openProp,
-      options: optionsProp,
+      options: rawOptionsProp,
       placeholder,
       plain,
       replace,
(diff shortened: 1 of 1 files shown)


jcfilben (commented):
Nice work on this so far! I think it would be good to include a new storybook example that demonstrates one of the loading cases in the original issue. 

When `options` is set to null or undefined the SelectMultiple shows 'No matches found' text at the bottom of the drop. I think we need to adjust the 'No matches found' to only show when we there is actually text in the search input.

jcfilben:
Hi @BrunoViveiros just wanted to check in on this pull request. I think it is pretty close to the finish line and there are just a few things to tie off. Is this something that you are still interested in working on?

PR #6500, opened 2022-11-16
fix:prevent-extra-div-container-in-drop-component

### LINKED ISSUE #6371: Tip - hovering over child creates empty divs that aren't cleaned up
Hovering over the child component wrapped by Tip will create empty `<div></div>
`
### Expected Behavior

If a div is created for whatever reason, it's also cleaned up afterwards

### Actual Behavior

Hovering over the child component wrapped in Tip will endlessly create more divs

https://codesandbox.io/s/bitter-sunset-edyo97?file=/src/App.tsx

### Steps to Reproduce (In the Codesandbox example)

1. Wrap an element with Tip (example uses `<button/>`)
2. Open devtools and inspect elements
3. Navigate to the `<iframe/>`  element of the codesandbox preview window
4. Navigate to the `<body/>` element within that
5. Now hover over the component wrapped in Tip
6. Notice the empty div elements are being appended to the end of the `<body/>`

### Your Environment

- Grommet version:2.26.0 in Codesandbox; Also experienced in 2.25.1
- Operating System and version (desktop or mobile): macOS 11.6.5

Locen9:
#### What does this PR do?
It prevent Drop.js Component from creating multiple divs when running App in StrictMode

From Drop.js in src/js/components/Drop/Drop.js

I tested it on a local environment adding Grommet as a library and creating a very simple react App (in StrictMode) with one Tip component that uses the Drop component inside. (I basically reproduced the codebox example of the issue #6371 ). Eventually I ran storybook to see possible regressions and found none.

I was not able to reproduce it without wrapping the App in StrictMode

I encountered this issue on  a personal project by using the Tip Component, I found out that only in a dev-build I was seeing this issue and prod build were bug free.

it closes #6371

it is backwards compatible

### DIFF
--- a/src/js/components/Drop/Drop.js
+++ b/src/js/components/Drop/Drop.js
@@ -1,4 +1,10 @@
-import React, { forwardRef, useEffect, useState, useContext } from 'react';
+import React, {
+  forwardRef,
+  useEffect,
+  useState,
+  useContext,
+  useRef,
+} from 'react';
 import { createPortal } from 'react-dom';
 
 import { ThemeContext } from 'styled-components';
@@ -24,13 +30,15 @@ const Drop = forwardRef(
     useEffect(() => setOriginalFocusedElement(document.activeElement), []);
     const [dropContainer, setDropContainer] = useState();
     const containerTarget = useContext(ContainerTargetContext);
-    useEffect(
-      () =>
+    const containerChildNodesLenght = useRef(null);
+    useEffect(() => {
+      if (!containerChildNodesLenght?.current) {
+        containerChildNodesLenght.current = containerTarget.childNodes.length;
         setDropContainer(
           !inline ? getNewContainer(containerTarget) : undefined,
-        ),
-      [containerTarget, inline],
-    );
+        );
+      }
+    }, [containerTarget, inline]);
 
     // just a few things to clean up when the Drop is unmounted
     useEffect(
--- a/src/js/components/Drop/Drop.js
+++ b/src/js/components/Drop/Drop.js
@@ -30,10 +30,13 @@ const Drop = forwardRef(
     useEffect(() => setOriginalFocusedElement(document.activeElement), []);
     const [dropContainer, setDropContainer] = useState();
     const containerTarget = useContext(ContainerTargetContext);
-    const containerChildNodesLenght = useRef(null);
+    const containerChildNodesLength = useRef(null);
     useEffect(() => {
-      if (!containerChildNodesLenght?.current) {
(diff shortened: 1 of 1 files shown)


src/js/components/Drop/Drop.js
```diff
@@ -24,13 +30,15 @@ const Drop = forwardRef(
     useEffect(() => setOriginalFocusedElement(document.activeElement), []);
     const [dropContainer, setDropContainer] = useState();
     const containerTarget = useContext(ContainerTargetContext);
-    useEffect(
-      () =>
+    const containerChildNodesLenght = useRef(null);
```
jcfilben, line 33: Why does this need to be a ref? I think it could just be a regular variable. Also can we rename to `containerChildNodesLength` and add a comment about why we need this check for the Tip component

jcfilben (commented):
Nice work so far, just a few comments...

jcfilben:
I realized that strict mode intentionally renders things twice so I wonder if this issue is to be expected or if it is still something that should be fixed

src/js/components/Drop/Drop.js
Locen9: Hi jcfilben, thanks for your feedback!

I added a comment and changed the misspelled variable name.

Unfortunately using a variable will not resolve this: 

- if we define the variable inside the useEffect and then assign it inside the if condition, the issue will not be solved when the component is destroyed and then re-created, as everytime the useEffect runs,  the variable will be initialised again and the if condition will always be true.
- if we define the variable outside the useEffect and assign a value inside, it will lose its assignment after each render which will fix this issue as long as the drop component is without any side-effects.(A tooltip generating plain text will work fine). 
In a bit more complex case in which the parent element of the Drop component has sideEffects that updates the dropped element (i.e. a Calendar), the variable will be re-initialised and we will incur in the bug again.

I would prefer useRef as it is guaranteed not to change between render(s)

Locen9:
> I realized that strict mode intentionally renders things twice so I wonder if this issue is to be expected or if it is still something that should be fixed

Regarding this, as in the future React may apply [new features](https://reactjs.org/docs/strict-mode.html#ensuring-reusable-state), in order to maintain a reusable state and to prevent future headaches, I would go ahed and resolve it.

src/js/components/Drop/Drop.js
jcfilben: Thanks for the explanation, that makes sense

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

PR #6508, opened 2022-11-22
first draft for feedback component

britt6612:
#### What does this PR do?
Adds Feedback Component:

There is essentially 3 components in this PR 

- Feedback
- ThumbsRating
- StarRating

The following props will align with Feedback component
This component includes the container for feedback for inline and modal use. 
When someone sets `feedbackButton=true` they will get the floating `button` which they can place anywhere on the page through theme. They can also change how the button looks through `ButtonProps` in the theme. For example they can change the `roundness` `color` etc...  Clicking on this `button` will open up the Feedback component

Someone can use the modal portion of the feedback without the floating button by setting `modal=true` This would then give he user control as to how they open and close the Layer by the `show` prop 
```
  children?: React.ReactNode;
  feedbackButton?: boolean;
  layerProps?: LayerProps;
  onSubmit?: (event: FormExtendedEvent<T>) => void;
  messages?: object;
  modal?: boolean;
  show?: boolean;
  title?: string;

```

ThumbRating
This component is essentially an Input kind that has the options of dislike and like. It is made up of a RadioButtonGroup and the user can give 2 options which the first would be connected to `like` and the second would be associated with `dislike`. The user can then wrap this in a Grommet `FormField` as they wish.

```
  formFieldProps?: FormFieldProps;
  color?: string;
  name?: string;
  label?: string;
  value?: number;
  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
  outlineColor?: string;
  options: (
    | string
    | number
    | boolean
    | {
        disabled?: boolean;
        id?: string;
      }
  )[];
```

StarRating:
This component is essentially an Input that has the StarRatings. It is made up of a RadioButtonGroup.

```
  fillColor?: string;
  outlineColor?: string;
  name?: string;
  label?: string;
  value?: number;
  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
```

```
  <Feedback
    modal
    title="We’d love your feedback"
    //   onSubmit={onSubmit}
  >
    <FeedbackRating label="Was this content helpful?" />
    <FormField
      htmlFor="comments"
      name="comments"
      label="Any additional comments?"
    >
      <TextArea id="comments" name="comments" placeholder="Comments" />
    </FormField>
  </Feedback>

```

TODO:

- [x] Fix PropTypes & index.d.ts file for Feedback 
- [x] Review with team on the approach 
- [ ] Add a kind to feedback so then `formfield` can know the formContext so if `kind = feedback` theme label for formfield can be used.
- [ ] Condense FeedbackNegPos and FeedbackRating to one file since they are the same props 

Feedback
TODO
Storybook
TODO
- [ ] write tests

closes https://github.com/grommet/hpe-design-system/issues/3037
yes
yes
backwards compatible


src/js/components/Feedback/Feedback.js
```diff
@@ -0,0 +1,140 @@
...
+  const onOpen = () => setOpen(true);
+  const onClose = () => setOpen(undefined);
+
+  // Want to show Thank you message so close the modal after 2 seconds
+  // should we change the timing to be in theme to have user control?
+  const closeFeedbackModal = () => {
+    setTimeout(() => {
+      setOpen(false);
+    }, 2000);
+  };
```
jcfilben, line 52: Not sure if this rule would apply to the time limit set here https://www.w3.org/WAI/WCAG21/Understanding/timing-adjustable.html

src/js/components/Feedback/stories/Feedback.js
```diff
@@ -0,0 +1,46 @@
+import React, { useContext } from 'react';
```
jcfilben, line 1: The Layer background can be 'pushed' vertically

Screen Shot 2022-11-23 at 2 26 23 PM

src/js/components/Feedback/stories/FeedbackInline.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
```
jcfilben, line 1: For the inline version of the Feedback what should happen when you click 'Cancel'? Is the Feedback component no longer displayed?

src/js/components/Feedback/stories/FeedbackNegPos.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
```
jcfilben, line 1: In the grommet theme the spacing feels weird on this. I would expect the Icons to be aligned with the "Was this content helpful?" text and have a little spacing between the FormField underline.

Screen Shot 2022-11-23 at 2 29 36 PM

src/js/components/StarRating/StarRating.js
```diff
@@ -0,0 +1,43 @@
...
+}) => {
+  const [rating, setRating] = useState(value);
+
+  const options = [];
+  for (let i = 0; i < scale; i += 1) {
+    options.push(i);
+  }
+
+  return (
+    <RadioButtonGroup
```
jcfilben, line 23: I noticed some odd behavior when using voice over and trying to navigate with my keyboard

- If 5/5 stars were selected I could no longer tab to the RadioButtonGroup component.
- When 4/5 stars were selected voice over announced that 5/5 were selected.
- Using my left and right arrow keys I could select up to 4/5, but was unable to select the last star.

src/js/components/StarRating/StarRating.js
```diff
@@ -0,0 +1,43 @@
+import React, { useState } from 'react';
+import { Star, StarOutline } from 'grommet-icons';
+import { RadioButtonGroup } from '../RadioButtonGroup';
+
+export const StarRating = ({
+  fillColor,
+  name,
+  label,
+  scale = 5,
```
jcfilben, line 9: It might be nice to have a story that passes a different value for scale

jcfilben (commented):
Nice work, just left a few comments after a first pass

src/js/components/Feedback/Feedback.js
britt6612: oh yeah I was going to see if we should change this or can always have it in the theme one of the questions I have for the research that was done

src/js/components/Feedback/stories/Feedback.js
britt6612: might need a `flex` here Ill look

src/js/components/Feedback/stories/FeedbackInline.js
britt6612: oh right good point maybe only show cancel if `modal`

britt6612: I still need to clarify the inline flow with designer but will bring up the `cancel`

ericsoderberghp:
I'm going to stir things up a bit ...

Given the direction of having StarRating and ThumbRating being their own components, and allowing any generic FormField + input component within it as well, the Feedback component feels like it could be more generic than for just feedback. It seems like a floating Button that opens a Form in a Layer. I'm wondering if we should revise it along those lines, perhaps calling it something like "ButtonForm", to make it clear that its children are in a Form?

Thoughts?

halocline:
> I'm going to stir things up a bit ...
> 
> Given the direction of having StarRating and ThumbRating being their own components, and allowing any generic FormField + input component within it as well, the Feedback component feels like it could be more generic than for just feedback. It seems like a floating Button that opens a Form in a Layer. I'm wondering if we should revise it along those lines, perhaps calling it something like "ButtonForm", to make it clear that its children are in a Form?
> 
> Thoughts?

I like the thought of decoupling the button mechanism to display the feedback.

To me, the scope of the Feedback should be limited to being a layout which serves as a means for presenting a specific type/style of form + form fields.

britt6612:
going to close going with second approach

Issue #2595, opened 2018-12-16
Anchor should have a pad property


shai32:
### Expected Behavior

Anchor should have a pad property
### Actual Behavior

Anchor has only support for margin

shai32:
I want the entire Anchor will be clickable, if the padding will be with a Box. all the padding space is not part of the Anchor and it will not be part of the link

shai32:
@ShimiSun nice solution, I can use it.
but still, I think if Anchor support margin, it should support padding as well

ShimiSun:
@petercpwong thank you for the feedback, we'll discuss and consider.

ShimiSun:
We haven't seen community enthusiasm to support this feature, and we are keeping in mind that Anchor is not a component that is intended to drive Layout so `pad` shouldn't be necessary. 
I'll be closing this issue unless an interest will rise.

issogno:
Saying that there is no "community enthusiasm" on a fairly unkown framework, does not feel right.

I started experimenting with grommet for a new website I am working on, and just one week after I am already falling into this issue and encountering a "lack of enthusiasm" on listening to your users. 

This is not a good sign for a new framework with a small user-base.

jcfilben:
Hi @issogno just wanted to give some context on why this enhancement was passed over. We try to evaluate enhancements based on usefulness and interest from the community. We try to balance adding new features with not bloating the repository with too many props and making it more difficult to maintain and use. (Side note: this is an interesting [podcast episode](https://kentcdodds.com/chats/02/09/erik-rasmussen-chats-with-kent-about-maintaining-open-source-libraries) with Kent C. Dodds and Erik Rasmussen related to this topic)

For this specific feature it felt like adding an additional prop wasn't necessary since Anchor shouldn't be driving the layout and the desired behavior can be achieved in another way. We also didn't see +1's or requests from the community for this besides the initial author. However, that being said even if we close an issue and then interest arises in it later, we are always willing to revisit and discuss. The door is alway open :)

mrmagcore:
I'm with this very old thread.  Why does anchor have margin if it doesn't have pad?  Even wrapped in a box with pad="none", it still sets a large pad around the anchor.  Either don't add a pad around the anchor or allow us to set pad.

mrmagcore:
Sorry, looks like anchor only has a pad around icons, not text.  Still weird that it has margin at all if it doesn't have pad.

mrmagcore:
[...]

PR #6514, opened 2022-11-26
Added `pad` prop in the anchor component.

umeshiscreative:
#### What does this PR do?
Fix the bug when using the Anchor component containing padding.

Anchor.js

I am using the storybook by adding a custom props `pad` in the stories.

Yes, I removed the padding CSS and made this CSS dependent on the props passing from the stories/usage side.

#2595

No

No

backwards compatible.

### DIFF
--- a/src/js/components/Anchor/Anchor.js
+++ b/src/js/components/Anchor/Anchor.js
@@ -34,6 +34,7 @@ const Anchor = forwardRef(
       onClick: onClickProp,
       onFocus,
       reverse,
+      pad,
       ...rest
     },
     ref,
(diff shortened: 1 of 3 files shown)


jcfilben:
Thanks for working on this. I think there was some confusion about the extra space I mentioned in the original issue. I wasn't refering to the padding around the icon only anchor but rather the extra space shown in the blue area of the screenshot that is below the icon. My expectation would be that the blue area shown in the screenshot should end at the edge of the bottom of the icon as it does at the top, left, and right. I think for this issue it would be good to understand where the extra space (not the padding) below the anchor is coming from. Based on [previous discussions](https://github.com/grommet/grommet/issues/2595) we have been leaning away from adding the pad prop to Anchor.

umeshiscreative:
@jcfilben, Sorry for the late update, I was a little busy with release-related things.
I fixed this issue and attached it with the relevant issue.

Thanks,

Issue #6301, opened 2022-08-25
Button - dark mode label color inconsistent


taysea:
### Expected Behavior

If the theme doesn't specify a label color for a specific button state (e.g., active), then the active button should have the same label color as the default state.

### Actual Behavior

In dark mode, when a button's default state has a background with opacity < 1 (such as background-contrast) the label color is coming through differently than when that button is active.

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/active-button-color-pmujf7?file=/index.js

### Steps to Reproduce

Notice the difference in color in the dark mode of the "test button" vs "active test button"

### Your Environment

- Grommet version: 2.25.1
- Browser Name and version: MacOS Chrome
- Operating System and version (desktop or mobile): MacOS Mojave

paola-machado:
Hi! I would like to work on this, if that's ok

jcfilben:
Hey @paola-machado feel free to work on this!

PR #6529, opened 2022-12-06
add form context kind for survey

britt6612:
#### What does this PR do?
This PR is a second approach at a Feedback component 

This PR includes:
StarRating: This is similar to the previous StarRating in which it is made from `RadioButtonGroup` and can accept all props 
ThumbsRating: This is similar to the previous ThumbsRating in which it is made from `RadioButtonGroup` and can accept all props 

**NEW**
`kind` prop was introduced to Form in which `formfield` uses in order to style the label correctly. If a user passes `kind='survey'` to `form` then `formfield` label will use `theme.formfield.survey.label`

The 2 main stories are `unsolicited` and `solicited` in Form folder on storybook

local
local
TODO:

From a discussion with the Grommet maintainers we decided to go with the following in order to simplify the process and make things for flexible in Grommet:

- [x] add kind to Form
- [x] if FormField is in a Form with certain kind use that kind styling for the FormField
- [x] Add theme.formField.feedback to define styling for that kind in the theme (similar to button kinds)

### Add `dropButton` with floating prop
**Pros:** 
- A user would not have to style a grommet Button inorder for it to be "floating"
- Out of the `Box` with `open` and `close` behavior 
- Less code on the users side as well as less state being updated on their end

**Cons:**
- DropButton is pretty specific in its props with being used with Drop.
- per backward compatibility we would need to keep `dropAlign` `dropContent` so would it make sense to also add `layerContent` or `layerPosition`
- The component it self `DropButton` name is specific to Drop

### Add `layerButton` with floating prop
**Pros:** 
- A user would not have to style a grommet Button inorder for it to be "floating"
- Out of the `Box` with `open` and `close` behavior 
- Less code on the user's side as well as less state being updated on their end

**Cons:**
- Essentially the code would be very similar to `DropButton` so not sure it's worth having another component that does the same behavior. 
- Behavior of the button if the Layer is not interactive is that to opinionated from the Grommet side in how we handle that would be disable the `button`?

### Add `floating` prop to Button
**Pros:** 
- A user would not have to style a grommet Button inorder for it to be "floating"

**Cons:**

- the risk of it having it obscure content (which we have been wanting to avoid).
- maybe there isn't a lot of cases to have a floating button. (not sure if this is a con though)

 [Here](https://deploy-preview-6529--sad-tereshkova-173d07.netlify.app/?path=/story/input-form-unsolicited-feedback--un-solicited-feedback) is the story with a floating `Button` and `Layer` just to show to help us decide on what route to go.
- [ ] Explore LayerButton with floating prop or DropButton with floating prop 

closes https://github.com/grommet/hpe-design-system/issues/3037
no
yes
yes
 backwards compatible


src/js/components/Form/Form.js
```diff
@@ -655,9 +657,11 @@ const Form = forwardRef(
           }
         }}
       >
-        <FormContext.Provider value={formContextValue}>
-          {children}
-        </FormContext.Provider>
+        <FormTypeContext.Provider value={kind}>
+          <FormContext.Provider value={formContextValue}>
```
taysea, line 661: Should we just add `kind` to the FormContext? Instead of making its own

src/js/components/FormField/FormField.js
```diff
@@ -334,7 +336,10 @@ const FormField = forwardRef(
       borderColor = (themeBorder && themeBorder.color) || 'border';
     }
 
-    const labelStyle = { ...formFieldTheme.label };
+    let labelStyle;
+    if (formKind === 'survey') {
```
taysea, line 340: Do we want kind === 'survey' to be hardcoded or do we want something a bit more loose like how button kind works?

src/js/components/StarRating/StarRating.js
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  return (
+    <RadioButtonGroup
+      direction="row"
+      options={options}
+      value={rating}
+      name="starRating"
+      onChange={(event) => {
+        const adjustedRating = parseInt(event.target.value, 10) + 1;
```
taysea, line 29: what happens if someone passes their own `options`  here? will this calculation cause any issues?

taysea: or are we discouraging people from passing their own options? which I'm fine with too.

src/js/components/StarRating/StarRating.js
```diff
@@ -0,0 +1,43 @@
...
+import { RadioButtonGroup } from '../RadioButtonGroup';
+
+export const StarRating = ({
+  fillColor,
+  name,
+  label,
+  scale = 5,
+  value,
+  onChange,
+  outlineColor,
```
taysea, line 12: instead of having `fillColor` and `outlineColor`, it'd be nice to just have a single `color` prop. Where color could be:
- a string (applies to both fill and outline)
- object: { fill: string, outline: string }

src/js/components/StarRating/StarRating.js
```diff
@@ -0,0 +1,43 @@
...
+  for (let i = 0; i < scale; i += 1) {
+    options.push(i);
+  }
+
+  return (
+    <RadioButtonGroup
+      direction="row"
+      options={options}
+      value={rating}
+      name="starRating"
```
taysea, line 27: I don't think we have default `name` on other inputs, so thinking we should remove. But you might want to take a peak at how other inputs are using formContext:

TextInput.js for example
```
    const [value, setValue] = formContext.useFormInput({
      name: readOnly ? undefined : name,
      value: valueProp,
    });
```

Select
```
    const [value, setValue] = formContext.useFormInput({
      name,
      value: valueProp,
      initialValue: defaultValue || '',
    });
```

src/js/components/ThumbsRating/ThumbsRating.js
```diff
@@ -0,0 +1,39 @@
...
+export const ThumbsRating = ({
+  fillColor,
+  outlineColor,
+  formProps,
+  name,
+  label,
+  options,
+  ...rest
+}) => {
+  const likeOption = options[0];
```
taysea, line 14: should we have some default options, in the case that the user doesn't provide any?

src/js/components/__tests__/__snapshots__/components-test.js.snap
```diff
@@ -1247,6 +1247,15 @@ exports[`Components loads 1`] = `
           "margin": {
             "bottom": "small",
           },
+          "survey": {
+            "label": {
+              "margin": {
+                "vertical": "xsmall",
+              },
+              "size": "large",
+              "weight": 500,
```
taysea, line 1256: Figma seems to be using default weight (400) -- would be good to double check the desired value.

src/js/components/__tests__/__snapshots__/components-test.js.snap
```diff
@@ -1247,6 +1247,15 @@ exports[`Components loads 1`] = `
           "margin": {
             "bottom": "small",
           },
+          "survey": {
+            "label": {
+              "margin": {
+                "vertical": "xsmall",
```
taysea, line 1253: The spacing between question and input and also between formfields looks off -- would be good to double check the desired value.

taysea (commented):
Nice progress, I'm liking the `kind` approach on Form and then just using FormField inside.

src/js/components/StarRating/StarRating.js
britt6612: yeah Im thinking of changing this to either have this as default or accept `options`

src/js/components/StarRating/StarRating.js
britt6612: good idea

src/js/components/StarRating/StarRating.js
britt6612: aw okay yeah that makes sense but since its StarRating do we want to use that as a default?

since the other inputs are more flexible with how the user wants

src/js/components/__tests__/__snapshots__/components-test.js.snap
britt6612: yeah you are correct and `medium` these values were more just to show it works since I would take them out anyways here and put in hpe theme

src/js/components/Form/Form.js
britt6612: yeah I tried this and ran into issues... I couldn't figure some of it out I can take another look though

src/js/components/FormField/FormField.js
britt6612: aw okay so if `(kind)` then do styles rather then `=== survey`

src/js/components/StarRating/StarRating.js
britt6612: Ill also do the same for thumbs

src/js/components/ThumbsRating/ThumbsRating.js
britt6612: yes I can add 
thought I did but might of been in other PR

src/js/components/StarRating/StarRating.js
jcfilben: This might be what is causing the weird keyboard behavior where it is off by one

taysea:
Between <DropButton floating /> or <LayerButton floating /> -- are there any accessibility benefits to using `Layer` for the feedback vs `Drop`? I noticed in the DS site we use Layer but we do `modal={false}`. I don't know the intricacies of accessibility concerns off the top of my head, but if we can implement it the same with Drop then I might be included to use DropButton?

OR

<Button floating onClick={() => setOpen} /> and then just let the user handle their layer as they usually would.

I'm leaning away from Layer button at the moment, pending the accessibility questions I had above.

src/js/components/FormField/FormField.js
jcfilben: I think it might be nice to have a little more flexibility here like how button handles kinds. This would allow people to have 'marketing' forms or other styling variations we haven't thought about

jcfilben (commented):
I like the idea of adding the `floating` prop to Button. It feels like the other options have more downsides associated with them and may not be routes we want to go. Adding `floating` to Button seems like a reasonable solution that doesn't add to much extra code or unnecessary work

src/js/components/ThumbsRating/ThumbsRating.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { Like, LikeFill, Dislike, DislikeFill } from 'grommet-icons';
+import { RadioButtonGroup } from '../RadioButtonGroup';
+
+export const ThumbsRating = ({
+  fillColor,
+  outlineColor,
+  formProps,
```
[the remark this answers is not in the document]
britt6612: no I can take it out

britt6612:
> Between or -- are there any accessibility benefits to using `Layer` for the feedback vs `Drop`? I noticed in the DS site we use Layer but we do `modal={false}`. I don't know the intricacies of accessibility concerns off the top of my head, but if we can implement it the same with Drop then I might be included to use DropButton?
> 
> OR
> 
> <Button floating onClick={() => setOpen} /> and then just let the user handle their layer as they usually would.
> 
> I'm leaning away from Layer button at the moment, pending the accessibility questions I had above.

So also with `Layer` there is an animation that is nice when it opens and closes. Sort of like a `slide` in which Drop does not have animation Im sure we could add this though as an enhancement. As far as accessibility Drop does do a better job with having the focus around the whole Drop content then you can down arrow to see what is in the content. Where as `Layer` we have to add the `autoFocus` to the X element which is in our DS guidance..

src/js/components/FormField/FormField.js
britt6612: yes changed this to be whatever `kindArg` they pass

src/js/components/StarRating/StarRating.js
britt6612: Thinking we dont let someone pass `options` for now first pass since Im not sure what other `options` they would be passing since this is just a rating and they can change the `scale`

src/js/components/StarRating/StarRating.js
britt6612: done

src/js/components/StarRating/StarRating.js
britt6612: Took out `name` and since Im reusing RadioButtonGroup this works fine with putting `name` on formfield with default values

src/js/components/ThumbsRating/ThumbsRating.js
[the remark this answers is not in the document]
britt6612: done

src/js/components/FormField/FormField.js
```diff
@@ -334,7 +335,10 @@ const FormField = forwardRef(
       borderColor = (themeBorder && themeBorder.color) || 'border';
     }
 
-    const labelStyle = { ...formFieldTheme.label };
+    let labelStyle;
+    if (formKind) {
+      labelStyle = { ...formFieldTheme[formKind].label };
```
jcfilben, line 340: Are there other style props that formKind could control besides just `label`?

src/js/components/ThumbsRating/index.d.ts
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+import { RadioButtonGroupProps } from '../RadioButtonGroup/index';
+export interface ThumbsRating {
+  color?: string | object;
+  options: (
```
jcfilben, line 5:
```suggestion
  options?: (
```
Should this be like this?

src/js/components/StarRating/__tests__/StarRating-tests.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
```
jcfilben, line 1: Can we make this a typescript file? Trying to make sure any new test files that are added are typescript

src/js/components/ThumbsRating/__tests__/ThumbsRating.js
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
```
jcfilben, line 1: Can we make this a typescript file? Trying to make sure any new test files that are added are typescript

src/js/contexts/FormTypeContext.js
```diff
@@ -0,0 +1,3 @@
+import React from 'react';
+
+export const FormTypeContext = React.createContext(undefined);
```
jcfilben, line 3: Are we still using this?

src/js/components/Form/stories/UnsolicitedFeedback.js
```diff
@@ -0,0 +1,137 @@
...
+            </Box>
+          </Box>
+        </Layer>
+      )}
+    </>
+    // </Grommet>
+  );
+};
+
+UnSolicitedFeedback.storyName = 'UnSolicited feedback';
```
jcfilben, line 133: I think we can exclude this story from having a chromatic snapshot

src/js/components/FormField/FormField.js
britt6612: good point I could see us expanding on the `kind` for `border` or `roundness` I think we can brainstorm more how flexible we want `kind` but I agree on expanding

src/js/components/Form/stories/UnsolicitedFeedback.js
```diff
@@ -0,0 +1,137 @@
...
+          elevation="large"
+          color="purple!"
+          label="Submit Feedback"
+          a11yTitle="This button launches a modal to give feedback."
+          primary
+          alignSelf="start"
+        />
+      </PositionedFeedbackBox>
+      {open && (
+        <Layer modal={false} onEsc={onClose}>
```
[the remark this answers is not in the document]
britt6612: changed that on this story since that is more hpe design system guidance

src/js/contexts/FormTypeContext.js
britt6612: nope taking out

src/js/components/Form/stories/UnsolicitedFeedback.js
britt6612: done

src/js/components/Form/stories/SolicitedFeedback.js
```diff
@@ -0,0 +1,81 @@
...
+          method="post"
+          validate="submit"
+          kind="survey"
+        >
+          <FormField
+            htmlFor="star-rating"
+            name="rating"
+            label="Was this content helpful?"
+          >
+            <StarRating id="star-rating" name="rating" color="pink" />
```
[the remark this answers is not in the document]
britt6612: need to look into this behavior more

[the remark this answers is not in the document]
britt6612: Ill also attempt a keyboard test so we can have coverage there

src/js/components/Form/stories/UnsolicitedFeedback.js
```diff
@@ -0,0 +1,137 @@
+import React, { useContext, useState } from 'react';
```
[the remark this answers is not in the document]
britt6612: ill change the story to reset the successful

src/js/themes/base.js
```diff
@@ -937,6 +937,13 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       },
       margin: { bottom: 'small' },
       // round: undefined,
+      survey: {
```
taysea, line 940: I'm pausing on the naming and alignment with our Figma direction of calling it "Feedback" -- is this a spot where we are okay naming the Grommet thing differently than the Figma thing? Open to discussion, just want to prompt the topic.

britt6612: honestly would this even be in here? I put it so we can test it out but should this be a default for grommet or just have Feedback in our hpe theme

src/js/components/ThumbsRating/__tests__/ThumbsRating.js
britt6612: done

src/js/components/StarRating/__tests__/StarRating-tests.js
britt6612: done

src/js/components/Form/Form.js
```diff
@@ -569,9 +570,10 @@ const Form = forwardRef(
         };
       };
 
-      return { useFormField, useFormInput };
+      return { useFormField, useFormInput, kindArg };
```
ericsoderberghp, line 573: I think the context should call it "kind". I think "*Arg" should primarily be used within the function component to distinguish from other variables in scope. Any reason we cannot just call it "kind" everywhere?

src/js/components/FormField/FormField.js
ericsoderberghp: I think it's fine to start with just what we need now, as long as we think the theme structure is reasonably future-proof, and add more later as tangible needs arise.

src/js/components/StarRating/StarRating.js
```diff
@@ -0,0 +1,46 @@
+import React, { useState } from 'react';
+import { Star, StarOutline } from 'grommet-icons';
+import { StarRatingPropTypes } from './propTypes';
+import { RadioButtonGroup } from '../RadioButtonGroup';
+
+const StarRating = ({ color, scale = 5, ...rest }) => {
```
ericsoderberghp, line 6: Do we need the `color` prop for the first version? Just checking.

ericsoderberghp, line 6: Do we need the `scale` property for the first version? Just checking.

src/js/components/StarRating/StarRating.js
```diff
@@ -0,0 +1,46 @@
...
+  const [rating, setRating] = useState();
```
ericsoderberghp, line 7: Can we call this "value" instead of "rating"? A soft opinion but it might make it clearer that this is an input.

src/js/components/ThumbsRating/ThumbsRating.js
```diff
@@ -0,0 +1,45 @@
...
+    colorProp = color;
+  } else if (color && typeof color === 'object') {
+    colorfill = color.fill;
+    coloroutline = color.outline;
+  } else colorProp = undefined;
+
+  return (
+    <RadioButtonGroup
+      direction="row"
+      options={options || ['like', 'dislike']}
```
ericsoderberghp, line 22: Given how the child objects are created, I don't think we should allow customizing the `options` via a property.

src/js/components/ThumbsRating/ThumbsRating.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { Like, LikeFill, Dislike, DislikeFill } from 'grommet-icons';
+import { ThumbsRatingPropTypes } from './propTypes';
+import { RadioButtonGroup } from '../RadioButtonGroup';
+
+const ThumbsRating = ({ color, options, ...rest }) => {
```
ericsoderberghp, line 6: Similar question about `color` in the first version as with star rating.

src/js/components/__tests__/__snapshots__/components-test.js.snap
ericsoderberghp: I think we should include "survey" in the base grommet theme. It's a reasonable concept that I think would be good to provide.

src/js/themes/base.js
ericsoderberghp: I'm inclined to use "survey" in the base grommet theme.

src/js/components/Form/__tests__/Form-test-controlled.js
```diff
@@ -51,6 +52,30 @@ describe('Form controlled', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
+  test('custom theme', () => {
+    const customTheme = {
+      formField: {
+        feedback: {
```
ericsoderberghp, line 58: Can we use "survey" here, to align with using it elsewhere?

ericsoderberghp (commented):
I'm also liking the "kind" approach.

src/js/components/StarRating/StarRating.js
britt6612: yes that aligns more with other areas I can change

src/js/components/StarRating/StarRating.js
britt6612: not necessarily

src/js/components/ThumbsRating/ThumbsRating.js
britt6612: we could leave out designs for hpe use `color` are you opposing it as a prop and maybe have it in theme ?

src/js/components/Form/__tests__/Form-test-controlled.js
britt6612: yes

src/js/components/Form/Form.js
britt6612: agree Ill change it back to `kind`

src/js/components/ThumbsRating/ThumbsRating.js
ericsoderberghp: yes

britt6612: sounds good I can adjust this 👍

src/js/components/ThumbsRating/ThumbsRating.js
```diff
@@ -0,0 +1,49 @@
...
+                theme.thumbsRating?.like?.icon?.color ||
+                theme.thumbsRating?.like?.icon?.color?.outline
+              }
+            />
+          );
+        }
+        return checked ? (
+          <DislikeFill
+            color={
+              theme.thumbsRating?.dislike?.icon?.color ||
```
britt6612, line 31: feel free to add ideas for naming in theme.

src/js/themes/base.js
```diff
@@ -1501,6 +1508,18 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       xlarge: { ...fontSizing(2) },
       xxlarge: { ...fontSizing(4) },
     },
+    thumbsRating: {
```
britt6612, line 1511: I can strip these out just left them incase we want to see story and change naming

src/js/themes/base.js
```diff
@@ -1663,6 +1682,11 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         size: 'medium',
       },
     },
+    starRating: {
```
britt6612, line 1685: I can strip these out just left them incase we want to see story and change naming

src/js/components/Form/stories/SolicitedFeedback.js
```diff
@@ -0,0 +1,81 @@
...
+          Let us know how your expirence was!
+        </Heading>
+      </Header>
+      <Box
+        // Padding used to prevent focus from being cutoff
+        pad={{ horizontal: 'xxsmall' }}
+      >
+        <Form
+          onSubmit={({ value }) => console.log('Submit', value)}
+          method="post"
```
ericsoderberghp, line 39: Do we need `method="post"`?

src/js/components/Form/stories/SolicitedFeedback.js
```diff
@@ -0,0 +1,81 @@
...
+        align="start"
+        gap="xxsmall"
+        pad={{ horizontal: 'xxsmall' }}
+      >
+        <Heading level={2} size="small" margin="none">
+          Let us know how your expirence was!
+        </Heading>
+      </Header>
+      <Box
+        // Padding used to prevent focus from being cutoff
```
ericsoderberghp, line 34: This seems a bit odd to me. What is cutting the focus off?

src/js/components/Form/stories/SolicitedFeedback.js
```diff
@@ -0,0 +1,81 @@
...
+          <FormField
+            htmlFor="star-rating"
+            name="rating"
+            label="Was this content helpful?"
+          >
+            <StarRating id="star-rating" name="rating" color="pink" />
+          </FormField>
+          <FormField
+            label="What would have improved your experience"
+            htmlFor="better-experience"
```
ericsoderberghp, line 52:
```suggestion
            htmlFor="experience"
```

src/js/components/Form/stories/UnsolicitedFeedback.js
```diff
@@ -0,0 +1,136 @@
...
+  Heading,
+  ThumbsRating,
+  ResponsiveContext,
+} from 'grommet';
+
+// floating button purposes
+const PositionedFeedbackBox = styled(Box)`
+  bottom: 0px;
+  right: 0px;
+  position: ${(props) => props.theme.feedback?.buttonContainer?.position};
```
ericsoderberghp, line 21: Why does the story need to reach into the theme like this? I would have either expected that to be built in somehow or that those things shouldn't be in the theme at all.

src/js/components/Form/stories/UnsolicitedFeedback.js
```diff
@@ -0,0 +1,136 @@
...
+      setOpen(false);
+      setIsSuccessful(false);
+    }, 3000);
+  };
+
+  return (
+    // Uncomment <Grommet> lines when using outside of storybook
+    // <Grommet theme={...}>
+    <>
+      <PositionedFeedbackBox>
```
ericsoderberghp, line 46: Why do we need the Box? Can we just use a Button directly?

src/js/themes/base.js
ericsoderberghp: I'm OK with the naming being broader than "feedback" on the grommet side. I think it aligns better with what it is for generally.

src/js/themes/base.js
ericsoderberghp: I'm inclined to not have these colors here as part of the base theme.

src/js/themes/base.js
ericsoderberghp: I'm inclined to not have this color here as part of the base theme

ericsoderberghp (commented):
Looking good! Just a few more comments.

src/js/themes/base.js
britt6612: yes I can remove you okay with theme structure?

src/js/components/Form/stories/SolicitedFeedback.js
britt6612: no

src/js/components/Form/stories/SolicitedFeedback.js
britt6612: taken out

src/js/components/Form/stories/UnsolicitedFeedback.js
britt6612: fixed

src/js/components/Form/stories/UnsolicitedFeedback.js
```diff
@@ -0,0 +1,133 @@
...
+            </Header>
+            <Box
+              // Padding used to prevent focus from being cutoff
+              pad={{ horizontal: 'xxsmall' }}
+            >
+              <Form
+                method="post"
+                validate="submit"
+                kind="survey"
+                onSubmit={(value) => {
```
jcfilben, line 82: After clicking Submit I am forced to wait until the Layer closes itself. I feel like you should still be able to close the Layer by hitting 'esc' or something so you are not forced to wait

src/js/components/Form/stories/UnsolicitedFeedback.js
```diff
@@ -0,0 +1,133 @@
...
+
+  return (
+    // Uncomment <Grommet> lines when using outside of storybook
+    // <Grommet theme={...}>
+    <>
+      <PositionedFeedbackButton
+        onClick={onOpen}
+        margin={{ vertical: 'medium', horizontal: 'medium' }}
+        elevation="large"
+        color="brand"
```
jcfilben, line 49: Not sure why but in storybook with the hpe theme this button is showing up with dark text. Then when the Layer is open the button has light text. Seems like it should always have light text so there is enough contrast

Screen Shot 2022-12-15 at 2 43 18 PM

Screen Shot 2022-12-15 at 2 43 59 PM

src/js/components/StarRating/__tests__/StarRating-tests.js
jcfilben: Looks like this got switched back to a `.js` file

src/js/components/ThumbsRating/__tests__/ThumbsRating.jsx
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
```
jcfilben, line 1: Can we switch this file to `.tsx`?

src/js/components/ThumbsRating/ThumbsRating.js
jcfilben: I'm fine with the naming that is already here. Maybe we could shorten it to `thumbsRating.dislike.color` because there is nothing else I would expect to be able to change the color on here besides the icon, so icon is already implied. Soft opinion though, flexible either way.

jcfilben (commented):
Just a few more comments, I think it is really close

src/js/components/Form/stories/UnsolicitedFeedback.js
britt6612: I can add `esc` thats just the pattern I was following from the DS example

src/js/components/Form/stories/UnsolicitedFeedback.js
britt6612: yeah here is the issue for this 
https://github.com/grommet/grommet/issues/6301

src/js/components/Form/stories/UnsolicitedFeedback.js
britt6612: added `onClickOutside={onClose}`

src/js/components/StarRating/StarRating.js
```diff
@@ -0,0 +1,40 @@
...
+import { ThemeContext } from 'styled-components';
+import { Star, StarOutline } from 'grommet-icons';
+import { RadioButtonGroup } from '../RadioButtonGroup';
+
+const StarRating = ({ ...rest }) => {
+  const theme = useContext(ThemeContext);
+  const [value, setValue] = useState();
+
+  const options = [];
+  for (let i = 0; i < 5; i += 1) {
```
taysea, line 11: I see here that we're indexing from 0 (which I know is standard for an array), but I'm wondering if it would make more sense to start at `1`? Because a value of 0 almost feels like the user didn't select anything. Soft opinion.

britt6612: Im good either way Ive seen both ways when I was looking around

britt6612: changed it 👍

package.json
```diff
@@ -205,7 +205,7 @@
   "bundlesize": [
     {
       "path": "./dist/grommet.min.js",
-      "maxSize": "145 kB"
+      "maxSize": "245 kB"
```
jcfilben, line 208: Seems like we shouldn't need to increase by this much.

britt6612: fixed!

jcfilben (approved):
Looks good!

src/js/components/StarRating/StarRating.js
```diff
@@ -0,0 +1,41 @@
...
+      options={options}
+      onChange={(event) => {
+        setValue(event.target.value);
+      }}
+      {...rest}
+    >
+      {(option) =>
+        option <= value ? (
+          <Star
+            color={theme.starRating?.color || theme.starRating?.color?.fill}
```
ericsoderberghp, line 28: This use of `theme.starRating.color.fill` doesn't align with the typescript definition, where `ColorType` is just a string or an object with dark/light keys. Can we simplify this to just `theme.starRating.color`?

ericsoderberghp (commented):
Just a few.

src/js/components/StarRating/StarRating.js
britt6612: aw yes I see your point wondering if I should make `theme.starRating.color.fill` its own type so we can have `theme.starRating.color` where someone can pass dark/light keys and also have the `theme.starRating.color.fill` and `theme.starRating.color.outline` if they want different colors depending on which icon is selected or you saying for now just allow one color for the icons regardless of fill out outline?

ericsoderberghp: I think the theme should just be asking for a color and we should use it appropriately within the StarRating to apply the color where people would typically expect it, as outline for unselected, as fill for selected.

taysea (approved):

ericsoderberghp (approved):

Issue #6524, opened 2022-12-05
FormField - `required` prop should set `aria-required=true`


jcfilben:
### Expected Behavior
A screenreader user should be notified that a field is required.

### Actual Behavior
Without `aria-required` a screenreader doesnt give an indication that the field is required.

### URL, screen shot, or Codepen exhibiting the issue
The issue can be see in the following story, the email field is required but the screenreader announcement doesn't indicate that.
https://storybook.grommet.io/?path=/story/input-form-controlled-lazy--controlled-lazy

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.28.0
- Browser Name and version:
- Operating System and version (desktop or mobile):

britt6612:
If someone passed the `required` prop to the FormField the user will also need to pass `aria-required` to whatever input they are using. 

```
The aria-required attribute indicates that user input is required on the element before a form may be submitted.
```

For now we are going to add the `aria-required` in all of our stories to show this practice. Since we'd have to look at if the `required` prop was passed to FormField in all of our inputs at this time we are not going to automatically add this in to Grommet it will be upon the user to add.

PR #6540, opened 2022-12-14
FormField -- include required text in screen reader only

taysea:
#### What does this PR do?

This updates our approach to marking required fields with a required indicator. Previously, we relied on aria-label to put a "required" string on the * span. However, certain screen readers like JAWS/NVDA do not read aria-labels on spans. This takes an improved approach (mirroring an approach [Bootstrap](https://getbootstrap.com/docs/4.0/utilities/screenreaders/) has taken) to include "required" in the DOM but keeping it only visible for screen readers.

Relates to #6524 

No.

Yes. Improved screen reader support for required form fields.

Backwards compatible.

### DIFF
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -69,6 +69,17 @@ const RequiredText = styled(Text)`
   line-height: inherit;
 `;
 
+const ScreenReaderOnly = styled(Text)`
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  border: 0;
+`;
+
 const Message = ({ error, info, message, type, ...rest }) => {
   const theme = useContext(ThemeContext) || defaultProps.theme;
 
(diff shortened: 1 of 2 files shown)


src/js/components/Form/stories/Controlled.js
```diff
@@ -47,8 +47,9 @@ export const Controlled = () => {
           <FormField label="Name" name="name">
             <TextInput name="name" suggestions={suggestions} />
           </FormField>
-          <FormField label="Email" name="email" required>
+          <FormField label="Email" htmlFor="email" name="email" required>
```
jcfilben, line 50: Do we need to have the `htmlFor` prop for 'required' to get read out correctly? I noticed required is being read out for this story but not for other Form stories

taysea: Yes, but FormField always should have `htmlFor` and `id` on input in order to associate the label with the input, regardless of it's required or not. (you'll see not associated label flagged in the accessibility panel for all form stories). It's something we should fix in our stories but I think should be handled in a separate PR.

jcfilben: Sounds good. Can we open up a separate issue to change the stories to use `htmlFor` and `id`?

jcfilben (approved):
Looks good!

src/js/components/Form/stories/Controlled.js
taysea: https://github.com/grommet/grommet/issues/6542

ericsoderberghp (approved):

Issue #6515, opened 2022-11-28
SelectMultiple: Clear All does not work with reduce: true in valueKey


TimPchelintsev:
### Expected Behavior

"Clear All" button should appear instead of "Select All" in SelectMultiple component after "Select All" was applied with "reduce: true" option set for "valueKey"

### Actual Behavior

Component was stuck in "Select All" mode even with all options already selected

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/s/grommet-v2-template-forked-l8b45k?file=/index.js

### Steps to Reproduce

1. Set "reduce: true" for "valueKey" with object options in SelectMultiple component
2. Click "Select All" button - it works as expected
3. BUT button will not change to "Clear All" - bug

### Your Environment

- Grommet version: 2.28.0
- Browser Name and version: Google Chrome 107.0.5304.110 
- Operating System and version (desktop or mobile): Mac OS Ventura 13.0.1

feliperli:
Hey! I'll try to look into this :)

PR #6557, opened 2023-01-02
fix selected multiple when option is no object

feliperli:
#### What does this PR do?

This PR fixes https://github.com/grommet/grommet/issues/6515 by checking if the option is truly an object before accessing it's property.

Look at `src/js/components/Select/utils.js`

I added `valueKey={{ key: 'value', reduce: true }}` on `select all and clear` test

By creating something similar to https://codesandbox.io/s/grommet-v2-template-forked-l8b45k?file=/index.js

https://github.com/grommet/grommet/issues/6515

No

I don't think so because it's a bug 🤔 

It's compatible


jcfilben (approved):
Looks good!

ericsoderberghp (approved):

PR #6558, opened 2023-01-02
Added DataTableColumns

ericsoderberghp:
#### What does this PR do?

Add DataTableColumns as a component, allowing users to customize which columns to show in a DataTable and what order to show them.
It integrates with Data, DataForm, and DataTable.
It is almost a direct port from the implementation on the hpe-design-system site.

Check the stories to see how it works.
The bulk of the work is in the new DataTableColumns.js

An initial unit test and a couple of stories so far.

stories

Screenshot 2023-01-02 at 2 59 38 PM

yes

yes

backwards compatible


src/js/components/DataTableColumns/stories/Inline.js
```diff
@@ -0,0 +1,33 @@
...
+import { DataTableColumns } from '../DataTableColumns';
+import { columns, DATA } from '../../DataTable/stories/data';
+
+export const Inline = () => (
+  // Uncomment <Grommet> lines when using outside of storybook
+  // <Grommet theme={...}>
+  <Grid pad="large" columns={['large']} justifyContent="center">
+    <Data data={DATA} updateOn="change">
+      <DataFilters>
+        <DataTableColumns
```
jcfilben, line 14: Screen Shot 2023-01-03 at 12 39 25 PM

In the grommet theme it looks like there isn't enough spacing between the search and the CheckBoxGroup or the CheckBoxGroup and the DataTable

src/js/components/DataTableColumns/stories/Inline.js
```diff
@@ -0,0 +1,33 @@
...
+          options={columns.map(({ header, property }) => ({
+            property,
+            label: header,
```
jcfilben, line 17: Screen Shot 2023-01-03 at 12 42 21 PM

In the grommet theme the bolding on 'Location' and 'Date' doesn't match up with the table headers like it does in the HPE theme

Screen Shot 2023-01-03 at 12 43 03 PM

In addition when the label is not bolded it's associated down/up buttons are not read correctly by the screenreader.

Bolded label (Location) up button screenreader output: '1 Location move up button'
Non bolded label (Name with extra) up button screenreader output: '2 [object, object] move up button'

src/js/components/DataTableColumns/DataTableColumns.js
```diff
@@ -0,0 +1,163 @@
...
+      <Tabs {...tabsProps[drop ? 'drop' : 'noDrop']} {...rest}>
+        <Tab
+          id={`${dataId}--select-columns-tab`}
+          title={format({
+            id: 'dataTableColumns.select',
+            messages: messages?.dataTableColumns,
+          })}
+        >
+          <Box pad={{ top: 'small' }}>
+            <TextInput
```
jcfilben, line 88: I am unable to type certain characters into the search. For example on the inline story I can't type 'p' into the search

MikeKingdom:
If you hide the name column then re-show it, it ends up on the right.
image

src/js/components/DataTable/DataTable.js
```diff
@@ -90,19 +90,29 @@ const DataTable = ({
...
       }));
-    if (data.length) return Object.keys(data[0]).map((p) => ({ property: p }));
-    return [];
-  }, [columnsProp, data, properties]);
+    else if (data.length)
+      result = Object.keys(data[0]).map((p) => ({ property: p }));
+    if (view?.columns)
+      result = result
+        .filter((c) => view.columns.includes(c.property))
+        .sort(
```
MikeKingdom, line 109: Is there a way to make columns keep track of their last position and let them sort close to where they last were? It's confusing to hide then show a column and have it kicked to the end.

src/js/components/DataTableColumns/DataTableColumns.js
```diff
@@ -0,0 +1,163 @@
...
+            pad="none"
+            primaryKey={(objectOptions && 'label') || undefined}
+          />
+        </Tab>
+      </Tabs>
+    </Box>
+  );
+};
+
+export const DataTableColumns = ({ drop, options, ...rest }) => {
```
MikeKingdom, line 132: How could you persist the changes in visibility or order? Give access to onClose? Have an onChange?

ericsoderberghp: That functionality is coming in a separate pull request, tied to "named views".

MikeKingdom (approved):
Looks good. The only oddness for me is the Search box for inline mode. The way is presented isn't clearly constrained within the context/tabs for column order/visibility so I really wondered if the search would search the table contents rather than column headings.

ericsoderberghp:
> Looks good. The only oddness for me is the Search box for inline mode. The way is presented isn't clearly constrained within the context/tabs for column order/visibility so I really wondered if the search would search the table contents rather than column headings.

Fair point. I'm inclined to focus on the functionality for this first pass and then work on stylistic improvements later. I don't think the inline version will be used much or at all.

jcfilben (commented):
Minor comment...
If we only have one column displayed should we even show the 'order columns' tab?

Everything else looks good so I'm okay if we want to move forward without this change

ericsoderberghp:
> Minor comment... If we only have one column displayed it seems like we don't need to show the 'order columns' tab
> 
> Everything else looks good so I'm okay if we want to move forward without this change

I can see the concept. My one concern is re-laying out the Drop as the user toggles between one and two columns in the first tab. I think it's OK to leave it there, partly since a one column table is a rarer case.

ericsoderberghp:
> Looks good. The only oddness for me is the Search box for inline mode. The way is presented isn't clearly constrained within the context/tabs for column order/visibility so I really wondered if the search would search the table contents rather than column headings.

Decided to just remove the Inline story and turn off the functionality for that. We can bring it back when the need arises.

jcfilben (approved):
Looks good!

PR #6562, opened 2023-01-04
Update DOM functions to fix shadow root bug

### LINKED ISSUE #6232: document.activeElement is not working when using Grommet in shadow Dom. This means a form field is not getting highlighted when clicked.
When using  formfield and a form element component in shadow Dom and clicking it, the highlighted border does not appear.

### Expected Behavior

The behavior should be the same for web component usages as it is in the normal scenario

### Actual Behavior

The border is missing when clicked. Styles are never added to stylesheetmanager outlet when the element is clicked. When not using shadow Dom, I can see the styles being added correctly.

https://codesandbox.io/s/grommet-v2-template-forked-cycmxn?file=/index.js

### Steps to Reproduce

Add formfield and form element to shadowDom with a stylesheemanager and click on text input. Notice the highlighted border is missing only when using shadowDom.

### Your Environment

- Grommet version: 2.24.0
- Operating System and version (desktop or mobile): OSX

taysea:
#### What does this PR do?

Enhances certain DOM functions that FormField relies on to ensure that activeElement within a shadow DOM is detected.

Reference: https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode, https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/activeElement

Tested in local storybook where I set up a FormField in a shadow dom (mirroring the scenario from #6232)

The Box inside the shadow DOM now receives the proper highlighting.

https://user-images.githubusercontent.com/12522275/210661218-34ed2c19-c0b4-4c38-833e-bcc525c223d9.mov

document.activeElement will not look into a shadow root, so getRootNode() needs to be used to determine if root is document or shadow root.

Closes #6232 

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -494,7 +494,10 @@ const FormField = forwardRef(
         {...outerProps}
         style={outerStyle}
         onFocus={(event) => {
-          setFocus(containsFocus(formFieldRef.current) && shouldKeepFocus());
+          const root = formFieldRef.current.getRootNode();
+          setFocus(
+            containsFocus(formFieldRef.current) && shouldKeepFocus(root),
+          );
           if (onFocus) onFocus(event);
         }}
         onBlur={(event) => {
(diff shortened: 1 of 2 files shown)


ericsoderberghp (approved):

jcfilben (approved):
Looks good!

PR #6563, opened 2023-01-05
Added StarWars story for Data

ericsoderberghp:
#### What does this PR do?

Added StarWars story for Data.

This is to help explore and use Data with a variety of REST API endpoints.

no

no

backwards compatible


MikeKingdom:
Pretty cool and nice implementation. The API is pretty slow. I think the search needs a debounce. Does the API support a sort?

I did run into a CORS complaint from the browser while doing a couple searches. Not sure why since the same call seemed to work just before.

src/js/components/Data/stories/StarWars.js
```diff
@@ -0,0 +1,99 @@
...
+
+export const StarWars = () => {
+  const [result, setResult] = useState({});
+  const [view, setView] = useState({ search: '' });
+  const abortRef = useRef();
+
+  useEffect(() => {
+    // This API is a bit slow, abort any uncompleted requests.
+    abortRef?.current?.abort();
+    abortRef.current = new AbortController();
```
MikeKingdom, line 37: Cool.

MikeKingdom (approved):
Nice improvements. LGTM

src/js/components/DataSearch/DataSearch.js
```diff
@@ -19,12 +24,15 @@ export const DataSearch = ({ ...rest }) => {
       id={`${dataId}--search`}
       name="_search"
       icon={<Search />}
+      reverse={addSubmit}
```
taysea, line 27: I think I was expecting the search icon to always be at the start, regardless of if the data submits on change or on submit. Having it on the left seems like it aids quick identification of the search field

src/js/components/DataSearch/DataSearch.js
```diff
@@ -19,12 +24,15 @@ export const DataSearch = ({ ...rest }) => {
...
       type="search"
       {...rest}
     />
   );
 
-  if (noForm) content = <DataForm footer={false}>{content}</DataForm>;
+  if (noForm) {
```
taysea, line 33: Any reason this line changed? Seems like we can keep in the simplified format from before.

ericsoderberghp: reverted to simplified syntax

src/js/components/DataSearch/DataSearch.js
ericsoderberghp: Changed to always be at the start. I think I misinterpreted one of your earlier messages.

taysea (approved):
LGTM!

Issue #6551, opened 2022-12-26
useForwardedRef incorrectly updates the ref during the render phase


denis-sokolov:
In #5462 the function useForwardedRef was changed so that now the ref is mutated during the render phase of React. This violates React’s expected invariant that render functions have no side-effects and conflicts with Concurrent React.

In some cases React warns about that in the console with: “Cannot update a component while rendering a different component. To locate the bad setState() ...”

In the latest master, notice how one updateRef is correctly in an effect, and the other is outside of an effect:

https://github.com/grommet/grommet/blob/7c45cc933a67c02688eee584086f372f857118e2/src/js/utils/refs.js#L15-L21

[PR change](https://github.com/grommet/grommet/pull/5462/files#diff-85218aec9285d16ac3e83982890ff1160b83befbbeaa9d6fb44f5018e6b5f60bR17).

### Expected Behavior

useForwardedRef should only mutate ref in the useEffect.

### Actual Behavior

useForwardedRef mutates ref outside of an effect.

jcfilben:
Linking react docs section that advises not to do this: https://beta.reactjs.org/reference/react/useRef#caveats

MikeKingdom:
> Linking react docs section that advises not to do this: https://beta.reactjs.org/reference/react/useRef#caveats

Doesn't this statement allow this usage, ie. for initialization?

> Normally, writing or reading ref.current during render is not allowed. However, it’s fine in this case because the result is always the same, and the condition only executes during initialization so it’s fully predictable.

denis-sokolov:
@MikeKingdom, the code in useForwardedRef updates ref potentially at any render, not only the first one. If innerRef has been changed, the next time we render, useForwardedRef will mutate ref.

MikeKingdom:
> @MikeKingdom, the code in useForwardedRef updates ref potentially at any render, not only the first one. If innerRef has been changed, the next time we render, useForwardedRef will mutate ref.

I think you could avoid this and make it just like the example in the react docs reference if you changed useForwardedRef to do 
```jsx
if (ref?.current === null) updateRef(ref, innerRef);
```
but leave the use*Effects as is.
It would take some testing to see if this caused issues. I've run into some tricky issues with the timing of this as it is.

denis-sokolov:
@MikeKingdom, I can confirm that does not fix it.

[...]

PR #6564, opened 2023-01-05
fix useForwardedRef

MikeKingdom:
#### What does this PR do?
Change the implementation of `useForwardedRef` to not modify the associated ref during render. The preferred method of updating the forwarded ref according to React is its `useImperativeHandle` hook.

See #6551 for details on the issue.

The change to the DateInput-test is basically because the `.focus()` done via a setTimeout() can actually happen after the DOM has gone away (even though the timeout is only 1ms.)

refs.js

jest, manual and storybook

Go to the DateInput "Format" story and click on the icon to open the Calendar. Then click on a date in the calendar and close it.  It should use the ref that the Calendar forwarded from DateInput to set focus on the DateInput input field.

pass a ref into any component that uses `useForwardedRef()` and then display that ref from a `useEffect`.

#6551 

No

Yes

Backwards compatible

### DIFF
--- a/src/js/components/DateInput/DateInput.js
+++ b/src/js/components/DateInput/DateInput.js
@@ -193,7 +193,7 @@ Use the icon prop instead.`,
                 if (onChange) onChange({ value: normalizedValue });
                 if (open && !range) {
                   closeCalendar();
-                  setTimeout(() => ref.current.focus(), 1);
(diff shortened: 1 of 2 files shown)


ericsoderberghp (approved):

jcfilben (approved):
Looks good!

Issue #5538, opened 2021-08-24
Dropdown component: When it is required clicking on the dropdown itself is giving red border


AlokKrSinha:
Expected Behavior
Required error should only show up when on blur out  if item  not selected

Actual Behavior
Even when select  is open  required shows up under a specific usecase.

URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/s/compassionate-cori-oo0xo?file=/src/App.js

 Steps to Reproduce

1. form has attribute as validate="blur"
2. formfield required 
3. Open select option by clicking 
4. than click outside

 Your Environment

Grommet version: latest
Browser Name and version: Chrome, latest
Operating System and version (desktop or mobile): desktop, Mac

PR #6566, opened 2023-01-06
Fix Form onBlur validation with drop components

jcfilben:
#### What does this PR do?
Fixes an issue where onBlur validation was running prematurely on components with drops (Select, SelectMultiple, DateInput) when the drop is opened.

Tested with storybook story and added a jest test

Closes https://github.com/grommet/grommet/issues/5538

No

Yes

Backwards compatible


src/js/utils/DOM.js
```diff
@@ -65,6 +65,16 @@ export const containsFocus = (node) => {
   return !!element;
 };
 
+export const focusWithinDropPortal = (node) => {
```
taysea, line 68: small naming thought -- we have `containsFocus` function already so wondering if we name this `portalContainsFocus` or something that aligns with that. or `dropContainsFocus`

src/js/components/FormField/FormField.js
```diff
@@ -209,6 +213,7 @@ const FormField = forwardRef(
     const formKind = formContext.kind;
     const [focus, setFocus] = useState();
     const formFieldRef = useForwardedRef(ref);
+    const [dropFocused, setDropFocused] = useState();
```
taysea, line 216: I interpret this variable to be a boolean based on the naming but it seems like we're giving it a DOM element. naming ideas:
- focusedDrop
- focusedDropElement
- focusedElement

taysea (commented):
just one small comment on naming, otherwise looks good

src/js/components/FormField/FormField.js
jcfilben: Good note, I changed this to `focusedDropElement`

src/js/components/Form/stories/ValidateOnBlur.js
```diff
@@ -43,6 +43,25 @@ export const ValidateOnBlur = () => (
           <TextInput name="email" aria-label="email" type="email" />
         </FormField>
 
+        <FormField
+          label="Size"
+          name="select-size"
+          htmlFor="select-size"
```
taysea, line 49:
```suggestion
          htmlFor="select-size__input"
```
small detail to properly associate label with input

taysea (approved):
LGTM!

src/js/utils/DOM.js
ericsoderberghp, line 68: I don't see anything within this function that seems to have anything to do with "focus". It looks like this should be "withinDropPortal"?

src/js/components/FormField/FormField.js
```diff
@@ -502,7 +507,26 @@ const FormField = forwardRef(
...
-          if (contextOnBlur) contextOnBlur(event);
+
+          // if input has a drop and focus is within drop
+          // prevent onBlur validation from running until
+          // focus is no longer within the drop or input
+          if (
+            event.relatedTarget &&
+            focusWithinDropPortal(event.relatedTarget)
+          ) {
+            setFocusedDropElement(event.relatedTarget);
```
ericsoderberghp, line 518: It seems a bit odd to me that we are storing this in state. Can we just do the calculations each time this event handler is called?

jcfilben:
Converting this PR to a draft, I realized this solution isn't working well when a Form is contained within a Layer due to the `withinDropPortal` function.

src/js/utils/DOM.js
jcfilben: Renamed to `withinDropPortal`

src/js/components/FormField/FormField.js
jcfilben: Fixed

src/js/components/FormField/FormField.js
```diff
@@ -502,7 +509,20 @@ const FormField = forwardRef(
...
           setFocus(false);
-          if (contextOnBlur) contextOnBlur(event);
+
+          // if input has a drop and focus is within drop
+          // prevent onBlur validation from running until
+          // focus is no longer within the drop or input
+          const dropPortalId = withinDropPortal(event.relatedTarget);
+          if (
+            (dropPortalId === undefined ||
+              portalContext.indexOf(parseInt(dropPortalId, 10)) !== -1) &&
```
taysea, line 519: Can we add a small comment speaking to what portalContext is doing here?

src/js/utils/DOM.js
```diff
@@ -65,6 +65,17 @@ export const containsFocus = (node) => {
   return !!element;
 };
 
+export const withinDropPortal = (node) => {
```
ericsoderberghp, line 68: Since this is returning the drop portal id, what about calling this function `getDropPortalId`?

src/js/components/FormField/FormField.js
```diff
@@ -502,7 +509,20 @@ const FormField = forwardRef(
         }}
         onBlur={(event) => {
           setFocus(false);
-          if (contextOnBlur) contextOnBlur(event);
+
+          // if input has a drop and focus is within drop
+          // prevent onBlur validation from running until
+          // focus is no longer within the drop or input
+          const dropPortalId = withinDropPortal(event.relatedTarget);
```
ericsoderberghp, line 516: What about pushing more of this logic into the utility function? For example:
```
if (contextOnBlur && !formFieldRef.current.contains(event.relatedTarget) && !withinPortalContext(event.relatedTarget, portalContext)) {
  contextOnBlur(event);
}
```

src/js/components/FormField/FormField.js
```diff
@@ -214,6 +219,8 @@ const FormField = forwardRef(
     const { border: themeBorder } = formFieldTheme;
     const debounce = useDebounce();
 
+    const portalContext = useContext(PortalContext) || [];
```
ericsoderberghp, line 222: I'm wondering if we should have the default PortalContext value be `[]`, so we don't need the `|| []` here.

ericsoderberghp (commented):
Definitely improving!

src/js/components/FormField/FormField.js
jcfilben: Yes, some code moved around after I incorporated Eric's feedback so I ended up adding a comment in the utils/dom.js file

src/js/utils/DOM.js
jcfilben: This function changed around a bit in my latest commit. I left the name as `withinDropPortal` but I think with the changes this name now accurately reflects what the function is doing

src/js/components/FormField/FormField.js
jcfilben: done

src/js/components/FormField/FormField.js
jcfilben: done

ericsoderberghp (approved):
Thanks for moving more to DOM.js, I think it makes the component code read easier.

Issue #6565, opened 2023-01-06
Drop for TextInput with suggestions does not close when clicking on Layer


halocline:
### Expected Behavior

Any click outside the TextInput or its suggestions drop should close the drop.

### Actual Behavior

Clicks outside the input or suggestions drop do not close if the click target is on a layer.

"The crux of the issue is that the suggestions do not go away when you click on the layer - so that is what is up. Specifically this is an issue with we have with [home.hpe.com](http://home.hpe.com/) and search. User enters a search term and the results are on the layer."

### URL, screen shot, or Codepen exhibiting the issue

https://codesandbox.io/p/github/matsawka/publiclayer/main?file=%2Fsrc%2FApp.js

### Steps to Reproduce

1. Click within the search input - the suggestions Drop should open.
2. Click on the Layer - the drop does not close.

### Your Environment

- Grommet version: v2.28.0
- Browser Name and version: FireFox, Chrome, Safari
- Operating System and version (desktop or mobile): MacOS, Windows

PR #6567, opened 2023-01-06
Fix allowing suggestions drop to close when click occurs on layer

halocline:
#### What does this PR do?

Closes https://github.com/grommet/grommet/issues/6565

TextInput.js

Storybook using the following:
```
/* eslint-disable arrow-body-style */
import { FormClose, Search } from 'grommet-icons';
import React from 'react';
import {
  Box,
  Button,
  Header,
  Heading,
  Layer,
  Paragraph,
  TextInput,
} from '../..';

const allSuggestions = ['Apples', 'Oranges', 'Bananas'];

export const Temp = () => {
  const bodyRef = React.useRef();
  const [show, setShow] = React.useState(false);

  // Initializing layer as open for demo purposes.
  React.useEffect(() => {
    setShow(true);
  }, []);

  return (
    <Box height="100vh">
      <Header background="background-contrast" pad="small" justify="evenly">
        <SearchBar background="background-front" />
      </Header>
      <Box ref={bodyRef} background="background-front" fill />
      {show && (
        <Layer
          target={bodyRef.current}
          position="top"
          full
          modal
          margin="medium"
        >
          <Box as="section" pad="medium">
            <Header>
              <Heading level={2} margin="none">
                Search results layer
              </Heading>
              <Button icon={<FormClose />} onClick={() => setShow(false)} />
            </Header>
            <Paragraph>
              Place focus on the search input, then click on the layer.
              Previously, the suggestions drop would not close when a click
              occurred on the layer. A click on the layer should now close the
              drop.
            </Paragraph>
          </Box>
        </Layer>
      )}
    </Box>
  );
};

const SearchBar = ({ ...rest }) => {
  const [suggestions, setSuggestions] = React.useState(allSuggestions);

  const onChange = (event) => {
    const nextValue = event.target.value;
    setValue(nextValue);
    if (!nextValue) setSuggestions(allSuggestions);
    else {
      const regexp = new RegExp(`^${nextValue}`);
      setSuggestions(allSuggestions.filter((s) => regexp.test(s)));
    }
  };

  return (
    <Box width={{ min: 'medium', max: '33%' }} {...rest}>
      <TextInput
        placeholder="Search..."
        value={value}
        onChange={onChange}
        suggestions={suggestions}
        icon={<Search />}
        reverse
      />
    </Box>
  );
};

export default {
  title: 'Input/TextInput/Temp',
};

```

Storybook

 https://github.com/grommet/grommet/issues/6565

No

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -500,6 +500,7 @@ const TextInput = forwardRef(
                 event.relatedTarget !== dropRef.current
               ) {
                 setFocus(false);
+                closeDrop();
                 if (onBlur) onBlur(event);
               }
             }}


jcfilben (commented):
This solution is working but I'm wondering if we should be adjusting the way `onClickDocument` is working in DropContainer so that we can prevent this issue from happening in any component that uses a Drop. Currently `onClickDocument` doesn't trigger `onClickOutside` if a click is outside the drop but within a portal.

halocline:
> This solution is working but I'm wondering if we should be adjusting they way `onClickDocument` is working in DropContainer so that we can prevent this issue from happening in any component that uses a Drop. Currently `onClickDocument` doesn't trigger `onClickOutside` if a click is outside the drop but within a portal.

I'll check that out.

src/js/components/Drop/DropContainer.js
```diff
@@ -259,19 +259,25 @@ const DropContainer = forwardRef(
...
+            clickedPortal = node;
+          }
           node = node.parentNode;
         }
+
         if (
           clickedPortalId === null ||
-          portalContext.indexOf(clickedPortalId) !== -1
+          portalContext.indexOf(clickedPortalId) !== -1 ||
+          (clickedPortal && clickedPortal !== (ref || dropRef).current)
```
jcfilben, line 280: I think we might also need to check if `clickedPortal` is a child of `dropRef.current`. If you look at the Drop/Multiple story, there are examples of a drop within a drop. Clicking within the child drop closes the parent drop.

PR #6568, opened 2023-01-06
Enhanced Pagination to integrate with Data

ericsoderberghp:
#### What does this PR do?

Integrated Pagination with Data.
Of note:
- `step` and `page` are now available within the `view` object
- Pagination page changes trigger `onView` at the Data level containing the new `page`
- SpaceX story was changed to remove its managing of `page`

Still to consider:
- Should the pagination text summary shown in the SpaceX story be integrated into Pagination? Or, should we add a new PaginationSummary component that works like DataSummary?
- Wondering whether `step` belongs inside the `view` object or not. I'm leaning that direction but not 100% sure.
- For cases where the UI doesn't specify the `step` but the back-end returns it, how should that get fed in? Perhaps treat `step` more like `total` and less like `page`?

Data/index.d.ts

added a unit test
changed SpaceX story

^^^

followed existing pattern of unit tests

yes

yes

backwards compatible


MikeKingdom (approved):
Looks good. I was kinda on the fence about step being there since it seems kind of consumer specific but I think it fits pretty nicely in view.

src/js/components/Data/stories/SpaceX.js
```diff
@@ -185,50 +180,34 @@ export const Table = () => {
...
-          }}
+          onView={setView}
           toolbar
         >
           <DataTable
             columns={columns}
-            sort={{ ...sort, external: true }}
-            onSort={(opts) => setSort(opts)}
+            sort={{ ...view.sort, external: true }}
+            // onSort={(opts) => setSort(opts)} TODO: build inside DataTable
```
taysea, line 191: Is this a TODO for this PR or future?

src/js/components/Data/stories/SpaceX.js
```diff
@@ -185,50 +180,34 @@ export const Table = () => {
...
           />
+          {total > view.step && (
+            <Footer>
+              <Text>
+                Showing {pageResultStart}-{pageResultEnd} of {total}
+              </Text>
+              <Pagination />
```
taysea, line 198: It's nice how much this gets cleaned up.

taysea:
On the initial SpaceX story render, it says I have a filter applied (I would've expected no filters were applied and couldn't figure out what was filtered):
Screen Shot 2023-01-06 at 11 36 42 AM


When I clear filters, I lose the pagination control:

Screen Shot 2023-01-06 at 11 36 53 AM

src/js/components/Data/stories/SpaceX.js
ericsoderberghp: future pull request

ericsoderberghp:
> On the initial SpaceX story render, it says I have a filter applied (I would've expected no filters were applied and couldn't figure out what was filtered): Screen Shot 2023-01-06 at 11 36 42 AM
> 
> When I clear filters, I lose the pagination control:
> 
> Screen Shot 2023-01-06 at 11 36 53 AM

Should be fixed now.

src/js/components/Data/stories/SpaceX.js
```diff
@@ -131,38 +132,43 @@ const columns = [
...
-    rockets?.docs?.map(({ name, id }) => ({ value: id, label: name })) || [];
-
-  const docs = data?.docs || [];
+    fetchLaunches(view).then((response) => {
+      setResult({
+        data: response.docs,
+        filteredTotal: response.totalDocs,
+        page: response.page,
+      });
+      setTotal((prevTotal) => Math.max(prevTotal, response.totalDocs));
```
MikeKingdom, line 160: I'm not really sure why this is doing the max with the previous.

src/js/components/Data/stories/SpaceX.js
```diff
@@ -181,54 +187,42 @@ export const Table = () => {
...
           <DataTable
             columns={columns}
-            sort={{ ...sort, external: true }}
-            onSort={(opts) => setSort(opts)}
+            sort={{ ...view.sort, external: true }}
+            // TODO: in some other pull request,
+            // integrate sorting between Data and DataTable
+            // onSort={(opts) => setSort(opts)}
           />
+          {result.filteredTotal > view.step && (
```
MikeKingdom, line 207: Playing with the filter popup then clearing filters I got it into a state without the pagination controls:
image

src/js/components/Data/stories/SpaceX.js
ericsoderberghp: It appeared to me that you don't get the total number of items that are available in a request that has filtered some out. So, if your first request is for all Falcon 1 launches, you don't get anything in the response indicating that there are 205 total launches even though there are only 5 Falcon 1 launches. The Math.max() is a weak safety net in case the first request has a filter that doesn't get everything. I guess it would be more accurate to detect this case and do a subsequent query with a small limit just to get the total.

src/js/components/Data/stories/SpaceX.js
taysea: I ran into this as well. Steps were:
1. Filter by something
2. Clear filters

ericsoderberghp: Good catch. Added `defaultView` to allow the caller to control what "Clear filters" returns to. This allows the caller to persist `step` in the view.

src/js/components/Data/stories/SpaceX.js
ericsoderberghp: I've added a comment to explain.

MikeKingdom (approved):
This works pretty nicely. Great to see something with a fair amount of data work well with good performance

taysea (approved):
LGTM! Really nice how simplified the Pagination surface becomes when you use it in Data.

Issue #6344, opened 2022-09-21
Datatable rowDetails poor component interaction and visual bugs


aheintz:
The behaviour of rowDetails in Datatable is very limited and have visual issues. 

*Visual issues*
1. Hover highlight (background) of the angle/open icon is misaligned. visual1.png attached.
2. Using rowDetails and opening any row and switching window and back makes the FormDown/FormUp icon (open/close) on the first row highlight, with a different alignment than hover. visual2.png attached. In the Storybook example the entire first row is highlighted for no reason.

*Functional shortcoming*
It's not possible to have rowDetails controlled, i.e. using state to control which row is open, for example to use `onClickRow` to open/close and allow only one rowDetails open through a state variable. 

### Expected Behavior

*Visual issues*
1. Hover background color should be aligned
2. Only highlight Angle (open/close) button on hover, it's a 
3. FormDown and FormUp as default is IMHO the wrong way (<- opinion)

*Functional shortcomings*
1. It should be possible to control which row is open to ensure details are only shown for one row, and not have to have an open/close button (to be able to to use `onClickRow`. It does _not_ work returning null in the `rowDetail` to handle this, using this approach still shows an empty row (in my case 12 pixels high) and no way of controlling Open/Close button.

### URL, screen shot, or Codepen exhibiting the issue

See screenshots.

Cannot initialize codesandbox for Datatable example exhibiting the problems above, but the Storybook have similar problems for no 2.

### Steps to Reproduce

1. Cannot see this in the storybook, I use a very vanilla version with only color changes, and no datatable theming.
2. Go in to Storybook example, open a row, hover over table, switch to another window and switch back.

For the uncontrolled issue, I'm looking into creating a PR, but that'll take time to make something that's backwards compatible.
<img width="147" alt="visual1" src="https://user-images.githubusercontent.com/4854621/191614037-cbe4a05b-4381-4822-9189-59fbe9c1d44c.png">
<img width="250" alt="visual2" src="https://user-images.githubusercontent.com/4854621/191614040-4426c206-5db3-4159-ac3b-133eafbeeadf.png">

### Your Environment

- Grommet version: 2.23.0
- Browser Name and version: Chrome 105.0.5195.125 (arm64)
- Operating System and version (desktop or mobile): MacOS X Monterey 12.5.1

PR #6569, opened 2023-01-06
Fix for visual rowDetails in DataTable

aheintz:
#### What does this PR do?
1. Added a `justify="center"` to the ExpanderControl component to make the icon vertically centered and to handle non-standard themes or styling.
2. Fixed a rendering problem where the first line of the Datatable was selected if using Cmd-Tab to move back to the browser window and triggering an event in `use-keyboard.js` which caused this effect. Removed the "usingKeyboard", the only case I could find where this would be applicable would be using the `onSelect` use-case, but that still works as intended after manual testing. I would appreciate some input whether which is correct though.

`src/components/DataTable/ExpanderCell.js` and `src/components/DataTable/Body.js` are the only files changed except the updated snapshot.

1. Manual testing with different themes to verify the icon's position. The snapshot changes also indicate that the only affected are vertical (flexbox) alignment.
2. Manual tests of the removal of the `usingKeyboard`, I might need help here to verify it's not affecting something I'm not aware of, I'm not too familiar with grommet.

The storybooks should suffice.

This partially resolved https://github.com/grommet/grommet/issues/6344

No

No

Backwards compatible

### DIFF
--- a/src/js/components/DataTable/Body.js
+++ b/src/js/components/DataTable/Body.js
@@ -2,7 +2,6 @@
 import React, { forwardRef, memo, useContext } from 'react';
 import { ThemeContext } from 'styled-components';
 
-import { useKeyboard } from '../../utils';
 import { CheckBox } from '../CheckBox';
 import { InfiniteScroll } from '../InfiniteScroll';
 import { TableRow } from '../TableRow';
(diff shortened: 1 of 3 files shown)


aheintz:
Sorry, had a bit of a git issue. Closing this and reopening.

PR #6577, opened 2023-01-13
(FormField) fix: guard against Uncaught TypeError

buberdds:
In v2.29.0 FormField crashes with `Uncaught TypeError: Cannot read properties of null (reading 'getRootNode')
    at onFocus (FormField.js:500:1)` when a field uses `autoFocus` attr.  

#### What does this PR do?
Prevents FormField errors

manual 

Update `components/FormField/stories/Simple.js` with 

```
      <FormField label="Label" htmlFor="select">
        <TextArea autoFocus />
      </FormField>
    </Box>
```

Relates to [6562](https://github.com/grommet/grommet/pull/6562)

-
-
no
yes 
yes


src/js/components/FormField/FormField.js
```diff
@@ -494,6 +494,9 @@ const FormField = forwardRef(
         {...outerProps}
         style={outerStyle}
         onFocus={(event) => {
+          if (!formFieldRef.current) {
```
jcfilben, line 497: I'm thinking that even if `formFieldRef.current` is undefined we should still run line 504 `if (onFocus) onFocus(event);`.

Maybe we could change this section to be:
``` javascript
onFocus={(event) => {
  const root = formFieldRef.current?.getRootNode();
  if (root) {
    setFocus(
      containsFocus(formFieldRef.current) && shouldKeepFocus(root),
    );
  }
  if (onFocus) onFocus(event);
}}
```

jcfilben (commented):
Thanks for reporting and filing a fix for this bug! Added a comment below

src/js/components/FormField/FormField.js
buberdds: ahh yeah looks safer, thanks

src/js/components/FormField/FormField.js
```diff
@@ -501,10 +501,13 @@ const FormField = forwardRef(
         {...outerProps}
         style={outerStyle}
         onFocus={(event) => {
-          const root = formFieldRef.current.getRootNode();
-          setFocus(
-            containsFocus(formFieldRef.current) && shouldKeepFocus(root),
-          );
+          const root =
+            formFieldRef.current && formFieldRef.current.getRootNode();
```
MikeKingdom, line 505: Could simplify this to
```jsx
const root = formFieldRef.current?.getRootNode();
```

ericsoderberghp (approved):

PR #6579, opened 2023-01-13
Remove unnecessary focus() on List click

### LINKED ISSUE #6574: Cursor got misplaced and list automatically scrolls down when we have large number of items in List with Selection
### Expected Behavior
When we select an item in List with selection, that item should be selected and cursor should not get misplaced

### Actual Behavior
When we select an item, list scrolls down and cursor position got misplaced to some other item in the list.

https://codesandbox.io/s/suspicious-mccarthy-g94eib?file=/src/List.js

### Steps to Reproduce

1. Render List with Selection having large number of items in list so that we have scroll.
2. Select any item in the list For instance Item1
3. The cursor scrolls down and pointing to some other item in the list.

### Your Environment

- Grommet version: 2.25.3
- Browser Name and version: 108.0.5359.124
- Operating System and version (desktop or mobile): macOS Monterey Version 12.6

MikeKingdom:
When a List allows clicking on items via the `onClickItem`, it sets focus on the List container for various reasons including accessibility and WCAG compliance. However that .focus() causes the the browser to automatically scroll the container in some cases even though it's not necessary in this situation.

#### What does this PR do?
This removes the call to set focus() on the List container on item click. It turns out this really doesn't seem to be needed anymore. Without this extra focus() call, clicking on an item does cause the container to get focus, and it does it without causing an extra scroll attempt to make sure it's visible (which isn't needed in this situation.)  

List.js

Manual, jest, existing storybook.

- In storybook go to the List/onClickItem.
- Make the  window sized so that the list has a scrollbar
- click on an item
Before the change the List will scroll so it's at the very top of the viewport. After the change the List will not scroll (arrowing around or using the Tab key to navigate can still cause the viewport to scroll as needed.)

Fixes #6574 

No

Not required but...yes

Backwards compatible

### DIFF
--- a/src/js/components/List/List.js
+++ b/src/js/components/List/List.js
@@ -478,9 +478,6 @@ const List = React.forwardRef(
                         adjustedEvent.item = item;
                         adjustedEvent.index = index;
                         onClickItem(adjustedEvent);
-                        // put focus on the List container to meet WCAG
-                        // accessibility guidelines that focus remains on `ul`
-                        listRef.current.focus();
                         sendAnalytics({
                           type: 'listItemClick',
                           element: listRef.current,


taysea (approved):
LGTM, thanks Mike.

halocline (approved):

ericsoderberghp (approved):

Issue #6265, opened 2022-08-09
`WorldMap` component goes off after state update.


Ojas-Sri:
- `WorldMap` component's labels misalign after a state update to the `places` props.
- Code style : `<WorldMap places={state? places1 : places2} />`

### Expected Behavior

- The Place labels should be placed at their correct coordinates after state updates.

### Actual Behavior

- The Place labels misalign after state update to the `places` props.

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/s/grommet-world-map-issue-zyhtu5?file=/src/App.js

![PR (33)](https://user-images.githubusercontent.com/80472243/183626088-63e325ec-db78-41ae-bc80-2d148e911533.gif)

### Your Environment

- Grommet version: 2.25.1
- Browser Name and version: Chromium
- Operating System and version (desktop or mobile): Not tested

uldz:
Hi @jcfilben! I'd like to try it! Will it be the correct way to test this issue by adding a button in the local storybook story? Will it be a similar env to reproduce the issue?

jcfilben:
Hi @uldz awesome that you are interested in working on this! You're right, I think the best way to test this is to add a state button to a WorldMap storybook story. Make sure your branch is up to date with master and then run `yarn storybook` and test out the issue.

jcfilben:
Hey @uldz just checking in on this to see if there were any questions and if it something you are still interested in working on

uldz:
Hey, yes I’m in. Just had some hard times with my main project. I will try my best in few days🙂

jcfilben:
No worries! Just wanted to check in

uldz:
It looks like there are issues with render order - Drops render faster than ref places appear on WorldMap. Trying to figure out how to make render with already rendered dropTargets.

uldz:
Sorry, probably don't have much time for this now. If someone will get to this ticket, pls, reassign it. Thank you.

jcfilben:
No worries, appreciate the update! I will remove your assignment from this issue

hwdsjtu97:
I am interested in this issue. Can I have a try?

jcfilben:
Hey @hwdsjtu97 feel free to work on this and reach out if you have any questions!

PR #6587, opened 2023-01-19
fix: WorldMap placeContent not show in right position after re-render

hwdsjtu97:
#### What does this PR do?
Fix https://github.com/grommet/grommet/issues/6265
WorldMap component
`yarn test`
Test with the example in https://github.com/grommet/grommet/issues/6265

https://github.com/grommet/grommet/issues/6265

https://user-images.githubusercontent.com/60074937/213358789-4bfa3313-3d4d-47f7-8ecb-2f0588b3c59e.mov

No
Compatible


hwdsjtu97:
The issue in https://github.com/grommet/grommet/issues/6265 is due to the `targets` state is not cleaned up before each time the `placeProps` changes. The consequence is that when `placeProps` changes to `[]`, the `placeElements` becomes empty, but `targets` are still refs to the previous `placeElements` which are unmounted `<path>`. When the `placeProps` is set to places again, the `targets` won't update because [this condition ](https://github.com/grommet/grommet/blob/master/src/js/components/WorldMap/WorldMap.js#L544) will be false, so update to new placeElements will never happen

jcfilben (approved):
Nice work, this looks good!
Appreciate the explanation of the changes made 😃

ericsoderberghp (approved):

PR #6590, opened 2023-01-19
Changed DataTable to not default to height: 100%

ericsoderberghp:
#### What does this PR do?

Changed DataTable to not default to `height: 100%;`

I was noticing odd layout behavior across a variety of screen resolutions with DataTable and traced it down to all DataTables having `height: 100%;` causing them not to get laid out correctly. In many circumstances this wouldn't be noticed. It also explain the need at times to wrap DataTable in a Box to get it to behave better.

one line change

- manual testing in scenarios I was seeing problems in, particularly in grommet-designer using the new `<Data />` component
- review storybook stories

evaluate storybook stories at various screen resolutions

no test changes

prior Data -> DataTable story:
Screenshot 2023-01-19 at 12 14 09 PM

new Data -> DataTable story:
Screenshot 2023-01-19 at 12 14 51 PM

no

yes

If `fill` is set then there is no change. If `fill` is not set, then some of the workarounds to get the table to layout correctly will not be needed anymore. There could be visual changes at certain resolutions but I would argue those are defects.


jcfilben:
I'm having a hard time seeing the difference between the two screenshots in the PR description

ericsoderberghp:
> I'm having a hard time seeing the difference between the two screenshots in the PR description

The rows are different heights. In the prior version, the table would stretch to 100% height, causing the rows to stretch out taller. In the newer version, the rows control their own height and they remain so.

jcfilben:
> The rows are different heights. In the prior version, the table would stretch to 100% height, causing the rows to stretch out taller. In the newer version, the rows control their own height and they remain so.

Ah okay I see it now, thanks

jcfilben (approved):
I checked the DataTable stories in storybook and everything looked fine there. I think once this gets merged maybe we can just check that the examples on the design system site to make sure everything looks okay there too

taysea (approved):
I had the same thought as Jessica re: using Design System site to observe any unexpected changes. I'm comfortable moving forward for now.

PR #6604, opened 2023-01-26
Enhance button theme to support icon-only padding specification

taysea:
#### What does this PR do?

Adds support for custom padding for icon-only buttons (buttons that have `icon` and no `label`). Having the ability to specify icon only padding is desirable to create uniform dimension icon only buttons.

Additionally, it is desirable to be able to manage icon size from the theme and vary the icon's size depending on what size button it is contained in.

This PR adds:

- theme.button.icon.size (allows caller to specify what size the icon should render as when contained in a small, medium, or large button
```
  icon?: {
    size?: {
      small?: string;
      medium?: string;
      large?: string;
    };
  };
```

- theme.button.size[size].iconOnly.pad which allows the caller to specify pad for iconOnly pad across button sizes.

base.js

Tested in local story:

Screen Shot 2023-01-26 at 2 44 10 PM

Related to: https://github.com/grommet/hpe-design-system/issues/3117

Yes.

Yes.

Backwards compatible.


ericsoderberghp (approved):

src/js/components/Button/StyledButtonKind.js
```diff
@@ -37,7 +37,16 @@ const fontStyle = (props) => {
...
 
-const padFromTheme = (size = 'medium', theme, themeObj, kind) => {
+const padFromTheme = (size = 'medium', theme, themeObj, kind, iconOnly) => {
+  if (size && iconOnly && themeObj?.size?.[size]?.iconOnly?.pad) {
+    const pad = themeObj?.size?.[size]?.iconOnly?.pad;
+
+    return {
+      vertical: typeof pad === 'string' ? pad : pad.vertical,
+      horizontal: typeof pad === 'string' ? pad : pad.horizontal,
+    };
```
halocline, line 47: Minor. Not sure if either of these helps readability... 

```suggestion
    return typeof pad === 'object' ? pad : {
      vertical: pad,
      horizontal: pad,
    };
```

OR 

```suggestion
    return {
      vertical: pad.vertical || pad,
      horizontal: pad.horizontal || pad,
    };
```

src/js/components/Button/stories/CustomThemed/Active.js
```diff
@@ -1,26 +1,318 @@
...
-  >
-    <Box pad="large" align="center">
-      <Button hoverIndicator="light-1" onClick={() => {}} active>
-        {/*  When kind Button include children, it is treated as plain */}
-        <Box pad="small" direction="row" align="center" gap="small">
-          <Add />
-          <Text>Kind</Text>
-        </Box>
-      </Button>
+          height: '24px',
```
halocline, line 144: Noticing this comment and the height value. Is this working? I don't see the effect coming through. Also don't see it represented in base.d.ts or base.js.

taysea: Thanks for catching, this story was staged with the brand refresh theme and not intended to stay when we merge. The line-height question can be handled in a separate PR if necessary. Reverting the story to the original.

src/js/components/Button/StyledButtonKind.js
taysea: I'm inclined to leave as is.

ericsoderberghp:
Chromatic picked up some things that look a bit off.

Pagination large vertical alignment:
Screenshot 2023-02-01 at 5 17 03 PM

Button large icon and padding don't look right here:
Screenshot 2023-02-01 at 5 18 18 PM

Custom themed Button changed a bit unexpectedly:
Screenshot 2023-02-01 at 5 19 23 PM

Screenshot 2023-02-01 at 5 20 12 PM

src/js/components/Button/Button.js
```diff
@@ -242,6 +242,10 @@ const Button = forwardRef(
       return undefined; // pre-default, no kind
     }, [kindArg, kindObj, primary, secondary, theme]);
 
+    // for backwards compatibility, no-kind button theme did not
```
halocline, line 245: 👍

halocline (approved):

ericsoderberghp (approved):

PR #6606, opened 2023-01-27
Enhance button theme to support direction

taysea:
#### What does this PR do?

Adds `theme.button[kind].direction`to allow theme to drive the direction of icon + label.

src/js/themes/base.js

Locally in storybook:

Screen Shot 2023-01-26 at 4 08 19 PM

Added jest test as well.

Related to https://github.com/grommet/hpe-design-system/pull/3117

Yes.

Yes.

Backwards compatible.


src/js/components/Button/StyledButtonKind.js
```diff
@@ -30,7 +30,7 @@ const radiusStyle = (props) => {
 
 const fontStyle = (props) => {
   const size = props.sizeProp || 'medium';
-  const data = props.theme.text[size];
+  const data = props.theme.text[props.kind?.font?.size || size];
```
ericsoderberghp, line 33: Button index.d.ts has `kind?: string;`, which doesn't align with the property being an object.

ericsoderberghp: Also, shouldn't the size be coming from the theme instead of the property? Should it be something like `props.theme.text[props.theme?.[props.kind]?.font?.size`?

taysea: Removing this adjustment, it turns out the font size for the label was already supported so this change was not driving the styling.

ericsoderberghp (approved):

src/js/themes/base.js
```diff
@@ -469,7 +469,9 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       //   background: undefined,
       //   border: undefined,
       //   color: undefined,
+      //   direcion: undefined,
```
halocline, line 472:
```suggestion
      //   direction: undefined,
```

src/js/themes/base.js
```diff
@@ -482,11 +484,13 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       // },
       // primary: {
       //   font: {
+      //     size: undefined,
       //     weight: undefined,
       //   },
       //   background: undefined,
       //   border: undefined,
       //   color: undefined,
+      //   direcion: undefined,
```
halocline, line 493:
```suggestion
      //   direction: undefined,
```

src/js/themes/base.js
```diff
@@ -497,11 +501,13 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       // },
       // secondary: {
       //   font: {
+      //     size: undefined,
       //     weight: undefined,
       //   },
       //   background: undefined,
       //   border: undefined,
       //   color: undefined,
+      //   direcion: undefined,
```
halocline, line 510:
```suggestion
      //   direction: undefined,
```

src/js/themes/base.js
```diff
@@ -514,6 +520,7 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
```
halocline, line 523:
```suggestion
      //   direction: undefined,
```

halocline (commented):
Typos, otherwise looks good.

ericsoderberghp:
> Typos, otherwise looks good.

accidental Spanish version :)

halocline (approved):

Issue #6495, opened 2022-11-15
Dateinput component gets cropped for certain resolution, ideally 100% of the date input component should be seen in viewport.


vinayravindrajartarkar:
* Visit url https://v2.grommet.io/dateinput on chrome browser.
* On chrome tab at 100% zoom on click on select date icon in the rendered Dateinput component.
* You will see bottom 20% of the date input component hidden.
* The same can be seen if there is content below on scroll.

### Expected Behavior

* The complete Date input component should be shown.

### Actual Behavior

* At 100% zoom level in chrome in the url https://v2.grommet.io/dateinput, you will see 80% of the Date input component seen.
* Remaining 20% can only be seen on scroll.

### URL, screen shot, or Codepen exhibiting the issue

- Grommet version: 2.27.0
- Browser Name and version: Chrome 106.0.5249.119
- Operating System and version (desktop or mobile): MAC

vinayravindrajartarkar:
https://user-images.githubusercontent.com/97430588/201988827-018cfe9b-d916-49a5-99e0-0647a5c3cad5.mov

jcfilben:
Thanks for raising this issue! Currently the Drop has a threshold, 20% of the Drop can be out of view before it tries to change direction. I think we can change the behavior of Drop to adhere to the following instead:
- If the Drop is aligned in a direction and there is enough room to display the entire Drop in that direction, leave as is.
- If the Drop doesn't have enough room to be fully visible and there is more room in the other direction, change the Drop to display above/below.
- If the Drop doesn't have enough room to be fully visible and there is less room in the other direction leave the Drop in its current position.

The logic for the 20% threshold for drop was added in this PR: https://github.com/grommet/grommet/pull/3045. As we move towards this new method we may find that there was an important reason for having the 20% threshold, but willing to explore this new direction and see what happens.

jcfilben:
Update 1/20: I should have a PR up for his soon just playing around in storybook to make sure there aren't any weird side effects from this change

PR #6611, opened 2023-01-27
Drop - flip drop position if part of the drop isn't visible

jcfilben:
#### What does this PR do?
Changed Drop behavior to flip the drop if part of the drop isn't visible and there is more room in the opposite direction.

I also set the `responsive` prop to true by default. This aligns with what the documentation shows.

Closes https://github.com/grommet/grommet/issues/6495

No

Yes

Backwards compatible

### DIFF
--- a/src/js/components/Drop/DropContainer.js
+++ b/src/js/components/Drop/DropContainer.js
@@ -48,7 +48,7 @@ const DropContainer = forwardRef(
       onKeyDown,
       overflow = 'auto',
       plain,
-      responsive,
+      responsive = true,
       restrictFocus,
       stretch = 'width',
       trapFocus,
@@ -166,32 +166,20 @@ const DropContainer = forwardRef(
           let bottom;
           let maxHeight = containerRect.height;
           if (align.top) {
-            if (align.top === 'top') {
-              ({ top } = targetRect);
-            } else {
-              top = targetRect.bottom;
-            }
+            const alignToTop = align.top === 'top';
+            const switchTop = !alignToTop ? targetRect.top : targetRect.bottom;
+            top = alignToTop ? targetRect.top : targetRect.bottom;
 
-            // Calculate visible area underneath the control w.r.t window height
-            const percentVisibleAreaBelow =
-              100 - (targetRect.bottom / windowHeight) * 100;
-
-            // Check whether it is within 20% from bottom of the window or
-            // visible area to flip the control
-            // DropContainer doesn't fit well within visible area when
-            // percentVisibleAreaBelow value<=20%
-            // There is enough space from DropContainer to bottom of the window
-            // when percentVisibleAreaBelow>20%.
-
-            if (windowHeight === top || percentVisibleAreaBelow <= 20) {
+            if (
+              responsive &&
+              (windowHeight === top ||
+                (windowHeight < top + containerRect.height &&
+                  windowHeight - top < switchTop))
+            ) {
               // We need more room than we have.
               // We put it below, but there's more room above, put it above
               top = '';
-              if (align.top === 'bottom') {
-                bottom = targetRect.top;
-              } else {
-                ({ bottom } = targetRect);
-              }
+              bottom = alignToTop ? targetRect.bottom : targetRect.top;
               maxHeight = bottom;
               container.style.maxHeight = `${maxHeight}px`;
             } else if (top > 0) {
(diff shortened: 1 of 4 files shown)


britt6612:
one thing I saw in the storybook is since we have `responsive = true,` now set which is fine since docs stated that responsive is default to `true` thinking we need to add responsive={false} to [this story](https://deploy-preview-6611--sad-tereshkova-173d07.netlify.app/?path=/story/controls-drop-overflow--overflow).

src/js/components/Drop/DropContainer.js
```diff
@@ -166,32 +166,20 @@ const DropContainer = forwardRef(
...
-              ({ top } = targetRect);
-            } else {
-              top = targetRect.bottom;
-            }
-
-            // Calculate visible area underneath the control w.r.t window height
-            const percentVisibleAreaBelow =
-              100 - (targetRect.bottom / windowHeight) * 100;
+            const alignToTop = align.top === 'top';
+            const switchTop = !alignToTop ? targetRect.top : targetRect.bottom;
```
britt6612, line 170: just opinion not sure about the naming for `switchTop`

src/js/components/Drop/DropContainer.js
```diff
@@ -166,32 +166,20 @@ const DropContainer = forwardRef(
...
 
-            // Check whether it is within 20% from bottom of the window or
-            // visible area to flip the control
-            // DropContainer doesn't fit well within visible area when
-            // percentVisibleAreaBelow value<=20%
-            // There is enough space from DropContainer to bottom of the window
-            // when percentVisibleAreaBelow>20%.
-
-            if (windowHeight === top || percentVisibleAreaBelow <= 20) {
+            if (
```
britt6612, line 173: can we add a new comment for this section 
the old comment was clear but since the behavior changed just for future reference

britt6612 (commented):
Just a few comments but I think the direction of this approach is a good change.

src/js/components/Drop/DropContainer.js
jcfilben: Agree the name is confusing. I changed it to `switchTarget` since this props is new target position the drop will be aligned to if we flip the drop.

jcfilben:
> one thing I saw in the storybook is since we have responsive = true, now set which is fine since docs stated that responsive is default to true thinking we need to add responsive={false} to [this story](https://deploy-preview-6611--sad-tereshkova-173d07.netlify.app/?path=/story/controls-drop-overflow--overflow).

Changed the story to have `responsive={false}`

src/js/components/Drop/DropContainer.js
```diff
@@ -166,32 +166,27 @@ const DropContainer = forwardRef(
...
               // We put it below, but there's more room above, put it above
               top = '';
-              if (align.top === 'bottom') {
-                bottom = targetRect.top;
-              } else {
-                ({ bottom } = targetRect);
-              }
+              bottom = alignToTop ? targetRect.bottom : targetRect.top;
               maxHeight = bottom;
               container.style.maxHeight = `${maxHeight}px`;
```
ericsoderberghp, line 191: How come sometimes we set `container.style.maxHeight` without honoring `preserveHeight`? I was expecting we'd only set the style within `if (!preserveHeight) {` later down.

src/js/components/Drop/DropContainer.js
```diff
@@ -166,32 +166,27 @@ const DropContainer = forwardRef(
...
-            }
-
-            // Calculate visible area underneath the control w.r.t window height
-            const percentVisibleAreaBelow =
-              100 - (targetRect.bottom / windowHeight) * 100;
+            const alignToTop = align.top === 'top';
+            const switchTarget = !alignToTop
+              ? targetRect.top
+              : targetRect.bottom;
+            top = alignToTop ? targetRect.top : targetRect.bottom;
```
ericsoderberghp, line 173: My head is a bit confused trying to keep track of all of this. I'm wondering if there's a way to make it a bit clearer, perhaps by renaming some variables, or perhaps by restructuring the code. For instance, would it be clearer if we just checked `if (align.top === 'top') {` and then put everything for that case under it and then `} else {` for the other version, rather than trying to merge them together?

ericsoderberghp (commented):
A few initial comments to get started with.

ericsoderberghp (approved):

britt6612 (approved):
alot cleaner 🙌  and thanks for adding comment

taysea (approved):
Good work!

Issue #6589, opened 2023-01-19
List - focusIndicator should only be displayed when navigating with a keyboard


jcfilben:
### Expected Behavior
The focusIndicator (green outline) on the List container should only be displayed when interacting with the List using a keyboard. If a mouse is used the focusIndicator shouldn't be displayed.

### Actual Behavior
The focusIndicator is displayed both when interacting with a mouse and a keyboard.

### URL, screen shot, or Codepen exhibiting the issue
https://storybook.grommet.io/?path=/story/visualizations-list-pinned--pinned

### Steps to Reproduce

1. Go to the storybook link above
2. Click anywhere within the list container

### Your Environment

- Grommet version: 2.29.0
- Browser Name and version:
- Operating System and version (desktop or mobile):

jcfilben:
Resolved by #7143

Issue #6588, opened 2023-01-19
List - key render story error


jcfilben:
### Expected Behavior
Storybook stories shouldn't have any console errors

### Actual Behavior
The list key render storybook story has the following error: `Encountered two children with the same key`

### URL, screen shot, or Codepen exhibiting the issue
<img width="315" alt="Screen Shot 2023-01-19 at 11 35 55 AM" src="https://user-images.githubusercontent.com/54560994/213531009-da8742a0-8b44-429b-b38a-6531beb7274d.png">

### Steps to Reproduce

1. Run storybook locally
2. Navigate to the List/Key render story
3. Open the console

### Your Environment

- Grommet version: 2.29.0
- Browser Name and version:
- Operating System and version (desktop or mobile):

MadhurSikko:
@jcfilben I would like to be assigned this issue, can you give a little more insight as I am relatively new to open source? I was able to replicate the issue in List/Key_Render.

PR #6621, opened 2023-02-05
fix: List render with and focus only from keyboard.

umeshiscreative:
#### What does this PR do?
fix the bug when the List item key is unique and the List focus from the keyboard.

component/List/List.js

I used the storybook to test this component.

The list item key was accepted as an object from key render, so destructure it from an object. and mouse event intercept for accepting only keyboard.

#6589 #6588

No

No

No

### DIFF
--- a/src/js/components/List/List.js
+++ b/src/js/components/List/List.js
@@ -337,6 +337,7 @@ const List = React.forwardRef(
               setLastActive(active);
               updateActive(undefined);
             }}
+            onMouseDown={(event) => event.preventDefault()}
             {...ariaProps}
             {...rest}
           >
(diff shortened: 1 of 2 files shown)


src/js/components/List/List.js
```diff
@@ -337,6 +337,7 @@ const List = React.forwardRef(
               setLastActive(active);
               updateActive(undefined);
             }}
+            onMouseDown={(event) => event.preventDefault()}
```
jcfilben, line 340: This change breaks the drag and drop re-ordering functionality in list. To prevent the focus indicator from being displayed if we are interacting with the component using our mouse I think we just need to add the following to the StyledList:
```
&:focus:not(:focus-visible) {
    ${unfocusStyle()}
  }
```
See StyledButton for an example

umeshiscreative: ok

umeshiscreative:
Moving PR to the https://github.com/grommet/grommet/pull/7143.

PR #6622, opened 2023-02-06
Fix DataTable size parameter to allow for css string

wrdaigle:
#### What does this PR do?

Currently, the DataTable size parameter only allows for a size that is defined in the theme (i.e. small, medium, large, etc.). This PR allows for a custom css string to be provided instead. 

- Checkout the PR branch. 
- Run storybook
- Navigate to http://localhost:9001/?path=/story/visualizations-datatable-sized--sized-data-table
- Update to the [associated story](https://github.com/grommet/grommet/blob/master/src/js/components/DataTable/stories/Sized.js) to test a standard css value (i.e. '200px')

I have not developed test for this PR since it is a pretty simple change.  Let me know if you would like me to do so

I have a use case in which I need to fill a container (Box) with a data table.  Without this PR, I haven't found another way to accomplish this. 

I don't think so. The docs already say that a 'string' can be provided. I'm not sure if 'Any CSS value' would be more accurate. 

It it backward compatible.

### DIFF
--- a/src/js/components/DataTable/StyledDataTable.js
+++ b/src/js/components/DataTable/StyledDataTable.js
@@ -94,7 +94,7 @@ const StyledDataTableBody = styled(TableBody)`
     `
     display: block;
     width: 100%;
-    max-height: ${props.theme.global.size[props.size]};
+    max-height: ${props.theme.global.size[props.size] || props.size};
     overflow: auto;
   `}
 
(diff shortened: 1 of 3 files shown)


jcfilben (commented):
Nice enhancement, would you be willing to add a unit test for this?

wrdaigle:
OK, @jcfilben. The tests are added. I'm a newbie at jest snapshots, so any critical review is welcome.  Any idea why my CircleCI checks are failing?

jcfilben:
> OK, @jcfilben. The tests are added. I'm a newbie at jest snapshots, so any critical review is welcome. Any idea why my CircleCI checks are failing?

Thanks for adding the tests! We recently updated our circle ci configuration file. I think if we update your branch with master that will probably fix the issue

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

PR #6629, opened 2023-02-08
Button text color handle dark and transparent backgrounds

jcfilben:
#### What does this PR do?
Fixes an issue on dark transparent background where the text is the wrong color.

Just a note about the snapshot changes: for background colors that are opaque the snapshots are changing to use the css property `background-color` instead of `background`. In addition the text color is added in the snapshot, note that in most cases the snapshot already includes the text color so the text color is being duplicated in some cases. I couldn't find a good way around this. Visually I didn't notice any problems from this change but since this code is used across so many components it would be good to get extra eyes here.

Added a unit test and tested with the following story:
```javascript
import React from 'react';

import {
  Box,
  Grommet,
  Button,
} from 'grommet';
import { hpe } from 'grommet-theme-hpe';

const myTheme = deepMerge(hpe, {
  button: {
    'background-contrast': {
      background: 'background-contrast',
      color: 'text-strong',
    },
    active: {
      'background-contrast': {
        color: 'text-strong',
      },
    },
  },
});

export const Test = () => (
  <Grommet theme={myTheme}>
    <Box
      background={{ dark: true, color: 'background' }}
      pad="large"
      gap="medium"
      align="start"
    >
      <Button label="Test button" kind="background-contrast" />
      <Button label="Active Test button" kind="background-contrast" active />
    </Box>
    <Box pad="large" gap="medium" align="start">
      <Button label="Test button" kind="background-contrast" />
      <Button label="Active Test button" kind="background-contrast" active />
    </Box>
  </Grommet>
);

export default {
  title: 'Controls/Button/Test',
};
```

closes https://github.com/grommet/grommet/issues/6301

Maybe? Not sure if we should mention that we support `button.kind.background` and `button.kind.background.color`? Currently we just document `button.kind.background.color`.

Yes

Backwards compatible

### DIFF
--- a/src/js/components/Pagination/__tests__/__snapshots__/Pagination-test.tsx.snap
+++ b/src/js/components/Pagination/__tests__/__snapshots__/Pagination-test.tsx.snap
@@ -1596,7 +1596,8 @@ exports[`Pagination should apply button kind style when referenced by a string 1
   border-width: 2px;
   border-color: skyblue;
   padding: 2px 20px;
-  background: #33333310;
+  background-color: #33333310;
+  color: #000000;
   background-color: rgba(202,156,234,1);
   color: #444444;
   border-color: transparent;
--- a/src/js/utils/background.js
+++ b/src/js/utils/background.js
@@ -169,6 +169,7 @@ export const backgroundAndTextColors = (backgroundArg, textArg, theme) => {
 
     if (backgroundColor && canExtractRGBArray(backgroundColor)) {
       const colorArray = getRGBArray(backgroundColor);
+      // check if the alpha value is less than 0.5
       if (colorArray[3] < 0.5) transparent = true;
     }
     if (shade) {
--- a/src/js/utils/background.js
+++ b/src/js/utils/background.js
@@ -1,6 +1,12 @@
 import { css } from 'styled-components';
 
-import { colorIsDark, getRGBA, normalizeColor } from './colors';
+import {
+  colorIsDark,
+  getRGBA,
+  normalizeColor,
+  canExtractRGBArray,
+  getRGBArray,
+} from './colors';
 
 // evalStyle() converts a styled-components item into a string
 const evalStyle = (arg, theme) => {
@@ -159,8 +165,16 @@ export const backgroundAndTextColors = (backgroundArg, textArg, theme) => {
   } else {
     backgroundColor = normalizeBackgroundColor(background, theme);
     const shade = darkContext(backgroundColor, theme);
+    let transparent;
+
+    if (backgroundColor && canExtractRGBArray(backgroundColor)) {
+      const colorArray = getRGBArray(backgroundColor);
+      if (colorArray[3] < 0.5) transparent = true;
+    }
     if (shade) {
       textColor = normalizeColor(text[shade] || text, theme, shade === 'dark');
+    } else if (transparent && text) {
+      textColor = normalizeColor(text, theme);
     } else {
       // If we can't determine the shade, we assume this isn't a simple color.
       // It could be a gradient. backgroundStyle() will take care of that case.
--- a/src/js/components/Accordion/__tests__/__snapshots__/Accordion-test.tsx.snap
+++ b/src/js/components/Accordion/__tests__/__snapshots__/Accordion-test.tsx.snap
@@ -4259,7 +4259,8 @@ exports[`Accordion custom accordion 1`] = `
 }
 
 .c3:hover {
-  background: #33333310;
+  background-color: #33333310;
+  color: #000000;
 }
 
 .c3:focus {
@@ -4572,7 +4573,8 @@ exports[`Accordion custom accordion 2`] = `
 }
 
 .c3:hover {
-  background: #33333310;
+  background-color: #33333310;
+  color: #000000;
 }
 
 .c3:focus {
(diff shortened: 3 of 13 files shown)


src/js/utils/background.js
```diff
@@ -159,8 +165,16 @@ export const backgroundAndTextColors = (backgroundArg, textArg, theme) => {
   } else {
     backgroundColor = normalizeBackgroundColor(background, theme);
     const shade = darkContext(backgroundColor, theme);
+    let transparent;
+
+    if (backgroundColor && canExtractRGBArray(backgroundColor)) {
+      const colorArray = getRGBArray(backgroundColor);
+      if (colorArray[3] < 0.5) transparent = true;
+    }
     if (shade) {
```
taysea, line 174: Can it ever be a shade and transparent? if so will the current logic catch that?

src/js/utils/background.js
```diff
@@ -159,8 +165,16 @@ export const backgroundAndTextColors = (backgroundArg, textArg, theme) => {
   } else {
     backgroundColor = normalizeBackgroundColor(background, theme);
     const shade = darkContext(backgroundColor, theme);
+    let transparent;
+
+    if (backgroundColor && canExtractRGBArray(backgroundColor)) {
+      const colorArray = getRGBArray(backgroundColor);
+      if (colorArray[3] < 0.5) transparent = true;
```
taysea, line 172: add a comment that `[3]` refers to the alpha value?

src/js/utils/background.js
jcfilben: if the alpha value is < 0.5 shade will be undefined so shade and transparent will never both be defined.

src/js/utils/background.js
jcfilben: I added a comment in the latest commit

src/js/utils/background.js
```diff
@@ -159,8 +165,17 @@ export const backgroundAndTextColors = (backgroundArg, textArg, theme) => {
   } else {
     backgroundColor = normalizeBackgroundColor(background, theme);
     const shade = darkContext(backgroundColor, theme);
+    let transparent;
+
+    if (backgroundColor && canExtractRGBArray(backgroundColor)) {
+      const colorArray = getRGBArray(backgroundColor);
+      // check if the alpha value is less than 0.5
+      if (colorArray[3] < 0.5) transparent = true;
```
taysea, line 173: Further up in this function we use `0.3` as a threshold if something is translucent or not. Should we align or is this intentionally different?

src/js/utils/background.js
```diff
@@ -159,8 +165,17 @@ export const backgroundAndTextColors = (backgroundArg, textArg, theme) => {
...
+    }
     if (shade) {
       textColor = normalizeColor(text[shade] || text, theme, shade === 'dark');
+    } else if (transparent && text) {
+      textColor = normalizeColor(text, theme);
```
taysea, line 178: The parent container has `inherit` for text color. Not sure if it's preferred to just use `inherit` here as opposed to recalculating as value?
```suggestion
      textColor = 'inherit';
```

src/js/utils/background.js
taysea, line 173: Further up in this function we use `0.3` to determine if a background is translucent. Should we align or is this intentionally different?

jcfilben: The section above that has `0.3` is used when we define `background.color` in the theme and it looks like there are some different issues with the code in this section. For example if a background color has vey little transparency it doesn't take that into account and shows white background & white text, so that might be a separate issue to look into.

We could change `colorArray[3] < 0.5` to `colorArray[3] < 0.3`, this will affect the transpancy level we flip the text color at. However I think `0.5` works better than `0.3`

Here is an example where we have `colorArray[3] < 0.3`
Screen Shot 2023-02-13 at 12 28 50 PM


and here is the same example with `colorArray[3] < 0.5`
Screen Shot 2023-02-13 at 12 29 17 PM

taysea (approved):
Thanks for handling this!

src/js/components/Pagination/__tests__/__snapshots__/Pagination-test.tsx.snap
```diff
@@ -1596,7 +1596,8 @@ exports[`Pagination should apply button kind style when referenced by a string 1
   border-width: 2px;
   border-color: skyblue;
   padding: 2px 20px;
-  background: #33333310;
+  background-color: #33333310;
+  color: #000000;
   background-color: rgba(202,156,234,1);
```
ericsoderberghp, line 1601: Something looks a little strange here. We seem to be setting `background-color` and `color` twice each.

jcfilben: Yes, I agree that this behavior isn't great. While I was working on this it felt like the entire way we are setting background and colors in `background.js` needs an overhaul. If you look at the snapshots prior to the changes made in this PR, there are lots places where we set background and color multiple times within the same css class (for an example look at the snapshot for 'Button kind primary button 1' class 'c3' another good place to look is at the Pagination snapshots). I wasn't sure if we want to handle that in this PR or separately?

ericsoderberghp: OK. Thanks for the explanation.

ericsoderberghp (approved):

PR #6631, opened 2023-02-09
Changed DataFilter to fix an issue with the range property.

ericsoderberghp:
#### What does this PR do?

Changed DataFilter to fix an issue with the range property.
It was already correctly converting the range from Data which starts as an object: `{ min: 1, max: 100 }` and is converted to an array `[1, 100]`.
But, it wasn't converting the `range` property value similarly, until now. :) 

the very beginning, it's a good place to start

added a unit test

manually

followed existing test patterns

no

yes

backwards compatible


MikeKingdom (approved):
LGTM

Issue #6626, opened 2023-02-07
DataFilters layer should have overflow property


JCoder121:
### Expected Behavior
When there are many filters, the DataFilters component should overflow so that the user can see all filter content.
<img width="1101" alt="image" src="https://user-images.githubusercontent.com/23144209/217326783-2ca92a6c-d026-4960-9fd7-4c33a8c0bb0e.png">

### Actual Behavior
There is no overflow behavior, so some content exceeds the container
<img width="1131" alt="image" src="https://user-images.githubusercontent.com/23144209/217326659-d46e3aff-056d-494c-825f-b3103ba0e90b.png">

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #6634, opened 2023-02-10
Changed DataFilters to have overflow

ericsoderberghp:
#### What does this PR do?

Changed DataFilters to have overflow when using `drop` or `layer`.

The snapshot changes are primarily due to two reasons:
- the restructuring of DataForm
- adding `flex={false}` to FormField

DataForm.js

storys

storys

no tests were changed in the creation of this pull request

#6626 

no

yes

backwards compatible

### DIFF
--- a/src/js/components/DataFilters/DataFilters.js
+++ b/src/js/components/DataFilters/DataFilters.js
@@ -63,7 +63,7 @@ export const DataFilters = ({ drop, children, heading, layer, ...rest }) => {
     </Box>
   );
 
-  let filters;
+  let content = children;
   if (Children.count(children) === 0) {
     let filtersFor;
     if (!properties && data && data.length)
@@ -75,18 +75,17 @@ export const DataFilters = ({ drop, children, heading, layer, ...rest }) => {
     else if (typeof properties === 'object')
       filtersFor = Object.keys(properties);
     else filtersFor = [];
-    filters = filtersFor.map((property) => (
+    content = filtersFor.map((property) => (
       <DataFilter key={property} property={property} />
     ));
     if (view?.sort) {
-      filters.push(<DataSort key="_sort" />);
+      content.push(<DataSort key="_sort" />);
     }
   }
 
-  const content = (
+  content = (
     <DataForm
       pad={controlled ? 'medium' : undefined}
-      gap="xsmall"
       onDone={() => setShowContent(false)}
       onTouched={
         controlled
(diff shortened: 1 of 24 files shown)


src/js/components/DataFilters/DataFilters.js
```diff
@@ -86,7 +86,7 @@ export const DataFilters = ({ drop, children, heading, layer, ...rest }) => {
   const content = (
     <DataForm
       pad={controlled ? 'medium' : undefined}
-      gap="small"
+      gap="xsmall"
```
taysea, line 89: HPE FormFields already have spacing built into FormField styling, so `gap` shouldn't be used at all. Wondering if this value should be removed or theme driven?

src/js/components/DataFilters/DataFilters.js
```diff
@@ -100,7 +100,7 @@ export const DataFilters = ({ drop, children, heading, layer, ...rest }) => {
               }))
           : undefined
       }
-      {...(!controlled ? rest : {})}
+      {...(!controlled ? rest : { fill: 'vertical' })}
```
taysea, line 103: Is controlled when the user isn't passing their own DataFilter? it feels like fill: vertical would always be helpful if that enables the overflow scroll.

src/js/components/DataFilters/DataFilters.js
```diff
@@ -49,7 +49,7 @@ export const DataFilters = ({ drop, children, heading, layer, ...rest }) => {
   );
 
   const clearControl = badge && (
-    <Box flex={false}>
+    <Box flex={false} margin={{ start: 'small' }}>
```
taysea, line 52: Not sure where to comment on it, but the gap between the layer footer buttons is too large. In HPE patterns, we use gap="small"

src/js/components/DataFilters/DataFilters.js
ericsoderberghp: When rendering inline, it allows the caller to decide how to handle it. Do you have a specific scenario you are envisioning?

src/js/components/DataFilters/DataFilters.js
ericsoderberghp: Good catch. I've made a few layout adjustments to address.

src/js/components/DataFilters/DataFilters.js
ericsoderberghp: FYI, the default `gap` in Footer is "medium". Explicitly made it "small" for DataForm.

taysea (approved):
LGTM!

Issue #6637, opened 2023-02-13
Calendar - Replace Heading with Text


jcfilben:
Follow on to: https://github.com/grommet/hpe-design-system/issues/3016
### Expected Behavior
The Calendar's month and year shouldn't be a Heading. This causes the screenreader to interpret it as a heading landmark on the page.

### Actual Behavior
We should use the Text component instead of Heading for the month and year. The month and year container should be a Header component.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #6639, opened 2023-02-14
Calendar - add `calendar[size].title` prop to theme

jcfilben:
#### What does this PR do?
Adds a new prop `calendar[size].headingSize` to the theme. Specifying this prop will override the `calendar.heading.level` prop and change the calendar to use `<Text>` instead of `<Heading>` to display the date and year. This prop also gives the flexibility to specify different font sizes for each calendar size (small, medium, and large).

In addition to this PR I will file a PR on grommet-theme-hpe to opt into this behavior for the hpe theme.

Added a jest test and tested by adding a custom theme to calendar stories

Related to https://github.com/grommet/grommet/issues/6637

Yes, we need to document the new theme prop as well as mention that using `calendar[size].headingSize` is preferred over `calendar.heading.level`

Yes

Backwards compatible

### DIFF
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -14,6 +14,7 @@ import { defaultProps } from '../../default-props';
 
 import { Box } from '../Box';
 import { Button } from '../Button';
+import { Header } from '../Header';
 import { Heading } from '../Heading';
 import { Keyboard } from '../Keyboard';
 import { Text } from '../Text';
@@ -587,19 +588,24 @@ const Calendar = forwardRef(
           ? theme.calendar.icons.small.next
           : theme.calendar.icons.next;
 
+      const monthAndYear = reference.toLocaleDateString(locale, {
+        month: 'long',
+        year: 'numeric',
+      });
+
       return (
         <Box direction="row" justify="between" align="center">
-          <Box flex pad={{ horizontal: headingPadMap[size] || 'small' }}>
-            {theme.calendar.heading && theme.calendar[size].titleSize ? (
-              <Text weight="bold" size={theme.calendar[size].titleSize}>
-                {reference.toLocaleDateString(locale, {
-                  month: 'long',
-                  year: 'numeric',
-                })}
+          <Header flex pad={{ horizontal: headingPadMap[size] || 'small' }}>
+            {theme.calendar[size]?.title ? (
+              <Text
+                weight={theme.calendar[size].title?.weight || 'normal'}
+                size={theme.calendar[size].title?.size || 'medium'}
+              >
+                {monthAndYear}
               </Text>
             ) : (
               // theme.calendar.heading.level should be removed in v3 of grommet
-              // theme.calendar[size].titleSize should be used instead
+              // theme.calendar[size].title should be used instead
               <Heading
                 level={
                   size === 'small'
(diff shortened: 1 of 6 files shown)


taysea:
I'm wondering if `headingSize` might be a little misleadinng since it's not actually a Heading -- will people assume that their theme's Heading sizes would be used instead of text?

jcfilben:
> I'm wondering if headingSize might be a little misleadinng since it's not actually a Heading -- will people assume that their theme's Heading sizes would be used instead of text?

Good note, I'll see if I can find a different name for this

jcfilben:
> > I'm wondering if headingSize might be a little misleadinng since it's not actually a Heading -- will people assume that their theme's Heading sizes would be used instead of text?
> 
> Good note, I'll see if I can find a different name for this

renamed to titleSize

src/js/components/Calendar/Calendar.js
```diff
@@ -589,23 +590,36 @@ const Calendar = forwardRef(
...
-              margin="none"
-              overflowWrap="normal"
-            >
-              {reference.toLocaleDateString(locale, {
-                month: 'long',
-                year: 'numeric',
-              })}
-            </Heading>
+            {theme.calendar.heading && theme.calendar[size].titleSize ? (
+              <Text weight="bold" size={theme.calendar[size].titleSize}>
```
taysea, line 594: Enforcing bold here might not be the best idea because in v5 of HPE theme, we are leaning away from use of bold in typography. Wondering if we should have `theme.calendar[size].title` where any text props are supported?

In base.js, the default weight might be bold but then in HPE theme we can override with 500 or whatever is appropriate?

src/js/components/Calendar/Calendar.js
```diff
@@ -589,23 +590,36 @@ const Calendar = forwardRef(
...
-              size={size}
-              margin="none"
-              overflowWrap="normal"
-            >
-              {reference.toLocaleDateString(locale, {
-                month: 'long',
-                year: 'numeric',
-              })}
-            </Heading>
+            {theme.calendar.heading && theme.calendar[size].titleSize ? (
```
ericsoderberghp, line 593: Why do we need the check for `theme.calendar.heading`?

src/js/components/Calendar/Calendar.js
```diff
@@ -589,23 +590,36 @@ const Calendar = forwardRef(
...
+              <Text weight="bold" size={theme.calendar[size].titleSize}>
+                {reference.toLocaleDateString(locale, {
```
ericsoderberghp, line 595: Since we're duplicating this code, can we put it in a const variable first?

src/js/components/Calendar/Calendar.js
ericsoderberghp: good idea!

src/js/components/Calendar/Calendar.js
jcfilben: Yes, fixed in latest commit

src/js/components/Calendar/Calendar.js
jcfilben: Removed this extra check in latest commit

src/js/components/Calendar/Calendar.js
```diff
@@ -589,23 +590,36 @@ const Calendar = forwardRef(
       return (
         <Box direction="row" justify="between" align="center">
```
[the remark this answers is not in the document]
jcfilben: Changed Box to Header in latest commit

src/js/components/Calendar/Calendar.js
jcfilben: > Wondering if we should have theme.calendar[size].title where any text props are supported?

Done in latest commit

taysea (approved):
LGTM.

ericsoderberghp (approved):

Issue #6638, opened 2023-02-14
Data and DataFilters should accept nested objects


JCoder121:
### Expected Behavior
Data and DataFilters should handle nested objects as properties/filters.

### Actual Behavior
Errors out/does not reflect the correct filter/table column

### URL, screen shot, or Codepen exhibiting the issue
When specifying properties using one of the following methods below,
<img width="494" alt="image" src="https://user-images.githubusercontent.com/23144209/218827738-f027e4e2-d174-4fa6-b224-8f5b4fc3b9f9.png">

<img width="528" alt="image" src="https://user-images.githubusercontent.com/23144209/218827793-6063863c-086f-4341-9185-1eaa9e6f0f69.png">

Results in default filter behavior with no affect on filter content.
![image](https://user-images.githubusercontent.com/23144209/218827874-6fe191c3-d556-4e5e-ab6b-960b241c675f.png)

This is the format of the mock data we are using:
<img width="271" alt="image" src="https://user-images.githubusercontent.com/23144209/218828723-cac8fdd2-9d3c-476f-af50-6cfa89310800.png">

We want to filter on `hardware.health.summary`, `state.connected`, `hardware.powerState`, and `hardware.model`. In the live site example without refactoring, this is how we designated what to filter on:
<img width="494" alt="image" src="https://user-images.githubusercontent.com/23144209/218829712-622ec7c3-01f3-4add-a79c-f9824981ca02.png">

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #6645, opened 2023-02-22
Changed Data to handle property paths

ericsoderberghp:
#### What does this PR do?

Changes Data to handle nested objects in the data as properties.
Now, you can use `properties={ 'location.city': { ...} }` and it will do what you might suspect.

Also, I combined `options` and `range` into a single `useMemo()` in DataFilter to be more efficient.

Take a look at the new Complex story under Data.

Added the Complex story to test interaction with both range and options versions.

Complex story

no jest changes

#6638 

yes

yes

backwards compatible

### DIFF
--- a/src/js/components/Data/DataForm.js
+++ b/src/js/components/Data/DataForm.js
@@ -79,7 +79,7 @@ const unflatten = (formValue) => {
       let parent = result;
       while (parts.length > 1) {
         const sub = parts.shift();
-        parent[sub] = {};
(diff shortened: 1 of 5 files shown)


jcfilben (commented):
Functionality looks good! Can we add a unit test for this?

taysea (approved):
Looks good, thanks for adding story and test.

jcfilben (approved):
Looks good!

PR #6647, opened 2023-02-22
make changes based on storybook check

britt6612:
#### What does this PR do?
This PR puts `toolBar` under Data stories and adds proptype for `Skeleton` on Box
general-propTypes.js
storybook

toolbar should live under Data 
error on story because we are passing a ternary(object) to skeleton on Box
Screen Shot 2023-02-22 at 2 20 38 PM

release prep https://github.com/grommet/hpe-design-system/issues/3086
no
no
backwards compatible

### DIFF
--- a/src/js/utils/general-prop-types.js
+++ b/src/js/utils/general-prop-types.js
@@ -217,8 +217,6 @@ export const skeletonPropType = PropTypes.oneOfType([
   }),
 ]);
 
-// { message: { start: 'Loading', end: 'Content Loaded' } }
-
 const dimSizeType = PropTypes.oneOf([
   'xxsmall',
   'xsmall',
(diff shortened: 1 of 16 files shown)


src/js/utils/general-prop-types.js
```diff
@@ -203,6 +203,7 @@ export const skeletonColorsPropType = PropTypes.shape({
 
 export const skeletonPropType = PropTypes.oneOfType([
   PropTypes.bool,
+  PropTypes.object,
```
jcfilben, line 206: I think instead of adding `PropTypes.object` if we change `end: PropTypes.end` to `end: PropTypes.string` that should resolve the issue

src/js/utils/general-prop-types.js
```diff
@@ -217,6 +218,8 @@ export const skeletonPropType = PropTypes.oneOfType([
   }),
 ]);
 
+// { message: { start: 'Loading', end: 'Content Loaded' } }
+
```
jcfilben, line 222:
```suggestion
```

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

PR #6648, opened 2023-02-22
fix: update "children" type

vgpena:
#### What does this PR do?
This PR updates the `children` type of the Checkbox to include `React.ReactNode`. This fixes a regression introduced when bringing in explicit typings for React Children. It also makes Checkbox consistent with the other changes in https://github.com/grommet/grommet/pull/6066 and with React's `PropsWithChildren` type signature ([link](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L800)).

Compare the type update here to `RadioButton > children` and React's `PropsWithChildren`.

- tested locally in a linked environment
- manual testing steps (see below)

1. On the `master` branch, rename `Checkbox/stories/Children.js` to  `Checkbox/stories/Children.tsx`.
2. On lines 17-19 of `Checkbox/stories/Children.tsx`, substitute the function child component with a plain `React.ReactNode`. (e.g., `<span>foo</span>`). You should get a TypeScript error.
3. Stash that change, switch to this branch, and apply the change (or some equivalent -- we want the updated story file and the updated type definition on the same branch). You should no longer see a TypeScript error in `Checkbox/stories/Children.tsx`.

This issue was discovered because of a project that consumes Grommet and uses React's `PropsWithChildren` type.

N/A

N/A

N/A

Yes

Backwards-compatible


jcfilben:
Hey @vgpena I looked a bit more at the change I made to the RadioButton `children` type in the React 18 pull request and I think that I shouldn't have set `children?: React.ReactNode | Function;` and it should just be `children?: Function;`. I thought about this more and couldn't think of a case where we would want to pass a react node. In addition, the documentation for RadioButton and CheckBox the `children` prop only show `function` as a valid type to pass.

Could you share the use case for wanting to pass a react node? Maybe there is a valid case that I am just not thinking of at the moment.

vgpena:
Hi @jcfilben , I see the point you're making, and the primary use case definitely seems to be using a function instead of a `React.ReactNode`.

The actual issue I'm trying to solve is about backwards compatibility. With implicit React children, `React.ReactNode` is allowed to be passed in. The React 18 PR in Grommet removes support for `React.ReactNode` as a child of the Checkbox. From the perspective of an application that consumes Grommet and uses TypeScript, this is a breaking API change. In our case since we're using React's `PropsWithChildren` type ([link](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L800)), upgrading past v2.23 of Grommet requires code changes on our end which is normally only required for major-version upgrades.

The official recommendation of how to add explicit child types is [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210#issue-1016584444)  (this ignores functions but in some testing locally it seems like `React.ReactElement` allows for `() => React.ReactElement` as well). 

Within Grommet, looking at the `children` type in RadioButton [link](https://github.com/grommet/grommet/pull/6066/files#diff-cea3da7744bc08d052d75b4f297b3012f87451d639e243123e190dd881c2b4e2R7) is `React.ReactNode | Function`, which made me think that Checkbox should be the same and the lack of `React.ReactNode` was a transcription error.

Hope this clears things up!

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

PR #6650, opened 2023-02-23
Changed Distribution to handle undefined values better

ericsoderberghp:
#### What does this PR do?

Changed Distribution to handle undefined values better.

Found an issue using grommet-designer when the first value is undefined and there are subsequent values.

in the code

customized a story by hand to verify

???

Cannot seem to use jest as the scenario fails type checking.

no

yes

backwards compatible


taysea (approved):
Looks good.

PR #6656, opened 2023-02-28
Added button busy animation

jcfilben:
#### What does this PR do?
Adds a `busy` prop to button to support an animation while button is loading or waiting for something. The animation also indicates when the button action succeeds.

Added a story

Planning to go back in and add a jest test

https://github.com/grommet/hpe-design-system/issues/3090
closes https://github.com/grommet/hpe-design-system/issues/3174
Yes

Yes

backwards compatible


src/js/components/Button/stories/Busy.js
```diff
@@ -0,0 +1,36 @@
...
+export const Busy = () => {
+  const [busy, setBusy] = useState();
+
+  return (
+    <Box align="center" pad="medium">
+      <Button
+        primary
+        busy={{
+          state: busy,
+          messages: {
```
taysea, line 13: It might be nice for these messages to have a reasonable default string in Button so callers wouldn't have to implement each time

src/js/components/Button/stories/Busy.js
```diff
@@ -0,0 +1,36 @@
...
+
+export const Busy = () => {
+  const [busy, setBusy] = useState();
+
+  return (
+    <Box align="center" pad="medium">
+      <Button
+        primary
+        busy={{
+          state: busy,
```
taysea, line 12: Would it make more sense for `busy` to be a boolean. Then the "success" state would automatically rendered for "X" amount of time when busy changes from true to false?

taysea: Pausing to think about what happens if a button is busy, then the process fails..

src/js/components/Button/BusyAnimation.js
```diff
@@ -0,0 +1,68 @@
...
+
+export const EllipsisAnimation = ({ color }) => (
+  <Box
+    style={{ position: 'absolute' }}
+    fill
+    alignContent="center"
+    justify="center"
+  >
+    <Box alignSelf="center" direction="row" gap="small">
+      <Dot color={color} delay="-0.32s" />
```
taysea, line 37: Maybe a comment talking about how these delays relate to the overall animation length/effect?

src/js/components/Button/BusyAnimation.js
```diff
@@ -0,0 +1,68 @@
...
+    opacity: 1;
+    transform: scale(1.15);
+  }
+  30% { transform: scale(.9); }
+  45% { transform: scale(1.05); }
+  55% { transform: scale(1); }
+  100% { transform: scale(1); }  
+`;
+
+export const GrowCheckmark = styled(FormCheckmark)`
```
taysea, line 59: We've been talking about leaning away from use of "Form" icons --> wondering if we should use the normal "Checkmark" here. @ericsoderberghp

src/js/components/Button/stories/Busy.js
taysea: I see the default strings now -- you can ignore this comment. Wondering if we should have a story example without custom message overrides to show you don't have to override?

taysea: And if that's the case can the API for that version be simplified to:

`<Button busy={busy} />` ?

src/js/components/Button/BusyAnimation.js
jcfilben: I tried using the regular 'Checkmark' icon but the sizing looks off. Here is the default size:
Screen Shot 2023-03-03 at 9 38 22 AM

And here is the small size:
Screen Shot 2023-03-03 at 9 45 06 AM

This is what is looks like with FormCheckmark:
Screen Shot 2023-03-03 at 9 46 26 AM

ericsoderberghp: I think we should ensure it looks reasonable with the grommet theme as well.

For the HPE theme, this will be resolved with the icon sizing work Taylor is doing.

src/js/components/Button/stories/Busy.js
jcfilben: Based on our discussion yesterday evening I changed `busy` to be a boolean. I changed `messages` to no longer be nested within busy and instead just a prop on Button. I also added a `success` prop that is a boolean.

src/js/components/Button/stories/Busy.js
jcfilben: I removed `messages` from the story example

src/js/components/Button/Button.js
```diff
@@ -399,6 +433,31 @@ const Button = forwardRef(
...
+            (hover && getIconColor(themePaths.hover, theme)) ||
+            getIconColor(themePaths.base, theme, color, kind);
+        }
+      } else if (primary) {
+        animationColor =
+          theme.global.colors.text[isDarkBackground() ? 'dark' : 'light'];
+      }
+
+      contents = (
+        <Box style={{ position: 'relative' }}>
```
taysea, line 451: Add comment about why style tag and position is necessary

src/js/components/Button/stories/Busy.js
```diff
@@ -0,0 +1,33 @@
...
+        primary
+        busy={busy}
+        success={success}
+        label="Button Busy"
+        onClick={() => {
+          if (success) {
+            setSuccess(false);
+          } else {
+            setBusy(true);
+            setTimeout(() => {
```
taysea, line 20: Do we need a clearTimeout somewhere?

jcfilben: It sounds like when you use `setInterval` you need to clear it but if you just use `setTimeout` it is not necessary to clear it. Related stackoverflow thread: https://stackoverflow.com/questions/7391567/when-using-settimeout-do-you-have-to-cleartimeout

src/js/components/Button/BusyAnimation.js
jcfilben: Here are the different options in the grommet theme:

Checkmark medium size:
Screen Shot 2023-03-06 at 12 36 23 PM

Checkmark small size:
Screen Shot 2023-03-06 at 12 38 10 PM


FormCheckmark:
Screen Shot 2023-03-06 at 12 35 25 PM

src/js/components/Button/Button.js
jcfilben: added comment

src/js/components/Button/stories/Busy.js
taysea: Great, thanks for clarifying!

src/js/components/Button/BusyAnimation.js
jcfilben: I tested with the grommet-theme-hpe next-stable and grommet-icons stable. Here is the `Checkmark` icon with the default size:

Screen Shot 2023-03-08 at 4 30 19 PM
Screen Shot 2023-03-08 at 4 30 12 PM

This feels reasonable to me

src/js/components/Button/Button.js
```diff
@@ -402,6 +438,42 @@ const Button = forwardRef(
...
+        }
+      } else if (primary) {
+        animationColor =
+          theme.global.colors.text[isDarkBackground() ? 'dark' : 'light'];
+      }
+
+      contents = (
+        // position relative is necessary to have the animation
+        // display over the button content
+        <Box style={{ position: 'relative' }}>
```
ericsoderberghp, line 458:
```suggestion
        <Box style={{ position: 'relative' }} flex={false}>
```

ericsoderberghp, line 458: Maybe do `const RelativeBox = styled(Box)``position: relative``;` rather than inline style?

jcfilben: Fixed in latest commit

jcfilben: Done

src/js/components/Button/Button.js
```diff
@@ -216,16 +250,18 @@ const Button = forwardRef(
...
       (event) => {
-        sendAnalytics({
-          type: 'buttonClick',
-          element: findButtonParent(event.target),
-          event,
-          href,
-          label: typeof label === 'string' ? label : undefined,
-        });
-        if (onClickProp) onClickProp(event);
+        if (!busy && !success) {
```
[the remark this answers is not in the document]
jcfilben: Yes, done in latest commit

src/js/components/Button/Button.js
```diff
@@ -203,6 +214,29 @@ const Button = forwardRef(
     const theme = useContext(ThemeContext) || defaultProps.theme;
     const [focus, setFocus] = useState();
     const [hover, setHover] = useState(false);
+    const announce = useContext(AnnounceContext);
+    const { format } = useContext(MessageContext);
+
+    if (busy && success) {
+      console.warn('Button can not have both busy and success set to true.');
```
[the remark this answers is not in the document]
jcfilben: Done

src/js/components/Button/Button.js
```diff
@@ -402,6 +436,46 @@ const Button = forwardRef(
...
+          animationColor =
+            (hover && getIconColor(themePaths.hover, theme)) ||
+            getIconColor(themePaths.base, theme, color, kind);
+        }
+      } else if (primary) {
+        animationColor =
+          theme.global.colors.text[isDarkBackground() ? 'dark' : 'light'];
+      }
+
+      const RelativeBox = styled(Box)`
```
ericsoderberghp, line 453: This should be outside of the component, perhaps around line 39?

src/js/components/Button/stories/Busy.js
```diff
@@ -0,0 +1,32 @@
...
+          }, 2000);
+          setTimeout(() => {
+            setSuccess(false);
+          }, 4000);
+        }}
+      />
+    </Box>
+  );
+};
+
```
ericsoderberghp, line 29: Maybe skip chromatic for this, since it doesn't show anything unique statically?

src/js/components/Button/Button.js
jcfilben: Fixed

taysea (approved):
LGTM!

PR #6686, opened 2023-03-16
Add pad property to Button

### LINKED ISSUE #6506: Add `pad` prop to Button
Add a `pad` prop on Button to allow adjustments to a Button's padding. See related conversation https://github.com/grommet/grommet/issues/2595#issuecomment-1322780890

taysea:
#### What does this PR do?

Adds `pad` to Button which accepts all pad types supported by `pad` on Box. Does not apply if Button is `plain` in alignment with docs.

Snapshot changes related to styled components class name generation, but no values changed.

src/js/components/Button/Button.js

Closes #6506 

Yes.

Yes. Add `pad` property to Button.

Backwards compatible.


src/js/components/Button/Button.js
```diff
@@ -475,7 +477,7 @@ const Button = forwardRef(
           }}
           onMouseOver={onMouseOverButton}
           onMouseOut={onMouseOutButton}
-          pad={!plain}
+          pad={pad || !plain}
```
ericsoderberghp, line 480: Why do we need the `!plain`, given that the Styled button code always checks the `plain` property?

ericsoderberghp (approved):

PR #6687, opened 2023-03-16
Fixed Type Error in List Component 

Prakhargoel8c:
#### What does this PR do?  
Fixed Type Error in List Component 

src/js/components/List/index.d.ts

Manually tested

Check type error when using children in List component  

No

Yes

Backwards compatible.


jcfilben (commented):
It looks like `ulProps` also defined children. We should omit the children type definition from `ulProps` and just have it defined in ListProps.

jcfilben (approved):
Thank you for fixing this!

ericsoderberghp (approved):

Issue #6678, opened 2023-03-10
NameValueList - `layout='grid'` not working in Safari


jcfilben:
### Expected Behavior
Safari should match the behavior that chrome has for this story.

### Actual Behavior
Content is difficult to read an is overlapping other content.

### URL, screen shot, or Codepen exhibiting the issue
https://storybook.grommet.io/?path=/story/visualizations-namevaluelist-layout--layout

<img width="659" alt="Screen Shot 2023-03-10 at 11 10 39 AM" src="https://user-images.githubusercontent.com/54560994/224392154-9dca6b6a-9374-4b69-bc90-6c45a501d86f.png">

### Steps to Reproduce
Go to the NameValueList layout story in Safari

### Your Environment

- Grommet version:
- Browser Name and version: Safari
- Operating System and version (desktop or mobile):

taysea:
Closed by #6691

PR #6691, opened 2023-03-17
Fix NameValueList grid layout in Safari

buberdds:
Tested in three browsers. Not sure how users override styles etc so It's hard to tell if any regression can occur.

#### What does this PR do?
Fix NameValueList rendering in Safari

NameValueList.js

Manual using Safari/Chrome/FF

Check story in Safari

https://github.com/grommet/grommet/issues/6678

yes

dunno


src/js/components/NameValueList/NameValueList.js
```diff
@@ -57,6 +57,7 @@ const NameValueList = forwardRef(
           ref={ref}
           columns={columns}
           gap={gap}
+          height="auto"
           fill={layout === 'grid'}
```
jcfilben, line 61:
```suggestion
```
I'm wondering if instead of adding `height='auto'` we could remove `fill={layout === 'grid'}`?

buberdds: good point,`height` is not needed as we can control height via fill prop. I think we have two options here

1) remove `fill={layout === 'grid'}` completely (we will remove height and width)
2) change fill to prop to `fill={layout === 'grid' && 'horizontal'}` (we will keep width: 100%)

Does grid width depend on the context and layout? Does width 100% is necessary or redundant? no idea :)
Let me know which option you prefer, thanks!

jcfilben: From what I am seeing I don't think we need to have `width: 100%` so I think removing `fill` completely should be okay. Appreciate the thoughts and different options proposed though!

jcfilben (approved):
Looks good!

ericsoderberghp (approved):

PR #6692, opened 2023-03-17
Improve badge dimension calculations and add theme.button.badge.align

taysea:
#### What does this PR do?

This PR fixes some issues with how badge was calculating its dimensions. Because of the use of `useLayoutEffect`, instead of handling height/width of badge with `useState`, we apply the style directly to the DOM.

Additionally, `theme.button.badge.align` was added. If the value is specified as `container` then the badge will always offset from the button container.

Snapshot changes are due to the offset/width/height being applied directly to the DOM as opposed to in the CSS styles.

src/js/components/Button/Badge.js

Local storybook with HPE next theme, grommet theme

Yes. Added `theme.button.badge.align`.

Yes.


src/js/components/Button/Badge.js
```diff
@@ -24,47 +25,55 @@ export const Badge = ({ children, content }) => {
...
     // width since badge offset is reliant on its dimensions.
     const onResize = () => {
-      if (contentRef && contentRef.current) {
-        if (
-          typeof content === 'number' ||
-          (typeof content === 'object' && content.value)
-        ) {
-          const { height: contentHeight, width: contentWidth } =
-            contentRef.current.getBoundingClientRect();
+      if (containerRef && containerRef.current) {
```
ericsoderberghp, line 35:
```suggestion
      if (containerRef?.current) {
```

src/js/components/Button/Badge.js
```diff
@@ -24,47 +25,55 @@ export const Badge = ({ children, content }) => {
...
+        containerRef.current.style.minHeight = '';
+        containerRef.current.style.minWidth = '';
+
+        if (contentRef && contentRef.current) {
```
ericsoderberghp, line 39:
```suggestion
        if (contentRef?.current) {
```

ericsoderberghp (commented):
very small suggestions

ericsoderberghp (approved):

PR #6693, opened 2023-03-17
Fix issue with DateInput

coltonw:
#### What does this PR do?
There is an issue in an inline ranged DateInput where if you type a date and then click that date in the calendar to deselect it, the component simply breaks.  It will return nonsensical values to its onChange handler and will display NaN in the date text input.  It is not recoverable from this state.

This PR fixes that issue.

Testing in storybook I suppose.

It has been tested using a jest test and manually tested in storybook.

You can repro by going to this story: https://storybook.grommet.io/?path=/story/input-dateinput-format-inline--format-inline and then type a date and then click that date in the calendar to deselect it.

Screen Shot 2023-03-17 at 2 28 06 PM

No

Maybe?

Backwards compatible.


jcfilben (approved):
Thank you for fixing this and adding a test! This looks good to me

PR #6704, opened 2023-03-24
Data Array Enhancement

### LINKED ISSUE #6701: Data Enhancement with Array
### Expected Behavior
On properties that contain array values, `Data` should:
1. recognize array
2. do matching for values **inside** the array, not try to match the filter's value to the array object directly

image
Ex: Filtering on the "Categories" object above, selecting "Machine learning" OR "Developer tools" should cause this object to be returned when filtering.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

JCoder121:
[Complex Story Deploy Preview](https://deploy-preview-6704--sad-tereshkova-173d07.netlify.app/?path=/story/data-data-complex--complex)

#### What does this PR do?
Add functionality to Data's `filter()` for when a property is a `list`, rather than a `string`, `bool`, or nested object. Allow the user to select filter values that filter upon list objects.

src/js/components/Data/filter.js
src/js/components/Data/stories/Complex.js

Manual and visual testing to ensure filtering upon list elements works as intended.

Check that existing filtering works as expected in addition to the new list filtering capabilities.

Closes #6701

Yes - update after initial code review to include example case.


src/js/components/Data/filter.js
```diff
@@ -60,10 +60,19 @@ export const filter = (data, view, properties) => {
           );
 
         // options case
-        if (Array.isArray(filterValue)) return !filterValue.includes(value);
+        if (Array.isArray(filterValue)) {
+          // value is array case
+          if (Array.isArray(value))
+            // from https://stackoverflow.com/a/39893636
```
taysea, line 66:
```suggestion
```
It is fine to use the `.some` without referencing the stack overflow source since it's a generic method.

src/js/components/Data/filter.js
```diff
@@ -60,10 +60,19 @@ export const filter = (data, view, properties) => {
           );
 
         // options case
-        if (Array.isArray(filterValue)) return !filterValue.includes(value);
+        if (Array.isArray(filterValue)) {
```
taysea, line 63: Can we simplify format to:
```
if (Array.isArray(filterValue)) return Array.isArray(value) ? !value.some((v) => filterValue.includes(v)) : !filterValue.includes(value);
```

yarn.lock
```diff
@@ -1280,16 +1280,16 @@
   integrity sha512-OWORNpfjMsSSUBVrRBVGECkhWcULOAJz9ZW8uK9qgxD+87M7jHRcvh/A96XXNhXTLmKcoYSQtBEX7lHMO7YRwg==
 
 "@eslint-community/eslint-utils@^4.2.0":
-  version "4.3.0"
-  resolved "https://registry.yarnpkg.com/@eslint-community/eslint-utils/-/eslint-utils-4.3.0.tgz#a556790523a351b4e47e9d385f47265eaaf9780a"
-  integrity sha512-v3oplH6FYCULtFuCeqyuTd9D2WKO937Dxdq+GmHOLL72TTRriLxz2VLlNfkZRsvj6PKnOPAtuT6dwrs/pA5DvA==
+  version "4.4.0"
```
taysea, line 1283: For PRs on grommet, we don't want to adjust the yarn.lock. Can you undo the changes here?

taysea (commented):
Functionality looks good! Couple small cleanup comments.

src/js/components/Data/filter.js
JCoder121: Originally included it because it was the exact code needed to directly solve the issue.

src/js/components/Data/stories/Complex.js
```diff
@@ -10,19 +10,32 @@ const data = [
...
+    colors: ['Red', 'Yellow', 'Black'],
   },
 ];
 
 const properties = {
   name: { label: 'Name', search: true },
   'location.city': { label: 'City' },
   'economy.GDP': { label: 'GDP' },
+  colors: {
+    label: 'Flag Colors',
```
taysea, line 36:
```suggestion
    label: 'Flag colors',
```
sentence casing to align with DataTable column header/best practices.

ericsoderberghp:
Screenshot 2023-03-24 at 12 21 52 PM

Something looks off with the vertical alignment here. I don't think this is related to the changes in this pull request but wondering if you can take a quick look.

ericsoderberghp:
Can you enhance the story so that searching matches on the color names, in addition to the filtering?

ericsoderberghp:
I think we should consider using `verticalAlign` for the columns to align them to the top instead of centering the cell contents.

JCoder121:
> Screenshot 2023-03-24 at 12 21 52 PM
> 
> Something looks off with the vertical alignment here. I don't think this is related to the changes in this pull request but wondering if you can take a quick look.

image
This screenshot is from the Views story [here](https://storybook.grommet.io/?path=/story/data-data-views--views). It's looking like whenever the `Clear Filters` button is inserted in the toolbar, it isn't centered inside its parent Box (maybe aligned to start?). Additionally, when tabbing through the controls, the Filter button's focus indicator box seems to be very slightly larger than the DataSearch component (and in this case the Select component as well).

ericsoderberghp (approved):

PR #6725, opened 2023-04-03
Data Array Enhancement

JCoder121:
#### What does this PR do?
follow up from https://github.com/grommet/grommet/pull/6704, fixes filtering when filter value is an object with {label: __, value: __}

src/js/components/Data/filter.js
src/js/components/Data/stories/Complex.js

manual + visual testing

ensure that filtering works as intended


src/js/components/Data/filter.js
```diff
@@ -60,10 +60,16 @@ export const filter = (data, view, properties) => {
...
         // options case
-        if (Array.isArray(filterValue))
-          return Array.isArray(value)
-            ? !value.some((v) => filterValue.includes(v))
-            : !filterValue.includes(value);
+        if (Array.isArray(filterValue)) {
+          if (Array.isArray(value)) {
+            return !filterValue.some((f) =>
+              typeof f === 'object'
+                ? value.includes(f.value) // from {label: ___, value: ___}
```
taysea, line 67: It looks like previous we were doing a "filterValue.includes(value)" pattern and now we're doing "value.includes(filter)" -- can you explain why we made this switch? Was that a bug in the initial implementation?

src/js/components/Data/stories/Complex.js
```diff
@@ -76,7 +85,9 @@ export const Complex = () => (
       status="info"
       message="Data is in 'beta'. The API surface is subject to change."
     />
-    <Data data={data} properties={properties} toolbar>
+    {/* <Data data={data} properties={properties} toolbar> */}
```
taysea, line 88: Cleanup comment - Remove commented line

src/js/components/Data/filter.js
JCoder121: Connected over call - `value.includes(filter)` is used in this case because we iterate over `filterValue` and we use `typeof f` to determine if each `f` is in the form of an object `{label: ___, value: __}`. If so, we want to check whether or not `value` includes `f.value`. Previously, there was no check for `typeof f`, which resulted in incorrect behavior when each `value` was a `{label: __, value: __}` object.

taysea (approved):
Nice work!

src/js/components/Data/stories/Complex.js
```diff
@@ -76,7 +85,8 @@ export const Complex = () => (
       status="info"
       message="Data is in 'beta'. The API surface is subject to change."
     />
-    <Data data={data} properties={properties} toolbar>
+    <Data data={data} properties={properties} updateOn="change">
+      <DataFilters layer />
```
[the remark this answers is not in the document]
JCoder121: image
I realized after I just tried testing the original layout -- in fullscreen interaction, there is extremely minimal scroll space for the Select options, and `<DataFilters layer />` somewhat solves the problem. I think this might be an issue unrelated to this PR?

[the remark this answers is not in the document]
taysea: I'm fine adjusting this story to the DataFilters layer, but I'd probably recommend we don't add "updateOn=change" in that case.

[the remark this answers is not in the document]
JCoder121: Thanks for clarifying! Will make that final edit.

JCoder121:
4/14/23 status:

Final stages of review, waiting upon final review to wrap things up and should be merged today or early on next week.

src/js/components/Data/filter.js
jcfilben, line 67: Can we change this to be `f?.value`? Just so if the object doesn't have a value it doesn't cause the app to crash
```suggestion
                ? value.includes(f?.value) // from {label: ___, value: ___}
```

jcfilben (commented):
Minor comment, otherwise looks good

jcfilben (approved):
Looks good!

taysea (approved):

ericsoderberghp (approved):

Issue #6228, opened 2022-07-11
Storybook - copy code shouldn't include line numbers


ShimiSun:
### Expected Behavior
When copying code from storybook, it shouldn't include the line numbers of the code. This is a regression.

### Actual Behavior
Copy code include the line numbers.

### URL, screen shot, or Codepen exhibiting the issue
Copy code from https://storybook.grommet.io/?path=/story/visualizations-list-onclickitem--on-click-item-list
And receive this when pasting:

26  const [clicked, setClicked] = React.useState();
27  const [show, setShow] = React.useState();
28
29  return (
30    <Box align="center" pad="large" gap="large">
31      <List
32        data={data.slice(0, 10)}
33        onClickItem={(event) => {
34          setShow(true);
35          setClicked(event.item);
36        }}
37      />
38
39      {show && (
40        <Layer
41          position="center"
42          onEsc={() => setShow(false)}
43          onClickOutside={() => setShow(false)}
44        >
45          <Box margin="medium">
46            <Text>{clicked && JSON.stringify(clicked, null, 2)}</Text>
47            <Button
48              margin={{ top: 'medium' }}
49              label="close"
50              onClick={() => setShow(false)}
51            />
52          </Box>
53        </Layer>
54      )}
55    </Box>
56  );
57};

### Steps to Reproduce

1. Go to any storybook code example, e.g. https://storybook.grommet.io/?path=/story/visualizations-list-onclickitem--on-click-item-list
2. Copy code from storybook 
3. Paste code on your editor

### Your Environment
Mac + Chrome

- Grommet version: Recents
- Browser Name and version: N/A
- Operating System and version (desktop or mobile): N/A

taysea:
As a workaround (not a solve, but takes some of the tediousness out), you can put your cursor on a certain line, then hold alt + shift, and click on a new line -- this will select all the lines in between:

https://user-images.githubusercontent.com/12522275/181321740-eb318250-99e9-416a-88b2-072ec2cdbb6c.mov

jcfilben:
Should be fixed by the storybook 7 upgrade https://github.com/grommet/grommet/pull/6726

PR #6726, opened 2023-04-03
Storybook 7 upgrade

jcfilben:
#### What does this PR do?
Upgrades to use the latest storybook version

https://github.com/grommet/hpe-design-system/issues/3252
Closes https://github.com/grommet/grommet/issues/6228

no
no
backwards compatible


jcfilben:
Marking as a draft while I look into the chromatic changes

jcfilben:
> Marking as a draft while I look into the chromatic changes
Should be fixed now. I just had to update where storybook was grabbing the arguments we are passing to the grommet wrapper

taysea (approved):

halocline (commented):
This is looking good, however I am noticing a couple things:

1) When retrieving the root index (e.g. `https://[domain]/`) I land on the first story under Accordion. Do we want the to resolve to "All"?
2) When running `build-storybook` I am receiving a handful of warnings in my terminal. They seem related to other storybook dependencies, but am curious if you had run into / investigated these warnings?

jcfilben:
> 

> This is looking good, however I am noticing a couple things:
> 
> 1. When retrieving the root index (e.g. `https://[domain]/`) I land on the first story under Accordion. Do we want the to resolve to "All"?
> 2. When running `build-storybook` I am receiving a handful of warnings in my terminal. They seem related to other storybook dependencies, but am curious if you had run into / investigated these warnings?

1. This should be fixed now, 'All' should be the story we land on.
2. I did run into these warnings. Storybook 7 dropped support for webpack 4, I think these warnings are coming from storybook dependencies and addons that still have some cleanup to do from moving to webpack 5. I didn't see anything that looked too concerning in the warnings, but if we want we can always wait a little longer to upgrade and give more time for storybook 7 to become a little more solid.

halocline:
> 2. I did run into these warnings. Storybook 7 dropped support for webpack 4, I think these warnings are coming from storybook dependencies and addons that still have some cleanup to do from moving to webpack 5. I didn't see anything that looked too concerning in the warnings, but if we want we can always wait a little longer to upgrade and give more time for storybook 7 to become a little more solid.

I agree with your assessment. I feel comfortable proceeding.

halocline (approved):

Issue #6737, opened 2023-04-06
SelectMultiple when using object values and expanded handles click wrongly


urubatan:
There is an error in select multiple when using object values in conjunction with showSelectedInline, when the inline item is clicked instead of de-selecting that item it ends up being the only item selected

### Expected Behavior

Unselect an inline item when clicked

### Actual Behavior

When clicking an inline item it ends up being the only item selected

### URL, screen shot, or Codepen exhibiting the issue

The error happens all the time, it is easy to check it in the storybook 
(https://storybook.grommet.io/?path=/story/input-selectmultiple-object-options--object-options)

the error is in the line (https://github.com/grommet/grommet/blob/bcf727d1ea1734f64dab91173d033a1c9c2bee2c/src/js/components/SelectMultiple/SelectMultipleValue.js#L94-L96)

the object comparison should be !== like the string value comparison, but it is using ===

jcfilben:
Fixed in #6738

PR #6738, opened 2023-04-06
fix for SelectMultiple when using object values inline clicks

urubatan:
When clicking on an inline option for a SelectMultiple that uses objects as options the behavior should be the same as when the options are strings.
This PR fixes the comparison for object values to be the same as for string values

#### What does this PR do?
changes one line in SelectMultipleValue to make sure the inline option clicked is deselected instead of being the only one selected for object options

you can check the current error in storybook 

https://storybook.grommet.io/?path=/story/input-selectmultiple-object-options--object-options

when clicking in any inline option that is the only one that is still selected, instead of deselecting only that option

I tested manually on my machine

all existing tests are still working

I didn't add more tests to confirm the bug because my javascript skills aren't good enough and I didn't know how to do that

opening the storybook for this PR you can verify the correct behavior 

This PR fixes the issue #6737 

No, after this PR the docs are correct for all cases

no need

it just fixes a bug


jcfilben (commented):
Thanks for fixing this! Would you be willing to add a unit test for this?

urubatan:
Sure, I'll check how to and will push the tests later today

Em qui., 6 de abr. de 2023 17:37, Jessica Filben ***@***.***>
escreveu:

jcfilben:
> Sure, I'll check how to and will push the tests later today

Thanks! Feel free to reach out if you run into any issues

urubatan:
@jcfilben just pushed the test to validate the fix

jcfilben (approved):
Nice work!

taysea (approved):

PR #6740, opened 2023-04-07
Add noForm to DataFilter

JCoder121:
#### What does this PR do?

Similar to code pattern found in DataView and DataSearch, add `noForm` from `FormContext` to DataFilter. 

anywhere

Visual + manual testing
Added a unit test

Manually test for screen reader experience. Ensure that functionality works as intended

Initially this was a PR to add an `inputOnly` prop to DataFilter that allowed user to manually specify whether or not to wrap DataFilter `content` in a FormField. However, the PR was changed to use `noForm` instead, which is a context provided by other Data components that consume `DataFilter` that determines whether or not the content returned is wrapped in a FormField, which matches the design pattern with other Data components.

Related to https://github.com/grommet/hpe-design-system/pull/3225. This issue came up as I was progressing through the PR.

no

no

backwards compatible


ericsoderberghp:
Can you take a look at how `noForm` is used within DataView.js to see if that pattern would work?

JCoder121:
> Can you take a look at how `noForm` is used within DataView.js to see if that pattern would work?

Taking a look

src/js/components/DataFilter/DataFilter.js
```diff
@@ -121,9 +129,11 @@ export const DataFilter = ({
     }
   }
 
-  return inputOnly ? (
-    content
-  ) : (
+  if (noForm) content = <DataForm footer={false}>{content}</DataForm>;
```
ericsoderberghp, line 132: To align with DataView, we'd wrap with DataForm instead of FormField when `noForm`.

ericsoderberghp (approved):

PR #6746, opened 2023-04-13
Add Default Tooltip to DataTableColumns

### LINKED ISSUE #6743: DataTableColumns - Allow caller to specify tooltip.
**Expected**
Allow caller to specify tooltip. Consider having as a default similar to  the filter button.

**Actual**
Column customization button icon is not immediately recognizable. A tooltip would be helpful for users to identify.

**Related issues**

Closes https://github.com/grommet/hpe-design-system/issues/3260

JCoder121:
* USE SENTENCE CASE FOR MANAGE COLUMNS

#### What does this PR do?
Adds default tooltip using `messages` and contexts defined in DataTableColumns.js. 
Updated the DataTableColumns `Simple` story - Adds `DataSearch` below the Toolbar, which matches the pattern in the [Data/DataTable story](https://storybook.grommet.io/?path=/story/data-data-datatable--table), allowing for the DataTableColumns tip to not overlap with content in the DataTable.

src/js/components/DataTableColumns/DataTableColumns.js
src/js/components/DataTableColumns/stories/Simple.js

Manual + visual testing

Ensure that DataTable functionality remains the same. Also, visually test for if the DataTableColumns tip overlaps with any content.

Following up, the current message is "Open column selector". Guidance on the Design Site [here](https://design-system.hpe.design/templates/datatable-customization#an-action-to-manage-column-visibility-and-order-controls) indicates that the message should be "Manage Columns". 

Closes #6743. 

image
DataSearch components creates the space for the tip to occupy, otherwise, the tip overflows into the table content.

no

yes

backwards compatible


src/js/components/DataTableColumns/DataTableColumns.js
```diff
@@ -170,6 +170,7 @@ export const DataTableColumns = ({ drop, options, ...rest }) => {
       })}
       kind="toolbar"
       icon={<Splits />}
+      tip="Manage Columns"
```
ericsoderberghp, line 173: I think we should consider doing this via `messages` to follow how we localize other messages.

ericsoderberghp: Maybe look at how `tip` is done with DataFilters.js

JCoder121: Taking a look!

ericsoderberghp (approved):

PR #6715, opened 2023-03-29
Changed Calendar to honor reference when set

ericsoderberghp:
#### What does this PR do?

Changed Calendar to honor reference when set

#### Where should the reviewer start?

Calendar.js

#### What testing has been done on this PR?

Tried assorted stories.

#### How should this be manually tested?

Try various Calendar and DateInput stories. I'm sure I haven't covered everything from a testing perspective.

#### Do Jest tests follow these best practices?

no test changes

#### What are the relevant issues?

#6609 

#### Do the grommet docs need to be updated?

no

#### Should this PR be mentioned in the release notes?

yes

#### Is this change backwards compatible or is it a breaking change?

backwards compatible

PR #6766, opened 2023-04-24
Revert "Changed Calendar to honor reference when set"

jcfilben:
Reverts grommet/grommet#6715

### DIFF
--- a/src/js/components/Calendar/stories/Dual.js
+++ b/src/js/components/Calendar/stories/Dual.js
@@ -34,11 +34,9 @@ export const Dual = () => {
         onReference={(reference) => {
           const refDate = new Date(reference);
           const nextDate = new Date(refDate);
-          console.log(refDate);
-          console.log(nextDate);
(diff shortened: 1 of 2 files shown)


src/js/components/Calendar/stories/Dual.js
```diff
@@ -42,9 +34,11 @@ export const Dual = () => {
         onReference={(reference) => {
           const refDate = new Date(reference);
           const nextDate = new Date(refDate);
+          console.log(refDate);
+          console.log(nextDate);
           nextDate.setMonth(refDate.getMonth() + 1, 1);
-          setReference1(refDate.toISOString());
-          setReference2(nextDate.toISOString());
+          setReference1(refDate.toString());
```
ericsoderberghp, line 40: This seems incorrect to me. I think `toISOString` would be better here.

jcfilben: fixed in latest commit

ericsoderberghp (approved):

PR #6775, opened 2023-05-01
Changed DataContext to fix an issue with using the default value

ericsoderberghp:
#### What does this PR do?

Changed DataContext to fix an issue with using the default value

DataContext.js

added a unit test
tested with grommet-designer

followed existing tests

detected by grommet designer use using DataSearch outside of Data.

no

yes

backwards compatible


halocline (commented):
This looks good.

There also exists a scenario where `<Data />` may have "no data" causing an error at https://github.com/grommet/grommet/blob/a577633f6e1a32aa6af91b7100368aa94b73b28d/src/js/components/Data/filter.js#L28.

Scenario: In Grommet Designer --> add Data --> edit the property 'data' so that the input is something other than an array (e.g. undefined, valid JSON but not an array,)

Screenshot 2023-05-01 at 4 19 58 PM

PR #6777, opened 2023-05-01
Specify resource classes for CircleCI jobs

jcfilben:
#### What does this PR do?
Adds a resource class size to jobs. See https://circleci.com/docs/configuration-reference/#resourceclass for more detail.

no
no
backwards compatible


jcfilben:
Duration is ~30% less than jobs from the past 7 days
Screen Shot 2023-05-01 at 2 06 31 PM

jcfilben:
> How did you determine the sizes to use? Can you add some comments about that? I'm wondering why test is "small".

I used the run time of the jobs to determine the resource classes to assign them to. For example, the lint job was taking around 45 seconds and would finish much faster than the other jobs. So I assigned lint to a small resource class so that it would use less resources and possibly run a little longer. For jobs that were taking longer such as jest and chromatic I assigned these to a large resource class.

The test job has a small resource class because this step is running our end to end tests. Right now we only have 3 end to end tests, so this step doesn't take very long. As we add more end to end tests we may want to adjust the resource class accordingly.

jcfilben:
If it would be helpful I can add a comment in the config file about this too

ericsoderberghp (approved):

PR #6795, opened 2023-05-09
Changed DataFilter to handle properties with zero or one values

ericsoderberghp:
#### What does this PR do?

Changed DataFilter to handle properties with zero or one values.

Of note, there is a change to the unit test but not a snapshot change. This is because the "blank" and "zero" properties are not rendered by DataFilter. Before the change to DataFilter.js here, the snapshots would have changed to show awkward filter controls that didn't do anything.

DataFilter.js

Manually used a data set provided by @samuelgoff.
Enhanced a unit test to include these cases.

Try with more variations of data.

no change to existing test patterns

See notes above.

no

yes

backwards compatible, since filtering on properties with less than two possible values is meaningless.

### DIFF
--- a/src/js/components/DataFilter/DataFilter.js
+++ b/src/js/components/DataFilter/DataFilter.js
@@ -62,8 +62,10 @@ export const DataFilter = ({
 
     // generate options from all values for property
     const uniqueValues = generateOptions(unfilteredData || data, property);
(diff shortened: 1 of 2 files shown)


MikeKingdom (approved):
Looks good. Darn 0 = falsey :)

PR #6632, opened 2023-02-09
Change logic for restrict scroll

britt6612:
#### What does this PR do?
This changes the logic of when to `restrictScroll` 
#### Where should the reviewer start?
LayerContainer.js
#### What testing has been done on this PR?
Storybook
#### How should this be manually tested?
```
import React from 'react';
import { Box, Grommet, Button, Layer, Header, Heading } from 'grommet';
import { hpe } from 'grommet-theme-hpe';

export const TargetLayer = () => {
  const [open, setOpen] = React.useState();
  const ref = React.useRef();
  const onOpen = () => setOpen(true);

  return (
    // Uncomment <Grommet> lines when using outside of storybook
    <Grommet theme={hpe}>
      <Box height="xxlarge">
        <Box
          ref={ref}
          height={{ min: 'small' }}
          gap="medium"
          width="medium"
          background="brand"
        >
          <Button label="Open" onClick={onOpen} />
          {open && (
            <Layer position="center" target={ref.current}>
              <Box
                gap="medium"
                pad="medium"
                width={{ min: 'medium' }}
                flex="grow"
              >
                <Header
                  flex={false}
                  align="start"
                  gap="small"
                  justify="between"
                >
                  <Box>
                    <Heading
                      id="layer-title"
                      level={2}
                      margin="none"
                      size="small"
                    >
                      test
                    </Heading>
                  </Box>
                </Header>
              </Box>
            </Layer>
          )}
        </Box>
      </Box>
    </Grommet>
  );
};

TargetLayer.storyName = 'Target';

TargetLayer.args = {
  full: 'min',
};

export default {
  title: 'Layout/Layer/Target',
};

```
#### Do Jest tests follow these best practices?

- [ ] `screen` is used for querying.
- [ ] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [ ] `userEvent` is used in place of `fireEvent`.
- [ ] `asFragment()` is used for snapshot testing.

#### Any background context you want to provide?
When hitResponsiveBreakpoint is `small` this line 
`!layerTarget || hitResponsiveBreakpoint ? true : undefined` turns true however when we have a `layerTarget` this should not be `true` so instead of OR this should be AND
#### What are the relevant issues?
Scroll issue in DS site
Closes https://github.com/grommet/hpe-design-system/issues/3111
#### Do the grommet docs need to be updated?
no
#### Should this PR be mentioned in the release notes?
yes
[...]

PR #6801, opened 2023-05-15
fix: Layer with modal prop does not restrict scroll

CopyJosh:
Bug introduced in v2.30.0

#### What does this PR do?
Fixes a bug introduced in v2.30.0 where the Layer would not restrict scroll when passed the modal prop.

https://github.com/grommet/grommet/compare/v2.29.1...v2.30.0#

Local testing against our website

This can be manually tested in [storybook](https://storybook.grommet.io/?path=/story/layout-layer-target--target-layer), by toggling on "modal" and adjusting the screen width such that `hitResponsiveBreakpoint` returns false.

This can be manually tested in the docs at https://v2.grommet.io/layer by adding the `modal` prop and observing the window still scrolls.

Up until now the `modal` prop alone would appear to restrict the scroll of the page, but maybe that was not a widespread intention given the code comments. If that is the case this PR could be ignored..

https://github.com/grommet/grommet/pull/6632


taysea (approved):
LGTM.

halocline (approved):

halocline:
Thank you, @CopyJosh. We appreciate the contribution.

jcfilben (approved):

Issue #6798, opened 2023-05-12
Card - only show focusIndicator when using keyboard


jcfilben:
### Expected Behavior
The focusIndicator should only be displayed on Card when the user is navigating with a keyboard.

### Actual Behavior
The focusIndicator shows up when clicking on a card with the mouse.

### URL, screen shot, or Codepen exhibiting the issue
https://storybook.grommet.io/?path=/story/layout-card-custom-themed-clickable--clickable&globals=theme:base

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #6805, opened 2023-05-16
Remove focusIndicator from Box when interacting with a mouse

jcfilben:
#### What does this PR do?
Removes the focus indicator from box when using a mouse.

Closes https://github.com/grommet/grommet/issues/6798
no
yes
backwards compatible


jcfilben:
Holding off on this change for now, it touches a lot of areas and I think we need more time to evaluate the direction.

PR #6815, opened 2023-05-25
Adjust empty search message styling on Select

jcfilben:
#### What does this PR do?
This PR changes the 'no options found' container in Select and SelectMultiple so that it is no longer a disabled Button and is instead just Text within a Box. The Text and Box can be styled with `theme.select.options.container` and `theme.select.options.text`. 

For themes that do not have any padding defined in `theme.select.options.container` and have a `button kind='option'` defined, the padding defined by `button kind='option'` is used on the Box around the 'no options found' text. The options in the Select are styled as `button kind='option'` so this ensures that the options in the select and the 'no options found' text have the same amount of padding applied.

no
yes
backwards compatible

### DIFF
--- a/src/js/components/Select/SelectContainer.js
+++ b/src/js/components/Select/SelectContainer.js
@@ -390,14 +390,14 @@ const SelectContainer = forwardRef(
               theme={theme}
             />
           )}
-          <OptionsContainer
-            role="listbox"
-            tabIndex="-1"
-            ref={optionsRef}
-            aria-multiselectable={multiple}
-            onMouseMove={() => setKeyboardNavigation(false)}
-          >
-            {options.length > 0 ? (
+          {options.length > 0 ? (
+            <OptionsContainer
+              role="listbox"
+              tabIndex="-1"
+              ref={optionsRef}
+              aria-multiselectable={multiple}
+              onMouseMove={() => setKeyboardNavigation(false)}
+            >
               <InfiniteScroll
                 items={options}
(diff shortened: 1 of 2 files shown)


src/js/components/Select/SelectContainer.js
```diff
@@ -390,14 +390,14 @@ const SelectContainer = forwardRef(
...
-          <OptionsContainer
-            role="listbox"
-            tabIndex="-1"
-            ref={optionsRef}
-            aria-multiselectable={multiple}
-            onMouseMove={() => setKeyboardNavigation(false)}
-          >
-            {options.length > 0 ? (
+          {options.length > 0 ? (
+            <OptionsContainer
```
MikeKingdom, line 394: EmptySearchOption is also used in SelectMultipleContainer.js in a similar way. Do you need to pull it out of the container there also?

MikeKingdom (commented):
It looks great and seems to work well. I'm mainly curious about if the SelectMultiple case needs to pull the EmptySearchOption out of the OptionsContainer also.

src/js/components/Select/SelectContainer.js
jcfilben: Good call, I moved the EmptySearchOption outside of the options container in SelectMultipleContainer.js

MikeKingdom (approved):
Looks good!

halocline (approved):

PR #6823, opened 2023-06-02
Add eslint-plugin-grommet

jcfilben:
#### What does this PR do?
Adds the eslint-plugin-grommet package as a dev dependency. For now I disabled some of the rules just until we get a chance to go through and make fixes. There were 409 issues so this will take some time to cleanup.

no
no
backwards compatible


jcfilben:
We should do a quick audit of the rules in eslint-plugin-grommet to make sure they are setup how we would like them to be before adding this plugin to grommet. Closing this PR for now.

Issue #6829, opened 2023-06-05
Select - Having `value` 0 (zero) prevents `valueLabel` function from being called


pascallapradepodform3d:
On the `Select` component, if I provide `0` (zero) as the value, and also provide a `valueLabel` rendering function, the `valueLabel` function is not called, and my value is not rendered as I would like.

### Expected Behavior

If I provide a `valueLabel` to the `Select` component, it should be called for any `value` that I pass to the component.

### Actual Behavior

When I provide a `valueLabel` to the `Select` component, it is not called if the `value` is `0` (which I expect also happens with other falsy values, though I did not test it).

### URL, screen shot, or Codepen exhibiting the issue

Here is minimal reproduction in Codepen: https://codesandbox.io/s/grommet-v2-template-forked-6dn7wd?file=/index.js

It is also reproducible directly on the v2.grommet.io Select component page.

### Steps to Reproduce

1. In the Codepen, notice that the current option is `0`, and the component displays `0`.
2. Notice in the code I have a `valueLabel` function that should display "Value: [the value]".
3. In the rendered component, select a non-0 value.
4. The component changes to display the desired label defined by `valueLabel`.
5. Select 0 again.
6. The label switches back to something other than what is defined by `valueLabel`.

### Your Environment

- Grommet version: 2.32.1 and the version used by v2.grommet.io and the version used by Codepen
- Browser Name and version: Google Chrome 113.0.5672.126 (arm64)
- Operating System and version (desktop or mobile): macOS Ventura 13.4 (arm64)

### My understanding of why this is happening

I looked into the implementation of the Select component here on Github, and I think I have an idea where the issue is happening. I may or may not be correct about this, but hopefully this can help pinpoint and correct the bug.

At lines 235:237 (https://github.com/grommet/grommet/blob/v2.32.2/src/js/components/Select/Select.js#LL235C7-L237C8) we see that an if statement causes a coercion of the user-provided value before `valueLabel` is called:

```javascript
if (valueLabel instanceof Function) {
  if (value) return valueLabel(value);
}
```

As such, a falsy value, such as `0`, will be treated by JS as `false`, and the `valueLabel` function will not be called.

I don't know if this is the intended behavior, blocking all falsy values (in which case maybe it should be documented under the `valueLabel` property, to explain that falsy JS values will prevent the function to be called).

If it is not, then changing the condition to a stricter null and/or undefined check would likely fix the issue while preserving the intended behavior of the nullish guard.

[...]

PR #6833, opened 2023-06-06
Fix issue where valueLabel doesn't show if value is 0

jcfilben:
#### What does this PR do?
Resolves an issue where using `valueLabel` on Select and `value={0}` results in the valueLabel not being displayed correctly.

Tested with the following storybook story and added a jest test.
```javascript
import React from 'react';

import { Box, Select } from 'grommet';

export const Test = () => {
  const [value, setValue] = React.useState(0);
  return (
    <Box fill align="center" justify="start" pad="large" gap="medium">
      <Select
        options={[0, 1, 2]}
        value={value}
        valueLabel={(val) => `Value: ${val}`}
      />
    </Box>
  );
};

export default {
  title: 'Input/Select/Test',
};
```

Closes https://github.com/grommet/grommet/issues/6829

no
yes
backwards compatible

### DIFF
--- a/src/js/components/Select/Select.js
+++ b/src/js/components/Select/Select.js
@@ -233,7 +233,7 @@ const Select = forwardRef(
     // element to show, trumps inputValue
     const selectValue = useMemo(() => {
       if (valueLabel instanceof Function) {
-        if (value || value === 0) return valueLabel(value);
+        if (value || value === 0 || value === false) return valueLabel(value);
       } else if (valueLabel) return valueLabel;
       else if (React.isValidElement(value)) return value; // deprecated
       return undefined;
--- a/src/js/components/Select/Select.js
+++ b/src/js/components/Select/Select.js
@@ -233,7 +233,7 @@ const Select = forwardRef(
     // element to show, trumps inputValue
     const selectValue = useMemo(() => {
       if (valueLabel instanceof Function) {
-        if (value) return valueLabel(value);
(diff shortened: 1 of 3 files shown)


src/js/components/Select/Select.js
```diff
@@ -233,7 +233,7 @@ const Select = forwardRef(
     // element to show, trumps inputValue
     const selectValue = useMemo(() => {
       if (valueLabel instanceof Function) {
-        if (value) return valueLabel(value);
+        if (value !== '' && value !== null) return valueLabel(value);
```
MikeKingdom, line 236: There are other 'falsey' values like `undefined` or `false` that this handled before. Should we be calling valueLabel for `undefined` or `false`? Might be better to change this to:
```javascript
if (value || value === 0) return valueLabel(value)
```

taysea: Not sure if this is a use case, but is it possible that someone might have some options like `[true, false]` ?

jcfilben: ^ good point, we do support having `[true, false]` as options. Maybe we could do something like `if (value || value === 0 || value === false) return valueLabel(value);`? Not sure if there is a cleaner way to write this to get our desired behavior.

jcfilben: Cool, just pushed up that change

taysea (approved):

MikeKingdom (approved):

PR #6841, opened 2023-06-11
Deactivated Tip wrapper with a conditional prop

LegerG:
#### What does this PR do?

Deactivate the Tip when the props `deactivated` is set to `true` 

Tip.js

Manual and automatic tests

With storybook

Yes

Yes

Backward compatible

### DIFF
--- a/src/js/components/Tip/Tip.js
+++ b/src/js/components/Tip/Tip.js
@@ -12,75 +12,77 @@ import { Drop } from '../Drop';
 import { useForwardedRef, useKeyboard } from '../../utils';
 import { TipPropTypes } from './propTypes';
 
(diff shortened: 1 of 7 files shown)


jcfilben:
Hi @LegerG I'm not sure I understand the need for this prop, it seems like you could just conditionally render the tip instead

LegerG:
Hello @jcfilben, yes it's to render the Tips wrapper conditionally. It's to write it more concisely.

jcfilben:
Hey @LegerG appreciate your desire to improve Grommet! In this case I think we should avoid adding props like this. Since the desired behavior can be achieved relatively easily without this property it just ends up polluting the API.

Issue #5961, opened 2022-02-18
Form should run validation when it's set from outside the form


vujevits:
### Expected Behavior

If I have a button which changes the form's content, validation logic won't run.

### Actual Behavior

When `change` is the validation trigger, and I have a button which updates the field to a different value, it won't trigger the validation.

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/s/hopeful-poitras-i32fk?file=/src/Form.js

### Steps to Reproduce
The above form expects numeric input

1. Press the button to add text
2. no validation is applied

narayand16:
The PR for this issue is already merged. What causing the issue to still pop up its head ? ;)

jcfilben:
The pull request that is showing up in this issue wasn't a pull request made on Grommet. The pull request is just on a repo using Grommet and they mentioned this issue in it. I believe this issue is not fixed yet.

PR #6850, opened 2023-06-17
fix: Form validation when it's set from outside the form

g4rry420:
#### What does this PR do?
- Added a console error when validate prop is added to FormField  and name is not provided to it
- Another console error when FormField Children's name doesn't match with the FormField's name, if that happens then validations doesn't work
- Added name prop to FormField's Children, if it is not provided before so that Form validation can trigger when it's set from outside the form

FormField.js

All the pre-written tests are passing correctly

I have tested this manually using this code:

```js
import { useState } from 'react';

import {
  Box,
  Form,
  FormField,
  TextInput,
  Button,
} from 'grommet'

export const Valiidate = () => {
    const [nameInput, setNameInput] = useState("");
  
    const handleInputChange = (e) => {
      const val = e.target.value;
      setNameInput(val);
    };
  
    return (
          <Button label="Set id to ASDF" onClick={() => setNameInput("ASDF")} />
  
          <Box width="medium">
            <Form
              validate="change"
              onValidate={(validationResults) => {
                console.log("validationResults = ", validationResults.errors);
              }}
            >
              <FormField
                label="ID"
                name="firstName"
                validate={[{ regexp: /^[0-9]/i }]}
              >
                <TextInput name="firstNameTEXTINput" value={nameInput} validate onChange={handleInputChange} />
              </FormField>
            </Form>
          </Box>
        </Box>
    );
};

export default {
  title: 'Input/Form/Validate',
};
```

N/A

We are specifying in the docs, to provide `name` with the FormField, But I think it's better to show error when it doesn't happen to provider better DX, please let me know your thoughts on this. https://v2.grommet.io/formfield#validate

#5961

N/A

No

Not sure

backwards compatible


src/js/components/FormField/FormField.js
```diff
@@ -236,6 +241,16 @@ const FormField = forwardRef(
       (themeBorder &&
         children &&
         Children.map(children, (child) => {
```
g4rry420, line 243: I have few questions over this code:
- Why is themeBorder is added in this condition over here ?
-  Is there a sceneario where multiple children are required for one FormField component ?

jcfilben: Good questions!

> Why is themeBorder is added in this condition over here ?

This code checks if a border for formFields is defined in theme. If so, any border on the component that is a child of FormField should be removed and instead a border should be placed on the FormField accordion to the specification defined in `theme.formField.border`.

> Is there a sceneario where multiple children are required for one FormField component ?

There wasn't a case I could think of where we would end up with multiple children for FormField. I assume we must have run into this case somewhere though based on the map function that is being used.

src/js/components/FormField/FormField.js
```diff
@@ -250,6 +265,19 @@ const FormField = forwardRef(
...
+              return cloneElement(child, {
+                plain: true,
+                focusIndicator: false,
+                name,
+                pad:
+                  'CheckBox'.indexOf(child.type.displayName) !== -1
+                    ? formFieldTheme?.checkBox?.pad
+                    : undefined,
+              });
+            }
```
jcfilben, line 280: Instead of attaching the name to the FormField child I wonder if we should just have a warning if `name` is not specified on FormField. If the FormField is correctly setup so that the input has an associated label and meets accessibility requirements (by using `htmlFor` on the FormField and `id` on the input). Then the `name` prop only needs to be set on the FormField and not the input.

src/js/components/FormField/FormField.js
```diff
@@ -196,6 +196,11 @@ const FormField = forwardRef(
     },
     ref,
   ) => {
+    if (validate && !name) {
+      console.error(`For 'validate' prop to work properly,
+       you must specify 'name' 
+      prop to FormField and it must be unique for all FormFields.`);
+    }
```
g4rry420, line 203: Over here

src/js/components/FormField/FormField.js
g4rry420: Even if FormField is setup correctly (by using `htmlFor` on the FormField and `id` on the input), then this particular scenario will not work - #5961, besides, we are already showing error if `name` prop is not setup for FormField - https://github.com/grommet/grommet/pull/6850/files#r1242173825

Please, let me know if I am missing something :)

jcfilben: I went back and tested and I'm not sure what I was seeing when I commented before, it looks like we do need the name prop on the input. Sorry for the confusion!
Now I'm just spending some time thinking through if there are any potential issues that could arise from us automatically setting the name prop for the user.

jcfilben: Hey @g4rry420 thanks for your patience on this! I had a chance to chat with my team and we are thinking that we should not automatically add the `name` prop for the user and instead give a warning. Generally Grommet takes the approach of trying to avoid touching any native html properties in a way that would have them behave unexpectedly for the user. I think this scenario falls under that case. It would be unexpected for a `name` prop to appear on the dom for an element it was not specified on. Another consideration is if we automatically populate the `name` prop on the child of the FormField the child element may not actually be the input element. We could have a case where the input element is wrapped by a Box. With the current approach the name prop would end up on the Box element instead of the actual input.

I do think that it would still be helpful for grommet to warn users if they forgot to include the `name` prop on their input element. Ideally I think we should warn users if an input doesn't have a `name` prop within the actual input instead of at the FormField level. This is beneficial because it keeps the warning closer to the actual issue as well as avoids a situation where the FormField component needs to dig through its children's props. To accomplish this I think we could check within Grommet's input components if they are within a Form by using `FormContext` and then if they are contained in a Form throw a warning if the name prop is not defined.

g4rry420: Thank you for your explanation and I do understand the your concern of changing native html properties.
I will make the changes. Just to clarify, in every Grommet's input component, shall I add this warning ?
```js
if (!formContext.noForm && !name) {
      console.warn("'name' prop must be defined for TextInput inside Form component.")
}
```

Grommet's input components:
```js
const grommetInputNames = [
  'CheckBox',
  'CheckBoxGroup',
  'TextInput',
  'Select',
  'MaskedInput',
  'SelectMultiple',
  'TextArea',
  'DateInput',
  'FileInput',
  'RadioButtonGroup',
  'RangeInput',
  'RangeSelector',
  'StarRating',
  'ThumbsRating',
];
```

jcfilben: Yes we should add the warning to each input component. Also can we change the text of the warning to be `console.warn("The 'name' prop must be defined for TextInput while it is within a Form.")`

src/js/components/CheckBox/CheckBox.js
```diff
@@ -59,6 +59,12 @@ const CheckBox = forwardRef(
   ) => {
     const theme = useContext(ThemeContext) || defaultProps.theme;
     const formContext = useContext(FormContext);
+    if (formContext.noForm !== undefined && !formContext.noForm && !name) {
+      console.warn(
```
g4rry420, line 63: Over here, `undefined` is added because there is no `name` prop  added for DataTable component

src/js/components/TextInput/TextInput.js
```diff
@@ -106,6 +106,12 @@ const TextInput = forwardRef(
     const { format } = useContext(MessageContext);
     const announce = useContext(AnnounceContext);
     const formContext = useContext(FormContext);
+    if (formContext.noForm !== undefined && !formContext.noForm && !name) {
```
g4rry420, line 109: Over here, `undefined` is added because there is no name prop added for TextInput in DataTableColumns component

jcfilben, line 109:
```suggestion
    if (!formContext.noForm && !name) {
```
I think we should remove the `formContext.noForm !== undefined` because if the input is within a form the `noForm` in formContext will be undefined.

src/js/components/FormField/FormField.js
```diff
@@ -236,6 +241,16 @@ const FormField = forwardRef(
       (themeBorder &&
         children &&
         Children.map(children, (child) => {
+          if (
+            validate &&
+            name &&
+            child.props.name &&
+            name !== child.props.name
+          ) {
+            console.error(`${child.type.displayName} 'name' prop value -
```
jcfilben, line 250: I think we should just do a console.warn here instead of console.error

src/js/components/FormField/FormField.js
```diff
@@ -196,6 +196,11 @@ const FormField = forwardRef(
     },
     ref,
   ) => {
+    if (validate && !name) {
+      console.error(`For 'validate' prop to work properly,
```
jcfilben, line 200: I think we should just do a console.warn here instead of console.error

src/js/components/TextInput/TextInput.js
g4rry420: Over here, `undefined` is added because there is no name prop added for TextInput in DataTableColumns component and tests failed because of that

jcfilben:
Hey @g4rry420 I connected with some other members of the grommet team, we discussed cases where you could end up with inputs within a Form that shouldn't be passed though the actual Form value. For example you may have a progressive disclosure pattern where a field is only shown if the user opts-in to it. The opt-in could be a CheckBox, the CheckBox is technically an input within the form but it is only relevant to showing/hiding the additional field and its value doesn't need to be passed through the Form value. In this case it would be reasonable to not have a name on the CheckBox so that its value doesn't get passed through the Form, but with this PR anyone trying to do this will have a warning in their console. I think that a warning might be frustrating in this case because what the developer is trying to do here is valid.

I think we can't make the assumption that all inputs within a Form are intended to have their values included in the Form value. Since the behavior of the Grommet form in this case aligns with how a native html form handles `name` I feel like it is a reasonable expectation to allow users to have inputs with or without name and understand how they will behave without showing a warning. I do see value in enhancing the Grommet FormField `name` docs to help communicate this. Another possibility would be creating an opt-in eslint rule on the [eslint-plugin-grommet](https://github.com/grommet/eslint-plugin-grommet) repo so that if people wanted to opt into these warning messages they could.

Apologize for the change in direction on this PR. Having inputs within a Form that shouldn't be included in the Form value was not a case I had thought about until recently.

g4rry420:
okay, sounds good

So, shall I only warning when user is using `validate` prop with formfield and `name` isn't set ?

jcfilben:
I think we could have this as a lint rule on https://github.com/grommet/eslint-plugin-grommet that flags this for people, that way they can choose to opt-in or out of the rule

jcfilben:
Closing, I think we can enhance eslint-plugin-grommet instead

PR #6861, opened 2023-07-04
Fix accessibility violations in List Storybook

navedqb:
#### What does this PR do?
Fixes accessibility violations in 2 List storybooks

List component storybooks

-  List/Pinned
-  List/Order

#6124


jcfilben (commented):
It looks like when List has `onClickItem` or `onOrder` defined the role of the list component is set to 'listbox' I think we should always have `role='list'` regardless of `onClickItem` or `onOrder`. I think we can make this change in the List.js file (see line 222).

jcfilben:
It looks like some of the List tests are failing because they are trying to reference the 'listbox' role. We should change those tests to use `screen.getByRole('list')` (See `List/__tests__/List-test.js`).

The test snapshots will also need updated. To do this you can run `yarn jest -u`.

jcfilben:
Closing List storybook automated accessibility violations have been resolved

PR #6863, opened 2023-07-05
feat: Arrow for Diagram Component

g4rry420:
#### What does this PR do?
- Added `arrow` prop for the connection of Diagram component
- API of arrow: `from` | `to` | true
- This is the initial design which is coming from my earlier pr - https://github.com/grommet/grommet/pull/5950

Diagram.js

Currently, only manually tested using Arrow.js and Animated.js files

- Arrow.js and Animated.js files

N/A

#5882 

154091194-1c155fed-f663-48e0-9626-e8f4c6d39f04

Yes, after this pr is merged

Yes, please

backwards compatible

### DIFF
--- a/src/js/components/Diagram/Diagram.js
+++ b/src/js/components/Diagram/Diagram.js
@@ -1,4 +1,5 @@
 import React, {
+  Fragment,
   forwardRef,
   useCallback,
   useContext,
@@ -66,6 +67,19 @@ const findTarget = (target) => {
   return target;
 };
 
+const openArrow = (color, index, name = 'openArrowEnd', orient = 'auto') => (
+  <marker
+    id={`${name}-${index}`}
+    markerWidth="7"
+    markerHeight="9"
+    refX="2"
+    refY="6"
+    orient={orient}
+  >
+    <path d="M1,4 L3,6 L1,8" stroke={color} fill="none" />
+  </marker>
+);
+
 const Diagram = forwardRef(({ connections, ...rest }, ref) => {
   const theme = useContext(ThemeContext) || defaultProps.theme;
   const [dimensions, setDimensions] = useState({ width: 0, height: 0 });
@@ -117,7 +131,7 @@ const Diagram = forwardRef(({ connections, ...rest }, ref) => {
   const placeConnections = useCallback(() => {
     const containerRect = svgRef.current.getBoundingClientRect();
     const updatedConnectionPoints = connections.map(
-      ({ anchor, fromTarget, toTarget }) => {
+      ({ anchor, fromTarget, toTarget, arrow }) => {
         let points;
         const fromElement = findTarget(fromTarget);
         const toElement = findTarget(toTarget);
@@ -145,16 +159,52 @@ const Diagram = forwardRef(({ connections, ...rest }, ref) => {
             toPoint[0] += toRect.width / 2;
             if (fromRect.top < toRect.top) {
               fromPoint[1] += fromRect.height;
+
+              if (arrow === 'from') {
+                fromPoint[1] += 10;
+              } else if (arrow === 'to') {
+                toPoint[1] -= 10;
+              } else if (arrow) {
+                fromPoint[1] += 10;
+                toPoint[1] -= 10;
+              }
             } else {
               toPoint[1] += toRect.height;
+
+              if (arrow === 'from') {
+                fromPoint[1] -= 10;
+              } else if (arrow === 'to') {
+                toPoint[1] += 10;
+              } else if (arrow) {
+                fromPoint[1] -= 10;
+                toPoint[1] += 10;
+              }
             }
           } else if (anchor === 'horizontal') {
             fromPoint[1] += fromRect.height / 2;
             toPoint[1] += toRect.height / 2;
             if (fromRect.left < toRect.left) {
               fromPoint[0] += fromRect.width;
+
+              if (arrow === 'from') {
+                fromPoint[0] += 10;
+              } else if (arrow === 'to') {
+                toPoint[0] -= 10;
+              } else if (arrow) {
+                fromPoint[0] += 10;
+                toPoint[0] -= 10;
+              }
             } else {
               toPoint[0] += toRect.width;
+
+              if (arrow === 'from') {
+                fromPoint[0] -= 10;
+              } else if (arrow === 'to') {
+                toPoint[0] += 10;
+              } else if (arrow) {
+                fromPoint[0] -= 10;
+                toPoint[0] += 10;
+              }
             }
           } else {
             // center
@@ -190,6 +240,7 @@ const Diagram = forwardRef(({ connections, ...rest }, ref) => {
           round,
           thickness,
           type,
+          arrow,
           ...connectionRest
         },
         index,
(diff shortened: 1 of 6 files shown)


halocline:
Hi @g4rry420 - Thank you so much for this. This will be very useful. From an API standpoint, I am thinking about the following structure for the connection object. I think it will allow us the flexibility to extend in the future while aligning with the diagram design tools such as Figma, Adobe, Lucidchart, etc.

```
connection: {
  type: 'direct' | 'curved' | 'rectilinear',
  endpoint: 'arrow' | { from: 'arrow', to: 'arrow' },  // if string, defaults to both from and to, else use what is specified in object
  ...
}
```

which in the future could be extended to allow for other endpoint types such as "square", "round", etc.

```
connection: {
  type: 'direct' | 'curved' | 'rectilinear',
  endpoint: 'arrow' | 'square' | 'round' | { 
    from: 'arrow' | 'square' | 'round', 
    to: 'arrow' | 'square' | 'round' 
  },
  ...
}
```

g4rry420:
Sounds good, I have made changes and also added a snapshot test, please let me if there are any further changes required :)

src/js/components/Diagram/Diagram.js
```diff
@@ -222,20 +273,49 @@ const Diagram = forwardRef(({ connections, ...rest }, ref) => {
...
+                stroke={normalizeColor(colorName, theme)}
+                strokeWidth={strokeWidth}
+                strokeLinecap={round ? 'round' : 'butt'}
+                strokeLinejoin={round ? 'round' : 'miter'}
+                fill="none"
+                d={d}
+                markerStart={`url("#openArrowStart-${index}")`}
+                markerEnd={`url("#openArrowEnd-${index}")`}
+              />
+              {endpoint && arrowMarker && <defs>{arrowMarker}</defs>}
```
taysea, line 317: Why do we wrap in `defs`?

src/js/components/Diagram/Diagram.js
```diff
@@ -222,20 +273,49 @@ const Diagram = forwardRef(({ connections, ...rest }, ref) => {
             colorName = colors[index % colors.length];
           }
 
+          let arrowMarker = null;
```
taysea, line 276: This is a smaller comment but to be more flexible for future, should we name this variable `endpoint`? If needed we can alias the prop as `endpointProp`.

src/js/components/Diagram/Diagram.js
```diff
@@ -222,20 +273,49 @@ const Diagram = forwardRef(({ connections, ...rest }, ref) => {
...
+
+          if (typeof endpoint === 'object' && endpoint?.from === 'arrow') {
+            arrowMarker = openArrow(
+              normalizeColor(colorName, theme),
+              index,
+              'openArrowStart',
```
taysea, line 282: also to be more flexible for future, should we name this `endpointStart` since it might not always be an arrow?

taysea:
https://github.com/grommet/grommet/assets/12522275/ac1b31da-7814-4092-9300-103ce1d96e3a



@jcfilben @halocline I'm curious your thoughts from a design perspective. Do you think the arrow endpoints should already appear when the line isn't drawn? Or appear only once the line has met the end? Or something else?

src/js/components/Diagram/Diagram.js
```diff
@@ -145,16 +159,52 @@ const Diagram = forwardRef(({ connections, ...rest }, ref) => {
             toPoint[0] += toRect.width / 2;
             if (fromRect.top < toRect.top) {
               fromPoint[1] += fromRect.height;
+
+              if (typeof endpoint === 'object' && endpoint?.from) {
+                fromPoint[1] += 10;
```
taysea, line 164: How was `10` determined? It might be helpful to add a code comment for future.

src/js/components/Diagram/Diagram.js
```diff
@@ -222,20 +273,49 @@ const Diagram = forwardRef(({ connections, ...rest }, ref) => {
...
+                // eslint-disable-next-line react/no-unknown-property
+                animation={animation}
+                {...cleanedRest}
+                stroke={normalizeColor(colorName, theme)}
+                strokeWidth={strokeWidth}
+                strokeLinecap={round ? 'round' : 'butt'}
+                strokeLinejoin={round ? 'round' : 'miter'}
+                fill="none"
+                d={d}
+                markerStart={`url("#openArrowStart-${index}")`}
```
taysea, line 314: Is there any downside to having the `markerStart`/`markerEnd` here if the path doesn't have an arrow?

taysea: Also, if multiple diagrams are rendered on a page, will this index id approach work still?

jcfilben:
> Screen.Recording.2023-07-17.at.1.18.59.PM.mov 
> @jcfilben @halocline I'm curious your thoughts from a design perspective. Do you think the arrow endpoints should already appear when the line isn't drawn? Or appear only once the line has met the end? Or something else?

I think my expectation would be that if the arrows are at the endpoints the should only appeal once the line has been drawn to the end.

src/js/components/Diagram/Diagram.js
g4rry420: Since, `endpoint` is already a prop we are adding in Diagram component, naming the `arrowMarker` variable to `endpointProp` will create a confusion as it's expected type to be `'arrow' | { from?: 'arrow'; to?: 'arrow' }`. How about if we name this variable as `endpointMarker` ?
The reason, I have choosed this name - `endpointMarker` variable is because we are storing the `marker` tag with the openArrow in it and in future I believe for other `endpoints` we will be using `marker` for it. Please, let me know your thoughts

src/js/components/Diagram/Diagram.js
g4rry420: No, there is no downside of having `markerStart` and `markerEnd` over here if the path doesn't have an arrow.
Yes, currently, if multiple diagrams are rendered on page, the index approach will not work. I am working on new approach.

src/js/components/Diagram/Diagram.js
g4rry420: Sure, will do

src/js/components/Diagram/Diagram.js
g4rry420: The main use case for `defs`  was to holds a set of reusable definitions for the SVG markers. But, it isn't serving it's purpose right now, I will push some change in which we will using arrow based on it's open, close and color.

src/js/components/Diagram/Diagram.js
```diff
@@ -251,6 +415,14 @@ const Diagram = forwardRef(({ connections, ...rest }, ref) => {
       connections={paths}
       {...rest}
     >
+      {markerElements.length > 0 && (
+        <defs>
```
jcfilben, line 419: Do we need `defs` here?

src/js/components/Diagram/__tests__/__snapshots__/Diagram-test.tsx.snap
```diff
@@ -361,6 +361,171 @@ exports[`Diagram color 1`] = `
...
+    </div>
+    <div
+      class="c5"
+    >
+      <svg
+        class="c6"
+        preserveAspectRatio="xMinYMin meet"
+        viewBox="0 0 0 0"
+      >
+        <defs>
```
jcfilben, line 463: Do we need `defs` here?

taysea:
> > Screen.Recording.2023-07-17.at.1.18.59.PM.mov
> > @jcfilben @halocline I'm curious your thoughts from a design perspective. Do you think the arrow endpoints should already appear when the line isn't drawn? Or appear only once the line has met the end? Or something else?
> 
> I think my expectation would be that if the arrows are at the endpoints they should only appear once the line has been drawn to the end.

@g4rry420 I'm still noticing the arrow appearing before the animation is finished. Can you update so the end point arrows only appear once the animation is done?

g4rry420:
> I'm still noticing the arrow appearing before the animation is finished. Can you update so the end point arrows only appear once the animation is done?

@taysea I am unable to figure out the animation part over here. Can you please help me ?
I have tried using `animateMotion` tag with `mpath` but it doesn't seems to work. Please, let me know if I am missing something

taysea:
> > I'm still noticing the arrow appearing before the animation is finished. Can you update so the end point arrows only appear once the animation is done?
> 
> @taysea I am unable to figure out the animation part over here. Can you please help me ? I have tried using `animateMotion` tag with `mpath` but it doesn't seems to work. Please, let me know if I am missing something

I haven't tried this locally, but is there a way to conditionally render the arrows based on the animation duration/etc. So, don't have it rendered (or have it visibly hidden) until the animation completes?

PR #6866, opened 2023-07-07
Fix issue in DataSort when data is empty

jcfilben:
#### What does this PR do?
Fixes an issue where when the Data array is empty DataSort will throw an error.

Added a jest test and tested in storybook

This issue can be seen in the Data/DataTable story by searching with a string that doesn't match anything then trying to open the filters.

no
yes
backwards compatible


taysea (approved):
Good fix.

halocline (approved):

PR #6872, opened 2023-07-11
Updates Text to use transient props

taysea:
#### What does this PR do?

Updates Text to use transient props

### DIFF
--- a/src/js/components/Text/Text.js
+++ b/src/js/components/Text/Text.js
@@ -73,8 +73,8 @@ const Text = forwardRef(
         as={!as && tag ? tag : as}
         aria-label={a11yTitle}
         $colorProp={color}
-        $truncate={truncate}
         $size={size}
+        $truncate={truncate}
         {...rest}
         ref={textRef}
       >
--- a/src/js/components/Text/Text.js
+++ b/src/js/components/Text/Text.js
@@ -4,7 +4,7 @@ import { useLayoutEffect } from '../../utils/use-isomorphic-layout-effect';
 
 import { StyledText } from './StyledText';
 import { Tip } from '../Tip';
-import { useForwardedRef } from '../../utils';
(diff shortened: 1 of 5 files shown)


src/js/components/Text/Text.js
```diff
@@ -70,10 +71,10 @@ const Text = forwardRef(
     const styledTextResult = (
       <StyledText
         as={!as && tag ? tag : as}
-        colorProp={color}
         aria-label={a11yTitle}
-        truncate={truncate}
-        size={size}
+        $colorProp={color}
```
jcfilben, line 75:
```suggestion
        $color={color}
```
Should we just rename this to `$color`?

src/js/components/Text/Text.js
```diff
@@ -70,10 +71,10 @@ const Text = forwardRef(
...
+        $truncate={truncate}
+        $size={size}
```
jcfilben, line 77: Can we alphabetize these while we are here

src/js/components/Text/Text.js
taysea: For some reason I was getting a type error on that one:
Screen Shot 2023-07-11 at 1 28 45 PM

taysea: Not sure if it's a special case?

src/js/components/Text/Text.js
taysea: Yes, good call

taysea: Done.

src/js/components/Text/Text.js
jcfilben: Hmm interesting, in that case I'm fine with leaving it as `colorProp` for now

jcfilben:
Looks good just need to fix the merge conflict

jcfilben (approved):

PR #6873, opened 2023-07-11
Changed Box to use transient props

jcfilben:
#### What does this PR do?
Converts Box component to use transient props. I updated the snapshots for the Box tests and reviewed the changes there, mostly it was formatting related. Also the snapshot changes cleaned up some duplicated props (for example `display` being defined multiple times)

no
as part of the larger styled components upgrade

### DIFF
--- a/src/js/components/Box/Box.js
+++ b/src/js/components/Box/Box.js
@@ -17,36 +17,38 @@ import { BoxPropTypes } from './propTypes';
 import { SkeletonContext, useSkeleton } from '../Skeleton';
 import { AnnounceContext } from '../../contexts/AnnounceContext';
 import { OptionsContext } from '../../contexts/OptionsContext';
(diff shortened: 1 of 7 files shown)


src/js/components/Box/Box.js
```diff
@@ -230,15 +233,14 @@ const Box = forwardRef(
...
-        widthProp={width}
-        heightProp={height}
-        responsive={responsive}
-        tabIndex={adjustedTabIndex}
+        $kind={kind}
+        $overflow={overflow}
+        $wrap={wrap}
+        $width={width}
+        $height={height}
+        $responsive={responsive}
```
taysea, line 241: small detail but can we alphabetize these?

jcfilben: Yes, good note

taysea (approved):

PR #6874, opened 2023-07-11
Adjust Paragraph to use transient props

taysea:
#### What does this PR do?

Adjust Paragraph to use transient props


src/js/components/Paragraph/Paragraph.js
```diff
@@ -21,9 +23,9 @@ const Paragraph = forwardRef(
     return (
       <StyledParagraph
         ref={ref}
-        colorProp={color}
-        fillProp={fill}
-        size={size}
+        $colorProp={color}
```
jcfilben, line 26: Since we have `Omit<JSX.IntrinsicElements['p'], 'color'> ` in the index.d.ts file for Paragraph I think we can use the name `$color` instead of `$colorProp`

taysea: For some reason, still getting this flag: 

Screen Shot 2023-07-12 at 10 22 43 AM

jcfilben: Ah weird, okay with moving forward as it

jcfilben (approved):
Looks good!

PR #6881, opened 2023-07-12
Fix DataTableColumns search filter

kyuohoh:
#### What does this PR do?
This PR has a fix for the DataTableColumns filter functionality where it will look at the label first to match the searchTerm.
Initially, it was looking at the property and this was causing the search filter to improperly filter the options.

Issue example: https://f4phv2.sse.codesandbox.io/datatable

- Storybook


MikeKingdom (approved):
Looks great. I tested it locally with the codesandbox example and a few of our storybook examples.

jcfilben (approved):
Looks good, thanks for contributing!

Issue #6843, opened 2023-06-12
DataTable - option to remove the select all header cell


jcfilben:
For use cases where only a limited number of rows in the DataTable can be selected to perform actions on we should give the option of not showing the select all control in the header cell.

Currently if `onSelect` is defined the select all checkbox at the header level of the table is added by default.

Related slack thread: https://hpe.slack.com/archives/C04LMJWQT/p1685641505818049

jcfilben:
Closed by https://github.com/grommet/grommet/pull/6887

PR #6887, opened 2023-07-18
Disabled the header checkbox to (un)select all lines in DataTable

LegerG:
#### What does this PR do?

Add an option to disable the "Select All" checkbox in the header when `select` or `onSelect` props are set in `DataTable`
This prop is named `allSelectedDisabled`. I'm not sure about the naming, don't hesitate to share yours ;)

src/js/components/DataTable/DataTable.js

Only manual testing with storybook

I have this need in my projects, and see there is an open issue.

https://github.com/grommet/grommet/issues/6843

image

Yes

Yes

Backward compatible


taysea:
Thanks for filing this PR! We are having a few discussions on the design direction/experience of this kind of flow. Once we have more clear direction there, we'll drop any comments here for items to be adjusted.

LegerG:
Thanks for your response, I keep in touch.

jcfilben:
Hey @LegerG the team evaluated the use case in the original ticket (#6843) and realized that for that particular case where we limit the number of rows that can be selected we wanted to go a different direction than removing the select all checkbox. However, I saw in your PR description you mentioned that you needed this functionality for one of your projects. Would you be willing to share your use case details? The reason I ask is we like any enhancements that get added to grommet to be use case driven. This helps ensure we keep the grommet api clean.

LegerG:
The use case is the following:
I have a list of room reservations, each row as the id of the reservation, the id of the room, the datetime and the duration.
I want to select some of them with a limited criterion based on the room id. If I select the room number 12 at 4pm, I want my user to only select a reservation for the room number 12, and not for others.
In this case, the "all select" checkbox is not wanted because some rows must not be selected together.
I think this use case could be generalized as a qualitative restriction of selection, and not quantitative as the limitation on the number of rows.

For now, I have used CSS tricks to remove the checkbox and manage which lines could be selected or not. 

I hope this explanation has enough details.

jcfilben:
Thanks for sharing your use case! I'm comfortable with moving forward with this feature. @taysea @MikeKingdom @halocline any concerns with adding this?

src/js/components/DataTable/DataTable.js
```diff
@@ -87,6 +87,7 @@ const DataTable = ({
   rowDetails,
   step = 50,
   verticalAlign,
+  allSelectedDisabled = false,
```
jcfilben, line 90:
```suggestion
  allowSelectAll = true,
```
Can we change the prop name here to `allowSelectAll`?

jcfilben (commented):
Can we add a jest test for this?
Also can we add the new prop to the `DataTable/propTypes.js` and `DataTable/index.d.ts` files?

LegerG:
I have written a test, added types, and renamed the field :)

src/js/components/DataTable/index.d.ts
```diff
@@ -142,6 +142,7 @@ export interface DataTableProps<TRowType = any> {
         body?: VerticalAlignType;
         footer?: VerticalAlignType;
       };
+  allowSelectAll?: boolean;
```
jcfilben, line 145: Super minor comment but can we alphabetize this prop (it should appear under `alignSelf` and above `background`)

src/js/components/DataTable/propTypes.js
```diff
@@ -179,6 +179,7 @@ if (process.env.NODE_ENV !== 'production') {
         footer: PropTypes.oneOf(verticalAlign),
       }),
     ]),
+    allowSelectAll: PropTypes.bool,
```
jcfilben, line 182: Same request here, can we alphabetize this so it appears above `background`

src/js/components/DataTable/stories/AllowSelectAll.js
```diff
@@ -0,0 +1,103 @@
...
+        select={select}
+        onSelect={(selected) => setSelect(selected)}
+        allowSelectAll={allowSelectAll}
+      />
+    </Box>
+  );
+};
+
+AllowSelectAll.storyName = 'Allow Select All';
+
```
jcfilben, line 100:
```suggestion

AllowSelectAll.parameters = {
  chromatic: { disable: true },
};

```
Can we disable this story from chromatic snapshots?

jcfilben (commented):
Just some minor comments otherwise looks good

jcfilben (approved):
Looks good!

Issue #6888, opened 2023-07-18
Layer causes DOMException


stephanpelikan:
### Expected Behavior

Hiding `Layer`s should not break rendering, even in case of race-conditions.

### Actual Behavior

In some situations hiding a `Layer` causes DOMException because the child was already removed (or maybe its parent changed).

### URL, screen shot, or Codepen exhibiting the issue

![grommet](https://github.com/grommet/grommet/assets/5190548/d428c99a-60f2-414c-89a2-b4027c149727)

### Steps to Reproduce

In my application on certain pages it is reproducable every ~ 4th render (Ctrl-R), but only with dev-server.

It seems to be a race condition. My use-case is that I load data in `useEffect` showing a ´Layer´ for a loading-indicator. On response the `Layer` is hidden.

Using production-builts it works without error because they load faster. In Firefox it works fine even with dev-server, but in Chrome I get those errors.

### Your Environment

- Grommet version: 2.33.0
- Browser Name and version: Chrome 114.0.5735.199 (Offizieller Build) (64-Bit)
- Operating System and version (desktop or mobile): Windows 10 Desktop

jcfilben:
Fixed in #6889

PR #6889, opened 2023-07-18
Suppress DOMExceptions on hiding Layer (#6888)

stephanpelikan:
#### What does this PR do?

Suppress DOMExceptions on hiding `Layer` (see #6888).

Good question. There are only minor code changes which and this should not effect functionality. But maybe suppressing the DOMException unhides other problems. In my manual testing I didn't see this, but maybe I overlooked something.

Rerun automated tests. As well as manual tests.

Since this is hard to reproduce, just hiding a `Layer`.

#6888

This screenshot shows that the error is gone using this PR:

grommet3

One the left side you can see the header and the menu which proofs that rendering is fine on using this PR.

No

Yes

It is backwards compatible

### DIFF
--- a/src/js/components/Layer/Layer.js
+++ b/src/js/components/Layer/Layer.js
@@ -58,12 +58,16 @@ const Layer = forwardRef((props, ref) => {
               .getRootNode()
               .getElementById('layerClone');
             if (clone) {
-              containerTarget.removeChild(clone);
+              if (containerTarget.contains(clone)) {
+                containerTarget.removeChild(clone);
+              }
               layerContainer.remove();
             }
           }, animationDuration);
         } else {
-          containerTarget.removeChild(layerContainer);
+          if (containerTarget.contains(layerContainer)) {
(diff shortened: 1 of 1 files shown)


taysea:
Thanks for filing this PR. It would be ideal if we could fix the Layer behavior in a way that no console errors or warnings would occur.

I took a look at the original issue you filed, and based on that it seems like before we try to `removeChild` we should check if `clone` is still a child of the `containerTarget`. (see code below).

current layer code:
```
 if (clone) {
              containerTarget.removeChild(clone);
              layerContainer.remove();
            }
```

I think we could leverage something like `containerTarget.contains(clone)`. Do you mind trying that out and updating your PR if that works?

stephanpelikan:
I updated as you suggested and tested the code and it works perfectly fine. 😄

jcfilben (approved):
Looks good!

taysea (approved):
Great, thanks!

stephanpelikan:
Thank you 👍 😄

PR #6890, opened 2023-07-18
Convert List to transient props

taysea:
#### What does this PR do?

Convert List to use transient props.

### DIFF
--- a/src/js/components/List/List.js
+++ b/src/js/components/List/List.js
@@ -9,6 +9,7 @@ import { Keyboard } from '../Keyboard';
 import { Pagination } from '../Pagination';
 import { Text } from '../Text';
 import {
+  convertRestToTransientProps,
   focusStyle,
   genericStyles,
   normalizeColor,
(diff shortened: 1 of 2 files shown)


src/js/components/List/List.js
```diff
@@ -679,9 +681,9 @@ const List = React.forwardRef(
                     tag="li"
```
jcfilben, line 681: Can we alphabetize

src/js/components/List/List.js
```diff
@@ -679,9 +681,9 @@ const List = React.forwardRef(
...
                     background={adjustedBackground}
                     border={adjustedBorder}
-                    isDisabled={isDisabled}
                     flex={false}
                     pad={pad || theme.list.item.pad}
+                    $isDisabled={isDisabled}
                     {...defaultItemProps}
                     {...boxProps}
                     {...clickProps}
```
jcfilben, line 689: Do any of the props passed through here need to be transient?

taysea: Great question, I will take a look.

jcfilben:
Closing in favor of styled components approach in https://github.com/grommet/grommet/pull/6947

PR #6900, opened 2023-07-21
Fix 'undefined' value in MaskedInput

### LINKED ISSUE #6884: MaskedInput displays "undefined" when unsupported key is entered
### Expected Behavior

MaskedInput shouldn't display "undefined" as text value.

### Actual Behavior

When MaskedInput is in a FormField and a user types in an invalid character, the text value becomes "undefined".

Note: This only happens when MaskedInput is within a Form.

https://codesandbox.io/s/maskedinput-undefined-wtlqp6?file=/src/App.js

### Steps to Reproduce

1. Type "m" into MaskedInput
2. Notice how "undefined" becomes text in masked input

### Your Environment

- Grommet version: stable

MikeKingdom:
#### What does this PR do?
Fixes an issue where the MaskedInput value shows as `undefined` when the `value` property isn't supplied and an non-matching character is input.

Fixes #6884

MaskedInput.js

Manual, storybook and jest

```jsx
export const Barcode = () => {
  const mask = [
    {
      length: [1, 14],
      regexp: /^(3|38|38[0,1]|38[0,1][0-9]{1,11})$/,
    },
  ];

  return (
    // <Grommet theme={...}>
      <Box width="medium">
        <MaskedInput
          mask={mask} 

       />
      </Box>
    </Box>
    // </Grommet>
  );
};
```

#6884

No

Yes

Backwards compatible

### DIFF
--- a/src/js/components/MaskedInput/MaskedInput.js
+++ b/src/js/components/MaskedInput/MaskedInput.js
@@ -185,6 +185,7 @@ const MaskedInput = forwardRef(
     const [value, setValue] = formContext.useFormInput({
       name,
       value: valueProp,
+      initialValue: '',
     });
 
     const valueParts = useMemo(() => parseValue(mask, value), [mask, value]);
--- a/src/js/components/MaskedInput/__tests__/__snapshots__/MaskedInput-test.js.snap
+++ b/src/js/components/MaskedInput/__tests__/__snapshots__/MaskedInput-test.js.snap
@@ -1062,7 +1062,7 @@ exports[`MaskedInput event target props are available option via mouse 4`] = `
     id="item"
     name="item"
     placeholder="!"
-    value=""
+    value="aa!"
   />
 </div>
 `;
@@ -2519,7 +2519,7 @@ exports[`MaskedInput option via mouse 4`] = `
     id="item"
     name="item"
     placeholder="!"
-    value=""
+    value="aa!"
   />
 </div>
 `;


src/js/components/MaskedInput/MaskedInput.js
```diff
@@ -185,6 +185,7 @@ const MaskedInput = forwardRef(
     const [value, setValue] = formContext.useFormInput({
       name,
       value: valueProp,
+      initialValue: '',
```
taysea, line 188: Should we take `defaultValue` into account here at all?

MikeKingdom: > Should we take `defaultValue` into account here at all?

What defaultValue? MaskedInput doesn't have a defaultValue prop. I think at best the caller would pass that in as a `value`

jcfilben (approved):
Looks good!

src/js/components/MaskedInput/MaskedInput.js
taysea: TextInput supports a `defaultValue` -- I'm assuming the "rest" from MaskedInput ends up on the TextInput it's built with. https://github.com/grommet/grommet/blob/master/src/js/components/TextInput/TextInput.js#L477

In some cases like in CheckBoxGroup we do something like `initialValue: defaultValue || []` so wasn't sure if that was important to consider here or not. https://github.com/grommet/grommet/blob/master/src/js/components/CheckBoxGroup/CheckBoxGroup.js#L45

We don't have `defaultValue` documented on MaskedInput at the moment so I would be fine leaving it separate.

MikeKingdom: > TextInput supports a `defaultValue` -- I'm assuming the "rest" from MaskedInput ends up on the TextInput it's built with. https://github.com/grommet/grommet/blob/master/src/js/components/TextInput/TextInput.js#L477
> 
Interesting. Note that it's not documented for TextInput either (https://v2.grommet.io/textinput) and it's not in TextInput's propTypes or typescript from what I can tell. I could add it to the formInput initialValue though

MikeKingdom: > TextInput supports a `defaultValue` -- I'm assuming the "rest" from MaskedInput ends up on the TextInput it's built with. https://github.com/grommet/grommet/blob/master/src/js/components/TextInput/TextInput.js#L477

And actually it doesn't look like MaskedInput uses TextInput. It passes `rest` along to `StyledMaskedInput` which is just a `styled.input`

So I think we're good here (although poor TextInput probably should have defaultValue documented.)

taysea (approved):

Issue #6852, opened 2023-06-20
Bad user experience with long tags


briancoburn:
The HPE Tag spec allows 128 characters for the Tag name and 256 characters for the Tag value. Longer Tags that approach these limits can cause ui problems and lead to a bad user experience.

### Expected Behavior

Minimally, the Tag component should accept props for a character limit. If the limit is exceeded, the Tag component should truncate the characters exceeding the limit and display the entire string in a tooltip. Another approach would be to allow a user to override the internal Text elements in the Tag by passing their own custom elements as props.

### Actual Behavior

See attached screenshot. For very long tag names, it might overflow the viewport and the user cannot even see the tag value. In the attached image, the Tag has white-space: no-wrap and width:fit-content styles applied. Without these, sometimes the Tag width matches that of the container rather than the content, and creates awkward line breaks that can result in a tag spanning many lines.

### URL, screen shot, or Codepen exhibiting the issue
<img width="1530" alt="Screen Shot 2023-06-15 at 3 09 15 PM" src="https://github.com/grommet/grommet/assets/16736249/2c06d9fc-efec-4539-aec4-895c3f96d6b5">

### Steps to Reproduce

Create a Tag with a very long name(128 characters) and a very long value(256 characters). Put that Tag in a width-constrained container such as a cell in a Table Column. Note the Text elements in the Tag overflow the enclosing container

### Your Environment

- Grommet version: 2.32.0
- Browser Name and version:Chrome 114.0.5735.133 (Official Build) (arm64)
- Operating System and version (desktop or mobile): Mac OS Monterey 12.6

jcfilben:
Hey @briancoburn thanks for raising this issue! Could you share a use case where you do end up with really long tag names and values? I think seeing an actual use case will help us determine if this is something that is frequently happening as well as how best to handle it from a user experience perspective.

briancoburn:
hey @jcfilben - the image I included demonstrates the issue. The use case is that we need to support the HPE tagging standard, which says we must support 128 character names and 256 character values for the Tag. The attached image shows what that looks like should the user create Tags with really long names and/or values. We don't yet have data as to how frequently this might occur, but the issue is that we need to allow the user to enter these long names and values, and if they do, it can break the ui. If the Tag component took props which allowed the name and value to be truncated and show a tooltip, the issues shown in the attached image could be avoided.

samuelgoff:
[...]

PR #6906, opened 2023-07-26
add truncation support and tip for long tags

briancoburn:
#### What does this PR do?

HPE Tag spec supports tag names of up to 128 characters and tag values of up to 256 characters. If either value is anywhere near the max length, it can overflow containers and break ui. 

This pr adds support for truncation and showing tip for long tag names and values. Addresses this issue:
https://github.com/grommet/grommet/issues/6852

Tested in ui using the updated grommet tag component. It takes a truncate prop, and optional truncateNumChars prop, which determines the maximum number of characters allowed before truncation occurs.


jcfilben:
Hey @briancoburn saw you marked this as a draft, just wanted to check if you are still working on this and we should hold off on reviewing

src/js/components/Tag/Tag.js
```diff
@@ -23,27 +35,61 @@ const Tag = forwardRef(
...
     };
-    const contents = (
+    const displayName =
+      truncate && name.length > truncateNumChars
+        ? `${name.substring(0, truncateNumChars - 3)}...`
+        : name;
+    const displayValue =
+      truncate && value.length > truncateNumChars
+        ? `${value.substring(0, truncateNumChars - 3)}...`
+        : value;
```
MikeKingdom, line 45: FYI I'd rather see this done via some element width constraint and the css ellipsis truncation rather than string truncation based on number of chars.

samuelgoff: I agree with @MikeKingdom's feedback. You should be using CSS-only truncation, not `substring`. In general, if it is possible to accomplish something using CSS-only solutions, you should try to make it work that way.

briancoburn: Hey @MikeKingdom and @samuelgoff thanks for the feedback. I have updated the pr to use width based constraints and let the the Text elements handle the truncation internally. Also, tips are "stacked" vertically so they don't get crazy wide and I've added window.resize update for responsive truncation. If you feel like this is ready to go, I'll take it out of draft status, or if you have further feedback, I would be happy to address any other issues.

jcfilben (commented):
Hi @briancoburn thanks for working on this! Here are some comments so far:

- The tag should wrap by default and only truncate if the truncate prop is set to `true` or ‘tip’. Currently the tag is truncating by default and no longer wrapping.

- If we have a really long name and a very short value only the name should truncate (or vice versa).

- There is a bit of code that only needs to be run if the truncate prop is `true` or `‘tip’`. Adding a check before this code is run would be beneficial.

- In addition to supporting the truncate prop on Tip we should allow callers to specify `tag.name.truncate` or `tag.value.truncate` in the theme since we already document this as a valid option.

- The `propTypes.js` and `index.d.ts` files for Tag will need to be updated to reflect the new prop.

- I don’t think we should enforce the MAX_NAME_LENGTH within Grommet. It feels like this is only specific to hpe cases and feel a bit too strict for Grommet.

- Can we add a storybook example that demonstrates the tag truncation?

- Having a padding default value feels a bit brittle. I think there are better ways to determine if the content is overflowing and should be truncated. One potential route you could go is you could prevent the tag from wrapping if truncate is defined and then do a check like `containerRef.current.scrollWidth > containerRef.current.offsetWidth`.

samuelgoff:
> * The tag should wrap by default and only truncate if the truncate prop is set to `true` or ‘tip’. Currently the tag is truncating by default and no longer wrapping.

Hey @jcfilben while I agree with most of your feedback, I disagree that the tag should wrap by default, because doing so makes it more difficult to read. I would invert the default behavior such that it **_does not wrap by default_**, and `wrap` is an optional prop that can be set to `true`

samuelgoff:
> I think we need to keep the wrap by default for backwards compatibility purposes

Respectfully, tags that wrap impede readability -- which is the most important criterion for usability of this type of component. If the existing design defaults to wrapping, IMO this is a **_usability defect_** & a fix to this defect that breaks backward compatibility is **_not_** out of the realm of reason -- especially if this change is clearly communicated in release notes

jcfilben:
> > I think we need to keep the wrap by default for backwards compatibility purposes
> 
> Respectfully, tags that wrap impede readability -- which is the most important criterion for usability of this type of component. If the existing design defaults to wrapping, IMO this is a **_usability defect_** & a fix to this defect that breaks backward compatibility is **_not_** out of the realm of reason -- especially if this change is clearly communicated in release notes

That's a fair stance to take, I'm okay with us considering the wrapping to be a defect and make the default behavior truncation. And agree we can just make sure to communicate this in the release notes so that people are not surprised.

briancoburn:
hey Jessica- 

Sorry for the delay. You guys did create some challenging requirements.
I hope to have another update for you soon. 

Brian

On 2023-08-21 08:02, Jessica Filben wrote:

> Hey @briancoburn [1] just wanted to check in on this PR and see if there were any questions or clarifications needed about the review comments 
> 
> --
> Reply to this email directly, view it on GitHub [2], or unsubscribe [3].
> You are receiving this because you were mentioned.Message ID: ***@***.***>
 

Links:
------
[1] https://github.com/briancoburn
[2] https://github.com/grommet/grommet/pull/6906#issuecomment-1686500674
[3]

jcfilben:
Hey Brian
No worries, just wanted to check in!

jcfilben:

PR #6941, opened 2023-09-11
Styled components withConfig approach

jcfilben:
#### What does this PR do?
Upgrade to use styled components v6

Styled components that are styling native html components are now using `withConfig(styledComponentsConfig)`. `styledComponentsConfig` is defined in `utils/styles.js`. The config filters out any props that are not recognized as native html props. For styled components that are styling Grommet components we are not filtering out anything and letting all props pass through. Each Grommet component eventually gets rendered as a native html element and this is where we want the prop filtering to occur, not before then.

Stylis selector patterns require a prepended `&` or `&&`. I made this adjustment in some files. See `StyledCheckBox.js` as an example. Styled components documentation on this:
https://styled-components.com/docs/basics#pseudoelements-pseudoselectors-and-nesting

Tested that `@container` works with this approach
Tested with the following story and compared to how this is currently working in grommet (see [codesandbox](https://codesandbox.io/s/grommet-v2-template-forked-7lkg4z?file=/index.js))
```
import React from 'react';
import { Box } from 'grommet';

const CustomBox = styled(Box)`
  padding: 10px;
`;

export const SimpleBox = () => (
  <>
    <CustomBox forwardedAs="span" background="pink" border>
      box content
    </CustomBox>
    <Box as="span" border>
      regular box
    </Box>
  </>
);

SimpleBox.storyName = 'Simple';

export default {
  title: 'Layout/Box/Simple',
};
```


package.json
```diff
@@ -51,15 +51,15 @@
   "peerDependencies": {
     "react": "^16.6.1 || ^17.0.0 || ^18.0.0",
     "react-dom": "^16.6.1 || ^17.0.0 || ^18.0.0",
-    "styled-components": "^6.0.0"
+    "styled-components": "^5.1.0 || ^6.0.0",
+    "stylis": "^4.0.0"
```
MikeKingdom, line 55: styled-components 6 lists stylis ^4.3.0 as a dependencies. Does this need to be a peer dependency? (especially since it isn't a dependency for styled-components 5)

jcfilben: Good note, I'll remove stylis from the peer dependencies

package.json
```diff
@@ -136,7 +137,8 @@
     "require-reload": "^0.2.2",
     "simple-git": "^3.19.1",
     "storybook": "7.0.20",
-    "styled-components": "5.3.9",
+    "styled-components": "^6.0.0",
+    "stylis": "^4.0.0",
```
taysea, line 141: Will `stylis` be included when we install `styled-components 6`?

I think we can remove.

jcfilben:
> @jcfilben Can you add links to the issues/docs you found related to use of & / &&?

Yes, added a link to the PR description

package.json
jcfilben: I was thinking stylis was a PeerDependency of Styled Components but looks like it is just a regular dependency so yes I think we can remove it here.

jcfilben:
Closing in favor of https://github.com/grommet/grommet/pull/6947

Issue #6730, opened 2023-04-04
DataFilter Range problems


LiveFreeAndRoam:
### Slider and Position of Range Labels

The range boundary values are rendered _inline_ with the slider control (see image). When those numbers become large, the slider control becomes very compressed, to the point it is difficult to use.

This is the current slider control with boundary values rendered inline:

![image](https://user-images.githubusercontent.com/8231064/229915943-9b312cbe-2eb9-4c39-ad39-36fe283c3708.png)

When range values become large, it would be better if the range appeared under the control.  Perhaps a property something like: `{range: { labelAlign: "[top|bottom|inline]" }}`.  E.g. if it were aligned to the bottom, it might look something like this:

![image](https://user-images.githubusercontent.com/8231064/229917150-60b8c2ae-c64a-4510-9804-b3af2c49b2e8.png)

### Floating Point
The slider displays floating point values when the range is not evenly divisible by 20.

![image](https://user-images.githubusercontent.com/8231064/229915650-a6f7e24e-e5a1-4fe8-8ad2-1f6f9ea22c86.png)

### Step size
The step size is currently hardcoded to be `(max-min)/20`.  With 5000 items, the slider granularity is 250. Worst case, my min is 250 below and max 250 above where I want to filter.  i.e. it means I have a range of 500 items that I'm filtering.

The step-size (or number of steps) needs to be made a property so I have control over the granularity of slider steps.

When the range is fewer than 20, the step size should become 1.

### Control the type of filter
The type of filter is determined by the data type and there is no way to override that.

When the data set contains numbers, the DataFilter only provides a Range selector.  However, if the range is very small, (maybe just 5 items or fewer), I'd rather display it as a Select list or CheckBoxGroup.

Perhaps add a property to allow the user to set `filterType="RangeSelector|CheckBoxGroup|SelectMultiple"`

### Filter Range Excludes boundary values
The filter boundary values are exclusive, when they should be inclusive.  E.g. filter 0-3, the filter omits 0 and 3.  This actually makes it impossible to filter only on the min value (e.g. 0) or the max value.

The only workaround to this is to have the filter display values beyond the dataset.  E.g. to filter for items with a value of 0, I need to configure the {min: -1, max: 1}.  Meanwhile, negative values are not permitted in my dataset, so this is a hack to workaround the exclusive behavior.

There is also greater cognitive load required to mentally calculate values either side of the actual range to be filtered, rather than just displaying all the values to be filtered.

taysea:
[...]

PR #6942, opened 2023-09-11
Allow caller to specify step for DataFilter range

taysea:
#### What does this PR do?

Allows caller to specify step for a property in Data by:
1. specifying in `properties` on Data
2. specifying in `range` on a given DataFilter.

Tested in a storybook example. Added jest test.

- [ ] `userEvent` is used in place of `fireEvent`. NOTE: Used fireEvent to be able to do drag event.

Related to #6730 

Yes.

Yes. Added `step` to Data `properties.range` and DataFilter `range`.

Backwards compatible.


src/js/components/DataFilter/__tests__/DataFilter-test.tsx
```diff
@@ -201,6 +203,54 @@ describe('DataFilter', () => {
...
+        </Data>
+      </Grommet>,
+    );
+
+    const lowerBound = screen.getByRole('slider', { name: 'Lower Bounds' });
+    fireEvent.mouseDown(lowerBound);
+    fireEvent.mouseMove(lowerBound, { clientX: 31, clientY: 20 });
+    fireEvent.mouseUp(lowerBound);
+    expect(lowerBound.getAttribute('aria-valuenow')).toEqual('45');
+    expect(asFragment()).toMatchSnapshot();
```
jcfilben, line 225: Do we need a snapshot test here or can we just rely on the check above `expect(lowerBound.getAttribute('aria-valuenow')).toEqual('45');`?

jcfilben (commented):
Code looks good
In addition to the `index.d.ts` files should the `propTypes.js` files also be updated?

src/js/components/DataFilter/__tests__/DataFilter-test.tsx
taysea: The snapshot is just capturing styling info, while the previous expect is more so testing fuctionailty.

Given there are other snapshots capturing range, should I remove this one to keep things more lean?

jcfilben: Yeah let's remove the snapshot. This article has some good points about snapshot testing: https://kentcdodds.com/blog/effective-snapshot-testing. My main takeaway was that snapshot testing can be very useful, but I think in Grommet we have a tendency to overuse it a bit when the test could be written using other assertions.

taysea: Totally agree, I'll remove.

jcfilben (approved):
Looks good!

taysea:
Converting to draft because some of the alphabetically sorted filters appear to be missorted in chromatic.

src/js/components/DataFilter/DataFilter.js
```diff
@@ -116,7 +128,13 @@ export const DataFilter = ({
           label
           min={range[0]}
           max={range[1]}
-          step={(range[1] - range[0]) / 20}
+          step={
+            // from `range` on DataFilter
+            rangeProp?.step ||
+            // from range in Data `properties`
+            properties?.[property]?.range?.step ||
+            (range[1] - range[0]) / 20
```
halocline, line 136: Where does "20" come from? What is the rationalization behind that value?

taysea: I think it was just a value that Eric chose as a consistent number of steps for a range. Not sure if there was any further rationalization beyond that.

halocline: can we do something like `const defaultSteps = 20;` to use a variable to be self documenting?

halocline (commented):
small comment related to implementation documentation

halocline (approved):

jcfilben (approved):

Issue #6940, opened 2023-09-08
DataFilters do not show an filterable properties when result set is zero


halocline:
### Expected Behavior

When applying too restrictive filter criteria yielding a result set of zero, a user should be able open the data filters and modify the filter criteria.

### Actual Behavior

With a filter result set of zero, when re-opening the data filters, no filters are shown and therefor a user can't modify their filter criteria.

### URL, screen shot, or Codepen exhibiting the issue

![Screenshot 2023-09-08 at 3 40 52 PM](https://github.com/grommet/grommet/assets/1756948/a8dc53ba-fbfd-4fc1-acdf-cebdd4291f1e)

### Steps to Reproduce

1. Storybook Data/Data/Simple
1. Open filters
1. Apply filters for location = Boise and date = 2018-06-09
1. Click apply filters
1. Result set should yield zero records
1. Reopen the filters 

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #6943, opened 2023-09-11
Fixed DataFilters to show all filter options

halocline:
#### What does this PR do?

Closes https://github.com/grommet/grommet/issues/6940.

Uses `unfilteredData` instead of `data` as source to build up default filter options. Previously, the logic was built up from `data`, however if `data` reached a length of zero, then no object was available to build up the options.

- review issue
- DataFilters.js

- Storybook
- Added unit tests

Use the following stories:
- https://storybook.grommet.io/?path=/story/data-data-simple--simple
- Story 1 - when `properties` is not specified, [data provides default filters](https://v2.grommet.io/data#properties).
- Story 2 - when `properties` is specified, `properties` becomes [the schema definition for which properties are filterable](https://v2.grommet.io/data#properties) by filters.

For each:
1. Use a combination of filters to get zero results.
2. Re-open filters, all filter properties should be options.

Story w/out `properties` prop:
```
 <Data
    data={[
      {
        name: 'aa',
        enabled: true,
        rating: 2.3,
        sub: { note: 'ZZ' },
        tags: ['qa', 'staging', 'prod'],
      },
      {
        name: 'bb',
        enabled: false,
        rating: 4.3,
        sub: { note: 'YY' },
        tags: ['qa', 'staging'],
      },
      { name: 'cc', sub: {}, tags: ['qa'] },
      { name: 'dd' },
    ]}
    toolbar
  >
    <DataTable
      columns={[
        { property: 'name', header: 'Name' },
        { property: 'sub.note', header: 'Note' },
        {
          property: 'tags',
          header: 'Tags',
          render: ({ tags }) => (tags ? tags.join(', ') : null),
        },
      ]}
    />
  </Data>
```

Story with `properties` prop:
```
 <Data
    data={[
      {
        name: 'aa',
        enabled: true,
        rating: 2.3,
        sub: { note: 'ZZ' },
        tags: ['qa', 'staging', 'prod'],
      },
      {
        name: 'bb',
        enabled: false,
        rating: 4.3,
        sub: { note: 'YY' },
        tags: ['qa', 'staging'],
      },
      { name: 'cc', sub: {}, tags: ['qa'] },
      { name: 'dd' },
    ]}
    toolbar
    properties={{
      name: { label: 'Name' },
      'sub.note': { label: 'Note' },
      tags: {
        label: 'Tags',
        options: [
          { label: '01 - Development', value: 'dev' },
          { label: '02 - QA', value: 'qa' },
          { label: '03 - Staging', value: 'staging' },
          { label: '04 - Production', value: 'prod' },
        ],
      },
    }}
  >
    <DataTable
      columns={[
        { property: 'name', header: 'Name' },
        { property: 'sub.note', header: 'Note' },
        {
          property: 'tags',
          header: 'Tags',
          render: ({ tags }) => (tags ? tags.join(', ') : null),
        },
      ]}
    />
  </Data>
```

https://github.com/grommet/grommet/issues/6940

No.

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/DataFilters/DataFilters.js
+++ b/src/js/components/DataFilters/DataFilters.js
@@ -26,10 +26,10 @@ const layerProps = {
 export const DataFilters = ({ drop, children, heading, layer, ...rest }) => {
   const {
     clearFilters,
-    data,
     id: dataId,
     messages,
     properties,
+    unfilteredData,
     view,
   } = useContext(DataContext);
   const { format } = useContext(MessageContext);
@@ -66,10 +66,10 @@ export const DataFilters = ({ drop, children, heading, layer, ...rest }) => {
   let content = children;
   if (Children.count(children) === 0) {
     let filtersFor;
-    if (!properties && data && data.length)
+    if (!properties && unfilteredData && unfilteredData.length)
       // build from a piece of data, ignore object values
-      filtersFor = Object.keys(data[0]).filter(
-        (k) => typeof data[0][k] !== 'object',
+      filtersFor = Object.keys(unfilteredData[0]).filter(
+        (k) => typeof unfilteredData[0][k] !== 'object',
       );
     else if (Array.isArray(properties)) filtersFor = properties;
     else if (typeof properties === 'object')
--- a/src/js/components/Data/__tests__/Data-test.tsx
+++ b/src/js/components/Data/__tests__/Data-test.tsx
@@ -30,6 +30,7 @@ const data = [
 ];
 
 describe('Data', () => {
+  window.scrollTo = jest.fn();
   beforeEach(createPortal);
 
   test('renders', () => {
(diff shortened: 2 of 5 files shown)


src/js/components/DataFilters/DataFilters.js
```diff
@@ -66,10 +66,10 @@ export const DataFilters = ({ drop, children, heading, layer, ...rest }) => {
   let content = children;
   if (Children.count(children) === 0) {
     let filtersFor;
-    if (!properties && data && data.length)
+    if (!properties && unfilteredData && unfilteredData.length)
       // build from a piece of data, ignore object values
-      filtersFor = Object.keys(data[0]).filter(
-        (k) => typeof data[0][k] !== 'object',
+      filtersFor = Object.keys(unfilteredData[0]).filter(
```
halocline, line 71: Is it reasonable for us to only be looking at the first object to build up the default options? This assumes the first object is representative of all objects in the data set and has all properties. What if other objects have properties not represented?

MikeKingdom: I think it's reasonable given they can specify properties if they have sparse objects for some reason

MikeKingdom (approved):

taysea (approved):

jcfilben (approved):
Looks good!

PR #6947, opened 2023-09-18
Styled components v6

jcfilben:
#### What does this PR do?

Upgrade to support both styled components v5 and v6

Update 10/21: Upon further exploration of the StyleSheetManager behavior in styled components I think we could run into situations where having a built in StyleSheetManager results in confusing behavior for the user. The benefits of the built in StyleSheetManager are outweighed by the potential risks and added confusion it may cause for the user.

Styled components that are styling native html components are now using withConfig(styledComponentsConfig). styledComponentsConfig is defined in utils/styles.js. The config filters out any props that are not recognized as native html props. For styled components that are styling Grommet components we are not filtering out anything and letting all props pass through. Each Grommet component eventually gets rendered as a native html element and this is where we want the prop filtering to occur, not before then.

Stylis selector patterns require a prepended & or &&. I made this adjustment in some files. See StyledCheckBox.js as an example. Styled components documentation on this:
https://styled-components.com/docs/basics#pseudoelements-pseudoselectors-and-nesting

Tested that `@container` works with this approach
Tested with the following story and compared to how this is currently working in grommet (see [codesandbox](https://codesandbox.io/s/grommet-v2-template-forked-7lkg4z?file=/index.js))
```
import React from 'react';
import { Box } from 'grommet';

const CustomBox = styled(Box)`
  padding: 10px;
`;

export const SimpleBox = () => (
  <>
    <CustomBox forwardedAs="span" background="pink" border>
      box content
    </CustomBox>
    <Box as="span" border>
      regular box
    </Box>
  </>
);

SimpleBox.storyName = 'Simple';

export default {
  title: 'Layout/Box/Simple',
};
```


MikeKingdom:
When in the shadow dom I get an odd drop shadow in some cases. 
image

The same code when not in the shadow dom looks fine. Here's the styles difference. Notice also the lack of -webkit* styles when in the shadow DOM
Shadow dom:
image

Normal dom:
image

jcfilben:
> When in the shadow dom I get an odd drop shadow in some cases. image
> 
> The same code when not in the shadow dom looks fine. Here's the styles difference. Notice also the lack of -webkit* styles when in the shadow DOM Shadow dom: image
> 
> Normal dom: image

The lack of `-webkit` styles is expected in the shadow dom if you don't specify `enableVenderPrefixes`. I assume if you have `enableVendorPrefixes` set on the styleSheetManager then the `-webkit` prefixes will show up again.

The box shadow issue does seem weird though

MikeKingdom:
> The lack of `-webkit` styles is expected in the shadow dom if you don't specify `enableVenderPrefixes`. I assume if you have `enableVendorPrefixes` set on the styleSheetManager then the `-webkit` prefixes will show up again.

Right but that means our workaround isn't backward compatible plus it's affecting a part of the DOM it's not supposed to (outside the shadow root since the StyleSheetManager has not target)

> The box shadow issue does seem weird though

taysea (approved):

PR #6955, opened 2023-10-02
Fixed key render story console error

### LINKED ISSUE #6588: List - key render story error

HardikChoudhary24:
#### What does this PR do?
Fixes #6588
- Fixes key render story console error

- Run storybook locally
- Navigate to the List/Key render story
- Open the console
- No console errors are found

- Ran locally on storybook.

#6588

Screenshot 2023-10-02 at 12 05 23 PM

- Yes


src/js/components/List/stories/RenderedList.js
```diff
@@ -22,16 +22,8 @@ export const RenderedList = () => (
...
-          {item.entry}
-        </Text>
-      )}
-      secondaryKey={(item) => (
-        <Text key={item.location} size="small" color="dark-4">
-          {item.location}
-        </Text>
-      )}
+      primaryKey={(item) => item.entry}
+      secondaryKey={(item) => item.location}
```
jcfilben, line 26: Can we preserve the `Text` component being passed to the primary and secondaryKey properties along with the styling?

HardikChoudhary24: @jcfilben yes we can preserve it but instead of doing  
```
 primaryKey={(item) => (
        <Text key={item.entry} size="large" weight="bold">
          {item.entry}
        </Text>
 )}
```  
we can do
 ```
primaryKey={(item) => [
        item.entry,
        <Text size="large" weight="bold">
          {item.entry}
        </Text>,
 ]}
```  
so that it takes the returning value of the getValue function as an object instead of number or string in 👇
        https://github.com/grommet/grommet/blob/420cecb3148948c9efbe3635b51926324135b9f0/src/js/components/List/List.js#L364-L374
        
And then we can change the above conditional render as
```
           else if (primaryKey) {
                  const primary = getValue(item, index, primaryKey);
                  content =
                    typeof primary === 'string' ||
                    typeof primary === 'number' ? (
                      <Text key="p" weight="bold">
                        {primary}
                      </Text>
                    ) : (
                      <Text key="p" {...primary[1].props}>
                        {primary[0]}
                      </Text>
                    );
```
This approach allows us to preserve the ` Text ` component along with its styling. Additionally, I've observed that we don't need to modify how we pass the ` secondaryKey `. The warning in the console is specifically related to the ` primaryKey `, so we can leave the ` secondaryKey ` unchanged.

**Comment by HardikChoudhary24** [comment]
@jcfilben I have added the `itemKey`

**Review by britt6612 — APPROVED** [review]

File: src/js/components/List/stories/RenderedList.js [hunk 2]
```diff
@@ -32,6 +32,7 @@ export const RenderedList = () => (
           {item.location}
         </Text>
       )}
+      itemKey="key"
```
jcfilben, line 35:
```suggestion
      itemKey="entry"
```
Can we change this to map to the entry property in data?

jcfilben (approved):
Looks good!

PR #6957, opened 2023-10-02
Accessiblity Violation for box/Simple storybook #6124

MadhurSikko:
#### What does this PR do?
Solved the Accessibility violation for box/Simple storybook.
The violation was regarding the minimum colour contrast ratio thresholds

/components/Box/stories/Simple.js

`yarn test`

In the storybook, inside the accessibility tab

Screenshot (79)

### DIFF
--- a/src/js/components/Box/stories/Simple.js
+++ b/src/js/components/Box/stories/Simple.js
@@ -11,13 +11,13 @@ export const SimpleBox = () => (
     justify="center"
     align="center"
     pad="xlarge"
-    background="dark-2"
+    background="dark-3"
     gap="medium"
   >
     <Box
       pad="large"
       align="center"
-      background={{ color: 'light-2', opacity: 'strong' }}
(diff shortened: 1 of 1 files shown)


jcfilben (commented):
Thanks for working on this! I think there is value to showing one card with a light color scheme and one with a dark color scheme. Can we keep the left card in a light color scheme but just adjust what light colors we are using to fix the accessibility violations?

MadhurSikko:
> Thanks for working on this! I think there is value to showing one card with a light color scheme and one with a dark color scheme. Can we keep the left card in a light color scheme but just adjust what light colors we are using to fix the accessibility violations?

I have been trying to figure out a combination which keeps the left card light colour and the right card dark colour using the in-build colours (e.g.: light-1 to light-6, dark-1 to dark6) however I have not been able to find such a combination.
https://dequeuniversity.com/rules/axe/4.7/color-contrast?application=axeAPI According to the following website we can add such a configuration however we can't use the predefined colors.

PR #6971, opened 2023-10-13
update minor and patch dependences

britt6612:
#### What does this PR do?
Updates all of the minor and patch dependencies will do major in another PR 
package.json
yarn build 
storybook 


package.json
```diff
@@ -101,29 +101,29 @@
...
     "elliptic": "^6.5.4",
-    "eslint": "^8.48.0",
+    "eslint": "^8.51.0",
     "eslint-config-airbnb": "^19.0.4",
     "eslint-config-prettier": "^9.0.0",
     "eslint-plugin-babel": "^5.3.1",
     "eslint-plugin-import": "^2.28.1",
     "eslint-plugin-jsx-a11y": "^6.7.1",
-    "eslint-plugin-prettier": "^5.0.0",
+    "eslint-plugin-prettier": "^5.0.1",
```
jcfilben, line 114: I know we were at `v5.0.0` previously but can we actually bump this down to `v4.2.1`? I realized that eslint-plugin-prettier v5 is expecting that we have prettier v3 as a peerDependency

britt6612: have we tried upgrading to prettier v3?

jcfilben: Yeah I tried upgrading to v3 last month but it wasn't working

jcfilben (approved):

Issue #6901, opened 2023-07-24
Select searchable: search input text is focussed twice when using keyboard 


Simondin:
Using the Select component with multiple and onSearch props defined the search input text box is focussed twice using the keyboard.

### Expected Behavior

When I open the Select component (with _multiple_ and _onSearch_ props defined) and use the _KeyDown_ to reach the first option I expect the first option stay focussed.

### Actual Behavior

When  I open the Select component and try to reach the first option using the keyboard the focus moves from input text field to first option and the get back to the input text without doing nothing. If I use again the keyboard the issue doesn't occur again.

### URL, screen shot, or Codepen exhibiting the issue

Select story book: 
https://storybook.grommet.io/?path=/story/input-select-search--search
(Input->Select->Search)

### Steps to Reproduce

1. Open the documentation link
2.  Click on the second _Select_ (select multiple options)
3. Use the _KeyDown_ to reach the first option
4. The focus will pass from search text input to the first option and it will get back to the text input.

### Your Environment

- Grommet version: 2.33.1
- Browser Name and version: Chrome 114
- Operating System and version (desktop or mobile): Windows 11

jcfilben:
Resolved by #6982

PR #6982, opened 2023-10-21
Fix Select focus when opening drop with search (#6901)

miqh:
Hello there,

These changes are intended to address #6901.

Happy to amend as necessary.

---

#### What does this PR do?

* Resolves the automatic double focus on the search input of Select described in the linked issue.
* Ensures the focus behaviour within the Select dropdown is consistent when the latter is opened via mouse click or by keyboard control. Currently, opening the dropdown via keyboard control will automatically focus the search input as well as the first option.

Changes should be confined to files for the Select component.

Confirmed the changes manually through Storybook locally. I've also added two additional tests to cover the focus behaviours going forward.

Follow the steps outlined in the linked issue.

(None of the existing tests for Select make use of `userEvent` and so I've followed the surroundings for consistency.)

None.

#6901

Keyboard control:

demo1

Mouse control:

demo2

Unlikely.


src/js/components/Select/__tests__/Select-test.js
```diff
@@ -14,6 +14,7 @@ import { Select } from '..';
 
 describe('Select', () => {
   window.scrollTo = jest.fn();
+  window.HTMLElement.prototype.scrollIntoView = jest.fn();
```
miqh, line 17: I've relocated this here because the tests I've introduced are also affected by this function, which doesn't appear to exist in JSDOM.

I noticed the previous location made no attempt at restoring the original contents, though I suspect it's an inconsequential matter.

src/js/components/Select/SelectContainer.js
```diff
@@ -136,7 +136,7 @@ const SelectContainer = forwardRef(
         const clearButton = clearRef.current;
         if (onSearch) {
           const searchInput = searchRef.current;
-          if (searchInput && searchInput.focus) {
+          if (searchInput && searchInput.focus && !activeRef.current) {
```
jcfilben, line 139: When I tested with only the change on line 118, it appeared to fix the issue so just wanted to know why this change was also necessary

jcfilben (commented):
Thanks for adding a test!
Functionality looks good just a small question

jcfilben:
You can fix the failing bundlesize test by increasing thebundleSize in the `package.json` file from 161 to 162.

```
  "bundlesize": [
    {
      "path": "./dist/grommet.min.js",
      "maxSize": "162 kB"
    }
  ],
```

src/js/components/Select/__tests__/Select-test.js
```diff
@@ -1592,5 +1589,76 @@ describe('Select', () => {
...
+        <Select
+          id="test-select"
+          options={['one', 'two']}
+          onSearch={onSearch}
+          placeholder="test select"
+        />
+      </Grommet>,
+    );
+    const select = getByPlaceholderText('test select');
+    fireEvent.click(select);
```
taysea, line 1606: @jcfilben Do you feel that this test is a candidate to use userEvent instead of fireEvent?

I know userEvent is preferred as a best practice but also has some shortcomings. Wanted to get your opinion before suggesting the PR author to make any changes.

jcfilben: Good question, I think places where we use useFakeTimers we've run into issues with `userEvent`. But there is a chance that `userEvent` could work here. I haven't nailed down the exact situations that it does and doesn't work I just know useFakeTimers + userEvent has caused issues on other tests

taysea: Thanks, Jessica. @miqh would you be open to trying `userEvent` in place of `fireEvent` in your tests? Let us know if you run into any issues related to the points Jessica mentioned above

miqh: @taysea, I've given it a shot, but to no ease of success.

The best I could arrive at was something like the following:

```js
// Test itself is marked async
const user = userEvent.setup();
...
const select = getByPlaceholderText('test select');
await user.click(select);
await waitFor(() => expect(getAllByRole('searchbox')[0]).toHaveFocus());
// Select does not respond to navigation if event is directed at the search input
await user.type(document.getElementById('test-select__select-drop'), '[ArrowDown]');
// Search input does unfocus
await waitFor(() => expect(getAllByRole('searchbox')[0]).not.toHaveFocus());
// But the focus is instead directed at a container element and not the first option
await waitFor(() => expect(getAllByRole('option')[0]).toHaveFocus());
```
I should also add that I tried `user.keyboard`, which fell short of even being able to shift focus (i.e. respond to the arrow down) once the drop appears.

This is presumably because the latter acts on the element in focus. However, as I've noted, dispatching the arrow down key while the search input is in focus causes the Select to not respond (i.e. something must be swallowing up that event?).

If I contrast this with my existing test, that single `fireEvent.keyDown` invocation, plus a delay, was able to correctly (or at least for the purpose of the test) simulate a keyboard navigation step and to also focus the first option as expected.

Any suggestions on your part? I'm not sure what to do at this stage if there's insistence on `userEvent` being used instead.

Otherwise I'll update the bundle size test and call it a day.

jcfilben: Hey @miqh thanks for trying `userEvent`, since it is not working I am okay with us sticking with the original test

src/js/components/Select/SelectContainer.js
jcfilben: Ah gotcha, that make sense. Thanks for the explanation!

jcfilben (approved):
Looks good!

taysea (approved):
LGTM! Thank you for the contribution.

Issue #6925, opened 2023-08-29
unable to style data table group headers


tedinspace:
I would like the group headers of my datatable to be styled differently but have been unable to achieve this

### Expected Behavior

including

` dataTable: {
    groupHeader: {
      background: "red"
    },
  }`
in the grommet custom grommet theme object, should make the group header red 

`<DataTable columns={columns} data={DATA} sortable onSelect={()=>{}} groupBy={"location"}/>`

according to my understanding of the docs
https://v2.grommet.io/datatable#dataTable.groupHeader.background

### Actual Behavior

the group header remains unstyled

### URL, screen shot, or Codepen exhibiting the issue

you can find repo here: [X](https://github.com/tedinspace/grommet-datatable-issue)

grommet theme call: [X](https://github.com/tedinspace/grommet-datatable-issue/blob/main/src/App.tsx)

datatable call: [X](https://github.com/tedinspace/grommet-datatable-issue/blob/main/src/GroupedTable.tsx)

### Steps to Reproduce

1. used create react app with typescript: `npx create-react-app $APP_NAME --template typescript`
2. installed packages `npm install grommet styled-components --save` and `npm install grommet-icons grommet-styles --save`
3. rolled back `styled-components` to version `^5.3.11` for compatibility with grommet 

### Your Environment

I've also tested in a windows dev environment with same result 

- Grommet version: 2.33.1
- Browser Name and version: Chrome Version 116.0.5845.110 (Official Build) (x86_64)
- Operating System and version (desktop or mobile): Mac OS 10.14.6 desktop

jcfilben:
For anyone else looking into this issue here is a codesandbox that reproduces the problem: https://codesandbox.io/s/grommet-v2-template-forked-kwc6xn?file=/index.js

jcfilben:
Resolved by https://github.com/grommet/grommet/pull/6986

PR #3146, opened 2019-06-06
update DataTable groupby to fold only groups with >1 member

dgennetten:
#### What does this PR do?
DataTable groupBy should only fold groups with >1 member
#### Where should the reviewer start?
https://github.com/grommet/grommet/issues/2661
### What testing has been done on this PR?
Chrome/Edge-chromium
#### How should this be manually tested?
storybook sample
#### Any background context you want to provide?
https://github.com/grommet/grommet/issues/2661
#### What are the relevant issues?
There are pre-existing minor layout issues which this PR does not fix
#### Do the grommet docs need to be updated?
No
#### Should this PR be mentioned in the release notes?
Yes
#### Is this change backwards compatible or is it a breaking change?
non-breaking

PR #6986, opened 2023-10-22
Fix groupHeader theming properties for DataTable (#6925)

miqh:
Hello there,

Another pull request, this time aimed at addressing #6925.

A noteworthy side effect of my proposed fix involved updating existing test snapshots for DataTable. In particular, those which used `groupBy` for row grouping. I can only guess that this was because theme properties relating to `groupHeader` were never fed through properly.

---

#### What does this PR do?

Ensures that any theme properties (e.g. https://v2.grommet.io/datatable#dataTable.groupHeader.background) used to customise a DataTable component are incorporated for rendering.

Making sure that the test snapshots I updated are within reason.

I used the available DataTable stories in addition to adding an extra test that covers the provision of `groupHeader` theme properties.

Any DataTable story that involves `groupBy` would be affected by this change (i.e. because even the default `groupHeader` theme properties didn't appear to be passed along).

None.

#6925

This should now be the default appearance when using `groupBy`:

image

Edited the custom theme story for DataTable to demonstrate `groupHeader` theme properties being changed:

image

Likely, as it involves a visual change that would immediately affect anybody using `groupBy` with DataTable.

As mentioned above, this change will alter the default look of DataTable when using `groupBy`, but it's in the interest of allowing people the ability to customise the appearance of row groups if they don't want the default.

### DIFF
--- a/src/js/components/DataTable/DataTable.js
+++ b/src/js/components/DataTable/DataTable.js
@@ -419,7 +419,10 @@ const DataTable = ({
   const bodyContent = groups ? (
     <GroupedBody
       ref={bodyRef}
-      cellProps={{ body: cellProps.body, groupHeader: cellProps.groupHeader }}
+      cellProps={{
+        body: cellProps.body,
+        groupHeader: { ...cellProps.body, ...cellProps.groupHeader },
+      }}
       columns={columns}
       disabled={disabled}
       groupBy={typeof groupBy === 'string' ? { property: groupBy } : groupBy}
--- a/src/js/components/DataTable/DataTable.js
+++ b/src/js/components/DataTable/DataTable.js
@@ -419,7 +419,7 @@ const DataTable = ({
   const bodyContent = groups ? (
     <GroupedBody
       ref={bodyRef}
-      cellProps={cellProps.body}
+      cellProps={{ body: cellProps.body, groupHeader: cellProps.groupHeader }}
       columns={columns}
       disabled={disabled}
       groupBy={typeof groupBy === 'string' ? { property: groupBy } : groupBy}
--- a/src/js/components/DataTable/GroupedBody.js
+++ b/src/js/components/DataTable/GroupedBody.js
@@ -165,7 +165,9 @@ export const GroupedBody = forwardRef(
             } = row;
             const cellProps = normalizeRowCellProps(
               rowProps,
-              cellPropsProp,
+              context === 'groupHeader'
+                ? cellPropsProp.groupHeader
+                : cellPropsProp.body,
               primaryValue,
               index,
             );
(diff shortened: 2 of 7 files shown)


jcfilben (commented):
Nice work, I'm thinking since previously the dataTable.groupHeader content defined in the base theme was not being used let's change it to:

```javascript
groupHeader: {
//  background: undefined,
//  border: undefined,
//  pad: undefined
},
```

This way people using grouped DataTable will be able to customize this but it won't change by surprise when this change gets released.

MikeKingdom (commented):
Like you hint at in the description, for backward compatibility, I think this needs to fallback to use cellProps.body like before if the groupHeader props aren't defined.

image

miqh:
@jcfilben, regarding your suggested adjustment:

> ```js
> groupHeader: {
> //  background: undefined,
> //  border: undefined,
> //  pad: undefined
> },
> ```

This won't preserve existing style compatibility.

The suggested adjustment would instead yield something like the following (take note of the misaligned expansion buttons):

image

The `pad` member needs to be set to the same default as the `body` context (as hinted to by @MikeKingdom as well) because this what's currently passed through. 

In any case, I've adjusted the default theme settings, updating test snapshots as well.

---

While doing this update, I noticed `groupEnd` is present in the thematic variables, but there's no documentation regarding its customisation potential (perhaps right so, because it'll need some extra change to support).

https://github.com/grommet/grommet-site/blob/master/src/screens/DataTable.js

Similar to what's happened with `groupHeader`, the changes (https://github.com/grommet/grommet/pull/3146) which introduced `groupEnd` didn't appear to adjust DataTable accordingly to feed through the correct thematic variables.

If you want this done for this pull request too, I'm able to add more changes.

miqh:
Hi @jcfilben,

I've been cleaning out some stale pull request contributions and noticed this one.

Based on my last comment, I don't believe any further changes on my part are needed (unless I've missed something).

The `groupEnd` matter can be raised as a separate issue if need be.

miqh:
@jcfilben, I feel a bit silly glossing over the point about `cellProps.body` as a fallback.

It was raised much earlier too, so I wasn't reading carefully enough.

So, with the fallback to `cellProps.body`, I've been able to get rid of the default theme value for `dataTable.groupHeader.pad` and for DataTable itself to preserve its presentation.

Hope that does it.

jcfilben:
> @jcfilben, I feel a bit silly glossing over the point about cellProps.body as a fallback.
> 
> It was raised much earlier too, so I wasn't reading carefully enough.
> 
> So, with the fallback to cellProps.body, I've been able to get rid of the default theme value for dataTable.groupHeader.pad and for DataTable itself to preserve its presentation.
> 
> Hope that does it.

No worries! Thanks for getting these changes in, code looks good!

jcfilben (approved):

MikeKingdom (approved):
LGTM! Thanks!

PR #6834, opened 2023-06-07
Reset DateInput textValue on form reset

taysea:
#### What does this PR do?
The MaskedInput inside of DateInput is controlled. Therefore, in an uncontrolled form, when the form is reset, the text value in the DateInput isn't. This PR updates DateInput to listen for the `reset` event to reset the textValue.

#### Where should the reviewer start?
src/js/components/DateInput/DateInput.js

#### What testing has been done on this PR?
Tested with code snippet provided in #6818 

Also tested with:

```
import React from 'react';

import { DateInput, Form, TextInput, Button } from 'grommet';

export const Simple = () => (
  <>
    <Form
      onChange={(value) => {
        console.log(JSON.stringify(value));
      }}
      onReset={(e) => {
        console.log('my reset');
        e.stopPropagation();
      }}
    >
      <DateInput format="mm/dd/yyyy" name="date" />
      <TextInput name="text" />
      <Button label="clear" type="reset" />
    </Form>

    <Form
      onChange={(value) => {
        console.log(JSON.stringify(value));
      }}
    >
      <DateInput format="mm/dd/yyyy" name="date-2" />
      <TextInput name="text-2" />
      <Button label="clear" type="reset" />
    </Form>
  </>
);

Simple.parameters = {
  chromatic: { disable: true },
};

export default {
  title: 'Input/DateInput/Simple',
};

```
Added jest test.

#### Do Jest tests follow these best practices?

- [x] `screen` is used for querying.
- [x] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [x] `userEvent` is used in place of `fireEvent`.
- [ ] `asFragment()` is used for snapshot testing.

#### What are the relevant issues?
Closes #6818 

#### Do the grommet docs need to be updated?
No.

#### Should this PR be mentioned in the release notes?
Yes. Fixed bug where DateInput text value did not reset with Form reset.

#### Is this change backwards compatible or is it a breaking change?
Backwards compatible.

PR #7005, opened 2023-11-02
Stop form submit event from bubbling up

### LINKED ISSUE #6975: DataTable triggers 'Submit' event which propagates to its parent "wizard" component
hi All, I recently started working on React... I created a "wizard" template to take multiple inputs in forms in different steps of the wizard based on the template documented in the HPE design templates-->https://design-system.hpe.design/templates/wizard.
In one of the steps of the wizard, I am using another template called "data table customization" documented at -> https://design-system.hpe.design/templates/datatable-customization
this data table template allows filtering the data based on the filters. But when this data table is placed inside the wizard template, and when I navigate to the filters layer which shows user data to apply the filter, and when the "apply filters"  button is clicked, it triggers a "submit" event which is reaching all the up to the wizard form which is incrementing the active step and moving to the next step.
is there any way or configuration available in the component (I believe the data table is using a data filters component to filter the data) to stop this event from reaching its parent, so that only the filters get applied and the related data is shown in the table?
Please let me know if there are any other issues that I could be overlooking?

### Expected Behavior

the submit event should not propagate to the wizard . It should stop at the data table level and show the selected filtered result

### Actual Behavior

The submit event on data table template triggers submit event which propagates to its parent wizard component . This submit event in wizards moves the wizard to the next step which is not desired.

taysea:
#### What does this PR do?

Adds logic to Form onSubmit handler to ensure there are no undesired form submissions in cases where a portal containing a form (such as DataFilters as layer, Layer) is the child of another form in the React tree.

From React docs: "An event fired from inside a portal will propagate to ancestors in the containing React tree, even if those elements are not ancestors in the DOM tree."

This approach ensures that the submission event was related to the form itself before advancing with the form submission logic. Follows pattern used in this PR: https://github.com/grommet/grommet/pull/6834

Additionally, this approach ensures the event still bubbles up to the window, which can be relevant for use cases like analytics that listen for events at the window.

https://github.com/grommet/grommet/assets/12522275/2e90fc63-df5d-4fef-adbd-1f9a5f5e39a7

Tested locally in storybook by editing the Full Layer story:

```

import {
  Box,
  Button,
  Layer,
  Grommet,
  Form,
  FormField,
  TextInput,
  grommet,
} from 'grommet';

export const FullLayer = () => {
  const [open, setOpen] = useState(false);

  React.useEffect(() => {
    const handleSubmit = (e) => console.log('hey', e);
    document.addEventListener('submit', handleSubmit);

    return () => document.removeEventListener('submit', handleSubmit);
  });

  return (
    <Grommet
      theme={grommet}
      // onAnalytics={(value) => console.log(value)}
    >
      <Box align="start" gap="large">
        <Form id="wizard-form" onSubmit={({ value }) => console.log(value)}>
          <FormField label="Label" htmlFor="text-input" name="text-input">
            <TextInput id="text-input" name="text-input" />
          </FormField>
          <Button label="Open layer" secondary onClick={() => setOpen(true)} />
          {open ? <LayerExample setOpen={setOpen} /> : null}
        </Form>
        <Button label="Submit" type="submit" primary form="wizard-form" />
        {/* {open ? <LayerExample setOpen={setOpen} /> : null} */}
      </Box>
    </Grommet>
  );
};

const LayerExample = ({ setOpen }) => (
  <Layer onEsc={() => setOpen(false)}>
    <Form onSubmit={({ value }) => console.log(value)}>
      <FormField label="Label 2" htmlFor="text-input-2" name="text-input-2">
        <TextInput name="text-input-2" id="text-input-2" />
      </FormField>
      <Button label="Submit" primary type="submit" />
    </Form>
  </Layer>
);

FullLayer.storyName = 'Full';

FullLayer.parameters = {
  chromatic: { disable: true },
};

FullLayer.args = {
  full: true,
};

export default {
  title: 'Layout/Layer/Full',
};

```

Original issue filed had a scenario of a Wizard (which contains a form that wraps each step) where one wizard step contained a data collection and toolbar which would open a layer of filters. Clicking "Apply filters" (which would submit the DataFilters form) was causing a submit of the wizard form also.

Closes #6975 

No.

Yes.

Backwards compatible.


MikeKingdom (commented):
I think this is probably the right thing, but I'm curious if anyone would register a submit handler somewhere in the tree above this expecting it to bubble. It's a technique used by things like analytics. Any chance we can just make Form recognize if this event is against this form?

halocline (approved):

MikeKingdom (approved):
Thanks Taylor!

jcfilben (approved):
Looks good!

RohanEmmanuel07:
Hi @taysea @jcfilben ,
i had raised this bug as discussed on the Grommet slack channel, now with the fix when I try out the filters, i don't see the wizard containing the data table template moving to the next step. But when the search box is used to search the data and Enter Key is pressed , this triggers the submit event which is reaching the parent (the wizard in this case) and this is making the wizard move to  the next step. any thoughts on this?

taysea:
> any thoughts on this?

Hi @RohanEmmanuel07 Could you please provide a codesandbox demonstrating the search box behavior?

RohanEmmanuel07:
hi @taysea ,
Thanks for responding. I reused the sandbox that you had created --> https://codesandbox.io/s/interesting-mayer-vgdpgf?file=/src/App.js

jcfilben:
@RohanEmmanuel07 thanks for reporting this. Some additional notes, when I am on a screen size medium or above, hitting 'Enter' while focus is on the search box is resulting in the Wizard moving to the next step. If I am on a small screen size and the search is displayed in a drop hitting 'Enter' doesn't result in anything and there is not way for me to actually search the data.

I'll create a separate issue for this so we can get this resolved.

PR #7007, opened 2023-11-04
added 500ms delay before tooltip is closed

### LINKED ISSUE #6998: Tip - add slight delay before tooltip is closed
In https://github.com/grommet/grommet/pull/6954 the behavior of Tip was adjusted so that the tooltip stays open when the cursor is moved over the tooltip. In cases where the Tip has a slight gap between the Tip target and Tip content, when you try and move the mouse over the Tip content it closes too soon. Maybe we can add a slight delay to the Tip closing.

https://storybook.grommet.io/?path=/story/controls-tip-simple--simple&globals=theme:hpe

### Steps to Reproduce

1. In the storybook example above try and move your cursor over the tooltip content and notice that it closes before you can move your cursor there.

HardikChoudhary24:
#### What does this PR do?
Fixes #6998
- Added 500ms delay before tooltip is closed.

- In the stories

- Ran locally on storybook.

- Previously, the tooltip would close immediately when the mouse left.

- #6998 

- yes


taysea:
Thanks for the contribution! I'm thinking about if a timeout is the best approach for this or if there's any way to make the component smarter to know the mouse is in the region between the control and the tip.

For example, APG (an accessibility best practices group through WCAG) has been exploring a pattern that leverages CSS :after selector to have an invisible block for the region between the control and the tip: https://codepen.io/Moiety/pen/LaPvWy

I'm wondering if there's a way to explore a similar approach in Grommet. Right now, the space between the control and the tip is driven by a margin value in the theme. Is there anyway to use that value and incorporate that into the broader "click target" so that if the mouse is a) over the control, b) over the tip, or c) in the space between the tip and control, the tip will remain visible?

jcfilben:

jonscottclark:
@taysea That suggestion would be amazing, we're running up against this issue now

I think the `dropProps` of the Tip would need to inform the position of the `::after` relative to the tip trigger. Here's a little screenshot from the [updated demo of the APG pattern](https://zoebijl.github.io/apg-tooltip/) you referenced.

Screenshot 2024-11-07 at 9 28 21 AM

PR #7010, opened 2023-11-07
Enhance Data screen reader announcements and add search debounce

### LINKED ISSUE #6630: Data Component Enhancing Accessibility
Enhancing Data component search and filter to meet better screen reader 
### Expected Behavior

- [x] After pressing "enter" in the search it should read out "number of results found" or "no results found"
- [x] On the filters button have an `aria-label` to let the user know how many filters are applied. (This has already been enhanced)
- [x] When clicking on `apply filters` it would be great if the results were ready out "number of results found" or "no results found"

### Actual Behavior
1. After pressing  "enter" in the search nothing came back on the screen reader so it was not aware if there were results unless the user arrows down.
2. On the filters button, it does not let you know how many filters are applied.
3.  When clicking on `apply filters` nothing came back on the screen reader so it was not aware if there were results unless the user arrows down.

### Steps to Reproduce
Use JAWS or voiceover
1. Type something into the search and press enter. Nothing is read out by the screenreader.
2. Focus on the filters button. No information about the number of filters applied is read out.
3. Press the apply filters button. No information about the # of results is read out.

https://storybook.grommet.io/?path=/story/layout-data-list--example

### Your Environment

- Grommet version:
- Browser Name and version:

taysea:
#### What does this PR do?

This PR enhances Data to add announcements to screen reader users how the number of results has changed based on filters applied or search text.

The general approach is that a `useEffect` is added to Data that listens for changes to the `filteredTotal` or `total` of the Data context value. Anytime these values change, AnnounceContext is used to announce the number of results in the format that mirrors that of DataSummary.

AnnounceContext is intentionally used in a `useEffect` at the parent Data level as opposed to an aria-live directly on DataSummary `Text` in order to ensure that even if a caller does not display DataSummary visually on the screen that a screen reader user will still receive the announcement.

Additionally, this PR adds debounce for DataSearch with a delay of 300ms. This helps to ensure that the announcement is not made after every keystroke but instead will wait for a short period of inactivity. 300ms was chosen empirically and through reference of other search debounce examples.

Due to the differences of screen reader behaviors, there are limitations to the consistency that the "1 result of 8 items" (for example) message will be announced. This is due to a) other announcements that may already be read by the screen reader, b) the internal logic of the screen reader and how it manages its queue of messages/clears it out after a certain amount of time. For example, the pattern implemented sometimes is read out properly by VoiceOver but sometimes doesn't announce due to the aria-label on the Filter button being read out first. On the other hand, ChromeVox plugin is fairly consistent at announcing the "1 result of 8 items" message.

That being said, I think this pattern is in best alignment with aria best practices as possible, so we should still move forward with it.

src/js/components/Data/Data.js

Locally with Data stories.

Using Data stories + screen reader.

NOTE: Adjusted tests that were failing to use `fireEvent` due to repeated timeout issues locally and in Circle CI. The snapshot changes make sense given the state of the components when the snapshots are taken.

Closes #6630 

No.

Yes.

Backwards compatible.


jcfilben (approved):
Looks good!

src/js/components/Data/Data.js
```diff
@@ -51,6 +53,41 @@ export const Data = ({
     return filter(dataProp, view, properties);
   }, [dataProp, filteredTotal, onView, properties, total, view]);
 
+  const announce = useContext(AnnounceContext);
+  const { format } = useContext(MessageContext);
+  useEffect(() => {
```
halocline, line 58: Add some sort of comment explaining the intent behind this useEffect

```suggestion
  // Announce to screen readers when search or filters are applied and affect the underlying result set
  useEffect(() => {
```

halocline (commented):
Looks good. Small request to add comment to useEffect

src/js/components/Data/DataForm.js
```diff
@@ -229,6 +229,23 @@ const normalizeValue = (nextValue, prevValue, views) => {
   return result;
 };
 
+// 300ms was chosen empirically as a reasonable default
+const DEBOUNCE_TIMEOUT = 300;
```
halocline, line 233: If needed in the future, is there an approach where we can we allow the caller to specify the debounce value if desired? I believe there is, but just confirming.

MikeKingdom: Theme maybe?

halocline: > Theme maybe?

As a default, I think the theme could make sense. Was wondering about specific instances where the performance of the API being interacted with might also be a consideration and the implementor might want control to tune for that instance. Thoughts @MikeKingdom ?

halocline: Not saying this needs to be implemented in this PR, rather just a line of sight to how we might approach if needed.

MikeKingdom: Yeah definitely valid. Might be worth a prop

taysea: Yeah I think there are a variety of approaches we could take to allow caller to drive (whether via prop, or theme, etc.). And then code would just modify where default value is used to something like `propValue || DEBOUNCE_TIMEOUT`

halocline (approved):

Issue #7002, opened 2023-11-01
VoiceOver doesn't always navigate Radio Button Group successfully 


phinch:
Hi! When using VoiceOver with a radio button group, and using VoiceOver commands to navigate through the labels, we're running into some strange behavior where the focus is sometimes forced back to the first radio button. I work on an enterprise software team, and our accessibility auditors have told us that this is violating our accessibility requirements :( 

### Expected Behavior

When viewing a radio button group, the user should be able to scan through the radio buttons and labels by using the VoiceOver arrow key commands.

### Actual Behavior

When moving through the radio button group using the VoiceOver VO+→ and VO+←, the focus occasionally jumps back to the first radio button, which takes the user off of their focused element and back to the beginning of the radio group.

I'm wondering if the root cause is [this imperative focus behavior in the radio group button](https://github.com/grommet/grommet/blob/master/src/js/components/RadioButtonGroup/RadioButtonGroup.js#L71-L77)...it seems like this effect would fire often, since i think onFocus and onBlur are firing often as we move through the component in VO, and since `valueIndex` is 0 when the first option is selected, we would satisfy the condition where `(focus && !valueIndex)` and force focus back to the first radio button? But i'm not sure i'm fully understanding all the behavior there.
![image](https://github.com/grommet/grommet/assets/9158038/e886a133-2463-43ed-9268-399ea528d4a4)

### URL, screen shot, or Codepen exhibiting the issue

I was able to reproduce this issue on the [RadioGroup example on the Grommet website](https://v2.grommet.io/radiobuttongroup?) The steps might be hard to communicate because it requires VoiceOver (and tbh i'm not a VO expert), so I'll also attach a video that hopefully shows the behavior.

### Steps to Reproduce

1. Focus on the radio group
1. Use VO+→ and and VO+← to move through the buttons and labels of the component
1. Occasionally, notice that the focus shifts back to the first radio button

#### Reproduction example

At 0:08 and again at 0:19, you can see the focus jumps back to the first radio button even though i'm scanning the second element of the radio group

(As a bonus, our accessibility auditors pointed out that ideally, we should add `focusable=false aria-hidden=true` to the "checked" icon that shows up when a button is selected, so that the screenreader doesn't read out the additional "image" text when scanning through the options - but that's less critical than the forced focus issue)

https://github.com/grommet/grommet/assets/9158038/401150af-d064-4925-aeef-0522732a2996

### Your Environment

- Grommet version: latest
[...]

PR #7011, opened 2023-11-08
Fix focus issue on RadioButtonGroup

jcfilben:
#### What does this PR do?
Resolves an issue with RadioButtonGroup. When using a screenreader and navigating through the RadioButtonGroup options sometimes the focus jumps back to the first radio button.

This PR removes an unnecessary useEffect. The useEffect was added to ensure that when a RadioButtonGroup doesn't have a value and it is focused, the first option will receive focus. This behavior is already covered here: https://github.com/grommet/grommet/blob/master/src/js/components/RadioButtonGroup/RadioButtonGroup.js#L156-L158. The useEffect is unnecessary and introduces the screenreader issue.

Closes https://github.com/grommet/grommet/issues/7002

no
yes
backwards compatible

### DIFF
--- a/src/js/components/RadioButtonGroup/RadioButtonGroup.js
+++ b/src/js/components/RadioButtonGroup/RadioButtonGroup.js
@@ -1,10 +1,4 @@
-import React, {
-  forwardRef,
-  useContext,
-  useEffect,
-  useRef,
-  useState,
-} from 'react';
+import React, { forwardRef, useContext, useRef, useState } from 'react';
 
 import { ThemeContext } from 'styled-components';
 import { FormContext } from '../Form/FormContext';
@@ -67,14 +61,6 @@ const RadioButtonGroup = forwardRef(
       });
       return result;
     }, [options, value]);
-
-    useEffect(() => {
-      // if tab comes back to RadioButtonGroup when there still is no selection,
-      // we want focus to be on the first RadioButton
-      if (focus && !valueIndex) {
-        optionRefs.current[0].focus();
-      }
-    }, [focus, valueIndex]);
 
     const onNext = () => {
       if (valueIndex !== undefined && valueIndex < options.length - 1) {


src/js/components/RadioButtonGroup/RadioButtonGroup.js
```diff
@@ -68,14 +62,6 @@ const RadioButtonGroup = forwardRef(
...
     }, [options, value]);
 
-    useEffect(() => {
-      // if tab comes back to RadioButtonGroup when there still is no selection,
-      // we want focus to be on the first RadioButton
-      if (focus && !valueIndex) {
-        optionRefs.current[0].focus();
-      }
-    }, [focus, valueIndex]);
-
```
ShimiSun, line 78: Nice cleanup. Did it pass all the tests w/o it? Maybe we can try deploying it on the DS to ensure it works as part of an integral solution?

taysea: I just tested using keyboard for a RadioButtonGroup on its own as well as a RadioButtonGroup in a FormField. The behavior seems to be working as expected for Chrome, Firefox, Safari, and with VoiceOver. I wish I could remember why in the initial PR I added the useEffect since it seems like my other addition to `focusable` logic was sufficient.

I am comfortable with removing the useEffect.

taysea (approved):

halocline (approved):
🎉  Yay! Much needed.

src/js/components/RadioButtonGroup/RadioButtonGroup.js
jcfilben: Hey @ShimiSun! Good questions, just confirming I also tested the keyboard behavior across browsers and with a RadioButtonGroup within and without Form/FormField.

ShimiSun: Wonderful! Thank you ladies. Missed you - you two superwomen!

PR #7013, opened 2023-11-12
Fix tip truncate mode for Text using dynamic content (#6910)

### LINKED ISSUE #6910: Style to show/hide tip on truncate='tip' in Text tag doesn't work on content update
Using below code doesn't show tip or hid tip, if name state updated in the background.

```
<Box width={{ max: "120px" }}>
    <Text truncate="tip">{name}</Text>
</Box>
```
Here is the code with issue reproduced.
https://codesandbox.io/s/grommet-v2-template-forked-8zf93c?file=/MyApp.js

image

### Steps to Reproduce

1. Have above code as part of the component.
2. Try updating the name state to in the useEffect in the background. Either update it from small name to big or big name to small.
3. If updated small to big name, tip should appear
4. If updated big to small name, tip should not appear

You can download this code also:
[grommet-v2-template-forked - issue reproduced.zip](https://github.com/grommet/grommet/files/12192135/grommet-v2-template-forked.-.issue.reproduced.zip)

### Your Environment

Here is the code with issue reproduced.
https://codesandbox.io/s/grommet-v2-template-forked-8zf93c?file=/MyApp.js

Tried with both 2.24.0 and 2.33.1 version of grommet.

miqh:
Hello,

This is a minor adjustment that's intended to resolve #6910.

My understanding of why this particular issue happens is because the truncation testing logic inside Text isn't being triggered when it should be (i.e. when text content gets updated).

Note that even without my proposed change here, the problem resolves itself when a user triggers a viewport resize event. That's because the latter is another way the truncation testing logic may be retriggered.

(I should also add that at this time of writing, tests on `master` are failing, even without my proposed change.)

---

#### What does this PR do?

Ensures that Text usages with `truncate="tip"` display a tooltip when its content changes (e.g. stateful text being updated).

By having a look at the Text component implementation.

I sunk a big chunk of time into investigating how to write a test to replicate the problematic scenario. Alas, I wasn't able to arrive at something satisfactory. I saw two sticking points:

1. Truncation testing logic inside Text relies on layout properties (e.g. `textRef.current.scrollWidth`), which I don't believe jsdom copes with.
2. It's not clear to me where the tooltip (i.e. Drop) attaches, and it looks too difficult to hijack and mock refs that deep down.

The best assurance I can provide with my change is through attempting to reproduce the issue with a custom story (see below).

Otherwise, I'm happy to be schooled on how might to get a test going for this.

I'm unsure the following story merits a permanent place, which is why I've not included it with the fix.

However, one should be able to copy and paste the story to test locally if desired:

```js
import React from 'react';

import { Box, Text } from 'grommet';

export const DynamicTruncateTip = () => {
  const [text, setText] = React.useState('');
  React.useEffect(() => {
    setText('expecting this text to be truncated');
  });
  return (
    <Box width="small">
      <Text truncate="tip">{text}</Text>
    </Box>
  );
};

DynamicTruncateTip.parameters = {
  chromatic: { disable: true },
};

export default {
  title: `Type/Text/DynamicTruncateTip`,
};
```

N/A.

#6910

Don't think so, it's a bug fix.

As above.

I don't believe there should be compatibility issues.

### DIFF
--- a/src/js/components/Text/Text.js
+++ b/src/js/components/Text/Text.js
@@ -53,7 +53,7 @@ const Text = forwardRef(
         window.removeEventListener('resize', updateTip);
         window.removeEventListener('pagechange', updateTip);
       };
-    }, [textRef, truncate]);
+    }, [textRef, children, truncate]);
 
     if (skeleton) {
       return (


jcfilben (commented):
Thanks for working on this!
It looks like our linter is flagging an issue with this solution:
`React Hook useLayoutEffect has a missing dependency: 'textRef'. Either include it or remove the dependency array`

miqh:
@jcfilben,

Ah, my bad for missing the lint! I'd assumed the `test` command implicitly covered linting as well.

In any case, I should explain why I redacted `textRef` originally: it's because I don't believe it ever changes once the `Text` component mounts (i.e. the `StyledText` wrapping the tooltip content, which on the contrary can possibly change), so inclusion in the dependencies array for `useLayoutEffect` seemed inconsequential to me.

MikeKingdom (commented):
Seems like it could use a comment about the children dependency since it's not quite obvious why it's there. Apart from that it seems reasonable.
Having the textRef as a dependency is good since it makes lint happy even though we don't think it is required.

PR #7020, opened 2023-11-17
fix datachart bug for single chart data

### LINKED ISSUE #6991: undefined error of DataChart horizontal bar when the input length is 1
### Expected Behavior
When the input length of data is 1, the data chart should render a single horizontal bar.

### Actual Behavior
undefined error: Cannot read properties of undefined

https://codesandbox.io/s/grommet-v2-template-forked-tjjym6?file=/index.js
image

### Steps to Reproduce
- Choose `direction='horizontal'` in DataChart
- Declare the y column in `series` and `axis`.
- if the length of data is 1, it will be an undefined error, if the length is larger than 1, the rendered date is wrong.

### Your Environment
- Grommet version: 2.33.1
- Operating System and version (desktop or mobile): Win 10

Sulaymon333:
#### What does this PR do?
This pull request is meant to fix datachart bug with single data item

closes #6991

No

Yes

Backward compatible

### DIFF
--- a/src/js/components/DataChart/DataChart.js
+++ b/src/js/components/DataChart/DataChart.js
@@ -527,7 +527,7 @@ const DataChart = forwardRef(
       series.forEach(({ property, render }) => {
         if (
           !render &&
-          data.length > 1 &&
+          data.length >= 1 &&
           typeof data[0][property] === 'string'
         ) {
           result[property] = createDateFormat(
@@ -547,7 +547,8 @@ const DataChart = forwardRef(
         if (!property || (!horizontal && y) || (horizontal && !y)) {
           if (render) return render(value);
         } else {
-          const datum = data[axisValue];
+          // const datum = data[axisValue];
+          const datum = data[0];
           value = datum[property];
           if (render) return render(value, datum, property);
         }
--- a/src/js/components/DataChart/utils.js
+++ b/src/js/components/DataChart/utils.js
@@ -78,7 +78,13 @@ export const createDateFormat = (firstValue, lastValue, full) => {
   ) {
     const delta = Math.abs(endDate - startDate);
     let options;
-    if (delta < 60000)
+    if (delta === 0)
+      // equal 0 - single data point
+      options = {
+        month: full ? 'short' : 'numeric',
+        day: 'numeric',
+      };
+    else if (delta < 60000)
       // less than 1 minute
       options = full
         ? {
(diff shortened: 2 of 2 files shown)


src/js/components/DataChart/DataChart.js
```diff
@@ -547,7 +547,8 @@ const DataChart = forwardRef(
         if (!property || (!horizontal && y) || (horizontal && !y)) {
           if (render) return render(value);
         } else {
-          const datum = data[axisValue];
+          // const datum = data[axisValue];
+          const datum = data[0];
```
jcfilben, line 551: This change causes an issue with the axis labels. If we have the following code we end up with two xaxis labels that are the same value (Jun 23) coming from data[0].

```
const data = [
  { a: 1, b: 'one', c: 111111, d: '2020-06-24' },
  { a: 2, b: 'two', c: 222222, d: '2020-06-23' },
];

export const Simple = () => (
  <Box align="center" justify="start" pad="large">
    <DataChart
          data={data}
          series={['d', 'a']}
          axis={{
            x: { property: 'd' },
            y: { property: 'a' },
          }}
        />
  </Box>
);
```
Screen Shot 2023-11-20 at 12 32 56 PM

src/js/components/DataChart/DataChart.js
```diff
@@ -527,7 +527,7 @@ const DataChart = forwardRef(
       series.forEach(({ property, render }) => {
         if (
           !render &&
-          data.length > 1 &&
+          data.length >= 1 &&
```
jcfilben, line 530:
```suggestion
          data.length > 1 &&
```
I think we should revert this change back. If we only have 1 data point we don't need to the createDateFormat code

src/js/components/DataChart/utils.js
```diff
@@ -78,7 +78,13 @@ export const createDateFormat = (firstValue, lastValue, full) => {
...
     const delta = Math.abs(endDate - startDate);
     let options;
-    if (delta < 60000)
+    if (delta === 0)
+      // equal 0 - single data point
+      options = {
+        month: full ? 'short' : 'numeric',
+        day: 'numeric',
+      };
+    else if (delta < 60000)
```
jcfilben, line 87:
```suggestion
    if (delta < 60000)
```
Lets revert this change back, I don't think we need to adjust this

src/js/components/DataChart/utils.js
```diff
@@ -96,7 +102,7 @@ export const createDateFormat = (firstValue, lastValue, full) => {
     else if (delta < 86400000)
       // less than 1 day
       options = { hour: 'numeric' };
-    else if (delta < 2592000000)
+    else if (delta < 2592000000 || delta === 0)
```
jcfilben, line 105:
```suggestion
    else if (delta < 2592000000)
```
Let's revert this change as well

britt6612:

PR #7026, opened 2023-11-20
Make DataFilter range inclusive of bounds

jcfilben:
#### What does this PR do?
Changes the DataFilter behavior so that if the filter is a range the results will be inclusive of the range bounds. For example if the range bounds are min: 2, max: 5 anything between 2 and 5 will be included in the filtered results and 2 and 5 themselves will also be included.

https://github.com/grommet/grommet/issues/6730

no
yes
backwards compatible


src/js/components/DataFilter/__tests__/DataFilter-test.tsx
```diff
@@ -271,6 +273,30 @@ describe('DataFilter', () => {
...
+
+    const lowerBound = screen.getByRole('slider', { name: 'Lower Bounds' });
+    act(() => {
+      lowerBound.focus();
+    });
+    fireEvent.keyDown(lowerBound, { key: 'Right', keyCode: 39 });
+    const applyFiltersButton = getByRole('button', { name: 'Apply filters' });
+    fireEvent.click(applyFiltersButton);
+
+    expect(screen.queryByText('1')).not.toBeInTheDocument();
```
taysea, line 296:
```suggestion
    expect(screen.queryByText('1')).not.toBeInTheDocument();
    expect(screen.queryByText('2')).toBeInTheDocument();
```
add this to ensure the "inclusive" aspect is checked?

taysea (commented):
one small comment on the test, otherwise this looks good

src/js/components/DataFilter/__tests__/DataFilter-test.tsx
```diff
@@ -271,6 +273,30 @@ describe('DataFilter', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
+  test('range inclusive', () => {
```
taysea, line 276: can we make a slightly more descriptive test name? like "includes range min and max in filtered results"

src/js/components/DataFilter/__tests__/DataFilter-test.tsx
jcfilben: Added this in latest commit, I also removed the snapshot since I think the other assertions are effectively testing the behavior we want

src/js/components/DataFilter/__tests__/DataFilter-test.tsx
jcfilben: Yes, done in latest commit

taysea (approved):
LGTM!

britt6612 (approved):

Issue #7030, opened 2023-11-22
RangeSelector -- floating point step issue


taysea:
### Expected Behavior

RangeSelector should support floating point step cleanly.

### Actual Behavior

When step is a value like `0.15`, for example, there can be resulting issues with the floating point addition that results in an undesired value. As this compounds, it can lead to cases where the keyboard can't get back to the "min".  Notice the console log statement in below video which is the [min, max], step.

### URL, screen shot, or Codepen exhibiting the issue

https://github.com/grommet/grommet/assets/12522275/0b96d2c3-a288-47c7-9306-517d7a817f6e

### Steps to Reproduce

1. Step min to 0, max to 3
2. Set step to 0.15
3. Set `label` true on RangeSelector to see bounds
4. Use arrow keys to move lower bound to the right a handful of times

### Your Environment

- Grommet version: 2.34.1
- Browser Name and version: MacOS Chrome
- Operating System and version (desktop or mobile):

joaoGabriel55:
I can work on that

taysea:
Closed by https://github.com/grommet/grommet/pull/7033

PR #7031, opened 2023-11-24
RangeSelector - floating point step issue

joaoGabriel55:
#### What does this PR do?

When step is a value like 0.15, for example, there can be resulting issues with the floating point addition that results in an undesired value. As this compounds, it can lead to cases where the keyboard can't get back to the "min". Now RangeSelector supports floating point step cleanly.

```
http://localhost:9001/?path=/story/input-rangeselector-floatstep--float-step
```

Related issue: https://github.com/grommet/grommet/issues/7030


src/js/components/RangeSelector/RangeSelector.js
```diff
@@ -63,8 +63,10 @@ const RangeSelector = forwardRef(
 
     const change = useCallback(
       (nextValues) => {
-        setValues(nextValues);
-        if (onChange) onChange(nextValues);
+        const newRange = nextValues.map((n) => Number(Number(n).toFixed(2)));
```
taysea, line 66: Thanks for looking at this issue! Could there be issues with selecting `2` as the fixed length/are we being opinionated? What if the step is 0.135? etc.

joaoGabriel55: Do you think that it would be good to capture the number of decimal places from step number informed?

PR #7033, opened 2023-11-27
Improve DataFilter and RangeSelector handling of floating point values

taysea:
#### What does this PR do?

This PR resolves a handle of bugs in DataFilter and RangeSelector related to floating point values as well as enhances some of the keyboard behaviors of RangeSelector to better align with the mouse behavior.

The modifications in DataFilter aim to do floating point calculations in integers then to restore the original order of magnitude if necessary. When a caller does not define a min/max for their DataFilter range, DataFilter makes some "smart" decisions to look through the data and determine a friendly min/max. From that, DataFilter divides that range by 20 to determine the default step. In the case of floating point math in Javascript, something benign like `2.8 / 20` returns `0.13999999999999999` as opposed to `0.14`. The adjustments to the code instead do the calculation as `28 / 200`.

The adjustments to the floating point addition aligns with something the browser does on a generic input of type="number" with step="0.15" where 0.3 + 0.15 renders cleanly as 0.45 as opposed to 0.44999999999999996

The modifications in RangeFilter fixes a bug to ensure that:
- As a keyboard user, I can always reach my min and max again. Previously due to floating point errors or cases where the default value might not have been on an exact interval of "step", there were cases where the keyboard could not reach the min/max. This is achieved by modifying the `change` function.

Before

https://github.com/grommet/grommet/assets/12522275/34d7fd52-c1cc-47f0-a62c-80c2e1d3173a

After

https://github.com/grommet/grommet/assets/12522275/a36dd42f-e9b7-40fb-9e26-73cd2f59815c

Before

Screen Shot 2023-12-06 at 2 55 37 PM

After

Screen Shot 2023-12-06 at 3 01 09 PM

src/js/components/DataFilter/DataFilter.js

Tested in storybook

storybook

Snapshot changes show expected, improved behavior.

Related to #6730 

Yes.

Yes.

Backwards compatible.


jcfilben (commented):
Code looks good!

Just flagging an edge case. If we have something like `min: 10.3, max: 50.7, step: 1` the min and max will be rounded and not display a decimal point. I think I am okay with this behavior for know knowing that later down the road we will likely add more control over how the label is displayed. Could be an issue for currency though. Maybe if we have zero padding in the step we should respect that? If we have something like `min: 10.00, max: 50.00, step: 1.00` this should be displayed with two decimal points. Right now the zero padding is ignored. 

Not sure if it is worth doing this at the moment. I would be okay with moving forward with this PR as is.

MikeKingdom (commented):
This looks good. Although I wonder if given how wonky precision of a float can be in javavscript, if a better default behavior (when no step is given) would be to make a step that is a more integral value based on the range. E.g. if you divide the range by 20 or whatever you can get a large number of digits still for some min/max values. Can we be a little more picky about only 2 significant digits? Like if you get .0033333333 can that be cut down to .0033? Maybe this is still a bit of a corner case but it still seems like we have less usuable situations

src/js/components/DataFilter/__tests__/__snapshots__/DataFilter-test.tsx.snap
```diff
@@ -1000,17 +1000,17 @@ exports[`DataFilter noForm 1`] = `
             </div>
             <div
               class="c19"
-              style="flex: 3.8 0 0px; cursor: ew-resize;"
+              style="flex: 3.8000000000000003 0 0px; cursor: ew-resize;"
```
taysea, line 1003: Note to reviewers, the floating point here I think is something specific to jest. In the browser, this renders as 3.8:

Screen Shot 2023-12-06 at 2 04 26 PM

jcfilben (approved):
Nice work! Code looks good and I tested various cases with strange values as the min, max, and step

PR #7035, opened 2023-11-28
When RangeSelector is reset to min/max set value to undefined 

jcfilben:
#### What does this PR do?
When RangeSelector is reset so that the value is equal to the min and max set the value to undefined. In a DataFilters context this ensures that the DataFilter doesn't show a filter being applied when the range is set to the min/max

https://github.com/grommet/grommet/issues/6730

no
yes


jcfilben:
Closing, I don't think this approach is backwards compatible

PR #7038, opened 2023-11-29
add aria-label for inputs

britt6612:
#### What does this PR do?
This PR adds the `aria-label` to the inputs in the DataFilter input 

HTML best practice for inputs is to provide labels, a fundamental principle outlined in the WCAG guidelines. It is possible to make an input accessible without the visible label. We can do this by using an Aria-label. The goal is to make the input clear to all users.

datafilter.js
updated snapshots
use voice over on mac for how the inputs are read out for keyboard users. 

When DataFilter is used alone depending on the data it will use a `checkboxgroup` or a `range` these however do not have `labels` because in some cases they may be used in a `toolbar` for design purposes the `label` is left off. In order to still have these inputs accessible they need to have an `aria-label` associated with the input. 
Data&Friends
Accessibility -- When DataFilter isn't wrapped in DataFilters, there's no label. Should it always be wrapped in DataFilters?
no
no

### DIFF
--- a/src/js/components/DataFilter/DataFilter.js
+++ b/src/js/components/DataFilter/DataFilter.js
@@ -121,7 +121,7 @@ export const DataFilter = ({
   // only add aria-label for no form examples
   const ariaLabel = noForm
     ? `${properties?.[property]?.label || property}`
-    : null;
+    : undefined;
 
   let content = children;
   if (!content) {
--- a/src/js/components/DataFilter/DataFilter.js
+++ b/src/js/components/DataFilter/DataFilter.js
@@ -123,6 +123,7 @@ export const DataFilter = ({
     if (range) {
       content = (
         <RangeSelector
+          aria-label={property}
           id={id}
           name={`${property}._range`}
           defaultValues={range}
@@ -144,10 +145,22 @@ export const DataFilter = ({
       if (options.length === 2 && options[1] === true && options[0] === false) {
         // special case boolean properties
         content = (
-          <CheckBoxGroup id={id} name={property} options={booleanOptions} />
+          <CheckBoxGroup
+            aria-label={property}
+            id={id}
+            name={property}
+            options={booleanOptions}
+          />
         );
       } else if (options.length <= maxCheckBoxGroupOptions) {
-        content = <CheckBoxGroup id={id} name={property} options={options} />;
+        content = (
+          <CheckBoxGroup
+            aria-label={property}
+            id={id}
+            name={property}
+            options={options}
+          />
+        );
       } else {
         content = (
           <SelectMultiple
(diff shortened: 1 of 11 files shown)


src/js/components/DataFilter/DataFilter.js
```diff
@@ -123,6 +123,7 @@ export const DataFilter = ({
     if (range) {
       content = (
         <RangeSelector
+          aria-label={property}
```
MikeKingdom, line 126: Shouldn't this do what the FormField would do for label and try to use a label defined in the properties? In many real world cases that property name isn't all that friendly/usable.
```jsx
label={properties?.[property]?.label || property}
```

MikeKingdom (commented):
Should this only be in the `noForm` case? I would think the FormField `htmlFor` would take care of the connection to the label for screen readers.

src/js/components/DataFilter/DataFilter.js
britt6612: good point yes the `property` isnt always the best

src/js/components/DataFilter/DataFilter.js
```diff
@@ -118,11 +118,17 @@ export const DataFilter = ({
 
   const id = `${dataId}-${property}`;
 
+  // only add aria-label for no form examples
+  const ariaLabel = noForm
```
britt6612, line 122: open to naming suggestion if we wanted to change here

MikeKingdom (commented):
Looking pretty good. Does the SelectMultiple need it too?

src/js/components/DataFilter/DataFilter.js
```diff
@@ -118,11 +118,17 @@ export const DataFilter = ({
...
+    ? `${properties?.[property]?.label || property}`
+    : null;
```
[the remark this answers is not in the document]
britt6612: Ill change

britt6612:
> Looking pretty good. Does the SelectMultiple need it too?

SelectMultiple under the hood already has an `aria-label` with "open drop" so when I did the voice over I thought that was fine since that's our default behavior.

MikeKingdom (approved):

src/js/components/DataFilter/DataFilter.js
```diff
@@ -144,10 +150,22 @@ export const DataFilter = ({
...
+          <CheckBoxGroup
+            aria-label={ariaLabel}
+            id={id}
+            name={property}
+            options={options}
+          />
+        );
       } else {
         content = (
           <SelectMultiple
```
taysea, line 171: `ariaLabel` seems to be missing from here?

src/js/components/DataFilter/DataFilter.js
taysea: This variable name is fine to me

src/js/components/DataFilter/DataFilter.js
britt6612: That's fine I can add it here.

jcfilben (approved):
Looks good!

PR #7045, opened 2023-12-04
add check if values are arrays

### LINKED ISSUE #7034: RangeSelector - controlled form
### Expected Behavior
RangeInput should work without errors when it is used in a controlled lazy Form

### Actual Behavior
This pattern results in a `Maximum update depth exceeded` error

https://codesandbox.io/p/sandbox/grommet-v2-template-forked-shq84y?file=%2Findex.js%3A27%2C63

britt6612:
#### What does this PR do?
This PR adds a check inside of useEffect where the controlled input component is driving the value. 
form.js
```     
     <Form
            onChange={(nextValue) => {
              console.log("Change", nextValue);
            }}
            onSubmit={(event) => console.log("Submit", event.value)}
          >
            <Text>Controlled RangeSelector has problems</Text>
            <FormField name="range" htmlFor="range" label="Range">
              <RangeSelector
                id="range"
                name="range"
                min={0}
                max={100}
                label
                defaultValues={[0, 100]}
                values={range}
                onChange={(nextRange) => {
                  setRange(nextRange);
                }}
              />
            </FormField>
            <Box direction="row" justify="between" margin={{ top: "medium" }}>
              <Button type="submit" label="Update" primary />
            </Box>
          </Form>
```
use the story above to manually test 

The componentValue and the formValue were both coming back as an array however this check `componentValue !== formValue` was not looking inside of the array to check whether or not they were equal so this was causing this useEffect to continuously run in an infinite loop. 
closes #7034 
no
yes
backwards compatible


src/js/components/Form/Form.js
```diff
@@ -460,9 +460,19 @@ const Form = forwardRef(
             componentValue !== undefined && // input driving
             componentValue !== formValue // don't already have it
           ) {
-            setValueState((prevValue) =>
-              setFieldValue(name, componentValue, prevValue),
-            );
+            // Check if componentValue & formValue are arrays
+            if (Array.isArray(componentValue) && Array.isArray(formValue)) {
```
taysea, line 464: It feels like the string comparison should happen as a condition to enter the block at all. Maybe altering line 461 to something like:

 ```
((Array.isArray(componentValue) && Array.isArray(formValue) && componentValue.toString() !== formValue.toString()) || componentValue !== formValue)
```
then leaving everything inside the block as it was before this PR

britt6612: makes sense and simplifies I'll make adjustments

jcfilben (commented):
I'm still getting the maximum update depth exceeded issue in the console

britt6612:
> I'm still getting the maximum update depth exceeded issue in the console

aw okay yeah the last change it came back Ill readjust

Issue #6871, opened 2023-07-10
Data - allow search across more properties than filter


taysea:
### Expected Behavior

When custom `properties` are passed to `Data`, it would still be ideal to search across all fields. For example, in [this filtering example](https://design-system.hpe.design/templates/filtering?q=filter#filtering-with-selectable-results), I would still like to be able to search by "Name" even though I can't filter by name.

### Actual Behavior

I can only search across properties that are also "filterable".

### URL, screen shot, or Codepen exhibiting the issue

https://github.com/grommet/grommet/assets/12522275/d142f52a-acf7-428c-8dd7-6d35ccf38d0f

### Steps to Reproduce

1. search for "Server 1"
2. Notice how there are "0 results"

### Your Environment

- Grommet version: stable
- Browser Name and version:
- Operating System and version (desktop or mobile):

halocline:
MVP - Should be implemented for MVP. Search should be independently configurable and decoupled from filters.

PR #7046, opened 2023-12-05
Adjust Data properties to add filter, sort, and search booleans

jcfilben:
#### What does this PR do?
Changes the Data `properties` prop to include a `filter`, `sort`, `search` boolean for each property. If not specified they will resolve to `true`.

Added a new storybook story `Data/Properties`

Closes https://github.com/grommet/grommet/issues/6871

Yes
Yes
Data is still in beta so this is backwards compatible


MikeKingdom (approved):
LGTM!

taysea:
TLDR to the below: I understand the direction and desire here, but I'm pausing to wonder if it feels like `properties` is taking on too much? Also, this is a pretty large shift in the role of `properties` vs the existing documented behavior, so maybe setting up a review session for this would be valuable to have all devs on same page to discuss?
------------------------------------------

I understand the direction and desire of these changes in terms of providing more fine grain control over which attributes are filterable, searchable, and sortable. Although, I'm wondering if we're now trying to have `properties` do "too much" and losing some of the niceties of the existing documented behavior that `properties` "allows caller to specify how property label renders and which properties should be filterable".

Let's say I have a dataset where a single record is as follows:

```
{
  id: number;
  name: string;
  location: string;
  status: string;
  cost: number;
  quantity: number;
}
```

In the scenario where I want to search across everything except id, and only filter on location, status, and cost my configuration of properties would be as follows. Which maybe is completely fine, but I think my initial reaction was that it felt like a lot.

```
properties={{
   id: { searchable: false, filterable: false },
   name: { filterable: false },
   location: { label: "Location" },
   status: { label: "Status" },
   cost: { label: "Cost" },
   quantity: { filterable: false }
}}
```

It also feels a little weird to be mixing the presentation of the "label" which only is relevant to the filter presentation alongside the configurations of what is searchable/sortable/filterable.

MikeKingdom:
> Posing this scenario as a question: Given the proposal in this PR, if `properties` is configured in a way with X columns as filterable and Y as not filterable, and some time down the line another attribute is added to the data-- with the current setup, that attribute would appear in the UI as filterable as soon as the data contained that attribute. If this is not desired, then a change to the UI code would be needed to set filterable to false on that attribute. Or, to avoid unexpected data being filterable and presented to an end user, there would need to be some coordination by the backend and frontend team so that additional attribute wouldn't be added until the front-end code had been modified. In a real world scenario, to avoid undesired filters being presented to an end user, would this coordination be expected?

Great questions. I might be missing something, but assuming there is a `properties` attribute, until the attribute is defined in `properties` it wouldn't appear in the Filter dialog (or be searchable). Once you have a properties attribute, only things that appear in there would be filterable. Any that don't appear in there or ones that have `filterable:false` in there.

taysea:
> Great questions. I might be missing something, but assuming there is a `properties` attribute, until the attribute is defined in `properties` it wouldn't appear in the Filter dialog (or be searchable). Once you have a properties attribute, only things that appear in there would be filterable. Any that don't appear in there or ones that have `filterable:false` in there.

Thanks for clarifying. I think I understand now. So, one shift in this approach is prior `properties` was optional, but this switches the behavior to be required. I can see the argument for that in terms of alignment with the `columns` concept.

jcfilben:
> Thanks for clarifying. I think I understand now. So, one shift in this approach is prior properties was optional (if not provided, show all attributes in filters and search), but this switches the behavior to be required. I can see the argument for that in terms of alignment with the columns concept.

In this approach `properties` is still optional. If you don't provide it everything default to true (filterable, searchable, sortable) same as the current behavior.

taysea (approved):
LGTM! Can you just adjust the PR description to reflect the final property values?

britt6612 (approved):

Issue #6944, opened 2023-09-12
NameValueList and NameValuePair don't handle size xsmall from hpe theme


MikeKingdom:
With the hpe theme, when a window containing a NameValueList is made smaller, at the small breakpoint it changes to a column layout. But when you get to size xsmall it switches back to the row layout.

Size checks for `small` are hardcoded in NameValuePair. They should be in the theme really or at least include xsmall.
 
### Expected Behavior

NameValueList should switch to column layout at size xsmall

### Actual Behavior

NameValueList should reverts to row layout at size xsmall

### URL, screen shot, or Codepen exhibiting the issue
Medium:
![image](https://github.com/grommet/grommet/assets/12054362/5311ff5e-2b9d-41f5-8db4-c711b139904c)

Small:
![image](https://github.com/grommet/grommet/assets/12054362/7b5a2ef2-aa43-45e8-a3c4-b466a7dcf666)

Xsmall:
![image](https://github.com/grommet/grommet/assets/12054362/86d50b71-6381-4285-9e0d-40ff52cffceb)

### Your Environment

- Grommet version: 2.33.1
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #7047, opened 2023-12-05
Adjust NameValueList to account for xsmall screen size

jcfilben:
#### What does this PR do?
Adjusts the code in NameValueList and NameValuePair to account for both 'xsmall' and 'small' screen sizes instead of just 'small'.

Tested with existing NameValueList/Simple story with the hpe theme

Closes https://github.com/grommet/grommet/issues/6944
no
no (only relevant to hpe theme users)
backwards compatible

### DIFF
--- a/src/js/components/DataSearch/DataSearch.js
+++ b/src/js/components/DataSearch/DataSearch.js
@@ -11,6 +11,7 @@ import { TextInput } from '../TextInput';
 import { MessageContext } from '../../contexts/MessageContext';
 import { ResponsiveContext } from '../../contexts/ResponsiveContext';
 import { DataSearchPropTypes } from './propTypes';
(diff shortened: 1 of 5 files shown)


MikeKingdom (commented):
Looks good. There is a bit or repeated logic. It might be worth defining something like
```jsx
const isSmall = (size) => ['xsmall', 'small'].includes(size);
```
or something similar

jcfilben:
> Looks good. There is a bit or repeated logic. It might be worth defining something like
> 
> ```js
> const isSmall = (size) => ['xsmall', 'small'].includes(size);
> ```
> 
> or something similar

Good note, implemented this in latest commit

britt6612 (approved):
Nice clean up

MikeKingdom (approved):
Thanks!

PR #7049, opened 2023-12-06
Fixed handling of multi-level property names that are used as RangeSelectors in filters

### LINKED ISSUE #7037: Data - complex story range filter
### Expected Behavior
On the Data/Complex story when a range filter is applied a badge should show that a filter has been applied and the 'clear filters' button should appear.

### Actual Behavior
There is no indication a filter has been applied and the 'clear filters' button doesn't show up.

This issue seems isolated to the Data/Complex story, other stories with range filters don't have this issue.

https://storybook.grommet.io/?path=/story/data-data-complex--complex

### Steps to Reproduce

1. Visit the story above
2. Filter by GDP

MikeKingdom:
Fixed handling of multi-level property names that are used as RangeSelectors in filters

Note that the badge doesn't go away when the filter is opened again and the range set back to its original value. It is possible to fix this, and I have a working fix, but it is a little complicated and inefficient so far. The user can still hit `Clear filters` to clear the badge

DataForm.js

Manual via Storybook

In the `Data/Complex` story in storybook change the GDP range in the filter dialog to some other value, hit `Apply Filters` and observe that the badge on the filter toolbar icon appears.

Closes #7037 

image
image


src/js/components/Data/DataForm.js
```diff
@@ -181,12 +181,18 @@ const resetPage = (nextFormValue, prevFormValue) => {
...
 
 const transformTouched = (touched, value) => {
+  // DataFilters expects values for keys touched to evaluate to falsey to 
+  // not cause a badge. However any property value that is set back to its
+  // default/initial value that isn't undefined/null/false/0 will cause a
+  // badge this is particulary true for a 'range' which will always have
+  // a value.
+  //
+  // Should this instead determine touched by comparing against
+  // initial/default values?
```
taysea, line 191: From a design standpoint (and to be handled in separate PR), but I might argue that if there are default values applied, the badge should appear. Rationale there being badge is indicating if filters have been applied, regardless of it that's due to default values/views or the user touching the filter directly.

Agree that in all of these cases, "range" is an outlier since it technically always has a value and we want a way not to badge it when the value is equal to min and max.

src/js/components/Data/DataForm.js
```diff
@@ -181,12 +181,18 @@ const resetPage = (nextFormValue, prevFormValue) => {
...
+  //  
   const result = {};
   Object.keys(touched).forEach((key) => {
-    // special case _range fields
-    const parts = key.split('.');
-    if (parts[1] === formRangeKey) result[key] = value[parts[0]];
```
taysea, line 188: Any rationale why _range was initially written with a special case but now isn't needed?

MikeKingdom: The _xxxx property names are needed for the things in the forms that don't have an associated data property from the user like _search, etc. But the _range is used in a couple other spots to just treat the _range selector specially even though it has a property name from the data. Here it doesn't seem to be really needed since the flattened version of the property with _range on the end seems to be used as is later rather than strip off the _range suffix. Best I can tell...

src/js/components/Data/DataForm.js
jcfilben: I agree that if the default values are something other than the min/max then a badge should appear. I think it makes sense to just check against the min/max for range and if the values are not equal to that show a badge.

jcfilben (approved):
Looks good!

src/js/components/Data/DataForm.js
taysea: It looks like in `viewToFormValue` function there is this logic which would translate to the nested `_range` as opposed to flat:
```
Object.keys(result).forEach((key) => {
    if (
      typeof result[key]?.min === 'number' ||
      typeof result[key]?.max === 'number'
    ) {
      result[key] = { [formRangeKey]: [result[key].min, result[key].max] };
    }
  });
....
```

Will the changes cause any issues? Just trying to piece it all together in my head to understand where it's expecting `_range` to be a nested object inside of a property..

taysea: Also investigating a bit myself to understand better.

taysea (approved):

src/js/components/Data/DataForm.js
MikeKingdom: Thanks for checking!

PR #7052, opened 2023-12-08
Range input #6739

### LINKED ISSUE #6739: RangeInput not respecting background color opacity
### Expected Behavior

RangeInput should accurately render track colors, and include any opacity information (such as a HEX string that includes opacity)

### Actual Behavior

The opacity is not being applied. For example, if a theme specifies `#0000005C`, the `5C` opacity of 36% isn't being honored.

The HPE theme defines the track as "border" (#0000005C), but it's being rendered as #000000.

https://design-system.hpe.design/components/rangeinput?q=rangein

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

Sulaymon333:
#### What does this PR do?
fix opacity not honoured for 8-hex+alpha for range input component track color.

Closes #6739 

Yes

Background compatible


src/js/components/RangeInput/StyledRangeInput.js
```diff
@@ -64,7 +64,7 @@ const trackColorStyle = (props) => {
     else {
       defaultTrackColor = getRGBA(
         normalizeColor(props.theme.rangeInput.track.color, props.theme),
-        props.theme.rangeInput.track.opacity || 1,
+        props.theme.rangeInput.track.opacity || undefined,
```
taysea, line 67: If we remove the fallback to `1` here will this be backwards compatible?

src/js/components/List/index.d.ts
```diff
@@ -43,7 +43,11 @@ export interface ListProps<ListItemType> {
     | string[]
     | { light: string | string[]; dark: string | string[] };
   border?: BorderType;
-  children?: (...args: any[]) => any;
+  children?: (
```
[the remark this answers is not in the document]
sulaymon-tajudeen-hpe: Sorry, I branched from my last PR instead of just the main line. I will have to clean that up and updat the PR.

src/js/components/RangeInput/StyledRangeInput.js
sulaymon-tajudeen-hpe: I belive this should be fine becuase the original implementation was shortcircuting the normalizedAlpha inside getRGBA function to pass alpha channel value as 1 always and it was simply not honouring the 8-hex value with alpha value in the first place. with the updated code being changed to undefined allows the alpha channel value to be passed appropriately and regular 6-hex value works as well. Of course, this is my understanding. we can further think through it again just to ensure the updated code is still backward compatible.

taysea: Thank you for clarifying, I see that a fallback opacity is built into `getRGBA` function that falls back to 1, so it seems reasonable to just have:

`getRGBA(normalizeColor(props.theme.rangeInput.track.color, props.theme), props.theme.rangeInput.track.opacity)`

I don't think we need the `|| undefined`

taysea:
@sulaymon-tajudeen-hpe Can you also add a snapshot test that captures the 8-digit HEX use case ?

taysea (approved):
Thank you for the fix!

jcfilben (commented):
There are some places in the `StyledRangeInput.js` file where we still have `props.theme.rangeInput.track.opacity || 1` or `opacity || 1` (see lines 82 and 120). Just wanted to understand why we are keeping it on those lines

sulaymon-tajudeen-hpe:
Both lines 82 and 120 did not seems to make any difference with || 1 on the 8-digit hex code, that is why I did not remove them initially. For the purpose of cnsistency and since || 1 is already taken care of from the original getRGBA function that falls back to 1, the || 1 is now removed from both lines 82 and 120.

jcfilben (commented):
I'm not seeing the opacity coming through on cases where we specify an 8 digit hex within the RangeInput theme. For example if we specify something like this in the theme only the thumb color is respecting the opacity
```
rangeInput: {
  thumb: {
    color: '#7f03fc20',
  },
  track: {
    lower: {
      color: '#fc03a925',
    },
    upper: {
      color: '#705a6923',
    },
  },
}
```

britt6612:

PR #7053, opened 2023-12-08
add dataFormContext

### LINKED ISSUE #7054: If DataSearch is rendered in a Layer nested in another form, it always has a form label and doesn't filter data properly
If DataSearch is rendered in a Layer nested in another form in the React Tree (such as a scenario where Data is rendered within a wizard form), it always has a form label and doesn't filter data properly. It seems there's an error with the `noForm` condition.

Expected behavior would be that DataSearch only renders with label when it's inside a DataForm specifically.

Seen here: https://codesandbox.io/s/hardcore-spence-rdgvql?file=/src/App.js

britt6612:
#### What does this PR do?
This PR creates a new `DataFormContext` 
previously there was a context in Form to determine if something was wrapped in Form.
However there was an issue when `DataSearch was rendered in a Layer nested in another form`. This is because the Layer in rendered in the DOM at the bottom so its parent was not the `Form` 

Closes #7054 

DataForm.js
storybook
storybook

The context is used internally within Grommet so it made more sense for this context to be within DataForm so that any of the other components that are used in addition to Data can access this Context instead of being in Form. 
closes `If DataSearch is rendered in a Layer nested in another form, it always has a form label and doesn't filter data properly`
no
no
backwards compatible


src/js/components/Form/FormContext.js
```diff
@@ -24,5 +24,4 @@ const useFormField = ({ error, info, disabled }) => ({ error, info, disabled });
 export const FormContext = React.createContext({
   useFormField,
   useFormInput,
-  noForm: true,
```
jcfilben, line 27: I'm a little confused where in the current implementation `noForm` gets set to false

taysea: Right now, when you wrap something in a Form then it completely overrides this default context value so `noForm` just becomes undefined. And this was only added for the Data + friends scenario, but since we need a bit more specificity that we are in a DataForm, we don't need this generic `noForm` value in FormContext anymore.

src/js/components/DataSearch/DataSearch.js
```diff
@@ -19,16 +19,16 @@ const dropProps = {
...
   const { format } = useContext(MessageContext);
   const size = useContext(ResponsiveContext);
   const skeleton = useSkeleton();
   const [showContent, setShowContent] = useState();
   const id = idProp || `${dataId}--search`;
 
   useEffect(() => {
-    if (noForm) addToolbarKey('_search');
-  }, [addToolbarKey, noForm]);
+    if (inDataForm) addToolbarKey('_search');
```
taysea, line 30: Should it be this?

```suggestion
    if (!inDataForm) addToolbarKey('_search');
```

britt6612: good catch

src/js/components/Form/FormContext.js
britt6612: ☝️

taysea (approved):
LGTM.

jcfilben (approved):
Looks good!

PR #7057, opened 2023-12-11
Change datafilters default presentation within toolbar to `layer`

### LINKED ISSUE #7056: DataFilters default presentation of DataFilters to be `layer`
To align with 80% case for presentation format of filters, make DataFilters default behavior when using `toolbar` prop `layer`.

taysea:
#### What does this PR do?

In 80% of use cases where filters are hidden behind a filter button, they should open into a side drawer. More often than not, there are many attributes to be filtered on (especially in the default case before any `properties` have been specified. This ensures that we don't provide a default experience where there are tons of filters displayed in a drop which breaks our design recommendations.

This does not alter the default behavior for use cases not leveraging the `toolbar` prop. In those scenarios, the default behavior remains as inline and can be opted into via `layer`.

src/js/components/Data/Data.js

Local in storybook Data Simple.js

Any storybook that's using `toolbar` on Data.

Closes #7056 

Breaking change, but this is fine because Data + friends are still in beta.

### DIFF
--- a/src/js/components/Data/Data.js
+++ b/src/js/components/Data/Data.js
@@ -145,7 +145,7 @@ export const Data = ({
       <Toolbar key="toolbar">
         {(toolbar === true || toolbar === 'search') && <DataSearch />}
         {(toolbar === true || toolbar === 'view') && <DataView />}
-        {(toolbar === true || toolbar === 'filters') && <DataFilters drop />}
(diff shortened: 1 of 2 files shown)


britt6612 (approved):

jcfilben (approved):
Looks good!

Issue #7042, opened 2023-11-30
Enhancement: Change MaskedInput icon click to open suggestions drop-down


kimkscott:
### Expected Behavior

When the user clicks the icon shown in MaskedInput, the list of suggestions is shown.

### Actual Behavior

After the user enters a value in MaskedInput, if they click on the icon in MaskedInput, the list of suggestions isn't shown.  Instead the cursor is positioned to the right of the current value.

### URL, screen shot, or Codepen exhibiting the issue

![image](https://github.com/grommet/grommet/assets/91684949/c501c3b3-a441-4444-87a9-f1ca7fe8a135)

A code sandbox is here: https://codesandbox.io/p/sandbox/date-time-form-q3c4yj?file=%2Findex.js%3A118%2C36

The associated COM Jira is here: https://jira-pro.it.hpe.com:8443/browse/CCSE-15989

### Steps to Reproduce

The COM UI has a modal for creating a schedule that uses DateInput to enter the day and MakedInput to enter the time.  

1.  When the user clicks the calendar icon in DateInput, the calender opens and the user can select a day.  Later if the user wants to change the day, they can click on the calendar icon and the calendar re-opens so the user can pick a different day.

2. When the user clicks on the clock icon (inside the input field) in the MaskedInput, the drop-down showing choices opens and the user can select options or enter a time.  However, if later the user wants to change the time and clicks on the clock icon, the option list doesn't open.  MaskedInput just positions the cursor to the right of the previously entered time.  The user either has to backspace over the time or click to the left of the time to open the list of options.

### Your Environment

- Grommet version: 2.33.1
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): Windows

abdulbasithqb:
This is possible enhancement for this issue : PR - https://github.com/grommet/grommet/pull/7066 
It would be appreciated if you could provide a feedback on these changes
Thanks

jcfilben:
The existing behavior where clicking the icon has the same result as clicking the input makes sense to me. @taysea maybe we can chat about this enhancement? I think it is confusing if clicking the icon has a different behavior than clicking the input and we are introducing two separate click targets.

I think it is also confusing which suggestions list should be displayed when the icon is clicked. In the time example should the hours, minutes, or am/pm suggestions show when the icon is clicked if all fields are already filled out?

jcfilben:
[...]

PR #7066, opened 2023-12-13
ehancement: add MaskedInput icon click-ability

abdulbasithqb:
#### What does this PR do?
Introduces the ability to make the MaskedInput icon clickable, triggering the display of suggestion options.

Review changes in the MaskedInput component related to icon click-ability(components/MaskedInput/MaskedInput.js).

- Manually clicking the icon to ensure the expected behavior.
- Local Jest tests.

Manually test by clicking the icon in a MaskedInput field with the `maskIconClickable` prop set to `true` and verify the suggestion options appear. Then, test with `maskIconClickable` set to `false` and ensure the icon does not trigger the suggestions.

This enhancement improves user interaction with MaskedInput by allowing suggestions to be triggered by clicking the icon.

fixes: #7042

#### Screen-record (if appropriate)

https://github.com/grommet/grommet/assets/134603758/33ef4c65-339d-41ea-aadd-ff0bca396c4a

No

Not sure.

No.

### DIFF
--- a/src/js/components/MaskedInput/MaskedInput.js
+++ b/src/js/components/MaskedInput/MaskedInput.js
@@ -163,6 +163,7 @@ const MaskedInput = forwardRef(
       focus: focusProp,
       focusIndicator = true,
       icon,
+      maskIconClickable = false,
       id,
       mask = defaultMask,
       name,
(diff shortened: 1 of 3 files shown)


taysea:
Thanks for this proposal. I'm tagging this as "discussion" because there is a bit to discuss here from an API standpoint and accessibility standpoint as well since we would need to consider rendering this as a Button/avoiding nested interactive elements/etc.

Issue #7072, opened 2023-12-19
StarRating - typescript issues


jcfilben:
### Expected Behavior
I should be able to provide a `name` and `id` to StarRating

### Actual Behavior
I get the following typescript warning
```
Type '{ id: string; name: string; }' is not assignable to type 'IntrinsicAttributes & RadioButtonGroup'.
```

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/p/sandbox/grommet-v2-template-forked-tdlvg8?file=%2Findex.tsx%3A42%2C14

PR #7073, opened 2023-12-20
fix: starRating - typescript issues

abdulbasithqb:
#### What does this PR do?
fix: starRating - typescript issues

StarRating/index.d.ts

yarn test

create starRating with props id

fixes: #7072

Screenshot 2023-12-20 at 6 15 08 PM

No

No

No


src/js/components/StarRating/index.d.ts
```diff
@@ -1,7 +1,9 @@
 import * as React from 'react';
 import { RadioButtonGroupProps } from '../RadioButtonGroup/index';
 
-type RadioButtonGroup = Omit<RadioButtonGroupProps, 'options'>;
+type RadioButtonGroup = Omit<RadioButtonGroupProps, 'options'> & {
+  id?: string;
+};
```
jcfilben, line 6: I think we can just import `RadioButtonGroupExtendedProps` instead of `RadioButtonGroupProps`. This way we don't need to explicitly define a type for `id`.

abdulbasithqb: Thanks for this info, i just updated as requested,

jcfilben (approved):
Looks good!

taysea (approved):

PR #7069, opened 2023-12-15
fix: Resolve hover styling and tooltip issue for DataSort control

abdulbasithqb:
#### What does this PR do?
Fix: Resolve hover styling and tooltip issue for DataSort control

#### Where should the reviewer start?
Review changes in the DataSort component(components/DataSort/DataSort.js)

#### What testing has been done on this PR?
 1.  Manually clicking the icon to ensure the expected behavior.
 2. Local Jest tests.

#### How should this be manually tested?
Manually test the DataSort control by hovering over sort button and checking if the tooltips appear as intended.

#### Do Jest tests follow these best practices?

- [ ] `screen` is used for querying.
- [ ] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [ ] `asFragment()` is used for snapshot testing.

#### What are the relevant issues?
fixes #7064 

#### Screenshots (if appropriate)
Screenshot 2023-12-15 at 11 51 59 AM

#### Do the grommet docs need to be updated?
No

#### Should this PR be mentioned in the release notes?
No

#### Is this change backwards compatible or is it a breaking change?
Yes, this change is backwards compatible, and Jest tests have been updated accordingly.

PR #7078, opened 2024-01-05
Remove use of `hoverIndicator` within Data

taysea:
#### What does this PR do?

According to button docs, `hoverIndicator` should only be used with `plain` Buttons. Removed improper use of `hoverIndicator` within Data components to allow theme to drive styling.

Related to https://github.com/grommet/grommet/pull/7069

Backwards compatible, these components are still in beta.


jcfilben (commented):
Looks like there are some snapshot issues when the tests run on circleci.

Also should we add the theme hover styling in this PR?

taysea:
> Looks like there are some snapshot issues when the tests run on circleci.
> 
I'll resolve those.

> Also should we add the theme hover styling in this PR?

In base/grommet theme, icon only buttons don't have hover styling. In HPE theme, they will with the `toolbar` kind defined my HPE theme. So, I don't know that any theme changes are needed since it's behaving in accordance with the rest of grommet theme. What do you think?

jcfilben:
> In base/grommet theme, icon only buttons don't have hover styling. In HPE theme, they will with the toolbar kind defined my HPE theme. So, I don't know that any theme changes are needed since it's behaving in accordance with the rest of grommet theme. What do you think?

Makes sense, agree I don't think we need theme changes then.

jcfilben (approved):
Looks good!

britt6612 (approved):

Issue #7074, opened 2023-12-20
SelectMultiple - adding additional options within `onSearch`


jcfilben:
### Expected Behavior
When I search within SelectMultiple I should be able to use my search term to query a backend to fetch more options that were not available already in the `options` array.

### Actual Behavior
The internal `allOptions` prop doesn't get updated to reflect the new options. SelectMultiple is currently built in a way that `onSearch` can't change `options` to something that isn't a subset of the previous `options`.

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/p/sandbox/grommet-v2-template-forked-ht9y6h?file=%2Findex.js%3A16%2C34

### Steps to Reproduce
In the above codesandbox example search for 'cd', notice new options were added but the new options can not be selected

### Your Environment
- Grommet version: 2.34.2
- Browser Name and version:
- Operating System and version (desktop or mobile):

MikeKingdom:
To support scalability, both Select and SelectMultiple need to be changed so they don't rely on having all the options in allOptions . I looked at this a bit before the break and it's not trivial, but the code that tries to map indexes from one to the other shouldn't have to do it that way. We shouldn't attempt to somehow add new stuff to allOptions based on an async search query and a selection list shouldn't be index based.

PR #7079, opened 2024-01-05
SelectMultiple new options on search

jcfilben:
#### What does this PR do?
Enables new options to be added to the SelectMultiple within the onSearch function

Closes https://github.com/grommet/grommet/issues/7074

yes
yes
backwards compatible


src/js/components/SelectMultiple/SelectMultiple.js
```diff
@@ -152,6 +152,23 @@ const SelectMultiple = forwardRef(
       if (!search) setAllOptions(optionsProp);
     }, [optionsProp, search]);
 
+    useEffect(() => {
+      if (search && optionsProp && optionsProp.length > 0) {
+        const additionalOptions = [...allOptions];
+        optionsProp.forEach((i) =>
+          additionalOptions.some((j) => {
```
taysea, line 159: Syntax question, could this also be written as `!additionalOptions.some(...) && additionalOptions.push(i)` -- or is the `null` return needed specifically?

src/js/components/SelectMultiple/stories/SearchFetch.js
```diff
@@ -0,0 +1,56 @@
+import React, { useState } from 'react';
+
+import { Box } from 'grommet';
+import { SelectMultiple } from '../SelectMultiple';
+
+const locationList = [
```
taysea, line 6: From offline discussion, I agree with not adding the storybook to the published set. Or if we do, build it out to be more representative of fetching from an API.

src/js/components/SelectMultiple/SelectMultiple.js
```diff
@@ -152,6 +152,23 @@ const SelectMultiple = forwardRef(
...
+            if (typeof i === 'object')
```
taysea, line 160: Can this be rewritten to:

```
return typeof i === 'object' ? applyKey(i, valueKey) === applyKey(j, valueKey) : i === j;
```

taysea: Although if this feels less readable, i'm fine leaving the initial implementation

taysea (commented):
Looking good, some minor comments!

src/js/components/SelectMultiple/SelectMultiple.js
jcfilben: Good suggestion, the `null` return is not needed. Incorporated this suggestion in latest commit

src/js/components/SelectMultiple/SelectMultiple.js
jcfilben: Incorporated this in the latest commit. I think you suggestion doesn't hurt readability

src/js/components/SelectMultiple/stories/SearchFetch.js
jcfilben: Yeah I'm thinking we will remove this story before committing

taysea (approved):
I think this approach is reasonable for the time being. Once others approve, let's remove the storybook file.

Side note -- I was trying to think of an approach that might be a little more performant than mapping over the array with `.some` for each loop iteration, like potentially using a Set or some other structure. Although, since options can be an array of objects, the Set approach didn't exactly work. Not a blocker to moving forward with this, but just something I was thinking about.

MikeKingdom (approved):

PR #7080, opened 2024-01-05
Decouple DataFilters and DataSort, refine toolbar layout

### LINKED ISSUE #7063: DataSort and DataFilters - determine a reasonable default behavior
Having both `DataSort` and `DataFilters` results in some strange behavior. After selecting ‘sort by’ and ‘sort direction’ those fields show up in DataFilters. It feels like they should not show up in DataFilters.

This behavior is coming from https://github.com/grommet/grommet/blob/master/src/js/components/DataFilters/DataFilters.js#L110-L112
If `DataFilters` doesn't have children and `view.sort` is defined, `DataFilters` will include `DataSort` within.

What is the 80% case when using DataSort. Do we expect it should be within DataFilters or a separate control?

https://github.com/grommet/grommet/assets/54560994/7eca31ee-f346-4f8c-9923-7b789e4c62b6

### Steps to Reproduce

1. https://storybook.grommet.io/?path=/story/data-data-properties--properties
2. Notice the DataFilters drop initially only shows a location filter
3. Open the DataSort drop and select a property to sort by and a direction
4. Open the DataFilters drop. Notice the drop content now includes DataSort.

taysea:
#### What does this PR do?

This PR:
- Decouples DataSort from DataFilters. In cases where a view applied "sort", then the sort controls were appearing as part of filters. Now, DataSort will appear as a peer to DataFilters, keeping DataFilters purely focused on filter inputs.
- Aligns default `toolbar` layout with [Figma designs](https://www.figma.com/file/bMVkdgHkimCENGHiKPGcYH/Maps?type=design&node-id=1260-91540&mode=design&t=ABgPyLwvKbDaPeLj-0). 
   - DataView should appear in a separate section and after filter controls.
- Adjusts DataClearFilters behavior to only clear filters and restore dataset to page 1, but does not clear search, sort, step, etc.

Screen Shot 2024-01-08 at 11 41 40 AM

Storybook.

Closes #7063 

Backwards compatible, these components are still in beta.

### DIFF
--- a/src/js/components/Data/Data.js
+++ b/src/js/components/Data/Data.js
@@ -109,6 +109,10 @@ export const Data = ({
       const nextView = { ...view };
       delete nextView.properties;
       delete nextView.page;
+      // by clearing the properties from a view, it is no
+      // longer reflecting the view
+      delete nextView.name;
+      delete nextView.view;
       setFiltersCleared(true);
       setView(nextView);
       if (onView) onView(nextView);
--- a/src/js/components/Data/Data.js
+++ b/src/js/components/Data/Data.js
@@ -3,6 +3,7 @@ import { AnnounceContext } from '../../contexts';
 import { Box } from '../Box';
 import { DataFilters } from '../DataFilters';
 import { DataSearch } from '../DataSearch';
+import { DataSort } from '../DataSort';
 import { DataSummary } from '../DataSummary';
 import { DataView } from '../DataView';
 import { Toolbar } from '../Toolbar';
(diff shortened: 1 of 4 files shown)


jcfilben (commented):
The code looks good, one thing that is a little confusing (and maybe a separate issue) is when I have options configured in the DataSort drop but nothing selected in the DataFilters drop there is no 'Clear filters' button, which leads me to believe that the 'Clear filters' button is only associated with DataFilters. But when I do have something selected in DataFilters and I click 'Clear filters' it will also reset DataSort.

taysea:
> The code looks good, one thing that is a little confusing (and maybe a separate issue) is when I have options configured in the DataSort drop but nothing selected in the DataFilters drop there is no 'Clear filters' button, which leads me to believe that the 'Clear filters' button is only associated with DataFilters. But when I do have something selected in DataFilters and I click 'Clear filters' it will also reset DataSort.

Good comment. A separate note I have in the "Data + friends" board is to modify the "Clear filters" control to only clear out filters (not search, and based on these changes likely not sort either). However, given that, I'm wondering how someone might "clear" out sorting if desired. We can discuss more next week.

taysea:
After discussion with the dev group, we will:
- Not have DataSort appear as part of toolbar by default, even if a view has a sort. This is because a view might contain a sort config but that doesn't necessarily mean it's desired for the user to have control over it.
- "Clear filters" should not clear out Sort or Search, or other view configurations. It should just clear out `properties`
   - One nuance to explore is if it should clear `page` to 1 or not

jcfilben (commented):
**Update** Just saw your draft PR (https://github.com/grommet/grommet/pull/7084) I think that will address some of the things I mentioned below

Maybe unrelated to this PR but something odd I observed while testing this is that if I have a view applied and try to 'unselect' a filter option in that view I am not able to. I am able to select new filters to apply to the view. Then when i hit 'clear filters' the view is cleared. I think the behavior isn't intuitive and there should be a separate way to clear the view


https://github.com/grommet/grommet/assets/54560994/26a24c06-cf7a-4dc0-b664-dae8ffaf550a

jcfilben (approved):
Looks good!

taysea:
> **Update:** Just saw your draft PR (#7084) I think that will address some of the things I mentioned below
> 
> Maybe unrelated to this PR but something odd I observed while testing this is that if I have a view applied and try to 'unselect' a filter option in that view I am not able to. I am able to select new filters to apply to the view. Then when i hit 'clear filters' the view is cleared. I think the behavior isn't intuitive and there should be a separate way to clear the view
> 

Totally agree it's unintuitive, but yes that draft PR should address this.

MikeKingdom (approved):

halocline (approved):

PR #7083, opened 2024-01-09
Refine DataFilters badge behavior for RangeSelector reset to min/max

### LINKED ISSUE #7082: DataFilters -- when RangeSelector value returns to min/max it is still being flagged as filter applied
Since RangeSelector always has a "value", there are scenarios where RangeSelector is restored to its min/max value in DataFilters but it still contributes to the badge count as if it's a filter applied.

When RangeSelector is at its min/max values, it should not cause a badge on DataFilters.

taysea:
#### What does this PR do?

When RangeSelector is reset to its min/max, DataFilters should not "badge" that a filter has been applied. Putting this as a draft to get reactions to the approach.

Prior to this, a badge would apply even if RangeSelector is at its min/max. This is because technically RangeSelector always has a value, so once it was view as "touched" by DataFilters, there was no way to get it to be removed from the "touched" object without clearing all filters.

This is done by:
1. Creating a DataFiltersContext that has a value of a `pendingReset` ref in DataFilters.
2. `pendingReset` ref in DataFilters whose value is a Javascript Set, to track which range selectors should not be included in the "touched" object.
3. on change of its value, RangeSelector will add/remove itself from pendingReset depending on if its value is at its min/max.

Closes #7082

### DIFF
--- a/src/js/components/DataFilters/DataFilters.js
+++ b/src/js/components/DataFilters/DataFilters.js
@@ -12,7 +12,7 @@ import { Box } from '../Box';
 import { Button } from '../Button';
 import { DataClearFilters } from '../DataClearFilters';
 import { DataFilter } from '../DataFilter';
-import { DataForm, formRangeKey } from '../Data/DataForm';
+import { DataForm } from '../Data/DataForm';
 import { DataSort } from '../DataSort';
 import { DropButton } from '../DropButton';
 import { Header } from '../Header';
@@ -21,6 +21,7 @@ import { Layer } from '../Layer';
 import { DataContext } from '../../contexts/DataContext';
 import { MessageContext } from '../../contexts/MessageContext';
 import { DataFiltersPropTypes } from './propTypes';
+import { DataFiltersContext } from './DataFiltersContext';
 
 const dropProps = {
   align: { top: 'bottom', right: 'right' },
@@ -53,6 +54,9 @@ export const DataFilters = ({
   const { format } = useContext(MessageContext);
   const theme = useContext(ThemeContext);
   const [showContent, setShowContent] = useState();
+  // special case for range selectors which always have a value.
+  // when value returns to its min/max, mark it to be removed from `touched`
+  // so it doesn't contribute to the badge count
   const pendingReset = React.useRef(new Set());
   // touched is a map of form field name to its value, it only has fields that
   // were changed as part of the DataForm here. This is so we can track based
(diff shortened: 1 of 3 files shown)


MikeKingdom (commented):
Seems like a decent workaround. I'm wondering if it's worth generalizing the concept a little, like maybe just name the event something a little more generic in case we need to apply it to another component

MikeKingdom:
> > Seems like a decent workaround. I'm wondering if it's worth generalizing the concept a little, like maybe just name the event something a little more generic in case we need to apply it to another component
> 
> That makes sense, I don't think it would hurt to use a more generic name. Maybe something like `inputvaluereset`?

Cool. I think even `valuereset` works given value is the attr on most inputs.

jcfilben (commented):
If I slide the RangeInput so that one side is at the max/min, the badge doesn't show that that filter is applied. In the clip below notice the badge shows 2 instead of 3.


https://github.com/grommet/grommet/assets/54560994/ff486ab5-a4a8-4442-a318-2d0a609da72a

taysea:
> If I slide the RangeInput so that one side is at the max/min, the badge doesn't show that that filter is applied. In the clip below notice the badge shows 2 instead of 3.
> 
Thanks for noting, I also realized this approach is a little brittle if somebody drags it to its min/max but hasn't submitted the form, then changes the value again. It still considers the range selector to be pendingReset, which causes a mismatch in the badge count. Converting to draft again.

taysea:
@jcfilben @MikeKingdom -- I've altered the approach based on the issues Jessica and I ran into while testing. Let me know your thoughts.

jcfilben (approved):
Looks good!

britt6612 (approved):
LGTM.

src/js/components/DataFilters/DataFilters.js
```diff
@@ -112,44 +117,59 @@ export const DataFilters = ({
     }
   }
 
+  const contextValue = useMemo(() => ({ pendingReset }), []);
```
MikeKingdom, line 120: So the context itself never changes and won't cause a re-render since it's a ref that itself never changes. So it's relying on the `touched` to cause any updates on the badge and in theory the pending reset flag on that field in the Set would have been updated before that. Did I get that right?

MikeKingdom (approved):
Looks good. Just left a question.

src/js/components/DataFilters/DataFilters.js
taysea: Yes, that is correct. Since changing the range selector value will trigger a "touched" change, that's how the process is working.

PR #7084, opened 2024-01-09
DataFilters -- display badge when view applied

### LINKED ISSUE #7081: DataFilters -- badge should appear when default filters are applied via view
If there are filters, yes apply the badge.

Expected behavior based on initial Data Collection Toolbar designs: https://www.figma.com/file/bMVkdgHkimCENGHiKPGcYH/Maps?type=design&node-id=1260-91540&mode=design&t=ABgPyLwvKbDaPeLj-4

- [ ] Apply badging when view applies default filters

taysea:
#### What does this PR do?

The purpose of the `badge` on DataFilters is to notify the user that filters hidden in a drop/layer have values applied. The scenarios when this could occur are:
- The user has interacted directly with the filters and set their values
- A view has been set (either programmatically or by the user via DataView)

In all of these scenarios, a badge should be presented to the user and the number in the badge should align with the number of hidden filters with values applied.

A new configuration for Data properties is added. Now, users can configure if a property should be included in the `badge` of not. For example:
```
<Data
   properties={{
      location: { label: 'Location', badge: false },
      name: { label: 'Name' },
   }}
  ...
/>
```
This configuration is needed in cases where a filter is exposed in the toolbar (as opposed to hidden in the drop/layer). If a view is set, where the view sets a value on the exposed filter and a filter hidden in the layer, we don't want the badge to include a count from this exposed filter.

This PR moves away from the previous mentality of `touched` as a means of representing purely what the user had directly interacted with and moves more towards badging more holistically based on the view.

Lastly, I updated tests and adjusted the DataClearFilters story to more accurately reflect this use case.

https://github.com/grommet/grommet/assets/12522275/c9b13db8-9718-421b-bbf8-7ee2565a9b88

DataFilters

Local storybook + added jest tests.

DataClearFilters Simple storybook.

Closes #7081 

Yes.

Backwards compatible.


jcfilben (commented):
I'm seeing an issue with the filters badge. 

Here is the Data/Complex story with the changes from this PR. Notice that at the end of the video the badge says there are '2' filters applied when really there is only 1.
https://github.com/grommet/grommet/assets/54560994/3c6f5452-79ae-46ac-bba3-fbbfbde025ef

Here is the same flow and story on our regular storybook. In this example the badge correctly shows '1' filter being applied at the end.
https://github.com/grommet/grommet/assets/54560994/8e255ee1-1e3c-45c9-bc99-5587f3e47485

taysea:
> I'm seeing an issue with the filters badge.
> 
> Here is the Data/Complex story with the changes from this PR. Notice that at the end of the video the badge says there are '2' filters applied when really there is only 1.
> 

Good catch, Brittany flagged the same. Just fixed this in latest commit!

src/js/components/DataFilters/DataFilters.js
```diff
@@ -77,6 +76,30 @@ export const DataFilters = ({
...
+    // when view changes via DataView or user interacting with filters,
+    // adjust badge to reflect that
+    if (controlled && view.properties) {
+      const nextTouched = { ...view.properties };
+      Object.keys(nextTouched).forEach((k) => {
+        if (
+          properties?.[k]?.badge === false ||
+          (configured && properties && !properties?.[k])
+        ) {
+          delete nextTouched[k];
```
jcfilben, line 90: Do we need to do `delete view.properties[k];` here too?

src/js/components/Data/DataForm.js
```diff
@@ -262,15 +246,14 @@ export const DataForm = ({
...
       setFormValue(nextValue);
-      if (onTouched) onTouched(transformTouched(touched, nextValue));
       onView(formValueToView(nextValue, views));
       if (onDone) onDone();
     },
-    [formValue, onDone, onTouched, onView, views],
+    [formValue, onDone, onView, views],
   );
 
   const onChange = useCallback(
```
jcfilben, line 259: Do we still need `touched` in this function?

src/js/components/DataFilters/DataFilters.js
taysea: No we don't need it here. For "non RangeSelector" cases, view already removes properties that don't have a value. So we just need this special case for RangeSelector. Although, part of me is wondering if when `onView` is called it should have already sifted the RangeSelector out if it's at its min/max... so it behaves more in alignment with the other properties. Let me explore this briefly.

src/js/components/Data/DataForm.js
taysea: Currently we do because touched is still used for the `debounceSearch` function in onChange.

taysea: In that instance, `touched` is unrelated to `onTouched` and is coming directly from form.

src/js/components/DataFilters/DataFilters.js
taysea: Just pushed the cleanups related to this in my latest commit.

MikeKingdom (commented):
Looks good from my testing. It does seem to still have the range selector issue though where you have to clear filters to get rid of the badge. Was the pending supposed to avoid that?

britt6612:
> Looks good from my testing. It does seem to still have the range selector issue though where you have to clear filters to get rid of the badge. Was the pending supposed to avoid that?

I did not see this issue can you let me know how to reproduce?

MikeKingdom:
> > Looks good from my testing. It does seem to still have the range selector issue though where you have to clear filters to get rid of the badge. Was the pending supposed to avoid that?
> 
> I did not see this issue can you let me know how to reproduce?

- Go to Data/Complex story
- Click on filter icon
- Change range on the range selector
- Apply filters (badge appears)
- Click on filter icon
- Set range back to full/original
- Apply filters (badge remains)

britt6612:
> Details

good catch in Simple story it is fine.. so I did not see this 🙌

src/js/components/Data/__tests__/Data-test.tsx
```diff
@@ -438,4 +439,61 @@ describe('Data', () => {
     expect(screen.getByRole('button', { name: 'Apply filters' })).toBeTruthy();
     expectPortal('data--filters-control').toMatchSnapshot();
   });
+
+  test('should badge based on view', () => {
```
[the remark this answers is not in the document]
taysea: Will include this in a subsequent commit that also fixes Mike's comment.

taysea:
> > > Looks good from my testing. It does seem to still have the range selector issue though where you have to clear filters to get rid of the badge. Was the pending supposed to avoid that?
> > 
> > 
> > I did not see this issue can you let me know how to reproduce?
> 
> * Go to Data/Complex story
> * Click on filter icon
> * Change range on the range selector
> * Apply filters (badge appears)
> * Click on filter icon
> * Set range back to full/original
> * Apply filters (badge remains)

This has been resolved in the latest commit.

src/js/components/Data/__tests__/Data-test.tsx
taysea, line 443:
```suggestion
  test('should include badge based on view', () => {
```

britt6612 (approved):

src/js/components/Data/DataForm.js
```diff
@@ -159,10 +160,22 @@ const formValueToView = (value, views) => {
...
 // remove any empty arrays of property values by deleting the key for
 // that property in the view properties
-const clearEmpty = (formValue) => {
+const clearEmpty = (formValue, pendingReset) => {
   const value = formValue;
   Object.keys(value).forEach((k) => {
     if (Array.isArray(value[k]) && value[k].length === 0) delete value[k];
+    // special case for when range selector returns to its min/max
+    // flat format with full: true needed to match filter name structure
+    // { a: b: { _range: [0, 100] } } ==> a.b._range: [0, 100]
```
britt6612, line 169: nice comment :)

MikeKingdom (approved):
Awesome

jcfilben (approved):
Looks good!

PR #7091, opened 2024-01-18
DataSearch: Removed dependence on form

MikeKingdom:
#### What does this PR do?
Remove use of DataForm from DataSearch and instead directly modify the `view`

DataSearch.js

Each of the uses of Data in storybook that have the search ability

Storybook Data examples

No

No need

Backwards compatible

### DIFF
--- a/src/js/components/DataSearch/DataSearch.js
+++ b/src/js/components/DataSearch/DataSearch.js
@@ -72,7 +72,10 @@ export const DataSearch = ({ drop, id: idProp, responsive, ...rest }) => {
     />
   );
 
-  if (inDataForm)
+  if (!inDataForm)
+    // likely in Toolbar
+    content = <Box>{content}</Box>;
+  else
     content = (
       <FormField
         htmlFor={id}
--- a/src/js/components/DataSearch/DataSearch.js
+++ b/src/js/components/DataSearch/DataSearch.js
@@ -28,6 +28,7 @@ export const DataSearch = ({ drop, id: idProp, responsive, ...rest }) => {
     addToolbarKey,
     onView,
     view,
+    views,
   } = useContext(DataContext);
   const { inDataForm } = useContext(DataFormContext);
   const { format } = useContext(MessageContext);
@@ -46,8 +47,18 @@ export const DataSearch = ({ drop, id: idProp, responsive, ...rest }) => {
   useEffect(() => setValue(view?.search), [view?.search]);
 
   const onChange = (e) => {
+    const nextValue = { ...view, search: e.target?.value };
+
+    // If there's a named view in effect that has a search term
+    // we'll clear the named view (but leave it's other filters)
+    const currentView =
+      nextValue.view && views?.find((v) => v.name === nextValue.view);
+    if (currentView?.search) {
+      delete nextValue.view;
+      delete nextValue.name;
+    }
     setValue(e.target?.value);
-    debounce(() => () => onView({ ...view, search: e.target?.value }));
+    debounce(() => () => onView(nextValue));
   };
 
   let content = skeleton ? null : (
--- a/src/js/components/DataSearch/DataSearch.js
+++ b/src/js/components/DataSearch/DataSearch.js
@@ -12,7 +12,7 @@ import { MessageContext } from '../../contexts/MessageContext';
 import { ResponsiveContext } from '../../contexts/ResponsiveContext';
 import { DataSearchPropTypes } from './propTypes';
 import { isSmall } from '../../utils/responsive';
-import { useDebounce } from '../../utils';
+import { useDebounce } from '../../utils/use-debounce';
 
 const dropProps = {
   align: { top: 'bottom', left: 'left' },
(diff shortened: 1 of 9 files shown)


taysea:
Looking at the chromatic changes, seems like Form was enforcing some kind of default max width that now is not being set because DataSearch is growing quite wide. It'd be nice to enforce a reasonable default width still.

src/js/components/DataSearch/DataSearch.js
```diff
@@ -42,17 +67,14 @@ export const DataSearch = ({ drop, id: idProp, responsive, ...rest }) => {
       name="_search"
       icon={<Search />}
       type="search"
+      onChange={onChange}
```
taysea, line 70: This is looking good for `onChange` approach. Do you envision there would be a way to incorporate the `updateOn="submit"` functionality with this direction? https://github.com/grommet/grommet/pull/7029

MikeKingdom: Yes. I think it just needs a Keyboard onEnter handler since it can't rely on the form submit anymore

src/js/components/DataSearch/DataSearch.js
```diff
@@ -3,7 +3,7 @@ import { Search } from 'grommet-icons/icons/Search';
 import { ThemeContext } from 'styled-components';
 import { Box } from '../Box';
 import { DataContext } from '../../contexts/DataContext';
-import { DataForm } from '../Data/DataForm';
+// import { DataForm } from '../Data/DataForm';
```
taysea, line 6:
```suggestion
```

src/js/components/DataSearch/DataSearch.js
```diff
@@ -18,8 +18,29 @@ const dropProps = {
   align: { top: 'bottom', left: 'left' },
 };
 
+// 300ms was chosen empirically as a reasonable default
+const DEBOUNCE_TIMEOUT = 300;
```
taysea, line 22: Instead of repeating some of the debounce functions in DataSearch + DataForm, can we have the logic live in a shared util file and import them into each component accordingly?

src/js/components/DataSearch/DataSearch.js
```diff
@@ -42,17 +67,14 @@ export const DataSearch = ({ drop, id: idProp, responsive, ...rest }) => {
...
   );
 
   if (!inDataForm)
     // likely in Toolbar
-    content = (
-      <DataForm footer={false} updateOn="change">
-        {content}
-      </DataForm>
-    );
+    content = <Box>{content}</Box>;
```
taysea, line 77: So we don't lose the reasoning in the future, can we add a comment mentioning that presence of this Box ensures the right default width.

taysea (commented):
Functionality looks good. Small refinement comments.

jcfilben (commented):
In the current implementation on master when I have a search term and I select a view the view is applied and my search term is cleared.
In this PR when I select a view my search term is not cleared, the view is displayed but it is not being filtered by the search term. This feels confusing. I think we should either clear the search term or display the view filtered by the search term.

src/js/components/DataSearch/DataSearch.js
MikeKingdom: I do have updateOn tweaks stashed. I can either add them to this PR or wait until this merges and make them a commit to the updateOn PR since it will have to be updated or closed anyway.

src/js/components/DataSearch/DataSearch.js
MikeKingdom: makes sense.

src/js/components/DataSearch/DataSearch.js
MikeKingdom: Certainly

MikeKingdom:
> In the current implementation on master when I have a search term and I select a view the view is applied and my search term is cleared. In this PR when I select a view my search term is not cleared, the view is displayed but it is not being filtered by the search term. This feels confusing. I think we should either clear the search term or display the view filtered by the search term.

Great question. I'll see if I can make this still happen

src/js/components/DataSearch/DataSearch.js
taysea: Let's handle it on the updateOn PR for clarity.

taysea:
When I select a view that configures a search term, then I clear that search term, the view does not clear. If we are holding the mindset that "if the current state of search/sort/filters does not match the view, clear the view", should the view clear out here?

https://github.com/grommet/grommet/assets/12522275/03f27e35-374f-4042-a564-021a34a46920

MikeKingdom:
> When I select a view that configures a search term, then I clear that search term, the view does not clear. If we are holding the mindset that "if the current state of search/sort/filters does not match the view, clear the view", should the view clear out here?

Are you thinking it should only clear the named view if that view has a search term? In general I think you still want the ability to add/remove a search term with a named view still in effect, at least if that named view doesn't have a search term component.

taysea (approved):
Behavior looks good!

jcfilben (approved):
Looks good!

PR #7094, opened 2024-01-24
Set min-width of DataFilters layer

taysea:
#### What does this PR do?

Sets min-width of medium on DataFilters layer for more reasonable default width. With removal of "undo changes" which previously was holding some space to make the layer wider, the DataFilter layer was rendering quite narrow if the data labels were short. This sets a reasonable default for consistency.

src/js/components/DataFilters/DataFilters.js

BEFORE
Screen Shot 2024-01-23 at 5 23 50 PM

AFTER
Screen Shot 2024-01-23 at 5 19 40 PM


britt6612 (approved):
LGTM! Thanks for adding

jcfilben (approved):
Looks good!

PR #7099, opened 2024-01-26
fix missing props and typings for tag component

Sulaymon333:

### DIFF
--- a/src/js/components/Tag/propTypes.js
+++ b/src/js/components/Tag/propTypes.js
@@ -1,15 +1,8 @@
 import PropTypes from 'prop-types';
-import {
-  backgroundPropType,
-  genericProps,
-} from '../../utils/general-prop-types';
 
 let PropType = {};
 if (process.env.NODE_ENV !== 'production') {
   PropType = {
-    ...genericProps,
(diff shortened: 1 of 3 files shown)


britt6612 (approved):
Thanks for cleaning this up

src/js/components/Tag/propTypes.js
```diff
@@ -1,11 +1,8 @@
 import PropTypes from 'prop-types';
-import { genericProps } from '../../utils/general-prop-types';
 
 let PropType = {};
 if (process.env.NODE_ENV !== 'production') {
   PropType = {
-    ...genericProps,
```
taysea, line 7: Why do we remove genericProps / as here? If so, should we be spreading Box prop types?

britt6612:
https://github.com/grommet/grommet-site/pull/504 documentation for PR

src/js/components/Tag/propTypes.js
taysea: Since Tag is built on box, it will inherit the props. Resolving this comment.

taysea (approved):
LGTM.

jcfilben (approved):
Looks good!

Issue #7024, opened 2023-11-19
Enhancement : DataSearch to trigger search onenter, instead of onchange


anoopadvaitha:
### Expected Behavior
DataSearch Component should give an option to choose us to select whether to search onchange or onsubmit/or enter.

### Actual Behavior
DataSearch triggers search as soon as I start typing, this causes issue if we are using REST API triggers a lot of API calls.

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.34.1
- Browser Name and version:
- Operating System and version (desktop or mobile):

anoopadvaitha:
Open to contribute if accepted.

jcfilben:
Hi @anoopadvaitha this seems like a reasonable enhancement

taysea:
This is closed by #7110

PR #7110, opened 2024-01-30
Implement DataSearch updateOn property

MikeKingdom:
This is an updated version of PR #7029 originally raised by @anoopadvaitha.

Currently DataSearch filters data as soon as we start typing (onchange). This becomes challenge in scenarios where the data is fetched from API. 

#### What does this PR do?
The PR provides updateOn parameter which is passed to DataForm so that developer can leverage to either use change or submit. 

\src\js\components\DataSearch\DataSearch.js

Tested when no value is provided for DataSearch and when submit is provided and works in both scenarios. 

#7024 

Yes, will be raising the PR on that

No.

Yes

### DIFF
--- a/src/js/components/DataSearch/DataSearch.js
+++ b/src/js/components/DataSearch/DataSearch.js
@@ -8,6 +8,7 @@ import { DataFormContext } from '../../contexts/DataFormContext';
 import { FormField } from '../FormField';
 import { useSkeleton } from '../Skeleton';
 import { TextInput } from '../TextInput';
+import { Keyboard } from '../Keyboard';
 import { MessageContext } from '../../contexts/MessageContext';
 import { ResponsiveContext } from '../../contexts/ResponsiveContext';
 import { DataSearchPropTypes } from './propTypes';
@@ -21,7 +22,13 @@ const dropProps = {
 // 300ms was chosen empirically as a reasonable default
 const DEBOUNCE_TIMEOUT = 300;
 
-export const DataSearch = ({ drop, id: idProp, responsive, ...rest }) => {
+export const DataSearch = ({
+  drop,
+  id: idProp,
+  responsive,
+  updateOn,
+  ...rest
+}) => {
   const {
     id: dataId,
     messages,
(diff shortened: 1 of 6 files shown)


src/js/components/DataSearch/index.d.ts
```diff
@@ -4,6 +4,7 @@ import { TextInputProps } from '../TextInput/index';
 export interface DataSearchProps {
   drop?: boolean;
   responsive?: boolean;
+  updateOn?: string;
```
taysea, line 7:
```suggestion
  updateOn?: 'change' | 'submit';
```

src/js/components/DataSearch/stories/ChangeonEnter.js
```diff
@@ -0,0 +1,30 @@
...
+  </Grid>
+  // </Grommet>
+);
+
+ChangeOnEnter.args = {
+  full: true,
+};
+
+export default {
+  title: 'Data/DataSearch/ChangeOnEnter',
```
taysea, line 29: Maybe calling this "updateOn" or "UpdateOnSubmit" might be more aligned to prop?

src/js/components/DataSearch/DataSearch.js
```diff
@@ -77,6 +90,9 @@ export const DataSearch = ({ drop, id: idProp, responsive, ...rest }) => {
     />
   );
 
+  if (updateOn === 'submit')
+    content = <Keyboard onEnter={updateView}>{content}</Keyboard>;
```
taysea, line 94: Is there anything from a screen reader standpoint that we should add to clarify how to submit the search? Or is hitting enter already an assumed way to submit? @jcfilben do you have insights from your work with Bill?

jcfilben: I tested the SelectMultiple search with Bill and he mentioned that for a search input he always expects to be able to hit enter and then the screenreader should give some indication that something actually happened. So I think hitting enter is already the assumed way to submit.

MikeKingdom: So for the default of `change` the search will happen before you'd hit enter (and hitting enter won't actually do anything since it's already done). But does that mean we need some announce context?

jcfilben: Just tested with voiceover, when you hit enter (regardless of whether the search is trigger onChange or onSubmit) the screenreader will read out the summary text. I think this is a good screenreader experience but we can always validate with Bill. Either way I think we can move forward with this PR and if we need to make any screenreader adjustments we can do that separately.

taysea (approved):
Thanks, Mike!

PR #7111, opened 2024-01-30
Fix undefined case for DataFilter

taysea:
#### What does this PR do?

When testing Grommet designer, if DataFilter was added and no `property` had been defined yet, the app would crash.

In a Typescript project, the typings would already flag that `property` is required, so this is a fix for non-Typescript projects.

Tested locally in storybook.

```
<DataFilter />
```

Got this error in Grommet designer:

Screen Shot 2024-01-29 at 4 18 27 PM

Backwards compatible.

### DIFF
--- a/src/js/components/DataFilter/DataFilter.js
+++ b/src/js/components/DataFilter/DataFilter.js
@@ -15,8 +15,7 @@ const minSelectSearchOptions = 10;
 const defaultRangeSteps = 20;
 
 const getValueAt = (valueObject, pathArg) => {
-  if (!pathArg) return undefined;
-  if (valueObject === undefined) return undefined;
+  if (!pathArg || valueObject === undefined) return undefined;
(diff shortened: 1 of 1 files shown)


src/js/components/DataFilter/DataFilter.js
```diff
@@ -15,6 +15,7 @@ const minSelectSearchOptions = 10;
 const defaultRangeSteps = 20;
 
 const getValueAt = (valueObject, pathArg) => {
+  if (!pathArg) return undefined;
   if (valueObject === undefined) return undefined;
```
jcfilben, line 19:
```suggestion
  if (!pathArg || valueObject === undefined) return undefined;
```
Could we change to this so we only need one `return undefined` statement?

britt6612 (approved):

jcfilben (approved):
Looks good!

PR #7123, opened 2024-02-07
Readonly enhancement for TextInput and DateInput

jcfilben:
#### What does this PR do?
Adds the ability to style readonly inputs in the theme. In addition a new prop was added `readOnlyCopy` to allow copy to clipboard button to be added to the TextInput and DateInput.

Added storybook examples and jest tests.
Tested styling by adding a background and border to the readOnly theme

Closes https://github.com/grommet/hpe-design-system/issues/3695

Yes

Yes

Backwards compatible


src/js/themes/base.js
```diff
@@ -315,6 +315,12 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         },
         // deprecate in v3
         // weight: undefined,
+        readOnly: {
```
taysea, line 318: if this isn't applicable to all inputs, would it be better to just have it under the components that are affected?

taysea: or is it implied that this only will apply to inputs that can support readonly? I could be okay with that too

src/js/components/TextInput/TextInput.js
```diff
@@ -129,6 +132,25 @@ const TextInput = forwardRef(
     );
 
     const [suggestionsAtClose, setSuggestionsAtClose] = useState();
+    const [tipContent, setTipContent] = useState(
+      format({ id: 'input.readOnlyCopyValidation', messages }),
+    );
+
+    const copyOnClick = () => {
```
taysea, line 139:
```suggestion
    const onClickCopy = () => {
```
small naming suggestion for handler to start with `on`

src/js/components/TextInput/TextInput.js
```diff
@@ -129,6 +132,25 @@ const TextInput = forwardRef(
...
+      global.navigator.clipboard.writeText(value);
+      announce(
+        format({ id: 'input.readOnlyCopyValidation', messages }),
+        'assertive',
+      );
+      setTipContent(format({ id: 'input.readOnlyCopyValidation', messages }));
+    };
+
+    const copyOnBlur = () => {
```
taysea, line 148:
```suggestion
    const onBlurCopy = () => {
```
similar suggestion to above

src/js/components/TextInput/TextInput.js
```diff
@@ -129,6 +132,25 @@ const TextInput = forwardRef(
     );
 
     const [suggestionsAtClose, setSuggestionsAtClose] = useState();
+    const [tipContent, setTipContent] = useState(
+      format({ id: 'input.readOnlyCopyValidation', messages }),
```
taysea, line 136: could we create variables for the formatted messages once and just reference those later in the code?

```
const readOnlyCopyValidation = format({ id: 'input.readOnlyCopyValidation', messages });
const readOnlyCopyPrompt =  format({ id: 'input.readOnlyCopy', messages });

File: src/js/components/TextInput/TextInput.js [hunk 5]
```diff
@@ -129,6 +132,25 @@ const TextInput = forwardRef(
     );
 
     const [suggestionsAtClose, setSuggestionsAtClose] = useState();
+    const [tipContent, setTipContent] = useState(
```
taysea, line 135:
```suggestion
    const [tip, setTip] = useState(
```
would this be sufficient?

src/js/components/TextInput/TextInput.js
```diff
@@ -444,12 +466,35 @@ const TextInput = forwardRef(
...
+      </Tip>
+    );
+
     return (
-      <StyledTextInputContainer plain={plain}>
+      <StyledTextInputContainer
+        readOnlyCopy={readOnlyCopy}
+        plain={plain}
+        border={!plain}
+        round={theme.dateInput.container.round}
```
taysea, line 491: It seems odd to have dateInput theme here

src/js/components/TextInput/TextInput.js
```diff
@@ -465,7 +510,7 @@ const TextInput = forwardRef(
             placeholder={
               typeof placeholder === 'string' ? placeholder : undefined
             }
-            icon={icon}
+            icon={readOnly ? undefined : icon}
```
taysea, line 513: Can this be `!readOnly && icon`?

src/js/components/TextInput/TextInput.js
```diff
@@ -526,8 +572,8 @@ const TextInput = forwardRef(
             }
           />
         </Keyboard>
-
-        {drop}
+        {!reverse && readOnlyCopy === true && CopyButton}
```
taysea, line 575: Do we need the `=== true`?

src/js/languages/default.json
```diff
@@ -82,6 +82,10 @@
     "invalid": "invalid",
     "required": "required"
   },
+  "input": {
+    "readOnlyCopy": "Copy to clipboard",
```
taysea, line 86: Wondering if we could do:
```
"input": {
   "readOnlyCopy": {
       "prompt": "Copy to clipboard",
       "validation": "Copied",
    }
}
```

src/js/utils/readOnly.js
```diff
@@ -0,0 +1,18 @@
...
+  if (theme.global.input.readOnly?.border?.color)
+    styles.push(
+      css`
+        border-color: ${theme.global.input.readOnly.border.color};
+      `,
+    );
+  if (theme.global.input.readOnly?.background)
+    styles.push(
+      css`
+        background-color: ${theme.global.input.readOnly.background};
```
taysea, line 14:
```suggestion
        background: ${theme.global.input.readOnly.background};
```
Is there a case where this isn't a color and we'd want to be slightly more general with the CSS?

taysea (commented):
Looking good, dropped some naming suggestions.

Also, I see we have this on TextInput and DateInput. Should this be supported on MaskedInput also?

britt6612:
> Looking good, dropped some naming suggestions.
> 
> Also, I see we have this on TextInput and DateInput. Should this be supported on MaskedInput also?

@jcfilben I'll hold off on reviewing the code until you have time to incorporate Taylors' comments but also think this could potentially be added to TextArea as well.

src/js/themes/base.js
jcfilben: Eventually the idea would be that we would apply this styling to all inputs that support read-only. In this PR though we are just adding support for TextInput and DateInput

src/js/components/TextInput/TextInput.js
jcfilben: Good idea, changed in latest commit

src/js/components/TextInput/TextInput.js
jcfilben: done

src/js/components/TextInput/TextInput.js
jcfilben: Yes, done

src/js/components/TextInput/TextInput.js
jcfilben: done

src/js/components/TextInput/TextInput.js
jcfilben: Oops this was a mistake, I removed it

src/js/components/TextInput/TextInput.js
jcfilben: done

src/js/components/TextInput/TextInput.js
jcfilben: fixed

src/js/languages/default.json
jcfilben: done

src/js/utils/readOnly.js
jcfilben: good note, I changed this to be `background` instead of `background-color` just to allow a little more flexibility

jcfilben:
Agree that MaskedInput and TextArea should support readOnly styling as well. For now Kenny and I were just going to focus on TextInput and DateInput and then do the rest of the readOnly inputs as follow ons to this work.

src/js/components/DateInput/__tests__/DateInput-test.tsx
```diff
@@ -1164,4 +1164,41 @@ describe('DateInput', () => {
...
   });
+
+  test('read only', () => {
+    const { asFragment } = render(
+      <Grommet>
+        <DateInput
+          format="mm/dd/yyyy"
+          value="01/01/2024"
+          readOnly
+          aria-readonly
```
britt6612, line 1175: did we talk about automatically putting `aria-readonly` if `readOnly` is passed seems like a user would always want this here or we can leave it up to them

src/js/components/TextInput/TextInput.js
```diff
@@ -444,12 +468,31 @@ const TextInput = forwardRef(
...
+        />
+      </Tip>
+    );
+
     return (
-      <StyledTextInputContainer plain={plain}>
+      <StyledTextInputContainer
+        readOnlyCopy={readOnlyCopy}
+        plain={plain}
+        border={!plain}
```
britt6612, line 489: what is this? why is the `border` needed?

src/js/components/DateInput/__tests__/DateInput-test.tsx
jcfilben: I was considering adding it automatically but I think it only needs to be added if something isn't a semantic html form element. I forgot to remove it from my story though, I'll go back and do that.

More details:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-readonly
> Note: When using semantic HTML form controls, if you set the readonly attribute, you don't need to include aria-readonly="true".

src/js/components/TextInput/TextInput.js
jcfilben: This container goes around the Button and the input. I have a border on it so that the border doesn't just go around the input and the button looks like it is outside of the container.

Screen Shot 2024-02-16 at 11 13 11 AM

src/js/components/DateInput/DateInput.js
```diff
@@ -244,7 +264,29 @@ Use the icon prop instead.`,
       );
     }
 
-    const calendarButton = (
+    const onClickCopy = () => {
+      global.navigator.clipboard.writeText(textValue);
+      announce(readOnlyCopyValidation, 'assertive');
```
taysea, line 269: Just confirming "assertive" is the correct announce to use? I know that is supposed to be used sparingly

src/js/components/DateInput/DateInput.js
```diff
@@ -262,19 +304,23 @@ Use the icon prop instead.`,
...
+            }
           }}
         >
-          <Box
+          <StyledDateInputContainer
             ref={containerRef}
             border={!plain}
             round={theme.dateInput.container.round}
             direction="row"
+            readOnlyProp={readOnly}
```
taysea, line 318: Are we calling this `readOnlyProp` to avoid it being passed through to DOM? If so, maybe a comment just so it doesn't get changed in future

src/js/components/DateInput/DateInput.js
```diff
@@ -244,7 +264,29 @@ Use the icon prop instead.`,
...
+      global.navigator.clipboard.writeText(textValue);
+      announce(readOnlyCopyValidation, 'assertive');
+      setTip(readOnlyCopyValidation);
+    };
+
+    const onBlurCopy = () => {
+      if (tip === readOnlyCopyValidation) setTip(readOnlyCopyPrompt);
+    };
+
+    const calendarButton = readOnlyCopy ? (
```
taysea, line 277: For clarity, should we rename this from "calendarButton" to something else? From a readability standpoint, it was a little strange in the logic below to see `readOnlyCopy && calendarButton`

src/js/components/TextInput/TextInput.js
britt6612: got it thank you for clarifying I was a little lost

src/js/components/DateInput/DateInput.js
```diff
@@ -262,19 +304,23 @@ Use the icon prop instead.`,
...
             border={!plain}
             round={theme.dateInput.container.round}
             direction="row"
+            readOnlyProp={readOnly}
             fill
           >
-            {reverse && calendarButton}
+            {reverse && (!readOnly || readOnlyCopy) && calendarButton}
             <MaskedInput
+              readOnly={readOnly}
```
taysea, line 323: Are we expecting people to pass `readOnly` and `readOnlyCopy` if they want the copy behavior? Or if they say `readOnlyCopy` should we also enforce `readOnly` for them?

src/js/components/DateInput/__tests__/DateInput-test.tsx
britt6612: ok thanks 👍

src/js/components/DateInput/DateInput.js
jcfilben: Initially I didn't have the 'assertive' set and was not getting the announcement to come through on voice over so I switched to use assertive

src/js/components/DateInput/DateInput.js
jcfilben: Yes, I'll add a comment

src/js/components/DateInput/DateInput.js
britt6612: I was kinda also thinking the same I felt like if you had `readOnlyCopy` it should trigger `ReadOnly`

jcfilben: > Or if they say readOnlyCopy should we also enforce readOnly for them?

Good idea, that would make things simpler. I'll make that change

jcfilben: done

src/js/components/DateInput/DateInput.js
jcfilben: done, renamed to `DateInputButton`. I feel like the naming isn't perfect but hopefully this makes it more clear that this is just the button within the DateInput rather than the button to open the calendar.

src/js/components/DateInput/DateInput.js
jcfilben: done

src/js/components/TextInput/TextInput.js
```diff
@@ -444,12 +469,31 @@ const TextInput = forwardRef(
 
     const textInputIcon = useSizedIcon(icon, rest.size, theme);
 
+    const CopyButton = (
```
britt6612, line 472: I assume this same component is also going to be used for MaskedInput and potentially TextArea wonder if we pull it out to its separate file and then import that way if we need to change anything it's in one place. just thought for future doesn't necessarily have to be in this PR

jcfilben: Going to hold off consolidating until we add read only to more inputs, I was running into some issues trying to move the copy button to the utils/readonly file. It seems like there are problems importing grommet components with utils.

britt6612: thats fine!

britt6612 (approved):

taysea:
This is a minor thing, and maybe makes sense to handle separately since this behavior already exists in DateInput, but it would be nice if clicking into the field would show the focus indicator around the whole field even if the actual focus is just on the input (more similar to how grommet theme looks, see screenshots).

Screen Shot 2024-02-22 at 8 30 49 AM

Screen Shot 2024-02-22 at 8 30 56 AM

We do this in FileInput where we show focus around the whole field
Screen Shot 2024-02-22 at 8 31 07 AM

taysea (changes requested):
I have some changes staged locally to address styling that wasn't coming through properly for TextInput with readonly that was outside of FormField. Will plan to refine those fixes and either push directly to this PR or wait for when Jessica is back.

taysea (commented):
I pushed changes to this PR to address increasing the button click target.

britt6612:
If I have this in my theme 
```
        readOnly: {
          background: 'light-2',
          border: {
           color: 'brand',
          },
        },
```
below is what I get would we not expect the first TextInput that is using the `readOnlyCopy` wrapped in `FormField` to have the same style?
Screen Shot 2024-02-28 at 10 45 03 PM

src/js/components/TextInput/CopyButton.js
```diff
@@ -0,0 +1,45 @@
...
+  const theme = useContext(ThemeContext);
+
+  return (
+    <Tip dropProps={{ align: { bottom: 'top' } }} content={tip}>
+      <StyledButton
+        onClick={onClickCopy}
+        icon={<Copy />}
+        // only apply horizontal padding since button will
+        // fill height of input
+        pad={{
```
britt6612, line 34: spacing looks much nicer 🙌

src/js/components/SelectMultiple/__tests__/__snapshots__/SelectMultiple-test.tsx.snap
```diff
@@ -182,6 +182,7 @@ exports[`SelectMultiple children 1`] = `
   margin: 0;
   border: 1px solid rgba(0,0,0,0.33);
   border-radius: 4px;
+  border: none;
```
britt6612, line 185: Do we want this? I thought I read in the code 
`readOnly && border: none`
so why would this show up in other snapshots?

taysea: These components are using a readOnly TextInput internally

taysea:
> below is what I get would we not expect the first TextInput that is using the `readOnlyCopy` wrapped in `FormField` to have the same style? Screen Shot 2024-02-28 at 10 45 03 PM

Fixed this in latest commit! @britt6612 

Screen Shot 2024-02-29 at 12 11 20 PM

taysea (approved):
LGTM!

src/js/components/TextInput/TextInput.js
```diff
@@ -444,12 +468,28 @@ const TextInput = forwardRef(
...
+
     return (
-      <StyledTextInputContainer plain={plain}>
+      <StyledTextInputContainer
+        readOnlyProp={readOnly} // readOnlyProp to avoid passing to DOM
+        readOnlyCopy={readOnlyCopy}
+        plain={plain}
+        border={!plain}
+      >
+        {reverse && readOnlyCopy === true && ReadOnlyCopyButton}
```
jcfilben, line 488:
```suggestion
        {reverse && readOnlyCopy && ReadOnlyCopyButton}
```

jcfilben (commented):
Minor comment (I think it was left over from my code), everything else looks good

jcfilben (commented):
I can't officially 'approve' since I'm the original author but everything looks good

britt6612 (approved):
LGTM Thank you for fixing!

Issue #7055, opened 2023-12-11
Calendar range picker behaviour is not conventional


sabarnix:
### Expected Behavior

<img width="474" alt="image" src="https://github.com/grommet/grommet/assets/13603907/c0e40104-d6a8-43d0-a3df-c0e1810b3a43">

on clicking 7th at this point it is expected that the selection should get reset and a new selection should start but currently the start date shifts.

<img width="482" alt="image" src="https://github.com/grommet/grommet/assets/13603907/a7057d3b-1e94-483f-84f5-04c772b194ab">

at this point on clicking 30th the date selection gets reset as expected but after that on selection any date before 30th the date range is not selected. It works only for forward date selection

### Your Environment

- Grommet version: latest
- Browser Name and version: chrome
- Operating System and version (desktop or mobile): desktop

sabarnix:
For anyone looking for a fix can try this workaround
https://codesandbox.io/p/sandbox/quizzical-cannon-qzw7sl

PR #7135, opened 2024-02-15
added a new prop for simple calendar selection #7055

sabarnix:
#### What does this PR do?
Current calendar range selection is ambiguous without the start date and end date inputs. Added a prop simpleSelection which is more predictive selection of date.

test cases are pending

Visualizations/Calendar/RangeSimple
this storybook has the working example

#7055
yes
yes

### DIFF
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -268,6 +268,7 @@ const Calendar = forwardRef(
       showAdjacentDays = true,
       size = 'medium',
       timestamp: timestampProp,
+      simpleSelection = false,
       ...rest
     },
     ref,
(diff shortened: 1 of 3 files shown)


britt6612:
@sabarnix thank you for starting this PR just want to let you know that we are going to discuss this behavior a little more among our team.. [Here](https://github.com/grommet/hpe-design-system/issues/3600) is another issue as well that we have been dropping some comments in regarding this same behavior.

sabarnix:
@britt6612 I was able to achieve the functionality by using activeDate prop.
https://codesandbox.io/p/sandbox/grommet-y546i
This breaks on selecting same date as start and end. Is there a possible solution to this?

sabarnix:
@britt6612 sure

Issue #7131, opened 2024-02-14
List - `onOrder` buttons should be selectable by hitting the 'Space' key


jcfilben:
### Expected Behavior
The onOrder buttons in the list should respond to pressing 'Enter' or 'Space'.

### Actual Behavior
Pressing 'Enter' works but 'Space' does not work.

### URL, screen shot, or Codepen exhibiting the issue
https://storybook.grommet.io/?path=/story/visualizations-list-order--order&globals=theme:hpe

### Steps to Reproduce

1. Go to the above story
2. Navigate to a up/down button on the list and hit 'Space' on the keyboard.

### Your Environment
- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #7140, opened 2024-02-16
fix: onOrder buttons should be selectable by hitting the 'Space' key

abdulbasithqb:
#### What does this PR do?
fix: onOrder buttons should be selectable by hitting the 'Space' key

List Component

1. Manually clicking the start icon to ensure the expected behavior and set default value.
2. Local Jest tests.

1. Run storybook and open List Component section
2. Select Order section Test with Enter and Space

No
fixes: #7131
No

### DIFF
--- a/src/js/components/List/List.js
+++ b/src/js/components/List/List.js
@@ -1,4 +1,11 @@
-import React, { Fragment, useContext, useMemo, useRef, useState } from 'react';
+import React, {
+  Fragment,
+  useCallback,
+  useContext,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 import styled, { ThemeContext } from 'styled-components';
 
 import { DataContext } from '../../contexts/DataContext';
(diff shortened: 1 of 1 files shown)


src/js/components/List/List.js
```diff
@@ -249,54 +256,60 @@ const List = React.forwardRef(
...
-                      event: adjustedEvent,
-                      item: data[active],
-                      index: active,
-                    });
-                  }
-                }
-              : undefined
-          }
+          onEnter={onSelectOption}
+          onSpace={onSelectOption}
```
jcfilben, line 312: Can we add `event.preventDefault` to `onSpace` so that the screen doesn't scroll down when we hit space? It also might be nice to add `event.preventDefault` to `onUp` and `onDown`, I realize that is going outside of the scope of the original issue though so I will leave it up to you if you want to include that part or not.
```suggestion
          onSpace={(event) => {
            event.preventDefault();
            onSelectOption(event);
          }}
```

jcfilben (commented):
Nice work! Just dropped a small comment

src/js/components/List/List.js
abdulbasithqb: hey @jcfilben  for now, added event.preventDefault to `onSpace`
`onUp` and `onDown` will raise a different PR

MikeKingdom (approved):
LGTM

jcfilben (approved):
Looks good!

PR #7141, opened 2024-02-16
Remove engine-strict: true from package.json

taysea:
#### What does this PR do?

Removes `engine-strict` true from package.json so that projects running node versions less than what is defined in `engines.node` of package.json don't fail with an error on install. This aligns back with the [default package.json behavior](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines).

We recently changed our package.json to increase our supported node version from >=16 to >=18 since node 16 has already reached its end of life date. That being said, we don't want to cause errors on dependency install for projects that still use node 16.

package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
     "type": "git",
     "url": "https://github.com/grommet/grommet.git"
   },
-  "engine-strict": true,
   "engines": {
     "node": ">= 18"
   },


jcfilben (approved):
Looks good!

britt6612 (approved):
🙌

PR #7142, opened 2024-02-16
revert node version back to 16

jcfilben:
#### What does this PR do?

In yarn this causes an error when for people using v16 of node, which would break backwards compatibility. Reverting to ensure backwards compatibility.


taysea (approved):
Thank you for the fix.

PR #7143, opened 2024-02-19
fix: List focus via keyboard

umeshiscreative:
#### What does this PR do?
fix the bug when the List focus from the keyboard.

component/List/List.js

I used the storybook to test this component.

The list focuses on accepting only from the keyboard.

#6589

No

No

No

### DIFF
--- a/src/js/components/List/List.js
+++ b/src/js/components/List/List.js
@@ -45,6 +45,10 @@ const StyledList = styled.ul`
     props.itemFocus &&
     focusStyle({ forceOutline: true, skipSvgChildren: true })}}
   ${(props) => props.theme.list && props.theme.list.extend}}
+
+  &:focus:not(:focus-visible) {
+    ${unfocusStyle()}
+  }
 `;
 
 const StyledItem = styled(Box)`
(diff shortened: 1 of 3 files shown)


umeshiscreative:
@jcfilben Can you please review this PR?

src/js/components/List/List.js
```diff
@@ -52,6 +52,10 @@ const StyledList = styled.ul`
     props.itemFocus &&
```
taysea, line 52: This is looking great, although it looks like this block from props.itemFocus is causing the focus indicator to still appear when interacting with mouse. I don't believe this block is necessary anymore given the latest focus/unfocus styles. Can you remove this block entirely? I checked on Chrome, Safari, and Firefox and it behaved as expected for mouse + keyboard navigation:

comment + code to be removed:
```
 // during the interim state when a user is holding down a click,
  // the individual list item has focus in the DOM until the click
  // completes and focus is placed back on the list container.
  // for visual consistency, we want to keep the focus indicator on the
  // list container the whole time.
  ${(props) =>
    props.itemFocus &&
    focusStyle({ forceOutline: true, skipSvgChildren: true })}}
```

jcfilben (approved):
Looks good!

taysea (approved):
Looks good! Thank you

PR #7147, opened 2024-02-22
Enhance Box `gap` to accept object with `row` and `column`

britt6612:
#### What does this PR do?
This PR enhances `gap` on Box so a user can pass in {{ row: small, column: 'large' }}
styledBox.js
local storybook also wrote snapshot

yes
yes
backwards compatible

### DIFF
--- a/src/js/components/Box/Box.js
+++ b/src/js/components/Box/Box.js
@@ -221,7 +221,7 @@ const Box = forwardRef(
         fillProp={fill}
         focus={focus}
         gap={
-          (boxOptions?.cssGap || cssGap) &&
+          (boxOptions?.cssGap || cssGap || typeof gap === 'object') &&
           gap &&
           gap !== 'none' &&
           border !== 'between' &&
--- a/src/js/components/Box/Box.js
+++ b/src/js/components/Box/Box.js
@@ -113,7 +113,7 @@ const Box = forwardRef(
     if (
       gap &&
       gap !== 'none' &&
-      (!(boxOptions?.cssGap || cssGap) ||
+      (!(boxOptions?.cssGap || cssGap || typeof gap === 'object') ||
         // need this approach to show border between
         border === 'between' ||
         border?.side === 'between' ||
--- a/src/js/components/Box/StyledBox.js
+++ b/src/js/components/Box/StyledBox.js
@@ -213,12 +213,25 @@ const gapStyle = (directionProp, gap, responsive, wrap, theme) => {
           theme.global.edgeSize[gap.column] || gap.column
         };`,
       );
+      if (responsiveMetric) {
+        styles.push(breakpointStyle(breakpoint, `gap: ${responsiveMetric};`));
+      }
     } else if (gap.row !== undefined) {
       styles.push(`row-gap: ${theme.global.edgeSize[gap.row] || gap.row};`);
+      if (responsiveMetric) {
+        styles.push(
+          breakpointStyle(breakpoint, `row-gap: ${responsiveMetric};`),
+        );
+      }
     } else if (gap.column !== undefined) {
       styles.push(
         `column-gap: ${theme.global.edgeSize[gap.column] || gap.column};`,
       );
+      if (responsiveMetric) {
+        styles.push(
+          breakpointStyle(breakpoint, `column-gap: ${responsiveMetric};`),
+        );
+      }
     }
   }
   if (directionProp === 'column' || directionProp === 'column-reverse') {
--- a/src/js/components/Box/StyledBox.js
+++ b/src/js/components/Box/StyledBox.js
@@ -213,11 +213,9 @@ const gapStyle = (directionProp, gap, responsive, wrap, theme) => {
           theme.global.edgeSize[gap.column] || gap.column
         };`,
       );
-    }
-    if (gap.row !== undefined && gap.column === undefined) {
+    } else if (gap.row !== undefined) {
       styles.push(`row-gap: ${theme.global.edgeSize[gap.row] || gap.row};`);
-    }
-    if (gap.row === undefined && gap.column !== undefined) {
+    } else if (gap.column !== undefined) {
       styles.push(
         `column-gap: ${theme.global.edgeSize[gap.column] || gap.column};`,
       );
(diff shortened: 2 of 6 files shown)


src/js/components/Box/StyledBox.js
```diff
@@ -206,6 +206,23 @@ const gapStyle = (directionProp, gap, responsive, wrap, theme) => {
...
   const styles = [];
+  if (typeof gap === 'object') {
+    if (gap.row !== undefined && gap.column !== undefined) {
+      styles.push(
+        `gap: ${theme.global.edgeSize[gap.row] || gap.row} ${
+          theme.global.edgeSize[gap.column] || gap.column
+        };`,
+      );
+    }
+    if (gap.row !== undefined && gap.column === undefined) {
```
taysea, line 217: can we simplify the logic a bit by making these `else if` statements and removing the second check?
```
if (gap.row !== undefined && gap.column !== undefined) { ... }
else if (gap.row !== undefined) { ... }
else if (gap.column !== undefined) { ... }
```

src/js/components/Box/StyledBox.js
```diff
@@ -206,6 +206,23 @@ const gapStyle = (directionProp, gap, responsive, wrap, theme) => {
   const responsiveMetric = responsive && breakpoint && breakpoint.edgeSize[gap];
 
   const styles = [];
+  if (typeof gap === 'object') {
```
taysea, line 209: This is looking good. We are missing the `breakPointStyle` that the other gap values have. I think we should add that, can follow pattern of what is done below in this function:
```
if (responsiveMetric) ...
```

src/js/components/Box/__tests__/Box-layout-test.tsx
```diff
@@ -229,6 +229,23 @@ describe('Box', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
+  test('row and column gap', () => {
+    const { container } = render(
```
taysea, line 233:
```suggestion
    const { asFragment } = render(
```

src/js/components/Box/__tests__/Box-layout-test.tsx
```diff
@@ -229,6 +229,23 @@ describe('Box', () => {
...
+          <Box />
+        </Box>
+        <Box as="span" gap="small">
+          <span>first</span>
+          <span>second</span>
+        </Box>
+      </Grommet>,
+    );
+
+    expect(container.firstChild).toMatchSnapshot();
```
taysea, line 246:
```suggestion
    expect(asFragment()).toMatchSnapshot();
```
more recently, we've been using `asFragment` instead of container -- can you regenerate snapshots when you merge this?

src/js/components/Box/index.d.ts
```diff
@@ -38,7 +38,7 @@ export interface BoxProps {
   flex?: 'grow' | 'shrink' | boolean | { grow?: number; shrink?: number };
   fill?: FillType;
   focusIndicator?: boolean;
-  gap?: GapType;
+  gap?: GapType | object;
```
taysea, line 41:
```suggestion
  gap?: GapType | { row?: string; gap?: string };
```
can we be more specific?

src/js/components/Box/propTypes.js
```diff
@@ -153,6 +153,7 @@ if (process.env.NODE_ENV !== 'production') {
         'xlarge',
       ]),
       PropTypes.string,
+      PropTypes.object,
```
taysea, line 156:
```suggestion
      PropTypes.shape({ row: PropTypes.string, column: PropTypes.string }),
```
also just being more specific

src/js/components/Box/Box.js
```diff
@@ -221,7 +221,7 @@ const Box = forwardRef(
         fillProp={fill}
         focus={focus}
         gap={
-          (boxOptions?.cssGap || cssGap) &&
+          (boxOptions?.cssGap || cssGap || typeof gap === 'object') &&
```
taysea, line 224: On line 117 above, I think we also want it to read:
```
!(boxOptions.cssGap || cssGap || typeof gap === 'object')
```

Screen Shot 2024-02-22 at 4 40 05 PM

Also, from a documentation standpoint, we should call out that this only works if you are using `cssGap`

src/js/components/Box/propTypes.js
britt6612: good idea

src/js/components/Box/Box.js
britt6612: I wonder if we are having two different thoughts here.. I was thinking you don't need `cssGap` for this to work it looks if `gap` is an object if it is then it uses the `gapStyle` from the styledButton. Are you thinking we need `cssGap` as well as `gap={{row: 'small', column: 'medium'}}`

src/js/components/Box/StyledBox.js
britt6612: Ill take a look

britt6612: done

src/js/components/Box/StyledBox.js
```diff
@@ -211,6 +211,34 @@ const gapStyle = (directionProp, gap, responsive, wrap, theme) => {
     if (responsiveMetric) {
```
taysea, line 211: Should the check of if `typeof gap === 'object'` come before this `if (directionProp === 'column' || directionProp === 'column-reverse')` block?

src/js/components/Box/index.d.ts
```diff
@@ -38,7 +38,7 @@ export interface BoxProps {
   flex?: 'grow' | 'shrink' | boolean | { grow?: number; shrink?: number };
   fill?: FillType;
   focusIndicator?: boolean;
-  gap?: GapType;
+  gap?: GapType | { row?: string; column?: string };
```
taysea, line 41:
```suggestion
  gap?: GapType | { row?: GapType; column?: GapType };
```
realizing we could be a bit more specific here.

britt6612: good call!

taysea (approved):
LGTM, thanks!

jcfilben (approved):
Looks good!

PR #7148, opened 2024-02-23
Run form submit validation when submit is fired directly on input

taysea:
#### What does this PR do?

Fixing bug introduced by https://github.com/grommet/grommet/pull/7005, where if a submit event is fired directly on an input element, the form validations won't run.

This ensures that the event is either being trigged by the form or an input within the form, but still keeps the initial fix to avoid the React portals issue noted in the original issue.

Locally created test and added jest test:

```
test('should validate when submit is fired directly on input', () => {
    render(
      <Grommet>
        <Form>
          <FormField
            label="My field"
            name="my-field"
            htmlFor="my-field"
            validate={{
              regexp: /^some-value/,
              message: 'Invalid email address',
              status: 'error',
            }}
          >
            <TextInput
              name="my-field"
              id="my-field"
              placeholder="placeholder text"
            />
          </FormField>
          <Button label="Submit" type="submit" />
        </Form>
      </Grommet>,
    );

    const element = screen.getByPlaceholderText('placeholder text');
    fireEvent.submit(element);
    expect(screen.getByText('Invalid email address')).toBeTruthy();
  });
```

Backwards compatible.

### DIFF
--- a/src/js/components/Form/Form.js
+++ b/src/js/components/Form/Form.js
@@ -714,7 +714,11 @@ const Form = forwardRef(
           // bubbles up to the "outer" form even though in the DOM the portal
           // doesn't render as child of the "outer" form.
           // https://legacy.reactjs.org/docs/portals.html#event-bubbling-through-portals
-          if (formRef.current && event.target === formRef.current) {
+          if (
+            formRef.current &&
+            (event.target === formRef.current ||
+              event.target.form === formRef.current)
+          ) {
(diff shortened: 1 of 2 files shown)


MikeKingdom (approved):
This makes a lot of sense. My hope was element.form was set like I think it is :) and that seems to jive with this solution.

halocline (approved):

PR #7155, opened 2024-03-05
add toggleButtonGroup

britt6612:
#### What does this PR do?
This is a proposal for a ToggleButtonGroup component to follow on which design work that has been done.

The ToggleButtonGroup is a RadioButtonGroup if `type=single` and a CheckboxGroup if `type=multiple'. This component provides the layout of the container which has rounded edges, direction='row', and a right border that is served as a separator since `gap='none`. 

```
// API
  options: (
    | string
    | number
    | boolean
    | {
        disabled?: boolean;
        id?: string;
      }
  )[];
value?: string | number | boolean | object;
onChange?:  (event: React.ChangeEvent<HTMLInputElement>) => void;
type?: single | multiple
```

// default Props 
```
  const toggleButtonGroupProps = {
    direction: 'row',
    margin: 'none',
    gap: 'none',
    border: true,
    round: 'xsmall',
  };
```

We can move default props into theme for users to change how the styling will be. We can also discuss what we want in base.js

To-do:

- [ ] add testing
- [ ] add stories

## Questions

1. should the size of the box stay the same what if one label is larger than others? 
2. As of now I had to have a toggleButtonGroup Context because checkbox always shows value and label. (is this correct?)
4. Do we like the prop `type` for changing between `checkboxGroup` and `RadioButtonGroup` 
5. Need to have an option of selecting just one but you can also select none. So one or none.
6. what hover behavior do we want?

[ToggleButtonGroup.js](https://github.com/grommet/grommet/compare/master...britt6612:Add-toggleButtonGroup?expand=1#diff-8002567d9fcaff5455a46ebc6975147afce9ac0de841e8f7140021f9b39d613a)
storybook
storybook

- [ ] - Need to add tests

NEW toggleButtonGroup component was needed 
https://github.com/grommet/hpe-design-system/issues/3739
yes
yes
 backwards compatible

### DIFF
--- a/src/js/components/Button/stories/CustomThemed/Typescript.tsx
+++ b/src/js/components/Button/stories/CustomThemed/Typescript.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 
 import { ThemeType } from 'grommet/themes';
-import { Box, Button, Grommet } from 'grommet';
+import { Box, Button, Tag, Grommet } from 'grommet';
 
 // Type annotations can only be used in TypeScript files.
 // Remove ': ThemeType' if you are not using Typescript.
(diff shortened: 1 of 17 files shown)


jcfilben:
> 1. should the size of the box stay the same what if one label is larger than others? 

No, I think it is okay for the boxes to be different sizes based on the content

> 2. As of now I had to have a toggleButtonGroup Context because checkbox always shows value and label. (is this correct?)

I think this is okay, I need to think about it and look at the code a little more

> 3. Do we like the prop `type` for changing between `checkboxGroup` and `RadioButtonGroup` 

It seems like `type` could support 4 different options (not sure about naming)
`single` - one option at most can be selected, but there can also be no options selected.
`singleAlwaysActive` - only one option can be selected and there always needs to be an active option.
`multiple` - multiple or no options can be selected.
`multipleAlwaysActive` - at least one option must always be selected but multiple can be selected.

> 4. Need to have an option of selecting just one but you can also select none. So one or none.

See response to above question

> 5. what hover behavior do we want?

Let's check with designers but I personally like the background change to a grey color and no border changes.

---
I'm wondering if internally we should support a `plain` prop for CheckBoxGroup and RadioButtonGroup that removes the pad and border so that those can be driven within the ToggleButtonGroup

britt6612:
closing in favor of #7172

PR #7158, opened 2024-03-06
List keyboard behavior test and fix lint issues

jcfilben:
#### What does this PR do?
Modifies one of the list tests to use the 'space' key on the onOrder buttons. The test covers behavior added in https://github.com/grommet/grommet/pull/7140.

https://github.com/grommet/grommet/pull/7140 introduced a lint error because the dependency array for onSelectOption didn't include everything. I ended up removing the `useCallback` for that function. I don't think we were getting much value from `useCallback` in this case.

I modified our lint rules to flag react-hooks/exhaustive-deps as an error instead of just a warning. I was not able to find a way to have lint fail on warnings (but I didn't spend a ton of time looking, so I could have missed something).

no
no
backwards compatible

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -64,7 +64,7 @@
     "react/sort-comp": 0,
     "react/state-in-constructor": 0,
     "react/static-property-placement": 0,
-    "react-hooks/exhaustive-deps": 1,
+    "react-hooks/exhaustive-deps": 2,
     "react-hooks/rules-of-hooks": 2,
     "semi": [2, "always"]
   },
--- a/src/js/components/List/List.js
+++ b/src/js/components/List/List.js
@@ -1,11 +1,4 @@
-import React, {
-  Fragment,
-  useCallback,
-  useContext,
-  useMemo,
-  useRef,
-  useState,
-} from 'react';
+import React, { Fragment, useContext, useMemo, useRef, useState } from 'react';
 import styled, { ThemeContext } from 'styled-components';
 
 import { DataContext } from '../../contexts/DataContext';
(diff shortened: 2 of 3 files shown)


src/js/components/List/__tests__/List-test.js
```diff
@@ -581,7 +581,7 @@ describe('List onOrder', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
-  test('Keyboard move up', () => {
+  test.only('Keyboard move up', () => {
```
jcfilben, line 584:
```suggestion
  test('Keyboard move up', () => {
```

src/js/components/List/__tests__/List-test.js
```diff
@@ -600,11 +600,11 @@ describe('List onOrder', () => {
     // beta's up arrow control should be active
     expect(container.firstChild).toMatchSnapshot();
     fireEvent.keyDown(getByText('alpha'), {
-      key: 'Enter',
-      keyCode: 13,
-      which: 13,
+      key: 'Space',
```
taysea, line 603: Did we need to change this test? Seems like `Enter` is supported

src/js/components/List/List.js
```diff
@@ -252,54 +245,41 @@ const List = React.forwardRef(
...
-            onClickItem(adjustedEvent);
-            sendAnalytics({
-              type: 'listItemClick',
-              element: listRef.current,
-              event: adjustedEvent,
-              item: data[active],
-              index: active,
-            });
-          }
+    const onSelectOption = (event) => {
```
taysea, line 248: Pausing about removing `useCallback` vs just adding the needed dependencies. Reading docs about function caching, is there value to keeping it if it wasn't causing bugs?

https://react.dev/reference/react/useCallback

.eslintrc
```diff
@@ -64,7 +64,7 @@
     "react/sort-comp": 0,
     "react/state-in-constructor": 0,
     "react/static-property-placement": 0,
-    "react-hooks/exhaustive-deps": 1,
+    "react-hooks/exhaustive-deps": 2,
```
taysea, line 67: I'm good with this.

src/js/components/List/__tests__/List-test.js
britt6612: nice ☝️ thinking

taysea (approved):
Looks good!

britt6612 (approved):

PR #7164, opened 2024-03-14
fix(list): updates typescript types for children prop

### LINKED ISSUE #7160: [TypeScript] List component types don't correctly support render props
### Expected Behavior

When using the `List` component in TS with a render prop function as `props.children`, the types should be correctly inferred, either from an explicit passing of the generic type or inference from the passing of typed data.

### Actual Behavior

The types for `props.children` seem to have been set in such a way that inference expects a union of `(((...args: any[]) => Element) & (string | number | boolean | ReactElement<any, string | JSXElementConstructor<any>> | Iterable<ReactNode> | ReactPortal | null)) | undefined`.

<figure>
TS errors when passing a generic
<br />
<figcaption>In the current state, it seems that typescript expects a string for the children prop.</figcaption>
</figure>

<br />
<br />
<br />

<figure>
TS errors when passing a string as recommended in the above screenshot
<br />
<figcaption>When passing a string as recommended above, TypeScript provides a more useful response showing the intersection type expected.</figcaption>
</figure>

### Steps to Reproduce

See screenshots above - PR with fix incoming

### Your Environment

- Grommet version: 2.35.0
- TypeScript version: 4.9.5

Tbhesswebber:
#### What does this PR do?

This PR fixes the TypeScript types for `ListExtendedProps` to align with the documentation.

There isn't much to this PR.  I would start with reading through https://github.com/grommet/grommet/issues/7160 and then pulling the code down to test in a development environment.

Migrated the tests to Typescript and modified the object children test

Use it in a development environment.

Closes #7160

<figure>
Screenshot of intellisense in VSCode
<figcaption>Type inference works properly with this change in the types</figcaption>
</figure>

No, although I'm not sure what the type for the third argument should be based on the docs, so I left it a vague `Record<string, any>`

No

Backwards compatible unless people leveraged module augmentation or similar to get around the issue.


src/js/components/List/index.d.ts
```diff
@@ -54,9 +58,9 @@ export interface ListProps<ListItemType> {
   };
   margin?: MarginType;
   onActive?: (index: number) => void;
-  onClickItem?:
-    | ((event: React.MouseEvent) => void)
-    | ((event: { item?: ListItemType; index?: number }) => void);
+  onClickItem?: (
+    event: React.MouseEvent & { item: ListItemType; index: number },
+  ) => void;
```
Tbhesswebber, line 63: Based on a read-through of the code, the correct type is a function with the intersection of the event and what it gets modified with as the single argument

src/js/components/List/index.d.ts
```diff
@@ -76,7 +80,7 @@ export interface ListExtendedProps<ListItemType>
     ulProps {}
 
 declare const List: <ListItemType = string | {}>(
-  p: React.PropsWithChildren<ListExtendedProps<ListItemType>>,
+  p: ListExtendedProps<ListItemType>,
```
Tbhesswebber, line 83: This is what was clobbering the type in such a weird way

src/js/components/List/__tests__/List-test.tsx
```diff
@@ -454,7 +454,7 @@ describe('List events', () => {
   test('should apply pagination styling', () => {
     const { container } = render(
       <Grommet>
-        <List data={data} paginate={{ background: 'red', margin: 'large' }} />
+        <List data={data} paginate={{ margin: 'large' }} />
```
Tbhesswebber, line 457: It looks like `background` isn't actually a valid paginate prop

src/js/components/List/__tests__/List-test.tsx
```diff
@@ -552,9 +552,13 @@ describe('List onOrder', () => {
 
   test('Mouse move down', () => {
     const { container } = render(<App />);
+    const $element = container.querySelector('#alphaMoveDown');
+
+    if (!$element)
+      throw new Error('Cannot find element with id "alphaMoveDown"');
```
Tbhesswebber, line 558: This short-circuits so that `fireEvent` doesn't get passed `null`

src/js/components/List/__tests__/List-test.tsx
```diff
@@ -627,7 +631,7 @@ describe('List onOrder with action', () => {
               onOrder(newData);
             }}
             // eslint-disable-next-line react/no-unstable-nested-components
-            action={(item, index) => (
+            action={(_, index) => (
```
Tbhesswebber, line 634: There was a lint error about unused arguments and the `_` fixes that

src/js/components/List/index.d.ts
```diff
@@ -43,7 +43,11 @@ export interface ListProps<ListItemType> {
     | string[]
     | { light: string | string[]; dark: string | string[] };
   border?: BorderType;
-  children?: (...args: any[]) => any;
+  children?: (
+    item: ListItemType,
+    index: number,
+    state?: { active: boolean },
+  ) => any;
```
jcfilben, line 50: I am getting a typescript error on the `active` prop if I try to do something like this: `(item, index, {active}) => {}`.

src/js/components/List/__tests__/List-test.tsx
jcfilben: Good catch, I think removing the background here is fine

src/js/components/List/index.d.ts
Tbhesswebber: Ah, I misread the documentation - I thought the entire state object was optional rather than just the properties within.  Fix incoming!

Tbhesswebber: Just ran into an issue - I modified the test to validate and it seems that it's actually the entire state object that is optional.  Pinging on Slack for follow-up semi-synchronous conversation and will report back the decision for posterity's sake.

image

jcfilben: Discussed offline, we are going to hold off on trying to support destructuring the active prop in the typescript definition. I think we can move forward with the types that are defined in this PR.

jcfilben (approved):
Looks good!

taysea (approved):

src/js/components/List/index.d.ts
```diff
@@ -54,9 +58,9 @@ export interface ListProps<ListItemType> {
   };
   margin?: MarginType;
   onActive?: (index: number) => void;
-  onClickItem?:
-    | ((event: React.MouseEvent) => void)
-    | ((event: { item?: ListItemType; index?: number }) => void);
+  onClickItem?: (
+    event: React.MouseEvent & { item: ListItemType; index: number },
```
britt6612, line 62: should this be optional? 
`event: React.MouseEvent & { item?: ListItemType; index?: number },`

Tbhesswebber: Not according to [the documentation](https://v2.grommet.io/list#onClickItem), but happy to change if desired

britt6612: I think that's fine I read through and matched the documentation.

britt6612 (approved):

PR #7165, opened 2024-03-15
Styled components v6 support

jcfilben:
#### What does this PR do?
Upgrade to support both styled components v5 and v6.

The code in this PR was previously released in https://github.com/grommet/grommet/releases/tag/v2.35.0-alpha.1. This PR updates the code with the latest changes since that release and is aimed towards the master branch.

#### Linking to previous PRs and issues which have more context:

Approved direction:
* Current accepted approach (released in beta) - https://github.com/grommet/grommet/pull/6947

Explored approaches, but ultimately not chosen:
* StyleSheetManager (confusing behavior, potentially risky in different configurations) - see description of https://github.com/grommet/grommet/pull/6947 for more detail
* Transient props approach (too much work required) - https://github.com/grommet/grommet/pull/6873
* Word doc for transient approach - https://docs.google.com/document/d/1SLnJLBdVdpBByUTPaCzlwKksViRl9Q_cZ1yagnU7C7o/edit#heading=h.t103iljemjoj

https://github.com/grommet/hpe-design-system/issues/3528
https://github.com/grommet/hpe-design-system/issues/3476

no
yes
backwards compatible


package.json
```diff
@@ -189,6 +190,9 @@
         "statements": 86.6
       }
     },
+    "moduleNameMapper": {
+      "^styled-components": "styled-components/dist/styled-components.browser.cjs.js"
```
taysea, line 194: I forget why we needed this -- is it documented anywhere why we have this? Would be helpful to keep note for future reference.

src/js/components/Select/StyledSelect.js
```diff
@@ -8,9 +9,10 @@ import {
   selectedStyle,
   controlBorderStyle,
   sizeStyle,
+  styledComponentsConfig,
 } from '../../utils';
 
-export const StyledContainer = styled.div`
+export const StyledContainer = styled(Box)`
```
taysea, line 15: @jcfilben Can you remind me why in some cases we changed to "Box" vs others we kept as div?

would be good to document this decision for future reference.

jcfilben: There were some places that were using `styled.div` and `as={Box}` so I think we decided to just simplify to `styled(Box)`

package.json
jcfilben: I believe we were having issues with the jest tests, the suggestion to add this line came from https://github.com/styled-components/jest-styled-components/issues/144#issuecomment-394184823

.circleci/config.yml
```diff
@@ -117,6 +117,7 @@ jobs:
             git config --global user.email "grommet@hpe.com"
             yarn release-stable
   release-styled-components-beta-stable:
+    # needed for styled components beta pre-release
```
britt6612, line 120: do we still need? of we are going to release regular minor version in grommet?

britt6612 (approved):

package.json
```diff
@@ -51,9 +51,10 @@
   "peerDependencies": {
     "react": "^16.6.1 || ^17.0.0 || ^18.0.0",
     "react-dom": "^16.6.1 || ^17.0.0 || ^18.0.0",
-    "styled-components": "^5.1.0"
+    "styled-components": "^5.1.0 || ^6.0.0"
   },
   "dependencies": {
+    "@emotion/is-prop-valid": "^1.2.1",
```
Tbhesswebber, line 57: Is there some rationale for this being a direct dependency rather than a peer dependency?  It is unintuitive (to me, at least) that a package used specifically for a peer dependency would be a direct dependency.

package.json
```diff
@@ -51,9 +51,10 @@
   "peerDependencies": {
     "react": "^16.6.1 || ^17.0.0 || ^18.0.0",
     "react-dom": "^16.6.1 || ^17.0.0 || ^18.0.0",
-    "styled-components": "^5.1.0"
+    "styled-components": "^5.1.0 || ^6.0.0"
```
Tbhesswebber, line 54: Supporting both seems like it might be risky with significant breaking changes specifically because of how v6 handles nesting as compared to v5.  You can find that documentation [here](https://github.com/styled-components/styled-components/blob/22e8b7f233e12500a68be4268b1d79c5d7f2a661/packages/styled-components/src/types.ts#L152).

package.json
taysea: @jcfilben Do we need to add this directly? Or is it included with styled-components v6?

package.json
taysea: Hey @Tbhesswebber thanks for commenting. Could you elaborate on your concerns? Definitely want to hear them out as we're considering adding support for v6 as well.

With these changes, we are taking care of shouldForwardProp (mentioned in your link above) on all styled instances within grommet components (see styles.js under files changed). This will ensure proper styling if someone is using styled components v6.

package.json
Tbhesswebber: It's not included, but it can be configured globally through styled components (and likely will in many cases). The alternative is to migrate to transient props, which a lot of people seem to be balking at

package.json
Tbhesswebber: I updated the link above (I was hasty with copy pasta, it seems).

```ts
const Foo = styled.button`
  :hover {
    background: black;
  }
`;
```

The above snippet outputs different CSS in v5 and v6. In v5, there's an implicit `&` prepended to `:hover`, so it's applied directly to the element. In v6, it's prepended with `& ` to align with the nesting spec for CSS, so the hover state applies to children of the button rather than the button itself.

package.json
jcfilben: @emotion/is-prop-valid is a dependency of styled-components but since we need to import and use `isPropValid` within our own code I added it as a dependency on Grommet. Although it is a styled-components dependency, I didn't see a way to import `isPropValid` through styled-components

package.json
jcfilben: I went through and added `&` to the places in Grommet where the `&` was being implicitly added by styled components v5 so that we will get the same behavior in v5 and v6. (See the changes in StyledCheckBox.js for an example)

package.json
Tbhesswebber: v6 dropped the direct dependency, making it a peer dependency. They migrated to transient props instead

https://styled-components.com/docs/faqs#shouldforwardprop-is-no-longer-provided-by-default

package.json
Tbhesswebber: Sorry for not being clear, my concern is moreso in the realm of consumers having this syntax in the `.extends` properties in their configuration or in `styled(SomeGrommetComponent)` declarations. Breaking changes that likely won't be documented in Grommet's changelog because they aren't actually changes to Grommet

package.json
jcfilben: We explored transient props but ultimately decided to use is-prop-valid to filter out styling props from being passed to the dom. Some more context on our explorations:
https://github.com/grommet/grommet/pull/6947
https://github.com/grommet/grommet/pull/6941
https://github.com/grommet/grommet/tree/styled-components-v6 - branch where we started converting to transient props

package.json
jcfilben: I think we can add a note to the Grommet release where we add styled components v6 support, so that this doesn't catch anyone by surprise. We have a similar message on the alpha release we did for styled components v6 support: https://github.com/grommet/grommet/releases/tag/v2.35.0-alpha.1.
But ultimately since the breaking change is not caused by Grommet and is a result of the styled-components upgrade then I think the person upgrading would just have to refer to the styled-components upgrade documentation which calls out the breaking changes.
Since we will have support for both v5 and v6 of styled components it is up to the consumer if they want to upgrade to v6 or not.

package.json
taysea: Thanks for jogging my memory Jessica!

src/js/components/FormField/__tests__/__snapshots__/FormField-test.js.snap
```diff
@@ -522,7 +434,6 @@ exports[`FormField contentProps 1`] = `
           <input
             autocomplete="off"
             class="c5"
-            value=""
```
taysea, line 525: @jcfilben any concerns around this attribute being removed? I'm going to look into it a bit more

taysea: I looked a little more, this is a change in the DOM although there's no break to functionality. This is in cases where the input is in a form and the input value is never updating in the DOM since it's managed by the form itself (this is the case regardless of styled-components version) so it moreso seems like it's just cleaning up an empty DOM attribute?

jcfilben: Agree that it looks like a cleanup. There are other snapshots in this PR that have the same change. It looks like styled components is just stricter and doesn't let an empty value pass into the DOM. I think this snapshot change is fine

taysea: I feel comfortable with this cleanup. It mirrors the behavior of [Radix forms](https://www.radix-ui.com/primitives/docs/components/form) as well.

taysea (approved):
This looks good. We just need to update our bundleSize in package.json before we merge.

halocline (approved):
Got all my questions answered. Looks solid. Thank you.

PR #7168, opened 2024-03-19
Fix vertical pad on readOnlyCopy button

taysea:
#### What does this PR do?

`readOnlyCopy` button should not apply vertical pad, and should instead just fill the height of input. This was the originally intended behavior that seemed to not be implemented as expected.

TextInput or DateInput

Local in storybook

Captura de Pantalla 2024-03-19 a la(s) 2 03 46 p m

Backwards compatible.


jcfilben (approved):
Thanks for catching and fixing this

halocline (approved):

PR #7172, opened 2024-03-22
New ToggleButtonGroup/ToggleGroup

britt6612:
#### What does this PR do?
This is a proposal for a ToggleButtonGroup component to follow on which design work that has been done.

The ToggleButtonGroup is a Box with Buttons within it. The default type is a user can select one option from the group as far as if a user wants more options they pass `multiple` similar to how Select uses the `multiple` prop. This component provides the layout of the container which has rounded edges, direction='row', and a right border that is served as a separator since `gap='none`. 

```
// API
  options: (
    | string
    | number
    | boolean
    | {
        disabled?: boolean;
        id?: string;
      }
  )[];
onChange?:  (event: React.ChangeEvent<HTMLInputElement>) => void;
multiple?: boolean
```

// default Props 
```
  const toggleButtonGroupProps = {
        round="xsmall"
        direction="row"
        border
        style={{ width: 'fit-content' }}
  };
```

We can move default props into theme for users to change how the styling will be. We can also discuss what we want in base.js

To-do:

- [ ] add testing
- [ ] add stories
- [ ] keyboard func

[ToggleButtonGroup.js](https://github.com/grommet/grommet/compare/master...britt6612:Add-toggleButtonGroup?expand=1#diff-8002567d9fcaff5455a46ebc6975147afce9ac0de841e8f7140021f9b39d613a)
storybook
storybook

- [ ] - Need to add tests

NEW toggleButtonGroup component was needed 
https://github.com/grommet/hpe-design-system/issues/3739
yes
yes
 backwards compatible

### DIFF
--- a/src/js/components/ToggleButtonGroup/README.md
+++ b/src/js/components/ToggleButtonGroup/README.md
@@ -1,0 +1,2 @@
+## ToggleButtonGroup
+Documentation for this component: https://v2.grommet.io/toggleButtonGroup
--- a/src/js/components/ToggleButtonGroup/ToggleButtonGroup.js
+++ b/src/js/components/ToggleButtonGroup/ToggleButtonGroup.js
@@ -1,0 +1,90 @@
+import React, { useState } from 'react';
+import styled from 'styled-components';
+import { Box } from '../Box';
+import { Button } from '../Button';
+import { Keyboard } from '../Keyboard';
+import { edgeStyle } from '../../utils/styles';
+
+// to overcome `plain` styling due to (icon && !label) condition
+// in buttons without theme.button.default, apply the padding here
+const StyledButton = styled(Button)`
+  border-radius: ${(props) => props.theme.global.control.border.radius};
+  ${(props) =>
+    !props.theme.button.default
+      ? edgeStyle('padding', props.pad, false, undefined, props.theme)
+      : ''}
+`;
+
+// default
+// make sure defaultValue html native behavior works? if it does work maybe not having defaultOption
+
+const ToggleButtonGroup = ({
+  options,
+  onChange,
+  multiple,
+  ...rest
+}) => {
+  const [selectedOption, setSelectedOption] = useState([]);
+
+  const handleToggle = (option) => {
+    if (!multiple) {
+      setSelectedOption([option]);
+      if (onChange) {
+        onChange([option]);
+      }
+    } else {
+      const newSelectedOptions = selectedOption.includes(option)
+        ? selectedOption.filter((item) => item !== option)
+        : [...selectedOption, option];
+
+      setSelectedOption(newSelectedOptions);
+      if (onChange) {
+        onChange(newSelectedOptions);
+      }
+    }
+  };
+
+  const flatOptions = options.map((option) =>
+    typeof option === 'object' ? option.label : option,
+  );
+
+  return (
+    <Keyboard target="document">
+      <Box
+        round="xsmall"
+        direction="row"
+        border
+        style={{ width: 'fit-content' }}
+      >
+        {options.map((option, index) => (
+            <Box
+              key={index}
+              role="group"
+              // should have color in theme
+              border={
+                flatOptions.indexOf(option.label || option) <
+                flatOptions.length - 1
+                  ? { side: 'right', color: 'border' }
+                  : undefined
+              }
+              {...rest}
+            >
+              <StyledButton
+                role="radio"
+                onClick={() => handleToggle(option)}
+                icon={option.label}
+                label={option.label ? undefined : option}
+                pad="small"
+                onChange={onChange}
+              />
+            </Box>
+          ))}
+      </Box>
+    </Keyboard>
+  );
+};
+
+ToggleButtonGroup.displayName = 'ToggleButtonGroup';
+// ToggleButtonGroup.propTypes = ToggleButtonGroupPropTypes;
+
+export { ToggleButtonGroup };
--- a/src/js/components/ToggleButtonGroup/index.d.ts
+++ b/src/js/components/ToggleButtonGroup/index.d.ts
@@ -1,0 +1,27 @@
+import * as React from 'react';
+import { BoxProps } from '../Box/index';
+
+export interface ToggleButtonGroupProps {
+  multiple?: boolean;
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  options: (
+    | string
+    | number
+    | boolean
+    | {
+        disabled?: boolean;
+        id?: string;
+        label?: string | React.ReactNode;
+        value: string | number | boolean;
+      }
+  )[];
+}
+
+export interface ToggleButtonGroupExtendedProps
+  extends ToggleButtonGroupProps,
+    BoxProps,
+    Omit<JSX.IntrinsicElements['div'], 'children' | 'onClick' | 'onChange'> {}
+
+declare const ToggleButtonGroup: React.FC<ToggleButtonGroupExtendedProps>;
+
+export { ToggleButtonGroup };
--- a/src/js/components/ToggleButtonGroup/propTypes.js
+++ b/src/js/components/ToggleButtonGroup/propTypes.js
@@ -1,0 +1,27 @@
+import PropTypes from 'prop-types';
+
+let PropType = {};
+if (process.env.NODE_ENV !== 'production') {
+  PropType = {
+    multiple: PropTypes.bool,
+    onChange: PropTypes.func,
+    options: PropTypes.oneOfType([
+      PropTypes.arrayOf(PropTypes.string),
+      PropTypes.arrayOf(PropTypes.number),
+      PropTypes.arrayOf(PropTypes.bool),
+      PropTypes.arrayOf(
+        PropTypes.shape({
+          disabled: PropTypes.bool,
+          id: PropTypes.string,
+          label: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+          value: PropTypes.oneOfType([
+            PropTypes.string,
+            PropTypes.number,
+            PropTypes.bool,
+          ]).isRequired,
+        }),
+      ),
+    ]).isRequired,
+  };
+}
+export const ToggleButtonGroupPropTypes = PropType;
--- a/src/js/components/ToggleButtonGroup/index.js
+++ b/src/js/components/ToggleButtonGroup/index.js
@@ -1,0 +1,1 @@
+export { ToggleButtonGroup } from './ToggleButtonGroup';
--- a/src/js/components/ToggleButtonGroup/stories/newstory.js
+++ b/src/js/components/ToggleButtonGroup/stories/newstory.js
@@ -1,12 +1,9 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
 import { Box, Heading, ToggleButtonGroup } from 'grommet';
 import {
   List,
   MapLocation,
   Table,
-  Bold,
-  Italic,
-  Underline,
 } from 'grommet-icons';
 
 
@@ -29,8 +26,6 @@ const optionsIcons = [
 export const newStory = () => {
   const [multipleSelectedColors, setMultipleSelectedColors] = useState(['Red']);
   const [selectedOption, setSelectedOption] = useState('');
-
-  console.log(selectedOption);
 
   const handleToggle = (option) => {
     setSelectedOption(option);
--- a/src/js/components/ToggleButtonGroup/stories/newstory.js
+++ b/src/js/components/ToggleButtonGroup/stories/newstory.js
@@ -1,0 +1,70 @@
+import React, { useState, useEffect } from 'react';
+import { Box, Heading, ToggleButtonGroup } from 'grommet';
+import {
+  List,
+  MapLocation,
+  Table,
+  Bold,
+  Italic,
+  Underline,
+} from 'grommet-icons';
+
+
+const optionsIcons = [
+  {
+    label: <List />,
+    value: 'list',
+  },
+  {
+    label: <Table />,
+    value: 'table',
+  },
+  {
+    label: <MapLocation />,
+    value: 'map',
+  },
+];
+
+
+export const newStory = () => {
+  const [multipleSelectedColors, setMultipleSelectedColors] = useState(['Red']);
+  const [selectedOption, setSelectedOption] = useState('');
+
+  console.log(selectedOption);
+
+  const handleToggle = (option) => {
+    setSelectedOption(option);
+  };
+
+  const handleColorsToggle = (colors) => {
+    setMultipleSelectedColors(colors);
+  };
+
+  return (
+    <Box gap="large" pad="large">
+      <Box>
+        <Heading>Selected colors: {multipleSelectedColors}</Heading>
+        <ToggleButtonGroup
+          options={['Red', 'Green', 'Blue']}
+          onChange={handleColorsToggle}
+          multiple
+        />
+      </Box>
+      <Box>
+        <Heading>Selected color: {selectedOption}</Heading>
+        <ToggleButtonGroup
+          options={['Option 1', 'Option 2', 'Option 3']}
+          onChange={handleToggle}
+          defaultValue="Option 3"
+        />
+      </Box>
+      <ToggleButtonGroup
+          options={optionsIcons}
+        />
+    </Box>
+  );
+};
+
+export default {
+  title: 'Input/ToggleButtonGroup/newStory',
+};
--- a/src/js/components/__tests__/__snapshots__/components-test.js.snap
+++ b/src/js/components/__tests__/__snapshots__/components-test.js.snap
@@ -5034,6 +5034,7 @@ exports[`Components loads 1`] = `
     },
     "render": [Function],
   },
+  "ToggleButtonGroup": [Function],
   "Toolbar": [Function],
   "Video": {
     "$$typeof": Symbol(react.forward_ref),
--- a/src/js/components/index.d.ts
+++ b/src/js/components/index.d.ts
@@ -88,6 +88,7 @@ export * from './TextArea';
 export * from './TextInput';
 export * from './ThumbsRating';
 export * from './Tip';
+export * from './ToggleButtonGroup';
 export * from './Toolbar';
 export * from './Video';
 export * from './WorldMap';
--- a/src/js/components/index.js
+++ b/src/js/components/index.js
@@ -90,6 +90,7 @@ export * from './TextArea';
 export * from './TextInput';
 export * from './Tip';
 export * from './Toolbar';
+export * from './ToggleButtonGroup';
 export * from './ThumbsRating';
 export * from './Video';
 export * from './WorldMap';


britt6612:
- [ ] Naming of onChange (technically this is not an onChange event so should we rename it?)
- [x] Should we have `kind` on the button to override some of the grommet default behavior. Right now we have
Kind={theme.toggleButtonGroup} however this opens up more ways to add styles should we just have `pad` and `border` in the styledButton UPDATE: decided to go with putting these in `styledButton` for now to have a more narrow start.
- [x] Some of the names in theme make sure we all agree and it makes sense. UPDATE: instead of `border.color` we are going to have `divider.color` to make more sense to the user in what is changing within the component.

src/js/components/ToggleButtonGroup/stories/controlledGroup.js
```diff
@@ -0,0 +1,44 @@
...
+          }}
+          multiple
+        />
+      </Box>
+    </Box>
+  );
+};
+
+export default {
+  title: 'Input/ToggleButtonGroup/ControlledGroup',
```
taysea, line 43:
```suggestion
  title: 'Controls/ToggleButtonGroup/Controlled',
```

src/js/components/ToggleButtonGroup/stories/controlledGroup.js
```diff
@@ -0,0 +1,44 @@
+import React, { useState } from 'react';
```
taysea, line 1: For the storybook examples, can we:
- Put these under "Controls" section as opposed to "Inputs"? 
- Capitalize the file names to be consistent with other stories?
- For story naming, maybe "Controlled", "Multiple", "Simple", "Icons"

src/js/components/ToggleButtonGroup/stories/controlledGroup.js
```diff
@@ -0,0 +1,44 @@
...
+import { Box, ToggleButtonGroup, Text } from 'grommet';
+
+export const ControlledGroup = () => {
+  const [singleControlled, setSingleControlled] = useState('c2');
+  const [multipleControlled, setMultipleControlled] = useState();
```
taysea, line 6:
```suggestion
  const [multipleControlled, setMultipleControlled] = useState([]);
```

src/js/components/ToggleButtonGroup/stories/simple.js
```diff
@@ -0,0 +1,19 @@
...
+export const Simple = () => (
+  <Box gap="large" pad="large">
+    <ToggleButtonGroup
+      options={[
+        { label: 'Choice 1', value: 'c1' },
+        { label: 'Choice 2', value: 'c2' },
+        { label: 'Choice 3', value: 'c3' },
+      ]}
+    />
+    <ToggleButtonGroup options={['Option 1', 'Option 2', 'Option 3']} />
```
taysea, line 13: Should we put this "simplest" one first?

src/js/components/ToggleButtonGroup/index.d.ts
```diff
@@ -0,0 +1,27 @@
...
+  options: (
+    | string
+    | {
+        icon?: React.ReactNode | JSX.Element;
+        id?: string;
+        label?: string | React.ReactNode;
+        value: string;
+      }
+  )[];
+  value?: string | [];
```
taysea, line 17:
```suggestion
  value?: string | string[];
```

src/js/components/ToggleButtonGroup/propTypes.js
```diff
@@ -0,0 +1,24 @@
...
+let PropType = {};
+if (process.env.NODE_ENV !== 'production') {
+  PropType = {
+    multiple: PropTypes.bool,
+    onChange: PropTypes.func,
+    options: PropTypes.oneOfType([
+      PropTypes.arrayOf(PropTypes.string),
+      PropTypes.arrayOf(
+        PropTypes.shape({
+          id: PropTypes.string,
```
taysea, line 12:
```suggestion
```

src/js/components/ToggleButtonGroup/propTypes.js
```diff
@@ -0,0 +1,24 @@
...
+          label: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
```
taysea, line 13: Missing `icon` in this object

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
```diff
@@ -0,0 +1,490 @@
...
+
+    expect(container.firstChild).toMatchSnapshot();
+  });
+
+  test('object options', () => {
+    const { container } = render(
+      <Grommet>
+        <ToggleButtonGroup
+          options={[
+            { id: 'onE', label: 'One', value: 'one' },
```
taysea, line 42:
```suggestion
            { label: 'One', value: 'one' },
```
can we remove `id` from these since we're not using it in code?

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
```diff
@@ -0,0 +1,490 @@
...
+      <Grommet>
+        <ToggleButtonGroup options={['one', 'two']} />
+      </Grommet>,
+    );
+
+    expect(container.firstChild).toMatchSnapshot();
+  });
+
+  test('object options', () => {
+    const { container } = render(
```
taysea, line 38:
```suggestion
    const { asFragment } = render(
```
We should use `asFragment` instead of `container` now.

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
```diff
@@ -0,0 +1,490 @@
...
+        <ToggleButtonGroup
+          options={[
+            { id: 'onE', label: 'One', value: 'one' },
+            { id: 'twO', label: 'Two', value: 'two' },
+          ]}
+        />
+      </Grommet>,
+    );
+
+    expect(container.firstChild).toMatchSnapshot();
```
taysea, line 49:
```suggestion
    expect(asFragment()).toMatchSnapshot();
```

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
```diff
@@ -0,0 +1,490 @@
...
+      <Grommet>
+        <ToggleButtonGroup options={['one', 'two']} defaultValue="one" />
+      </Grommet>,
+    );
+
+    const activeToggleButton = screen.getByRole('radio', { name: /one/i });
+    // Assert that the active toggle button exists
+    expect(activeToggleButton).toBeInTheDocument();
+    // Assert that the aria-checked attribute of the active toggle button is true
+    expect(activeToggleButton).toHaveAttribute('aria-checked', 'true');
```
taysea, line 63: Nice! 👍

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
```diff
@@ -0,0 +1,490 @@
...
+      'box-shadow',
+      '0 0 2px 2px #6FFFB0',
+      {
+        modifier: ':focus',
+      },
+    );
+    expect(container.firstChild).toMatchSnapshot();
+  });
+
+  test('Should allow caller to clear all values in controlled scenarios', () => {
```
taysea, line 151: Now that we changed the default behavior, let's update this test to be `should allow caller to enforce a value in controlled scenarios`

taysea: Can use the code from your "Controlled" storybook example.

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
```diff
@@ -0,0 +1,490 @@
...
+        <ToggleButtonGroup
+          options={[
+            { icon: <Add />, value: 'one' },
+            { icon: <Add />, value: 'two' },
+          ]}
+          defaultValue="one"
+        />
+      </Grommet>,
+    );
+
```
taysea, line 246: Can we add an expect here that the button with value "one" should have aria-checked?

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
```diff
@@ -0,0 +1,490 @@
...
+    expect(container).toMatchSnapshot();
+  });
+
+  test('string options with multiple prop', () => {
```
taysea, line 250:
```suggestion
  test('should render string options with multiple prop', () => {
```

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
```diff
@@ -0,0 +1,490 @@
...
+    expect(container.firstChild).toMatchSnapshot();
+  });
+
+  test('Should render when options is array of objects with multiple prop', () => {
+    const { container } = render(
+      <Grommet>
+        <ToggleButtonGroup
+          multiple
+          options={[
+            { id: 'onE', label: 'One', value: 'one' },
```
taysea, line 266:
```suggestion
            {  label: 'One', value: 'one' },
```
Remove id here

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
```diff
@@ -0,0 +1,490 @@
...
+    render(
+      <Grommet>
+        <ToggleButtonGroup
+          options={[
+            { label: 'Choice 1', value: 'c1' },
+            { label: 'Choice 2', value: 'c2' },
+            { label: 'Choice 3', value: 'c3' },
+          ]}
+          multiple
+          defaultValue="c2"
```
taysea, line 305: In `multiple` scenarios, value/defaultValue should be an array of strings

taysea: For this, can we also have multiple values as default value --> ['c2', 'c3']

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
```diff
@@ -0,0 +1,490 @@
...
+  });
+
+  test('Should put tab focus on active button when value with multiple prop', async () => {
+    const user = userEvent.setup();
+
+    const { container } = render(
+      <Grommet>
+        <ToggleButtonGroup
+          options={['one', 'two', 'three']}
+          defaultValue="three"
```
taysea, line 355: Default value should be array. Also, to check behavior can we have `['two', 'three']` then assert that "two" is the one that gets focus

taysea: Can you double check for all of the `multiple` tests that defaultValue/value are array (even array with single element) vs just a string

src/js/components/ToggleButtonGroup/index.d.ts
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import { BoxProps } from '../Box/index';
+
+export interface ToggleButtonGroupProps {
+  multiple?: boolean;
```
taysea, line 5: We should add `defaultValue` here

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
britt6612: I updated this however since onChange type is wrong Im going to comment out.

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
britt6612: added

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
britt6612: done

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
britt6612: done

src/js/components/ToggleButtonGroup/index.d.ts
britt6612: done

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
britt6612: done

src/js/components/ToggleButtonGroup/__tests__/ToggleButtonGroup-test.tsx
britt6612: done

src/js/components/ToggleButtonGroup/propTypes.js
britt6612: done

src/js/components/ToggleButtonGroup/index.d.ts
britt6612: done

src/js/components/ToggleButtonGroup/stories/simple.js
britt6612: done

src/js/components/ToggleButtonGroup/stories/controlledGroup.js
britt6612: done

jcfilben (commented):
Screen Shot 2024-04-03 at 1 52 56 PM

The hover and selected styling has rounded corners so it doesn't go all the way to the edge of the button group container

src/js/components/ToggleButtonGroup/ToggleButtonGroup.js
```diff
@@ -0,0 +1,205 @@
...
+    box-shadow: none;
+    background: ${(props) =>
+      normalizeColor(
+        props.theme.global.colors['background-contrast'],
+        props.theme,
+      )};
+    color: ${(props) =>
+      normalizeColor(props.theme.toggleButtonGroup.button.color, props.theme)};
+  }
+`;
```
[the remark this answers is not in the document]
britt6612: if we remove the extra &hover it will just be 
But i can move to another file just put it here since its on the shorter end.
```
const StyledButton = styled(Button)`
  ${(props) =>
    edgeStyle(
      'padding',
      props.theme.toggleButtonGroup.button.pad,
      false,
      undefined,
      props.theme,
    )}
  border-radius: ${(props) => props.theme.global.control.border.radius};
  border: none;
`;

```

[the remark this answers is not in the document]
britt6612: done

src/js/components/ToggleButtonGroup/StyledToggleButtonGroup.js
```diff
@@ -0,0 +1,17 @@
+import styled from 'styled-components';
+import { Button } from '../Button';
+import { roundStyle, edgeStyle } from '../../utils';
+
+export const StyledButton = styled(Button)`
+  border-radius: ${(props) => props.theme.global.control.border.radius};
```
taysea, line 6:
```suggestion
  border-radius: 0;
```
I believe this should be 0 to baseline the border-radius, then props.round should just apply to the first/last item

src/js/components/ToggleButtonGroup/ToggleButtonGroup.js
```diff
@@ -0,0 +1,187 @@
...
+            ? !!value.includes(optionValue)
+            : value === optionValue;
+          let round = 0;
+
+          // remove round corners of first and last buttons to match container
+          if (typeof theme.toggleButtonGroup.container.round === 'string') {
+            if (index === 0) {
+              round = {
+                corner: 'right',
+                size: 'none',
```
taysea, line 141:
```suggestion
                corner: 'left',
                size: theme.toggleButtonGroup.container.round,
```
This block is supposed to apply rounding based on that of the container

src/js/components/ToggleButtonGroup/ToggleButtonGroup.js
```diff
@@ -0,0 +1,187 @@
...
+              };
+            } else if (index === options.length - 1) {
+              round = {
+                corner: 'left',
```
taysea, line 145:
```suggestion
                corner: 'right',
                size: theme.toggleButtonGroup.container.round,
```
This block is supposed to apply rounding based on that of the container

src/js/components/ToggleButtonGroup/ToggleButtonGroup.js
taysea: We could actually go back to this logic from before to simplify:

```
round = {
               corner: index === 0 ? 'left' : 'right',
              size: theme.toggleButtonGroup.container.round,
}

File: src/js/components/ToggleButtonGroup/stories/Icons.js [hunk 23]
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { Box, ToggleButtonGroup } from 'grommet';
+import { List, MapLocation, Table } from 'grommet-icons';
+
+const optionsIcons = [
+  {
+    icon: <List />,
```
halocline, line 7: We should ensure our examples are demonstrating A11y best practices.

```suggestion
    icon: <List a11yTitle="List view" />,
```

src/js/components/ToggleButtonGroup/stories/Icons.js
```diff
@@ -0,0 +1,28 @@
...
+    value: 'list',
+  },
+  {
+    icon: <Table />,
```
halocline, line 11:
```suggestion
    icon: <Table a11yTitle="Table view" />,
```

src/js/components/ToggleButtonGroup/stories/Icons.js
```diff
@@ -0,0 +1,28 @@
...
+    value: 'Table',
+  },
+  {
+    icon: <MapLocation />,
```
halocline, line 15:
```suggestion
    icon: <MapLocation a11yTitle="Map view" />,
```

src/js/components/ToggleButtonGroup/stories/Icons.js
```diff
@@ -0,0 +1,28 @@
...
+    value: 'map',
+  },
+];
+
+export const UncontrolledIcons = () => (
+  <Box pad="large">
+    <ToggleButtonGroup options={optionsIcons} />
```
halocline, line 22:
```suggestion
    <ToggleButtonGroup a11yTitle="Choose view" options={optionsIcons} />
```

halocline, line 22: I would think that this example would behave as a radio button group because I believe there should always be one option selection.

src/js/components/ToggleButtonGroup/stories/multiple.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { Box, ToggleButtonGroup } from 'grommet';
+
+export const Multiple = () => (
+  <Box pad="large">
+    <ToggleButtonGroup options={['Red', 'Green', 'Blue']} multiple />
```
halocline, line 6: Let's make these examples more representative of real world use cases. This one seems like it lends itself to a text formatting use case e.g. "bold, italics, underline."

src/js/components/ToggleButtonGroup/stories/simple.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { Box, ToggleButtonGroup } from 'grommet';
+
+export const Simple = () => (
+  <Box gap="large" pad="large">
+    <ToggleButtonGroup options={['Option 1', 'Option 2', 'Option 3']} />
```
halocline, line 6: Let's talk about what stories we want to represent as a whole.

src/js/components/ToggleButtonGroup/ToggleButtonGroup.js
```diff
@@ -0,0 +1,182 @@
...
+      onUp={onPrevious}
+      onDown={onNext}
+      onLeft={onPrevious}
+      onRight={onNext}
+    >
+      <Box
+        ref={ref}
+        direction="row"
+        alignSelf="start"
+        role="group"
```
halocline, line 110: I think this role will need to be conditional based on whether it is in a radio vs. multiple mode.

Per spec `role="radio"` must be contained in a `role="radiogroup"` ([reference](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/radio_role#description))

When in `multiple`, `role="group"` is appropriate.

src/js/components/ToggleButtonGroup/ToggleButtonGroup.js
```diff
@@ -0,0 +1,182 @@
...
+                      side: 'right',
+                      color: theme.toggleButtonGroup.divider.color,
+                    }
+                  : undefined
+              }
+              key={optionValue || index}
+            >
+              <StyledButton
+                active={active}
+                aria-checked={active}
```
halocline, line 160: I believe we will need to choose `aria-checked` when in radio mode and `aria-pressed` when in multiple mode. Aria-pressed will announce the control as a "toggle button."

[reference](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-pressed)

src/js/components/ToggleButtonGroup/README.md
```diff
@@ -0,0 +1,2 @@
+## ToggleButtonGroup
```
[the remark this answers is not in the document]
britt6612: done

taysea:
@britt6612 For HPE cases, I think we need to support an `toggleGroup.button.iconOnly.pad` in the theme for icon only vs label buttons

src/js/components/ToggleGroup/__tests__/ToggleButtonGroup-test.tsx
```diff
@@ -0,0 +1,498 @@
...
+      'box-shadow',
+      '0 0 2px 2px #6FFFB0',
+      {
+        modifier: ':focus',
+      },
+    );
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  // test('should allow caller to enforce a value in controlled scenarios', () => {
```
taysea, line 147: This is commented out, do we want to enable it or remove it?

src/js/components/ToggleGroup/propTypes.js
```diff
@@ -0,0 +1,29 @@
...
+    multiple: PropTypes.bool,
+    onChange: PropTypes.func,
+    options: PropTypes.oneOfType([
+      PropTypes.arrayOf(PropTypes.string),
+      PropTypes.arrayOf(
+        PropTypes.shape({
+          id: PropTypes.string,
+          icon: PropTypes.oneOfType([PropTypes.element, PropTypes.node]),
+          label: PropTypes.oneOfType([PropTypes.string, PropTypes.element]),
+          value: PropTypes.oneOfType([PropTypes.string]),
```
taysea, line 19: Can this just be PropTypes.string?

src/js/components/ToggleGroup/propTypes.js
```diff
@@ -0,0 +1,29 @@
...
+      PropTypes.string,
+      PropTypes.arrayOf(PropTypes.string),
+    ]),
+    multiple: PropTypes.bool,
+    onChange: PropTypes.func,
+    options: PropTypes.oneOfType([
+      PropTypes.arrayOf(PropTypes.string),
+      PropTypes.arrayOf(
+        PropTypes.shape({
+          id: PropTypes.string,
```
taysea, line 16:
```suggestion
```
We can remove this?

src/js/components/ToggleGroup/stories/Icons.js
```diff
@@ -0,0 +1,28 @@
...
+import { List, MapLocation, Table } from 'grommet-icons';
+
+const optionsIcons = [
+  {
+    icon: <List a11yTitle="List view" />,
+    value: 'list',
+  },
+  {
+    icon: <Table a11yTitle="Map view" />,
+    value: 'Table',
```
taysea, line 12: just lowercasing to be consistent with other options
```suggestion
    value: 'table',
```

src/js/components/ToggleGroup/__tests__/ToggleButtonGroup-test.tsx
britt6612: This is another TODO once we decide on name for `onChange` if we keep it then I need to update TS then I will enable

src/js/components/ToggleGroup/propTypes.js
britt6612: yes

src/js/components/ToggleGroup/StyledToggleGroup.js
```diff
@@ -0,0 +1,19 @@
...
+import { Button } from '../Button';
+import { roundStyle, edgeStyle } from '../../utils';
+
+export const StyledButton = styled(Button)`
+  border-radius: 0;
+  ${(props) => roundStyle(props.round, true, props.theme)};
+  ${(props) =>
+    edgeStyle(
+      'padding',
+      props.icon !== undefined && props.theme.toggleGroup.iconOnly?.pad
```
taysea, line 11: I think we want it only when it's iconOnly but not icon + label. Also, I wonder if it'd be cleaner though to have the logic in ToggleGroup itself so this is purely styles focused.

So in ToggleGroup where we're mapping and creating the Buttons:
```
const iconOnly = icon && !label;
let pad = theme.toggleGroup.button.pad;
if (iconOnly && theme.toggleGroup.button?.iconOnly?.pad) pad = theme.toggleGroup.button.iconOnly.pad;

...

<Button
... 
pad={pad}
/>
```

src/js/components/ToggleGroup/ToggleGroup.js
```diff
@@ -0,0 +1,179 @@
...
+  const value = controlled ? prop : uncontrolledProp;
+  const handleChange = useCallback(onChange, [onChange]);
+
+  const setValue = useCallback(
+    (nextValue) => {
+      // only update internal value in uncontrolled cases
+      if (!controlled) {
+        setUncontrolledProp(nextValue);
+      }
+      handleChange(nextValue);
```
taysea, line 20: Personally, I'd still recommend we pass this as an object to give use more flexibility in future to pass other things, something like:

```
handleChange({ value: nextValue })
```

src/js/components/ToggleGroup/propTypes.js
britt6612: done

src/js/components/ToggleGroup/StyledToggleGroup.js
britt6612: done

src/js/components/ToggleGroup/ToggleGroup.js
britt6612: agree forgot we talked about this!

britt6612:
- Why named “ToggleGroup”
  - When viewing Google search trends it was found that `toggleGroup` was used more across the wider region.
  - Another downside to using ToggleButtonGroup is how long the word is and in many places in the code base we would need to disable eslint for the `imports` since the line would be over a amount of characters.
  - Did `button` actually have any meaning even though this component was made with a Button this did not necessarily need to be made known to the user.

- We enforce a value once one is selected.
  - 80% of use cases all have one selected option for a toggleGroup.
  - For our specific need for this component a use case was toggling through either a `map` `list` or `data` view where at least one of those needed to be selected.

src/js/components/ToggleButtonGroup/stories/Icons.js
taysea: Done.

src/js/components/ToggleButtonGroup/README.md
```diff
@@ -0,0 +1,2 @@
...
+Documentation for this component: https://v2.grommet.io/toggleButtonGroup
```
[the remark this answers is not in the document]
taysea: Done.

src/js/components/ToggleButtonGroup/stories/Icons.js
taysea: Done.

src/js/components/ToggleButtonGroup/stories/Icons.js
taysea: Done.

src/js/components/ToggleButtonGroup/ToggleButtonGroup.js
taysea: Done.

src/js/components/ToggleGroup/stories/Controlled.js
```diff
@@ -0,0 +1,30 @@
+import React, { useState } from 'react';
+import { Box, ToggleGroup, Text } from 'grommet';
+
+export const ControlledGroup = () => {
+  const [multipleControlled, setMultipleControlled] = useState(['c2']);
+
+  return (
+    <Box direction="row" gap="xlarge" overflow="auto">
+      <Box gap="large" pad="large">
+        <Text>An option always has to be checked</Text>
```
halocline, line 10: How is this different than the default/simple behavior? Shouldn't this story be updated to show only one or none option can be selected?

src/js/components/ToggleGroup/stories/Icons.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import { Box, ToggleGroup } from 'grommet';
+import { List, MapLocation, Table } from 'grommet-icons';
+
+const optionsIcons = [
```
halocline, line 5: I think this story should be the "simple." Then this file Icons story can go away.

src/js/components/ToggleGroup/stories/multiple.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { Box, ToggleGroup } from 'grommet';
+
+export const Multiple = () => (
+  <Box pad="large">
+    <ToggleGroup options={['Bold', 'Italic', 'Underline']} multiple />
```
halocline, line 6: These should be icons, not text.

Please check that the screenreader announces these properly.

src/js/components/ToggleGroup/stories/simple.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { Box, ToggleGroup } from 'grommet';
+
+export const Simple = () => (
+  <Box gap="large" pad="large">
+    <ToggleGroup options={['Option 1', 'Option 2', 'Option 3']} />
```
halocline, line 6: Replace this story with the "choose view" execution?

src/js/components/ToggleGroup/stories/Controlled.js
taysea: I believe in the `multiple` case we are allowing removal of all selections.

However, for clarity, I'd be supportive of changing this to be single select + ability to remove all selections.

taysea: After discussion, we'll leave as is and just adjust the intro text for clarity.

src/js/components/ToggleGroup/stories/Icons.js
taysea: Done.

src/js/components/ToggleGroup/stories/multiple.tsx
taysea: Done.

src/js/components/ToggleGroup/stories/simple.js
taysea: Done.

taysea (approved):

src/js/components/ToggleGroup/ToggleGroup.js
```diff
@@ -0,0 +1,190 @@
...
+          }
+
+          const iconOnly = icon && !label;
+          let pad;
+          if (iconOnly && theme.toggleGroup.button?.iconOnly?.pad) {
+            pad = theme.toggleGroup.button.iconOnly.pad;
+          } else pad = theme.toggleGroup.button.pad;
+
+          return (
+            <Box
```
halocline, line 153: Do we need to think about some sort of flex={false}?

Screenshot 2024-04-17 at 3 35 34 PM

src/js/themes/base.js
```diff
@@ -1935,6 +1935,28 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         margin: 'none',
       },
     },
+    toggleGroup: {
+      button: {
+        iconOnly: {
+          pad: {
+            horizontal: 'small',
+            vertical: 'small',
```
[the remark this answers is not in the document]
taysea: Agree they should. I'll update

src/js/components/ToggleGroup/stories/simple.js
```diff
@@ -0,0 +1,32 @@
...
+  },
+  {
+    icon: <MapLocation a11yTitle="Map view" />,
+    value: 'map',
+  },
+];
+
+export const Simple = () => (
+  <Box gap="large" pad="large">
+    <ToggleGroup
```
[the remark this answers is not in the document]
taysea: This is because we are just inheriting the button theme as is and that's how Grommet theme styles buttons. Are we okay with that? We'd discussed not customizing the button styling beyond padding/rounding.

taysea:
Dismissing my review, I noticed the button pad + border of container is causing the height to be 2px higher than usual button. Should we be subtracting border width from pad? or should caller be responsible for the calculation?

src/js/components/ToggleGroup/stories/simple.js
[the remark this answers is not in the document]
halocline: The icon only versions are not receiving the purple hover state. Even though we are inconsistent in Button, similar to height, should we be consistent within ToggleGroup?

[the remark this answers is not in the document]
taysea: @halocline Do you want us to enforce either the border or no border hover styling across the board for this component?

taysea:
> Dismissing my review, I noticed the button pad + border of container is causing the height to be 2px higher than usual button. Should we be subtracting border width from pad? or should caller be responsible for the calculation?

After offline discussion, we will align with how Button works --> For "base" theme, the caller is responsible for managing the padding + border calculations. For "kind" themes (like HPE), the button handles the subtraction of border-width from padding. So, for "base" theme, ToggleGroup will base leave responsible for managing.

src/js/components/ToggleGroup/stories/simple.js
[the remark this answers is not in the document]
halocline: I was thinking the purple border seemed to be more in alignment with the other button hover states. What's your thinking?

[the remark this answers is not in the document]
taysea: After discussion with Matt offline, we determined we should not have the purple border (see rationale in PR description under style overrides).

src/js/components/ToggleGroup/ToggleGroup.js
taysea: Added `flex={false}`.

src/js/components/ToggleGroup/StyledToggleGroup.js
```diff
@@ -1,10 +1,43 @@
...
+        : props.theme?.toggleGroup.button?.pad;
+
+    // adjust pad for "kind" themes to align with how "kind" themes
+    // manages this calculation
+    const pad = props.theme?.button?.default
+      ? adjustPad(themePad, props.theme)
+      : themePad;
+    return edgeStyle('padding', pad, false, undefined, props.theme);
+  }}
+  // remove hover style from StyledButton/StyledButtonKind theme
```
britt6612, line 38: 👍

src/js/themes/base.js
```diff
@@ -1935,6 +1935,24 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         margin: 'none',
       },
     },
+    toggleGroup: {
+      button: {
+        // iconOnly: {
+        //   pad: undefined,
+        // },
+        pad: {
+          horizontal: '11px',
```
taysea, line 1944: @halocline @britt6612 Final item to decide on for base theme, what should the default horizontal width be?

Here, it's the same for with label or icon only and looks like this:
Screen Shot 2024-04-19 at 11 32 48 AM

Alternately, if we aligned to the larger horizontal pad for default buttons it would have a value of `23px` and look like this (where iconOnly ones would still retain their 11px):

Screen Shot 2024-04-19 at 11 34 57 AM

I was questioning the larger horizontal pad because that is geared more towards buttons with full rounding. Let me know which value you both are good with.

britt6612: I agree 
`that the larger horizontal pad because that is geared more towards buttons with full rounding`

I'd like the smaller option.

halocline: Probably the smaller

taysea: Great, then ready for final review.

halocline (approved):

PR #7173, opened 2024-03-23
DO NOT MERGE Temporary DataTable example of multi select across pagination

MikeKingdom:
This is a DataTable example that uses pagination of data that is only partial (ie as if it only grabs a page at a time of data from the backend) and allows selection across pages and the items are selected by default.

To do this in a scalable way you instead need to track what items get deselected. This also means that at the time you want to submit this result to a backend API, the API needs to support being told what items you **don't** want selected.

To avoid this you would need a reasonable api to get the entire list of primary keys for the selected items which isn't scalable.

**Note** the select/deselect all checkbox in the header of Grommet DataTable component won't be able to support this across pagination without the entire data set. So if you go this route you'd likely have to disable that checkbox via `allowSelectAll={false}`

### DIFF
--- a/src/js/components/DataTable/stories/Multisel.js
+++ b/src/js/components/DataTable/stories/Multisel.js
@@ -1,0 +1,77 @@
+import React, { useMemo, useState } from 'react';
+import { Box, Data, DataTable, Pagination } from 'grommet';
+
+const DATA = [];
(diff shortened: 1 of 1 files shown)


jcfilben:
Closing since this was only for demo purposes. @MikeKingdom feel free to reopen if this is still needed

PR #6987, opened 2023-10-22
Fix #6981

aditya000099:
changed const autoFocusingTags = /(a|area|input|select|textarea|button|iframe)$/; to  const autoFocusingTags = /(a|area|input|select|textarea|button|iframe)(?![\w-])$/;

hacktoberfest

Issue #6981, opened 2023-10-20
autoFocusingTags regex unintentionally matches `<data` and `<meta`


lukaw3d:
https://github.com/grommet/grommet/blob/7ad63c7326a531cf8b4997143346c204291420a1/src/js/utils/DOM.js#L160

matches `data` and `meta` elements

aditya000099:
can you assign this to me?

lukaw3d:
Update: this would match various web-components too

jcfilben:
Resolved by #7174

PR #7174, opened 2024-03-26
Fix autoFocusingTags regex

lukaw3d:
Prevents unintentionally matching `data` and `meta` elements, and possibly custom web-components.

- Issue https://github.com/grommet/grommet/issues/6981
- Comments on previous incorrect fix https://github.com/grommet/grommet/pull/6987

Not certain, but I doubt it affects any real-world usecase

### DIFF
--- a/src/js/utils/DOM.js
+++ b/src/js/utils/DOM.js
@@ -157,7 +157,8 @@ export const makeNodeFocusable = (node) => {
   }
 };
 
-const autoFocusingTags = /(a|area|input|select|textarea|button|iframe)$/;
+// Using ^ and $ to match the whole tagName, and not e.g. <meta> and <data>.
+const autoFocusingTags = /^(a|area|input|select|textarea|button|iframe)$/;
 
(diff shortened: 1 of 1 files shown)


src/js/utils/DOM.js
```diff
@@ -157,7 +157,7 @@ export const makeNodeFocusable = (node) => {
   }
 };
 
-const autoFocusingTags = /(a|area|input|select|textarea|button|iframe)$/;
+const autoFocusingTags = /^(a|area|input|select|textarea|button|iframe)$/;
```
britt6612, line 160: can we add a comment? just for future reference 
`maybe something like to avoid unintentionally matching <data and <meta tags we need the ^`

lukaw3d: Added, though it's just essential regex knowledge

britt6612 (approved):

MikeKingdom (approved):
lgtm. Thanks!

taysea (approved):

Issue #7170, opened 2024-03-21
Enhance DataTableColumns to disallow changing the order of some columns


kimkscott:
Requesting a DataTableColumns enhancement to disallow the user from changing the order of a subset of the columns.  Today I don't think DataTableColumns supports this.

We would like to disallow changing the order of some of the columns.  We would like to pin some of the leftmost columns during horizontal scrolling.  But I am not sure this is feasible if the user can reorder the columns.

Currently I think all columns can be reordered.  I don't think there's a way to prevent the user from changing the order of one or more columns.

N/A

### Your Environment

- Grommet version: 2.33.1
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): Windows 10 22H2

PR #7178, opened 2024-03-28
Enhance DataTableColumns ordering options

kimkscott:
This is a partial, draft, PR to enhance DataTableColumns to disallow changing the order of selected columns.

This PR addresses this issue: https://github.com/grommet/grommet/issues/7170

A new attribute, pinOrder, can be specified in an object in the options property.  This attribute pins the order of this column in DataTableColumns.

This is implemented using the existing pinned property of the List component.

Here is a description of requirements provided by Taylor Seamans:
1. Whichever item is effectively the “first” reorderable item should have the up arrow button disabled
2. If an item that isn’t the leftmost is not reorderable, then the items around it should shift while keeping its order fixed
3. This behavior should work for both drag + drop as well as using the arrow keys

image

### DIFF
--- a/src/js/components/DataTableColumns/index.d.ts
+++ b/src/js/components/DataTableColumns/index.d.ts
@@ -2,7 +2,15 @@ import * as React from 'react';
 
 export interface DataTableColumnsProps {
   drop: boolean;
-  options: (string | { property: string; label: string; disabled?: boolean })[];
+  options: (
+    | string
+    | {
+        property: string;
+        label: string;
+        disabled?: boolean;
+        pinOrder?: boolean;
+      }
+  )[];
 }
 
 declare const DataTableColumns: React.FC<DataTableColumnsProps>;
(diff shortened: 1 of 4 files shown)


src/js/components/DataTableColumns/index.d.ts
```diff
@@ -2,7 +2,15 @@ import * as React from 'react';
...
 export interface DataTableColumnsProps {
   drop: boolean;
-  options: (string | { property: string; label: string; disabled?: boolean })[];
+  options: (
+    | string
+    | {
+        property: string;
+        label: string;
+        disabled?: boolean;
+        pinOrder?: boolean;
```
britt6612, line 11:
```suggestion
        pinned?: boolean;
```

src/js/components/DataTableColumns/propTypes.js
```diff
@@ -10,6 +10,7 @@ if (process.env.NODE_ENV !== 'production') {
         PropTypes.shape({
           disabled: PropTypes.bool,
           label: PropTypes.string,
+          pinOrder: PropTypes.bool,
```
britt6612, line 13:
```suggestion
          pinned: PropTypes.bool,
```

britt6612:
Looking at enhancing/changing current style 

Screen Shot 2024-04-08 at 8 28 26 PM

taysea:
Closing this PR since the work will be continued by https://github.com/grommet/grommet/pull/7193

PR #7186, opened 2024-04-04
update step prop

britt6612:
#### What does this PR do?
Fixes a bug where step was not being updated based on `stepProp` 
pagination.js
added a jest test 
```
  const [itemsPerPage, setItemsPerPage] = useState(20);
  const [totalItems, setTotalItems] = useState(itemsPerPage);

  useEffect(() => {
    setTotalItems(250);
  }, []);

  return (
    <Grommet theme={theme}>
      <Select
        testId="rows-per-page-dropdown"
        defaultVal={20}
        value={itemsPerPage}
        options={[10, 20, 50, 100]}
        onChange={({ option }) => {
          console.log(option);
          setItemsPerPage(option);
        }}
      />
      <Pagination numberItems={totalItems} step={itemsPerPage} />
    </Grommet>
  );
```

We added a step state then set that value within the PaginationStep so since I changed step from being directly assigned to stepProp to using useState it was only getting the initial value of stepProp during the initial render and was not updating
team need
no
yes
backwards compatible

### DIFF
--- a/src/js/components/Pagination/Pagination.js
+++ b/src/js/components/Pagination/Pagination.js
@@ -48,6 +48,11 @@ const Pagination = forwardRef(
     const total = numberItems ?? filteredTotal ?? 0;
     const page = pageProp || view?.page || 1;
 
+    // watch for changes to stepProp
+    useEffect(() => {
+      if (stepProp) setStep(stepProp);
+    }, [stepProp]);
+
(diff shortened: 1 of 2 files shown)


britt6612:
closing in favor of https://github.com/grommet/grommet/pull/7188

PR #7187, opened 2024-04-04
Add check for responsive in `roundStyle`

### LINKED ISSUE #6445: Box "round" prop not respecting whether its responsive
It is not currently possible to disable the responsive adjustments for border-radius on `Box` with `responsive={false}`. 

### Expected Behavior
The size for `round` should reference the responsive value unless `responsive={false}` is set.

### Actual Behavior
The size for the `round` prop always references the responsive value when given a non-object type value. 

https://codesandbox.io/s/grommet-v2-template-forked-d914qf

### Steps to Reproduce

1. create a box 
2. set responsive to false
3. give it round prop with a string value
4. resize the screen (should see radius not change, but it does).

### Your Environment

Codesandbox env has the bug.

- Grommet version:
- Browser Name and version:

britt6612:
#### What does this PR do?
fixes Box `round` prop not respecting `responsive`
styles.js
locally
```
    <Box pad="medium" gap="small">
      <Box pad="large" border="all" round="large">
        <Heading>Responsive Box</Heading>
      </Box>

      <Box responsive={false} pad="large" border="all" round="large">
        <Heading>Non-Responsive Box</Heading>
        <Paragraph>round with string prop ❌</Paragraph>
      </Box>

      <Box
        responsive={false}
        pad="large"
        border="all"
        round={{ corner: "top", size: "large" }}
      >
        <Heading>Non-Responsive Box</Heading>
        <Paragraph>round with object prop ✅</Paragraph>
      </Box>
    </Box>
```

see issue https://github.com/grommet/grommet/issues/6445

For Skeleton there was `  ${(props) =>
    props.round && roundStyle(props.round, props.responsive, props.theme)}` 
however there is no responsive prop on Skeleton so just changed to `true`
closes #6445 
no
yes
backwards compatible


src/js/components/Skeleton/StyledSkeleton.js
```diff
@@ -36,7 +36,6 @@ export const StyledSkeleton = styled.div`
       props.theme.box.responsiveBreakpoint,
       props.theme,
     )}
-  ${(props) =>
-    props.round && roundStyle(props.round, props.responsive, props.theme)}
+  ${(props) => props.round && roundStyle(props.round, true, props.theme)}
```
taysea, line 39: Why did we change this prop `props.responsive` to "true"?

britt6612: because there is no `props.responsive` well I guess there is `...rest` on Box but this is default to `false`

britt6612: changing back!

src/js/components/Skeleton/__tests__/__snapshots__/Skeleton-test.tsx.snap
```diff
@@ -706,12 +706,6 @@ exports[`Skeleton with styling props 1`] = `
   -webkit-font-smoothing: antialiased;
 }
 
-@media only screen and (max-width:768px) {
```
taysea, line 709: Why are we losing this here? I couldn't see a responsive false anywhere that would cause this to be removed? Or did I miss something

britt6612: I see the issue so.. 
```
  ${(props) => props.round && roundStyle(props.round, true, props.theme)}

```

We do need to set true because if we just have props. responsive that's undefined. hence the snap shot change because I added ` responsive && breakpoint && breakpoint.edgeSize[size];` however it was no longer going in here because responsive is not true for skeleton its undefined. StyledSkeleton is a `styledDiv` so that `responsive` prop is not pulled out from Box and is being used. The only place we have 
`  ${(props) =>
    props.round && roundStyle(props.round, props.responsive, props.theme)}` is in Box which makes sense otherwise we have `true` in other cases such as `StyledDrop` `styledTable` `styledTag` since `responsive` is not a prop on those components.

taysea: Gotcha, thanks for clarifying. Setting `true` makes sense then in this case for backwards compatibility. Thanks for digging.

taysea (approved):
Thank you!

src/js/components/Box/__tests__/Box-style-test.tsx
```diff
@@ -198,6 +198,15 @@ describe('Box', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
+  test('round with responive equal to false', () => {
```
jcfilben, line 201: Does the snapshot effectively test the `responsive={false}` behavior or do we need to adjust the screen size in the jest test before taking the snapshot?

jcfilben:
In the PR description it says the docs do not need to be updated for this change but I think we would need to update the `responsive` prop documentation

src/js/components/Box/__tests__/Box-style-test.tsx
britt6612: we can make the screen smaller and then take a snap shot however if we did not have `responsive={false}` then snapshot would automatically add the 

```
@media only screen and (max-width:768px) {
  .c1 {
    border-radius: 24px;
  }
}
```

britt6612:
> In the PR description it says the docs do not need to be updated for this change but I think we would need to update the `responsive` prop documentation

I can update!

src/js/components/Box/__tests__/Box-style-test.tsx
jcfilben: Ah gotcha, I think the test is good as is then

jcfilben (approved):
Looks good!

PR #7188, opened 2024-04-04
Fixed Pagination step to allow it to be controlled

### LINKED ISSUE #7201: Passing dynamic value to `step` prop in pagination no longer updates
Using Pagination component from grommet & passing the dynamic value(dropdown selection value) to the `step` prop. I see that when I change the `step` value using dropdown & pass it to Pagination component. The pagination page count is not getting changed.

Grommet added a step state than set that value within the PaginationStep so since I changed `step` from being directly assigned to stepProp to using useState it was only getting the initial value of stepProp during the initial render and was not updating
### Expected Behavior
The pagination page count should change since the `step` value is changing

### Actual Behavior
The pagination `step` prop is not being updated if we pass a dynamic value to Pagination `step`

https://codesandbox.io/p/sandbox/grommet-box-forked-ns453z?file=%2Fpackage.json%3A11%2C28

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.36.1

britt6612:
#### What does this PR do?
Fixes a bug where step was not being updated based on `stepProp` 
pagination.js
added a jest test 
```
  const [itemsPerPage, setItemsPerPage] = useState(20);
  const [totalItems, setTotalItems] = useState(itemsPerPage);

  useEffect(() => {
    setTotalItems(250);
  }, []);

  return (
    <Grommet theme={theme}>
      <Select
        testId="rows-per-page-dropdown"
        defaultVal={20}
        value={itemsPerPage}
        options={[10, 20, 50, 100]}
        onChange={({ option }) => {
          console.log(option);
          setItemsPerPage(option);
        }}
      />
      <Pagination numberItems={totalItems} step={itemsPerPage} />
    </Grommet>
  );
```

We added a step state then set that value within the PaginationStep so since I changed step from being directly assigned to stepProp to using useState it was only getting the initial value of stepProp during the initial render and was not updating.

This is an alternative approach to https://github.com/grommet/grommet/pull/7186
This way we are leveraging the `useControlled` func which will cause one less re-render.
closes #7201 
no
yes
backwards compatible

### DIFF
--- a/src/js/components/Pagination/Pagination.js
+++ b/src/js/components/Pagination/Pagination.js
@@ -8,6 +8,7 @@ import { PageControl } from './PageControl';
 import { PaginationStep } from './PaginationStep';
 import { PaginationSummary } from './PaginationSummary';
 import { PaginationPropTypes } from './propTypes';
+import { useControlled } from '../../utils/useControlled';
 
 const StyledPaginationContainer = styled(Box)`
   ${(props) =>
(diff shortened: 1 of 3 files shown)


halocline (approved):
Thank you.

taysea (approved):

PR #7193, opened 2024-04-09
Enhance DataTableColumns ordering options

ganondev:
Clones #7178 

This is a partial, draft, PR to enhance DataTableColumns to disallow changing the order of selected columns.

Closes https://github.com/grommet/grommet/issues/7170

A new attribute, pinOrder, can be specified in an object in the options property. This attribute pins the order of this column in DataTableColumns.

This is implemented using the existing pinned property of the List component.

Here is a description of requirements provided by Taylor Seamans:

1. Whichever item is effectively the “first” reorderable item should have the up arrow button disabled
2. If an item that isn’t the leftmost is not reorderable, then the items around it should shift while keeping its order fixed
3. This behavior should work for both drag + drop as well as using the arrow keys

image

### DIFF
--- a/src/js/components/DataTableColumns/__tests__/DataTableColumns-test.tsx
+++ b/src/js/components/DataTableColumns/__tests__/DataTableColumns-test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { act, fireEvent, render } from '@testing-library/react';
+import { act, fireEvent, render, screen, within } from '@testing-library/react';
 import 'jest-styled-components';
 import '@testing-library/jest-dom';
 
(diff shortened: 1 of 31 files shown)


ganondev:
Screenshot after renaming `pinOrder` to `pinned`
image

britt6612:
@ganondev 
Thinking of a good test for this maybe we can add
scenario: 'pinned' second column
```
1. pressing `up arrow` on third column should result in third column now as first and first column now as third.
2. dragging third column up should result in third column now as first and first column now as third.
3. up/down arrows should not exist for `pinned` column

```

Similar to this test lmk if you need help!
https://github.com/grommet/grommet/blob/master/src/js/components/List/__tests__/List-test.js#L921

Also we decided on a styling but I will do those changes in a separate PR so we don't hold this up.

src/js/components/DataTableColumns/stories/Simple.js
```diff
@@ -16,6 +16,9 @@ const options = columns.map(({ header, property }) => ({
   label: property === 'name' ? 'Name' : header,
 }));
 
+options[0].pinned = true;
```
taysea, line 19: Instead of this code pattern, can we incorporate the `pinned` logic in the object return of options? Something like:
```
const options = columns.map(({ header, property }, index) => ({
  property,
  label: property === 'name' ? 'Name' : header,
  // demoing pinned behavior
  pinned: index === 0 || index === 1 ? true : undefined,
}));

Also, adjusting to the first 2 columns to be a bit more "realistic" to common use cases

Comment (britt6612) [reply src/js/components/DataTableColumns/stories/Simple.js:19]: to add to this can we revert `Simple` story back and have a new story named Pinned

Comment (edwardhpe) [reply src/js/components/DataTableColumns/stories/Simple.js:19]: done

Comment (taysea) [reply src/js/components/DataTableColumns/stories/Simple.js:19]: @edwardhpe It looks like there are still changes in the "Simple" story with the addition of `.pinned` -- can you remove those lines so Simple story is unchanged?

File: src/js/components/DataTableColumns/stories/Pinned.js [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+  DataTableColumns,
+  Toolbar,
+} from 'grommet';
+import { columns, DATA } from '../../DataTable/stories/data';
+
+// simplify option label for name property
+const options = columns.map(({ header, property }, i) => ({
+  property,
+  label: property === 'name' ? 'Name' : header,
+  pinned: i === 0 || i === 2,
```
taysea, line 17:
```suggestion
  pinned: i === 0 || i === 1,
```
Tiny suggestion to be a bit more aligned with common use cases.

src/js/components/DataTableColumns/stories/Simple.js
taysea, line 19:
```suggestion
```

src/js/components/DataTableColumns/stories/Simple.js
```diff
@@ -16,6 +16,9 @@ const options = columns.map(({ header, property }) => ({
...
+options[2].pinned = true;
```
taysea, line 20:
```suggestion
```

src/js/components/DataTableColumns/stories/Simple.js
```diff
@@ -16,6 +16,9 @@ const options = columns.map(({ header, property }) => ({
...
+
```
taysea, line 21:
```suggestion
```

britt6612 (approved):
LGTM

src/js/components/DataTableColumns/__tests__/DataTableColumns-test.tsx
```diff
@@ -123,4 +123,84 @@ describe('DataTableColumns', () => {
       expect.objectContaining({ columns: ['size'] }),
     );
   });
+
+  test('pinned', async () => {
```
taysea, line 127:
```suggestion
  test('pinned', () => {
```
Do we need async? It doesn't seem like anything is using `await` syntax. I tried locally without async and all passed.

taysea (approved):
LGTM, thank you!

halocline (approved):
Thank you @ganondev for your contribution!

Issue #7199, opened 2024-04-11
Data table selections should persist across DataViews


halocline:
### Expected Behavior

When using DataTable's `onSelect` with Data, my selections across views should be persist and accumulate as I change my view and make additional selections. For example, I should be able to:

**Scenario 1**
- Alter my data view - for example's sake, let's say the view returned 7 unselected records.
- Use the select all control in the table's header.
- Total selected should be 7.
- Adjust my view again, this time returning 5 unselected records.
- Select all in the current view.
- Total selected should be 12.

**Scenario 2**
- Alter my data view - for example's sake, let's say the view returned 7 unselected records.
- Use the select all control in the table's header.
- Total selected should be 7.
- Adjust my view again, this time returning 5 records (3 selected + 2 unselected).
- Select all in the current view.
- Total selected should be 9.

**Scenario 3**
- Alter my data view - for example's sake, let's say the view returned 7 unselected records.
- Use the select all control in the table's header.
- Total selected should be 7.
- Adjust my view again, this time returning 5 records (3 selected + 2 unselected).
- Select all in the current view.
- Total selected should be 9.
- Unselect all in the current view.
- Total selected should be 4.

### Actual Behavior

Currently, my previously selected records are not retained after changing view and then selecting all. Instead, the previously selected records are replaced with my new selections.

### URL, screen shot, or Codepen exhibiting the issue

https://github.com/grommet/grommet/assets/1756948/607a1c47-66ae-4636-9b8d-8697c483adc7

### Steps to Reproduce

1. Modify the Data/SpaceEx story by adding `select` and `onSelect`.
2. Select all
3. Change view by selecting another page or applying a new search or filter
4. Select all again

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #7200, opened 2024-04-11
Fixed select all behavior across Data views

halocline:
#### What does this PR do?

Closes https://github.com/grommet/grommet/issues/7199 by persisting selected records across various Data views.

DataTable/Header.js

Using the story Data/OnSelect

Select all & unselect all across a variety of views, altering current page, filters, search, etc.

Using the story Data/OnSelect

https://github.com/grommet/grommet/issues/7199

No.

Yes.

Backwards compatible.


src/js/components/DataTable/DataTable.js
```diff
@@ -222,7 +222,7 @@ const DataTable = ({
       }
       const nextPinnedOffset = {};
 
-      if (columnWidths !== []) {
+      if (columnWidths.length !== 0) {
```
taysea, line 225: Tiny question, is this same as `if (columnWidths?.length)`

src/js/components/DataTable/DataTable.js
```diff
@@ -422,6 +422,13 @@ const DataTable = ({
     );
   }
 
+  const handleSelect = (nextSelected, row) => {
+    setSelected(nextSelected);
+    if (setSelectedDataContext) setSelectedDataContext(nextSelected.length);
+    if (row) onSelect(nextSelected, row);
```
taysea, line 428: Do we need this added condition or can we stick with old logic of `else onSelect(nextSelected, row)`?

It seems like if `row` is undefined it would be the same either way

src/js/components/DataTable/Header.js
```diff
@@ -195,35 +195,28 @@ const Header = forwardRef(
...
-        nextSelected = disabled
-          ? primaryValues.filter(
-              (v) => disabled.includes(v) && selected.includes(v),
-            )
-          : [];
-        nextGroupSelected[''] = 'none';
+      // if all enabled are already selected, remove them from selected;
+      // otherwise add them.
+      if (
+        selected.filter((s) => enabledSelected.includes(s)).length ===
```
taysea, line 201: Just trying to sort out the logic here. Does `enabledSelected` already contain the filtered array of selected values? If so can we simply do `if (enabledSelect.length === enabled.length)`?

src/js/components/DataTable/DataTable.js
halocline: Given
```
const columnWidths = [];
```

```
if (columnWidths?.length)
```
would evaluate to `0` which would be falsey and not `false`, right? Thinking I'd prefer the explicit. Thoughts?

src/js/components/DataTable/Header.js
```diff
@@ -195,35 +195,28 @@ const Header = forwardRef(
...
-        nextGroupSelected[''] = 'all';
-        groupBy?.expandable?.forEach((key) => {
-          nextGroupSelected[key] = 'all';
+        enabled.forEach((p) => {
+          if (!nextSelected.includes(p)) {
+            nextSelected.push(p);
+          }
         });
       }
-      if (groupBy?.onSelect) {
```
taysea, line 223: Is it backwards compatible to remove this condition? Feels like we might be losing support for something. Doesn't look like we currently have a unit test for `groupBy.onSelect` so would be worth adding to confirm we're staying backwards compatible here.

src/js/components/DataTable/Header.js
```diff
@@ -261,17 +254,22 @@ const Header = forwardRef(
                       ? 'unselect all'
                       : 'select all'
                   }
+                  // should the logic for these states live here or
```
taysea, line 257: I'm supportive of pulling them above so the return is cleaner, but also fine leaving as is if you don't want to add more changes to the PR.

Once you decide, we can remove this comment.

src/js/components/DataTable/Header.js
```diff
@@ -261,17 +254,22 @@ const Header = forwardRef(
...
-                        totalSelected === data.length
+                        // totalSelected === data.length
+                        totalSelected === (contextTotal || data.length)
                   }
                   indeterminate={
                     groupBy?.select
                       ? groupBy.select[''] === 'some'
-                      : totalSelected > 0 && totalSelected < data.length
+                      : totalSelected > 0 &&
+                        // totalSelected < data.length
```
taysea, line 271:
```suggestion
```

src/js/components/DataTable/Header.js
```diff
@@ -261,17 +254,22 @@ const Header = forwardRef(
...
                   }
+                  // should the logic for these states live here or
+                  // in a state variable?
                   checked={
                     groupBy?.select
                       ? groupBy.select[''] === 'all'
                       : totalSelected > 0 &&
                         data.length > 0 &&
-                        totalSelected === data.length
+                        // totalSelected === data.length
```
taysea, line 264:
```suggestion
```

src/js/components/DataTable/Header.js
halocline: Agree that a unit test would be beneficial here.

Is this story reasonable to confirm desired behavior? https://5d9774839a6eff00203f5cbf-wkumjyibuc.chromatic.com/?path=/story/visualizations-datatable-grouped-and-onselect--grouped-on-select-data-table

src/js/components/DataTable/DataTable.js
halocline: Fails a Jest test titled "DataTable › onSelect select/unselect all."

Can look at whether the test assertion is correct or not.

halocline: Confirmed we do need this condition and did exist previously in the body of the table.

src/js/components/DataTable/Header.js
halocline: done.

src/js/components/DataTable/Header.js
halocline: done

src/js/components/DataTable/Header.js
halocline: done

src/js/components/DataTable/Header.js
halocline: done

src/js/components/DataTable/DataTable.js
taysea: Good with the more explicit.

src/js/components/DataTable/Header.js
taysea: This story doesn't quite look like it's using `groupBy.onSelect`. It just seems to be using `groupBy` and separately `onSelect` prop but docs say you can have an `onSelect` handler in your groupBy object.

I'll add a unit test and we can check if it's breaking on this branch.

britt6612:
One thing I was looking at what if we have 1 `selected` but `disabled` 

is this the correct behavior? As you can see the `Amos-17` is `selected` and `disabled` Is this still the correct behavior for the intermediate or would we expect this to be checked?

Screen Shot 2024-04-12 at 2 18 34 PM

halocline:
> One thing I was looking at what if we have 1 `selected` but `disabled`
> 
> is this the correct behavior? As you can see the `Amos-17` is `selected` and `disabled` Is this still the correct behavior for the intermediate or would we expect this to be checked?
> 
> Screen Shot 2024-04-12 at 2 18 34 PM
> The count seems off when nothing is selected besides the one that is `disabled` and `selected` it shows one however when I select everything is shows one less.
> 
> Screen Shot 2024-04-12 at 2 20 15 PM

No. You are correct, we would expect the selected count to equal the total records.

halocline:
> > One thing I was looking at what if we have 1 `selected` but `disabled`
> > is this the correct behavior? As you can see the `Amos-17` is `selected` and `disabled` Is this still the correct behavior for the intermediate or would we expect this to be checked?
> > 
> > The count seems off when nothing is selected besides the one that is `disabled` and `selected` it shows one however when I select everything is shows one less.
> 
> No. You are correct, we would expect the selected count to equal the total records.

@britt6612 - Actually, I believe it is behaving correctly. The story had multiple items disabled which is why when 'Amos-17' was selected and disabled, 'CRS-21' was still just disabled. ('CRS-22' doesn't actually exist in the data set).
```
const disabled = ['Amos-17', 'CRS-21', 'CRS-22'];
```


I'm going to remove the 'disabled' from the story after you complete your review. It's helpful for QA but not for the story itself.

britt6612:
> > > One thing I was looking at what if we have 1 `selected` but `disabled`
> > > is this the correct behavior? As you can see the `Amos-17` is `selected` and `disabled` Is this still the correct behavior for the intermediate or would we expect this to be checked?
> > > The count seems off when nothing is selected besides the one that is `disabled` and `selected` it shows one however when I select everything is shows one less.
> > 
> > 
> > No. You are correct, we would expect the selected count to equal the total records.
> 
> @britt6612 - Actually, I believe it is behaving correctly. The story had multiple items disabled which is why when 'Amos-17' was selected and disabled, 'CRS-21' was still just disabled. ('CRS-22' doesn't actually exist in the data set).
> 
> ```
> const disabled = ['Amos-17', 'CRS-21', 'CRS-22'];
> ```
> 
> I'm going to remove the 'disabled' from the story after you complete your review. It's helpful for QA but not for the story itself. since the `'CRS-21'` was disabled that was missing from count

your correct

src/js/components/DataTable/Header.js
halocline: addressed

taysea (approved):
LGTM!

PR #7204, opened 2024-04-16
Add DataTableColumns `pinned` styling

britt6612:
#### What does this PR do?
Adds a `DataTableColumnContext` this is used so within LIst we know when List is being used in a DataTableColumn to have the Lock styling 
List.js DataTableColumns.js
locally storybook
storybook
TODO: add test

Need DataTableColumns pinned style
Ability to theme:
pinned icon (for HPE, Lock)
ability to style text/icon color (for HPE, text-weak)
yes
yes
 backwards compatible

### DIFF
--- a/src/js/components/DataTableColumns/DataTableColumns.js
+++ b/src/js/components/DataTableColumns/DataTableColumns.js
@@ -6,7 +6,7 @@ import { Box } from '../Box';
 import { CheckBoxGroup } from '../CheckBoxGroup';
 import { DataForm, formColumnsKey } from '../Data';
 import { DataFormContext } from '../../contexts/DataFormContext';
-import { DataTableColumnContext } from '../../contexts/DataTableColumnContext';
+import { DataTableColumnsContext } from '../../contexts/DataTableColumnsContext';
 import { FormContext } from '../Form/FormContext';
 import { DropButton } from '../DropButton';
 import { List } from '../List';
@@ -165,7 +165,7 @@ export const DataTableColumns = ({ drop, options, ...rest }) => {
   const { format } = useContext(MessageContext);
   const theme = useContext(ThemeContext);
   const [showContent, setShowContent] = useState();
-  const contextValue = useMemo(() => ({ inDataColumn: true }), []);
+  const contextValue = useMemo(() => ({ inDataTableColumns: true }), []);
 
   const tip = format({
     id: 'dataTableColumns.tip',
@@ -173,9 +173,9 @@ export const DataTableColumns = ({ drop, options, ...rest }) => {
   });
 
   let content = (
-    <DataTableColumnContext.Provider value={contextValue}>
+    <DataTableColumnsContext.Provider value={contextValue}>
       <Content drop={drop} options={options} />
-    </DataTableColumnContext.Provider>
+    </DataTableColumnsContext.Provider>
   );
   if (!inDataForm)
     content = (
--- a/src/js/components/DataTableColumns/DataTableColumns.js
+++ b/src/js/components/DataTableColumns/DataTableColumns.js
@@ -6,6 +6,7 @@ import { Box } from '../Box';
 import { CheckBoxGroup } from '../CheckBoxGroup';
 import { DataForm, formColumnsKey } from '../Data';
 import { DataFormContext } from '../../contexts/DataFormContext';
+import { DataTableColumnContext } from '../../contexts/DataTableColumnContext';
 import { FormContext } from '../Form/FormContext';
 import { DropButton } from '../DropButton';
 import { List } from '../List';
@@ -164,13 +165,18 @@ export const DataTableColumns = ({ drop, options, ...rest }) => {
   const { format } = useContext(MessageContext);
   const theme = useContext(ThemeContext);
   const [showContent, setShowContent] = useState();
+  const contextValue = useMemo(() => ({ inDataColumn: true }), []);
 
   const tip = format({
     id: 'dataTableColumns.tip',
     messages: messages?.dataTableColumns,
   });
 
-  let content = <Content drop={drop} options={options} />;
+  let content = (
+    <DataTableColumnContext.Provider value={contextValue}>
+      <Content drop={drop} options={options} />
+    </DataTableColumnContext.Provider>
+  );
   if (!inDataForm)
     content = (
       <DataForm footer={false} updateOn="change">
--- a/src/js/components/DataTableColumns/DataTableColumns.js
+++ b/src/js/components/DataTableColumns/DataTableColumns.js
@@ -6,7 +6,6 @@ import { Box } from '../Box';
 import { CheckBoxGroup } from '../CheckBoxGroup';
 import { DataForm, formColumnsKey } from '../Data';
 import { DataFormContext } from '../../contexts/DataFormContext';
-import { DataTableColumnsContext } from '../../contexts/DataTableColumnsContext';
 import { FormContext } from '../Form/FormContext';
 import { DropButton } from '../DropButton';
 import { List } from '../List';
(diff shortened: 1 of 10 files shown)


britt6612:
@taysea 

- Need test
- Need to change `Lock` to be able to be themed if that's what we want
- put `text-weak` in base.theme you can change or keep we can discuss.

src/js/components/List/List.js
```diff
@@ -349,6 +352,20 @@ const List = React.forwardRef(
...
+                if (pinned.length > 0) {
+                  if (typeof item === 'object' && !itemKey) {
+                    console.error(
+                      // eslint-disable-next-line max-len
+                      `Warning: Missing prop itemKey. Prop pin requires itemKey to be specified when data is of type 'object'.`,
+                    );
+                  }
+                  isPinned = pinned?.includes(key);
+                }
+                const pinTextColor = theme.list.item.pinned.text.color;
```
taysea, line 367: It feels a bit odd that this is in `list` theme but it only applies when in DataTableColumns. I thought we were planning on just hard coding the value for now as opposed to having it in theme?

So on line 381 below, just have `inDataColumn ? `text-weak` : undefined`

src/js/themes/base.js
```diff
@@ -1113,6 +1113,9 @@ export const generate = (baseSpacing = 24, scale = 6) => {
             size: 'medium',
             pad: 'small',
           },
+          text: {
+            color: 'text-weak',
```
taysea, line 1117: This doesn't seem backwards compatible

taysea: I thought we'd just inline this for DataTableColumns use cases but leave base.js as is?

src/js/components/List/List.js
taysea: I also wonder if there's anyway to consolidate that DataTableColumns logic in one place. Feels a little odd having it dispersed throughout what otherwise is just generic list code.

taysea (commented):
Thanks for getting this started! Just some initial thoughts/feedback.

src/js/contexts/DataTableColumnContext.js
```diff
@@ -0,0 +1,3 @@
+import React from 'react';
+
+export const DataTableColumnContext = React.createContext({});
```
taysea, line 3:
```suggestion
export const DataTableColumnsContext = React.createContext({});
```
Can we call it this to align with component name?

src/js/components/DataTableColumns/DataTableColumns.js
```diff
@@ -164,13 +165,18 @@ export const DataTableColumns = ({ drop, options, ...rest }) => {
   const { format } = useContext(MessageContext);
   const theme = useContext(ThemeContext);
   const [showContent, setShowContent] = useState();
+  const contextValue = useMemo(() => ({ inDataColumn: true }), []);
```
taysea, line 168:
```suggestion
  const contextValue = useMemo(() => ({ inDataTableColumns: true }), []);
```
Can we call it this to align with component name?

src/js/components/List/List.js
britt6612: yeah I can consolidate as well as just have `text-weak` I wasn't sure if we wanted it in theme but agree on naming maybe `dataTableColumns.pinned.text.color`

src/js/themes/base.js
britt6612: can do

src/js/contexts/DataTableColumnContext.js
britt6612: done

src/js/components/DataTableColumns/DataTableColumns.js
britt6612: done

src/js/components/List/List.js
```diff
@@ -640,9 +650,15 @@ const List = React.forwardRef(
                 let displayPinned;
                 if (isPinned) {
                   // Pinned icon and settings
-                  const Pin = theme.list.icons.pin;
+                  // If pinned is passed in dataTableColumns use Lock icon.
+                  const Icon = inDataTableColumns ? Lock : theme.list.icons.pin;
                   const pinSize = theme.list.item.pinned.icon.size;
                   const pinPad = theme.list.item.pinned.icon.pad;
+                  const iconColor = inDataTableColumns ? pinTextColor : null;
```
taysea, line 657:
```suggestion
                  const iconColor = inDataTableColumns ? pinTextColor : undefined;
```
Subtle nuance that might not have an effect on the behavior, but I think `undefined` is more fitting here. Also, even though we don't use Typescript, I know that they tend to lean away from `null` to favor `undefined`

src/js/components/List/List.js
```diff
@@ -640,9 +650,15 @@ const List = React.forwardRef(
...
+
+                  orderProps = {
```
taysea, line 659: Is the `onOrder + !isPinned` logic on line 499 not working as expected? I wouldn't have expected to need to add this here based on how the logic is written above.

src/js/themes/base.d.ts
```diff
@@ -1079,6 +1079,13 @@ export interface ThemeType {
         color?: ColorType;
         cursor?: string;
       };
+      pinned?: {
```
taysea, line 1082: Good catch that this was missing. Thanks for adding.

src/js/components/List/List.js
britt6612: not sure Im following the question above in the `onOrder + !isPinned` we have the logic for the dragging so that is where draggable is set to `true` so we need to set to `false` when pinned is true.

taysea: Right, I suppose what I meant is if an item `isPinned`, why did it get the draggable true applied to it? I would expect it to fail the `onOrder && !isPinned` condition

src/js/components/DataTableColumns/__tests__/__snapshots__/DataTableColumns-test.tsx.snap
```diff
@@ -302,14 +302,6 @@ exports[`DataTableColumns pinned 1`] = `
 
 }
 
-@media only screen and (max-width:768px) {
```
taysea, line 305: Is this expected?

britt6612: trying to see where this came from

src/js/themes/base.d.ts
britt6612: thanks!

taysea:
Screen Shot 2024-04-23 at 2 18 29 PM

Should the Lock padding match the button padding so it's in alignment? Maybe it's something we need to adjust in HPE theme?

britt6612:
> Screen Shot 2024-04-23 at 2 18 29 PM
> Should the Lock padding match the button padding so it's in alignment? Maybe it's something we need to adjust in HPE theme?

Let me double check but I beleive we would need to update `pin` pad in hpe theme

src/js/components/DataTableColumns/__tests__/DataTableColumns-test.tsx
```diff
@@ -203,4 +203,63 @@ describe('DataTableColumns', () => {
...
+
+    // Find the element by its role
+    const list = screen.getByRole('listbox');
+    const listItems = within(list).getAllByRole('listitem');
+
+    // Assert that the element is truthy
+    expect(listItems).toBeTruthy();
+
+    const nonDragElement = listItems[0];
+    expect(nonDragElement).not.toHaveAttribute('draggable', 'true');
```
taysea, line 257: Can we add an assertion that checks that the "Lock" icon is there? Can likely use the aria-label from the lock to getBy

britt6612:
closing in favor of https://github.com/grommet/grommet/pull/7213

PR #7205, opened 2024-04-16
Fix pagination step and stepOptions behaviors when step is controlled

taysea:
#### What does this PR do?
This PR fixes handling on `step` to ensure that in a controlled scenario, the internal state value is updated. It also ensures that `step` + `stepOptions` works.

Previously, https://github.com/grommet/grommet/pull/7188 caused the internal management of `stepOptions` to break.

For context, we had explored introducing a `useControlled` hook to reduce 1 extra re-render caused by the `useEffect` pattern. That said, this assumes that the `prop` being controlled (in this case `step`) has an equivalent change handler (which doesn't currently exist). Therefore, by passing `onChange` to the `useControlled` hook, we were incorrectly calling a change event more frequently than previously as well as not allowing `stepOptions` to update the internal step in cases where `step` was also used.

Given we follow the `useEffect` pattern elsewhere in Grommet, we will still with that for the time being as opposed to extending Pagination's API surface to include a step handler like `onStep`.

src/js/components/Pagination/Pagination.js

Locally in storybook.

```
import React from 'react';

import { Box, Pagination, Text, Select } from 'grommet';

export const Simple = () => {
  const [page, setPage] = React.useState(1);
  const [step, setStep] = React.useState(10);

  return (
    // <Grommet theme={...}>
    <Box align="start" pad="small" gap="medium">
      <Box>
        <Select
          options={[10, 25, 50, 100]}
          onChange={({ option }) => setStep(option)}
          value={step}
        />
        <Text>Default</Text>
        <Pagination
          numberItems={237}
          page={page}
          onChange={({ page: nextPage }) => {
            console.log(nextPage);
            setPage(nextPage);
          }}
          step={step}
          stepOptions
        />
      </Box>
    </Box>
    // </Grommet>
  );
};

export default {
  title: 'Controls/Pagination/Simple',
};
```

Added jest tests.

In storybook locally as mentioned above.

Closes https://github.com/grommet/grommet/issues/7201

No.

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/Pagination/__tests__/Pagination-test.tsx
+++ b/src/js/components/Pagination/__tests__/Pagination-test.tsx
@@ -444,6 +444,46 @@ describe('Pagination', () => {
     expect(updatedSelectButton).toBeTruthy();
   });
 
+  test('should respect stepOptions and controlled step', async () => {
+    window.scrollTo = jest.fn();
+    const user = userEvent.setup();
+    const App = () => {
+      const [step, setStep] = useState(10);
+      return (
+        <Grommet>
+          <Select
+            a11yTitle="Controlled select"
+            options={[10, 25, 50, 100]}
+            onChange={({ option }) => setStep(option)}
+          />
+          <Pagination numberItems={NUM_ITEMS} step={step} stepOptions />
+        </Grommet>
+      );
+    };
+    render(<App />);
+    await user.click(
+      screen.getByRole('button', { name: /Controlled select/i }),
+    );
+    await user.click(screen.getByRole('option', { name: '50' }));
+    // stepOptions should have updated
+    const updatedSelectButton = screen.getByRole('button', {
+      name: 'Open Drop; Selected: 50',
+    });
+    expect(updatedSelectButton).toBeTruthy();
+    expect(screen.getByRole('button', { name: 'Go to page 5' })).toBeTruthy();
+    // there should only be 5 pages based on step = 50 and NUM_ITEMS
+    expect(
+      screen.queryByRole('button', { name: 'Go to page 6' }),
+    ).not.toBeInTheDocument();
+    await user.click(updatedSelectButton);
+    await user.click(screen.getByRole('option', { name: '25' }));
+    expect(screen.getByRole('button', { name: 'Go to page 10' })).toBeTruthy();
+    // there should only be 10 pages based on step = 25 and NUM_ITEMS
+    expect(
+      screen.queryByRole('button', { name: 'Go to page 11' }),
+    ).not.toBeInTheDocument();
+  });
+
   test('should apply a text component with summary', () => {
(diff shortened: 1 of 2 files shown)


src/js/components/Pagination/__tests__/Pagination-test.tsx
```diff
@@ -444,6 +444,46 @@ describe('Pagination', () => {
     expect(updatedSelectButton).toBeTruthy();
   });
 
+  test('should respect stepOptions and controlled step', async () => {
+    window.scrollTo = jest.fn();
+    const user = userEvent.setup();
```
halocline, line 449: Related to our discussion yesterday, because we are using both fireEvent and userEvent in the same test script then we will receive test timeouts when running in parallel (e.g. `yarn test-watch`). This does not occur when using `--runInBand` which executes the tests serially.

Let's chat whether we want to refactor all tests in Pagination to use either fireEvent or userEvent. Or, explore the leads below.

Potentially a couple leads here:
https://github.com/testing-library/user-event/issues/833

https://github.com/testing-library/user-event/issues/833#issuecomment-1171452841

```
Using userEvent instead of fireEvent in the same test script can cause performance issues and timeout errors in the Testing Library. This is because userEvent provides more advanced browser interaction simulation than fireEvent, which can overcome the default Jest timeout limit of 5 seconds per test. 

github.com
Performance issues · testing-library user-event · Discussion #1144
Recently, in one of new tests, I've noticed that it often overcomes the default Jest timeout limit of 5s per test. After investigating for a while, I've discovered that the performance decrease was unfortunately caused by using userEvent instead of fireEvent. An example: a simple test ...

testing-library.com
user-event v13 - Testing Library
Nov 23, 2023 — user-event is a companion library for Testing Library that provides more advanced simulation of browser interactions than the built-in fireEvent method.
One solution to this issue is to set the delay option on the userEvent to null. For example, const user = userEvent.setup({ delay: null });. 
Here are some other tips for using the Testing Library:
Avoid wrapping fireEvent in act or await waitFor.
Use the async RTL methods that return promises and are resolved when an element is found. 
The Testing Library is a lightweight solution for testing that provides utilities for querying for nodes. It can fire events like a button click or a form submit using the fireEvent function.
```

taysea: I tried refactoring the code to all use `userEvent`, but now I'm getting errors where onChange is not being called the correct number of times. Seems as though counts are bleeding over from other tests.

Should we use fireEvent for now or just leave as is with mix of fireEvent/userEvent and improve later?

taysea: Going to leave as is for now.

halocline (approved):
Thank you!

britt6612 (approved):

Issue #6741, opened 2023-04-11
remove defaultProps from Grommet


samuelgoff:
Current version of React throws this error:
`Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.`

This occurs in over 100+ Grommet components because they're still using `defaultProps`

### Expected Behavior
Convert defaultProps to JavaScript default parameters instead.

### URL, screen shot, or Codepen exhibiting the issue
![Screenshot 2023-04-11 at 3 28 16 PM](https://user-images.githubusercontent.com/2007084/231280909-9eec7215-f84f-49f9-8021-f72bb93da078.png)
![Screenshot 2023-04-11 at 3 28 29 PM](https://user-images.githubusercontent.com/2007084/231280931-c451921c-cc0b-4642-bab3-df09d92f0e37.png)

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

ericsoderberghp:
I think the main source of this is the use of `defaultProps` within the Styled* components as a means of at least passing the base theme in when there is no theme context. This happens when grommet components are used outside of a `<Grommet />` context. I am hoping we can find an alternative to deliver the base theme in these contexts as it would break backward compatibility to require an outer Grommet context.

chrisphillers:
I am getting the same issue and it is crashing my Next.js 13.5 app.
Wondered if this was likely to be resolved? Thanks

undeletable:
@ericsoderberghp I've started working on this, and I think I'll have a PR in a few days

britt6612:
@undeletable thank you looking forward to your PR

PR #7224, opened 2024-05-09
Remove defaultProps usages for functional components

undeletable:
#### What does this PR do?

Starting from React 19, `defaultProps` for functional components is deprecated. Starting from React 18.3, the corresponding error is thrown for a functional component with `defaultProps` defined.

Changes in this PR:
- replace `defaultProps` for functional components with default function parameters values;
- use `attrs` method of `styled-components` to pass theme to styled components which might be outside of theme context (required per https://github.com/grommet/grommet/issues/6741#issuecomment-1533232289);
- add custom hook to obtain theme from context with fallback to default value;
- configure eslint rule to not require `defaultProps` for functional components;
- update snapshots.

No specific place to start.

- manual testing on storybook and real project using grommet;
- tests run result:
```
Test Suites: 97 passed, 97 total
Tests:       1548 passed, 1548 total
Snapshots:   1777 passed, 1777 total
```

Use React 18.3+ in an app. Check that styles are applied correctly and no errors about `defaultProps` are thrown into console.

[React 19 upgrade guide](https://react.dev/blog/2024/04/25/react-19-upgrade-guide)

Closes https://github.com/grommet/grommet/issues/6741

N/A.

Just the internal ones (if any).

Yes.

Changes are backwards compatible.


britt6612:
@undeletable Thank you for your contribution! We are in the process of looking at this just wanted to let you know!

jcfilben (commented):
Thanks for working on this!

In the changes I saw that you are using the `attrs` method of styled-components to pass the theme. In the PR notes you referenced this comment https://github.com/grommet/grommet/issues/6741#issuecomment-1533232289 as the reason for those changes. I think this comment was misinterpreting the original intent of removing `defaultProps`.

There is some confusing naming where we have `defaultProps` in the react sense and then we have `defaultProps` in Grommet that contains the base theme. I think https://github.com/grommet/grommet/issues/6741#issuecomment-1533232289, misunderstood the issue and was thinking that we would remove the `defaultProps` that has the Grommet theme instead of removing the react `defaultProps`.

All this to say I don't think the comment is relevant for what we are trying to do here with removing the react defaultProps and we don't need to use the `attrs` method.

We did also test to confirm that the base theme is being passed through correctly to components that are not in a Grommet wrapper in react 18.3.0 without using `attrs`.

undeletable:
@jcfilben 
> There is some confusing naming where we have defaultProps in the react sense and then we have defaultProps in Grommet that contains the base theme. 

Actually these two entities are connected: Grommet's `defaultProps` is used for as prototype for components' `defaultProps`.

Another hint for me was that component snapshots are way too different without both React's `defaultProps` and `attrs` usages, and almost equivalent with `attrs`.

> We did also test to confirm that the base theme is being passed through correctly to components that are not in a Grommet wrapper in react 18.3.0 without using attrs.

Do I understand correctly that I should just remove `attrs` usages from my PR?

undeletable:
@jcfilben 
As far as I see, there are two ways of using theme for components.
**Way 1. Using inside the component function:**
```
const Card = forwardRef(({ ...rest }, ref) => {

  return (...);
});
```
Here, the cause of using component outside of theme context is handled by using fallback vaue.
**Way 2. Using inside arguments of styled-components' template function:**
```
const StyledMinute = styled.line.withConfig(styledComponentsConfig)`
  stroke-width: ${(props) => props.theme.clock.analog.minute.width};
  stroke: ${(props) =>
    normalizeColor(props.theme.clock.analog.minute.color, props.theme)};
  transition: stroke 1s ease-out;
`;

StyledMinute.defaultProps = {};
Object.setPrototypeOf(StyledMinute.defaultProps, defaultProps);
```
If `defaultProps` is removed for such a component, we need another way to pass default theme as fallback. For such cases I'm using `attrs()`. In the most recent commit I just removed `attrs()` calls for the places where `props.theme` is not used.

ALso, I've tried to remove all `attrs()` calls from my changes. As result, several test suites failed: this was caused by failed attempts to get properties of theme, which was not defined.

Please correct me if I'm wrong.

jcfilben:
Thanks for the explanation, and sorry for the confusion on my end. Yesterday my team and I had tested in storybook by removing the Grommet wrapper and `attrs` and was still seeing the styles coming through so I was thinking we were good. But I just ran the unit tests and I see what you are saying about the difference in styles. I also had missed that we are currently using react defaultProps at the end of each Styled components file.
```
StyledMinute.defaultProps = {};
Object.setPrototypeOf(StyledMinute.defaultProps, defaultProps);
```
I think the direction you've taken with `attrs` makes sense.

src/js/default-props.js
```diff
@@ -8,3 +8,11 @@ export const defaultProps = {
 export const extendDefaultTheme = theme => {
   defaultProps.theme = deepMerge(base, theme);
 };
+
+/*
+  Pass `theme` for component which can be loacted outside of theme context.
+  To be used as argument for `attrs` method from `styled-components`.
+ */
+export const ehnancePropsWithTheme = props => ({
```
jcfilben, line 16: Minor spelling change
```suggestion
export const enhancePropsWithTheme = props => ({
```

undeletable: Such a silly thing :) Thanks for pointing, fixed that.

jcfilben:
If you update your branch with the latest changes from the master branch, the chromatic check should pass

jcfilben:
I'm going to try and push a no-op commit to your PR to see if I can re-trigger the chromatic step on circleci

undeletable:
@jcfilben 
> locally when I pull down this PR I have some tests failing

For me, some tests for Pagination are slowly running locally - both in master and my branch

undeletable:
@jcfilben 
From what I see, there's a different CircleCI project used for the fork:
https://app.circleci.com/pipelines/github/grommet/grommet
https://app.circleci.com/pipelines/github/undeletable/grommet

mshepel:
@jcfilben On my local env the following test is failing almost always:
```
  test('should update page range based on step prop dynamically changing', async () => {
    render(<TestComponent />);
    // Open select
    await userEvent.click(screen.getByRole('button', { name: /Open Drop/i }));
    // Click on the option '10'
    await userEvent.click(screen.getByRole('option', { name: '10' }));

    // Expect input value to be 10
    const updatedSelectButton = screen.getByRole('button', {
      name: 'Open Drop; Selected: 10',
    });
    expect(updatedSelectButton).toBeTruthy();
    expect(screen.getByText(`Showing 1-10 of 100`)).toBeTruthy();

    // Open select again
    await userEvent.click(screen.getByRole('button', { name: /Open Drop/i }));

    // Click on the option '50'
    await userEvent.click(screen.getByRole('option', { name: /50/i }));

    // Expect input value to be 50
    const updatedSelectButton1 = screen.getByRole('button', {
      name: 'Open Drop; Selected: 50',
    });
    expect(updatedSelectButton1).toBeTruthy();
    expect(screen.getByText(`Showing 1-50 of 100`)).toBeTruthy();

    // Open select again
    await userEvent.click(screen.getByRole('button', { name: /Open Drop/i }));

    // Click on the option '100'
    await userEvent.click(screen.getByRole('option', { name: /100/i }));

    // Expect input value to be 100
    const updatedSelectButton2 = screen.getByRole('button', {
      name: 'Open Drop; Selected: 100',
    });
    expect(updatedSelectButton2).toBeTruthy();
    expect(screen.getByText(`Showing 1-100 of 100`)).toBeTruthy();
  });
```
Thrown error:
```
"Exceeded timeout of 5000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."
```
Comparing to other tests in at least that module, there are many actions performed inside this one. Looks like this is the cause.

undeletable:
@jcfilben On my local env the following test is failing almost always:
```
  test('should update page range based on step prop dynamically changing', async () => {
    render(<TestComponent />);
    // Open select
    await userEvent.click(screen.getByRole('button', { name: /Open Drop/i }));
    // Click on the option '10'
    await userEvent.click(screen.getByRole('option', { name: '10' }));

    // Expect input value to be 10
    const updatedSelectButton = screen.getByRole('button', {
      name: 'Open Drop; Selected: 10',
    });
    expect(updatedSelectButton).toBeTruthy();
    expect(screen.getByText(`Showing 1-10 of 100`)).toBeTruthy();

    // Open select again
    await userEvent.click(screen.getByRole('button', { name: /Open Drop/i }));

    // Click on the option '50'
    await userEvent.click(screen.getByRole('option', { name: /50/i }));

    // Expect input value to be 50
    const updatedSelectButton1 = screen.getByRole('button', {
      name: 'Open Drop; Selected: 50',
    });
    expect(updatedSelectButton1).toBeTruthy();
    expect(screen.getByText(`Showing 1-50 of 100`)).toBeTruthy();

    // Open select again
    await userEvent.click(screen.getByRole('button', { name: /Open Drop/i }));

    // Click on the option '100'
    await userEvent.click(screen.getByRole('option', { name: /100/i }));

    // Expect input value to be 100
    const updatedSelectButton2 = screen.getByRole('button', {
      name: 'Open Drop; Selected: 100',
    });
    expect(updatedSelectButton2).toBeTruthy();
    expect(screen.getByText(`Showing 1-100 of 100`)).toBeTruthy();
  });
```
Thrown error:
```
"Exceeded timeout of 5000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."
```
Comparing to other tests in at least that module, there are many actions performed inside this one. Looks like this is the cause.

src/js/components/DataFilter/__tests__/__snapshots__/DataFilter-test.tsx.snap
jcfilben, line 1: Do you know why this snapshot is changing?

undeletable: I'm not sure about that yet. In fact, the changes are about reordering styles and removing the following one:
```
.c8:hover {
  box-shadow: 0px 0px 0px 2px #7D4CDB;
}
```
I'll check why that happens and if that's expected

src/js/components/Menu/__tests__/Menu-test.js
```diff
@@ -49,7 +49,7 @@ describe('Menu', () => {
   test('should have no accessibility violations', async () => {
     const { container } = render(
       <Grommet>
-        <Menu />
+        <Menu items={[]} />
```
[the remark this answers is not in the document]
undeletable: At the moment I've added it, using `Menu` with no `items` prop passed resulted in test failure. I'll check if that's still the case and why, if yes.

src/js/components/DataFilter/__tests__/__snapshots__/DataFilter-test.tsx.snap
jcfilben: I looked into this test issue a bit more, I think what is happening is when the SelectMultiple drop in this test opens sometimes the cursor lands on the 'Select All' button and sometimes it does not which causes the snapshot difference. I think this issue stems from using `userEvent`, there are other tests that are similar (they take a snapshot of the SelectMultiple dropdown) that use `fireEvent` and the snapshots there are consistent. We like to try and use `userEvent` when possible because it mimics user behavior more realistically but we have run into issues in the past with using it instead of `fireEvent`. Another thing to note is that this snapshot difference is not isolated to your PR and while testing it we did see it pop up in other branches as well.

We could solve this issue by swapping out `userEvent` for `fireEvent` in this test, but ultimately I think the snapshot of the SelectMultiple drop down doesn't add much value to what the original intent of the test is trying to capture. In Grommet we have been trying to move away from snapshot testing unless it actually provides value to the test. I will put up a PR to remove this test snapshot.

undeletable: @jcfilben 
I've detected that the difference in styles is coming from the following callback in `StyledButton`:
```
  ${(props) =>
    !props.disabled &&
    !props.selected &&
    !props.focus &&
    !props.busy &&
    !props.success &&
    hoverStyle}
```
Should I proceed with any changes related to DataFilter tests in scope of current PR?

jcfilben: Thanks for looking into this, no need to make changes related to the DataFilter tests in this PR

britt6612:
@undeletable 
Hey I was looking over this PR and Im glad to see that you caught `Pagination` in that it is throwing an error with not wrapped in `<grommet> wrapper so I went ahead and added tests for the components on which other components also have this issue and found quite a few I was wondering if you would be open to fixing these and adding the `attrs` to these components 
`List` `ToggleGroup` `DataChart` `NameValueList` `RangeSelector` `WorldMap` `FileInput``PageHeader` `StarRating` 
These are just a few there are a few more.

mshepel:
@britt6612 Pagination tests fail sometimes because of different reason (slowness), but I got your point. I'll take a look.

undeletable:
@britt6612 Pagination tests fail sometimes because of different reason (slowness), but I got your point. I'll take a look.

undeletable:
@jcfilben @britt6612 I've noticed that target branch in https://github.com/grommet/grommet/pull/7249 is different fro master. Should I change target branch to `react-18.3.0-upgrade` in my PR?

undeletable:
@britt6612 I've added fallbacks to default theme to all the remaining places I saw that might be needed. Tests on both my and your branch passed.

britt6612:
> @jcfilben @britt6612 I've noticed that target branch in #7249 is different from master. Should I change target branch to `react-18.3.0-upgrade` in my PR?

@undeletable so yes what I did was make that new PR which will get merged into `react-18.3.0-upgrade` then we will point your PR to `react-18.3.0-upgrade` that way we can make sure tests are good before merging to master!

Thank you for adding this for the rest of the components

britt6612:
@undeletable thank you! Hoping my PR gets merged early next week! then we can merge yours!

jcfilben (approved):
Looks good!

src/js/components/Diagram/Diagram.js
```diff
@@ -66,198 +63,200 @@ const findTarget = (target) => {
   return target;
 };
 
-const Diagram = forwardRef(({ connections, ...rest }, ref) => {
-  const theme = useContext(ThemeContext) || defaultProps.theme;
-  const [dimensions, setDimensions] = useState({ width: 0, height: 0 });
-  const [connectionPoints, setConnectionPoints] = useState();
-  const svgRef = useForwardedRef(ref);
+const defaultConnectionsValue = [];
```
britt6612, line 66: why was this added?

src/js/components/Distribution/propTypes.js
```diff
@@ -43,7 +43,7 @@ if (process.env.NODE_ENV !== 'production') {
       PropTypes.shape({
         value: PropTypes.number,
       }),
-    ).isRequired,
```
britt6612, line 46: 👍

src/js/components/Grid/StyledGrid.js
```diff
@@ -216,6 +216,7 @@ const areasStyle = (props) => {
 const StyledGrid = styled.div
   .withConfig(styledComponentsConfig)
   .attrs((props) => ({
+    ...enhancePropsWithTheme(props),
```
britt6612, line 219: why is this one formatted different just trying to understand

src/js/components/Diagram/Diagram.js
undeletable: It's added in order to keep parity with `defaultProps` approach and avoid potentially redundant rerenders or even rerender loops.
Let's consider a prop with value of object type. Assume it has no value passed and falls back to value from `defaultProps`. In this case, the value would be equal between renders (`props.prop === defaultProps.prop`), since `defaultProps` is static.
If we replace `defaultProps` with default function parameters, and the default value of mentioned prop is defined as object literal, this means that on each render its values would be different. On each render component function is called again, so default parameter value is re-evaluated. If component state depends on such prop (e.g., via `useEffect` or `useMemo`), this would result in infinite rerenders loop.
That's why I've initialized such props with stable links. Actually this approach was already used in some other places of grommet.

src/js/components/Grid/StyledGrid.js
undeletable: Because `attrs` was already used for passing `aria-label` (see the line below). So using `attrs(enhancePropsWithTheme)` would result in losing that property.

britt6612: oh gotcha thank you for clarifying

src/js/components/Diagram/Diagram.js
britt6612: Thank you ☝️

britt6612 (approved):

britt6612:
Thank you @undeletable for all your help! we have a few more things to add but this was a big help!

PR #7226, opened 2024-05-13
add selected prop to be passed through

britt6612:
#### What does this PR do?
adds the `selected` prop to be passed through to styledButtonResult
button.js
locally
locally

Users can leverage this `selected` prop. This is passed to styledButton but not to styledButtonResult
Related to https://github.com/grommet/grommet-theme-hpe/issues/390

no
yes
backwards compatible

### DIFF
--- a/src/js/components/Button/Button.js
+++ b/src/js/components/Button/Button.js
@@ -517,6 +517,7 @@ const Button = forwardRef(
           pad={pad}
           plain={plain || Children.count(children) > 0}
           primary={primary}
+          selected={selected}
           sizeProp={size}
           success={success}
           type={!href ? type : undefined}


jcfilben (dismissed):
Looks good!

jcfilben:
I initially approved but then looking at the docs and the types for button I noticed we are only using the selected prop internally. From what I understand the intent is to use this prop to flag styling in the extend for the hpe theme. Just pausing to think if it is okay to use an internal variable within the theme

taysea:
> I initially approved but then looking at the docs and the types for button I noticed we are only using the selected prop internally. From what I understand the intent is to use this prop to flag styling in the extend for the hpe theme. Just pausing to think if it is okay to use an internal variable within the theme

Agree it's not the ideal solution, but aligning to what is already being passed through to StyledButton. I'd say if a theme (in this case HPE) leverages an undocumented prop then they are responsible for the risk that might be associated with that.

In future (when we have more bandwidth), we should enhance the button theme to support selected + hover.

jcfilben (approved):

Issue #7127, opened 2024-02-09
Can not upgrade certain dependencies 


britt6612:
Currently we are unable to upgrade the following dependencies at this time. 
```
    "@testing-library/dom": "9.3.0",
    "babel-plugin-styled-components": "2.0.6",
    "del": "^6.1.1",
    "jest-styled-components": "7.0.5",
   "jsdom": "^21.1.2",
    "styled-components": "5.3.9",
    "testcafe": "3.3.0",
```

- "@testing-library/dom": "9.3.0",

```
 Uncaught [Error: Uncaught 'Warning: An update to %s inside a test was not wrapped in act(...).\n' 
```

-  "babel-plugin-styled-components": "2.0.6",
    -  Test suite failed to run

- "testcafe": "3.3.0",
<img width="996" alt="Screen Shot 2024-02-07 at 6 38 32 PM" src="https://github.com/grommet/grommet/assets/42451602/2886a4df-cb03-43e2-9990-9569bd0ebb33">

- "jest-styled-components": "7.0.5",
This was getting rid of classnames in the snap shots
<img width="588" alt="Screen Shot 2024-02-07 at 6 44 20 PM" src="https://github.com/grommet/grommet/assets/42451602/34e34604-90ca-474c-9025-6929706d1fc6">

- "styled-components": "6.1.8",

    thrown: "Warning: React does not recognize the `%s` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `%s` instead. If you accidentally passed it from a parent component, remove it from the DOM element.%s"

britt6612:
Adding to the dependencies that we can not update at this time 
```
    "react-dom": "^18.2.0",
    "react-is": "^18.2.0",

```

  thrown: "Warning: `ReactDOMTestUtils.act` is deprecated in favor of `React.act`. Import `act` from `react` instead of `react-dom/test-utils`. See https://react.dev/warnings/react-dom-test-utils for more info."

We will need to go over 200+ tests if we want to update we will need to update our tests to use act and import from React.
`Warning: An update to %s inside a test was not wrapped in act(...).\n`

https://react.dev/warnings/react-dom-test-utils

https://github.com/facebook/react/releases

jcfilben:
> Adding to the dependencies that we can not update at this time
> 
> ```
>     "react-dom": "^18.2.0",
>     "react-is": "^18.2.0",
> ```
> 
[...]

PR #7227, opened 2024-05-13
Temporarily freeze react version, upgrade other dependencies

jcfilben:
#### What does this PR do?
Temporarily freeze react version at v18.2.0 until we have a chance to do that upgrade (tracked here: https://github.com/grommet/grommet/issues/7127).

This PR upgrades some other dependencies

no
no
backwards compatible


src/js/components/Anchor/__tests__/__snapshots__/Anchor-test.tsx.snap
```diff
@@ -464,8 +464,8 @@ exports[`Anchor matches icon size to size prop when theme.icon.matchSize is true
   -webkit-flex: 0 0 auto;
   -ms-flex: 0 0 auto;
   flex: 0 0 auto;
-  width: NaNpx;
-  height: NaNpx;
+  width: NaNxsmall;
```
britt6612, line 467: do we want these? I remember they are from `grommet-icons` wasn't sure if we looked into why they changed

jcfilben: I'm guessing it is coming from this PR, but not positive: https://github.com/grommet/grommet-icons/pull/325/files
Either way I think visually this test wouldn't change. The width is just being replaced with another invalid option.
At some point we should look at the `useSizedIcon` function in Grommet, I think this is where the issue may be coming from. I think the icons are supposed to scale to the size of the Anchor. This isn't working currently and this PR doesn't change that it just changes the incorrect value to another incorrect value.
Screen Shot 2024-05-14 at 9 00 58 AM

taysea: Do we have a ticket somewhere to capture?

jcfilben: Issue created: https://github.com/grommet/grommet/issues/7231

taysea (approved):

Issue #7177, opened 2024-03-28
DataTable: changing size of cells does not work in every situation


stephanpelikan:
### Expected Behavior

I need to show details for a record by expanding it's row size (changing the content of one column). On collapsing the row's size decreases to the content height of the cell. This should work for normal-, plain-columns and also having multiple plain plain-columns.

### Actual Behavior

Expanding and collapsing works if there is no or only one plain-column. Once more than one plain-column is used, the columns can be expanded but won't reduce their height any more on collapsing the content.

The reason for this is because for plain-columns an additional div is inserted: https://github.com/grommet/grommet/blob/master/src/js/components/TableCell/TableCell.js#L128. On expanding heights all those div-elements will be set based on the surrounding cell: https://github.com/grommet/grommet/blob/master/src/js/components/TableCell/TableCell.js#L73. So, if the expanded plain-cell collapses it's content, then the others plain-cell's height is expanded because their cell-height changed according to the expanded cell (all cells of a row have the same height). In the end, the row can only expand it's height.

Actually, a fix is if the height is not measured based on the cell but rather on the additional div-element. But that might break the issue mentioned in this comment, which is the reason for using this additional div-element: https://github.com/grommet/grommet/blob/master/src/js/components/TableCell/TableCell.js#L129.

### Steps to Reproduce

I created a sandbox using one storybook's example:

https://codesandbox.io/p/sandbox/datatable-collapse-issue-2jq6d9

In the "percent complete" column click on the percent value to change to the graphical representation what expands the cell. Another click on the graphical representation will collapse the cell.

This percent-column is of type plain. Once change another column to plain (see line 322), you will experience the bug.

### Your Environment

- Grommet version: 2.34.2
- Browser Name and version: Chrome, Firefox, Edge
- Operating System and version (desktop or mobile): Windows, MacOS

stephanpelikan:
Based on a Google search "div fit into td" there first a hit showing a CSS solution (https://jsfiddle.net/3mx8hvpo/12/). I applied this to grommet code-base and it works fine. I tested with Chrome, Firefox and Safari.

Would you mind to accept this https://github.com/stephanpelikan/grommet/commit/7586a993a0e14445bbcb3984f7d05b7793aa5881 as a PR?

PR #7229, opened 2024-05-14
DataTable: changing size of cells does not work in every situation

stephanpelikan:
#### What does this PR do?

In DataTable the height is calculated in code what does not work on resizing any column's content (see https://github.com/grommet/grommet/issues/7177) if there is another plain-column. This PR is a pure CSS solution for the problem.

See section "Actual Behavior" of https://github.com/grommet/grommet/issues/7177.

We manually tested both situations: On initializing the table and on changing cells content. We use this fork already in production.

In "Steps to reproduce" of https://github.com/grommet/grommet/issues/7177 a sandbox is provided. To test correct sizing on table initialization one can set the default-value to true at https://codesandbox.io/p/sandbox/datatable-collapse-issue-2jq6d9?file=%2Findex.js%3A255%2C1-256%2C1. Additionally, one can remove the line 
https://codesandbox.io/p/sandbox/datatable-collapse-issue-2jq6d9?file=%2Findex.js%3A310%2C1-311%2C1 to also test for non-plain columns.

The author of the initial implementation didn't see that there is a pure CSS solution (see comment in https://github.com/grommet/grommet/blob/master/src/js/components/TableCell/TableCell.js#L129). Meanwhile there were changes in the same section https://github.com/grommet/grommet/commit/10c46fd10273788d5ef86c00d46403b537f84d44 which should be tested. I updated the Sandbox to Grommet's current version and the intermediate change does not effect the behavior of this PR.

https://github.com/grommet/grommet/issues/7177

No

No

As you wish

Was backward compatible before https://github.com/grommet/grommet/commit/10c46fd10273788d5ef86c00d46403b537f84d44. If the underlying issue of this change https://github.com/grommet/grommet/pull/6086 is still solved with the PR in place, then it is confirmed that it still is backward compatible.


taysea:
Hi @stephanpelikan, thanks for putting up this PR. There seems to be a bit of history in the repo when it comes to table sizing, so we're taking a bit more time to understand that before we drop PR comments. Just wanted to let you know, thank you!

stephanpelikan:
Closed in favor to https://github.com/grommet/grommet/pull/7283 which is rebased onto the current release branch.

Issue #7225, opened 2024-05-10
Button -- `busy` indicator gets squished when button is narrow


taysea:
### Expected Behavior

The "..." in `busy` Button indicator should stay as circles, not get squished.

### Actual Behavior

The "..." gets squished when the button is narrow

### URL, screen shot, or Codepen exhibiting the issue

https://github.com/grommet/grommet/assets/12522275/48fae3fb-d711-4b01-a65f-fcd5e9bfeb64

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version: 2.37.0
- Browser Name and version: Chrome
- Operating System and version (desktop or mobile): MacOS

britt6612:
The padding isnt allowing for those `dots` to be the correct size.

PR #7235, opened 2024-05-20
fix busy button dots to be smaller when button is small

britt6612:
#### What does this PR do?
This PR changes the size of the dots in the busy animation button
busyAnimation.js
storybook
storybook

https://github.com/grommet/grommet/issues/7225
closes https://github.com/grommet/grommet/issues/7225
no
maybe
backwards compatible

### DIFF
--- a/src/js/components/Button/BusyAnimation.js
+++ b/src/js/components/Button/BusyAnimation.js
@@ -15,17 +15,18 @@ const bounceDelayRule = css`
   animation: ${bounceDelay} 1.4s infinite ease-in-out both;
 `;
 
+/* When button is small size we need half the dot size to fit properly */
 const Dot = styled(Box)`
   background-color: currentColor;
-  width: 8px;
-  height: 8px;
+  width: ${(props) => (props.size.size === 'small' ? '4px' : '8px')};
+  height: ${(props) => (props.size.size === 'small' ? '4px' : '8px')};
   border-radius: 100%;
   display: inline-block;
   ${bounceDelayRule}
   ${(props) => props.delay && `animation-delay: ${props.delay};`}
 `;
 
-export const EllipsisAnimation = () => (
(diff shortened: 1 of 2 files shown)


src/js/components/Button/BusyAnimation.js
```diff
@@ -15,17 +15,18 @@ const bounceDelayRule = css`
   animation: ${bounceDelay} 1.4s infinite ease-in-out both;
 `;
 
+/* When button is small size we need half the dot size to fit properly */
 const Dot = styled(Box)`
   background-color: currentColor;
-  width: 8px;
-  height: 8px;
+  width: ${(props) => (props.size.size === 'small' ? '4px' : '8px')};
```
taysea, line 21: It's not necessarily that the button has prop `size="small"` but rather that if the label is really short, the width of the button is more narrow, causing the "dots" to squish.

I'm wondering if instead these "Dot"s should just have a flex={false}?

jcfilben: If the dots have `flex={false}` and the label is really short would the Button size grow horizontally when it enters the busy state?

britt6612: aw I didnt check the label so even on a medium size button with a small label this is happening

britt6612: ☝️ 
so yes when we have `flex={false}` then the dots are not in the correct space. We may need to check the size of the button if it is under 100px seems to be where it is happening then change the values. 

https://github.com/grommet/grommet/assets/42451602/cc8068be-aadb-4448-afff-796f6eb2a837

britt6612:
Converting to draft for now..
The issue is even if we have a `min-height` and `min-width` for the dots there is still not enough room in the middle (because of the `pad`) of the bottom therefore the 3 dots are overflowing and getting shifted to the right.

Issue #7215, opened 2024-04-25
Data not being updated correctly


britt6612:
### Expected Behavior
I expect that if you click on the last page in pagination than change your items per page it should load to show page 1 on its own 

### Actual Behavior
I expect that if you click on the last page in pagination than change your items per page you have to manually click on page 1 to see the correct data. 

If you go to the following story 
https://storybook.grommet.io/?path=/story/data-data-onselect--on-select
Then click on the last page in pagination then change the `items per page` to 100 then it keeps you on the last page where you can not see data. You have to click back on page 1.
### Steps to Reproduce

1. https://storybook.grommet.io/?path=/story/data-data-onselect--on-select
2. click on last page in pagination 21
3. change items per page to 100
4. see no data in the DataTable
5. have to click back on page 1

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #7237, opened 2024-05-21
Pagination - fix issue when step size changes and current page isn't visible

jcfilben:
#### What does this PR do?
Fixes an issue in Pagination when you are on a page and then change the step size in such a way that the current page you are on no longer exists. The current behavior is that no page will be selected, this PR changes the behavior so that the user lands on the last page (this is so that they will be closest to the data they were previously viewing before changing the step size).

Added jest test and tested with existing storybook stories

Closes https://github.com/grommet/grommet/issues/7215

no
yes
backwards compatible

### DIFF
--- a/src/js/components/Pagination/Pagination.js
+++ b/src/js/components/Pagination/Pagination.js
@@ -269,7 +269,14 @@ const Pagination = forwardRef(
               messages={messages}
               options={Array.isArray(stepOptions) ? stepOptions : undefined}
               step={step}
-              onChange={({ value }) => setStep(value)}
+              onChange={({ value }) => {
+                setStep(value);
+                const newTotal = Math.ceil(total / value);
+                if (activePage > newTotal) {
+                  setActivePage(newTotal);
+                  if (onView) onView({ ...view, page: newTotal });
+                }
+              }}
             />
           )}
           {paginationControls}
--- a/src/js/components/Pagination/__tests__/Pagination-test.tsx
+++ b/src/js/components/Pagination/__tests__/Pagination-test.tsx
@@ -484,6 +484,28 @@ describe('Pagination', () => {
     ).not.toBeInTheDocument();
   });
 
+  test.only('move to last page when current page is > than available pages', async () => {
+    window.scrollTo = jest.fn();
+    const user = userEvent.setup();
+
+    const { asFragment } = render(
+      <Grommet>
+        <Pagination stepOptions={[1, 3, 5]} numberItems={10} step={1} />
+      </Grommet>,
+    );
+
+    await user.click(screen.getByRole('button', { name: /10/i }));
+    await user.click(screen.getByRole('button', { name: /Open Drop/i }));
+    await user.click(screen.getByRole('option', { name: /5/i }));
+
+    expect(screen.getByRole('button', { name: 'Go to page 2' })).toBeTruthy();
+    expect(
+      screen.queryByRole('button', { name: 'Go to page 3' }),
+    ).not.toBeInTheDocument();
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+
   test('should apply a text component with summary', () => {
     const { asFragment } = render(
       <Grommet>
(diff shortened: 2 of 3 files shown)


britt6612:
more of a question.. should the user be put back to page one? after re-selecting `items per page`


https://github.com/grommet/grommet/assets/42451602/ac4a5b58-81d8-453f-b9c9-0ce592211887

jcfilben:
> more of a question.. should the user be put back to page one? after re-selecting items per page

I put them on the last page because then they would be closer to the data that they were previously viewing before changing the step

britt6612:
> > more of a question.. should the user be put back to page one? after re-selecting items per page
> 
> I put them on the last page because then they would be closer to the data that they were previously viewing before changing the step

gotcha just wanted to make sure that was the correct pattern!

britt6612 (approved):

taysea:
To just speak out loud, is it correct to say our logic is "as `step` changes, we keep the previously active `page` active. For example, if user was on page 5 and they change from step of 10 to 25, we'll stay on step 5. In cases where the page no longer exists because there are less total pages, we'll fallback to the last page"?

I'm trying to figure out what I'd expect in a scenario like this where I was on page 14, then increased the step, and then was on the last page of data. 
https://github.com/grommet/grommet/assets/12522275/34f41ec8-4f2f-4af1-b637-31dda33ba14a

@halocline any thoughts on behavior?

halocline:
> To just speak out loud, is it correct to say our logic is "as `step` changes, we keep the previously active `page` active. For example, if user was on page 5 and they change from step of 10 to 25, we'll stay on step 5. In cases where the page no longer exists because there are less total pages, we'll fallback to the last page"?
> 
> I'm trying to figure out what I'd expect in a scenario like this where I was on page 14, then increased the step, and then was on the last page of data.

> @halocline any thoughts on behavior?

I agree that we should pause and think through this a bit further. The page 14 to last page scenario isn't right - because last page is much further away.

We are either "smart" and do something like ensure the index of the first item in the current page is in the resulting subset in the next page.

Or, always go back to page 1.

We also need to consider the controlled scenario as well.

jcfilben:
> I agree that we should pause and think through this a bit further. The page 14 to last page scenario isn't right - because last page is much further away.

I'm confused by the statement 'last page is much further away', if we have a scenario where we are on page 14 and the step changes in such a way that page 14 no longer exists and we are moved to the last page then the data on the last page will include data from page 14

MikeKingdom:
> > I agree that we should pause and think through this a bit further. The page 14 to last page scenario isn't right - because last page is much further away.
> 
> I'm confused by the statement 'last page is much further away', if we have a scenario where we are on page 14 and the step changes in such a way that page 14 no longer exists and we are moved to the last page then the data on the last page will include data from page 14

I think point is that isn't necessarily true. You probably could do the math to figure out what page in the new step size contains the top item from the previous current page. But we do need to think a little more about the various cases and if we're screwing up the controlled situations (ie. not letting the caller control where it ends up when step changes.)

jcfilben:
> I think point is that isn't necessarily true. You probably could do the math to figure out what page in the new step size contains the top item from the previous current page. But we do need to think a little more about the various cases and if we're screwing up the controlled situations (ie. not letting the caller control where it ends up when step changes.)

Ah okay I was thinking of page 14 as the last page, but if page 14 is a middle page I see what you are saying.

jcfilben:
Slow brain haha sorry, should we just default to going back to the first page?

jcfilben:
Writing out logic for going back to Page 1
```
If the current page no longer exists
  If `pageProp` and `onChange` are defined
    Do nothing, Pagination is in a controlled state
  Else set page to the first page
```
If a different behavior is desired then it can be achieved through a controlled version of Pagination

taysea:
Writing out logic for keeping previously visible data visible

```
If the current page no longer exists
  Determine what was the "first" record that was previously visible
  Set page to the page that now contains record
  If `pageProp` and `onChange` are defined
    Do nothing, Pagination is in a controlled state
```

If a different behavior is desired then it can be achieved through a controlled version of Pagination

PR #7238, opened 2024-05-21
ToggleGroup - fix issue where button rounding doesn't align with toggle group border-radius

jcfilben:
#### What does this PR do?
There is a slight gap in the rounding between the border of the ToggleGroup container and the buttons within. It is easiest to see this in the dark mode grommet theme. This PR fixes this so that there isn't a gap.
before:
Screen Shot 2024-05-21 at 12 16 11 PM
after:
Screen Shot 2024-05-21 at 12 35 06 PM

no
yes
backwards compatible

### DIFF
--- a/src/js/components/ToggleGroup/ToggleGroup.js
+++ b/src/js/components/ToggleGroup/ToggleGroup.js
@@ -115,6 +115,7 @@ const ToggleGroup = ({
         // match button rounding
         responsive={false}
         role={multiple ? 'group' : 'radiogroup'}
+        overflow="hidden"
         {...rest}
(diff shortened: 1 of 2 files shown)


taysea:
This solution seems to crop out the focus indicator.

https://github.com/grommet/grommet/assets/12522275/33ae9a85-3c64-4531-a9ee-d57814568a5b

jcfilben:
> This solution seems to crop out the focus indicator.

Oops, good catch. Marking this PR as a draft. I'll have to think about other approaches we can take to solve this

Issue #7242, opened 2024-05-28
Storybook - RangeInput/Bounds example


jcfilben:
### Expected Behavior
The '+' and '-' buttons on the RangeInput/Bounds have some odd behavior, sometimes the value jumps to unexpected places.

### Actual Behavior
The '+' button should always just increase the value, if the value is at the max it should be disabled. The '-' button should always just decrease the value, if the value is at the minimum the button should be disabled.

### URL, screen shot, or Codepen exhibiting the issue

https://github.com/grommet/grommet/assets/54560994/8ed68fd6-26fc-44de-9867-37f221ced801

### Steps to Reproduce

https://storybook.grommet.io/?path=/story/input-rangeinput-custom-themed-bounds--bounds&globals=theme:base

PR #7244, opened 2024-05-29
fix: Storybook - RangeInput/Bounds example

abdulbasithqb:
#### What does this PR do?
fix: Storybook - RangeInput/Bounds example

Storybook - RangeInput/Bounds example

Manually clicking the + and - to ensure the expected behavior.

go to Storybook - RangeInput/Bounds example then click  + and - and validate it

fixes: #7242

https://github.com/grommet/grommet/assets/134603758/21d4e066-e40c-43fa-a9b2-802e90eb4b46

No


jcfilben (commented):
Thanks for working on this!

One small comment, as soon as the value hits the min or max the button should become disabled. In this PR the button only becomes disabled when we hit the min/max and then try to click the button again.

One potential solution is instead of the `setIsSubtractDisabled(true)` or `setIsAddDisabled(true)` being called in the onClick you could use a `useEffect` that gets called when the value changes and then check if the value is equal to the max/min and if it is set the button to be disabled.

abdulbasithqb:
> Thanks for working on this!
> 
> One small comment, as soon as the value hits the min or max the button should become disabled. In this PR the button only becomes disabled when we hit the min/max and then try to click the button again.
> 
> One potential solution is instead of the `setIsSubtractDisabled(true)` or `setIsAddDisabled(true)` being called in the onClick you could use a `useEffect` that gets called when the value changes and then check if the value is equal to the max/min and if it is set the button to be disabled.

Done, please check

abdulbasithqb:
> [ ](/abdulbasithqb)
> ` [fix: lint fix](/grommet/grommet/pull/7244/commits/c6f14e8214ddf3263948a4de0783bcb174f50bd2)`
Fixed

jcfilben (approved):

halocline (approved):
@abdulbasithqb, thank you for your contribution!

PR #7252, opened 2024-06-12
Remove isRequired from propTypes

jcfilben:
#### What does this PR do?
Remove `isRequired` from propTypes as part of the react 18.3.0 upgrade. See https://github.com/grommet/grommet/pull/7224#discussion_r1624433552 for more details.

There are some places that don't have default props so I kept the `isRequired` because we will not run into the situation mentioned in https://github.com/grommet/grommet/pull/7224#discussion_r1624433552 in these cases.

no
yes
backwards compatible


src/js/components/RoutedButton/RoutedButton.js
```diff
@@ -5,7 +5,7 @@ import { Button } from '../Button';
 
 class RoutedButton extends Component {
   static contextTypes = {
-    router: PropTypes.shape({}).isRequired,
```
britt6612, line 8: there is also some in the tests for RoutedButton do we need to take out?

src/js/components/RoutedAnchor/RoutedAnchor.js
```diff
@@ -5,7 +5,7 @@ import { Anchor } from '../Anchor';
 
 class RoutedAnchor extends Component {
   static contextTypes = {
-    router: PropTypes.shape({}).isRequired,
```
britt6612, line 8: there is also some in the tests for RoutedAnchor do we need to take out?

britt6612:
Do we need to remove `isRequired` from stories?

src/js/components/RoutedButton/RoutedButton.js
jcfilben: I didn't remove `isRequired` from places where it would not cause backwards compatibility issues for users

jcfilben:
> Do we need to remove isRequired from stories?

Keeping it won't cause regressions so I didn't remove it from stories

britt6612 (approved):

taysea (approved):
Looks good!

britt6612 (approved):

PR #7253, opened 2024-06-14
Refactor routed components for react 19 compatibility

jcfilben:
#### What does this PR do?
Refactor RoutedAnchor and RoutedButton to remove `contextTypes` since it will not be supported in react 19. See
https://react.dev/blog/2024/04/25/react-19-upgrade-guide#removed-removing-legacy-context for more details.

Existing jest tests and storybook

no
no
backwards compatible


src/js/components/RoutedButton/RouterContext.js
```diff
@@ -0,0 +1,3 @@
+import React from 'react';
+
+export const RouterContext = React.createContext();
```
britt6612, line 3: do we need both these files? can we just have one `RouterContext` under our context folder?

jcfilben: Yeah I was considering having one RouterContext within the context folder, the reason I ended up creating two RouterContexts and putting them within the RoutedButton and RoutedAnchor was just to keep them a bit more contained and tied to those specific components. Since RoutedButton and RoutedAnchor are deprecated I wanted to keep the RouterContext relatively contained and nearby to those components so that it doesn't have as much visibility as something in the context folder might have. Just a soft opinion though, I'm good either way though if we want to move it to the context folder.

britt6612: gotcha no that makes sense that way if we get rid of those files I understand your thinking..

britt6612 (approved):

britt6612:
opps one small comment shouldnt this point to our React branch?

PR #7255, opened 2024-06-17
defaultVisible prop

jpranays:
#### What does this PR do?
Optional defaultVisible prop on Tip component

Tip component folder

Tested e2e

Inside e2e folder, import Tip component in App.js and pass  **defaultVisible** prop

yes for an additional prop entry

backwards compatible


src/js/components/Tip/index.d.ts
```diff
@@ -4,6 +4,7 @@ import { DropType } from '../Drop';
 export interface TipProps {
   children?: React.ReactNode;
   content?: React.ReactNode;
+  defaultVisible?: boolean;
```
britt6612, line 7: wondering if this is the best naming let us think on this!

britt6612:
@jpranays anyway you can add a test for this ?

britt6612:
@jpranays do you also have a use case for this desired behavior?

jpranays:
> @jpranays anyway you can add a test for this ?
sure

jpranays:
Any update on this ?

jcfilben:
Hey @jpranays thanks for working on this and sharing your ideas with us. A few thoughts:

You mentioned that part of the desire for adding this is to improve accessibility, usability, and overall experience. However,  having the tooltip visible by default seems to go against the W3C pattern for how a tooltip should behave:
- https://w3c.github.io/aria/#tooltip
- https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/

Currently our Tip component is accessible via keyboard focus and mouse hover which aligns with the accessibility guidance we have seen.

You also mentioned that having the tooltip visible by default will enable 'immediate access to essential information, thereby making the interface more intuitive and inclusive'. I think if information is essential it should not be displayed in a tooltip and a different component should be used so that the information is always visible.

jpranays:
Thank you for your thoughtful feedback and your commitment to accessibility and adherence to the W3C guidelines for tooltips.

The current implementation of the Tip component aligns well with these guidelines, ensuring both accessibility and user-friendliness. However, I don’t believe that having a defaultVisible prop would violate W3C guidelines. On the contrary, it would enhance the component’s accessibility. Additionally, other major libraries include a defaultVisible or defaultOpen prop, further extending the component’s accessibility.

jcfilben:
> Additionally, other major libraries include a defaultVisible or defaultOpen prop, further extending the component’s accessibility.

Could you share what other libraries are supporting this? I think that would be helpful for us to looks at as we evaluate this enhancement

jpranays:
1. https://react-tooltip.com/docs/options#props
2. https://2x.ant.design/components/tooltip/
3. https://mui.com/material-ui/api/tooltip/
4. https://v1.chakra-ui.com/docs/components/overlay/tooltip#props
5. https://github.com/mohsinulhaq/react-popper-tooltip/blob/v3/README.md#defaulttooltipshown

jcfilben:
Thanks for sharing, we will take a look at these

jcfilben:
The Grommet team is planning to discuss this a bit more on Thursday, I'll keep you in the loop with any updates and post them here.

jpranays:
Thanks for letting me know.

jcfilben:
Hey @jpranays after talking through the grommet team is supportive of this enhancement. Thanks again for providing links and detailed responses to our questions.

jcfilben (commented):
Can you add `defaultVisible` to the `Tip/propTypes.js` file?

jpranays:
That’s great to hear! 
Welcome!!

src/js/components/Tip/Tip.js
```diff
@@ -12,78 +13,84 @@ import { Drop } from '../Drop';
...
-        onMouseLeave={() => setTooltipOver(false)}
-      >
-        {plain ? content : <Box {...theme.tip.content}>{content}</Box>}
-      </Drop>
-    ),
-  ];
-});
+    useEffect(() => {
+      setOver(defaultVisible);
+    }, [defaultVisible]);
```
jcfilben, line 74: Instead of the useEffect could we just have the initial state of `over` set to `defaultVisible`?
For example:
```
const [over, setOver] = useState(defaultVisible);
```

jpranays: ya initially, I tried that approach, but in the Drop component, on the first render, it returned from line 82 because the dropContainer was empty. Therefore, I needed to wait until the child component was mounted, which is why useEffect was used.

image

jcfilben: Gotcha, thanks for the explanation

jcfilben (approved):
Looks good!

britt6612 (approved):

Issue #7240, opened 2024-05-23
Dependencies - upgrade to react 18.3.0


jcfilben:
We temporarily froze react at v18.2.0 (see https://github.com/grommet/grommet/pull/7227) because there are a number of cleanups we need to do to upgrade to v18.3.0 without getting warnings.

PR #7249, opened 2024-06-06
Remove grommet wrapper

britt6612:
#### What does this PR do?
While looking at upgrading React and the changes need we a far as defaultProps in components should be replaced with default parameters.

Regarding this there are components that never had `deafultProps` For example Pagination is one of many components that will throw an error when it is used alone. Although we technically support using Grommet components outside of the Grommet wrapper this isn’t something we check or test for. All our storybook stories and the large majority of our jest tests have the Grommet wrapper. Another reason we might have missed this is if a component is wrapped in another component that has theme props then it will get theme props. 

This PR is adding tests that will render components outside of the Grommet wrapper.
#### Where should the reviewer start?
all tests 
#### Do Jest tests follow these best practices?

- [ ] `screen` is used for querying.
- [ ] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [ ] `asFragment()` is used for snapshot testing.

#### Any background context you want to provide?
see above 
#### Do the grommet docs need to be updated?
no
#### Should this PR be mentioned in the release notes?
no

PR #7257, opened 2024-06-18
React 18.3.0 upgrade

jcfilben:
#### What does this PR do?
Upgrade to react 18.3.0

This PR includes the work from a few other PRs (see their descriptions for more details):
Jest tests for components outside of Grommet wrapper - https://github.com/grommet/grommet/pull/7249
Remove defaultProps - https://github.com/grommet/grommet/pull/7224
Remove isRequired - https://github.com/grommet/grommet/pull/7252
Refactor routed components - https://github.com/grommet/grommet/pull/7253

Closes https://github.com/grommet/grommet/issues/7240
Closes https://github.com/grommet/grommet/issues/6741

no
yes
backwards compatible


src/js/default-props.js
```diff
@@ -8,3 +8,11 @@ export const defaultProps = {
 export const extendDefaultTheme = theme => {
   defaultProps.theme = deepMerge(base, theme);
 };
+
+/*
+  Pass `theme` for component which can be located outside of theme context.
+  To be used as argument for `attrs` method from `styled-components`.
+ */
+export const enhancePropsWithTheme = props => ({
```
britt6612, line 16: just flagging if we want to chat here about naming

jcfilben: Naming suggestions from grommet checkpoint:
* withTheme
* baseGrommetProps
* baseProps
* grommetProps
* baseGrommet

britt6612: went with `withTheme`

britt6612 (approved):

src/js/components/Menu/Menu.js
```diff
@@ -57,6 +58,8 @@ To make a selection:
 - Space is pressed.
 */
 
+const defaultMenuItemsValue = [];
```
halocline, line 61: Nit. Doesn't need to be addressed, but noticing we are deviating a bit from concise variable names where `defaultMenuItemsValue` could be `defaultItems`. Perhaps that namespace is elsewhere in the file that I'm not seeing, but we are already in the Menu file for context and value maybe extraneous.

Also caught my eye earlier in Distribution, but didn't comment there. Just noticed are style seemed to be drifting a bit.

halocline (commented):
Looks good. I had to put on my "lawyer hat" to find anything to comment on. Real small and don't necessarily need addressing. Just caught my eye while reviewing.

src/js/components/Anchor/__tests__/Anchor-test.tsx
```diff
@@ -35,6 +35,11 @@ describe('Anchor', () => {
     expect(container).toMatchSnapshot();
   });
 
+  test('renders anchor without grommet wrapper', () => {
+    const { container } = render(<Anchor />);
+    expect(container).toMatchSnapshot();
```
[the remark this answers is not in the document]
jcfilben: Good question, for a while we were using `asFragment` instead of `container` in our new snapshot tests based on a suggestion from a Kent C Dodds article on testing. The guidance has changed since then and the suggestion is to not use container if you are trying to query for elements in the dom. So we have been just using `asFragment` or `container.firstChild` for tests based on what the testing file is already using.

If you take a snapshot of the container your snapshot will include an extra div wrapper. If you take a snapshot of `container.firstChild` you won't get the div and will get the root element of your rendered element. I think generally it is a little cleaner to use `container.firstChild`

Additional info & links:

Container:
https://testing-library.com/docs/react-testing-library/api/#container-1

asFragment:
https://testing-library.com/docs/react-testing-library/api/#asfragment

Testing articles:
https://kentcdodds.com/blog/common-mistakes-with-react-testing-library

https://kentcdodds.com/blog/effective-snapshot-testing

src/js/components/Menu/Menu.js
jcfilben: Good note, I'll change this to be more concise

halocline (approved):
Thank you, @jcfilben !

Issue #7266, opened 2024-07-01
List onOrder - Hide index of items


stcsaba:
### Expected Behavior

The List component should be able to hide the index of the items if the onOrder function is given.

### Actual Behavior

On each line there is a row number (index + 1 value) if the onOrder function is given.

### URL, screen shot, or Codepen exhibiting the issue

![image](https://github.com/grommet/grommet/assets/21088623/478f49ca-cef3-4a3b-9336-f39291ececdd)

```
                  <StyledItem
                    key={key}
                    tag="li"
                    background={adjustedBackground}
                    border={adjustedBorder}
                    isDisabled={isDisabled}
                    flex={false}
                    pad={pad || theme.list.item.pad}
                    {...defaultItemProps}
                    {...boxProps}
                    {...clickProps}
                    {...orderProps}
                    {...itemAriaProps}
                  >
                    {onOrder && <Text color={pinnedColor}>{index + 1}</Text>} // this line should be optional
                    {content}
                    {displayPinned}
                    {orderControls}
                  </StyledItem>
```

PR #7267, opened 2024-07-02
enhancement: List onOrder - Hide index of items

abdulbasithqb:
#### What does this PR do?
enhancement: List onOrder - Hide index of items

List Grommet Component

manual testing and jest

Go to List component onOrder function new props added 

fixes: #7266


jcfilben (commented):
Can you add a jest test for this?
Also we would need to add this prop to the index.d.ts file and propTypes file for List

abdulbasithqb:
> Can you add a jest test for this? Also we would need to add this prop to the index.d.ts file and propTypes file for List

sure will do it

abdulbasithqb:
> Can you add a jest test for this? Also we would need to add this prop to the index.d.ts file and propTypes file for List

Done, could you please check it out

src/js/components/List/List.js
```diff
@@ -148,6 +148,7 @@ const List = React.forwardRef(
       onKeyDown,
       onMore,
       onOrder,
+      hideIndex = false,
```
jcfilben, line 151: @grommet/devs any thoughts on naming for this prop?

britt6612: showIndex,
displayIndex
displayIndexNumber
isVisable 


just some ideas we can think more but I like this functionality

jcfilben: I like `showIndex` and then just have it default to true. There are other props in grommet where we use the word `show` so this might help with consistency in the language we use.

jcfilben (commented):
Almost there @abdulbasithqb! Just a few more things...

- See the conversation above about the prop naming
- Looks like we are just missing the prop definition in the List/\propTypes.js file.
- Could you sign your commits so the dco check passes? https://github.com/grommet/grommet/pull/7267/checks?check_run_id=27145859579

src/js/components/List/List.js
britt6612: sounds good ☝️

abdulbasithqb:
> Almost there @abdulbasithqb! Just a few more things...
> 
> * See the conversation above about the prop naming
> * Looks like we are just missing the prop definition in the List/\propTypes.js file.
> * Could you sign your commits so the dco check passes? https://github.com/grommet/grommet/pull/7267/checks?check_run_id=27145859579

closed this PR and created a new pr with sign-off added : https://github.com/grommet/grommet/pull/7267#pullrequestreview-2169809312

Issue #7278, opened 2024-07-25
DataTableColumns - add support for managing initial column visibility


jcfilben:
### Expected Behavior
I should be able to set a column to initially not be visible in DataTableColumns.

### Actual Behavior
By default all the columns are visible unless toggled off by the user.

We could add a field to the DataColumns options prop to manage initial visibility.
Requested by Amanda V. in office hours 7/25/24

jcfilben:
Resolved by #7281

PR #7281, opened 2024-08-02
Add columns to Data View.

coltonw:
#### What does this PR do?

Add columns as an option to data views, allowing for having a default set of columns on or off through the defaultView Data prop.

index.d.ts

Manual testing, storybook testing, and automated tests.

Try a view with the new property.

https://github.com/grommet/grommet/issues/7278

Probably

Yes probably.

Backwards compatible.


coltonw:
Just to be clear, the work here was basically to do nothing. All form data, including columns, is already controllable by the view, so all I had to do was fix some types and add examples and testing.

jcfilben:
Hey @coltonw thank you for working on this! I chatted with the rest of the grommet team about this PR, here is some of the feedback:

- Adding `columns` to the view in the Data component might be too specific to the DataTable use case because `columns` only applies to DataTable views and the Data component not only supports DataTable but also List, Cards, and other data driven components.

- Regardless of whether we support `columns` in `view` we should support having a way to control the initial visibility of the column from the `options` prop of the DataTableColumns component. This allows someone to manage the visibility without having to enter into a controlled state of the Data component via `view` and `onView`.

Can we explore an approach where we don't add `columns` to `view` (we might consider this at a later point in time), and instead handle managing the initial visibility of the columns within the DataTableColumns `options` property? For example something like this:
```
<DataTableColumns 
  options={
    [
      { label: 'Name', property: 'name' },
      { label: 'Age', property: 'age', defaultVisible: false },
    ]
  }
/>
```

jcfilben:
I realized that the data view is already passing `view.columns` through `onView`, so we might reconsider the above comment. I'll keep you posted

jcfilben:
We are going to move forward with this approach

src/js/components/Data/Data.js
```diff
@@ -128,6 +128,7 @@ export const Data = ({
       // longer reflecting the view
       delete nextView.name;
       delete nextView.view;
+      delete nextView.columns;
```
britt6612, line 131:
```suggestion

```

britt6612: I don't beleive we need this..
Otherwise looks good

taysea: I would agree with removing this. The "Clear filters" button should be scoped to clearing out filters. That said, I wouldn't expect that clearing my filters would also reset the visibility of columns.

britt6612 (approved):

taysea (approved):
Tested locally as well. This behavior makes:
1. Select "view" with predefined filters and columns
2. Change visible columns --> view not longer in data view
3. "Clear filters" --> selected columns are still reflected rather than being cleared out

https://github.com/user-attachments/assets/b9e91de2-e4a8-4f65-b9d4-d4d2774767fb



```
import React, { useEffect, useState } from 'react';

import {
  Box,
  DataTable,
  Data,
  Grid,
  Pagination,
  Text,
  Tip,
  DataSearch,
  DataFilters,
  DataView,
  DataTableColumns,
  Toolbar,
} from 'grommet';

import { StatusCritical } from 'grommet-icons';

const buildQuery = (view) => {
  const query = {};
  const properties = view?.properties || [];
  Object.keys(properties).forEach((property) => {
    switch (property) {
      case 'success':
        if (properties.success.length === 1) {
          query[property] = properties.success[0] === 'Successful';
        }
        break;
      case 'rocket':
        query.rocket = {
          $in: properties.rocket,
        };
        break;
      default:
        query[property] = properties[property];
    }
  });
  if (view?.search) query.$text = { $search: view.search };
  return query;
};

const fetchLaunches = async (view) => {
  const query = buildQuery(view);
  const sort = {
    [view?.sort?.property || 'name']: view?.sort?.direction || 'asc',
  };

  const body = {
    options: {
      populate: [
        {
          path: 'rocket',
          select: { name: 1 },
        },
      ],
      sort,
      select: ['name', 'success', 'failures'],
      limit: view?.step || 10,
      page: view?.page || 1,
    },
    query,
  };
  return fetch('https://api.spacexdata.com/v4/launches/query', {
    method: 'POST',
    headers: {
      'Content-Type': 'application/json',
    },
    body: JSON.stringify(body),
  }).then((response) => response.json());
};

const fetchRockets = async () => {
  const body = {
    options: {
      sort: { name: 'asc' },
      select: ['name', 'id'],
    },
  };
  return fetch('https://api.spacexdata.com/v4/rockets/query', {
    method: 'POST',
    headers: {
      'Content-Type': 'application/json',
    },
    body: JSON.stringify(body),
  }).then((response) => response.json());
};

const columns = [
  {
    property: 'name',
    header: 'Name',
    size: 'small',
    primary: true,
  },
  {
    property: 'rocket.name',
    header: 'Rocket',
    size: 'xsmall',
    sortable: false,
  },
  {
    property: 'success',
    header: 'Success',
    size: 'xsmall',
    align: 'center',
    sortable: false,
    render: (datum) => {
      if (datum.success === false) {
        const content = (
          <Box width={{ max: 'medium' }}>
            {datum.failures?.map(({ reason }) => (
              <Text key={reason}>{reason}</Text>
            ))}
          </Box>
        );
        return (
          <Tip
            plain
            content={content}
            dropProps={{
              round: 'medium',
              pad: 'small',
              background: 'background-back',
            }}
          >
            <Box>
              <StatusCritical color="red" />
            </Box>
          </Tip>
        );
      }
      return undefined;
    },
  },
];

const options = columns.map(({ header, property }) => ({
  property,
  label: header,
}));

const defaultView = {
  search: '',
  sort: { property: 'name', direction: 'asc' },
  step: 10,
};

export const SpaceX = () => {
  const [total, setTotal] = useState(0);
  const [result, setResult] = useState({ data: [] });
  const [rockets, setRockets] = useState([]);
  const [view, setView] = useState(defaultView);

  useEffect(() => {
    fetchRockets().then((response) =>
      setRockets(
        response.docs.map(({ name, id }) => ({ value: id, label: name })),
      ),
    );
  }, []);

  useEffect(() => {
    fetchLaunches(view).then((response) => {
      setResult({
        data: response.docs,
        filteredTotal: response.totalDocs,
        page: response.page,
      });
      // The REST API doesn't return the unfiltered total in responses.
      // Since the first request likely has no filtering, we'll likely use
      // response.totalDocs the first time and prevTotal thereafter.
      setTotal((prevTotal) => Math.max(prevTotal, response.totalDocs));
    });
  }, [view]);

  return (
    // <Grommet theme={...}>
    <Grid
      flex={false}
      pad="large"
      columns={[['small', 'large']]}
      justifyContent="center"
    >
      <Data
        properties={{
          rocket: { label: 'Rocket', options: rockets },
          success: { label: 'Success', options: ['Successful', 'Failed'] },
        }}
        data={result.data}
        total={total}
        filteredTotal={result.filteredTotal}
        defaultView={defaultView}
        views={[
          {
            name: 'My view',
            columns: ['name', 'rocket.name'],
            properties: {
              success: ['Successful'],
            },
          },
        ]}
        view={view}
        onView={setView}
      >
        <Toolbar>
          <DataSearch />
          <DataFilters layer />
          <DataView />
          <DataTableColumns options={options} drop />
        </Toolbar>
        <DataTable columns={columns} sortable />
        {result.filteredTotal > view.step && (
          <Pagination summary border="top" pad={{ vertical: 'xsmall' }} />
        )}
      </Data>
    </Grid>
    // </Grommet>
  );
};

SpaceX.storyName = 'SpaceX';

SpaceX.args = {
  full: true,
};

export default {
  title: 'Data/Data/SpaceX',
};
```

MikeKingdom (approved):

jcfilben (approved):
Looks good!

PR #7290, opened 2024-08-08
Upgrade styled components

jcfilben:
#### What does this PR do?
Upgrade to latest styled components version

### DIFF
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
     "regenerator-runtime": "^0.14.1",
     "simple-git": "^3.25.0",
     "storybook": "7.0.20",
-    "styled-components": "^6.1.11",
+    "styled-components": "^6.1.12",
     "tarball-extract": "^0.0.6",
     "testcafe": "3.3.0",
     "testcafe-react-selectors": "^5.0.3",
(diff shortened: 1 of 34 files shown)


britt6612 (dismissed):
all good on my end!

britt6612:
Im getting the following:

● DateInput › format with date bounds

    thrown: "Warning: An update to %s inside a test was not wrapped in act(...).

    When testing, code that causes React state updates should be wrapped into act(...):

    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */

    This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act%s"

      at _getError (node_modules/jest-circus/build/utils.js:432:18)
          at Array.map (<anonymous>)

  ● DateInput › format with date bounds

    Uncaught 'Warning: An update to %s inside a test was not wrapped in act(...).\n' +
      '\n' +
      'When testing, code that causes React state updates should be wrapped into act(...):\n' +
      '\n' +
      'act(() => {\n' +
      '  /* fire events that update state */\n' +
      '});\n' +
      '/* assert on the output */\n' +
      '\n' +
      "This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act%s"

britt6612 (approved):

taysea (approved):
Tested on node 18 & 20. Looks good

taysea:
The chromatic change is one that's been a bit flaky over recent dependency upgrades but from what i remember had to do with what chrome version chromatic is pulling in?

I'm okay with the change.

jcfilben:
> The chromatic change is one that's been a bit flaky over recent dependency upgrades but from what i remember had to do with what chrome version chromatic is pulling in?

Yeah we've had the same issue with this snapshot in the past. I'm not remembering 100% but I know we looked into it and the chrome version problem sounds familiar

Issue #7274, opened 2024-07-17
Select multiple do not honour the order of selection


sabarnix:
https://storybook.grommet.io/?path=/story/input-selectmultiple-inside-a-formfield--inside-form-field
in the example if your order of selection is "Grape", "Banana" and then "Kiwi", the dropdown selection will show "Banana", "Grape" , "Kiwi". It orders the selection according to list.

It's not a big deal in the example as checkboxes are shown besides the options. Image if we don't show the checkbox and hide the selected value from the list. In that case the order of selection matters. There could be use cases where order of selection is important.

### Is this something you or your team is interested in contributing a pull request towards?

yes
### Expected Behavior

The selected value should be shown based on the order of selection

### Actual Behavior

The selection is ordered based on the list.

### URL, screen shot, or Codepen exhibiting the issue

https://storybook.grommet.io/?path=/story/input-selectmultiple-inside-a-formfield--inside-form-field

### Your Environment

- Grommet version: latest
- Browser Name and version: chrome
- Operating System and version (desktop or mobile): desktop mbp

jcfilben:
Hi @sabarnix thanks for sharing your proposal. I'm not sure if I am understanding the use case correctly. 

> It's not a big deal in the example as checkboxes are shown besides the options. Image if we don't show the checkbox and hide the selected value from the list. In that case the order of selection matters. There could be use cases where order of selection is important.

It seems like there could be usability and accessibility issues with a design like this. I'm also not understanding why the selection order is important in the scenario you described

sabarnix:
![image](https://github.com/user-attachments/assets/08c62c5d-7fcd-4148-9122-0d7013ca1f4c)
Our select roughly looks like this. And in our case the order matters. 

One of the popular multiselect component https://react-select.com/home also retains the order. I think that should be the expected behaviour. The way react-select uses multiselect is very common. I don't see any usability or accessibility issue with it.

jcfilben:
Thanks for sharing the image, that is helpful to visualize. I'm not sure that it is currently possible to achieve this type of design using the grommet multi-select component, regardless of wether the order of selection is retained or not. My concerns with accessibility stemmed from this comment:
>  Image if we don't show the checkbox and hide the selected value from the list.
In the image above it looks like you have two lists, one of selected items and one of unselected items. When someone makes a selection and the item moves from the unselected list to the selected list does the keyboard focus move with it? How does the screenreader provide context to a user that the item has moved and that item has been selected but moved to a new list

[...]

PR #7291, opened 2024-08-09
fix(Select): Preserve order of selection #7274

sabarnix:
#### What does this PR do?
The select component of grommet do not preserve the order of selection. Even this behaviour is not consistent. Order for first 2 selection are maintained but as you select the next one, the earlier selection gets sorted. 

It is expected the dropdown will preserve the order of selection.

Run the storybook for renderChildren the issue is reproducible there.

Manual testing and existing tests

src/js/components/Select/stories/ChildrenRender.js
open this storybook and select values from bottom to up

#7274

No

Not sure

It's backward compatible


jcfilben (commented):
Modifying the order of the options seems like it may cause backwards compatibility issues for some users. Since `onChange` is passed both the value and the option that was selected it seems like organizing the value by order of selection could be handled by the user within the onChange function.

sabarnix:
There is a bug in the existing behavior. The order is not sorted always it only happens after a new value is selected. 
let's say the options are 1, 2, 3, 4 if you select 3 then 2. The order will be preserved 3, 2. But if you select 1 after this then the selection would become 2, 3, 1.

jcfilben:
> There is a bug in the existing behavior. The order is not sorted always it only happens after a new value is selected.
> let's say the options are 1, 2, 3, 4 if you select 3 then 2. The order will be preserved 3, 2. But if you select 1 after this then the selection would become 2, 3, 1.

Could you clarify which example you are looking at or codesandbox?

sabarnix:
Screen Recording 2024-08-20 at 12 52 30 PM (1)

codesandbox 
https://codesandbox.io/p/live/364c3ca6-c33f-4804-867a-e2bc537f6086

The code is same as the select storybook named children.
I have only remove the sort on values

jcfilben:
I wasn't able to see your codesandbox (looks like it isn't shared publicly) but I was able to reproduce what you were seeing by removing the sort from the Select/Children story. I see what you mean about the behavior being inconsistent

sabarnix:
I have updated the link.
I think the solution for this is to preserve the order of selection. Let me know if you think otherwise. Either way if we try to fix this issue it will break the existing behaviour.

jcfilben:
> I have updated the link.
I think the solution for this is to preserve the order of selection. Let me know if you think otherwise. Either way if we try to fix this issue it will break the existing behaviour.

Thanks for the updated link. I think the existing behavior is odd enough it could be considered a bug. I'm good with moving forward with this we will just want to make sure we announce it in the release notes. @grommet/devs let me know if anyone has concerns.

jcfilben (approved):
Looks good to me

src/js/components/Select/stories/ChildrenRender.js
```diff
@@ -85,7 +85,7 @@ export const Children = () => {
         selected={selected}
```
britt6612, line 85: @sabarnix would you be open to changing the story a bit..

When this story was made since then `selected` is actually going to be deprecated. I figured since we are in the files if you can make a few quick adjustments. 
 https://v2.grommet.io/select#selected

remove the `selected={selected}` then use `value={selected}` then with what we have in our `value` now can you move that to be 
```
        valueLabel={
          <Box wrap direction="row" width="small">
            {selected && selected.length ? (
              selected.map((index) => renderSeason(allSeasons[index]))
            ) : (
              <Box
                pad={{ vertical: 'xsmall', horizontal: 'small' }}
                margin="xsmall"
              >
                Select Season
              </Box>
            )}
          </Box>
        }
```

then update the `onChange` I believe this will work the same just more aligned to what our current guidance is.

britt6612:
@sabarnix if you also mind adding a test that would be great! Then I think this will be good to merge!

src/js/components/Select/stories/ChildrenRender.js
sabarnix: done

sabarnix:
I have added the fix to SelectMultiple too and added test cases for both the components

sabarnix:
Thanks for pointing out @britt6612. I had missed updating the remove function. I have pushed the fix. It should work now.

britt6612 (approved):
LGTM thank you @sabarnix

sabarnix:
Hi, there has been no activity in last 3 week. Is there anything I can help with?

jcfilben:
Hi @sabarnix apologies for the long wait on the review, I think this got lost in my notifications. I am confused on the select multiple behavior, it looks like value that is passed to onChange is the same order as before, maybe I'm missing something?

sabarnix:
> Screen Recording 2024-08-20 at 12 52 30 PM (1) [ Screen Recording 2024-08-20 at 12 52 30 PM (1) ](https://private-user-images.githubusercontent.com/13603907/359379928-26b7be63-2f4e-4d30-b753-dc21719255bc.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjcxNTA1NTEsIm5iZiI6MTcyNzE1MDI1MSwicGF0aCI6Ii8xMzYwMzkwNy8zNTkzNzk5MjgtMjZiN2JlNjMtMmY0ZS00ZDMwLWI3NTMtZGMyMTcxOTI1NWJjLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA5MjQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwOTI0VDAzNTczMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTI1ODc5ODU2OWNlNDRkNDM4YmFkOGMxMDZmNDY4NDQ2MTkyMzY3MmJhYmZhOTkzYjNhOGI3NTY5ZTFjYzg5MzMmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.R7WiBDulhRismxWkIrIUoEA1EHty63gf11VszCJTpU4) [ ](https://private-user-images.githubusercontent.com/13603907/359379928-26b7be63-2f4e-4d30-b753-dc21719255bc.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjcxNTA1NTEsIm5iZiI6MTcyNzE1MDI1MSwicGF0aCI6Ii8xMzYwMzkwNy8zNTkzNzk5MjgtMjZiN2JlNjMtMmY0ZS00ZDMwLWI3NTMtZGMyMTcxOTI1NWJjLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA5MjQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwOTI0VDAzNTczMVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTI1ODc5ODU2OWNlNDRkNDM4YmFkOGMxMDZmNDY4NDQ2MTkyMzY3MmJhYmZhOTkzYjNhOGI3NTY5ZTFjYzg5MzMmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.R7WiBDulhRismxWkIrIUoEA1EHty63gf11VszCJTpU4)
> 
> codesandbox https://codesandbox.io/p/devbox/adoring-dust-55fxs8
> 
> The code is same as the select storybook named children. I have only remove the sort on values
> 
> edit: updated with working link

Hi @jcfilben, refer to this screenshot and codesandbox

sabarnix:
There is not existing storybook for SelectMultiple to demonstrate the issue. Since both has almost same code the issue is there in SelectMultiple too.

sabarnix:
Also is there somewhere documented why SelectMultiple is needed, because Select also supports multiple selection. The code duplication is difficult to maintain.

jcfilben:
> @jcfilben There is no existing storybook for SelectMultiple to demonstrate the issue. Since both has almost same code the issue is there in SelectMultiple too.

Oops I didn't have the latest version pulled down, but pulled down the changes and then tested with console logs to show the order of options passed to onChange in SelectMultiple and looks good.

> Also is there somewhere documented why SelectMultiple is needed, because Select also supports multiple selection. The code duplication is difficult to maintain.

The Select component does support multiple selection but the ui isn't ideal for this. We wanted a better experience for Select with multiple selection, but we couldn't make all the changes we wanted to on the Select component without breaking backwards compatibility. So we created a new SelectMultiple component.

src/js/components/SelectMultiple/__tests__/SelectMultiple-test.tsx
```diff
@@ -159,6 +159,38 @@ describe('SelectMultiple', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
+  test('Value should be of same order as the order of selection', async () => {
```
jcfilben, line 162: Can we swap the order of selections in this test so that we select something that would end up in a different order than alphabetical? For example Apple, Orange, then Banana

sabarnix:
> The Select component does support multiple selection but the ui isn't ideal for this. We wanted a better experience for Select with multiple selection, but we couldn't make all the changes we wanted to on the Select component without breaking backwards compatibility. So we created a new SelectMultiple component.

I don't think it's good idea to maintain two component with similar code. Maybe the common code can be extracted as a hook and used in the components.

britt6612:
> > The Select component does support multiple selection but the ui isn't ideal for this. We wanted a better experience for Select with multiple selection, but we couldn't make all the changes we wanted to on the Select component without breaking backwards compatibility. So we created a new SelectMultiple component.
> 
> I don't think it's good idea to maintain two component with similar code. Maybe the common code can be extracted as a hook and used in the components.

I think for this PR it would be okay to just change the test that @jcfilben recommended then at least merge the functionality.

jcfilben (approved):
Looks good!

PR #7295, opened 2024-08-14
Fix for Accordion accessibility issue

ldoliv:
#### What does this PR do?
Fixes the accessibility violation "Ensure landmarks are unique" that can be seen in the storybook accessibility tab.

Any story under the `Accordion` component

Was run before commiting:
`yarn test-update`
`yarn test`

Check the accessibility tab for any `Accordion` component story

https://github.com/grommet/grommet/issues/6124

Before:
issue

After:
resolved

No

No

backwards compatible


src/js/components/AccordionPanel/AccordionPanel.js
```diff
@@ -148,7 +151,7 @@ const AccordionPanel = forwardRef(
             </Box>
           )}
         </Button>
-        <Box role="region" border={contentBorder}>
+        <Box role="region" border={contentBorder} aria-labelledby={uniqueId}>
```
jcfilben, line 154: Instead of assigning an id and setting `aria-labelledby` could we pull `aria-label` out of `..rest` and apply it to this Box? If `aria-label` is not defined we can try to fall back to `label` if it is a string.

```
<Box 
  role="region" 
  border={contentBorder} 
  aria-label={ariaLabel || (typeof label === 'string' ? label : undefined)}
>
```

ldoliv: I'm following the accordion pattern from [ARIA Authoring Practices Guide (APG)](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/) where they use "aria-labelledby" and link it to the Button.
Their example https://www.w3.org/WAI/ARIA/apg/patterns/accordion/examples/accordion/.

To use aria-label on the Box component we would have to guarantee that it's value be unique otherwise the same violation shows.

Duplicate label show as a violation here: [landmark-unique rule](https://dequeuniversity.com/rules/axe/4.9/landmark-unique?application=axeAPI)
image

jcfilben: Thanks for the explanation and linking the APG guidance, this is helpful. Chatting with some of the other grommet devs on this, I'll keep you posted...

ldoliv:
Updated the branch

britt6612 (approved):
LGTM

jcfilben (approved):
Looks good!

PR #7305, opened 2024-09-04
fix trap focus

britt6612:
#### What does this PR do?
This PR is a POC of how we can trapFocus within a `drop`
 focusedContainer
locally storybook

TrapFocus:
In our current focusedContainer.js if 
Hidden is false, trapFocus is true, roots is defined 
Then it goes over all the elements in the roots array and call makeNofeUnfocusable on each in the array. 

makeNodeUnfocusable is given a node and then makes the node and the children unfocusable. 

Sets the aria-hidden to true on the grommet wrapper
Ensures that inherently faceable elements become unfocusable and adding -1 to anything that has a tabIndex.

Using this approach this just makes everything not tab-able inside of the grommet wrapper (besides the things in drop) so when you don’t have anything else to tab it goes back up to the window

POC: 
We know that in our FocusedContainer component the following `div` 

    <div ref={ref} aria-hidden={hidden} {...rest}>
        {children}
      </div>

Will always wrap the children so that div will always be the last one in the roots array so we can access that div then look at the first child in that div (which would be the container of the drop. 

We then get the first and last focusable elements. 

In our tabkeyHandler we ensure that the user can not tab out of the set. For example 
If the user presses Tab and the focus is on the last element, the focus will move back to the first element.

needed for DS site Popover
potentially 
yes


britt6612:
https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/

The document above in how they are trapFocus was what was referenced for this PR.


We are still leaving our current `roots.forEach(makeNodeUnfocusable);` the same we are just adding to this same `useEffect` 
We have an event.listener that will then call our `trapFocusHandler` function. 

It uses our roots to get the correct root in order to `trapFocus`
It then finds all focusableElements within that root, sets the first and last focus.

src/js/components/FocusedContainer.js
```diff
@@ -43,19 +44,48 @@ export const FocusedContainer = ({
...
+    }
+  };
+
   useEffect(() => {
     const timer = setTimeout(() => {
-      if (!hidden && trapFocus && roots && roots[0]) {
+      if (!hidden && trapFocus && nextRoots && nextRoots.length > 0) {
+        // Make all nodes unfocusable except the last one in the list
         roots.forEach(makeNodeUnfocusable);
+        document.addEventListener('keydown', trapFocusHandler, true);
```
MikeKingdom, line 76: `keydown` doesn't make any sense to me. Shouldn't this be `focus`?

src/js/components/FocusedContainer.js
```diff
@@ -43,19 +44,48 @@ export const FocusedContainer = ({
     };
   }, [bodyOverflowStyle, hidden, trapFocus, restrictScroll]);
 
+  const trapFocusHandler = (event) => {
```
MikeKingdom, line 47: This needs to be memoized with useCallback or something like that so it doesn't change on every render

src/js/components/FocusedContainer.js
```diff
@@ -43,19 +44,48 @@ export const FocusedContainer = ({
...
-      if (roots && roots[0]) makeNodeFocusable(roots[roots.length - 1]);
+      document.removeEventListener('keydown', trapFocusHandler, true);
+      // Restore focusability when component is unmounted or updated
       clearTimeout(timer);
+      if (roots && roots.length > 0) {
+        makeNodeFocusable(roots[roots.length - 1]);
+      }
     };
-  }, [hidden, roots, trapFocus]);
+  }, [hidden, nextRoots, trapFocus, roots]);
```
MikeKingdom, line 88: this seems to be missing trapFocusHandler in the dependencies

taysea:
Superseded by https://github.com/grommet/grommet/pull/7324

PR #7308, opened 2024-09-09
add first element focus option to Drop

britt6612:
#### What does this PR do?
This PR adds an option to pass `firstElement` to restricFocus prop
DropContainer.js
locally 
TODO

in order for focus to get put back in the trigger element in which opened drop restrictFocus needs to be set to `true` however this also puts focus on drop. This allows the user to have the restrictFocus true but also focus on the first interactive element
DS site popover
yes
yes
 backwards compatible


src/js/components/DateInput/__tests__/__snapshots__/DateInput-test.tsx.snap
```diff
@@ -8258,37 +8258,57 @@ exports[`DateInput handle focus in FormField 2`] = `
...
+.c3 >polyline,
+.c3 >rect {
+  outline: none;
+  box-shadow: 0 0 2px 2px #6FFFB0;
+}
+
+.c3 ::-moz-focus-inner {
+  border: 0;
+}
+
```
jcfilben: I'm a little confused on this snapshot change

britt6612: this is no longer relevant

src/js/components/Drop/DropContainer.js
```diff
@@ -15,6 +15,30 @@ import { StyledDrop } from './StyledDrop';
 import { OptionsContext } from '../../contexts/OptionsContext';
 import { useThemeValue } from '../../utils/useThemeValue';
 
+// Check if the element.tagName is an input, select or textarea
```
jcfilben, line 18: This comment doesn't seem accurate to what the function is doing

src/js/components/Drop/DropContainer.js
```diff
@@ -15,6 +15,30 @@ import { StyledDrop } from './StyledDrop';
...
+export const isFocusableElements = (element) => {
```
jcfilben, line 19: Do we need to export this? Looks like we are only referencing it in this file

src/js/components/Drop/DropContainer.js
```diff
@@ -343,8 +367,23 @@ const DropContainer = forwardRef(
...
 
+    // if restrictFocus is true place focus on the drop container
+    // if restrictFocus is firstElement put focus on first
+    // interactive element if none fall back to drop container
     useEffect(() => {
-      if (restrictFocus) {
+      if (restrictFocus === true) {
+        dropRef.current.focus();
+      }
+      if (
```
jcfilben, line 377: Should this be an else if?

src/js/components/Drop/__tests__/__snapshots__/Drop-test.tsx.snap
jcfilben, line 1: Not sure that I am following the snapshot changes, could you explain a little what is causing these

src/js/components/Drop/DropContainer.js
britt6612: yes your correct

jcfilben (commented):
Might be worth adding a storybook story for this

src/js/components/Drop/DropContainer.js
britt6612: done

src/js/components/Drop/__tests__/__snapshots__/Drop-test.tsx.snap
britt6612: yeah It is a little strange as soon as I add `restrictFocus` it adds the body so will look more into this

src/js/components/Drop/DropContainer.js
britt6612: updated

src/js/components/Drop/__tests__/__snapshots__/Drop-test.tsx.snap
jcfilben: Maybe in the 'restrict focus === firstElement' test instead of using the TextInput component, just having a component that is isolated to that test?

britt6612: I can do this but just not sure why adding that would effect the current snapshot.

src/js/components/Drop/DropContainer.js
```diff
@@ -343,9 +367,21 @@ const DropContainer = forwardRef(
...
         dropRef.current.focus();
+      } else if (restrictFocus === 'firstElement') {
+        const firstFocusableElement = getFirstFocusableDescendantElement(
+          dropRef.current,
+        );
+        if (firstFocusableElement !== undefined) {
+          firstFocusableElement.focus();
+        } else {
+          dropRef.current.focus();
+        }
```
jcfilben, line 384: Do we need this last else statement? It seems odd that it executes the same code as the first if statement

britt6612: I added it incase someone has `restrictFocus='firstElement'`  but they might not realize they don't have any interactive elements

britt6612: I can take out just put it there to be cautious

britt6612:
```

  // if restrictFocus is true place focus on the drop container
    // if restrictFocus is firstElement put focus on first
    // interactive element if none fall back to drop container

```

Follow what I had commented

jcfilben: Yeah I think we should take it out just to ensure we align to whatever the previous behavior is when restrictFocus is false or undefined.

britt6612: so it will never come into this whole block of code when restrictFocus is false just like before. 

I can take it out though.

jcfilben (approved):
Looks good!

britt6612:
closing in favor of #7322

PR #7313, opened 2024-09-13
Fix DataFilter SelectMultiple when options are object

taysea:
#### What does this PR do?

Fixes an issue when options are passed to SelectMultiple as object in DataFilter.

Locally:

```
import range from 'lodash/range';

import { Box, Data, DataFilters, Text } from 'grommet';
import { DATA } from '../../DataTable/stories/data';

export const Simple = () => {
  const [view, setView] = useState({
    properties: {
      selectMultiple: ['selectmultiple5'],
      checkBoxGroup: ['checkboxgroup2'],
      selectMultipleSimple: ['selectmultiple-simple2'],
    },
  });
  return (
    // <Grommet theme={...}>
    <Box pad="large" fill>
      <Data
        data={[]}
        properties={{
          selectMultiple: {
            label: 'SelectMultiple Name',
            options: range(9).map((_, i) => ({
              value: `selectmultiple${i}`,
              label: `SelectMultiple Name ${i}`,
            })),
          },
          selectMultipleSimple: {
            label: 'SelectMultiple Simple Name',
            options:
              console.log(range(9).map((i) => `selectmultiple-simple${i}`)) ||
              range(9).map((i) => `selectmultiple-simple${i}`),
          },
          checkBoxGroup: {
            label: 'CheckBoxGroup Name',
            options: range(4).map((i) => ({
              value: `checkboxgroup${i}`,
              label: `CheckBoxGroup Name ${i}`,
            })),
          },
        }}
        view={{
          properties: {
            selectMultiple: ['selectmultiple5'],
            checkBoxGroup: ['checkboxgroup2'],
            selectMultipleSimple: ['selectmultiple-simple2'],
          },
        }}
        onView={setView}
      >
        <DataFilters layer />
      </Data>
      <Text>{JSON.stringify(view)}</Text>
    </Box>
    // </Grommet>
  );
};

Simple.args = {
  full: true,
};

export default {
  title: 'Data/DataFilters/Simple',
};

```

https://github.com/user-attachments/assets/bbb6694b-1c57-428f-8588-44917fc5dfac

Bug can be seen here: https://codesandbox.io/p/sandbox/hungry-mountain-wnkyvx?file=%2Fsrc%2FApp.js%3A18%2C36-23%2C4

No.

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/DataFilter/DataFilter.js
+++ b/src/js/components/DataFilter/DataFilter.js
@@ -227,6 +227,8 @@ export const DataFilter = ({
                 : undefined
             }
             onClose={() => setSearchText('')}
+            labelKey="label"
+            valueKey={{ key: 'value', reduce: true }}
           />
         );
       }
--- a/src/js/components/Data/__tests__/Data-test.tsx
+++ b/src/js/components/Data/__tests__/Data-test.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import range from 'lodash/range';
 import '@testing-library/jest-dom';
 import { act, fireEvent, render, screen } from '@testing-library/react';
 import 'jest-styled-components';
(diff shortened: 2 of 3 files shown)


src/js/components/DataFilter/DataFilter.js
```diff
@@ -227,6 +227,8 @@ export const DataFilter = ({
                 : undefined
             }
             onClose={() => setSearchText('')}
+            labelKey="label"
+            valueKey={{ key: 'value', reduce: true }}
```
jcfilben, line 231: Should we only pass these if options is an array of objects?

taysea: Good question -- I had thought about the same, but tested it out with just an array of strings and didn't see any issues with functionality. This is because the logic inside SelectMultiple will check if the option is an object or not.

Let me know if you're comfortable with leaving as is. Otherwise, I could add a check to see if the first option is an object and only pass the props through in that case.

jcfilben: Sounds good, just wanted to verify that the code in SelectMultiple is checking that options is an object before trying to use labelKey or valueKey. I think it is fine to leave as is

jcfilben (approved):
Looks good!

Issue #6997, opened 2023-10-31
Drop position is sometimes incorrect 


britt6612:
### Expected Behavior
https://storybook.grommet.io/?path=/story/visualizations-list-action--action

 if you click on the drop in the list item near the bottom, the drop  should open upwards

### Actual Behavior
The drop open below with a scroll bar in the drop. This previously worked in v2.30.0 

### URL, screen shot, or Codepen exhibiting the issue

https://storybook.grommet.io/?path=/story/visualizations-list-action--action

### Steps to Reproduce
go to this story and open the last one near the bottom of the page.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #7320, opened 2024-09-19
Fix drop positioning

jcfilben:
#### What does this PR do?
Resolves an issue where drop positioning was not correctly flipping when there was not enough room to display in one direction.

See code comments in DropContainer for a more thorough explanation of what is happening. In cases where the drop has a smaller height than the target and the position is either `top: top` or `bottom: bottom` I decided to not try and flip the drop position. We can change this if we want. I just couldn't think of use cases where we would want this behavior.

This PR covers use cases where the drop doesn't have a scroll within it. Drops that do have a scroll within them don't respond well to the screen size adjusting. The work to address the issues with Drop position when they have a scroll will be handled as a separate issue/PR.

Tested with the existing stories (Drop/All not stretched, List/Action, Form/Controlled) as well as this story:
```
import React from 'react';

import { grommet, Box, Button, Grommet, Tip } from 'grommet';

export const Simple = () => (
  // <Grommet theme={...}>
  <Box align="center" justify="center">
    <Box background="pink" height="large">
      spacer
    </Box>
    <Tip
      defaultVisible
      content={
        <Box width="30px" background="green">
          tip
        </Box>
      }
      dropProps={{
        align: { bottom: 'bottom' },
      }}
      plain
    >
      <Button label="action" size="large" />
    </Tip>
    <Box background="pink" height="large">
      spacer
    </Box>
  </Box>
  //</Grommet>
);

Simple.args = {
  full: true,
};

Simple.parameters = {
  chromatic: { disable: true },
};

export default {
  title: 'Controls/Tip/Simple',
};


```

Closes https://github.com/grommet/grommet/issues/6997

no
yes
backwards compatible

### DIFF
--- a/src/js/components/Drop/DropContainer.js
+++ b/src/js/components/Drop/DropContainer.js
@@ -209,59 +209,55 @@ const DropContainer = forwardRef(
             less room in the other direction leave the Drop in its current
             position. */
           if (
-            /* The bottom of the drop is aligned to the top of the target 
-            and the top of the drop is either at the top edge of the viewport
-            or beyond it. If there is room to display the drop below the target
-            change the drop's position so that the top of the drop is aligned 
-            to the bottom of the target. */
             responsive &&
+            // drop is above target
             align.bottom === 'top' &&
+            // drop is overflowing above window
             targetRect.top - containerRect.height <= 0 &&
+            // there is room to display the drop below the target
             targetRect.bottom + containerRect.height < windowHeight
           ) {
+            // top of drop is aligned to bottom of target
             top = targetRect.bottom;
             maxHeight = top;
           } else if (
-            /* The top of the drop is aligned to the top of the target, the 
-            bottom of the drop is either at the bottom edge of the viewport
-            or beyond it, and the height of the drop is larger than the target. 
-            If there is room to display the drop above the target, change the 
-            drop's position so that the bottom of the drop is aligned to the 
-            bottom of the target. */
             responsive &&
+            // top of drop is aligned to top of target
             align.top === 'top' &&
+            // drop is overflowing below window
             targetRect.top + containerRect.height >= windowHeight &&
+            // height of the drop is larger than the target.
             targetRect.top + containerRect.height > targetRect.bottom &&
+            // there is room to display the drop above the target
             targetRect.bottom - containerRect.height > 0
           ) {
+            // bottom of drop is aligned to bottom of target
             bottom = targetRect.bottom;
             maxHeight = top;
           } else if (
-            /* The top of the drop is aligned to the bottom of the target 
-            and the bottom of the drop is either at the bottom edge of the 
-            viewport or beyond it. If there is room to display the drop 
-            above the target change the drop's position so that the bottom 
-            of the drop is aligned to the top of the target. */
             responsive &&
+            // top of drop is aligned to bottom of target
             align.top === 'bottom' &&
+            // drop is overflowing below window
             targetRect.bottom + containerRect.height >= windowHeight &&
+            // there is room to display the drop above the target
             targetRect.top - containerRect.height > 0
           ) {
+            // bottom of drop is aligned to top of target
             bottom = targetRect.top;
             maxHeight = bottom;
           } else if (
-            /* The bottom of the drop is aligned to the bottom of the target, 
-            the top of the drop is either at the top edge of the viewport
-            or beyond it, and the height of the drop is larger than the target. 
-            If there is room to display the drop below the target change the 
-            drop's position so that the top of the drop is aligned  to the top 
-            of the target. */
             responsive &&
+            // bottom of drop is aligned to bottom of target
             align.bottom === 'bottom' &&
+            // drop is overflowing above window
             targetRect.bottom - containerRect.height <= 0 &&
+            // height of the drop is larger than the target.
             targetRect.bottom - containerRect.height > targetRect.top &&
+            // there is room to display the drop below the target
             targetRect.top + containerRect.height > 0
           ) {
+            // top of drop is aligned to top of target
             top = targetRect.top;
             maxHeight = bottom;
           } else if (align.top === 'top') {
--- a/src/js/components/Drop/DropContainer.js
+++ b/src/js/components/Drop/DropContainer.js
@@ -209,22 +209,60 @@ const DropContainer = forwardRef(
             less room in the other direction leave the Drop in its current
             position. */
           if (
+            /* The bottom of the drop is aligned to the top of the target 
+            and the top of the drop is either at the top edge of the viewport
+            or beyond it. If there is room to display the drop below the target
+            change the drop's position so that the top of the drop is aligned 
+            to the bottom of the target. */
             responsive &&
-            ((align.top === 'top' && targetRect.top < 0) ||
-              (align.bottom === 'top' &&
-                targetRect.top - containerRect.height <= 0 &&
-                targetRect.bottom + containerRect.height < windowHeight))
+            align.bottom === 'top' &&
+            targetRect.top - containerRect.height <= 0 &&
+            targetRect.bottom + containerRect.height < windowHeight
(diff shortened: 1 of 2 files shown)


src/js/components/Drop/DropContainer.js
```diff
@@ -209,23 +209,61 @@ const DropContainer = forwardRef(
             less room in the other direction leave the Drop in its current
             position. */
           if (
+            /* The bottom of the drop is aligned to the top of the target
```
britt6612, line 212: I wonder if we can simplify I know you added to help but maybe just saying if  `align.bottom === 'top'` then the top of the drop is at the edge of the viewport or beyond....

src/js/components/Drop/DropContainer.js
```diff
@@ -209,23 +209,61 @@ const DropContainer = forwardRef(
...
+            above the target change the drop's position so that the bottom 
+            of the drop is aligned to the top of the target. */
+            responsive &&
+            align.top === 'bottom' &&
+            targetRect.bottom + containerRect.height >= windowHeight &&
+            targetRect.top - containerRect.height > 0
           ) {
             bottom = targetRect.top;
             maxHeight = bottom;
+          } else if (
```
britt6612, line 252: code looks good though these are the same checks I was thinking when I was looking at this breaking everything apart thank you 👍

src/js/components/Menu/Menu.js
```diff
@@ -439,7 +439,9 @@ const Menu = forwardRef((props, ref) => {
               background={dropBackground || theme.menu.background}
               {...passThemeFlag}
             >
-              {alignControlMirror === 'top' && align.top === 'top'
+              {alignControlMirror === 'top' &&
+              align.bottom !== 'top' &&
```
britt6612, line 443: why was this added?

taysea:
In this scenario, it feels like it would be better for the Drop to stay at the top and have a scroll rather than flip below. Should we consider this kind of case?

Example code:

```

import { Box, Text } from 'grommet';
import { SelectMultiple } from '../SelectMultiple';

const defaultOptions = [
  'Apple',
  'Orange',
  'Banana',
  'Grape',
  'Melon',
  'Strawberry',
  'Kiwi',
  'Mango',
  'Raspberry',
  'Rhubarb',
];

export const Default = () => {
  const [valueMultiple, setValueMultiple] = useState([]);

  return (
    // <Grommet theme={...}>
    <Box fill align="center" justify="end" pad="large" gap="large">
      <Text>SelectMultiple Default</Text>
      <SelectMultiple
        value={valueMultiple}
        placeholder="Select"
        options={options}
        onSearch={(text) => {
          // The line below escapes regular expression special characters:
          // [ \ ^ $ . | ? * + ( )

          // Create the regular expression with modified value which
          // handles escaping special characters. Without escaping special
          // characters, errors will appear in the console
          setOptions(defaultOptions.filter((o) => exp.test(o)));
        }}
        onChange={({ value }) => {
          setValueMultiple(value);
        }}
        // dropHeight="small"
      />
    </Box>
    // </Grommet>
  );
};

Default.parameters = {
  chromatic: { disable: true },
};

Default.args = {
  full: true,
};

export default {
  title: 'Input/SelectMultiple/Default',
};
```

https://github.com/user-attachments/assets/2710ea8f-7d77-46df-8e75-b7b7b8ab92ba

src/js/components/Drop/DropContainer.js
jcfilben: Yeah I can go back and simplify the comments

src/js/components/Menu/Menu.js
jcfilben: To ensure that when `align.bottom !== 'top'` or `align.top !== 'bottom'` we don't add the controlMirror

jcfilben:
> In this scenario, it feels like it would be better for the Drop to stay at the top and have a scroll rather than flip below. Should we consider this kind of case?

Yeah I think this work might fit better in with the exploration on drops without set heights, and how to manage scroll and positioning in those situations. Let me know what you think or if you think it should be in this PR

src/js/components/Drop/DropContainer.js
taysea: Screenshot 2024-09-24 at 3 14 09 PM

@jcfilben @britt6612 Found myself still trying to parse the conditions as I reviewed. Here's a suggestion of a different way to comment -- basically inline with each condition instead of above. Do you find this readable/helpful? Or do you prefer with the comments above each block?

taysea:
> > In this scenario, it feels like it would be better for the Drop to stay at the top and have a scroll rather than flip below. Should we consider this kind of case?
> 
> Yeah I think this work might fit better in with the exploration on drops without set heights, and how to manage scroll and positioning in those situations. @taysea Let me know what you think or if you think it should be in this PR

Thanks for clarifying, yes I'm comfortable handling in a separate PR and keeping this PR scoped to Drops with defined heights/those that do not use InfiniteScroll internally.

src/js/components/Drop/DropContainer.js
```diff
@@ -209,23 +209,61 @@ const DropContainer = forwardRef(
...
+            align.bottom === 'top' &&
+            targetRect.top - containerRect.height <= 0 &&
+            targetRect.bottom + containerRect.height < windowHeight
           ) {
             top = targetRect.bottom;
             maxHeight = top;
           } else if (
+            /* The top of the drop is aligned to the top of the target, the 
+            bottom of the drop is either at the bottom edge of the viewport
+            or beyond it, and the height of the drop is larger than the target.
```
taysea, line 227: I know we discussed it but can you remind me why it matters that the height of the drop is larger than the target?

src/js/components/Drop/DropContainer.js
```diff
@@ -209,23 +209,61 @@ const DropContainer = forwardRef(
             less room in the other direction leave the Drop in its current
             position. */
           if (
+            /* The bottom of the drop is aligned to the top of the target 
+            and the top of the drop is either at the top edge of the viewport
+            or beyond it. If there is room to display the drop below the target
+            change the drop's position so that the top of the drop is aligned 
+            to the bottom of the target. */
             responsive &&
```
taysea, line 217: The top few conditions check for `responsive &&` then the bottom ones don't. Would it be better to write this logic as:
```
if (responsive) {
   if (
      align.bottom === 'top' && 
      // drop is overflowing above window
      targetRect.top - containerRect.height <= 0 && 
      // there's space below the target 
      targetRect.bottom + containerRect.height < windowHeight
   ) { 
       ... 
   } else if () {
   } ...
} else if (align.top === 'top') {...}
...
```
just would save a few condition checks for the non-responsive cases

taysea (commented):
Dropped some cleanup comments but the behavior is looking good.

src/js/components/Drop/DropContainer.js
jcfilben: ^ I like this format, switched to this in latest commit

src/js/components/Drop/DropContainer.js
jcfilben: If the height of the drop is smaller than the target and the drop is positioned so that it is displayed over the target then it doesn't do much good to try and reposition the drop unless we had a really large target. This check just prevents unnecessary re-positioning that doesn't really help make the drop more visible

src/js/components/Drop/DropContainer.js
jcfilben: If we restructure this then we run into issues where if responsive is true but it doesn't meet any of those top 4 conditions then it will never reach the other else if statements or the else statement at the end.

taysea: Ah thank you for clarifying.

taysea (approved):
LGTM, and super helpful additions with the comments. Thank you!

britt6612 (approved):
Thanks for cleaning up comments 🙌

PR #7321, opened 2024-09-19
Exploring how to trap focus in Drop

taysea:
#### What does this PR do?

This is an exploration off of https://github.com/grommet/grommet/pull/7305 of how to trap focus within the "top-most" drop.

There are some issues with re-renders/functionality, but putting this up for people to check out.

The general approach intent is:
1. Track a global array of any "FocusedContainer" that has been added
   - This differs from our current behavior where "RootsContext" only extends the array as nested "FocusedContainers" are added but therefore not all FocusedContainers have access to the same array
   - This updated approach is what the [WCAG pattern ](https://codepen.io/pen) and focus-trap library seem to be doing
2. Use this global array to:
   - Add focus event listener to the most recently opened FocusContainer
   - Remove focus event listener from any other open FocusContainers (this is only possible if FocusContainer has access to a shared global array so it can know when it's no longer the most recently added one)

That said with the approach of storing this global array, I'm running into a few issues with React lifecycle management:
- It's causing more re-renders to the top Grommet component which is not ideal
- State lags one render behind, so running into some competing focus between Drop's `restrictFocus` and the container that triggered the Drop to open (in the cases where a Drop is nested inside of a layer)

I feel like if we could find a way to reliable store the global array of "FocusContainers" (roots) without triggering re-renders from setState then the rest might fall into place.

Open to any feedback/suggestions.

### DIFF
--- a/src/js/components/FocusedContainer.js
+++ b/src/js/components/FocusedContainer.js
@@ -12,13 +12,113 @@ export const FocusedContainer = ({
 }) => {
   const [bodyOverflowStyle, setBodyOverflowStyle] = useState('');
   const ref = useRef(null);
+  const preNodeRef = useRef(null);
+  const postNodeRef = useRef(null);
 
-  const roots = useContext(RootsContext);
-  const [nextRoots, setNextRoots] = useState(roots);
+  const { roots, setRoots } = useContext(RootsContext);
+  const [active, setActive] = useState(true);
+
+  // only the most recent container in roots should be active and listening
+  // for focus events. when active is false, event listeners will be removed
   useEffect(() => {
-    // make sure value of null is not added to array
-    if (ref.current) setNextRoots([...roots, ref.current]);
+    if (ref.current && roots[roots.length - 1] === ref.current) {
+      setActive(true);
+    } else setActive(false);
   }, [roots]);
+
+  useEffect(() => {
+    const container = ref.current;
+    if (container) {
+      // avoid infinite loop by only adding it if it's not already there
+      const nextRoots = !roots.includes(container)
+        ? [...roots, container]
+        : roots;
+      setRoots(nextRoots);
+    }
+    // on unmount, remove this container from roots
+    return () => {
+      const nextRoots = [...roots.filter((root) => root !== container)];
+      setRoots(nextRoots);
+      // ensure ability to focus is restored to what will now be the
+      // top-most container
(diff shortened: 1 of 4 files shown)


src/js/components/Grommet/Grommet.js
```diff
@@ -150,10 +150,16 @@ const Grommet = forwardRef((props, ref) => {
...
+
+  // TO DO this global roots approach causes extra re-renders at
+  // this level which isn't ideal given it's global
+  console.log(roots);
+
   return (
     <ThemeContext.Provider value={theme}>
       <ResponsiveContext.Provider value={responsive}>
-        <RootsContext.Provider value={roots}>
+        <RootsContext.Provider value={rootsContextValue}>
```
jcfilben, line 162: If we are outside of the grommet container how do we track the roots?

britt6612: this is a really good point!

MikeKingdom: Contexts can have a default value. Can we make the root context have a default value based on `document`?

taysea: Good point. Need to think about where the RootsContext Provider lives/how that would work. To clarify, Jessica you're meaning a scenario like this:
```
<Grommet>
...
</Grommet>
<Drop>
// I am outside of Grommet wrapper so I don't have access to RootsContext.Provider from Grommet
</Drop>

Comment (jcfilben) [reply src/js/components/Grommet/Grommet.js:162]: ^ Yes that type of scenario

File: src/js/components/FocusedContainer.js [hunk 2]
```diff
@@ -12,14 +12,114 @@ export const FocusedContainer = ({
...
   }, [roots]);
 
+  useEffect(() => {
+    const container = ref.current;
+    if (container) {
+      // avoid infinite loop by only adding it if it's not already there
+      const nextRoots = !roots.includes(container)
+        ? [...roots, container]
+        : roots;
+      setRoots(nextRoots);
```
taysea, line 36: // try this approach to avoid the infinite loop

setRoots(prevValue => {

});

src/js/components/Grommet/Grommet.js
taysea: Maybe explore the context to be a ref instead of an array.

Have the event listener look at the ref to decide if it's the current container.

taysea:
Closing, superseded by #7324

PR #7324, opened 2024-09-23
Drop trap focus -- ref approach

taysea:
#### What does this PR do?

Approach 2 to https://github.com/grommet/grommet/pull/7321

This is a modified approach to minimize re-rendering which aims to store a global array of all open drops in a ref (as opposed to state which caused unneccessary rerenders). The behavior is working well for the most part (still some refinements to look into). That said, there are complaints in jest which  I think point to potentially deeper considerations we need to be aware of:

Screenshot 2024-09-23 at 4 30 04 PM

Screenshot 2024-09-23 at 4 30 18 PM

Need to figure out how to properly cleanup things in the useEffects to not run into these issues.

### DIFF
--- a/src/js/components/Drop/stories/Simple.js
+++ b/src/js/components/Drop/stories/Simple.js
@@ -1,36 +1,118 @@
-import React, { useEffect, useRef, useState } from 'react';
+import React, { useRef, useState } from 'react';
 
-import { Box, Drop } from 'grommet';
+import { Box, DropButton, Button, Anchor, Layer, Drop } from 'grommet';
+
+const InteractiveContent = () => (
+  <Box gap="medium" pad="large">
+    Drop Contents
+    <Button label="testing button" />
+    <Anchor href="#" label="testing anchor" />
+  </Box>
+);
 
 const align = { top: 'bottom', left: 'left' };
 
 const SimpleDrop = () => {
-  const targetRef = useRef();
+  const [showDrop, setShowDrop] = useState(false);
+  const [showLayer, setShowLayer] = useState(false);
+  const [showDropInModal, setShowDropInModal] = useState(false);
+  const [showRestrictFocusDrop, setShowRestrictFocusDrop] = useState(false);
+  const layerRef = useRef();
+  const buttonRef = useRef();
+  const layerButtonRef = useRef();
+  const restrictFocusButtonRef = useRef();
 
-  const [, setShowDrop] = useState(false);
-  useEffect(() => {
-    setShowDrop(true);
-  }, []);
+  const onClose = () => setShowLayer(false);
 
   return (
     // Uncomment <Grommet> lines when using outside of storybook
     // <Grommet theme={...}>
-    <Box fill align="center" justify="center">
-      <Box
-        background="dark-2"
-        pad="medium"
-        align="center"
-        justify="start"
-        ref={targetRef}
-      >
-        Target
+    <>
+      <Box direction="row" pad="large" gap="small">
+        <DropButton
+          label="Drop button"
+          dropContent={<InteractiveContent />}
+          dropAlign={align}
+        />
+        <Button
+          ref={buttonRef}
+          label="Regular button"
+          onClick={() => setShowDrop(true)}
+        />
+        <Button
+          ref={layerRef}
+          label="Open layer"
+          onClick={() => setShowLayer(true)}
+        />
+        {showLayer && (
+          <Layer
+            position="right"
+            full="vertical"
+            modal
+            onClickOutside={onClose}
+            onEsc={onClose}
+          >
+            <Box
+              fill="vertical"
+              overflow="auto"
+              width="medium"
+              pad="medium"
+              gap="small"
+            >
+              {/* <Anchor href="#" label="testing anchor" /> */}
+              <DropButton
+                label="Another drop button"
+                dropContent={<InteractiveContent />}
+                dropAlign={align}
+              />
+              <Button
+                ref={layerButtonRef}
+                label="Another regular button"
+                onClick={() => setShowDropInModal(true)}
+              />
+              {showDropInModal && (
+                <Drop
+                  target={layerButtonRef.current}
+                  align={align}
+                  onClickOutside={() => setShowDropInModal(false)}
+                  onEsc={() => setShowDropInModal(false)}
+                >
+                  <InteractiveContent />
+                </Drop>
+              )}
+              <Button
+                ref={restrictFocusButtonRef}
+                label="A regular button with restrict focus"
+                onClick={() => setShowDropInModal(true)}
+              />
+              {showRestrictFocusDrop && (
+                <Drop
+                  target={restrictFocusButtonRef.current}
+                  align={align}
+                  onClickOutside={() => setShowRestrictFocusDrop(false)}
+                  onEsc={() => setShowRestrictFocusDrop(false)}
+                  // TO DO onEsc isn't restoring
+                  // focus to the button as I'd expect
+                  restrictFocus
+                >
+                  <InteractiveContent />
+                </Drop>
+              )}
+            </Box>
+          </Layer>
+        )}
       </Box>
-      {targetRef.current && (
-        <Drop align={align} target={targetRef.current}>
-          <Box pad="large">Drop Contents</Box>
+      {showDrop && (
+        <Drop
+          target={buttonRef.current}
+          align={align}
+          onClickOutside={() => setShowDrop(false)}
+          onEsc={() => setShowDrop(false)}
+        >
+          <InteractiveContent />
         </Drop>
       )}
-    </Box>
+    </>
     // </Grommet>
   );
 };
--- a/src/js/components/FocusedContainer.js
+++ b/src/js/components/FocusedContainer.js
@@ -60,29 +60,26 @@ export const FocusedContainer = ({
       document.removeEventListener('focus', handleTrapFocus, true);
     };
 
-    const timer = setTimeout(() => {
-      // add container to the global roots
-      if (container) roots.push(container);
+    // add container to the global roots
+    if (container) roots.push(container);
 
-      // Create and insert focusable nodes to help track when focus
-      // has left this container but without letting focus be noticeably placed
-      // on anything outside the container
-      const preDiv = document.createElement('div');
-      const postDiv = document.createElement('div');
+    // Create and insert focusable nodes to help track when focus
+    // has left this container but without letting focus be noticeably placed
+    // on anything outside the container
+    const preDiv = document.createElement('div');
+    const postDiv = document.createElement('div');
 
-      preNodeRef.current = container.parentNode.insertBefore(preDiv, container);
-      postNodeRef.current = container.parentNode.insertBefore(
-        postDiv,
-        container.nextSibling,
-      );
-      preNodeRef.current.tabIndex = 0;
-      postNodeRef.current.tabIndex = 0;
+    preNodeRef.current = container.parentNode.insertBefore(preDiv, container);
+    postNodeRef.current = container.parentNode.insertBefore(
+      postDiv,
+      container.nextSibling,
+    );
+    preNodeRef.current.tabIndex = 0;
+    postNodeRef.current.tabIndex = 0;
 
-      addListeners();
-    }, 0);
+    addListeners();
 
     return () => {
-      clearTimeout(timer);
       // remove from global roots array
       if (roots.indexOf(container)) roots.splice(roots.indexOf(container), 1);
 
--- a/src/js/components/FocusedContainer.js
+++ b/src/js/components/FocusedContainer.js
@@ -15,110 +15,83 @@ export const FocusedContainer = ({
   const preNodeRef = useRef(null);
   const postNodeRef = useRef(null);
 
-  const { roots, setRoots } = useContext(RootsContext);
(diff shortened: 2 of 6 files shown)


src/js/components/FocusedContainer.js
```diff
@@ -12,13 +12,81 @@ export const FocusedContainer = ({
...
+      container.nextSibling,
+    );
+    preNodeRef.current.tabIndex = 0;
+    postNodeRef.current.tabIndex = 0;
+
+    addListeners();
+
+    return () => {
+      // remove from global roots array
+      if (roots.indexOf(container)) roots.splice(roots.indexOf(container), 1);
```
taysea, line 82: Try to simplify this down to only the roots part to start, to see if it resolves the issue.

Maybe separate out the useEffects with the preNode/postNode or check if they still exist before calling `.remove()`

taysea: Also could explore using `.filter()` on the array instead of `splice`

taysea: If we end up keeping the handler function here, then we might be able to initialize the RootsContext array to include `document` (to handle cases where Drop/Layer is rendered outside of Grommet tag).

src/js/contexts/RootsContext.js
```diff
@@ -1,6 +1,21 @@
...
+  // If FocusedContainer is within Grommet React tree, there will be a value
+  // from RootsContext. If not, set up a default value that can be passed
+  // to drops that might open from within a drop.
+  const defaultRoots = useRef([]);
+  const defaultValue = useMemo(() => ({ roots: defaultRoots }), []);
+  const existingRoots = useContext(RootsContext);
+
+  return {
+    contextValue: existingRoots || defaultValue,
+    withinGrommet: existingRoots?.roots !== undefined,
```
MikeKingdom, line 19: I'm wondering if this should be renamed to something like `isDefault`? Is it true in the "not within grommet" case that the first FocusedContainer will add the context but any children in that container that do a drop or Layer will result in useRoots returning `true` for `withinGrommet` and nicely not add another provider? (ie. `withinGrommet` is a little misleading.)

src/js/components/FocusedContainer.js
```diff
@@ -12,13 +12,92 @@ export const FocusedContainer = ({
...
+     textarea:not([tabindex="-1"]), [tabindex]:not([tabindex="-1"])`,
+        );
+
+        if (focusableElements.length === 0) return;
+
+        const firstElement = focusableElements[0];
+        const lastElement = focusableElements[focusableElements.length - 1];
+
+        // only perform focus if this is the most recently opened drop
+        if (roots[roots.length - 1] === container) {
```
MikeKingdom, line 39: would it be worth moving this check up between 26 and 27 so it avoids doing the query when it's not the top container?

MikeKingdom (commented):
I like it. Cool stuff. I'd still like to test it out a little but it's looking good

src/js/components/FocusedContainer.js
```diff
@@ -44,24 +123,31 @@ export const FocusedContainer = ({
...
       }
     }, 0);
 
-    return () => {
-      // remove trap and restore ability to focus on the last root only
-      if (roots && roots[0]) makeNodeFocusable(roots[roots.length - 1]);
-      clearTimeout(timer);
-    };
-  }, [hidden, roots, trapFocus]);
+    return () => clearTimeout(timer);
```
britt6612, line 136: within here do we need to make the nodesFocusable again? 
above we have 
```
  // make every root before this one unfocusable
        roots.forEach((root, index) => {
          if (index < roots.length - 1) makeNodeUnfocusable(root);
        });

```

so in the return do we need 

      // Make all roots focusable again
      roots.forEach((root) => {
        makeNodeFocusable(root);
      });

taysea: I've done it in the other useEffect once we splice the roots array to the updated array on unmount.

src/js/contexts/RootsContext.js
taysea: Correct -- "any children in that container that do a drop or Layer will result in useRoots returning true nicely not add another provider". Open to changing the name of the variable-- agree in that case it's a bit misleading.

taysea: Renamed to `hasRoots`

src/js/components/FocusedContainer.js
taysea: Done.

jcfilben (commented):
Stil looking through the code but one thing I noticed is if we are outside the grommet container and I test with the Drop/Simple story in this PR, I get some unexpected behavior with the focus order on the regular button drop.


https://github.com/user-attachments/assets/c5a0e8c5-cc50-459c-9dd9-036f56db0ca8

Notice how after opening the 'regular button' drop my next tab stop is the 'open layer' button

src/js/components/Drop/stories/Simple.js
```diff
@@ -1,36 +1,118 @@
...
+                label="A regular button with restrict focus"
+                onClick={() => setShowDropInModal(true)}
+              />
+              {showRestrictFocusDrop && (
+                <Drop
+                  target={restrictFocusButtonRef.current}
+                  align={align}
+                  onClickOutside={() => setShowRestrictFocusDrop(false)}
+                  onEsc={() => setShowRestrictFocusDrop(false)}
+                  // TO DO onEsc isn't restoring
```
britt6612, line 94: just saw this locally adding comment so I remember to look into this more as well!

britt6612: The 'original element' is coming back as the whole `body` in the dom so the focus is not going back on the `button` thats not the correct 'original element'

britt6612:
also see this in the console on when opening the Layer 

Screenshot 2024-09-27 at 1 47 10 PM

src/js/components/Drop/stories/Simple.js
taysea: This is resolved now, realized it was actually an error in how I'd implemented the story.

taysea:
> also see this in the console on when opening the Layer
> 
> Screenshot 2024-09-27 at 1 47 10 PM

This issue is happening on master as well (I'm wondering if it's a newer error from a dependency because we've had this functionality in Grommet for a while). @jcfilben Are you comfortable addressing the fix in a separate PR if so?

taysea:
> Stil looking through the code but one thing I noticed is if we are outside the grommet container and I test with the Drop/Simple story in this PR, I get some unexpected behavior with the focus order on the regular button drop.
> 

Fixed.

jcfilben:
> This issue is happening on master as well (I'm wondering if it's a newer error from a dependency because we've had this functionality in Grommet for a while). @jcfilben Are you comfortable addressing the fix in a separate PR if so?

Yeah I think this can be handled in a separate PR

jcfilben:
When I open the 'regular button' drop while in the netlify deploy preview focus moves to the netlify toolbar first before going into the drop. This isn't an issue with the 'drop button' or 'open layer' examples.


https://github.com/user-attachments/assets/82722276-8606-4395-ad67-18f20da32aa1

taysea:
> When I open the 'regular button' drop while in the netlify deploy preview focus moves to the netlify toolbar first before going into the drop. This isn't an issue with the 'drop button' or 'open layer' examples.
> 

I debugged why this is happening. What happens is:
- The "regular button" in that story does not have `restrictFocus`, therefore when you open the Drop, focus stays on the button itself.
- The Netlify deploy is rendered in an iframe with its own internal html document
- The "Drop" that opened as a result of the button is appended at the end of the overall DOM (after the Netlify iframe)
- When you press "Tab", the Netlify iframe is technically the next thing in the DOM. The Grommet 'focus' event listener which is placed on the document doesn't apply to the Netlify iframe (since it has its own document) so the handleTrapFocus function is never called to "trapFocus" within the Drop instead.

This is an interesting edge case (the same behavior actually exists on master branch right now so it's not a bug introduced in this PR). That said, I'm not sure what the right solution to this would be given `document` feels like the reasonable place to attach the event listener (grommet tries not to affect things outside of itself). I'm tempted to argue in that case, someone should just use `restrictFocus` or `autoFocus` on something within their Drop.

jcfilben:
> This is an interesting edge case. That said, I'm not sure what the right solution to this would be given document feels like the reasonable place to attach the event listener (grommet tries not to affect things outside of itself). I'm tempted to argue in that case, someone should just use restrictFocus or autoFocus on something within their Drop.

Thanks for looking into this. Agree with leaving the event listener as is and that people should just use restrict focus or autoFocus if they run into this edge case.

jcfilben (approved):
This looks good to me!
Nice work on this

jcfilben:
Linking to https://github.com/grommet/grommet/issues/5896. Once we merge I think we can close #5896

britt6612 (approved):
LGTM

Thank you for assisting in this one!

src/js/components/FocusedContainer.js
```diff
@@ -12,13 +30,112 @@ export const FocusedContainer = ({
...
+      if (
+        !hidden &&
+        trapFocus &&
+        ignoreUtilFocusRef.current === false &&
+        container &&
+        // only perform focus if this is the most recently opened drop
+        roots[roots.length - 1] === container
+      ) {
+        if (container.contains(e.target)) {
+          container.lastFocus = e.target;
```
MikeKingdom, line 90: I believe this is adding an property to an element object which may not be an OK practice since we don't own it, especially if the object is locked (firefox?). The original case where this came from I believe was able to store this in somewhere it owned. Should we move this to a ref in here?

src/js/components/FocusedContainer.js
```diff
@@ -44,24 +161,31 @@ export const FocusedContainer = ({
   }, [bodyOverflowStyle, hidden, trapFocus, restrictScroll]);
 
   useEffect(() => {
+    const roots = contextRoots.current;
     const timer = setTimeout(() => {
-      if (!hidden && trapFocus && roots && roots[0]) {
-        roots.forEach(makeNodeUnfocusable);
+      if (!hidden && trapFocus && roots?.[0]) {
```
MikeKingdom, line 166: I don't think you need the `roots?.[0]` check, at least if you change line 68 to do `roots?.forEach`

src/js/components/FocusedContainer.js
```diff
@@ -12,13 +30,112 @@ export const FocusedContainer = ({
...
+      }
+    };
+
+    const addListeners = () => {
+      document.addEventListener('focus', handleTrapFocus, true);
+    };
+
+    const removeListeners = () => {
+      document.removeEventListener('focus', handleTrapFocus, true);
+    };
```
MikeKingdom, line 107: Since these are trivial and only used on one spot you could just inline them where used and avoid creating functions as well as simplify this already large useEffect

src/js/components/Grommet/Grommet.js
```diff
@@ -146,14 +145,18 @@ const Grommet = forwardRef((props, ref) => {
 
   const grommetRef = useForwardedRef(ref);
 
+  // track open FocusedContainers in a global array to manage
+  // focus event listeners for trapFocus
+  const roots = useRef([]);
   useEffect(() => {
-    if (grommetRef.current) setRoots([grommetRef.current]);
+    if (grommetRef.current) roots.current.push(grommetRef.current);
   }, [grommetRef]);
```
MikeKingdom, line 153: I think really you'd want this to depend on `grommetRef.current`. I'm wondering if you're getting lucky because this useEffect is happening after the current gets populated, but I'm curious if there's an unmount/remount that could break it (ie. thinking ahead to why React has the StrictMode checks in new versions.)

MikeKingdom (commented):
It's looking really good. There's a couple of questions I had (and a couple cosmetic things.)

src/js/components/FocusedContainer.js
taysea: You're right, I think this was lingering from old logic.

src/js/components/FocusedContainer.js
taysea: Agree, will update.

britt6612:
@taysea also need to update bundle size

src/js/components/Grommet/Grommet.js
taysea: I get this eslint error when I try to switch it do grommetRef.current.

Screenshot 2024-10-02 at 11 10 43 AM

src/js/components/FocusedContainer.js
taysea: In this case `container` is `ref.current` where the "ref" is on the FocusedContainer. Let me know if the scenario is still a concern.

taysea: Are you suggesting we have a `lastFocusRef = useRef(null)` instead?

MikeKingdom: That was my first thought. I think it would serve the same purpose and it's local and basically has the same scope as the container involved

src/js/components/FocusedContainer.js
taysea: Done.

src/js/components/FocusedContainer.js
taysea: Done -- This direction works as well. I like the useRef pattern.

MikeKingdom (approved):
Thanks!

PR #7334, opened 2024-10-04
Replace useId in AccordionPanel

jcfilben:
#### What does this PR do?
Replaces useId in AccordionPanel because useId is only available in React 18 and we need to support react 16 and 17 as well.

no
yes
backwards compatible


src/js/components/AccordionPanel/AccordionPanel.js
```diff
@@ -10,11 +10,19 @@ import { AccordionContext } from '../Accordion/AccordionContext';
 import { AccordionPanelPropTypes } from './propTypes';
 import { useThemeValue } from '../../utils/useThemeValue';
 
+// Ideally we should use `useId` from react but this is only available
+// in React 18 and we want to keep compatibility with React 17 and 16.
+function getUID() {
+  const timestamp = Date.now().toString(36);
```
jcfilben, line 16: consider appending math.random

jcfilben:
Closing in favor of https://github.com/grommet/grommet/pull/7345

PR #7345, opened 2024-10-07
Accordion - Replace useId

jcfilben:
#### What does this PR do?
Replaces useId in AccordionPanel because useId is only available in React 18 and we need to support react 16 and 17 as well.

This PR is targeted to merge into the react-compatibility branh so we can do a patch release from that branch

no
yes
backwards compatible

### DIFF
--- a/src/js/components/Accordion/__tests__/Accordion-test.tsx
+++ b/src/js/components/Accordion/__tests__/Accordion-test.tsx
@@ -19,6 +19,16 @@ const customTheme = {
 };
 
 describe('Accordion', () => {
+  beforeEach(() => {
+    jest.spyOn(Date, 'now').mockImplementation(() => 12345678);
+    jest.spyOn(global.Math, 'random').mockReturnValue(0.123456789);
+  });
+
+  afterEach(() => {
+    jest.spyOn(Date, 'now').mockRestore();
+    jest.spyOn(global.Math, 'random').mockRestore();
+  });
+
   test('should have no accessibility violations', async () => {
     const { container, asFragment } = render(
       <Grommet>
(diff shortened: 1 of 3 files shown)


src/js/components/AccordionPanel/AccordionPanel.js
```diff
@@ -10,11 +10,19 @@ import { AccordionContext } from '../Accordion/AccordionContext';
 import { AccordionPanelPropTypes } from './propTypes';
 import { useThemeValue } from '../../utils/useThemeValue';
 
+// Ideally we should use `useId` from react but this is only available
+// in React 18 and we want to keep compatibility with React 17 and 16.
+function getUID() {
+  const timestamp = Date.now().toString(36);
```
taysea, line 16: Should we do:

`Date.now() + Math.random()` instead of just appending Math.random?

https://stackoverflow.com/questions/8012002/create-a-unique-number-with-javascript-time/40591207#40591207

src/js/components/Accordion/__tests__/Accordion-test.tsx
```diff
@@ -19,6 +19,16 @@ const customTheme = {
 };
 
 describe('Accordion', () => {
+  beforeEach(() => {
+    jest.spyOn(Date, 'now').mockImplementation(() => 12345678);
+    jest.spyOn(global.Math, 'random').mockReturnValue(0.123456789);
```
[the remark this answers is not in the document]
jcfilben: Added a comment in latest commit

src/js/components/AccordionPanel/AccordionPanel.js
jcfilben: Changed in latest commit

taysea (approved):

britt6612 (approved):

PR #7348, opened 2024-10-07
fix #7347 - use alternative useId implementation supporting older React versions

### LINKED ISSUE #7347: AccordionPanel - `useId` breaks renders in supported React versions below 18
In "supported" versions of React below 18, the `AccordionPanel` will fail to render due to the use of the `useId` hook.

Yes

### Expected Behavior

`AccordionPanel` should render in all supported versions

### Actual Behavior

`AccordionPanel` only renders in react >= 18.

### Steps to Reproduce

Use `AccordionPanel` in a React version below 18.

### Your Environment

- Grommet version: 2.40.1

ganondev:
Changes the use of React 18's `useId` hook to a third part implementation, which I've audited on my own (it's quite straightforward). Using an existing implementation was the most straightfowrad solution, however I realize that may not be what the maintainers want to do. In any case, I wanted to speed things up by trying to fix this issue on my own.

#### What does this PR do?
Adds the `react-use-id-hook` package and uses it in `AccordionPanel`.

Tested by running relevant tests within jest under React 16.1, 17, and 18 to ensure the render passes.

#7347

Backwards compatible, React 18 users should not notice a difference outside of snapshot tests.


britt6612:
@ganondev going to close in favor of https://github.com/grommet/grommet/pull/7345

PR #7353, opened 2024-10-08
Select - `focusIndicator` prop should only work if `plain=true`

### LINKED ISSUE #4165: focusIndicator on Select shouldn't work unless Select uses `plain`
### Expected Behavior
focusIndicator on Select shouldn't work unless Select uses `plain.

See the docs: 
https://v2.grommet.io/select#focusIndicator

### Actual Behavior
focusIndicator should affect the focus behavior only when the component is using `plain`.

https://v2.grommet.io/select#focusIndicator

```
function Example() {
  const [value, setValue] = React.useState('medium');
  return (
    <Select   
      focusIndicator={false}
      options={['small', 'medium', 'large']}
      value={value}   
    />
  );
}
```

### Steps to Reproduce

1. Open the code example
2. Notice how the focus is gone 
3. Focus should be there unless the component is using `plain`

### Your Environment
any

- Grommet version: latest

avantika-codes:
#### What does this PR do?

This PR modifies the Select component to ensure that the focusIndicator prop is only evaluated if plain={true}. If plain={false}, the focusIndicator is ignored, and the focus indicator is displayed by default. Additionally, it includes a new test case to validate this behavior.

The reviewer should start by examining the changes made to the Select component, particularly around the conditional handling of the focusIndicator prop. Then, the reviewer can review the newly added test case to ensure it properly validates the behavior when plain={false}.

Manual and automated testing using Jest has been done. A new test case was added to ensure that when plain={false}, the focusIndicator defaults to true regardless of the passed prop value.

- Render the Select component with plain={false} and check that the focus indicator is displayed by default, even if focusIndicator={false} is passed.
- Use tools like browser dev tools to verify the presence of the box-shadow or other focus styles on the Select component's button.
- Run the updated Jest tests to ensure they pass successfully.

The change was prompted by the need to ensure that the focusIndicator prop is correctly handled when plain={false}, as the default behavior should show the focus indicator in such cases.

There may not be an explicitly tracked issue for this, but the change addresses a behavioral inconsistency where the focusIndicator was not being correctly handled when plain={false}.

Focus is indicated by default for when focusIndicator is false, and plain isn't specified:

```
function Example() {
  const [value, setValue] = React.useState('medium');
  return (
    <Select   
      focusIndicator={false}
      options={['small', 'medium', 'large']}
      value={value}   
    />
  );
}
```

Screenshot 2024-10-09 at 4 38 27 AM

Yes, the documentation might need a slight update to reflect the fact that focusIndicator is ignored when plain={false}, and the focus indicator is always displayed in that case.

Yes, it should be mentioned, especially if developers were relying on the behavior of focusIndicator with plain={false}. This clarifies how the prop is handled going forward.

This change is backwards compatible. It only affects cases where plain={false}, ensuring that the focus indicator is always shown, which aligns with expected behavior.


britt6612 (approved):
LGTM

jcfilben (approved):
Looks good!

jcfilben:
Added the hacktoberfest-approved label, waiting to merge until after our next release

jcfilben:
Closes https://github.com/grommet/grommet/issues/4165

PR #7356, opened 2024-10-09
Fix `trapFocus` unwanted page scroll

taysea:
#### What does this PR do?

NOTE: Drop "Simple" and "Simple2" stories are purely for QA in review. Changes to those stories and `storybook/preview.js` should be reverted before merging.

When an element receives focus, the browser will automatically scroll to attempt to bring that element into view. In the case of `trapFocus`, we don't want this to happen and instead it can cause a confusing/unexpected page scroll. Therefore, these adjustments listen for the `keydown` tab event in the cases where a Drop/Layer has `trapFocus` and the focus is:

- a. not in the container yet (`restrictFocus={false}`)
- b. on the last focusable element and will loop with tab
- c. on the first focusable element and will loop with shift + tab.

In the behavior merged in https://github.com/grommet/grommet/pull/7324, we hadn't tested the scenario where focusable elements existed outside of the Grommet wrapper (aka `makeNodeUnfocusable` wouldn't have been called on them). 

Additionally, the code pattern we referenced from APG always placed focus within the dialog when it opens. In our cases (`restrictFocus={false}`), that isn't guaranteed. When modifying their code to align with our cases, the same behavior can be seen (ensure that the bottom "Add Delivery Address" button is out of view, then press "Enter" on the first Add Delivery Address button, then press "Tab" and see how the page jumps down): https://zzcdcj.csb.app/

Similar to the change made in TextInput in #7324, MaskedInput should have `trapFocus={false}` since it maintains focus on the input itself.

FocusedContainer.js

Locally in the scenario where focusable elements exist outside of the Grommet wrapper.

Using Drop Simple/Simple2 stories.

BEFORE (with undesired page scroll)

https://github.com/user-attachments/assets/58d68034-e147-40a8-af4c-297bb67eaca7

AFTER

https://github.com/user-attachments/assets/ae65917f-6fc9-4cf6-a3ef-3e39ff3ff165

https://github.com/user-attachments/assets/ab5b88cf-e2e5-43e4-9487-3866b75f3c4c

No.

Backwards compatible.

### DIFF
--- a/src/js/components/Drop/stories/Simple.js
+++ b/src/js/components/Drop/stories/Simple.js
@@ -1,40 +1,124 @@
 import React, { useEffect, useRef, useState } from 'react';
 
-import { Box, Drop } from 'grommet';
+import { Box, DropButton, Button, Anchor, Layer, Drop } from 'grommet';
+
+const InteractiveContent = () => (
+  <Box gap="medium" pad="large">
+    Drop Contents
+    <Button label="testing button" />
+    <Anchor href="#" label="testing anchor" />
+  </Box>
+);
 
 const align = { top: 'bottom', left: 'left' };
 
 const SimpleDrop = () => {
-  const targetRef = useRef();
+  const [showDrop, setShowDrop] = useState(false);
+  const [showLayer, setShowLayer] = useState(false);
+  const [showDropInModal, setShowDropInModal] = useState(false);
+  const [showRestrictFocusDrop, setShowRestrictFocusDrop] = useState(false);
+  const layerRef = useRef();
+  const buttonRef = useRef();
+  const layerButtonRef = useRef();
+  const restrictFocusButtonRef = useRef();
 
-  const [, setShowDrop] = useState(false);
+  const onClose = () => setShowLayer(false);
+
   useEffect(() => {
-    setShowDrop(true);
+    const timer = setTimeout(() => buttonRef.current.focus(), 5000);
+    return () => clearTimeout(timer);
   }, []);
 
   return (
     // Uncomment <Grommet> lines when using outside of storybook
     // <Grommet theme={...}>
-    <Box fill align="center" justify="center">
-      <Box
-        background="dark-2"
-        pad="medium"
-        align="center"
-        justify="start"
-        ref={targetRef}
-      >
-        Target
+    <>
+      <Box direction="row" pad="large" gap="small">
+        <DropButton
+          label="Drop button"
+          dropContent={<InteractiveContent />}
+          dropAlign={align}
+        />
+        <Button
+          ref={buttonRef}
+          label="Regular button"
+          onClick={() => setShowDrop(true)}
+        />
+        <Button
+          ref={layerRef}
+          label="Open layer"
+          onClick={() => setShowLayer(true)}
+        />
+        {showLayer && (
+          <Layer
+            position="right"
+            full="vertical"
+            modal
+            onClickOutside={onClose}
+            onEsc={onClose}
+          >
+            <Box
+              fill="vertical"
+              overflow="auto"
+              width="medium"
+              pad="medium"
+              gap="small"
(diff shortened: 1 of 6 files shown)


storybook/preview.js
```diff
@@ -79,9 +79,22 @@ export const decorators = [
     }
 
     return (
-      <Grommet theme={THEMES[state]} full={full} dir={dir} options={options}>
-        <Story state={THEMES[state]} />
-      </Grommet>
+      <>
```
taysea, line 82: NOTE: these changes should be undone before merging

src/js/components/Drop/stories/Simple.js
```diff
@@ -1,40 +1,124 @@
 import React, { useEffect, useRef, useState } from 'react';
 
-import { Box, Drop } from 'grommet';
+import { Box, DropButton, Button, Anchor, Layer, Drop } from 'grommet';
```
taysea, line 3: NOTE: these changes should be undone before merging

src/js/components/Drop/stories/Simple2.js
```diff
@@ -0,0 +1,253 @@
+import React, { useEffect, useRef, useState } from 'react';
```
taysea, line 1: NOTE: this file should be deleted before merging.

jcfilben (commented):
In the Drop/Multiple if I open the first drop and then press tab twice it scrolls the page down

https://github.com/user-attachments/assets/7cec758b-6d04-4cb9-98dc-f65dfa8a4bc2

jcfilben:
I'm also seeing an issue in DateInput where I can no longer access the dates in the Calendar via keyboard, my focus is instead looped through the forward/back buttons for the Calendar

britt6612:
> In the Drop/Multiple if I open the first drop and then press tab twice it scrolls the page down
> 
>  Oct-09-2024.16-55-47.mp4

this should be fixed added a check for if first and last tab able element were the same

britt6612:
not trapping focus correctly in Layer example
https://github.com/user-attachments/assets/49948a7e-369f-4748-a4c6-3d616794d2b1

britt6612:
> not trapping focus correctly in Layer example https://github.com/user-attachments/assets/49948a7e-369f-4748-a4c6-3d616794d2b1

behavior is already existing and was not introduced because of any of our new changes behavior is also in v2.40.0 so we will not worry about it here. Whats happening is the Anchors are the next tab-able element because drop is going to be the last element in the dom it will go through the Anchor first

src/js/components/FocusedContainer.js
```diff
@@ -60,41 +68,75 @@ export const FocusedContainer = ({
...
       ) {
-        if (container.contains(e.target)) {
-          lastFocusRef.current = e.target;
-        } else {
-          focusFirstDescendant(container);
-          if (lastFocusRef.current === document.activeElement) {
-            focusLastDescendant(container);
-          }
-          lastFocusRef.current = document.activeElement;
+        e.preventDefault();
```
jcfilben, line 135: I think this preventDefault might be preventing more than we want it to. When I tab to the TextInput within the first Drop in the Drop/Multiple story the suggestions drop doesn't open until I press another key


https://github.com/user-attachments/assets/a3e660b7-f5fb-4a5d-a413-5516c8131c5e

src/js/components/FocusedContainer.js
```diff
@@ -16,11 +16,19 @@ const isFocusable = (element) => {
     case 'SELECT':
     case 'TEXTAREA':
       return true;
+    case 'DIV':
```
taysea, line 19: should this be any element with tabIndex >= 0? just to be safe if someone has programmatically added it

britt6612: yes thats on the next line

britt6612:
closing in favor of #7360

PR #7359, opened 2024-10-11
Defined Select-onChange Type and Fixed the Test Cases

TheThinker277:
#### What does this PR do?
This PR defines the Type used in onChange, onClose, onOpen function of the Select Component.

Yes.

No, I don't think so.
No.


src/js/components/Select/index.d.ts
```diff
@@ -8,6 +8,15 @@ import {
   PlaceHolderType,
 } from '../../utils';
 
+type OptionType = string | number | { label: string; value: string | number };
+
+// Define the event type for the onChange function
+interface SelectOnChangeEvent {
+  option?: number; // Option selected from the dropdown
```
britt6612, line 15: I think some of these types are to narrow. 

For example if you now look at the CreatOption.tsx story we are getting an error on the `option` since it is not a number

britt6612: I also wonder if it would help to test this if we either change locally the stories we have to be typescript to ensure we are not breaking anything as well as the tests we can change to be typescript as well to test.

TheThinker277: Hi @britt6612 ,
Thanks for the point out,  Will look into this.

TheThinker277: @britt6612 
I have fixed the issues on CreateOptions.tsx, Please have a look.
Thanks

src/js/components/Select/index.d.ts
```diff
@@ -8,6 +8,18 @@ import {
   PlaceHolderType,
 } from '../../utils';
 
+export type OptionType =
```
britt6612, line 11: I believe there is still some missing from here
https://v2.grommet.io/select#options

src/js/components/SelectMultiple/__tests__/SelectMultiple-test.tsx
```diff
@@ -37,8 +37,8 @@ const defaultOptions = [
 ];
 
 const TestManyOptions = () => {
-  const [options, setOptions] = useState(defaultOptions);
-  const [valueMultiple, setValueMultiple] = useState([]);
+  const [options, setOptions] = useState<string[]>(defaultOptions);
```
britt6612, line 40: why was the `<string[]>` added here?

britt6612: do we just in to expand on types so we dont need this?

TheThinker277: I added the Type because of the error I was getting while setting the options.

TheThinker277: I am not sure if any other Type is left. We need to decide the Types that are required and then make appropriate Changes to support those.
If there's some Trivial Case left which you can point out, I'll add those.

Thanks

src/js/components/Pagination/__tests__/Pagination-test.tsx
```diff
@@ -456,13 +456,15 @@ describe('Pagination', () => {
...
-      const [step, setStep] = useState(10);
+      const [step, setStep] = useState<number>(10);
       return (
         <Grommet>
           <Select
             a11yTitle="Controlled select"
             options={[10, 25, 50, 100]}
-            onChange={({ option }) => setStep(option)}
+            onChange={({ option }) => {
+              if (typeof option === 'number') setStep(option);
```
britt6612, line 466: why did we add this?

src/js/components/Pagination/__tests__/Pagination-test.tsx
```diff
@@ -530,8 +532,7 @@ describe('Pagination', () => {
           options={[10, 20, 50, 100]}
           value={itemPerPage} // Set value of Select to itemPerPage
           onChange={({ option }) => {
-            // Update itemPerPage state when Select value changes
-            setItemPerPage(option);
+            if (typeof option === 'number') setItemPerPage(option);
```
britt6612, line 535: also why are we checking the type of option

src/js/components/Select/index.d.ts
```diff
@@ -32,10 +47,10 @@ export interface BasicSelectProps {
   margin?: MarginType;
   messages?: { multiple?: string };
   name?: string;
-  onChange?: (...args: any[]) => void;
-  onClose?: (...args: any[]) => any;
+  onChange?: (event: SelectOnChangeEvent) => void;
+  onClose?: React.MouseEventHandler<HTMLButtonElement>;
   onMore?: (...args: any[]) => any;
-  onOpen?: (...args: any[]) => any;
+  onOpen?: React.MouseEventHandler<HTMLButtonElement>;
```
britt6612, line 53: the `onClose` and `onOpen` make sense to me just need to make sure we arent missing anything in the `onChange`

src/js/components/Select/index.d.ts
```diff
@@ -32,10 +47,10 @@ export interface BasicSelectProps {
   margin?: MarginType;
   messages?: { multiple?: string };
   name?: string;
-  onChange?: (...args: any[]) => void;
-  onClose?: (...args: any[]) => any;
+  onChange?: (event: SelectOnChangeEvent) => void;
```
britt6612, line 50: do we also need `event?: SelectOnChangeEvent`

src/js/components/Pagination/__tests__/Pagination-test.tsx
TheThinker277: I added that because OnChange will accept an argument of type OptionType, and step is a type of number, if we don't check the type, it will show an error.

src/js/components/Select/index.d.ts
TheThinker277: I don' think so, because we already have onChange? and we have used onChange as a functional Prop to Select.

TheThinker277: I don't think that there will be a case when we use OnChange and don't have arguments to pass it to.

src/js/components/Pagination/__tests__/Pagination-test.tsx
TheThinker277: I added that because OnChange will accept an argument of type OptionType, and option is a type of number, if we don't check the type, it will show an error.

TheThinker277:
Hi @Sulaymon333, Can you please review this PR?

TheThinker277:
@britt6612 , @Sulaymon333  Bumping for review

src/js/components/Select/index.d.ts
jcfilben, line 53: I would expect `onClose` and `onOpen` to both be functions

src/js/components/Select/index.d.ts
```diff
@@ -56,7 +71,13 @@ export interface SelectProps extends BasicSelectProps {
   defaultValue?: string | number | object | (string | number | object)[];
   multiple?: boolean;
   selected?: number | number[];
-  value?: string | JSX.Element | number | object | (string | number | object)[];
+  value?:
+    | boolean
```
jcfilben, line 75: Why did we include `boolean` here? Looking at the docs it isn't listed there

src/js/components/SelectMultiple/index.d.ts
```diff
@@ -5,7 +5,7 @@ export interface SelectMultipleProps extends BasicSelectProps {
   defaultValue?: (string | number | object)[];
   help?: string | React.ReactNode;
   limit?: number;
-  value?: (string | number | object)[];
+  value?: string | number | boolean | JSX.Element | object;
```
jcfilben, line 8: Why are we including boolean and JSX.Element here?

src/js/components/Select/index.d.ts
TheThinker277: Made the required change.

src/js/components/SelectMultiple/index.d.ts
TheThinker277: I added the JSX.Element because it was there in Select Value type.

src/js/components/Select/index.d.ts
TheThinker277: I checked the Type for the other Components, they have used similar Type for onOpen and onClose

src/js/components/SelectMultiple/index.d.ts
britt6612: https://v2.grommet.io/selectmultiple#value

we dont need boolean here

src/js/components/Select/index.d.ts
britt6612: @TheThinker277 

I think this type if fine for the `onOpen` and `onClose`

React.MouseEventHandler<HTMLButtonElement> is a type that represents a function that handles mouse events specifically for a button.

src/js/components/Pagination/__tests__/Pagination-test.tsx
taysea: I don't think we need to adjust this test -- it looks like the test has all options defined as numbers, so I don't think an error will be thrown. Let's keep the PR lean if possible.

src/js/components/Pagination/__tests__/Pagination-test.tsx
taysea: I don't think we need to adjust this test -- it looks like the test has all options defined as numbers, so I don't think an error will be thrown. Let's keep the PR lean if possible.

src/js/components/Select/index.d.ts
```diff
@@ -8,6 +8,21 @@ import {
   PlaceHolderType,
 } from '../../utils';
 
+export type OptionType =
+  | string
+  | number
+  | boolean
+  | JSX.Element
+  | { label: string; value: string | number }
```
taysea, line 16: Is this correct/necessary? Or is `object` sufficient?

src/js/components/Select/index.d.ts
taysea: This event type seems to be missing the actual event itself. When logging the event from `onChange`, the following event is fired depending on if I interact with mouse or keyboard.

Screenshot 2024-10-29 at 3 01 00 PM
Screenshot 2024-10-29 at 3 01 24 PM

src/js/components/Select/index.d.ts
taysea: `onOpen isn't called with any event so I don't think an event handler makes sense here.

If anything, it seems like `() => void` would be more appropriate for `onOpen` and `onClose`

Screenshot 2024-10-29 at 3 05 52 PM

src/js/components/Select/stories/typescript/CreateOption.tsx
```diff
@@ -1,19 +1,20 @@
...
 for (let i = 1; i <= 5; i += 1) {
   defaultOptions.push(`option ${i}`);
 }
 
 const updateCreateOption = (text: string) => {
   const len = defaultOptions.length;
-  if (defaultOptions[len - 1].includes(prefix)) {
-    // remove Create option before adding an updated one
-    defaultOptions.pop();
+  if (typeof defaultOptions[len - 1] === 'string') {
```
taysea, line 14: Can we simplify logic to `if (typeof defaultOptions[len - 1] === 'string' && defaultOptions[len - 1].includes(prefix))`?

src/js/components/SelectMultiple/index.d.ts
taysea: It also doesn't look like JSX.Element is a supported SelectMultiple value

britt6612:
closing due to inactivity feel free to reopen.

Issue #7344, opened 2024-10-07
Layer - when focus loops through layer the screen is scrolled down


jcfilben:
### Is this something you or your team is interested in contributing a pull request towards?
Yes

### Expected Behavior
Moving focus in Layer should not cause the screen behind the Layer to scroll

### Actual Behavior
When tabbing through the focusable items in a Layer when focus loops back up to the first element the screen behind the Layer scrolls down

### URL, screen shot, or Codepen exhibiting the issue

https://github.com/user-attachments/assets/c474d34e-972f-4dd4-97b2-2792a7e0efd1

### Steps to Reproduce

This issue can be seen with the stable branch of grommet on the Design System site Layer example

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

vijaykr338:
I'd like to be assigned on this

PR #7360, opened 2024-10-11
Fix Trapfocus scroll alternate approach

jcfilben:
#### What does this PR do?
This is an alternate approach to https://github.com/grommet/grommet/pull/7356

Notes from #7456 that are also relevant to this PR:
NOTE: Drop "Simple" and "Simple2" stories are purely for QA in review. Changes to those stories and storybook/preview.js should be reverted before merging.

When an element receives focus, the browser will automatically scroll to attempt to bring that element into view. In the case of trapFocus, we don't want this to happen and instead it can cause a confusing/unexpected page scroll. **This PR adds styles to the divs that we use to determine if the focus moves outside of the drop to ensure that they are always within the viewport so we don not need to scroll.**

Similar to the change made in TextInput in https://github.com/grommet/grommet/pull/7324, MaskedInput should have trapFocus={false} since it maintains focus on the input itself.

Closes https://github.com/grommet/grommet/issues/7344

no
yes
backwards compatible


storybook/preview.js
```diff
@@ -79,9 +79,22 @@ export const decorators = [
     }
 
     return (
-      <Grommet theme={THEMES[state]} full={full} dir={dir} options={options}>
-        <Story state={THEMES[state]} />
-      </Grommet>
+      <>
```
britt6612, line 82: note: to take out

src/js/components/MaskedInput/MaskedInput.js
```diff
@@ -431,6 +431,8 @@ const MaskedInput = forwardRef(
               target={inputRef.current}
               onClickOutside={onHideDrop}
               onEsc={onHideDrop}
+              // MaskedInput manages its own keyboard behavior via Keyboard
```
britt6612, line 434: are there any other components were missing?

src/js/components/FocusedContainer.js
```diff
@@ -103,12 +105,23 @@ export const FocusedContainer = ({
...
+      preDiv.style.position = 'absolute';
+      preDiv.style.height = '1px';
+      preDiv.style.top = '0';
+      preDiv.style.left = '0';
+      preDiv.style.right = '0';
+      postDiv.style.position = 'absolute';
+      postDiv.style.height = '1px';
+      postDiv.style.left = '0';
+      postDiv.style.right = '0';
+      postDiv.style.bottom = '0';
```
britt6612, line 123: I like this better since we are not having to add a keydown event listener.. can we just clean this up a little by abstracting the common styles out

const commonStyles = {
    position: 'absolute',
    height: '1px',
    left: '0',
    right: '0',
};

Object.assign(preDiv.style, {
    ...commonStyles,
    top: '0',
});

same thing for postDiv...

just so if we need to edit anything its in one place ( dont think we ever would) but may be cleaner

src/js/components/MaskedInput/MaskedInput.js
jcfilben: I looked through and I don't think there is

src/js/components/FocusedContainer.js
jcfilben: Yeah agree that would be cleaner, I'll do that

jcfilben: Fixed in latest commit

taysea: Is the 1px needed specifically to ensure it's "in view"?

jcfilben: Yeah, if I don't have the 1px it tries to scroll

taysea (approved):
I'm good with this approach! Let's just remove the storybook changes before merging

britt6612 (approved):
also just missing sign off

britt6612 (approved):

PR #7365, opened 2024-10-15
fixed #7323 issue List - when using keyboard, active list item isn't …

### LINKED ISSUE #7323: List - when using keyboard, active list item isn't always in view
### Expected Behavior

When using arrow keys to get to list item, would expect active list item to be in view on the screen.

### Actual Behavior

Active list item can end up being off screen but page doesn't scroll down.

https://storybook.grommet.io/?path=/story/visualizations-list-onclickitem--on-click-item-list

https://github.com/user-attachments/assets/b5ca4265-db00-4132-8cda-89b263866b01

### Steps to Reproduce

1. Make window smaller than length of list
2. Tab to List
3. Use arrow keys to move down

### Your Environment

- Grommet version: Latest (on stable of storybook)

vijaykr338:
The PR fixes the issue #7323, where using the keyboard to traverse the list items in the List would cause the active item to go off-screen.

#### What does this PR do?
In this PR, I simply attached a ref to each list item which would utilize the scrollIntoView method to scroll the screen as soon as the item which is not visible come into the view and become the active item.

Please check the src/js/components/List/List.js
It is the only file I worked on

The tests are failing, before this I was using the useEffect hook to achieve the same, but then I switched to using a function instead.

I have tested it on StoryBook

I don't think there is any

The tests related to List are failing


src/js/components/List/List.js
```diff
@@ -174,6 +174,18 @@ const List = React.forwardRef(
     // List items are likely selectable), active will be the
     // index of the item which is currently active.
     const [active, setActive] = useState();
+    const activeItemRef = useRef(null);
+
+    const setActiveItemRef = (element) => {
+      if (element) {
+        element.scrollIntoView({
```
jcfilben, line 181: Can we only scroll the element into view if we are navigating with a keyboard?
You can use `useKeyboard` from 'src/js/utils' to help determine if someone is navigating with a keyboard or if they are using a mouse.

Also can we check if the list item is fully visible or not before triggering the scrolIntoView function so that we don't call it if we don't need to

CONTRIBUTING.md
```diff
@@ -7,6 +7,9 @@ us make Grommet the best tool for streamlining the way you develop apps. You
 came to the right place to start your contribution! Follow the guidelines and
 let us know if we can help with anything else.
 
+Hello there this is vijay ifufbwudfbweufbweoufbweoifnewfnewofnw
+this is fix branch
+
```
jcfilben, line 12:
```suggestion
```

vijaykr338: sorry about this one

jcfilben:
Closing #7323 has been resolved by https://github.com/grommet/grommet/pull/7448

PR #7366, opened 2024-10-15
Fixes #7231 : Anchor - icon size

### LINKED ISSUE #7231: Anchor - icon size
### Expected Behavior
If the icon is not passed a size then the icon size should scale based on the Anchor size.
Potentially related to the `useSizedIcon` function in Grommet.

### Actual Behavior
The icon size does not scale with the Anchor size.

image

avantika-codes:
#### What does this PR do?

This PR fixes an issue in the Anchor component where the icon size was not properly scaling based on the Anchor size. It ensures that when the icon is not passed an explicit size, it will now scale in accordance with the size of the Anchor.

The reviewer should begin by looking at the logic related to how the icon size is determined, specifically the changes in the Anchor.js file. Pay close attention to the logic where we check for sizeProp or size, and how the useSizedIcon function is applied.

Manual testing was performed to verify that the icon correctly scales with the Anchor size, particularly for xsmall, which was initially not working. Debugging outputs were used to trace the size calculations. Tests were run for all standard sizes (small, medium, etc.) to ensure the fix works across the board.

Manually test by rendering the Anchor component with different size props, including:

- No size passed (default behavior)
- Explicit small, medium, and large sizes. Ensure the icon scales appropriately with the Anchor size and that the component behaves as expected when sizes are dynamically changed.

The main issue addressed is the incorrect scaling of the icon within the Anchor component when an explicit size isn’t provided.

Screenshot 2024-10-15 at 11 32 55 PM

No updates to the Grommet docs are necessary as this change fixes existing functionality rather than adding new features.

Yes, it should be mentioned as a bug fix: "Fixed an issue where icons in the Anchor component did not scale correctly with the Anchor size, including xsmall."

This change is backwards compatible, as it only fixes the internal logic for size scaling without affecting the existing API.

### DIFF
--- a/src/js/components/Anchor/Anchor.js
+++ b/src/js/components/Anchor/Anchor.js
@@ -66,16 +66,25 @@ const Anchor = forwardRef(
     }, [children, icon, label]);
 
     let coloredIcon = icon;
-    if (icon && !icon.props.color) {
-      coloredIcon = cloneElement(icon, {
-        color: normalizeColor(
-          color ||
-            theme.anchor?.icon?.color ||
-            theme.anchor?.size?.[sizeProp || size]?.color ||
-            theme.anchor.color,
-          theme,
-        ),
-      });
+    if (icon) {
+      if (!icon.props.color) {
+        coloredIcon = cloneElement(icon, {
+          color: normalizeColor(
+            color ||
+              theme.anchor?.icon?.color ||
+              theme.anchor?.size?.[sizeProp || size]?.color ||
+              theme.anchor.color,
+            theme,
+          ),
+        });
+
+        if (!icon.props.size) {
+          const scaledSize = sizeProp || size || 'medium'; // fallback size
+          coloredIcon = cloneElement(coloredIcon, {
+            size: theme.anchor?.size?.[scaledSize]?.iconSize || scaledSize,
+          });
+        }
+      }
     }
 
     const anchorIcon = useSizedIcon(coloredIcon, sizeProp || size, theme);
--- a/src/js/components/Anchor/__tests__/__snapshots__/Anchor-test.tsx.snap
+++ b/src/js/components/Anchor/__tests__/__snapshots__/Anchor-test.tsx.snap
@@ -920,8 +920,8 @@ exports[`Anchor renders size specific theming 1`] = `
 .c20 {
   display: inline-block;
   flex: 0 0 auto;
-  width: 24px;
-  height: 24px;
+  width: 48px;
+  height: 48px;
   fill: #7D4CDB;
   stroke: #7D4CDB;
 }


britt6612:
@jcfilben @taysea 
one thing to point out here is that Anchor supports size `xsmall` as well as `xxlarge` 
Im wondering if we should add a check if Anchor size is `xsmall` then use `small` icon as well as if Anchor size is `xxlarge` then use the `xlarge` icon. 

wanted to make sure we are aligned with what we want before we ask @avantika-codes to implement changes

britt6612:
@avantika-codes 
I think we agree that since the icons do not support the `xxlarge` and `xsmall` would you be able to add a check so that if Anchor size is `xsmall` then we use `small` icon size for now and same with `xxlarge`

avantika-codes:
> @avantika-codes I think we agree that since the icons do not support the `xxlarge` and `xsmall` would you be able to add a check so that if Anchor size is `xsmall` then we use `small` icon size for now and same with `xxlarge`

Yes sure, adding the rule.

src/js/components/Anchor/Anchor.js
```diff
@@ -66,16 +66,33 @@ const Anchor = forwardRef(
...
+          if (sizeProp === 'xsmall') {
+            iconSize = 'small';
+          } else if (sizeProp === 'xxlarge') {
+            iconSize = 'xlarge';
+          } else {
+            iconSize = theme.anchor?.size?.[scaledSize]?.iconSize || scaledSize;
+          }
+
+          coloredIcon = cloneElement(coloredIcon, { size: iconSize });
+        }
```
jcfilben, line 94: Should this block be outside of the `if (!icon.props.color)` block?

src/js/components/Anchor/Anchor.js
```diff
@@ -66,16 +66,33 @@ const Anchor = forwardRef(
...
+        if (!icon.props.size) {
+          const scaledSize = sizeProp || size || 'medium'; // fallback size
+
+          let iconSize;
+          if (sizeProp === 'xsmall') {
+            iconSize = 'small';
+          } else if (sizeProp === 'xxlarge') {
+            iconSize = 'xlarge';
+          } else {
+            iconSize = theme.anchor?.size?.[scaledSize]?.iconSize || scaledSize;
```
jcfilben, line 90: I'm not seeing that we have `theme.anchor.size.[scaledSize].iconSize` as an available theme property

avantika-codes:
@jcfilben sorry for the delay let me get this resolved by EOD today

src/js/components/Anchor/Anchor.js
britt6612: she has `scaledSize` defined on line 82 we can change naming or leave 

       ` const scaledSize = sizeProp || size || 'medium'; // fallback size`

also not sure if we need the `medium` for fall back but we can discuss

src/js/components/Anchor/Anchor.js
britt6612: moved

PR #7368, opened 2024-10-15
Added test for Drop trap focus

jcfilben:
#### What does this PR do?
Adds a test for https://github.com/grommet/grommet/pull/7324

no
no
backwards compatible

### DIFF
--- a/src/js/components/Drop/__tests__/Drop-test.tsx
+++ b/src/js/components/Drop/__tests__/Drop-test.tsx
@@ -18,6 +18,7 @@ import { expectPortal } from '../../../utils/portal';
 import { Grommet } from '../../Grommet';
 import { Drop, DropExtendedProps } from '..';
 import { ThemeType } from '../../../themes';
+import userEvent from '@testing-library/user-event';
 
 const customTheme = {
   global: {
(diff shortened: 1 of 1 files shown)


src/js/components/Drop/__tests__/Drop-test.tsx
```diff
@@ -282,6 +315,34 @@ describe('Drop', () => {
...
+    await userEvent.tab();
+
+    // Expect the focus to go to the second button
+    expect(document.activeElement).toBe(secondButton);
+
+    // Simulate tab
+    await userEvent.tab();
+
+    // Expect the focus to go back to the first button
+    expect(document.activeElement).toBe(firstButton);
```
britt6612, line 343: is there a way to include 
```
await userEvent.tab({ shift: true });

```
does that work just to check if after its then on the secondButton 

not super high need but would be nice to check

britt6612, line 343:
```suggestion
    expect(document.activeElement).toBe(firstButton);
    
     // Simulate backwards tab with Shift key
    await userEvent.tab({ shift: true });
    // Expect the focus to be on the second button again
    expect(document.activeElement).toBe(secondButton);
    
```

jcfilben: Good idea, added this in latest commit

britt6612 (approved):
LGTM

PR #7357, opened 2024-10-10
Calendar - 'space' key can be used to select dates

anvesh-singh:
#### What does this PR do?
solves Calendar - date should be selectable by pressing 'space' #7337
#### Where should the reviewer start?
calendar.js
#### What testing has been done on this PR?
passes all premade tests
#### How should this be manually tested?
by pressing space it selects the dates
#### Do Jest tests follow these best practices?

- [ ] `screen` is used for querying.
- [ ] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [ ] `asFragment()` is used for snapshot testing.

#### What are the relevant issues?
Closes #7337

PR #7369, opened 2024-10-16
Select and SelectMultiple accessibility improved

nishantdecode:
What does this PR do?
If the Select or SelectMultiple is in a FormField with a label and no placeholder text the label will be read out by a screenreader when the Select or SelectMultiple receives focus.

Since Select and SelectMultiple are complex components htmlFor doesn’t work as intended, so I added a aria-labelledby for Select and SelectMultiple to correctly associate to external label in the FormField.

FormField/stories/simple.js

Locally in the scenario where screenreader focuses on Select or SelecMultiple.

Create a Select component in a FormField with a label and no placeholder text. Then use a screenreader to focus on the Select Component

Do Jest tests follow these best practices?

What are the relevant issues?

Screenshots (if appropriate)
Before with incorrect association: 
Screenshot 2024-10-16 at 4 29 25 PM
 After with correct association:
Screenshot 2024-10-16 at 4 18 22 PM

no

no

backwards compatible


britt6612:
You will need the -s flag and a remediation commit. See https://github.com/grommet/grommet/pull/7357/checks?check_run_id=31375867254

package.json
```diff
@@ -55,6 +55,7 @@
   },
   "dependencies": {
     "@emotion/is-prop-valid": "^1.2.1",
+    "grommet": "^2.40.2",
```
britt6612, line 58: there shouldnt be any package.json changes

nishantdecode: @britt6612 
I have made changes in the latest commit so that there no changes to either, package.json or yarn.lock

src/js/components/Select/Select.js
```diff
@@ -294,6 +294,7 @@ const Select = forwardRef(
                 })
               : ''
           }`}
+          aria-labelledby={id ? `${id}__label` : undefined}
```
jcfilben, line 297: I think we only want to set aria-labelled if we are within a formfield, and if ariaLabel and a11yTitle are unidefined. We can check if we are within a formfield with `formContext.useFormField({}).inForm`

nishantdecode: @jcfilben 
I was testing the above and found that using formContext.useFormField({}).inForm would not work:

1. If the FormField itself is not wrapped by a Form as is the case in Simple.js(FormField stories Folder); but Select has a intended label using FormField label and htmlFor property.
2. Additionally if the Select is directly a child of Form and no FormField wrapped around Select, it would lead to aria-labelledby property on Select which is not associated to any external label.

Please let me know if these concerns are genuine, if they are I think the most effective way to solve it is by making a separate FormField context and wrapping it around FormField children.

nishantdecode: @jcfilben @britt6612 
I have now followed the second approach using prop passing to get the intended results, let me know if it is correct and if any other changes are needed.

src/js/components/SelectMultiple/SelectMultiple.js
```diff
@@ -124,6 +125,19 @@ const SelectMultiple = forwardRef(
...
     });
 
+    // inFormField is used to determine if this component is wrapped by
+    // FormField component and has label and htmlFor props defined,
+    // since htmlFor doesn't work as expected in the case of
+    // SelectMultiple. LabelId is the id referring to label element
+    // defined in FormField Component
+    const { inFormField, labelId } = formFieldProps;
+
+    // ariaLabelledBy determines if this component if this component
```
nishantdecode, line 135: wrote "if this component" twice by mistake, let me know if I should change it.

src/js/components/Select/Select.js
```diff
@@ -109,6 +110,19 @@ const Select = forwardRef(
...
     });
 
+    // inFormField is used to determine if this component is wrapped by
+    // FormField component and has label and htmlFor props defined,
+    // since htmlFor doesn't work as expected in the case of
+    // Select. LabelId is the id referring to label element
+    // defined in FormField Component
+    const { inFormField, labelId } = formFieldProps;
+
+    // ariaLabelledBy determines if this component if this component
```
nishantdecode, line 120: wrote "if this component" twice by mistake, let me know if I should change it.

src/js/components/Select/Select.js
jcfilben: > 1. If the FormField itself is not wrapped by a Form as is the case in Simple.js(FormField stories Folder); but Select has a intended label using FormField label and htmlFor property.
> 2. Additionally if the Select is directly a child of Form and no FormField wrapped around Select, it would lead to aria-labelledby property on Select which is not associated to any external label.

I think both of these cases fall outside of the expected use for Form and FormField so I would be okay with this. @britt6612 does this behavior sound okay to you too?

package.json
nishantdecode: @britt6612 is this conversation resolved for now?

src/js/components/Select/Select.js
MikeKingdom: > > 1. If the FormField itself is not wrapped by a Form as is the case in Simple.js(FormField stories Folder); but Select has a intended label using FormField label and htmlFor property.
> > 2. Additionally if the Select is directly a child of Form and no FormField wrapped around Select, it would lead to aria-labelledby property on Select which is not associated to any external label.
> 
> I think both of these cases fall outside of the expected use for Form and FormField so I would be okay with this. @britt6612 does this behavior sound okay to you too?

#2 is a valid use case and at least in that situation if the developer provides their own aria-labelledby property it needs to override this

nishantdecode: > > > 1. If the FormField itself is not wrapped by a Form as is the case in Simple.js(FormField stories Folder); but Select has a intended label using FormField label and htmlFor property.
> > > 2. Additionally if the Select is directly a child of Form and no FormField wrapped around Select, it would lead to aria-labelledby property on Select which is not associated to any external label.
> > 
> > 
> > I think both of these cases fall outside of the expected use for Form and FormField so I would be okay with this. @britt6612 does this behavior sound okay to you too?
> 
> #2 is a valid use case and at least in that situation if the developer provides their own aria-labelledby property it needs to override this

so should I go ahead with this?

jcfilben: Hey @nishantdecode were gonna think on this a little as a dev group and then get back to you on a direction. Thanks for your patience here!

jcfilben: Ok I think we are okay to use `formContext.useFormField({}).inForm`. Can we also include the original id in the aria-labelledby so that if a value is selected that will be read out in addition to the formfield label?
For example:
`${id}__label ${id}`

taysea:
I'm seeing this from VoiceOver -- is that expected? Would've expected to see "Size" 

Screenshot 2025-06-26 at 9 20 50 AM

britt6612:
> I'm seeing this from VoiceOver -- is that expected? Would've expected to see "Size"
> 
> Screenshot 2025-06-26 at 9 20 50 AM

@taysea 
This is because of the story 
we have 
```
        <FormField label="Size" htmlFor="size" name="size">
            <Select
              id="size"
              name="size"
              options={['small', 'medium', 'large']}
            />
          </FormField>
```

and in my code for it to work we need to add the `__input` 

```
       <FormField label="Size" htmlFor="size__input" name="size">
            <Select
              id="size"
              name="size"
              options={['small', 'medium', 'large']}
            />
          </FormField>
```

@taysea 
EDIT I have added `__input` to our examples so this should read out correctly now

src/js/components/SelectMultiple/SelectMultiple.js
```diff
@@ -126,6 +127,10 @@ const SelectMultiple = forwardRef(
       initialValue: defaultValue || '',
     });
 
+    let ariaLabelledBy;
+    if (formContext?.useFormField({})?.inForm && id && !ariaLabel) {
+      ariaLabelledBy = `${id}__input__label`;
```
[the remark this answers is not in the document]
britt6612: we need the `__input` 
     The id of the input element contained in this field. If the input
            element is Select or SelectMultiple, append "__input" to the id.

src/js/components/Select/Select.js
```diff
@@ -284,6 +285,11 @@ const Select = forwardRef(
 
     const iconColor = getIconColor(theme);
 
+    let ariaLabelledBy;
+    if (formContext?.useFormField({})?.inForm && id && !ariaLabel) {
+      ariaLabelledBy = `${id}__input__label`;
```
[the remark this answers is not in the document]
britt6612: The id of the input element contained in this field. If the input
            element is Select or SelectMultiple, append "__input" to the id.

src/js/components/Select/Select.js
```diff
@@ -284,6 +285,11 @@ const Select = forwardRef(
 
     const iconColor = getIconColor(theme);
 
+    let ariaLabelledBy;
+    if (formContext?.useFormField({})?.inForm && id && !ariaLabel) {
```
jcfilben: I think we wanted to check if there was a placeholder

src/js/components/Select/Select.js
jcfilben: I think we also wanted to try including the original id, see comment https://github.com/grommet/grommet/pull/7369#discussion_r1815792748

britt6612: added

src/js/components/Select/Select.js
britt6612: added

jcfilben (approved):
Looks good!

src/js/components/Select/Select.js
```diff
@@ -284,6 +285,16 @@ const Select = forwardRef(
...
     const iconColor = getIconColor(theme);
 
+    let ariaLabelledBy;
+    if (
+      formContext?.useFormField({})?.inForm &&
+      id &&
+      !ariaLabel &&
+      !placeholder
+    ) {
+      ariaLabelledBy = `grommet-${id}__input__label ${id}`;
```
taysea: How come we need both the grommet prefixed one and standalone id?

britt6612: from Jessica comment above
```
Can we also include the original id in the aria-labelledby so that if a value is selected that will be read out in addition to the formfield label?
For example:
${id}__label ${id}
```
so when value is there it gets read out

taysea: @jcfilben @britt6612 Sorry for the last minute review comment, but was wondering why it's `grommet-${id}__input__label` and not just `grommet-${id}__label`?

britt6612: because of the `_input` 
https://v2.grommet.io/formfield#htmlFor

src/js/components/FormField/FormField.js
```diff
@@ -597,6 +597,31 @@ const FormField = forwardRef(
...
+    if (htmlFor) {
+      let isSelectComponent = false;
+
+      // Check if children contain Select or SelectMultiple
+      if (children) {
+        Children.forEach(children, (child) => {
+          if (
+            child &&
+            child.type &&
+            (child.type.displayName === 'Select' ||
```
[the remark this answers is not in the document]
britt6612: since were already doing this in other places we decided it was valid for this

src/js/components/Select/Select.js
```diff
@@ -284,6 +285,16 @@ const Select = forwardRef(
 
     const iconColor = getIconColor(theme);
 
+    let ariaLabelledBy;
+    if (
+      formContext?.useFormField({})?.inForm &&
+      id &&
+      !ariaLabel &&
+      !placeholder
+    ) {
```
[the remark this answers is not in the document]
britt6612: changed

src/js/components/SelectMultiple/SelectMultiple.js
```diff
@@ -126,6 +127,16 @@ const SelectMultiple = forwardRef(
       initialValue: defaultValue || '',
     });
 
+    let ariaLabelledBy;
+    if (
+      formContext?.useFormField({})?.inForm &&
+      id &&
+      !ariaLabel &&
+      !placeholder
+    ) {
```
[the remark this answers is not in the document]
britt6612: changed

src/js/components/FormField/FormField.js
```diff
@@ -597,6 +597,31 @@ const FormField = forwardRef(
...
+              child.type.displayName === 'SelectMultiple')
```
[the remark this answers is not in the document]
britt6612: leaving as is since we compare displayName

src/js/components/Select/Select.js
```diff
@@ -284,6 +286,11 @@ const Select = forwardRef(
 
     const iconColor = getIconColor(theme);
 
+    let ariaLabelledBy;
```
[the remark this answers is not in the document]
britt6612: not going to implement

src/js/components/FormField/FormField.js
```diff
@@ -597,6 +597,31 @@ const FormField = forwardRef(
...
+          ) {
+            isSelectComponent = true;
+          }
+        });
+      }
+
+      // If it's a Select component and htmlFor doesn't end with __input, add it
+      if (isSelectComponent && !htmlFor.endsWith('__input')) {
+        adjustedHtmlFor = `${htmlFor}__input`;
```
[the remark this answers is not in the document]
britt6612: we just need the `__input` string

src/js/components/SelectMultiple/SelectMultiple.js
```diff
@@ -126,6 +128,11 @@ const SelectMultiple = forwardRef(
       initialValue: defaultValue || '',
     });
 
+    let ariaLabelledBy;
+    if (formFieldData?.inForm && id && !ariaLabel && !placeholder) {
+      ariaLabelledBy = `grommet-${id}__input__label ${id}`;
```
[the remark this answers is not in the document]
britt6612: we dont need the grommet prefix in both places

britt6612:
> Code looks good just wondering if we can add jest tests for Select and SelectMultiple

cool added those!

jcfilben (approved):
Looks good!

src/js/components/FormField/FormField.js
```diff
@@ -597,6 +597,31 @@ const FormField = forwardRef(
...
+            (child.type.displayName === 'Select' ||
+              child.type.displayName === 'SelectMultiple')
+          ) {
+            isSelectComponent = true;
+          }
+        });
+      }
+
+      // If it's a Select component and htmlFor doesn't end with __input, add it
+      if (isSelectComponent && !htmlFor.endsWith('__input')) {
```
taysea, line 621: I like the benefit for standard Select/SelectMultiple usage, that it ensures FormField label gets properly associated with the input. Is there any potential concern of people expecting their htmlFor to translate to the DOM exactly as they define it? Is there a risk we'd break something for queries or tests people might already have set up?

I think if we move forward with this, we should clearly define this manipulation in our docs.

@MikeKingdom @britt6612 @jcfilben @halocline

src/js/components/Select/Select.js
```diff
@@ -284,6 +286,11 @@ const Select = forwardRef(
...
+    if (formFieldData?.inForm && id && !ariaLabel && !placeholder) {
+      ariaLabelledBy = `grommet-${id}__input__label ${id}`;
```
taysea: If someone has a custom Select, our FormField code won't add the `__input` so I think we should make sure this stays in alignment with the `htmlFor` of the formfield it's in.

For example, the htmlFor on this won't be updated by our internal logic, so we'd be creating a broken aria-labelledby:

```

const MyCustomSelect = ({...rest}) => <Select {...rest} />
...

<FormField label="Label" htmlFor="myselect">
   <MyCustomSelect id="myselect" options={[...]} />
</FormField>
```

src/js/components/FormField/FormField.js
MikeKingdom: I don't think we should be simply checking for something ending in `__input`. We'd want to check for the exact id of the select  + `__input`.

src/js/components/Select/Select.js
britt6612: talked about this offline and decided to add a check for `document.getElementById(`grommet-${id}__input__label`,);` before adding the ariaLabelledBy

src/js/components/SelectMultiple/SelectMultiple.js
```diff
@@ -126,6 +128,19 @@ const SelectMultiple = forwardRef(
       initialValue: defaultValue || '',
     });
 
+    const [ariaLabelledBy, setAriaLabelledBy] = useState();
```
taysea, line 131: Any chance we could do this as a useMemo instead of a setState + useEffect to avoid extra re-render from the useEffect?

britt6612: good call!

britt6612: Im thinking actually `useMemo` is running to early then the `document.getElementById` is coming back as null so then the `ariaLabelledBy` doesnt get set

britt6612: from co-pilot

I recommend Option 1 (going back to useEffect) because:

The DOM query needs to happen after the label element is rendered
FormField renders the label, then the Select component
useEffect ensures the DOM is ready
The extra re-render is minimal compared to the complexity of timing the DOM query correctly

taysea: Good point. Even though it would be nice to rely on more React things like refs/props rather than document query selectors I'm okay with the tradeoffs given the current architecture and agree this feels more lightweight than introducing a new context or trying to pass a ref property down to children.

taysea: It feels like we'd want to rethink a variety of things in FormField logic in the future, but this at least feels like an improvement on what we have currently.

taysea: We can see if Mike has any remaining concerns or if he's okay with the approach.

src/js/components/Form/stories/FieldWithChildren.stories.js
```diff
@@ -88,7 +88,7 @@ export const FieldWithChildren = () => (
             options={['morning', 'evening']}
           />
         </FormField>
-        <FormField label="Size" htmlFor="size" name="size">
+        <FormField label="Size" htmlFor="size__input" name="size">
```
taysea: Just to confirm, these story changes aren't required for this functionality (because now grommet would add the __input automatically), but we're adding it since we document as a best practice. Correct?

taysea (approved):
I'm comfortable with these changes since I think they improve the accessibility experience. Although I do look forward to when we can refactor this component to better handle the custom input cases.

src/js/components/Form/stories/FieldWithChildren.stories.js
britt6612: exactly.

jcfilben (approved):

src/js/components/Form/stories/FieldWithChildren.stories.js
MikeKingdom: If we don't require it now, why perpetuate the complexity in examples?

src/js/components/FormField/FormField.js
MikeKingdom: This is still just looked at a part of the htmlFor. Wouldn't it be much more accurate and safer to make sure the htmlFor is referring to Select ID? What if the formfield has a few children, only one of which is the Select/SelectMultiple and that htmlFor is trying to refer to one of the other children?

src/js/components/SelectMultiple/SelectMultiple.js
MikeKingdom: Seems fair for now. We do need to schedule some time to look at FormField and how it's pretty inflexible to custom children.

src/js/components/FormField/FormField.js
britt6612: we would have to dig into the `child.props` of this `isSelectComponent` we felt like that was getting a little messy and wasnt sure if we need to do that. 

what you mean if FormField has a few children? like this?
```
<Formfield>
<CheckBox/>
<Select/>
</Formfield>
``
???

MikeKingdom: Basically yes. One of the examples for Morpheus was to create a custom formfield that was a textinput for typing in a size plus a select for setting the units (bytes/kbytes...)

src/js/components/Form/stories/FieldWithChildren.stories.js
britt6612: thats a good point. Since we add automatically now should we change the docs? or still show this in the stories?
@taysea @jcfilben

src/js/components/FormField/FormField.js
britt6612: hmm yeah you got me there.. in that case we would have to look into the childs.props and the `id` and match them.

src/js/components/Form/stories/FieldWithChildren.stories.js
jcfilben: We should update the docs so that people understand that we add this automatically. I could go either way on the stories, don't have too strong of an opinion there. One one hand it is nice to show the `__input` so that people hopefully understand that concept better. But I could see how it adds unnecessary complexity

britt6612: yeah I wonder to keep it clean we remove in the story if we update docs

britt6612: removed the `__input` in stories

src/js/components/FormField/FormField.js
jcfilben: I'm curious how Morpheus is handling their FormField label linking for that case. Do they just link the label to one of the fields or try to link it to both? I'm not sure what the best practice is there to make 2 input fields within one FormField accessible. Maybe we could do some research on that?

MikeKingdom: I'm not how they are handling this today. I think they've avoided Grommet so far in a small part due to how they felt it was hard to do custom form fields.

But give `htmlFor` basically makes the browser read that label for a value but also allows clicks on the label to click the related field (in some cases at least, especially checkbox/radiobutton) I'd say in this case it would be set to the text input id. Really for screen reader purposes they'd want to set an aria label that includes the current value plus units.

src/js/components/SelectMultiple/SelectMultiple.js
britt6612: okay I added the check to see if the child `id` matches the `htmlFor`

src/js/components/FormField/FormField.js
britt6612: @MikeKingdom okay I added the check to see if the child `id` matches the `htmlFor`

PR #7376, opened 2024-10-18
Fix a11y aria-label for select

NOMADE55:
#### What does this PR do?
This PR adds a new function `formatValueForA11y` which uses both the value of the select option(s) and labelKey to extract a Human Readable value for the `aria-label` attribute. [9f986298af444e52777c404e99852fd008f6bd89]
~~Also, as a plus, Button tests were breaking. Updating the snapshot this was properly fixed. [3d5e88750fe4aa53decc992b03b1ba21281634d6]~~

- Start in the context that runs the replacements over at: [src/js/contexts/MessageContext/MessageContext](https://github.com/grommet/grommet/blob/9f986298af444e52777c404e99852fd008f6bd89/src/js/contexts/MessageContext/MessageContext.js)
- Then check `formatValueForA11y` at the [select utils](https://github.com/grommet/grommet/blob/9f986298af444e52777c404e99852fd008f6bd89/src/js/components/Select/utils.js)

- Used npm link to test the components in a test project
- Used Snapshots/Jest to test the output of the components

Using `Select` Component:
- Add options which contain objects
- Use `multiple` prop and select more than one value

- Fixes https://github.com/grommet/grommet/issues/6393

No
Not sure
Yes

### DIFF
--- a/src/js/contexts/MessageContext/MessageContext.js
+++ b/src/js/contexts/MessageContext/MessageContext.js
@@ -1,6 +1,5 @@
 import React from 'react';
 
-import { formatValueForA11y } from '../../components/Select/utils';
 import defaultMessages from '../../languages/default.json';
 
 // options:
@@ -40,13 +39,13 @@ export const format = (options, messages) => {
     messageObj ||
     options.defaultMessage;
 
-  const { labelKey, values } = options;
+  const { values } = options;
 
   let newMessage = message;
   const tokens = message?.match(/\{(.+?)\}/g);
   tokens?.forEach((token) => {
     const names = token.substr(1, token.length - 2);
-    const value = formatValueForA11y(values[names], labelKey);
+    const value = values[names];
     newMessage = newMessage.replace(token, value);
   });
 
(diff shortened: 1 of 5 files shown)


NOMADE55:
This is my first commit here, please, let me know if I naming is ok, or something else is needed.

src/js/contexts/MessageContext/MessageContext.js
```diff
@@ -39,13 +40,13 @@ export const format = (options, messages) => {
     messageObj ||
     options.defaultMessage;
 
-  const { values } = options;
+  const { labelKey, values } = options;
```
jcfilben, line 43: The format function is used by a lot of components in Grommet besides just Select and SelectMultiple. These other components don't all have the concept of a `labelKey` prop. Can we explore changes that are more localized to the Select and SelectMultiple cases so that we don't have unintended changes in other components?

NOMADE55: Sounds reasonable, I'll see what I can do!

NOMADE55:
@jcfilben I've moved the formatting logic to the Select component,

jcfilben (approved):
Looks good!

MikeKingdom (approved):
LGTM

Issue #7375, opened 2024-10-17
Pagination - <ul> and <ol> must only directly contain <li>, <script> or <template> elements


jcfilben:
### Expected Behavior
The `<ul>` element in Pagination should only have children that are `<li>`, `<script>` or `<template>` elements

### Actual Behavior
The `<ul>` element in the Pagination component has `<div>` elements as children

### URL, screen shot, or Codepen exhibiting the issue
https://storybook.grommet.io/?path=/story/controls-pagination-simple--simple

### Steps to Reproduce
Look at the accessibility panel on the above story

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

PR #7379, opened 2024-10-19
Fix using div gaps inside pagination lists

NOMADE55:
#### What does this PR do?
Using the prop `cssGap`, I avoid using the StyledBoxGap element to add gaps between the elements. 
This was as per the Box component comment: 'an internal prop'.

[Pagination Component](https://github.com/grommet/grommet/blob/4a970d1eb0a176e072910d251f300d719e52552b/src/js/components/Pagination/Pagination.js)
Is also worth checking if cssGap should be added to the declarations file and docs.

- Used npm link to test the components in a test project
- Used Snapshots/Jest to test the output of the components

- Measure Pagination before/after pulling the branch. **Shouldn't have changed.**
- Avoid any extra customization to the component to get a proper measurement

- Fixes https://github.com/grommet/grommet/issues/7375

|before *Before*|
|:--:|
|after *After*|

### DIFF
--- a/src/js/components/Data/__tests__/__snapshots__/Data-test.tsx.snap
+++ b/src/js/components/Data/__tests__/__snapshots__/Data-test.tsx.snap
@@ -1814,12 +1814,7 @@ exports[`Data pagination 1`] = `
   min-height: 0;
   flex-direction: row;
   padding: 0px;
-}
-
-.c15 {
-  flex: 0 0 auto;
-  align-self: stretch;
-  width: 3px;
+  column-gap: 3px;
 }
 
 .c9 {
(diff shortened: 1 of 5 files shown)


jcfilben (approved):
Looks good!

jcfilben:
> Is also worth checking if cssGap should be added to the declarations file and docs.

Good question, it looks like we have it documented on the site https://v2.grommet.io/box#cssGap, but we do not have it in the Box index.d.ts or propTypes.js files. I'll file an issue to add it to those places

MikeKingdom (approved):
LGTM

PR #7394, opened 2024-10-23
Fixed the DisabledKey Type in Select Component

TheThinker277:
#### What does this PR do?
I have defined the Type of DisabledKey Prop in the Select Component. I also have defined the the type as custom, so in future if we want to support more/different types, we can easily integrate the Type.
Yes.

I checked the Util.js file and SelectContainer.js file for checking the Types of the disabledKey.
No, I don't think so.
No


src/js/components/Select/index.d.ts
```diff
@@ -7,14 +7,14 @@ import {
   MarginType,
   PlaceHolderType,
 } from '../../utils';
-
+type disabledKeyType = string | undefined | object | boolean | Function;
```
jcfilben, line 10: Looking at the docs it doesn't look like we support object or boolean for this prop

TheThinker277: In the File \src\js\components\Select\utils.js, we have a function applyKey in which the second argument we are passing is the disabledkey and there we are checking if the typeof Key === 'object'. That is why I have added a type of object.

For the Boolean Type, I made the same mistake of thinking that as we are using it in if condition it can be boolean

I have made the relevant changes. Thanks

TheThinker277:
@jcfilben , Could you please merge if no issues?
Thanks

src/js/components/Select/index.d.ts
```diff
@@ -7,14 +7,14 @@ import {
   MarginType,
   PlaceHolderType,
 } from '../../utils';
-
+type disabledKeyType = string | undefined | object | Function;
```
britt6612, line 10: why did we include `object` here?

britt6612:
closing due to inactivity feel free to reopen

src/js/components/Select/index.d.ts
TheThinker277: In the File \src\js\components\Select\utils.js, we have a function applyKey in which the second argument we are passing is the disabledkey and there we are checking if the typeof Key === 'object'. That is why I have added a type of object.

TheThinker277: @britt6612 , Can you please reopen this?

PR #7395, opened 2024-10-23
Defined the Type of OnMore in Select Component

TheThinker277:
#### What does this PR do?
I have defined the Type of OnMore Prop in the Select Component.
Yes

I checked the infiniteScroll file and got the usage of OnMore in the SelectComponent. I have defined the Type which supports this usage.
No, I don't think so.
No


src/js/components/Select/index.d.ts
```diff
@@ -34,7 +34,7 @@ export interface BasicSelectProps {
   name?: string;
   onChange?: (...args: any[]) => void;
   onClose?: (...args: any[]) => any;
-  onMore?: (...args: any[]) => any;
+  onMore?: () => void | boolean;
```
jcfilben, line 37: I would expect onMore to be a function

TheThinker277: I originally thought that onMore could be a boolean because it's used within an useEffect hook in the src\js\components\InfiniteScroll\InfiniteScroll.js file. However, I have realized that onMore is actually declared as a function. This means that onMore must be a function in order for the useEffect hook to execute the code block within the if condition.

Thanks.

TheThinker277:
@jcfilben en , Could you please merge if no issues?
Thanks

jcfilben (approved):
Looks good!

britt6612 (approved):

PR #7398, opened 2024-10-23
Tag — Apply default size of medium

taysea:
#### What does this PR do?

Tag should fallback to size "medium" so that anything in `tag.size.medium` automatically applies. Currently when `size` is not passed directly by caller, the "medium" object won't apply even though "medium" technically the default size.

Note: Snapshot change is expected because it's applying the "medium" sizing to a Tag that doesn't have size explicitly defined via prop.

src/js/components/Tag/Tag.js

Locally.

No.

Yes.

Backwards compatible.


MikeKingdom (approved):
LGTM

halocline (approved):

Issue #7341, opened 2024-10-07
Select - clear selection + search keyboard interactions


jcfilben:
### Expected Behavior
The clear and search should be accessible via keyboard in the Select component

### Actual Behavior
When you have both clear and search in a select the keyboard behavior doesn't work very well. If you arrow down to clear you can't get back up to search.

### URL, screen shot, or Codepen exhibiting the issue
https://codesandbox.io/p/sandbox/grommet-v2-template-forked-jwn8q2?file=%2Findex.js%3A7%2C24

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

miqh:
@jcfilben,

I've got an idea of how to address this issue, but first I'd ask for confirmation on how keyboard navigation is expected to behave for the Select component.

At present, with the Select component in focus:

- When both clear (top position) and search are enabled, <kbd>Down</kbd> focuses search, another <kbd>Down</kbd> focuses the first option. Is skipping clear desirable here? (i.e. Potential counter-intuitiveness that <kbd>Down</kbd> navigation skips clear, but <kbd>Up</kbd> navigation won't)
- When both clear (bottom position) and search are enabled, <kbd>Down</kbd> while on the last option won't move the focus to clear, so the only way to get there is via <kbd>Tab</kbd>. Is this current behaviour desirable to retain?

jcfilben:
> When both clear (top position) and search are enabled, Down focuses search, another Down focuses the first option. Is skipping clear desirable here? (i.e. Potential counter-intuitiveness that Down navigation skips clear, but Up navigation won't)

I don't think we should skip the clear button here. I think the ideal behavior is that down arrow will focus search and then pressing down arrow again will focus the clear button.

> When both clear (bottom position) and search are enabled, Down while on the last option won't move the focus to clear, so the only way to get there is via Tab. Is this current behaviour desirable to retain?

No, I don't think the current behavior is correct. If we press down arrow while on the last option I would expect to move to the clear option. However we should maintain the behavior of pressing tab will also move to the clear button.

jcfilben:
We should use tab only (not arrow keys) to move to and from the search

PR #7399, opened 2024-10-24
Fix keyboard navigation for Select with clear and search (#7341)

miqh:
#### What does this PR do?

It addresses incomplete keyboard navigation support for Select when both `clear` and `onSearch` are configured.

Noteworthy changes are concentrated in `SelectContainer`.

I've added two additional test cases that emulate vertical navigation through constituents of the Select dropdown configured with both `clear` and `onSearch`.

Modifying the "Search" story for Select to add `clear` for tinkering is an expedient option. I did this myself as I was trying to find fringe cases that would break my proposed changes.

At first I had a trivial fix in mind, but it wasn't until I started trying to break it that I discovered Select configurations involving disabled options introduced some complications.

#7341

`onSearch` + `clear` bottom position + disabled boundary options:

`onSearch` + `clear` top position + disabled boundary options:

`clear` top position:

`clear` bottom position:

`clear` mouse over focus:

Maintainers to decide.

Maintainers to decide.

It shouldn't be a breaking change.

### DIFF
--- a/src/js/components/Select/SelectContainer.js
+++ b/src/js/components/Select/SelectContainer.js
@@ -136,7 +136,12 @@ const SelectContainer = forwardRef(
         const clearButton = clearRef.current;
         if (onSearch) {
           const searchInput = searchRef.current;
-          if (searchInput && searchInput.focus && !activeRef.current) {
+          if (
+            searchInput &&
+            searchInput.focus &&
+            !activeRef.current &&
+            document.activeElement !== clearButton
+          ) {
             setFocusWithoutScroll(searchInput);
           }
         } else if (
@@ -246,39 +251,66 @@ const SelectContainer = forwardRef(
         ) {
           nextActiveIndex += 1;
         }
-        if (nextActiveIndex !== options.length) {
+
+        const searchInput = searchRef.current;
+        const clearButton = clearRef.current;
+        if (
+          // Expect search input to immediately precede the clear button
+          ((shouldShowClearButton('top') &&
+            document.activeElement === searchInput) ||
+            // Expect clear button to immediately follow the last
+            // selectable option
+            (shouldShowClearButton('bottom') &&
+              nextActiveIndex === options.length)) &&
+          clearButton &&
+          clearButton.focus
+        ) {
+          setActiveIndex(nextActiveIndex);
+          setKeyboardNavigation(true);
+          setFocusWithoutScroll(clearButton);
+        } else if (nextActiveIndex < options.length) {
           setActiveIndex(nextActiveIndex);
           setKeyboardNavigation(true);
         }
       },
-      [activeIndex, options, isDisabled],
+      [activeIndex, options, isDisabled, shouldShowClearButton],
     );
 
     const onPreviousOption = useCallback(
       (event) => {
         event.preventDefault();
         let nextActiveIndex = activeIndex - 1;
-
-        if (nextActiveIndex === -1) {
-          const searchInput = searchRef.current;
-          const clearButton = clearRef.current;
-          if (
-            clearButton &&
-            clearButton.focus &&
-            shouldShowClearButton('top')
-          ) {
-            setActiveIndex(nextActiveIndex);
-            setFocusWithoutScroll(clearButton);
-          } else if (searchInput && searchInput.focus) {
-            setActiveIndex(nextActiveIndex);
-            setFocusWithoutScroll(searchInput);
-          }
-        }
-
         while (nextActiveIndex >= 0 && isDisabled(nextActiveIndex)) {
           nextActiveIndex -= 1;
         }
-        if (nextActiveIndex >= 0) {
+
+        const searchInput = searchRef.current;
+        const clearButton = clearRef.current;
+        if (
+          // Expect search input to immediately precede the clear button
+          ((shouldShowClearButton('top') &&
+            document.activeElement === clearButton) ||
+            // Expect search input to immediately precede the first
+            // selectable option
+            (!shouldShowClearButton('top') && nextActiveIndex === -1)) &&
+          searchInput &&
+          searchInput.focus
+        ) {
+          setActiveIndex(-1);
+          setKeyboardNavigation(true);
+          setFocusWithoutScroll(searchInput);
+        } else if (
+          // Expect clear button to immediately precede the first
+          // selectable option
+          shouldShowClearButton('top') &&
+          nextActiveIndex === -1 &&
+          clearButton &&
+          clearButton.focus
+        ) {
+          // Clear buttons already have focus handlers that set the active index
+          setKeyboardNavigation(true);
+          setFocusWithoutScroll(clearButton);
+        } else if (nextActiveIndex >= 0) {
           setActiveIndex(nextActiveIndex);
           setKeyboardNavigation(true);
         }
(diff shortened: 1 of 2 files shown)


src/js/components/Select/SelectContainer.js
```diff
@@ -136,7 +136,12 @@ const SelectContainer = forwardRef(
         const clearButton = clearRef.current;
         if (onSearch) {
           const searchInput = searchRef.current;
-          if (searchInput && searchInput.focus && !activeRef.current) {
+          if (
+            searchInput &&
+            searchInput.focus &&
+            !activeRef.current &&
+            document.activeElement !== clearButton
```
miqh, line 143: This additional condition is intended to prevent an unwanted refocus of the search input when the Select dropdown is triggered by a click, then a user initiates keyboard navigation.

Without it, we'd be back at a similar situation I fixed in #6982.

Only `SelectOption` elements are considered for `activeRef`.

src/js/components/Select/SelectContainer.js
```diff
@@ -246,39 +251,66 @@ const SelectContainer = forwardRef(
...
+        const searchInput = searchRef.current;
+        const clearButton = clearRef.current;
+        if (
+          // Expect search input to immediately precede the clear button
+          ((shouldShowClearButton('top') &&
+            document.activeElement === searchInput) ||
+            // Expect clear button to immediately follow the last
+            // selectable option
+            (shouldShowClearButton('bottom') &&
+              nextActiveIndex === options.length)) &&
```
miqh, line 264: I tried to not stray far from the design of having an active index of `-1` represent a special boundary position for the options collection. Here, `options.length` represents the opposite boundary position.

Whenever navigation needs to take place to or from boundary indices, the idea is to use the focussed element to inform the next keyboard navigation destination. I did this to avoid wrangling checks involving index offsets.

jcfilben (commented):
Behavior looks good when the clear select is positioned at the top of the drop, but when the clear select is positioned at the bottom of the drop and it gets hovered or focused the drop options are scrolled all the way to the bottom. Can we prevent this scroll from happening?

src/js/components/Select/SelectContainer.js
```diff
@@ -376,6 +408,7 @@ const SelectContainer = forwardRef(
                   setActiveIndex(-1);
                   onSearch(nextSearch);
                 }}
+                onFocus={() => setActiveIndex(-1)}
```
miqh, line 411: This was something I inadvertently missed in my original changeset.

Without it, after cycling focus back to the search input using the keyboard, the down key won't navigate focus down to the first option in the list as one might expect.

src/js/components/Select/SelectContainer.js
```diff
@@ -403,7 +442,11 @@ const SelectContainer = forwardRef(
                 step={theme.select.step}
                 onMore={onMore}
                 replace={replace}
-                show={activeIndex !== -1 ? activeIndex : undefined}
+                show={
+                  ![-1, options.length].includes(activeIndex)
```
miqh, line 446: The `show` prop was responsible for undesirable auto-scroll to the end of the list of options when focus was applied to the clear button in bottom position.

Like the search input, it should be exempt from the behaviour.

src/js/components/Select/__tests__/Select-test.js
```diff
@@ -1709,6 +1715,208 @@ describe('Select', () => {
     expect(getAllByRole('option')[0]).toHaveFocus();
   });
 
+  test('allows keyboard navigation with search and top clear', async () => {
+    // Neither user.type() nor user.keyboard() trigger arrow keys as needed
```
miqh, line 1719: Specifically, I noticed events generated from either would lack both `keyCode` and `which`, which the Keyboard implementation relies on for routing callbacks.

src/js/components/Select/__tests__/Select-test.js
```diff
@@ -1709,6 +1715,208 @@ describe('Select', () => {
...
+    upKey();
+    expect(options[0]).toHaveFocus();
+    upKey();
+    expect(search).toHaveFocus();
+    upKey();
+    // Focus should stay on the search input
+    expect(search).toHaveFocus();
+  });
+
+  test('does not scroll on search or bottom clear focus', async () => {
```
miqh, line 1855: This test wasn't part of the original changeset and I've added it to cover the case concerning the undesirable scrolling related to the clear button focus.

src/js/components/Select/__tests__/Select-test.js
```diff
@@ -1709,6 +1715,208 @@ describe('Select', () => {
...
+
+    const search = screen.getByPlaceholderText('test search');
+    const clear = screen.getByLabelText(/Clear selection/);
+
+    await waitFor(() => expect(search).toHaveFocus());
+
+    // Simulate focus shift via tab key
+    fireEvent.focus(clear);
+
+    expect(screen.queryByText('item1000')).not.toBeInTheDocument();
```
miqh, line 1906: At first, I had assertions to expect the first option (i.e. `item1`) would no longer be present in the DOM because of how InfiniteScroll does paging. However, I suspect limitations with JSDOM prevents dimension calculations from happening properly, resulting in no paging at all.

miqh:
@jcfilben,

I put aside some more time to look more closely at the behaviour you pointed out with the undesirable scrolling.

Turns out that it was a case that I didn't actively test for while making my changes (i.e. I wasn't testing with enough options to get a scrollbar appearing).

In fixing the keyboard navigation behaviour, I identified what I thought was a bug with how InfiniteScroll treated the show prop.

Even without all my changes here, I found that once navigation was far enough along a list of options, and the clear in top position was put into focus again via the tab key, pressing down once won't scroll to reveal the first option. Pressing down another time will scroll to reveal the second option.

If the above is actually intentional, I can revert my changes to InfiniteScroll. However, in doing so, beware that there will be an asymmetry in the scroll to reveal behaviour (e.g. clear in bottom position).

Hopefully I've covered it this time.





---

As an aside, I noticed another potentially unexpected behaviour with cycling focus with the tab key when the dropdown is open.

If a selected option is in view, focus should be able to cycle directly onto the option. If it's out of view (i.e. outside of InfiniteScroll's current page), focus only cycles between the search and clear elements, or whatever is available.

One might argue that if an option is selected:
- The dropdown should focus and scroll into view this option when it's opened.
- Cycling focus using the tab key should always focus the selected option, even if it's outside the current frame of results.

If this is an issue, I think it should be raised as a separate matter, rather than burdening this changeset with more.

jcfilben (approved):
Nice work, this looks good to me!
Thanks for adding tests

src/js/components/Select/SelectContainer.js
```diff
@@ -397,7 +430,13 @@ const SelectContainer = forwardRef(
               name={name}
               onClear={onClear}
               onFocus={() => setActiveIndex(-1)}
-              onMouseOver={() => setActiveIndex(-1)}
+              onMouseOver={() => {
+                setActiveIndex(-1);
+                const clearButton = clearRef.current;
+                if (clearButton && clearButton.focus) {
+                  clearButton.focus();
```
taysea, line 437: @jcfilben doing a programmatic focus here as a result of mousing over feels a little bit strange to me.

jcfilben: But agree it is a little odd, maybe we can explore another way to prevent the above behavior

MikeKingdom (approved):
LGTM

Issue #7374, opened 2024-10-17
Select - nested interactive elements


jcfilben:
### Expected Behavior
Automated accessibility tests shouldn't flag Select component as violating the nested interactive rule.

### Actual Behavior
We have a button with an input inside of it in the select and component. The input is needed to pass a value to a Form and the button is needed to open the select dropdown. Even though the input has tabIndex="-1" automated accessibility tools will still flag this as a nested interactive element. See more details here https://github.com/dequelabs/axe-core/issues/3227.

We could potentially explore a solution where we have a hidden input that is outside of the button, use this for the Form. Then just have text within the button where the input previously was to give the visual indication of what has been selected.

### URL, screen shot, or Codepen exhibiting the issue
https://storybook.grommet.io/?path=/story/input-select-simple--simple

### Steps to Reproduce

Go to the above story and look at the accessibility panel

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

NOMADE55:
I've been investigating this. I've found with a roadblock that might need some extra thought.
Grommet relies on user-agent's stylesheet min width for select inputs, as we do not define a size by default.

This doesn't fly with a span or div replacing the input's place., where we're going to need an specific value size if we want the content to fill as much as the input did previously.
![image](https://github.com/user-attachments/assets/ad23aec7-12c3-4eb0-baa3-c565de80a78e)

jcfilben:
Good comments @NOMADE55, agree the input width makes this a bit tricky. Not sure at the moment how we might want to handle this but I'll post any updates here if I have any thoughts

PR #7420, opened 2024-11-07
Fix nested interactive issue on select component

jcfilben:
#### What does this PR do?
Resolves the issue in select where automated accessibility tools flagged that we have nested interactive elements by adding `inert` to the input field.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert#browser_compatibility

Because of a react issue I had to pass `inert=""` instead of just `inert`. More details here: https://github.com/facebook/react/issues/17157

I added a note in the code that we may need to revisit this when React 19 comes out because for there is a chance `inert=""` might no longer work but it is not guaranteed either way at the moment so we would potentially would need to pass just `inert` for react 19.

Tested with react 18, 17, and 16 to confirm `inert=""` works in each of those
Tested with existing storybook stories and confirmed voiceover is working the same as it was previously

In the accessibility tab on select stories the nested interactive issue should no longer be flagged

Closes https://github.com/grommet/grommet/issues/7374

no
yes
backwards compatible


taysea (approved):
Nice solution

MikeKingdom (approved):
Nice

Issue #7419, opened 2024-11-07
Long option text sometimes gets rendered right-justified in Select


MikeKingdom:
### Expected Behavior

![image](https://github.com/user-attachments/assets/c4f3d259-ba39-448c-8353-917382b07cb1)

### Actual Behavior

![image](https://github.com/user-attachments/assets/18e4ee96-a340-4285-8f29-429d18154df6)

### URL, screen shot, or Codepen exhibiting the issue
```jsx
import React, { useState } from 'react';

import { Box, Form, FormField, Heading, Layer, Select, Text } from 'grommet';

export const Long = () => {
  const options = [
    {
      category: 'SCDEV01',
      label: 'Disaster Recovery Viewer',
      description:
        'View only permissions with no ability to perform operations.',
      location: 'Built-in',
    },
    {
      category: 'HPE GreenLake Platform',
      label: 'Identity domain and SSO administrator',
      description:
        // eslint-disable-next-line max-len
        'Full management of DNS domains and SSO profiles across the organization',
      location: 'Built-in',
    },
  ];
  const [value, setValue] = useState('');
  return (
    // Uncomment <Grommet> lines when using outside of storybook
    // <Grommet theme={...}>
    <Box fill align="center" justify="start" pad="large" gap="medium">
      <Layer position="right" full="vertical">
        <Box fill pad="medium" width="medium" gap="small">
          <Heading level={2} margin="none">
            Assign role
          </Heading>
          <Text>Grant access to by assigning</Text>
          <Form>
            <FormField label="Role" required name="select" htmlFor="select">
              <Select
                id="select"
                name="select"
                placeholder="Select"
                valueKey="label"
                labelKey="label"
                value={value}
                options={options}
                onChange={({ option }) => setValue(option)}
                onSearch={() => {}}
              >
                {(option, index, _options, { selected }) => (
                  <Box
                    direction="row"
                    width="large"
                    justify="between"
                    align="center"
                    pad="small"
                    background={selected ? "brand" : undefined}
                  >
                    <Box>
                      <Text>{option.category}</Text>
                      <Text weight="bold">{option.label}</Text>
                      <Text>{option.description}</Text>
                    </Box>
                    <Text>{option.location}</Text>
                  </Box>
                )}
              </Select>
            </FormField>
          </Form>
        </Box>
      </Layer>
    </Box>
    // </Grommet>
  );
};

Long.args = {
  full: true,
};

export default {
  title: 'Input/Select/Long',
};

```

[...]

PR #7427, opened 2024-11-12
reset scroll position for readOnly inputs with long values

NOMADE55:
#### What does this PR do?
When using the Select Component, the readonly input used to show the value is scrolled when this value is too long.
This PR leverages the input's scroll attribute to understand if the readonly input is being show properly.
Whenever scrollLeft is more than zero, we reset it back to beginning.

- In this particular case, is worth checking and testing the issue https://github.com/grommet/grommet/issues/7419
- Then check [TextInput's file](https://github.com/grommet/grommet/blob/07f5feff44f7cfe6673a608132e5e68e39b9b233/src/js/components/TextInput/TextInput.js)

-  Tried the fix in a test project.

- Create a test case with a Select which value once displayed, is longer than the available space.

- Fixes https://github.com/grommet/grommet/issues/7419

- No


jcfilben (approved):
Looks good!

MikeKingdom (approved):
Looks good. Tested with the original problem situation as well as typical readonly TextInputs.

I was a little worried it would prevent manual scrolling of long values but it did fine.

NOMADE55:
> Looks good. Tested with the original problem situation as well as typical readonly TextInputs.
> 
> I was a little worried it would prevent manual scrolling of long values but it did fine.

Great!
The only quirk is that it works on blur. 
It's a good compromise. If the user is not focusing on the input anymore, resets its scroll.

PR #7430, opened 2024-11-13
Fix nested interactive issue for SelectMultiple

jcfilben:
#### What does this PR do?
Follow on to https://github.com/grommet/grommet/pull/7420, fixes the nested interactive issue for SelectMultiple

In addition to adding `inert` to the inputs within the drop button we also needed to add `inert` to the CheckBoxes because they are nested within a Button element. I also added `pointerEvents: 'none'` to the CheckBoxes to prevent an issue in FireFox where the click event was swallowed by the CheckBox instead of triggering the click event on the Button.

To add the `pointerEvents: 'none'` style onto the StyledCheckBoxContainer I needed to add containerProps onto CheckBox, for now leaving this as an internal only prop.

Normally we avoid inline styles but in this case it was the solution that felt like it was the most readable and least intrusive to the CheckBox component.

Check that SelectMultiple Storybook stories no longer have an error in the accessibility tab for nested interactive elements

no
yes
backwards compatible

### DIFF
--- a/src/js/components/CheckBox/CheckBox.js
+++ b/src/js/components/CheckBox/CheckBox.js
@@ -32,6 +32,7 @@ const CheckBox = forwardRef(
       'aria-label': ariaLabel,
       checked: checkedProp,
       children,
+      containerProps, // internal only for now, used by SelectMultiple
       defaultChecked = false,
       disabled,
       fill,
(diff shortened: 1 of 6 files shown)


britt6612 (dismissed):

britt6612:
https://deploy-preview-7430--sad-tereshkova-173d07.netlify.app/?path=/story/input-selectmultiple-disabled--disabled
still see error here

taysea (approved):
Looks good, thank you.

britt6612 (approved):

Issue #7303, opened 2024-08-31
UI issue when FormField label is a node and required


nnurmano:
I am using HPE theme and when FormField label is a node and required, the required indicator slips to the next row. 
<img width="359" alt="image" src="https://github.com/user-attachments/assets/f475ab9d-81ed-4d07-ab03-e1eb7ebeb736">

Here is a sandbox that illustrates the issue https://codesandbox.io/p/sandbox/grommet-v2-template-forked-d87lgp?file=%2Findex.js%3A34%2C34&workspaceId=724eced4-5631-43d4-a90e-1f99ee73f924

britt6612:
@nnurmano Im not able to see the codesandbox can you make it public or paste the code snippet?

nnurmano:
Made it public, sorry I did not know that it is private by default

britt6612:
@nnurmano I see the issue now thank you for bringing up! Wondering if that container with the `label` and `required` needs a direction `row` would you be interested in submitting a PR?

NOMADE55:
@britt6612 Please check my PR observations regarding this fix in https://github.com/grommet/grommet/pull/7435

NOMADE55:
> > @britt6612 Please check my PR observations regarding this fix in #7435
> 
> Your node child could have `style="display: inline-flex;"` and get the desired effect.

I agree.

jcfilben:
> > @britt6612 Please check my PR observations regarding this fix in #7435
> 
> Your node child could have `style="display: inline-flex;"` and get the desired effect.

Good comment, I think it makes sense to avoid building this style into the grommet component and instead people can add this to their custom FormField label node if needed. Going to mark this issue as closed.

nnurmano:
Unfortunately, this does not solve the issue
When there is an icon, the asterisk is not above the label. 
<img width="412" alt="image" src="https://github.com/user-attachments/assets/b23ee13c-03e5-4662-af47-95e9c1890987">

nnurmano:
Used the changes by @NOMADE55 & styled the FormField component

const StyledFormField = styled(FormField)`
  & label {
    display: flex;
  }
`;

NOMADE55:
@nnurmano my changes have been dismissed.

Another solution could be to prevent the showIndicator from appearing, *even if required*, so you can control the asterisk's position manually within the label you created.

PR #7435, opened 2024-11-16
fix: requirement indicator below label

NOMADE55:
#### What does this PR do?
- Adds flex (row) for labels in [FormField](https://github.com/grommet/grommet/blob/a3a17c16ce36c49eedc1ea9e8e61b75cfbb83cb9/src/js/components/FormField/FormField.js) in case the value is required. 

- This [code sandbox](https://codesandbox.io/p/sandbox/grommet-v2-template-forked-d87lgp?file=%2Findex.js%3A34%2C34&workspaceId=724eced4-5631-43d4-a90e-1f99ee73f924) shows a possible issue when the label is `display: block` type element.
- Check the file change for [FormField](https://github.com/grommet/grommet/blob/a3a17c16ce36c49eedc1ea9e8e61b75cfbb83cb9/src/js/components/FormField/FormField.js) 

- Ran Jest
- Tested in a mocked test project 

- The required indicator being a span will go always go below a block type component. 
- Maybe also align with `baseline` could be an improvement.

- Fixes https://github.com/grommet/grommet/issues/7303

image

I'm not sure if this should actually be "fixed". I'm creating the PR because is a simple change.
That said, adding a flex could break previous implementations of this label in FormField. 
Another solution is adding a prop that will remove the required indicator visually and this enables the developer to add them by themselves in the element.

### DIFF
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -506,6 +506,10 @@ const FormField = forwardRef(
     if (typeof required === 'object' && required.indicator === false)
       showRequiredIndicator = false;
 
+    if (showRequiredIndicator) {
+      labelStyle.style = { display: 'flex' };
+    }
+
     return (
       <FormFieldBox
         ref={formFieldRef}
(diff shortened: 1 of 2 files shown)


NOMADE55:
Closing this, please refer to: https://github.com/grommet/grommet/issues/7303#issuecomment-2483581827

PR #7437, opened 2024-11-18
fix calendar accessibility issue

### LINKED ISSUE #7424: Calendar - Ensure elements with an ARIA role that require parent roles are contained by them
### Expected Behavior
Calendar should pass all automated accessibility checks

### Actual Behavior
Calendar has the following accessibility error: 'Ensure elements with an ARIA role that require parent roles are contained by them'
An element in Calendar has `role="row"` and doesn't have the correct parent role.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/row_role

https://dequeuniversity.com/rules/axe/4.10/aria-required-parent?application=axeAPI

https://storybook.grommet.io/?path=/story/visualizations-calendar-simple--simple

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

britt6612:
#### What does this PR do?
This PR adds a `box` around the `daysoftheweek` as well as `styledweekscontainer` 

Since `styledweekscontainer` contains the role `rowgroup` when using the role `rowgroup` this needs to be wrapped within either a role `grid` `table` `treegrid` previously we had the role `grid` on the styledCalendar however with the role `grid` this can not contain more than one tabbable element so we needed to move it down after the previous and next buttons. 
calendar.js
locally 
storybook in calendar for accessibility 

```
Calendar has the following accessibility error: 'Ensure elements with an ARIA role that require parent roles are contained by them'
An element in Calendar has role="row" and doesn't have the correct parent role.
```
closes #7424 
no
no
backwards compatible


jcfilben:
I'm seeing an unexpected snapshot change on the calendar/fill story

jcfilben (approved):
Looks good!

MikeKingdom (approved):
Looks good!

PR #7445, opened 2024-11-21
add support for messages for meter

### LINKED ISSUE #7326: Meter - screenreader accessibility
### Expected Behavior
Meter information should be read out for a screenreader user.

### Actual Behavior
Currently, Voiceover doesn't have access to read chart text information; just read out as "image" with no content information.

This issue can be seen in any of the Meter storybook stories
https://storybook.grommet.io/?path=/story/visualizations-meter-bar--bar

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:

britt6612:
#### What does this PR do?
adds `messages` to be passed through the `Grommet` tag for Meter component
meter.js
storybook
storybook
TODO:

previously screen reader just read out that our meter was an image. We need to provide better context
closes #7326 
yes
yes
backwards compatible


src/js/components/Meter/__tests__/__snapshots__/Meter-test.tsx.snap
```diff
@@ -24,6 +24,7 @@ exports[`Meter background 1`] = `
   class="c0"
 >
   <svg
+    aria-label="bar meter with undefined% filled"
```
MikeKingdom: What's happening with all the `undefined` substitution values?

src/js/components/Meter/Meter.js
```diff
@@ -46,12 +50,41 @@ const Meter = forwardRef(
...
+      ariaLabel ||
+      format({
+        id: meterType,
+        messages: messages?.meter,
+        values: {
+          percentage: value,
+          type,
+        },
+      });
+
```
MikeKingdom:
```suggestion
    const meterAriaLabel =
      ariaLabel ||
      format({
        id: `meter.${type ?? 'default'}`,
        messages: messages?.meter,
        values: {
          percentage: value ?? 0,
          type,
        },
      });

```

MikeKingdom (commented):
Needs the messages property in Meter/index.d.ts and propTypes.js

src/js/components/Meter/__tests__/__snapshots__/Meter-test.tsx.snap
britt6612: its when there is no `value` fixing now

britt6612: in one test we have <Meter /> without any `value` passed in

britt6612: There shouldnt be anymore undefined

src/js/components/Meter/__tests__/__snapshots__/Meter-test.tsx.snap
```diff
@@ -249,6 +256,7 @@ exports[`Meter boundary values 1`] = `
   class="c0"
 >
   <svg
+    aria-label="circle meter with value 2.26% out of 10.299999999999999"
```
britt6612, line 259: @jcfilben still not sure about the wording.. what you think

britt6612: okay I wonder if we take out the `%` and just have the user pass in whatever units they want

jcfilben: Yeah I think we should take out the `%`

britt6612: agree took out in last commit

src/js/components/Meter/__tests__/__snapshots__/Meter-test.tsx.snap
```diff
@@ -249,6 +256,7 @@ exports[`Meter boundary values 1`] = `
   class="c0"
 >
   <svg
+    aria-label="circle meter with value 2.26 out of 10.299999999999999"
```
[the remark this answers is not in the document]
britt6612: I can add the 8.04 in the message in the plural one..

so the max is coming from here: 
```
const deriveMax = (values) => {
  let max = 100;
  if (values && values.length > 1) {
    max = 0;
    values.forEach((v) => {
      max += v.value;
    });
  }
  return max;
};
```

`    const memoizedMax = useMemo(() => deriveMax(values), [values]);`

the `memoizedMax` is what Im passing through to the message.
if a user was passing `max` this was being put through the `...rest` so we didnt have access to this value pulling out the `max` so we can use that if it is being passed through

src/js/languages/default.json
```diff
@@ -104,6 +104,10 @@
     "openMenu": "Open Menu",
     "closeMenu": "Close Menu"
   },
+  "meter": {
+    "singular": "{type} meter with value {meterValue} filled out of {max}",
+    "plural": "{type} meter with values {meterValue} filled out of {max}"
```
jcfilben, line 109:
```suggestion
    "singular": "{type} meter with value {meterValue} out of {max}",
    "plural": "{type} meter with values {meterValue} out of {max}"
```

src/js/components/Meter/__tests__/__snapshots__/Meter-test.tsx.snap
```diff
@@ -306,6 +364,7 @@ exports[`Meter default 1`] = `
   class="c0"
 >
   <svg
+    aria-label="bar meter with values  filled out of 100"
```
jcfilben, line 367: Can we have a message that handles when there is no value?

jcfilben:
Looks like the max bundlesize value needs bumped up a little

src/js/components/Meter/__tests__/__snapshots__/Meter-test.tsx.snap
britt6612: done

jcfilben (approved):
Looks good!

src/js/components/Meter/Meter.js
```diff
@@ -19,9 +20,12 @@ const deriveMax = (values) => {
 const Meter = forwardRef(
   (
     {
+      'aria-label': ariaLabel,
       background = { color: 'light-2', opacity: 'medium' },
       color,
       direction = 'horizontal',
+      max,
```
halocline, line 27: Was `max` documented, but not an actual prop?

halocline, line 27: `max` documentation states "The maximum value for the Meter." Does this imply that if I specify `max={10}`, then set `value={20}`, the 20 won't be respected? If not, should the definition of max be clarified?

halocline: Are there other props documented but not represented as props here?

src/js/components/Meter/Meter.js
```diff
@@ -47,12 +52,28 @@ const Meter = forwardRef(
...
+      });
+
     let content;
     if (type === 'bar') {
       content = (
         <Bar
           ref={ref}
-          max={memoizedMax}
+          aria-label={meterAriaLabel}
+          max={max || memoizedMax}
```
MikeKingdom, line 76: This isn't quite the same but probably an improvement.
Previously if the max prop was in `rest` I think it would override this property. Now you will get `memoizedMax` if max is specified but falsey (e.g. 0, undefined, null). Maybe that's better? That does assume we can't have negative values (e.g. say a range of values from -10 to 0)?

I think it's little fragile that the rules for what max is are now split into 4 places:

1. deriveMax
2. the message max param
3. Bar max property (does logic)
4. Cirlce max property (does logic)

This is alot of risk for disagreement. Right now if there are multiple `values` but they are 0 you will get a `memoizedMax` of 0. The message will say 100 but the Bar and Circle will be passed `max=0`

`memoizedMax` will be 100 or if there are at least 2 `values` it adds them up. That also means if `values` only contains 1 thing it will return 100 or if there's a `value` it will return 100.

To avoid this we might want to refactor to normalize max in one spot and pass that one value to the message, Bar and Circle. Maybe something like:

Change the component prop to be `max: maxProp` and then
```javascript
const max = useMemo(() => {
  let maxValue = 100;
  if (values?.length > 1) {
    maxValue = values.reduce((total, currentValue) => total + currentValue, 0);
  }
  return maxValue || maxProp || 100;
}, [maxProp, values]);  
```
 And then just use `max` for the 3 places it's needed (message param, Bar prop and Circle prop).

src/js/components/Meter/__tests__/Meter-test.tsx
```diff
@@ -248,4 +248,22 @@ describe('Meter', () => {
...
   });
+
+  test('custom message', () => {
+    const value = 30;
+    const { container } = render(
+      <Grommet
+        messages={{
+          messages: {
+            meter: {
+              singular: `bar value is ${value}% out of 100.`,
```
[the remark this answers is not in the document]
britt6612: adding.

src/js/components/Meter/__tests__/Meter-test.tsx
```diff
@@ -248,4 +248,22 @@ describe('Meter', () => {
...
+            },
+          },
+        }}
+      >
+        <Meter value={value} />
```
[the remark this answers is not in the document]
britt6612: yes correct a user can add on the Meter directly adding

src/js/components/Meter/Meter.js
britt6612: yes there are other props documented but not represented here I had pulled out the `max` so I could use that in the message since that was just being passed in the `...rest`

britt6612: correct we can clarify our documentation. 

so if we have the following: 
```
      <Meter type="bar" max={50} value={60} />
```

The whole meter will be filled. 

Screenshot 2024-11-25 at 7 18 01 PM

src/js/components/Meter/Meter.js
britt6612: @MikeKingdom 

yes agree there is alot being done here 

```
deriveMax
the message max param
Bar max property (does logic)
Cirlce max property (does logic)
```

I added the suggestion you provided with a small change. 

1. I needed 
` maxValue = values.reduce((total, currentValue) => total + currentValue, 0);`
currentValue.value

2. I put the `maxProp` to be the first in our || statement so its the following 
`  return maxProp || maxValue || 100;`

for example if I have 
```
      <Meter type="bar" max={40} value={30} />
```
with having maxValue || maxProp || 100 
it was taking the maxValue of 100 and not respecting what the user passed in for `maxProp`

src/js/languages/default.json
```diff
@@ -104,6 +104,10 @@
     "openMenu": "Open Menu",
     "closeMenu": "Close Menu"
   },
+  "meter": {
+    "singular": "{type} meter with value {meterValue} out of {max}",
+    "plural": "{type} meter with values {meterValue} out of {max}"
```
MikeKingdom, line 109: One thing I just realized is the `type` isn't currently localizable. We should have keys in here for type. Although a bit wordy, I think a more flexible thing for localizers would be:
```suggestion
    "bar": {
      "singular": "Bar meter with value {meterValue} out of {max}",
      "plural": "Bar meter with values {meterValue} out of {max}"
    },
    "circle": {
      "singular": "Circle meter with value {meterValue} out of {max}",
      "plural": "Circle meter with values {meterValue} out of {max}"
    }
```

britt6612: aw okay good point I can restructure

britt6612: @MikeKingdom updated structure

MikeKingdom (approved):
Thanks!

jcfilben (approved):
Looks good!

PR #7446, opened 2024-11-26
announce the error in max character count

britt6612:
#### What does this PR do?
This PR adds an `announcement` so the screen reader will read out the error on the validation for the max character count
formfield.js
screen reader locally

Announce the max character allowed before the user begins typing.
Announce when the user has typed beyond the threshold of the max allowed.
https://github.com/grommet/hpe-design-system/issues/3787

no
yes
backwards compatible


MikeKingdom (approved):

taysea (approved):
Tested on https://deploy-preview-7446--sad-tereshkova-173d07.netlify.app/?path=/story/input-formfield-thresholdvalidation--threshold-validation and it worked.

PR #7448, opened 2024-11-26
fix list out of view for upkey and downKey

### LINKED ISSUE #7323: List - when using keyboard, active list item isn't always in view

Sulaymon333:
#### What does this PR do?
- Fixes list out of view for up key and down key for List component

closes #7323


src/js/components/List/List.js
```diff
@@ -315,7 +315,20 @@ const List = React.forwardRef(
               event.preventDefault();
               if (active) {
                 const min = onOrder ? 1 : 0;
-                updateActive(Math.max(active - 1, min));
+                const activeElementIndex = Math.max(active - 1, min);
+                updateActive(activeElementIndex);
+
+                // Ensure the active item is in view
+                // setTimeout for activeElement to be updated
+                setTimeout(() => {
```
jcfilben, line 323: Do we need the timeout? I tested without it and wasn't seeing a difference in behavior

Sulaymon333: Indeed the behavior is the same without the timeout, timeout is implemented to guarantee that the activeElement is updated  prior to executing any operations on it and eventually it ensures that all test cases pass. I have shown a screenshot where the timeout is commented and the test case is failing.

I hope this clarifies the need for the timeout. Looking through the code base, a similar approach is employed for example, in the SkipLinks component, Layer component, and some other places.

image

jcfilben: Thanks for the info!

jcfilben (approved):
Looks good!

MikeKingdom (approved):

PR #7461, opened 2024-12-11
Fix additional accessibility issues in storybook part 3

jcfilben:
#### What does this PR do?
Fixes storybook accessibility violations flagged by automated testing

https://github.com/grommet/grommet/issues/6124

no
yes, change to base theme

there is a change to the base theme but I would consider it to still be backwards compatible because what was there was not meeting accessibility requirements

### DIFF
--- a/src/js/components/Form/stories/CustomThemed/FieldSpacingOptions.js
+++ b/src/js/components/Form/stories/CustomThemed/FieldSpacingOptions.js
@@ -60,7 +60,7 @@ export const FieldSpacingOptions = () => (
             Default Spacing
           </Heading>
           <FormField
-            htmlFor="example1-id"
+            htmlFor="textinput1-id"
             name="example1"
             label="Field Label"
             help="Some helpful descriptive text"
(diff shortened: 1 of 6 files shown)


src/js/components/Form/stories/CustomThemed/FieldSpacingOptions.js
```diff
@@ -98,28 +98,28 @@ export const FieldSpacingOptions = () => (
             }}
           >
             <FormField
-              htmlFor="example1-id"
-              name="example1"
+              htmlFor="example3-id"
```
britt6612, line 101: can we do anything more descriptive then `example?`

britt6612:
just one small comment

src/js/components/Form/stories/CustomThemed/FieldSpacingOptions.js
jcfilben: Yeah was kinda tricky to come up with descriptive things to put here since they are all just generic examples, but I described what kind of form element they were

britt6612 (approved):

taysea (approved):

PR #7310, opened 2024-09-12
Add announce to DataChart Detail for series label or property and value

mcbouslog:
#### What does this PR do?
- adds `AnnounceContext` and `announce` to DataChart Detail to announce Detail content with screen reader

#### Where should the reviewer start?
- review issue #7309 

#### What testing has been done on this PR?
- I'm having difficulties creating test(s), suggestions welcome!

#### How should this be manually tested?
- locally, review http://localhost:9001/?path=/story/visualizations-datachart-detail--detail with screen reader, noting announcement of Detail content as arrow or mouseover series values

#### Do Jest tests follow these best practices?

- [ ] `screen` is used for querying.
- [ ] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [ ] `asFragment()` is used for snapshot testing.

#### Any background context you want to provide?
- I'm not sure if the related issue is valid or if this is the best solution, ok to close if not

#### What are the relevant issues?
- issue #7309 

#### Do the grommet docs need to be updated?
- I don't think so

Issue #7309, opened 2024-09-12
DataChart with detail should announce Detail content with screen reader


mcbouslog:
Using a DataChart, with `detail`, when I use arrow keys or mouseover a series value the Detail component shows the serie label or property and serie value, and maybe it should also announce that information with a screen reader?

### Is this something you or your team is interested in contributing a pull request towards?
I have started a related pull request, will open as draft shortly, and will link to this issue.

### Expected Behavior

The content shown in the DataChart Detail component (on arrow or mouseover series data) is announced with a screen reader.

### Actual Behavior

The content shown in the DataChart Detail component (on arrow or mouseover series data) is not announced with a screen reader.

### URL, screen shot, or Codepen exhibiting the issue
Grommet DataChart Detail story - https://storybook.grommet.io/?path=/story/visualizations-datachart-detail--detail

### Steps to Reproduce

1. from https://storybook.grommet.io/?path=/story/visualizations-datachart-detail--detail
2. turn on screen reader
3. navigate to DataChart
4. use arrow keys to focus different bars within DataChart
5. note Detail component shows bar content
6. note bar content is not announced with screen reader

### Your Environment

- Grommet version: 2.40.1
- Browser Name and version: Chrome v128
- Operating System and version (desktop or mobile): desktop

jcfilben:
Resolved by #7310

PR #7465, opened 2025-01-05
a11y: announce DataChart Detail in VoiceOver

eatyourgreens:
Some changes to DataChart Detail so that its live regions are announced in VoiceOver/Safari.
- make the live region assertive, so that it's announced immediately on pressing a key.
- implicitly call `announce` when handling the left/right cursor keys to navigate a chart.
- prevent the default arrow key behaviour, which is to navigate a node's content character-by-character.
- add full stops to the announced message, to distinguish between the x- and y-axis values.
- give the browser some time to register the `AnnounceContext` live region, before making announcements.
- don't trap focus in the popup when it's open. The keyboard controls rely on `DetailControl` having focus even when the `Drop` is open.
- add list semantics to the `DetailControl` component, so that chart details can be navigated as a list in screen readers.
- fix missing `aria-pressed` state for selected legend buttons.

Start with the storybook. You'll probably need to test in Safari with VoiceOver, to get full support for live regions.
http://localhost:9001/?path=/story/visualizations-datachart-detail--detail

I've tested with VoiceOver/Safari, navigating to the chart in the storybook, then using the left/right arrow keys to navigate  the chart and verify that each bar's data is announced by VO. I've tested both by tabbing/shift-tabbing through the page, like a sighted user would do, and navigating through the page in VoiceOver with ctrl-opt-left arrow and ctrl-opt-right arrow, like a blind user would. Navigating via the virtual cursor can be buggy, since the keyboard controls depend on the `DetailControl` component being focussed. If you don't tab to the focusable `DetailControl`, the left/right keys may not work as expected.

This screen recording shows me navigating the chart with the left/right keys in Safari. VoiceOver correctly announces each bar as I navigate to it.

https://github.com/user-attachments/assets/3e9fa8be-24c2-455d-8dbe-e96ceb4bc782

I've also tested quickly with Chrome/VoiceOver (works pretty much the same as Safari) and Firefox/VoiceOver (ignores the live region changes when I navigate the chart.)
 
See above.

- #7310 added live regions, but they were polite and don't update immediately after using the keyboard. It also didn't override the default arrow key behaviour in VO, which is to navigate a node's content character-by-character.

- #7309 should be fully fixed by this. Previously, VO worked but only for sighted users, who can navigate the chart with a mouse or other pointing device. Keyboard navigation didn't announce the chart details.
- this is still somewhat buggy, though better than before. ~~The first bar isn't consistently announced, so you might need to navigate back and forward through the chart a couple of times to hear all the bars.~~ I've fixed missing announcements by making sure that the live region is created, and registered with the browser, when `AnnounceContext` is first imported. That said,  a better implementation might be to drop live regions altogether. Instead, implement a data chart as a grouped control, where each detail bar is a focusable, labelled region. In that case, a screen reader should announce each bar's role and accessible name/label when it receives focus. I didn't really have time to make such a sweeping change to the `DataChart` component here.

No.

Maybe mention this as an accessibility fix for screenreader users.

It should be completely backwards compatible.

### DIFF
--- a/src/js/components/DataChart/Detail.js
+++ b/src/js/components/DataChart/Detail.js
@@ -140,11 +140,9 @@ const Detail = ({
         <DetailControl
           key="band"
           fill
-          role="region"
+          role="list"
           tabIndex={0}
-          aria-label={`
-            Data chart: Use the left and right arrow keys to navigate the data.
-          `}
+          aria-label="Data chart"
           justify="between"
           responsive={false}
           {...(horizontalProp
@@ -155,7 +153,11 @@ const Detail = ({
                 direction: 'row',
                 pad,
               })}
-          onFocus={() => {}}
+          onFocus={() => {
+            announce(
+              'Use the left and right arrows to navigate the data chart.',
+            );
+          }}
           onBlur={() => setDetailIndex(undefined)}
         >
           {data.map((_, i) => {
@@ -167,6 +169,8 @@ const Detail = ({
               <Box
                 // eslint-disable-next-line react/no-array-index-key
                 key={i}
+                role="listitem"
+                aria-label={getContent(i)}
                 responsive={false}
                 {...(horizontalProp
                   ? {
--- a/src/js/components/DataChart/Detail.js
+++ b/src/js/components/DataChart/Detail.js
@@ -170,7 +170,6 @@ const Detail = ({
                 // eslint-disable-next-line react/no-array-index-key
                 key={i}
                 role="listitem"
-                aria-label={getContent(i)}
                 responsive={false}
                 {...(horizontalProp
                   ? {
@@ -191,6 +190,8 @@ const Detail = ({
                 onBlur={() => {}}
               >
                 <Box
+                  role="img"
+                  aria-label={getContent(i)}
                   // for horizontal, ref will be placed on child box so
                   // drop is restricted to drop dimensions as opposed
                   // to filling the chart width
--- a/src/js/components/DataChart/Detail.js
+++ b/src/js/components/DataChart/Detail.js
@@ -104,7 +104,7 @@ const Detail = ({
             return `${serie.label || serie.property} ${renderValue(
               serie,
               axisValue,
-            )}`;
+            )}.`;
           })
           .join(' ');
       }
--- a/src/js/components/DataChart/Detail.js
+++ b/src/js/components/DataChart/Detail.js
@@ -7,7 +7,8 @@ import React, {
   useState,
 } from 'react';
 import styled from 'styled-components';
-import { AnnounceContext } from '../../contexts';
+import { AnnounceContext } from '../../contexts/AnnounceContext';
+import { MessageContext } from '../../contexts/MessageContext';
 import { Box } from '../Box';
 import { Drop } from '../Drop';
 import { Grid } from '../Grid';
@@ -38,6 +39,7 @@ const Detail = ({
   thickness,
 }) => {
   const announce = useContext(AnnounceContext);
+  const { format } = useContext(MessageContext);
   const { theme } = useThemeValue();
   const [detailIndex, setDetailIndex] = useState();
   const activeIndex = useRef();
@@ -142,7 +144,7 @@ const Detail = ({
           fill
           role="list"
           tabIndex={0}
-          aria-label="Data chart"
+          aria-label={format({ id: 'dataChart.detailTitle' })}
           justify="between"
           responsive={false}
           {...(horizontalProp
@@ -154,9 +156,7 @@ const Detail = ({
                 pad,
               })}
           onFocus={() => {
-            announce(
-              'Use the left and right arrows to navigate the data chart.',
-            );
+            announce(format({ id: 'dataChart.detailFocus' }));
           }}
           onBlur={() => setDetailIndex(undefined)}
         >
--- a/src/js/components/DataChart/Detail.js
+++ b/src/js/components/DataChart/Detail.js
@@ -180,6 +180,7 @@ const Detail = ({
                 onMouseOver={(event) => {
                   activeIndex.current = event.currentTarget;
                   setDetailIndex(i);
+                  announce(getContent(i), 'assertive');
                 }}
                 onMouseLeave={onMouseLeave}
                 onFocus={() => {}}
--- a/src/js/components/DataChart/Detail.js
+++ b/src/js/components/DataChart/Detail.js
@@ -214,6 +214,7 @@ const Detail = ({
           align={dropAlign}
           plain
           onMouseLeave={onMouseLeave}
+          trapFocus={false}
         >
           <Box pad="small" background={{ color: 'background-back' }}>
             <Grid
--- a/src/js/components/DataChart/Detail.js
+++ b/src/js/components/DataChart/Detail.js
@@ -2,7 +2,6 @@ import React, {
   Fragment,
   useCallback,
   useContext,
-  useEffect,
   useMemo,
   useRef,
   useState,
@@ -113,12 +112,6 @@ const Detail = ({
     },
     [activeProperty, axis, data, horizontalProp, renderValue, series],
   );
-
-  useEffect(() => {
-    if (detailIndex !== undefined) {
-      announce(getContent(detailIndex), 'assertive');
-    }
-  }, [announce, detailIndex, getContent]);
 
   return (
     <>
--- a/src/js/components/DataChart/Detail.js
+++ b/src/js/components/DataChart/Detail.js
@@ -88,55 +88,70 @@ const Detail = ({
     return res;
   }, [data.length, detailIndex, horizontalProp]);
 
+  const getContent = useCallback(
+    (index) => {
+      if (index !== undefined) {
+        return series
+          .filter(
+            ({ property }) =>
+              ((!activeProperty || activeProperty === property) &&
+                data?.[index]?.[property] !== undefined) ||
+              (axis && axis.x && axis.x.property === property),
+          )
+          .map((serie) => {
+            const axisValue = horizontalProp
+              ? data[index][serie.property]
+              : index;
+            return `${serie.label || serie.property} ${renderValue(
+              serie,
+              axisValue,
+            )}`;
+          })
+          .join(' ');
+      }
+      return undefined;
+    },
+    [activeProperty, axis, data, horizontalProp, renderValue, series],
+  );
+
   useEffect(() => {
     if (detailIndex !== undefined) {
-      const content = series
-        .filter(
-          ({ property }) =>
-            ((!activeProperty || activeProperty === property) &&
-              data?.[detailIndex]?.[property] !== undefined) ||
-            (axis && axis.x && axis.x.property === property),
-        )
-        .map((serie) => {
-          const axisValue = horizontalProp
-            ? data[detailIndex][serie.property]
-            : detailIndex;
-          return `${serie.label || serie.property} ${renderValue(
-            serie,
-            axisValue,
-          )}`;
-        })
-        .join(' ');
-      announce(content);
+      announce(getContent(detailIndex), 'assertive');
     }
-  }, [
-    activeProperty,
-    announce,
-    axis,
-    data,
-    detailIndex,
-    horizontalProp,
-    renderValue,
-    series,
-  ]);
+  }, [announce, detailIndex, getContent]);
 
   return (
     <>
       <Keyboard
-        onLeft={() => {
-          if (detailIndex === undefined) setDetailIndex(data.length - 1);
-          else if (detailIndex > 0) setDetailIndex(detailIndex - 1);
+        onLeft={(event) => {
+          event.preventDefault();
+          if (detailIndex === undefined) {
+            setDetailIndex(data.length - 1);
+            announce(getContent(data.length - 1), 'assertive');
+          } else if (detailIndex > 0) {
+            setDetailIndex(detailIndex - 1);
+            announce(getContent(detailIndex - 1), 'assertive');
+          }
         }}
-        onRight={() => {
-          if (detailIndex === undefined) setDetailIndex(0);
-          else if (detailIndex < data.length - 1)
+        onRight={(event) => {
+          event.preventDefault();
+          if (detailIndex === undefined) {
+            setDetailIndex(0);
+            announce(getContent(0), 'assertive');
+          } else if (detailIndex < data.length - 1) {
             setDetailIndex(detailIndex + 1);
+            announce(getContent(detailIndex + 1), 'assertive');
+          }
         }}
       >
         <DetailControl
           key="band"
-          tabIndex={0}
           fill
+          role="region"
+          tabIndex={0}
+          aria-label={`
+            Data chart: Use the left and right arrow keys to navigate the data.
+          `}
           justify="between"
           responsive={false}
           {...(horizontalProp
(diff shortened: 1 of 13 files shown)


src/js/components/DataChart/Detail.js
```diff
@@ -88,55 +88,70 @@ const Detail = ({
...
       >
         <DetailControl
           key="band"
-          tabIndex={0}
           fill
+          role="region"
+          tabIndex={0}
+          aria-label={`
+            Data chart: Use the left and right arrow keys to navigate the data.
+          `}
```
eatyourgreens, line 154: Whenever you make a generic element, like `div`, focusable, you must also override the generic role and add an accessible name/label to the element.

src/js/components/DataChart/Detail.js
```diff
@@ -88,55 +88,70 @@ const Detail = ({
...
-    activeProperty,
-    announce,
-    axis,
-    data,
-    detailIndex,
-    horizontalProp,
-    renderValue,
-    series,
-  ]);
+  }, [announce, detailIndex, getContent]);
```
eatyourgreens, line 121: I'm not sure if this effect hook is still needed, when `announce` is called directly from the keyboard handlers.

eatyourgreens: Testing shows it doesn't add anything in Chrome or Safari, so I've removed it.

src/js/components/DataChart/Detail.js
```diff
@@ -172,6 +180,7 @@ const Detail = ({
                 onMouseOver={(event) => {
                   activeIndex.current = event.currentTarget;
                   setDetailIndex(i);
+                  announce(getContent(i), 'assertive');
```
eatyourgreens, line 183: Totally blind users won't use a mouse (or any other visual pointing device) but I've left the previous mouse hover behaviour in, in case it's useful to announce the detail content to sighted VoiceOver users.

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -13,6 +13,12 @@ const createAnnouncer = () => {
 
   return announcer;
 };
+/*
+Prime the live region so that it's registered with the browser
+when we first call the announce function.
+See https://tetralogical.com/blog/2024/05/01/why-are-my-live-regions-not-working/
+*/
+createAnnouncer();
```
eatyourgreens, line 21: This should fix a bug where the first data bar isn't announced, when you first press the right arrow, because the live region hasn't been registered with the browser yet.

src/js/components/DataChart/Detail.js
jcfilben: I like this change, makes it much clearer when the focus lands on this div.

To allow this aria-label to be internationalized can we have this string come from the `languages/default.json` file and then use `MessageContext`

src/js/components/DataChart/Detail.js
```diff
@@ -88,55 +87,64 @@ const Detail = ({
...
+          } else if (detailIndex > 0) {
+            setDetailIndex(detailIndex - 1);
+            announce(getContent(detailIndex - 1), 'assertive');
+          }
         }}
-        onRight={() => {
-          if (detailIndex === undefined) setDetailIndex(0);
-          else if (detailIndex < data.length - 1)
+        onRight={(event) => {
+          event.preventDefault();
```
jcfilben, line 130: In the PR description you mentioned: 

> prevent the default arrow key behaviour, which is to navigate a node's content character-by-character.

I assume that is what the preventDefault here is trying to do. However, when I try to navigate with arrow key while using a screen reader I am still unable to reach the detail section

eatyourgreens: I find it works in VoiceOver if I use Tab to focus the `DetailControl` div. 

https://support.apple.com/en-gb/guide/voiceover/vo2753/mac

If I use VO-right arrow to navigate the page, then keyboard focus doesn't follow the virtual cursor, so `Keyboard` here won't intercept the left and right arrow keys. That issue exists on the main branch too. 

A possible fix for that might be to use the `list` role for the container, and `listitem` for individual detail items. Then, a screen reader user would hear that they'd entered a list, and could navigate the individual details using the keyboard commands for lists.

src/js/components/DataChart/Detail.js
eatyourgreens: Sure. No problem.

src/js/components/DataChart/Detail.js
eatyourgreens: I haven't tried this, but this says that VO-cmd-f4 will move keyboard focus to the VoiceOver cursor. 

https://www.applevis.com/guides/mac-voiceover-keystrokes-alphabet-function-keys-macos-13-ventura

Still seems kind of clunky to me, though. I'm not sure how someone with a vision impairment would discover that they need to focus the component in order to use the custom keyboard controls.

eatyourgreens: I very quickly tried adding list semantics to `DetailControl` and its children. Here's a screen recording where I'm navigating with VO-left arrow and VO-right arrow in Safari. I hear that 'Data chart' is a list, and I can hear the content of each data bar by navigating through the list with standard VoiceOver commands. This might be a better approach to making the chart accessible, rather than trying to force the popups to be announced by a screen reader.

https://github.com/user-attachments/assets/a4778962-28c6-4926-9ba3-d04b743dac21

eatyourgreens: Here's a more complicated example, using the Everything story but in Chrome with VoiceOver. I've still got the chart marked up with list semantics here. I can find the chart with VoiceOver and navigate it as a list, hearing each group of points as a list item. The popups aren't triggered, but that might not be a big deal if I'm listening to the chart and can't see them. 

I can also toggle each data series via the buttons below the chart, although I did notice that each button is missing a pressed state to tell me which data series is currently selected.

https://github.com/user-attachments/assets/dbb1624e-388b-4215-a6ba-1f546a33d736

eatyourgreens: Sorry, I should have added that I experimented with using list roles locally, but I haven't pushed those changes to this PR.

src/js/components/DataChart/Detail.js
eatyourgreens: Done! I've added `dataChart.detailTitle` and `dataChart.detailFocus` with a default title for the chart and keyboard instructions to announce when `Detail` has focus.

src/js/components/DataChart/Detail.js
```diff
@@ -172,12 +183,15 @@ const Detail = ({
...
                   setDetailIndex(i);
+                  announce(getContent(i), 'assertive');
                 }}
                 onMouseLeave={onMouseLeave}
                 onFocus={() => {}}
                 onBlur={() => {}}
               >
                 <Box
+                  role="img"
+                  aria-label={getContent(i)}
```
eatyourgreens, line 194: VoiceOver announces empty list items as 'empty group'. This works around that by adding the content for each detail as a child of each list item.

src/js/components/DataChart/Legend.js
```diff
@@ -52,6 +52,7 @@ const Legend = ({
             <Button
               key={property}
               active={isActive}
+              aria-pressed={isActive ? 'true' : 'false'}
```
eatyourgreens, line 55: You can test this with the legend controls for the Everything story.

eatyourgreens:
Spotted while I was working on this: `DataChart` renders the `a11yTitle` prop as `aria-label` on `Grid` here:
https://github.com/grommet/grommet/blob/1bb5f345d1113dbb5070b8b422a745dba516aa03/src/js/components/DataChart/DataChart.js#L731-L734

`Grid` renders a div by default, which has the generic role. The ARIA spec explicitly forbids generic elements from being labelled, so `a11yTitle` on a chart here effectively does nothing in screen readers.

>Because the generic role is nameless, the [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) and [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) attributes are prohibited. Because the role is generic, the [aria-roledescription](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-roledescription) and [aria-brailleroledescription](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-brailleroledescription) attributes are also prohibited.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/generic_role#description

I suspect there's a few other places in the Grommet code where `a11yTitle` is used to add `aria-label` to HTML elements that can't be labelled.

src/js/components/DataChart/Detail.js
```diff
@@ -88,55 +89,62 @@ const Detail = ({
...
+  );
 
   return (
     <>
       <Keyboard
-        onLeft={() => {
-          if (detailIndex === undefined) setDetailIndex(data.length - 1);
-          else if (detailIndex > 0) setDetailIndex(detailIndex - 1);
+        onLeft={(event) => {
+          event.preventDefault();
```
jcfilben, line 122: I'm not seeing a difference in behavior when I have `event.preventDefault` vs when I do not. Let me know if I'm missing anything here or not testing correctly. Tested with voice over on Chrome.

jcfilben (commented):
Thanks for updating the code and adding additional comments to the PR, very helpful to understand the changes and your thought process!

Just a few more comments...

src/js/components/DataChart/Detail.js
```diff
@@ -147,7 +155,9 @@ const Detail = ({
                 direction: 'row',
                 pad,
               })}
-          onFocus={() => {}}
+          onFocus={() => {
+            announce(format({ id: 'dataChart.detailFocus' }));
```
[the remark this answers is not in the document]
eatyourgreens: I hear it after tabbing to the detail component, but the live region is polite by default. That means it won’t interrupt if VoiceOver is already speaking. Maybe that’s the problem? I’m not sure. 

I’m not sure if we need to tell a blind user about the custom left/right keys either, so I might remove this.

src/js/components/DataChart/Detail.js
eatyourgreens: Early on, I was hearing VO read out the newlines in the empty detail divs, when I used the arrow keys. You’re right that this line might not be needed any more.

src/js/contexts/AnnounceContext/AnnounceContext.js
eatyourgreens: I’m inclined to revert this commit (and the snapshot tests) and focus this PR just on the `DataChart` component.

src/js/components/DataChart/Detail.js
eatyourgreens: If I remove the `event.preventDefault()`, the left and right keys update the live region, but they also navigate the list, so I hear each detail item twice. Here's a screen recording where I tab to the chart details, then navigate with the cursor keys (without cancelling the default keyboard events.)


https://github.com/user-attachments/assets/c843299f-6010-4580-b69c-1fe9f5bab167

I'm wondering if I should just remove the live region announcement, and rely on the default screen reader keyboard behaviour to announce each list item instead. That seems like a more maintainable solution.

eatyourgreens: Here's a screen recording after trying that. I'm seeing the newlines announced again, and I have to press left or roight arrow twice for each detail item. I'm going to leave the code as-is, otherwise the custom control seems to be clashing with default cursor key behaviour in MacOS.


https://github.com/user-attachments/assets/3c10215b-980b-4091-a901-d6d07cbe435f

If it helps to test this, I'm using one of the newer M3 MacBooks with Full Keyboard Access, so MacOS might be mapping the cursor keys to spatial navigation for the page.

MikeKingdom:
@eatyourgreens Thank you for working on this. It's a big help and definitely needed.

I've been looking at this set of changes, but it's having a fair number of issues with NVDA on Windows. I don't have a license for JAWS to see how it behaves. But it definitely needs some work to be reasonable on Windows. I'll spend some time taking a closer look at the issues. The annouce for the chart about using the arrow keys happens, but when I use the arrow keys it's not really reading the detail involved. It may be an NVDA control key mapping difference

eatyourgreens:
@MikeKingdom thanks for testing with NVDA. I only have access to VoiceOver. 

I've been using [Patrick Lauke's live region tests](https://tetralogical.com/blog/2024/05/01/why-are-my-live-regions-not-working/) in JAWS and NVDA as a reference. 

I'm curious, are you able to navigate the details as a list in NVDA?

MikeKingdom:
> @MikeKingdom thanks for testing with NVDA. I only have access to VoiceOver.
> 
> I've been using [Patrick Lauke's live region tests](https://tetralogical.com/blog/2024/05/01/why-are-my-live-regions-not-working/) in JAWS and NVDA as a reference.
> 
> I'm curious, are you able to navigate the details as a list in NVDA?

@eatyourgreens  Thanks. I'll check it out. I can do arrow key navigation through detail items with the arrow keys in NVDA if I force it to switch to Browse mode instead of Reading mode (Reading mode uses the arrow keys for reading navigation.)

Although I'm wondering about other accessibility improvements we should make for DataChart. Like would it be better to add `title` and `description` blocks to the SVG content and also potentially `graphics-object` and maybe `graphics-symbol` roles (plus aria-labels) to appropriate pieces rather than the listitem/announce strategy. That seems like the more standard way to make charts more accessible

eatyourgreens:
>I can do arrow key navigation through detail items with the arrow keys in NVDA if I force it to switch to Browse mode instead of Reading mode (Reading mode uses the arrow keys for reading navigation.)

VoiceOver uses the arrow keys for reading navigation too, but doesn't seem to override any custom key handlers. I think that's my biggest concern with #7310. It added live region announcements to keys that are already used by screen readers for page navigation.

>Although I'm wondering about other accessibility improvements we should make for DataChart. Like would it be better to add title and description blocks to the SVG content and also potentially graphics-object and maybe graphics-symbol roles (plus aria-labels) to appropriate pieces rather than the listitem/announce strategy. That seems like the more standard way to make charts more accessible.

Absolutely. The chart semantics should be expressed in the SVG, and read by screen readers. I think using a live region to compensate for missing roles and accessible names is always going to run into a couple of problems:
- you don't have any control over when a live region is announced. Polite regions will wait until the browser isn't busy (user isn't pressing any keys, screen reader isn't already speaking, etc.)
- live regions ignore structured markup and are really designed to announce short strings like 'You have 3 new messages.'

eatyourgreens:
[How to make diagrams more accessible](https://www.jointjs.com/blog/diagram-accessibility) suggests marking up a bar chart with `graphics-symbol` for each bar and `graphics-object` for each axis.

src/js/languages/default.json
```diff
@@ -19,6 +19,10 @@
     "enterCalendar": "Calendar is open, use arrow keys and enter to select a date.",
     "exitCalendar": "Exited calendar dialog"
   },
+  "dataChart": {
+    "detailTitle": "Data chart",
+    "detailFocus": "Use the left and right arrow keys to navigate the chart data."
+  },
```
MikeKingdom, line 25: These need to be added to the messages property definition in Grommet/index.d.ts and propTypes.js.

MikeKingdom (commented):
Looks good and is a nice improvement. The new message props need to be documented in the TS and prop types.

We also should make sure to schedule improvements to the rest of the DataChart accessibility (SVG title/desc and graphic roles for the various pieces.)

src/js/languages/default.json
eatyourgreens: Done!

MikeKingdom (approved):
Thanks!

jcfilben (approved):
Looks good!

Issue #7464, opened 2024-12-19
Select - active, hover, and focus styling issues


jcfilben:
When I select an option and then hover over it and move my cursor outside of the Drop, the selected option still has the 'active' styling. I would expect that when the cursor leaves the drop the selected option should just have the 'selected' styling (in this case a purple background).
https://github.com/user-attachments/assets/6f1983ac-6491-41ab-852e-6b82792f999b

We should simplify to have 'hover' styling that is applied when a select option is hovers and 'focus' styling that is applied to a select option when it receives keyboard focus. I don't think we need to have 'active' styling as well.

In addition when I hover over option 2 and my cursor leaves the drop I would expect option 2 to no longer have the hover styling.
<img width="373" alt="Screenshot 2024-12-19 at 3 59 17 PM" src="https://github.com/user-attachments/assets/28906a6b-773e-4e53-9acc-a7dee93c2630" />

This expectation aligns with the native html behavior:

https://github.com/user-attachments/assets/d825d417-b7b0-4800-af4d-183a7ccaa576

On focus via keyboard I would expect a focus indicator to be displayed and the 'hover' and 'active' styling not to show.

umeshiscreative:
Hi @jcfilben, Please check the attached image, Is this the expected behavior?
![grommet2](https://github.com/user-attachments/assets/1c27299c-bdb3-4cab-aaad-50c14984a028)

PR #7469, opened 2025-01-16
Fix Select - active, hover, styling issues on mouseOut

umeshiscreative:
#### What does this PR do?
Has fix for the Hover styling on Select component.

components/select/select-container.js

Tested with all the theme levels with storybook

Reset the onMouse movement

#7464 

NA

NA

NA

### DIFF
--- a/src/js/components/Select/SelectContainer.js
+++ b/src/js/components/Select/SelectContainer.js
@@ -334,13 +334,6 @@ const SelectContainer = forwardRef(
       [keyboardNavigation],
     );
 
-    const onResetActiveOption = useCallback(
-      () => () => {
-        if (!keyboardNavigation) setActiveIndex(-1);
-      },
-      [keyboardNavigation],
-    );
-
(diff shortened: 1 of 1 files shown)


britt6612 (approved):

MikeKingdom (approved):
LGTM

src/js/components/Select/SelectContainer.js
```diff
@@ -483,6 +490,9 @@ const SelectContainer = forwardRef(
                       onMouseOver={
                         !optionDisabled ? onActiveOption(index) : undefined
                       }
+                      onMouseOut={
+                        !optionDisabled ? onResetActiveOption(index) : undefined
```
jcfilben, line 494: Instead of adding `onResetActiveOption` what if we just did something like this instead?
```suggestion
                        !optionDisabled ? onActiveOption(-1) : undefined
```

jcfilben (approved):
Looks good!

Issue #7467, opened 2025-01-14
Calendar - update colors to be accessible


jcfilben:
Based on the work done in https://github.com/grommet/grommet/issues/7423 the Calendar colors should be updated to meet accessibility requirements.

Color recommendations here: https://www.figma.com/design/OBxHRrYylrg1IyNetpK7OF/%237423---Calendar-contrast?node-id=45-31

jcfilben:
Re-opened #7423, need that ticket closed before continuing

Issue #7423, opened 2024-11-08
Calendar - days of the week text doesn't meet required contrast ratio


jcfilben:
### Expected Behavior
Calendar days of the week text should meet accessibility contrast ratios.

### Actual Behavior
There is a 50% opacity applied to the days of the week above the calendar. This text color with the opacity doesn't meet contrast requirements.

### URL, screen shot, or Codepen exhibiting the issue
<img width="782" alt="Screenshot 2024-11-08 at 2 43 13 PM" src="https://github.com/user-attachments/assets/b58b7e6e-fa5f-412b-8f32-cad677185273">

### Steps to Reproduce

https://storybook.grommet.io/?path=/story/visualizations-calendar-fill--fill

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

NOMADE55:
Seems to only be the days outside of the month. 
Which makes sense because the opacity _has_ to be low if these are from previous/next month.

jcfilben:
Agree, it makes sense visually to have the opacity low to communicate that the days are outside of the month but then we run into issues because these controls are still interactive and therefore need to meet contrast requirements for WCAG. Not sure what the correct answer is here but what we have at the moment isn't meeting accessibility requirements.

NOMADE55:
Maybe a sr-only span with a comment remarking how these dates are outside of the month.
Other than that, I'm not sure if this is something that can be solved without seriously impacting the design.

luketa8:
I think we have two options here:

### Option #1

Change -> `opacity: 0,5;` to `opacity: 0,81;`.

This passes WCAG AA and AAA Text Contrast with 4.6:1.

### Option #2

Use `color.text.weak` Token: `#757575`

This passes WCAG AA and AAA Text Contrast with 4.6:1.

luketa8:
Tagging @jcfilben @oliverHPE and @taysea for review. Thanks!

jcfilben:
Thanks for sharing the options, I feel like it is a little hard to distinguish the color of the dates inside the month vs the dates outside of the month. I know that that we don't have a lot of options here to make sure we meet color contrast while also making the dates outside the month visually distinct from the rest of the days. I'm wondering how other component libraries are handling this issue? Is there anything we could learn from their approaches

[...]

PR #7473, opened 2025-01-21
Calendar style a11y updates

jcfilben:
#### What does this PR do?
Updates the Calendar styling to meet accessibility requirements. Based on work done in https://github.com/grommet/grommet/issues/7423.

The `calendar.day.adjacent.color` theme prop was added because in the HPE theme we want the calendar dates outside of the month to be `text-weak` instead of `text-xweak`.
Open to naming suggestions on the theme prop. I went with the `adjacent` naming because this aligns with the `showAdjacentDays` prop on the Calendar component which controls if days outside of the month should be displayed or not.

https://github.com/grommet/grommet/issues/7467

yes
yes
backwards compatible because previously we weren't meeting WCAG AA contrast requirements

### DIFF
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -708,7 +708,7 @@ const Calendar = forwardRef(
             sizeProp={size}
             fillContainer={fill}
           >
-            <StyledDay otherMonth sizeProp={size} fillContainer={fill}>
+            <StyledDay sizeProp={size} fillContainer={fill}>
               {day.toLocaleDateString(locale, { weekday: 'narrow' })}
             </StyledDay>
           </StyledDayContainer>,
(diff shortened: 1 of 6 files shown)


src/js/components/Calendar/Calendar.js
```diff
@@ -708,7 +708,7 @@ const Calendar = forwardRef(
             sizeProp={size}
             fillContainer={fill}
           >
-            <StyledDay otherMonth sizeProp={size} fillContainer={fill}>
```
britt6612, line 711: why was this taken out?

jcfilben: Previously the days of the week were receiving the `otherMonth` styling, but we don't want this to happen anymore and instead they should get the same styling as the regular days of the month.

britt6612: okay thank you for clarifying

britt6612 (approved):

MikeKingdom (approved):
LGTM

PR #7479, opened 2025-01-24
Add ability for compound focus indicator

### LINKED ISSUE #7475: Focus indicator -- ability to have outline + shadow
Grommet `theme.global.focus` should support the ability to define outline + shadow in order to achieve the technique recommended by WCAG: https://www.w3.org/WAI/WCAG21/Techniques/css/C40

Right now, the theme only supports having one or the other.

taysea:
#### What does this PR do?

Enhances Grommet theme to allow for user to do compound focus style that combines both "outline" and "box-shadow" -- this is useful to enable alignment to WCAG recommendation: https://www.w3.org/WAI/WCAG21/Techniques/css/C40

Previously, Grommet would should outline or shadow, so the caller's theme must "opt-in" via `theme.global.focus.allowAll`.

Locally and added jest test

https://github.com/user-attachments/assets/db5c323c-f13b-46aa-a1bf-d8cf0556a405

```
 focus: {
        // shadow or outline are required for accessibility
        // border: {
        //   // remove to only have shadow
        //   color: 'focus',
        // },
        outline: { color: 'black', size: '2px', offset: '2px' },
        shadow: {
          color: 'skyblue',
          size: '2px',
          blur: '0px',
          // inset: true,
        },
        allowAll: true,
      },
```

Closes #7475


src/js/themes/base.d.ts
```diff
@@ -403,12 +403,14 @@ export interface ThemeType {
...
         size?: string;
+        offset?: string;
       };
       shadow?:
         | string
         | {
             color?: ColorType;
             size?: string;
+            blur?: string;
           };
```
jcfilben, line 414: Do we need to include `allowAll` here?

taysea: Yes, thank you. Will add.

taysea: Done.

jcfilben (dismissed):
Looks good!

src/js/themes/base.js
```diff
@@ -281,6 +281,7 @@ export const generate = (baseSpacing = 24, scale = 6) => {
           color: 'focus',
           size: '2px',
         },
+        allowAll: undefined,
```
halocline, line 284: Pausing on the name of this property. From a Grommet convention, typically we haven't used verbs, instead use a noun / specific property.

How were we thinking of describing this in the documentation? Something like?:
```
If true, applies both outline and shadow styles to create a compound focus ring.
```
OR,
```
If true, applies both outline and shadow styles to create a two-color focus ring.
```

A couple naming options:
- `twoColor`
- `compound`

"Two-color" seemed to be the most common terminology used in search results.

halocline (commented):
Implementation and functionality look good.

One question re: theme property name.

src/js/themes/base.js
taysea: Good feedback, let's use the term `twoColor` since it speaks to the technique this is seeking to enable.

halocline (approved):

Issue #6213, opened 2022-06-29
a11y: expand is not specific (breaks 3.3.2)


dgreene1:
This is basically the same concept as https://github.com/grommet/grommet/issues/6204 but it's for the expand button instead of the onSelect checkbox.

So we we want to make sure that the aria label says "expand ${primaryValue}` instead of "expand" so that we can meet the WCAG's [Success Criterion 3.3.2: Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html)

### How to implement this

I think we would want modify the a11yTitle line in ExpanderCell https://github.com/grommet/grommet/blob/04505a6838827267168b75164af799af06e41835/src/js/components/DataTable/ExpanderCell.js#L43

It should follow the same logic as the select line, but we really should wait until we have a clear solution for the NVDA issue describe in https://github.com/grommet/grommet/issues/6204#issuecomment-1170320696. But here's the logic anyway.

https://github.com/grommet/grommet/blob/04505a6838827267168b75164af799af06e41835/src/js/components/DataTable/Body.js#L90

### Expected Behavior

When navigating the [group/on-select example](https://storybook.grommet.io/?path=/story/visualizations-datatable-grouped-and-onselect--grouped-on-select-data-table) with the NVDA screen reader, I would be able to hear context about the row each expand button will be expanding (or collapsing)

### Actual Behavior

It just says "expand"

### URL, screen shot, or Codepen exhibiting the issue

n.a. it's a screen reader

### Steps to Reproduce

Sorry I'm in a rush so I can't provide this step by step. You just need to use the screen reader to navigate to the cell that contains the expand/collapse arrows.

### Your Environment

Grommet version: whatever was in the storybook website on 6/27/2021
Browser Name and version: 101.0.1 (64-bit)
Operating System and version (desktop or mobile): Windows Version 10.0.19042

Hriishikeshh:
hey would like to work on this
assign it to me if suited

britt6612:
@Hriishikeshh we usually dont assign tickets but feel free to work on this and push up a PR! Let us know if you have any questions

jcfilben:
Check which wcag rule this is related to

PR #7480, opened 2025-01-24
Fix DataTable expand aria label

jcfilben:
#### What does this PR do?
Fixed DataTables expand cell aria label to be more descriptive.

Closes https://github.com/grommet/grommet/issues/6213

no
yes
backwards compatible

### DIFF
--- a/src/js/components/DataTable/ExpanderCell.js
+++ b/src/js/components/DataTable/ExpanderCell.js
@@ -9,7 +9,14 @@ import { useThemeValue } from '../../utils/useThemeValue';
 
 // ExpanderControl is separated from ExpanderCell to give TableCell a chance
 // to set the ThemeContext dark context.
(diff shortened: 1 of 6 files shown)


src/js/components/DataTable/ExpanderCell.js
```diff
@@ -38,7 +45,7 @@ const ExpanderControl = ({ context, expanded, onToggle, pad, ...rest }) => {
     content = (
       <Button
         fill
-        a11yTitle={expanded ? 'collapse' : 'expand'}
+        a11yTitle={`${expanded ? 'collapse' : 'expand'} ${primaryValue}`}
```
MikeKingdom, line 48: This may not be human friendly (really in most cases it should be non visible id or guid)

jcfilben:
Closing, doesn't feel like there is a good way for us to try and figure out which value should be read out in the aria-label

PR #7482, opened 2025-01-24
Update dependencies post 2.44.0 release

jcfilben:
#### What does this PR do?
Upgrades dependencies

Froze eslint-plugin-react-hooks to stay at 5.0.0 until the fix for https://github.com/facebook/react/issues/31687 is released

no
no
backwards compatible


MikeKingdom (approved):
LGTM. Looks like it will cause some style class renaming that will make partners have to update snapshots also.

britt6612 (approved):

PR #7488, opened 2025-01-27
add notification.message.fill

### LINKED ISSUE #7485: Notification -- enhance theme to support `notification.message.fill`
In cases where Notification direction is "column" (so the message renders as a paragraph), the theme should have the ability to set paragraph `fill={true}`.

Right now the paragraph always wraps:

Image

britt6612:
#### What does this PR do?
This PR enhances Notification theme to be able to pass `fill: true` to the theme for when Paragraph is used. 
notification.js
locally
locally 

In cases where Notification direction is "column" (so the message renders as a paragraph), the theme should have the ability to set paragraph fill={true}.
closes #7485 
yes
yes
backwards compatible


src/js/themes/base.d.ts
```diff
@@ -1235,7 +1235,7 @@ export interface ThemeType {
     iconContainer?: BoxProps;
     textContainer?: BoxProps;
     title?: TextProps;
-    message?: TextProps;
+    message?: TextProps | { fill?: boolean };
```
taysea, line 1238: Should this be TextProps and fill? instead of TextProps or fill

britt6612: aw yes

src/js/components/Notification/Notification.js
```diff
@@ -172,7 +172,15 @@ const Notification = ({
...
       typeof message === 'string' ? (
-        <Message {...theme.notification.message} color={messageColor}>
+        <Message
+          {...theme.notification.message}
+          color={messageColor}
+          fill={
+            Message === Paragraph
+              ? theme.notification.message?.fill || false
+              : undefined
+          }
```
MikeKingdom, line 182: I'm trying to think of a cleaner way to do this. It seems pretty awkward to first set Message to Paragraph or Text and then compare against Paragraph here. I think it would at least be better to set up props at the same time you define whether it's Paragraph or Text. It would probably be cleaner to actually pull this up into a separate Message component.

britt6612: I can take a stab at making this a bit cleaner

src/js/components/Notification/Notification.js
```diff
@@ -168,18 +194,20 @@ const Notification = ({
...
-          {/* include actions with message so it wraps with message */}
-          {actions}
-        </Message>
+        <MessageComponent
+          message={message}
+          actions={actions}
+          theme={theme}
+          direction={direction}
+          messageColor={messageColor}
+        />
```
MikeKingdom, line 206:
```suggestion
<Message {...theme.notification.message} color={messageColor} direction={direction} >
  <Text margin={{ right: 'xsmall' }}>{message}</Text>
  {/* include actions with message so it wraps with message */}
  {actions}
</Message>
```

src/js/components/Notification/Notification.js
```diff
@@ -17,6 +17,32 @@ import { Text } from '../Text';
...
+      {...theme.notification.message}
+      color={messageColor}
+      fill={messageFill}
+    >
+      <Text margin={{ right: 'xsmall' }}>{message}</Text>
+      {actions}
+    </Message>
+  );
+};
+
```
MikeKingdom, line 45: I think you could simplify to this (plus the suggestion where it's used).
The fill prop should end up getting passed in from the fact that theme.notification.message is already spread into the props. This versions also avoids passing fill to Text. And it avoids making the paragraph or text decision twice (once for the component and once for the fill.) Also by just letting the children come in via `rest` it avoids needing the `message`, and `actions` parameters (or similarly `color`), keeping it more agnostic. I'd considered renaming direction to a different property name since this really doesn't care about direction but rather something telling it to use Paragraph. But this is probably fine.
```suggestion
const Message = ({
  fill,
  direction,
  ...rest,
}) => {
  return direction === 'row'
    ? <Text {...rest} />
    : <Paragraph {...rest} fill={fill || false} />;
};
```

britt6612: Im going to leave the name `direction` as is just because we have in the theme `notification.direction` 
just so we dont get confused I did think we could change it to `messageLayout` but going to leave off for now since that also might confuse.

src/js/components/Notification/Notification.js
```diff
@@ -17,6 +17,12 @@ import { Text } from '../Text';
 import { NotificationType } from './propTypes';
 import { useThemeValue } from '../../utils/useThemeValue';
 
+const Message = ({ fill, direction, ...rest }) =>
+  direction === 'row' ? (
+    <Text {...rest} />
+  ) : (
+    <Paragraph {...rest} fill={fill || false} />
+  );
```
taysea, line 25:
```suggestion
  );
  
```

taysea: Just adding an empty line between this and the next function

taysea (approved):
Tiny comment but otherwise looks good!

jcfilben (approved):
Looks good!

Issue #7476, opened 2025-01-23
DataTable -- enhance theme to allow overrides for last row of page


taysea:
We have a desired theming where data cells have a bottom border. That said, to avoid a double border between the last row and footer cells or pagination, the theme should provide the ability to provide specific overrides for the last row of a data table page.

MikeKingdom:
Is `border-collapse: collapse` an option for the `<table>`?

jcfilben:
`border-collapse: collapse` works for when there is a footer in the table but it doesn't work for the Pagination case because the Pagination is outside of the table

PR #7489, opened 2025-01-27
Added `theme.datatable.body.row.extend`

jcfilben:
#### What does this PR do?
Adds `theme.datatable.body.row.extend` so that styles can be added to rows within the DataTable body. This enhancement allows for changes like removing the bottom border of the last row in the body of the datatable so we don't get a double border when we have a footer or pagination.

This can be achieved with:
```
dataTable: {
  body: {
    row: {
      extend: `&:last-child td {
        border: none;
        }
        &:last-child th {
        border: none;
        }`,
    },
  },
},
```

Added jest test and modified custom theme datatable story

Closes https://github.com/grommet/grommet/issues/7476

yes
yes
backwards compatible


MikeKingdom (approved):
LGTM

PR #7490, opened 2025-01-28
Add focus indicator theme option

### LINKED ISSUE #7486: FormField -- theme should support ability for focus to go directly on input
Right now, Grommet always places focus on the formfield container as opposed to the native input. It's desirable to be able to set a flag in the theme to allow focus to be placed on the individual Checkbox/RadioButton/etc directly (following native browser behavior) rather than stealing it to place on formfield.

Current:

Image

Desired:

Image

britt6612:
#### What does this PR do?
This PR adds an option in the theme to remove the `focusIndicator` from Formfield.
FormField.js
locally & wrote a test

test could use some work
locally 

It's desirable to be able to set a flag in the theme to allow focus to be placed on the individual Checkbox/RadioButton/etc directly (following native browser behavior) rather than stealing it to place on formfield.
closes #7486 
yes
yes
backwards compatible


src/js/components/FormField/__tests__/FormField-test.tsx
```diff
@@ -158,6 +160,39 @@ describe('FormField', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
+  test.only('does not apply focusIndicator when formField.focusIndicator is false', () => {
+    const { container } = render(
+      <Grommet
+        theme={{
+          formField: {
+            focusIndicator: false,
```
britt6612, line 168: does this make sense to be `false` 
@taysea @jcfilben

src/js/components/FormField/__tests__/FormField-test.tsx
```diff
@@ -158,6 +160,39 @@ describe('FormField', () => {
...
+      </Grommet>,
+    );
+
+    // Simulate focus
+    fireEvent.focus(screen.getByRole('checkbox'));
+    fireEvent.focus(screen.getByLabelText('Option 1'));
+    fireEvent.focus(screen.getByLabelText('Option 2'));
+    fireEvent.focus(screen.getByRole('slider'));
+
+    expect(container).toMatchSnapshot();
```
britt6612, line 193: Ill look back at the test see if I can check directly for border to be more accurate

src/js/components/FormField/FormField.js
```diff
@@ -526,6 +549,19 @@ const FormField = forwardRef(
...
+          // components that have their own focus indicator
+          const shouldSkipFocus = [
+            'checkbox', // For CheckBox
+            'radio', // For RadioButton
+            'range', // For RangeInput
+          ].includes(targetType.toLowerCase());
+
+          // skip focus
+          if (shouldSkipFocus && theme.formField.focusIndicator === false) {
+            return;
```
jcfilben, line 563: It seems like the onFocus event should still trigger we just shouldn't get the styling

src/js/themes/base.d.ts
```diff
@@ -935,6 +935,7 @@ export interface ThemeType {
         color?: ColorType;
       };
     };
+    focusIndicator?: boolean;
```
jcfilben, line 938: Seems a little weird to have a `focus` object and then separately have `focusIndicator`

src/js/components/FormField/FormField.js
```diff
@@ -261,13 +261,36 @@ const FormField = forwardRef(
...
-          if (
+
+          const isInputComponent =
             child &&
             child.type &&
-            grommetInputNames.indexOf(child.type.displayName) !== -1 &&
+            grommetInputNames.indexOf(child.type.displayName) !== -1;
+          // inputs that would have their own focus indicator
+          // if theme.formField.focusIndicator is false
+          const wantInputFocusIndicator =
```
jcfilben, line 271: Instead could we have something like this at the top of the file similar to how we set up `grommetInputNames` and `grommetInputPadNames`
```
const grommetInputFocusNames = [
  'CheckBox',
  'CheckBoxGroup',
  'RadioButton',
  'RadioButtonGroup',
  'RangeInput',
  'RangeSelector',
  'StarRating',
  'ThumbsRating',
];
```

src/js/themes/base.d.ts
britt6612: yeah maybe put focusIndicator in the focus object or open to other ideas or naming

src/js/components/FormField/FormField.js
```diff
@@ -61,13 +72,24 @@ const isGrommetInput = (comp) =>
   (grommetInputNames.indexOf(comp.displayName) !== -1 ||
     grommetInputPadNames.indexOf(comp.displayName) !== -1);
 
+const getFocusStyle = (props) => {
+  if (
+    props.focus &&
+    props.shouldSkipFocus &&
+    props.theme.formField.focus.childFocus
```
jcfilben, line 79: Do we need `?`s here
```suggestion
    props.theme.formField?.focus?.childFocus
```

src/js/components/FormField/__tests__/FormField-test.tsx
```diff
@@ -158,6 +160,41 @@ describe('FormField', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
+  test.only('does not apply focusIndicator when formField.focusIndicator is false', () => {
```
jcfilben, line 163:
```suggestion
  test('does not apply focusIndicator when formField.focusIndicator is false', () => {
```

src/js/components/FormField/__tests__/FormField-test.tsx
```diff
@@ -158,6 +160,41 @@ describe('FormField', () => {
...
+    const { container } = render(
+      <Grommet
+        theme={{
+          formField: {
+            focus: {
+              nativeFocus: false,
```
jcfilben, line 169: I know the name of this prop might change but just flagging so we don't forget to also change the name here.
```suggestion
              childFocus: false,
```

src/js/components/FormField/FormField.js
```diff
@@ -261,13 +300,25 @@ const FormField = forwardRef(
...
+          if (
+            isInputComponent &&
             child.props.plain === undefined &&
             child.props.focusIndicator === undefined
           ) {
+            // Apply the modified focusIndicator
+            if (wantInputFocusIndicator) {
+              const modifiedFocusIndicator =
+                theme.formField?.focus?.childFocus === true;
+              return cloneElement(child, {
```
taysea, line 318: Generally we try to avoid cloneElement where possible. Wonder if there's any other patterns to explore here.

Should it just be FormField's responsibility to "turn off" its own focus indicator then have the child inputs be responsible for "turning on" their own?

Haven't looked into the code possibility but just a reaction.

src/js/components/FormField/FormField.js
britt6612: yes thought I had them think when I pulled out they got lost Thanks!

src/js/components/FormField/FormField.js
britt6612: yeah.. I talked a little about this with Mike as well and even looking at the children felt a little weird the reason I did it this way is because this was prior code for cloneElement and then setting the following:

```
     return cloneElement(child, {
              plain: true,
              focusIndicator: false,
              pad:
                'CheckBox'.indexOf(child.type.displayName) !== -1
                  ? formFieldTheme?.checkBox?.pad
                  : undefined,
            });

```

that being said I know its not the best but wasnt sure timing if we have to to re-think how we are doing this as well as refactor and even then maybe change FormField to not look at children but maybe have the input check if they are in a FormField then turn on/off focus based on that but again with timing not sure where we stand if we wanted to swarm this or leave the pattern as is.

taysea: I see, thanks for clarifying. I hadn't noticed we were already doing this pattern in this file. I'm comfortable leveraging the existing pattern then and we can always come back in future to revisit it.

src/js/components/FormField/FormField.js
```diff
@@ -30,6 +30,17 @@ import { FormFieldPropTypes } from './propTypes';
...
 
+const grommetInputFocusNames = [
+  'CheckBox',
+  'CheckBoxGroup',
+  'RadioButton',
+  'RadioButtonGroup',
+  'RangeInput',
+  'RangeSelector',
+  'StarRating',
+  'ThumbsRating',
```
taysea, line 41: Should "ToggleGroup" be part of this list?

src/js/components/FormField/FormField.js
```diff
@@ -241,6 +263,20 @@ const FormField = forwardRef(
       return readOnly;
     }, [children]);
 
+    const wantFocusIndicator = useMemo(() => {
```
taysea, line 266:
```suggestion
    const containerFocus = useMemo(() => {
```

taysea: Trying to align logic with theme namespace.

src/js/components/FormField/FormField.js
```diff
@@ -241,6 +263,20 @@ const FormField = forwardRef(
...
+      let focusIndicatorFlag = false;
```
taysea, line 267:
```suggestion
      let focusIndicatorFlag = true;
```

src/js/components/FormField/FormField.js
```diff
@@ -241,6 +263,20 @@ const FormField = forwardRef(
...
+      Children.forEach(children, (child) => {
+        if (
+          child &&
+          child.type &&
+          grommetInputFocusNames.includes(child.type.displayName)
+        ) {
+          focusIndicatorFlag = true;
```
taysea, line 274:
```suggestion
          focusIndicatorFlag = false;
```

src/js/components/FormField/FormField.js
```diff
@@ -428,6 +476,7 @@ const FormField = forwardRef(
           {...themeContentProps}
           {...innerProps}
           {...contentProps}
+          shouldSkipFocus={wantFocusIndicator} // internal prop
```
taysea, line 479:
```suggestion
          containerFocus={containerFocus} // internal prop
```

src/js/components/FormField/FormField.js
```diff
@@ -525,6 +574,7 @@ const FormField = forwardRef(
         margin={abut ? abutMargin : margin || { ...formFieldTheme.margin }}
         {...outerProps}
         style={outerStyle}
+        shouldSkipFocus={wantFocusIndicator} // internal prop
```
taysea, line 577:
```suggestion
        containerFocus={containerFocus} // internal prop
```

src/js/components/FormField/FormField.js
```diff
@@ -261,13 +297,25 @@ const FormField = forwardRef(
...
+            if (wantFocusIndicator) {
+              const modifiedFocusIndicator =
+                theme.formField?.focus?.conatinerFocus === false;
+              return cloneElement(child, {
+                focusIndicator: modifiedFocusIndicator,
+              });
+            }
             return cloneElement(child, {
               plain: true,
               focusIndicator: false,
```
taysea, line 321: Could the above "if (wantFocusIndicator) condition be removed and just change the existing return to be:

```
return cloneElement(child, {
              plain: true,
              focusIndicator: theme.formField?.focus?.containerFocus === true ? false : true,
...
```

src/js/components/FormField/FormField.js
```diff
@@ -61,13 +72,24 @@ const isGrommetInput = (comp) =>
   (grommetInputNames.indexOf(comp.displayName) !== -1 ||
     grommetInputPadNames.indexOf(comp.displayName) !== -1);
 
+const getFocusStyle = (props) => {
+  if (
+    props.focus &&
+    props.shouldSkipFocus &&
+    props.theme.formField?.focus?.conatinerFocus === false
```
taysea, line 79:
```suggestion
    props.theme.formField?.focus?.containerFocus === false
```
typo -- double check other places that might be relying on this logic

src/js/components/FormField/FormField.js
britt6612: yeah good point

taysea: Although wondering if maybe a better name could be chosen? Just a suggestion to try to minimize all the different terms.

taysea (approved):
Behavior looks good.

jcfilben:
I was just expecting the focus to change for these inputs
```
const grommetInputFocusNames = [
  'CheckBox',
  'CheckBoxGroup',
  'RadioButton',
  'RadioButtonGroup',
  'RangeInput',
  'RangeSelector',
  'StarRating',
  'ThumbsRating',
];
```

I didn't expect the focus to change for inputs outside of this list when `focus.containerFocus = false`

This PR:
Screenshot 2025-01-31 at 9 54 17 AM

Previously:
Screenshot 2025-01-31 at 9 55 11 AM

britt6612:
> I was just expecting the focus to change for these inputs
> 
> ```
> const grommetInputFocusNames = [
>   'CheckBox',
>   'CheckBoxGroup',
>   'RadioButton',
>   'RadioButtonGroup',
>   'RangeInput',
>   'RangeSelector',
>   'StarRating',
>   'ThumbsRating',
> ];
> ```
> 
> I didn't expect the focus to change for inputs outside of this list when `focus.containerFocus = false` Here is the Select component This PR: Screenshot 2025-01-31 at 9 54 17 AM
> 
> Previously: Screenshot 2025-01-31 at 9 55 11 AM

good catch @jcfilben fixed now!

jcfilben (approved):
Looks good!

PR #7491, opened 2025-01-28
add option in anchor for gap per size

### LINKED ISSUE #7477: Anchor -- theme should support size-specific gap
Right now, the gap between anchor label + icon can only be defined once (`theme.anchor.gap`) but it would be useful to be able to define a gap per anchor size.

britt6612:
#### What does this PR do?
This PR  adds `anchor.size.gap` define a gap per anchor size.
Anchor.js
locally and test 
locally 

Right now, the gap between anchor label + icon can only be defined once (theme.anchor.gap) but it would be useful to be able to define a gap per anchor size.
closes #7477
yes
yes
backwards compatible


src/js/components/Anchor/StyledAnchor.js
```diff
@@ -17,9 +17,11 @@ const sizeStyle = (props) => {
   if (props.size) {
     const size = props.size || 'medium';
     const data = props.theme.text[size];
+    const gap = props.theme.anchor?.size?.[size]?.gap;
```
jcfilben, line 20: If we are passing `gap` to the Box component in Anchor.js why do we also need to include it here?

britt6612: yeah good call

MikeKingdom (approved):
LGTM

taysea (approved):
Looks good!

src/js/components/Anchor/__tests__/Anchor-test.tsx
```diff
@@ -236,10 +236,23 @@ describe('Anchor', () => {
         textDecoration: 'underline',
         fontWeight: 700,
         size: {
```
jcfilben, line 238: This test might not be the most effective because when it maps through the different sizes it don't create an anchor that would have a gap because the anchors don't have an icon. So we don't actually test the anchor gap.

britt6612: good catch I think Im just missing `small` though 

in the test there is 

```
    <Anchor
          label="Large anchor with icon should receive color on icon"
          icon={<LinkNext />}
          size="large"
        />
```

then 
```
        <Anchor size="medium" label="Anchor with icon" icon={<LinkNext />} />
```

britt6612: ☝️  default is `medium`

src/js/components/Anchor/__tests__/Anchor-test.tsx
```diff
@@ -272,6 +285,11 @@ describe('Anchor', () => {
         </Paragraph>
         <Anchor label="Default anchor with no size prop should receive medium" />
         <Anchor label="Anchor with icon" icon={<LinkNext />} />
+        <Anchor
+          label="Large anchor with icon should receive color on icon"
```
jcfilben, line 289: Should update the anchor name since it isn't a large anchor

jcfilben (approved):
Looks good!

PR #7492, opened 2025-01-29
Fix unFocusStyle for twoColor focus

taysea:
#### What does this PR do?
In https://github.com/grommet/grommet/pull/7479, the twoColor focus strategy was added to focusStyles but the mirrored adjustments were not made in unfocusStyles. This adds the necessary code to ensure focus styles are removed appropriately in the case of twoColor focus approach.

Locally

BEFORE (part of the two color focus still remained when interacting with mouse)

https://github.com/user-attachments/assets/ac7fd563-d22c-41c9-a40c-645968fb30bf

AFTER (focus does not appear when interacting with mouse)

https://github.com/user-attachments/assets/f7f53a9d-e481-492d-be0a-9589c79af4dc

twoColor focus still appears when interacting with keyboard

Screenshot 2025-01-29 at 12 14 28 PM

### DIFF
--- a/src/js/utils/styles.js
+++ b/src/js/utils/styles.js
@@ -292,6 +292,7 @@ const unfocusStyles = (props, { forceOutline, justBorder } = {}) => {
       global: { focus },
     },
   } = props;
+  let compoundFocusStyle = '';
   if (!focus || (forceOutline && !focus.outline)) {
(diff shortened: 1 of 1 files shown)


jcfilben (approved):
Looks good!

halocline (approved):

PR #7494, opened 2025-01-29
Add `theme.menu.container`

taysea:
#### What does this PR do?

This PR adds `theme.menu.container` which styles the Box directly surrounding the menu items. If menu groups are provided, then `theme.menu.group.container` will override theme.menu.container where appropriate.

Adjusted Custom Themed menu storybook.

Added jest test.

Yes.

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -387,7 +387,7 @@ const Menu = forwardRef((props, ref) => {
             />
           </Box>
         )}
-        <Box {...theme.menu.group?.container}>
+        <Box {...theme.menu.container} {...theme.menu.group?.container}>
           {group.map((item) => {
             // item index needs to be its index in the entire menu as if
             // it were a flat array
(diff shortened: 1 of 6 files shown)


src/js/components/Menu/Menu.js
```diff
@@ -444,7 +445,12 @@ const Menu = forwardRef((props, ref) => {
               align.top !== 'bottom'
                 ? controlMirror
                 : undefined}
-              <Box overflow="auto" role="menu" a11yTitle={a11y}>
+              <Box
+                overflow="auto"
+                role="menu"
+                a11yTitle={a11y}
+                {...(!grouped ? theme.menu.container : {})}
```
jcfilben, line 452: Would it make sense to do something like this

If I have `theme.menu.container` defined and `theme.menu.group.container` is undefined then use use `theme.menu.container` regardless of whether the Menu is grouped or not.

If I have `theme.menu.container` defined and `theme.menu.group.container` is also defined then don't use `theme.menu.container` for grouped Menus.

taysea: Good comment -- I think that makes sense. Will adjust the logic

taysea: Done.

jcfilben (approved):
Looks good!

MikeKingdom (approved):
LGTM

PR #7496, opened 2025-01-30
Select, SelectMultiple -- Add extend to listbox

taysea:
#### What does this PR do?

This PR adds an extend to the theme for the "listbox" container directly surrounding the options. Open to naming suggestions, but things like "option.container" etc were already taken and apply to different things. "Listbox" is because this is the container with role="listbox".

Had to pass through `selectMultiple` as a prop on OptionsContainer so that the extend could differentiate between desired styling for Select vs SelectMultiple

src/js/components/Select/StyledSelect.js

Locally. Also added jest tests.

```
theme: {
   select: {
      listbox: {
          extend: `padding: 24px;`,
      },
   },
  selectMultiple: {
      listbox: {
          extend: `padding: 12px;`,
      },
   },
}
```

Yes.

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/Select/__tests__/Select-test.js
+++ b/src/js/components/Select/__tests__/Select-test.js
@@ -1815,6 +1815,39 @@ describe('Select', () => {
     expectPortal('test-clear-selection__drop').toMatchSnapshot();
   });
 
+  test('renders custom listbox styling', () => {
+    jest.useFakeTimers();
+    const customTheme = {
+      select: {
+        listbox: {
+          extend: `padding: 24px;`,
+        },
+      },
+    };
+
+    const { getByPlaceholderText, getByRole } = render(
+      <Grommet theme={customTheme}>
+        <Select
+          data-testid="test-select-style-open"
+          id="test-listbox"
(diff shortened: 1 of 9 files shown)


jcfilben (approved):
Looks good!

src/js/components/Select/__tests__/Select-test.js
```diff
@@ -1815,6 +1815,39 @@ describe('Select', () => {
...
+
+    fireEvent.click(getByPlaceholderText('Select time'));
+    act(() => {
+      jest.advanceTimersByTime(100);
+    });
+
+    expectPortal('test-listbox__drop').toMatchSnapshot();
+    const listbox = getByRole('listbox');
+    const styles = window.getComputedStyle(listbox);
+    expect(styles.padding).toBe('24px');
```
britt6612, line 1848: nice 🙌

britt6612:
LGTM!

Just one small comment! Also `listBox` makes sense to me I think the naming works

britt6612 (approved):

Issue #7495, opened 2025-01-30
FormField -- theme to support input specific container styling


taysea:
FormField should support the ability to modify the container styling based on the type of input. For example:

```
formField: {
   checkBoxGroup: {
      container: {
         border: {},
         disabled: {},
         error: {},
      }
   }
```

We should support this for all our input types.

PR #7498, opened 2025-01-30
Add extend to formfield container based on child input kind

jcfilben:
#### What does this PR do?
Adds an extend to the FormField container to style it differently based on the kind of child element (TextInput, TextArea, etc.).

Marking as a draft because I need to add the TS types and a jest test

Closes https://github.com/grommet/grommet/issues/7495

Yes
Yes
backwards compatible

### DIFF
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -349,7 +349,8 @@ const FormField = forwardRef(
     Children.forEach(children, (child) => {
       if (child && child.type) {
         childName = child.type.displayName;
-        if (childName?.length > 1)
+        // camelCase component name to match theme object key
+        if (childName?.length > 0)
           childName = childName.charAt(0).toLowerCase() + childName.slice(1);
       }
     });
--- a/src/js/components/FormField/FormField.js
+++ b/src/js/components/FormField/FormField.js
@@ -68,6 +68,15 @@ const FormFieldBox = styled(Box)`
 
 const FormFieldContentBox = styled(Box)`
   ${(props) => props.focus && focusStyle({ justBorder: true })}
+  ${(props) =>
+    props.theme.formField &&
+    props.theme.formField[props?.componentName]?.container?.extend}
+`;
+
+const StyledContentsBox = styled(Box)`
+  ${(props) =>
+    props.theme.formField &&
+    props.theme.formField[props?.componentName]?.container?.extend}
 `;
 
 const StyledMessageContainer = styled(Box)`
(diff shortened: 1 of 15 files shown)


src/js/components/FormField/FormField.js
```diff
@@ -336,11 +345,26 @@ const FormField = forwardRef(
       isFileInputComponent = true;
     }
 
+    let childName;
+    Children.forEach(children, (child) => {
+      if (child && child.type) {
+        childName = child.type.displayName;
+        if (childName?.length > 1)
```
taysea, line 352:
```suggestion
        // camelcase component name to match theme object key
        if (childName?.length > 1)
```

taysea: Why do we check > 1?

taysea (commented):
Behavior looking good here. The aesthetics of `disabledProp` and `componentName` feel a bit awkward to me, but I'm comfortable with the trade-off given it'll only be engaged with via `extend`. Agree that this feels like a more reasonable starting point as well than introducing to many theme "container" objects.

src/js/components/FormField/FormField.js
britt6612: ☝️ also was wondering the same thing. I guess Im assuming to ensure that only components with valid names but if we document this as working with only our inputs then Im not sure wed need the check?

jcfilben: Switched this to > 0 instead of > 1

jcfilben: Just wanted to check that the string wasn't `''` or undefined before trying to do `childName.charAt(0).toLowerCase() + childName.slice(1)`

src/js/themes/base.d.ts
```diff
@@ -982,6 +985,71 @@ export interface ThemeType {
         color?: ColorType;
       };
     };
+    checkBoxGroup?: {
```
britt6612, line 988: can we pull out the inputs then do something like 

```
 [(or just check input list): string]: {
    container?: {
      extend?: ExtendType;
    };
  };
```

I know we do something similar with `[size]` and other areas

src/js/components/FormField/__tests__/FormField-test.tsx
```diff
@@ -632,4 +632,27 @@ describe('FormField', () => {
...
+
+  test('Custom theme TextInput FormField container', () => {
+    const { container } = render(
+      <Grommet
+        theme={{
+          formField: {
+            textInput: {
+              container: {
+                extend: (props: any) => css`
+                  color: ${props['href'] ? 'green' : 'red'};
```
britt6612, line 644: are we just showing here how we can use props? I just saw this and thought why would a textInput have `href` or am I thinking of this wrong

src/js/themes/base.d.ts
jcfilben: I originally tried this approach but was getting type errors on the other `theme.formfield.` objects because everything was trying to match the 
```
 container?: {
      extend?: ExtendType;
    };
```
type

src/js/components/FormField/__tests__/FormField-test.tsx
jcfilben: Oops this should be disabled not href, I'll go back and fix

britt6612: all good!

britt6612 (approved):
LGTM

taysea (approved):
lgtm!

PR #7499, opened 2025-01-31
Add theme.global.radius and theme.global.breakpoint[size].radius

taysea:
#### What does this PR do?

Adds `theme.global.radius` and `theme.global.breakpoint[size].radius` to allow for radius values to be styled independently from "edgeSize" (which controls gap, margin, pad).

Adjusts logic to fallback to "edgeSize" if theme.global.radius is not provided.

Add theme.global.radius and theme.global.breakpoint[size].radius

Locally by modifying base theme.

```
theme: {
   global: {
      breakpoint: {
        small: {
            radius: {
              small: '2px',
            },
         },
      },
     radius: {
         small: '38px',
     },
}
```

Yes.

Yes.

Backwards compatible.

### DIFF
--- a/src/js/utils/styles.js
+++ b/src/js/utils/styles.js
@@ -700,17 +700,20 @@ const ROUND_MAP = {
 
 export const roundStyle = (data, responsive, theme) => {
   const breakpoint = getBreakpointStyle(theme, theme.box.responsiveBreakpoint);
+  // fallback to edgeSize for backwards compatibility
+  const radius = theme.global.radius ? 'radius' : 'edgeSize';
+
   const styles = [];
   if (typeof data === 'object') {
     const size =
       ROUND_MAP[data.size] ||
-      theme.global.edgeSize[data.size || 'medium'] ||
+      theme.global[radius][data.size || 'medium'] ||
       data.size;
     const responsiveSize =
       responsive &&
       breakpoint &&
-      breakpoint.edgeSize[data.size] &&
-      (breakpoint.edgeSize[data.size] || data.size);
+      breakpoint[radius]?.[data.size] &&
+      (breakpoint[radius][data.size] || data.size);
     if (data.corner === 'top') {
       styles.push(css`
         border-top-left-radius: ${size};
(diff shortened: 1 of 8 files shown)


src/js/utils/styles.js
```diff
@@ -807,10 +810,10 @@ export const roundStyle = (data, responsive, theme) => {
   } else {
     const size = data === true ? 'medium' : data;
     styles.push(css`
-      border-radius: ${ROUND_MAP[size] || theme.global.edgeSize[size] || size};
+      border-radius: ${ROUND_MAP[size] || theme.global[radius][size] || size};
```
britt6612, line 813:
```suggestion
      border-radius: ${ROUND_MAP[size] || theme.global[radius]?.[size] || size};
```

britt6612, line 813:
```suggestion
      border-radius: ${ROUND_MAP[size] || theme.global[radius]?.[size] || size};
```

src/js/utils/styles.js
```diff
@@ -700,17 +700,20 @@ const ROUND_MAP = {
...
       ROUND_MAP[data.size] ||
-      theme.global.edgeSize[data.size || 'medium'] ||
+      theme.global[radius][data.size || 'medium'] ||
       data.size;
     const responsiveSize =
       responsive &&
       breakpoint &&
-      breakpoint.edgeSize[data.size] &&
-      (breakpoint.edgeSize[data.size] || data.size);
+      breakpoint[radius]?.[data.size] &&
```
jcfilben, line 715: What happens if `theme.global.radius` is defined but `theme.global.breakpoints[size].radius` isn't defined? Should we fall back to looking at `theme.global.breakpoints[size].edgeSize`

taysea: If `responsiveSize` comes back as undefined, then it won't apply the "responsniveStyling" and just will apply from `theme.global.radius` (this would be in alignment with how currently edgeSize and responsive edge size behaves).

Let me know if that makes sense.

jcfilben: Ah okay that makes sense thanks!

jcfilben (approved):
Looks good!

britt6612 (approved):

halocline (approved):

Issue #7233, opened 2024-05-15
Deoendencies - upgrade markdown-to-jsx to latest


jcfilben:
markdown-to-jsx v7.4.5 is causing snapshot changes, verify that the snapshot changes look okay and then upgrade.

localrobot:
@jcfilben `markdown-to-jsx` needs to update to latest. There is a bug in older versions that fails to compile some markdown content. This can be reproduce in [this stackblitz](https://stackblitz.com/edit/vitejs-vite-aeyvogap?file=package.json)

The example should work as is. To reproduce the bug, remove the `overrides` in `package.json`, run `npm install` and then restart the dev server.

PR #7501, opened 2025-02-04
update markdown-to-jsx to latest

localrobot:
#### What does this PR do?

Updates `markdown-to-jsx` to latest

package.json

All automated tests. Some snapshots were updated.

There are some cases where `markdown-to-jsx` fails to compile some valid markdown text. See [this reproduction](https://stackblitz.com/edit/vitejs-vite-aeyvogap?file=package.json)

To Reproduce the issue, remove `overrides` from `package.json`, run npm install and restart the dev server.

#7233 

Yes

Backwards Compatible


src/js/components/Markdown/__tests__/__snapshots__/Markdown-test.tsx.snap
jcfilben, line 1: Some of the snapshot changes seem odd, do you know what is causing the extra spacing?

src/js/components/Markdown/__tests__/Markdown-test.tsx
```diff
@@ -17,16 +17,18 @@ Paragraph
...
 
 > i carry your heart with me
 
 ![alt text](//v2.grommet.io/assets/IMG_4245.jpg "Markdown Image")
 
-Markdown | Less | Pretty
---- | --- | ---
-*Still* | \`renders\` | **nicely**
-1 | 2 | 3
+| Markdown | Less | Pretty |
```
britt6612, line 28: why the change here?

src/js/components/Markdown/__tests__/__snapshots__/Markdown-test.tsx.snap
jcfilben: Ah I see the test changed above so that caused some snapshot changes, does the test still work if we leave it as it was before?

src/js/components/Markdown/__tests__/Markdown-test.tsx
```diff
@@ -17,16 +17,18 @@ Paragraph
 
 #### H4
 
+http://somehere.to.something/
+
```
britt6612, line 21:
```suggestion

```

src/js/components/Markdown/__tests__/__snapshots__/Markdown-test.tsx.snap
anuraghpe: It does not. If leading and trailing `|` is not used, it's no longer recognised as a table.

localrobot: It does not work. If the leading and trailing `|` are removed, it is no longer recognised as a table.

jcfilben: Gotcha, I'm a little worried with the backwards compatibility of this change if this type of table format is no longer supported. I wonder if there is anyway for us to upgrade while maintaining this functionality

PR #7502, opened 2025-02-04
Upgraded version of grommet-icons

jcfilben:
#### What does this PR do?
Bumps grommet-icons dependency to 4.12.4 because 4.12.2 and 4.12.3 have a bug in them

no
yes
backwards compatible


britt6612 (approved):
Looks good

PR #7503, opened 2025-02-06
Rework column widths used for pinning

MikeKingdom:
#### What does this PR do?
This PR improves how column widths are tracked in a DataTable and how they are used to pin columns.

The previous implementation would only record the width of a header cell the first time it was rendered. If something changed later to cause that header cell to change size then where the column is pinned can be wrong. This was causing gaps between pinned columns in cases like the "DataTable/Multiple pins" story:
image

The reason is the HPE theme would load the metric font, but the components get drawn first with a default font (Arial?) while the Metric font is being loaded. Once the font is loaded the columns shrink since the Metric font is a bit narrower.

The `ResizeObserver` is used when available to get notified when the header cell widths change size. This works better than the `element.getClientBoundingRect()` check typically done in a `useLayoutEffect()` because the render of the cell doesn't get retrigged when the css (ie. font finally loading) is what causes the width change. We still fallback to the `getClientBoundingRect()` when the ResizeObserver isn't available (like when server-side rendering is done.)

AFTER fix
image

This also includes a fix for grouped DataTables where pinned columns weren't being rendered as pinned when they should have been.

Note: There is still another problem prexisting where `select` checkbox columns bleed through a pinned footer. This will likely be addressed in a follow on PR:
image

storybook DataTable/Multiple pins

storybook plus a couple hand crafted examples that try to duplicate the COM example (adding and removing columns) as well as pinned columns in grouped DataTables.

grommet/grommet-theme-hpe#422


src/js/components/TableCell/TableCell.js
```diff
@@ -48,13 +48,39 @@ const TableCell = forwardRef(
...
     const containerRef = useRef();
+    const widthRef = useRef();
 
     useLayoutEffect(() => {
+      let resizeObserver;
+      const element = cellRef.current;
       if (onWidth) {
-        const { width } = cellRef.current.getBoundingClientRect();
-        onWidth(width);
+        if (window?.ResizeObserver) {
```
jcfilben, line 57: Should we do `typeof window !== 'undefined'` instead of `window?.`
I know we ran into an issue recently with using `window?.` https://github.com/grommet/grommet-icons/pull/349/

src/js/components/TableCell/TableCell.js
```diff
@@ -48,13 +48,39 @@ const TableCell = forwardRef(
     const tableContext = useContext(TableContext);
     const cellRef = useForwardedRef(ref);
     const containerRef = useRef();
+    const widthRef = useRef();
 
     useLayoutEffect(() => {
+      let resizeObserver;
+      const element = cellRef.current;
```
jcfilben, line 55: Should this be 
```suggestion
      const element = cellRef?.current;
```

jcfilben (commented):
Widths of the pinned columns look good, tested on chome, safari, and firefox

src/js/components/TableCell/TableCell.js
MikeKingdom: cellRef is always defined. It's current may not be (although should be).

src/js/components/TableCell/TableCell.js
MikeKingdom: Great point. Yeah if the variable doesn't exist at all it might be a problem.

jcfilben (approved):
Looks good!

src/js/components/TableCell/TableCell.js
```diff
@@ -48,13 +48,39 @@ const TableCell = forwardRef(
...
+              widthRef.current = width;
+              onWidth(width);
+            }
+          });
+          if (element) {
+            resizeObserver.observe(cellRef.current);
+          }
+        } else {
+          // fallback for server side rendering
+          const { width } = cellRef.current.getBoundingClientRect();
```
halocline, line 71: Above we are getting width from `contentRect` which is getting the inner dimensions of the content (excludes borders and margins). Should we be doing the same here using `clientWidth`? Or is `getBoundingClientRect` what we want?

halocline (commented):
Looking great. Just a couple clarifying questions for my understanding.

src/js/components/DataTable/Header.js
```diff
@@ -151,18 +149,28 @@ const Header = forwardRef(
...
-    useEffect(() => {
-      if (onWidths && cellWidths.length !== 0) {
-        onWidths(cellWidths);
+    const cellWidthsRef = useRef({});
+    const timerRef = useRef();
+
+    const handleWidths = () => {
+      const cellWidths = cellWidthsRef.current;
+      if (onWidths && cellWidths) {
+        const internalColumnWidths =
```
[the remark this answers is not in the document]
MikeKingdom: This is only for the select column so far. Currently the grouped expander cell is never pinned so it never needs to contribute to the pinned offset.

src/js/components/TableCell/TableCell.js
MikeKingdom: That's a great question. getBoundingClientRect() is what it did before and does include borders. The ResizeObserverEntry.contentRect does not include borders as you noted and I should probably be using ResizeObserverEntry.borderBoxSize.  It doesn't matter for our existing styles EXCEPT for resizable columns. Resizable columns end up reporting a smaller width in contentRect (the difference is that `th` has padding.)

halocline (approved):
Thank you, @MikeKingdom !

PR #7505, opened 2025-02-07
redundant state updates, set as `context.globals`

Montana:
Added `useMemo` to prevent unnecessary re-renders when switching themes, ensuring that the `theme` value only recalculates when `selectedTheme` changes. 

Avoiding recursiveness, implemented`useCallback` which I used for setting `rootRef`, avoiding unnecessary function re-creation in the `shadow` root implementation. 

The redundant state updates (`setState` and `setRoot`) have been removed, as `context.globals`.


jcfilben (commented):
When I use the shadowroot mode and refresh the page I get the following error:
Screenshot 2025-02-10 at 5 27 02 PM

PR #7509, opened 2025-02-15
NameValueList should accept width {min: '' max:''}

britt6612:
#### What does this PR do?
This PR adds an option for `max` and `min` to pass through `width` of NameValueList
NameValueList.js
locally 
storybook 

pass `width: [min: 'xsmall', max: 'max-content'] ` to either Grid or NameValueList since under the hood also uses Grid

Ran into the error on the DS site
no
yes
backwards compatible


jcfilben (commented):
Do we need to support an array when we already support an object where min and max can be set?
Screenshot 2025-02-18 at 12 05 00 PM

britt6612:
example of how we are using it. 
```
const nameProps = {
  width: ['xsmall', 'max-content']
};

<NameValueList nameProps={nameProps}>

```

 I think Grommet should update the width type definition for NameValueList as it using Grid underneath and an array with min/max values should be valid.

jcfilben (approved):
Looks good!

src/js/components/NameValueList/index.d.ts
```diff
@@ -7,14 +7,14 @@ export interface NameValueListProps {
   layout?: 'column' | 'grid';
   nameProps?: {
     align?: AlignType;
-    width?: WidthType;
+    width?: WidthType | string[];
```
halocline, line 10: Can we check whether or not `WidthType` is the correct type here? `WidthType` supports accepting an object of
```
{
  min: ...,
  max: ...,
  width: ...,
}
```

Is NameValueList handling this object properly?

I am open to discussion as to whether or not `string[]` is appropriate or not. The only reason we are trying to call `<NameValueList nameProps={[ 'small', 'max-content' ]} ... />` is that we know that `nameProps.width` is being passed to `Grid.columns` and Grid `columns` accepts a variety of configurations including a `string[]`. This doesn't mean that NameValueList's API needs to support this structure, it very well could accept a `{min: ..., max: ...}` object instead, but we'd need to ensure that the object is handled properly in NameValueList implementation, which I don't think it does at the moment.

Is WidthType the proper type?

halocline (changes requested):
Before we move forward with this, I think we should pause and have an conversation about what the API should look like.

I think there are potential advantages to aligning to more of a `{min: ..., max: ...}` which is a convention which is potentially more well understood throughout Grommet. Whereas, the `['minString', 'maxString']` seems to be less common. It is convenient to reach for `string[]` since the underlying implementation is using Grid, but is it the right API for the caller's dev experience?

cc: @jcfilben @MikeKingdom @taysea

halocline:
> But Grid doesn't support an array either, it uses the object approach. So I'm wondering why we wouldn't just do `const nameProps = { width: {min: 'xsmall', max: 'max-content'} };`

Grid `width` does not, but `columns` does which is where this is being consumed. Left comment for discussion around "what the desired API should look like," establish that, then move forward based on that decision.

jcfilben:
Could you update the description no that this PR has changed direction?

jcfilben (approved):

src/js/components/NameValueList/__tests__/NameValueList-test.tsx
```diff
@@ -113,6 +113,24 @@ describe('NameValueList', () => {
     expect(asFragment()).toMatchSnapshot();
   });
 
+  test(`should render correct width with max and min`, () => {
+    const { asFragment } = render(
+      <Grommet>
+        <NameValueList
+          nameProps={{ width: { min: 'xsmall', max: 'max-content' } }}
```
halocline, line 120: Do we have a test scenario which covers `valueProps.width`?

halocline (commented):
Looks good. Just a question about test coverage.

Also, do we have a issue to document this on Grommet site?

src/js/components/NameValueList/__tests__/NameValueList-test.tsx
britt6612: will add one! Also will also add to grommet site

britt6612: https://github.com/grommet/grommet-site/pull/537

src/js/components/NameValueList/__tests__/NameValueList-test.tsx
```diff
@@ -113,6 +113,42 @@ describe('NameValueList', () => {
...
+            </NameValuePair>
+          ))}
+        </NameValueList>
+      </Grommet>,
+    );
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  test(`should render correct width with max and min`, () => {
```
halocline, line 134:
```suggestion
  test(`should render correct valueProps width with max and min`, () => {
```

src/js/components/NameValueList/__tests__/NameValueList-test.tsx
```diff
@@ -113,6 +113,42 @@ describe('NameValueList', () => {
     expect(asFragment()).toMatchSnapshot();
   });
 
+  test(`should render correct width with max and min`, () => {
```
halocline, line 116: Unique test names.

```suggestion
  test(`should render correct nameProps width with max and min`, () => {
```

src/js/components/NameValueList/__tests__/NameValueList-test.tsx
```diff
@@ -113,6 +113,42 @@ describe('NameValueList', () => {
...
+    );
+
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  test(`should render correct width with max and min`, () => {
+    const { asFragment } = render(
+      <Grommet>
+        <NameValueList
+          valueProps={{ width: { min: 'xsmall', max: 'max-content' } }}
```
halocline, line 138: Changing up the values a bit.

```suggestion
          valueProps={{ width: { min: 'small', max: 'medium' } }}
```

halocline (approved):

PR #7523, opened 2025-02-20
need to export dataContext for typescript

britt6612:
#### What does this PR do?
This PR adds a types file for the dataContext so that it can be exported in a TS project
context/dataContext
tested in the DS repo using `yalc`
tested in the DS repo using `yalc`

Im currently trying to build a page in the DS repo and need to access the 
DataContext in order to implement the Selector component we have with being tied to the data. 
no
sure
backwards compatible


taysea (commented):
Question -- once we start to use "DataContext" inside of Data.js, we add a variety of other things to the DataContext: https://github.com/grommet/grommet/blob/master/src/js/components/Data/Data.js#L109

Should each of these be part of the index.d.ts? If we don't include them but then someone tries to use "view" from DataContext, etc. will it throw a TypeScript error?

britt6612:
> Question -- once we start to use "DataContext" inside of Data.js, we add a variety of other things to the DataContext: https://github.com/grommet/grommet/blob/master/src/js/components/Data/Data.js#L109
> 
> Should each of these be part of the index.d.ts? If we don't include them but then someone tries to use "view" from DataContext, etc. will it throw a TypeScript error?

Thats a great point locally let me try and use `view` and check it out

britt6612:
@taysea I checked locally and you are on the right path if I tried to pull out 
```
  const { view, total } = useContext(DataContext);

```
which I had locally I do get a TS error which 
Property 'view' does not exist on type 'DataContextType'
Property 'total' does not exist on type 'DataContextType'.

so that being said Im thinking we would have to provide the other types in this as well.

src/js/contexts/DataContext/index.d.ts
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
+import { View } from '../../components/Data';
+
+interface DataContextType {
+  data: any[];
```
jcfilben, line 5: Will this always be an array of objects or could it be other types?

src/js/contexts/DataContext/index.d.ts
```diff
@@ -0,0 +1,92 @@
...
+            max?: number;
+            min?: number;
+            step?: number;
+          };
+          // for internal filtering only, should searching evaluate this property
+          search?: boolean;
+          sort?: boolean;
+        };
+      };
+  result: any[];
```
jcfilben, line 85: Is result an object or is it an array?

src/js/contexts/DataContext/index.d.ts
```diff
@@ -0,0 +1,92 @@
...
+  selected: any;
```
jcfilben, line 86: Should selected be an array?

src/js/contexts/DataContext/index.d.ts
britt6612: I had it as this since this was the prop in the code
```
  data: dataProp = [],
```

src/js/contexts/DataContext/index.d.ts
britt6612: aw yes I think array should be 

```
  const [selected, setSelected] = useState([]);
```

src/js/contexts/DataContext/index.d.ts
britt6612: object of arrays I can change

src/js/contexts/DataContext/index.d.ts
jcfilben: Seems like it is valid for it to be either an empty array or an array of objects
Screenshot 2025-03-03 at 2 45 49 PM

jcfilben: If there isn't a good way to set the type for that then we can leave as `data: any[];`

britt6612: let me know I changed it to the following:

```
  data: Array<{}> | [];
```

jcfilben (approved):
Looks good!

halocline (approved):
Thank you, @britt6612 .

Can you make sure the public API topic comes up in our next PRty / Grommet working group?

britt6612:
> Thank you, @britt6612 .
> 
> Can you make sure the public API topic comes up in our next PRty / Grommet working group?

will do! @jcfilben do not merge until we discuss 3/13

PR #7524, opened 2025-02-20
extend box types for pagination

britt6612:
#### What does this PR do?
This extends the pagination types to include BoxTypes
index.d.ts
locally 
locally

if you have a typescript project and pass
`border="top"` in which we do for DS styles you will get an error
the `StyledPaginationContainer` is built off of Box which has a `...rest` so we should include boxtypes

DS site
no
yes
 backwards compatible


jcfilben (approved):
Looks good!

MikeKingdom (approved):

PR #7533, opened 2025-03-01
Responsive container

MikeKingdom:
UPDATED: The responsive container functionality is now done via
```jsx
<Box responsive="container">
```

TODO: add jest test case(s)

#### What does this PR do?
`<Box responsive="container"> sets up the ResponsiveContext and any responsive descendents such that their breakpoints are based on the width of this container rather than the entire window. This allows everything inside the container to do responsive things without knowing about what else takes up space around them in the window.

The ResponsiveContainer sets up a "container query" in the CSS which is used instead every where it would normally use a "media query" to change padding, font size etc based on breakpoints.

Note that Layers are still responsive to the whole window since today they still will be modal to the entire window and positioned relative to the window frame.

components/Box/stories/Responsive container

This example shows the same component in two different situations, one inside a `<Box responsive="container">`. The panel on the right can be toggled on and off via the arrow button at the top. You can resize the window and toggle the right panel to see at which points each copy of the component on the left reacts in a responsive way. 
image

Storybook

Storybook

Yes

Yes

Backwards compatible but requires styled-components v6 to be used


taysea (commented):
Thanks for getting this started, Mike. I'm excited about it. A couple of initial questions about API/expected behaviors.

1. Do we want to introduce an entirely new component, or would it be nice to be able to turn any Box into a responsive container via a prop like `container`?
2. Do we expect responsive styling that typically steps down at window breakpoints (spacing, heading sizing, etc.) to apply to content in responsive containers? I suppose my initial thought was people might what to use ResponsiveContainerContext to drive layout things like Grid columns, etc. but I wasn't necessarily expecting the "sizing down" that happens at window breakpoints to apply to responsive containers.

MikeKingdom:
> Thanks for getting this started, Mike. I'm excited about it. A couple of initial questions about API/expected behaviors.
> 
> 1. Do we want to introduce an entirely new component, or would it be nice to be able to turn any Box into a responsive container via a prop like `container`?

Good question and I went back and forth on this. So far I landed on separate component because it adds a lot of special code and useEffects (which themselves can't be conditional, although their contents can be.) Seems like it was a lot of extra stuff in a rather large component. Plus it's something that really is likely only at the top level of an application so far. That said we can certainly discuss more what makes the most sense.

> 2. Do we expect responsive styling that typically steps down at window breakpoints (spacing, heading sizing, etc.) to apply to content in responsive containers? I suppose my initial thought was people might what to use ResponsiveContainerContext to drive layout things like Grid columns, etc. but I wasn't necessarily expecting the "sizing down" that happens at window breakpoints to apply to responsive containers.

The primary use case is basically what you see in the storybook example: one container at the top level of a large piece of functionality so that it's normal window breakpoint behavior doesn't have to worry about something else getting added around it. This originally came about due to how the contextual help side panel totally invalidating the main GLP application's responsiveness. Trying to make all of GLP understand when the panel is visible and somehow make other responsive choices when it is was basically impossible. This is also exactly the use case COM wants it for. 

Of course Matt and I have been also thinking about other situations where this might be useful (and you'd likely want to wrap it in a ThemeContext.Extend that has breakpoints that are best suited for the situation.)

src/js/utils/mixins.js
```diff
@@ -29,12 +29,23 @@ export const fontSize = (size, lineHeight) => css`
...
-  @media only screen ${breakpoint.value &&
-    `and (max-width: ${breakpoint.value}px)`} {
-    ${content};
-  }
-`;
+export const breakpointStyle = (breakpoint, content, responsive) => {
+  const st = 
+  responsive === 'container'
+    ? css`
+        @container ${breakpoint.value && `(max-width: ${breakpoint.value}px)`} {
```
jcfilben, line 36: Will using `@container` cause issues for anyone using v5 of styled-components with Grommet?

MikeKingdom: If I remember correctly it won't create useful css and simply won't work. Either way we'll have to document that this functionality requires styled-components v6+ if you want to use it

src/js/components/Box/Box.js
```diff
@@ -54,6 +56,10 @@ const Box = forwardRef(
 
     const skeleton = useSkeleton();
 
+    const responsiveContainer = useContext(ResponsiveContainerContext);
+    const responsive =
+      responsiveContainer && responsiveProp ? 'container' : responsiveProp;
```
jcfilben, line 61: I'm a little confused about this line, I would expect `responsive={true}` and `responsive='container'` to mean different things. It seems like this line would result in responsive being set to 'container' when the value is `true`.

src/js/utils/mixins.js
```diff
@@ -29,12 +29,23 @@ export const fontSize = (size, lineHeight) => css`
...
-  @media only screen ${breakpoint.value &&
-    `and (max-width: ${breakpoint.value}px)`} {
-    ${content};
-  }
-`;
+export const breakpointStyle = (breakpoint, content, responsive) => {
+  const st =
+    responsive === 'container'
+      ? css`
+          @container ${breakpoint.value &&
```
jcfilben, line 36: Could we add a console info message for people that try to use this feature with styled components v5?

jcfilben (commented):
Originally you had built responsive container as a separate component. You mentioned the reasoning for that was: 

> Good question and I went back and forth on this. So far I landed on separate component because it adds a lot of special code and useEffects (which themselves can't be conditional, although their contents can be.) Seems like it was a lot of extra stuff in a rather large component. Plus it's something that really is likely only at the top level of an application so far. That said we can certainly discuss more what makes the most sense.

The PR has changed direction since then, I'm not opposed to it being built as a prop within Box but wanted to understand the reasoning for the change in direction.

MikeKingdom:
> The PR has changed direction since then, I'm not opposed to it being built as a prop within Box but wanted to understand the reasoning for the change in direction.

Having it just be a simple prop on Box was an original goal but I was worried about the added complexity/bloat to Box. I figured out a decent way to avoid a lot of the complexity and given folks were commenting on avoiding a separate component it seemed preferable to just stick with Box.

src/js/components/Box/Box.js
MikeKingdom: It would. This value is what is passed to the various apis used by the styled components involved (breakpoint css, edge, and round etc.) If `responsiveProp` is true or 'container' (basically not false) this Box needs to do responsive style stuff. If this Box is supposed to be responsive it needs to know whether it's inside a ResponsiveContainerContext (ie some Box way above it in the parentage has `responsive="container"`. The breakpoint style apis need to know this but can't check a context. So this is how we tell them which type of responsive.

src/js/utils/mixins.js
MikeKingdom: Any thoughts on how to detect styled-components v5 vs v6?

src/js/utils/mixins.js
jcfilben: Ah okay thanks for the explaination

src/js/utils/mixins.js
jcfilben: I think we could do:
```
const styled = require('styled-components');
console.log(styled.version);
```

halocline: I think checking the version directly can lead to build errors for folks who might be doing transforms in their bundling. As alternative to directly checking the version, feature detection is another potential pattern --> detect features that differ between v5 and v6. I've seen folks doing something like:

```
// tests `shouldForwardProp`'s behavior change from v5 to v6

export function detectStyledComponentsVersion() {
  const TestComponent = styled.div.attrs({ testProp: 'mytest' })`
   ...
  `;

  const testDiv = document.createElement('div');
  ReactDOM.render(<TestComponent ref={(el) => {
    if (el) {
      if (el.hasAttribute('testProp')) {
        return "v5 or earlier";
      } else {
        return "v6 or later";
      }
    }
  }} />, testDiv);
  return "";
}
```

britt6612: @MikeKingdom did you try something similar to above ☝️

src/js/components/Box/ResponsiveContainer.js
```diff
@@ -0,0 +1,63 @@
...
+} from '../../contexts';
+import { StyledResponsiveContainer } from './StyledResponsiveContainer';
+import { deviceResponsive, getBreakpoint } from '../../utils';
+
+const responsiveContainerValue = true;
+
+export const ResponsiveContainer = ({ children }) => {
+  const theme = useContext(ThemeContext);
+  const ref = useRef(undefined);
+  const [stateResponsive, setResponsive] = useState();
```
britt6612, line 15:
```suggestion
  const [stateResponsive, setStateResponsive] = useState();
```

britt6612: if we change this then have to change below as well

src/js/components/Skeleton/__tests__/__snapshots__/Skeleton-test.tsx.snap
```diff
@@ -634,12 +646,6 @@ exports[`Skeleton with styling props 1`] = `
   -webkit-font-smoothing: antialiased;
 }
 
-@media only screen and (max-width: 768px) {
```
britt6612, line 637: this went away because of this change correct?

```
 ${(props) => props.round && roundStyle(props.round, true, props.theme)}
  ${(props) =>
    props.round && roundStyle(props.round, props.responsive, props.theme)}
```

src/js/components/Box/stories/ResponsiveContainer.js
```diff
@@ -0,0 +1,118 @@
...
+        if (widthRef.current) {
+          widthRef.current.innerText = `Container width: ${width.toFixed()}px`;
+        }
+      });
+      if (element) {
+        resizeObserver.observe(containerRef.current);
+      }
+    } else {
+      // fallback for server side rendering
+      const width = containerRef.current.getBoundingClientRect();
```
britt6612, line 57:
```suggestion
      const width = containerRef.current.getBoundingClientRect().width;
```

would they just want the width of the getBoundingClientRect()??

src/js/utils/mixins.js
MikeKingdom: The solution that seemed to work best was detect that it's pre-v6 by checking for the `withComponent` api that was removed in v6

src/js/components/Skeleton/__tests__/__snapshots__/Skeleton-test.tsx.snap
MikeKingdom: Yes. It wasn't passing along the responsive like it really should.

src/js/components/Box/ResponsiveContainer.js
MikeKingdom: Done

britt6612:
also one thing to increase bundle size

src/js/components/Box/stories/ResponsiveContainer.js
MikeKingdom: Good point. Thanks. This is code a normal human can't hit but I was able to test the fix via the debugger.

src/js/components/Box/Box.js
```diff
@@ -249,6 +256,16 @@ const Box = forwardRef(
       </StyledBox>
     );
 
+    if (responsiveProp === 'container') {
```
britt6612, line 259: nice addition

britt6612 (approved):
LGTM 

Thanks Mike looking forward to this!

src/js/components/Box/StyledBox.js
```diff
@@ -222,22 +230,33 @@ const gapStyle = (directionProp, gap, responsive, wrap, theme) => {
...
         );
       }
     }
   } else if (directionProp === 'column' || directionProp === 'column-reverse') {
     styles.push(`row-gap: ${metric};`);
     if (responsiveMetric) {
-      styles.push(breakpointStyle(breakpoint, `row-gap: ${responsiveMetric};`));
+      styles.push(
+        breakpointStyle(breakpoint, `row-gap: ${responsiveMetric};`),
+        responsive,
```
jcfilben, line 246: Should this be 
```suggestion
        breakpointStyle(breakpoint, `row-gap: ${responsiveMetric};`, responsive),
```

MikeKingdom: Oooo great catch. A couple of spots just above too for gap. And there's also a pre-existing bug here we should fix, maybe as a separate PR? If gap is an object like `gap={{ column: "medium", row: "xsmall" }}` then line `responsiveMetric` at line 200 is always `undefined`.

jcfilben: good to know, I feel like since it is a pre-existing issue it could be a separate PR

jcfilben (approved):
Looks good!

PR #7546, opened 2025-03-18
Fix css gap responsiveness

### LINKED ISSUE #7543: When gap for Box is specified as an object it won't be adjusted at the small breakpoint.
For Box, it's possible to specify separate row and column gaps via an object like `gap={{ row: 'small', column: 'large' }}`. However it doesn't get modified at the responsive breakpoint like a simple string gap.

What's happening is line 200 of StyledBox.js is trying to find the edge size in the breakpoint info from the theme as:
 
```jsx
197  const gapStyle = (directionProp, gap, responsive, wrap, theme) => {
198  const metric = theme.global.edgeSize[gap] || gap;
199  const breakpoint = getBreakpointStyle(theme, theme.box.responsiveBreakpoint);
200  const responsiveMetric = responsive && breakpoint && breakpoint.edgeSize[gap];
201
202  const styles = [];
203  if (typeof gap === 'object') {
204    if (gap.row !== undefined && gap.column !== undefined) {
```
But since gap is an object it never finds anything.
It should probably instead resolve responsiveMetric individually inside the block that starts at 203 (where it knows it is an object)

MikeKingdom:
#### What does this PR do?
This PR fixes Box gap's which have been specified as an object to correctly have `@media` or `@container` styles which modify the gap at the small breakpoint like it would for a non-object gap.

StyledBox.js

Storybook

Storybook Layout/Box/Responsive Container

Fixes #7543 

No

Yes

Backwards compatible

### DIFF
--- a/src/js/components/Box/StyledBox.js
+++ b/src/js/components/Box/StyledBox.js
@@ -201,24 +201,34 @@ const gapStyle = (directionProp, gap, responsive, wrap, theme) => {
 
   const styles = [];
   if (typeof gap === 'object') {
+    const responsiveColumnMetric =
+      responsive && breakpoint && breakpoint.edgeSize[gap.column];
+    const responsiveRowMetric =
+      responsive && breakpoint && breakpoint.edgeSize[gap.row];
     if (gap.row !== undefined && gap.column !== undefined) {
       styles.push(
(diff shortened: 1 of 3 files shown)


britt6612 (approved):
Looks good!

jcfilben (approved):
Looks good!

PR #7547, opened 2025-03-19
Make DataTable rowDetails expansion tracked by primary key rather than index

MikeKingdom:
#### What does this PR do?
This PR fixes an issue from Kari where DataTables rowDetails expansion was tied to a table index. Being based on index wasn't working very well with filter/sort or pagination.

These changes also introduce the ability to use the details row expansion in a "controlled" mode. For example:
```jsx
const [expand, setExpand] = useState([]);
...
<DataTable rowDetails={{
  render: (datum) => <SomeComponent data={datum} />,
  expand,
  onExpand: (expanded, datum) => setExpand(expanded),
}} />
```

This pattern more closely follows how we do selections as well as grouped row expansions.

I re-wrote the rowDetails story to better show a rowDetails example using the StarWars api. It has more interesting details worth showing in a detail row.

DataTable/Body.js

storybook and manual

Storybook Visualizations/DataTable/rowDetails

From Kari's request:

We've been implementing a feature using the rowDetails prop on DataTable.  There are some oddities about how it works, so I'd like to suggest a few enhancements:
- Which rows are expanded is tracked via their index instead of by the primary key. This means if you expand the fifth row and then resort, the fifth row is still expanded even if it has completely different data.  If you tracked expanded rows using primary key, then this would lead to more natural result of the same data being expanded even if it is in a new location in the table
- Similarly, the indexes expanded is the same when you go to a new page of data in a paginated table. So if you had the 2nd row expanded and then go to the next page, the 2nd row of the new page is expanded. This doesn't seem obvious.
- It would also be nice to support get/set props like for select to track and control which rows are expanded.... Similar to select and onSelect props. And/or allow controlling this via Data context props.

image

Yes

Yes

This change should be backwards compatible since we didn't expose the expansion keys before now. It will also use the index as a key if for some reason it can't find a primary key


src/js/components/DataTable/propTypes.js
```diff
@@ -132,7 +132,14 @@ if (process.env.NODE_ENV !== 'production') {
       PropTypes.func,
       PropTypes.oneOf(['select']),
     ]),
-    rowDetails: PropTypes.func,
+    rowDetails: PropTypes.oneOfType([
+      PropTypes.func,
+      PropTypes.shape({
+        render: PropTypes.func.isRequired,
+        expand: PropTypes.arrayOf(PropTypes.string),
```
britt6612, line 139: should we also add number for the case if they use index

src/js/components/DataTable/index.d.ts
```diff
@@ -99,7 +99,13 @@ export interface DataTableProps<TRowType = any> {
       pad?: PadType;
     };
   };
-  rowDetails?: (row: TRowType) => React.ReactNode;
+  rowDetails?:
+    | ((row: TRowType) => React.ReactNode)
+    | {
+        render: (row: TRowType) => React.ReactNode;
+        expand?: Array<string>;
```
britt6612, line 106: should we also add number for the case if they use index

src/js/components/DataTable/Body.js
```diff
@@ -157,7 +163,7 @@ const Row = memo(
         <StyledDataTableRow key={`${index.toString()}_expand`}>
           {(selected || onSelect) && <TableCell />}
           <TableCell colSpan={columns.length + 1}>
-            {rowDetails(data[index])}
+            {rowDetails.render ? rowDetails.render(datum) : rowDetails(datum)}
```
britt6612, line 166: are we sure that datum would be equivalent to `data[index]` or why did you change from `data[index]`

britt6612 (commented):
Thanks Mike for working on this and getting something up I just have a couple questions as far as code but I like the behavior thank you for changing the story its nice to have!

src/js/components/DataTable/Body.js
MikeKingdom: Should be. Everything above this including ExpanderCell, selection Cell, column Cells, etc pass the `datum`. `datum` basically gets set by line 294 of InfiniteScroll.js:
```jsx
items.slice(firstIndex, lastIndex + 1).forEach((item, index) => {
    const itemsIndex = firstIndex + index;
    const child = children(item, itemsIndex);
    result.push(child);
  });
```
the `item` passed to `children` as the first parameter is basically `items[index]` with index corrected for the pagination slice.

It seemed simpler to rely on `datum` consistently in each of these spots.

src/js/components/DataTable/index.d.ts
MikeKingdom: We certainly can although we currently don't for `groupBy.expand`, `expandable` or `select` which are arrays of primary keys

src/js/components/DataTable/Body.js
```diff
@@ -118,10 +117,17 @@ const Row = memo(
             context={isRowExpanded ? 'groupHeader' : 'body'}
             expanded={isRowExpanded}
             onToggle={() => {
```
britt6612, line 119: just wondering if we can make this `onToggle` a little cleaner 


```
           onToggle={() => {
              const rowKey = primaryValue || index;
              const nextRowExpand = isRowExpanded
                ? rowExpand.filter((s) => s !== rowKey)
                : [...rowExpand, rowKey];

              rowDetails.onExpand
                ? rowDetails.onExpand(nextRowExpand, datum)
                : setRowExpand(nextRowExpand);
            }}
            
```

britt6612:
I was trying to think of a way we can also write a test for this behavior but that doesnt have to block this PR. One small comment then everything looks good!

src/js/components/DataTable/Body.js
MikeKingdom: How so? It has to conditionally add/remove the key then conditionally call the callback (controlled) or just do the expand change.

britt6612: aw yes conditionally add/remove ignore the comment

britt6612 (approved):

src/js/components/DataTable/DataTable.js
```diff
@@ -191,7 +191,12 @@ const DataTable = ({
     }
   }, [select, setSelectedDataContext]);
 
-  const [rowExpand, setRowExpand] = useState([]);
+  const [rowExpand, setRowExpand] = useState(rowDetails?.expand || []);
+  useEffect(() => {
```
taysea, line 195: Does this pattern introduce an additional render? Will default state value be set and then also call the useEffect which will set the state value?

taysea (commented):
Overall looks great and storybook example is super helpful/realistic thank you, one small comment.

src/js/components/DataTable/DataTable.js
MikeKingdom: Verified...same number of renders unless caller changes rowDetails.expand

britt6612 (approved):

jcfilben (approved):
Looks good!

PR #7565, opened 2025-04-01
Fix a responsive container issue for sized boxes

MikeKingdom:
With the initial responsive container implementation, it didn't track fixed Box sizes correctly.

#### What does this PR do?
This moves the `container-type` CSS style from a wrapping div to the actual StyledBox div and uses the forwarded ref to track the size of the StyledBox. This size is used to set the breakpoint value provided in the ResponsiveContext at this level as well as the `@container` query for adjusting various edge sizes such for responsive breakpoint behavior.

The generated CSS class names on Box components changed due to the addition of the potential `container-type` style even when it doesn't add the style. I tried a few ways to avoid having it change the actual CSS class name although it still causes the name to be different.

**Note**: The Box test for `as` being a function was giving a warning (thrown as an exception) because refs in that case apparently have no effect. Anytime we use `useForwardedRef()` it always creates a ref. So if in those situations if there is an `as` property that is a function it will throw that warning/error if we pass that ref to the styled component with the `as` property defined. There is no issue if the `as` is a string. So in Grommet as it exists prior to this PR, the following will fail:
```jsx
  const ref = useRef(null);
  useEffect(() => console.log("Ref", ref), [ref]);
  return (
    <Grommet theme={hpe}>
      <Box ref={ref} as={(props) => <header className={props.className} />}>
        <Text>Some content</Text>
      </Box>
    </Grommet>
  );
```

So for any component that is also passed `as` that is a function, it cannot be passed a ref or internally cause a ref to be passed via useForwardedRef (whether or not the component itself was passed a ref.)

In this responsive container use case we simply won't attempt to useForwardedRef if `as` is a function and in that case the container can't really be declared `responsive="container"` 
 
Box.js

Storybook and manual responsive containers

Storybook Layout/Box/Responsive container and 
```jsx
const MyContent = () => {
  const size = React.useContext(ResponsiveContext);
  return (
    <p>
      I am in a <b>{size}</b> container.
    </p>
  );
};

export const FixedSize = () => {
  return (
    <Box responsive="container" border width="300px" pad="medium">
      <MyContent />
    </Box>
  );
```


src/js/components/Box/ResponsiveContainerProvider.js
```diff
@@ -0,0 +1,55 @@
...
+          if (size) {
+            setValue(getBreakpoint(size, theme));
+          }
+        });
+      });
+      resizeObserver.observe(container);
+    }
+    // Initial size for ResponsiveContext in responsive container mode.
+    // Also is a fallback for server side rendering
+    const containerWidth = container.getBoundingClientRect().width;
```
taysea, line 38: @MikeKingdom Is it correct to say that this logic to handle the fixed width is the main difference from the existing useEffect?

src/js/components/Box/Box.js
```diff
@@ -223,7 +235,7 @@ const Box = forwardRef(
         aria-label={a11yTitle}
         background={background}
         border={border}
-        ref={ref}
+        ref={typeof as === 'function' ? undefined : containerRef}
```
taysea, line 238: Can you add an inline comment for why this condition is necessary?

taysea:
Behavior looks as expected 👍 

Screenshot 2025-04-02 at 10 18 29 AM

src/js/components/Box/Box.js
MikeKingdom: Certainly. I have changes in my sandbox also that will do a console.warn in this situation (both the pre-existing ref+as function situation and the responsive="container" + as function situation)

src/js/components/Box/ResponsiveContainerProvider.js
MikeKingdom: I'm not sure what you mean by existing useEffect. This line isn't specific to fixed width. It gives the context an initial width to work with (before the observer gets triggered if at all.)

taysea: "existing useEffect" refers to the useEffect that was in ResponsiveContainer file that was deleted, but this file seems to be largely the same code with some modifications (so that's what I meant by that).

Then for my own understanding, was trying to figure out how we're accounting for fixed width, and I was interpreting that this line accounted for fixed width (where observer might not get triggered), among other potential cases. If that's not the case, can you point me to the logic that does so I can understand?

src/js/components/Box/__tests__/Box-layout-test.tsx
```diff
@@ -31,6 +36,56 @@ describe('Box', () => {
...
+      bottom: 0,
+      x: 0,
+      y: 0,
+      toJSON: () => {},
+    });
+    const { container } = render(
+      <Grommet>
+        <Box responsive="container" width="1280px">
+          <ResponsiveContext.Consumer>
+            {(size2) => size2}
```
taysea, line 55: How come "size2" instead of just "size"?

MikeKingdom: Not needed at this point but I wasn't convinced both size vars weren't in the same scope. Long story, jest dom doesn't really render content fully. As part of the test `render` the getBoundingClientRect will always return 0's and no observers will get triggered (hence the spy.) So while checking this I tried using different vars for the return value to make sure one didn't overwrite the other. The seem to end up as differently scoped vars in the JSX :)

src/js/components/Box/ResponsiveContainerProvider.js
MikeKingdom: Gotcha. This code actually isn't specific to fixed width. There are cases where the StyledBox size will be known and static on the first render (just after which this useEffect is triggered) and the Observer probably won't get triggered. SSR is probably the best example of this. In most browser situations, the Observer will get triggered just after the useEffect. This is why I initially had the useEffect only use the getBoundingClientRect if we didn't have the ability to use an observer. But having it get an initial size is a simple way to make sure.

Here's some debug output on first render of a `<Box responsive="container" width="300px" />`
image

taysea: Awesome, thanks for the thorough explanation.

taysea (approved):
Looks good to me, thank you Mike.

britt6612 (approved):

jcfilben (approved):
Looks good!

Issue #7552, opened 2025-03-21
Tip — Enhance to support ability to dismiss and ability to move cursor into content


taysea:
Tip is currently not passing these WCAG points for WCAG 1.4.13:

1. Dismissible
- A [mechanism](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html#dfn-mechanism) is available to dismiss the additional content without moving pointer hover or keyboard focus, unless the additional content communicates an [input error](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html#dfn-input-error) or does not obscure or replace other content;
2. Hoverable
- If pointer hover can trigger the additional content, then the pointer can be moved over the additional content without the additional content disappearing;

halocline:
Affects tips on ToggleGroup as well.

jcfilben:
Affects FileInput https://github.com/grommet/hpe-design-system/issues/4813

jcfilben:
@britt6612 now that this has been fixed can we go back and update the WCAG documentation on the DS site?

jcfilben:
Actually was cleaning up some other stuff on DS site and just made a PR to update this

PR #7568, opened 2025-04-02
add intelligent margin to Tip

britt6612:
#### What does this PR do?
This PR fixes the `hover` behavior so that when the mouse is moved from the target to the tooltip the tooltip stays open
WCAG 1.4.13 
If pointer hover can trigger the additional content, then the pointer can be moved over the additional content without the additional content disappearing;

tip.js
locally storybook 
storybook mouse over the target then move over the additional content
TODO:

Currently in the grommet/base theme when you hover over a target that has a `tooltip` you can freely move your mouse to the `tooltip` content. 

In the `hpe` theme there is a difference in which when you hover over a target and move your mouse to hover over the `tooltip` content the `tooltip` closes before you can get to the content. 

**The difference between the `hpe` theme and the `base` theme is that in `hpe` theme we are setting `global.drop.margin` in our hpe theme and that is getting applied to the Drop margin component.** 

Screenshot 2025-04-03 at 9 12 47 AM

This margin is getting set on the Drop. What this is doing is adding a little AREA between the target and the `tooltip` so that once the mouse hits that `6px` margin the `tooltip` closes. 

Instead of adding the margin to the Drop via `global.drop.intelligentMargin` and `global.drop.margin` the idea is to take the value from the `global.drop.margin` and if  `global.drop.intelligentMargin` is set to true we are going to add that `margin` to the `margin` around the `div` that is containing the `tooltip` so that it stays open and a user can freely move back and forth from the target to the `tooltip` 

See below for the difference on where the `margin` is hpe theme vs grommet theme

hpe theme `margin` is set on `drop`

Screenshot 2025-04-03 at 9 12 47 AM

grommet theme `margin` is set on `div` around the content of the tip

Screenshot 2025-04-03 at 11 13 25 PM

Current behavior in grommet theme the user can move freely with mouse between the target and tooltip

https://github.com/user-attachments/assets/b095ef32-01e1-4537-af20-dfab64fc8c19

Current behavior in hpe theme the `tooltip` closes when trying to move mouse to content. 

https://github.com/user-attachments/assets/597d1cf6-9610-441b-8834-7ff3282132fd

closes part of #7552 
no
yes
backwards compatible


src/js/components/Tip/Tip.js
```diff
@@ -21,6 +21,90 @@ const Tip = forwardRef(
...
+    // if margin is a string, parse it or use the theme value
+    const normalizeMargin = (margin) => {
+      if (typeof margin === 'string') {
+        const value = theme.global.edgeSize[margin] || parseInt(margin, 10);
+        return { top: value, bottom: value, left: value, right: value };
+      }
+      return { top: 0, bottom: 0, left: 0, right: 0 };
+    };
+
+    const tipMarginStyle = (theme, align) => {
```
britt6612, line 33: passing this theme in is causing a `no-shadow` lint error but when I take it out its not working property any devs have thoughts?

src/js/components/Drop/StyledDrop.js
```diff
@@ -88,7 +88,7 @@ const StyledDrop = styled(Box)`
     marginStyle(
       props.theme,
       props.alignProp,
-      props.theme.global.drop.margin,
+      props.margin || props.theme.global.drop.margin,
```
taysea, line 91:
```suggestion
      props.theme.global.drop.margin,
```
It doesn't seem like we need to pass this through here because it is already passed to the function below on line 93?

src/js/components/Drop/StyledDrop.js
```diff
@@ -54,7 +54,7 @@ const marginStyle = (theme, align, data, responsive, marginProp) => {
   }
   return edgeStyle(
     'margin',
-    adjustedMargin,
+    marginProp || adjustedMargin,
```
taysea, line 57: @jcfilben Curious to get your thoughts, for a theme that has [intelligentMargin: true](https://v2.grommet.io/drop?theme=hpe&c=DwEQTg9gDgQgrgFwRAdgKAAQYDYEMBGAptgLwBEAYrigMYCeGAysYTcmGZhgCaRQDCqBIRQISAby5ZgMCAA8MUXN3J4wAc0JkM+XDQDW6yHBQqy2AJbqAFggC0AJm0YA9AD4uAXy69oAQUt1FAlxZCgALgwAcnwIJAgAWyiAGhxCADMESKjsDIQoz28sXygABT4AZxCMBNwNCxRshMJuCzgkwq5oETR3IA#global.drop.intelligentMargin) (such as hpe), would you expect margin prop when it is a string to only apply to the side between the drop and the target? or all sides?

I'm thinking drop and target.

Right now in the PR it goes around all sides:

Screenshot 2025-04-04 at 2 04 15 PM

src/js/components/Tip/Tip.js
```diff
@@ -82,8 +166,15 @@ const Tip = forwardRef(
           {...dropProps}
           onMouseEnter={() => setTooltipOver(true)}
           onMouseLeave={() => setTooltipOver(false)}
+          margin="none"
```
taysea, line 169: Inline comment for why we are setting this to none

taysea:
I was exploring possible cleanups to the logic for this -- mainly aiming to have one central function that calculates the desired combinedMargin. Still more edge case testing needs to be done, but wondering if this draft PR might help with code refinements? https://github.com/grommet/grommet/pull/7572

I was finding the logic currently a bit difficult to follow the conditions.

src/js/components/Drop/StyledDrop.js
britt6612: @taysea are you asking if intelligentMargin is true and then someone passes in `margin` for a prop?

src/js/components/Drop/StyledDrop.js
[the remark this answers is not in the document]
britt6612: @MikeKingdom I added this because I saw that if someone passed `margin` prop to drop the theme was taking precedence over the prop which we should have the other way around.

src/js/components/Text/__tests__/__snapshots__/Text-test.tsx.snap
```diff
@@ -416,6 +416,18 @@ exports[`renders tip 1`] = `
 
 }
 
+@media only screen and (max-width: 768px) {
+
+}
```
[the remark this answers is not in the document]
britt6612: TODO

src/js/components/Drop/StyledDrop.js
britt6612: If thats the case thats how it works now is its added to every side. Below locally before these changes just on master (with drop change to make prop override theme I have intelligent set to true in theme and then `margin:10px` in theme but in the story Im doing

    <Tip dropProps={{ margin: '20px' }} content="action info">
      <Button label="action" />
    </Tip>

```
and you can see the '20px` is getting set to every side. So thats why I left it like that. to mimic the behavior we have today
<img width="709" alt="Screenshot 2025-04-04 at 5 14 14 PM" src="https://github.com/user-attachments/assets/9cf85d50-4e81-4f69-854d-c809d32c1a98" />

File: src/js/components/Drop/StyledDrop.js (hunk 2 above)
Comment (britt6612) [reply src/js/components/Drop/StyledDrop.js:91]: I was getting an error when I tried removing here. I can look further though if we want to remove
Screenshot 2025-04-04 at 5 27 05 PM

File: src/js/components/Drop/StyledDrop.js (hunk 3 above)
Comment (britt6612) [reply src/js/components/Drop/StyledDrop.js:57]: @MikeKingdom @jcfilben and I talked through this and made sense for us that is someone passes in `margin` as prop respect that they are passing it in and do not change the default behavior so apply margin to all sides

File: src/js/components/Tip/__tests__/__snapshots__/Tip-test.tsx.snap [hunk 6]
```diff
@@ -455,7 +500,25 @@ exports[`Tip should be visible by default 1`] = `
 
 @media only screen and (max-width: 768px) {
   .c2 {
-    margin: 3px;
+    margin-top: 6px;
```
britt6612, line 503: currently not respecting responsive

src/js/components/Tip/__tests__/__snapshots__/Tip-test.tsx.snap
```diff
@@ -567,7 +630,7 @@ exports[`Tip themed 1`] = `
   display: flex;
   box-sizing: border-box;
   max-width: 100%;
-  margin: 21px;
+  margin: 0px;
```
britt6612, line 633: what if we have `theme.tip.drop.margin`

britt6612:
closing in favor of #7589

Issue #6530, opened 2022-12-07
Image component has fallback prop but Avatar component hasn't.


boraoren:

jcfilben:
Resolved by https://github.com/grommet/grommet/pull/7570

PR #7570, opened 2025-04-02
add `imageProps` prop to `Avatar`

max-capozzi-hpe:
#### What does this PR do?
adds the `imageProps` prop to `Avatar` to add missing props for the `Image` used internally by `Avatar`

`Avatar.js`

manually tested with Storybook

Storybook

#6530
Screen Recording 2025-04-02 at 5 06 14 PM
Yes
Yes
Backwards compatible

### DIFF
--- a/src/js/components/Avatar/Avatar.js
+++ b/src/js/components/Avatar/Avatar.js
@@ -11,6 +11,7 @@ const Avatar = ({
   align = 'center',
   children,
   height, // for warning check and discarding the value
+  imageProps,
   justify = 'center',
   round = 'full',
   size = 'medium',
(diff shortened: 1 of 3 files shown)


taysea (commented):
This seems like a reasonable enhancement, thank you!

@jcfilben Are you comfortable with this API? Would there be any reason instead to structure as `imageProps` similar to `dropProps` on other components?

So instead of

```
<Avatar fallback="" />
```

it would be 

```
<Avatar imageProps={{ fallback: '' }} />
```

Not sure if any of the other image props would ever be relevant in Avatar scenario.

jcfilben:
> This seems like a reasonable enhancement, thank you!
> 
> @jcfilben Are you comfortable with this API? Would there be any reason instead to structure as imageProps similar to dropProps on other components?
> 
> So instead of
> `<Avatar fallback="" />`
> it would be
> `<Avatar imageProps={{ fallback: '' }} />`
> Not sure if any of the other image props would ever be relevant in Avatar scenario.

I like the idea of `imageProps` I could see people potentially wanting to pass an aria-label or opacity here in addition to fallback

jcfilben (approved):
Looks good! Thanks @max-capozzi-hpe

jcfilben:
Waiting to merge until after we release v2.46.1

PR #7571, opened 2025-04-03
add aria label for prop to remove button

### LINKED ISSUE #7555: Tag - close button aria-label
On the Tag component there is no aria-label on the close icon button and no way for the implementer to provide one.

This relates to WCAG rule 1.1.1 and 1.3.6

britt6612:
#### What does this PR do?
adds a prop for `aria-label` to be passed in
tag.js
locally
locally
I didnt add one since it got added to other snapshots

On the Tag component there is no way for the implementer to provide an aria-label
closes #7555 
yes
yes
backwards compatible


src/js/components/Tag/Tag.js
```diff
@@ -9,7 +9,10 @@ import { StyledRemoveButton, StyledTagButton } from './StyledTag';
 import { useThemeValue } from '../../utils/useThemeValue';
 
 const Tag = forwardRef(
-  ({ name, value, size = 'medium', onRemove, onClick, ...rest }, ref) => {
+  (
+    { name, value, size = 'medium', onRemove, onClick, ariaLabel, ...rest },
```
taysea, line 13: This approach seems like it would apply to the tag, not the subcomponent. Wondering if we should support

```
removeProps= { 'aria-label': '' }
```

For now I'd say we only let aria-label be the prop in there.

src/js/components/Tag/Tag.js
```diff
@@ -66,6 +69,7 @@ const Tag = forwardRef(
         {onRemove && (
           <StyledRemoveButton
             onClick={onRemove}
+            aria-label={ariaLabel || 'Remove tag'}
```
taysea, line 72: If we add "Remove tag" as a string, we should include it in default.json message and allow for internationalization.

taysea: I also wonder if instead of "Remove tag" as the fallback if it should be `Remove ${name} ${value}`

britt6612: good thinking yeah I can do that

src/js/components/Tag/Tag.js
MikeKingdom: I would agree with this and in fact given how `rest` is put on the container you'd potentially break existing usage of ariaLabel by apps.

src/js/components/Tag/Tag.js
MikeKingdom: Note that to properly localize this it would be no `$` in default.json and the format call would be passed the values:
In default.json
```json
"tag": {
  "removeLabel": "Remove {name} {value}"
},
```
And in Tag.js something like:
```jsx
format({
  id: 'tag.removeLabel',
  messages,
  values: {
    name,
    value
  },
})
```
And you have to document it for Grommet's index.d.ts/prop-types.

Typically we've also added a messages prop to the component so that it can be locally set/overridden (not just globally set in Grommet.) Not sure that's always required.

src/js/components/Tag/Tag.js
```diff
@@ -47,7 +67,7 @@ const Tag = forwardRef(
       console.warn('Tag cannot combine "onClick" and "onRemove".');
     }
 
-    const removeProps = !theme.tag.remove.kind
+    const computedRemoveProps = !theme.tag.remove.kind
```
MikeKingdom, line 70: seems like  better name for this would have "theme" in it.

britt6612: yeah I was wondering on the naming thank you!

britt6612: I changed to `themeTagRemoveProps`

src/js/components/Tag/Tag.js
```diff
@@ -7,12 +7,32 @@ import { Text } from '../Text';
...
-  ({ name, value, size = 'medium', onRemove, onClick, ...rest }, ref) => {
+  (
+    {
+      name,
+      value,
+      size = 'medium',
+      onRemove,
+      onClick,
+      messages,
+      removeProps = { 'aria-label': '' },
```
taysea, line 21:
```suggestion
```
Do we even need this anymore since it's handled my `messages`?

britt6612: aw yes good catch

src/js/components/Tag/Tag.js
```diff
@@ -47,7 +66,7 @@ const Tag = forwardRef(
       console.warn('Tag cannot combine "onClick" and "onRemove".');
     }
 
-    const removeProps = !theme.tag.remove.kind
+    const themeTagRemoveProps = !theme.tag.remove.kind
```
taysea, line 69: I don't think we need to rename this anymore

src/js/components/Tag/Tag.js
```diff
@@ -66,7 +85,9 @@ const Tag = forwardRef(
         {onRemove && (
           <StyledRemoveButton
             onClick={onRemove}
+            {...themeTagRemoveProps}
             {...removeProps}
```
taysea, line 89: Related to above comment, but `removeProps` was renamed and now no longer exists. Getting an error locally for missing variable.

I'd say we don't even need to renamed `removeProps` anymore, but can delete "themeTagRemoveProps"

src/js/components/Tag/__tests__/__snapshots__/Tag-test.tsx.snap
```diff
@@ -551,6 +551,7 @@ exports[`Tag onRemove 1`] = `
       </span>
     </div>
     <button
+      aria-label="Remove undefined Test"
```
taysea, line 554: If name doesn't exist, I wouldn't expect it to include "undefined" in the aria-label. I'd expect this to read "Remove Test"

src/js/components/Tag/Tag.js
britt6612: aw yes will revert

src/js/components/Data/__tests__/__snapshots__/Data-test.tsx.snap
```diff
@@ -400,7 +400,7 @@ exports[`Data controlled search 1`] = `
     <span
       class="c11"
     >
-      4 results of undefined items
+      4 results of  items
```
taysea, line 403: This read a little odd. It feels like ideally if the total wasn't provided it'd read either:
- "4 results" OR
- "4 results of unknown total items" (or something along those lines)

@MikeKingdom you're more expert in localization -- is this something we should worry about or accept this change?

MikeKingdom: Yeah. In general we should never put "undefined" or "null" in a message. But we should also make sure we're not creating a message that isn't "friendly". Typically in this situation the guidance is if you don't know the total then only say "4 results".

src/js/languages/default.json
```diff
@@ -161,6 +161,9 @@
   "tabs": {
     "tabContents": "Tab Contents"
   },
+  "tag": {
+    "removeLabel": "Remove {name} {value}"
```
MikeKingdom, line 165: This probably is a bit simplistic. I think we probably need 3 variations of this message based on whether name or value is defined or both. And the ':' is there in the visible rendering for a reason. Without it these can get confusing.

src/js/components/Tag/__tests__/__snapshots__/Tag-test.tsx.snap
```diff
@@ -551,6 +551,7 @@ exports[`Tag onRemove 1`] = `
       </span>
     </div>
     <button
+      aria-label="Remove Test"
```
MikeKingdom, line 554: We should probably have a test case for the other 2 variations

src/js/components/Tag/index.d.ts
```diff
@@ -5,6 +5,9 @@ export interface TagProps {
   name?: string;
   onClick?: (...args: any[]) => any;
   onRemove?: (...args: any[]) => any;
+  messages?: {
+    removeLabel?: string;
```
MikeKingdom, line 9: this needs to be updated for the new keys

src/js/components/Tag/propTypes.js
```diff
@@ -7,6 +7,9 @@ if (process.env.NODE_ENV !== 'production') {
     value: PropTypes.string.isRequired,
     onClick: PropTypes.func,
     onRemove: PropTypes.func,
+    messages: PropTypes.shape({
+      removeLabel: PropTypes.string,
```
MikeKingdom, line 11: same

MikeKingdom (commented):
Thanks Brittany. Looks like it needs some types updates.

src/js/components/Tag/__tests__/__snapshots__/Tag-test.tsx.snap
britt6612: will do

britt6612:
@MikeKingdom 

glad I added the tests. I saw that `value` is actually required so I changed the json file to just be 2 use cases. 
```
  "tag": {
    "removeLabel": {
      "nameAndValue": "Remove {name}: {value}",
      "valueOnly": "Remove {value}"
    }
  },
```
since `value` is always required we only need `valueOnly` and `nameAndValue` So updates propTypes, TS and tests.

MikeKingdom:
> since `value` is always required we only need `valueOnly` and `nameAndValue` So updates propTypes, TS and tests.

Fair enough although I think that's a silly requirement and people tend to pick one or the other based on whether or not they want it to be bold.

MikeKingdom (approved):

taysea (approved):

jcfilben (approved):
Looks good!

jcfilben:
Holding off on merging until after the upcoming patch release

jcfilben:
Updated with master but looks like we need to bump the bundlesize up

PR #7581, opened 2025-04-08
Make Calendar able to scale down responsively

### LINKED ISSUE #7557: Date Input - 1.4.10 Reflow
**Summary of the issue**
Unable to access calendar when scrolling on small viewport

**Expected Outcome**
The date input calendar must be fully accessible without requiring horizontal scrolling at a viewport width of 320px, ensuring all functionality remains usable.

MikeKingdom:
#### What does this PR do?
This PR makes Calendar able to scale down if the specified size is smaller than the viewport.
This is to meet **WCAG 1.4.10 Reflow**. See #7557 .

Calendar accepts a `size` property that defaults to `medium` (384px). WCAG 1.4.10 requires that certain chunks of information be able to reflow and only require scrolling in 1 direction to see them down to 320px wide. 320px is basically equivalent to a 400% zoom of a 1280 display.

The change adds a `@media` query at the `size` of the Calendar. If the viewport width is smaller than this size this adjusts the styles of the Calendar to narrow. This can be disabled by setting `responsive={false}` or by using `fill`. 

~~Only the width squishes and not the height so the result can be cells that are non-square. Currently I feel like this is acceptable since it would be a lot more work to scale the height also (although ideas are more than welcome.)~~

**UPDATE: Now the Calendar maintains aspect ratio as it shrinks,**

Here's an example using the Input/DateInput/Simple story:
Normal size:
image

At 320px screen width:
image

At 320px screen width with a highlighted/button date:
image

StyledCalendar.js

Jest and Storybook

Any of the Calendar or DateInput storys in storybook

Fixes #7557 

Yes

Yes

Backwards compatible


MikeKingdom:
Just a note: The CSS property `aspect-ratio` seems to be an option in pretty much all browsers now. Using it so the calendar remains square as it scales width down may be a good final piece of this solution.

MikeKingdom:
Merge was hung so replaced this with #7583

PR #7583, opened 2025-04-09
Make Calendar able to scale down responsively

### LINKED ISSUE #7557: Date Input - 1.4.10 Reflow

MikeKingdom:
Replaces #7581 

#### What does this PR do?
This PR makes Calendar able to scale down if the specified size is smaller than the viewport.
This is to meet **WCAG 1.4.10 Reflow**. See #7557 .

Calendar accepts a `size` property that defaults to `medium` (384px). WCAG 1.4.10 requires that certain chunks of information be able to reflow and only require scrolling in 1 direction to see them down to 320px wide. 320px is basically equivalent to a 400% zoom of a 1280 display.

The change adds a `@media` query at the `size` of the Calendar. If the viewport width is smaller than this size this adjusts the styles of the Calendar to narrow. This can be disabled by setting `responsive={false}` or by using `fill`. 

~~Only the width squishes and not the height so the result can be cells that are non-square. Currently I feel like this is acceptable since it would be a lot more work to scale the height also (although ideas are more than welcome.)~~

**UPDATE: Now the Calendar maintains aspect ratio as it shrinks,**

Here's an example using the Input/DateInput/Simple story:
Normal size:
image

At 320px screen width:
image

At 320px screen width with a highlighted/button date:
image

StyledCalendar.js

Jest and Storybook

Any of the Calendar or DateInput storys in storybook

Fixes #7557 

Yes

Yes

Backwards compatible


src/js/components/Calendar/StyledCalendar.js
```diff
@@ -134,6 +195,22 @@ const daySizeStyle = (props) => {
   `;
 };
 
+const responsiveDaySizeStyle = (props) => {
+  const breakpoint = props.theme.global.size[props.sizeProp];
+  const data = props.theme.calendar[props.sizeProp];
+  // try width: 100%; max-width: data.daySize
```
taysea, line 201:
```suggestion
```
This is implemented, can we remove?

src/js/components/Calendar/StyledCalendar.js
```diff
@@ -10,6 +11,20 @@ import {
   styledComponentsConfig,
 } from '../../utils';
 import { activeStyle } from '../../utils/background';
+import { breakpointStyle } from '../../utils/mixins';
+
+const responsiveSizeStyle = (props) => {
+  const breakpoint = props.theme.global.size[props.sizeProp];
+  // try width: 100%; max-width: width
```
taysea, line 18:
```suggestion
```

src/js/utils/mixins.js
```diff
@@ -30,17 +30,21 @@ export const fontSize = (size, lineHeight) => css`
 `;
 
 export const breakpointStyle = (breakpoint, content, responsive) => {
+  const px =
+    typeof breakpoint?.value === 'string' && breakpoint.value.endsWith('px')
```
taysea, line 34: Do we need this additional check? In typescript, the theme-defined breakpoint values only accept number

taysea: Doesn't hurt having it here to be more friendly, but just checking if it's needed.

src/js/components/Calendar/StyledCalendar.js
```diff
@@ -40,9 +56,23 @@ const weeksContainerSizeStyle = (props) => {
...
   `;
 };
+
+const weeksContainerResponsiveSizeStyle = (props) => {
+  const breakpoint = props.theme.global.size[props.sizeProp];
+  return breakpointStyle(
+    { value: breakpoint },
+    `
+    height: auto;
+    aspect-ratio: 7/6;
```
taysea, line 66: Remind me why 7/6? Could we add a comment for future reference?

taysea (commented):
Behavior and styling looks great. Just added a few small comments. Nice find with `aspect-ratio`.

src/js/utils/mixins.js
MikeKingdom: Existing breakpoints used by `breakpointStyle` are numbers so the existing code made them into a string with `px` on the end. The t-shirt size values used by this are already strings with `px` on the end. We're using the latter as the value for the media query for this situation

src/js/components/Calendar/StyledCalendar.js
MikeKingdom: Answered via slack but replying here also for others benefit. 7 days per week, 6 weeks shown. Pre-existing code does 14.6% width (100%/7) and 6*daySize height.

jcfilben: Good to know, agree adding a comment in the code would be helpful for future

jcfilben (commented):
Code looks good

src/js/components/Calendar/StyledCalendar.js
MikeKingdom: Done

src/js/components/Calendar/StyledCalendar.js
MikeKingdom: Done

src/js/components/Calendar/StyledCalendar.js
MikeKingdom: Done

jcfilben (approved):
Looks good!

taysea (approved):
Thanks, Mike!

PR #7586, opened 2025-04-11
add aria-role for when defaultsuggestion is applied

### LINKED ISSUE #7551: TextInput — defaultSuggestion causes unclear screen reader experience
When a screenreader user moves to the textinput with defaultSuggestion defined, they don't get information about the input field and instead the screenreader jumps to reading out 'suggestion 2'. Could make it confusing for a screen reader user to understand what element they're on. Related to WCAG 3.2.1

https://github.com/user-attachments/assets/20b25427-5880-44b3-85a9-1992ea3d271f

britt6612:
#### What does this PR do?
This PR fixes the voice over on the TextInput with `defaultSuggestion`
textinput.js
locally with voice over
locally with voice over

The moment `aria-expanded=true`, and `aria-activedescendant` is set (which happens when activeSuggestionIndex is not -1),
 VoiceOver:
* Assumes the focus is within the listbox,
* Skips past the text input, and
* Announces the active option right away.

Right now I added when a user is using `keyboard` the focus is set on the `textInput` until they `arrowdown` then set the `aria-activedescendant` to whatever the `defaultSuggestion` is 
closes #7551 
no
yes
 backwards compatible

### DIFF
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -363,7 +363,8 @@ const TextInput = forwardRef(
           <ContainerBox
             aria-label={
               defaultSuggestion !== undefined
-                ? 'Suggestions list for text input'
+                ? `combobox with suggestions type
+                 text to display a list of suggestions`
                 : undefined
             }
             id={id ? `listbox__${id}` : undefined}
@@ -408,11 +409,6 @@ const TextInput = forwardRef(
                       ref={itemRef}
                     >
                       <Button
-                        aria-label={
-                          defaultSuggestion !== undefined
-                            ? `You are on the default suggestion ${index + 1}`
-                            : undefined
-                        }
                         id={id ? `listbox-option-${index}__${id}` : undefined}
                         role="option"
                         aria-selected={selected ? 'true' : 'false'}
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -363,8 +363,10 @@ const TextInput = forwardRef(
           <ContainerBox
             aria-label={
               defaultSuggestion !== undefined
-                ? `combobox with suggestions type
-                 text to display a list of suggestions`
+                ? format({
+                    id: 'textInput.defaultSuggestion',
+                    messages,
+                  })
                 : undefined
             }
             id={id ? `listbox__${id}` : undefined}
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -362,7 +362,7 @@ const TextInput = forwardRef(
         >
           <ContainerBox
             aria-label={
-              defaultSuggestion !== -1
+              defaultSuggestion !== undefined
                 ? 'Suggestions list for text input'
                 : undefined
             }
@@ -409,7 +409,7 @@ const TextInput = forwardRef(
                     >
                       <Button
                         aria-label={
-                          defaultSuggestion !== -1
+                          defaultSuggestion !== undefined
                             ? `You are on the default suggestion ${index + 1}`
                             : undefined
                         }
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -23,7 +23,6 @@ import {
   sizeStyle,
   useForwardedRef,
   useSizedIcon,
-  useKeyboard,
 } from '../../utils';
 
 import {
@@ -122,22 +121,6 @@ const TextInput = forwardRef(
 
     const [focus, setFocus] = useState();
     const [showDrop, setShowDrop] = useState(false);
-    const [arrowDownInitiated, setArrowDownInitiated] = useState(false);
-
-    const usingKeyboard = useKeyboard();
-    useEffect(() => {
-      const handleKeyDown = (event) => {
-        if (event.key === 'ArrowDown') {
-          setArrowDownInitiated(true);
-        }
-      };
-
-      document.addEventListener('keydown', handleKeyDown);
-
-      return () => {
-        document.removeEventListener('keydown', handleKeyDown);
-      };
-    }, []);
 
     const handleSuggestionSelect = useMemo(
       () => (onSelect && !onSuggestionSelect ? onSelect : onSuggestionSelect),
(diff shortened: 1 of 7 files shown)


britt6612:
TODO: update bundlesize

jcfilben (commented):
Actually just noticed a few things

On the TextInput/Suggestions story I have to press down arrow twice to get into the suggestion list.

In the TextInput/Default Suggestion story the first time I tab to the textinput and press down arrow to move into the drop down, things work well. However if I move focus away from the textinput and then move back again, the default suggestion gets autofocused as soon as my focus moves to the textinput

src/js/components/TextInput/TextInput.js
```diff
@@ -121,6 +122,22 @@ const TextInput = forwardRef(
...
+
+    const usingKeyboard = useKeyboard();
+    useEffect(() => {
+      const handleKeyDown = (event) => {
+        if (event.key === 'ArrowDown') {
+          setArrowDownInitiated(true);
+        }
+      };
+
+      document.addEventListener('keydown', handleKeyDown);
```
MikeKingdom, line 135: Shouldn't we be using Keyboard instead of directly doing window listeners like this, especially since this depends on events bubbling all the way to the window level

src/js/components/TextInput/TextInput.js
```diff
@@ -121,6 +122,22 @@ const TextInput = forwardRef(
 
     const [focus, setFocus] = useState();
     const [showDrop, setShowDrop] = useState(false);
+    const [arrowDownInitiated, setArrowDownInitiated] = useState(false);
+
+    const usingKeyboard = useKeyboard();
+    useEffect(() => {
+      const handleKeyDown = (event) => {
+        if (event.key === 'ArrowDown') {
+          setArrowDownInitiated(true);
```
MikeKingdom, line 131: I'm not really seeing how this is reliable. The first time an ArrowDown happens the arrowDownInitiated is set true and never changes after that. Nothing sets it back to false. Does that mean we never expect to be able to get back to a state where this should be false?

britt6612: good call out I would set back the state to false when the drop closes but going to do a deeper look into this with Bill

src/js/components/TextInput/TextInput.js
```diff
@@ -361,6 +361,11 @@ const TextInput = forwardRef(
           {...dropProps}
         >
           <ContainerBox
+            aria-label={
```
britt6612, line 364: @jcfilben @taysea 
I can go back and add this into our default.json message and allow for internationalization but wanted to have you guys check first if the voice over made sense while going through the default suggestion story

src/js/components/TextInput/TextInput.js
```diff
@@ -403,6 +408,11 @@ const TextInput = forwardRef(
                       ref={itemRef}
                     >
                       <Button
+                        aria-label={
+                          defaultSuggestion !== undefined
+                            ? `You are on the default suggestion ${index + 1}`
```
taysea, line 413: Why do we need to add this here rather than just rely on what was already being read out by the button label / children?

britt6612: So while I was going through the component with Bill we were looking at our `defaultSuggestion` story in which we are setting `suggestion 2` as the `defaultSuggestion` when Bill was going through the component he was very confused as to why he was on `suggestion 2` and not `suggestion 1` I was trying to explain to him that we had the prop and that we are putting the user on `suggestion 2` so he had said if there was a way for us to make that clear for a screen reader so that is why I added the `You are on the default suggestion` I would change `${index + 1}` to be whatever the defaultSuggestion is I just had it set to `2` so that way you can test with the storybook example

britt6612: This isnt new to the changes I have made ☝️ I know its not ideal so I will check in this PR if I can take that option out or I can create a new issue and handle it separately

taysea: I see, I wonder if the fact that these are "suggestion 1", "suggestion 2", etc. exacerbated the confusion. Whereas if the example was just "Apples", "Bananas", etc. he maybe wouldn't have been as focused on if he was on the first one or not. 

In the real world, it feels like `defaultSuggestion` would be leveraged in cases where it would be useful for someone to already be on that suggestion (meaning someone wouldn't really be thinking "where is the first suggestion?").

I think the aria-label on the container is helpful (and agree about moving it to "default.json" so it can be internationalized, but I'm not feeling like it's required to add the aria-label to the individual suggestion at the moment. And if we did -- we wouldn't want it to key off of the index but rather read out the text within the suggestion I'd think.

britt6612: thats fair I see your point and yeah this story isnt really "real life" or simulates a real experience. Will move things to "default.json"

src/js/languages/default.json
```diff
@@ -165,7 +165,8 @@
     "enterSelect": "(Press Enter to Select)",
     "suggestionsCount": "suggestions available",
     "suggestionsExist": "This input has suggestions use arrow keys to navigate",
-    "suggestionIsOpen": "Suggestions drop is open, continue to use arrow keys to navigate"
+    "suggestionIsOpen": "Suggestions drop is open, continue to use arrow keys to navigate",
+    "defaultSuggestion": "TextInput with suggestions type text to display a list of suggestions"
```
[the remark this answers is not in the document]
britt6612: I did notice that all of those announcements are not being read out I talked to Jessica about that but she said since some of them have `roles` that they might be overriding our announcements so their not getting read out.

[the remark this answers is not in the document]
britt6612: @jcfilben @taysea to discuss next week.

britt6612:
closing in favor of https://github.com/grommet/grommet-site/pull/540

During testing, we observed a discrepancy in how screen readers interpret the component behavior—specifically between VoiceOver and JAWS. With JAWS, the behavior is correct and expected: the screen reader first announces the TextInput field before moving to the defaultSuggestion, providing a clear and accessible user experience.

However, when testing with VoiceOver, we found that it bypasses the TextInput entirely and jumps straight to the defaultSuggestion. This behavior results in a confusing experience for users relying on VoiceOver, as it obscures the main input field and focuses prematurely on the suggestion, which is meant to be supplementary.

We discussed implementing a workaround or fix to address this inconsistency in VoiceOver. However in this case, attempting to correct the VoiceOver behavior risked breaking the already functional and compliant behavior in JAWS.

Because of this, we decided against implementing a technical fix that might degrade the experience for JAWS users. Instead, we opted to document the limitations `defaultSuggestion` prop to be used with caution.

PR #7588, opened 2025-04-15
Add combobox props to MaskedInput with suggestions

### LINKED ISSUE #7573: MaskedInput -- Should announce when masked options drop is open
The "options" for MaskedInput mask sections should behave more closely to TextInput suggestions with `comboxBoxProps`. This will help inform screen reader users about the presences of suggestions for each masked section.

Related to WCAG 1.3.1

taysea:
#### What does this PR do?

Adds [`combobox`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role) props to MaskedInput (similar to TextInput logic).

Slight modifications to the logic so that each mask section can appropriately inform screen reader if it's expanded.

According to ARIA documentation, an "id" is needed to associate the input with the combobox, so added "id" to the stories in MaskedInput and TextInput that involve the suggestions.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role

Locally in SizeUnits story with VoiceOver.

Same as above.

Closes #7573 

Yes. Fixed MaskedInput to support `combobox` aria properties for dropdown options.
 
Backwards compatible.

### DIFF
--- a/src/js/components/MaskedInput/MaskedInput.js
+++ b/src/js/components/MaskedInput/MaskedInput.js
@@ -369,6 +369,34 @@ const MaskedInput = forwardRef(
 
     const maskedInputIcon = useSizedIcon(icon, rest.size, theme);
 
+    /*
+    If the masked input has a list of options, add the WAI-ARIA 1.2
+    combobox role and states.
+    */
+    let comboboxProps = {};
+    let activeOptionID;
+    const options = useMemo(() => {
+      let res;
+      if (!activeMaskIndex)
+        // ensures that comboboxProps gets set on input initially
+        res = mask.find((item) => item?.options?.length > 0)?.options;
+      else res = mask[activeMaskIndex]?.options;
+      return res;
+    }, [mask, activeMaskIndex]);
+
+    if (id && options?.length > 0) {
+      if (showDrop && options) {
+        activeOptionID = `listbox-option-${activeOptionIndex}__${id}`;
+      }
+      comboboxProps = {
+        'aria-activedescendant': activeOptionID,
+        'aria-autocomplete': 'list',
+        'aria-expanded': showDrop ? 'true' : 'false',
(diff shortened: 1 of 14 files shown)


src/js/components/MaskedInput/MaskedInput.js
```diff
@@ -369,6 +369,34 @@ const MaskedInput = forwardRef(
...
+      return res;
+    }, [mask, activeMaskIndex]);
+
+    if (id && options?.length > 0) {
+      if (showDrop && options) {
+        activeOptionID = `listbox-option-${activeOptionIndex}__${id}`;
+      }
+      comboboxProps = {
+        'aria-activedescendant': activeOptionID,
+        'aria-autocomplete': 'list',
```
MikeKingdom, line 393: cool.

MikeKingdom (approved):
lgtm

jcfilben (commented):
Looks like we are missing an id in the MaskedInput Date, DateRange, and Filtered stories

jcfilben (commented):
when I arrow through the items in the drop it reads out the item and says 'completion selected', I'm confused why it is reading that out
Screenshot 2025-04-16 at 1 44 34 PM

taysea:
> completion selected

I believe that is the automatic readout that happens when you've selected something and then it's active again. For example, here's what happens in Headless UI combobox

Screenshot 2025-04-16 at 1 28 38 PM

jcfilben:
Taylor and I chatted more about the 'completion selected' announcement. Seems like something was changed recently in voiceover with this behavior. It also isn't consistent in how it reads out even with suggestion drop downs outside of Grommet. This behavior seems outside of our control so leaving things as they are for now.

jcfilben (approved):
Looks good!

PR #7589, opened 2025-04-16
add buffer box to keep tooltip open

### LINKED ISSUE #7552: Tip — Enhance to support ability to dismiss and ability to move cursor into content

britt6612:
#### What does this PR do?
Added a `bufferBox` to lay on top of `margin` so that when the user hovers over that `bufferBox` the tip stays open.
tip.js
storybook in hpe theme 
storybook in hpe theme

In the current hpe theme we are passing `theme.global.drop.margin` in which a `margin` is being placed on the Drop component. When the user was hovering over this `margin` the tooltip was being closed. 

I have no added a `bufferBox` that will lay over the `margin` that is being set to that way when the user hovers over this space the tooltip stays open.
closes #7552
no
yes
backwards compatible

### DIFF
--- a/src/js/components/Tip/Tip.js
+++ b/src/js/components/Tip/Tip.js
@@ -23,14 +23,15 @@ const BufferBox = styled(Box)`
   width: ${(props) => props.width}px;
   height: ${(props) => props.height}px;
 `;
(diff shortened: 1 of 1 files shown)


britt6612:
closing in favor of #7596

PR #7592, opened 2025-04-17
add second option for adding another div for margin

### LINKED ISSUE #7552: Tip — Enhance to support ability to dismiss and ability to move cursor into content

britt6612:
#### What does this PR do?
This PR adds the additional of 2 divs in order to have the `margin` set on the outer `div` in order to allow the user to move their mouse from the target to the tooltip without the tooltip closing. 
tip.js
storybook
storybook

If `theme.global.drop.margin` is set and is a string as well as `theme.global.drop.intelligentMargin`  is true then we will go through to calculate where the `margin` should be placed depending on the alignment of the drop. 

Now if `theme.global.drop.intelligentMargin` is false then we just take the value that is either passed into `theme.global.drop.margin` or `theme.tip.drop.margin` or `dropProps.margin` 
closes #7552 
no
yes
 backwards compatible

### DIFF
--- a/src/js/components/Button/__tests__/__snapshots__/Button-test.tsx.snap
+++ b/src/js/components/Button/__tests__/__snapshots__/Button-test.tsx.snap
@@ -5139,6 +5139,22 @@ exports[`Button tip 1`] = `
 
 }
 
+@media only screen and (max-width: 768px) {
+
+}
+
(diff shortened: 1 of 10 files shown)


britt6612:
closing in favor of #7596

PR #7595, opened 2025-04-22
use timeout approach to keep the tooltip opened

britt6612:
#### What does this PR do?
use `timeout` approach to keep the `tooltip` open when the user hovers from target to tooltip
tip.js
locally 
locally in hpe theme

hpe theme has `margin` around Drop so when the user hovers over that `margin` the tooltip closes


britt6612:
closing in favor of #7596

PR #7596, opened 2025-04-22
look at marginProp before setting adjustedMargin

### LINKED ISSUE #7552: Tip — Enhance to support ability to dismiss and ability to move cursor into content

britt6612:
#### What does this PR do?
This PR address that `tip.drop.margin` should take priority over `global.drop.margin` currently in our hpe theme we have `global.drop.margin` set as well as `intelligentMargin` so whats happening is those are being applied to Tip even though we have `tip.drop.margin` set to "none"

I have add the check that if `marginProp` is set to something then that should come before the adjustedMargin in the `marginStyle` function in Drop. 
styledDrop
locally 
locally 

in hpe theme you will see that the only margin being applied is the `theme.tip.content.margin` in which we will update this value in the hpe them to either `6px` or `9px` depending on discussion with designers. 

`margin` on the drop was causing the tooltip not to stay open as a user hovered from the target to the tooltip
closes #7552 
no
yes
 backwards compatible

### DIFF
--- a/src/js/components/Drop/StyledDrop.js
+++ b/src/js/components/Drop/StyledDrop.js
@@ -29,6 +29,8 @@ const dropKeyFrames = keyframes`
 
 // The desired margin may be adjusted depending on drops alignment
 const marginStyle = (theme, align, data, responsive, marginProp) => {
+  // NOTE: If marginProp is passed, it overrides the alignment-aware
+  //  margin logic and uses the provided value instead.
   const margin = theme.global.edgeSize[data] || data;
   let adjustedMargin = {};
   // if user provides CSS string such as '50px 12px', apply that always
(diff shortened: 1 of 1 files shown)


src/js/components/Drop/StyledDrop.js
```diff
@@ -54,7 +54,7 @@ const marginStyle = (theme, align, data, responsive, marginProp) => {
   }
   return edgeStyle(
     'margin',
-    adjustedMargin,
+    marginProp || adjustedMargin,
```
britt6612, line 57: this way `marginProp` that is set to 'none' will come before `adjustedMargin`

taysea (approved):
I am comfortable with this. Upon initial review, I was wondering if `theme.tip.drop.margin` should also adhere to "intelligentMargin" (which would make the margin alignment aware). In looking at the implementation of Tip, "theme.tip.drop" is passed onto Drop as a prop, so allowing for that would further complicate this logic.

Given the use case of Tip with mouseover should ideally be leaving margin as none anyway, I'm comfortable with this.

jcfilben (approved):
Looks good!

PR #7597, opened 2025-04-22
#7562 fix file-input accessibility iasue 1.4.13

### LINKED ISSUE #7562: File Input - WCGA Rule 1.4.13 Content on Hover or Focus
Issue:
Keyboard users can’t access the tooltip that shows the full file name when it gets cut off, which fails WCAG 1.4.13.

Fix:
Make the tooltip accessible with keyboard focus, or show the full file name in another accessible way (like in a focusable element or aria-label).

Image

Sulaymon333:
#### What does this PR do?
- Fix accessibillity issue 1.4.13 for file input component
- fixes #7562 
- Integrate file input to use Grommet Tip for file name instead of native browser solution
image
image

### DIFF
--- a/src/js/components/FileInput/FileInput.js
+++ b/src/js/components/FileInput/FileInput.js
@@ -18,6 +18,7 @@ import { Button } from '../Button';
 import { FormContext } from '../Form/FormContext';
 import { Keyboard } from '../Keyboard';
 import { Text } from '../Text';
+import { Tip } from '../Tip';
 
 import { StyledFileInput } from './StyledFileInput';
 import { FileInputPropTypes } from './propTypes';
@@ -461,29 +462,42 @@ const FileInput = forwardRef(
                 justify="between"
                 direction="row"
                 align="center"
+                style={{ zIndex: 0 }} // make the tip appear on file name hover
               >
                 {renderFile ? (
                   renderFile(file)
                 ) : (
-                  <Box
-                    {...theme.fileInput.label}
-                    gap="xsmall"
-                    align="center"
-                    direction="row"
+                  <Tip
+                    content={file.name}
+                    dropProps={{
+                      align: { top: 'bottom' },
+                      width: 'small',
+                      round: 'xsmall',
+                      background: { ...theme.tip.drop },
+                    }}
                   >
-                    {((maxSize && file.size > maxSize) ||
-                      (max && index >= max)) && <CircleAlert />}
-                    <Label
-                      weight={
-                        theme.global.input.weight ||
-                        theme.global.input.font.weight
-                      }
-                      truncate
-                      {...passThemeFlag}
+                    <Box
+                      {...theme.fileInput.label}
+                      gap="xsmall"
+                      align="center"
+                      direction="row"
+                      tabIndex={0} // make name focusable with keyboard
+                      flex // make file with shortname have larger hover area
                     >
-                      {file.name}
-                    </Label>
-                  </Box>
+                      {((maxSize && file.size > maxSize) ||
+                        (max && index >= max)) && <CircleAlert />}
+                      <Label
+                        weight={
+                          theme.global.input.weight ||
+                          theme.global.input.font.weight
+                        }
+                        truncate
+                        {...passThemeFlag}
+                      >
+                        {file.name}
+                      </Label>
+                    </Box>
+                  </Tip>
                 )}
                 <Box flex={false} direction="row" align="center">
                   <Button


Sulaymon333:
I guess, we might need to update the focus style when using keyboard, right?

src/js/components/FileInput/FileInput.js
```diff
@@ -461,29 +462,42 @@ const FileInput = forwardRef(
...
-                        theme.global.input.font.weight
-                      }
-                      truncate
-                      {...passThemeFlag}
+                    <Box
+                      {...theme.fileInput.label}
+                      gap="xsmall"
+                      align="center"
+                      direction="row"
+                      tabIndex={0} // make name focusable with keyboard
```
britt6612, line 484: why do we need to make the name focusable?
even currently on storybook it still reads out the full time even if visually you cant see the whole file name

Sulaymon333: I believe is for keyboard only users who don't use a screenreader and hence they won't have access to the full name when they navigate with Keyboard only

britt6612: that makes sense.

So then in that case can we only show the `tip` if the text is truncated? right now its showing on everything.

src/js/components/FileInput/FileInput.js
```diff
@@ -461,29 +462,42 @@ const FileInput = forwardRef(
...
                   renderFile(file)
                 ) : (
-                  <Box
-                    {...theme.fileInput.label}
-                    gap="xsmall"
-                    align="center"
-                    direction="row"
+                  <Tip
+                    content={file.name}
+                    dropProps={{
```
britt6612, line 472: I dont think we need any of these `dropProps` can we just let the drop be styled from the theme with whats already there

src/js/components/FileInput/FileInput.js
britt6612, line 484: in the case you mentioned 
`keyboard only users who don't use a screenreader and hence they won't have access to the full name when they navigate with Keyboard only
`

can we only set this `tabIndex={0} // make name focusable with keyboard` for when the text is truncated

MikeKingdom: That doesn't seem like the right thing to do. Also, instead of an explicitly adding Tip you should just be able to change Label's truncate to `truncate="tip"` We shouldn't have to one-off solutions like this

MikeKingdom:
Brittany and I did some experimenting with using the `truncate="tip"`. It has some conflicts with the built in behavior from the input due to how we're rendering the filenames by stacking them on top of the input with its content hidden via opacity.  There is a built in tooltip for the FileInput that we would want to disable somehow. It seems to still be triggerable and wins in some situations (and there is a tooltip that comes up on hover when no files are selected.)

src/js/components/FileInput/FileInput.js
britt6612: I think we need to have a larger discussion between devs on this issue since this rule would effect more then just fileinput.

1. In dataTable what if we have a name/label and it has a tooltip do we now need to make that name/label focusable as well? 
2. If we have a long text on a page and it is truncated and shows a tooltip does that text then have to be focusable?

☝️ just a couple things Mike and I thought of so think @Sulaymon333 if you wanted to have a dev call and discuss this further with the team.

Sulaymon333:
@MikeKingdom @britt6612 I have experimented on the built in tooltip as suggested but based on the internal meeting we agreed to do more reseaarch into this first. So my findings from some other react frameworks are summarized below.

[Prime React](https://primereact.org/fileupload/#accessibility) - FileUpload component

- Long filename is NOT truncated, long file names is wrapped to next line
- file name is not focusable just the remove button
- voiceover DOES NOT announce filename when navigating with tab and remove icon is in focus. it just says image, button group.
- *no default file input tooltip observed. input filetype visually hidden*

[Carbon Design FileUploader ](https://react.carbondesignsystem.com/?path=/story/components-fileuploader--default)component
- long name is truncated as well
- file name is not focusable just the remove button
- voiceover announces FULL filename when navigating with keyboard and remove icon is in focus. Announces Delete file <full file name>
- _no default file input tooltip observed. input filetype visually hidden_

[Chakra UI FileUpload component](https://www.chakra-ui.com/docs/components/file-upload)

- long name is truncated as well
- file name is not focusable just the remove button
- voiceover announces FULL filename when navigating with keyboard and remove icon is in focus. Announces Delete file <full file name>
-  _no default file input tooltip observed. input filetype visually hidden_

[Ant Design fie upload](https://ant.design/components/upload) component

- long name is truncated as well
- file name is not focusable but image preview and remove button are focusable
- voiceover DOES NOT announce filename when navigating with tab and remove icon is in focus. it just says delete file
- _no default file input tooltip observed. input filetype visually hidden_

[GOV UK Design System file upload component](https://design-system.service.gov.uk/components/file-upload/)

- for new improvement they made recently, Long filename is not truncated. long file names is wrapped to next line
- no remove button, did not see multiple file upload
- no default file input tooltip observed. input filetype visually hidden. hidden=true
- [Mozilla Protocol from Mozilla file component](https://protocol.mozilla.org/components/detail/input--file.html)
- very simple and basic
- _Default file input tooltip is present_

[Syncfusion file uploader](https://ej2.syncfusion.com/react/documentation/uploader/getting-started)

- long name is truncated in the middle (shows starts and end)
- voiceover DOES NOT announce filename when navigating with tab and remove icon is in focus. it just says remove empty group
- _Default file input tooltip is present_

[Polaris file upload component](https://polaris-react.shopify.com/components/selection-and-input/drop-zone?example=drop-zone-with-custom-file-upload-text)

- Long filename is NOT truncated, long file names is wrapped to next line
- _no default file input tooltip observed. input filetype visually hidden_

**My take**

1. We can follow the multiline filename and fix height for filenames container box with vertical scroll. Already file more than ten are not displayed, only 11 files is selected.
2. to get rid of native file input tooltip, we have to hide it and then rework the drop zone to work. I am experimenting with associating a label via htmlFor and id but I don't know how this will play well if file upload component is placed within a Grommet form component.
3. My observation is that, what is common is to truncate the name in some sorts either at the end, mostly common or just in-between showing start and end. I believe this is not very accessible for keyboard only users although for voiceover, Grommet file upload announces full file name when navigating with only keyboard when delete icon is in focus. but keyboard only users loos out on the full name.

**Other information**

to get rid of native file input tooltip,
setting title='' or title='U+2800' is not working on safari
preventing default behavior on mousever is not working either

let me know your thoughts moving forward. Thank you.

image
image
image
image
image
image

Issue #7556, opened 2025-03-26
Date Input - 2.1.3 Keyboard No exception


CillianHPE:
**Summary of the issue**
When the calendar closes, the focus doesn't return to the input.

**Expected Outcome**

When the calendar popup closes:
	•	After selecting a date: Focus should return to the date input field.
	•	After pressing Escape or clicking away: Focus should return to the triggering element (input or toggle button).
	•	Ensure focus is not lost or trapped, maintaining a clear, logical flow for keyboard users.

jcfilben:
Resolving this issue will likely resolve [4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG22/#name-role-value) on the DateInput component as well

jcfilben:
In addition we should fix the focus indicator so that it moves to the dates on calendar instead of being on the calendar container

PR #7612, opened 2025-04-29
Change Calendar keyboard behavior to roving tab index and clean up DateInput focus

jcfilben:
#### What does this PR do?
This PR does two things.

The majority of the changes are related to changing the Calendar from having focus on the entire calendar container to focus on the individual dates. This change is needed to give people enough of a visual indication of which date is active/in focus.

The second change was to DateInput. This change ensures that when the drop is closed focus is returned back to the input.

https://github.com/grommet/grommet/issues/7556
Closes https://github.com/grommet/grommet/issues/5677

no
yes
backwards compatible


src/js/components/Calendar/Calendar.js
```diff
@@ -962,66 +973,73 @@ const Calendar = forwardRef(
...
                 aria-label={`${reference.toLocaleDateString(locale, {
                   month: 'long',
                   year: 'numeric',
                 })}; ${currentlySelectedString(value, locale)}`}
-                ref={daysRef}
                 sizeProp={size}
                 fillContainer={fill}
                 responsive={responsive}
-                focus={focus}
                 onFocus={() => {
```
taysea, line 1038: Are the onFocus/onBlur handlers needed anymore if this div won't have `tabIndex={0}`?

src/js/components/Calendar/Calendar.js
```diff
@@ -200,59 +199,67 @@ export const getOutputFormat = (dates) => {
...
-          >
-            {children}
-          </StyledDay>
-        )}
-      </StyledDayButton>
-    </StyledDayContainer>
-  );
-};
+        <StyledDayButton
+          role="gridcell"
```
taysea, line 231: Can you explain why we moved `role="gridcell"` to the button instead of the container?

src/js/components/Calendar/Calendar.js
```diff
@@ -847,21 +847,24 @@ const Calendar = forwardRef(
         if (!children) {
           days.push(
             <CalendarDay
+              activeProp={active}
+              day={day}
               key={day.getTime()}
+              disabledProp={dayDisabled && !!dayDisabled}
```
taysea, line 853: Why do we move the disabled logic here instead of passing through buttonProps?

src/js/components/Calendar/Calendar.js
```diff
@@ -911,6 +915,11 @@ const Calendar = forwardRef(
       </StyledWeek>,
     );
 
+    useEffect(() => {
```
taysea, line 918: Can we add a comment for what this useEffect is responsible for?

src/js/components/DateInput/DateInput.js
```diff
@@ -396,7 +395,14 @@ Use the icon prop instead.`,
     if (open && !readOnly) {
       return [
         input,
-        <Keyboard key="drop" onEsc={() => ref.current.focus()}>
+        <Keyboard
+          key="drop"
+          onEsc={() => {
+            setTimeout(() => {
+              ref?.current.focus();
+            }, 1);
```
taysea, line 403: Generally it feels like we try to avoid timeouts like this. Is the calendar capturing the Esc event? or placing focus elsewhere in response?

Something feels a little fishy here.

src/js/components/Calendar/Calendar.js
```diff
@@ -200,59 +199,67 @@ export const getOutputFormat = (dates) => {
...
-      </StyledDayButton>
-    </StyledDayContainer>
-  );
-};
+        <StyledDayButton
+          role="gridcell"
+          aria-current={activeProp?.getTime() === day?.getTime()}
+          aria-disabled={disabledProp}
+          aria-selected={isSelected}
+          disabledProp={disabledProp}
```
taysea, line 235:
```suggestion
          disabled={disabledProp}
```
Should this be passed to `disabled` so it's passing through as a button prop?

src/js/components/Calendar/Calendar.js
```diff
@@ -847,21 +847,24 @@ const Calendar = forwardRef(
...
+                    ? () => {}
+                    : () => onClick(dateObject),
               }}
               isInRange={inRange}
               isSelected={selected}
               otherMonth={day.getMonth() !== reference.getMonth()}
               rangePosition={rangePosition}
               size={size}
               fill={fill}
+              ref={focusableDateRef}
```
taysea, line 867: To make sure I'm understanding, we track this as a ref so we can programmatically focus the "active" day? If so, how come the `tabIndex={0}` can't just take care of that?

src/js/components/Calendar/Calendar.js
```diff
@@ -847,21 +847,24 @@ const Calendar = forwardRef(
...
+              disabledProp={dayDisabled && !!dayDisabled}
               buttonProps={{
                 a11yTitle: day.toDateString(),
-                active: active && active.getTime() === day.getTime(),
-                disabled: dayDisabled && !!dayDisabled,
-                onClick: () => onClick(dateObject),
-                onMouseOver: () => setActive(dateObject),
-                onMouseOut: () => setActive(undefined),
+                onClick:
+                  dayDisabled && !!dayDisabled
```
taysea, line 857: Shouldn't `onClick` not have an effect is the button is disabled? Seems like we're doubling up on that logic here

src/js/components/Calendar/Calendar.js
jcfilben: aria-selected needs to go on the same element as `role="gridcell"` so I moved it to the button. I can explore adding the aria attributes to the container instead.

src/js/components/DateInput/DateInput.js
jcfilben: From what I understand when we press the escape key the drop closes causing a re-render. At the same time the onEsc function gets triggered. If we try to do `ref?.current.focus();`, ref.current won't be defined yet because of the re-render caused by the drop closing. However if we add a timeout, then ref.current will be defined and we can set focus back on the DropButton.

Basically just moved what was on line 244 to line 401

src/js/components/Calendar/Calendar.js
jcfilben: I didn't want the button to get the disabled prop so that we could still navigate to it and instead use aria-disabled to signify that it can't be interacted with. See https://github.com/grommet/grommet/pull/7612#discussion_r2073752541

src/js/components/Calendar/Calendar.js
jcfilben: > To make sure I'm understanding, we track this as a ref so we can programmatically focus the "active" day?
Correct

> If so, how come the tabIndex={0} can't just take care of that?
Setting tabIndex to 0 allows something to receive focus but adding/removing `tabIndex={0}` doesn't automatically change where the focus is placed

src/js/components/Calendar/Calendar.js
jcfilben: I'm not setting the button to disabled but rather using aria-disabled on the button so that's why I added this check

src/js/components/Calendar/StyledCalendar.js
```diff
@@ -180,9 +176,11 @@ const responsiveDayButtonStyle = (props) => {
 
 // when caller opts in to day hover styling, apply all state styles
 // on CalendarDay instead of active state on CalendarDayButton
+// ${(props) =>
+//   props.theme.calendar?.day?.hover?.background && 'background: inherit;'}
 const StyledDayButton = styled(Button)`
   ${(props) =>
-    props.theme.calendar?.day?.hover?.background && 'background: inherit;'}
+    props.disabledProp && disabledStyle(props.theme.button.disabled.opacity)}
```
taysea, line 183: Since some themes (like HPE Theme) don't use opacity for disabled button but rather disabled color, we should make sure to apply the entirety of disabled theming here (given we aren't passing `disabled` through to Button directly:

I added bounds to calendar range story:

Screenshot 2025-05-06 at 4 13 00 PM

src/js/components/Calendar/StyledCalendar.js
```diff
@@ -10,7 +11,7 @@ import {
   roundStyle,
   styledComponentsConfig,
 } from '../../utils';
-import { activeStyle } from '../../utils/background';
+// import { activeStyle } from '../../utils/background';
```
taysea, line 14:
```suggestion
```
Remove?

src/js/components/Calendar/StyledCalendar.js
```diff
@@ -180,9 +176,11 @@ const responsiveDayButtonStyle = (props) => {
 
 // when caller opts in to day hover styling, apply all state styles
 // on CalendarDay instead of active state on CalendarDayButton
+// ${(props) =>
```
taysea, line 179: Remove?

src/js/components/Calendar/Calendar.js
taysea: When I use keyboard (not screen reader), I get the following behavior:
- First when focus moves into the calendar it's on a day that's outside the bounds I've defined (visually disabled, has aria-disabled="true").
- I can't use keyboard to move to any other disabled days, but I can move to a non-disabled day that's adjacent.

My experience is that the disabled dates are not navigable (this feels right to me, but is counter to what you described in your comment as the desired behavior, so I might need to understand that better) but I would've expected the initial focus to go onto the first enabled day.

https://github.com/user-attachments/assets/6feae5bd-f9cd-41eb-9e79-4dbc8fa15e10

src/js/components/Calendar/Calendar.js
taysea: I see -- let's continue this question in the comment thread you linked!

src/js/components/Calendar/Calendar.js
taysea: I supposed my question is why we need to programmatically focus a day at all?

src/js/components/Calendar/Calendar.js
taysea: I see -- let's continue this thread about how to treat "disabled" in this comment thread: https://github.com/grommet/grommet/pull/7612#discussion_r2076519357

src/js/components/DateInput/DateInput.js
taysea: I see, thanks for clarifying! Hadn't realized the logic was also already part of the component.

src/js/components/Calendar/Calendar.js
jcfilben: Discussed on a call, just summarizing here for other people. Because the roving tab index pattern moves focus via arrow keys instead of the tab button, focus does need to be handled programmatically

src/js/components/Calendar/StyledCalendar.js
jcfilben: Taylor and I discussed this more. The issue is that StyledCalendar is looking for disabled styles to be at the root level of `props.theme.button.disabled` for themes that have kind buttons. In the hpe theme we don't define disable styles at `props.theme.button.disabled` but instead at the `props.theme.button.disabled[kind]` level. We decided to leave the grommet code as is and change the hpe theme to have disabled styles defined at `props.theme.button.disabled`. I will put up a separate PR for this.

src/js/components/Calendar/StyledCalendar.js
jcfilben: Removed

src/js/components/Calendar/Calendar.js
jcfilben: Taylor and I discussed this on a call and over slack messages. Summarizing here for anyone else.

Disabled dates and out of bounds dates have different behavior when it comes to keyboard navigation.

A disabled date can receive focus and be navigated to via arrow keys

An out of bounds date can not receive focus and can't be navigated to via arrow keys.

The idea here is that a disabled date could land in the middle of dates that are enabled and it is reasonable to expect that someone needs to navigate past them. It is helpful for them to still be read out to screenreader users and not have the keyboard navigation jump past them so that we don't break up the flow of the calendar.

However if a date is out of bounds then people don't have a need to navigate past it so we don't support focus or navigation on these dates.

src/js/components/Calendar/Calendar.js
taysea: So the `onFocus` is being called here more so as a function of the focus event bubbling up rather than the actual container receiving focus?

src/js/components/Calendar/__tests__/Calendar-test.tsx
```diff
@@ -916,3 +791,104 @@ describe('Calendar Keyboard events', () => {
     expect(asFragment()).toMatchSnapshot();
   });
 });
+
+describe('Calendar Keyboard events', () => {
```
taysea, line 795: Is there a reason we had to move the test order? Or just preference?

src/js/themes/base.js
```diff
@@ -617,10 +617,10 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         // adjacent: {
         //   color: undefined,
         // },
-        // hover: {
-        //   background: undefined,
-        //   color: undefined,
-        // },
+        hover: {
+          background: 'active-background',
```
taysea, line 621: Just confirming this is because we're removing the concept of "active" and now we want to enforce it here?

For best semantic use of tokens, can you also include a change to grommet-theme-hpe in the PR you plan to file which adjust this to:
```
hover: {
   background: 'background-hover',
}
```

taysea (commented):
This is behaving well / how I would expect based on discussions. I dropped a few last comments, but I think it's in a great direction for another dev to pick up for further reviewing.

src/js/components/Calendar/__tests__/Calendar-test.tsx
jcfilben: I wanted the 'heading font size' test to not be grouped into the keyboard events describe block and instead be grouped in the main calendar test describe block

src/js/themes/base.js
jcfilben: > Just confirming this is because we're removing the concept of "active" and now we want to enforce it here?

Yes

> For best semantic use of tokens, can you also include a change to grommet-theme-hpe

Yes, I'll plan to make that change

src/js/components/Calendar/Calendar.js
```diff
@@ -847,21 +846,24 @@ const Calendar = forwardRef(
...
```
MikeKingdom, line 856: You do this pattern in a lot of places and it could be error prone or a maintenance issue if somewhere else forgets to force it to a boolean. Maybe it would be best to change lines 835-837 to take care of it so you can just use `dayDisabled`
```jsx
        const dayDisabled = !!(
          withinDates(day, normalizeInput(disabled))[0] ||
          (bounds && !betweenDates(day, normalizeInput(bounds))));
```

src/js/components/Calendar/Calendar.js
```diff
@@ -847,21 +846,24 @@ const Calendar = forwardRef(
...
+                    ? () => {}
+                    : () => onClick(dateObject),
```
MikeKingdom, line 858: With the above suggestion this could be simplified:
```suggestion
                onClick: () => if (dayDisabled) onClick(dateObject),
```

src/js/components/Calendar/Calendar.js
```diff
@@ -847,21 +846,24 @@ const Calendar = forwardRef(
...
               }}
               isInRange={inRange}
               isSelected={selected}
               otherMonth={day.getMonth() !== reference.getMonth()}
               rangePosition={rangePosition}
               size={size}
               fill={fill}
+              ref={focusableDateRef}
```
MikeKingdom, line 866: you aren't tracking custom days so does that mean they don't get the same navigation?

src/js/components/DateInput/DateInput.js
MikeKingdom, line 403: Any reason this is 1 rather than 0? I think the goal is just to schedule this immediately after the current thread of execution

MikeKingdom (commented):
The focus border gets cut off in various ways for the days and even the month arrows (all themes and sizes I believe)
image
We should probably add an issue to address this.

Also we definitely need to refactor Calendar.js. It is pretty huge these days and has a lot of complicated logic making it really hard to analyze.

src/js/components/Calendar/Calendar.js
jcfilben: Yeah I didn't go into custom dates for this PR. Seems like the customDates needs a decent amount of work o held off for now. If a calendar is build with customDates it is all tab based navigation.

src/js/components/DateInput/DateInput.js
jcfilben: Updated to use 0 instead of 1

src/js/components/Calendar/Calendar.js
jcfilben: fixed in latest commit

src/js/components/Calendar/Calendar.js
jcfilben: Good call, fixed in latest commit

jcfilben:
> The focus border gets cut off in various ways for the days and even the month arrows (all themes and sizes I believe)

I forgot that Taylor and I discussed this and looked at adding position relative and a setting z-index of the calendar date button in focus to 1. Added this in the latest commit.

The focus indicator does still get cut off on the edges of the calendar but we couldn't find a good way around this and it seems like other calendar components have this same issue.

While not ideal, if the focus indicator does get cut off the we are not meeting WCAG AAA, but still meeting WCAG AA.

src/js/themes/base.js
jcfilben: Just checked, looks like we already have `calendar.day.hover.background` set to 'background-hover' in the hpe theme

src/js/components/Calendar/StyledCalendar.js
jcfilben: PR created for grommet-theme-hpe https://github.com/grommet/grommet-theme-hpe/pull/469

src/js/components/Calendar/StyledCalendar.js
```diff
@@ -123,12 +123,7 @@ const rangeRoundStyle = (props) => {
     themeObj =
       props.theme.calendar?.[props.sizeProp]?.range?.round ||
       props.theme.calendar?.medium?.range?.round;
-  return (
-    themeObj && [
-      roundStyle(themeObj, props.responsive, props.theme),
-      'overflow: hidden;',
```
britt6612, line 129: why did we remove the overflow?

jcfilben: The overflow hidden was preventing the focus indicator from being displayed, and when I removed it I didn't see any negative side effects from not having it

src/js/components/Calendar/Calendar.js
```diff
@@ -911,6 +915,32 @@ const Calendar = forwardRef(
...
+        if (!betweenDates(active, normalizedBounds)) {
+          const diff1 = Math.abs(
+            active.getTime() - normalizedBounds[0].getTime(),
+          );
+          const diff2 = Math.abs(
+            active.getTime() - normalizedBounds[1].getTime(),
+          );
+          let closerDate = 1;
+          if (diff1 < diff2) {
+            closerDate = 0;
```
britt6612, line 937: are we ever reassigning this? could we clean up?
```const closerDate = diff1 < diff2 ? 0 : 1;```

**Review by britt6612 — COMMENTED** [review]
A couple comments but behavior looks good!

File: src/js/components/Calendar/Calendar.js [hunk 19]
```diff
@@ -962,66 +994,73 @@ const Calendar = forwardRef(
...
+                  }
                 }
               }}
               onRight={() => {
-                setActive(addDays(active, 1));
-                if (!betweenDates(addDays(active, 2), displayBounds)) {
-                  changeReference(active);
+                const nextActive = addDays(active, 1);
+                if (
+                  !!betweenDates(nextActive, normalizeInput(bounds)) ||
```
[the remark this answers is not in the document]
jcfilben: betweenDates returns 2, 1, or undefined so it doesn't return a boolean

src/js/components/Calendar/Calendar.js
jcfilben: Good call, just pushed up a change to fix this

src/js/components/Calendar/Calendar.js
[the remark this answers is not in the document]
britt6612: ah okay thank you for confirming

britt6612 (approved):

src/js/components/Calendar/Calendar.js
```diff
@@ -200,61 +199,69 @@ export const getOutputFormat = (dates) => {
...
-          >
-            {children}
-          </StyledDay>
-        )}
-      </StyledDayButton>
-    </StyledDayContainer>
-  );
-};
+        <StyledDayButton
+          aria-disabled={disabledProp}
```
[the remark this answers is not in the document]
jcfilben: Done intentionally, see conversation here: https://github.com/grommet/grommet/pull/7612#discussion_r2069462130

src/js/components/DateInput/DateInput.js
```diff
@@ -203,9 +205,14 @@ Use the icon prop instead.`,
...
     const closeCalendar = useCallback(() => {
+      if (usingKeyboard && !inline && ref?.current) {
+        setTimeout(() => {
+          ref?.current?.focus();
+        }, 0);
+      }
       setOpen(false);
       announce(formatMessage({ id: 'dateInput.exitCalendar', messages }));
-    }, [announce, formatMessage, messages]);
+    }, [announce, formatMessage, messages, usingKeyboard, ref, inline]);
```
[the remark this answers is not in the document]
jcfilben: Done intentionally to prevent the focus indicator from showing when we are navigating via mouse

taysea:
On the "Dual" story, I lose focus indicator (it seems to skip the calendar all together):

https://github.com/user-attachments/assets/c26a0830-e33c-4f16-9d1a-81a898db17eb

jcfilben:
Just pushed up changes to resolve how the focus is handled when showAdjacentDays is false. The Dual story still doesn't work well but I think minly due to pre-existing issues with that story.

I tested by adding `showAdjacentDays={false}` to the simple calendar story

taysea:
> Just pushed up changes to resolve how the focus is handled when showAdjacentDays is false. The Dual story still doesn't work well but I think mainly due to pre-existing issues with that story.
> 
> I tested by adding `showAdjacentDays={false}` to the simple calendar story

In the case of showAdjacentDays={true}, I'd expect it to go onto the first visible day. The current behavior always puts it on the first of the month.

https://github.com/user-attachments/assets/a9d982ca-e203-4a6a-9b7e-3cb7dbfba257

taysea:
In cases where there are multiple calendars on the screen, once I select a day, the focus is returned to the body for some reason. I would expect it to still be on the day I just selected. When I inspect, it looks like the day I just selected has `tabIndex={-1}`

https://github.com/user-attachments/assets/c8a2a46a-6546-458c-9609-494f0f7ff03c

jcfilben:
> In cases where there are multiple calendars on the screen, once I select a day, the focus is returned to the body for some reason. I would expect it to still be on the day I just selected. When I inspect, it looks like the day I just selected has tabIndex={-1}

Taylor and I chatted offline, summarizing our discussion here. The issue isn't related to having multiple calendars in this story but rather the use of `toDateString` to set the initial value. Initial values should be passed in as iso strings as noted in the documentation otherwise the internal calculations will not function as intended. I have update the story to use `toISOString` instead of `toDateString` in the latest commit

src/js/components/Calendar/Calendar.js
```diff
@@ -132,14 +132,27 @@ const normalizeRange = (value, activeDate) => {
...
 };
 
-const getReference = (reference, value) => {
+const getReference = (reference, value, activeDate) => {
   let nextReference;
   if (value) {
     if (Array.isArray(value)) {
       if (value[0] instanceof Date) {
-        [nextReference] = value;
+        // if we just selected an end date, active date will be 'end'
```
taysea: Should this read "active date will be 'end'"?

src/js/components/Calendar/Calendar.js
```diff
@@ -132,14 +132,27 @@ const normalizeRange = (value, activeDate) => {
...
+        // and we should set the reference to the end date
+        if (activeDate === 'start' && value[1] instanceof Date) {
+          [, nextReference] = value;
+        } else {
+          [nextReference] = value;
+        }
       } else if (Array.isArray(value[0])) {
-        nextReference = value[0][0] ? value[0][0] : value[0][1];
+        // if we just selected an end date, active date will be 'end'
```
taysea: Should this read "active date will be 'end'"?

jcfilben: Oops it should be 'start', pushing up that change now

jcfilben: fixed

src/js/components/Calendar/Calendar.js
jcfilben: fixed

britt6612:
not sure if you discussed this already but on the first Calendar when I go to May then hit `tab` it focuses on the last day of the month. Should this be the first day of the month?


https://github.com/user-attachments/assets/79b6cdf3-2611-4876-a466-dc7d96885d30

jcfilben:
> not sure if you discussed this already but on the first Calendar when I go to May then hit tab it focuses on the last day of the month. Should this be the first day of the month?

Is this just happening on the Dual story? That story needs fixed it has been broken for a while now

britt6612:
> > not sure if you discussed this already but on the first Calendar when I go to May then hit tab it focuses on the last day of the month. Should this be the first day of the month?
> 
> Is this just happening on the Dual story? That story needs fixed it has been broken for a while now

yes I verified that is the only story so agree that is separate issue

taysea (approved):
Looks good! Thank you

MikeKingdom:
The Dual calendar example in storybook seems a bit broken. I'm not sure if it's the story more than the Calendar itself, but it behaves differently in top-of-tree compared to this PR. Navigating with the prev/next arrow buttons gets you in a weird pair of months (not consecutive months) and you can't get back to a normal state. Moving up and down in a month with the arrow keys can cause it to lose the focus and stop moving between days or months (with no visible focus). Selecting a date with the keyboard cause the other calendar to flip to that same date.

With code on master you can at least keep moving up or down with the arrow keys and traverse months

MikeKingdom:
If you arrow down to a date in the bottom left corner, you have to hit the down arrow key twice to navigate off the date into the next month. You can see this in Input->DateInput->European

taysea:
> The Dual calendar example in storybook seems a bit broken. I'm not sure if it's the story more than the Calendar itself, but it behaves differently in top-of-tree compared to this PR. Navigating with the prev/next arrow buttons gets you in a weird pair of months (not consecutive months) and you can't get back to a normal state. Moving up and down in a month with the arrow keys can cause it to lose the focus and stop moving between days or months (with no visible focus). Selecting a date with the keyboard cause the other calendar to flip to that same date.
> 
> With code on master you can at least keep moving up or down with the arrow keys and traverse months

I think in some previous comment threads Jessica felt like the existing behavior of that story wasn't ideal either and had mentioned not to be concerned on the behavior of that story. But it does feel worth looking into if we can fix the story with the new logic / if we should be considered about breaking changes for any teams that might have implemented this kind of design. @britt6612 Let's check this out a bit.

EDIT: After looking, I think the main part we might consider looking into on this PR is "Moving up and down in a month with the arrow keys can cause it to lose the focus and stop moving between days or months (with no visible focus)." The rest is existing issues/rooted in the story setup.

src/js/components/Calendar/Calendar.js
```diff
@@ -529,6 +549,15 @@ const Calendar = forwardRef(
 
     const changeCalendarMonth = (messageId, newMonth) => {
       handleReference(newMonth);
+      const firstOfMonth = new Date(
+        newMonth.getFullYear(),
+        newMonth.getMonth(),
+        1,
+        startOfMonth(reference).getHours(),
+        startOfMonth(reference).getMinutes(),
+        startOfMonth(reference).getSeconds(),
```
MikeKingdom: Would be better to pull this out into a var and use it rather than re-calc the start of the month 3 times

taysea: Done.

jcfilben:
> After looking, I think the main part we might consider looking into on this PR is "Moving up and down in a month with the arrow keys can cause it to lose the focus and stop moving between days or months (with no visible focus)." The rest is existing issues/rooted in the story setup.

Just pushed up a commit that I think should fix this

MikeKingdom (approved):

PR #7631, opened 2025-05-06
update dependencies

britt6612:
#### What does this PR do?
update dependencies post grommet v2.47.0 release
package.json
locally
locally 


src/js/components/Accordion/__tests__/__snapshots__/Accordion-test.tsx.snap
jcfilben, line 1: Do we know where this change is coming from?

britt6612: yes this change is coming from here
https://github.com/facebook/react/pull/32001

this was noted in the release notes

jcfilben (commented):
Could we break out dependencies that cause snapshot changes into a separate PR so it's easier to link the snapshot just with that change? All other dependency upgrades can be grouped together

britt6612:
> Could we break out dependencies that cause snapshot changes into a separate PR so it's easier to link the snapshot just with that change? All other dependency upgrades can be grouped together

sure can do

jcfilben (approved):
Looks good!

PR #7641, opened 2025-05-15
fix file input 1.1.1

### LINKED ISSUE #7561: File Input - WCAG Rule 1.1.1 Non-text content
The error icon is being read by screen readers, but it doesn’t convey meaningful information to users.

Sulaymon333:
#### What does this PR do?
- Fix accessibility issue 1.1.1 for file input component
- closes #7561

### DIFF
--- a/src/js/components/FileInput/FileInput.js
+++ b/src/js/components/FileInput/FileInput.js
@@ -472,7 +472,7 @@ const FileInput = forwardRef(
                     direction="row"
                   >
                     {((maxSize && file.size > maxSize) ||
-                      (max && index >= max)) && <CircleAlert />}
+                      (max && index >= max)) && <CircleAlert a11yTitle="" />}
                     <Label
                       weight={
                         theme.global.input.weight ||
@@ -488,10 +488,25 @@ const FileInput = forwardRef(
                 <Box flex={false} direction="row" align="center">
                   <Button
                     ref={index ? undefined : removeRef}
-                    a11yTitle={`${format({
-                      id: 'fileInput.remove',
-                      messages,
-                    })} ${file.name}`}
+                    // a11yTitle={`${format({
+                    //   id: 'fileInput.remove',
+                    //   messages,
+                    // })} ${file.name}`}
+                    a11yTitle={
+                      (maxSize && file.size > maxSize) || (max && index >= max)
+                        ? format({
+                            id: 'fileInput.removeMaxFileSize',
+                            messages,
+                            values: {
+                              fileName: file.name,
+                              maxSize: formatBytes(maxSize),
+                            },
+                          })
+                        : `${format({
+                            id: 'fileInput.remove',
+                            messages,
+                          })} ${file.name}`
+                    }
                     icon={<RemoveIcon />}
                     hoverIndicator
                     disabled={disabled}
--- a/src/js/languages/default.json
+++ b/src/js/languages/default.json
@@ -80,7 +80,8 @@
       "plural": "{numOfFiles} files are too large. Select files which are no larger than {maxSize}."
     },
     "remove": "remove",
-    "removeAll": "remove all"
+    "removeAll": "remove all",
+    "removeMaxFileSize": "Error - remove {fileName} (exceeds max file size) {maxSize}."
   },
   "form": {
     "invalid": "invalid",


crashuniverse:
@Sulaymon333 To make this pull request complete, we should make sure the commit is signed off - that will make the first check "DCO" go through. It is simple. On the very branch `#7561-fix-file-input-1.1.1`, just do: `git commit --amend -s`. The `-s` would sign the commit and this check should pass then.

src/js/components/FileInput/FileInput.js
```diff
@@ -455,101 +455,125 @@ const FileInput = forwardRef(
...
+                    fileName: file.name,
+                    maxNumber: max,
+                  },
+                });
+              } else {
+                removeIconA11yTitle = `${format({
+                  id: 'fileInput.remove',
+                  messages,
+                })} file: ${file.name}`;
+              }
```
MikeKingdom, line 483: We can simplify this. Will require adding a file name param to the fileInput.remove message. I know it was pre-existing, but when internationalizing, you should never concatenate parts of a message together (instead use substitution vars.)
```suggestion
              let messageId = 'fileInput.remove';
              if (maxSize && file.size > maxSize) {
                messageId = 'fileInput.removeMaxFileSize';
              }
              else if (max && index >= max) {
                messageId = 'fileInput.removeMaxFileCount';
              }
```
Then change the point where it puts in in prop to:
```jsx
                format({
                  id: messageId,
                  messages,
                  values: {
                    fileName: file.name,
                    maxSize: formatBytes(maxSize),
                    maxNumber: max,
                  },
                });
```

src/js/components/FileInput/__tests__/__snapshots__/FileInput-test.tsx.snap
```diff
@@ -1564,7 +1564,7 @@ exports[`FileInput disabled with file selected 1`] = `
               class="c10"
             >
               <button
-                aria-label="remove file-size-okay.csv"
+                aria-label="remove file: file-size-okay.csv"
```
MikeKingdom, line 1567: Should have a test that hits those other cases if possible

src/js/languages/default.json
```diff
@@ -80,7 +83,9 @@
       "plural": "{numOfFiles} files are too large. Select files which are no larger than {maxSize}."
     },
     "remove": "remove",
```
MikeKingdom, line 85:
```suggestion
    "remove": "remove file: {fileName}",
```

src/js/languages/default.json
```diff
@@ -104,6 +109,9 @@
       "validation": "Copied"
     }
   },
+  "list": {
+    "pinned": "Item pinned, order cannot be changed."
+  },
```
MikeKingdom, line 114: Shouldn't be in this PR?

src/js/languages/default.json
```diff
@@ -126,6 +134,9 @@
       "plural": "Semicircle meter with values {meterValue} out of {max}"
     }
   },
+  "notification": {
+    "close": "Dismiss notification"
+  },
```
MikeKingdom, line 139: Shouldn't be in this PR?

src/js/languages/default.json
```diff
@@ -80,7 +83,9 @@
       "plural": "{numOfFiles} files are too large. Select files which are no larger than {maxSize}."
     },
     "remove": "remove",
-    "removeAll": "remove all"
+    "removeAll": "remove all",
+    "removeMaxFileSize": "Error - remove file: {fileName} (exceeds max file size) of {maxSize}.",
+    "removeMaxFileCount": "Error - remove file: {fileName}, (maximum number of files) - {maxNumber} exceeded."
```
MikeKingdom, line 88: It seems odd to put this message in the remove button. I think the remove button should just say "Remove file: {fileName}". Shouldn't the validation error text be elsewhere, like the alert icon (ie. tells you what the alert icon means")

```suggestion
    "removeMaxFileSize": "Remove file: {fileName}. This file exceeds max file size of {maxSize}",
    "removeMaxFileCount": "Remove file: {fileName}. This file exceeds the maximum number of files) - {maxNumber}"
```

Sulaymon333: The word `Error` was used for comparison with the [draft PR ](https://github.com/grommet/grommet/pull/7674/files#diff-ecd05461ba0fd4b2a00e66e9708f8d8a59bf3dde54716424f8aadaa0a4750430)that Bill will be testing against, the draft PR has a better message that is concise and direct. Please see below and let me know what you think

```
"removeMaxFileSize": "Remove file: {fileName}. This file is too large. Maximum allowed size is {maxSize}.",
"removeMaxFileCount": "Remove file: {fileName}. Too many files uploaded. Maximum allowed is {maxNumber}."
```

src/js/languages/default.json
Sulaymon333: noted

src/js/languages/default.json
Sulaymon333: noted

src/js/languages/default.json
Sulaymon333: I will update. I was thinking it was used elsewhere. But it is better we append with file right insde the translation file

src/js/components/FileInput/__tests__/__snapshots__/FileInput-test.tsx.snap
Sulaymon333: ok noted

Sulaymon333:
On another note, there is no way for visual users as well to tell if a file is too large or too many files uploaded. We can have for example a small text below each file name to state if file is too large or too many files are uploaded already. I have seen similar implementation from other libraries. Does this sound like what we would like to add as well.

image

src/js/languages/default.json
MikeKingdom: Better thanks. I still think it's not right that it's on the remove button and should be instead on the alert icon (ie. why I'm getting this icon). Testing with Bill will help

MikeKingdom:
I think I forgot to mention this, but the typescript types and prop-types for the message parameter for this component as well as the Grommet global messages prop need to be updated anytime you add a key

MikeKingdom:
> On another note, there is no way for visual users as well to tell if a file is too large or too many files uploaded. We can have for example a small text below each file name to state if file is too large or too many files are uploaded already. I have seen similar implementation from other libraries. Does this sound like what we would like to add as well.
We originally chose not to do that because of the clutter and it didn't seem necessary. I don't think we need to revisit that UX discussion

Sulaymon333:
I understand the decision around this. However, [from my research](https://github.com/grommet/grommet/pull/7597#issuecomment-2900075466), a number of frameworks seems to provide some sort of feedback to users in these two cases mentioned. 

What support do we have for implemeneter that has a use case for this?

I know `renderFile prop` exposes the native file object that implementer can use to render custom name and  check file sizethen  render whatever is desired when file size is too large. What about if the total number of file count exceeds what is set? Is there a way for implementer to access `max file count prop` somehow?

sulaymon-tajudeen-hpe:
@MikeKingdom I address the feedback you gave on separating the error message for files that are too large or for when too many files are attached here in this [draft PR](https://github.com/grommet/grommet/pull/7679). I separted it on a different draft PR for accessibility testing with JAWS by Bill. I should get Bill's review on this new version next week when he is back.

I have requested your revirew on it as well. 

P.S. still need to add more testing for the new draft PR (v3) at some point when direction is clear and fix circle alert icon focus indicator.

MikeKingdom:
> I understand the decision around this. However, [from my research](https://github.com/grommet/grommet/pull/7597#issuecomment-2900075466), a number of frameworks seems to provide some sort of feedback to users in these two cases mentioned.
> 
> What support do we have for implemeneter that has a use case for this?
> 
> I know `renderFile prop` exposes the native file object that implementer can use to render custom name and check file sizethen render whatever is desired when file size is too large. What about if the total number of file count exceeds what is set? Is there a way for implementer to access `max file count prop` somehow?

Short answer really is that is outside the scope of this effort and not something We'll spend time on here. And food for thought...just because a particular file is beyond the max file count in the list of files doesn't mean that is a file that should be removed from the list. Really the problem is they added more files than they are allowed to for whatever reason and let them decide which files to keep or drop. Don't overly bias that decision arbitrarily.

britt6612:
@Sulaymon333 can we close this PR?

Sulaymon333:
Closed in favour of https://github.com/grommet/grommet/pull/7679

PR #7642, opened 2025-05-16
Fixes #7418 Add onClickOutside handler for Tip component

### LINKED ISSUE #7418: Button Tip Will Not Go Away
I am seeing instances when the button tip does not go away. It happens when your mouse leaves the button as soon as the tip changes color. Take the two examples below, one is from HPE design system and the other is from my [codesandbox](https://codesandbox.io/p/sandbox/grommet-v2-template-forked-p7rfzt?workspaceId=d6eb95fa-558f-459c-a396-97c7bbb32b9a) minimum replicable example. Note the HPE design system video is trimmed, as I am only able to achieve a hanging tip about 3/100 times. However, because the codesandbox is a much slower application the hanging tip effect is much more probable to occur.

image002

https://github.com/user-attachments/assets/9ebc4576-178d-49a8-80b8-5ebc2da66dac

https://github.com/user-attachments/assets/b896f108-fe06-41df-b1bd-1f663b986aaa

Happy to if it's a good first issue. 

### Expected Behavior

The tip should disappear as soon as you stop hovering over the button.

### Actual Behavior

The tip will not disappear if its state changes, i.e background color, at the same moment your mouse moves away from the button. 

[Codesandbox](https://codesandbox.io/p/sandbox/grommet-v2-template-forked-p7rfzt?workspaceId=d6eb95fa-558f-459c-a396-97c7bbb32b9a)
It is easier to replicate this behavior when the theme has to change the color for a lot of components because it creates a lag.

### Steps to Reproduce

1. Click the change theme button
2. At the moment the color changes from light to dark or vis versa move your mouse away from the theme button. This may take a few tries to get the timing down.
3. Observe the tip does not disappear until you hover over the button again 

### Your Environment

- Grommet version: 2.37 - 2.41
- Browser Name and version: Chrome 130

crashuniverse:
Making clone of children in slow processing environment leads to a situation where mouseEnter event happens for a previous element and so for a new cloned element that is trigger for Tip component, mouseLeave never triggers leaving Tip component hanging. Although the situation may not be improved however a missing prop of onClickOutside is not used yet, so by adding this prop we eliminate this problem if user clicks anywhere on document.

#### What does this PR do?
This PR adds a onClickOutside prop to Tip component so that in cases Tip is persistent on screen, at least clicking outside the element closes the Tip.

The reviewer should catch up with https://github.com/grommet/grommet/issues/7418#issuecomment-2883744959 first so that emulating the problem becomes very straightforward. Once we see the problem, we should add a console.log to onMouseEnter and onMouseLeave event handlers so that we can see which event handler gets called. This should let us see that onMouseLeave is never called in slower CPU environments because of how we clone elements (which we should not fiddle with).

I've ran `yarn test` and on storybook, I've seen all 9 variations of Tip component to keep all the features of Tip alive.

With this branch, we can modify file `e2e/src/App.js` so that it is very similar to codesandbox example.

<details><summary>Click here to see entire code</summary>

```
import React, { useState } from "react";
import { Box, Button, Header, Text } from "../../src/js/components";
import { Sun, Moon } from "grommet-icons";

const App = () => {
  const [mode, setMode] = useState("light");

  const body = Array.from(new Array(2000), (element, i) => (
    <Text key={i}>
      {mode}-{i}
    </Text>
  ));

  return (
    <div>    
      <Header justify="end">
        <Button
          icon={mode === "light" ? <Sun /> : <Moon />}
          tip={
            mode === "light" ? "Switch to Dark Mode" : "Switch to Light Mode"
          }
          onClick={() => {
            mode === "light" ? setMode("dark") : setMode("light");
          }}
        />
      </Header>
      <Box pad="large">{body}</Box>
    </div>
  );
};

export default App;

```

</details>

- [] `screen` is used for querying.

No

https://github.com/grommet/grommet/issues/7418 

None

No

No

Yes, is backward compatible


jcfilben (commented):
I'm wondering if we should explore memoizing the cloned element

PR #7656, opened 2025-05-22
Skeleton wcag

### LINKED ISSUE #7600: Skeleton - WCAG rule 1.1.1, 1.3.1, & 1.3.2
Skeletons are purely visual and contain no relevant meaning for screen readers, so should be given aria-hidden=\"true\. [Supporting documentation](https://adrianroselli.com/2020/11/more-accessible-skeletons.html)

Because Skeletons are appearing on screen readers, it is breaking three WCAG rules:
- 1.1.1 Non-text content
- 1.3.1 Info and Relationships
     - Because skeletons are not hidden, they may mess with the structure of the page for screen readers.
- 1.3.2 Meaningful Sequence
     - Hide skeletons from screen readers to preserve logical flow.

britt6612:
#### What does this PR do?
add aria-hidden to skeleton
skeleton.js
locally 
updated snapshots

Skeletons are purely visual and contain no relevant meaning for screen readers, so should be given aria-hidden="true. [Supporting documentation](https://adrianroselli.com/2020/11/more-accessible-skeletons.html)

Because Skeletons are appearing on screen readers, it is breaking three WCAG rules:

1.1.1 Non-text content
1.3.1 Info and Relationships
Because skeletons are not hidden, they may mess with the structure of the page for screen readers.
1.3.2 Meaningful Sequence
Hide skeletons from screen readers to preserve logical flow.
closes #7600 
no but DS doc need to be updated
yes
backwards compatible


jcfilben (dismissed):
Look good!

jcfilben (commented):
It looks like we have announcements for skeleton (for example see the skeleton code in Box.js) I'm wondering if we should remove these

src/js/components/Box/Box.js
```diff
@@ -93,14 +90,6 @@ const Box = forwardRef(
       }
     }, [refProp, as, responsiveProp]);
 
-    useEffect(() => {
-      if (skeletonProp?.message?.start) announce(skeletonProp.message.start);
-      else if (typeof skeletonProp?.message === 'string')
```
jcfilben, line 98: I realized we have a prop on Box for passing messages to skeleton, we'll need to figure out how we want to handle the backwards compatibility of that

jcfilben: I'm thinking if someone passes skeleton.message we should go ahead and leave the announcement to maintain backwards compatibility. Also I was thinking about this more and even though the skeleton itself should be hidden from the screenreader user, an announcement indicating something has loaded could be useful in certain situations. The implementor should use it sparingly though so they don't overload the screen reader user with unnecessary announcements

britt6612: done

jcfilben (approved):
Looks good!

taysea (approved):

PR #7666, opened 2025-05-30
change aria labels for tabs

### LINKED ISSUE #7647: DataTableColumns - WCAG 1.3.2 1.3.3

None

### Expected Behavior

The title "order of columns" is not clear for screen readers as well as "selected columns"

- [ ] After fixing this issue go back to DS and update the rules 1.3.2 1.3.3

### Actual Behavior

not passing

_No response_

### Steps to Reproduce

_No response_

### Your Environment

_No response_

britt6612:
#### What does this PR do?
This PR updates the `aria-label` to give a better description of what these items do for screen readers
datatablecolumns.js
storybook with voice over
voice over with storybook 

Went through DataTableColumns component with Bill and he mentioned 

``` 
some feedback it would be nice if it actually told you what checkboxes did as well as the list. Ordering columns did not make it clear to Bill as what he was actually suppose to do he mentioned to say something like "here where you can select what data elements you want to show in the datatable"

closes #7647 

yes
yes
backwards compatible

### DIFF
--- a/src/js/components/DataTableColumns/DataTableColumns.js
+++ b/src/js/components/DataTableColumns/DataTableColumns.js
@@ -112,6 +112,10 @@ const Content = ({ drop, options = [], ...rest }) => {
             id: 'dataTableColumns.select',
             messages: messages?.dataTableColumns,
           })}
+          aria-label={format({
+            id: 'dataTableColumns.selectAria',
+            messages: messages?.dataTableColumns,
+          })}
         >
           <Box pad={{ vertical: 'small' }} gap="xsmall">
             <TextInput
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Review by jcfilben — COMMENTED** [review]
Can we add the messages to Data proptypes and index.d.ts?

**Comment by britt6612** [comment]
> Can we add the messages to Data proptypes and index.d.ts?

done

File: src/js/components/Data/index.d.ts [hunk 1]
```diff
@@ -104,6 +104,8 @@ export interface DataProps {
       order?: string;
       select?: string;
       tip?: string;
+      orderAria?: string;
```
taysea, line 108: For SelectMultiple, we use the term a11y ("clearAllA11y") instead of aria for the aria specific label -- do we want to standardize on one direction moving forward?

aka if we prefer "aria", I'm good with this.

taysea: I'm leaning towards standardizing on aria I think.

britt6612: ah I didnt realize we do that for SelectMultiple but I would agree standardizing on aria

jcfilben: I feel like if we already use `a11y` we could standardize to that

PR #7673, opened 2025-06-05
Update SelectMultiple to use roving tab index

### LINKED ISSUE #7671: SelectMultiple - focus order 2.4.3

None

### Expected Behavior

When I tab into the list of options in the drop the first option should receive focus

### Actual Behavior

The first option does not receive focus until down arrow is pressed

https://storybook.grommet.io/?path=/story/input-selectmultiple-default--default

### Steps to Reproduce

1. Open the Select multiple drop down
2. Tab to the options
3. First option doesn't receive focus

### Your Environment

_No response_

taysea:
#### What does this PR do?

This PR resolves focus visibility bug on SelectMultiple by:
- Removing nested interactive elements (previously we had tabIndex="0" on the listbox and on any selected option inside the listbox).
- Switching to a roving tab index approach to ensure there's still only one-ish tab stop (see sub-bullets here) for the listbox and arrow keys are used to navigate among options)
   - listbox has tabIndex="-1" (not in tab order)
   - If nothing is selected, the first option has tabIndex="0"
   - Using arrow keys updates the "active" option to tabIndex="0" and the rest to tabIndex="-1"
   - NOTE: If an option is selected, it also remains part of the tab order. This is existing behavior that I didn't change, but it does mean that it's possible for the listbox to have more than one tab stop (which goes against WCAG APG pattern recommendation. I can update if desired)
- Removes aria-activedescendant from listbox because this is not necessary when listbox doesn't have tabIndex="0"
- Removes incorrect keyboard behavior that was allowing arrow keys to be used to move from "Select all/Clear all" button to options / search input to options

NOTE: VoiceOver reads out options as "{label} not selected, selected" (This is existing behavior and also happens on WCAG APG pattern, so it seems like it's a VoiceOver bug or decision outside of our control. Just flagging that I saw and assessed if this was something we should try to overcome, but it seems not.)

src/js/components/SelectMultiple/SelectMultipleContainer.js

Locally across the various storybook examples.

Storybook.

Closes #7671 

No.

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/SelectMultiple/SelectMultipleContainer.js
+++ b/src/js/components/SelectMultiple/SelectMultipleContainer.js
@@ -591,7 +591,7 @@ const SelectMultipleContainer = forwardRef(
                       tabIndex={
                         optionSelected ||
                         activeIndex === index ||
-                        // when notthing is selected and entering listbox
+                        // when nothing is selected and entering listbox
                         // first option should be focused
                         (value.length === 0 &&
                           activeIndex === -1 &&
--- a/src/js/components/SelectMultiple/SelectMultipleContainer.js
+++ b/src/js/components/SelectMultiple/SelectMultipleContainer.js
@@ -69,12 +69,18 @@ const SelectMultipleContainer = forwardRef(
     ref,
   ) => {
     const { theme } = useThemeValue();
+    // the currently active option based on keyboard navigation
+    // or mouse hover, -1 means no active option
     const [activeIndex, setActiveIndex] = useState(-1);
     const [keyboardNavigation, setKeyboardNavigation] = useState(usingKeyboard);
     const { format } = useContext(MessageContext);
     const searchRef = useRef();
     const optionsRef = useRef();
     const [disabled, setDisabled] = useState(disabledProp);
+    // the node of the currently active option, as activeIndex changes
+    // this is updated too and the useEffect below ensures the
+    // active option remains in keyboard focus since we're
+    // following roving tab index pattern
     const activeRef = useRef();
     const [showA11yLimit, setShowA11yLimit] = useState();
     const clearRef = useRef();
@@ -187,7 +193,10 @@ const SelectMultipleContainer = forwardRef(
       (event) => {
         event.preventDefault();
         const nextActiveIndex = activeIndex + 1;
-        if (nextActiveIndex !== options?.length) {
+        // checking activeIndex > -1 ensures arrow keys don't
+        // move focus when select all/clear all button or search input
+        // are focused
+        if (nextActiveIndex !== options?.length && activeIndex > -1) {
           setActiveIndex(nextActiveIndex);
           setKeyboardNavigation(true);
         }
@@ -200,16 +209,11 @@ const SelectMultipleContainer = forwardRef(
         event.preventDefault();
         const nextActiveIndex = activeIndex - 1;
 
-        if (nextActiveIndex === -1) {
-          const searchInput = searchRef.current;
-          if (searchInput && searchInput.focus) {
-            setActiveIndex(nextActiveIndex);
-            setFocusWithoutScroll(searchInput);
-          }
-        }
-
-        if (nextActiveIndex >= 0) {
-          setActiveIndex(nextActiveIndex);
+        // checking activeIndex > -1 ensures arrow keys don't
+        // move focus when select all/clear all button or search input
+        // are focused
+        if (activeIndex > -1) {
+          setActiveIndex(Math.max(nextActiveIndex, 0));
           setKeyboardNavigation(true);
         }
       },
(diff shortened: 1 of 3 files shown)


src/js/components/SelectMultiple/__tests__/__snapshots__/SelectMultiple-test.tsx.snap
```diff
@@ -1045,7 +1037,7 @@ exports[`SelectMultiple keyboard interactions 1`] = `
     <button
       aria-expanded="true"
       aria-haspopup="listbox"
-      aria-label="Open drop. 2 selected"
+      aria-label="Open drop. 0 selected"
```
jcfilben: Looks like we may need to update the keyboard interactions test. It looks like it is trying to navigate into the list items by pressing down arrow from the search input, which we are no longer supporting

taysea: Oops, thank you I will fix.

src/js/components/SelectMultiple/__tests__/__snapshots__/SelectMultiple-test.tsx.snap
```diff
@@ -1037,6 +1037,7 @@ exports[`SelectMultiple keyboard interactions 1`] = `
 }
 
 <div
+  aria-hidden="true"
```
taysea, line 1040: @jcfilben FYI I think these additions make sense because the "Drop" is open (so other things on the screen should be hidden from aria.

I'm not sure why these weren't previously part of the snapshot but could've been due to lack of jest.useFakeTimers() to allow the drop to fully open? Idk.

src/js/components/SelectMultiple/__tests__/__snapshots__/SelectMultiple-test.tsx.snap
```diff
@@ -1037,9 +1046,10 @@ exports[`SelectMultiple keyboard interactions 1`] = `
     <button
       aria-expanded="true"
       aria-haspopup="listbox"
-      aria-label="Open drop. 0 selected"
+      aria-label="Open drop. 2 selected"
       class="c2 c3"
-      tabindex="0"
+      data-g-tabindex="0"
+      tabindex="-1"
```
jcfilben, line 1052: I'm wondering why the tabindex and data-g-tabindex don't match

src/js/components/SelectMultiple/__tests__/__snapshots__/SelectMultiple-test.tsx.snap
jcfilben: Could we try changing the test and checking locally so that the drop is closed before we take the snapshot just to confirm that at that point the aria-hidden gets removed?

taysea: I tested locally and that's the case (aria-hidden removed)

src/js/components/SelectMultiple/__tests__/__snapshots__/SelectMultiple-test.tsx.snap
taysea: I believe the way those attributes are set up is that data-g-tabindex stores what the tabindex of the element was before the drop was open, so we can properly restore it once the drop closes.

So this is saying "this element previously had tabIndex 0 (data-g-tabindex)"

src/js/components/SelectMultiple/__tests__/__snapshots__/SelectMultiple-test.tsx.snap
taysea: Done.

src/js/components/SelectMultiple/__tests__/__snapshots__/SelectMultiple-test.tsx.snap
jcfilben: Cool, thanks for the explaination

jcfilben (approved):
Looks good!

src/js/components/SelectMultiple/SelectMultipleContainer.js
```diff
@@ -584,7 +588,17 @@ const SelectMultipleContainer = forwardRef(
                         if (optionRef) optionRef.current = node;
                         if (optionActive) activeRef.current = node;
                       }}
-                      tabIndex={optionSelected ? '0' : '-1'}
+                      tabIndex={
+                        optionSelected ||
+                        activeIndex === index ||
+                        // when notthing is selected and entering listbox
```
taysea:
```suggestion
                        // when nothing is selected and entering listbox
```

britt6612 (approved):

PR #7675, opened 2025-06-05
add fix for select focus

### LINKED ISSUE #7670: Select - focus order 2.4.3

None

### Expected Behavior

Pressing tab after opening the drop should either do nothing or more to another item in the drop (ex. the clear button)

### Actual Behavior

After opening the drop if I press 'tab' I can't get back into the drop

https://storybook.grommet.io/?path=/story/input-select-simple--simple

### Steps to Reproduce

1. Open the drop
2. Press tab a few times and try to get back into the drop

### Your Environment

_No response_

britt6612:
#### What does this PR do?
This PR changes the tab index from `-1` to `0` when keyboard is true and the option is the currently active option. 
The Focus was leaving the drop because once it opens we were forcing the focus to be on the first option in the select however once someone hit `tab` there is nothing actually focusable in the select so that is why the focus would move we need to make an option have tab index in order to keep the focus inside and trap it. 
selectContainer.js
locally 
focus on select open it with enter then tab focus again it should not move out of the drop

see above
closes #7670 
no
yes
 backwards compatible

### DIFF
--- a/src/js/components/Select/SelectContainer.js
+++ b/src/js/components/Select/SelectContainer.js
@@ -460,7 +460,7 @@ const SelectContainer = forwardRef(
                         if (optionRef) optionRef.current = node;
                         if (optionActive) activeRef.current = node;
                       }}
-                      tabIndex={optionSelected ? '0' : '-1'}
+                      tabIndex={usingKeyboard && activeIndex === index ? 0 : -1}
                       role="option"
                       aria-setsize={options.length}
                       aria-posinset={index + 1}
--- a/src/js/components/Select/__tests__/__snapshots__/Select-multiple-test.js.snap
+++ b/src/js/components/Select/__tests__/__snapshots__/Select-multiple-test.js.snap
@@ -973,7 +973,7 @@ exports[`Select Controlled multiple onChange toggle with valueKey reduce 2`] = `
         aria-setsize="2"
         class="c5 c6"
         role="option"
-        tabindex="0"
+        tabindex="-1"
         type="button"
       >
         <div
@@ -2461,7 +2461,7 @@ exports[`Select Controlled multiple onChange without labelKey 4`] = `
         aria-setsize="2"
         class="StyledButton-sc-323bzc-0 brhmKp StyledSelect__SelectOption-sc-znp66n-3 jdCSKT"
         role="option"
-        tabindex="0"
+        tabindex="-1"
         type="button"
       >
         <div
@@ -2534,7 +2534,7 @@ exports[`Select Controlled multiple onChange without labelKey 6`] = `
         aria-setsize="2"
         class="StyledButton-sc-323bzc-0 brhmKp StyledSelect__SelectOption-sc-znp66n-3 jdCSKT"
         role="option"
-        tabindex="0"
+        tabindex="-1"
         type="button"
       >
         <div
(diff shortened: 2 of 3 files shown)


jcfilben (commented):
It looks like we lost the behavior where you could press tab to jump to a selected date if you have multiple selected dates

jcfilben (commented):
looks good!

jcfilben (approved):

britt6612:
> It looks like we lost the behavior where you could press tab to jump to a selected option if you have multiple selected options

fixed

taysea:
In the story with onSearch, I can't tab from search to options (but I would expect to be able to). Is that something this PR should handle or take care of it in a separate PR?

britt6612:
> In the story with onSearch, I can't tab from search to options (but I would expect to be able to). Is that something this PR should handle or take care of it in a separate PR?

yes thats the same behavior now I wonder then would you expect after the tab on the search to be the first one? I think thats a little different from the issue here but I can add in this PR

britt6612:
> In the story with onSearch, I can't tab from search to options (but I would expect to be able to). Is that something this PR should handle or take care of it in a separate PR?

@taysea added this should work with onSearch now

src/js/components/Select/SelectContainer.js
```diff
@@ -460,7 +460,13 @@ const SelectContainer = forwardRef(
                         if (optionRef) optionRef.current = node;
                         if (optionActive) activeRef.current = node;
                       }}
-                      tabIndex={optionSelected ? '0' : '-1'}
+                      tabIndex={
```
taysea, line 490: If I:
- open select
- tab to options
- use arrow keys to go down some options
- use shift tab

Then I would expect the shift tab to go back to search, but it goes back to the first option (even though it's not selected?)

https://deploy-preview-7675--sad-tereshkova-173d07.netlify.app/?path=/story/input-select-search--search&globals=theme:hpe

https://github.com/user-attachments/assets/080eeb0f-f2ea-4760-9968-aa6e6f98013d

britt6612: ah let me look into this good catch

jcfilben (approved):
Looks good!

taysea:
Right now the keyboard behavior for Select vs SelectMultiple is different for the cases where options had keyboard focus, then focus was moved back to the search input (we should align on one direction).

In the case of SelectMultiple, after the "Search" input receives focus, then if focus is given back to the options when nothing is selected, **the first item always receives focus again.**

In the case of this PR for Select, after the "Search" input receives focus, then if focus is given back to the options when nothing is selected, **the previously active item is given focus.**

We should align on a single desired behavior, then update whichever component to match.

britt6612:
> Right now the keyboard behavior for Select vs SelectMultiple is different for the cases where options had keyboard focus, then focus was moved back to the search input (we should align on one direction).
> 
> In the case of SelectMultiple, after the "Search" input receives focus, then if focus is given back to the options when nothing is selected, **the first item always receives focus again.**
> 
> In the case of this PR for Select, after the "Search" input receives focus, then if focus is given back to the options when nothing is selected, **the previously active item is given focus.**
> 
> We should align on a single desired behavior, then update whichever component to match.

aligned behavior to align with SelectMultiple @taysea

halocline (approved):
Behavior looks good to me.

taysea (approved):
LGTM!

PR #7676, opened 2025-06-06
RangeSelector - add theme value for the edge size

### LINKED ISSUE #7669: RangeSelector - WCAG 2.5.8 target size

Yes

### Expected Behavior

RangeSelector fails 2.5.8

The circular buttons on either side of the slider do not meet the minimum interactive target size of 24×24. 

once PR is merged need to update DS site.

- [ ] Update DataFilters, DataFilter rule 2.5.8

### Actual Behavior

The circular buttons are 12×12.

rangeselector component 

### Steps to Reproduce

_No response_

### Your Environment

_No response_

britt6612:
#### What does this PR do?
This PR adds a theme value for the theme to allow users to pass a value for `edge` size
EdgeControl.js
locally and wrote a test 
pass in a value for `theme.rangeSelector?.edge?.size;`

The size needs to be at least 24X24 
right now we have it at 12X12 so we dont break the current size I have added a theme option for us to use in the hpe theme 
closes #7669 
yes
yes
 backwards compatible

### DIFF
--- a/src/js/components/RangeSelector/EdgeControl.js
+++ b/src/js/components/RangeSelector/EdgeControl.js
@@ -29,7 +29,11 @@ const EdgeControl = forwardRef(
     const { theme } = useThemeValue();
     const [focus, setFocus] = useState(false);
     const { cursor, fill } = DIRECTION_PROPS[direction];
-    const size = parseMetricToNum(theme.global.spacing) / 2;
+    const themeSize = theme.rangeSelector?.edge?.size;
+    const size =
+      typeof themeSize === 'string'
+        ? parseMetricToNum(theme.global.edgeSize?.[themeSize] || themeSize)
+        : parseMetricToNum(theme.global.spacing) / 2;
     const keyboardProps =
       direction === 'vertical'
         ? { onUp: onDecrease, onDown: onIncrease }
--- a/src/js/components/RangeSelector/EdgeControl.js
+++ b/src/js/components/RangeSelector/EdgeControl.js
@@ -30,10 +30,11 @@ const EdgeControl = forwardRef(
     const [focus, setFocus] = useState(false);
     const { cursor, fill } = DIRECTION_PROPS[direction];
     const themeEdgeSize = theme.rangeSelector?.edge?.size;
-    const size =
-      typeof themeSize === 'string'
-        ? parseMetricToNum(theme.global.edgeSize?.[themeSize] || themeSize)
-        : parseMetricToNum(theme.global.spacing) / 2;
+    const size = themeEdgeSize
+      ? parseMetricToNum(
+          theme.global.edgeSize?.[themeEdgeSize] || themeEdgeSize,
+        )
+      : parseMetricToNum(theme.global.spacing) / 2;
     const keyboardProps =
       direction === 'vertical'
         ? { onUp: onDecrease, onDown: onIncrease }
--- a/src/js/components/RangeSelector/EdgeControl.js
+++ b/src/js/components/RangeSelector/EdgeControl.js
@@ -31,7 +31,7 @@ const EdgeControl = forwardRef(
     const { cursor, fill } = DIRECTION_PROPS[direction];
     const themeEdgeSize = theme.rangeSelector?.edge?.size;
     let size;
-    if (themeEdgeSize != null) {
+    if (themeEdgeSize) {
       // Try to look up the value in theme.global.edgeSize
       // If not found, assume it's a raw CSS value like '10px'.
       const themeEdge = theme.global.edgeSize?.[themeEdgeSize] || themeEdgeSize;
--- a/src/js/components/RangeSelector/EdgeControl.js
+++ b/src/js/components/RangeSelector/EdgeControl.js
@@ -29,7 +29,7 @@ const EdgeControl = forwardRef(
     const { theme } = useThemeValue();
     const [focus, setFocus] = useState(false);
     const { cursor, fill } = DIRECTION_PROPS[direction];
-    const themeSize = theme.rangeSelector?.edge?.size;
+    const themeEdgeSize = theme.rangeSelector?.edge?.size;
     const size =
       typeof themeSize === 'string'
         ? parseMetricToNum(theme.global.edgeSize?.[themeSize] || themeSize)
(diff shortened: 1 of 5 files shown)


src/js/components/RangeSelector/EdgeControl.js
```diff
@@ -29,7 +29,11 @@ const EdgeControl = forwardRef(
     const { theme } = useThemeValue();
     const [focus, setFocus] = useState(false);
     const { cursor, fill } = DIRECTION_PROPS[direction];
-    const size = parseMetricToNum(theme.global.spacing) / 2;
+    const themeSize = theme.rangeSelector?.edge?.size;
+    const size =
```
jcfilben: What if they want to pass a px value instead of a t-shirt size?

britt6612: It still comes through

src/js/components/RangeSelector/EdgeControl.js
```diff
@@ -29,7 +29,11 @@ const EdgeControl = forwardRef(
...
+      typeof themeSize === 'string'
```
taysea: Do we need this check? Can we enforce this in typescript definition and propTypes?

src/js/components/RangeSelector/EdgeControl.js
```diff
@@ -29,7 +29,11 @@ const EdgeControl = forwardRef(
     const { theme } = useThemeValue();
     const [focus, setFocus] = useState(false);
     const { cursor, fill } = DIRECTION_PROPS[direction];
-    const size = parseMetricToNum(theme.global.spacing) / 2;
+    const themeSize = theme.rangeSelector?.edge?.size;
```
taysea: naming suggestion for clarity.
```suggestion
    const themeEdgeSize = theme.rangeSelector?.edge?.size;
```

taysea (commented):
Behavior looks good, just some minor comments.

src/js/components/RangeSelector/EdgeControl.js
britt6612: yes we dont need the check

src/js/components/RangeSelector/EdgeControl.js
```diff
@@ -29,7 +29,12 @@ const EdgeControl = forwardRef(
     const { theme } = useThemeValue();
     const [focus, setFocus] = useState(false);
     const { cursor, fill } = DIRECTION_PROPS[direction];
-    const size = parseMetricToNum(theme.global.spacing) / 2;
+    const themeEdgeSize = theme.rangeSelector?.edge?.size;
+    const size = themeEdgeSize
+      ? parseMetricToNum(
+          theme.global.edgeSize?.[themeEdgeSize] || themeEdgeSize,
```
halocline: So this logic says "look up the string to see if it is defined in the theme's global edgeSize, if so, use the referenced value; if not, its probably a CSS value and use it." Is that correct?

Should we be thinking about stronger type safety for `rangeSelector.edge.size`? Something like `EdgeSizeType | string` and introducing `EdgeSizeType`?

src/js/components/RangeSelector/__tests__/RangeSelector-test.tsx
```diff
@@ -22,6 +22,24 @@ describe('RangeSelector', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
+  test('basic with theme edge size', () => {
```
halocline, line 25: Please consider the following options for additional test coverage.



Here are example test cases you can add to further ensure the RangeSelector component correctly supports the edge size theme property:

---

### 1. Test each supported edge size

```js
const edgeSizes = ['small', 'medium', 'large', 'xlarge']; // use all supported sizes

edgeSizes.forEach(size => {
  test(`renders with edge size: ${size}`, () => {
    const theme = {
      rangeSelector: {
        edge: { size },
      },
    };
    const { container } = render(
      <Grommet theme={theme}>
        <RangeSelector values={[10, 40]} />
      </Grommet>
    );
  });
});
```

---

### 2. Test default edge size

```js
test('renders with default edge size when not specified', () => {
  const theme = {
    rangeSelector: {},
  };
  const { container } = render(
    <Grommet theme={theme}>
      <RangeSelector values={[5, 25]} />
    </Grommet>
  );
});
```

---

### 3. Test with an invalid edge size

```js
test('renders gracefully with invalid edge size', () => {
  const theme = {
    rangeSelector: {
      edge: { size: 'not-a-real-size' },
    },
  };
  const { container } = render(
    <Grommet theme={theme}>
      <RangeSelector values={[15, 35]} />
    </Grommet>
  );
});
```

---

src/js/themes/base.d.ts
```diff
@@ -1672,6 +1672,7 @@ export interface ThemeType {
     };
     edge?: {
       type?: string;
+      size?: string;
```
halocline: See comment about whether `EdgeSizeType` would be a valuable addition.

src/js/components/RangeSelector/EdgeControl.js
britt6612: yes that is correct I added some comments as well

src/js/themes/base.d.ts
britt6612: used our `EdgeSizeType` here

src/js/components/RangeSelector/EdgeControl.js
```diff
@@ -29,7 +29,23 @@ const EdgeControl = forwardRef(
     const { theme } = useThemeValue();
     const [focus, setFocus] = useState(false);
     const { cursor, fill } = DIRECTION_PROPS[direction];
-    const size = parseMetricToNum(theme.global.spacing) / 2;
+    const themeEdgeSize = theme.rangeSelector?.edge?.size;
+    let size;
+    if (themeEdgeSize != null) {
```
MikeKingdom: Any reason this is != rather than !==? Or should it simply be `if (themeEdgeSize)`? According to the type it is supposed to always be a tshirt size or some other string (probably non-empty or 0). So it seems like `if (themeEdgeSize)` should be good. If you need to support 0 it should still be something more robust (it seems like this may be possible given how you default to using `themeEdgeSize` if there's not a mapping in `global.edgeSize`.)

britt6612: agree it can be simplified

MikeKingdom (approved):
Thanks!

PR #7679, opened 2025-06-10
[DRAFT PR] - Fix/a11y file input 1.1.1 v3

Sulaymon333:
#### What does this PR do?
- Fix accessibility issue 1.1.1 for file input component
- closes https://github.com/grommet/grommet/issues/7561

- This is a draft PR with changes suggestion from #7641 
- This PR is to be ran through Accessibility test using JAWS 

- JAWS with the used to check accessibility
- TODO - jest tests to cover changes

Yes

### DIFF
--- a/src/js/components/FileInput/FileInput.js
+++ b/src/js/components/FileInput/FileInput.js
@@ -455,101 +455,127 @@ const FileInput = forwardRef(
           )}
           {files.length > 0 &&
             files.length <= aggregateThreshold &&
-            files.map((file, index) => (
-              <Box
-                key={file.name}
-                justify="between"
-                direction="row"
-                align="center"
-              >
-                {renderFile ? (
-                  renderFile(file)
-                ) : (
-                  <Box
-                    {...theme.fileInput.label}
-                    gap="xsmall"
-                    align="center"
-                    direction="row"
-                  >
-                    {((maxSize && file.size > maxSize) ||
-                      (max && index >= max)) && <CircleAlert />}
-                    <Label
-                      weight={
-                        theme.global.input.weight ||
-                        theme.global.input.font.weight
-                      }
-                      truncate
-                      {...passThemeFlag}
+            files.map((file, index) => {
+              let messageId = '';
+              if (maxSize && file.size > maxSize) {
+                messageId = 'fileInput.alert.maxSize';
+              } else if (max && index >= max) {
+                messageId = 'fileInput.alert.maxFile';
+              }
+              return (
+                <Box
+                  key={file.name}
+                  justify="between"
+                  direction="row"
+                  align="center"
+                >
+                  {renderFile ? (
+                    renderFile(file)
+                  ) : (
+                    <Box
+                      {...theme.fileInput.label}
+                      gap="xsmall"
+                      align="center"
+                      direction="row"
                     >
-                      {file.name}
-                    </Label>
-                  </Box>
-                )}
-                <Box flex={false} direction="row" align="center">
-                  <Button
-                    ref={index ? undefined : removeRef}
-                    a11yTitle={`${format({
-                      id: 'fileInput.remove',
-                      messages,
-                    })} ${file.name}`}
-                    icon={<RemoveIcon />}
-                    hoverIndicator
-                    disabled={disabled}
-                    onClick={(event) => {
-                      if (confirmRemove) {
-                        event.persist(); // necessary for when React < v17
-                        setPendingRemoval({ event, index });
-                        setShowRemoveConfirmation(true);
-                      } else removeFile(index);
-                    }}
-                  />
-                  {files.length === 1 && (
-                    <Keyboard
-                      onSpace={(event) => {
-                        if (controlRef.current === event.target)
-                          inputRef.current.click();
-                      }}
-                      onEnter={(event) => {
-                        if (controlRef.current === event.target)
-                          inputRef.current.click();
-                      }}
-                    >
-                      {theme.fileInput.button ? (
-                        <Button
-                          // The focus here is redundant for keyboard users
-                          tabIndex={-1}
-                          ref={controlRef}
-                          kind={theme.fileInput.button}
-                          label={format({
-                            id: 'fileInput.browse',
+                      {((maxSize && file.size > maxSize) ||
+                        (max && index >= max)) && (
+                        <CircleAlert
+                          tabIndex={0}
+                          onClick={() => {}}
+                          hoverIndicator
+                          a11yTitle={format({
+                            id: messageId,
                             messages,
-                          })}
-                          onClick={() => {
-                            inputRef.current.click();
-                            inputRef.current.focus();
-                          }}
-                        />
-                      ) : (
-                        <Anchor
-                          // The focus here is redundant for keyboard users
-                          tabIndex={-1}
-                          ref={controlRef}
-                          margin="small"
-                          onClick={() => {
-                            inputRef.current.click();
-                            inputRef.current.focus();
-                          }}
-                          label={format({
-                            id: 'fileInput.browse',
-                            messages,
+                            values: {
+                              max,
+                              fileName: file.name,
+                              maxSize: formatBytes(maxSize),
+                            },
                           })}
                         />
                       )}
-                    </Keyboard>
+                      <Label
+                        weight={
+                          theme.global.input.weight ||
+                          theme.global.input.font.weight
+                        }
+                        truncate
+                        {...passThemeFlag}
+                      >
+                        {file.name}
+                      </Label>
+                    </Box>
                   )}
+                  <Box flex={false} direction="row" align="center">
+                    <Button
+                      ref={index ? undefined : removeRef}
+                      a11yTitle={format({
+                        id: 'fileInput.remove',
+                        messages,
+                        values: {
+                          fileName: file.name,
+                        },
+                      })}
+                      icon={<RemoveIcon />}
+                      hoverIndicator
+                      disabled={disabled}
+                      onClick={(event) => {
+                        if (confirmRemove) {
+                          event.persist(); // necessary for when React < v17
+                          setPendingRemoval({ event, index });
+                          setShowRemoveConfirmation(true);
+                        } else removeFile(index);
+                      }}
+                    />
+                    {files.length === 1 && (
+                      <Keyboard
+                        onSpace={(event) => {
+                          if (controlRef.current === event.target)
+                            inputRef.current.click();
+                        }}
+                        onEnter={(event) => {
+                          if (controlRef.current === event.target)
+                            inputRef.current.click();
+                        }}
+                      >
+                        {theme.fileInput.button ? (
+                          <Button
+                            // The focus here is redundant for keyboard users
+                            tabIndex={-1}
+                            ref={controlRef}
+                            kind={theme.fileInput.button}
+                            label={format({
+                              id: 'fileInput.browse',
+                              messages,
+                            })}
+                            onClick={() => {
+                              inputRef.current.click();
+                              inputRef.current.focus();
+                            }}
+                          />
+                        ) : (
+                          <Anchor
+                            // The focus here is redundant for keyboard users
+                            tabIndex={-1}
+                            ref={controlRef}
+                            margin="small"
+                            onClick={() => {
+                              inputRef.current.click();
+                              inputRef.current.focus();
+                            }}
+                            label={format({
+                              id: 'fileInput.browse',
+                              messages,
+                            })}
+                          />
+                        )}
+                      </Keyboard>
+                    )}
+                  </Box>
                 </Box>
-              </Box>
-            ))}
+              );
+            })}
         </ContentsBox>
         {showRemoveConfirmation && (
           <ConfirmRemove
--- a/src/js/components/FileInput/index.d.ts
+++ b/src/js/components/FileInput/index.d.ts
@@ -16,9 +16,18 @@ export interface FileInputProps {
     dropPrompt?: string;
     dropPromptMultiple?: string;
     files?: string;
+    maxFile?: string;
     remove?: string;
     removeAll?: string;
-    maxFile?: string;
+    maxSizeSingle?: string;
+    maxSizeMultiple?: {
+      singular?: string;
+      plural?: string;
+    };
+    alert?: {
+      maxSize?: string;
+      maxFile?: string;
+    };
   };
   multiple?: boolean | { aggregateThreshold?: number; max?: number };
   name?: string;
--- a/src/js/components/FileInput/propTypes.js
+++ b/src/js/components/FileInput/propTypes.js
@@ -11,8 +11,18 @@ if (process.env.NODE_ENV !== 'production') {
       dropPrompt: PropTypes.string,
       dropPromptMultiple: PropTypes.string,
       files: PropTypes.string,
+      maxFile: PropTypes.string,
       remove: PropTypes.string,
       removeAll: PropTypes.string,
+      maxSizeSingle: PropTypes.string,
+      maxSizeMultiple: PropTypes.shape({
+        singular: PropTypes.string,
+        plural: PropTypes.string,
+      }),
+      alert: PropTypes.shape({
+        maxSize: PropTypes.string,
+        maxFile: PropTypes.string,
+      }),
     }),
     multiple: PropTypes.oneOfType([
       PropTypes.bool,
--- a/src/js/components/Grommet/index.d.ts
+++ b/src/js/components/Grommet/index.d.ts
@@ -100,12 +100,16 @@ export interface GrommetProps {
         files?: string;
         maxFile?: string;
         maxSizeSingle?: string;
+        remove?: string;
+        removeAll?: string;
         maxSizeMultiple?: {
           singular?: string;
           plural?: string;
         };
-        remove?: string;
-        removeAll?: string;
+        alert?: {
+          maxSize?: string;
+          maxFile?: string;
+        };
       };
       form?: {
         invalid?: string;
--- a/src/js/components/Grommet/propTypes.js
+++ b/src/js/components/Grommet/propTypes.js
@@ -98,13 +98,17 @@ if (process.env.NODE_ENV !== 'production') {
           dropPromptMultiple: PropTypes.string,
           files: PropTypes.string,
           maxFile: PropTypes.string,
+          remove: PropTypes.string,
+          removeAll: PropTypes.string,
           maxSizeSingle: PropTypes.string,
           maxSizeMultiple: PropTypes.shape({
             singular: PropTypes.string,
             plural: PropTypes.string,
           }),
-          remove: PropTypes.string,
-          removeAll: PropTypes.string,
+          alert: PropTypes.shape({
+            maxSize: PropTypes.string,
+            maxFile: PropTypes.string,
+          }),
         }),
         form: PropTypes.shape({
           invalid: PropTypes.string,
--- a/src/js/languages/default.json
+++ b/src/js/languages/default.json
@@ -77,13 +77,17 @@
     "dropPromptMultiple": "Drop files here or",
     "files": "files",
     "maxFile": "Attach a maximum of {max} files only.",
+    "remove": "remove file: {fileName}",
+    "removeAll": "remove all",
     "maxSizeSingle": "The file is too large. Select a file no larger than {maxSize}.",
     "maxSizeMultiple": {
       "singular": "One file is too large. Select files which are no larger than {maxSize}.",
       "plural": "{numOfFiles} files are too large. Select files which are no larger than {maxSize}."
     },
-    "remove": "remove",
-    "removeAll": "remove all"
+    "alert": {
+      "maxSize": "Error - remove file: {fileName} to continue. This file is too large. Attach files which are no larger than {maxSize}.",
+      "maxFile": "Error - remove file: {fileName} to continue.  Attach a maximum of {max} files only."
+    }
   },
   "form": {
     "invalid": "invalid",
--- a/src/js/themes/__tests__/__snapshots__/theme-test.js.snap
+++ b/src/js/themes/__tests__/__snapshots__/theme-test.js.snap
@@ -1566,7 +1566,7 @@ exports[`Grommet hpe theme 1`] = `
   display: flex;
   box-sizing: border-box;
   max-width: 100%;
-  background-color: #3c3aa1;
+  background-color: #00384d;
   color: #ffffff;
   min-width: 0;
   min-height: 0;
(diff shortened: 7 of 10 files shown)


src/js/components/FileInput/index.d.ts
```diff
@@ -16,9 +16,18 @@ export interface FileInputProps {
...
-    maxFile?: string;
+    maxSizeSingle?: string;
+    maxSizeMultiple?: {
+      singular?: string;
+      plural?: string;
+    };
+    alert?: {
+      maxSize?: string;
+      maxFile?: string;
+    };
```
MikeKingdom: Try to keep all these in alphabetical order

src/js/components/FileInput/propTypes.js
```diff
@@ -11,8 +11,18 @@ if (process.env.NODE_ENV !== 'production') {
...
       removeAll: PropTypes.string,
+      maxSizeSingle: PropTypes.string,
+      maxSizeMultiple: PropTypes.shape({
+        singular: PropTypes.string,
+        plural: PropTypes.string,
+      }),
+      alert: PropTypes.shape({
+        maxSize: PropTypes.string,
+        maxFile: PropTypes.string,
+      }),
```
MikeKingdom: same

src/js/components/Grommet/propTypes.js
```diff
@@ -98,13 +98,17 @@ if (process.env.NODE_ENV !== 'production') {
...
           maxSizeMultiple: PropTypes.shape({
             singular: PropTypes.string,
             plural: PropTypes.string,
           }),
-          remove: PropTypes.string,
-          removeAll: PropTypes.string,
+          alert: PropTypes.shape({
+            maxSize: PropTypes.string,
+            maxFile: PropTypes.string,
+          }),
```
MikeKingdom: same

src/js/languages/default.json
```diff
@@ -77,13 +77,17 @@
...
     "maxSizeMultiple": {
       "singular": "One file is too large. Select files which are no larger than {maxSize}.",
       "plural": "{numOfFiles} files are too large. Select files which are no larger than {maxSize}."
     },
-    "remove": "remove",
-    "removeAll": "remove all"
+    "alert": {
+      "maxSize": "Error - remove file: {fileName} to continue. This file is too large. Attach files which are no larger than {maxSize}.",
+      "maxFile": "Error - remove file: {fileName} to continue.  Attach a maximum of {max} files only."
+    }
```
MikeKingdom: same. I still think we need to avoid saying "Error - remove file"
```suggestion
    "alert": {
      "maxSize": "File {fileName} is too large. Attach files which are no larger than {maxSize}.",
      "maxFile": "Attach a maximum of {max} files only. File {fileName} is beyond this limit"
    }
```

src/js/themes/__tests__/__snapshots__/theme-test.js.snap
MikeKingdom, line 1: What caused these diffs? Did we change a theme color somewhere?

MikeKingdom (commented):
Thanks for the changes. It's looking a little cleaner.

src/js/components/FileInput/index.d.ts
Sulaymon333: done

src/js/languages/default.json
Sulaymon333: I kept the "Error - remove file:..." because of last review I got from Bill below.


"Hello Sulaymon,
I like option 1.
You can learn the reason in a recording I did for both versions at the link below.

[Evaluating w JAWS Alerting Screen Readers Users that a file size is too large or too many files are uploaded-20250605_200407-Meeting Recording.mp4](https://hpe-my.sharepoint.com/:v:/p/bill_tipton/EXk8T06L1gtFr2SZS8gAizUBTJ5hDLDho99x4aYijjROCg?e=zDRIxE&nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJTdHJlYW1XZWJBcHAiLCJyZWZlcnJhbFZpZXciOiJTaGFyZURpYWxvZy1MaW5rIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXcifSwicGxheWJhY2tPcHRpb25zIjp7fX0%3D)

One suggestion: Although the button in focus says Remove: when telling the user about the error you may want to tell them at the end of that sentence what action must be done to proceed. Such as the file must be removed or something close to that. Then the user knows there is an error and what is needed to continue."

I have shared this version with him as well and I am expecting his evaluation when he is bank next week.

src/js/themes/__tests__/__snapshots__/theme-test.js.snap
Sulaymon333: I did not make any changes to theme color.

src/js/components/FileInput/propTypes.js
Sulaymon333: done

src/js/languages/default.json
MikeKingdom: > I kept the "Error - remove file:..." because of last review I got from Bill below.

Thanks for review with Bill. That's awesome. I think we can achieve the same thing but avoiding saying it as "Error ...." It's actually one of the standards across a number of our UX folks to avoid that kind of wording. I appreciate the comment from Bill and I'm sure we can still word this well but avoid conflicting with other messaging guidance that's more global to our product base.

src/js/themes/__tests__/__snapshots__/theme-test.js.snap
MikeKingdom: Please see if you can determine why.

sulaymon-tajudeen-hpe: OK. I am looking into it more closely.

src/js/languages/default.json
sulaymon-tajudeen-hpe: Thank you for the review. that makes sense. the alert messages are now updated.

britt6612: @MikeKingdom I do get your comment about using the word `error` Bill just mentioned that its nice to know that something is wrong right away instead of listening for the screen reader to finish the filename (especially if its long) to then say the file was to large. So I wonder if we can do something like 

`This file {name} is too large to upload. Remove it to continue.`
`File is too large. 'filename.pdf' must be removed to continue.`

Sulaymon333: @britt6612 thank you for the comment.

I would add something like `Alert` or `Attention` at the beginning of the text and possibly mention the `max size` as well.

So something like

maxSize: `Attention, this file {name} is too large. Attach files which are no larger than {maxSize} to continue`.
maxFile: `Attention, attach a maximum of {max} files only. File {fileName} is beyond this limit.`

For maxFile, I believe with the tone above, we are allowing the user to decide what they would like to do. Like [Mike pointed out earlier ](https://github.com/grommet/grommet/pull/7641#issuecomment-2964460837)in one of the previous comments for when users attached too many files.

MikeKingdom: > @MikeKingdom I do get your comment about using the word `error` Bill just mentioned that its nice to know that something is wrong right away instead of listening for the screen reader to finish the filename (especially if its long) to then say the file was to large. So I wonder if we can do something like
> 
> `This file {name} is too large to upload. Remove it to continue.` `File is too large. 'filename.pdf' must be removed to continue.`

This is more like how we've been asked to word things from a "tone" perspective. We used to have someone that focused on tone.

I'll let you folks work out what works best for Bill since he's obviously a better authority on that.

src/js/themes/__tests__/__snapshots__/theme-test.js.snap
britt6612: @sulaymon-tajudeen-hpe did you run `yarn` to fix snapshots?

Sulaymon333: I simply ran `yarn test-update` to update the snapshots and then those extra files showed up. I felt it is related to the react 19 vs 18 we were discussing on slack or so.

I will have to rebase to master branch or merge from master at some point.

britt6612: yes can you rebase to master now. or fix before we review?

britt6612: it would be nice to have files cleaned up for review

src/js/components/Grommet/index.d.ts
```diff
@@ -100,12 +100,16 @@ export interface GrommetProps {
...
         maxSizeSingle?: string;
+        remove?: string;
+        removeAll?: string;
         maxSizeMultiple?: {
           singular?: string;
           plural?: string;
         };
-        remove?: string;
-        removeAll?: string;
+        alert?: {
```
britt6612: nit pick but can you make these alphabetic order

src/js/components/FileInput/FileInput.js
```diff
@@ -455,101 +455,125 @@ const FileInput = forwardRef(
...
-                          margin="small"
-                          onClick={() => {
-                            inputRef.current.click();
-                            inputRef.current.focus();
-                          }}
-                          label={format({
-                            id: 'fileInput.browse',
+                      {((maxSize && file.size > maxSize) ||
+                        (maxFile && index >= maxFile)) && (
+                        <CircleAlert
```
Sulaymon333, line 492: @MikeKingdom @britt6612 will it be ok to update to Button component to have focus indicator like code below
```javascript
<Button
          plain
          a11yTitle={format({
            id: messageId,
            messages,
            values: {
              maxFile,
              fileName: file.name,
              maxSize: formatBytes(maxSize),
            },
          })}
          icon={<CircleAlert />}
        />
      )}
```

britt6612: why would we make this change? reasoning?

Sulaymon333: So that we can have the normal focus indicator on the alert icon like we have on the remove icon. 

image

At the moment, we have only a blue outline, the browser default I believe.

image

britt6612: ah yes should be the focus indicator from grommet or hpe and not native focus

MikeKingdom: I'm kinda missing something. Why would the alert icon ever be a button or focusable? The only clickable/focusable thing there should be the X icon

Sulaymon333: Based on your suggestion of separating the alert message vs the remove message read to by the screen reader, I added `tabIndex` to the alert icon so that sr users can tab into it and the message can be read out. This is the implementation we tested with Bill and he said he also likes that the two messages are separated now.

Do you have any implementation approach in mind?

Sulaymon333: [UPDATE] - circleAlert icon from is now removed from tab order.

src/js/languages/default.json
```diff
@@ -72,17 +72,21 @@
...
+    "alert": {
+      "maxSize": "Attention, this file {fileName} is too large. Attach files which are no larger than {maxSize} to continue.",
+      "maxFile": "Attention, attach a maximum of {maxFile} files only. File {fileName} is beyond this limit."
+    },
     "browse": "browse",
     "dropPrompt": "Drop file here or",
     "dropPromptMultiple": "Drop files here or",
     "files": "files",
-    "maxFile": "Attach a maximum of {max} files only.",
+    "maxFile": "Attach a maximum of {maxFile} files only.",
```
taysea: I think renaming the "max" variable in this message would be a breaking change for anyone relying on it in custom messages. Can you update so it's still called "max"?

src/js/languages/default.json
```diff
@@ -72,17 +72,21 @@
     "label": "View"
   },
   "fileInput": {
+    "alert": {
+      "maxSize": "Attention, this file {fileName} is too large. Attach files which are no larger than {maxSize} to continue.",
```
taysea: Can we try to make this more brief, while still conveying the meaning? Wording suggestion:

"{fileName} exceeds {maxSize} limit."

src/js/languages/default.json
```diff
@@ -72,17 +72,21 @@
...
     "files": "files",
-    "maxFile": "Attach a maximum of {max} files only.",
+    "maxFile": "Attach a maximum of {maxFile} files only.",
     "maxSizeSingle": "The file is too large. Select a file no larger than {maxSize}.",
     "maxSizeMultiple": {
       "singular": "One file is too large. Select files which are no larger than {maxSize}.",
       "plural": "{numOfFiles} files are too large. Select files which are no larger than {maxSize}."
     },
-    "remove": "remove",
+    "remove": "remove file: {fileName}",
```
taysea:
```suggestion
    "remove": "remove {fileName}",
```
Small suggestion for brevity.

src/js/languages/default.json
```diff
@@ -72,17 +72,21 @@
     "label": "View"
   },
   "fileInput": {
+    "alert": {
+      "maxSize": "Attention, this file {fileName} is too large. Attach files which are no larger than {maxSize} to continue.",
+      "maxFile": "Attention, attach a maximum of {maxFile} files only. File {fileName} is beyond this limit."
```
taysea: Suggestion for brevity:

"{fileName} exceeds limit of {maxFile} file limit."

taysea: What's the different between this "maxFile" message and when the other "maxFile" message will be read out? Any opportunity to just reuse the existing message?

src/js/components/FileInput/index.d.ts
```diff
@@ -12,13 +12,22 @@ export interface FileInputProps {
...
     browse?: string;
     dropPrompt?: string;
     dropPromptMultiple?: string;
     files?: string;
+    maxFile?: string;
+    maxSizeSingle?: string;
```
taysea, line 24: This doesn't seem to align with the messages in default.json -- can you align?

src/js/languages/default.json
britt6612: maybe still leave "Attention" at the front Bill said this was better for him to indicate he needs to listen to what is wrong instead of just the fileName since that can get long

britt6612: agree one being more brief though

src/js/languages/default.json
Sulaymon333: It is based on Bill's comment to have something like `Error, Alert, or Attention` at the beginning of the message and to let him know he has to do something by stating what needs to be done to continue. So creating a new variable gives room to meet this recommendation without breaking the current usage of established variables.

src/js/languages/default.json
Sulaymon333: So something like 
`Attention, {fileName} exceeds {maxFile} files limit.`
`Attention, {fileName} exceeds {maxSize} limit.`

src/js/languages/default.json
Sulaymon333: yes, sure and since we only support file uploads for now.

src/js/languages/default.json
Sulaymon333: ok noted.

src/js/languages/default.json
taysea: I don't think we need to say "Attention" at all.

taysea: The use of "Attention"/"Alert" feels a bit out of our standard messaging, but I am open to it given Bill's direct feedback. @jcfilben thoughts?

britt6612: maybe we can think of another word. I know Bill said initially `Error` but see Mike comment above we want to stay away from using that.

src/js/components/FileInput/index.d.ts
Sulaymon333: Sorry, I am not clear about this comment.

The two looks the same.

image

src/js/languages/default.json
taysea: I'd be fine with "Attention" or "Alert" -- when I said "standard messaging" I mostly meant we don't usually say anything like "attention", "alert", etc.

src/js/components/FileInput/index.d.ts
taysea: Thanks, I had missed the other messages since they were collapsed.

src/js/languages/default.json
jcfilben: I feel like there might be some wcag or apg recommendation for error messaging language, let me see if I can find that

jcfilben: Didn't find anything directly from wcag or apg but found these resources:
- https://uit.stanford.edu/accessibility/testing/errors
Stanford recommends using 'Error' in the message
- https://www.tpgi.com/writing-accessible-form-messages/
TPGI doesn't state that 'Error' or a similar alternative should be used but instead recommends avoiding robotic language, placing blame on the user, and jargon such as 'whoops' or 'uh-oh'

While I was looking around I found the aria-invalid attribute https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA21. Maybe we could explore using this and see how it affects the screenreader announcements

jcfilben: I think whatever word we land on using just make sure we are consistent with it and it is clear to users.

jcfilben: I think this is what Mike was referring to when he mentioned wanting to avoid the word 'error': https://design-system.hpe.design/foundation/voice-and-tone?q=voice%20an#objective
I feel like we might want to revisit this as 'error' feels more clear in this situation than a word like 'attention' or 'alert'.

src/js/languages/default.json
```diff
@@ -72,17 +72,21 @@
     "label": "View"
   },
   "fileInput": {
+    "alert": {
+      "maxFile": "Error, {fileName} exceeds {maxFile} files limit.",
```
britt6612, line 91: Im not hearing this message being read out is that intentional?

Sulaymon333: You can use `ctrl+option`.

It's agreed to be removed out of the tabIndex, since the alert icon didn't have user action on it. It's simply informational.

britt6612: I knew we removed from the tabIndex however I thought based on Taylors comment earlier it would be announced?

https://hpe.slack.com/archives/CPYJD4Q0K/p1750881349958279?thread_ts=1750880803.886469&cid=CPYJD4Q0K

> I agree that it seems odd to be making this focusable if it's not interactive. There are a lot of things on the screen that can be reachable by screenreader but not have tabindex 0 (such as the file names, amount of files, etc.).
> It feels like instead, maybe we could leverage the announce context or something to announce when there are too many files uploaded (since it's not really specific to that file but more so conveying there are 2 files more than the limit.

src/js/components/FileInput/utils/formatBytes.js
```diff
@@ -12,6 +12,9 @@ const getCurrentOS = () => {
 };
 
 const defaultFormat = (size) => {
+  if (size == null || size === undefined) {
```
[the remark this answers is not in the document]
MikeKingdom: Can size be an empty string? If so this will be false. Using == vs === is a bit subtle and its hard to tell sometimes if it really was intended. With the whole truthy/falsey ability in javascript you have to be careful what happens with 0, "0", "", undefined, null, false. For that reason we tend to always use === rather than ==, at least without commenting it.

src/js/components/FileInput/utils/formatBytes.js
```diff
@@ -24,6 +27,9 @@ const defaultFormat = (size) => {
 };
 
 const windowsFormat = (size) => {
+  if (size == null || size === undefined) {
```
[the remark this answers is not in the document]
MikeKingdom: same

src/js/languages/default.json
britt6612: talked through this offline
```We settle to drop the announceContext usage since that feels much like solving for 4.1.2.```

File: src/js/components/FileInput/utils/formatBytes.js (hunk 13 above)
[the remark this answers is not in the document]
Comment (Sulaymon333) [reply src/js/components/FileInput/utils/formatBytes.js]: This a good catch. I have updated formatBytes.js to cover more edge case

**Review by britt6612 — APPROVED** [review]
LGTM Thank you for the clean up

File: src/js/components/FileInput/utils/formatBytes.js [hunk 15]
```diff
@@ -24,9 +28,13 @@ const defaultFormat = (size) => {
 };
 
 const windowsFormat = (size) => {
-  const num = Math.ceil(size / IEC_CONVERSION_FACTOR);
+  if (size == null || size === '' || Number.isNaN(Number(size))) {
```
jcfilben: I'm wondering if it makes more sense to avoid calling formatBytes in these situations rather than returning '0 KB'. I feel like this could trip us up in the future if size is null but then gets transformed to '0 KB'

MikeKingdom (approved):
Thanks!

src/js/components/FileInput/utils/formatBytes.js
Sulaymon333: May be we can return an empty string?

Sulaymon333: I noticed there is some sort of early return for when there is any validation error on Submit.

This return some message if `maxSize` is not a number and it overrides `required` message when file input is empty. Can we prioritize showing this message before `required` when maxSize is not right type (number)? Will this be a pattern we would like to adopt?

image

image

image

Sulaymon333: Or may be we can just console.warn instead of sending a message. I have seen a similar pattern for  Grommet dataTable primaryKey type checking

image

britt6612: Im not sure Im following completely but can we just return `undefined` ? 

```
const windowsFormat = (size) => {
  if (size == null || size === '' || Number.isNaN(Number(size))) {
    return undefined; 
  }
```

jcfilben: I don't think we need to show a message in the UI since this is a error cause by the implementor not the user. I would be okay with returning undefined. I'm wondering why we are even getting into a situation where we call formatBytes with an invalid value. Seems like maybe we could have a check before we even get to that point

Sulaymon333: I removed the `null/undefined/empty string` check and replace with converting value passed down to a `Number` type.  The type conversion before rounding via `.toFixed()` helps to fix this invalid type issue.

Reason why we might end up with invalid value with `formatBytes` in the alert icon message is where either of the variables `max` and `maxSize` can be undefined, since they are not required values. so `.toFixed(1)` throws error trying yo convert the invalid types. now converting the value passed first to Number helps fixed this issue.

image

jcfilben: If we force conversion to a Number then I feel like we run into the same issue initially mentioned:

> I feel like this could trip us up in the future if size is null but then gets transformed to '0 KB'

Sulaymon333: Now updated with checking type of `maxSize` and `max` to be `number` before usage

Issue #7607, opened 2025-04-28
Notification - WCAG rule 4.1.3


SeamusLeonardHPE:
**Failed rule**

4.1.3
Status Messages
Level AA

In content implemented using markup languages, status messages can be programmatically determined through role or properties such that they can be presented to the user by assistive technologies without receiving focus.

---

**Notes**

Failing since the toast notification is not being announced

This rule is especially relevant for toast and inline notifications, because they:

Often appear dynamically (e.g., after form submission)
Do not receive focus, but still need to be communicated

To comply:

Use ARIA roles like:
- role="status" → polite updates (e.g., "Changes saved")
- role="alert" → urgent updates (e.g., "Connection lost")

Ensure the content is present in the DOM and updated when the message changes
The container should not receive focus, but must be in the accessibility tree

britt6612:
@jcfilben 

thoughts on the notes:

```

Notes

Failing since the toast notification is not being announced

This rule is especially relevant for toast and inline notifications, because they:

Often appear dynamically (e.g., after form submission)
Do not receive focus, but still need to be communicated

To comply:

Use ARIA roles like:

role="status" → polite updates (e.g., "Changes saved")
role="alert" → urgent updates (e.g., "Connection lost")
Ensure the content is present in the DOM and updated when the message changes
The container should not receive focus, but must be in the accessibility tree

```

from above but I know we talked about just only having the `role="status"` we still good with that route?

jcfilben:
I'm good with just using role="status"

britt6612:
closed via https://github.com/grommet/grommet/pull/7708
and updated on DS site

PR #7680, opened 2025-06-11
Improve announceContext utility

britt6612:
#### What does this PR do?
Improves the behavior of the AnnounceContext utility used across Grommet to trigger screen reader announcements. This includes:
1. Rendering a visually hidden div that screen readers can detect reliably.
2. Adding `aria-atmoic="true"` as a best practice to be used with aria-live. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Guides/Live_regions
 https://dequeuniversity.com/rules/axe/4.7/aria-valid-attr-value
4. Using `requestAnimationFrame` to ensure the DOM updates are flushed before setting the message, which improves consistency across screen readers
AnnounceContext.js
Testing with voiceover in storybook
screen reader on toast stories

The previous implementation relied on toggling `aria-live` and injecting content directly into a detached container. However, screen readers did not consistently pick up these announcements — particularly for aria-live="polite".

Previously, the announcements were not reliably triggered when using aria-live="polite", due to timing and DOM structure issues

https://www.accessibility-developer-guide.com/examples/sensible-aria-usage/
https://css-tricks.com/inclusively-hidden/
https://stackoverflow.com/questions/75520027/using-a-separate-status-div-with-aria-live-assertive-to-handle-toast-message-c
https://github.com/nvaccess/nvda/issues/12464
related to #7607 #7607
none
no
yes
 backwards compatible

### DIFF
--- a/src/js/components/Data/__tests__/Data-test.tsx
+++ b/src/js/components/Data/__tests__/Data-test.tsx
@@ -168,7 +168,7 @@ describe('Data', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
-  test('view sort', async () => {
+  test('view sort', () => {
     const { container } = render(
       <Grommet>
         <Data
@@ -187,7 +187,7 @@ describe('Data', () => {
       </Grommet>,
     );
 
-    await expectDataSummary('4 items');
+    expectDataSummary('4 items');
     expect(container.firstChild).toMatchSnapshot();
   });
 
--- a/src/js/components/Data/__tests__/Data-test.tsx
+++ b/src/js/components/Data/__tests__/Data-test.tsx
@@ -16,7 +16,7 @@ const DEBOUNCE_TIMEOUT = 300;
 
 // asserts that AnnounceContext aria-live region and visible DataSummary each have this text
 const expectDataSummary = (message: string) =>
-  expect(screen.getAllByText(message)).toHaveLength(2);
+  expect(screen.getByText(message)).toBeInTheDocument();
 
 const data = [
   {
@@ -168,7 +168,7 @@ describe('Data', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
-  test('view sort', () => {
+  test('view sort', async () => {
     const { container } = render(
       <Grommet>
         <Data
@@ -187,7 +187,7 @@ describe('Data', () => {
       </Grommet>,
     );
 
-    expectDataSummary('4 items');
+    await expectDataSummary('4 items');
     expect(container.firstChild).toMatchSnapshot();
   });
 
--- a/src/js/components/Grommet/__tests__/__snapshots__/Grommet-test.js.snap
+++ b/src/js/components/Grommet/__tests__/__snapshots__/Grommet-test.js.snap
@@ -22,19 +22,19 @@ exports[`Grommet announce 1`] = `
 
 exports[`Grommet announce 2`] = `
 <div
+  aria-atomic="true"
   aria-live="assertive"
   id="grommet-announcer"
-  style="left: -100%; right: 100%; position: fixed; z-index: -1;"
->
-  hello
-</div>
+  style="position: absolute; left: -1px; height: 1px; width: 1px; overflow: hidden; clip-path: rect(1px, 1px, 1px, 1px); white-space: nowrap;"
+/>
 `;
 
 exports[`Grommet announce 3`] = `
 <div
+  aria-atomic="true"
   aria-live="assertive"
   id="grommet-announcer"
-  style="left: -100%; right: 100%; position: fixed; z-index: -1;"
+  style="position: absolute; left: -1px; height: 1px; width: 1px; overflow: hidden; clip-path: rect(1px, 1px, 1px, 1px); white-space: nowrap;"
 />
 `;
 
--- a/src/js/components/Grommet/__tests__/__snapshots__/Grommet-test.js.snap
+++ b/src/js/components/Grommet/__tests__/__snapshots__/Grommet-test.js.snap
@@ -20,23 +20,9 @@ exports[`Grommet announce 1`] = `
 </div>
 `;
 
-exports[`Grommet announce 2`] = `
-<div
-  aria-atomic="true"
-  aria-live="assertive"
-  id="grommet-announcer"
-  style="position: absolute; left: -1px; height: 1px; width: 1px; overflow: hidden; clip-path: rect(1px, 1px, 1px, 1px); white-space: nowrap;"
-/>
-`;
+exports[`Grommet announce 2`] = `null`;
 
-exports[`Grommet announce 3`] = `
-<div
-  aria-atomic="true"
-  aria-live="assertive"
-  id="grommet-announcer"
-  style="position: absolute; left: -1px; height: 1px; width: 1px; overflow: hidden; clip-path: rect(1px, 1px, 1px, 1px); white-space: nowrap;"
-/>
-`;
+exports[`Grommet announce 3`] = `null`;
 
 exports[`Grommet background 1`] = `
 .c0 {
--- a/src/js/components/Grommet/__tests__/__snapshots__/Grommet-test.js.snap
+++ b/src/js/components/Grommet/__tests__/__snapshots__/Grommet-test.js.snap
@@ -20,9 +20,23 @@ exports[`Grommet announce 1`] = `
 </div>
 `;
 
-exports[`Grommet announce 2`] = `null`;
+exports[`Grommet announce 2`] = `
+<div
+  aria-atomic="true"
+  aria-live="assertive"
+  id="grommet-announcer"
+  style="position: absolute; left: -9999px; height: 1px; width: 1px; overflow: hidden;"
+/>
+`;
 
-exports[`Grommet announce 3`] = `null`;
+exports[`Grommet announce 3`] = `
+<div
+  aria-atomic="true"
+  aria-live="assertive"
+  id="grommet-announcer"
+  style="position: absolute; left: -9999px; height: 1px; width: 1px; overflow: hidden;"
+/>
+`;
 
 exports[`Grommet background 1`] = `
 .c0 {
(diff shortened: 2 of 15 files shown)


src/js/components/Notification/Notification.js
```diff
@@ -17,6 +17,38 @@ import { Text } from '../Text';
...
+  if (React.isValidElement(node)) return extractText(node.props.children);
+  return '';
+};
+
+const ScreenReaderAnnouncement = ({ title, message }) => {
+  const text = [title, message].map(extractText).filter(Boolean).join('. ');
+
+  return (
+    <div
+      role="alert"
```
britt6612: wonder if we should base the role off the `status` prop

britt6612: same with `aria-live`

src/js/components/Notification/Notification.js
```diff
@@ -252,14 +283,15 @@ const Notification = ({
     content = visible && (
       <Layer
         {...theme.notification.toast.layer}
-        role="log"
```
britt6612, line 255: not sure `role="log"` made sense here 

```
A type of live region where new information is added in meaningful order and old information may disappear. See related marquee.

Examples include chat logs, messaging history, game log, or an error log. In contrast to other live regions, in this role there is a relationship between the arrival of new items in the log and the reading order. The log contains a meaningful sequence and new information is added only to the end of the log, not at arbitrary points.

Elements with the role log have an implicit aria-live value of polite.

```

https://www.digitala11y.com/log-role/
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/log_role
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/log_role

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -4,30 +4,47 @@ import { AnnounceContextPropTypes } from './propTypes';
...
-    announcer.setAttribute('aria-live', 'off');
-    announcer.innerHTML = message;
+    // Update aria-live before setting content
     announcer.setAttribute('aria-live', mode);
-    setTimeout(() => {
-      announcer.innerHTML = '';
-    }, timeout);
+
+    // Use requestAnimationFrame to ensure DOM updates are complete
+    requestAnimationFrame(() => {
```
britt6612: requestAnimationFrame() schedules code to run just before the next screen repaint. It ensures the browser processes earlier DOM updates first

https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -4,30 +4,47 @@ import { AnnounceContextPropTypes } from './propTypes';
...
+      // Set content in next frame
+      requestAnimationFrame(() => {
```
britt6612: Using two nested requestAnimationFrame() calls ensures screen readers reliably detect
 content updates in the live region

 1. The first requestAnimationFrame() gives the browser a chance to complete layout and
accessibility tree updates after we clear the old content and update the aria-live attribute. Without this pause, screen readers might not register that anything changed.
 2. The second requestAnimationFrame() schedules the actual text content update (the new message)
    in the next frame. This guarantees that the browser has fully processed the previous state,
    and the screen reader is more likely to detect the change as new content to announce.



 In our issue with Toast
When a toast appears with `aria-live="polite"`, it's often competing with ongoing screen reader narration.
If the DOM update (clearing and setting the message) happens too fast, it may not be treated as a fresh change.
This results in the toast never being read aloud — even though the intent was for it to be politely queued.

src/js/components/Notification/Notification.js
jcfilben: Agree role='log' doesn't seem correct

src/js/components/Notification/Notification.js
```diff
@@ -252,7 +274,9 @@ const Notification = ({
...
+        role={status === 'critical' ? 'alert' : 'status'}
```
jcfilben: I'm wondering if toast notifications should always just be role='status' instead of using 'alert' since toast notifications should never be used to convey high attention, critical, or time sensitive information

jcfilben (commented):
On the AnnounceContext/Polite story the announcement is being read out twice for me (tested on chrome)

src/js/components/Notification/Notification.js
MikeKingdom: I agree you shouldn't given the transient nature but if status is critical that's what they did

britt6612: that is a good point.

``` 
since toast notifications should never be used to convey high attention, critical, or time sensitive information

```

britt6612: in this case @jcfilben if we keep the `status` of alert then we wouldnt need to add `assertive` just keep that as polite always?

jcfilben: Not sure if I am following, I feel like it probably still makes sense to have as assertive regardless of the role just so that as soon as the toast pops up the announcement gets read out

britt6612: oh really? I was thinking 'polite' what if its non critical? like `File uploaded successfully` then would that be assertive? or can that wait ?

jcfilben: If polite is reading out okay them I'm good with leaving as polite

britt6612: okay

britt6612: yes it is

jcfilben:
I'm still getting the announcement read out twice on AnnounceContext/Polite

src/js/components/Data/__tests__/Data-test.tsx
```diff
@@ -16,7 +16,7 @@ const DEBOUNCE_TIMEOUT = 300;
 
 // asserts that AnnounceContext aria-live region and visible DataSummary each have this text
 const expectDataSummary = (message: string) =>
-  expect(screen.getAllByText(message)).toHaveLength(2);
+  expect(screen.getByText(message)).toBeInTheDocument();
```
jcfilben, line 19: Why were some of the tests changed?

src/js/components/Grommet/__tests__/__snapshots__/Grommet-test.js.snap
```diff
@@ -22,19 +22,19 @@ exports[`Grommet announce 1`] = `
...
 exports[`Grommet announce 2`] = `
 <div
+  aria-atomic="true"
   aria-live="assertive"
   id="grommet-announcer"
-  style="left: -100%; right: 100%; position: fixed; z-index: -1;"
->
-  hello
-</div>
+  style="position: absolute; left: -1px; height: 1px; width: 1px; overflow: hidden; clip-path: rect(1px, 1px, 1px, 1px); white-space: nowrap;"
```
jcfilben: What was broken with the previous style?

britt6612: The previous style wasn't 'broken' in that it successfully moved the element off the visible screen. However, it wasn't the most robust or semantically clear way to create a 'screen reader only' element. 

`clip-path` and overflow: hidden creates a truly 'visually hidden' element that remains fully accessible, which is the recommended standard for screen-reader-only content.

https://www.a11yproject.com/posts/how-to-hide-content/

src/js/components/Data/__tests__/Data-test.tsx
britt6612: The test assertion changed because our announcer element is now correctly hidden including by testing tools. Previously, the older hiding styles could sometimes make the hidden announcer still "visible" to getAllByText, causing the test to look for two instances

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -16,18 +20,32 @@ const createAnnouncer = () => {
 
 export const AnnounceContext = React.createContext(
   (message, mode = 'polite', timeout = 500) => {
-    // we only create a new container if we don't have one already
-    // we create a separate node so that grommet does not set aria-hidden to it
     const announcer =
-      document.body.querySelector(`#grommet-announcer[aria-live]`) ||
-      createAnnouncer();
+      document.body.querySelector('#grommet-announcer') || createAnnouncer();
```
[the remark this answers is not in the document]
britt6612: fixed

src/js/components/Data/__tests__/Data-test.tsx
jcfilben: Thanks for the explanation!

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -16,18 +20,32 @@ const createAnnouncer = () => {
...
-    }, timeout);
+      announcer.textContent = message; // Set the new message
+
+      if (timeout > 0) {
+        announcer.timeoutId = setTimeout(() => {
+          announcer.textContent = ''; // Clear the message after timeout
+          delete announcer.timeoutId;
+        }, timeout);
+      }
+    }, 75); // Increased delay for better reliability
```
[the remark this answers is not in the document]
taysea: @britt6612 How was this number chosen?

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -16,18 +20,32 @@ const createAnnouncer = () => {
...
+
+    announcer.textContent = '';
+
+    // 2. Use a short delay before setting the new message.
+    // This allows screen readers to register the empty state.
+    // A 50-100ms delay is often more reliable than 10ms for this.
     setTimeout(() => {
-      announcer.innerHTML = '';
-    }, timeout);
+      announcer.textContent = message; // Set the new message
```
taysea: Why do we do textContent instead of innerHTML?

britt6612:
closing in favor of #7701

PR #7682, opened 2025-06-12
List -- Fix keyboard behavior with roving tab index

### LINKED ISSUE #7648: DataTableColumns WCAG - 2.4.7 focus visible

None

### Expected Behavior

1. need focus indicators on the arrows rather than the list itself (in the re-order mode)

- [ ] update rule 2.4.7 on DataTableColumns to passed when PR complete 

### Actual Behavior

not passing 

_No response_

### Steps to Reproduce

_No response_

### Your Environment

_No response_

taysea:
#### What does this PR do?

This PR improves the List interactive behavior to be more closely aligned to [WCAG APG best practices for keyboard vs mouse interactions](https://www.w3.org/WAI/ARIA/apg/patterns/listbox/examples/listbox-rearrangeable/), but implements the approach using roving tabindex instead of aria-activedescendant.

Expected `onOrder` behavior:
- List should have a singe tab stop and arrow keys should be used to navigate between the "arrow" move up/move down buttons. (This is in alignment with existing behavior).
- When entering the list for the first time, the first "move down" button should have tabIndex 0 and should receive focus. (NEW behavior in alignment with WCAP APG pattern).
- Once arrow keys have been moved to navigate among the "arrow" buttons, if tab is used to leave the list, then tab is used to return to the list, the previously active "arrow" button should have tabIndex 0 and receive focus. (This is existing behavior in alignment with WCAG APG).
- Pressing Enter or Space should click the "arrow" button and re-order the data. The focus should "move with" the re-ordered item. (This is in alignment with existing behavior).
- Using mouse to hover over the "arrow" buttons should not update the active "arrow" button. Instead, mouse / keyboard should operate independently. (NEW behavior in alignment with WCAG APG. The current behavior is that mousing over an "arrow" button _will_ update the active "arrow" button, but this differs from the WCAG APG pattern.)
- Using mouse to click on an "arrow" button should re-order the data. The focus, in this case, doesn't "move with" the re-ordered item but rather remains where the mouse is (This is in alignment with existing behavior).

Expected `onClickItem` behavior:
- List should have a singe tab stop and arrow keys should be used to navigate between the items. (This is in alignment with existing behavior).
- When entering the list for the first time, the first item should have tabIndex 0 and should receive focus. (NEW behavior in alignment with WCAP APG pattern).
- Once arrow keys have been moved to navigate among items, if tab is used to leave the list, then tab is used to return to the list, the previously active item should have tabIndex 0 and receive focus. (This is existing behavior in alignment with WCAG APG).
- Pressing Enter or Space should trigger `onClickItem`. (This is in alignment with existing behavior).
- Using mouse to hover over an item should not update the active item. Instead, mouse / keyboard should operate independently. (NEW behavior in alignment with WCAG APG. The current behavior is that mousing over an item _will_ update the active item, but this differs from the WCAG APG pattern.)
- Using mouse to click on an item should trigger `onClickItem`. (This is in alignment with existing behavior).

Key differences to internal logic of "active" and `onActive`:
- Mousing over an item or an "arrow" button no longer triggers a change to the "active" DOM element. I believe this is an improvement because "hover" is not the same as focus. That said, this will affect the amount of times that `onActive` is called to purely be when keyboard is used to navigate among items / "arrow" buttons or when an item / "arrow" button is clicked. Happy to have any discussion on concerns to this change in behavior.

Adjustments to Box:
- Enhanced Box to allow `onSpace` to also activate its click event as this is native browser behavior for clickable elements like Button.
- Enhanced Box so `unfocusStyles` is called if a click event is triggered by a mouse instead of keyboard (similar to logic in other interactive components).

TO DO:
- [ ] When `onOrder`, clicking anywhere on a list item except for the "arrow" buttons should not trigger the first list item's "Down" button to be active.
- [ ] If people are comfortable with the changes to the interactive behavior/logic, tests need to be updated.

List.js

Locally in List stories such as Order, Children, OnClickItem.

Deploy link in List stories such as Order, Children, OnClickItem.

Closes #7648 

No.

Yes.

### DIFF
--- a/src/js/components/Box/Box.js
+++ b/src/js/components/Box/Box.js
@@ -309,7 +309,11 @@ const Box = forwardRef(
     }
 
     if (onClick) {
-      content = <Keyboard onEnter={onClick}>{content}</Keyboard>;
+      content = (
+        <Keyboard onEnter={onClick} onSpace={onClick}>
+          {content}
+        </Keyboard>
+      );
     }
 
     return content;
--- a/src/js/components/Box/Box.js
+++ b/src/js/components/Box/Box.js
@@ -310,7 +310,13 @@ const Box = forwardRef(
 
     if (onClick) {
       content = (
-        <Keyboard onEnter={onClick} onSpace={onClick}>
+        <Keyboard
+          onEnter={onClick}
+          onSpace={(e) => {
+            e.preventDefault(); // prevent page scroll
+            onClick(e);
+          }}
+        >
           {content}
         </Keyboard>
       );
--- a/src/js/components/Box/StyledBox.js
+++ b/src/js/components/Box/StyledBox.js
@@ -346,6 +346,8 @@ const StyledBox = styled.div.withConfig({
     props.onClick &&
     props.focus &&
     props.focusIndicator !== false &&
+    // only show focus styles when using keyboard navigation
+    // but not with mouse
     css`
       ${focusStyle()}
       &:focus:not(:focus-visible) {
--- a/src/js/components/Box/StyledBox.js
+++ b/src/js/components/Box/StyledBox.js
@@ -11,6 +11,7 @@ import {
   elevationStyle,
   fillStyle,
   focusStyle,
+  unfocusStyle,
   genericStyles,
   getBreakpointStyle,
   getHoverIndicatorStyle,
@@ -345,7 +346,12 @@ const StyledBox = styled.div.withConfig({
     props.onClick &&
     props.focus &&
     props.focusIndicator !== false &&
-    focusStyle()}
+    css`
+      ${focusStyle()}
+      &:focus:not(:focus-visible) {
+        ${unfocusStyle()}
+      }
+    `}
   ${(props) => props.theme.box && props.theme.box.extend}
   ${(props) => props.kindProp && props.kindProp.extend}
   ${(props) => (props.responsiveContainer ? responsiveContainerStyle : '')}
--- a/src/js/components/List/List.js
+++ b/src/js/components/List/List.js
@@ -18,11 +18,9 @@ import { MessageContext } from '../../contexts/MessageContext';
 import { Pagination } from '../Pagination';
 import { Text } from '../Text';
 import {
-  // focusStyle,
   genericStyles,
   normalizeColor,
   normalizeShow,
-  // unfocusStyle,
   useForwardedRef,
   usePagination,
   styledComponentsConfig,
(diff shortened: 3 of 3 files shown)


src/js/components/Box/Box.js
```diff
@@ -309,7 +309,11 @@ const Box = forwardRef(
     }
 
     if (onClick) {
-      content = <Keyboard onEnter={onClick}>{content}</Keyboard>;
+      content = (
+        <Keyboard onEnter={onClick} onSpace={onClick}>
```
britt6612: agree with this!

britt6612:
One other thing I noticed was when you tab to the `onClick` story the native focus gets placed on the first option then when you arrow down it puts the focus style that we have for hpe on the first one then after that you can arrow down to the other options

https://github.com/user-attachments/assets/46e15049-16d7-43e5-b5f7-1fd563b6dc67

britt6612:
https://github.com/user-attachments/assets/9e2b3b35-9edd-4ff7-abb5-2db7f220b46c

src/js/components/List/List.js
```diff
@@ -534,40 +499,37 @@ const List = React.forwardRef(
                 if (onClickItem && !onOrder) {
                   clickProps = {
                     role: 'option',
-                    tabIndex: -1,
+                    tabIndex:
```
britt6612, line 516: think this is where Im seeing the issue of the native focus coming through

src/js/components/Box/StyledBox.js
```diff
@@ -345,7 +346,12 @@ const StyledBox = styled.div.withConfig({
     props.onClick &&
     props.focus &&
     props.focusIndicator !== false &&
-    focusStyle()}
```
britt6612, line 348: maybe just add comment

src/js/components/List/List.js
```diff
@@ -617,68 +579,84 @@ const List = React.forwardRef(
 
                   const Up = theme.list.icons.up;
                   const Down = theme.list.icons.down;
+                  const moveUpIndex = orderableIndex * 2;
```
britt6612, line 597: think we could clean this up a little to make it easier to read through

britt6612:
Overall I think this is the right approach and aligns more with the APG. I think there is some potential clean up and then the issues you mentioned in your description as well as I left a couple comments. If it helps to id be happy to clean up tests sometimes nice to have another eyes on tests that did not implement the code!

src/js/components/List/List.js
taysea: Thank you -- I will look into this.

src/js/components/Box/StyledBox.js
taysea: Done.

src/js/components/List/List.js
taysea: Fixed in latest commit.

src/js/components/List/List.js
```diff
@@ -534,40 +499,37 @@ const List = React.forwardRef(
...
+                        lastActive === undefined &&
+                        index === 0) ||
+                      // returning to list after already using keyboard
+                      (active === undefined &&
+                        lastActive !== undefined &&
+                        lastActive === index) ||
+                      // actively using keyboard
+                      active === index
+                        ? 0
+                        : -1,
```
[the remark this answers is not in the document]
taysea: Done.

src/js/components/List/List.js
taysea: Done. Pulled logic out into helper function.

taysea:
Closing in favor of #7684

PR #7683, opened 2025-06-17
Fix toast notification focus bug

### LINKED ISSUE #7115: Toast Notification focuses original element after disappearing
Changing focus after toast notification disappears is disruptive.

### Expected Behavior

Don't change focus when toast notification disappears, or make it configurable

### Actual Behavior

See steps to reproduce

See steps to reproduce

### Steps to Reproduce

- https://v2.grommet.io/notification
- click "Open Toast Notification"
- click empty space
- scroll away, read docs
- after 8 seconds notification disappears and page scrolls back up to "Open Toast Notification" https://github.com/grommet/grommet/blob/9aee30310819f75e770857d213ac164483996ea7/src/js/components/Layer/Layer.js#L31 

### Your Environment

- Grommet version: 2.34.2
- Browser Name and version:

Sulaymon333:
#### What does this PR do?
- Fix toast notification issue which focuses after original element after disappearing 
- closes #7115 

- Go to Toast with Content stoty
- click `Show Notification`
- click empty space
- scroll down the page
- after 8 seconds notification disappears and page **DOES NOT** scroll back up to "Show Notification" button. it stays where your current page scroll is.

yes


Sulaymon333:
The current solution explores checking if modal prop is set to `false` or not before restoring focus onto the element that was last in focus before the toast notification opened. But I am checking if this disrupt other usage of layer for example in `SkipLink`. but so far. I have not seen any issue.

Another possible solution is introducing a new prop like `restoreFocus` for Layer which consumer can explicitly opt into.

Sulaymon333:
The current solution explores checking if modal prop is set to `false` or not before restoring focus onto the element that was last in focus before the toast notification opened. But I am checking if this disrupt other usage of layer for example in `SkipLink`. but so far. I have not seen any issue.

Another possible solution is introducing a new prop like `restoreFocus` for Layer which consumer can explicitly opt into.

taysea:
This is feeling like a reasonable approach given the fact that non-modal dialogs means the rest of the page is still interactive.

Will wait to see what other devs think.

britt6612:
> The current solution explores checking if modal prop is set to `false` or not before restoring focus onto the element that was last in focus before the toast notification opened. But I am checking if this disrupt other usage of layer for example in `SkipLink`. but so far. I have not seen any issue.

@Sulaymon333 I would also agree with this approach I do think if you can double check the skip links that would be great! This would be a hard issue to write a test for and so we should definitely look at it at different angles. I think even using `yalc` to test this change in DS site would benefit. I do think using the `modal` would make more sense then introducing a new prop like  `restoreFocus`

britt6612:
@MikeKingdom would you have any objections here?

Sulaymon333:
I tested the skipLinks story with this changes and no issues found. But yes, @britt6612 I can double test again with DS site as well.

MikeKingdom:
> @MikeKingdom would you have any objections here?

No objections here. This seems to make sense. It definitely needs as much testing as possible given how core to Layer it is.

src/js/components/Notification/stories/TitleAndMessagWithContent.stories.js
```diff
@@ -0,0 +1,375 @@
+import React, { useState } from 'react';
+
+import { Notification } from 'grommet';
```
britt6612: after we do our testing I think we can remove the story

Sulaymon333:
I have a test done with DS site locally via yalc and I did not notice any regression

[Regression testing video](https://hpe-my.sharepoint.com/:v:/p/sulaymon_tajudeen/EbCmeBkifElHufK0BCfR_uUBwtaIvAtZLUyC-wNe1pvrog?e=9b2S8B&nav=eyJwbGF5YmFja09wdGlvbnMiOnt9LCJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJTdHJlYW1XZWJBcHAiLCJyZWZlcnJhbE1vZGUiOiJtaXMiLCJyZWZlcnJhbFZpZXciOiJwb3N0cm9sbC1jb3B5bGluayIsInJlZmVycmFsUGxheWJhY2tTZXNzaW9uSWQiOiIxYzZmOWJmNi01ZDc1LTQxM2EtYTkyMS0yYTZhZTBlZjdmZTcifX0%3D)

src/js/components/Notification/stories/TitleAndMessagWithContent.stories.js
Sulaymon333: the story is now removed and the PR is now ready to be merged

Sulaymon333:
Good catch @jcfilben. 

I am having a closer look.

Sulaymon333:
@jcfilben after having a closer look at the point you raised, few question comes to mind. How does it apply to the case of a toast that serves as alert in this case since according to [w3.org Authoring Practice Guide (APG)](https://www.w3.org/WAI/ARIA/apg/patterns/alert/#:~:text=they%20do%20not%20affect%20keyboard%20focus), Alert do not affect keyboard focus. Are we treating the situation you raised as an exception?
 
image

Sulaymon333:
Some updates on addressing [Jessica comment](https://github.com/grommet/grommet/pull/7683#pullrequestreview-2951228518)
There is a DRAFT PR for that for us to explore - https://github.com/grommet/grommet/pull/7696. I have not notice any major issue yet.

jcfilben:
> @jcfilben after having a closer look at the point you raised, few question comes to mind. How does it apply to the case of a toast that serves as alert in this case since according to [w3.org Authoring Practice Guide (APG)](https://www.w3.org/WAI/ARIA/apg/patterns/alert/#:~:text=they%20do%20not%20affect%20keyboard%20focus), Alert do not affect keyboard focus. Are we treating the situation you raised as an exception?

We shouldn't use toast as an alert since toast shouldn't be used to convey critical information. However similar to alert I agree that the toast shouldn't affect keyboard focus in unexpected ways such as moving keyboard focus when the toast is opened. If the toast contains keyboard focus and the toast closes then it still feels reasonable to move the focus back to the trigger. This doesn't feel unexpected and matches the behavior of drops and layers.

Sulaymon333:
Thank you for further details on similar behavior with drops and layer.

There is already a DRAFT PR for when the toast contains keyboard focus and the toast closes then it moves the focus back to the trigger - https://github.com/grommet/grommet/pull/7696. I have not noticed any major issue yet. Please have a look when you have time.

Sulaymon333:
[UPDATE] @jcfilben the PR is now updated based on your last comment. 
Nowe, we restore focus to trigger if within a toast before closing - both automatically and manually(https://github.com/grommet/grommet/pull/7683/commits/ca98dbca592a6961ea990952ea54caa31c293045)
No issue noticed so far.

jcfilben (commented):
If I move focus to a different element outside of the toast notification and then the toast closes it moves focus back to the button that triggers the toast notification. I would expect the focus to remain where it is unless the focus was within the toast notification

https://github.com/user-attachments/assets/94dd5eeb-553f-4c0a-9f1c-688373240a9f

Sulaymon333:
@jcfilben `focusout` event handler is now added to only restore focus to trigger if within the toast notification otherwise remain where it is.

https://github.com/user-attachments/assets/5d3dafe0-6234-49d4-be8d-542c9a1b89a0

Sulaymon333:
add comment to `shouldRestoreFocus` logic

src/js/components/Layer/Layer.js
```diff
@@ -74,12 +118,23 @@ const Layer = forwardRef((props, ref) => {
...
 
   return layerContainer
-    ? createPortal(<LayerContainer ref={ref} {...props} />, layerContainer)
+    ? createPortal(
+        <Keyboard
+          target="document"
+          onKeyDown={() => {
+            lastInputWasKeyboardRef.current = true;
+          }}
+        >
```
jcfilben: It seems once we trigger an `onKeyDown` `lastInputWasKeyboardRef` is set to `true` and it never gets set back to `false`. This behavior doesn't seem to align with the name of the variable. @Sulaymon333 can you help me understand what the intent here was?

Sulaymon333: It was an oversight that I did not restore the variable back while debugging the keyboard target. It is now fixed within the `mousedown` effect.

The intention around the `lastInputWasKeyboard` variable was to track the nature of input (keyboard vs mouse) in order to  decide how to restore the focus appropriately. 

See attached the[ explainer video](https://hpe-my.sharepoint.com/:v:/r/personal/sulaymon_tajudeen_hpe_com/Documents/layout%20component%20fix%20explainer%20video.webm?csf=1&web=1&e=40pLCj&nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJTdHJlYW1XZWJBcHAiLCJyZWZlcnJhbFZpZXciOiJTaGFyZURpYWxvZy1MaW5rIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXcifX0%3D) for more details.

[layout component fix explainer video.webm](https://github.com/user-attachments/assets/65237808-4b04-4799-8aa8-af9a2f6b91ac)

jcfilben: Thanks for the explanation, what's weird is if I remove the `lastInputWasKeyboardRef` it seems like everything is working correctly. Is there a test case I'm missing?

Sulaymon333: Yes, it does work without `lastInputWasKeyboardRef` when I checked as well but there is a very little difference without it. Based on my discovery, I recommend that we still keep it. I discovered a little bug which I addressed as well - resetting the focusWithinLayer onMouseDown.
 
image

I explain better in the [video](https://hpe-my.sharepoint.com/:v:/r/personal/sulaymon_tajudeen_hpe_com/Documents/Keyboard%20input%20tracking%20for%20layer%20component.webm?csf=1&web=1&e=ILJTis&nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJTdHJlYW1XZWJBcHAiLCJyZWZlcnJhbFZpZXciOiJTaGFyZURpYWxvZy1MaW5rIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXcifX0%3D) attached.

[Keyboard input tracking for layer component.webm](https://github.com/user-attachments/assets/8d2e757c-0a9c-43f5-93d8-80861848c24f)

jcfilben: Thanks for the video and explanation!
I think the behavior that you highlighted as a bug is what I would think should be the expected behavior.
If the notification is opened via keyboard and then focus is moved into the notification and then we click inside of the notification, when the notification closes I would expect focus to return to the button that opened the notification. If we are using mouse, the focusIndicator won't be displayed but the button should still have focus so that if we start navigating via keyboard we don't start at the top of the page but rather at the button that opened the notification.

Sulaymon333: @jcfilben I have removed the usage of lastInputWasKeyboardRef to track keyboard input. This is now addressed in this latest [commit](https://github.com/grommet/grommet/pull/7683/commits/9c9996ee21e474e6ab009e950c35de037a856021)

src/js/components/Layer/Layer.js
```diff
@@ -74,12 +113,18 @@ const Layer = forwardRef((props, ref) => {
...
       layerContainer,
+      modal,
       originalFocusedElement,
     ],
   );
 
   return layerContainer
-    ? createPortal(<LayerContainer ref={ref} {...props} />, layerContainer)
+    ? createPortal(
+        <Keyboard target="document">
```
jcfilben: I don't think we need the Keyboard wrapper anymore

jcfilben (approved):
Looks good!

Sulaymon333:
@MikeKingdom please can you review this?

MikeKingdom (approved):

Issue #7657, opened 2025-05-27
DataTable - WCAG Announcements 1.3.1 & 1.3.3


britt6612:
### Is this something you or your team is interested in contributing a pull request towards?

yes

### Expected Behavior

- rows added or removed due to "search" control or infinite scroll aren't announced.
- Rows added by group expand and/or row details expand aren't announced.

### Actual Behavior
- no announcement for these changes for the user to know 
not passing wcag

britt6612:
Once this is closed as well as https://github.com/grommet/grommet/issues/7625 then go back to DS site to mark the rules as passed.

PR #7686, opened 2025-06-23
Add aria-expanded to ExpanderCell in DataTable

taysea:
#### What does this PR do?

Adds `aria-expanded` for more clarity on ExpanderCell in cases of `groupBy` and `rowDetail`. Announces "expanded" or "collapsed" once user interacts for more confirmation of state.

src/js/components/DataTable/ExpanderCell.js

Locally.

Related to #7657 

No.

Yes.

Backwards compatible.

### DIFF
--- a/src/js/components/DataTable/ExpanderCell.js
+++ b/src/js/components/DataTable/ExpanderCell.js
@@ -38,6 +38,7 @@ const ExpanderControl = ({ context, expanded, onToggle, pad, ...rest }) => {
     content = (
       <Button
         fill
+        aria-expanded={expanded ? 'true' : 'false'}
         a11yTitle={expanded ? 'collapse' : 'expand'}
         hoverIndicator
         onClick={onToggle}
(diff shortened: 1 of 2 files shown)


britt6612 (approved):
This part makes sense and comes through for me!

jcfilben (commented):
I'm wondering if we could add more context to what is being expanded/collapsed. For example in the expand collapse control in the header maybe we could make it more clear that this is expanding/collapsing all rows in the datatable. For the groups in the datatable maybe we could use the property that the rowsare being grouped by to add more description? For example 'expand Fort Collins rows' or something like that

taysea:
> I'm wondering if we could add more context to what is being expanded/collapsed. For example in the expand collapse control in the header maybe we could make it more clear that this is expanding/collapsing all rows in the datatable. For the groups in the datatable maybe we could use the property that the rowsare being grouped by to add more description? For example 'expand Fort Collins rows' or something like that

Based on some of the discussions on this PR: https://github.com/grommet/grommet/pull/7480 and offline discussions, there's a variety of ways the primaryValue might be defined (and not rendered to the screen)-- so trying to make too many assumptions around what should be part of the aria-label probably isn't worth it. We'll leave this PR as if and in the future we can consider enhancing datatable to allow caller to pass in what property they want as the aria-label, but for now we won't handle that.

jcfilben (approved):
Looks good!

PR #7688, opened 2025-06-24
DataTable - Improve search and sort aria support

### LINKED ISSUE #7625: DataTable - WCAG
Currently in DataTable the Header sort buttons and search are failing the WCAG 1.1.1 

- These Header sort buttons and search need better a11y text or a way for user to pass something to them.

### LINKED ISSUE #7627: DataTable - WCAG 4.1.2
4.1.2 Name, Role, Value - Add `aria-sort` for ascending/descending this indicates if items in a table or grid are sorted in ascending or descending order.

taysea:
#### What does this PR do?
Introduce messages to better handle aria-labels/aria-sort for DataTable.

NOTE: aria-sort is not read out by [VoiceOver](https://discussions.apple.com/thread/250523941), but does have JAWS support. We can confirm the behavior with Bill if desired by I think looking at the DOM is enough to confirm alignment with [WCAG APG DOM ouput](https://www.w3.org/WAI/ARIA/apg/patterns/grid/examples/data-grids/).

Locally in storybook.

Storybook.

Closes #7625, Closes #7627

Yes.

Yes.

Backwards compatible.


src/js/components/DataTable/Header.js
```diff
@@ -378,7 +392,7 @@ const Header = forwardRef(
                       justify={align}
                     >
                       {content}
-                      {Icon && <Icon />}
+                      {Icon && <Icon a11yTitle={ariaSort} />}
```
britt6612: instead of `a11yTitle` can we use `aria-label` ?

taysea: a11yTitle is supported on icons, but aria-label doesn't get passed through: https://github.com/grommet/grommet-icons/blob/master/src/js/icons/index.d.ts#L4

taysea: Done.

britt6612 (approved):
LGTM

jcfilben (commented):
I think the aria-sort property should always be one of the following options: `ascending`, `descending`, `none` or `other`. If someone wants to change the aria-label of the sort for internationalization purposes or to give a different description then this change would also affect the aria-sort property, likely setting it to an invalid option. It seems like the aria-label and aria-sort shouldn't be set to the same variable

taysea:
> I think the aria-sort property should always be one of the following options: `ascending`, `descending`, `none` or `other`. If someone wants to change the aria-label of the sort for internationalization purposes or to give a different description then this change would also affect the aria-sort property, likely setting it to an invalid option. It seems like the aria-label and aria-sort shouldn't be set to the same variable

You're right, good call. Fixed in latest commit.

jcfilben (approved):
Looks good!

PR #7690, opened 2025-06-24
Fix incorrect formatting of video time

### LINKED ISSUE #7689: Video component time display is a minute too high half the time
The current/total time display on the right side of the controls accidentally is rounded up to the next minute when the number of seconds is 30-59. This makes it look like the time is jumping up and then back down while playing or scrubbing. When not playing, it displays the total duration which is 1 min greater than it should be if the number of seconds is between 30 and 59.

Reported by GLP. Here's a comparison of the same video in the Grommet Video component vs the stock `<video>` tag:

Image

The reason is Video.js line 36 rounds the minutes value instead of truncates like it should (since it's displaying seconds also)

```jsx
const formatTime = (time) => {
  let minutes = Math.round(time / 60);
  if (minutes < 10) {
    minutes = `0${minutes}`;
  }
  let seconds = Math.round(time) % 60;
  if (seconds < 10) {
    seconds = `0${seconds}`;
  }
  return `${minutes}:${seconds}`;
};
```
Should instead be:
```jsx
const formatTime = (time) => {
  let minutes = Math.trunc(time / 60);
  if (minutes < 10) {
    minutes = `0${minutes}`;
  }
  let seconds = Math.round(time) % 60;
  if (seconds < 10) {
    seconds = `0${seconds}`;
  }
  return `${minutes}:${seconds}`;
};
```

MikeKingdom:
#### What does this PR do?
This fixes the formatting of the current/total minutes display to truncate the number of minutes rather than accidentally rounding it (it shouldn't round because it displays the seconds too.)

You can see the issue with [this story](https://storybook.grommet.io/?path=/story/media-video-theater-mode--customizable-controls&globals=theme:hpe)

It's easier to see with a longer video like https://pavo.common.cloud.hpe.com/whatsnew/9731e661-5b03-4763-9fd9-2280b6ac87fb/whats-new-og-iam-rough-draft-80percent.mp4. If you hover along the scrubber line you can see at any intervals between 30-59 seconds it makes the minute digit too high

Fixes #7689 

Issue:
image

No

Yes

Backwards compatible

### DIFF
--- a/src/js/components/Video/Video.js
+++ b/src/js/components/Video/Video.js
@@ -33,7 +33,7 @@ import { useThemeValue } from '../../utils/useThemeValue';
 const VOLUME_STEP = 0.166667;
 
 const formatTime = (time) => {
-  let minutes = Math.round(time / 60);
+  let minutes = Math.trunc(time / 60);
   if (minutes < 10) {
     minutes = `0${minutes}`;
   }


jcfilben (approved):
Looks good!

PR #7692, opened 2025-06-25
DataTable - Update to use roving tabindex for `onClickRow`

### LINKED ISSUE #7629: DataTable - WCAG 1.4.11
The visual presentation of the following have a contrast ratio of at least 3:1 against adjacent color(s)

The following story we have the rows selectable and are using the rgba (0, 0, 0, 0.04) which is not passing the ratio of 3:1

Image

https://storybook.grommet.io/?path=/story/visualizations-datatable-clickable--clickable-data-table&globals=theme:hpe

taysea:
#### What does this PR do?

Updates DataTable to use focus on individual row when `onClickRow` rather than just "active" styling which may or may not meet contrast ratios (grommet / HPE theme don't for example).

Implements roving tab index similar to approach in recent List enhancements: https://github.com/grommet/grommet/pull/7684

Now, instead of table body having tabindex 0, an individual row has that and receives focus treatment on the row. To retain backwards compatible with previous meaning of "active" in DataTable, this required introducing a new state variable internal to List (focused, setFocused) to manage DOM focus (similar to List approach).

Local in storybook and reviewing jest changes.

https://github.com/user-attachments/assets/8026a863-9d02-4a73-9285-b91c8405688c

https://github.com/user-attachments/assets/fc02cb2f-f45b-4189-ab68-044f78f78b38

In snapshot changes, we can see the focus style now applied to `tr`:
Screenshot 2025-06-25 at 9 11 44 AM
Screenshot 2025-06-25 at 9 11 38 AM

Closes #7629 

No.

Yes.

Backwards compatible.


src/js/components/DataTable/__tests__/DataTable-test.tsx
```diff
@@ -2073,8 +2073,9 @@ describe('DataTable', () => {
 
     expect(container.firstChild).toMatchSnapshot();
 
-    fireEvent.mouseEnter(getByText('alpha'));
-    fireEvent.keyDown(getByText('alpha'), {
+    const row = getByText('alpha').closest('tr');
```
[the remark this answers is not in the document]
taysea: DOne.

src/js/components/DataTable/Body.js
```diff
@@ -22,32 +22,50 @@ const Row = memo(
...
     columns,
     pinnedOffset,
     primaryProperty,
     verticalAlign,
+    onRowRefChange,
   }) => (
     <>
       <StyledDataTableRow
-        ref={rowRef}
+        ref={(element) => {
```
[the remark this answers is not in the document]
taysea: Adjusted to account for this.

src/js/components/DataTable/StyledDataTable.js
```diff
@@ -74,6 +74,12 @@ const rowStyles = css`
 `;
 
 const StyledDataTableRow = styled(TableRow)`
+  &:focus {
+    ${focusStyle()}
+  }
+  &:focus:not(:focus-visible) {
```
britt6612, line 80: why do we need this part `:focus-visible`?

src/js/components/DataTable/Body.js
```diff
@@ -264,29 +324,50 @@ const Body = forwardRef(
...
+        }
         onDown={
-          onClickRow && data.length && active < data.length - 1
-            ? () => setActive((active ?? -1) + 1)
+          onClickRow && data.length && focused < data.length - 1
+            ? (event) => {
+                event.preventDefault();
+                const nextIndex = (focused ?? -1) + 1;
+                setFocused(nextIndex);
+                setActive(nextIndex);
```
britt6612, line 350: Good call keeping focused and active in sync

src/js/components/DataTable/Body.js
```diff
@@ -67,10 +92,29 @@ const Row = memo(
...
+          if (onClickRow && !isDisabled) {
+            setFocused(index);
+            setActive(index);
+          }
+        }}
+        onMouseOver={
           onClickRow && !isDisabled ? () => setActive(index) : undefined
         }
-        onMouseLeave={onClickRow ? () => setActive(undefined) : undefined}
+        tabIndex={
```
britt6612, line 104: this is more just suggestion for clean up if we pull out some of this logic

src/js/components/DataTable/Body.js
```diff
@@ -264,29 +324,50 @@ const Body = forwardRef(
...
+                  adjustedEvent.datum = data?.[focused];
                   onClickRow(adjustedEvent);
                 } else if (onClickRow === 'select') {
                   selectRow();
                 }
               }
             : undefined
         }
-        onUp={onClickRow && active ? () => setActive(active - 1) : undefined}
+        onUp={
```
britt6612, line 335: If focused is 0 (so the first row is focused), pressing the up arrow will set previousIndex to -1?
I wasnt sure about this but I did check locally if I was on the first row and pressed up arrow it stays on the first row but logic wise just wanted to make sure we don't get unintended behavior

britt6612:
Behavior looks good couple questions/comments on code

src/js/components/DataTable/StyledDataTable.js
taysea: This is so if someone is interacting with mouse, the focus indicator won't be applied (similar to what we do in Button/other interactive components)

britt6612: okay I didnt realize we do the `:focus-visible` as well thanks!

MikeKingdom (commented):
Look great.

There's a little oddness on the multiple pins story due to the user-define background for the pinned cells. We might want to test a little more with pinned, rowDetails or rowSelect in a grouped situation?

src/js/components/DataTable/Body.js
taysea: When I press up arrow from the first row, this won't get called because focused is already `0` (meaning the condition will resolve to false)

taysea:
> There's a little oddness on the multiple pins story due to the user-define background for the pinned cells. We might want to test a little more with pinned, rowDetails or rowSelect in a grouped situation?

Good call, looking at the multiple pins story, it seems like it's already breaking our recommendation not to have `onClickRow` and `onSelect` together (unless `onClickRow="select"`). 

In the general case of pinned, it is a little awkward how the "pinned" row sits on top of the row with focus indicator but I'm not quite sure a way around this given the "pinned" column is a different DOM element. We could make a "fake" focus indicator on that pinned column row, but that feels a little hacky.

Current behavior
Screenshot 2025-06-27 at 9 47 53 AM

I think the `onClickRow="select"` + `onSelect` behavior looks good:

Screenshot 2025-06-27 at 9 48 53 AM

We also don't support `rowDetails` + `onClickRow` at the moment because this causes nested interactive elements, so I don't think we have to worry currently about that with current behavior. We could always enhance in future though.

@MikeKingdom

src/js/components/DataTable/Body.js
taysea: Was gonna pull this out but decided to leave as is to not clutter the return

src/js/components/DataTable/__tests__/DataTable-test.tsx
```diff
@@ -2073,8 +2073,9 @@ describe('DataTable', () => {
 
     expect(container.firstChild).toMatchSnapshot();
 
-    fireEvent.mouseEnter(getByText('alpha'));
-    fireEvent.keyDown(getByText('alpha'), {
+    const row = screen.getByText('alpha').closest('tr');
+    fireEvent.focus(row!);
```
jcfilben, line 2077: What does the `!` mark do here?

src/js/components/DataTable/Body.js
```diff
@@ -198,16 +242,22 @@ const Body = forwardRef(
...
     const [scroll, setScroll] = React.useState();
     const containerRef = useForwardedRef(ref);
 
-    // Determine if using a keyboard to cover focus behavior
-    const usingKeyboard = useKeyboard();
-
-    const onFocusActive =
-      active ?? lastActive ?? (usingKeyboard && onClickRow ? 0 : undefined);
+    // Store refs for each row to enable direct focus management
+    const rowRefs = React.useRef(new Map());
```
jcfilben, line 252: It feels a little heavy to create refs for all the rows, wondering if we are able to do something similar to roving tab index in Calendar where we just have a ref for the item that can receive focus

src/js/components/DataTable/Body.js
britt6612: ah okay thank you for explanation

src/js/components/DataTable/__tests__/DataTable-test.tsx
taysea: It's a shorthand to mean "not null" -- so it'll only try to focus if row exists.

src/js/components/DataTable/Body.js
taysea: I believe [each row already has a ref](https://github.com/grommet/grommet/blob/master/src/js/components/DataTable/Body.js#L43) -- we're just taking that and storing in a map to know what to focus. I can look into another approach where the focusableRow ref (instead of rowRefs) is just a single row and we only apply it in the case where tabIndex=0 if you'd prefer though.

jcfilben: Ah okay, in that case I'm good with leaving this as is

src/js/components/DataTable/Body.js
britt6612: thats fine

Issue #7325, opened 2024-09-27
CheckBox - Providing a label on a FormField and on its contained Checkbox leads to an orphaned label on the FormField


jcfilben:
### Expected Behavior
Each FormField should have 1 label.

### Actual Behavior
Providing a label on a FormField and on its contained Checkbox leads to an orphaned label on the FormField. Once of the labels is not associated with anything.

Potentially RadioButton, RadioButtonGroup and CheckBoxGroup also have this issue

### Steps to Reproduce

1. 2. 3.

### Your Environment

- Grommet version:
- Browser Name and version:
- Operating System and version (desktop or mobile):

NOMADE55:
I could be wrong on what's being needed here, but I see an old part of the code where it checks if the prop component is `Checkbox`, and if it is, prevents `FormField` from adding a label.

![image](https://github.com/user-attachments/assets/7d1d3a13-1bd2-4bed-ba8c-0c0a1d471913)

PR #7695, opened 2025-06-25
RadioButtonGroup and CheckBoxGroup - associate FormField label with group

jcfilben:
#### What does this PR do?
When RadioButtonGroup or CheckBoxGroup are within a FormField that has a label, the label is linked to the `role='group'` or `role='radiogroup'` div so that the label isn't orphaned and is correctly associated with the input and read out by the screenreader.

If a user passes in an `aria-label` to RadioButtonGroup or CheckBoxGroup, this will take precedence

Closes https://github.com/grommet/grommet/issues/7325

no
yes
backwards compatible


src/js/components/RadioButtonGroup/RadioButtonGroup.js
```diff
@@ -101,6 +104,11 @@ const RadioButtonGroup = forwardRef(
 
     const onBlur = () => setFocus(false);
 
+    let ariaLabelledBy;
```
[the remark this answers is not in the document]
jcfilben: I feel like this probably isn't worth pulling into a util function

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -9,12 +9,15 @@ import { useThemeValue } from '../../utils/useThemeValue';
 const CheckBoxGroup = forwardRef(
   (
     {
+      'aria-label': ariaLabelProp,
```
taysea, line 12: Can you explain why aria-label takes precedence?

src/js/components/FormField/FormField.js
```diff
@@ -647,7 +647,12 @@ const FormField = forwardRef(
         {(label && component !== CheckBox) || help ? (
           <>
             {label && component !== CheckBox && (
-              <Text as="label" htmlFor={htmlFor} {...labelStyle}>
+              <Text
+                as="label"
+                id={htmlFor ? `${htmlFor}__label` : undefined}
+                htmlFor={htmlFor}
```
taysea, line 653: Wondering about keying off of `htmlFor` here, since technically in the case of CheckBoxGroup/RadioButtonGroup/CheckBox/RadioButton which have their own label, `htmlFor` and `id` isn't actually necessary (only `name` is necessary for Form logic).

Would it make more sense to key off of `name`?

src/js/components/FormField/__tests__/__snapshots__/FormField-test.tsx.snap
```diff
@@ -1718,6 +1718,7 @@ exports[`FormField disabled with custom label for help and info 1`] = `
       <label
         class="c2"
         for="text-input"
+        id="text-input__label"
```
taysea: Wondering why this got edited. I would've only expected the checkbox/radio related components.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
jcfilben: I was thinking that if someone intentionally passes an aria-label then we should announce that instead of the FormField label

src/js/components/CheckBoxGroup/CheckBoxGroup.js
```diff
@@ -65,9 +68,16 @@ const CheckBoxGroup = forwardRef(
       }
     };
 
+    let ariaLabelledBy;
+    if (formContext.useFormField({}).inForm && id && !ariaLabelProp) {
```
taysea:
```suggestion
    if (formContext?.useFormField({})?.inForm && id && !ariaLabelProp) {
```

src/js/components/FormField/FormField.js
jcfilben: I feel like if the implementer doesn't specify `htmlFor` and `id` then they wouldn't expect the label to be linked. So if we used name here instead I think the resulting behavior would feel a little unexpected

src/js/components/FormField/__tests__/__snapshots__/FormField-test.tsx.snap
jcfilben: In the FormField code I just added an `id` to the label reguardless of what the child component was. It didn't seem harmful to add an id on all the FormField labels but let me know if you think we should only add it for FormFields with specific children.

As a side note Brittany is working on a PR where we will need an `id` on the FormField label for FormFields with Select or SelectMultiple.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
jcfilben: That makes sense, I'll update

src/js/components/FormField/FormField.js
taysea: I see -- okay I'm okay with this given the direction Brittany's PR is also heading.

To help ensure we avoid any unexpected ID collisions, any thoughts on prefixing with:

`grommet-${htmlFor}__label` OR `g-${htmlFor}__label`? Low likelihood that there'd be collisions from something someone implemented but just a thought.

src/js/components/FormField/__tests__/__snapshots__/FormField-test.tsx.snap
taysea: Gotcha, okay I'm comfortable with this.

src/js/components/RadioButtonGroup/RadioButtonGroup.js
```diff
@@ -126,7 +137,7 @@ const RadioButtonGroup = forwardRef(
             (
               {
                 disabled: optionDisabled,
-                id,
+                id: optionId,
```
taysea, line 140: I'm not seeing where optionId gets created -- can you clarify why this gets updated?

jcfilben: Since I pulled out `id` from rest I just needed to change the name here to `optionId` so that it didn't conflict with the existing `id` namespace

src/js/components/CheckBoxGroup/CheckBoxGroup.js
taysea: From this MDN docs -- aria-labelledby should take precedence: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-labelledby

"aria-labelledby takes precedence over all other methods of providing an accessible name, including aria-label, [<label>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label), and the element's inner text."

Can you confirm we're not overriding this default behavior?

src/js/components/RadioButtonGroup/RadioButtonGroup.js
taysea: Ah understood, misread and didn't realize this was destructuring.

src/js/components/CheckBoxGroup/CheckBoxGroup.js
jcfilben: Yes so they way I have the precedence set up to work is in the following order (ordered from highest to lowest)

- User passes `aria-labelledby`
- User passes `aria-label`
- User passes neither `aria-label` or `aria-labelledby` so we internally set `aria-labelledby` to the FormField label

src/js/components/FormField/FormField.js
jcfilben: I would be open to adding a prefix. One hesitation is that there are other places in grommet where we set an internal `id`. In hindsight it would have been nice to have all internal `id`s be prefixed. Let me know your thoughts on if it is okay to just do this for new internally `id`s we create or if we should try and go back and update our other internal `id`s

taysea: I don't think we need to go back and update old ones. So if you like the prefix id we can just do it here and moving forward

src/js/components/CheckBoxGroup/CheckBoxGroup.js
taysea: Ah perfect. That's great. Thanks for explaining!

src/js/components/FormField/FormField.js
jcfilben: Cool, just pushed up changes to prefix the id with `grommet-`

taysea (approved):
LGTM!

britt6612 (approved):
LGTM

PR #7698, opened 2025-06-30
RangeSelector - add hidden input for safari screen reader

### LINKED ISSUE #7027: RangeSelector - screenreader issues
### Expected Behavior
When navigating with a keyboard and using a screenreader the screenreader should read out the value of the RangeSelector as it changes.

### Actual Behavior
Instead the screenreader reads out the previous value when the RangeSelector changes.

https://storybook.grommet.io/?path=/story/input-rangeselector-label--label

### Steps to Reproduce

1. Go to the story above
2. Using a screenreader and keyboard change the value of the RangeSelector

### Your Environment

- Grommet version:
- Browser Name and version:

britt6612:
#### What does this PR do?
This PR adds a `hiddenInput` in which provides the native range semantics that Safari voice over needs in order to announce slider values. 
edgeControl.js
local stories 
local stories in Safari

Safari has voiceover limitation. Safari has specific requirements for recognizing slider controls. It need the native `<input type="range">` to properly announce. 

The input is marked as `readOnly` to prevent interference with existing interaction logic so this input does not trigger the `onChange` that could interfere with the existing `onIncrease` and `onDecrease` logic 

The `aria-label` is moved now to the hiddenInput since that is what is used for the screen reader experience. The aria roles were taken off from the edgeControl since that control is just primary now used for the mouse interactions. 
closes #7027 
no
yes
backwards compatible


src/js/components/RangeSelector/EdgeControl.js
```diff
@@ -93,6 +117,19 @@ const EdgeControl = forwardRef(
           justify="center"
           alignSelf="stretch"
         >
+          {/* Add visually hidden range input for Safari VoiceOver */}
+          <VisuallyHiddenInput
```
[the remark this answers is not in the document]
britt6612: good point here thinking if we use the same message from RangeSelector aria-label
@taysea @jcfilben thoughts?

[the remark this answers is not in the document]
taysea: I like the existing wording of "Lower bounds", "Upper bounds" -- can we stick with that? key off of `edge` for which one?

src/js/components/RangeSelector/EdgeControl.js
```diff
@@ -107,7 +144,7 @@ const EdgeControl = forwardRef(
               minHeight: size,
               zIndex: 1,
             }}
-            tabIndex={0}
+            tabIndex={-1} // Remove from tab order
             onFocus={() => setFocus(true)}
```
taysea: If this now has tabIndex -1, does it need onFocus/onBlur?

taysea: Also how is the focus styling still going around the edge control even though the input has focus? just making sure I understand

britt6612: yes we still want the `onFocus` `onBlur`

britt6612: The `hiddenInput` is what gains the browser focus, but since the `onFocus` is called the `focus` state becomes true since this is true the styles then are applied on the StyledBox so we visually see the focus styles. 

So the thing that changed it that the `hiddenInput` is the focus target but the user will still see the focus style on the visible handle because of the StyledBox.

jcfilben: So the `onFocus` still gets called even though the Box isn't in the tabindex?

jcfilben: Still a little confused I guess, why do we have `onFocus` and `onBlur` on both this Box and the `VisuallyHiddenInput`?

jcfilben (commented):
Tested in safari, chrome, and firefox and screenreader behavior looks good! Just have some questions on the code (see thread above)

src/js/components/RangeSelector/EdgeControl.js
britt6612: Sorry for the confusion — I was referring to the `StyledBox` that has the focus styling. But @taysea , you’re correct — we no longer need the `onFocus` and `onBlur` on the Box with tabIndex={-1} since it’s no longer the element that gains focus. The `onFocus` and `onBlur` now only on the VisuallyHiddenInput. The hidden input is what actually receives browser focus, and when that happens, we set the focus state to true, which gets passed as focus={true} to the StyledBox to apply the visible focus styles on the handle.

src/js/components/RangeSelector/EdgeControl.js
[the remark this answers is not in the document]
jcfilben: @britt6612 looks like the aria-label is still "slider control", do we want to go ahead and change it to "Lower bounds" or "Upper bounds". Also we should allow for internationalization here

src/js/components/RangeSelector/RangeSelector.js
```diff
@@ -331,16 +328,14 @@ const RangeSelector = forwardRef(
...
-          role="slider"
-          aria-valuenow={upper}
-          aria-valuemin={min}
-          aria-valuemax={max}
-          tabIndex={0}
           color={color}
           direction={direction}
           thickness={thickness}
           edge="upper"
+          min={min}
```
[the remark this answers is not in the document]
britt6612: Im only passing the `messages` prop once because in the edgeControl.js Im checking to see which message is lower and upper 

```
     aria-label={format({
              id:
                edge === 'lower'
                  ? 'rangeSelector.lower'
                  : 'rangeSelector.upper',
              messages,
            })}
```

src/js/components/RangeSelector/EdgeControl.js
```diff
@@ -93,6 +119,29 @@ const EdgeControl = forwardRef(
           justify="center"
           alignSelf="stretch"
         >
+          {/* Add visually hidden range input for Safari VoiceOver */}
+          <VisuallyHiddenInput
+            type="range"
+            role="slider"
```
jcfilben: Do we need both `type="range"` and `role="slider"`, from a quick google search it seems like one or the other should be used

britt6612: yes you are correct thank you removed `role="slider"` since we need the `type="range"` for Safari voice over!

jcfilben (approved):
Looks good!

PR #7701, opened 2025-07-02
Improve AnnounceContext 

britt6612:
#### What does this PR do?
Improves the reliability and accessibility of the AnnounceContext utility in Grommet, which is used to trigger screen reader announcements. 
The previous implementation had structural and timing issues that caused announcements to be missed or inconsistently read by screen readers.

**Key improvements:**
1. Replaces innerHTML with textContent to avoid parsing and ensure clean text announcements.

2. Renders a persistently hidden live region using visually hidden styles that are more widely recognized by screen readers.

3. Adds aria-atomic="true" to ensure the entire message is read.

4. Forces a DOM reflow (offsetWidth) and uses a short timeout before setting the message content, which helps screen readers recognize announcements—especially when using aria-live="polite".

AnnounceContext.js

#### What testing has been done?
Manual testing in Storybook using macOS VoiceOver

#### Any background context?
The previous implementation:

inserted a live region with z-index: -1, which some screen readers ignored

Used innerHTML instead of textContent, which is not recommended for screen readers

**Resources:**

https://css-tricks.com/inclusively-hidden/

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-atomic

https://stackoverflow.com/questions/75520027/using-a-separate-status-div-with-aria-live-assertive-to-handle-toast-message

https://github.com/nvaccess/nvda/issues/7996

So with the AnnounceContext the story is rendered inside an iframe in the storybook.
There are known issues with `aria-live` and iframes suck that can result in messages being read twice or inconsistently.

While `aria-live="polite"` is designed to announce dynamic content changes without interrupting the user's current task, issues can arise when implementing it within an iframe, especially with VoiceOver. 

https://stackoverflow.com/questions/61244988/aria-live-content-read-twice-by-mac-chrome-voiceover#:~:text=According%20to%20a%20bug%20report%2C%20VoiceOver%20duplicates,in%20iframe**%20Aria%2Dlive%20reads%20content%20only%20once

While testing this open a new window so that the example is rendered without iframe

#### Related issues?
Related to #7607

#### Do the Grommet docs need to be updated?
No

#### Should this PR be mentioned in release notes?
Yes

#### Is this change backwards compatible?
Yes

### DIFF
--- a/src/js/components/DataClearFilters/__tests__/DataClearFilters-test.tsx
+++ b/src/js/components/DataClearFilters/__tests__/DataClearFilters-test.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { render, screen, fireEvent } from '@testing-library/react';
+import '@testing-library/jest-dom';
 import 'jest-styled-components';
 import { Data, View } from '../../Data';
 import { Grommet } from '../../Grommet';
@@ -8,7 +9,7 @@ import { DataSummary } from '../../DataSummary';
 
 // asserts that AnnounceContext aria-live region and visible DataSummary each have this text
 const expectDataSummary = (message: string) =>
-  expect(screen.getAllByText(message)).toHaveLength(2);
+  expect(screen.getByText(message)).toBeInTheDocument();
 
 const data = [
   {
--- a/src/js/components/DataSummary/__tests__/DataSummary-test.tsx
+++ b/src/js/components/DataSummary/__tests__/DataSummary-test.tsx
@@ -1,5 +1,6 @@
 import React, { useState } from 'react';
 import { fireEvent, render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom';
 import 'jest-styled-components';
 
 import { Data } from '../../Data';
@@ -10,7 +11,7 @@ import { DataSummary } from '..';
 
 // asserts that AnnounceContext aria-live region and visible DataSummary each have this text
 const expectDataSummary = (message: string) =>
-  expect(screen.getAllByText(message)).toHaveLength(2);
+  expect(screen.getByText(message)).toBeInTheDocument();
 
 const data = [{ name: 'a' }, { name: 'b' }];
 
--- a/src/js/components/Grommet/__tests__/Grommet-test.js
+++ b/src/js/components/Grommet/__tests__/Grommet-test.js
@@ -1,6 +1,5 @@
 import React from 'react';
 import { render } from '@testing-library/react';
-import '@testing-library/jest-dom';
 import 'jest-styled-components';
 
 import { hpe as hpeTheme } from 'grommet-theme-hpe';
@@ -62,9 +61,7 @@ describe('Grommet', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
-  test('announce', () => {
-    jest.useFakeTimers();
-
+  test('announce', (done) => {
     const { container } = render(
       <Grommet>
         <AnnounceContext.Consumer>
@@ -75,26 +72,18 @@ describe('Grommet', () => {
 
     expect(container.firstChild).toMatchSnapshot();
 
-    // Check that announcer element is created
-    const announcer = document.body.querySelector(
-      '#grommet-announcer[aria-live]',
-    );
-    expect(announcer).toBeInTheDocument();
-    expect(announcer).toMatchSnapshot();
+    // no style, no need for expectPortal
+    expect(
+      document.body.querySelector('#grommet-announcer[aria-live]'),
+    ).toMatchSnapshot();
 
-    jest.advanceTimersByTime(100);
-
-    // The message should now be announced
-    expect(announcer.textContent).toBe('hello');
-    expect(announcer.getAttribute('aria-live')).toBe('assertive');
-
-    // Fast-forward past the timeout
-    jest.advanceTimersByTime(500);
-
-    // The announcer should be cleared
-    expect(announcer.textContent).toBe('');
-
-    jest.useRealTimers();
+    setTimeout(() => {
+      // should clear the aria-live container
+      expect(
+        document.body.querySelector('#grommet-announcer[aria-live]'),
+      ).toMatchSnapshot();
+      done();
+    }, 600); // wait the aria-live container to clear
   });
 
   test('messages', () => {
--- a/src/js/components/Grommet/__tests__/Grommet-test.js
+++ b/src/js/components/Grommet/__tests__/Grommet-test.js
@@ -61,7 +61,9 @@ describe('Grommet', () => {
     expect(container.firstChild).toMatchSnapshot();
   });
 
-  test('announce', (done) => {
+  test('announce', () => {
+    jest.useFakeTimers();
+
     const { container } = render(
       <Grommet>
         <AnnounceContext.Consumer>
@@ -72,18 +74,25 @@ describe('Grommet', () => {
 
     expect(container.firstChild).toMatchSnapshot();
 
-    // no style, no need for expectPortal
-    expect(
-      document.body.querySelector('#grommet-announcer[aria-live]'),
-    ).toMatchSnapshot();
+    // Test announcer element properties
+    const announcer = document.body.querySelector(
+      '#grommet-announcer[aria-live]',
+    );
+    expect(announcer).toBeTruthy();
+    expect(announcer.getAttribute('aria-live')).toBe('assertive');
+    expect(announcer.getAttribute('aria-atomic')).toBe('true');
+    expect(announcer.id).toBe('grommet-announcer');
 
-    setTimeout(() => {
-      // should clear the aria-live container
-      expect(
-        document.body.querySelector('#grommet-announcer[aria-live]'),
-      ).toMatchSnapshot();
-      done();
-    }, 600); // wait the aria-live container to clear
+    jest.advanceTimersByTime(150);
+
+    // Now check that the message is announced
+    expect(announcer.textContent).toBe('hello');
+    jest.advanceTimersByTime(600);
+
+    // Check that content is cleared
+    expect(announcer.textContent).toBe('');
+
+    jest.useRealTimers();
   });
 
   test('messages', () => {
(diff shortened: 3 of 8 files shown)


src/js/components/Grommet/__tests__/__snapshots__/Grommet-test.js.snap
```diff
@@ -22,20 +22,26 @@ exports[`Grommet announce 1`] = `
 
 exports[`Grommet announce 2`] = `
 <div
+  aria-atomic="true"
   aria-live="assertive"
+  data-initial-announce-timeout-id="38"
   id="grommet-announcer"
-  style="left: -100%; right: 100%; position: fixed; z-index: -1;"
+  style="position: absolute; left: -9999px; height: 1px; width: 1px; overflow: hidden;"
```
britt6612: The previous style wasn't 'broken' in that it successfully moved the element off the visible screen. However, it wasn't the most robust or semantically clear way to create a 'screen reader only' element.

clip-path and overflow: hidden creates a truly 'visually hidden' element that remains fully accessible, which is the recommended standard for screen-reader-only content.

https://www.a11yproject.com/posts/how-to-hide-content/

src/js/components/DataSummary/__tests__/DataSummary-test.tsx
```diff
@@ -10,7 +11,7 @@ import { DataSummary } from '..';
 
 // asserts that AnnounceContext aria-live region and visible DataSummary each have this text
 const expectDataSummary = (message: string) =>
-  expect(screen.getAllByText(message)).toHaveLength(2);
+  expect(screen.getByText(message)).toBeInTheDocument();
```
britt6612, line 14: The test assertion changed because our announcer element is now correctly hidden including by testing tools. Previously, the older hiding styles could sometimes make the hidden announcer still "visible" to getAllByText, causing the test to look for two instances

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -1,33 +1,84 @@
...
+    const announceDelay = 100;
+
+    const initialAnnounceTimeout = setTimeout(() => {
+      // Set the actual message to be announced
+      announcer.textContent = message;
+
+      // Set timeout to clear the message (if timeout > 0)
+      if (timeout > 0) {
+        const clearAnnounceTimeout = setTimeout(() => {
+          announcer.textContent = '';
```
britt6612: Screen readers are designed to read plain text in live regions, and using textContent ensures the message is announced clearly without interference from HTML structure.

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -1,33 +1,84 @@
...
     announcer.setAttribute('aria-live', mode);
-    setTimeout(() => {
-      announcer.innerHTML = '';
-    }, timeout);
+
+    // have to reset the content with something different (non breaking space)
+    // to force the screen reader to recognize the change.
+    announcer.textContent = '\u00A0';
+
+    // Force DOM reflow to ensure screen readers detect the change
```
britt6612: Using JavaScript, you can dynamically change the content inside the element with role="alert". Note that if you need to fire the same alert multiple times (i.e., the content you're dynamically inserting is the same as before), this generally won't be seen as a change and will not lead to an announcement. For this reason, it's usually best to briefly "clear" the contents of the alert container before then injecting the alert message.



from MDN

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -1,33 +1,84 @@
...
+    }
 
-    announcer.setAttribute('aria-live', 'off');
-    announcer.innerHTML = message;
     announcer.setAttribute('aria-live', mode);
-    setTimeout(() => {
-      announcer.innerHTML = '';
-    }, timeout);
+
+    // have to reset the content with something different (non breaking space)
```
britt6612: found online others having to clear the region with a non-breaking space to force screen readers to re-announce or  toggle the aria-live attribute or briefly remove and re-add the live region element to trigger announcements.

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -1,33 +1,84 @@
...
 
+  // Position element off-screen to hide it visually
+  announcer.style.position = 'absolute';
+  announcer.style.left = '-9999px';
+  announcer.style.height = '1px';
+  announcer.style.width = '1px';
+  announcer.style.overflow = 'hidden';
+
+  // Add to DOM at the beginning of body for best screen reader support
+  document.body.appendChild(announcer);
```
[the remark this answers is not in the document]
britt6612: removed

src/js/components/Grommet/__tests__/Grommet-test.js
```diff
@@ -78,9 +78,10 @@ describe('Grommet', () => {
...
       // should clear the aria-live container
-      expect(
-        document.body.querySelector('#grommet-announcer[aria-live]'),
-      ).toMatchSnapshot();
+      const announcer = document.body.querySelector(
+        '#grommet-announcer[aria-live]',
+      );
+      expect(announcer.textContent).toBe('hello');
       done();
     }, 600); // wait the aria-live container to clear
```
britt6612: this test keeps timing out. going to revisit

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -1,33 +1,84 @@
...
+        delete announcer.dataset.timeoutId;
+      }
+
+      // Clean up initial timeout reference
+      delete announcer.dataset.initialAnnounceTimeoutId;
+    }, announceDelay);
+
+    // helps avoid multiple announcements stacking up
+    announcer.dataset.initialAnnounceTimeoutId =
+      initialAnnounceTimeout.toString();
```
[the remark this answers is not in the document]
britt6612: not sure about this change I can explore @taysea @jcfilben think you guys can still review while I look into these changes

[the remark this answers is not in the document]
britt6612: read a little more into this:
`requestAnumationFrame` aligns with the browsers refresh cycle which will provide a more consistent timing.

src/js/components/Grommet/__tests__/__snapshots__/Grommet-test.js.snap
```diff
@@ -22,20 +22,27 @@ exports[`Grommet announce 1`] = `
 
 exports[`Grommet announce 2`] = `
 <div
+  aria-atomic="true"
   aria-live="assertive"
+  data-initial-announce-timeout-id="rAF"
+  data-raf-id="1"
```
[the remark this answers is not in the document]
taysea: @britt6612 is this something to be accounted for?

taysea:
VoiceOver is reading things out 2 times.

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -1,33 +1,99 @@
...
-  announcer.style.right = '100%';
-  announcer.style.position = 'fixed';
-  announcer.style['z-index'] = '-1';
+  announcer.setAttribute('id', 'grommet-announcer');
+  announcer.setAttribute('aria-live', 'polite');
+  announcer.setAttribute('aria-atomic', 'true');
 
+  // Position element off-screen to hide it visually
+  announcer.style.position = 'absolute';
+  announcer.style.left = '-9999px';
```
taysea, line 10: Why did we change it from -100% to -9999px?

britt6612: -100% may not always move the element far enough off-screen, since it depends on the element's width. Using -9999px ensures the element is positioned well outside the viewport, making it a more reliable way to guarantee the element remains completely hidden visually.

src/js/components/DataClearFilters/__tests__/DataClearFilters-test.tsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import { render, screen, fireEvent } from '@testing-library/react';
+import '@testing-library/jest-dom';
```
taysea, line 3: Was this required? Just double checking

britt6612: because Im using `toBeInTheDocument()` it was yelling at me to import 

I could remove and use `expect(element).toBeTruthy();` if that is preferred

taysea: I'm fine including jest-dom. Thanks for clarifying.

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -16,18 +20,51 @@ const createAnnouncer = () => {
...
+    announcer.offsetWidth;
+
+    // Set a short delay before placing the actual message.
+    const announceDelay = 100;
+
+    const initialAnnounceTimeout = setTimeout(() => {
+      announcer.textContent = message;
+
+      // Set a timeout to clear the message after the specified duration
+      // (if timeout > 0)
```
jcfilben:
```suggestion
```

src/js/contexts/AnnounceContext/AnnounceContext.js
```diff
@@ -16,18 +20,51 @@ const createAnnouncer = () => {
...
-      announcer.innerHTML = '';
-    }, timeout);
+    announcer.textContent = '';
+
+    announcer.textContent = '\u00A0';
+
+    // Force a DOM reflow. This makes the browser immediately apply the DOM
+    // changes which can help screen readers detect the change more reliably.
+    // eslint-disable-next-line no-unused-expressions
+    announcer.offsetWidth;
```
taysea, line 35: Is this reliable / best practice ? Just am unfamiliar so it felt a bit awkward.

taysea: Are there any performance implications for this?

britt6612: no agree its strange when you see. while looking online I did find this as a common pattern other libraries were doing for the similar reasoning. Its used so that screen readers can detect the DOM changes. 


this would only trigger on announcements so the performance is minimal in this case. 

I can look at my history and link examples I found

britt6612:
closing in favor of https://github.com/grommet/grommet/pull/7708

Issue #7658, opened 2025-05-27
DataTable - resize control keyboard accessible


jcfilben:
### Is this something you or your team is interested in contributing a pull request towards?

None

### Expected Behavior

Related information in this ticket: https://github.com/grommet/grommet/issues/7273

The column re-sizer should be keyboard accessible

Relevant for WCAG rule 2.1.1
We should update the DS DataTable docs once this is resolved

jcfilben:
PR merged and DS site updated

Issue #7273, opened 2024-07-16
DataTable column resizer control does not provide enough affordance


halocline:
### Is this something you or your team is interested in contributing a pull request towards?

Yes.

### Expected Behavior

Resizer control should provide [adequate click target size](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html) and not interfere with the cells contents.

### Actual Behavior

DataTable column resizer control does not provide enough affordance between itself and the header cell content when columns are right-aligned. This is especially true when consuming `grommet-theme-hpe`. 

Reported in [this issue](https://hpe.atlassian.net/browse/GLCP-181403?focusedCommentId=1094729).

### URL, screen shot, or Codepen exhibiting the issue

[Storybook](https://storybook.grommet.io/?path=/story/visualizations-datatable-resizable-columns--resizable-data-table&globals=theme:hpe)

![Screenshot 2024-07-16 at 1 36 08 PM](https://github.com/user-attachments/assets/0b4edde9-c212-401f-89f7-0cc15d451396)

### Steps to Reproduce

1. View [Storybook](https://storybook.grommet.io/?path=/story/visualizations-datatable-resizable-columns--resizable-data-table&globals=theme:hpe)

### Your Environment

- Grommet version: 2.37.0
- Browser Name and version:
- Operating System and version (desktop or mobile):

jcfilben:
I think there are two accessibility issues here, the click target size is not large enough and the resizer is not keyboard accessible. 

I looked at some of the other design systems from this [list](https://designsystems.surf/components/data-table) to see how they are handling column resizing in DataTables. The majority of the ones I looked at do not offer column resizing as an option. There were two that support column resizing:
- https://ag-grid.com/react-data-grid/column-sizing/
- https://www.lightningdesignsystem.com/components/data-tables/

The salesforce column resizer doesn't seem to be working. The ag-grid column resizer looks nice visually but doesn't meet the click target requirements or support keyboard interaction.

### ag-grid column resizer

The ag-grid column resizer is nicely positioned directly between the columns with a little bit of click target overlap on either side. To achieve the visual alignment of the resizer control ag-grid is using negative pixel values, something I'm thinking we should avoid if possible.

<img width="513" alt="Screenshot 2024-08-02 at 3 56 24 PM" src="https://github.com/user-attachments/assets/7360c885-8cbc-4177-a136-8fd77036df1d">

[...]

PR #7702, opened 2025-07-02
DataTable - Add keyboard support for column resizing and improve resizing behavior

jcfilben:
#### What does this PR do?
This PR adds keyboard support for resizing DataTable columns. The resizer is now a Button that is in the tab order and when focused the left and right arrows can be pressed to decrease or increase the size of the column.

In addition I cleaned up the resize behavior to work a little better for both keyboard and mouse.

https://github.com/grommet/grommet/issues/7658

no
yes
backwards compatible


src/js/components/DataTable/DataTable.js
```diff
@@ -399,17 +399,13 @@ const DataTable = ({
...
+  // Add handler for column resizing
+  const onResize = useCallback((property, width) => {
+    setWidths((prevWidths) => ({
+      ...prevWidths,
+      [property]: width,
+    }));
+  }, []);
 
   if (size && resizeable) {
     console.warn('DataTable cannot combine "size" and "resizeble".');
```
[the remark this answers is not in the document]
jcfilben: Fixed in latest commit

src/js/components/DataTable/Resizer.js
```diff
@@ -81,27 +83,42 @@ const Resizer = ({ onResize, property }) => {
     };
   }
 
+  const onKeyDown = useCallback(
+    (event) => {
```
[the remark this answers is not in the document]
jcfilben: Added this in latest commit

src/js/components/DataTable/Resizer.js
```diff
@@ -2,10 +2,13 @@ import React, { useCallback, useEffect, useRef, useState } from 'react';
...
 
 import { Box } from '../Box';
+import { Button } from '../Button';
+import { Keyboard } from '../Keyboard';
 import { Stack } from '../Stack';
 import { useThemeValue } from '../../utils/useThemeValue';
 
-const InteractionBox = styled(Box)`
+const InteractionBox = styled(Button)`
+  min-width: 2px;
```
taysea, line 21: Can you add a comment for why this value?

src/js/components/DataTable/Resizer.js
```diff
@@ -81,27 +84,43 @@ const Resizer = ({ onResize, property }) => {
...
   }
 
+  const onKeyDown = useCallback(
+    (event) => {
+      event.preventDefault();
+      if (!ref.current) return;
+      let element = ref.current;
+      while (element && element.nodeName !== 'TH') element = element.parentNode;
+      const currentWidth = element.getBoundingClientRect().width;
+      const delta = event.key === 'ArrowLeft' ? -12 : 12;
```
taysea, line 106: Can you add a comment for why "12"?

taysea:
The hover indicator doesn't fully cover the initial state like it used to. also should we center the hover style rather than right align?

Screenshot 2025-07-02 at 1 17 14 PM

src/js/components/DataTable/DataTable.js
```diff
@@ -399,20 +399,16 @@ const DataTable = ({
...
-        const nextWidths = { ...widths };
-        nextWidths[property] = width;
-        setWidths(nextWidths);
-      }
-    },
-    [widths],
-  );
+  // Add handler for column resizing
+  const onResize = useCallback((property, width) => {
+    setWidths((prevWidths) => ({
```
taysea: Can you explain why we don't need the check anymore of widths[property] !== width?

src/js/components/DataTable/Resizer.js
```diff
@@ -81,27 +84,43 @@ const Resizer = ({ onResize, property }) => {
...
+      onResize(property, currentWidth + delta);
```
taysea, line 107: When I have my dev tools panel open, the behavior is odd. in this case before the recording started, I'd resized my column larger with mouse. Then, I was pressing right arrow (so I would've thought the column would get bigger) but it actually got smaller:


https://github.com/user-attachments/assets/737dfbde-2c58-4748-99d5-631b741e4dfa

src/js/components/DataTable/DataTable.js
jcfilben: Oops, I pulled this out while I was exploring some different solutions and forgot to add it back in

src/js/components/DataTable/Resizer.js
jcfilben: Good call, added comment in latest commit

src/js/components/DataTable/Resizer.js
jcfilben: Yes, added comment in latest commit

jcfilben:
> The hover indicator doesn't fully cover the initial state like it used to. also should we center the hover style rather than right align?

This should be resolved in latest commit

As far as center aligning vs right aligning I think we can explore that a bit more when we adjust the styling for the min target size requirement. I felt like the styling might change a bit when we adjust the target size so it might make more sense to adjust this within that PR instead of here.


Screenshot 2025-07-09 at 2 19 16 PM

src/js/components/DataTable/DataTable.js
jcfilben: Fixed in latest commit

src/js/components/DataTable/Resizer.js
jcfilben: Good catch, I observed the same behavior. This is happening because when the screen gets resize the column widths don't get recalculated.

While diving into this issue I noticed various places where the widths calculations and management in DataTable are problematic and need cleaned up. I spent a bit of time trying to see if I could clean up the code but was running into issues with existing behavior breaking (particularly with pinned columns).

In the end I opted to just resolve the issue noted above by adding a `handleResize` function to TableCell.js that is triggered when the window is resized.

At some point we should re-visit the width handling in DataTable. For now I decided to not refactor it as the risk for potentially breaking things felt high and the time commitment needed also felt like more than we are able to invest at the moment.

src/js/components/DataTable/Header.js
```diff
@@ -155,7 +155,7 @@ const Header = forwardRef(
     const cellWidthsRef = useRef({});
     const timerRef = useRef();
 
-    const handleWidths = () => {
+    const handleWidths = useCallback(() => {
```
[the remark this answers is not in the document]
jcfilben: I think this is incorrectly flagged. `onSelect` and `selected` are being used within `handleWidths`

src/js/components/DataTable/Resizer.js
```diff
@@ -81,27 +87,46 @@ const Resizer = ({ onResize, property }) => {
...
-        onMouseDown={onMouseDown}
-        onMouseMove={start !== undefined ? onMouseMove : undefined}
-        onMouseUp={start !== undefined ? onMouseUp : undefined}
-      >
-        <Box pad={{ vertical: 'small' }} border={border} />
-      </InteractionBox>
+      <Keyboard onLeft={onKeyDown} onRight={onKeyDown}>
+        {/* provides a wider, more accessible target to grab resizer */}
+        <InteractionBox
+          aria-label={`Resize ${property} column`}
```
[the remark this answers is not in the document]
jcfilben: Incorporated in latest commit

src/js/components/DataTable/Resizer.js
```diff
@@ -81,27 +87,46 @@ const Resizer = ({ onResize, property }) => {
...
+    (event) => {
+      event.preventDefault();
+      if (!ref.current) return;
+      let element = ref.current;
+      while (element && element.nodeName !== 'TH') element = element.parentNode;
+      const currentWidth = element.getBoundingClientRect().width;
+      const propertyWidth = widths?.[property];
+      // Used 12 here to align with the value set in onMouseMove
+      const delta = event.key === 'ArrowLeft' ? -12 : 12;
+      onResize(property, (propertyWidth || currentWidth) + delta);
```
taysea: One thing I'm noticing is if I use the left arrow to reduce the width of a column and the column reaches a min-width, then when I hit right arrow it doesn't immediately increase the width of the column.

Let's say I get to the min-width of a column (not necessarily a defined min-width but just what there's space for), and I hit left arrow 5 times. It will take me multiple presses to have the column grow again.

Seems odd given what I see in the logic but maybe the property width is coming back smaller than it should be?

taysea: https://github.com/user-attachments/assets/85418717-7642-4755-81c3-59775bd05c86

taysea: notice the arrow keys I'm pressing (in bottom left of recording)

src/js/components/DataTable/Header.js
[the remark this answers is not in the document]
taysea: I'd agree this seems incorrectly flagged.

src/js/components/DataTable/Resizer.js
jcfilben: Good catch, just pushed up a change that should resolve this

src/js/components/DataTable/Header.js
```diff
@@ -165,15 +165,20 @@ const Header = forwardRef(
...
-      if (timerRef.current) clearTimeout(timerRef.current);
-      timerRef.current = setTimeout(handleWidths, WIDTH_UPDATE_DELAY);
-    };
+    }, [columns, onSelect, onWidths, selected]);
+
+    const updateWidths = useCallback(
+      (property, width) => {
+        if (typeof width !== 'number') return;
+        // Only update if width actually changed
+        if (cellWidthsRef.current[property] !== width) {
```
britt6612: could `cellWidthsRef` ever be undefined? or null?

src/js/components/DataTable/Resizer.js
```diff
@@ -81,27 +87,45 @@ const Resizer = ({ onResize, property }) => {
...
-        onMouseDown={onMouseDown}
-        onMouseMove={start !== undefined ? onMouseMove : undefined}
-        onMouseUp={start !== undefined ? onMouseUp : undefined}
-      >
-        <Box pad={{ vertical: 'small' }} border={border} />
-      </InteractionBox>
+      <Keyboard onLeft={onKeyDown} onRight={onKeyDown}>
+        {/* provides a wider, more accessible target to grab resizer */}
+        <InteractionBox
+          aria-label={`Resize ${headerText} column`}
```
britt6612: internationalize?

src/js/components/DataTable/DataTable.js
```diff
@@ -412,7 +412,7 @@ const DataTable = ({
   );
 
   if (size && resizeable) {
-    console.warn('DataTable cannot combine "size" and "resizeble".');
+    console.warn('DataTable cannot combine "size" and "resizeable".');
```
britt6612, line 415: nice catch

britt6612:
noticed in smaller screens the focus indicator doesn't quite cover the whole area. 

Screenshot 2025-07-10 at 11 12 58 PM

Screenshot 2025-07-10 at 11 13 20 PM

src/js/components/DataTable/Header.js
jcfilben: I added a `?` just in case

src/js/components/DataTable/Resizer.js
jcfilben: Good call, internationalized this in the latest commit

jcfilben:
> noticed in smaller screens the focus indicator doesn't quite cover the whole area.

I'm thinking we can handle this in https://github.com/grommet/grommet/issues/7273 since the styling will likely change when we change the target size

britt6612:
> > noticed in smaller screens the focus indicator doesn't quite cover the whole area.
> 
> I'm thinking we can handle this in #7273 since the styling will likely change when we change the target size

good point yes that sounds like a plan.

src/js/languages/default.json
```diff
@@ -58,6 +58,7 @@
   "dataTable": {
     "ascending": "ascending",
     "descending": "descending",
+    "resizerAria": "Resize {headerText} column",
```
britt6612, line 61: do we need to add to proptypes and TS?

jcfilben: Good catch, just pushed up a commit to address this

src/js/components/DataTable/Resizer.js
britt6612: thank you for adding

src/js/components/DataTable/Header.js
britt6612: okay!

src/js/languages/default.json
britt6612: thanks! I know the behavior might be hard to test should we just add a test for adding a `message` for `resizerAria` ?

jcfilben: I think the aria-label on resizer behavior is already covered in some existing snapshots so don't feel like we need a specific test for this. Let me know if you think otherwise

Screenshot 2025-07-14 at 11 26 32 AM

britt6612: oh I meant like passing a custom `message` through grommet `messages` prop since we just added this just making sure it comes through. I caught a mistake in my PR I have so figured more of a nice to have not necessary though

jcfilben: Ah gotcha, I can add a test for that

jcfilben: Just pushed up a test

PR #7705, opened 2025-07-04
Fix controlled form invalid state update with complex object

battistar:
Controlled form fail to update state when using complex object and dot path. In particular this append when using an array with a nested object as parent or child.
This make difficult to update complex child form object (https://design-system.hpe.design/templates/forms/managing-child-objects?q=chil) forcing you to flat the state object.

An example:

```js
export const DotPathForm = () => {
  const [value, setValue] = useState({
    foo: {
      bar: [
        {
          baz: {
            qux: '',
          },
        },
      ],
    },
  });

  return (
    <Form value={value} onChange={setValue}>
      <FormField htmlFor="foo.bar[0].baz.qux" name="foo.bar[0].baz.qux">
        <TextInput id="foo.bar[0].baz.qux" name="foo.bar[0].baz.qux" />
      </FormField>
    </Form>
  );
};
```

The expected result on `onChange` if I had written "test" in the input field:
```js
{
  "foo": {
    "bar": [
      {
        "baz": {
          "qux": "test"
        }
      }
    ]
  }
}

```

The result:
```js
{
  "foo": {
    "bar": [
      {
        "baz": {
          "qux": ""
        }
      }
    ]
  },
  "foo.bar": [
    {
      "baz.qux": "test"
    }
  ]
}
```

In this pull request I updated the dot path navigation function used to set and get values from input to form state. Probably the best way is to use a consolidate library like Lodash get and set object function (https://lodash.com/docs/4.17.15#set), but I notice that you choose to not use any external deps, so I decided to write the entire functionality.

Cheers

### DIFF
--- a/src/js/components/Data/__tests__/Data-test.tsx
+++ b/src/js/components/Data/__tests__/Data-test.tsx
@@ -14,9 +14,10 @@ import { Data } from '..';
 import { createPortal, expectPortal } from '../../../utils/portal';
 const DEBOUNCE_TIMEOUT = 300;
 
-// asserts that AnnounceContext aria-live region and visible DataSummary each have this text
+// asserts that DataSummary has text. Previously this was set to 2 to include AnnounceContext
+// but that is no longer the case with improvments to AnnounceContext in "polite" mode.
 const expectDataSummary = (message: string) =>
-  expect(screen.getAllByText(message)).toHaveLength(2);
+  expect(screen.getAllByText(message)).toHaveLength(1);
 
 const data = [
   {
--- a/src/js/components/DataClearFilters/__tests__/DataClearFilters-test.tsx
+++ b/src/js/components/DataClearFilters/__tests__/DataClearFilters-test.tsx
@@ -6,9 +6,10 @@ import { Grommet } from '../../Grommet';
 import { DataClearFilters } from '..';
 import { DataSummary } from '../../DataSummary';
 
-// asserts that AnnounceContext aria-live region and visible DataSummary each have this text
+// asserts that DataSummary has text. Previously this was set to 2 to include AnnounceContext
+// but that is no longer the case with improvments to AnnounceContext in "polite" mode.
 const expectDataSummary = (message: string) =>
-  expect(screen.getAllByText(message)).toHaveLength(2);
+  expect(screen.getAllByText(message)).toHaveLength(1);
 
 const data = [
   {
--- a/src/js/components/DataSummary/__tests__/DataSummary-test.tsx
+++ b/src/js/components/DataSummary/__tests__/DataSummary-test.tsx
@@ -8,9 +8,10 @@ import { DataTable } from '../../DataTable';
 import { Grommet } from '../../Grommet';
 import { DataSummary } from '..';
 
-// asserts that AnnounceContext aria-live region and visible DataSummary each have this text
+// asserts that DataSummary has text. Previously this was set to 2 to include AnnounceContext
+// but that is no longer the case with improvments to AnnounceContext in "polite" mode.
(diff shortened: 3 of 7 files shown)


src/js/components/Form/Form.js
```diff
@@ -35,50 +35,95 @@ const stringToArray = (string) => {
...
+      segments.push(part.slice(1, -1)); // remove brackets from array segment
+    } else {
+      segments.push(part);
+    }
+
+    match = segmentRegex.exec(path);
+  }
+
+  return segments;
 };
```
MikeKingdom: I think we should simplify a lot of this. First lets replace this with something like:

```suggestion
const splitPath = (path) => {
  // First replace numeric array indexes with the literal index.
  // For example "foo.bar[0].baz" -> "foo.bar.0.baz"
  // Then split it into the individual segments. e.g. ["foo", "bar", "0", "baz"]
  return path?.replace(/\[(\d+)\]/g, '.$1')?.split('.');
};
```

src/js/components/Form/Form.js
```diff
@@ -35,50 +35,95 @@ const stringToArray = (string) => {
...
 
-const setValueAt = (valueObject, pathArg, value) => {
+const getValueAt = (valueObject, path) => {
+  if (valueObject === undefined) {
+    return undefined;
+  }
+
+  const splittedPath = Array.isArray(path) ? path : splitDotPath(path);
```
MikeKingdom: Use the recommended `splitPath` and rename (splitted is not a word)
```suggestion
  const pathSegments = Array.isArray(path) ? path : splitPath(path);
```

src/js/components/Form/Form.js
```diff
@@ -35,50 +35,95 @@ const stringToArray = (string) => {
...
+
+  const segment = splittedPath.shift();
+  const key = parseInt(segment, 10) || segment;
+
+  if (splittedPath.length === 0) {
+    return valueObject[key];
+  }
+
+  return getValueAt(valueObject[key], splittedPath);
```
MikeKingdom: This doesn't work like you intend. if segment is "0" `parseInt(segment, 10)` will be falsey and it will use the second part of the ||. You don't have to bother with the `parseInt`. This should work fine:
```suggestion
  if (splittedPath.length === 0) {
    return valueObject[segment];
  }

  return getValueAt(valueObject[segment], splittedPath);
```

src/js/components/Form/Form.js
```diff
@@ -35,50 +35,95 @@ const stringToArray = (string) => {
...
+
+  const splittedPath = Array.isArray(path) ? path : splitDotPath(path);
+
+  if (splittedPath.length === 1) {
+    object[splittedPath[0]] = value;
+  } else {
+    const segment = splittedPath.shift();
+    const key = parseInt(segment, 10) || segment;
+
+    const isNextValueArray = !!parseInt(segment, 10);
```
MikeKingdom: This doesn't seem to work for the same reason noted above. `!!parseInt("0", 10)` will be false. You probably would be fine doing `!Number.isNaN(parseInt(segment, 10))`. I'd also move this check inside the `if` at line 97. No point doing it when not needed

src/js/components/Form/Form.js
```diff
@@ -35,50 +35,95 @@ const stringToArray = (string) => {
...
+
+  const segment = splittedPath.shift();
+  const key = parseInt(segment, 10) || segment;
+
+  if (splittedPath.length === 0) {
+    return value[key];
   }
-  return getValueAt(value, name);
+
+  return getValueAt(value[key], splittedPath);
```
MikeKingdom: same comments as above

PR #7707, opened 2025-07-10
DataTable - add `expandLabel` for users to be able to pass aria label to `groupBy` or `rowDetails`

### LINKED ISSUE #6213: a11y: expand is not specific (breaks 3.3.2)

britt6612:
#### What does this PR do?
Adds a new `expandLabel` to `rowDetails` or `groupBy` object.

ExpanderCell.js 
Passed `rowExpandLabel` to one of the DataTable stories
look at the ControlledGrouped story in Storybook
Use a screen reader or inspect the expand/collapse buttons

TODO 

Previously, group expand/collapse buttons in DataTable only had generic "expand" and "collapse" labels, making it difficult for screen reader users to understand what content they were expanding.

What are the relevant issues?
closes #6213

Screenshots (if appropriate)

Yes

Yes 

backwards compatible


src/js/components/DataTable/ExpanderCell.js
```diff
@@ -35,11 +42,17 @@ const ExpanderControl = ({ context, expanded, onToggle, pad, ...rest }) => {
   );
 
   if (onToggle) {
+    let a11yTitle;
```
taysea, line 70: A logic suggestion:

```
let a11yTitle = expanded ? 'collapse' : 'expand';
if (rowExpandedLabel) a11yTitle = `${a11yTitle} ${rowExpandLabel}`;
```

src/js/components/DataTable/Body.js
```diff
@@ -33,6 +33,7 @@ const Row = memo(
     setActive,
     setRowExpand,
     rowExpand,
+    rowExpandLabel,
```
taysea: Thinking about naming suggestions (since this applies for `rowDetails` and `groupBy` and to clarify that its an ariaLabel not a visual label):
- `expandAriaLabel` (my top suggestion)
- `rowExpandAriaLabel` (kinda long)

@jcfilben

taysea (commented):
The approach seems reasonable, just a small suggestion on naming of prop

src/js/components/DataTable/Body.js
jcfilben: Good idea clarifying that this is an aria label. I think `expandAriaLabel` is clear enough. Agree `rowExpandAriaLabel` starts getting a little long

src/js/components/DataTable/ExpanderCell.js
jcfilben: Should we internationalize the `expand` and `collapse` strings?

src/js/components/DataTable/Body.js
britt6612: yeah thats fine `expandAriaLabel` makes sense

britt6612:
> Could we add a jest test?

yes I was kinda holding off on this until we were good with direction and naming but was in my TODO

src/js/components/DataTable/ExpanderCell.js
britt6612: we can? I think it would help forsure!

src/js/components/DataTable/Body.js
britt6612: changed name

britt6612:
> Could we add a jest test?

added the tests @jcfilben

src/js/components/DataTable/ExpanderCell.js
britt6612: implemented logic suggestion and internationalized the expand and collapse

jcfilben:
I wonder if we should pass back the row and the key to `expandAriaLabel` when it is a function similar to how we pass row through `rowDetails`. That way someone could do something like this: `expandAriaLabel={(row) => 'details for ${row.name}'}`

britt6612:
> I'm not sure I understand the utility of passing a string to expandAriaLabel. It seems like at the moment the string just gets read out exactly as is. I can't think of a case where someone would want each expand/collapse button to have the same arialabel. I could see the string having utility if we did something where we could pass expandAriaLabel="location" and we have a location column and the data in that column gets read out as the aria label.

my thinking here was if they only had one item that was expand/collapse then they can pass a string to just apply to that one aria-label.

britt6612:
> I wonder if we should pass back the row and the key to `expandAriaLabel` when it is a function similar to how we pass row through `rowDetails`. That way someone could do something like this: `expandAriaLabel={(row) => 'details for ${row.name}'}`

I can look at adding this if we think it will benefit

src/js/components/DataTable/propTypes.js
```diff
@@ -130,7 +130,9 @@ if (process.env.NODE_ENV !== 'production') {
     ]),
     messages: PropTypes.shape({
       ascending: PropType.string,
+      collapse: PropType.string,
```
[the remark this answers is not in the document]
britt6612: done

src/js/components/DataTable/propTypes.js
```diff
@@ -130,7 +130,9 @@ if (process.env.NODE_ENV !== 'production') {
...
       descending: PropType.string,
+      expand: PropType.string,
```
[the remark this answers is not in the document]
britt6612: done

src/js/components/DataTable/ExpanderCell.js
```diff
@@ -35,11 +47,45 @@ const ExpanderControl = ({ context, expanded, onToggle, pad, ...rest }) => {
...
+        id: 'dataTable.collapse',
+        messages: {
+          'dataTable.collapse': 'collapse',
+        },
+      });
+
+    let a11yTitle;
+
+    if (context === 'header') {
+      a11yTitle = expanded ? 'collapse all' : 'expand all';
```
jcfilben: Can we internationalize theses strings?

src/js/components/DataTable/ExpanderCell.js
```diff
@@ -35,11 +47,45 @@ const ExpanderControl = ({ context, expanded, onToggle, pad, ...rest }) => {
...
+    } else {
+      a11yTitle = expanded ? collapseText : expandText;
+
+      if (expandAriaLabel) {
+        const ariaLabel =
+          typeof expandAriaLabel === 'function'
+            ? expandAriaLabel(row, groupKey)
+            : expandAriaLabel;
```
jcfilben: If expandAriaLabel isn't a function it seems like `ariaLabel` should just be undefined

src/js/components/DataTable/index.d.ts
```diff
@@ -117,6 +119,7 @@ export interface DataTableProps<TRowType = any> {
         expand?: (string | number)[];
         onExpand?: (expandedKeys: (string | number)[], datum: TRowType) => void;
       };
+  expandAriaLabel?: (row: TRowType, groupKey: any) => string;
```
jcfilben: Will groupKey always be a string?
```suggestion
  expandAriaLabel?: (row: TRowType, groupKey: string) => string;
```

jcfilben (commented):
Could we add an expandAriaLabel to any DataTable story that has groups or rowDetails? That way we show best practice in our stories

src/js/components/DataTable/index.d.ts
britt6612: Im not sure I wonder if it could also be a number 🤔

src/js/components/DataTable/ExpanderCell.js
britt6612: done

src/js/components/DataTable/ExpanderCell.js
britt6612: in body.js Im passing 

```
    expandAriaLabel={
              expandAriaLabel ? expandAriaLabel(datum) : undefined
            }
```

so because of this in ExpanderCell.js `expandAriaLabel` gets passed a string. If I passed it directly 
`expandAriaLabel={ExpandAriaLabel}` the problem with this is that in ExpanderCell we do not have access to the specific datum of that row. So we need to do this ` expandAriaLabel ? expandAriaLabel(datum)` in which sends the specific row string to expandAriaLabel so in ExpanderCell the `expandAriaLabel` is a string not a func

let me know if this makes sense

src/js/languages/default.json
```diff
@@ -57,6 +57,8 @@
   },
   "dataTable": {
     "ascending": "ascending",
+    "collapse": "collapse",
+    "expand": "expand",
```
taysea: nit pick, alphabetize

src/js/components/DataTable/ExpanderCell.js
taysea: The collapse all and expand all don't seem to be internationalized

britt6612: weird these changes didnt commit

britt6612: thanks for catching ill push up!

src/js/languages/default.json
britt6612: done

src/js/components/DataTable/ExpanderCell.js
britt6612: @taysea pushed up

src/js/components/DataTable/Body.js
```diff
@@ -176,6 +177,9 @@ const Row = memo(
             }}
             pad={cellProps.pad}
             verticalAlign={verticalAlign}
+            expandAriaLabel={
+              expandAriaLabel ? expandAriaLabel(datum) : undefined
```
taysea: Is it confusing that for the grouped case we pass 2 things to the function but for the body we just pass the datum?

src/js/components/DataTable/index.d.ts
taysea: The way this is written, the non-grouped case would fail on Typescript because we're only passing it a single argument

src/js/components/DataTable/Body.js
britt6612: yes I see your point how this could be confusing.  @taysea 

GROUPED case (GroupedBody):
`expandAriaLabel(row, groupKey)`

ROW DETAILS case (Body):
`expandAriaLabel(datum)`


In the Row Details case what you think if we change to 

```
expandAriaLabel={
  expandAriaLabel ? expandAriaLabel(datum, undefined) : undefined
}
```

src/js/components/DataTable/index.d.ts
britt6612: see my comment below about updating the non-group case

src/js/components/DataTable/propTypes.js
```diff
@@ -174,6 +178,7 @@ if (process.env.NODE_ENV !== 'production') {
     replace: PropTypes.bool,
     resizeable: PropTypes.bool,
     rowProps: PropTypes.shape({}),
+    expandAriaLabel: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
```
taysea:
```suggestion
    expandAriaLabel: PropTypes.func,
```

src/js/components/DataTable/index.d.ts
```diff
@@ -117,6 +121,7 @@ export interface DataTableProps<TRowType = any> {
         expand?: (string | number)[];
         onExpand?: (expandedKeys: (string | number)[], datum: TRowType) => void;
       };
+  expandAriaLabel?: ({ row }: { row: TRowType }) => string | string;
```
taysea: I know I suggested this object approach, but now I'm thinking maybe we just match the structure of `rowDetails` and pass it back as `(row) => `

britt6612: changed back!

src/js/components/DataTable/ExpanderCell.js
```diff
@@ -35,11 +46,68 @@ const ExpanderControl = ({ context, expanded, onToggle, pad, ...rest }) => {
...
+        if (typeof expandAriaLabel === 'function') {
+          if (context === 'groupHeader') {
+            ariaLabel = expandAriaLabel(row.datum);
+          } else {
+            ariaLabel = expandAriaLabel(row);
+          }
+        } else {
+          ariaLabel = expandAriaLabel;
+        }
+        a11yTitle = `${a11yTitle} ${ariaLabel}`;
```
MikeKingdom: This looks to me like it's trying to compose a phrase by concatenating several parts. This is something we need to avoid when internationalizing. Text should be combined via substitution parameters for the overall phrase so localizers can order things as best makes sense for the language.

src/js/components/DataTable/ExpanderCell.js
```diff
@@ -35,11 +46,68 @@ const ExpanderControl = ({ context, expanded, onToggle, pad, ...rest }) => {
   );
 
   if (onToggle) {
+    const expandText =
+      messages?.expand ||
+      format({
+        id: 'dataTable.expand',
+        messages: {
+          'dataTable.expand': 'expand',
+        },
```
MikeKingdom: Why are these basically hardcoding messages? This parameter should simply be `messages`

britt6612: sorry was testing then forgot to revert back to `messages` changed now

taysea: Seems like the commit to revert this back didn't get pushed up

src/js/components/DataTable/ExpanderCell.js
```diff
@@ -35,11 +45,61 @@ const ExpanderControl = ({ context, expanded, onToggle, pad, ...rest }) => {
...
+        values: { label: expandLabel || '' },
+      });
+
+    const collapseText =
+      messages?.collapse ||
+      format({
+        id: 'dataTable.collapse',
+        messages: {
```
taysea: Same with here, seems like this should just be messages

taysea (commented):
I like the modified direction to include the property in the object of `rowDetails` and `groupBy`. A few small comments.

src/js/components/DataTable/ExpanderCell.js
britt6612: ah weird thought I pushed up! pushing up now

src/js/components/DataTable/ExpanderCell.js
```diff
@@ -35,11 +45,57 @@ const ExpanderControl = ({ context, expanded, onToggle, pad, ...rest }) => {
...
+      format({
+        id: 'dataTable.collapseAll',
+        messages,
+      });
+
+    let a11yTitle;
+    if (expandLabel) {
+      a11yTitle = format({
+        id: expanded ? 'dataTable.collapse' : 'dataTable.expand',
+        messages: {
```
taysea: Should this just be `messages` too?

taysea: Pushed a change for this.

src/js/components/DataTable/ExpanderCell.js
```diff
@@ -35,11 +45,54 @@ const ExpanderControl = ({ context, expanded, onToggle, pad, ...rest }) => {
   );
 
   if (onToggle) {
+    const expandText =
+      messages?.expand ||
```
taysea: Do we need this `||` condition or can it simply be `format({ ...})`?

taysea: Pushed fix.

MikeKingdom: Shouldn't need these checks. `format` was designed with the expectation that at least the base messages from grommet has the messages

taysea (approved):
LGTM, thanks!

PR #7713, opened 2025-07-15
Fix TextInput icon visibility when readOnly is true (#7668)

### LINKED ISSUE #7668: TextInput, when "readOnly" is true, the icon hides

Yes

### Expected Behavior

`<TextInput readOnly={true} icon={<UserExpert />}/>`
`<TextInput readOnly={false} icon={<UserExpert />}/>`

### Actual Behavior

`<TextInput readOnly={true} icon={undefined}/>`
`<TextInput readOnly={false} icon={<UserExpert />}/>`

### readOnly is false
Image

### readOnly is true
Image

### Steps to Reproduce

1. Render a button with custom icon
2. Place a boolean value controlled by react
3. Try to change the value of the boolean

### Your Environment

- Grommet version
    "grommet": "2.46.0",
    "grommet-icons": "4.12.4",

- Browser version: Version 136.0.7103.114 (Official Build) (arm64)

- Operating System and version: macOS Sequoia 15.3.1

avantikacode:
#### What does this PR do?

This PR ensures that the TextInput component continues to display the icon even when the readOnly prop is set to true. Previously, the icon would not render in read-only mode due to a conditional check that excluded it.

```
```

- Manual testing in a local storybook/example showing TextInput with readOnly and icon props.

- Ran full Jest test suite.

- Updated affected snapshot tests using jest test-update.

- Render a TextInput with both readOnly and icon props set.

- Confirm that the icon is visible while the input remains non-editable.

- Run yarn test or npm test to ensure all tests pass.

Issue #7668 noted that TextInput hides its icon when in readOnly mode, which was not the expected behavior. This fix simply removes that condition so the icon is always rendered.

Closes #7668

N/A – visual is unchanged aside from icon visibility in readOnly mode.

Behavior is aligned with existing props; no doc change required.

- [x] Yes
- [ ] No
“Fixed: TextInput now shows icon even when readOnly is true.”

Yes


britt6612:
@avantikacode looks like some snapshots are failing did you run `yarn jest -u` on your end?

avantika-msr:
> yarn jest -u

Yes @britt6612 I did and below is the output:

Test Suites: 97 passed, 97 total
Tests:       1682 passed, 1682 total
Snapshots:   1910 passed, 1910 total
Time:        12.135 s

All tests are passing locally. Could you please help me here?

britt6612:
Ill pull down locally and see what I can do!

britt6612:
@avantikacode can you locally run `yarn` then run `yarn jest -u` I did this locally and it reverted the snapshot changes back from your fix there shouldnt be this many snapshot changes

avantika-msr:
> @avantikacode can you locally run `yarn` then run `yarn jest -u` I did this locally and it reverted the snapshot changes back from your fix there shouldnt be this many snapshot changes

Sure @britt6612 will try it once today.

avantika-msr:
Hey @britt6612 the suggested changes were done and the build has passed, you can now have a look. Thanks!

taysea:
@jcfilben Was there any intentional reason when we introduced `readOnly` to exclude the icon? It seems reasonable to keep the icon visible, but just checking.

jcfilben:
> @jcfilben Was there any intentional reason when we introduced readOnly to exclude the icon? It seems reasonable to keep the icon visible, but just checking.

I think that was intentional, @KennyAtHPE not sure if I am remembering correctly but were you the designer on this? If so do you remember why we excluded the icon in readonly mode?

KennyAtHPE:
Hey @jcfilben and @taysea, I don't believe this was originally intentional. In my opinion, the icon should remain regardless of the state/readonly status, since it's purely decorative and not an interaction we'd want to limit based on state.

KennyAtHPE:
Hey @jcfilben @taysea @britt6612 , I had a chance to peruse the old exploration file, and I believe we came to the conclusion of disabling the icon within a Text Input readOnly mode because:
- The original exploration contained multiple designs for readOnly on other inputs besides just Text Input, such as Search and Select. In those cases, the search icon and caret icon would be omitted because they are affordances for interactions.
- To keep things lean and a low-risk decision, the Text Input would start out by omitting the icons for the same reason.

I think the main distinction here is that there are two different types of icons being used within our inputs.
- Purely decorative to help highlight what the content is related to.
- Both decorative and an interactive affordance to highlight the type of content and type of input.

I agree with the original rationale that icons such as search, the caret within select, and the calendar within Date Input should be hidden since they fall underneath the second category of icons. We'd want to avoid providing affordances that say "Hey I'm interactable, click on me." But for icons that fall underneath the first category, I would probably suggest exposing those, since they don't affect the user's perception of whether or not that particular input is interactable or not.

Regardless, I'm not sure if that nuance is too much or even doable within code.

jcfilben:
Linking this PR to https://github.com/grommet/grommet/issues/7668

PR #7714, opened 2025-07-16
DataTable -- Add touch gesture and dragging alternative support

### LINKED ISSUE #7711: DataTable - dragging movements wcag 2.5.7
Currently the resize columns functionality of the DataTable doesn't provide an alternative to dragging movements for mouse, touch screen, or stylus users. This violates WCAG rule 2.5.7.

Once this issue is resolved we should update the DataTable WCAG information on the design system site to set 2.5.7 to passing

taysea:
#### What does this PR do?

Updates Resizer to support both touch screen gestures and an alternative to dragging. The alternative is required to meet WCAG 2.5.7.

Now, if someone clicks on the Resizer, a dropdown appears with option to increase/decrease column width without dragging.

TO DO:
- [x] Add internationalization and aria-label.
- [x] Test touch gestures on mobile device

Closes #7711


src/js/components/DataTable/Resizer.js
```diff
@@ -40,50 +47,61 @@ const Resizer = ({ onResize, property, headerText, messages }) => {
...
-      document.removeEventListener('mousemove', onMouseMove);
+      document.removeEventListener('mouseup', onResizeEnd);
+      document.removeEventListener('mousemove', onResizeMove);
     };
 
     if (active) {
-      document.addEventListener('mouseup', onMouseUp);
-      document.addEventListener('mousemove', onMouseMove);
+      document.addEventListener('mouseup', onResizeEnd);
+      document.addEventListener('mousemove', onResizeMove);
```
britt6612, line 97: should we also add the following? 
```
  document.addEventListener('touchend', onResizeEnd);
  document.addEventListener('touchmove', onResizeMove);
```

I saw you added down below 
```
          onTouchStart={onResizeStart}
          onTouchMove={start !== undefined ? onResizeMove : undefined}
          onTouchEnd={start !== undefined ? onResizeEnd : undefined}
```

src/js/components/DataTable/Resizer.js
```diff
@@ -131,9 +149,52 @@ const Resizer = ({ onResize, property, headerText, messages }) => {
           margin={{ top: 'xsmall' }}
           ref={ref}
           responsive={false}
-          onMouseDown={onMouseDown}
-          onMouseMove={start !== undefined ? onMouseMove : undefined}
-          onMouseUp={start !== undefined ? onMouseUp : undefined}
+          onMouseDown={onResizeStart}
+          onMouseMove={start !== undefined ? onResizeMove : undefined}
```
britt6612: do we actually need these since we have the event listeners above?

taysea: I'm thinking that one of the 2 is redundant and I'm leaning towards remove the document event listeners and instead just leveraging these here. Will do some testing, just didn't want to remove anything since it was there before

src/js/components/DataTable/Resizer.js
taysea: See comment below but I think the event listeners might be redundant, will do some testing

src/js/components/DataTable/Resizer.js
britt6612: ah yes okay I can also test on my end if we remove!

taysea: Just pushed up a change -- we do need the global event listeners for mouse/touch move and mouseup/touchend in case the mouse or touch moves outside of the element while the person is interacting.

for mousedown and touchstart, we only want to listen for that inside the element itself!

src/js/components/DataTable/Resizer.js
```diff
@@ -40,50 +47,65 @@ const Resizer = ({ onResize, property, headerText, messages }) => {
...
-  const onMouseDown = useCallback((event) => {
+  useEffect(() => {
     if (ref.current) {
       let element = ref.current;
       // find TH parent
-      while (element && element.nodeName !== 'TH') element = element.parentNode;
+      while (element && element.nodeName !== 'TH') {
+        element = element.parentNode;
+      }
+      thRef.current = element;
```
jcfilben, line 54: This ref is the `th` parent, not the actual `th` right? I feel like the naming implies that this is the `th`

taysea: I believe it is the `th` -- the while loop moves up to the parent of the element until it is the `th`

jcfilben: Ah okay that makes sense!

jcfilben (approved):
Looks good!

jcfilben (approved):

src/js/components/DataTable/Resizer.js
```diff
@@ -51,48 +59,62 @@ const Resizer = ({ onResize, property, headerText, messages, headerId }) => {
     }
   }, [ref]);
 
-  const onMouseDown = useCallback((event) => {
+  useEffect(() => {
```
britt6612: there is a `useEffect` above that sets the initial width, can we not combine the 2 `useEffects` or do we need both?

src/js/components/DataTable/Resizer.js
```diff
@@ -141,28 +164,70 @@ const Resizer = ({ onResize, property, headerText, messages, headerId }) => {
...
+              autoFocus
+            />
+            <Button
+              aria-label={format({
+                id: 'dataTable.increase',
+                values: { headerText },
+                messages,
+              })}
+              icon={<Add />}
+              onClick={() => {
```
britt6612: should these calls be in a `useCallback` ?

taysea: Pulled them out to `useCallback` to optimize so that functions aren't recreated on every render

src/js/components/DataTable/Resizer.js
taysea: Consolidated and removed unneeded `ref` in dependency array since changes to `refs` don't cause re-renders.

src/js/components/DataTable/Resizer.js
```diff
@@ -6,29 +6,36 @@ import React, {
...
-import { focusStyle, unfocusStyle } from '../../utils/styles';
 
-const StyledResizer = styled(Box)`
+// We determined 12 empirically as being wide enough to hit but
+// not too wide to cause false hits.
+const STEP = 12; // Used to determine the width change on resize
+
+// Added a temporary min-width of 2px here so that the element doesn't
+// end up with a width of 0px. This is a placeholder solution until we
+// revisit this in https://github.com/grommet/grommet/issues/7273
```
jcfilben:
```suggestion
```
I think we can remove this comment

jcfilben (approved):

src/js/components/DataTable/Resizer.js
britt6612: thank you!

britt6612 (approved):
Thank you for cleaning up!

PR #7715, opened 2025-07-17
Prevent umounting of Accordion and Collapsible content

battistar:
When an accordion panel are callapsed its content are removed from DOM. If an accordion containing input fields within a form is collapsed, its inputs are unmounted, causing the form to submit an incomplete or invalid data set (https://design-system.hpe.design/templates/forms/managing-child-objects?q=child).

In this PR I added a `keepMount` prop to `Accordion` and `Collapsible` component to prevent content umounting.

`Accordion` and `Collapsible` docs may be update adding `keepMount` prop description.

### DIFF
--- a/src/js/components/Accordion/Accordion.js
+++ b/src/js/components/Accordion/Accordion.js
@@ -16,6 +16,7 @@ const Accordion = forwardRef(
       level,
       multiple,
       onActive,
+      keepMount,
       ...rest
     },
     ref,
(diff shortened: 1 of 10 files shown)


jcfilben:
Hey @battistar I tested collapsing the accordion after entering some form information into it on the Managing Child Objects - Required children example and it seems like the data that is submitted looks correct. Is there something else I need to test in order to see the behavior you're describing?

Here is a video of how I tested:

https://github.com/user-attachments/assets/095f4c8d-db4c-49f1-9f7b-df7b9598b143

battistar:
@jcfilben all examples in that documentation section don't use an accordion, but a custom solution to prevent the wrong behavior. Check the example code for more info.

src/js/components/Accordion/Accordion.js
```diff
@@ -16,6 +16,7 @@ const Accordion = forwardRef(
       level,
       multiple,
       onActive,
+      keepMount,
```
britt6612: We talked as a dev group about the naming and thought we should change the prop to `unmount` and default it to true.

PR #7716, opened 2025-07-17
Fix target size on resize control

jcfilben:
#### What does this PR do?
Changes the DataTable resizer to meet target size requirements (min. 24px x 24px)
I also improved the positioning of the control so that the resizer visually lines up with the edge of the column and the target area is on both the left and right of the visual line.

https://github.com/grommet/grommet/issues/7273

no
yes
backwards compatible


src/js/components/DataTable/Resizer.js
```diff
@@ -8,29 +8,28 @@ import React, {
...
 
-// Added a temporary min-width of 2px here so that the element doesn't
-// end up with a width of 0px. This is a placeholder solution until we
-// revisit this in https://github.com/grommet/grommet/issues/7273
-const InteractionBox = styled(Button)`
-  min-width: 2px;
+const StyledResizer = styled(Box)`
+  position: absolute;
+  right: 0;
+  margin-right: -12px;
```
britt6612: why 12px?

src/js/components/DataTable/Resizer.js
```diff
@@ -8,29 +8,28 @@ import React, {
...
-  > * {
-    opacity: 0;
+  z-index: 1;
+  &:focus {
+    ${(props) =>
+      (!props.plain || props.focusIndicator) &&
+      focusStyle({ inset: props.focusIndicator === 'inset' })}
   }
 
-  // when mouse down, we want to continue to display styling
```
britt6612, line 27: did we lose styling here by removing this?

src/js/components/DataTable/Resizer.js
britt6612: why 12?

src/js/components/DataTable/Resizer.js
```diff
@@ -109,36 +120,32 @@ const Resizer = ({ onResize, property, headerText, messages }) => {
...
-          onMouseDown={onMouseDown}
-          onMouseMove={start !== undefined ? onMouseMove : undefined}
-          onMouseUp={start !== undefined ? onMouseUp : undefined}
-        >
-          <Box pad={{ vertical: 'small' }} border={border} />
-        </InteractionBox>
-      </Keyboard>
-    </Stack>
+    <Keyboard onLeft={onKeyDown} onRight={onKeyDown}>
+      <StyledResizer
```
britt6612, line 143: did you look at adding an `aria-role` seperator?

src/js/components/DataTable/Resizer.js
```diff
@@ -109,36 +120,32 @@ const Resizer = ({ onResize, property, headerText, messages }) => {
...
+        onMouseEnter={() => setHover(true)}
+        onMouseLeave={() => setHover(false)}
+        onMouseDown={onMouseDown}
+        onMouseMove={start !== undefined ? onMouseMove : undefined}
+        onMouseUp={start !== undefined ? onMouseUp : undefined}
+        ref={ref}
+        pad={{ vertical: 'xsmall' }}
+      >
+        <Box
+          border={hover ? hoverBorder : border}
```
britt6612, line 165: looks like there isnt a border change on `hover` now why did we decide this?

Sulaymon333:
I notice the separator now grows vertically when the screen is smaller. Unlike before that a constant height is maintained. Is this the intended behavior?

image


Initial behavior
image

src/js/components/DataTable/Resizer.js
jcfilben: I just aligned with the styling that is on master branch. In the grommet theme there is no border change on hover, but in the hpe theme there is

src/js/components/DataTable/Resizer.js
jcfilben: The styling should still match what we have on master

src/js/components/DataTable/Resizer.js
jcfilben: Oops, I forgot to change this to a tshirt value, fixed in latest commit

src/js/components/DataTable/Resizer.js
jcfilben: Good call, added this in latest commit

src/js/components/DataTable/Resizer.js
jcfilben: Change this to a tshirt value in latest commit

jcfilben:
> I notice the separator now grows vertically when the screen is smaller. Unlike before that a constant height is maintained. Is this the intended behavior?

Good question, this is intentional. In smaller screens when the header text wraps the height of the header cell increases and the resizer height also increases to better visually align with the new height of the cell. I think this behavior looks better but open to any feedback if other's disagree

src/js/components/DataTable/Resizer.js
britt6612: ah yes your right in hpe theme

src/js/components/DataTable/Header.js
```diff
@@ -457,25 +446,19 @@ const Header = forwardRef(
...
-                      searcher || resizer
+                      searcher
                     )}
                   </Box>
                 );
               }
               const cellPin = [...pin];
               if (columnPin) cellPin.push('left');
 
+              const headerId = `grommet-data-table-header-${property}`;
```
britt6612, line 466: does this help with the accessibility I noticed we added the `id` below with the name?

jcfilben: I added it to be able to link the id to the `aria-controls` prop, however since we removed the role='seperator' we no longer need the `aria-controls` prop so I removed this

src/js/components/DataTable/Resizer.js
jcfilben: I actually went back and removed this. Brittany and I discussed this on a call, just summarizing for anyone else reviewing. `role='seperator'` requires that you also include `aria-valuemin` and `aria-valuemax` and I couldn't find a way to give an accurate min/max for the cell width so we decided to remove the role. I tested the screenreader announcement for the resizer control and I think it is working well as is. I also looked at other libraries and didn't see people using `role='seperator'` for their resize controls.

britt6612:
is this okay that the focus is right on the name of the next column?

Screenshot 2025-07-21 at 3 46 24 PM

jcfilben:
> is this okay that the focus is right on the name of the next column?

Yes, I think that's fine. The hpe focus indicator is built in such a way that having the focus indicator right against the text of the next column isn't an issue for accessibility

src/js/components/DataTable/Resizer.js
jcfilben: I looked into this more, actually we just need `aria-valuenow` and it is okay if we don't have `aria-valuemin` or `aria-valuemax`. Pushing up changes adding role seperator back in.

src/js/components/DataTable/Header.js
jcfilben: Added back role seperator (see comment here: https://github.com/grommet/grommet/pull/7716#discussion_r2220453783)

src/js/components/DataTable/Resizer.js
britt6612: oh good thank you for checking

britt6612:
> > is this okay that the focus is right on the name of the next column?
> 
> Yes, I think that's fine. The hpe focus indicator is built in such a way that having the focus indicator right against the text of the next column isn't an issue for accessibility

agree here I like the new behavior

britt6612:
> > is this okay that the focus is right on the name of the next column?
> 
> Yes, I think that's fine. The hpe focus indicator is built in such a way that having the focus indicator right against the text of the next column isn't an issue for accessibility

okay thank you for confirming

britt6612 (approved):

src/js/components/DataTable/Resizer.js
```diff
@@ -105,40 +125,44 @@ const Resizer = ({ onResize, property, headerText, messages }) => {
...
+        onMouseLeave={() => setHover(false)}
+        onMouseDown={onMouseDown}
+        onMouseMove={start !== undefined ? onMouseMove : undefined}
+        onMouseUp={start !== undefined ? onMouseUp : undefined}
+        ref={ref}
+        pad={{ vertical: 'xsmall' }}
+        margin={{ right: `-${theme.global.edgeSize.small}` }}
+        role="separator"
+        aria-valuenow={width}
+        aria-valuetext={width ? `${ariaLabel} ${width} pixels` : ariaLabel}
```
Sulaymon333: It will be nice for screenreader to say something like `Resizing column Name (50%) of original` for example. Now, the SR announces the current column width value in px value like `Resize column name 234.3444px` which I believe might be less comprehensible for a screen reader user.

Sulaymon333 (changes requested):

src/js/components/DataTable/Resizer.js
jcfilben: I feel like we don't have a great way to calculate what percentage they are at (see above comment thread about min/max values https://github.com/grommet/grommet/pull/7716#discussion_r2220390290)

src/js/components/DataTable/Header.js
```diff
@@ -457,25 +446,19 @@ const Header = forwardRef(
...
-                        flex="shrink"
-                        direction="row"
-                        align="center"
-                        gap={theme.dataTable.header.gap}
-                      >
-                        {searcher}
-                        {resizer}
-                      </Box>
+                    {searcher && onResize ? (
+                      <Box pad={{ right: 'xsmall' }}>{searcher}</Box>
```
taysea: Based on offline discussion, should we hardcode this  to 12px with a comment as to why?

I believe Sulaymon has the change on his branch but unsure if we want to treat this one as the "base" branch.

Fine either way if you want to fold it in here or wait.

src/js/components/DataTable/Header.js
```diff
@@ -498,12 +482,28 @@ const Header = forwardRef(
                       ? `${widths[property]}px`
                       : undefined,
                     boxSizing: onResize ? 'border-box' : undefined,
+                    position: onResize ? 'relative' : undefined,
+                    overflow: onResize ? 'visible' : undefined,
```
taysea, line 492: Any potential downsides to setting this to "visible" in terms of cell content overflowing? Is this still needed even though we have zIndex 1?

src/js/components/DataTable/Resizer.js
```diff
@@ -8,40 +8,48 @@ import React, {
...
-const Resizer = ({ onResize, property, headerText, messages }) => {
+const Resizer = ({ onResize, property, headerText, messages, headerId }) => {
   const { theme } = useThemeValue();
   const [active, setActive] = useState(false);
   const [start, setStart] = useState();
   const [width, setWidth] = useState();
   const ref = useRef();
   const { format } = useContext(MessageContext);
 
+  useEffect(() => {
```
taysea, line 43: Could you explain why this is needed?

src/js/components/DataTable/Header.js
jcfilben: If I don't have the overflow visible set then the focus indicator gets cut off and the portion of the click/interaction target that is outside of the cell no longer works.
Screenshot 2025-07-23 at 1 23 53 PM

As far as any downsides to setting this to visible, I was initially thinking we might get different behavior when the column size is small. But here is what I observed on this branch vs on the master branch. I could have missed something but from what I can tell I think we are okay changing the overflow here.

This branch:
Screenshot 2025-07-23 at 1 31 41 PM

Master:
Screenshot 2025-07-23 at 1 31 25 PM

src/js/components/DataTable/Resizer.js
jcfilben: This sets an initial width once the component is rendered. Without this we only `setWidth` after we try to interact with the resizer

src/js/components/DataTable/Header.js
jcfilben: Yes, I think it makes sense to include that change here too

jcfilben: Just pushed up a commit that includes this change

src/js/components/DataTable/Resizer.js
jcfilben: Pushed up a commit that adds a comment to this useEffect

src/js/components/DataTable/Header.js
taysea: I'm comfortable with the condition for the resize case then 👍

src/js/components/DataTable/Resizer.js
taysea: Thanks !

taysea (approved):
LGTM

src/js/components/DataTable/Resizer.js
jcfilben: I went ahead and removed the decimal from the width to improve the screenreader experience. I left it as reading out pixel values. I couldn't find examples of what other libraries were doing here. Based on our discussion I felt like we shouldn't invest too much time here, so left it as is for now. We can re-visit if desired.

Issue #7721, opened 2025-07-22
Calendar — when no `onSelect` days should have aria-disabled="true"


taysea:
### Is this something you or your team is interested in contributing a pull request towards?

None

### Expected Behavior

When no `onSelect` days should have aria-disabled="true" so screen reader uses know they can't be selected.

### Actual Behavior

aria-disabled="false" is on the calendar days, which could be misleading to a screen reader.

PR #7724, opened 2025-07-24
set aria-disabled to true if onSelect prop is not provided

Sulaymon333:
#### What does this PR do?
- sets aria-disabled to true if onSelect prop is not provide in alignment with [Grommet documentation](https://v2.grommet.io/calendar#onSelect)

- https://github.com/grommet/grommet/issues/7721

Yes

Yes

Yes


Sulaymon333:
The question that comes to mind for this fix is that should we disable the actual day picker as well or not? For example via the code below

I am researching into other libraries to see the common pattern.

image

image

PR #7725, opened 2025-07-28
add fix for storybook stories

britt6612:
#### What does this PR do?
This PR fixes 2 storybook violations
on the following story https://storybook.grommet.io/?path=/story/input-form-typed-form--typed-form this is the error
```
ARIA attribute element ID does not exist on the page: aria-labelledby="grommet-ampm__label"
```

https://storybook.grommet.io/?path=/story/input-form-controlled-input-lazy--controlled-input-lazy
#age
Critical
```ARIA attribute value is ignored while empty: aria-valuenow```

forms/stories
storybook
storybook

read the error
none pre work for release 
no
no
backwards compatible

### CODE REVIEW

**Review by taysea — APPROVED** [review]

File: src/js/components/RangeInput/RangeInput.js [hunk 1]
```diff
@@ -102,7 +102,7 @@ const RangeInput = forwardRef(
         aria-label={a11yTitle}
         aria-valuemax={max}
         aria-valuemin={min}
-        aria-valuenow={value}
+        aria-valuenow={value !== undefined && value !== '' ? value : undefined}
```
jcfilben: If value is undefined, wouldn't `value !== undefined && value !== '' ? value : undefined` and `value !== '' ? value : undefined` result in the same behavior? Just wanting to understand if we actually need the `value !== undefined` check

britt6612: yes you are correct just simplified

jcfilben (approved):
Looks good!

PR #7727, opened 2025-07-29
[FIX] layer to restore focus to trigger onClose

Sulaymon333:
#### What does this PR do?
- fixes layer restore focus to trigger onClose

#### Do Jest tests follow these best practices?

- [ ] `screen` is used for querying.
- [ ] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [ ] `asFragment()` is used for snapshot testing.

#### Do the grommet docs need to be updated?
No

#### Should this PR be mentioned in the release notes?
Yes

#### Is this change backwards compatible or is it a breaking change?
Yes

PR #7729, opened 2025-07-31
add test for pressing escape on layer

britt6612:
#### What does this PR do?
adds a test for this PR https://github.com/grommet/grommet/pull/7727
layer test
local testing
jest test 

the focus should be placed back on the trigger button when the layer is closed
none
no
no
backwards compatible

### DIFF
--- a/src/js/components/Layer/Layer.js
+++ b/src/js/components/Layer/Layer.js
@@ -93,14 +93,17 @@ const Layer = forwardRef((props, ref) => {
           }
           setTimeout(() => {
             // we add the id and query here so the unit tests work
-            const clone = containerTarget
-              .getRootNode()
-              .getElementById('layerClone');
-            if (clone) {
-              if (containerTarget.contains(clone)) {
-                containerTarget.removeChild(clone);
+            const rootNode = containerTarget.getRootNode();
+            // Not all root nodes (ShadowRoot, DocumentFragment)
+            //  have getElementById.
+            if (rootNode && typeof rootNode.getElementById === 'function') {
+              const clone = rootNode.getElementById('layerClone');
+              if (clone) {
+                if (containerTarget.contains(clone)) {
+                  containerTarget.removeChild(clone);
+                }
+                layerContainer.remove();
               }
-              layerContainer.remove();
             }
           }, animationDuration);
         } else if (containerTarget.contains(layerContainer)) {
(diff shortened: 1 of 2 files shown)


src/js/components/Layer/Layer.js
```diff
@@ -93,14 +93,17 @@ const Layer = forwardRef((props, ref) => {
           }
           setTimeout(() => {
             // we add the id and query here so the unit tests work
-            const clone = containerTarget
-              .getRootNode()
-              .getElementById('layerClone');
-            if (clone) {
-              if (containerTarget.contains(clone)) {
-                containerTarget.removeChild(clone);
+            const rootNode = containerTarget.getRootNode();
```
britt6612, line 96: When the test closes the Layer, it triggers this cleanup code in the Layer component that assumes `getRootNode` always returns a Document. However, in the JSDOM test environment,  `getRootNode` can return something like a DocumentFragment, which doesn’t have a `getElementById` method.

britt6612: error I was getting prior to changing this

```
  TypeError: containerTarget.getRootNode(...).getElementById is not a function

       96 |             const clone = containerTarget
       97 |               .getRootNode()
    >  98 |               .getElementById('layerClone');
          |                ^
       99 |             if (clone) {
      100 |               if (containerTarget.contains(clone)) {
      101 |                 containerTarget.removeChild(clone);

      at getElementById (src/js/components/Layer/Layer.js:98:16)
```

taysea: Thank you for clarifying the change!

src/js/components/Layer/Layer.js
```diff
@@ -93,14 +93,17 @@ const Layer = forwardRef((props, ref) => {
...
+            // Not all root nodes (ShadowRoot, DocumentFragment)
```
taysea, line 97: It seems like we should enhance the Layer behavior to work in ShadowRoot: https://developer.mozilla.org/en-US/docs/Web/API/Element/shadowRoot but that can be done in separate work

taysea (approved):
Changes look good, but see comment re: improving for shadow DOM. Could you make a ticket to explore that? Currently the modal trap focus and onEsc behavior don't seem to work properly for shadow DOM

src/js/components/Layer/Layer.js
britt6612: yes agree I will make a ticket.

Sulaymon333 (approved):
Thank you for adding the test.

LGTM.

PR #7731, opened 2025-08-05
fix DataTable columns do not receive the hover state when onClickRow …

### LINKED ISSUE #4877: Harden Grommet Site static site build and deploy
This issue should probably live under the grommet-site repo, but that isn't apart of the project board yet. Filing here for visibility and prioritization.

- [ ] Confirm root cause of Grommet site failure when deploying static site 
- [ ] Make recommendations to harden build process

britt6612:
#### What does this PR do?
This PR ensures that when a column is pinned, it maintains the same background color as the rest of the row.
body.js
locally 

DataTable columns do not receive the hover state when onClickRow + pinned use the following to test
```
{
      property: 'name',
      pin: true,
      primary: true,
      header: 'Name',
...
}
<DataTable
    onClickRow={}
```

```
{
      property: 'name',
      pin: true,
      primary: true,
      header: 'Name',
...
}
<DataTable
    onClickRow={}
```

This was found while working on the DS example pages. 
closes #4877
no
yes
backwards compatible

### DIFF
--- a/src/js/components/DataTable/Body.js
+++ b/src/js/components/DataTable/Body.js
@@ -184,7 +184,8 @@ const Row = memo(
             key={column.property}
             background={
               (isSelected && cellProps.selected.background) ||
-              (column.pin && cellProps.pinned.background) ||
+              (active && cellProps.active?.background) ||
+              (focused && cellProps.focused?.background) ||
               cellProps.background
             }
             border={(column.pin && cellProps.pinned.border) || cellProps.border}
--- a/src/js/components/DataTable/__tests__/DataTable-test.tsx
+++ b/src/js/components/DataTable/__tests__/DataTable-test.tsx
@@ -924,6 +924,54 @@ describe('DataTable', () => {
     expect(onExpand.mock.results[0].value).toMatchSnapshot();
   });
 
+  // first row is pinned
+  const columns = [
+    {
+      property: 'name',
+      header: 'Name',
+      pin: true,
+    },
+    {
+      property: 'location',
+      header: 'Location',
+    },
+  ];
+
+  const data = [
+    { name: 'Jet', location: 'Palo Alto' },
+    { name: 'Michael', location: 'Boise' },
+  ];
+
+  test('pinned columns have same background as selected row', () => {
+    const { getByText } = render(
+      <Grommet>
+        <DataTable
+          columns={columns}
+          data={data}
+          rowProps={{
+            Jet: { background: 'background-selected-primary' },
+          }}
+        />
+      </Grommet>,
+    );
+
+    // Find the pinned cell and a regular cell in the selected row
+    const pinnedText = getByText('Jet');
+    const regularText = getByText('Palo Alto');
+
+    const pinnedCell = pinnedText.closest('td') || pinnedText.closest('th');
+    const regularCell = regularText.closest('td') || regularText.closest('th');
+
+    expect(pinnedCell).not.toBeNull();
+    expect(regularCell).not.toBeNull();
+
+    const pinnedBg = window.getComputedStyle(pinnedCell!).backgroundColor;
+    const regularBg = window.getComputedStyle(regularCell!).backgroundColor;
+
+    // Assert that both backgrounds are the same
+    expect(pinnedBg).toBe(regularBg);
+  });
+
   test('replace', () => {
     const { container } = render(
       <Grommet>
--- a/src/js/components/DataTable/__tests__/__snapshots__/DataTable-test.tsx.snap
+++ b/src/js/components/DataTable/__tests__/__snapshots__/DataTable-test.tsx.snap
@@ -20039,26 +20039,13 @@ exports[`DataTable pin + background 1`] = `
   font-weight: inherit;
   text-align: inherit;
   text-align: start;
-  background: red;
   padding-left: 12px;
   padding-right: 12px;
   padding-top: 6px;
   padding-bottom: 6px;
 }
 
-.c14 {
-  margin: 0;
-  padding: 0;
-  font-weight: inherit;
-  text-align: inherit;
-  text-align: start;
-  padding-left: 12px;
-  padding-right: 12px;
-  padding-top: 6px;
-  padding-bottom: 6px;
-}
-
-.c16 {
+.c15 {
   margin: 0;
   padding: 0;
   font-weight: inherit;
@@ -20073,7 +20060,7 @@ exports[`DataTable pin + background 1`] = `
   padding-bottom: 6px;
 }
 
-.c19 {
+.c18 {
   margin: 0;
   padding: 0;
   font-weight: inherit;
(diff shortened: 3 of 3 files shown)


britt6612:
current focus for `hpe` theme this is because we are setting `backdrop-filter: blur(12px);`

Screenshot 2025-08-06 at 2 29 36 PM

taysea:
seems like we're getting a styling regression on the grommet theme example:

BEFORE
Screenshot 2025-08-06 at 5 17 16 PM

AFTER
Screenshot 2025-08-06 at 5 17 13 PM

britt6612:
> seems like we're getting a styling regression on the grommet theme example:
> 
> BEFORE Screenshot 2025-08-06 at 5 17 16 PM
> 
> AFTER Screenshot 2025-08-06 at 5 17 13 PM

good catch ill look into that!

src/js/components/DataTable/stories/Controlled.stories.js
```diff
@@ -14,6 +14,7 @@ delete controlledColumns[4].aggregate;
 
 export const ControlledDataTable = () => {
   const [checked, setChecked] = React.useState([]);
+  const [selectedRow, setSelectedRow] = React.useState();
```
britt6612, line 17: TODO 
revert story back just for demo on PR

src/js/components/DataTable/stories/data.js
```diff
@@ -14,6 +14,7 @@ export const columns = [
     header: <Text>Name with extra</Text>,
     primary: true,
     footer: 'Total',
+    pin: true,
```
britt6612, line 17: TODO revert back

britt6612:
I have added `rowProps={selectedRow ? { [selectedRow]: { background: 'blue' } } : {}}` in the story just for us to visually see and discuss. For this Im wondering if we need another theme value? or should we just match to the rest of the row. @taysea @jcfilben 

I wonder though if we match it automatically to the rest of the row would there be a use case someone wants it to stand out? or be different?

in the example as well Im passing a prop `rowProps` so think this should take precedence over anything

Screenshot 2025-08-13 at 2 34 07 PM

src/js/themes/base.js
```diff
@@ -920,10 +920,13 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       //   selected: undefined,
       // },
       pinned: {
-        // body: {
-        //    background: undefined,
-        //    extend: undefined,
-        // },
+        body: {
+          // hover: {
+          //   background: undefined,
```
britt6612: add color here to visually see the difference 

@jcfilben @taysea

src/js/components/DataTable/Body.js
```diff
@@ -43,177 +43,190 @@ const Row = memo(
...
+        >
+          {(selected || onSelect) && (
+            <Cell
+              background={
+                (isSelected && cellProps.selected.background) ||
+                (active &&
+                  columns.pin &&
+                  theme.dataTable?.pinned?.body?.hover?.background) ||
+                (active && cellProps.active?.background) ||
+                (columns.pin && cellProps.pinned.background) ||
```
jcfilben: Previously we were checking for `pinnedOffset` I'm just wondering why we don't check for that anymore

jcfilben:
> I have added rowProps={selectedRow ? { [selectedRow]: { background: 'blue' } } : {}} in the story just for us to visually see and discuss. For this Im wondering if we need another theme value? or should we just match to the rest of the row. @taysea @jcfilben
> 
> I wonder though if we match it automatically to the rest of the row would there be a use case someone wants it to stand out? or be different?
> 
> in the example as well Im passing a prop rowProps so think this should take precedence over anything

I was expecting the selected style to take precedence over the pinned styled because of this code here:
```
background={
  (isSelected && cellProps.selected.background) ||
  (active &&
    columns.pin &&
    theme.dataTable?.pinned?.body?.hover?.background) ||
  (active && cellProps.active?.background) ||
  (columns.pin && cellProps.pinned.background) ||
  cellProps.background
}
```

If we set a background on `rowProps` does this not get passed down to `cellProps`?

britt6612:
> > I have added rowProps={selectedRow ? { [selectedRow]: { background: 'blue' } } : {}} in the story just for us to visually see and discuss. For this Im wondering if we need another theme value? or should we just match to the rest of the row. @taysea @jcfilben
> > I wonder though if we match it automatically to the rest of the row would there be a use case someone wants it to stand out? or be different?
> > in the example as well Im passing a prop rowProps so think this should take precedence over anything
> 
> I was expecting the selected style to take precedence over the pinned styled because of this code here:
> 
> ```
> background={
>   (isSelected && cellProps.selected.background) ||
>   (active &&
>     columns.pin &&
>     theme.dataTable?.pinned?.body?.hover?.background) ||
>   (active && cellProps.active?.background) ||
>   (columns.pin && cellProps.pinned.background) ||
>   cellProps.background
> }
> ```
> 
> If we set a background on `rowProps` does this not get passed down to `cellProps`?


You're correct that the selected style should take precedence when `isSelected` is true. However, in the case where we're using onClickRow (like in the controlled story), the `isSelected` is not triggered because we're not using the built-in selection props `selected` `onSelect`
Instead, the background is set via `rowProps` which gets merged into cellProps.background. Since isSelected is undefined in this scenario, the background logic falls through to the last line and uses `cellProps.background`

src/js/components/DataTable/Body.js
britt6612: ah yes for this background I can revert back to checking the `pinnedOffset` then in the background where we are checking column.pin add there

src/js/components/DataTable/Body.js
```diff
@@ -43,177 +43,198 @@ const Row = memo(
...
-            }}
-            pad={cellProps.pad}
-            verticalAlign={verticalAlign}
-          />
-        )}
-        {columns.map((column) => (
-          <Cell
-            key={column.property}
-            background={
+              // should this background also be the same ??
```
britt6612: this background is this part of the DataTable, should we also account for the background here?

@taysea @jcfilben 

Screenshot 2025-08-18 at 10 03 09 PM

src/js/themes/base.js
```diff
@@ -920,10 +920,13 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       //   selected: undefined,
       // },
       pinned: {
-        // body: {
-        //    background: undefined,
-        //    extend: undefined,
-        // },
+        body: {
+          hover: {
+            background: '#89CFF0',
```
britt6612, line 925: TODO just here for demo

src/js/components/DataTable/Body.js
jcfilben: Yeah it seems like we should also get the blue background there too

britt6612: okay thats fine just wanted to confirm so Ill change that logic as well

britt6612: @jcfilben okay I have made the selected background the same for this case!

PR #7591, opened 2025-04-16
Add comments for places that are missing theme object for box props

taysea:
#### What does this PR do?

In a future major version of grommet-theme-hpe, we intend to expand/alter the values for edgeSize, size, etc. and we will need to have access to re-assign tshirt sizes for each of the commented styles.

#### Do Jest tests follow these best practices?

- [ ] `screen` is used for querying.
- [ ] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [ ] `asFragment()` is used for snapshot testing.

PR #7732, opened 2025-08-06
Add theme objects for Accordion

jcfilben:
#### What does this PR do?
Adds theme objects for the accordion component. Based on changes in https://github.com/grommet/grommet/pull/7591

Yes
Yes
Backwards compatible

### DIFF
--- a/src/js/components/AccordionPanel/AccordionPanel.js
+++ b/src/js/components/AccordionPanel/AccordionPanel.js
@@ -125,7 +125,7 @@ const AccordionPanel = forwardRef(
               {...rest}
             >
               {typeof label === 'string' ? (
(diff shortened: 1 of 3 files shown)


britt6612 (approved):
LGTM

One question I know for our theme we document `default` values Im wonder though since these were never documented in grommet docs since they were internal should we now document in grommet site as new theme value as well as the default?

jcfilben:
> One question I know for our theme we document default values Im wonder though since these were never documented in grommet docs since they were internal should we now document in grommet site as new theme value as well as the default?

Good question, I was thinking we would add them to the documentation. But if we want to keep these internal only for now and not add them I would be open to that. I think we could chat more as a dev team about this

PR #7733, opened 2025-08-06
Add theme objects for Box

jcfilben:
#### What does this PR do?
Adds theme objects for the Box component. Based on changes in https://github.com/grommet/grommet/pull/7591

I changed `theme.box.borderSize` to `theme.box.border.size` because I felt that it would give us more flexibility down the road if we ever wanted to add more theme support for box border beyond just size. Just a soft opinion, if we want I would be fine with switching back to `borderSize`

Yes
Yes
Backwards compatible

### DIFF
--- a/src/js/components/AccordionPanel/AccordionPanel.js
+++ b/src/js/components/AccordionPanel/AccordionPanel.js
@@ -125,7 +125,7 @@ const AccordionPanel = forwardRef(
               {...rest}
             >
               {typeof label === 'string' ? (
-                <Box pad={{ horizontal: 'xsmall' }}>
+                <Box pad={theme.accordion.label?.container?.pad}>
                   <Heading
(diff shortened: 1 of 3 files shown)


src/js/components/Box/StyledBox.js
```diff
@@ -399,7 +399,7 @@ const gapGapStyle = (directionProp, gap, responsive, border, theme) => {
   }
 
   if (border === 'between' || (border && border.side === 'between')) {
-    const borderSize = border.size || 'xsmall';
+    const borderSize = border.size || theme.box.border.size;
```
britt6612, line 402: I think this make sense!

britt6612 (approved):

taysea (approved):
I like the openness for future additions

PR #7734, opened 2025-08-06
Add Calendar theme objects

jcfilben:
#### What does this PR do?
Adds theme objects for the Calendar component. Based on changes in https://github.com/grommet/grommet/pull/7591

### DIFF
--- a/src/js/components/Calendar/Calendar.js
+++ b/src/js/components/Calendar/Calendar.js
@@ -44,12 +44,6 @@ import {
 import { setHoursWithOffset } from '../../utils/dates';
 import { useThemeValue } from '../../utils/useThemeValue';
 
-const headingPadMap = {
-  small: 'xsmall',
-  medium: 'small',
-  large: 'medium',
-};
-
 const getLocaleString = (value, locale) =>
   value?.toLocaleDateString(locale, {
     month: 'long',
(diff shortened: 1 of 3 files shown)


src/js/components/Calendar/Calendar.js
```diff
@@ -684,7 +678,7 @@ const Calendar = forwardRef(
 
       return (
         <Box direction="row" justify="between" align="center">
-          <Header flex pad={{ horizontal: headingPadMap[size] || 'small' }}>
+          <Header flex pad={theme.calendar?.[size]?.header.container.pad}>
```
taysea: I did this one I think, but commenting on my own but I'm wondering if we call this theme.calendar?.[size]?.title.container.pad?

Since technically this isn't swapping out the whole header or even wrapping around the title + before after buttons. It's just around the title.

jcfilben: I think that naming makes sense but on line 683 we are spreading `{...theme.calendar[size].title}` onto the Text component, so I feel like that namespace is already reserved

taysea: hmm good point -- this might be something we run into with places we're introducing the "container" concept -- where we can't just spread all the parent's object, but might need to do a bit of logic to pull the "container" subobject out of the larger theme object.

So that:
```
const { containerTheme, ...textTheme } = theme.calendar?.[size]?.title;
```

Then use those objects as needed?

jcfilben: Updated to use `theme.calendar?.[size]?.title` in latest commit

taysea (approved):

PR #7739, opened 2025-08-06
add DataSearch and DataSort theme objects

jcfilben:
#### What does this PR do?
Adds theme objects for the DataSearch and DataSort component. Based on changes in https://github.com/grommet/grommet/pull/7591

### DIFF
--- a/src/js/components/DataSearch/DataSearch.js
+++ b/src/js/components/DataSearch/DataSearch.js
@@ -128,7 +128,7 @@ export const DataSearch = ({
       kind={theme.data.button?.kind}
       icon={<Search />}
       dropProps={dropProps}
-      dropContent={<Box {...theme.data?.drop}>{content}</Box>}
+      dropContent={<Box pad={theme.data?.drop?.pad}>{content}</Box>}
       open={showContent}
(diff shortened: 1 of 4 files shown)


src/js/components/DataSearch/DataSearch.js
```diff
@@ -128,7 +128,7 @@ export const DataSearch = ({
       kind={theme.data.button?.kind}
       icon={<Search />}
       dropProps={dropProps}
-      dropContent={<Box pad="small">{content}</Box>}
+      dropContent={<Box {...theme.data?.drop}>{content}</Box>}
```
taysea: @jcfilben How do you feel about spreading all box props here vs only supporting pad to start?

I think I'd done this one, but just commenting now based on our discussion with Brittany.

britt6612: oh yeah I forgot I left this one I think we only wanted `pad`?

jcfilben: Oops yes, I think we only wanted pad. I'll update that

jcfilben: Updated in latest commit

PR #7740, opened 2025-08-07
add theme object for textinput

britt6612:
#### What does this PR do?
Adds theme objects for the TextInput component.
textinput.js

yes
yes
backwards compatible

### DIFF
--- a/src/js/components/TextInput/TextInput.js
+++ b/src/js/components/TextInput/TextInput.js
@@ -392,7 +392,10 @@ const TextInput = forwardRef(
                   else if (!theme.button.option)
                     // don't have theme support, need to layout here
                     child = (
-                      <Box align="start" pad="small">
+                      <Box
+                        align="start"
+                        pad={theme.textInput.suggestions?.container?.pad}
+                      >
                         {renderedLabel}
                       </Box>
                     );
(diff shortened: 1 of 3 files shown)


src/js/components/TextInput/TextInput.js
```diff
@@ -392,7 +392,10 @@ const TextInput = forwardRef(
                   else if (!theme.button.option)
                     // don't have theme support, need to layout here
                     child = (
-                      <Box align="start" pad="small">
+                      <Box
+                        align="start"
+                        pad={theme.textInput.suggestions?.container?.pad}
```
jcfilben: It looks like we only use this theme prop when `theme.button.option` isn't defined.

The name `theme.textInput.suggestions?.container?.pad` makes me think that this would be applied to all suggestions in the TextInput. So maybe we want a more specific name to indicate that this is only applied sometimes?

Not sure what name we want to use here, this is a pretty narrow case when we use this padding

britt6612: yeah open to naming. I just wanted to put something as a start. I also was not exactly sure how to make it clear this is when `theme.button.option` isnt defined

britt6612: add a comment that we are not going to add a theme object now since hpe theme it is not needed but future we may want

PR #7741, opened 2025-08-07
add tab theme object

britt6612:
#### What does this PR do?
Adds theme objects for the Tabs/Tab component.
tabs.js
tab.js

yes
yes
backwards compatible

### DIFF
--- a/src/js/components/Tabs/Tabs.js
+++ b/src/js/components/Tabs/Tabs.js
@@ -377,7 +377,7 @@ const Tabs = forwardRef(
               tabIndex={-1}
               onClick={() => moveByArrowKey('next')}
             >
-              <Box pad={theme.tabs.header?.overflowNavigation?.pad}>
+              <Box pad={theme.tabs.header?.nextTabButton?.pad}>
                 <Next
                   color={
                     disableRightArrow
--- a/src/js/components/Tabs/Tabs.js
+++ b/src/js/components/Tabs/Tabs.js
@@ -377,7 +377,7 @@ const Tabs = forwardRef(
               tabIndex={-1}
               onClick={() => moveByArrowKey('next')}
             >
-              <Box pad={theme.tabs.header?.nextTabButton?.pad}>
+              <Box pad={theme.tabs.header?.nextButton?.pad}>
                 <Next
                   color={
                     disableRightArrow
--- a/src/js/components/Tabs/Tabs.js
+++ b/src/js/components/Tabs/Tabs.js
@@ -377,7 +377,7 @@ const Tabs = forwardRef(
               tabIndex={-1}
               onClick={() => moveByArrowKey('next')}
             >
-              <Box pad={{ vertical: 'xsmall', horizontal: 'small' }}>
+              <Box pad={theme.tabs.header?.overflowNavigation?.pad}>
                 <Next
                   color={
                     disableRightArrow
--- a/src/js/components/Tabs/Tabs.js
+++ b/src/js/components/Tabs/Tabs.js
@@ -342,7 +342,7 @@ const Tabs = forwardRef(
               tabIndex={-1}
               onClick={() => moveByArrowKey('previous')}
             >
-              <Box pad={{ vertical: 'xsmall', horizontal: 'small' }}>
+              <Box pad={theme.tabs.header.previousButton.pad}>
                 <Previous
                   color={
                     disableLeftArrow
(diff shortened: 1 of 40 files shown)


src/js/components/Tabs/Tabs.js
```diff
@@ -377,7 +377,7 @@ const Tabs = forwardRef(
               tabIndex={-1}
               onClick={() => moveByArrowKey('next')}
             >
-              <Box pad={{ vertical: 'xsmall', horizontal: 'small' }}>
+              <Box pad={theme.tabs.header?.overflowNavigation?.pad}>
```
jcfilben: I wonder if we change `overflowNavigation` to `nextTabButton`? Not sure if that feels more clear

britt6612: oh good thinking!

src/js/themes/base.js
```diff
@@ -1967,6 +1967,7 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       //   color: undefined,
       // },
       // extend: undefined,
+      gap: 'small',
```
jcfilben, line 2079: Why did we add `gap`?

britt6612: for the `tab` change, sorry I think those changes did not come through they are there now

src/js/components/Tabs/Tabs.js
```diff
@@ -377,7 +377,7 @@ const Tabs = forwardRef(
               tabIndex={-1}
               onClick={() => moveByArrowKey('next')}
             >
-              <Box pad={{ vertical: 'xsmall', horizontal: 'small' }}>
+              <Box pad={theme.tabs.header?.nextTabButton?.pad}>
```
taysea: Soft opinion on naming - since this is already under "tabs" object, do we need to include that in the name? Ideas:
- `tabs.header.next.pad`
- `tabs.header.nextButton.pad`

britt6612: good point I like the nextButton for clarity

britt6612: changed naming.

jcfilben: Agree, I like nextButton

jcfilben (approved):

src/js/components/Tabs/Tabs.js
```diff
@@ -377,7 +377,7 @@ const Tabs = forwardRef(
               tabIndex={-1}
               onClick={() => moveByArrowKey('next')}
             >
-              <Box pad={{ vertical: 'xsmall', horizontal: 'small' }}>
+              <Box pad={theme.tabs.header?.nextButton?.pad}>
```
taysea, line 380: Realizing I missed a spot to add a comment for "previous" button. Could we either:
1. Add `theme.tabs.header.previousButton.pad` OR
2. Rename to a single theme object and have `theme.tabs.header.overflowButtons.pad`

britt6612: good catch ill update

src/js/components/Tabs/__tests__/Tabs-test.tsx
```diff
@@ -369,4 +369,52 @@ describe('Tabs', () => {
...
+    // Find previous and next tab control buttons by aria-label
+    const previousButton = screen.queryByLabelText('Previous Tab');
+    const nextButton = screen.queryByLabelText('Next Tab');
+
+    // Assert pad values on the Box inside the button
+    if (previousButton) {
+      const previousBox = previousButton.querySelector('div');
+      const previousBoxStyle =
+        previousBox && window.getComputedStyle(previousBox);
+      expect(previousBoxStyle?.paddingLeft).not.toBe('');
```
jcfilben: Is it possible to expect a certain value here rather than just expecting the styles not to be empty?
If it is tricky to do that I feel like this is a case were we could do snapshot testing

src/js/themes/base.js
```diff
@@ -1997,6 +1998,12 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         //   color: undefined,
         // },
         // extend: undefined,
+        nextButton: {
+          pad: { horizontal: 'small', vertical: 'xsmall' },
+        },
+        previousButton: {
+          pad: { vertical: 'xsmall', horizontal: 'small' },
+        },
```
jcfilben:
```suggestion
        nextButton: {
          pad: { horizontal: 'small', vertical: 'xsmall' },
        },
        previousButton: {
          pad: { horizontal: 'small', vertical: 'xsmall' },
        },
```
Super minor comment but makes it a little easier to read

Also wondering if we should consolidate into one object just to avoid adding extra things into the theme?

britt6612: hmm Id be open to one object @taysea objections here?

src/js/components/Tabs/__tests__/Tabs-test.tsx
britt6612: let me try!

jcfilben (approved):

PR #7742, opened 2025-08-07
add theme object rangeselector

britt6612:
#### What does this PR do?
Adds theme objects for the rangeselector component.
rangeselector.js

yes
yes
backwards compatible

### DIFF
--- a/src/js/components/RangeSelector/RangeSelector.js
+++ b/src/js/components/RangeSelector/RangeSelector.js
@@ -375,13 +375,18 @@ const RangeSelector = forwardRef(
           <Text
             ref={minRef}
             textAlign="end"
-            size="small"
-            margin={{ horizontal: 'small' }}
+            // do we need theme for size?
+            size={theme.rangeSelector.label.size}
+            margin={theme.rangeSelector.label.margin}
           >
             {typeof label === 'function' ? label(lower) : lower}
           </Text>
           {content}
-          <Text ref={maxRef} size="small" margin={{ horizontal: 'small' }}>
+          <Text
+            ref={maxRef}
+            size={theme.rangeSelector.label.size}
+            margin={theme.rangeSelector.label.margin}
+          >
             {typeof label === 'function' ? label(upper) : upper}
           </Text>
         </Box>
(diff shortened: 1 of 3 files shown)


src/js/components/RangeSelector/RangeSelector.js
```diff
@@ -375,13 +375,18 @@ const RangeSelector = forwardRef(
           <Text
             ref={minRef}
             textAlign="end"
-            size="small"
-            margin={{ horizontal: 'small' }}
+            // do we need theme for size?
```
jcfilben: Prop names look good, but I have the same question. Do we need to have a theme prop for size?

britt6612: yeah @taysea thoughts?

taysea: Good question -- I don't think text size is required in the scope of "breaking t-shirt sizes" (since we're more worried about margin/pad/gap.

I'd be fine if we want to leave as just the margin one for now, unless you think adding text is useful for flexibility.

britt6612: okay I can remove `size` I think its fine if we focus on just the `margin` for now not sure is the size is not as needed not add it at this moment

src/js/components/RangeSelector/RangeSelector.js
```diff
@@ -376,12 +376,16 @@ const RangeSelector = forwardRef(
...
+            margin={theme.rangeSelector.label.margin}
           >
             {typeof label === 'function' ? label(lower) : lower}
           </Text>
           {content}
-          <Text ref={maxRef} size="small" margin={{ horizontal: 'small' }}>
+          <Text
+            ref={maxRef}
+            size="small"
+            margin={theme.rangeSelector.label.margin}
```
britt6612, line 387: wondering if these names should actually be different from one another

britt6612: @taysea

britt6612: going to keep as the same we chatted through this we can expand in future if needed

PR #7744, opened 2025-08-07
add theme object radiobutton radiobuttongroup

britt6612:
#### What does this PR do?
Adds theme objects for the RadioButton and RadioButtonGroup component. Based on changes in https://github.com/grommet/grommet/pull/7591

RadioButton.js
RadioButtonGroup.js

yes
yes
backwards compatible

### DIFF
--- a/src/js/components/RadioButton/RadioButton.js
+++ b/src/js/components/RadioButton/RadioButton.js
@@ -82,7 +82,12 @@ const RadioButton = forwardRef(
         <StyledRadioButton
           flex={false}
           margin={
-            label ? { right: theme.radioButton.gap || 'small' } : undefined
+            label
+              ? {
+                  right:
+                    theme.radioButton.gap || theme.radioButton.labelMarginRight,
+                }
+              : undefined
           }
           {...passThemeFlag}
         >
--- a/src/js/components/RadioButton/RadioButton.js
+++ b/src/js/components/RadioButton/RadioButton.js
@@ -81,7 +81,7 @@ const RadioButton = forwardRef(
       >
         <StyledRadioButton
           flex={false}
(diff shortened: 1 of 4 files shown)


src/js/components/RadioButton/RadioButton.js
```diff
@@ -82,7 +82,12 @@ const RadioButton = forwardRef(
         <StyledRadioButton
           flex={false}
           margin={
-            label ? { right: theme.radioButton.gap || 'small' } : undefined
+            label
+              ? {
+                  right:
+                    theme.radioButton.gap || theme.radioButton.labelMarginRight,
+                }
+              : undefined
```
jcfilben: Could we change the theme object so that it is just `margin` and then in the base theme we just pass `margin={{ right: 'small' }}`?

britt6612: yes

jcfilben (approved):

src/js/components/RadioButton/RadioButton.js
```diff
@@ -81,9 +81,7 @@ const RadioButton = forwardRef(
       >
         <StyledRadioButton
           flex={false}
-          margin={
-            label ? { right: theme.radioButton.gap || 'small' } : undefined
-          }
+          margin={label ? theme.radioButton.label?.margin : undefined}
```
taysea: Changing this to move the "right" into base.js seems like it could have breaking changes for themes that just defined a string (expecting it to behave just on the right) and we're also changing an existing theme object name.

Feels like we should keep as is with "gap" for backwards compatibility and apply it on the right as it is.

Thoughts? @jcfilben @britt6612

jcfilben (approved):

taysea (approved):

PR #7746, opened 2025-08-07
add notification theme object

britt6612:
#### What does this PR do?
Adds theme objects for the Notification component.
notification.js

yes
yes
backwards compatible

### DIFF
--- a/src/js/components/Notification/Notification.js
+++ b/src/js/components/Notification/Notification.js
@@ -194,7 +194,7 @@ const Notification = ({
         <NotificationAnchor
           // create space between first anchor and
           // text content and next anchor
-          margin={{ right: 'xsmall' }}
+          margin={{ right: theme.notification.actions?.marginRight }}
           {...action}
           {...theme.notification.actions}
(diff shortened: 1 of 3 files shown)


src/js/components/Notification/Notification.js
```diff
@@ -210,7 +210,11 @@ const Notification = ({
           color={messageColor}
           direction={direction}
         >
-          <Text margin={{ right: 'xsmall' }}>{message}</Text>
+          <Text
+            margin={{ right: theme.notification.message?.text?.marginRight }}
```
britt6612: open to discussion on naming

MikeKingdom: I would argue this shouldn't specifically be doing just right margin here. How about the theme object is simply margin and it's default value is `{ right: 'xsmall' }`?

britt6612: yes thats fair I can change

jcfilben (approved):

MikeKingdom (approved):
Thanks!

PR #7748, opened 2025-08-07
add menu theme object

britt6612:
#### What does this PR do?
Adds theme objects for the Menu component.
menu.js

yes
yes
backwards compatible

### DIFF
--- a/src/js/components/Menu/Menu.js
+++ b/src/js/components/Menu/Menu.js
@@ -251,9 +251,11 @@ const Menu = forwardRef((props, ref) => {
   if (children) {
     content = children;
   } else if (!theme.button.default) {
-    // we are not going to change pad to a theme object
-    // since this will not be rendered in the hpe theme
-    // since we have theme.button.default
+    /*
+    Not adding a theme object now because this code path
+    is not used in the HPE theme, but we may add theme
+    support here in the future.
+    */
     content = (
       <Box
         direction="row"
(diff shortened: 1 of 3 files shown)


src/js/components/Menu/Menu.js
```diff
@@ -309,7 +309,7 @@ const Menu = forwardRef((props, ref) => {
     const child = !theme.button.option ? (
       <Box
         align={theme.menu.item?.align || 'start'}
-        pad="small"
+        pad={theme.menu.item?.pad}
```
jcfilben: Looks like this only gets applied when `theme.button.option` isn't defined, we may need to think about how we can communicate that in the prop name

britt6612: yes we can!

src/js/components/Menu/Menu.js
```diff
@@ -256,8 +256,8 @@ const Menu = forwardRef((props, ref) => {
         direction="row"
         justify={justifyContent}
         align="center"
-        pad="small"
-        gap={label && icon !== false ? 'small' : undefined}
+        pad={theme.menu.label?.pad}
```
taysea: Similar to the below content, thinking about how to better capture what this is. "label" is a subelement inside this container (this container contains label + icon). This really is the control/button that someone is clicking on. Not sure which of these names but some ideas:
- `menu.control.pad` / `menu.control.gap`
- `menu.button.pad` / `menu.button.gap` (although a little odd because it's not actually applying to a button`

src/js/components/Menu/Menu.js
```diff
@@ -251,6 +251,9 @@ const Menu = forwardRef((props, ref) => {
   if (children) {
     content = children;
   } else if (!theme.button.default) {
+    // we are not going to change pad to a theme object
```
britt6612: @jcfilben @taysea 

thoughts on wording?

jcfilben (approved):

taysea (approved):

PR #7749, opened 2025-08-07
Add DataSummary theme objects

jcfilben:
#### What does this PR do?
Adds theme objects for the DataSummary component. Based on changes in https://github.com/grommet/grommet/pull/7591

### DIFF
--- a/src/js/components/DataSummary/DataSummary.js
+++ b/src/js/components/DataSummary/DataSummary.js
@@ -12,7 +12,7 @@ export const DataSummary = ({ messages, ...rest }) => {
     messages: dataMessages,
     selected,
     total,
-  } = useContext(DataContext)
+  } = useContext(DataContext);
   const { theme } = useThemeValue();
 
   let messageId;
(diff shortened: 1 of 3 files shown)


britt6612 (approved):

src/js/components/DataSummary/DataSummary.js
```diff
@@ -28,7 +29,7 @@ export const DataSummary = ({ messages, ...rest }) => {
   });
 
   return (
-    <Text margin={{ vertical: 'xsmall' }} {...rest}>
+    <Text margin={theme.dataSummary?.container?.margin} {...rest}>
```
taysea: Naming convention -- does "container" feel reasonable here? It's technically just Text. Just testing our boundaries for what we need when we say "container"

britt6612: thats a good point. 🤔 wonder if we just leave it as text

jcfilben: Good call, removed the container wording in latest commit

PR #7750, opened 2025-08-07
Add DataTableColumns theme objects

jcfilben:
#### What does this PR do?
Adds theme objects for the DataTableColumns component. Based on changes in https://github.com/grommet/grommet/pull/7591

Note I removed `theme.dataTableColumns.tabs.justify` because i felt like we didn't need to add the justify prop to the theme for the t-shirt size changes

### DIFF
--- a/src/js/components/DataTableColumns/DataTableColumns.js
+++ b/src/js/components/DataTableColumns/DataTableColumns.js
@@ -21,11 +21,6 @@ const dropProps = {
   align: { top: 'bottom', left: 'left' },
 };
 
-const tabsProps = {
-  drop: { pad: 'small' },
-  noDrop: { justify: 'start' },
-};
-
 // options can either be an array of property names or an array of objects.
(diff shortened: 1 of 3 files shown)


src/js/components/DataTableColumns/DataTableColumns.js
```diff
@@ -52,6 +47,12 @@ const Content = ({ drop, options = [], ...rest }) => {
   const { id: dataId, messages } = useContext(DataContext);
   const { useFormInput } = useContext(FormContext);
   const { format } = useContext(MessageContext);
+  const { theme } = useThemeValue();
+
+  const tabsProps = {
```
taysea, line 52: This looks good, but where was tabsProps defined before? Cause it seems it was already used but can't find where it was defined.

jcfilben: Previously it was on line 24 of DataTableColumns.js

PR #7751, opened 2025-08-07
Add DateInput theme objects

jcfilben:
#### What does this PR do?
Adds theme objects for the DateInput component. Based on changes in https://github.com/grommet/grommet/pull/7591

### DIFF
--- a/src/js/components/DateInput/DateInput.js
+++ b/src/js/components/DateInput/DateInput.js
@@ -91,9 +91,7 @@ const DateInput = forwardRef(
     const announce = useContext(AnnounceContext);
     const { format: formatMessage } = useContext(MessageContext);
     const iconSize =
-      (theme.icon?.matchSize && rest.size) ||
-      theme.dateInput.icon?.size ||
-      'medium';
+      (theme.icon?.matchSize && rest.size) || theme.dateInput.icon?.size;
     const { useFormInput } = useContext(FormContext);
     const ref = useForwardedRef(refArg);
     const containerRef = useRef();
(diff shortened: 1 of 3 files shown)


src/js/themes/base.d.ts
```diff
@@ -840,6 +840,9 @@ export interface ThemeType {
     };
   };
   dateInput?: {
+    button?: {
+      margin?: MarginType;
```
taysea: Technically we're only supporting "left" and "right" at the moment right?

We won't cause any errors with this, so it's okay but if we wanted to be more specific we should probably break it down

jcfilben: Changed the type to just support passing a left and right margin

src/js/components/DateInput/DateInput.js
```diff
@@ -303,7 +301,11 @@ Use the icon prop instead.`,
         disabled={disabled}
         plain
         icon={icon || MaskedInputIcon || <CalendarIcon size={iconSize} />}
-        margin={reverse ? { left: 'small' } : { right: 'small' }}
+        margin={
+          reverse
```
taysea, line 305: Just thinking about how we'll document -- it seems like typically the value is the same and it's just the side that changes based on if `reverse` is used.

Do we want to have 2 different theme properties "left" and "right"? Or just a single one to start, then could always extend to support an object in future.

Thinking about it from the consumer perspective, I feel like it's unlikely that we'd be defining something different for when it's reverse vs not.

taysea: Although, soft opinion.

jcfilben: I'm open to just having one value for now

jcfilben: Changed this in latest commit

PR #7754, opened 2025-08-07
add list theme object

britt6612:
#### What does this PR do?
Adds theme objects for the List component.
list.js

yes
yes
backwards compatible

### DIFF
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -838,6 +838,18 @@ export interface ThemeType {
     button?: {
       kind?: string;
     };
+    drop?: {
+      pad?: PadType;
+    };
+  };
+  dataFilter?: {
+    rangeSelector?: {
+      size?: string;
+      round?: string;
+    };
+    selectMultiple?: {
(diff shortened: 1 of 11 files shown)


jcfilben (approved):

taysea (approved):

src/js/themes/base.d.ts
```diff
@@ -1256,6 +1256,7 @@ export interface ThemeType {
         color?: ColorType;
         cursor?: string;
       };
+      gap: GapType;
```
ShimiSun, line 1259: @jcfilben Is there a reason this isn’t an optional field? This change wasn’t backward compatible, and we had to update the themes for the upgrade to apply.

jcfilben: Hey @ShimiSun! This should be optional, PR to fix this here: https://github.com/grommet/grommet/pull/7857

ShimiSun: Appreciate the follow-up 🙏

PR #7756, opened 2025-08-07
Add SelectMultiple theme objects

jcfilben:
#### What does this PR do?
Adds theme objects for the SelectMultiple component. Based on changes in https://github.com/grommet/grommet/pull/7591

### DIFF
--- a/src/js/components/SelectMultiple/SelectMultipleContainer.js
+++ b/src/js/components/SelectMultiple/SelectMultipleContainer.js
@@ -365,7 +365,7 @@ const SelectMultipleContainer = forwardRef(
     if (help) {
       if (typeof help === 'string')
         helpContent = (
-          <Box flex={false} pad="xsmall">
+          <Box flex={false} pad={theme.selectMultiple?.helpText?.pad}>
             <Text size="small">{help}</Text>
           </Box>
         );
@@ -404,7 +404,14 @@ const SelectMultipleContainer = forwardRef(
         >
           {summaryContent}
           {onSearch && (
-            <Box pad={!customSearchInput ? 'xsmall' : undefined} flex={false}>
+            <Box
+              pad={
+                !customSearchInput
+                  ? theme.selectMultiple?.search?.pad
+                  : undefined
+              }
+              flex={false}
+            >
               <Keyboard
                 onEnter={(event) => {
                   onNextOption(event);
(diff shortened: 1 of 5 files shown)


src/js/components/SelectMultiple/SelectMultipleContainer.js
```diff
@@ -365,7 +365,7 @@ const SelectMultipleContainer = forwardRef(
     if (help) {
       if (typeof help === 'string')
         helpContent = (
-          <Box flex={false} pad="xsmall">
+          <Box flex={false} pad={theme.selectMultiple?.helpText?.pad}>
```
taysea: Is this a candidate for `helpText.container.pad`?

If in the future we ever wanted to allow theme to drive the size of the help text itself, we might want this to be the container?

jcfilben: Good call, changed this to add container

src/js/components/SelectMultiple/SelectMultipleContainer.js
```diff
@@ -365,7 +365,10 @@ const SelectMultipleContainer = forwardRef(
     if (help) {
       if (typeof help === 'string')
         helpContent = (
-          <Box flex={false} pad="xsmall">
+          <Box
+            flex={false}
+            pad={theme.selectMultiple?.helpText?.container?.pad}
```
taysea:
```suggestion
            pad={theme.selectMultiple?.help?.container?.pad}
```
Since the prop is "help" should we call it that?

src/js/components/SelectMultiple/SelectMultipleContainer.js
```diff
@@ -548,7 +558,7 @@ const SelectMultipleContainer = forwardRef(
                               </Text>
                             </Box>
                           }
-                          pad="xsmall"
+                          pad={theme.selectMultiple?.checkBox?.pad}
```
taysea: The line above called `theme.selectMultiple?.option?.pad` is also applying to a checkbox. It seems like both of these ultimately should follow the same styles and one is just handling the label differently in cases where search terms are being bolded.

Can we consolidate to one theme object? I'm slightly leaning towards calling in `theme.selectMultiple.option.pad` but could be swayed.

jcfilben: Good call, not sure why I made two theme objects. I updated to just use `theme.selectMultiple.option.pad`

src/js/components/SelectMultiple/SelectMultipleContainer.js
jcfilben: Yes, updated to `help` instead of `helpText`

britt6612 (approved):

PR #7759, opened 2025-08-08
Add Select theme objects

jcfilben:
#### What does this PR do?
Adds theme objects for the Select component. Based on changes in https://github.com/grommet/grommet/pull/7591

### DIFF
--- a/src/js/components/Select/SelectContainer.js
+++ b/src/js/components/Select/SelectContainer.js
@@ -399,7 +399,10 @@ const SelectContainer = forwardRef(
           {...passThemeFlag}
         >
           {onSearch && (
-            <Box pad={!customSearchInput ? 'xsmall' : undefined} flex={false}>
+            <Box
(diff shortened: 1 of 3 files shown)


src/js/components/Select/SelectContainer.js
```diff
@@ -399,7 +399,10 @@ const SelectContainer = forwardRef(
           {...passThemeFlag}
         >
           {onSearch && (
-            <Box pad={!customSearchInput ? 'xsmall' : undefined} flex={false}>
+            <Box
+              pad={!customSearchInput ? theme.select?.search?.pad : undefined}
```
britt6612, line 403: just wondering about that ternary

britt6612: if user would know

taysea: We can capture the condition in docs I think!

britt6612: okay!

PR #7760, opened 2025-08-08
Add Pagination theme objects

jcfilben:
#### What does this PR do?
Adds theme objects for the Pagination component. Based on changes in https://github.com/grommet/grommet/pull/7591

### DIFF
--- a/src/js/components/Pagination/Pagination.js
+++ b/src/js/components/Pagination/Pagination.js
@@ -243,7 +243,7 @@ const Pagination = forwardRef(
       <StyledPaginationContainer
         direction="row"
         align="center"
-        gap={{ column: 'xsmall', row: 'small' }}
+        gap={theme.pagination.withControls?.gap}
         wrap
         flex={false}
         {...theme.pagination.container}
@@ -263,7 +263,7 @@ const Pagination = forwardRef(
         <Box
           align="center"
           direction="row"
-          gap={{ column: 'xsmall', row: 'small' }}
+          gap={theme.pagination.withControls?.gap}
           wrap
         >
           {stepOptions && (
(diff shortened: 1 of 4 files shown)


src/js/components/Pagination/Pagination.js
```diff
@@ -263,7 +263,7 @@ const Pagination = forwardRef(
         <Box
           align="center"
           direction="row"
-          gap={{ column: 'xsmall', row: 'small' }}
+          gap={theme.pagination.withControls?.gap}
```
britt6612: would there be a case in which someone would want a different `gap` here than the container gap?

britt6612: ignore this comment, look again at if we can use the theme.pagination.container.gap if that causes snapshot changes

jcfilben: Just checked and using `theme.pagination.container.gap` is causing snapshot changes
Screenshot 2025-08-11 at 11 11 10 AM

taysea: Can we do this on ~ line 230:

```
// for backwards compatibility
    if (!summary && !stepOptions)
      return (
        <StyledPaginationContainer
          flex={false}
          // Set gap to undefined for backwards compatibility
          // There is only one child anyway, so gap would not have any effect
          {...{ ...theme.pagination.container, gap: undefined }}
          {...passThemeFlag}
          {...rest}
        >
          {paginationControls}
        </StyledPaginationContainer>
      );
```

It avoids any snapshot changes (even though the snapshot changes due to gap wouldn't have effected the layout anyway given that container only has a single child.

Then we don't have to introduce a `withControls` object and can just use the existing.

jcfilben: Yes, updated in latest commit

britt6612 (approved):
Looks good

PR #7761, opened 2025-08-08
Add MaskedInput theme objects

jcfilben:
#### What does this PR do?
Adds theme objects for the MaskedInput component. Based on changes in https://github.com/grommet/grommet/pull/7591

### DIFF
--- a/src/js/components/MaskedInput/MaskedInput.js
+++ b/src/js/components/MaskedInput/MaskedInput.js
@@ -485,9 +485,7 @@ const MaskedInput = forwardRef(
                   // Determine whether the label is done as a child or
                   // as an option Button kind property.
                   const child = !theme.button.option ? (
-                    <Box pad={{ horizontal: 'small', vertical: 'xsmall' }}>
-                      {option}
-                    </Box>
+                    <Box pad={theme.maskedInput?.option?.pad}>{option}</Box>
                   ) : undefined;
                   // if we have a child, turn on plain, and hoverIndicator
 
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -1275,6 +1275,9 @@ export interface ThemeType {
     disabled?: {
       opacity?: OpacityType;
     };
+    option?: {
+      pad?: PadType;
+    };
   };
   menu?: {
     background?: BackgroundType;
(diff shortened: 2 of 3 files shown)


src/js/components/MaskedInput/MaskedInput.js
```diff
@@ -485,9 +485,7 @@ const MaskedInput = forwardRef(
                   // Determine whether the label is done as a child or
                   // as an option Button kind property.
                   const child = !theme.button.option ? (
-                    <Box pad={{ horizontal: 'small', vertical: 'xsmall' }}>
```
britt6612, line 488: same comment from here https://github.com/grommet/grommet/pull/7748/files#r2260877763

what you think?

jcfilben: Oh yes good call, tagging with the discussion label

britt6612: just add comment we do not need for hpe theme just flag if we need future wise

jcfilben: Removed this change and added a comment in latest commit

britt6612: wonder if we should align on the comments 

I had 
```
        // We are not going to add a theme object now since HPE
                    // theme is not needed, but in the future we may want
                    // to support a theme prop.
```

but open to yours as well just thinking to have it the same across all

jcfilben: Agree, it would be good to have a consistent comment.
One note is the wording "since HPE theme is not needed" is a little confusing. I think we could clarify to say that in the HPE theme we won't enter the code block

britt6612: @jcfilben 

```
// Not adding a theme object now because this code path
// is not used in the HPE theme, but we may add theme
// support here in the future.
```

jcfilben: I like it, I'll make this change

jcfilben: Changed in latest commit

britt6612 (approved):

taysea (approved):

PR #7762, opened 2025-08-08
add chart theme object

britt6612:
#### What does this PR do?
Adds theme objects for the Chart/DataChart component.
chart.js
dataChart.js

yes
yes
backwards compatible

### DIFF
--- a/src/js/components/Chart/Chart.js
+++ b/src/js/components/Chart/Chart.js
@@ -12,7 +12,6 @@ import { useThemeValue } from '../../utils/useThemeValue';
 const gradientMaskColor = '#ffffff';
 
 // use constants so re-renders don't re-trigger effects
-const defaultSize = { height: 'small', width: 'medium' };
 const defaultValues = [];
 
 const Chart = React.forwardRef(
@@ -34,7 +33,7 @@ const Chart = React.forwardRef(
       point,
       round,
       size: sizeProp,
-      thickness = 'medium',
+      thickness,
       type = 'bar',
       values: valuesProp = defaultValues,
       ...rest
@@ -43,6 +42,8 @@ const Chart = React.forwardRef(
   ) => {
     const containerRef = useForwardedRef(ref);
     const { theme, passThemeFlag } = useThemeValue();
+
+    const thicknessValue = thickness || theme.chart?.thickness.default;
 
     const values = useMemo(() => normalizeValues(valuesProp), [valuesProp]);
 
@@ -55,8 +56,11 @@ const Chart = React.forwardRef(
     );
 
     const strokeWidth = useMemo(
-      () => parseMetricToNum(theme.global.edgeSize[thickness] || thickness),
-      [theme.global.edgeSize, thickness],
+      () =>
+        parseMetricToNum(
+          theme.global.edgeSize[thicknessValue] || thicknessValue,
+        ),
+      [theme.global.edgeSize, thicknessValue],
     );
 
     // inset is { top, left, bottom, right }
@@ -124,6 +128,10 @@ const Chart = React.forwardRef(
 
     // size is { width, height }
     const size = useMemo(() => {
+      const defaultSize = {
+        height: theme.chart?.height,
+        width: theme.chart?.width,
+      };
       const gapWidth = gap
         ? parseMetricToNum(theme.global.edgeSize[gap] || gap)
         : strokeWidth;
@@ -171,6 +179,8 @@ const Chart = React.forwardRef(
       strokeWidth,
       theme.global.edgeSize,
       theme.global.size,
+      theme.chart?.height,
+      theme.chart?.width,
       values,
     ]);
 
--- a/src/js/components/DataChart/DataChart.js
+++ b/src/js/components/DataChart/DataChart.js
@@ -45,7 +45,7 @@ const DataChart = forwardRef(
       data = [],
       detail,
       direction = 'vertical',
-      gap = 'small',
+      gap,
       guide: guideProp,
       legend,
       offset,
@@ -197,7 +197,9 @@ const DataChart = forwardRef(
         medium,
       };
       const granularity1 = {
-        ...(heightYGranularity[(size && size.height) || 'small'] || {
+        ...(heightYGranularity[
+          (size && size.height) || theme.dataChart.heightYGranularity
+        ] || {
           fine: 5,
           medium: 3,
         }),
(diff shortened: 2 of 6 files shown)


src/js/components/Chart/Chart.js
```diff
@@ -44,6 +43,8 @@ const Chart = React.forwardRef(
     const containerRef = useForwardedRef(ref);
     const { theme, passThemeFlag } = useThemeValue();
 
+    const thicknessValue = thickness || theme.chart?.thickness.default;
```
taysea: Should we do `useMemo` here?

src/js/components/DataChart/DataChart.js
```diff
@@ -197,7 +197,9 @@ const DataChart = forwardRef(
         medium,
       };
       const granularity1 = {
-        ...(heightYGranularity[(size && size.height) || 'small'] || {
+        ...(heightYGranularity[
+          (size && size.height) || theme.dataChart.heightYGranularity
```
taysea: The way I interpret what this "small" refers to is really the size of the chart (and depending on that the granularity changes). So if anything it seems like size.height should default to small?

taysea: And so maybe it should be a default theme object for dataChart size?

taysea (commented):
Overall direction here looks great. Added a few comments

src/js/components/Chart/Chart.js
MikeKingdom: ThicknessType is a simple atomic type (string or number) and the caculation here is minimal. I don't see any reason to memoize it here (ie not costly calculation and value won't cause a re-render in anything.)

src/js/themes/base.js
```diff
@@ -781,6 +781,18 @@ export const generate = (baseSpacing = 24, scale = 6) => {
     },
     chart: {
       color: 'graph-0',
+      height: 'small',
+      thickness: {
+        default: 'medium',
```
taysea: We don't call out "default" anywhere else in the theme (except for on default button). Typically the value of the theme object is assumed to be the default value. Trying to think through if there's any other object structure we could do to avoid this.

britt6612: yes thats a good point Ill think on this as well

britt6612: what about just `chart.thickness:` then set to `medium` just remove the `default`?

taysea: I liked that, but then was just figuring out what to do with `chart.thickness.pad`

britt6612: changed

src/js/components/DataChart/Detail.js
```diff
@@ -224,7 +224,7 @@ const Detail = ({
           <Box pad="small" background={{ color: 'background-back' }}>
             <Grid
               columns={['auto', 'auto', 'auto']}
-              gap="xsmall"
+              gap={theme.dataChart.detail.gap}
```
jcfilben, line 227: It looks like in the theme we define `theme.dataChart.detail.grid.gap` but here we just point to `theme.dataChart.detail.gap`

src/js/themes/base.d.ts
```diff
@@ -821,6 +831,38 @@ export interface ThemeType {
         huge?: string;
       };
     };
+    dataChart?: {
```
jcfilben, line 857: looks like we are missing detail.gap

britt6612: sorry not sure why they didnt push up

src/js/themes/base.js
```diff
@@ -903,6 +913,38 @@ export const generate = (baseSpacing = 24, scale = 6) => {
...
+        height: 'small',
+      },
+      thickness: {
+        veryDense: 'hair',
+        dense: 'xsmall',
+        heavy: 'small',
+        moderate: 'medium',
+        light: 'large',
+        sparse: 'xlarge',
+        pad: {
```
taysea: Should we make this `thicknessPad` and move as peer to `thickness` to align with the chart theme object?

britt6612: yes make sense to make it match chart theme object

src/js/components/Chart/calcs.js
```diff
@@ -194,24 +187,32 @@ export const calcs = (values = [], options = {}) => {
     x += xStepInterval;
   }
 
+  const thicknessPad = {
+    xlarge: theme.dataChart?.thicknessPad?.xlarge || 'large',
```
britt6612, line 191: since this is in the calcs.js file does theme.dataChart make sense @taysea @jcfilben

taysea (approved):

src/js/components/Chart/calcs.js
taysea: If this is shared across chart + datachart, I wonder if we should just define it in "theme.chart.thicknessPad" for now (and not have dataChart.thicknessPad?

soft opinion

britt6612: yes I mean its in the Chart folder so we can keep it as `theme.chart.thicknessPad`

britt6612: actually maybe I take that back because if we were to move the `thicknessPad` then I would think we would also move the 

```
   thickness: {
        veryDense: 'hair',
        dense: 'xsmall',
        heavy: 'small',
        moderate: 'medium',
        light: 'large',
        sparse: 'xlarge',
      },
```

and that would be a problem since we already have a theme.chart.thickness which is set to "medium"

jcfilben: Is there a way to distinguish the thickness used in calcs.js vs the thickness in chart.js? What do they each end up representing in the Chart?

It looks like the thickness in calcs.js is a fallback for when thickness isn't defined. Maybe we could call it `defaultThickness` or `fallbackThickness`?

I'm wondering since in Chart thickness will always be defined then we will never hit this theme object. We will only run into this theme object in DataChart so maybe it is okay to leave the namespace as datachart?

taysea (approved):

src/js/components/Chart/calcs.js
taysea: If it's just datachart then I'm good with leaving as is in datachart!

britt6612: yes this is correct 
`It looks like the thickness in calcs.js is a fallback for when thickness isn't defined. Maybe we could call it defaultThickness or fallbackThickness?`

Im not sure about the `default` word since we took that out earlier trying to avoid. right Im thinking as well we just run into this with DataChart so leave the theme as is

PR #7763, opened 2025-08-08
Add FileInput theme objects

jcfilben:
#### What does this PR do?
Adds theme objects for the FileInput component. Based on changes in https://github.com/grommet/grommet/pull/7591

### DIFF
--- a/src/js/components/FileInput/FileInput.js
+++ b/src/js/components/FileInput/FileInput.js
@@ -206,17 +206,14 @@ const FileInput = forwardRef(
         rightOffset = rightOffsetRemove + parseMetricToNum(rightPad);
       if (files.length === 1 || files.length > aggregateThreshold) {
         rightOffset =
-          rightOffsetBrowse +
-          rightOffsetRemove +
-          parseMetricToNum(theme.global.edgeSize.small) * 2;
+          rightOffsetBrowse + rightOffsetRemove + theme.fileInput.rightOffset;
       } else if (rightOffsetBrowse > rightOffsetRemove) {
-        rightOffset =
-          rightOffsetBrowse + parseMetricToNum(theme.global.edgeSize.small) * 2;
+        rightOffset = rightOffsetBrowse + theme.fileInput.rightOffset;
       } else rightOffset = rightOffsetRemove;
     } else if (!files.length && controlRef.current) {
       rightOffset =
         controlRef.current.getBoundingClientRect().width +
-        parseMetricToNum(theme.global.edgeSize.small) * 2;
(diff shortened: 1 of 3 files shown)


src/js/components/FileInput/FileInput.js
```diff
@@ -206,17 +206,14 @@ const FileInput = forwardRef(
...
       if (files.length === 1 || files.length > aggregateThreshold) {
         rightOffset =
-          rightOffsetBrowse +
-          rightOffsetRemove +
-          parseMetricToNum(theme.global.edgeSize.small) * 2;
+          rightOffsetBrowse + rightOffsetRemove + theme.fileInput.rightOffset;
       } else if (rightOffsetBrowse > rightOffsetRemove) {
-        rightOffset =
-          rightOffsetBrowse + parseMetricToNum(theme.global.edgeSize.small) * 2;
+        rightOffset = rightOffsetBrowse + theme.fileInput.rightOffset;
```
britt6612: maybe we can add to our discussion `rightOffset` not sure if I would know what that means.

the rest looks good though

britt6612:
https://github.com/grommet/grommet/pull/5094

lets look and see what it looked like previous to this fix

jcfilben:
I updated the logic for calculating the rightPad. With the updated logic we no longer need to add a theme object

src/js/components/FileInput/FileInput.js
```diff
@@ -264,6 +226,66 @@ const FileInput = forwardRef(
...
+          controlWidth +=
+            parseInt(
+              theme.global.edgeSize[theme.fileInput.anchor.margin.horizontal] ||
+                theme.fileInput.anchor.margin.horizontal,
+              10,
+            ) * 2;
+        }
+        if (theme.fileInput.anchor.margin?.right) {
+          controlWidth += parseInt(
+            theme.global.edgeSize[theme.fileInput.anchor.margin.right] ||
```
britt6612, line 269: where is `theme.fileInput.anchor.margin.right` defined?

jcfilben: We don't have it defined in our base theme but someone could pass a value to margin right

jcfilben: See TS definition for MarginType:
Screenshot 2025-08-12 at 1 56 19 PM

britt6612: ah I missed the
```
  if (theme.fileInput.anchor.margin?.right)

```
sorry I was like we dont have anything set

PR #7764, opened 2025-08-08
Add DataTable theme objects

jcfilben:
#### What does this PR do?
Adds theme objects for the DataTable component. Based on changes in https://github.com/grommet/grommet/pull/7591

### DIFF
--- a/src/js/components/DataTable/Footer.js
+++ b/src/js/components/DataTable/Footer.js
@@ -35,8 +35,8 @@ const Footer = forwardRef(
           {groups && (
             <TableCell
               plain
-              size={theme.dataTable.footer.cell.size}
-              pad={theme.dataTable.footer.cell.pad}
+              size={theme.dataTable.footer.groups.size}
+              pad={theme.dataTable.footer.groups.pad}
               verticalAlign="top"
               background={cellProps.background}
               border={cellProps.border}
--- a/src/js/components/DataTable/Footer.js
+++ b/src/js/components/DataTable/Footer.js
@@ -27,6 +27,7 @@ const Footer = forwardRef(
   ) => {
     const pin = pinProp ? ['bottom'] : [];
     const { passThemeFlag } = useThemeValue();
+    const { theme } = useThemeValue();
 
     return (
       <StyledDataTableFooter ref={ref} fillProp={fill} pin={pinProp} {...rest}>
@@ -34,8 +35,8 @@ const Footer = forwardRef(
           {groups && (
             <TableCell
               plain
-              size="xxsmall"
-              pad="none"
+              size={theme.dataTable.footer.cell.size}
+              pad={theme.dataTable.footer.cell.pad}
               verticalAlign="top"
               background={cellProps.background}
               border={cellProps.border}
--- a/src/js/components/DataTable/Header.js
+++ b/src/js/components/DataTable/Header.js
@@ -300,7 +300,13 @@ const Header = forwardRef(
               )}
             </StyledDataTableCell>
           )}
-          {rowDetails && <TableCell size="xxsmall" plain pad="none" />}
+          {rowDetails && (
+            <TableCell
+              size={theme.dataTable.rowDetails?.size}
+              plain
+              pad={theme.dataTable.rowDetails?.pad}
+            />
+          )}
           {columns.map(
             ({
               property,
@@ -398,7 +404,7 @@ const Header = forwardRef(
                     <Box
                       direction="row"
                       align="center"
-                      gap="xsmall"
+                      gap={theme.dataTable.sorter?.gap}
                       justify={align}
                     >
                       {content}
--- a/src/js/components/DataTable/Searcher.js
+++ b/src/js/components/DataTable/Searcher.js
@@ -5,7 +5,6 @@ import { FormSearch } from 'grommet-icons/icons/FormSearch';
 import { Box } from '../Box';
 import { Button } from '../Button';
 import { Keyboard } from '../Keyboard';
-import { Text } from '../Text';
 import { TextInput } from '../TextInput';
 import { MessageContext } from '../../contexts/MessageContext';
 import { normalizeColor } from '../../utils';
@@ -60,7 +59,7 @@ const Searcher = ({
         flex
         // padding right is not needed any longer. There is margin
         // right set already on the container, see Header.js
-        pad={{ left: 'small' }}
+        pad={theme.dataTable.searcher?.pad}
       >
         <TextInput
           name={`search-${property}`}
@@ -73,35 +72,21 @@ const Searcher = ({
       </Box>
     </Keyboard>
   ) : (
-    <>
-      {filters[property] ? (
-        <Box
-          flex={false}
-          pad={{ horizontal: 'small' }}
-          direction="row"
-          align="center"
-        >
-          <Text>{filters[property]}</Text>
-        </Box>
-      ) : null}
-      <Button
-        ref={buttonRef}
-        a11yTitle={a11yTitle}
-        icon={
-          <FormSearch
-            color={normalizeColor(
-              filtering === property ? 'brand' : 'border',
-              theme,
-            )}
-          />
-        }
-        hoverIndicator
-        focusIndicator={focusIndicator}
-        onClick={() =>
-          onFiltering(filtering === property ? undefined : property)
-        }
-      />
-    </>
+    <Button
+      ref={buttonRef}
+      a11yTitle={a11yTitle}
+      icon={
+        <FormSearch
+          color={normalizeColor(
+            filtering === property ? 'brand' : 'border',
+            theme,
+          )}
+        />
+      }
+      hoverIndicator
+      focusIndicator={focusIndicator}
+      onClick={() => onFiltering(filtering === property ? undefined : property)}
+    />
   );
 };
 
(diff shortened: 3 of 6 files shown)


src/js/components/DataTable/Searcher.js
```diff
@@ -73,35 +72,21 @@ const Searcher = ({
...
-      {filters[property] ? (
-        <Box
-          flex={false}
-          pad={{ horizontal: 'small' }}
-          direction="row"
-          align="center"
-        >
-          <Text>{filters[property]}</Text>
-        </Box>
-      ) : null}
```
jcfilben: It didn't seem like we were ever using this code block so I removed it

britt6612: weird.. 🤔

src/js/components/DataTable/Footer.js
```diff
@@ -27,15 +27,16 @@ const Footer = forwardRef(
...
 
     return (
       <StyledDataTableFooter ref={ref} fillProp={fill} pin={pinProp} {...rest}>
         <TableRow>
           {groups && (
             <TableCell
               plain
-              size="xxsmall"
-              pad="none"
+              size={theme.dataTable.footer.groups.size}
```
britt6612: Im not sure Im following the `groups`

src/js/components/DataTable/Header.js
```diff
@@ -300,7 +300,13 @@ const Header = forwardRef(
               )}
             </StyledDataTableCell>
           )}
-          {rowDetails && <TableCell size="xxsmall" plain pad="none" />}
+          {rowDetails && (
+            <TableCell
```
britt6612: looks good

src/js/components/DataTable/Header.js
```diff
@@ -398,7 +404,7 @@ const Header = forwardRef(
                     <Box
                       direction="row"
                       align="center"
-                      gap="xsmall"
+                      gap={theme.dataTable.sorter?.gap}
```
britt6612, line 407: i wonder if we need the `sorter` or could be just `sort.gap`

src/js/components/DataTable/Footer.js
jcfilben: We only enter this block of code if groups is true (if we are using the groupBy prop)

britt6612: so would groupBy make more sense? just thinking as a user groups Im not sure would mean anything to me off the bat

jcfilben: Yeah I'm open to changing this to groupBy

jcfilben: Pushed up changes

src/js/components/DataTable/Searcher.js
britt6612: I looked back I think this block of code could get hit idk how often but I think we should leave as is.

I added a background pink to the box so I could see 
```
      {filters[property] ? (
        <Box
          flex={false}
          pad={{ horizontal: 'small' }}
          direction="row"
          align="center"
          background="pink"
        >
          <Text>{filters[property]}</Text>
        </Box>
      ) : null}
```


example story

```
import { Box, DataTable, Heading, Meter, Text } from 'grommet';

const amountFormatter = new Intl.NumberFormat('en-US', {
  style: 'currency',
  currency: 'NIS',
  minimumFractionDigits: 2,
});

const columns = [
  {
    property: 'name',
    header: <Text>Name with extra</Text>,
    primary: true,
    search: true,
  },
  {
    property: 'location',
    header: 'Location',
  },
];

const DATA = [
  { key: 1, name: 'Ilana', location: 'Bay Area', paid: 0 },
  { key: 2, name: 'Chris', location: 'Paris', paid: 1234 },
];

export const Show = () => {
  const step = 10;

  const [filters, setFilters] = useState({ name: 'Chris' });
  const [filtering, setFiltering] = useState(undefined);

  return (
      <Heading level={3}>
        <Box gap="small">
          <strong>DataTable with filter text showing</strong>
        </Box>
      </Heading>
      <DataTable
        columns={columns}
        data={DATA}
        step={step}
        filters={filters}
        filtering={filtering}
        onFilter={(property, value) =>
          setFilters((prev) => ({ ...prev, [property]: value }))
        }
        onFiltering={(prop) => setFiltering(prop)}
        show={20}
      />
    </Box>
  );
};

export default {
  title: 'Visualizations/DataTable/Show',
};
```

https://github.com/user-attachments/assets/e78a158d-459b-4814-b775-4b7e3217a00b

src/js/components/DataTable/Footer.js
britt6612: looks good thank you

src/js/components/DataTable/Searcher.js
jcfilben: Good catch! Let me go back in and look at this

jcfilben: Just pushed up changes to restore this code block

src/js/components/DataTable/Footer.js
```diff
@@ -27,15 +27,16 @@ const Footer = forwardRef(
...
 
     return (
       <StyledDataTableFooter ref={ref} fillProp={fill} pin={pinProp} {...rest}>
         <TableRow>
           {groups && (
             <TableCell
               plain
-              size="xxsmall"
-              pad="none"
+              size={theme.dataTable.footer.groupBy.size}
```
britt6612:
```suggestion
              size={theme.dataTable.footer?.groupBy?.size}
```

src/js/components/DataTable/Footer.js
```diff
@@ -27,15 +27,16 @@ const Footer = forwardRef(
...
     return (
       <StyledDataTableFooter ref={ref} fillProp={fill} pin={pinProp} {...rest}>
         <TableRow>
           {groups && (
             <TableCell
               plain
-              size="xxsmall"
-              pad="none"
+              size={theme.dataTable.footer?.groupBy?.size}
+              pad={theme.dataTable.footer.groupBy.pad}
```
britt6612:
```suggestion
              pad={theme.dataTable.footer?.groupBy?.pad}
```


one more comment

britt6612 (approved):

src/js/components/DataTable/Header.js
```diff
@@ -300,7 +300,13 @@ const Header = forwardRef(
               )}
             </StyledDataTableCell>
           )}
-          {rowDetails && <TableCell size="xxsmall" plain pad="none" />}
+          {rowDetails && (
+            <TableCell
+              size={theme.dataTable.rowDetails?.size}
```
taysea: General naming question -- it seems like this and the footer groupBy one are all related to how to size a cell when there is an "expand" cell. Is that correct?

If so, seems like all those would move in unison -- should we simplify to `theme.dataTable.expand.size` ? or something

src/js/components/DataTable/Header.js
```diff
@@ -300,7 +300,13 @@ const Header = forwardRef(
...
+              plain
+              pad={theme.dataTable.rowDetails?.pad}
```
taysea: Since pad is none should we even create a theme object for it?

src/js/components/DataTable/Searcher.js
```diff
@@ -77,7 +77,7 @@ const Searcher = ({
       {filters[property] ? (
         <Box
           flex={false}
-          pad={{ horizontal: 'small' }}
+          pad={theme.dataTable.searcher?.text?.pad}
```
taysea:
```suggestion
          pad={theme.dataTable.search?.text?.pad}
```
feels like "searcher" is the internal component but we don't really use that kind of naming elsewhere. Should we just call "search"?

src/js/components/DataTable/Sorter.js
```diff
@@ -33,7 +33,7 @@ const Sorter = ({
       direction="row"
       justify={align}
       align="center"
-      gap="xsmall"
+      gap={theme.dataTable.sorter?.gap}
```
taysea: This component doesn't even seem to be used anywhere in grommet -- can one of you confirm? And if so, lets not even add a theme object for it.

jcfilben: Just checked and it looks like we are not using it. I'll remove the theme object

src/js/components/DataTable/Searcher.js
jcfilben: Yes, search makes more sense from a user's perspective, I'll update

src/js/components/DataTable/Header.js
jcfilben: I'll remove

src/js/components/DataTable/Header.js
jcfilben: Yeah I like `theme.dataTable.expand.size`, I'll update to use that

src/js/components/DataTable/Sorter.js
jcfilben: We do still need the theme object to use in the Header.js file. But I removed the reference to it in Sorter.js since we don't use code from this file.

src/js/components/DataTable/Header.js
```diff
@@ -398,7 +400,7 @@ const Header = forwardRef(
                     <Box
                       direction="row"
                       align="center"
-                      gap="xsmall"
+                      gap={theme.dataTable.sorter?.gap}
```
taysea:
```suggestion
                      gap={theme.dataTable.sort?.gap}
```
not sure if this naming would be better ?

jcfilben: Yes, just pushed up changes

PR #7767, opened 2025-08-12
created a11Props id for Text with tip combo - issue: Text - Tip is not accessible #7654

### LINKED ISSUE #7654: Text - Tip is not accessible
When the Text component has the `tip` prop defined the tooltip is not accessible to keyboard users. The text needs to be included in the tab order and the Tip needs an `id` in addition to the text having an `aria-describedby` attribute that points to the id for the tooltip to be read out by a screenreader

Danny-Web-Dev:
#### What does this PR do?

- I have fixed the accessibility issue by adding the id to the Tip component and the Text component reference to this id.

src/js/components/Text/Text.js

I have updated the Text-test.tsx file to match the new way of the Text component with the Tip and ran the tests. (also did not cancel the pre-commit)

Add a Text component with Tip prop and check if tooltip is accessible to keyboard users, also check via screen reader.

I would like to get a feedback if possible.

closes #7654 

No

No

Is compatible.


src/js/components/Text/Text.js
```diff
@@ -71,6 +71,17 @@ const Text = forwardRef(
       );
     }
 
+    let tipId;
+    let extraA11yProps = {};
+
+    if (tipProp || textTruncated) {
+      tipId = `tip-${Math.random().toString(36).slice(2, 9)}`;
```
jcfilben: I think we can use `useId` here (just import it from utils instead of directly from React so that we pull in a version that is backwards compatible with older versions of React). See AccordionPanel.js for an example, we had to do something similar there.

Danny-Web-Dev: Ok, thanks for the feedback. I will get to it.

Danny-Web-Dev: @jcfilben done :)

src/js/themes/base.d.ts
```diff
@@ -1875,6 +1875,10 @@ export interface ThemeType {
     gap?: GapType;
     pad?: PadType;
   };
+  sidebar?: {
```
britt6612: these changes look unrelated can you just remove?

Danny-Web-Dev: Sorry about that, done.

britt6612: all good thank you for cleaning up!

Danny-Web-Dev:
@britt6612 I think this is not in a waiting status anymore :)

Sulaymon333:
@Danny-Web-Dev thank you for putting this up.

The PR looks good to me. the only thing I notice is that tip with content drop voice message using native voiceover from Macbook says `"object object, group"` instead of announcing the actual tip message like `"tooltip, group"`

tip with dropProps and truncate="tip" are both working fine with the voice over.

britt6612:
@Danny-Web-Dev @Sulaymon333 

https://deploy-preview-7767--sad-tereshkova-173d07.netlify.app/?path=/story/controls-tip-info--info

with this story I dont hear how someone using voiceover with keyboard only would be able to get to the Tip or hear the content information in the tip.

britt6612:
@Sulaymon333 
can you clean up merge conflicts as well as maybe leverage some behavior from this PR https://github.com/grommet/grommet/pull/7807 if that tooltip was being read out correctly can not remember off the top of my head so if you can test it out.

Sulaymon333:
> @Danny-Web-Dev @Sulaymon333
> 
> https://deploy-preview-7767--sad-tereshkova-173d07.netlify.app/?path=/story/controls-tip-info--info
> 
> with this story I dont hear how someone using voiceover with keyboard only would be able to get to the Tip or hear the content information in the tip.

I did not hear either but I understood the issue #7654 is meant to address Text component with `tip` prop. And this PR only made changes to `Text.js` only and no changes are made to `Tip.js`. The story you shared is about the `Tip` component itself. What is the expectation? Is the issue expected to fix for both cases or we can track the issue for `Tip` in another ticket?

Sulaymon333:
> @Sulaymon333 can you clean up merge conflicts as well as maybe leverage some behavior from this PR #7807 if that tooltip was being read out correctly can not remember off the top of my head so if you can test it out.

PR https://github.com/grommet/grommet/pull/7807 did not work as expected when I pulled it down and tested it. I did not hear any voiceover announcement. I cannot use even the keyboard navigation either if I remember correctly.

I will fix the merge conflicts and push it up.

Sulaymon333:
I cleaned up the test and also resolve the conflict issue.

Sulaymon333:
> @Danny-Web-Dev thank you for putting this up.
> 
> The PR looks good to me. the only thing I notice is that tip with content drop voice message using native voiceover from Macbook says `"object object, group"` instead of announcing the actual tip message like `"tooltip, group"`
> 
> tip with dropProps and truncate="tip" are both working fine with the voice over.

This is fixed now. 

@britt6612 should we track the `Tip` component issue you highlighted in another ticket?

britt6612 (approved):
LGTM 

reads out clear thanks!

src/js/components/Text/Text.js
```diff
@@ -31,11 +31,11 @@ const Text = forwardRef(
...
     ref,
   ) => {
-    const { passThemeFlag } = useThemeValue();
-    const textRef = useForwardedRef(ref);
     const [textTruncated, setTextTruncated] = useState(false);
     const textContextValue = useMemo(() => ({ size }), [size]);
-
+    const { passThemeFlag } = useThemeValue();
+    const textRef = useForwardedRef(ref);
+    const tipId = useId();
```
jcfilben: let's pull id out of rest, then if it is undefined we can call useId

src/js/components/Text/Text.js
```diff
@@ -21,7 +21,7 @@ const Text = forwardRef(
       tip: tipProp,
       // can't alphabetize a11yTitle before tip is defined
       a11yTitle = (typeof tipProp === 'string' && tipProp) ||
-        tipProp?.content ||
+        (tipProp?.content && getTextFromReactNode(tipProp?.content)) ||
```
jcfilben: Why do we need this if we are using the aria-describedby pattern?

jcfilben:
> @britt6612 should we track the Tip component issue you highlighted in another ticket?

What issue are you referencing?

britt6612:
> > @britt6612 should we track the Tip component issue you highlighted in another ticket?
> 
> What issue are you referencing?

@jcfilben previously I was not getting the voiceover Im not sure if it was my end or what but thats what I was referring to but its coming through on my end now.

src/js/components/Text/Text.js
britt6612: good call ☝️

Sulaymon333:
@jcfilben I made some changes

- I replace `aria-label` with `aria-labelledby` and link to tip content via id prop. 
- I updated the `Text-test.tsx` test covering  `aria-labelledby`
- I removed ununsed util function `getTextFromReactNode.js`

src/js/components/Text/Text.js
```diff
@@ -19,10 +19,6 @@ const Text = forwardRef(
       tag,
       as,
       tip: tipProp,
-      // can't alphabetize a11yTitle before tip is defined
-      a11yTitle = (typeof tipProp === 'string' && tipProp) ||
```
jcfilben, line 23: I think we may run into backwards compatibility issues removing the a11yTitle

src/js/components/Tip/Tip.js
```diff
@@ -104,6 +107,7 @@ const Tip = forwardRef(
             target={componentRef.current}
             trapFocus={false}
             key="tip-drop"
+            id={id || undefined}
```
jcfilben: Do we need the explicit undefined here?
```suggestion
            id={id}
```

src/js/components/Text/Text.js
jcfilben, line 1: When the text receives focus can we style the focus indicator so that it matches our other focus indicators instead of the system default?

src/js/components/Text/Text.js
jcfilben: I think we could keep the a11yTitle but only use it if someone explicitly passes a string to it

Sulaymon333: I added logic to respect `a11yTitle` as long as is defined as prop for `Text`. It takes precedence over `aria-labelledby`

src/js/components/Text/Text.js
Sulaymon333: theme focus style is added as well. It generate more snapshots though for when `Text` was used as expected. Conditionally adding focus for only when truncate/tip props are defined does not make any difference

src/js/components/Text/Text.js
jcfilben: Why do we still have `typeof tipProp === 'string' && tipProp`?

src/js/components/Text/Text.js
```diff
@@ -32,10 +30,13 @@ const Text = forwardRef(
...
   ) => {
     const { passThemeFlag } = useThemeValue();
+    const usingKeyboard = useKeyboard();
+    const [focus, setFocus] = useState(false);
     const textRef = useForwardedRef(ref);
     const [textTruncated, setTextTruncated] = useState(false);
     const textContextValue = useMemo(() => ({ size }), [size]);
-
+    const generatedId = useId();
+    const tipId = a11yTitle ? undefined : `${rest.id || generatedId}-tipId`;
```
jcfilben: What is someone passes both a11yTitle and an id?

src/js/components/Text/Text.js
```diff
@@ -79,8 +89,13 @@ const Text = forwardRef(
         level={level}
         truncate={truncate}
         size={size}
+        focus={focus && usingKeyboard}
```
jcfilben: It seems like we only need to set this if there is a tip

src/js/components/Tip/Tip.js
Sulaymon333: the undefined is removed now.

src/js/components/Text/Text.js
Sulaymon333: 'a11yTitle` takes precedence whenever it is defined. We will mot build the `aria-labelledby` and will not pass the id to the tip content element either.  But the `id` prop still remains on the actual `Text` component itself. see attached images for more info. 

image

image


image

image


image



```js
const tipId = a11yTitle ? undefined : `${rest.id || generatedId}-tipId`;

  if (tipProp || textTruncated) {
      extraA11yProps = {
        tabIndex: 0,
        'aria-labelledby': tipId, // tipId/aria-labelledby will be undefined if a11yTitle prop is used on `Text` component
      };
    }

if (tipProp || textTruncated) {
      let tipProps = tipProp && typeof tipProp === 'object' ? tipProp : {};
      tipProps = { ...tipProps, id: tipId }; // id/tipId will will will be undefined if a11yTitle prop is used on `Text` component
```

src/js/components/Text/Text.js
Sulaymon333: The focus is set conditionally to only apply for either tip or truncate props present.

src/js/components/Text/Text.js
jcfilben: If someone manually passes an id I feel like we should avoid modifying by appending `-tipId`

src/js/components/Text/Text.js
```diff
@@ -79,8 +89,13 @@ const Text = forwardRef(
         level={level}
         truncate={truncate}
         size={size}
+        focus={(tipProp || truncate) && focus && usingKeyboard}
+        tip={tipProp}
         {...passThemeFlag}
         {...rest}
+        {...extraA11yProps}
+        onFocus={() => setFocus(true)}
+        onBlur={() => setFocus(false)}
```
jcfilben: Why not include focus, onFocus, and onBlur in extraA11yProps?

src/js/utils/useId.d.ts
```diff
@@ -0,0 +1 @@
+export function useId(): string;
```
jcfilben: Why not add this to `utils/index.d.ts`?

jcfilben (commented):
Kind of an edge case but when I tab to the text and then click on it I get the browser focus style

https://github.com/user-attachments/assets/c2f97b07-e052-4319-aff5-d5de44e6b109

src/js/components/Text/Text.js
Sulaymon333: this is to be backward compatible and respect `tip prop` [api contract](https://v2.grommet.io/text#tip) that states _If the value is a string and no a11yTitle value is provided, tip value will be used for the a11yTitle default value_.

src/js/components/Text/Text.js
Sulaymon333: I had the same thought as well. But I pulled back from it since the variable name is tied to a11yProps specifically. In addition, focus, onFocus and onBlur are not always exclusively related to a11y. 

I think current implementation follows general pattern of separating general UI interaction event handlers from a11y props and it helps with code readability as well and separation of concerns. 

What do you think?

src/js/components/Text/Text.js
Sulaymon333: this is now fixed via this [commit](https://github.com/grommet/grommet/pull/7767/commits/49a82947e9f62140c6ac83093e95152946067eea)

user's specified id will be preserved in its original form.

src/js/utils/useId.d.ts
Sulaymon333: this is resolved via this [commit](https://github.com/grommet/grommet/pull/7767/commits/49a82947e9f62140c6ac83093e95152946067eea)

It is now rehoused into `utils/index.d.ts` and the original file has been removed

Sulaymon333:
> Kind of an edge case but when I tab to the text and then click on it I get the browser focus style
> 
>  Screen.Recording.2025-12-04.at.4.12.48.PM.mov

This is now fixed via this [commit](https://github.com/grommet/grommet/pull/7767/commits/12d77d9aa6be68db5b58f8a120ecfc5103213a80). The theme focus style will now be retained when clicked on, and focus is active. 

https://github.com/user-attachments/assets/58d2009d-6ea4-47c3-b48f-2d09098447d5

Sulaymon333:
@jcfilben, this is ready for review and merging.

sulaymon-tajudeen-hpe:
@jcfilben @britt6612 please can you have another round of review?

src/js/components/Tip/Tip.js
```diff
@@ -104,6 +107,7 @@ const Tip = forwardRef(
             target={componentRef.current}
             trapFocus={false}
             key="tip-drop"
+            id={id}
```
britt6612: @sulaymon-tajudeen-hpe do we need to add the new `id` type to the propTypes and TS file?

britt6612: @sulaymon-tajudeen-hpe should we add a test for tip as well

```
  test('Tip should have proper ARIA attributes for screen readers', async () => {
    render(
      <Grommet>
        <Text 
          id="test-text"
          tip={{
            id: "test-tooltip",
            content: "Accessible tooltip"
          }}
        >
          Screen reader accessible text
        </Text>
      </Grommet>,
    );

    const textElement = screen.getByText('Screen reader accessible text');
    
    // Text should have aria-labelledby pointing to tooltip
    expect(textElement).toHaveAttribute('aria-labelledby');
    expect(textElement).toHaveAttribute('tabIndex', '0');
    
    // Focus the text to show tooltip
    fireEvent.focus(textElement);
    
    await waitFor(() => {
      const tooltip = screen.getByText('Accessible tooltip');
      expect(tooltip).toBeInTheDocument();
      
      // Check that the Drop has the correct id
      const dropElement = tooltip.closest('[id="test-tooltip"]');
      expect(dropElement).toBeInTheDocument();
    });
  });
```
something similar

Sulaymon333: Yes, we should. Good catch! This is fixed via this [commit](the test is added via this [commit](https://github.com/grommet/grommet/pull/7767/changes/5e269df42eea1c708805d0e2afec56061446d935))

jcfilben (commented):
Storybook flagged some accessibility stuff on the Text/Tip story
Image

Sulaymon333:
@jcfilben storybook violations now fixed. updated tests and snapshots as well.

src/js/components/Text/__tests__/Text-test.tsx
```diff
@@ -6,179 +6,262 @@ import 'jest-axe/extend-expect';
 import 'regenerator-runtime/runtime';
 
 import { Grommet } from '../../Grommet';
+import { Box } from '../../Box';
```
britt6612, line 10: can we add a test for the case of when `truncate="tip"`

Sulaymon333: the test is added via this [commit](https://github.com/grommet/grommet/pull/7767/changes/5e269df42eea1c708805d0e2afec56061446d935)

src/js/components/Text/StyledText.js
```diff
@@ -56,6 +58,14 @@ const StyledText = styled('span').withConfig(styledComponentsConfig)`
...
     props.theme.text.font && props.theme.text.font.family && fontFamily}
 
+  &:focus {
+    ${focusStyle()}
+  }
+
+  &:focus:not(:focus-visible) {
+    ${unfocusStyle()}
+  }
+
```
britt6612, line 72: @sulaymon-tajudeen-hpe I don't think we need the `focusStyle` on every `text` this is setting we only want this ideally for if `text` has `tip`

Sulaymon333: @britt6612 focus is now scoped to just when tip or truncate is active. This is addressed via this [commit](https://github.com/grommet/grommet/pull/7767/changes/c1d38633ad444ca12680c0873ba0b9b43243b9a1).

src/js/components/Tip/Tip.js
```diff
@@ -109,7 +141,20 @@ const Tip = forwardRef(
             onMouseEnter={() => setTooltipOver(true)}
             onMouseLeave={() => setTooltipOver(false)}
           >
-            {plain ? content : <Box {...theme.tip.content}>{content}</Box>}
+            {plain ? (
+              <span
```
britt6612: why did we add the `span` here?

src/js/components/Text/Text.js
britt6612: @sulaymon-tajudeen-hpe isnt this still an issue?

src/js/components/Tip/Tip.js
sulaymon-tajudeen-hpe: So the span ( updated to Box using as="span") does two things. When plain is active, it 

1. Provides a DOM target for aria-describedby when content is a React node, without adding block-level layout or padding.
2. Ensures string content is a valid, identifiable child within the Drop.

I updated to use Box as="span" also to follow codebase convension

```js
<Box as="span"
        {...(isStringContent ? {} : { id: tooltipId, role: 'tooltip' })}
   >
         {content}
  </Box>
```

so basically we just need some element to anchor/wrap even when plain is true to provide minimal non styled wrapper that consumers can build on that still ensuring id is available for `aria-describedby` and it is safe for when just string and not react node is passed down to content while plain is true.

Updated via this [commit](https://github.com/grommet/grommet/pull/7767/changes/11f4a9aea494b3bedccbcaba439cd2b6261e1ebc)

britt6612:
Screenshot 2026-04-17 at 4 15 47 PM

src/js/components/Data/__tests__/__snapshots__/Data-test.tsx.snap
```diff
@@ -5171,6 +5176,7 @@ exports[`Data toolbar 1`] = `
           class="c9"
         >
           <button
+            aria-description="Open filters"
```
britt6612: hmm now on voice over this will repeat "open filters open filters" I dont think we want that behavior

src/js/components/Tip/Tip.js
Sulaymon333: So the span ( updated to Box using as="span") does two things. When plain is active, it 

1. Provides a DOM target for aria-describedby when content is a React node, without adding block-level layout or padding.
2. Ensures string content is a valid, identifiable child within the Drop.

I updated to use Box as="span" also to follow codebase convension

```js
<Box as="span"
        {...(isStringContent ? {} : { id: tooltipId, role: 'tooltip' })}
   >
         {content}
  </Box>
```

so basically we just need some element to anchor/wrap even when plain is true to provide minimal non styled wrapper that consumers can build on that still ensuring id is available for `aria-describedby` and it is safe for when just string and not react node is passed down to content while plain is true.

Updated via this [commit](https://github.com/grommet/grommet/pull/7767/changes/11f4a9aea494b3bedccbcaba439cd2b6261e1ebc)

src/js/components/Data/__tests__/__snapshots__/Data-test.tsx.snap
Sulaymon333: Good catch. Thank you. I am having a look.

Sulaymon333:
> Screenshot 2026-04-17 at 4 15 47 PM

This check is no longer needed in this implementation. I tested with `a11yTitle` on Tip story and it is coming through.

src/js/components/Data/__tests__/__snapshots__/Data-test.tsx.snap
Sulaymon333: this is addressed via this [commit](https://github.com/grommet/grommet/pull/7767/changes/39e4b57e6ab1d7b6a24d760024470901d7159577)

jcfilben (commented):
The aria-description was reading nicely for me in voiceover, but I looked into it a bit more and it is not recommended to be used to make tooltips accessible. Posting some relevant info:

Image

Image

Could be worth chatting with Bill Tipton, he might have more context.

Sulaymon333:
> The aria-description was reading nicely for me in voiceover, but I looked into it a bit more and it is not recommended to be used to make tooltips accessible. Posting some relevant info:
> 
> Image Image
> Could be worth chatting with Bill Tipton, he might have more context.

Thank you @jcfilben. I will reach out to Bill for more context and possible testing.

Sulaymon333:
> > The aria-description was reading nicely for me in voiceover, but I looked into it a bit more and it is not recommended to be used to make tooltips accessible. Posting some relevant info:
> > Image Image
> > Could be worth chatting with Bill Tipton, he might have more context.
> 
> Thank you @jcfilben. I will reach out to Bill for more context and possible testing.

I tested with Bill and we found out that one of the examples did not announce the tip text. 
I can have a closer look at the issue when we have bandwidth.

Image

PR #7771, opened 2025-08-19
add anchor theme object

britt6612:
#### What does this PR do?
Adds theme objects for the Anchor component.
anchor.js

yes
yes
backwards compatible

### DIFF
--- a/src/js/components/Anchor/StyledAnchor.js
+++ b/src/js/components/Anchor/StyledAnchor.js
@@ -75,7 +75,7 @@ const StyledAnchor = styled.a.withConfig(styledComponentsConfig)`
     edgeStyle(
       'padding',
       props.theme.anchor.iconOnly.pad,
-      props.responsive,
+      false,
       props.theme.global.edgeSize.responsiveBreakpoint,
       props.theme,
     )}
--- a/src/js/components/Anchor/StyledAnchor.js
+++ b/src/js/components/Anchor/StyledAnchor.js
@@ -72,7 +72,7 @@ const StyledAnchor = styled.a.withConfig(styledComponentsConfig)`
     props.hasIcon &&
     !props.hasLabel &&
     `
-    padding: ${props.theme.global.edgeSize.small};
+    padding: ${props.theme.anchor.iconOnly.pad};
   `}
   ${(props) => props.disabled && disabledStyle}
   ${(props) => props.focus && focusStyle()}
--- a/src/js/components/Anchor/StyledAnchor.js
+++ b/src/js/components/Anchor/StyledAnchor.js
@@ -1,6 +1,7 @@
 import styled, { css } from 'styled-components';
 
 import {
+  edgeStyle,
   focusStyle,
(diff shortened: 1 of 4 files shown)


src/js/components/Anchor/__tests__/__snapshots__/Anchor-test.tsx.snap
```diff
@@ -2048,7 +2048,7 @@ exports[`Anchor warns about invalid icon render 1`] = `
   font-weight: 600;
   text-decoration: none;
   cursor: pointer;
-  padding: 12px;
+  padding: small;
```
britt6612: does this look correct @taysea

jcfilben: I just looked through our other snapshots and I couldn't find any other places where this happens. It seems like in the snapshots things are always translated into px values, so it seems like something is off here

src/js/components/Anchor/StyledAnchor.js
```diff
@@ -72,7 +72,7 @@ const StyledAnchor = styled.a.withConfig(styledComponentsConfig)`
     props.hasIcon &&
     !props.hasLabel &&
     `
-    padding: ${props.theme.global.edgeSize.small};
+    padding: ${props.theme.anchor.iconOnly.pad};
```
jcfilben: I wonder if we need to do something like this
```
  ${(props) =>
    props.pad &&
    edgeStyle(
      'padding',
      props.pad,
      props.responsive,
      props.theme.global.edgeSize.responsiveBreakpoint,
      props.theme,
    )}
```

src/js/components/Anchor/__tests__/__snapshots__/Anchor-test.tsx.snap
britt6612: ah okay so yeah we might have to convert it using edge size or something

src/js/components/Anchor/StyledAnchor.js
britt6612: ill look

jcfilben (approved):

src/js/components/Anchor/StyledAnchor.js
```diff
@@ -71,9 +72,13 @@ const StyledAnchor = styled.a.withConfig(styledComponentsConfig)`
   ${(props) =>
     props.hasIcon &&
     !props.hasLabel &&
-    `
-    padding: ${props.theme.global.edgeSize.small};
-  `}
+    edgeStyle(
+      'padding',
+      props.theme.anchor.iconOnly.pad,
+      props.responsive,
```
taysea: It seems like this wasn't responsive previously, so we should probably just set to false?
```suggestion
      false,
```

britt6612: yes I was going to ask you your thoughts

taysea (approved):

PR #7772, opened 2025-08-19
add layer theme object

britt6612:
#### What does this PR do?
Adds theme objects for the Layer component.
layer.js

yes
yes
backwards compatible

### DIFF
--- a/src/js/components/Layer/StyledLayer.js
+++ b/src/js/components/Layer/StyledLayer.js
@@ -777,7 +777,7 @@ const StyledContainer = styled.div.withConfig({
   ${(props) => (!props.modal ? baseStyle : '')}
   display: flex;
   flex-direction: column;
-  min-height: ${(props) => props.theme.global.size.xxsmall};
+  min-height: ${(props) => props.theme.layer.container.height};
   ${(props) =>
     !props.plain &&
     (props.background || props.theme.layer.background) &&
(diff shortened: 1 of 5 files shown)


src/js/components/Layer/StyledLayer.js
```diff
@@ -777,7 +777,7 @@ const StyledContainer = styled.div.withConfig({
   ${(props) => (!props.modal ? baseStyle : '')}
   display: flex;
   flex-direction: column;
-  min-height: ${(props) => props.theme.global.size.xxsmall};
+  min-height: ${(props) => props.theme.layer.container.height};
```
jcfilben: Might need to do something like this
`${(props) => props.heightProp && heightStyle(props.heightProp, props.theme)}`

src/js/themes/base.js
```diff
@@ -1317,6 +1317,7 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       container: {
         // elevation: undefined,
         zIndex: '20',
+        height: 'xxsmall',
```
taysea: Should this be min-height right?
```suggestion
        height: { min: 'xxsmall' },
```

taysea: Oops didn't see the way the logic was implemented above. Wondering which makes more sense for a consumer?

britt6612: ah yes thats fine

britt6612: makes more sense

taysea (approved):

jcfilben (approved):

PR #7773, opened 2025-08-19
add theme value table

britt6612:
#### What does this PR do?
Adds theme objects for the table component.
table.js

yes
yes
backwards compatible

### DIFF
--- a/src/js/components/Table/StyledTable.js
+++ b/src/js/components/Table/StyledTable.js
@@ -67,7 +67,14 @@ const StyledTableCell = styled.td.withConfig(styledComponentsConfig)`
 const StyledTableDataCaption = styled.caption.withConfig(
   styledComponentsConfig,
 )`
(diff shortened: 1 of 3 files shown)


src/js/components/Table/StyledTable.js
```diff
@@ -67,7 +67,14 @@ const StyledTableCell = styled.td.withConfig(styledComponentsConfig)`
 const StyledTableDataCaption = styled.caption.withConfig(
   styledComponentsConfig,
 )`
-  margin-bottom: ${(props) => props.theme.global.edgeSize.xxsmall};
+  ${(props) =>
+    edgeStyle(
+      'margin',
+      props.theme.table.caption.margin,
+      props.responsive,
```
jcfilben: I wonder if we should not include `props.responsive` since we don't document this as an available option

PR #7775, opened 2025-08-20
Add `theme.spinner.responsive`

taysea:
#### What does this PR do?

In HPE theme, we don't want Spinner to drop down its border thickness at small breakpoints. Previously, "medium" border for desktop and mobile was the same. With t-shirt size adjustments, this isn't the case.
 
### DIFF
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -1942,6 +1942,7 @@ export interface ThemeType {
       | { color?: ColorType }
       | { size?: 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | string };
     icon?: React.ReactNode | Icon;
+    responsive?: true;
     size?: {
       xsmall?: string;
       small?: string;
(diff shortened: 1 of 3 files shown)


britt6612 (approved):

src/js/themes/base.d.ts
```diff
@@ -1942,6 +1942,7 @@ export interface ThemeType {
       | { color?: ColorType }
       | { size?: 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | string };
     icon?: React.ReactNode | Icon;
+    responsive?: true;
```
jcfilben: Should the type be a boolean?
```suggestion
    responsive?: boolean;
```

britt6612: good catch ☝️

taysea: whoops yes

britt6612 (approved):

PR #7777, opened 2025-08-21
add table cell theme object

britt6612:
#### What does this PR do?
Adds theme objects for the table cell component.
tablecell.js

yes
yes
backwards compatible

### DIFF
--- a/src/js/components/TableCell/TableCell.js
+++ b/src/js/components/TableCell/TableCell.js
@@ -111,7 +111,7 @@ const TableCell = forwardRef(
 
             // height must match cell height otherwise table will apply some
             // margin around the cell content
-            const borderSizeKey = theme.table.cell?.border?.size;
+            const borderSizeKey = theme.table?.[tableContext]?.border?.size;
             const borderSizeValue =
               theme.global.borderSize[borderSizeKey] || borderSizeKey;
             const borderSizeNum = parseInt(borderSizeValue, 10);


src/js/components/TableCell/TableCell.js
```diff
@@ -111,10 +111,15 @@ const TableCell = forwardRef(
 
             // height must match cell height otherwise table will apply some
             // margin around the cell content
+            const borderSizeKey = theme.table.cell?.border?.size;
```
taysea: Should we have this key off of `tableContext` -- which can be `header`, `body`, or `footer` (I think?)
```suggestion
            const borderSizeKey = theme.table?.[tableContext]?.border?.size;
```

taysea: Also seems like we've lost whatever logic for the first part of the condition of `border || theme.table[tableContext].border` -- what is `border`? A prop? Seems like we should make sure to still account for it.

britt6612: ah good thinking `tableContext` didnt realize we had that.

britt6612: @taysea 
yes `border` is a prop
```
const TableCell = forwardRef(
  (
    {
      align,
      'aria-disabled': ariaDisabled,
      background,
      border,
      children,
      className, // so StyledDataTableCell is applied to td/th
      colSpan,
      onResize, // for DataTable
      onWidth,
      pad,
      plain,
      rowSpan,
      scope,
      size,
      verticalAlign,
      property, // for DataTable
      ...rest
    },
    ref,
  ) => {

```

the `border` is still there though so not sure what you mean?

```

            container.style.height = `${Math.max(
              cellRect.height -
                (border || theme.table[tableContext].border
                  ? borderSizeNum
                  : 0),
              0,
            )}px`;
```

also there is already a theme a border for the header and footer

```
     header: {
        align: 'start',
        border: 'bottom',
        // verticalAlign: undefined,
        // background: undefined,
        // extend: undefined,
      },
      body: {
        align: 'start',
        // background: undefined,
        // border: undefined,
        // extend: undefined,
      },
      // row: {
      //   hover: {
      //     background: undefined,
      //     color: undefined,
      //   },
      // },
      footer: {
        align: 'start',
        border: 'top',
        // verticalAlign: undefined,
        // background: undefined,
        // extend: undefined,
      },
    },
```

britt6612:
going to close since users using hpe theme should not hit this block of code

PR #7781, opened 2025-08-28
Add `theme.tag.hover` support

taysea:
#### What does this PR do?

Add theme support for `theme.tag.hover` (background and border) when `onClick` is applied.

There have been some design discussions around potential Tag style changes for HPE theme. This won't be part of the initial landmark release, but might be a modification in future once more discussion has been had.

So while nothing has been "settled", I wanted to make sure Grommet was in a place to support the style updates if needed (and not require a peerDependency bump again in the theme).

Included a storybook example with the potential desired styles.

https://github.com/user-attachments/assets/0fed83ca-e98c-4a0e-b0cb-21cd921bdffe

Yes.

Yes.

Yes, backwards compatible.

### DIFF
--- a/src/js/components/Tag/StyledTag.js
+++ b/src/js/components/Tag/StyledTag.js
@@ -1,12 +1,26 @@
 import styled from 'styled-components';
 
-import { borderStyle, roundStyle } from '../../utils';
+import { backgroundStyle, borderStyle, roundStyle } from '../../utils';
 
 import { Button } from '../Button';
 
 export const StyledTagButton = styled(Button)`
+  ${(props) =>
+    props.background && backgroundStyle(props.background, props.theme)}
   ${(props) => props.border && borderStyle(props.border, true, props.theme)}
   ${(props) => props.round && roundStyle(props.round, true, props.theme)}
+  &:hover {
+    ${(props) =>
+      props.background &&
+      props.onClick &&
+      props.theme?.tag?.hover?.background &&
+      backgroundStyle(props.theme.tag.hover.background, props.theme)}
+    ${(props) =>
+      props.border &&
+      props.onClick &&
+      props.theme?.tag?.hover?.border &&
+      borderStyle(props.theme.tag.hover.border, props.theme)}
+  }
 `;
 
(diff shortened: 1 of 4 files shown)


src/js/components/Tag/stories/OnClick.stories.js
```diff
@@ -1,14 +1,32 @@
 import React from 'react';
 
-import { Box, Tag } from 'grommet';
+import { Box, Tag, Grommet } from 'grommet';
+import { hpe } from 'grommet-theme-hpe';
+import { deepMerge } from '../../../utils';
+
+const customTheme = deepMerge(hpe, {
```
taysea: Revert before merging.

src/js/components/Tag/StyledTag.js
```diff
@@ -1,12 +1,26 @@
 import styled from 'styled-components';
 
-import { borderStyle, roundStyle } from '../../utils';
+import { backgroundStyle, borderStyle, roundStyle } from '../../utils';
 
 import { Button } from '../Button';
 
 export const StyledTagButton = styled(Button)`
+  ${(props) =>
+    props.background && backgroundStyle(props.background, props.theme)}
```
jcfilben, line 9: Do we need this so that the hover style doesn't get overridden by button styles?

taysea: Instead, it's because that button is a `plain` button (via prop and passing children). So no button styles come through. So if we want to add any styling, it needs to be here.

jcfilben: Gotcha, thanks for the explanation

src/js/components/Tag/stories/OnClick.stories.js
jcfilben, line 1: Just need to revert this then I'll merge

britt6612: @jcfilben the story has been reverted

jcfilben: Thanks @britt6612!

PR #7784, opened 2025-09-03
Added theme objects for icons

jcfilben:

### DIFF
--- a/src/js/components/Tabs/Tabs.js
+++ b/src/js/components/Tabs/Tabs.js
@@ -47,6 +47,8 @@ const Tabs = forwardRef(
     const [focusIndex, setFocusIndex] = useState(-1);
     const headerRef = useRef();
     const size = useContext(ResponsiveContext);
+    const PreviousIcon = theme.tabs.header?.previousButton?.icon || Previous;
+    const NextIcon = theme.tabs.header?.nextButton?.icon || Next;
 
     const sendAnalytics = useAnalytics();
 
@@ -343,7 +345,7 @@ const Tabs = forwardRef(
               onClick={() => moveByArrowKey('previous')}
             >
               <Box pad={theme.tabs.header.previousButton.pad}>
-                <Previous
+                <PreviousIcon
                   color={
                     disableLeftArrow
                       ? theme.button.disabled.color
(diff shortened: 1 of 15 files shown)


src/js/components/Tabs/Tabs.js
```diff
@@ -47,6 +47,8 @@ const Tabs = forwardRef(
     const [focusIndex, setFocusIndex] = useState(-1);
     const headerRef = useRef();
     const size = useContext(ResponsiveContext);
+    const PreviousIcon = theme.tabs.header?.previousButton?.icon || Previous;
+    const NextIcon = theme.tabs.header?.nextButton?.icon || Next;
```
britt6612, line 51: this format seems different any reason why its not 

```
    const PreviousIcon = theme.tabs.header?.icons?.previousButton || Previous;
    const NextIcon = theme.tabs.header?.icons?.nextButton || Next;
```

src/js/themes/base.js
```diff
@@ -488,6 +488,9 @@ export const generate = (baseSpacing = 24, scale = 6) => {
       },
       busy: {
         gap: 'small',
+        // icons: {
+        //   success: Checkmark,
```
britt6612: should this be undefined?

britt6612:
couple small comments

src/js/themes/base.js
jcfilben: Good call, just fixed

src/js/components/Tabs/Tabs.js
jcfilben: Is it the theme format that you expected to be different?

britt6612: you have 
```
theme.tabs.header?.nextButton?.icon
```

in other places though its 
```theme.tabs.header?.icons?...```

Comment (jcfilben) [reply src/js/components/Tabs/Tabs.js:51]: Discussed offline, since `previousButton` and `nextButton` objects were already available in the theme, we were going to leave `icon` under those instead of creating a new icons object. Open to other thoughts here.

File: src/js/themes/base.d.ts [hunk 3]
```diff
@@ -187,6 +187,9 @@ interface ButtonType {
   };
   busy?: {
     gap?: GapType;
+    icons?: {
+      success?: React.ReactNode | Icon;
```
halocline, line 191: A caller would be able use an `<svg>` in their theme, correct? Is it worthwhile to include a test(s) covering icon types other than `Icon`?

jcfilben: Added a test covering the svg case. For now I just added it to one component (DataFilters), let me know if you think we should include it on all of them.

PR #7779, opened 2025-08-22
Add theme support for `theme.dataChart.granularity.y`

taysea:
#### What does this PR do?

Since grommet-theme-hpe is expanding the range of t-shirt sizes, need a way to redefine our desired granularities.

#### Do Jest tests follow these best practices?

- [ ] `screen` is used for querying.
- [ ] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [ ] `asFragment()` is used for snapshot testing.

PR #7778, opened 2025-08-21
add theme values

britt6612:
#### What does this PR do?
Adds theme objects for the dataChart component.
Based on changes in https://github.com/grommet/grommet/pull/7591
#### Where should the reviewer start?
DataChart
#### Do Jest tests follow these best practices?

- [ ] `screen` is used for querying.
- [ ] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [ ] `asFragment()` is used for snapshot testing.

#### Any background context you want to provide?
https://github.com/grommet/grommet/pull/7591
#### Do the grommet docs need to be updated?
yes
#### Should this PR be mentioned in the release notes?
yes
#### Is this change backwards compatible or is it a breaking change?
backwards compatible

PR #7791, opened 2025-09-10
Add test for DataChart theme obj

jcfilben:
#### What does this PR do?
Adds a test for the following PRs
https://github.com/grommet/grommet/pull/7779
https://github.com/grommet/grommet/pull/7778

### DIFF
--- a/src/js/themes/base.d.ts
+++ b/src/js/themes/base.d.ts
@@ -857,51 +857,6 @@ export interface ThemeType {
         huge?: string;
       };
     };
-    dataChart?: {
-      gap?: GapType;
-      granularity?: {
-        y?: {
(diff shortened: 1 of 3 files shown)


src/js/themes/base.d.ts
```diff
@@ -923,6 +878,51 @@ export interface ThemeType {
       pad?: PadType;
     };
   };
+  dataChart?: {
```
britt6612, line 881: why you move this?

jcfilben: Before it was under `clock`

britt6612: ah good okay!

britt6612 (approved):

PR #7797, opened 2025-09-18
[UPGRADE] packageS that do not cause snapshot changes

Sulaymon333:
#### What does this PR do?
- upgrade packageS that do not cause snapshot changes


package.json
```diff
@@ -54,81 +54,80 @@
...
     "@babel/preset-react": "^7.27.1",
     "@chromatic-com/storybook": "^3.2.6",
-    "@eslint/compat": "^1.2.9",
+    "@eslint/compat": "^1.3.2",
     "@eslint/eslintrc": "^3.3.1",
-    "@eslint/js": "^9.32.0",
+    "@eslint/js": "^9.35.0",
     "@storybook/addon-a11y": "^8.6.14",
     "@storybook/addon-storysource": "^8.6.14",
-    "@storybook/addon-toolbars": "^8.6.14",
```
jcfilben, line 83: Can we group this removal in with the storybook upgrade PR instead of in this PR?

Sulaymon333: done.

package.json
```diff
@@ -87,48 +87,48 @@
...
     "@types/jest-axe": "^3.5.9",
-    "@types/node": "^24.1.0",
-    "@types/react": "^19.1.9",
-    "@types/react-dom": "^19.1.7",
+    "@types/node": "^24.5.2",
+    "@types/react": "^19.1.13",
+    "@types/react-dom": "^19.1.9",
     "@types/styled-components": "^5.1.34",
-    "babel-core": "^7.0.0-bridge.0",
+    "babel-core": "^6.26.3",
```
jcfilben: Can you help me understand this change?

jcfilben (commented):
Looks like the bundlesize check is failing, you can just bump up the bundlesize in package.json

package.json
sulaymon-tajudeen-hpe: that is the available pacakage version for upgrade 
image

sulaymon-tajudeen-hpe:
Although I was getting 181.7KB for bundele size locally. I bumped the size up to 185KB now

image

jcfilben:
Looks like the package size is jumping to 191, this seems a little unexpected. This isn't necessarily an issue but it would be good to understand what's causing the change

Screenshot 2025-09-18 at 4 44 29 PM

package.json
jcfilben: I'm thinking we leave this as `"^7.0.0-bridge.0"`

Sulaymon333: Restored `babel-core` back from `^6.26.3`  to `^7.0.0-bridge.0`

Sulaymon333:
As I mentioned. I was getting 181.7 KB when I checked the bundlesize locally. 191.14 KB is from the circleCI build?

jcfilben:
> 191.14 KB is from the circleCI build?

Yes, see https://app.circleci.com/jobs/github/grommet/grommet/90100

Sulaymon333:
The increase in gzip is coming from webpack upgrade from `5.97.1` to `5.101.3`

### **With webpack `5.101.3`**
image
image


### **With webpack `5.97.1`**
image
image

jcfilben:
> The increase in gzip is coming from webpack upgrade from 5.97.1 to 5.101.3

Thanks for looking into this, do the webpack release notes give any indication of what might be causing this?

Sulaymon333:
No. Webpack release note does not mention that specifically but bundlephobia shows that the minified + Gzipped sizes are a bit more even when the minified webpack size might be the same. What I observed is that actual bundle size of `grommet.min.js` itself does not change for either v5.97.1 or v5.98.0 or v5.101.3. The extra KB shows up when converted to `gzip` format. I also noticed that webpack v5.98.0 as well causes this extra KB. So, it is not only the latest version of webpack that makes `gzip` to increase. See attached video below for more details.


- [webpack release](https://github.com/webpack/webpack/releases?page=2) notes for different versions. Look at v5.97.1 and v5.98.0 in particular. Nothing suggest increase in gzip specifically. But bundlephobia shows a slight increase as shown bleow

Bundlephobia analysis
- v5.97.1 bundlephobia analysis - https://bundlephobia.com/package/webpack@5.97.1. Minified + Gzipped size is 795.7kB
- v5.98.0 bundlephobia analysis - https://bundlephobia.com/package/webpack@5.98.0. Minified + Gzipped  size is 802kB
- v5.101.3 bundlephobia analysis - https://bundlephobia.com/package/webpack@5.98.0. Minified + Gzipped  size is 819.7kB

P.S I realised I was comparing the wrong versions from the recorded video.

https://github.com/user-attachments/assets/d5d58125-7e6a-4f36-8b3c-b9b0159a9181

MikeKingdom:
Any chance the bundle is already effectively compressed? Typically if you try to compress something that is already compressed it gets larger

jcfilben:
@Sulaymon333 any response here?

> Any chance the bundle is already effectively compressed? Typically if you try to compress something that is already compressed it gets larger

Sulaymon333:
> Any chance the bundle is already effectively compressed? Typically if you try to compress something that is already compressed it gets larger

I did a quick test on `grommet.min.js`. 
`grommet.min.gzip` is compressed version of `grommet.min.js`
`grommet.min-1.gzip` is acompressed from `grommet.min.gzip`
`grommet.min.gzip.zip ` is compressed from `grommet.min.gzip`

As shown below, I did not see any difference in original size of 192k for `grommet.min.gzip`

image

britt6612:
closing due to 

https://github.com/grommet/grommet/pull/7825
https://github.com/grommet/grommet/pull/7824

PR #7801, opened 2025-09-25
add fix for datatable pinned position

### LINKED ISSUE #7800: Data table visual issue

No

### Expected Behavior

Expected behavior
Image

The data table should look like the image above

### Actual Behavior

The issue below was seen when upgrading Grommet to v2.49.0

Image

I spoke to Matt Glissman and he said it was due to having the `pin` property on the `Serial number` column and having the `resizeable` property on the data table

Image

### Steps to Reproduce

Add `pin` property to a data table column object
Add `resizeable` property to the data table

### Your Environment

Grommet version: 2.49.0
Browser: Chrome 140.0.7339.185
OS: Windows

britt6612:
#### What does this PR do?
Fix position offset bug for pinned resizable DataTable columns.

The bug: When both `position: relative` (resize) and `position: sticky` (pin) applied to the same element, they conflicted and caused visual offset.

The fix: `position: onResize && !columnPin ? 'relative' : undefined` lets pinned columns use their existing sticky positioning while non-pinned columns get the relative positioning needed for the Resizer.
header.js
locally see `resizeable` story 
added a test to look specifically at the `position` property
storybook

The Resizer uses `position: absolute`.

Non-pinned columns: Need `position: relative` to serve as the positioning context
Pinned columns: Already have `position: sticky` from the pin styling, which can serve as the positioning context
closes #7800 
none
no
yes
backwards compatible


src/js/components/DataTable/stories/ResizableColumns.stories.js
```diff
@@ -54,26 +54,60 @@ const DATA = [
...
+    // Uncomment <Grommet> lines when using outside of storybook
+    // <Grommet theme={grommet}>
+    <Box align="center" pad="large">
+      <Heading level="3">Table with resizable & column sizes</Heading>
+      <Box overflow="auto" width="medium" height="medium">
+        <DataTable
+          columns={columnsResize}
+          data={DATA}
+          primaryKey={false}
+          resizeable
```
britt6612: TODO revert changes

src/js/components/DataTable/Header.js
```diff
@@ -500,8 +500,9 @@ const Header = forwardRef(
                       ? `${widths[property]}px`
                       : undefined,
                     boxSizing: onResize ? 'border-box' : undefined,
-                    position: onResize ? 'relative' : undefined,
+                    position: onResize && !columnPin ? 'relative' : undefined,
                     overflow: onResize ? 'visible' : undefined,
+                    zIndex: onResize && columnPin ? 2 : undefined,
```
jcfilben: Why a value of '2' here?

src/js/components/DataTable/stories/ResizableColumns.stories.js
jcfilben, line 1: It seems like the resize behavior is not working on this story, I can't change the column size

britt6612: the `resizer` has `z-index: 1;`
so I just did one above that.

src/js/components/DataTable/Header.js
britt6612: the resizer has z-index: 1;
so I just did one above that.

src/js/components/DataTable/stories/ResizableColumns.stories.js
britt6612: The issue is that the DataTable container has a fixed `width="small"` which constrains the table, when you try to resize columns, there's no room for them to expand.

halocline: Right, need to constrain the table to test the pin behavior, correct?

Can we modify the story to include a wider table too? That way we can verify both are behaving as expected. Just for testing purposes, can remove afterwards.

src/js/components/DataTable/Header.js
halocline: Can we define this in a variable and add a code comment regarding the relationship?

```
// Cells and their re-sizers need to scroll beneath pinned cells.
// StyledResizer has a z-index of 1.
const PINNED_ZINDEX = 2;
```

britt6612: yes will do

src/js/components/DataTable/stories/ResizableColumns.stories.js
britt6612: story is modified so now we can test!

src/js/components/DataTable/Header.js
britt6612: changed!

britt6612:
> Nice work! Just one last thing, the story you made is really nice for testing, I'm thinking we could create a new story ('Resize pinned' or something) and copy that code into there as a permanent addition to our storybook

yes that would be something we could do I know our stories are alot but I do think this one helps

jcfilben (approved):
Code looks good, would just want to create a new story and move the story code there before merging

src/js/components/DataTable/Header.js
```diff
@@ -500,8 +504,13 @@ const Header = forwardRef(
...
                       : undefined,
                     boxSizing: onResize ? 'border-box' : undefined,
-                    position: onResize ? 'relative' : undefined,
+                    // Don't override positioning when DataTable has pin prop
+                    position:
+                      onResize && !columnPin && !pinProp
+                        ? 'relative'
+                        : undefined,
                     overflow: onResize ? 'visible' : undefined,
+                    zIndex: onResize && columnPin ? PINNED_ZINDEX : undefined,
```
MikeKingdom: The `zIndex` in `StyledDataTableCell` is set to `z-index: ${Object.keys(props.pin).length};`

How is this interacting with that? It seems like it would override it which makes me wonder about multiple pins

britt6612: ah good point I could see this causing conflict

MikeKingdom (approved):

PR #7804, opened 2025-09-30
DataTable - add resizer theme objects for t-shirt values

jcfilben:
#### What does this PR do?
Adds theme objects for the DataTable resizer margin and padding.

I adjusted the margin from -12px to -11px because the border was sometimes getting cut off since the resizer has 24px width and the border is 1px. We want the center of the resizer to align with the edge of the cell. So we divide the resizer width by 2 and subtract 1 to account for the border width. The -11px margin-right value means that the border ends up at the edge of the cell and doesn't get cut off.

Yes, theme props should be marked as beta

backwards compatible

### DIFF
--- a/src/js/themes/base.js
+++ b/src/js/themes/base.js
@@ -1127,9 +1127,6 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         //     size: undefined,
         //   },
         // },
-        margin: {
-          right: '-11px', // offset 12 px padding + 1px border
-        },
         padding: {
           vertical: 'xsmall',
         },
--- a/src/js/themes/base.js
+++ b/src/js/themes/base.js
@@ -1127,6 +1127,12 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         //     size: undefined,
         //   },
         // },
+        margin: {
+          right: '-11px', // offset 12 px padding + 1px border
+        },
+        padding: {
+          vertical: 'xsmall',
+        },
       },
       search: {
         pad: {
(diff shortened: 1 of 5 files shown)


britt6612 (approved):

src/js/themes/base.js
```diff
@@ -1127,6 +1127,12 @@ export const generate = (baseSpacing = 24, scale = 6) => {
         //     size: undefined,
         //   },
         // },
+        margin: {
+          right: '-11px', // offset 12 px padding + 1px border
```
MikeKingdom: Can this be calculated instead? Don't we do this for secondary buttons?

jcfilben: Good call, just pushed up changes

MikeKingdom: The border width is 1 because the default border size is `theme.global.borderSize['xsmall']`. Whether there is a border seems to come from `theme.dataTable.resize.border` and the border has a `side`. To be accurate and correctly respond to theme settings shouldn't we get that width from here instead of assuming 1px?

jcfilben: Just pushed up a fix to use `theme.global.borderSize.xsmall` instead of `1px`. 

Border size 'xsmall' didn't change with the hpe v7 t-shirt sizes so I did not pull this out as a theme variable and just referenced `theme.global.borderSize.xsmall` directly in the resizer code.

jcfilben: My latest push also includes a fix for `theme.dataTable.resize.border.side = 'start'`, which was previously not working

MikeKingdom (approved):
Thanks!

PR #7808, opened 2025-10-06
(fix): #6991 undefined error of DataChart horizontal bar when the inp…

avantikacode:
#### What does this PR do?

This PR fixes a critical bug in the DataChart component where passing a single data point would cause an "undefined error: Cannot read properties of undefined" instead of rendering a horizontal bar as expected. The fix ensures that DataChart properly handles single data points by adding special case logic in the renderValue function to use the correct data index.

Start by reviewing the changes in src/js/components/DataChart/DataChart.js in the renderValue function (lines 548-571). The key change is the addition of a special case that ensures single data points always use index 0 when accessing data properties, preventing out-of-bounds access that caused the undefined error.

The existing DataChart test suite was run, specifically the "single" test case
The test now passes without the undefined error
Verified that the component renders correctly with single data points
Additional improvements were made to the Chart component's axis generation logic for better single data point handling

1. Create a DataChart component with a single data point:

`<DataChart
  data={[{ value: 10, label: 'Single Point' }]}
  series={['value']}
  axis={{ x: 'label', y: 'value' }}
/>`

2. Verify that:
- No undefined errors are thrown
- A horizontal bar is rendered
- The chart displays the single data point correctly
- Axis labels are properly rendered

[x] screen is used for querying.
[x] The correct query is used. (Refer to this list of queries)
[x] asFragment() is used for snapshot testing.

The issue occurred because when DataChart has only one data point, the axis generation logic creates multiple axis values (0 and 1) to provide a visual range, but the data array only contains one element at index 0. When the renderValue function tried to access data[1], it returned undefined, causing the error. The fix ensures that for single data points, we always use the valid data index (0) regardless of the axis value.

Fixes the issue where DataChart throws "undefined error: Cannot read properties of undefined" when data length is 1, preventing the component from rendering a single horizontal bar as expected.

Data provided: 

  const data = [
    { date: "2020-08-20", amount: 22 },
    // { date: "2020-08-21", amount: 33 },
    // { date: "2020-08-22", amount: 44 }
  ];

image

No, this is a bug fix that doesn't change the API or expected behavior of the DataChart component.

Yes, this should be mentioned as a bug fix in the release notes since it resolves a critical error that prevented DataChart from working with single data points.

This is a backwards compatible bug fix. It doesn't change the component's API or expected behavior - it only fixes the undefined error that was preventing the component from working correctly with single data points.


avantikacode:
Checking why the build failed

avantika-msr:
@britt6612, @jcfilben I have added the fix to this issue ([https://github.com/grommet/grommet/issues/6991](url)) here. Could you please review once and let me know. Thanks!

britt6612:
@avantikacode We will be doing reviews mid October, thank you for contributing!

britt6612:
@avantika-msr 

Hello sorry for the delay but can you help me understand the problem 

I have the following in my code:

```
import React from 'react';

import { Box, DataChart } from 'grommet';

const data = [
  { date: '2020-08-20', amount: 22 },
  // { date: "2020-08-21", amount: 33 },
  // { date: "2020-08-22", amount: 44 }
];

export const Simple = () => (
  // <Grommet theme={grommet}>
  <Box align="center" justify="start" pad="large">
    <DataChart data={data} series={['date', 'amount']} />
  </Box>
  // </Grommet>
);

export default {
  title: 'Visualizations/DataChart/Simple',
};

```

however this is what I see rendered

Screenshot 2025-10-29 at 12 20 54 PM


I just want to make sure I understand the bug. I do not see a 
`"undefined error: Cannot read properties of undefined" `

PR #7825, opened 2025-10-24
update webpack dependency

britt6612:
#### What does this PR do?
update webpack dependency
package.json
locally


britt6612:
@jcfilben 

The small gzip increase comes from a Webpack change in how module IDs determine the final render order. The runtime code is emitted slightly differently, so gzip compresses it a bit less efficiently. The bundle content itself hasn’t changed.

https://github.com/webpack/webpack/pull/19184
https://github.com/webpack/webpack/releases/tag/v5.98.0

jcfilben:
> @jcfilben
>
> The small gzip increase comes from a Webpack change in how module IDs determine the final render order. The runtime code is emitted slightly differently, so gzip compresses it a bit less efficiently. The bundle content itself hasn’t changed.

Thanks for looking into this
I think we just need to bump up the bundlesize in package.json and resolve the merge conflicts

jcfilben (approved):

PR #7830, opened 2025-10-28
Enforce readonly on range calendar when missing onSelect

### LINKED ISSUE #7720: Calendar — `range` without `onSelect` shouldn't change value

None

### Expected Behavior

If Calendar doesn't have `onSelect`, it should be read only: https://v2.grommet.io/calendar#onSelect

### Actual Behavior

Right now, for single dates this behaves as expected. However, for `range` cases, the value still updates: https://storybook.grommet.io/?path=/story/visualizations-calendar-range--range

_No response_

### Steps to Reproduce

1. Go to story
2. Click on calendar day
3. See how range changes

### Your Environment

_No response_

NOMADE55:
#### What does this PR do?
Prevents the use of the calendar when no `onSelect` function was passed as prop.

- [src/js/components/Calendar/Calendar.js](https://github.com/grommet/grommet/blob/d4519a42e4423dd88ed838ea2d999ce197d1b582/src/js/components/Calendar/Calendar.js#L229)

- Tested on Storybook
- Tested on a dummy project

1. Create a range calendar component
2. onSelect should be undefined or not set.
3. Calendar should be 'read only'

- Fixes #7720


src/js/components/Calendar/Calendar.js
```diff
@@ -876,10 +876,16 @@ const Calendar = forwardRef(
               day={day}
               key={day.getTime()}
               disabledProp={dayDisabled}
-              buttonProps={{
-                a11yTitle: day.toDateString(),
-                onClick: dayDisabled ? () => {} : () => onClick(dateObject),
-              }}
+              buttonProps={
```
britt6612: @NOMADE55 

Lets update the selectDate function above to prevent state changes:

```
const selectDate = useCallback(
  (selectedDate) => {
    // If no onSelect prop is provided, the calendar should be read-only
    if (!onSelect) return;
    
    let nextValue;

    if (range || Array.isArray(value?.[0])) {
      nextValue = handleRange(selectedDate);
    } else {
      nextValue = selectedDate;
    }

    nextValue = normalizeOutput(nextValue, outputFormat);
    onSelect(nextValue);
  },
  [handleRange, onSelect, outputFormat, range, value],
);
```

britt6612:
> Hello, thank you for working on this. The changes looks good to me.
> 
> The only thing I notice is that the aria-labels are missing on the buttons from the snapshot.
> 
> @jcfilben @britt6612 is this an expected change?

I left one comment as well but I think the `aria-label` should always be there

maybe we can do 

```
buttonProps={{
  a11yTitle: day.toDateString(),
  ...(onSelect && !dayDisabled && { onClick: () => onClick(dateObject) }),
}}
```

NOMADE55:
Got it @britt6612 , @Sulaymon333.

On the aria topic, there is another ticket which has an open PR https://github.com/grommet/grommet/issues/7721 regarding `aria-disabled`.

Want this PR to also address that issue?

britt6612:
@NOMADE55 you are correct we can take a look at that PR separate then do not worry about it in here.

britt6612:
@NOMADE55 PR looks good I need to look at why the test is failing on https://app.circleci.com/jobs/github/grommet/grommet/90599

@jcfilben ideas?

jcfilben:
Thanks for working on this! A few thoughts here...

If we want the component to be 'readonly' if someone doesn't pass `onSelect` then it seems like we should use the readonly and aria-readonly attributes to indicate to screenreader users that this component is not interactive.

In the original ticket it says "If Calendar doesn't have onSelect, it should be read only... Right now, for single dates this behaves as expected. However, for range cases, the value still updates". However this doesn't seem to always be the case, in this storybook example we have a single select calendar where we do not pass `onSelect`, but the Calendar is not read only and is still interactive.
https://storybook.grommet.io/?path=/story/visualizations-calendar-custom-day--custom-day-calendar

@grommet/devs We should align on the desired behavior here to ensure we don't break existing functionality and move in a direction that feels correct for the long term. Ideally I think that `readonly` should be a separate prop. Not passing `onSelect` shouldn't necessarily default you to read only mode. Open to other thoughts on this though/

Our current documentation for the prop does say that not passing `onSelect` should result in a readonly Calendar, but I'm not sure that this feels correct to me and doesn't align with the components behavior.
Screenshot 2025-11-03 at 4 10 59 PM

NOMADE55:
Agree with @jcfilben thoughts, a readonly prop is more intuitive, I'd keep current behavior with onSelect or be very clear about the new changes in the patch notes though.

britt6612:
> Agree with @jcfilben thoughts, a readonly prop is more intuitive, I'd keep current behavior with onSelect or be very clear about the new changes in the patch notes though.

We are planning to discuss more as a dev group next week will keep you updated!

src/js/components/Calendar/Calendar.js
```diff
@@ -897,18 +903,18 @@ const Calendar = forwardRef(
             <CalendarCustomDay
               key={day.getTime()}
               aria-selected={selected}
-              buttonProps={
-                onSelect
+              buttonProps={{
+                a11yTitle: day.toDateString(),
+                ...(onSelect
```
jcfilben, line 908: Nice fix!

jcfilben:
Discussed as a dev group and we are going to align with the documented behavior, so not passing `onSelect` will result in readOnly behavior. With that I think we are good to move forward with this PR

jcfilben (commented):
In the Calendar/CustomDay example I am still able to select a date even though `onSelect` is not defined

NOMADE55:
> In the Calendar/CustomDay example I am still able to select a date even though `onSelect` is not defined

Weird, I'll check and push a fix!

NOMADE55:
@jcfilben:  the `Calendar` component is not the one handling selection behavior, but the Box within it:

https://github.com/grommet/grommet/blob/11f7778a3098a201db5918c2c1057abf38b538e4/src/js/components/Calendar/stories/CustomDay.stories.js#L10-L28

jcfilben:
> @jcfilben: the Calendar component is not the one handling selection behavior, but the Box within it:

Ah okay, that makes sense

NOMADE55:
🫡

PR #7843, opened 2025-12-04
add tag to anchor story

britt6612:
#### What does this PR do?
This PR adds a `tag` to the Anchor story to test if it works correctly with only being deployed on chromatic and dev mode and not the actual deploy.
anchor/gap.stories
locally see if the story shows up in the deployed storybook 

Look at default tag "test" and see if story does not show up in deployed story. POC
closes https://github.com/grommet/hpe-design-system/issues/5620
no
no
backwards compatible

### DIFF
--- a/storybook/main.js
+++ b/storybook/main.js
@@ -40,9 +40,4 @@ module.exports = {
   typescript: {
     reactDocgen: 'react-docgen-typescript',
   },
-
-  // Exclude internal stories only in production builds
-  tags: {
-    exclude: process.env.NODE_ENV === 'production' ? ['internal'] : [],
-  },
 };
--- a/storybook/preview.js
+++ b/storybook/preview.js
@@ -88,6 +88,12 @@ export const decorators = [
 
 export const parameters = {
   layout: 'fullscreen',
+  tags: {
+    exclude:
+      process.env.NODE_ENV === 'production' && !isChromatic()
+        ? ['internal']
+        : [],
+  },
   options: {
     storySort: (first, second) => {
       /**


src/js/components/Anchor/stories/Gap.stories.js
```diff
@@ -15,6 +15,8 @@ const GapAnchor = () => (
 
 export const Gap = () => <GapAnchor />;
 
+Gap.tags = ['!dev'];
```
jcfilben: Is it possible to have a tag where the story doesn't get included in the deployed version but does get included when we run storybook locally as well as included in chromatic?

britt6612: None of the default Storybook tags will work:

- dev – Added to every story by default. Excluding it (!dev) would hide the story everywhere (local, Chromatic, and production).
- test – Also applied to all stories, so it doesn’t provide any differentiation.
- autodocs – Controls docs generation only; it does not affect story visibility.
- play-fn and test-fn – Added automatically based on whether a story uses those features.

the default tags are either:
applied universally to every story,
applied automatically based on story behavior, or
unrelated to visibility entirely.

I think we need a custom tag to be able to have the story not deployed but included in chromatic.

jcfilben (approved):

storybook/main.js
```diff
@@ -40,4 +40,9 @@ module.exports = {
   typescript: {
     reactDocgen: 'react-docgen-typescript',
   },
+
+  // Exclude internal stories only in production builds
+  tags: {
+    exclude: process.env.NODE_ENV === 'production' ? ['internal'] : [],
+  },
```
jcfilben: Was just looking at the storybook documentation and I also saw this pattern:

```
  tags: {
    internal: {
      defaultFilterSelection: 'exclude',
    },
  },
```

britt6612: I think that would still be excluded from Chromatic but happy to test it out

jcfilben: Sounds good we can move forward with this PR as is, then test this out later if needed

britt6612: @jcfilben actually just looked into this pattern 

```
  tags: {
    internal: {
      defaultFilterSelection: 'exclude',
    },
  },

```
looks like we could do that but not on v8

jcfilben: Ah okay good to know

halocline (approved):

PR #7853, opened 2025-12-19
[Fix] datafilter bug for single data

### LINKED ISSUE #7852: DataFilter bug when there is one item

Yes

### Expected Behavior

When there is only one data item, opening DataFilters should still display the available filter option(s) for that item.

### Actual Behavior

When there is only one data item and the user opens DataFilters, the panel only shows:

The Filters title

The Apply Filters button

The available filter value for the single data item is not shown.

Even with a single data item, users should be able to see and interact with the available filter options. Hiding them creates confusion and makes the filter UI appear empty or broken.

https://codesandbox.io/p/sandbox/v8p5mt?file=%2Fsrc%2FApp.js%3A25%2C1

### Steps to Reproduce

look at the codesandbox then open the filters button

### Your Environment

_No response_

Sulaymon333:
#### What does this PR do?
Consumption Analytics team needs this fix. 

Fixes an issue where the DataFilters panel appears empty when single data item or items with one unique values are present. Previously, opening DataFilters with one item showed only the title and “Apply Filters” button, hiding the available filter option. This update ensures filter options are still displayed and usable even with a single data item or items with one unique values are present, preventing confusion and making the filter UI behave consistently.

**Root Cause**
The original logic checked if (uniqueValues.length < 2) which prevented any filter from rendering when there was fewer than 2 unique values. This was too restrictive.

**Solution**
Modified the range calculation and step normalization logic:

**Key Changes:**
1. Critical Logic Fix - Condition Change
```js
// Before: if (uniqueValues.length < 2) return [undefined, undefined];
// After:  if (uniqueValues.length === 0) return [undefined, undefined];
```
- Now only hides filter when there are truly no values (empty dataset)
- Allows single-value and multi-value filters to render consistently
- Fixes the empty panel issue for single data points

2. Simplified Range Calculation
- Removed complex alignMin/alignMax normalization that was expanding ranges (e.g., [-19.99, -4.895] → [-20, -9.79])
- Preserves actual data values (e.g., [-19.99, -4.859]  becomes [-19.99, -4.86] (maximum of 2 decimal places)

3. Floating-Point Precision Cleanup
- Rounds to maximum 2 decimal places to improve range selector label readability. Before, we can have 2.44434863, After, it will be normalized to 2.44 (maximum of 2 decimal places)
- Applied to both min and max range bounds
- Example: -13.8739999999999999 becomes -13.87

4. Step Value Normalization (RangeSelector rendering)
- Capped step to maximum 2 decimal places
- Added safeguard to ensure step is never <= 0 (RangeSelector requirement)
- Handles single-value case [min === max] gracefully

Benefits:
✅ Fixes original issue: Filters display with single data points
✅ Works with all numeric types: Single values, decimals, zero, negative
✅ Clean UI display: Maximum 2 decimal places everywhere
✅ No prop errors: RangeSelector always receives valid min/max/step
✅ Consistent behavior: RangeSelector used for all numeric ranges

closes #7852 

See [changes](https://github.com/grommet/grommet/blob/a4a9d4a934f461132b373420fbe88a7d5ae50abd/src/js/components/DataFilter/DataFilter.js#L77C1). the condition is now updated to only hide filter if there is truly no values to filter on. for example, empty dataset or empty values/undefined,

image

Yes

Yes


src/js/components/DataFilter/DataFilter.js
```diff
@@ -104,8 +104,8 @@ export const DataFilter = ({
 
     // generate options from all values for property
     const uniqueValues = generateOptions(unfilteredData || data, property);
-    // if less than two values, nothing to filter
-    if (uniqueValues.length < 2) return [undefined, undefined];
+    // if no values, nothing to filter
+    if (uniqueValues.length === 0) return [undefined, undefined];
```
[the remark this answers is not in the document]
britt6612: we can ignore this comment since we talked about it in the office hours

src/js/components/DataFilter/__tests__/__snapshots__/DataFilter-test.tsx.snap
```diff
@@ -2874,6 +2874,109 @@ exports[`DataFilter renders 1`] = `
                 </div>
               </div>
             </div>
+            <div
```
britt6612, line 6173: is this snap shot because we had a test with 1 item?

Sulaymon333: @britt6612 it is due to the property `zero` from the data object used for the test, since all 3 items have: `[zero: 0]`. so unique values generated is `[0]`(only ONE unique value) and it does not satisfy the new changes from DataFilter.js `if (uniqueValues.length === 0) return [undefined, undefined]; `

With the previous code `if (uniqueValues.length < 2) return [undefined, undefined];`, the RangeSlector won't be rendered  because of early return of `[undefined, undefined]` since `[0].length < 2` is truthy but with new changes (`if (uniqueValues.length === 0) return [undefined, undefined];`), the RangeSelector is rendered now and hence changes in the snapshot since `[0].length !== 0`. A closer look at the code execution code flow below for more details. 


```js
// DataFilter.js
if (uniqueValues.length === 0) return [undefined, undefined];
// 1 === 0 is FALSE 
// Does NOT return, continues to next check

if (uniqueValues.some((v) => v !== undefined && typeof v !== 'number'))
  // [0].some(...) → false (0 IS a number)
  // Doesn't return here

// Falls through to numeric range handling
const delta = uniqueValues[1] - uniqueValues[0];  // 0 - 0 = 0
return [undefined, [0, 0]];  //  Returns range!

// so finally
if (range) {  // [0, 0] is truthy 
  content = <RangeSelector ... />  //  RENDERS!
}
```

**Bottom line:** the snapshot chnages is expected and it is OK based on the test data.
image

britt6612: great thank you for this!

britt6612 (approved):

src/js/components/DataFilter/DataFilter.js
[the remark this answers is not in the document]
halocline: Can you add the summary for what the discussion and resulting rationale was? Without the explanation, I'm not sure I agree with this conclusion because of the concerns stated above.

Also, where is this issue documented?

halocline, line 77: Do we have test coverage for the various scenarios this condition is evaluating?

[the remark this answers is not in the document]
britt6612: Here is the issue https://github.com/grommet/grommet/issues/7852

@Sulaymon333 can you put in the PR description about the conversation that was had in office hours?

Sulaymon333: yes we have some coverage based on the current test. that is why there is a snapshot change. We can add more cases like specifically testing with a single data point and so on.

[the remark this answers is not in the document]
britt6612: issue here @halocline 
https://github.com/grommet/grommet/issues/7852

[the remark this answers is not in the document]
Sulaymon333: @halocline the description of this PR is updated now to reflect the rationale.

Sulaymon333: @halocline the range normalization logic is now simplified to cover several cases (negative, zero, floating numbers and so on) and more test cases added.

Sulaymon333: the PR description is further fleshed out to for easier review

halocline:
@Sulaymon333 @britt6612 - I need to think through this a bit more and think this would benefit from a broader discussion.

I'm not convinced that this is the only solution and should consider other design oriented solutions as well. DataFilters are built with a "hybrid" approach which by default provides the current behavior, but also provides the API to allow callers to have full control over the values presented should they need it.

Secondly, this change fundamentally changes existing behavior and will result in breaking changes to existing implementations. We need to ensure that we are very confident in the point of view that the current behavior is truly a bug vs. design intent.

After discussion, we may end up with the point of view that this is in fact the proper direction, but I am reluctant to rush in this direction until the broader team has had a chance to evaluate, discuss, and explore alternatives.

cc: @jcfilben , @MikeKingdom , @oliverHPE , @SOjaHPE

Sulaymon333:
```js
const data = [  
  { price: -15.895, name: "Thin", location: "Warehouse C" },
  { price: -4.895, name: "Thins", location: "Warehouse B" }
]
```

**Before fix**
image
image
image

**After fix**
image
image
image

Sulaymon333:
```js
const data = [  
  { value: 1.1234567, name: "Doodad", location: "Warehouse B" },
  { value: 2.9876543, name: "Doodad Pro", location: "Warehouse A" },
]
```

**Before**
image

**After**
image

src/js/components/DataFilter/DataFilter.js
Sulaymon333: more tests added. so resolving this comment.

Sulaymon333:
based on offline discussion with Brittany.  The direction for this PR is fine based on internal discussion within the team. 

@halocline please can you have another round of review?

master is merged into the branch as well.

jcfilben (commented):
Created a test story with a value with 3 decimal points, when I adjust the filter in this example I would still expect to see `-4.859` in the second screenshot but it doesn't show up.
```
const negativeDecimalData = [{ price: -19.99 }, { price: -4.859 }];

export const Simple = () => (
  <Data data={negativeDecimalData}>
    <DataFilters>
      <DataFilter property="price" />
    </DataFilters>
    <DataTable />
  </Data>
);
```
Screenshot 2026-02-03 at 10 45 55 AM

Screenshot 2026-02-03 at 10 45 48 AM

jcfilben (approved):

src/js/components/DataFilter/__tests__/DataFilter-test.tsx
```diff
@@ -449,4 +449,249 @@ describe('DataFilter', () => {
...
+    );
+
+    const lowerBound = screen.getByRole('slider', {
+      name: 'Lower Bounds',
+    });
+    expect(lowerBound).toBeTruthy();
+    expect(lowerBound.getAttribute('aria-valuenow')).toBe('2.5');
+
+    expect(asFragment()).toMatchSnapshot();
+  });
```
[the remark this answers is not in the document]
Sulaymon333: @britt6612 should we consider adding a test for single positive integer. Although it looks ok but test coverage might be useful.

`test('single positive integer [5] - handles integer correctly')

For a single positive Data I have a the following screenshot. Is this what we are expecting?

image

Sulaymon333:
@britt6612 should we consider adding a story for a single data point  so consumers can see visually what a single data now look like. See use case below.


```js
const negativeDecimalData = [{ price: -19.99 }, { price: -4.859 }];

export const Simple = () => (
  <Data data={negativeDecimalData}>
    <DataFilters>
      <DataFilter property="price" />
    </DataFilters>
    <DataTable />
  </Data>
);
```

Sulaymon333:
Everything else looks good to me aside those two comments above. I am not able to approve the PR since I raise the PR initially.

britt6612:
> @britt6612 should we consider adding a story for a single data point so consumers can see visually what a single data now look like. See use case below.
> 
> ```js
> const negativeDecimalData = [{ price: -19.99 }, { price: -4.859 }];
> 
> export const Simple = () => (
>   <Data data={negativeDecimalData}>
>     <DataFilters>
>       <DataFilter property="price" />
>     </DataFilters>
>     <DataTable />
>   </Data>
> );
> ```

I think the existing stories cover the range concept well enough. If we do add it, it might be more useful as part of a validation or edge case section rather than a standalone story. 
@jcfilben thoughts here

jcfilben:
I'm good with not adding a story for now, this is a bit of an edge case so don't feel like we have to show a story for it

Issue #7851, opened 2025-12-15
autoFocus prop does not focus Select


sktw:
### Is this something you or your team is interested in contributing a pull request towards?

None

### Expected Behavior

Setting the `autoFocus` prop on a `Select` component should focus it on render. This apparently worked previously, according to this [comment](https://github.com/grommet/grommet/issues/3119#issuecomment-620854598).

### Actual Behavior

The `Select` component does not receive focus on render. In a normal browser environment (not CodeSandbox) the active element remains `body`. The react devtools show that the `autoFocus` prop is successfully passed to `TextInput` sub-component.

### URL, Screenshot, or CodeSandbox

https://codesandbox.io/p/sandbox/grommet-v2-template-forked-dwvc8h

### Steps to Reproduce

1. Render a `Select` with the `autoFocus` prop set.
2. Try to open the dropdown by pressing `Space`.

### Your Environment

Grommet version 2.52.0 (sandbox uses 2.44.0 from template)
Chrome, Firefox
Ubuntu 20.04.6

PR #7854, opened 2026-01-07
Fix/select autofocus

rickyentzer:
#### What does this PR do?
Add the autoFocus prop to the Select component so that it can pass it down correctly.
Addresses https://github.com/grommet/grommet/issues/7851

Select.js and propTypes.js

Add test to Select-Test.js

Add a temporary story to storybook

### DIFF
--- a/src/js/components/Select/__tests__/Select-test.js
+++ b/src/js/components/Select/__tests__/Select-test.js
@@ -135,6 +135,9 @@ describe('Select', () => {
     );
 
     expect(container.firstChild).toMatchSnapshot();
+
+    const selectButton = container.querySelector('button');
+    expect(document.activeElement).toBe(selectButton);
   });
 
   [0, null].forEach((value) =>
--- a/src/js/components/Select/index.d.ts
+++ b/src/js/components/Select/index.d.ts
@@ -11,6 +11,7 @@ import {
 export interface BasicSelectProps {
   a11yTitle?: A11yTitleType;
   alignSelf?: AlignSelfType;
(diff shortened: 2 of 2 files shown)


src/js/components/Select/propTypes.js
```diff
@@ -3,6 +3,7 @@ import { genericProps } from '../../utils/general-prop-types';
 
 export const genericSelectProps = {
   ...genericProps,
+  autoFocus: PropTypes.bool,
```
britt6612, line 6: did we add to our TS file as well?

src/js/components/Select/__tests__/Select-test.js
```diff
@@ -122,6 +122,21 @@ describe('Select', () => {
...
+          id="test-select"
+          autoFocus
+          placeholder="test select"
+          options={['one', 'two']}
+        />
+      </Grommet>,
+    );
+
+    expect(container.firstChild).toMatchSnapshot();
+  });
```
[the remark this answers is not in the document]
britt6612: good comment can we add a assertion that the focus style is there?

.gitattributes
jcfilben, line 1: Can we remove this change from this PR and instead just handle it in https://github.com/grommet/grommet/pull/7845/

rickyentzer:
Removed the .gitattributes commit (branched off the wrong branch) and added the focus test Brittany suggested.

jcfilben:
Looks like some extra changes are getting pull into the PR, maybe try updating your branch with the master branch

PR #7857, opened 2026-01-20
fix gap type to be optional

britt6612:
#### What does this PR do?
fixes the `gap` on List to be an optional type
base.d.ts
locally 

https://github.com/grommet/grommet/pull/7754

There was a theme object added for users to be able to control the `gap` via theme and the `gap` was declared as required in the TS theme file. This should be optional for users to be able to change the `gap` if they need/want.
brought up in this ticket 
https://github.com/grommet/grommet/pull/7754
no
yes
backwards compatible

### DIFF
--- a/src/js/components/List/__tests__/List-test.tsx
+++ b/src/js/components/List/__tests__/List-test.tsx
@@ -363,9 +363,6 @@ describe('List', () => {
   test('renders custom theme for primaryKey', () => {
     const theme = {
       list: {
-        item: {
-          gap: 'xsmall',
-        },
         primaryKey: {
           color: 'brand',
           weight: 500,


src/js/components/List/__tests__/List-test.tsx
```diff
@@ -363,6 +363,9 @@ describe('List', () => {
   test('renders custom theme for primaryKey', () => {
     const theme = {
       list: {
+        item: {
```
jcfilben: Why doesn't this change result in a snapshot change?

britt6612: sorry used value that was default changing now

britt6612:
@ShimiSun don't feel like you need to review just tagging for visibility

PR #7873, opened 2026-02-19
Revert back to node 20 docker img

jcfilben:
#### What does this PR do?
Revert back to cimg/node:20.19.0-browsers to try and resolve this issue:

Screenshot 2026-02-19 at 2 40 45 PM

https://app.circleci.com/pipelines/github/grommet/grommet/17333/workflows/b7113869-da46-4e1c-91d5-02cc100d0d51/jobs/91564

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ orbs:
 defaults: &defaults
   working_directory: ~/grommet-ci
   docker:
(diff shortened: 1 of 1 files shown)


britt6612:
> I don't want to hold this up, but I'm still curious about why newer docker node images have this problem

agree maybe we can make a draft ticket in backlog so we can understand better

PR #7905, opened 2026-03-27
fix(Button): color prop on default Button sets text color, not background

### LINKED ISSUE #7880: Button - default color
According to the documentation when the color prop is set for a default Button the text color will change. Instead the background color is changing.

Documentation: https://v2.grommet.io/button#color

Codesandbox example: https://codesandbox.io/p/sandbox/grommet-v2-template-forked-5ds92d?file=%2Findex.js%3A8%2C22

Gdhanush-13:
## What

Fixes #7880 - Setting `color` prop on a default Button incorrectly changes the background color instead of the text color.

## Root Cause

In `kindPartStyles()` (`styles.js`), when the theme object has a `background` property, `colorValue` replaces it via `backgroundStyle(colorValue || obj.background, ...)`. This is correct for primary/secondary buttons (where `color` should set the background), but wrong for default buttons (where `color` should set the text color per the docs).

The same issue exists in `getIconColor()` (`Button.js`), where `colorProp` replaces the background for icon color calculation.

## Fix

**StyledButtonKind.js** - `kindStyle()`: Detect default kind theme paths and:
- Don't pass `colorValue` to `kindPartStyles()` (prevents background replacement)
- Apply `colorValue` as text color via `normalizeColor()`  

**Button.js** - `getIconColor()`: For default kind paths:
- Don't use `colorProp` as background replacement
- Use `colorProp` as text color so icon matches the label

## Behavior

| Scenario | Before | After |
|----------|--------|-------|
| `<Button color='red' label='Test' />` (default) | Background turns red | Text turns red |
| `<Button primary color='red' label='Test' />` | Background turns red | Background turns red (unchanged) |

## Tests

Added test `default button color prop sets text color, not background` with a theme defining `button.default` with background, verifying the rendered output via snapshot.

### DIFF
--- a/src/js/components/Button/__tests__/Button-test.tsx
+++ b/src/js/components/Button/__tests__/Button-test.tsx
@@ -682,4 +682,29 @@ describe('Button', () => {
 
     expect(screen.getByLabelText('Add')).toBeInTheDocument();
   });
+
+  test('default button color prop sets text color, not background', () => {
+    const theme: ThemeType = {
+      button: {
+        default: {
+          background: { color: 'white' },
+          border: { color: 'border', width: '2px' },
+          color: 'text',
+        },
+      },
+    };
+
+    const { container } = render(
+      <Grommet theme={theme}>
+        <Button label="Default" color="red" />
+      </Grommet>,
+    );
+    const button = container.querySelector('button')!;
+    // color prop should set text color, not background
+    expect(button).toHaveStyleRule('color', expect.stringContaining('red'));
+    expect(button).not.toHaveStyleRule(
+      'background-color',
+      expect.stringContaining('red'),
+    );
+  });
 });
--- a/src/js/components/Button/__tests__/Button-test.tsx
+++ b/src/js/components/Button/__tests__/Button-test.tsx
@@ -696,15 +696,15 @@ describe('Button', () => {
 
     const { container } = render(
       <Grommet theme={theme}>
-        <Button label="Default" color="red" />
+        <Button label="Default" color="#FF0000" />
       </Grommet>,
     );
     const button = container.querySelector('button')!;
     // color prop should set text color, not background
-    expect(button).toHaveStyleRule('color', expect.stringContaining('red'));
+    expect(button).toHaveStyleRule('color', expect.stringContaining('#FF0000'));
     expect(button).not.toHaveStyleRule(
       'background-color',
-      expect.stringContaining('red'),
+      expect.stringContaining('#FF0000'),
     );
   });
 });
--- a/src/js/components/Button/Button.js
+++ b/src/js/components/Button/Button.js
@@ -82,9 +82,19 @@ const getIconColor = (paths = [], theme, colorProp, kind) => {
     }
 
     if (obj) {
+      // For default kind buttons, color prop should be used as text color,
+      // not as background replacement.
+      const isDefaultKind =
+        paths[index] === 'default' ||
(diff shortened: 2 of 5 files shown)


src/js/components/Button/__tests__/Button-test.tsx
```diff
@@ -682,4 +682,26 @@ describe('Button', () => {
...
+
+    const { container } = render(
+      <Grommet theme={theme}>
+        <Button label="Default" color="red" />
+      </Grommet>,
+    );
+    const button = container.querySelector('button');
+    const style = window.getComputedStyle(button!);
+    // color prop should set text color, not background
+    expect(style.color).not.toBe('');
```
britt6612: This test doesn't actually catch the bug. `window.getComputedStyle` in jsdom doesn't resolve styled-components CSS class rules, so `style.color` is always so... `expect(style.color).not.toBe('')` is always true regardless of what the component renders.

I would recommend creating a snapshot or a more specific assertion for the style (for example to look at something like not to haveStyleRile (background-color: red)).

src/js/components/Button/__tests__/Button-test.tsx
```diff
@@ -682,4 +682,26 @@ describe('Button', () => {
 
     expect(screen.getByLabelText('Add')).toBeInTheDocument();
   });
+
+  test('default button color prop sets text color, not background', () => {
```
britt6612, line 686: can we also add a test to the `button-kind-test.js`

britt6612:
@Gdhanush-13 can you do me a favor and pull master from upstream on your fork then run `yarn` just testing something out with Chromatic

britt6612:
> @Gdhanush-13 can you do me a favor and pull master from upstream on your fork then run `yarn` just testing something out with Chromatic

@Gdhanush-13 Im so sorry that did not do what I was expecting is there any way you can undo the last commit without the `yarn.lock`

Gdhanush-13:
@britt6612 Done! I've removed the `yarn.lock` commit via rebase and force-pushed. The branch now only contains the fix commit and the test commit. Sorry about that!

britt6612 (approved):
Looks good!

britt6612:
@Gdhanush-13 can you look at the failing test also we are going to have a dev discussion regarding this behavior

Gdhanush-13:
Done - pulled upstream master, ran `yarn`, and pushed. Also found & fixed a bug where the `color` prop was leaking to `background-color` in active/disabled states for default kind buttons (the `isDefaultKind` check was per-themePath, so state paths like `active` bypassed it).

Gdhanush-13:
@britt6612 I've rebased onto upstream/master with DCO sign-off on all commits, and confirmed all Button tests pass locally (45 kind tests + 67 general tests, 0 failures).

Regarding the dev discussion - GitHub Discussions isn't enabled on this repo. Would you prefer I open a GitHub issue to discuss the `color` prop behavior on default-kind buttons, or would you like to discuss in a different channel? Happy to write up a summary of the current behavior vs proposed behavior to kick off the conversation.

britt6612:
@Gdhanush-13 thank you so much, and sorry for the confusion! I meant that the Grommet devs are planning to have a deeper discussion on this PR.

We’re a bit concerned about potential backward compatibility, especially since this behavior has existed for 5+ years even though your change aligns with the docs.

For now, there’s nothing further needed from your side. We just want to proceed cautiously while we make a final decision. We’ll keep you updated!

PR #7906, opened 2026-03-29
chore: add .gitattributes to enforce LF line endings

### LINKED ISSUE #3437: Add a .gitattributes file to enforce line endings
### Expected Behavior
People using Windows and Linux can commit with the same line endings (LF)

### Actual Behavior
Windows users with mis-configured git clients commit files with CRLF line endings.

Example:
https://github.com/grommet/grommet/issues/3394#issuecomment-539166085

### Steps to Reproduce
Clone grommet from a Windows PC, then try to commit some changes.

### Your Environment
- Operating System and version (desktop or mobile): Windows 

### Solution
A simple .gitattributes file that enforce line endings for text files. Github also has an article about it: https://help.github.com/en/articles/configuring-git-to-handle-line-endings

abhinandanas:
## What does this PR do?
Adds a `.gitattributes` file to enforce consistent LF line 
endings across the repository.

## Why?
Windows users with misconfigured git clients were committing 
files with CRLF line endings, causing inconsistencies across 
platforms.

## Changes
- Added `.gitattributes` with LF normalization for JS, TS, 
CSS, SCSS, JSON, YAML, MD and HTML files
- Windows-specific files (.bat, .cmd) retain CRLF

Closes #3437

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,18 @@
+# Enforce consistent line endings
+* text=auto eol=lf
+
+*.js text eol=lf
+*.jsx text eol=lf
+*.ts text eol=lf
+*.tsx text eol=lf
+*.css text eol=lf
+*.scss text eol=lf
+*.json text eol=lf
+*.md text eol=lf
+*.yml text eol=lf
+*.yaml text eol=lf
+*.html text eol=lf
+
+# Windows scripts keep CRLF
+*.bat text eol=crlf
+*.cmd text eol=crlf


britt6612 (commented):
Thanks for adding this! Do we need all the per-extension rules, or can we simplify to just `* text=auto`? Since the global rule already covers all text files, the individual lines seem redundant. The .bat/.cmd exceptions are also probably unnecessary since grommet doesn't have any Windows batch scripts. Happy to be corrected if there's a reason for the extra specificity!

PR #7912, opened 2026-04-03
fix(Tabs): remove state variables from useLayoutEffect deps to prevent infinite loop

Gdhanush-13:
## What
Fixes #7908 - Tabs component crashes with `Maximum update depth exceeded` due to infinite re-renders in `useLayoutEffect`.

## Root Cause
The `useLayoutEffect` at line 242 in `Tabs.js` had `overflow`, `disableLeftArrow`, and `disableRightArrow` in its dependency array. Inside the effect:
- `onResize()` calls `setOverflow()`
- `updateArrowState()` calls `setDisableLeftArrow()` and `setDisableRightArrow()`

Since these state variables are both **set inside** the effect and listed as **dependencies**, every state update triggers the effect again, creating an infinite loop.

## Fix
Removed `overflow`, `disableLeftArrow`, and `disableRightArrow` from the dependency array. These are **outputs** of the effect, not inputs. The effect still correctly re-runs when external inputs change (`tabRefs`, `activeIndex`, `headerRef`, `updateArrowState`).

## Files Changed
- `src/js/components/Tabs/Tabs.js` - useLayoutEffect dependency array (line 253)


src/js/components/Tabs/__tests__/Tabs-test.tsx
```diff
@@ -370,6 +370,28 @@ describe('Tabs', () => {
     expect(asFragment()).toMatchSnapshot();
   });
 
+  test('should not infinite loop when tabs overflow', () => {
```
jcfilben: This test is passing even if I revert the change made in Tabs.js, so I don't think it is acuratly testing the behavior that we want it to

Gdhanush-13:
Good catch @jcfilben - you're right, the previous test wasn't actually exercising the overflow path because JSDOM returns 0 for both scrollWidth and offsetWidth, so the if (scrollWidth > offsetWidth) branch never fired.

I've updated the test to mock HTMLElement.prototype.scrollWidth (1000) and offsetWidth (500) so the overflow detection actually triggers. With the old deps (overflow, disableLeftArrow, disableRightArrow), this would cause the state?effect?state infinite loop. With the fix (those deps removed), it renders cleanly.

jcfilben:
Thanks for updating the test! It looks like there's an issue with the Tabs theme tab gap snapshot not passing now: https://app.circleci.com/pipelines/github/grommet/grommet/17394/workflows/18e28c98-a6a6-4059-ae14-2f09afae9837/jobs/92010

Gdhanush-13:
@jcfilben Thanks for flagging! I've updated the Tabs theme tab gap snapshot — it should be passing now. Let me know if anything else needs attention.

britt6612:
> @jcfilben Thanks for flagging! I've updated the Tabs theme tab gap snapshot — it should be passing now. Let me know if anything else needs attention.

Two quick changes would make it more reliable:

Replace the `console.error` spy with `expect(() => render(...)).not.toThrow()`
Move the `Object.defineProperty` cleanup into a try/finally so it always restores even if the assertion fails

Gdhanush-13:
@britt6612 Done! Both changes applied:
- Replaced console.error spy with expect(() => render(...)).not.toThrow()
- Wrapped Object.defineProperty cleanup in 	ry/finally so it always restores

All 21 Tabs tests pass locally (including snapshots).

britt6612 (approved):
@Gdhanush-13 thank you so much looks good!

jcfilben (commented):
I noticed the test is still passing when I revert the change, but looking into it more it sounds like this might not be something that can be accurately tested with jsdom. I think we could just go ahead and remove he test. Sorry for the back and forth here!

src/js/components/Tabs/__tests__/Tabs-test.tsx
AntoineDuComptoirDesPharmacies: @Gdhanush-13 just in case you did not see the PR comment 👍

Gdhanush-13:
Hi @jcfilben and @AntoineDuComptoirDesPharmacies — pushed the final change now. Removed the regression test as requested since jsdom cannot accurately simulate scrollWidth/offsetWidth to reproduce the infinite loop. All 20 remaining Tabs tests pass locally with 25 clean snapshots. Should be ready to merge!

AntoineDuComptoirDesPharmacies (approved):
Looks good to me 👍

jcfilben (approved):
Looks good!

AntoineDuComptoirDesPharmacies:
Hi @britt6612, do you need anything to review/merge this pull request ?
Thank you !
Yours faithfully,
LCDP

britt6612 (approved):

britt6612:
> Hi @britt6612, do you need anything to review/merge this pull request ? Thank you ! Yours faithfully, LCDP

Yes sorry about that delay with will merge!

PR #7593, opened 2025-04-17
fix #7540 avatar role removal and auto generated aria-label for avatar text

Sulaymon333:
#### What does this PR do?
- fix #7540 
- avatar role removal 
- auto generated aria-label for avatar text using letters/{children} prop passed down 
image

#### Do Jest tests follow these best practices?

- [ ] `screen` is used for querying.
- [ ] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [ ] `asFragment()` is used for snapshot testing.

PR #7917, opened 2026-04-09
fix avatar accessibility

### LINKED ISSUE #7540: Avatar role update
### Link to problem

https://v2.grommet.io/avatar

### Current problem

Avatar currently attach `role=figure` to Avatar container and `role=presentation` which remove the image completely from the accessibility tree. Common pattern from popular React frameworks is that no role is used with the avatar component

### Expected behavior or solution

Proposal to remove the current role attributes on the Avatar component

### Screenshot or recording of problem

Image

### Steps to reproduce

1. Go to https://v2.grommet.io/avatar
2. Inspect and see the source code
3. check accessibility tree and you see that the Avatar image is completely missing

britt6612:
#### What does this PR do?
Fixes incorrect ARIA role usage on the Avatar component when a `src`  image is provided. Previously, the container had `role="figure"` and the inner  had `role="presentation", which completely hid the image from the accessibility tree. This caused VoiceOver to announce "empty figure" or "Eric Soderberg, image, empty figure" redundant, inconsistent, and confusing.

The fix moves the accessible name to the native `img alt` attribute and removes both roles entirely.
avatar.js
tested with voice over on macOS 
before: announced "empty figure" if no `a11ytitle` was given or the Eric Soderberg, image empty figure

after: 
Eric Soderberg image
this can be manually tested with storybook and voiceover

This is a semantics/quality fix rather than a strict WCAG failure. The accessible name was technically present via `aria-label` in the labeled case, but `role="figure"` with no caption relationship and `role="presentation"` on the inner image created AT inconsistencies. The native `` pattern is simpler and more robust across screen readers.

See WCAG 4.1.2 (Name, Role, Value) and ARIA best practices don't override native semantics when the native pattern already handles it correctly.
closes #7540 
yes
The visual is backwards compatible


src/js/components/Avatar/Avatar.js
```diff
@@ -63,19 +63,18 @@ const Avatar = ({
...
-        role={typeof src === 'string' ? 'figure' : undefined}
-        a11yTitle={a11yTitle || ariaLabel}
-        {...avatarProps}
-        {...passThemeFlag}
-        {...rest}
-      >
+      <StyledAvatar {...avatarProps} {...passThemeFlag} {...rest}>
         {content}
       </StyledAvatar>
     );
```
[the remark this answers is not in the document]
britt6612: fixed

src/js/components/Avatar/Avatar.js
jcfilben, line 61: Should the ariaLabel be passed here too?

britt6612: added

src/js/components/Avatar/Avatar.js
```diff
@@ -51,27 +55,34 @@ const Avatar = ({
   }
 
   let content;
+  let hasImage = false;
   if (typeof children === 'string') {
     content = (
       <StyledAvatarText
         alignSelf="center"
+        aria-hidden={a11yTitle || ariaLabel ? true : undefined}
```
jcfilben, line 63: Why do we have aria-hidden here?

src/js/components/Avatar/Avatar.js
```diff
@@ -51,27 +55,34 @@ const Avatar = ({
...
       >
         {children}
       </StyledAvatarText>
     );
   } else if (typeof src === 'string') {
+    hasImage = true;
     content = (
-      <Image role="presentation" fit="contain" src={src} {...imageProps} />
+      <Image
+        alt={a11yTitle || ariaLabel || ''}
```
jcfilben, line 74: If there isn't an a11yTitle or ariaLabel should alt just be undefined instead of `''`?

src/js/components/Avatar/Avatar.js
jcfilben, line 90: Does StyledAvatarText end up getting wrapped by StyledAvatar?

src/js/components/Avatar/Avatar.js
britt6612: The outer StyledAvatar already carries the accessible label via
`aria-label`. Hide the raw initials text so the voice over doesn't announce
 both the label and the initials characters separately.

src/js/components/Avatar/Avatar.js
britt6612: so this was a comment from co-pilot this is what it told me 

```
alt="" → attribute is present with empty string → screen reader treats the image as decorative and skips it entirely ✅
alt={undefined} → attribute is omitted from the DOM → screen reader falls back to announcing the image filename or [src] URL ❌
```

src/js/components/Avatar/Avatar.js
britt6612: yes:

 When children is a string, content is StyledAvatarText.
 When src is a string, content is Image.
 In both cases, content is rendered as a child of StyledAvatar below.

jcfilben (commented):
Minor comment but in the stories can we switch to use aria-label instead of a11yTitle. I think we should be encouraging the use of aria-label instead of a11yTitle

src/js/components/Avatar/stories/CustomThemed/Themed.stories.js
```diff
@@ -34,7 +34,7 @@ export const Themed = () => {
         pad="large"
         background="dark-2"
       >
-        <Avatar src={src} a11yTitle="Shimi" size="myLarge" />
+        <Avatar src={src} ariaLabel="Shimi" size="myLarge" />
```
jcfilben: Should these be aria-label instead of ariaLabel?

britt6612: yes updated

jcfilben (approved):

src/js/components/Avatar/Avatar.js
Sulaymon333, line 99: @britt6612 Since we are already touching this file, I would suggest a lightweight cleanup that makes the rendering logic easier to follow - with **zero regression:** See code below.

- Removed `useCallback` for `AvatarChildren` - `rest` in the dependency array is a new object every render since each render produces  a new object reference, so the memoization never actually cached. It was pure overhead. The JSX is inlined directly instead.
- Moved the `height/width` warning into `useEffect` - avoids firing on every re-render; triggers once on mount and only again if the values actually change.
- Extracted a label variable - `a11yTitle || ariaLabel || undefined` was repeated four times; single derivation eliminates duplication.
- Replaced intermediate `content/hasImage` variables with early returns - each branch is self-contained, no need to build up state across conditionals and render at the end.

All behavioral details are preserved - prop spreading order, default values, conditional `aria-label / passThemeFlag` per branch - so this is a pure refactor with zero reflection.

```js
import React, { useEffect, useMemo } from 'react';

import { Image } from '../Image';
import { StyledAvatar, StyledAvatarText } from './StyledAvatar';
import { AvatarPropTypes } from './propTypes';

const Avatar = ({
  a11yTitle,
  'aria-label': ariaLabel,
  align = 'center',
  children,
  height,
  imageProps,
  justify = 'center',
  round = 'full',
  size = 'medium',
  src,
  width,
  ...rest
}) => {
  const { theme, passThemeFlag } = useThemeValue();
  const avatarSize = theme.avatar.size[size] || size;
  const avatarTextSize = theme.avatar.text.size[size] || 'large';
  const label = a11yTitle || ariaLabel || undefined;

  useEffect(() => {
    if (height || width) {
      console.warn(
        'Avatar should use `size` instead of `height` or `width` props',
      );
    }
  }, [height, width]);

  const avatarProps = useMemo(
    () => ({
      align,
      height: avatarSize,
      justify,
      overflow: 'hidden',
      round,
      width: avatarSize,
    }),
    [align, avatarSize, justify, round],
  );

  const isTextContent = typeof children === 'string';
  const isImageContent = !isTextContent && typeof src === 'string';

  if (isTextContent) {
    return (
      <StyledAvatar
        aria-label={label}
        {...avatarProps}
        {...passThemeFlag}
        {...rest}
      >
        <StyledAvatarText
          alignSelf="center"
          aria-hidden={label ? true : undefined}
          size={avatarTextSize}
          {...passThemeFlag}
        >
          {children}
        </StyledAvatarText>
      </StyledAvatar>
    );
  }

  if (isImageContent) {
    return (
      <StyledAvatar
        {...avatarProps}
        {...passThemeFlag}
        {...rest}
      >
        <Image
          alt={label || ''}
          fit="contain"
          src={src}
          {...imageProps}
        />
      </StyledAvatar>
    );
  }

  return (
    <StyledAvatar
      aria-label={label}
      {...avatarProps}
      {...rest}
    >
      {children}
    </StyledAvatar>
  );
};

Avatar.propTypes = AvatarPropTypes;
export { Avatar };
```

Sulaymon333 (commented):
I would suggest a lightweight cleanup to `Avatar.js`. Please see the code suggestion attached. It is a zero regression refactor.

Sulaymon333:
I am closing #7593 since they are both solving same issue.

src/js/components/Avatar/Avatar.js
britt6612: @Sulaymon333  Appreciate the cleanup suggestions. I agree there are some nice improvements here, but I’d prefer to keep this PR scoped specifically to the issue being fixed to keep review/risk minimal. Since these are primarily refactor changes, I think a separate cleanup issue/PR would be the better place for them.


@jcfilben

Sulaymon333: Ok noted. Just to clarify, the refactor is on top of your changes and does not produce any regression.

Happy to move forward with what we have now in this PR if that is what the team lean towards.

Sulaymon333 (approved):

src/js/components/Avatar/Avatar.js
Sulaymon333: May be we can have a follow on ticket for refactor at some time.

britt6612:
https://github.com/grommet/grommet/issues/7949

PR #7921, opened 2026-04-13
Fix up bounds on calendar stories so they are usable

### LINKED ISSUE #5959: Fix Grommet Calendar Storybook examples and remove broken range/bounds states
## Summary
Current Calendar stories are creating confusion and making the component difficult to evaluate. Some examples appear broken, misleading, or dependent on stale date assumptions.

This work focuses on making Storybook a reliable evaluation surface again by removing or correcting broken examples, updating stale defaults, and ensuring at least one simple working date-range example is easy to understand.

## Goals
- Make Calendar stories usable in current Storybook
- Remove misleading or stale example behavior
- Provide a simple working example for common date-range usage

## Tasks
- Audit existing Calendar stories for broken or misleading states
- Remove or rewrite stories with outdated bounds or stale default dates
- Add one or more simple, realistic working date-range stories
- Ensure retained examples are understandable without code inspection
- Validate basic usability of the stories after updates

## Done when
- Calendar stories can be exercised successfully in Storybook
- Broken or misleading examples are removed or corrected
- At least one recommended working date-range example exists

MikeKingdom:
#### What does this PR do?
Currently a number of the `Calendar` stories have `bounds` which make the stories pretty much unusable and not good examples.
The bounds tend to be months to years ago such that the user is left not being able to navigate or select dates since they are disabled.

These changes just set the `bounds` on the problematic stories to 90 days that end on the current date. This makes a more practical example that allows you to select dates and/or navigate around that timeframe.

Storybook Visualizations->Calendar->*

Storybook

Storybook

Fixes [hpe-design/5959](https://github.com/grommet/hpe-design-system/issues/5959)


MikeKingdom:
We'll need to discuss what to do about the chromatic tests. These storybook examples basically display the current month and vary which dates are enabled. Can we remove them from chromatic and/or potentially make Simple and 1 of the heading related ones not have bounds?

jcfilben:
I think you can disable a particular part of the snapshot, so you can tell it to ignore changes in a specific div for example

src/js/components/Calendar/stories/CustomThemed/CustomSizeHeading.stories.js
```diff
@@ -11,6 +11,14 @@ const customHeading = deepMerge(grommet, {
   },
 });
 
+
+const end = new Date();
+const start = new Date(Date.now() - 1000 * 60 * 60 *24 * 90); // 90 days ago
+const bounds = [
+  `${start.getFullYear()}-${start.getMonth() + 1}-${start.getDate()}`,
```
britt6612, line 19: doesnt `start.getMonth() + 1` not return the 0 for the single digit months?

so for January it would be `1` and not `01` 
does that matter?

MikeKingdom: Doesn't need a leading 0. 
image

britt6612: got it thanks ☝️

MikeKingdom:
> I think you can disable a particular part of the snapshot, so you can tell it to ignore changes in a specific div for example

Thanks. Given that basically the whole thing will be different as months change, I chose to have the month it shows and bounds be fixed to a date so we can have stable chromatic screen shots. I think this will make it more useful as an example (you can click on dates on the month it shows and navigate around etc) but will be stable for the tests.

jcfilben (approved):

britt6612 (approved):
LGTM

PR #7931, opened 2026-04-15
fix(FormField): link error message to input via aria-describedby

### LINKED ISSUE #7927: FormField: error message should be linked to input via aria-describedby

Yes

### Expected Behavior

When a FormField has an error prop, the error message element should have a generated id (e.g. grommet-{id}__error) and the input should have aria-describedby pointing to it. This way a screen reader user landing on an invalid field will automatically hear the error message text, not just that the field is invalid.

### Actual Behavior

aria-invalid="true" is correctly set on the input, but the error message has no id and the input has no aria-describedby. The screen reader user is told the field is invalid but is not read the error message.

_No response_

### Steps to Reproduce

Render a FormField with an error prop. Inspect the error message element — no id attribute. Inspect the input — no aria-describedby attribute.

### Your Environment

latest grommet

akinshaywai:
## Summary

Fixes #7927.

When a `FormField` has an `error` prop and an `htmlFor` prop, screen readers now automatically read the error text when the invalid field is focused.

## What changed

**`FormField.js`**

1. The error `Message` element receives `id="grommet-{htmlFor}__error"` when both `error` and `htmlFor` are present.
2. The `Input` component receives `aria-describedby="grommet-{htmlFor}__error"` when in error state, falling back to any caller-supplied `aria-describedby` when there is no error.
3. The `Message` component now correctly places any `id` prop on the **outermost** rendered element in both the plain and icon/container variants.

## Before / After

| | Before | After |
|---|---|---|
| Input | `aria-invalid="true"` (no `aria-describedby`) | `aria-invalid="true" aria-describedby="grommet-email__error"` |
| Error message | no `id` | `id="grommet-email__error"` |
| Screen reader | announces "invalid" — no error text | announces "invalid — Please enter a valid email address" |

## Test plan

- [ ] Render a `FormField` with `htmlFor="email"` and `error="Please enter a valid email address"`
- [ ] Inspect the input — confirm `aria-describedby="grommet-email__error"` is present
- [ ] Inspect the error message element — confirm `id="grommet-email__error"` is present
- [ ] Tab to the field with VoiceOver or NVDA — confirm the error text is announced automatically
- [ ] Confirm no change when `error` is absent (no `aria-describedby` added)


britt6612:
@akinshaywai thanks for getting this started!
Can you look at the comments that co-pilot left I believe there are some valid points also looks like there are some linting errors

```
/home/circleci/grommet-ci/src/js/components/FormField/FormField.js
  147:12  error  Do not nest ternary expressions                      no-nested-ternary
  158:9   error  Do not nest ternary expressions                      no-nested-ternary
  357:28  error  'adjustedHtmlFor' was used before it was defined     no-use-before-define
  708:1   error  This line has a length of 81. Maximum allowed is 80  max-len
```

Sulaymon333:
@britt6612, I pushed a follow-up [commit](https://github.com/grommet/grommet/pull/7931/changes/0df37625146b9c24e723ba8e8162468dae8a2ce9) that extends this fix beyond the specific issues caught by copilot - the same "last spread wins" issue existed in 3 more places found doing a broader audit:

1. Internal `Input` fallback (what you flagged) — now merges `aria-describedby` instead of replacing it, so any caller-supplied value is preserved alongside the error id.
2. The error `Message`'s theme props (`{...formFieldTheme.error}`) were spread after the computed `id` - a theme with an `id` key would have silently broken the accessibility link. Fixed by reordering.
3. Same issue in the icon/container `Message` variant (`{...containerProps}`).
4. The `aria-describedby` injected onto children (the `TextInput`/`Select` pattern) also overwrote instead of merged - fixed to space-join with any existing value, same as (1).
5. CI then caught a remount regression from (4) - cloning children changed their reconciliation key once an error appeared, wiping input state. Fixed by merging it into the existing `plain`/`focusIndicator`/`pad` `Children.map` pass instead of a separate one.

Test coverage: 5 tests covering the children pattern, the merge-not-overwrite behavior in both code paths (children and internal `Input`), custom (non-Grommet) input components correctly *not* being linked, and the remount regression in (5).

src/js/components/FormField/FormField.js
```diff
@@ -342,6 +352,11 @@ const FormField = forwardRef(
           invalid={!!error}
           name={name}
           label={component === CheckBox ? label : undefined}
+          aria-describedby={
+            error && htmlFor
+              ? `grommet-${adjustedHtmlFor}__error`
```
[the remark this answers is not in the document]
Sulaymon333: Resolved in this [commit](https://github.com/grommet/grommet/pull/7931/changes/0df37625146b9c24e723ba8e8162468dae8a2ce9) by reordering the `adjustedHtmlFor` computation above its first use.

src/js/components/FormField/FormField.js
```diff
@@ -342,6 +352,11 @@ const FormField = forwardRef(
...
+              : rest['aria-describedby']
+          }
           {...rest}
```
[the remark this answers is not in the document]
Sulaymon333: Resolved in this [commit](https://github.com/grommet/grommet/pull/7931/changes/0df37625146b9c24e723ba8e8162468dae8a2ce9)

src/js/components/FormField/FormField.js
```diff
@@ -687,7 +702,12 @@ const FormField = forwardRef(
           </>
         ) : undefined}
         {contents}
-        <Message type="error" message={error} {...formFieldTheme.error} />
+        <Message
+          type="error"
+          message={error}
+          id={error && htmlFor ? `grommet-${adjustedHtmlFor}__error` : undefined}
+          {...formFieldTheme.error}
```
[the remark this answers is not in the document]
Sulaymon333: Resolved in this [commit](https://github.com/grommet/grommet/pull/7931/changes/0df37625146b9c24e723ba8e8162468dae8a2ce9)

britt6612:
@sulaymon-tajudeen-hpe thank you can you look into the test that is failing please?

Sulaymon333:
> @sulaymon-tajudeen-hpe thank you can you look into the test that is failing please?

The failing test is fixed now.

Sulaymon333:
Follow-up (out of scope here): 

I notice `aria-invalid` is only set on the no-children/internal-`Input` pattern, and not on the children pattern (`<FormField error><TextInput/></FormField>`), which is quite common usage. This is a pre-existing issue, not introduced by this PR. 

Worth a separate issue since `aria-invalid` is a distinct a11y signal (state) from `aria-describedby` (description) — We can have a look a follow up issue if it is worth separating otherwise, we can address in same PR since it is related enough.

cc: @jcfilben

britt6612:
Found a related gap while testing this: `aria-invalid` is only set on the internal `Input` wrapper (`FormField.js:204`) never on Grommet input children via the `cloneElement` path. So any `FormField` using `children` (`TextInput`, `MaskedInput`, `Select`, etc.) never gets `aria-invalid`, even with an active error.

Confirmed with VoiceOver on the Email field ( in the story `Input/Form/Controlled`, required error): it announces "Email, required, required, edit text" no "invalid data" at all. Users get no signal the field actually failed validation.

Since this directly affects the accessibility outcome we're fixing here, would you mind adding `aria-invalid` alongside `aria-describedby` in the same `cloneElement`?

Sulaymon333:
`aria-invalid` added via this [commit](https://github.com/grommet/grommet/pull/7931/changes/e1a808d847021f87f2be45c9aaec9489d417b673)

britt6612:
@Sulaymon333 thank you for adding that now the voice over is clear in saying "email required, required invalid data"

britt6612 (approved):

src/js/components/FormField/FormField.js
```diff
@@ -137,24 +137,40 @@ const Message = ({ error, info, message, type, ...rest }) => {
...
+    }
+    if (id) {
+      return typeof message === 'string' ? (
+        <Text id={id} {...contentRest}>
+          {message}
+        </Text>
+      ) : (
+        <Box id={id} {...contentRest}>
+          {message}
+        </Box>
```
MikeKingdom: This is a bit of awkward duplication and you basically do this work earlier but throw it away to add the id. Seems like this can be turned into a local component instead and used at line 158 and 163.

src/js/components/FormField/FormField.js
```diff
@@ -288,62 +304,133 @@ const FormField = forwardRef(
...
+    const errorId =
+      error && htmlFor ? `grommet-${adjustedHtmlFor}__error` : undefined;
+
+    // Combines aria-describedby linking and plain/focusIndicator/pad
+    // defaults into one Children.map pass, not two. The pass must always
+    // run the same way regardless of `error`, or the child's reconciliation
+    // key changes between renders - remounting it and losing state (e.g. a
+    // typed value) right when the error appears.
+    let contents;
+    if (children) {
```
MikeKingdom, line 393: This whole block of added code is hard to validate that it doesn't have a regression. Any way to simplify this whole area or break it up into easy to understand functions? It seems like alot of it mainly moved around so it's hard to analyze what bits are the new/different ones

Sulaymon333:
@MikeKingdom I have addressed both comment via this [commit](https://github.com/grommet/grommet/pull/7931/changes/ba1682121820d7f70dcd49ebbda210776524cae5)

src/js/components/FormField/FormField.js
```diff
@@ -297,53 +378,70 @@ const FormField = forwardRef(
...
-                  ? formFieldTheme?.checkBox?.pad
-                  : undefined,
-            });
-          }
-          return child;
-        })) ||
-      children;
+        const isInputComponent =
+          grommetInputNames.indexOf(child.type.displayName) !== -1;
+        if (!isInputComponent) return child;
```
MikeKingdom: Can probably simplify this to the following. It also helps minimize the multiple returns which can be a bug waiting to happen.
```jsx
        if (!child || !child.type ||
          grommetInputPadNames.indexOf(child.type.displayName) !== -1
        ) {
          return child;
        }
        
        if (themeBorder) {
          wantContentPad = true;
        }
```

src/js/components/FormField/FormField.js
```diff
@@ -193,6 +204,50 @@ const Input = ({ component, disabled, invalid, name, onChange, ...rest }) => {
...
+  )
+    return {};
+  return {
+    plain: true,
+    focusIndicator: !containerFocus,
+    pad:
+      'CheckBox'.indexOf(child.type.displayName) !== -1
+        ? formFieldTheme?.checkBox?.pad
+        : undefined,
+  };
```
MikeKingdom:
```suggestion
  return (
    !themeBorder ||
    child.props.plain !== undefined ||
    child.props.focusIndicator !== undefined
  )  ? {} :
  {
    plain: true,
    focusIndicator: !containerFocus,
    pad:
      'CheckBox'.indexOf(child.type.displayName) !== -1
        ? formFieldTheme?.checkBox?.pad
        : undefined,
  };
```

MikeKingdom (commented):
Looking better. I think it could still use a little simplification and I definitely still need to test it

src/js/components/FormField/FormField.js
Sulaymon333: Applying the code suggestion above produces a snapshot diff that breaks current implemnetation. 

So, I implemented code deduplication for instances when `child` is returned by getting rid of `isInputComponent ` and utilize `isGrommetInput()` instead, see this [commit](https://github.com/grommet/grommet/pull/7931/changes/372517e2d282e51cbb9340569f2374925d73a4ee)

```js
if (!child || !child.type || !isGrommetInput(child.type)) return child;
```

Sulaymon333:
Also cleaned up two nearby patterns discovered during audit (labelStyle's if/else and isFileInputComponent's empty-if-body side-effect trick) - both zero-behavior-change, verified against the full test suite with no snapshot diff.

commit - https://github.com/grommet/grommet/pull/7931/changes/5613127f50a189d7c62d7d14a5d363542d8dafc1

MikeKingdom (approved):

jcfilben (commented):
I'm not seeing the aria-described by coming through and the screenreader output doesn't read out the error message when the input is focused

Image
Image

jcfilben:
It looks like this is just an issue with the Form/Aggregate Validation story because the story doesn't use the error prop

Sulaymon333:
> It looks like this is just an issue with the Form/Aggregate Validation story because the story doesn't use the error prop

updated story to use `error` prop via this [commit](https://github.com/grommet/grommet/pull/7931/changes/0a9f057b87df936fb29b3bc8b9098af76f6844b5)

jcfilben (approved):

PR #7934, opened 2026-04-18
fix(DataTable): add aria-sort="none" and aria-hidden for unsorted columns

### LINKED ISSUE #7924: DataTable: sortable columns missing aria-sort="none" and unsorted icon should be aria-hidden

Yes

### Expected Behavior
Two related fixes needed for sortable columns that are not currently sorted:

The <th> element should have aria-sort="none" so AT users can discover the column is sortable before interacting with it. Currently only "ascending" and "descending" are set in Header.js the unsorted state is omitted entirely.

The neutral ↕ sort indicator icon shown on unsorted sortable columns should have aria-hidden="true". It is decorative in this context sortability will be communicated via aria-sort="none". Without aria-hidden, some AT may announce an unlabelled SVG since iconAriaLabel is undefined in this state.

### Actual Behavior

aria-sort is absent on sortable columns that are not currently sorted
The ↕ icon on unsorted sortable columns has no aria-hidden="true" and no label

n/a

### Steps to Reproduce

Render a DataTable with sortable columns. Inspect the <th> elements on unsorted columns — no aria-sort attribute is present and the sort icon SVG has no aria-hidden.

https://storybook.grommet.io/?path=/story/visualizations-datatable-sort--sort

### Your Environment

latest Grommet

mvanhorn:
Sortable columns that are not currently sorted were missing `aria-sort="none"`, so AT users couldn't discover that a column is sortable before interacting with it. The unsorted sort indicator icon was also exposed to assistive technology despite being decorative.

- Added `ariaSort = 'none'` in the unsorted-but-sortable branch of Header.js
- Added `aria-hidden={true}` on the sort indicator icon when no `aria-label` is set (unsorted state)

Follows the [WAI-ARIA sortable table pattern](https://www.w3.org/WAI/ARIA/apg/patterns/table/examples/sortable-table/).

Closes #7924

### DIFF
--- a/src/js/components/DataTable/Header.js
+++ b/src/js/components/DataTable/Header.js
@@ -381,6 +381,7 @@ const Header = forwardRef(
                     }
                   } else if (theme.dataTable.icons.sortable) {
                     Icon = theme.dataTable.icons.sortable;
+                    ariaSort = 'none';
                   }
                 }
 
@@ -404,7 +405,12 @@ const Header = forwardRef(
                       justify={align}
                     >
                       {content}
-                      {Icon && <Icon aria-label={iconAriaLabel} />}
+                      {Icon && (
+                        <Icon
+                          aria-label={iconAriaLabel}
+                          aria-hidden={!iconAriaLabel ? true : undefined}
+                        />
+                      )}
                     </Box>
                   </StyledHeaderCellButton>
                 );
--- a/src/js/components/DataTable/Header.js
+++ b/src/js/components/DataTable/Header.js
@@ -379,8 +379,10 @@ const Header = forwardRef(
                         messages,
                       });
                     }
-                  } else if (theme.dataTable.icons.sortable) {
-                    Icon = theme.dataTable.icons.sortable;
+                  } else {
+                    if (theme.dataTable.icons.sortable) {
+                      Icon = theme.dataTable.icons.sortable;
+                    }
                     ariaSort = 'none';
                   }
                 }
--- a/src/js/components/DataTable/Header.js
+++ b/src/js/components/DataTable/Header.js
@@ -316,6 +316,8 @@ const Header = forwardRef(
               units,
             }) => {
               let content;
+              const headerLabel =
+                typeof header === 'string' ? header : property;
               const unitsContent = units ? (
                 <Text {...textProps} {...theme.dataTable.header.units}>
                   {units}
@@ -390,6 +392,15 @@ const Header = forwardRef(
                 content = (
                   <StyledHeaderCellButton
                     plain
+                    a11yTitle={
+                      ariaSort === 'none'
+                        ? format({
+                            id: 'dataTable.sortable',
+                            messages,
+                            values: { label: headerLabel },
+                          })
+                        : undefined
+                    }
                     column={property}
                     fill="vertical"
                     focusIndicator={size ? 'inset' : undefined}
@@ -527,9 +538,7 @@ const Header = forwardRef(
                         onResize(prop, width);
                         updateWidths(prop, width);
                       }}
-                      headerText={
-                        typeof header === 'string' ? header : property
-                      }
+                      headerText={headerLabel}
                       messages={messages}
                       headerId={headerId}
                     />
--- a/src/js/components/DataTable/Header.js
+++ b/src/js/components/DataTable/Header.js
@@ -362,31 +362,27 @@ const Header = forwardRef(
               if (onSort && sortable !== false) {
                 let Icon;
                 let iconAriaLabel;
-                if (onSort && sortable !== false) {
-                  if (sort && sort.property === property) {
-                    Icon =
-                      theme.dataTable.icons[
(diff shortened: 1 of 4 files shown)


src/js/components/DataTable/Header.js
```diff
@@ -381,6 +381,7 @@ const Header = forwardRef(
                     }
                   } else if (theme.dataTable.icons.sortable) {
                     Icon = theme.dataTable.icons.sortable;
+                    ariaSort = 'none';
```
britt6612: @mvanhorn Thank you for jumping on this!

I think the `ariaSort` should be added outside the icon icon

```
} else {
  if (theme.dataTable.icons.sortable) {
    Icon = theme.dataTable.icons.sortable;
  }
  ariaSort = 'none'; // move here
}
```

mvanhorn:
Thanks @britt6612 - applied in 13c09e5. `ariaSort = 'none'` is now set outside the `theme.dataTable.icons.sortable` guard, so every sortable-but-unsorted column gets `aria-sort="none"` even when a theme omits the sortable icon. The icon assignment stays guarded as before.

britt6612:
@mvanhorn looks like the jest tests are failing you may have to run `yarn jest -u` also can we add a test for this behavior something along the lines of test `sortable columns have aria-sort="none" when unsorted`

mvanhorn:
Updated in 95c9788: ran `yarn jest -u` to regenerate the DataTable snapshots for the new `aria-sort` attribute, and added a `sortable columns have aria-sort="none" when unsorted` test that asserts both columns start with `aria-sort="none"` and the clicked column transitions to `"ascending"` while the other stays at `"none"`. Also signed off the two prior commits via rebase to clear the DCO check.

Verified locally: `yarn jest --testPathPatterns=DataTable-test` → 87 passed, 104 snapshots passed.

britt6612 (approved):

Sulaymon333 (commented):
Thank you @mvanhorn for working on this!

I tested `aria-sort="none"` with VoiceOver on Mac and it does not provide any signal that a column is sortable. According to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-sort#none), "none" is the default value - meaning the column has no active sort. VoiceOver does not announce it, so users still can not discover a column is sortable without interacting with it. NVDA/JAWS may behave differently, but this leaves inconsistent behavior across platforms - which doesn't fully address the original issue.

Additionally, `aria-hidden` on the neutral sort icon is a good cleanup, but worth noting that the neutral sort icon (`theme.dataTable.icons.sortable`) is undefined in the base grommet theme today (it is [commented out in base.js](https://github.com/grommet/grommet/blob/master/src/js/themes/base.js#L1136)), so in practice there's no icon to hide for most users unless their theme explicitly sets one.

I did a quick research on an approach that more directly addresses the core problem in the [ticket](https://github.com/grommet/grommet/issues/7924) - making sortability discoverable to SR/AT users regardless of platform. The idea is to add an `aria-label` on the sort button itself with context-aware messaging:

- Unsorted column: the button announces "Sort by {visible column name}" - immediately telling the user the column is sortable and what action it performs
- Sorted column: the button announces "{visible column name}, ascending" or "{visible column name}, descending" - communicating both the column name and its current sort state

This also should includes an aria-live="assertive" region inside the button so that when a user activates sorting (via keyboard or click), the new sort state is announced immediately without requiring the user to navigate away and back.

The labels should use grommet's existing `MessageContext` / `format()` system with new i18n keys (sortBy, sortedAscending, sortedDescending), so they are fully customizable via the DataTable `messages` prop.

This approach should work reliably across VoiceOver, NVDA, and JAWS because it does not rely on `aria-sort` attribute interpretation - which varies by screen reader technologies - and instead provides explicit, actionable labels that all SR/AT can read. 

Happy to share a possible code implementation approach if needed.


**Sortable Table Example from APG `aria-sort` value definition** 

image


cc: @britt6612 @jcfilben @halocline

mvanhorn:
Thanks for testing with VoiceOver, that's a real gap. `aria-sort="none"` is deliberately not announced by most screen readers (it's the ARIA default state), so on its own it exposes the state programmatically but doesn't make sortability discoverable in VoiceOver.

Addressed in d0f3b27: unsorted sortable header buttons now carry an accessible name of "{label}, sortable" via a new `dataTable.sortable` message (so it's localizable and overridable through the existing `messages` prop). VoiceOver now announces e.g. "A, sortable, button" before any interaction, and once a column is sorted the announcement comes from `aria-sort` as before. Full test suite passes (1758 tests), with new assertions covering the accessible names.

mvanhorn:
Thanks! Moved the aria-sort assignment out of the icon block — it's now set to `'none'` on any sortable columnheader and overridden to `ascending`/`descending` when active, independent of whether the sort icon renders. Added a test.

britt6612:
closing in favor of https://github.com/grommet/grommet/pull/7988

Issue #7936, opened 2026-04-23
NotFoundError on Button icon change while translating if cssGap is "true"


AntoineDuComptoirDesPharmacies:
### Is this something you or your team is interested in contributing a pull request towards?

Yes

### Expected Behavior

The button should change icon without crash.

### Actual Behavior

When the icon of the button change, the following crash occurs : 
```
Uncaught NotFoundError: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.
    at insertBefore (react-dom.development.js:11088:18)
    at insertOrAppendPlacementNode (react-dom.development.js:23900:7)
    at insertOrAppendPlacementNode (react-dom.development.js:23908:7)
    at insertOrAppendPlacementNode (react-dom.development.js:23908:7)
    at insertOrAppendPlacementNode (react-dom.development.js:23908:7)
    at commitPlacement (react-dom.development.js:23844:9)
    at commitReconciliationEffects (react-dom.development.js:24591:7)
    at commitMutationEffectsOnFiber (react-dom.development.js:24294:9)
    at recursivelyTraverseMutationEffects (react-dom.development.js:24273:7)
    at commitMutationEffectsOnFiber (react-dom.development.js:24576:9)
    at recursivelyTraverseMutationEffects (react-dom.development.js:24273:7)
    at commitMutationEffectsOnFiber (react-dom.development.js:24346:9)
    at recursivelyTraverseMutationEffects (react-dom.development.js:24273:7)
    at commitMutationEffectsOnFiber (react-dom.development.js:24293:9)
    at recursivelyTraverseMutationEffects (react-dom.development.js:24273:7)
    at commitMutationEffectsOnFiber (react-dom.development.js:24293:9)
    at recursivelyTraverseMutationEffects (react-dom.development.js:24273:7)
    at commitMutationEffectsOnFiber (react-dom.development.js:24346:9)
    at recursivelyTraverseMutationEffects (react-dom.development.js:24273:7)
    at commitMutationEffectsOnFiber (react-dom.development.js:24293:9)
    at recursivelyTraverseMutationEffects (react-dom.development.js:24273:7)
    at commitMutationEffectsOnFiber (react-dom.development.js:24293:9)
    at recursivelyTraverseMutationEffects (react-dom.development.js:24273:7)
    at commitMutationEffectsOnFiber (react-dom.development.js:24576:9)
    at recursivelyTraverseMutationEffects (react-dom.development.js:24273:7)
[...]

PR #7937, opened 2026-04-23
 #7936 Fix NotFoundError on insertBefore when page is translated. 

AntoineDuComptoirDesPharmacies:
#### What does this PR do?

Fixes a runtime crash that occurs in Anchor, Button and Tab when the page is translated by the browser (most commonly Chrome's Google Translate). When a component renders a plain string as a direct sibling of other React-managed elements (e.g. {icon}{label} inside a Box), the translator replaces the raw text node with a new node. React still holds a reference to the original text node and later calls insertBefore against it, throwing NotFoundError: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.

Impacted root components :
- Anchor
- Button
- Tab

The fix wraps the plain string label (and title for Tab) in a <span>. Translators replace the text node inside the span, but the span itself remains stable, so React's child references stay valid.

- src/js/components/Button/Button.js — original reported bug.
- src/js/components/Anchor/Anchor.js  — same pattern (string label/title rendered as a sibling of an icon).
- src/js/components/Tab/Tab.js — same pattern (string label/title rendered as a sibling of an icon).

Other components were audited (Menu, Notification, AccordionPanel, RadioButton, CheckBox, NameValuePair, Card, FormField, Heading, Text, Avatar) and either already wrap their string children in protective elements (e.g. <Text>, <Heading>, <Message>) or render the string as the only child, so they are not affected.

- Full Jest suite passes (1756 tests, 99 suites, 1968 snapshots).
- Snapshots that include the affected components were regenerated to reflect the added <span> wrapper.
- Two existing tests were updated to use accessibility-first queries (getByRole('button', { name: ... })) instead of getByText('label'), which previously returned the <button> element directly but now returns the inner <span>. The user-facing behavior (focus on click/tab, cursor inheritance) is unchanged

See issue here and associated code sandbox : https://github.com/grommet/grommet/issues/7936

- [ X ] `screen` is used for querying.
- [ X ] The correct query is used. (Refer to [this list of queries](https://testing-library.com/docs/queries/about/#priority))
- [ X ] `asFragment()` is used for snapshot testing.

This is a well-known React + Google Translate interaction issue. When a text node is a direct sibling of React-managed elements, the translator's DOM mutations invalidate React's internal references. 
The standard, community-recommended workaround is to wrap text nodes in an inert element (typically a  <span>), which acts as a stable anchor while the translator only swaps out the inner text.

Components like Menu, AccordionPanel, Notification, RadioButton, CheckBox, etc. were already wrapping their string content in <Text>, <Heading>, <Message> or similar elements, so they were already protected.

https://github.com/grommet/grommet/issues/7936

N/A

No

Yes — bug fix affecting users whose pages may be translated by the browser (e.g. multilingual audiences using Chrome's Google Translate).

Backwards compatible.
The only observable change is an extra inline <span> element wrapping the label text in the rendered DOM. This may affect:
  - Snapshot tests in downstream consumers (the wrapper element appears in serialized output).
  - Test queries that rely on getByText('label') to retrieve the parent button/anchor element — these should be migrated to getByRole('button', { name: 'label' }) (the Testing Library recommended approach).

No public API or prop signature changes.

### DIFF
--- a/src/js/components/Anchor/Anchor.js
+++ b/src/js/components/Anchor/Anchor.js
@@ -80,8 +80,14 @@ const Anchor = forwardRef(
 
     const anchorIcon = useSizedIcon(coloredIcon, sizeProp || size, theme);
 
-    const first = reverse ? label : anchorIcon;
-    const second = reverse ? anchorIcon : label;
+    // Wrap a plain string label in a span so browser translation tools
+    // (e.g. Chrome's Google Translate) replace the text node inside the span
+    // rather than a text node that is a direct sibling of React-managed
+    // elements, which would break React's DOM reconciliation
+    // (NotFoundError on insertBefore).
+    const safeLabel = typeof label === 'string' ? <span>{label}</span> : label;
+    const first = reverse ? safeLabel : anchorIcon;
+    const second = reverse ? anchorIcon : safeLabel;
 
     return (
       <StyledAnchor
--- a/src/js/components/Anchor/__tests__/__snapshots__/Anchor-test.tsx.snap
+++ b/src/js/components/Anchor/__tests__/__snapshots__/Anchor-test.tsx.snap
@@ -33,7 +33,9 @@ exports[`Anchor blur renders 1`] = `
     class="c1"
     href="#"
   >
-    Test
+    <span>
+      Test
+    </span>
   </a>
 </div>
 `;
@@ -125,7 +127,9 @@ exports[`Anchor focus renders 1`] = `
     class="c1"
     href="#"
   >
-    Test
+    <span>
+      Test
+    </span>
   </a>
 </div>
 `;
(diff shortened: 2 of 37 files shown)


MikeKingdom:
This fix seems too invasive and will cause a lot of issues with existing code and tests. This also seems like a significant issue with the browser's translate ability.

AntoineDuComptoirDesPharmacies:
> This fix seems too invasive and will cause a lot of issues with existing code and tests. This also seems like a significant issue with the browser's translate ability.

You are right, it is a bit invasive but it seems to be the only way to avoid Google Translation crashing the website.
If required, this project can add this specific Eslint rules to avoid patterns where the code can crash.

https://www.npmjs.com/package/eslint-plugin-react-google-translate

PR #7940, opened 2026-04-29
storybook upgrade phase 2

jcfilben:
#### What does this PR do?
Create code snippet for storybook examples and fixed timing issue in Form stories

### DIFF
--- a/src/js/components/Form/stories/CustomThemed/FieldCustomStates.stories.js
+++ b/src/js/components/Form/stories/CustomThemed/FieldCustomStates.stories.js
@@ -63,7 +63,11 @@ export const FieldCustomStates = () => {
   const inputRef = useRef();
 
   useEffect(() => {
-    inputRef.current.focus();
+    const focusTimeout = setTimeout(() => {
+      inputRef.current?.focus();
+    });
+
+    return () => clearTimeout(focusTimeout);
   }, []);
 
   return (
--- a/storybook/storySourceLoader.js
+++ b/storybook/storySourceLoader.js
@@ -1,0 +1,144 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+const { transformSync } = require('@babel/core');
+
+const PLUGIN_NAME = 'grommet-story-source-injector';
+const SOURCE_VAR = '__STORYBOOK_SOURCE_CODE__';
+const MERGE_VAR = '__withStorySource__';
+
+const getRuntimeExportNames = (declaration, t) => {
+  if (!declaration) return [];
+
+  if (
+    t.isFunctionDeclaration(declaration) ||
+    t.isClassDeclaration(declaration)
+  ) {
+    return declaration.id && declaration.id.name ? [declaration.id.name] : [];
+  }
+
+  if (t.isVariableDeclaration(declaration)) {
+    return declaration.declarations
+      .map((entry) => (entry.id && entry.id.name ? entry.id.name : null))
+      .filter(Boolean);
+  }
+
+  return [];
+};
+
+const storySourceInjectionPlugin = ({ types: t, template }) => ({
+  name: PLUGIN_NAME,
+  visitor: {
+    Program(programPath, state) {
+      const storyExports = [];
+
+      programPath.get('body').forEach((path) => {
+        if (!path.isExportNamedDeclaration()) return;
+
+        const { declaration } = path.node;
+
+        if (declaration) {
+          storyExports.push(...getRuntimeExportNames(declaration, t));
+          return;
+        }
+
+        if (path.node.specifiers && path.node.source == null) {
+          path.node.specifiers.forEach((specifier) => {
+            if (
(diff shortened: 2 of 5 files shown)


src/js/components/Form/stories/CustomThemed/FieldCustomStates.stories.js
```diff
@@ -63,7 +63,11 @@ export const FieldCustomStates = () => {
   const inputRef = useRef();
 
   useEffect(() => {
-    inputRef.current.focus();
+    const focusTimeout = setTimeout(() => {
```
jcfilben, line 66: In Storybook 10 the render timing is different, this effect could run before the TextInput ref had been attached, so reading current.focus threw when current was undefined. Deferring the call with a timeout preserves the intended focused visual state without relying on the ref being ready synchronously.

britt6612: good to know I wonder going forward how we can remember this, hard to say to put a comment in the story since users copy/paste but open to ideas

jcfilben: I think from what I read from copilot we shouldn't have been doing it this way even on previous versions of storybook. So maybe less related to storybook specifically and just changing the code to be a better practice

britt6612: ah okay gotcha then yeah leave as is

storybook/storySourceLoader.js
```diff
@@ -0,0 +1,144 @@
...
+  if (t.isVariableDeclaration(declaration)) {
+    return declaration.declarations
+      .map((entry) => (entry.id && entry.id.name ? entry.id.name : null))
+      .filter(Boolean);
+  }
+
+  return [];
+};
+
+const storySourceInjectionPlugin = ({ types: t, template }) => ({
```
britt6612, line 27: I dont think we would expect any stories to be written like
`export default ...`
if someone has a story or writes one in that format it would be skipped

src/js/components/Form/stories/CustomThemed/FieldCustomStates.stories.js
MikeKingdom: This definitely needs some kind of comment. useEffect runs after DOM has committed and painted so unless the component is conditionally rendered or the ref isn't correctly forwarded to the child component, it should definitely be defined in the useEffect.

Putting it in a timeout with no time value simply makes it run immediately after the thread running the useEffect. So what managed to happen in the rest of that path to make it defined?

storybook/storySourceLoader.js
jcfilben: I think having `export default` is a valid story format even though we aren't using that pattern in our grommet storybook stories, for example on the storybook site they show this:

Screenshot 2026-04-30 at 3 27 27 PM

src/js/components/Form/stories/CustomThemed/FieldCustomStates.stories.js
jcfilben: Thanks for the explanation, I'll look into this more

jcfilben: Here is the explaination copilot gave. I've also added a comment in the code for both stories.

> a timeout with no delay is not instant; it is deferred to a later task. That deferral is the key.
> 
> What changed between the failing version and the fixed version is timing:
> 
> - The story renders.
> - React commits that render.
> - Your effect runs and tries to call focus.
> - In this specific path, the ref target from TextInput is sometimes not ready yet.
> - With setTimeout(..., 0), the focus call is moved to the next macrotask, after React and child component side effects from the current commit have finished.
> - By then, inputRef.current is usually populated, so focus works.
> 
> Why this can happen even though useEffect is after commit:
> 
> - Commit means DOM updates are applied, but not every library component’s internal effect chain is guaranteed to be fully settled at the exact moment your parent effect runs.
> - Here, FormField + TextInput + Storybook 10 render timing can produce a brief window where the forwarded ref is still undefined.
> - Zero-delay timeout gives one more turn of the event loop for that chain to complete.

storybook/storySourceLoader.js
britt6612: so would that still work with this source loader?

jcfilben: Yes, it would show the code for the whole file

src/js/components/Form/stories/CustomThemed/FieldCustomStates.stories.js
MikeKingdom: Have to assume some useEffect in the children is delaying the DOM but if it works it works :)

britt6612 (approved):

MikeKingdom (approved):
LGTM.

PR #7950, opened 2026-05-12
Clean up Select and SelectMultiple children stories

MikeKingdom:
[Deploy Preview](https://5d9774839a6eff00203f5cbf-iijjefwoff.chromatic.com/?path=/story/input-selectmultiple-value-label--value-label&globals=theme:hpe)

#### What does this PR do?
This cleans up the `children` and `valueLabel` related stories for Select and SelectMultiple. Both try to show how to use `valueLabel` to render the selected items as tags but with custom versions (pre-Tag existing).

This also adds a separate SelectMultiple->valueLabel story that serves as a better example of doing tags (without the custom children in the drop.)

Finally this corrects a couple of the children render functions which we expecting the wrong arguments and incorrectly interpreting index as state.

Children.stories.tsx

Storybook.

Storybook

No

No
Backwards compatible

### DIFF
--- a/src/js/components/Select/stories/ChildrenRender.stories.js
+++ b/src/js/components/Select/stories/ChildrenRender.stories.js
@@ -1,7 +1,6 @@
 import React, { useState } from 'react';
 
-import { FormClose } from 'grommet-icons';
-import { Box, Button, Select, Text } from 'grommet';
+import { Box, Select, Tag, Text } from 'grommet';
 
 const allSeasons = [
   'S01',
@@ -23,37 +22,16 @@ export const Children = () => {
     setSelected(selected.filter((selectedSeason) => selectedSeason !== season));
   };
 
-  const renderSeason = (season) => (
-    <Button
-      key={`season_tag_${season}`}
-      href="#"
-      onClick={(event) => {
-        event.preventDefault();
-        event.stopPropagation();
-        onRemoveSeason(season);
-      }}
-      onFocus={(event) => event.stopPropagation()}
+  const renderOption = (option, index, options, state) => (
+    <Box
+      pad="small"
+      background={
+        state.selected ? 'brand' : undefined
+      }
     >
-      <Box
-        align="center"
-        direction="row"
-        gap="xsmall"
-        pad={{ vertical: 'xsmall', horizontal: 'small' }}
-        margin="xsmall"
-        background="brand"
-        round="large"
-      >
-        <Text size="small">{season}</Text>
-        <Box round="full" margin={{ left: 'xsmall' }}>
-          <FormClose size="small" style={{ width: '12px', height: '12px' }} />
-        </Box>
-      </Box>
-    </Button>
-  );
-
-  const renderOption = (option, state) => (
-    <Box pad="small" background={state.active ? 'active' : undefined}>
-      {option}
+      <Text color={state.disabled ? 'text-disabled' : 'text-strong'}>
+        {option}
+      </Text>
     </Box>
   );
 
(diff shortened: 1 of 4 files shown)


src/js/components/Select/stories/ChildrenRender.stories.js
```diff
@@ -64,20 +42,33 @@ export const Children = () => {
...
+            gap="xsmall"
+            cssGap
+            width="small"
+          >
+            {options && options.length ?
+              options.map((season) => (
+                <Tag
+                  key={`season_tag_${season}`}
+                  value={season}
+                  onRemove={(event) => {
```
jcfilben, line 60: If you are navigating via keyboard and a tag is removed focus is also removed and you have to tag all the way back down through the page. I'm wondering if we could do something similar to SelectMultiple showSelectedInline where when an option is removed the focus is placed on the next available option https://storybook.grommet.io/?path=/story/input-selectmultiple-showselectedinline--show-selected-inline

jcfilben (commented):
Image

Should the tags wrap instead of just being stacked vertically?

src/js/components/Select/stories/ChildrenRender.stories.js
MikeKingdom: Cool idea

MikeKingdom:
> Should the tags wrap instead of just being stacked vertically?

They do for me. The box is `row` and `wrap`. Note quite sure why it didn't seem to have space for you?
image

britt6612:
> Should the tags wrap instead of just being stacked vertically?


looks like this is the grommet theme from your screen shot @jcfilben its probably the `size` not coming through correctly since the sizes are different


@jcfilben @MikeKingdom 

for hpe theme looks how we want it.. 
Screenshot 2026-05-14 at 10 19 37 AM

britt6612 (approved):

MikeKingdom:
> > Should the tags wrap instead of just being stacked vertically?
> 
> looks like this is the grommet theme from your screen shot @jcfilben its probably the `size` not coming through correctly since the sizes are different

@britt6612 You're correct it looks like grommet theme, although it wraps at default zoom etc in all themes for me. I'm a little curious why it didn't for @jcfilben, especially since it looks like it had room for another tag on the same row. I'd have to be able to inspect the DOM to learn more.

britt6612:
Screenshot 2026-05-14 at 2 34 22 PM


@MikeKingdom the grommet theme the tags still wrap for you?

MikeKingdom:
> @MikeKingdom the grommet theme the tags still wrap for you?

They do, but I'm guessing it's a font thing between windows and mac.

Either way I changed it so the SelectMultiple is sized and the value label box just fills to that size. Hopefully it looks better for you now.

Note I also tweaked Tag typescript types since it wasn't allowing standard things like id etc.

src/js/components/Select/stories/ChildrenRender.stories.js
MikeKingdom: Done (just for the valueLabel example...seems a bit overwhelming for the intent of the other stories.)

src/js/components/SelectMultiple/stories/typescript/ValueLabel.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+      // if there is one, otherwise the previous one
+      const nextFocusIndex = index < nextSelected.length ? index : index - 1;
+      if (nextFocusIndex >= 0 && valuesRef.current) {
+        setTimeout(() => {
+          const element = valuesRef.current as HTMLElement;
+          const buttons = element?.querySelectorAll('button');
+          buttons[nextFocusIndex].focus();
+        }, 200);
+      }
+    }
```
jcfilben, line 36: Could we add this logic to the SelectMultiple children story too?

MikeKingdom: In my above comment I said I'd prefer not to just to keep the stories simpler since it's mainly showing children.

jcfilben: Oops sorry, missed that comment

jcfilben (approved):

PR #7952, opened 2026-05-13
Add support for children in Tag

### LINKED ISSUE #7951: Tag - Add support for children
Enhance the tag component so that children can be passed to the tag. For example:

```
<Tag>
  <>
    <Icon />
    <Text>Test</Text>
  </>
</Tag>
```

jcfilben:
#### What does this PR do?
Adds children prop to tag component to allow for customization of the tag

Also added node type to name and value

Closes https://github.com/grommet/grommet/issues/7951

Yes
Yes
Backwards compatible

### DIFF
--- a/src/js/components/Tag/index.d.ts
+++ b/src/js/components/Tag/index.d.ts
@@ -3,7 +3,7 @@ import { BoxProps } from '../Box';
 
 export interface TagProps {
   children?: React.ReactNode;
-  name?: string;
+  name?: string | React.ReactNode;
   onClick?: (...args: any[]) => any;
   onRemove?: (...args: any[]) => any;
   messages?: {
@@ -20,7 +20,7 @@ export interface TagProps {
     | 'xlarge'
     | 'xxlarge'
     | string;
-  value: string | number;
+  value: string | number | React.ReactNode;
 }
 
 export interface TagExtendedProps extends BoxProps, TagProps {}
--- a/src/js/components/Tag/index.d.ts
+++ b/src/js/components/Tag/index.d.ts
@@ -20,7 +20,7 @@ export interface TagProps {
     | 'xlarge'
     | 'xxlarge'
     | string;
-  value: string | number | React.ReactNode;
+  value?: string | number | React.ReactNode;
 }
 
 export interface TagExtendedProps extends BoxProps, TagProps {}
--- a/src/js/components/Tag/index.d.ts
+++ b/src/js/components/Tag/index.d.ts
@@ -2,6 +2,7 @@ import * as React from 'react';
 import { BoxProps } from '../Box';
 
 export interface TagProps {
+  children?: React.ReactNode;
   name?: string;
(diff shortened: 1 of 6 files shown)


MikeKingdom:
I like the children ability for flexibility. But I feel like we should support a react node for `name` and `value` too. It actually works today and would probably just be a doc and ts and prop types tweak. It's more convenient sometimes to just add an icon and some text in either place, especially in the example use case that still supports an icon and text in one of those but just text in the other.

With children you'd have to re-implement the `name : value` styling.

jcfilben:
> I like the children ability for flexibility. But I feel like we should support a react node for name and value too. It actually works today and would probably just be a doc and ts and prop types tweak. It's more convenient sometimes to just add an icon and some text in either place, especially in the example use case that still supports an icon and text in one of those but just text in the other.
> 
> With children you'd have to re-implement the name : value styling.

Yeah that seems reasonable. I would be supportive of that

britt6612 (approved):
Looks good

src/js/components/Tag/propTypes.js
```diff
@@ -3,8 +3,9 @@ import PropTypes from 'prop-types';
 let PropType = {};
 if (process.env.NODE_ENV !== 'production') {
   PropType = {
-    name: PropTypes.string,
-    value: PropTypes.string.isRequired,
+    children: PropTypes.node,
+    name: PropTypes.oneOfType([PropTypes.node, PropTypes.string]),
+    value: PropTypes.oneOfType([PropTypes.node, PropTypes.string]).isRequired,
```
MikeKingdom: just realized this probably shouldn't be required anymore. If you're doing children you're not going to pass this. We originally required it given the styling difference between name and value. And I always thought it was annoying since sometimes I just the name :)

src/js/components/Tag/index.d.ts
```diff
@@ -19,7 +20,7 @@ export interface TagProps {
     | 'xlarge'
     | 'xxlarge'
     | string;
-  value: string | number;
+  value: string | number | React.ReactNode;
```
MikeKingdom:
```suggestion
  value?: string | number | React.ReactNode;
```

MikeKingdom (commented):
Should Tag children also have a storybook story? Would be nice for testing/example.

src/js/components/Tag/propTypes.js
jcfilben: Good catch! I'll update this

jcfilben: Done

src/js/components/Tag/index.d.ts
jcfilben: Done

src/js/components/Tag/propTypes.js
britt6612: good catch ☝️

MikeKingdom (approved):

PR #7954, opened 2026-05-19
Fix Drop responsive placement when neither side fits

simleb:
#### What does this PR do?

Fixes `Drop` responsive placement when neither the preferred side nor the opposite side has room for the full content.

When `responsive` is true and the Drop's preferred side does not have room, `Drop` already flips to the opposite side **if** the full content fits there. When neither side fits, however, `Drop` kept the originally requested direction and clamped `maxHeight` to whatever space remained — which is the worst choice when the requested side is the smaller one (e.g. a `Select` opened near the bottom of the viewport with too many options to fit either above or below: the menu opens below into a tiny strip and scrolls).

This adds four new responsive branches in `DropContainer.place()` that fire after the existing "fully fits the other side" flips, picking whichever side has more available space and clamping `maxHeight` to it. The change is purely additive: every existing flip and clamp branch keeps its current behavior. `responsive={false}` is unaffected.

`src/js/components/Drop/DropContainer.js` — the `place()` function. The four new `else if` branches are inserted between the existing "fully fits other side" flip block and the fallback clamp block, so the diff reads as a clean addition. Each new branch mirrors the `top`/`bottom`/`maxHeight` formulas of its corresponding existing flip branch.

- Added three tests in `Drop-test.tsx` exercising the new branches: flip above when more room above, flip below when more room below, and a no-flip case where the requested side is already the larger one. They mock `getBoundingClientRect` on the trigger + drop container and `window.innerHeight` to set up the "neither side fully fits" geometry, then fire a `resize` event and assert on `dropEl.style.{top,bottom,maxHeight}`.
- One Menu snapshot (`with dropAlign bottom renders`) and three Drop snapshots (`align left left`, two `bottom: bottom` margin variants) were updated. In jsdom these tests run with all-zero rects; under the old behavior they clamped to `max-height: 0` pinned to the viewport edge — an invisible drop. The new behavior correctly flips to the side with more space; the snapshot updates record the fix rather than mask it.
- Full suite: 99 suites / 1771 tests pass locally.

1. Open Storybook and navigate to any Drop-backed component story (e.g. **Select**, **Menu**, **SelectMultiple**) with a long list of options.
2. Resize the browser window so the trigger sits where neither side has enough room for the full menu — typically a short window with the trigger near the middle, or just near either edge.
3. Open the menu.

Before this change: the menu opens in the originally requested direction and clamps to the (possibly very small) remaining space, producing a scrollable sliver. After: the menu opens on whichever side has more space.

- [ ] `asFragment()` is used for snapshot testing. — N/A, the new tests assert on inline style values directly (`dropEl.style.top`, `.bottom`, `.maxHeight`) since the behavior under test is the computed position.

Originally surfaced from a downstream report: a `Select` near the bottom of the viewport with ~40px of room below and ~500px of room above. Pre-fix, the menu opened into the 40px strip and clamped to a 1-option scrollable area. Post-fix, the same menu opens upward into the 500px above.

This affects every `Drop` consumer with a menu that can be taller than both available sides — `Select`, `Menu`, `SelectMultiple`, custom dropdowns. The trigger's vertical position determines which side has more space, so this is most visible on long forms and near the viewport edges.

None linked yet — happy to file one if the team prefers.

##### Before
before

##### After
after

No. `responsive=true` already documents itself as smart placement; this fills in a previously-missing case rather than introducing new behavior to document. `propTypes.js` is unchanged.

Yes — as a bug fix in `Drop` (and consequently `Select`, `Menu`, `SelectMultiple`, etc.) for the "menu near a viewport edge with a long option list" case.

Backwards compatible. The new logic only intercepts cases that previously fell through to the clamp branch with a too-small `maxHeight`; existing flip branches still match and fire first. Drops with `responsive={false}` are unaffected. No public API additions or changes.

### DIFF
--- a/src/js/components/Drop/DropContainer.js
+++ b/src/js/components/Drop/DropContainer.js
@@ -267,6 +267,9 @@ const DropContainer = forwardRef(
             align.top === 'bottom' &&
             // drop is overflowing below window
             targetRect.bottom + containerRect.height >= windowHeight &&
+            // the flip anchor (target's top) is on-screen — otherwise
+            // flipping pins the drop's bottom off-screen
+            targetRect.top > 0 &&
             // there is more room above the target than below it — neither
             // side fits the full drop, so prefer the larger side rather
             // than clamping to the smaller requested side.
@@ -281,6 +284,11 @@ const DropContainer = forwardRef(
             align.top === 'top' &&
             // drop is overflowing below window
             targetRect.top + containerRect.height >= windowHeight &&
+            // height of the drop is larger than the target — otherwise the
+            // drop already fits inside the target's vertical span and a
+            // flip would pin its bottom to a potentially off-screen
+            // target.bottom (mirrors the safety in branch #2 above).
+            targetRect.top + containerRect.height > targetRect.bottom &&
             // there is more room above the target than below it
             targetRect.bottom > windowHeight - targetRect.top
           ) {
@@ -293,6 +301,8 @@ const DropContainer = forwardRef(
             align.bottom === 'top' &&
             // drop is overflowing above window
             targetRect.top - containerRect.height <= 0 &&
+            // the flip anchor (target's bottom) is on-screen
+            targetRect.bottom < windowHeight &&
             // there is more room below the target than above it
             windowHeight - targetRect.bottom > targetRect.top
           ) {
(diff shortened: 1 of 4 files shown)


ShimiSun:
Looking good, thanks @simleb !! @taysea can we please take a look at this PR and prioritize it for next release 🙏

britt6612:
> Looking good, thanks @simleb !! @taysea can we please take a look at this PR and prioritize it for next release 🙏

I added both myself and @jcfilben to review so this is on our radar!

jcfilben (commented):
Nice enhancement! I noticed this issue (bit of an edge case) when the drop is aligned top: top, and the target is bigger than the drop, If I scroll so that part of the drop is covered on the bottom, it tries to flip the alignment of the drop to be bottom: bottom, which makes it so that I can't see the drop at all

https://github.com/user-attachments/assets/5e424cc1-22c8-4f1b-8486-0053bc14e3de

simleb:
@jcfilben good catch, all fixed!

britt6612 (approved):
Looks good thank you @simleb

jcfilben (approved):

PR #7958, opened 2026-05-26
fix file input cancel clears selection

britt6612:
#### What does this PR do?
Fixes a bug where cancelling out of the browser's file picker dialog would clear a previously selected file. When the user opens the file picker and cancels without selecting a new file, the browser fires a `change` event with an empty FileList. The `onChange` handler was processing this and wiping the existing file selection. This PR adds a guard to bail out early in that case.
fileinput.js
locally and added a test
1. Render a FileInput
2. Select a file  confirm filename appears
3. Click Browse again and cancel out of the dialog
4. Confirm the previously selected file is still shown

When the file picker is cancelled, browsers fire a `change` event with `event.target.files` as an empty FileList (length 0). This is browser-native behavior that cannot be suppressed, so the fix handles it in the handler itself.
n/a
n/a
n/a
sure
backwards compatible

### DIFF
--- a/src/js/components/FileInput/FileInput.js
+++ b/src/js/components/FileInput/FileInput.js
@@ -328,6 +328,9 @@ const FileInput = forwardRef(
             onChange={(event) => {
               event.persist();
               const fileList = event.target.files;
+              // If the user cancelled the file picker, fileList will be empty
+              // and we should not clear any previously selected files.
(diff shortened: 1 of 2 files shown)


halocline (approved):

jcfilben (approved):
Looks good!

PR #7959, opened 2026-05-27
Cards onOrder and sizeKey props

MikeKingdom:
[Deploy Preview](https://deploy-preview-7959--sad-tereshkova-173d07.netlify.app/?path=/story/visualizations-cards-onorder--on-order)

#### What does this PR do?

This PR adds support for reordering and sizing of items rendered by the Cards component.

Whenever the an `onOrder` callback is passed to the `Cards` component, items rendered for the `data` can be dragged to a new position in the grid that Cards uses to layout components. Since items are rendered in the grid across columns and dropping into a new row as needed. The onOrder callback is passed the data in the new order given where the item was dragged to. If `item A` is dragged over  `item B`, `item A` will now be where `item B` was and `item B` and any items after it in `data` will now be after `item A`. The `onOrder` callback is responsible for altering the `data` passed to the parent `Data` component (ie. a controlled value).

Whenever the `onOrder` callback is defined, the items will be focusable and the keyboard arrow keys can be used to move the focused item one position earlier or later in `data`. _Note: making the up and down arrow keys move the item to a previous or next row was considered but becomes rather problematic when items can span multiple rows or columns so I stuck with the simpler option for now._  

Whenever the `sizeKey` property is passed to `Cards`, any data items can specify a member with this key which is expected to be an object with `rows` and/or `columns` specified. These values will make the item span that number of rows or columns in the grid. For example if the `sizeKey` is `size` and a data item has
```javascript
{
  ...
  size: { columns: 2, rows: 2 }
} 
```
The resulting item will span 2 columns and 2 rows of space in the grid. This allows for a basic masonry type of layout. Note that if there is now additional row after this item it simply will only span 1 row and similarly if there is only a single column.

Cards.js

Storybook and manual testing

Storybook Visualizations/Cards/onOrder and Sizable stories.

See [Customizable dashboard: Cards component enhancements and reference implementation](https://github.com/grommet/hpe-design-system/issues/6201)

Yes

Yes

Backwards compatible


src/js/components/Cards/stories/onOrder.stories.js
jcfilben, line 1: The onOrder and Sizable stories both have accessibility issues
Image

jcfilben, line 1: I would expect to see the styled focus indicator defined in the theme instead of the browser default
Image

jcfilben, line 1: When I navigate to the cards my screenreader doesn't give me any indication that the cards are re-orderable or what keys I should press to change their order.

MikeKingdom: Any thoughts on what we'd want for the screenreader?

jcfilben: Hmm not sure off the top of my head if there is a standard pattern to follow here or what a screenreader user would expect. I wonder if APG has any patterns for this?

src/js/components/Cards/Cards.js
```diff
@@ -48,6 +63,123 @@ const Cards = React.forwardRef(
...
+          <CardBody>
+            {(typeof item === 'string' && item) ??
+              (typeof item === 'object' && Object.values(item)[0]) ??
+              index}
+          </CardBody>
+        </Card>
+      );
+
+      // If the items are orderable or sized by a property, wrap them in a
+      // StyledDraggableCell to apply the drag and drop handlers and/or
```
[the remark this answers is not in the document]
MikeKingdom: Done

src/js/components/Cards/StyledCellContainer.js
```diff
@@ -0,0 +1,16 @@
...
+      ${size?.rows ? `grid-row: span ${size.rows};` : ''}
+    `
+    : '';
+
+// display: grid makes the single child fill the wrapper (stretch in both axes)
+export const StyledCellContainer = styled.div`
+  display: grid;
+  ${(props) => (props.size ? sizeStyles(props.size) : undefined)}
+  ${(props) => (props.draggable ? 'cursor: move;' : undefined)}
+`;
```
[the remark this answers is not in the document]
MikeKingdom: Done

src/js/components/Cards/Cards.js
```diff
@@ -48,6 +63,123 @@ const Cards = React.forwardRef(
...
+        content = (
+          <StyledCellContainer
+            key={`_container_${item?.id || index}`}
+            {...wrapperProps}
+            size={sizeKey ? item[sizeKey] : undefined}
+          >
+            {content}
+          </StyledCellContainer>
+        );
+      }
```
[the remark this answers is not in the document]
MikeKingdom: Done

src/js/components/Cards/stories/onOrder.stories.js
MikeKingdom: Managed to move the focus to the actual child so focus gets rendered according to it's styles (ie. card rounding and focus styles will correctly apply). The keyboard events for moving the items bubbles to the container div where it's actually handled.

MikeKingdom: Footers with the same content in multiple cards is apparently a no-no. Rather annoying limitation so I simply just changed them to text in the stories.

src/js/components/Cards/Cards.js
```diff
@@ -48,6 +63,123 @@ const Cards = React.forwardRef(
...
+                // preventDefault is required to allow the drop to occur
+                event.preventDefault();
+              }
+            },
+            onDrop: () => {
+              if (orderedData) {
+                onOrder(orderedData);
+              }
+            },
+            tabIndex: 0,
```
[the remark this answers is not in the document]
MikeKingdom: Fixed. dragging/moving of items is tracked based on unfiltered/unpaginated data. onOrder will always return the original unfiltered/unpaginated data with the moved item.

src/js/components/Cards/stories/onOrder.stories.js
MikeKingdom: Given that we don't have a way to get a decent label for the children, I feel like the best answer would be for the developer to a decent label on the card when it's moveable?

MikeKingdom: Copilot's suggestion may be enough:

| What | How | Effect |
|---|---|---|
| **Live region** | Hidden `<span aria-live="polite" aria-atomic="true">` updated on each `move()` | Screen reader announces e.g. *"Card moved to position 3 of 8"* |
| **`aria-roledescription="sortable card"`** | Added to each card's `child` props | Screen reader announces role as *"sortable card"* instead of the generic element role |
| **`aria-describedby`** pointing at a hidden hint `<span>` | Unique ID per `Cards` instance via `useId()` | When a card is focused, screen reader reads *"Use arrow keys to reorder cards. Changes are applied immediately."* |

jcfilben: Copilot's suggestion seems reasonable

MikeKingdom: This seems to work pretty well and it still is easy to tell which card the "described_by" is referring to since it also reads whatever describes the card. You do have to be in "focus mode" in the screen reader for the arrow keys to move the cards rather than just cause the screen reader to read the next element. In NVDA you get into focus mode with Insert+Spacebar or CapsLock+Spacebar.

The announcement reads out when you move cards with the keyboard (but not when dragging since dragging with a mouse is a more visual feature.)

src/js/components/Cards/Cards.js
```diff
@@ -48,6 +63,123 @@ const Cards = React.forwardRef(
...
+                event.preventDefault();
+                move(-1);
+              },
+              onRight: (event) => {
+                event.preventDefault();
+                move(1);
+              },
+            },
+          }
+        : {};
```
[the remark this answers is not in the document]
MikeKingdom: Done

src/js/components/Cards/Cards.js
```diff
@@ -1,19 +1,58 @@
...
+};
+
+const reorder = (array, source, target) => {
+  const result = array.slice(0);
+  const tmp = result[source];
+  if (source < target)
+    for (let i = source; i < target; i += 1) result[i] = result[i + 1];
+  else for (let i = source; i > target; i -= 1) result[i] = result[i - 1];
+  result[target] = tmp;
+  return result;
```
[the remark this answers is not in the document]
MikeKingdom: Not needed

src/js/components/Cards/Cards.js
```diff
@@ -48,8 +93,164 @@ const Cards = React.forwardRef(
...
+        announce(
+          format({
+            id: 'cards.moved',
+            values: {
+              source: newIndex + 1,
+              target: currentItems.length,
+            },
+          }),
+        );
+      };
```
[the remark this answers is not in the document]
MikeKingdom: Intentional

jcfilben (commented):
When I'm using voiceover and moving a card, I sometimes get a focus indicator on 2 different cards at the same time
Image

jcfilben (commented):
I'm still seeing the browser default for focus styling when I'm navigating via keyboard
Image

src/js/components/Cards/__tests__/Cards-test.tsx
```diff
@@ -148,4 +148,42 @@ describe('Cards', () => {
...
+            { id: '3', name: 'three' },
+          ]}
+          onOrder={jest.fn()}
+        >
+          {(item) => <Card key={item.id}>{item.name}</Card>}
+        </Cards>
+      </Grommet>,
+    );
+    expect(container.firstChild).toMatchSnapshot();
+  });
```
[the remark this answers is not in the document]
britt6612: @MikeKingdom 

can we test the keyboard onOrder as well

```
  test('onOrder keyboard ArrowDown reorders items', () => {
    const onOrder = jest.fn();
    const data = [
      { id: '1', name: 'one' },
      { id: '2', name: 'two' },
      { id: '3', name: 'three' },
    ];
    const { container } = render(
      <Grommet>
        <Cards data={data} onOrder={onOrder}>
          {(item) => <Card key={item.id}>{item.name}</Card>}
        </Cards>
      </Grommet>,
    );
    const [firstCard] = container.querySelectorAll(
      '[aria-roledescription="sortable card"]',
    );
    fireEvent.keyDown(firstCard, { keyCode: 40 }); // ArrowDown
    expect(onOrder).toHaveBeenCalledTimes(1);
    expect(onOrder).toHaveBeenCalledWith([
      { id: '2', name: 'two' },
      { id: '1', name: 'one' },
      { id: '3', name: 'three' },
    ]);
  });
```

src/js/components/Cards/Cards.js
```diff
@@ -48,8 +93,164 @@ const Cards = React.forwardRef(
...
+
+    const currentItems = orderedData || (!paginate ? data : items);
+
+    const renderItem = (item, index) => {
+      const move = (count) => {
+        const newIndex = Math.max(
+          0,
+          Math.min(index + count, currentItems.length - 1),
+        );
+        onOrder(
```
britt6612, line 109: should we add a 
```
        if (newIndex === index) return;
```

when a user is on the first card and they press arrowUp then it gets repeated Card moved to position 1 of 8. which it would of already have been there so wondering if that would be confusing

src/js/components/Cards/Cards.js
```diff
@@ -48,8 +93,164 @@ const Cards = React.forwardRef(
...
+              }
+            },
+            onDrop: () => {
+              if (unfilteredOrder) {
+                onOrder(unfilteredOrder);
+              }
+            },
+            child: {
+              tabIndex: 0,
+              'aria-roledescription': 'sortable card',
```
britt6612: I believe you need to have a `role` defined when using the `aria-roledescription`

src/js/components/Cards/StyledCellContainer.js
```diff
@@ -0,0 +1,19 @@
+import styled from 'styled-components';
+import { styledComponentsConfig } from '../../utils';
+
+const sizeStyles = (size) =>
+  size && typeof size === 'object'
+    ? `
+      ${size?.columns ? `grid-column: span ${size.columns};` : ''}
```
britt6612, line 12: this is fine but as is but just redundant you already check `size` above
```suggestion
      ${size.columns ? `grid-column: span ${size.columns};` : ''}
```

src/js/components/Cards/StyledCellContainer.js
```diff
@@ -0,0 +1,19 @@
...
+      ${size?.rows ? `grid-row: span ${size.rows};` : ''}
+    `
+    : '';
+
+// display: grid makes the single child fill the wrapper (stretch in both axes)
+export const StyledCellContainer = styled.div.withConfig(
+  styledComponentsConfig,
```
britt6612:
```
export const StyledCellContainer = styled.div.withConfig({
  ...styledComponentsConfig,
  // 'size' passes isPropValid (valid for input/select), filter it to
  // prevent the object value from reaching the DOM as an attribute.
  shouldForwardProp: (prop) =>
    prop !== 'size' && styledComponentsConfig.shouldForwardProp(prop),
})`
```

do we need to prevent the side being passed to the dom

MikeKingdom: Good point. `size` actually is valid for some elements

src/js/components/Cards/StyledCellContainer.js
MikeKingdom: I think I'll leave this because `typeof null === 'object'` and the first check does get it but it doesn't hurt

src/js/components/Cards/Cards.js
MikeKingdom: Kinda true. Things with implicit role like button, li etc don't need it and we don't really know what the child is, if it already has a role and we don't really want to override it. So should we leave it up to the child to define its role?

src/js/components/Cards/Cards.js
MikeKingdom: I liked hearing the message anyway since it affirmed the situation but happy to avoid it if people prefer

britt6612: ah yeah that is another point it might be just personal preference @jcfilben thoughts here

jcfilben: Ok with leaving as is for now, we can change later if needed

src/js/components/Cards/Cards.js
britt6612: ah yes you are correct lets leave it up to the child to define then

src/js/components/Cards/Cards.js
britt6612: sounds good

src/js/components/Cards/Cards.js
```diff
@@ -48,8 +93,164 @@ const Cards = React.forwardRef(
...
+              'aria-describedby': hintId,
+            },
```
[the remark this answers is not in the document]
MikeKingdom: Done

src/js/components/Cards/StyledCellContainer.js
MikeKingdom: Done

src/js/components/Cards/__tests__/Cards-test.tsx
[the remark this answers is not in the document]
MikeKingdom: Done. Thanks for the code @britt6612  :)

jcfilben (approved):

src/js/components/Cards/StyledCellContainer.js
```diff
@@ -0,0 +1,39 @@
...
+    prop !== 'size' && styledComponentsConfig.shouldForwardProp(prop),
+})`
+  display: grid;
+  ${(props) => (props.size ? sizeStyles(props.size) : undefined)}
+  ${(props) => (props.draggable ? 'cursor: move;' : undefined)}
+  ${(props) =>
+    props.round && roundStyle(props.round, props.responsive, props.theme)}
+  &:focus {
+    ${(props) =>
+      (!props.plain || props.focusIndicator) &&
```
britt6612: wouldnt `plain` and `focusIndicator` always be undefined here? would `StyledCellContainer` receive these props?

MikeKingdom: That's certainly true in the current usage. I can remove it for now to clean it up.

britt6612 (approved):

src/js/components/Cards/StyledCellContainer.js
britt6612: thanks @MikeKingdom

PR #7960, opened 2026-05-28
GLCP-355080: Add Stepper Component

aishwaryasharmaccoew:
Storybook

Storybook


src/js/components/Stepper/index.d.ts
```diff
@@ -0,0 +1,65 @@
...
+  canNavigateTo: (stepId: string) => boolean;
+}
+
+declare const Stepper: React.ForwardRefExoticComponent<
+  StepperProps & React.RefAttributes<HTMLOListElement>
+>;
+
+declare const StepperContext: React.Context<StepperContextValue>;
+declare const useStepper: () => StepperContextValue;
+declare const StepperStep: React.FC<{ stepId: string }>;
```
jcfilben, line 47: It looks like we need to expand the typescript definition. In the StepperStep.js file we allow the following props:

Image

jcfilben (commented):
Don't forget to change the target branch so instead of merging into grommet:master we should merge into grommet:stepper

PR #7961, opened 2026-05-28
Remove overflow visible from DateInput

jcfilben:
#### What does this PR do?
Overflow visible was added in [this PR](https://github.com/grommet/grommet/pull/4382) in an effort to prevent the focus indicator from being cut off.
Having focus visible creates the following issue where the Calendar overflows outside it's container when the drop is constrained. Also focus visisble doesn't resolve the focus indicator issue, it gets cut off reguardless.

Before (with focus visible):
Screenshot 2026-05-28 at 3 16 36 PM

After (without focus visible):
Screenshot 2026-05-28 at 3 17 09 PM

Tested with the DateInput/Format story

no
yes
backwards compatible

### DIFF
--- a/src/js/components/DateInput/DateInput.js
+++ b/src/js/components/DateInput/DateInput.js
@@ -441,7 +441,6 @@ Use the icon prop instead.`,
         input,
         <Drop
           key="drop"
-          overflow="visible"
           id={id ? `${id}__drop` : undefined}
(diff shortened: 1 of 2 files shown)


MikeKingdom (approved):
Much better

britt6612 (approved):

